Issue #197, opened 2022-02-02
Copy over portions of the gravity module


evan-forbes:
The [QGB](https://github.com/celestiaorg/quantum-gravity-bridge) will necessitate several additions to the app. The largest of which is copying over portions of the gravity module. 

Since the the QGB is only a one way bridge and is not transferring assets, it doesn't require the portions of the gravity module that recreate state from the bridged chain. We only need to keep things relating to signing over the validator set (such as [`MsgSetOrchestratorAddress`](https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/proto/gravity/v1/msgs.proto#L56-L73) and [`MsgValsetConfirm`](https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/proto/gravity/v1/msgs.proto#L77-L97)) and relayer queries (such as [`ValsetConfirm`](https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/proto/gravity/v1/query.proto#L27-L29) and [`GetDelegateKeyByOrchestrator`](https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/proto/gravity/v1/query.proto#L81-L83))

tracking issue [#44](https://github.com/celestiaorg/quantum-gravity-bridge/issues/44)

SweeXordious:
@evan-forbes I guess for the QGB, it makes sense to start with this issue and similar ones in `celestia-app` before moving to the rest.
Will start drafting this and see. What do you thin? ?

evan-forbes:
to follow up a sync discussion with @SweeXordious, we decided that this issue is a good starting point, and that the step to completing this issue is to first scaffold out:

- a new `qgb` module, including initializing the app with this module installed
- Copy over the `[ValSetConfirm](https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/proto/gravity/v1/query.proto#L27-L29)` proto message
- create the proto message for Confirming `DataCommitments`
- and lastly we should scaffold out the keepers (but not actually introduce any state changes!)

[...]

PR #215, opened 2022-02-16
adds valset confirm logic and queries

SweeXordious:
Adds logic for `ValSetConfirm` without tests + asks some questions.
Contributes to https://github.com/celestiaorg/celestia-app/issues/197

### DIFF
--- a/proto/qgb/query.proto
+++ b/proto/qgb/query.proto
@@ -4,6 +4,7 @@ package qgb;
 import "google/api/annotations.proto";
 import "cosmos/base/query/v1beta1/pagination.proto";
 import "qgb/msgs.proto";
+import "gogoproto/gogo.proto";
 // this line is used by starport scaffolding # 1
 
 option go_package = "github.com/celestiaorg/celestia-app/x/qgb/types";
@@ -14,11 +15,38 @@ service Query {
   rpc ValsetConfirm(QueryValsetConfirmRequest) returns (QueryValsetConfirmResponse) {
     option (google.api.http).get = "/qgb/valset/confirm";
   }
+  // ValsetConfirmsByNonce
+  rpc ValsetConfirmsByNonce(QueryValsetConfirmsByNonceRequest) returns (QueryValsetConfirmsByNonceResponse) {
+    option (google.api.http).get = "/qgb/confirms/{nonce}";
+  }
   // GetDelegateKeyByOrchestrator Queries delegate key by orchestrator
   rpc GetDelegateKeyByOrchestrator(QueryGetDelegateKeysByOrchestratorAddress) returns (QueryGetDelegateKeyByOrchestratorResponse) {
     option (google.api.http).get = "/qgb/query_delegate_keys_by_orchestrator";
   }
   // this line is used by starport scaffolding # 2
+}
+
+// FIXME: Should we also have support for `Valset`s ? Or only `ValsetConfirm`s ?
+//// QueryCurrentValsetRequest
+//message QueryCurrentValsetRequest {}
+//// QueryCurrentValsetResponse
+//message QueryCurrentValsetResponse {
+//  Valset valset = 1 [(gogoproto.nullable) = false];
+//}
+//
+// QueryValsetRequestRequest
+//message QueryValsetRequestRequest {
+//  uint64 nonce = 1;
+//}
+
+// QueryValsetConfirmsByNonceRequest
+message QueryValsetConfirmsByNonceRequest {
+  uint64 nonce = 1;
+}
+
+// QueryValsetConfirmsByNonceResponse
+message QueryValsetConfirmsByNonceResponse {
+  repeated MsgValsetConfirm confirms = 1 [(gogoproto.nullable) = false];
 }
 
 // QueryValsetConfirmRequest ValSet confirm query request
--- a/x/qgb/client/cli/valsetconfirm.go
+++ b/x/qgb/client/cli/valsetconfirm.go
@@ -1,9 +1,28 @@
 package cli
 
 import (
+	"github.com/celestiaorg/celestia-app/x/qgb/types"
+	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/spf13/cobra"
+	"strconv"
 )
+
+func GetQueryCmd() *cobra.Command {
+	//nolint: exhaustivestruct
+	qgbQueryCmd := &cobra.Command{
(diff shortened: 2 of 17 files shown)


SweeXordious:
@evan-forbes Please take a look when you have time.
PS: I included some questions there next to `FIXME`s. However, the `TODO`s, I will address them in upcoming PRs.

proto/qgb/query.proto
```diff
@@ -14,13 +15,40 @@ service Query {
   rpc ValsetConfirm(QueryValsetConfirmRequest) returns (QueryValsetConfirmResponse) {
     option (google.api.http).get = "/qgb/valset/confirm";
   }
+  // ValsetConfirmsByNonce
+  rpc ValsetConfirmsByNonce(QueryValsetConfirmsByNonceRequest) returns (QueryValsetConfirmsByNonceResponse) {
+    option (google.api.http).get = "/qgb/confirms/{nonce}";
+  }
```
evan-forbes, line 21: adding this makes sense I think :+1:

proto/qgb/query.proto
```diff
@@ -14,13 +15,40 @@ service Query {
...
   // GetDelegateKeyByOrchestrator Queries delegate key by orchestrator
   rpc GetDelegateKeyByOrchestrator(QueryGetDelegateKeysByOrchestratorAddress) returns (QueryGetDelegateKeyByOrchestratorResponse) {
     option (google.api.http).get = "/qgb/query_delegate_keys_by_orchestrator";
   }
   // this line is used by starport scaffolding # 2
 }
 
+// FIXME: Should we also have support for `Valset`s ? Or only `ValsetConfirm`s ?
```
evan-forbes, line 29: We might need to add valsets when we implement slashing, but for now we use should only need the confirms.

The only reason that we're having validators submit transactions for the MVP is because it makes aggregation trivial for the relayer

x/qgb/client/cli/valsetconfirm.go
```diff
@@ -1,19 +1,55 @@
...
+				Nonce:   nonce,
+				Address: args[1],
+			}
+
+			res, err := queryClient.ValsetConfirm(cmd.Context(), req)
+			if err != nil {
+				return err
+			}
+
+			return clientCtx.PrintProto(res)
```
evan-forbes, line 52: we definitely do not have to do this in this PR or the next few, but eventually, we should include integration tests similar to what we do for the payment module. We might also want to write an issue to remind future us.

x/qgb/client/rest/tx.go
```diff
@@ -0,0 +1,3 @@
+package rest
+
+// TODO add ValSetConfirm REST methods
```
evan-forbes, line 3: grpc is already up and running, so I'm not sure adding a rest endpoint really gives us anything, but wdyt?

x/qgb/keeper/keeper.go
```diff
@@ -13,17 +12,59 @@ import (
 type Keeper struct {
 	cdc      codec.BinaryCodec
 	storeKey sdk.StoreKey
-	memKey   sdk.StoreKey
+
+	// FIXME: should these modules be added now ? Or we can keep them until later along with slashing? Probably, we would need some of them when adding tests
```
evan-forbes, line 16: later with slashing I think

x/qgb/keeper/keeper.go
```diff
@@ -13,17 +12,59 @@ import (
...
+
+// prefixRange turns a prefix into a (start, end) range. The start is the given prefix value and
+// the end is calculated by adding 1 bit to the start value. Nil is not allowed as prefix.
+// 		Example: []byte{1, 3, 4} becomes []byte{1, 3, 5}
+// 				 []byte{15, 42, 255, 255} becomes []byte{15, 43, 0, 0}
+//
+// In case of an overflow the end is set to nil.
+//		Example: []byte{255, 255, 255, 255} becomes nil
+// MARK finish-batches: this is where some crazy shit happens
+// FIXME: should we keep these funny comments ?
```
evan-forbes, line 43: we can obv keep the informative parts, but yeah I don't think we need the silly parts

x/qgb/keeper/query_valset_confirm_test.go
```diff
@@ -0,0 +1,132 @@
+package keeper
+
+// TODO uncomment after adding logic for other messages
```
evan-forbes, line 3: :+1:

x/qgb/types/keys.go
```diff
@@ -16,3 +22,29 @@ const (
...
+	// ValsetConfirmKey indexes valset confirmations by nonce and the validator account address
+	// FIXME: For our keys, should they be `gravity1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm` or `qgb1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm`?
+	// i.e gravity1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm
+	ValsetConfirmKey = "ValsetConfirmKey"
+)
+
+// GetValsetConfirmKey returns the following key format
+// prefix   nonce                    validator-address
+// [0x0][0 0 0 0 0 0 0 1][gravity1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm]
+// FIXME: For our keys, should they be `gravity1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm` or `qgb1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm`?
```
evan-forbes, line 36: great question, I'll have to dig into this more. We might want to write an issue for it as we might want to record our findings. 

could it also be `qgb1ahx7f8wyertuus9r20284ej0asrs085ceqtfnm`?

evan-forbes (approved):
great work, I left two comments that I think are important enough that we should write quick issues that point to them and include a small amount of context so that we don't forget to implement/research them later. 

pre-approving tho :+1: :shipit:

x/qgb/client/cli/valsetconfirm.go
SweeXordious: So no need to have a client module? 
Cool, I can remove this stuff and only add it later when needed.

x/qgb/client/rest/tx.go
SweeXordious: True that. I'll leave it for later if needed

x/qgb/types/keys.go
SweeXordious: I still don't have much information about those values. I'll keep the FIXME and come back to them later if I don't stumble upon them when addressing the other parts.

PR #216, opened 2022-02-18
Message inclusion check

### LINKED ISSUE #205: Check for message inclusion before committing to a block
Since we will be using ABCI++ soon, it should be much easier to implement a message inclusion check that can be performed when a validator processes a proposal block. This issue has been described in more depth previously in celestiaorg/celestia-core#520 and celestiaorg/celestia-core#519. We also have a pending ADR celestiaorg/celestia-core#559

This check should ensure that 
- each valid PayForMessage has a corresponding message with the correct commitment
- that there are no unaccounted for messages included in the square

blocked by celestiaorg/celestia-app#381 (might be able to be done in the same PR since they are so related)

blocked by celestiaorg/celestia-core#204

evan-forbes:
## Description

This PR creates a consensus time message inclusion check using the new `ProcessProposal` ABCI++ method.

It's worth noting that this check does not actually check that the commitment to the data is an actual subtree root to the data commitment, because 1) we aren't actually filling the square according to the non-interactive default rules yet and 2) the specs specifically describes the non-interactive rules as not consensus critical, unlike message inclusion. However, we can eventually make those rules consensus critical using the same ABCI method if that's something we want to do.

It's also worth noting that this check does not ensure that the `MsgPayForMessage` that it is checking the messages against is actually valid. This means that if a validator wanted to, they could produce a block that has an invalid `MsgPayForMessage`, which therefore wouldn't pay for the message, but the message does get included in the block. In order to check if the `MsgPayForMessage` is valid, I think we might have wait for immediate execution.

This PR also only uses the normal integration test, meaning that the happy case is tested for implicitly if the test passes, but the cases where the block should be rejected do not have integration/liveness tests. We can either write an issue to do this in a different PR or we can add these integration tests to this PR.

closes #205


x/payment/client/testutil/integration_test.go
```diff
@@ -131,7 +132,7 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForMessage() {
 
 				events := txResp.Logs[0].GetEvents()
 				for i := 0; i < len(events); i++ {
-					s.Equal("/payment.MsgPayForMessage", events[i].GetAttributes()[0].Value)
+					s.Equal(types.URLMsgPayforMessage, events[i].GetAttributes()[0].Value)
```
evan-forbes, line 135: this change is unrelated to this PR, but we should definitely be using a constant here.

adlerjohn:
Is there are outstanding TODOs can we convert this to draft?

app/process_proposal.go
```diff
@@ -0,0 +1,79 @@
...
+
+			pfm, ok := msg.(*types.MsgPayForMessage)
+			if !ok {
+				continue
+			}
+
+			commitments[string(pfm.MessageShareCommitment)] = struct{}{}
+		}
+
+	}
```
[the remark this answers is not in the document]
adlerjohn: I don't see any reason to forbid including two of the same message in the same block, so long as they're paid for. _That being said_, a block proposer could be tricky and make both PayForMessage txs in the block point to the same message! An interesting corner case. Can you file an issue...somewhere...to re-visit this? And document your decision (forbid or allow, doesn't really matter, impl detail) in the code with a link to the issue.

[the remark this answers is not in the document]
evan-forbes: #226 

will update this PR tmrw to make the default behavior to exclude the identical message, but keep the payments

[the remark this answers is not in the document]
evan-forbes: update to this, I just ended up taking the easier option, in which we count the commitments using a counter instead of relying on a map, this way we also count identical commitments and I think things work intuitively. Intuitively being that identical messages get included twice and each PayForData must have a corresponding message. 

we should still make a formal decision with #226

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			continue
+		}
+
+		for _, msg := range tx.GetMsgs() {
+			if sdk.MsgTypeURL(msg) != types.URLMsgPayForData {
+				continue
+			}
+
+			pfm, ok := msg.(*types.MsgPayForData)
```
liamsi, line 35:
```suggestion
			pfd, ok := msg.(*types.MsgPayForData)
```
?

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+			if !ok {
+				continue
+			}
+
+			commitments[string(pfm.MessageShareCommitment)] = struct{}{}
```
liamsi, line 40:
```suggestion
			commitments[string(pfd.MessageShareCommitment)] = struct{}{}
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+			commitmentCounter++
+		}
+
+	}
+
+	// quickly compare the number of PFMs and messages, if they aren't
```
liamsi, line 46:
```suggestion
	// quickly compare the number of PFDs and messages, if they aren't
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	rejectedPropBlockLog = "Rejected proposal block:"
+)
+
+func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponseProcessProposal {
+	// Check for message inclusion:
+	//  - each MsgPayForData included in a block should have a corresponding message also in the block data
+	//  - the commitment in each PFM should match that of its corresponding message
```
adlerjohn, line 16:
```suggestion
	//  - the commitment in each PFD should match that of its corresponding data
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+const (
+	rejectedPropBlockLog = "Rejected proposal block:"
+)
+
+func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponseProcessProposal {
+	// Check for message inclusion:
+	//  - each MsgPayForData included in a block should have a corresponding message also in the block data
```
adlerjohn, line 15:
```suggestion
	//  - each MsgPayForData included in a block should have a corresponding data also in the block body
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+	//  - the commitment in each PFM should match that of its corresponding message
+	//  - there should be no unpaid for messages
```
adlerjohn, line 17:
```suggestion
	//  - there should be no unpaid-for data
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// extract the commitments from any MsgPayForDatas in the block
+	commitments := make(map[string]struct{})
+	// we have a separate counter so that identical messages also get counted
```
adlerjohn, line 21:
```suggestion
	// we have a separate counter so that identical data also get counted
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+			rejectedPropBlockLog,
+			"reason",
+			"varying number of messages and payForData txs in the same block",
+		)
+		return abci.ResponseProcessProposal{
+			Result: abci.ResponseProcessProposal_REJECT,
+		}
+	}
+
+	// iterate through all of the messages and ensure that a PFM with the exact
```
adlerjohn, line 59: There are a bunch of these left. Search+replace.
```suggestion
	// iterate through all of the messages and ensure that a PFD with the exact
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+			pfd, ok := msg.(*types.MsgPayForData)
+			if !ok {
+				continue
+			}
+
+			commitments[string(pfd.MessageShareCommitment)] = struct{}{}
+			commitmentCounter++
+		}
+
+	}
```
adlerjohn, line 44:
```suggestion
		}
	}
```

app/process_proposal.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		if _, has := commitments[string(commit)]; !has {
+			app.Logger().Info(rejectedPropBlockLog, "reason", "missing MsgPayForData for included message")
+			return abci.ResponseProcessProposal{
+				Result: abci.ResponseProcessProposal_REJECT,
+			}
+		}
+
+	}
```
adlerjohn, line 83:
```suggestion
		}
	}
```

adlerjohn (commented):
nits

app/process_proposal.go
```diff
@@ -0,0 +1,86 @@
...
+
+		for _, msg := range tx.GetMsgs() {
+			if sdk.MsgTypeURL(msg) != types.URLMsgPayForData {
+				continue
+			}
+
+			pfd, ok := msg.(*types.MsgPayForData)
+			if !ok {
+				continue
+			}
```
adlerjohn, line 38: Shouldn't this not silently ignore the error? Is there ever a case where `ok` would be `false` here?

app/process_proposal.go
```diff
@@ -0,0 +1,86 @@
...
+				Result: abci.ResponseProcessProposal_REJECT,
+			}
+		}
+
+		if _, has := commitments[string(commit)]; !has {
+			app.Logger().Info(rejectedPropBlockLog, "reason", "missing MsgPayForData for included message")
+			return abci.ResponseProcessProposal{
+				Result: abci.ResponseProcessProposal_REJECT,
+			}
+		}
```
adlerjohn, line 80: This doesn't check the field in the wrapped tx of the starting location of the data.

adlerjohn (approved):
utACK

adlerjohn (changes requested):
Approved by accident 😂

app/process_proposal.go
evan-forbes: The only way that it would be false is if a malicious block proposer manufactured a transaction with a message that has a URL that matches the proto type, but the type is actually different. In that case, we have to handle that bool. We could 
- reject the block
- panic (default without checking the ok)
- ignore this tx and continue (perhaps log this?)

The cosmos-sdk only uses the URL when calling deliverTx or CheckTx, but they also catch panics.

imho ignoring the tx is fine, but I'm curious what you think, should we reject? Log it?

app/process_proposal.go
evan-forbes: wait, we don't have wrapped txs implemented yet?

adlerjohn: Yeah that's what I thought. Unless you have better memory than me :joy:, can you add a TODO here that once wrapped txs are implemented that field needs to be checked here? Or better yet, add it to the issue description.

app/process_proposal.go
adlerjohn: At the very least log it IMO.

app/process_proposal.go
evan-forbes: added a todo in https://github.com/celestiaorg/celestia-app/pull/216/commits/f6661b1c82739e4e78ff612c81c9d3b2e33d1a8e
https://github.com/celestiaorg/celestia-app/issues/383 to track this specific thing, but also had a tracking issue for the things that we need before we can do that in #382

app/process_proposal.go
evan-forbes: this is me logging that we are now logging https://github.com/celestiaorg/celestia-app/pull/216/commits/2713641910e2b32a9972c5e5601a119656d9ef99

:slightly_smiling_face:

adlerjohn (approved):

Issue #199, opened 2022-02-02
Add new transaction for validators to submit `DataCommitment`s


evan-forbes:
To accomodate the QGB, validators need a way to submit signatures for a data commitments so that relayers can easily find them and submit them to the bridged chain. To do this, we will introduce `MsgSubmitDataCommitment`

This new message type needs to convey 
- the signature
- that the validator's associated bridge address
- the commitment
- the range of blocks that that commitment was generated over

The business logic for this message needs to check that 
- the range of blocks is correct 
- the signature is valid
- that the provided validator's bridge address is the one associated with that validator

Lastly, relayers will need to be able to query these commitments in a similar fashion to how they query [`ValSetConfirms`](https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/proto/gravity/v1/query.proto#L18-L32)

tracking issue [#44](https://github.com/celestiaorg/quantum-gravity-bridge/issues/44)

PR #222, opened 2022-02-22
Add data commitment logic

SweeXordious:
Adds data commitment logic.
Closes: https://github.com/celestiaorg/celestia-app/issues/199


SweeXordious:
@evan-forbes Please take a look if this is in the right direction. And also couple questions when you have time:
- Are the queries' http endpoints, in this PR, well-defined ? Or that's not the right structure ? 
- What do you mean by: `that the provided validator's bridge address is the one associated with that validator` in this issue https://github.com/celestiaorg/celestia-app/issues/199 ?

evan-forbes:
> What do you mean by: that the provided validator's bridge address is the one associated with that validator in this issue https://github.com/celestiaorg/celestia-app/issues/199 ?

This just means that we have to check the current state to make sure that the message submitted is actually valid, and that the orchestrator is who they say they are. 

This is typically done in the gravity module via this func

https://github.com/Gravity-Bridge/Gravity-Bridge/blob/7a786d2ba999b45292c976282cf2aa88ba4cbba9/module/x/gravity/keeper/msg_server.go#L313-L353

SweeXordious:
@evan-forbes This means we need to add an extra field to `DataCommitmentConfirm` containing the `eth_address` ? 
Or, we need to have a KV store for these, no ?

x/qgb/keeper/msg_server.go
```diff
@@ -45,8 +47,56 @@ func (k msgServer) ValsetConfirm(c context.Context, msg *types.MsgValsetConfirm)
...
+
+	// TODO get eth address from store
+	//ethAddressFromStore, found := k.GetEthAddressByValidator(ctx, validator.GetOperator())
+	//if !found {
+	//	return nil, sdkerrors.Wrap(types.ErrEmpty, "no eth address set for validator")
+	//}
+	//
+	//if *ethAddressFromStore != *submittedEthAddress {
+	//	return nil, sdkerrors.Wrap(types.ErrInvalid, "submitted eth address does not match delegate eth address")
+	//}
```
evan-forbes, line 83: imho, I think we should actualy get rid of this, as our exact conditions, mechanisms, and fraud proofs for slashing are all tbd. While we will hopefully be able to reuse some code from the gravity module, we will also likely end up writing a our own custom code, and can rely on us making the necessary changes then.

x/qgb/keeper/msg_server.go
```diff
@@ -45,8 +47,56 @@ func (k msgServer) ValsetConfirm(c context.Context, msg *types.MsgValsetConfirm)
...
+
+	err = types.ValidateEthereumSignature([]byte(msg.Commitment), sigBytes, *ethAddress)
+	if err != nil {
+		return nil, sdkerrors.Wrap(types.ErrInvalid, fmt.Sprintf("signature verification failed expected sig by %s with checkpoint %s found %s", ethAddress, msg.Commitment, msg.Signature))
+	}
+
+	// TODO set data commitment confirm when we start persisting state
```
evan-forbes, line 90: I don't think it matters exactly when, but out of curiosity, will this be included in this PR or another?

evan-forbes (commented):
nice!

x/qgb/keeper/msg_server.go
SweeXordious: In this PR. This PR is still WIP. I just push to have a better vision on what is happening and how it looks when reviewing :D

x/qgb/keeper/msg_server.go
SweeXordious: cool, will remove them

x/qgb/config.go
```diff
@@ -0,0 +1,28 @@
...
+	bech32PrefixConsAddr = "qgbvalcons"
+	// bech32PrefixConsPub defines the bech32 prefix of a consensus node public key
+	bech32PrefixConsPub = "qgbvalconspub"
+)
+
+func init() {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount(bech32PrefixAccAddr, bech32PrefixAccPub)
+	config.SetBech32PrefixForValidator(bech32PrefixValAddr, bech32PrefixValPub)
+	config.SetBech32PrefixForConsensusNode(bech32PrefixConsAddr, bech32PrefixConsPub)
```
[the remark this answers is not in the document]
evan-forbes: while it might not be done using this exact mechanism, this is already done using `celes`, so we should avoid doing this again using `qgb`. Surprised this doesn't break anything lol

x/qgb/keeper/keeper.go
```diff
@@ -10,14 +12,16 @@ import (
 )
 
 type Keeper struct {
-	cdc      codec.BinaryCodec
-	storeKey sdk.StoreKey
+	cdc           codec.BinaryCodec
+	storeKey      sdk.StoreKey
+	StakingKeeper *stakingkeeper.Keeper
```
evan-forbes, line 17: its good practice in the sdk to define local interface that only have access to the methods that they need, for instance, in the payment module we have a [`BankKeeper`](https://github.com/celestiaorg/celestia-app/blob/6e0140b9c9434bf8b6a30855bfa884905ae80b9a/x/payment/keeper/keeper.go#L40-L44) to burn funds but it can't send tokens to arbitrary accounts

x/qgb/keeper/msg_server.go
```diff
@@ -45,8 +47,53 @@ func (k msgServer) ValsetConfirm(c context.Context, msg *types.MsgValsetConfirm)
...
+	if err != nil {
+		return nil, sdkerrors.Wrap(types.ErrInvalid, fmt.Sprintf("signature verification failed expected sig by %s with checkpoint %s found %s", ethAddress, msg.Commitment, msg.Signature))
+	}
+	ethAddressFromStore, found := k.GetEthAddressByValidator(ctx, validator.GetOperator()) // check what to do with operator
+	if !found {
+		return nil, sdkerrors.Wrap(types.ErrEmpty, "no eth address set for validator")
+	}
+	if *ethAddressFromStore != *ethAddress {
+		return nil, sdkerrors.Wrap(types.ErrInvalid, "submitted eth address does not match delegate eth address")
+	}
```
evan-forbes, line 85: nice, these are important to check! but I think they would better fit in the `ValidateBasic` method for `MsgDataCommitmentConfirm`. In general, if we can perform a quick stateless check on a message, it should be in `ValidateBasic`

x/qgb/keeper/query_data_commitment.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (k Keeper) DataCommitmentConfirmsByValidator(ctx context.Context, request *types.QueryDataCommitmentConfirmsByValidatorRequest) (*types.QueryDataCommitmentConfirmsByValidatorResponse, error) {
+	addr, err := sdk.AccAddressFromBech32(request.Address)
+	if err != nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "address invalid")
+	}
+	return &types.QueryDataCommitmentConfirmsByValidatorResponse{Confirms: k.GetDataCommitmentConfirmsByValidator(sdk.UnwrapSDKContext(ctx), addr)}, nil
+}
+
+func (k Keeper) DataCommitmentConfirmsByCommitment(ctx context.Context, request *types.QueryDataCommitmentConfirmsByCommitmentRequest) (*types.QueryDataCommitmentConfirmsByCommitmentResponse, error) {
```
evan-forbes, line 26: do we not have CI linters for line length? This is a bit of nit pick, but pretty sure these lines are too long. I forget what the limit is tbh

x/qgb/keeper/keeper_data_commitment.go
```diff
@@ -2,22 +2,110 @@ package keeper
...
+
+// GetDataCommitmentConfirmsByValidator Returns data commitment confirms by validator address
+func (k Keeper) GetDataCommitmentConfirmsByValidator(ctx sdk.Context, validator sdk.AccAddress) (confirms []types.MsgDataCommitmentConfirm) {
+	if err := sdk.VerifyAddressFormat(validator); err != nil {
+		ctx.Logger().Error("invalid validator address")
+		return nil
+	}
+
+	prefixStore := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.DataCommitmentConfirmKey))
+	start, end := prefixRange([]byte(types.DataCommitmentConfirmKey)) // FIXME can we make this faster ?
```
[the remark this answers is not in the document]
evan-forbes: I think so, but am still ivestigating/thinking about this

x/qgb/keeper/query_data_commitment.go
```diff
@@ -0,0 +1,32 @@
...
+	return &types.QueryDataCommitmentConfirmsByCommitmentResponse{Confirms: k.GetDataCommitmentConfirmsByCommitment(sdk.UnwrapSDKContext(ctx), request.Commitment)}, nil
+}
+
+func (k Keeper) DataCommitmentConfirmsByRange(ctx context.Context, request *types.QueryDataCommitmentConfirmsByRangeRequest) (*types.QueryDataCommitmentConfirmsByRangeResponse, error) {
+	return &types.QueryDataCommitmentConfirmsByRangeResponse{Confirms: k.GetDataCommitmentConfirmsByRange(sdk.UnwrapSDKContext(ctx), request.BeginBlock, request.EndBlock)}, nil
```
evan-forbes, line 31: again, a very small nit pick, but I think we lose a bit of readability by putting everything on one line

x/qgb/keeper/keeper_data_commitment.go
```diff
@@ -2,22 +2,110 @@ package keeper
...
+
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		confirm := types.MsgDataCommitmentConfirm{}
+		k.cdc.MustUnmarshal(iterator.Value(), &confirm)
+		if confirm.BeginBlock <= beginBlock && confirm.EndBlock >= endBlock {
+			confirms = append(confirms, confirm)
+		}
+	}
```
evan-forbes, line 80: note to self: dig into this more

evan-forbes (commented):
good stuff :+1:

x/qgb/config.go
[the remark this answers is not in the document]
SweeXordious: This is defined inside the module... maybe it is picked up per module?

x/qgb/keeper/keeper.go
SweeXordious: aaaaaah, beautiful stuff. Will change

x/qgb/keeper/msg_server.go
SweeXordious: We're doing the stateless ones in `ValidateBasic`. However, these require access to the state:
- `validator,` found := k.GetOrchestratorValidator(ctx, validatorAddress)`
- `ethAddressFromStore, found := k.GetEthAddressByValidator(ctx, validator.GetOperator())`

x/qgb/keeper/query_data_commitment.go
SweeXordious: Locally, my linter config complains. However, since it is passing GitHub one, I thought it's alright.

x/qgb/keeper/query_data_commitment.go
SweeXordious: will make the change then

x/qgb/keeper/msg_server.go
evan-forbes: makes sense

SweeXordious:
@evan-forbes If this PR is good. We can merge :D

x/qgb/config.go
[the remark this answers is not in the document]
evan-forbes: It makes sense for them to have it in their module, as they want to have grav addresses, but not ours as we will have celes. sdk.GetConfig is global (hence the "sealing"), and having different address types per module is not an sdk feature I'm aware of. The only addresses that we should have to manage are done so via the orchestrator and eth addresses, which we are keeping in the state. 

Do things break when we remove them?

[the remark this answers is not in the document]
SweeXordious: not at all. Ill remove them then.

evan-forbes (approved):
:+1:

PR #224, opened 2022-02-25
refactor prepare proposal to efficiently fill the square

evan-forbes:
## Description

This PR 
- moves the data root generation step to `PrepareProposal`
- estimates a square size in `PrepareProposal` based on the block data (errors on the side of over estimating)
- efficiently fills the square in an atomic and lazy fashion (only adds a pfd if its message can also fit, uses new lazy share writer from core)


app/prepare_proposal.go
```diff
@@ -1,30 +1,92 @@
...
+		app.Logger().Error(
+			"failure to erasure the data square while creating a proposal block",
+			"error",
+			err.Error(),
+		)
+		panic(err)
+	}
+
+	dah := da.NewDataAvailabilityHeader(eds)
+	data.Hash = dah.Hash()
```
liamsi, line 38: Nice. So this is not done in core at all anymore, right? I'm wondering if we should move some of the code that is currently in core to this repo actually 🤔

app/process_proposal.go
```diff
@@ -82,6 +86,49 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
+	if err != nil {
+		app.Logger().Error(rejectedPropBlockLog, "reason", "failure to unmarshal block data:", err)
+		return abci.ResponseProcessProposal{
+			Result: abci.ResponseProcessProposal_REJECT,
+		}
+	}
+
+	shares, _, err := data.ComputeShares(req.BlockData.OriginalSquareSize)
+	if err != nil {
+		app.Logger().Error(rejectedPropBlockLog, "reason", "failure to compute shares from block data:", err)
```
liamsi, line 99: how does this look like when printed? are the extra `:` needed?

x/payment/client/testutil/integration_test.go
```diff
@@ -83,19 +83,7 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForData() {
...
-			"valid transaction list of square sizes",
-			[]string{
-				hexNS,
-				hexMsg,
-				fmt.Sprintf("--from=%s", username),
-				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
-				fmt.Sprintf("--%s=%s", flags.FlagFees, sdk.NewCoins(sdk.NewCoin(s.cfg.BondDenom, sdk.NewInt(2))).String()),
-				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
-				fmt.Sprintf("--%s=%s", paycli.FlagSquareSizes, "256,128,64"),
+				fmt.Sprintf("--%s=%s", paycli.FlagSquareSizes, "2,4,8,16"),
```
liamsi, line 86: Q: shouldn't we at least test for 32 and 64 squares here as well?

app/split_shares.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			continue
+		}
+
+		authTx, ok := tx.(signing.Tx)
+		if !ok {
+			continue
+		}
+
+		// write the tx to the square if it normal
```
liamsi, line 43: if it normal?

app/split_shares.go
```diff
@@ -0,0 +1,298 @@
...
+// that message and their corresponding txs get written to the square
+// atomically.
+type shareSplitter struct {
+	txWriter  *coretypes.ContiguousShareWriter
+	msgWriter *coretypes.MessageShareWriter
+
+	// we aren't doing anything with evidence or isrs yet, so we don't need to
+	// lazily generate those shares lazily
+	evdShares [][]byte
+	isrShares [][]byte
```
liamsi, line 113: Didn't we remove ISRs everywhere already?

app/split_shares.go
```diff
@@ -0,0 +1,298 @@
...
+
+// shareSplitter write a data square using provided block data. It also ensures
+// that message and their corresponding txs get written to the square
+// atomically.
+type shareSplitter struct {
+	txWriter  *coretypes.ContiguousShareWriter
+	msgWriter *coretypes.MessageShareWriter
+
+	// we aren't doing anything with evidence or isrs yet, so we don't need to
+	// lazily generate those shares lazily
```
liamsi, line 111: _lazily_ generate those shares _lazily_?

app/split_shares.go
```diff
@@ -0,0 +1,298 @@
...
+	evdShares [][]byte
```
liamsi, line 112: Also, what do you mean by we are not doing anything with evidence?

app/split_shares.go
```diff
@@ -0,0 +1,298 @@
...
+		return false, nil, nil, err
+	}
+
+	wrappedTx, err := coretypes.WrapMalleatedTx(parentHash[:], rawProcessedTx)
+	if err != nil {
+		return false, nil, nil, err
+	}
+
+	// check if we have room for both the tx and message it is crucial that we
+	// add both atomically, otherwise the block is invalid
```
liamsi, line 189:
```suggestion
	// Check if we have room for both the tx and message. It is crucial that we
	// add both atomically, otherwise the block would be invalid.
```

app/process_proposal.go
evan-forbes: this exact error is difficult to reproduce, but they look like this

```
E[2022-05-11|20:46:44.403] Rejected proposal block:                     reason="missing MsgPayForMessage for included message" error="tacos are good on other days than tuesday"
```
I was however missing an error key, which has been added

x/payment/client/testutil/integration_test.go
evan-forbes: Now that we are efficiently filling the square and actually using a square size other than 64, messages have to be signed over accordingly. Since there is only a single transaction in this block, the square will be small.

You're right though, we should test with a larger message here. I tried and ran into #379 when adding the flag to increase the gas limit.

#400

app/split_shares.go
evan-forbes: :+1: good catch https://github.com/celestiaorg/celestia-app/pull/224/commits/d7be2b39692b00b575a4923c1071d9b5ebb3a9c6

app/split_shares.go
evan-forbes: whoops https://github.com/celestiaorg/celestia-app/pull/224/commits/ac75bde22a46a6a408f69bfe15e13a05f08ac24e

app/split_shares.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/224/commits/9dc8d123484dce907781f97a855657b239c26327

app/split_shares.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/224/commits/9dc8d123484dce907781f97a855657b239c26327

app/prepare_proposal.go
```diff
@@ -1,30 +1,92 @@
...
+	}
+
+	evdBytes := 0
+	for _, evd := range data.Evidence.Evidence {
+		evdBytes += evd.Size() + delimLen(uint64(evd.Size()))
+	}
+	evdShareEstimate := evdBytes / consts.TxShareSize
+	if evdBytes > 0 {
+		evdShareEstimate++ // add one to round up
+	}
```
adlerjohn, line 66: Do we still have evidence?

app/prepare_proposal.go
```diff
@@ -1,30 +1,92 @@
...
+
+	if estimatedSize > consts.MaxSquareSize {
+		return consts.MaxSquareSize
+	}
+
+	if estimatedSize < consts.MinSquareSize {
+		estimatedSize = consts.MinSquareSize
+	}
+
+	return estimatedSize
```
adlerjohn, line 82: This control flow is a bit weird because it's uneven: the first `if` returns and the second doesn't. If this were Rust you could use a `match`, but given the constraints of the language I would prefer an `if` that sets `estimatedSize` followed by an `else if` following by a `return`.

app/prepare_proposal.go
evan-forbes: I looked into this, but atm I think its best to keep it there. This is mainly because it makes it easier to test things. This could be a good goal for our next upgrade tho.

app/prepare_proposal.go
evan-forbes: we can use a switch statement, wdyt?

https://github.com/celestiaorg/celestia-app/pull/224/commits/a17beb0e99f9815842a260df74cbcc2261525d33

app/split_shares.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func delimLen(x uint64) int {
+	i := 0
+	for x >= 0x80 {
+		x >>= 7
+		i++
+	}
+	return i + 1
+}
```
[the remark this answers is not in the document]
evan-forbes: Nice!

originally, this was this was a modified version of [this function](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/encoding/binary/varint.go;l=41-50) from the standard library, which is used in core. 

https://github.com/celestiaorg/celestia-app/pull/224/commits/c77ea6e9a10309e198e1158d5cc7bf367abc6d0b

app/prepare_proposal.go
evan-forbes: wdym?

like, are we adding evidence to the square?
https://github.com/celestiaorg/celestia-app/blob/efdf2d6b320c4ce21a4c715e845d09bbbec1033d/app/split_shares.go#L133

adlerjohn (dismissed):

app/split_shares.go
liamsi: I see https://github.com/celestiaorg/celestia-app/commit/d7be2b39692b00b575a4923c1071d9b5ebb3a9c6 but I also see:
```suggestion
		// skip txs that don't contain messages
```

liamsi: What happened here? Did merging revert this or sth?

evan-forbes: I think that comment might have been there twice, or I fixed the wrong one?

liamsi (approved):

adlerjohn (approved):

PR #232, opened 2022-03-07
Record validator set during `EndBlock` for QGB orchestrator

evan-forbes:
## Description

Adds changes needed to record the validator set at the end of the block. We technically don't need this for MVP, as the orchestrator could keep track of changes to the validator set and submit  `ValsetConfirm`s on its own. However, we will need this for slashing in the future, and its makes writing the orchestrator significantly easier, so adding it now.


x/qgb/keeper/query_data_commitment.go
```diff
@@ -45,10 +46,10 @@ func (k Keeper) DataCommitmentConfirmsByCommitment(
...
 ) (*types.QueryDataCommitmentConfirmsByCommitmentResponse, error) {
 	return &types.QueryDataCommitmentConfirmsByCommitmentResponse{
-		Confirms: k.GetDataCommitmentConfirmsByCommitment(
-			sdk.UnwrapSDKContext(ctx),
-			request.Commitment,
-		),
+		// Confirms: k.GetDataCommitmentConfirmsByCommitment(
+		// 	sdk.UnwrapSDKContext(ctx),
+		// 	request.Commitment,
+		// ),
```
evan-forbes, line 52: commented this out for now until we figure out an ideal way of querying.

proto/qgb/genesis.proto
```diff
@@ -1,7 +1,26 @@
 syntax = "proto3";
 package qgb;
 
+import "gogoproto/gogo.proto";
```
adlerjohn, line 4: Nit, but can you format the alignment for this file, and add a newline at the end?

adlerjohn (commented):
Some nits

SweeXordious (approved):
Good job :+1: LGTM

SweeXordious:
@evan-forbes Can we merge this so that I include this stuff for the tests also ? Thanks

PR #234, opened 2022-03-08
orchestrator mvp

evan-forbes:
## Description

**work in progress** mvp qgb orchestrator and relayer

still needs some debugging, and we need commands to start each. We should also split this PR up for easier review


x/qgb/abci.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/celestiaorg/celestia-app/x/qgb/keeper"
+	"github.com/celestiaorg/celestia-app/x/qgb/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// EndBlocker is called at the end of every block
+func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
```
SweeXordious, line 11: So with every block, we will be updating the validator set if it is different ? Doesn't this happen only after certain period ? not on every block ?

app/app.go
```diff
@@ -284,10 +284,21 @@ func New(
...
+	qgbmodule := qgbmodule.NewAppModule(appCodec, app.QgbKeeper)
+
 	// register the staking hooks
 	// NOTE: stakingKeeper above is passed by reference, so that it will contain these hooks
 	app.StakingKeeper = *stakingKeeper.SetHooks(
-		stakingtypes.NewMultiStakingHooks(app.DistrKeeper.Hooks(), app.SlashingKeeper.Hooks()),
+		stakingtypes.NewMultiStakingHooks(app.DistrKeeper.Hooks(),
+			app.SlashingKeeper.Hooks(),
+			app.QgbKeeper.Hooks(),
+		),
```
SweeXordious, line 301:
```suggestion
		stakingtypes.NewMultiStakingHooks(
			app.DistrKeeper.Hooks(),
			app.SlashingKeeper.Hooks(),
			app.QgbKeeper.Hooks(),
		),
```

SweeXordious (commented):
Awesome stuff

x/qgb/abci.go
evan-forbes: it's possible that the validator set and how much voting power each has could change rapidly, so we have to check every block to ensure that we don't end up in a situation where the contracts can no longer be updated because the validator set changed too rapidly

proto/qgb/msgs.proto
```diff
@@ -8,16 +8,20 @@ option go_package = "github.com/celestiaorg/celestia-app/x/qgb/types";
...
+  // validator set.
   rpc ValsetConfirm(MsgValsetConfirm) returns (MsgValsetConfirmResponse) {
     option (google.api.http).post = "/qgb/valset_confirm";
   }
-  // DataCommitmentConfirm allows the validators to submit a confirmation for a data commitment.
-  rpc DataCommitmentConfirm(MsgDataCommitmentConfirm) returns (MsgDataCommitmentConfirmResponse) {
+  // DataCommitmentConfirm allows the validators to submit a confirmation for a
+  // data commitment.
+  rpc DataCommitmentConfirm(MsgDataCommitmentConfirm)
+      returns (MsgDataCommitmentConfirmResponse) {
```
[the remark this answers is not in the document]
evan-forbes: `make proto-format`

x/qgb/abci.go
SweeXordious: aaah, gotta make some more resources on validators in tendermint. Thanks

evan-forbes: would recommend looking into the [validator set](https://github.com/celestiaorg/celestia-core/blob/e0e519802a1a01059ac003ea627f3619b5d76214/types/validator_set.go#L113-L139), which is updated during `EndBlock` by the staking module. also related: the specs in the sdk are really great, and we will eventually add similar things to the qgb module https://github.com/cosmos/cosmos-sdk/tree/master/x/staking/spec

SweeXordious: Thanks

SweeXordious:
Trying to resolve the conflicts and make the project build in here: https://github.com/celestiaorg/celestia-app/pull/253

liamsi (commented):
Would it be possible to split this in a few smaller PRs?

evan-forbes:
closing in favor of #257, #255, and the upcoming relayer PR

evan-forbes:
closing in favor of #276

PR #242, opened 2022-03-11
Adds msg server data commitment test for QGB module

SweeXordious:
Adds a simplified integration test for data commitments in QGB module


evan-forbes (approved):
:+1: :+1: 

nice!

Do we have plans to add queries for the two new keeper getters? I don't think we need specific queries for them, so I don't really see the need atm, just curious.

SweeXordious:
@evan-forbes Only one is added. The other was just copied from another file to its right place.
I think we might need the new one when doing tests (I used it today to double check). So yes, I guess we can leave it.

Issue #241, opened 2022-03-11
Decide on additional indexing keys needed for qgb module's keeper


evan-forbes:
As pointed out by @SweeXordious in #240 and other PRs, we need to index certain data in the qgb module's state to quickly reference it when querying. We should decide on exactly which queries we want, and how to index them.

evan-forbes:
can we close this, or are we still working on indexing?

SweeXordious:
I think we have basic indexing now. We can close until we need better stuff and then think it more

PR #246, opened 2022-03-14
Adds simple indexing to qgb module

SweeXordious:
Uses index prefix to query for commitments.

Adds to: https://github.com/celestiaorg/celestia-app/issues/241

Currently, the keys are as follows: 
https://github.com/celestiaorg/celestia-app/blob/8209e23a3645209a8c81214026eadd8c9607521d/x/qgb/types/keys.go#L111

### DIFF
--- a/x/qgb/keeper/keeper_data_commitment.go
+++ b/x/qgb/keeper/keeper_data_commitment.go
@@ -31,7 +31,7 @@ func (k Keeper) GetDataCommitmentConfirmsByCommitment(
 	commitment string,
 ) (confirms []types.MsgDataCommitmentConfirm) {
 	store := ctx.KVStore(k.storeKey)
-	iterator := store.Iterator(nil, nil) // Can we make this faster?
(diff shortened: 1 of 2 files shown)


evan-forbes:
Indexing by commitment is useful, but it might be even more useful to index by nonce instead, mainly because we wouldn't have to generate the commitment ahead of time.

Specifically, I could see this being useful for both https://github.com/celestiaorg/quantum-gravity-bridge/issues/54 and https://github.com/celestiaorg/quantum-gravity-bridge/issues/55

evan-forbes:
we can keep this the way it is now though, just something to think about

SweeXordious:
What is a nonce for a data commitment message?

evan-forbes:
> What is a nonce for a data commitment message?
We don't have a nonce yet for the data commitment, but we will need one once this is expected by the contracts, and will make slashing easier. Currently the mvp relayer just generates the nonce, but that's not ideal.

Issue #217, opened 2022-02-18
Document QGB integration


evan-forbes:
The QGB integration to the app includes two new transactions that validators will have to create and submit on a routine basis. These need to be well documented along with any potential consequences for failure to do so properly. 

This needs to be done before we can merge the `qgb-integration` branch into master.

evan-forbes:
to clarify we need documentation for validators

evan-forbes:
todo: move to new repo

PR #250, opened 2022-03-15
Adds QGB related ADRs

SweeXordious:
Closes: https://github.com/celestiaorg/celestia-app/issues/217


SweeXordious:
@evan-forbes Please take a look on this if it is in the right direction. And, if you have any idea what needs to be added to this :D Thanks a lot

docs/architecture/ADR-002-QGB-Overview.md
```diff
@@ -0,0 +1,75 @@
...
+## Changelog
+
+- {date}: {changelog}
+
+## Context
+
+> This section contains all the context one needs to understand the current state, and why there is a problem.
+> It should be as succinct as possible and introduce the high level idea behind the solution.
+
+Following the [Quantum Gravity Bridge](https://github.com/celestiaorg/quantum-gravity-bridge) development, we decided to make the orchestrator and relayer part of Celestia-app.
```
evan-forbes, line 12: This is an important statement to discuss, but I think starting off with something more fundamental to why we could do this in the first place would be useful. Rewriting the relayer and the orchestrator was done because our relayer can be designed to be trust minimized, where a two way bridge requires would require each validator to run a trusted relayer. Since we had to rewrite them and they were simplified, we decided to move things to the app. Either in this section or another, we could also compare and contract the consequences of moving to the app.

We might also want to discuss how we are using a similar design to a portion of the normal gravity bridge's application integration. Technically, we don't even need the additional state for the MVP, we only need the state for slashing, but since we are planning on doing slashing in the future, the state changes make it trivial for the relayer to find the data needed to relay to the evm chain.

evan-forbes (commented):
This is a great start! :+1:  Originally I had only imagined a single ADR, but I like having three smaller ones. Left some comments

SweeXordious:
@evan-forbes Please take a look when you have time again

docs/architecture/ADR-002-QGB-Overview.md
```diff
@@ -0,0 +1,47 @@
+# ADR 003: ValSet
+
+## Changelog
+
+- {date}: {changelog}
+
+## Context
+Various Rollups on EVM L1s may have a need for using Celestia as a DA layer, while keeping fraud/validity proof verification in the EVM.
+To that end, two concrete possibilities exist: the Gravity bridge and a native EVM light client.
```
evan-forbes, line 9: as discussed in the sync, we want to focus on the app integration portion here. I'm still thinking on exactly what needs to be covered, because technically we don't need any of the app integration until slashing. If we didn't need to slash validators, then we would do everything off chain to save state bloat.

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+It works by relying on a set of signers to attest to some event on Celestia: the Celestia validator set.
+
+The QGB contract keeps track of the Celestia validator set by updating its view of the validator set with `updateValidatorSet()`.
+More than 2/3 of the voting power of the current view of the validator set must sign off on new relayed events, submitted with
+[`submitDataRootTupleRoot()`](https://github.com/celestiaorg/quantum-gravity-bridge/blob/980b9c68abc34b8d2e4d20ca644b8aa3025a239e/ethereum/solidity/src/QuantumGravityBridge.sol#L328).
+Each event is a batch of `DataRootTuples`, with each tuple representing a single data root (i.e. block header).
+Relayed tuples are in the same order as Celestia block headers.
+For more details, check the data commitment ADR.
+
+Finally, if there are no validator set updates for the unbonding window, the bridge must halt.
```
evan-forbes, line 28: A++

evan-forbes (commented):
thinking on this a bit more, we might just want to not include the overview ADR yet. We only need to add state to the app for the qgb when we have slashing, if we weren't planning on doing that, then we would do everything off chain. Unless we want to talk about that, I actually think its okay to get rid of the overview portion. Perhaps we can add in some grander overview later? over perhaps in the qgb repo. Still not sure yet.

The other two ADRs are good tho, and we should definitely include them

docs/architecture/ADR-002-QGB-Overview.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Approaches
+The following issue summarizes the two viable approaches to accommodate this: [#4](https://github.com/celestiaorg/quantum-gravity-bridge/issues/4)
+
+## Decision
+We decided to go for the gravity bridge approach and move the orchestrator/relayer logic to Celestia-app.
+
+## Detailed Design
+The QGB allows Celestia block header data roots to be relayed in one direction, from Celestia to an EVM chain. It does not support
+bridging assets such as fungible or non-fungible tokens directly, and cannot send messages from the EVM chain back to Celestia.
```
adlerjohn, line 19: Don't manually add newlines for markdown files, except maybe after sentences if you want each sentence on a different line for git diffs.
```suggestion
The QGB allows Celestia block header data roots to be relayed in one direction, from Celestia to an EVM chain. It does not support bridging assets such as fungible or non-fungible tokens directly, and cannot send messages from the EVM chain back to Celestia.
```

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+`MsgSetOrchestratorAddress` and `MsgValsetConfirm`) and relayer queries (such as `ValsetConfirm` and `GetDelegateKeyByOrchestrator`).
+
+It works by relying on a set of signers to attest to some event on Celestia: the Celestia validator set.
+
+The QGB contract keeps track of the Celestia validator set by updating its view of the validator set with `updateValidatorSet()`.
+More than 2/3 of the voting power of the current view of the validator set must sign off on new relayed events, submitted with
+[`submitDataRootTupleRoot()`](https://github.com/celestiaorg/quantum-gravity-bridge/blob/980b9c68abc34b8d2e4d20ca644b8aa3025a239e/ethereum/solidity/src/QuantumGravityBridge.sol#L328).
+Each event is a batch of `DataRootTuples`, with each tuple representing a single data root (i.e. block header).
+Relayed tuples are in the same order as Celestia block headers.
+For more details, check the data commitment ADR.
```
adlerjohn, line 26: Link to ADR?

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+
+Finally, if there are no validator set updates for the unbonding window, the bridge must halt.
+
+### When are validator sets created
```
adlerjohn, line 30: Do we need to submit a "new" validator set within the unbonding window even if it hasn't changed? Otherwise the bridge contract doesn't know if the bridge should shut down due to no updates within the unbonding window, or if there simply were no updates.

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+
+### Message types
+We added the following messages types:
+
+#### Bridge Validator
+The `BridgeValidator` represents a validator's ETH address and its power.
+```protobuf
+message BridgeValidator {
+  uint64 power            = 1;
+  string ethereum_address = 2;
```
adlerjohn, line 45: Why `string` and not `bytes`?

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+- `ethereum_address`: the Ethereum address that will be used by the validator to sign messages.
+
+#### ValSet
+`Valset` is the Ethereum Bridge Multsig Set, each qgb validator also maintains an ETH key
+to sign messages, these are used to check signatures on ETH because of the significant gas savings.
+```protobuf
+message Valset {
+  uint64                   nonce   = 1;
+  repeated BridgeValidator members = 2 [(gogoproto.nullable) = false];
+  uint64                   height  = 3;
```
adlerjohn, line 59: Alignment here is on types, while alignment about is on field number. Make uniform.

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+}
+```
+It contains:
+- `nonce`: a unique number referencing the `ValSet`.
+- `BridgeValidator`: a list of [BridgeValidator](#Bridge-Validator) containing the current validator set.
+- `height`: the current chain height.
```
adlerjohn, line 65: These should honestly be comments in the protobuf, so that when it's copy-pasted the comments are there.

docs/architecture/ADR-003-QGB-ValSet.md
SweeXordious: If I understand right, https://github.com/SweeXordious/celestia-app/blob/e4760741315840395c589d44414a185c1f641084/x/qgb/abci.go#L42
we will always post valsets requests even without it being changed. @evan-forbes Correct me if I am wrong please

docs/architecture/ADR-003-QGB-ValSet.md
SweeXordious: Because we will be submitting `0x...` hex encoded addresses

docs/architecture/ADR-003-QGB-ValSet.md
SweeXordious: Will put these in proto files also, then

docs/architecture/ADR-003-QGB-ValSet.md
SweeXordious: check now please if formatting is fixed ?

docs/architecture/ADR-003-QGB-ValSet.md
SweeXordious: We don't have an ADR in QGB, no ? I only found tracking issues. And these;
https://github.com/celestiaorg/quantum-gravity-bridge/issues/20
https://github.com/celestiaorg/quantum-gravity-bridge/issues/4
Do you mean one of them ?

docs/architecture/ADR-003-QGB-ValSet.md
evan-forbes: > Do we need to submit a "new" validator set within the unbonding window even if it hasn't changed? Otherwise the bridge contract doesn't know if the bridge should shut down due to no updates within the unbonding window, or if there simply were no updates.

it looks like it should work, but we don't have a test specifically for this. I'll write this down as a test case, and then @SweeXordious and I can plan around it today

docs/architecture/ADR-003-QGB-ValSet.md
```diff
@@ -0,0 +1,272 @@
...
+It contains:
+- `validator`: a `celesvaloper1` address referencing the validator in the current `ValSet`.
+- `orchestrator`: a `celes1` account address referencing the key that is being delegated to.
+- `eth_address`: the hex `0x` encoded Ethereum public key that will be used by this validator on Ethereum.
+
+#### ValSetConfirm
+`MsgValsetConfirm` is the message sent by the validators when they wish to submit their signatures
+over the validator set at a given block height. A validator must first call `MsgSetEthAddress` to
+set their Ethereum address to be used for signing.
+Then, someone (anyone) must make a `ValsetRequest`, the request is essentially a messaging mechanism
```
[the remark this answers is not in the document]
SweeXordious: maybe:
```
Then, using `EndBlock`, the protocol makes a `ValsetRequest`...
```
?

evan-forbes (approved):
nice! all the documentation in the proto files was a great idea/addition.

:+1: :+1: LGTM

SweeXordious:
@evan-forbes Can I merge then?

docs/architecture/ADR-003-QGB-ValSet.md
evan-forbes: I think John was referring simply to the other ADR included in this PR. We can just link to this that file in this PR I think

evan-forbes:
might want to wait on @adlerjohn approval, but all my nits were covered I think

adlerjohn (approved):

PR #255, opened 2022-03-28
Orchestrator and relayer client

evan-forbes:
## Description

This PR is the first of three PRs to add the MVP orchestrator and relayer. It contains the various rpc clients used by both the relayer and orchestrator to communicate with celestia-app, celestia-core, and ethereum.

part 1/3 of the orchestrator/relayer MVP


x/qgb/orchestrator/client.go
```diff
@@ -0,0 +1,171 @@
...
+	"context"
+	"crypto/ecdsa"
+	"fmt"
+	"math/big"
+	"os"
+	"strings"
+	"syscall"
+
+	wrapper "github.com/celestiaorg/quantum-gravity-bridge/ethereum/solidity/wrappers/QuantumGravityBridge.sol"
+	"github.com/celestiaorg/quantum-gravity-bridge/orchestrator/ethereum/keystore"
```
SweeXordious, line 13: Shouldn't we move this here? instead of importing ?

evan-forbes:
there's still a lot of unused code that will get used later, so the linter is failing

x/qgb/orchestrator/client.go
evan-forbes: good idea, https://github.com/celestiaorg/celestia-app/pull/255/commits/cc4f6009a65e1eb68b23b83032f17821346b4335

Issue #223, opened 2022-02-23
Add the qgb orchestrator as a sub command of the app


evan-forbes:
It would be nice for validators to not have to install/update separate matching binaries for the qgb and the app. Instead, we should include a sub command to the app that will run the orchestrator.

PR #257, opened 2022-03-30
MVP Orchestrator

evan-forbes:
## Description

This PR introduces the minimum viable orchestrator for the quantum gravity bridge. Still needs to tested with the most recent refactors, along with some documentation additions, but we can start review process.

closes: part 2/3 #223


evan-forbes:
I planned on having an additional command that queries for the necessary data and deploys the smart contracts, but ran into a really weird vs code bug when importing the go bindings that crippled my ide/go modules. We will likely want something similar to that command before merging to better test both the orchestrator and relayer.

proto/qgb/msgs.proto
```diff
@@ -61,14 +61,9 @@ message MsgSetOrchestratorAddressResponse {}
 // chain store and submit them to Ethereum to update the validator set
 // -------------
 message MsgValsetConfirm {
-  // Unique number referencing the `ValSet`.
```
SweeXordious, line 64: why removing the comments?

proto/qgb/query.proto
```diff
@@ -41,7 +48,11 @@ service Query {
     option (google.api.http).get = "/qgb/data_commitment/confirms/begin_block/"
                                    "{begin_block}/end_block/{end_block}";
   }
-
+  // LastValsetRequests queries for the latest valset requests
+  rpc LastValsetRequests(QueryLastValsetRequestsRequest)
+      returns (QueryLastValsetRequestsResponse) {
+    option (google.api.http).get = "/gravity/v1beta/valset/requests";
```
SweeXordious, line 54:
```suggestion
    option (google.api.http).get = "/qgb/valset/requests";
```

x/qgb/keeper/query_valset_confirm.go
```diff
@@ -25,3 +26,27 @@ func (k Keeper) ValsetConfirmsByNonce(
 	confirms := k.GetValsetConfirms(sdk.UnwrapSDKContext(c), req.Nonce)
 	return &types.QueryValsetConfirmsByNonceResponse{Confirms: confirms}, nil
 }
+
+const maxValsetRequestsReturned = 5
```
SweeXordious, line 30: Shouldn't we create a specific package containing the consts ?

x/qgb/keeper/query_valset_confirm.go
```diff
@@ -25,3 +26,27 @@ func (k Keeper) ValsetConfirmsByNonce(
...
+
+// LastValsetRequests queries the LastValsetRequests of the gravity module
```
SweeXordious, line 32:
```suggestion
// LastValsetRequests queries the LastValsetRequests of the qgb module
```

x/qgb/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,195 @@
...
+	queryClient := types.NewQueryClient(oc.qgbRPC)
+
+	resp, err := queryClient.Params(ctx, &types.QueryParamsRequest{})
+	if err != nil {
+		return err
+	}
+
+	params := resp.Params
+
+	results, err := oc.tendermintRPC.Subscribe(ctx, "height", coretypes.EventQueryNewBlockHeader.String())
```
SweeXordious, line 101: This means data commitment confirms will be consecutive

x/qgb/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := paytypes.BroadcastTx(ctx, oc.qgbRPC, 1, rawTx)
+	if err != nil {
+		return err
+	}
+
+	if resp.TxResponse.Code != 0 {
```
SweeXordious, line 190: We need some retry mechanism. Since, the data commitments confirms will be consecutive. If enough fo these fail, then it would create a gap

x/qgb/types/datacommitmentconfirm.go
```diff
@@ -50,3 +55,33 @@ func (msg *MsgDataCommitmentConfirm) ValidateBasic() (err error) {
 
 // Type should return the action
 func (msg *MsgDataCommitmentConfirm) Type() string { return "data_commitment_confirm" }
+
+// EncodeDomainSeparatedDataCommitment takes the required input data and produces the required signature to confirm a validator
+// set update on the Peggy Ethereum contract. This value will then be signed before being
```
SweeXordious, line 60:
```suggestion
// set update on the QGB Ethereum contract. This value will then be signed before being
```

proto/qgb/genesis.proto
```diff
@@ -23,4 +23,4 @@ message GenesisState { Params params = 1; }
 message GravityNonces {
   // the nonce of the last generated validator set
   uint64 latest_valset_nonce = 1;
-}
\ No newline at end of file
+}
```
SweeXordious, line 26: Shouldn't we change `GravityNonces` -> `QgbNonces` for example? or even remove the prefix ?

proto/qgb/types.proto
```diff
@@ -5,20 +5,15 @@ option go_package = "github.com/celestiaorg/celestia-app/x/qgb/types";
 
 // BridgeValidator represents a validator's ETH address and its power
 message BridgeValidator {
-  // Voting power of the validator.
   uint64 power = 1;
-  // Ethereum address that will be used by the validator to sign messages.
   string ethereum_address = 2;
 }
 
 // Valset is the Ethereum Bridge Multsig Set, each gravity validator also
```
SweeXordious, line 12:
```suggestion
// Valset is the Ethereum Bridge Multsig Set, each qgb validator also
```

SweeXordious (commented):
Great job :+1: Left some basic comments

proto/qgb/msgs.proto
adlerjohn: Yeah, please keep!

x/qgb/orchestrator/client.go
```diff
@@ -189,3 +189,20 @@ func ethPassFromStdin() (string, error) {
...
+
+// SigToVRS breaks apart a signature into its components to make it compatible with the contracts
+func SigToVRS(sigHex string) (v uint8, r, s ethcmn.Hash) {
+	signatureBytes := ethcmn.FromHex(sigHex)
+	vParam := signatureBytes[64]
+	if vParam == byte(0) {
+		vParam = byte(27)
+	} else if vParam == byte(1) {
+		vParam = byte(28)
+	}
```
adlerjohn, line 201: Is it guaranteed `vParam` is either zero or one? If not, can we check here?

x/qgb/orchestrator/orchestrator.go
evan-forbes: I agree! This is something that I figured that we could dig into and think about more after merging imho

We should definitely write an issue about this before merging

x/qgb/keeper/query_valset_confirm.go
evan-forbes: that's a good idea, tbh this was just copied over from the gravity module.

proto/qgb/genesis.proto
evan-forbes: Not sure how that snuck in there, we don't need this struct at all.

x/qgb/orchestrator/client.go
evan-forbes: I just copied this over from here https://github.com/celestiaorg/quantum-gravity-bridge/blob/6b6efce45c4ba67cf277d5e34c11c33e15084b25/orchestrator/ethereum/peggy/peggy_contract.go#L90-L104

I don't think it is guaranteed, but then we just want to use whatever is without changing it afaict. Again, I don't actually know why we have to replace these bytes. I tried to find some docs on the encoding of ethereum signatures, but didn't find any within a short period of time. I can look into this before merging if we want, or we can write an issue to investigate further.

x/qgb/orchestrator/orchestrator.go
evan-forbes: atm the moment yes, I figured this was a safe enough assumtion for mvp, but we do have an issue to consider other mechanisms https://github.com/celestiaorg/quantum-gravity-bridge/issues/54

x/qgb/orchestrator/client.go
adlerjohn: https://ethereum.stackexchange.com/a/113505

x/qgb/keeper/query_valset_confirm.go
SweeXordious: created issue: https://github.com/celestiaorg/celestia-app/issues/262

x/qgb/orchestrator/orchestrator.go
SweeXordious: this issue is enough I guess: https://github.com/celestiaorg/celestia-app/issues/260

x/qgb/orchestrator/client.go
evan-forbes: so is this behavior correct as is? do we need to investigate this further?

x/qgb/keeper/query_valset_confirm.go
evan-forbes: Thinking about this more, while it is a constant, it is not exported, so I'm not sure how much the net gain is by moving it. It seems like it is meant to be a local constant imho. That being said, it is still in a weird spot, as constants typically are found at the top of a go file. Moved it there for now, but still keeping #262 open so we can think about this more as we continually work on this module.

x/qgb/orchestrator/client.go
SweeXordious:
```
If block.number >= FORK_BLKNUM and CHAIN_ID is available, then when computing the hash of a transaction for the purposes of signing, instead of hashing only six rlp encoded elements (nonce, gasprice, startgas, to, value, data), you SHOULD hash nine rlp encoded elements (nonce, gasprice, startgas, to, value, data, chainid, 0, 0). If you do, then the v of the signature MUST be set to {0,1} + CHAIN_ID * 2 + 35 where {0,1} is the parity of the y value of the curve point for which r is the x-value in the secp256k1 signing process. If you choose to only hash 6 values, then v continues to be set to {0,1} + 27 as previously.

If block.number >= FORK_BLKNUM and v = CHAIN_ID * 2 + 35 or v = CHAIN_ID * 2 + 36, then when computing the hash of a transaction for purposes of recovering, instead of hashing six rlp encoded elements (nonce, gasprice, startgas, to, value, data), hash nine rlp encoded elements (nonce, gasprice, startgas, to, value, data, chainid, 0, 0). The currently existing signature scheme using v = 27 and v = 28 remains valid and continues to operate under the same rules as it did previously.
```
From [EIP-155](https://eips.ethereum.org/EIPS/eip-155).

IMO, we add an `else` branch where the bridge panics if it hits it (so we know if we're having other values than 0 and 1. Some threads say that 2 and 3 are also possible. More on this below).

Also, we need to check which fields we're signing, so that in case we're signing 9 fields, then we replace 27 and 28 with 37 and 38.

**On `v` value**: it is used for recovery, and  ECDSA allows having 4 points referring to the same signature [here](https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v/38909#38909). Thus, it's possible for `v` to be in `{0, 1, 2, 3}`. However, AFAIK, to solve ECDSA signatures malleability, Ethereum removed half the field. Thus, only two values for `v` should be valid.

x/qgb/types/abi_consts.go
```diff
@@ -0,0 +1,170 @@
...
+	ExternalQGBabi     abi.ABI
+	InternalQGBabi     abi.ABI
+	BridgeValidatorAbi abi.Arguments
+
+	VsDomainSeparator ethcmn.Hash
+	DcDomainSeparator ethcmn.Hash
+)
+
+func init() {
+	contractAbi, err := abi.JSON(strings.NewReader(wrapper.QuantumGravityBridgeMetaData.ABI))
```
SweeXordious, line 142: Any idea why this is empty: 
https://github.com/celestiaorg/quantum-gravity-bridge/blob/6b6efce45c4ba67cf277d5e34c11c33e15084b25/ethereum/solidity/wrappers/QuantumGravityBridge.sol/wrapper.go#L59
?

x/qgb/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,198 @@
...
+		case ev := <-results:
+			err = oc.processDataCommitmentEvents(ctx, params.DataCommitmentWindow, ev)
+			if err != nil {
+				oc.logger.Err(err)
+			}
+		}
+	}
+}
+
+func (oc *orchestrator) processDataCommitmentEvents(ctx context.Context, window uint64, ev rpctypes.ResultEvent) error {
```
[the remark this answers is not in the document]
SweeXordious: I guess it makes sense to change the name since we have a similar method for relayer and might lead to confusion (even if no exported).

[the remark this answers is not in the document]
evan-forbes: personally I think having one function called `orchestratrateDataCommitments` which calls another function called `orchestratrateDataCommitmentsEvents` even more confusing :shrug: In both the orchestrator and the relayer, they are processing those types of events. Ideally, we would have an interface for this and everything would be abstracted out cleanly, but I didn't think that was worth the time for the mvp.

its a toss up either way, so happy to rename if you still think otherwise

x/qgb/types/abi_consts.go
evan-forbes: Aren't we referencing this instead? 

https://github.com/celestiaorg/quantum-gravity-bridge/blob/6b6efce45c4ba67cf277d5e34c11c33e15084b25/ethereum/solidity/wrappers/QuantumGravityBridge.sol/wrapper.go#L765-L767

x/qgb/orchestrator/client.go
adlerjohn: I'm surprised there's no module in ethermint/some EVM compatibility layer that the Cosmos ecosystem provides to handle this properly.

adlerjohn: Note that 2 and 3 are rare:

> 27 = lower X even Y. 28 = lower X odd Y. 29 = higher X even Y. 30 = higher X odd Y. Note that 29 and 30 are exceedingly rarely, and will in practice only ever be seen in specifically generated examples. There are only two possible X values if r is between 1 and (p mod n), which has a chance of about 0.000000000000000000000000000000000000373 % to happen randomly – [Pieter Wuille](https://bitcoin.stackexchange.com/users/208/pieter-wuille)

SweeXordious: They're rare in Bitcoin. However, in Ethereum, they're considered invalid IIUR.

> All transaction signatures whose s-value is greater than secp256k1n/2 are now considered invalid. The ECDSA recover precompiled contract remains unchanged and will keep accepting high s-values; this is useful e.g. if a contract recovers old Bitcoin signatures.

and,

> Allowing transactions with any s value with 0 < s < secp256k1n, as is currently the case, opens a transaction malleability concern, as one can take any transaction, flip the s value from s to secp256k1n - s, flip the v value (27 -> 28, 28 -> 27), and the resulting signature would still be valid. This is not a serious security flaw, especially since Ethereum uses addresses and not transaction hashes as the input to an ether value transfer or other transaction, but it nevertheless creates a UI inconvenience as an attacker can cause the transaction that gets confirmed in a block to have a different hash from the transaction that any user sends, interfering with user interfaces that use transaction hashes as tracking IDs. Preventing high s values removes this problem.

from [here](https://eips.ethereum.org/EIPS/eip-2).

Thus, we shouldn't have `2` or `3` values... But, I guess adding an `else` branch is safer and will let us  know in case we encounter something similar.

adlerjohn: Doesn't hurt to catch that case, yeah.

evan-forbes:
closing in favor of #276

PR #196, opened 2022-01-30
Micro Celes

marbar3778:
## Description

closes: #XXXX

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #267, opened 2022-03-31
testutil/network: Refactor network test utility to add CELES and UCELES to genesis

renaynay:
Refactor network test utility to add both CELES and UCELES to genesis.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -88,6 +88,8 @@ import (
 const (
 	AccountAddressPrefix = "celes"
 	Name                 = "celestia-app"
+	// TokenDenom defines the name of the native staking token.
+	TokenDenom = "celes"
 	// BondDenom defines the native staking token denomination.
 	BondDenom = "uceles"
 	// DisplayDenom defines the name, symbol, and display value of the celes token.
--- a/testutil/network/network.go
+++ b/testutil/network/network.go
@@ -122,6 +122,7 @@ func newGenAccout(kr keyring.Keyring, name string, amount int64) (authtypes.Gene
 		sdk.NewCoin(fmt.Sprintf("%stoken", name), sdk.NewInt(amount)),
 		sdk.NewCoin(app.BondDenom, sdk.NewInt(amount)),
 		sdk.NewCoin(app.DisplayDenom, sdk.NewInt(amount)),
+		sdk.NewCoin(app.TokenDenom, sdk.NewInt(amount)),
 	)
 
 	bal := banktypes.Balance{
--- a/testutil/network/network.go
+++ b/testutil/network/network.go
@@ -121,6 +121,7 @@ func newGenAccout(kr keyring.Keyring, name string, amount int64) (authtypes.Gene
 	balances := sdk.NewCoins(
 		sdk.NewCoin(fmt.Sprintf("%stoken", name), sdk.NewInt(amount)),
 		sdk.NewCoin(app.BondDenom, sdk.NewInt(amount)),
+		sdk.NewCoin(app.DisplayDenom, sdk.NewInt(amount)),
 	)
 
 	bal := banktypes.Balance{
--- a/app/module.go
+++ b/app/module.go
@@ -42,6 +42,11 @@ func (bankModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
 				Exponent: 6,
 				Aliases:  []string{},
 			},
+			{
+				Denom:    TokenDenom,
+				Exponent: 6,
+				Aliases:  []string{},
+			},
 		},
 	}
 


evan-forbes (commented):
I'm tempted to merge this since its blocking, but we will change it very soon after without additional changes

testutil/network/network.go
```diff
@@ -121,6 +121,7 @@ func newGenAccout(kr keyring.Keyring, name string, amount int64) (authtypes.Gene
 	balances := sdk.NewCoins(
 		sdk.NewCoin(fmt.Sprintf("%stoken", name), sdk.NewInt(amount)),
 		sdk.NewCoin(app.BondDenom, sdk.NewInt(amount)),
+		sdk.NewCoin(app.DisplayDenom, sdk.NewInt(amount)),
```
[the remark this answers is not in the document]
evan-forbes: I think that DisplayDenom was meant to be a reference for any exchanges, but not actually used in code, but I could be wrong. 

@marbar3778 do you think we should keep this as is or use lower case?

testutil/network/network.go
```diff
@@ -121,6 +121,8 @@ func newGenAccout(kr keyring.Keyring, name string, amount int64) (authtypes.Gene
...
```
evan-forbes, line 124: we have the correct token denomination, so we don't have to hold up this PR, but we might also want to get rid of giving CELES tokens just to avoid any confusion for future readers.

approving without tho

evan-forbes (approved):
:+1:

app/app.go
```diff
@@ -88,6 +88,8 @@ import (
 const (
 	AccountAddressPrefix = "celes"
 	Name                 = "celestia-app"
+	// TokenDenom defines the name of the native staking token.
+	TokenDenom = "celes"
```
marbar3778, line 92: why is this being added? celes is not a denom on chain, 1,000,000uceles = 1 celes

testutil/network/network.go
```diff
@@ -121,6 +121,8 @@ func newGenAccout(kr keyring.Keyring, name string, amount int64) (authtypes.Gene
...
+		sdk.NewCoin(app.TokenDenom, sdk.NewInt(amount)),
```
marbar3778, line 125: there are two entries for celes here? one upper case and one lower,  is that on purpose?

marbar3778 (commented):
not sure I follow what this is aiming at solving, I think this will cause more confusion if used

evan-forbes (changes requested):
To attempt to clear up any confusion, all caps CELES was being added to the default genesis as a macro denomination of uceles. I think I was confused by this, so this PR changed that to lower case, that why we added 
`TokenDenom = "celes"`
https://github.com/renaynay/celestia-app/blob/0d496d75394aa9b003d1da7a0a30e4e4c8fb4e22/app/app.go#L92

to use in place of `DisplayDenom`. Tbh, I'm not sure why `DisplayDenom` exists in the first place, it was added in https://github.com/celestiaorg/celestia-app/pull/196 but I missed its purpose there. 

Its clear now that when creating a genesis file for testing, we should just be using `uceles`, which is the last change needed before this PR can be merged imho

renaynay:
@marbar3778 can you help here?

marbar3778:
you can have it as lower case or upper case for the display denom, Uppercase is inline with some chains but others like hub use lowercase. Uppercase is a bit better imo since that is what the token ticker will be. 


Using uceles as the default for testing should be done otherwise tooling you may build will rely on Celes but 1 Celes is not one celes but a micro celes, unless your equivalent of Wei or uatom will be Celes.

testutil/network/network.go
marbar3778: right now, if you try to use celes or CELES for staking it won't work, you will only be able to transfer the token since the system relies on uceles

evan-forbes: yes, we should remove this before merging

testutil/network/network.go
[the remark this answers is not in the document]
evan-forbes: I still think that we should remove this addition of Display denom to the default genesis file before merging

app/app.go
evan-forbes: This was added to replace the usage of DisplayDenom in the default genesis afaiu

evan-forbes:
> you can have it as lower case or upper case for the display denom, Uppercase is inline with some chains but others like hub use lowercase. Uppercase is a bit better imo since that is what the token ticker will be.

Using uceles as the default for testing should be done otherwise tooling you may build will rely on Celes but 1 Celes is not one celes but a micro celes, unless your equivalent of Wei or uatom will be Celes.

thanks for this clarification, I have no opinion on which we use, but lets pick one before merging

liamsi:
I think this can be closed as we are switching to [#token](https://github.com/celestiaorg/celestia-app/issues/279)

PR #273, opened 2022-04-05
Fix the ci linter

evan-forbes:
## Description

fixes the CI linter by specifying a specific version instead of automatically using the latest version of go

it also removes the specific version of tendermint in the replace directive, as we want to replace it across all inherited dependencies, not only for a specific verison


liamsi (approved):

adlerjohn (approved):
lfg

PR #293, opened 2022-04-12
doc: update PFM spec

renaynay:
I mostly made this PR to track some misunderstandings I had with the spec for using PFM.

### DIFF
--- a/x/payment/spec/docs.md
+++ b/x/payment/spec/docs.md
@@ -113,7 +113,7 @@ if err != nil {
 }
 
 // query for account information necessary to sign a valid tx
-err = keyringSigner.QueryAccount(ctx, grpcClientConn)
+err = keyringSigner.QueryAccountNumber(ctx, grpcClientConn)
 if err != nil {
     return err
 }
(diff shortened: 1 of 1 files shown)


x/payment/spec/docs.md
```diff
@@ -113,15 +113,15 @@ if err != nil {
 }
 
 // query for account information necessary to sign a valid tx
-err = keyringSigner.QueryAccount(ctx, grpcClientConn)
+err = keyringSigner.QueryAccountNumber(ctx, grpcClientConn)
```
renaynay, line 116: Is this the method we should be using?

evan-forbes: this brings up a really good point. This method appears to be vestigial, and should likely not be exported. Instead, it should be private and only called during the keyringSigner's BroadcastTx method.

evan-forbes: #294

renaynay: For the moment though, is this a necessary component to building and submitting a PFD?

liamsi:
what is the status of this PR?

evan-forbes:
I think we can close, as I believe we covered this issue synchronously but will defer to @renaynay

Issue #297, opened 2022-04-13
[bug]: Nondescript error code returned when trying to submit tx using reserved `namespaceID`


renaynay:
Attempted to submit a tx with `nID = []byte{0,0,0,0,0,0,0,0}`

Got: 
```
        Response:
          TxHash: BD4DDDEF78464185F1FB96F7904EDA568435E83DDC9CFA4EBDECAAE35751E83D
          Code: 1
          Raw Log: internal
          GasUsed: 6474
          Codespace: undefined
```

Instead of returning `internal`, it should return a `reserved namespace` error or something like that.

cc @liamsi 

Btw, when submitting a PFM with `nID = []byte{0, 0, 1, 0, 0, 0, 0, 0}`, I get success: 

```
        Response:
          Height: 8164
          TxHash: 8D7FBF77ACBAE66875438BBDD33C3DCB94C33882A3E396C52C52A87D4FAB456C
          Data: 0A1B0A192F7061796D656E742E4D7367506179466F724D657373616765
          Raw Log: [{"events":[{"type":"message","attributes":[{"key":"action","value":"/payment.MsgPayForMessage"}]},{"type":"payformessage","attributes":[{"key":"signer","value":"celes1vdjkcetnx9khvun3d4kx2emvx36hxertw3skwamp0furvvrnvaerqut5d4nns6nc8pmrvvcxu6e2w"},{"key":"size","value":"256"}]}]}]
          Logs: [{"events":[{"type":"message","attributes":[{"key":"action","value":"/payment.MsgPayForMessage"}]},{"type":"payformessage","attributes":[{"key":"signer","value":"celes1vdjkcetnx9khvun3d4kx2emvx36hxertw3skwamp0furvvrnvaerqut5d4nns6nc8pmrvvcxu6e2w"},{"key":"size","value":"256"}]}]}]
          GasWanted: 200000
          GasUsed: 38486
```

evan-forbes:
> Btw, when submitting a PFM with nID = []byte{0, 0, 1, 0, 0, 0, 0, 0}, I get success:

wait, should this fail or is this related to something else?

liamsi:
No, I think Rene is just confirming that this works with a proper non-reserved namespace.

renaynay:
No I'm just saying the happy case works fine @evan-forbes

liamsi:
We are trying to say the same thing Rene :D 

This should help resolve this:
[...]

PR #300, opened 2022-04-13
improve error handling

liamsi:
close: #297

### DIFF
--- a/x/payment/types/errors.go
+++ b/x/payment/types/errors.go
@@ -6,7 +6,12 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-// x/payment module sentinel errors
 var (
-	ErrSample = sdkerrors.Register(ModuleName, 1100, "sample error")
+	ErrReservedNamespace              = sdkerrors.Register(ModuleName, 11110, "cannot use reserved namespace IDs")
+	ErrInvalidNamespaceLen            = sdkerrors.Register(ModuleName, 11111, "invalid namespace length")
+	ErrInvalidDataSize                = sdkerrors.Register(ModuleName, 11112, "data must be multiple of shareSize")
+	ErrDeclaredActualDataSizeMismatch = sdkerrors.Register(ModuleName, 11113, "declared data size does not match actual size")
+	ErrCommittedSquareSizeNotPowOf2   = sdkerrors.Register(ModuleName, 11114, "committed to invalid square size: must be power of two")
+	ErrCalculateCommit                = sdkerrors.Register(ModuleName, 11115, "unexpected error calculating commit for share")
+	ErrInvalidShareCommit             = sdkerrors.Register(ModuleName, 11116, "invalid commit for share")
 )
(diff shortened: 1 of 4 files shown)


x/payment/types/payformessage_test.go
```diff
@@ -257,105 +256,6 @@ func TestSignMalleatedTxs(t *testing.T) {
 	}
 }
 
-func TestWirePayForMessage_ValidateBasic(t *testing.T) {
```
liamsi, line 260: Moved to x/payment/types/wirepayformessage_test.go

Test there also:
 - is slightly simpler
 - not string based
 - ensures sdk errors / abci codes

x/payment/types/wirepayformessage.go
```diff
@@ -83,38 +81,42 @@ func (msg *MsgWirePayForMessage) ValidateBasic() error {
...
+			return ErrCommittedSquareSizeNotPowOf2.Wrapf("committed to square size: %d", commit.K)
 		}
 
 		calculatedCommit, err := CreateCommitment(commit.K, msg.GetMessageNameSpaceId(), msg.Message)
 		if err != nil {
-			return err
+			return ErrCalculateCommit.Wrap(err.Error())
 		}
 
-		if string(calculatedCommit) != string(commit.ShareCommitment) {
```
liamsi, line 116: Was there ever a reason to cast this to string first?

x/payment/types/errors.go
```diff
@@ -6,7 +6,12 @@ import (
...
-// x/payment module sentinel errors
 var (
-	ErrSample = sdkerrors.Register(ModuleName, 1100, "sample error")
+	ErrReservedNamespace              = sdkerrors.Register(ModuleName, 11110, "cannot use reserved namespace IDs")
+	ErrInvalidNamespaceLen            = sdkerrors.Register(ModuleName, 11111, "invalid namespace length")
+	ErrInvalidDataSize                = sdkerrors.Register(ModuleName, 11112, "data must be multiple of shareSize")
+	ErrDeclaredActualDataSizeMismatch = sdkerrors.Register(ModuleName, 11113, "declared data size does not match actual size")
+	ErrCommittedSquareSizeNotPowOf2   = sdkerrors.Register(ModuleName, 11114, "committed to invalid square size: must be power of two")
+	ErrCalculateCommit                = sdkerrors.Register(ModuleName, 11115, "unexpected error calculating commit for share")
+	ErrInvalidShareCommit             = sdkerrors.Register(ModuleName, 11116, "invalid commit for share")
```
liamsi, line 16: These could also have the same err code and could be collapsed into a single error instead. Different error codes might be better if the caller wants to handle the errors differently (e.g. add a different errors string instead as user output).

x/payment/types/wirepayformessage.go
evan-forbes: non other to compare it afaict

evan-forbes (approved):
this is a really good addition

adlerjohn (approved):

Issue #338, opened 2022-04-19
QGB replaying relaying valsets 


SweeXordious:
Currently, we're listening for new valsets and check if we have the necessary threshold for submitting them to the EVM chain. However, for the previous valsets, we're currently ignoring them.
We can change the design of relaying valsets to the following:
- Query the EVM chain for the last nonce that was submitted
- Then, start the work from there and check if we have enough signatures for the next nonce, then relay them.

This is because the relayer doesn't  need to relay the current valsets while also replaying the old ones because the new valsets will not be accepted by the contract.

PR #342, opened 2022-04-21
Add valset request by nonce support for QGB

SweeXordious:
Add valset request by nonce support for QGB.

Closes: https://github.com/celestiaorg/celestia-app/issues/343

PR #344, opened 2022-04-21
Replay mechanism for QGB relayer

SweeXordious:
This change will make the relayer start from the QGB contract last state and relay from there.

Closes: https://github.com/celestiaorg/celestia-app/issues/338 and https://github.com/celestiaorg/celestia-app/issues/345


SweeXordious:
Can be merged after: https://github.com/celestiaorg/celestia-app/pull/342

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db v0.6.6
 	golang.org/x/net v0.0.0-20211015210444-4f30a5c0130f // indirect
-	golang.org/x/sys v0.0.0-20220412211240-33da011f77ad // indirect
+	golang.org/x/sys v0.0.0-20220422013727-9388b58f7150 // indirect
```
SweeXordious, line 21: Should we keep this outside this PR?

evan-forbes (approved):
I think this makes sense. I have a question or two, but let's sync on what we want to do next sometime soon. Pre-approving until then

go.mod
evan-forbes: ideally, yes, but we certainly don't have to if it's just the result of `go mod tidy` call

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -0,0 +1,174 @@
...
+				// TODO: store the nonce in the state somewhere, so that we don't have
+				// to assume what the nonce is
+				nonce := uint64(height) / types.DataCommitmentWindow
+
+				dataCommitments <- ExtendedDataCommitment{
+					Commitment: dcResp.DataCommitment,
+					Start:      startHeight,
+					End:        endHeight,
+					Nonce:      nonce,
+				}
```
evan-forbes, line 166: without any delay here, I think we will quickly fill this channel

x/qgb/orchestrator/querier.go
```diff
@@ -0,0 +1,269 @@
...
+				"foundValsetConfirms",
+				fmt.Sprintf(
+					"total power %d number of confirms %d",
+					currThreshHold,
+					len(confirmsResp.Confirms),
+				),
+			)
+		}
+		// TODO: make the timeout configurable
+		time.Sleep(time.Second * 30)
```
evan-forbes, line 211: I don't think these timeouts make sense here anymore

x/qgb/orchestrator/querier.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+}
+
+func (q *querier) QueryTwoThirdsValsetConfirms(
+	ctx context.Context,
+	timeout time.Duration,
+	valset types.Valset,
```
evan-forbes, line 159: I also think that we can simply rely on the context as long as we document that it will run forever until configured to stop.

go.mod
adlerjohn: At the risk of being anal, `go mod tidy` should be done in a separate PR.

SweeXordious: will revert then

x/qgb/orchestrator/broadcaster.go
```diff
@@ -0,0 +1,70 @@
...
+
+	builder := bc.signer.NewTxBuilder()
+	// TODO make gas limit configurable
+	builder.SetGasLimit(9999999999999)
+	// TODO: update this api
+	// via https://github.com/celestiaorg/celestia-app/pull/187/commits/37f96d9af30011736a3e6048bbb35bad6f5b795c
+	tx, err := bc.signer.BuildSignedTx(builder, msg)
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
SweeXordious: I think this can be addressed in a separate PR as this  PR only focuses on replaying missing attestations to the QGB contract

x/qgb/orchestrator/broadcaster.go
```diff
@@ -0,0 +1,70 @@
...
+	rawTx, err := bc.signer.EncodeTx(tx)
+	if err != nil {
+		return "", err
+	}
+
+	// TODO  check if we can move this outside of the paytypes
+	resp, err := paytypes.BroadcastTx(ctx, bc.qgbGrpc, 1, rawTx)
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
SweeXordious: Nope. I just think that if it's used by both modules, then it's better to have it defined in a more general scope.

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: That depends on the data window and the block time, I assume. If the data window is 100 blocks and block time is 15sec, then it wouldn't fill that fast. If it is 2 blocks and block time of 5sec then yes it will,  and we will risk of losing data commitments. 
However, this only stands for the current design. But, when we will have data commitment requests, I guess it will be better.

x/qgb/orchestrator/querier.go
SweeXordious: yes, sure. I guess this can be handled when we have the worker pool design.

SweeXordious: And also would allow us to optimize a lot about the workflows we're calling for different queries

x/qgb/orchestrator/querier.go
SweeXordious: But wouldn't that be too much ? If we keep sending requests non stop checking if we meet the threshold ? 
I guess we should create issues for this.

x/qgb/orchestrator/broadcaster.go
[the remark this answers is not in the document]
SweeXordious: @evan-forbes Can you please create an issue for this as I am not sure how to use these options.
Thanks a lot

PR #351, opened 2022-04-25
WIP: Orchestrator replay

SweeXordious:


SweeXordious:
This one is ready for review also

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -103,8 +110,65 @@ func (oc *orchestratorClient) SubscribeValset(ctx context.Context) (<-chan types
...
+			continue
+		}
+
+		valsetsChan <- *correspondingVs
+	}
+}
+
 func (oc *orchestratorClient) SubscribeDataCommitment(ctx context.Context) (<-chan ExtendedDataCommitment, error) {
-	dataCommitments := make(chan ExtendedDataCommitment)
+	dataCommitments := make(chan ExtendedDataCommitment, 100)
```
evan-forbes, line 168: any particular reason for the 100 here other than that we want a buffered channel?

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -103,8 +110,65 @@ func (oc *orchestratorClient) SubscribeValset(ctx context.Context) (<-chan types
...
+		// The valset signed by the orchestrator to get lastVsConfirm
+		// Used to get the height that valset waas first introduced
+		correspondingVs, err := oc.querier.QueryValsetByNonce(ctx, previousNonce)
+		if err != nil {
+			oc.logger.Error(err.Error())
+			return err
+		}
+		if correspondingVs.Height < lastUnbondingHeight {
+			// Most likely, we're up to date and don't need to catchup anymore
+			oc.logger.Info("Finished adding Valsets attestation to queue")
```
evan-forbes, line 155: slight nit pick: can we use a defer for this since we're calling it more than once only when we return?

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -103,8 +110,65 @@ func (oc *orchestratorClient) SubscribeValset(ctx context.Context) (<-chan types
...
+			return nil
+		}
+		previousNonce = previousNonce - 1
+		lastVsConfirm, err := oc.querier.QueryValsetConfirm(ctx, previousNonce, oc.orchestratorAddress)
+		if err != nil {
+			oc.logger.Error(err.Error())
+			return err
+		}
+		// The valset signed by the orchestrator to get lastVsConfirm
+		// Used to get the height that valset waas first introduced
```
evan-forbes, line 147:
```suggestion
		// Used to get the height that valset was first introduced
```

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -164,11 +228,89 @@ func (oc *orchestratorClient) SubscribeDataCommitment(ctx context.Context) (<-ch
...
-
 	}()
 
 	return dataCommitments, nil
 }
+
+func (oc *orchestratorClient) addOldDCAttestations(
+	ctx context.Context,
+	dataCommitmentsChan chan ExtendedDataCommitment,
+) error {
```
evan-forbes, line 241: if we're not going to handle this error, and only log, then I don't think we should be returning any

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -103,8 +110,65 @@ func (oc *orchestratorClient) SubscribeValset(ctx context.Context) (<-chan types
...
+			// Most likely, we're up to date and don't need to catchup anymore
+			oc.logger.Info("Finished adding Valsets attestation to queue")
+			return nil
+		}
+		if lastVsConfirm != nil {
+			// in case we have holes in the signatures
+			continue
+		}
+
+		valsetsChan <- *correspondingVs
```
evan-forbes, line 163: even tho this is temporary, do you mind adding a brief comment that describes where this valset is going and what happens to it? Its always difficult to track channels, hopefully we can get rid of all but 2 channels using the "fan out" worker pool approach.

evan-forbes (commented):
having to resort to things like this is not ideal, and rather messy imho, but I understand why and we should be able to clean up most of this using the worker pool model later

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -3,13 +3,12 @@ package orchestrator
...
 	"context"
 	"fmt"
-	"sync"
-
 	"github.com/celestiaorg/celestia-app/x/qgb/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	tmlog "github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/rpc/client/http"
 	coretypes "github.com/tendermint/tendermint/types"
+	"sync"
```
adlerjohn, line 11: Why was this moved here? It looks like other files separate standard from cosmos/celestia dependencies with a newline (as did this file before this PR). We should maintain a consistent style here, ideally enforced by CI.

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -103,8 +110,65 @@ func (oc *orchestratorClient) SubscribeValset(ctx context.Context) (<-chan types
 	return valsetsChan, nil
 }
 
+func (oc *orchestratorClient) addOldVSAttestations(ctx context.Context, valsetsChan chan types.Valset) error {
```
adlerjohn, line 113:
```suggestion
func (oc *orchestratorClient) addOldValsetAttestations(ctx context.Context, valsetsChan chan types.Valset) error {
```

x/qgb/orchestrator/querier.go
```diff
@@ -221,6 +229,11 @@ func (q *querier) QueryLastValset(ctx context.Context) (types.Valset, error) {
 		return types.Valset{}, err
 	}
 
+	if len(lastValsetResp.Valsets) == 1 && lastValsetResp.Valsets[0].Nonce == 1 {
```
adlerjohn, line 232: Is there a case where `len(lastValsetResp.Valsets) == 1 && lastValsetResp.Valsets[0].Nonce != 1` is `true`?

x/qgb/orchestrator/relayer.go
```diff
@@ -86,6 +86,7 @@ func (r *relayer) updateValidatorSet(
 		return err
 	}
 
+	// this should probably be valset with nonce -1
```
adlerjohn, line 89: Isn't nonce an unsigned int?
```suggestion
	// TODO should this be valset with nonce -1?
```

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: Cool,  will do. This channel is the ordinary valsets channel, the one used above to contain valsets when listening for new ones.

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: I added it so that when replaying old attestations, and we're so far behind,  not to have a huuuuge channel.
Do you think we should remove it ?

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: I think we will handle this error with the new design. But I can remove it, I'm okey with both.
Should I ?

x/qgb/orchestrator/querier.go
SweeXordious: Yes, when we just start the chain and only have one valset.

adlerjohn: In that case `lastValsetResp.Valsets[0].Nonce` would be `1`?

x/qgb/orchestrator/relayer.go
adlerjohn: Add a TODO and the edit pls

x/qgb/orchestrator/querier.go
SweeXordious: yes.
This case is for the following scenario:
- You start a Celestia chain. It creates on block 0 `valset.Nonce=1`
- You deploy a QGB contract using that chain
- The QGB contract will now have `valset.Nonce=1` as `valset0` 
- Then, when you want to relay `valset.Nonce=1` to the QGB
- To do so, you need to get the signatures of `valset1` using validators from `valset1` (which is `valset0` in the QGB contract)

Maybe we can handle this case differently somewhere else. Or we can update the QGB contract to have `valset.Nonce = 1` as the init valset.

Live example on Rinkeby: https://rinkeby.etherscan.io/address/0xEAcC249c0831f8DE0B6afe27e6179Ca29Dec9b74#events
Check the first two events:
- First event is the deployment event. It has nonce 0
- Second event is relaying `valset.Nonce=1`  
You can see that the threshold and the validator set hash are similar.

Now I think of it, this probably needs to be documented somewhere. Let me know if we want to document this, or we want to update the QGB contract to start at valset 1.

SweeXordious: Aaah now I see it. You're right, it is redundant. It can be as follows
```suggestion
	if len(lastValsetResp.Valsets) == 1 {
```

SweeXordious: Also, I created the following issue to track this: https://github.com/celestiaorg/celestia-app/issues/373

adlerjohn (commented):
lgtm, but will defer to @evan-forbes

x/qgb/orchestrator/orchestrator_client.go
evan-forbes: no I think that makes sense, it shouldn't be that much of a burden, and we can optimize later if we really want to "go golfing"

x/qgb/orchestrator/orchestrator_client.go
evan-forbes: yeah, lets get rid of it for now. We'll probably need to change this function a bit then anyway

SweeXordious: done

Issue #294, opened 2022-04-12
Investigate simplifying `KeyringSigner` api 


evan-forbes:
per dicsussion in https://github.com/celestiaorg/celestia-app/pull/293#discussion_r848826246 we should be able to simplify the API for the `KeyringSigner` by unexporting `QueryAccount`

evan-forbes:
after thinking about this more, I'm not sure that we should do this exact change, as we need to sign the share commitments after we query for the account number. We should still take another look at simplifying the API though

cc: @renaynay

renaynay:
It would make my life a lot easier if the canonical implementation of constructing + signing a PFD existed in celestia-app @evan-forbes

renaynay:
I have the initial implementation here: 

https://github.com/celestiaorg/celestia-node/pull/619

evan-forbes:
We should be able to not use the keyring signer after implementing #951

evan-forbes:
can close with addition of the canonical mechanism for submitting a PFB and the user package

Issue #239, opened 2022-03-10
Default to signing over all relevant square sizes when creating a `WirePayForMessage`


evan-forbes:
Currently we require a user to select which square sizes they want to sign over. This is a hostile UX, as the user must have extraneous knowledge on the data square, malleated transactions, and the non-interactive default rules to know which square sizes to sign over. For most cases, the user will likely want to sign all relevant square sizes.

as pointed out in #236, this does not mean that they have to sign over all square sizes

> if a message can fit on a single row of square size x, then a valid commitment for square size x is valid for all square sizes >= x.

Also, they don't have to sign over square sizes that don't have enough shares to include their message + `MsgPayForMessage` transaction.

toanngosy:
Hi @evan-forbes, can I work on this?

evan-forbes:
certainly!

Issue #236, opened 2022-03-09
Optimize number of share commitments signed over when creating a `WirePayForMessage`


evan-forbes:
Currently, when block producers are preparing a proposal, they are requiring that the user include a commitment for that exact square size. 

https://github.com/celestiaorg/celestia-app/blob/0363f0410d9d6bf0e51ac92afcaa9be7c0d1ba07/x/payment/types/wirepayformessage.go#L190-L193

This does not account for the fact that commitments do not change if the largest power of two that is smaller than the message can fit on a single row. Meaning that if the largest power of 2 that fits in a message can fit on a single row of square size x, then that commitment is valid for all square sizes >= x.

To be more efficient, block producers should also compare the the number shares taken up by the message. If a commitment for a square size that can fit the entire message in a single row is included in the WirePayForMessage, then that commitment can be used for any square size equal to or greater than that size.

we could replace the above code segment with something along the lines of
```go
for _, commit := range msg.MessageShareCommitment {
	if commit.K == squareSize {
		shareCommit = &commit
	}
	if msgShareCount <= commit.K && commit.K <= squareSize {
		shareCommit = &commit
	}
```

This will allow for a user to be more confident that their message will be included in an upcoming block. Instead of creating a `WirePayForMessage` that includes signatures for all possible square sizes, if their message is smaller than the max square size, then they only have to the sign over commitments for square sizes that are smaller than the total number of shares that their message takes up.

With our current max square size, this wouldn't save the user from creating and including that many share commitments, but in the future if we increase the max square size significantly, then this could help reduce the size of `WirePayForMessage`s.

evan-forbes:
we only use a single commitment now since #917

PR #352, opened 2022-04-25
x/payment/pfd.go: Add canonical `PFD` constructor/signer/submittor methods

renaynay:
This PR adds 3 methods: 

* `BuildPayForData`
* `SignPayForData`
* `SubmitPayForData`

Celestia-node (and other repos depending on this message type) should import these methods from app.

Once this PR is merged, I will close https://github.com/celestiaorg/celestia-node/pull/619 and reopen a PR that updates the celestia-app dep to the latest celestia-app release and uses these methods under the hood.

### DIFF
--- a/x/payment/pfd.go
+++ b/x/payment/pfd.go
@@ -1,0 +1,97 @@
+package payment
+
+import (
+	"context"
+	"google.golang.org/grpc"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdk_tx "github.com/cosmos/cosmos-sdk/types/tx"
+	"github.com/cosmos/cosmos-sdk/x/auth/signing"
+
+	"github.com/celestiaorg/celestia-app/x/payment/types"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+var (
+	// shareSizes includes all the possible share sizes of the given data
+	// that the signer must sign over.
+	shareSizes = []uint64{16, 32, 64, 128}
+)
+
+// SubmitPayForData constructs, signs and synchronously submits a PayForData
+// transaction, returning a sdk.TxResponse upon submission.
+func SubmitPayForData(
+	ctx context.Context,
+	signer *types.KeyringSigner,
+	conn *grpc.ClientConn,
+	nID namespace.ID,
+	data []byte,
+	gasLim uint64,
+) (*sdk.TxResponse, error) {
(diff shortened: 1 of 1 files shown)


x/payment/pfd.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/celestiaorg/celestia-app/x/payment/types"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+var (
+	// shareSizes includes all the possible share sizes of the given data
+	// that the signer must sign over.
+	shareSizes = []uint64{16, 32, 64, 128}
+)
```
evan-forbes, line 19: would prefer not to hardcode these, but I also think its ok to merge as is, considering that we have #239 and #236

x/payment/pfd.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	txResp, err := types.BroadcastTx(ctx, conn, sdk_tx.BroadcastMode_BROADCAST_MODE_BLOCK, rawTx)
+	if err != nil {
+		return nil, err
+	}
+	return txResp.TxResponse, nil
+}
```
evan-forbes, line 51: finally, the prophesied good api for submitting a payForData has come :raised_hands:

evan-forbes (commented):
I think we should also comment on #294 to see if we can also reduce the exposed API after introducing these.

Only one suggested change, wish we would have though to move this to the app after the dalc or even sooner. Really great work @renaynay . I think this will save countless headaches and foot guns.

evan-forbes:
oh shoot, one last thing minor nit pick, I would prefer to name the file something else instead of the shorthand version, just to be consistent.

x/payment/pfd.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return wpfd, nil
+}
+
+// SignPayForData signs a PayForData transaction.
+func SignPayForData(
+	signer *types.KeyringSigner,
+	pfd *types.MsgWirePayForData,
+	gasLimOption types.TxBuilderOption,
```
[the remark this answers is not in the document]
renaynay: Okay can you check this commit to see how i've implemented it: https://github.com/celestiaorg/celestia-app/pull/352/commits/931fbd5aedb36ed7a39db709af526bd36bd34a0f

evan-forbes (approved):
:weight_lifting: :100:

Issue #319, opened 2022-04-15
The QGB relayer should be able to wait for the transactions to be confirmed before moving to the next one


SweeXordious:
Currently, the QGB relayes the new confirms to the QGB contract. However, it does not wait for them to be confirmed to move to the next transaction. This might lead to inconsistencies. 
To have a complete QGB MVP, we need this functionality.

PR #354, opened 2022-04-26
adds evm_client wait for transaction confirmation

SweeXordious:
Adds a simple waiting mechanism to wait for evm client transaction confirmation before sending the next transaction.
Closes https://github.com/celestiaorg/celestia-app/issues/319

### DIFF
--- a/x/qgb/orchestrator/evm_client.go
+++ b/x/qgb/orchestrator/evm_client.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
+	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
@@ -37,20 +39,20 @@ type evmClient struct {
 	logger     tmlog.Logger
 	wrapper    wrapper.QuantumGravityBridge
 	privateKey *ecdsa.PrivateKey
-	evmRpc     string
+	evmRPC     string
 }
 
 func NewEvmClient(
 	logger tmlog.Logger,
 	wrapper wrapper.QuantumGravityBridge,
 	privateKey *ecdsa.PrivateKey,
-	evmRpc string,
+	evmRPC string,
 ) EVMClient {
 	return &evmClient{
 		logger:     logger,
 		wrapper:    wrapper,
 		privateKey: privateKey,
-		evmRpc:     evmRpc,
+		evmRPC:     evmRPC,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


x/qgb/orchestrator/evm_client.go
```diff
@@ -86,7 +89,22 @@ func (ec *evmClient) UpdateValidatorSet(
 	if err != nil {
 		return err
 	}
-	ec.logger.Info("ValSetUpdate", tx.Hash().String())
+
+	// TODO put this in a separate function and listen for new EVM blocks instead of just sleeping
+	for i := 0; i < 60; i++ {
```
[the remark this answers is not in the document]
SweeXordious: It was 6 at the beginning to wait for a minute. However, I am testing against rinkeby/ropsten, and they take a lot of time to include transactions in blocks. So, I changed it to 10 minutes of wait time.
In the future, when we have the worker pool design, we will change this by listening to new blocks and waiting for a number of confirmations before moving to the next attestation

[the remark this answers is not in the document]
SweeXordious: This is a QGB contract in action: 
https://ropsten.etherscan.io/address/0x820a20eb3c8f7789557595874b89dd71fea11a0e
I am generating a new valset every 15 blocks.
Don't mind the failing transactions, as those are duplicates (will be fixed when we implement the new design)

x/qgb/orchestrator/evm_client.go
```diff
@@ -86,7 +89,22 @@ func (ec *evmClient) UpdateValidatorSet(
...
+		lastNonce, err := ec.StateLastValsetNonce(&bind.CallOpts{Context: ctx})
+		if err != nil {
+			return err
+		}
+		if lastNonce == nonce {
+			ec.logger.Info(fmt.Sprintf("relayed valset %d: %s", nonce, tx.Hash().String()))
+			return nil
+		}
+		time.Sleep(10 * time.Second)
+	}
```
evan-forbes, line 105: as stated in the todo, I like the idea of a separate evm client method that waits for a specific update or something similar

evan-forbes (approved):
makes sense for a temporary solution

:+1:

x/qgb/orchestrator/evm_client.go
evan-forbes: let's write that down in an issue if we haven't already tho

SweeXordious: We can create a generic method that subscribes to new blocks and checks if a transaction hash is included or not.
Could be something like:
```golang
func (ec *EvmClient) waitForConfirmations(txHash string, numberOfConfirmations int) error
```
I wrote some snippets, but it would take much time debugging and making it work. Especially that subscribing to events requires using a web socket connection and that needs to be added to the config which itself needs to be updated to reflect the different commands.
Thus, I decided to leave it for later and do this instead.

SweeXordious: issue: https://github.com/celestiaorg/celestia-app/issues/356

evan-forbes: :shipit:

PR #358, opened 2022-04-27
Use uint64 instead of int64 in QGB types

rach-id:
To avoid confusion between data types, this PR changes int64 to uint64 in the right places


evan-forbes (approved):
the thinking around using int64 is that that is what tendermint uses for height, but I'm fine with keeping everything unsigned

evan-forbes:
I could be wrong, but I think its arbitrary for our use case since we will have to convert to an evm friendly encoding of a `big.Int` any way

SweeXordious:
uint64 make more sense for me. I will stick with it for now and in the future, if needed, we can switch again.

Issue #353, opened 2022-04-26
Data commitment roots for empty blocks returning same result


SweeXordious:
As per the change introduced in https://github.com/celestiaorg/celestia-core/pull/625, we have data commitment being generated from the data roots of a set of blocks. 
The problem is that for empty blocks, we end up having the same the commitment over different ranges. This creates an issue when signing data commitments by the QGB orchestrator, as it finds the same commitment and thinks it is a duplicate signature.
One possible solution is to prepend the generated data commitment with the range.

SweeXordious:
We don't have that yet. Cool, will use `endBlock % window` as it's probably what we will be using when we have the data commitment requests

evan-forbes:
Wait, do we need to include the commitment in the key at all?

SweeXordious:
You're right. `(nonce, orchestrator_address)` should be enough as `nonce ~ commitment`.
I'll look into changing that. Thanks for your help.

SweeXordious:
If we use `(nonce, orchestrator_address)` as a key, we won't be able to  have custom commitment ranges (not consecutive as they are now). Is that alright ?
I guess using: `(commitment, orchestrator_address, nonce)` makes more sense, as it will allow having custom commitment ranges and would solve our current issue with the empty blocks ranges.
@evan-forbes What do you think ?

evan-forbes:
I think we should use it because solves our current issue, and we can optimize later. I don't think we should use something for a potential feature we might not need. 

I think we could even get away with custom ranges only using nonces, but again, there's no reason to build those atm.

SweeXordious:
> I think we could even get away with custom ranges only using nonces

How? I think we will have duplicates as `nonce = endBlock%window`. Thus, for example, if `window=10`, all commitments the following commitments will have the same nonce:
- 10 ⇾ 20
- 11 ⇾ 21
- ...
- 19 ⇾ 29
Now, that I am thinking, even `(commitment, orchestrator_address, nonce)` wouldn't solve the indexing issue if we have non-consecutive ranges.

evan-forbes:
> How?

Having overlapping (custom) commitment ranges will require significant changes to the contracts, orchestrator, and relayer, but presumably we would still use a nonce to force each commitment to be unique. Simply indexing via nonce could work, and would incur less indexing cost, as we could have some DataCommitmentRequest (or similar) that is already kept in the state. Any needed info for that data commitment could be looked up simply by using the nonce. 

[...]

PR #359, opened 2022-04-27
Data commitment indexing key update

SweeXordious:
Uses `(end_block, begin_block, validator_address)` to index data commitment confirms.
Closes: https://github.com/celestiaorg/celestia-app/issues/353


evan-forbes (approved):
nice, makes sense and follows what we discussed earlier in #353

Issue #364, opened 2022-04-28
Add QGB last valset before height query


SweeXordious:
To implement https://github.com/celestiaorg/celestia-app/issues/363, we will need to be able to query the last valset before the data commitment end_block.

PR #365, opened 2022-04-28
adds last valset request before height query

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/364


x/qgb/keeper/query_valset.go
```diff
@@ -41,3 +42,22 @@ func (k Keeper) ValsetRequestByNonce(
...
 }
+
+// LastValsetBeforeHeight queries the last valset request before height
+func (k Keeper) LastValsetBeforeHeight(
+	c context.Context,
+	req *types.QueryLastValsetBeforeHeightRequest) (*types.QueryLastValsetBeforeHeightResponse, error) {
+	valReq := k.GetValsets(sdk.UnwrapSDKContext(c))
+	for _, valset := range valReq {
+		if !k.HasValsetRequest(sdk.UnwrapSDKContext(c), valset.Nonce+1) ||
+			(valset.Height <= req.Height && k.GetValset(sdk.UnwrapSDKContext(c), valset.Nonce+1).Height >= req.Height) {
```
adlerjohn, line 53:
```suggestion
			(valset.Height < req.Height && k.GetValset(sdk.UnwrapSDKContext(c), valset.Nonce+1).Height >= req.Height) {
```

x/qgb/keeper/query_valset.go
```diff
@@ -41,3 +42,22 @@ func (k Keeper) ValsetRequestByNonce(
...
 	return nil, sdkerrors.Wrap(sdkerrors.ErrNotFound, "valset request nonce not found")
 }
+
+// LastValsetBeforeHeight queries the last valset request before height
+func (k Keeper) LastValsetBeforeHeight(
+	c context.Context,
+	req *types.QueryLastValsetBeforeHeightRequest) (*types.QueryLastValsetBeforeHeightResponse, error) {
+	valReq := k.GetValsets(sdk.UnwrapSDKContext(c))
+	for _, valset := range valReq {
+		if !k.HasValsetRequest(sdk.UnwrapSDKContext(c), valset.Nonce+1) ||
```
adlerjohn, line 52: What if there's only a single valset saved, with height >= `req.Height`? Then this condition would incorrectly be `true`.

x/qgb/keeper/query_valset.go
```diff
@@ -41,3 +42,22 @@ func (k Keeper) ValsetRequestByNonce(
 
 	return nil, sdkerrors.Wrap(sdkerrors.ErrNotFound, "valset request nonce not found")
 }
+
+// LastValsetBeforeHeight queries the last valset request before height
+func (k Keeper) LastValsetBeforeHeight(
+	c context.Context,
+	req *types.QueryLastValsetBeforeHeightRequest) (*types.QueryLastValsetBeforeHeightResponse, error) {
+	valReq := k.GetValsets(sdk.UnwrapSDKContext(c))
+	for _, valset := range valReq {
```
adlerjohn, line 51: This seems rather inefficient, having to iterate over all valsets ever.

x/qgb/keeper/query_valset.go
SweeXordious: good spot :+1:

x/qgb/keeper/query_valset.go
SweeXordious: AFAIU, it can't be. We will always have a valset at block 0 or 1

SweeXordious: This is the condition creating valsets:
```
(latestValset == nil) || (lastUnbondingHeight == uint64(ctx.BlockHeight())) || significantPowerDiff
```
A valset will be created:
- If we have no valset 
- We're an unbonding height
- There  was a significant power difference in the validator set

adlerjohn: Please document this invariant here then in a comment.

x/qgb/keeper/query_valset.go
adlerjohn: Hmm I guess that's true. As Satoshi would say,

> With computer systems typically selling with 2GB of RAM as of 2008, and Moore's Law predicting current growth of
1.2GB per year, storage should not be a problem even if the [validator sets] must be kept in
memory.

evan-forbes: We still have a fair amount of work to do for optimizing these things. As we've previously discussed, we could change the way we index so that we make better use of maps instead of relying heavily on iterators, but this will require a significant refactor and departure from the gravity module's code. It won't hurt to keep track of instances where we can improve, so I'll link #369

x/qgb/keeper/query_valset.go
SweeXordious: Please check now if the comment is good.

adlerjohn (approved):

Issue #366, opened 2022-04-28
Add data commitment query by exact range to QGB


SweeXordious:
Following the development under https://github.com/celestiaorg/celestia-app/issues/353, the data commitment confirms are being indexed using `(begin_block, end_block, orchestrator_address)`. 
This requires adding a new query that will take an exact range: `(begin_block, end_block)` and return a set of signed data commitments. This latter will be used by the relayer when querying for a threshold of confirmations.

PS: we currently have `QueryDataCommitmentsByRange`. However, the purpose for this one is to return all commitments for a certain range (which can be bigger than the data commitment window). In our case, we will need to query for the exact range: `QueryDataCommitmentsByExactRange`

PR #367, opened 2022-04-28
adds query data confirm by exact range

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/366


evan-forbes (approved):
definitely makes sense for now! :+1: LGTM

I created to #368 and #369 to remind us to take another look at the existing query functionality before we merge, as the functionality here might leave other queries unused.

SweeXordious:
Yes, 100% we will need to clean up the queries and also the keepers

Issue #363, opened 2022-04-28
Query data commitment confirmations specific to a certain valset


SweeXordious:
Currently, when the QGB relayer is looking for more than 2/3 of the confirmations, it is checking them against the current valset. However, if the QGB is so far behind, the valset might have changed after that data commitment, and we might end up with a halting bridge.
Thus, we will need to get the confirmations for that specific valset.

PR #372, opened 2022-04-29
Relayer query right valset for confirmations

SweeXordious:
Queries the right valsets/data commitment confirmations when relaying + final changes for a working QGB contract.
Closes https://github.com/celestiaorg/celestia-app/issues/363

QGB contract in action: https://rinkeby.etherscan.io/address/0xEAcC249c0831f8DE0B6afe27e6179Ca29Dec9b74

Still investigating why it reverted at data commitment nonce 6... 

Edit: for some reason, I restarted the relayer after an hour, and it just continued. Needs to be investigated. The only difference between the two byte codes is the signatures. So, probably we have an issue when querying for the signatures (issue tracking this https://github.com/celestiaorg/celestia-app/issues/374)


SweeXordious:
Can be merged  after https://github.com/celestiaorg/celestia-app/pull/351

SweeXordious:
Nevermind the ugly history, everything will be squashed.
It is because I was working on branches that we squashed into the QGB integration branch that had tons of commits.

evan-forbes (approved):
LGTM :+1:

Issue #379, opened 2022-05-04
Submitting a PayForData tx from the cli fails when using --gas-prices flag


evan-forbes:
The cli sub command fails when attempting to submit a pay for data transaction while using the --gas-prices flag. This can be recreated in the cli integration test and when using the latest version of master locally using the `single-node.sh` script located in contrib.

evan-forbes:
just to update on what I found today: 

I no longer think that this is caused by adding specific flags, as the same command will create, sign, and broadcast a transaction just fine if ran again. It's only the first time that the transaction is ran do we get the error above. 

The account number and chain-id that are being used as input to create a transaction are also what should be expected.

fwiw, this same error occurs with the #399 branch that updates the cosmos-sdk to v0.46.0

PR #380, opened 2022-05-04
Deduct fees for PayForData transactions 

### LINKED ISSUE #48: Fees: actually enforce deducting those
As far as I remember, currently no fees are deducted for PayForMessage. Is that right? Or do we actually make use of the default fee-machanism of the SDK?

evan-forbes:
## Description

This PR consumes gas for PayForData transactions. This should work, as it's a fairly standard operation, but it would be nice to have more integration testing. When attempting to add this to the cli integration test, I encountered #379. We might want to wait until that bug is fixed just to be sure.

closes #48

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -341,7 +341,6 @@ func New(
 
 	app.PaymentKeeper = *paymentmodulekeeper.NewKeeper(
 		appCodec,
-		app.BankKeeper,
 		keys[paymentmoduletypes.StoreKey],
 		keys[paymentmoduletypes.MemStoreKey],
 	)
--- a/x/payment/keeper/gas_test.go
+++ b/x/payment/keeper/gas_test.go
@@ -1,0 +1,27 @@
+package keeper
+
+import (
+	"testing"
+
+	"github.com/celestiaorg/celestia-app/x/payment/types"
+	"github.com/cosmos/cosmos-sdk/simapp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+func TestPayForDataGas(t *testing.T) {
+	app := simapp.Setup(false)
+	ctx := app.BaseApp.NewContext(false, tmproto.Header{})
+	k := Keeper{}
+
+	messageSize := uint64(100)
+
+	msg := types.MsgPayForData{
+		MessageSize: messageSize,
+	}
+
+	_, err := k.PayForData(sdk.WrapSDKContext(ctx), &msg)
+	require.NoError(t, err)
+	require.Equal(t, messageSize, ctx.GasMeter().GasConsumed())
+}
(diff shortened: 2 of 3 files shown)


app/app.go
```diff
@@ -341,7 +341,6 @@ func New(
 
 	app.PaymentKeeper = *paymentmodulekeeper.NewKeeper(
 		appCodec,
-		app.BankKeeper,
```
evan-forbes, line 344: This was prematurely added over a year ago to do fee burning. Since we won't be doing fee burning until we have some form of immediate execution, this is being removed

x/payment/keeper/gas_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	messageSize := uint64(100)
+
+	msg := types.MsgPayForData{
+		MessageSize: messageSize,
+	}
+
+	_, err := k.PayForData(sdk.WrapSDKContext(ctx), &msg)
+	require.NoError(t, err)
+	require.Equal(t, messageSize, ctx.GasMeter().GasConsumed())
```
liamsi, line 26: Isn't that nice? 🤩

x/payment/keeper/keeper.go
```diff
@@ -35,15 +35,12 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 //  MsgPayForData moves a user's coins to the module address and burns them.
 func (k Keeper) PayForData(goCtx context.Context, msg *types.MsgPayForData) (*types.MsgPayForDataResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	ctx.GasMeter().ConsumeGas(msg.MessageSize, payForDataGasDescriptor)
```
liamsi, line 39: Qs: 
- Can a block proposer still include larger blobs than the size that was committed to in the PFD? Is there a test somewhere for that?
- Can a block proposer include a blob without a PFD? (my understanding is no; just double checking)

liamsi (commented):
Looking good!

x/payment/keeper/keeper.go
adlerjohn: Isn't including larger blobs and including blob without PFD isomorphic?

adlerjohn (approved):
utACK

x/payment/keeper/keeper.go
evan-forbes: After we add the message inclusion check, it will not be possible without 2/3s of the voting power, as validators recompute the commitment for each message and compare them with those in the pfds

SweeXordious (approved):

x/payment/keeper/keeper.go
liamsi: > Isn't including larger blobs and including blob without PFD isomorphic?

Yeah, good point. 

> After we add the message inclusion check, it will not be possible without 2/3s of the voting power, as validators recompute the commitment for each message and compare them with those in the pfds

That is in line with the honest majority assumption for state. So I think that is good enough.

Issue #302, opened 2022-04-14
Update the CLI flags to correctly reflect the parameters needed for QGB


SweeXordious:
Currently, we are passing the same configuration struct for all the QGB commands. For example, we are allowing the `--contract-address` to be passed when deploying a new QGB contract, and it is not used.
Thus, we will need to create separate structs to elegantly reflect the parameters needed for each command.

PR #402, opened 2022-05-12
better structure QGB CLI flags

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/302

### DIFF
--- a/x/qgb/orchestrator/config.go
+++ b/x/qgb/orchestrator/config.go
@@ -28,10 +28,10 @@ func init() {
 
 const (
 	// cosmos-sdk keyring flags
-	keyringBackendFlag = "keyring-backend"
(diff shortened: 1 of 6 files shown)


x/qgb/orchestrator/config.go
```diff
@@ -28,10 +28,10 @@ func init() {
```
SweeXordious, line 28: @evan-forbes  Parsing the flags in this file is a bit repetitive. However, I couldn't find a way to better structure them.
Please let me know if you have an idea on how to do it better.

evan-forbes (approved):
I think this makes sense. There are worse ways of organizing it, and while being brutally simple might not be able to serve our needs in the future, it certainly does now, along with sticking to the "pattern" that is already there!

Issue #398, opened 2022-05-11
QGB E2E test proposal


SweeXordious:
## Proposal
We will create a docker-compose network containing the following:
- 4 Celestia-app nodes that can be validators
- 4 Orchestrator nodes that will each run aside of a celestia-app
- 1 Ethereum node. Probably Ganache as it is easier to set up
- 1 Relayer node that will listen to Celestia chain and relay attestations

The validator nodes need to be able to join the network on a per-need basis. The reason for this is to be able to test valset changes thoroughly.
For example, the network starts with only one validator.  A Valset is created and a series of Data commitments will be orchestrated. Then, after a few blocks, a second validator needs to join the network and start validating blocks. Now, the QGB should be able to detect this change and update the valset accordingly, all while relaying attestations and testing against them.
Thus, we will need a deployment that would allow validators to join sequentially.

### Stretch goal
The integration tests should also test for the hooks defined under `x/qgb/keeper/hooks.go` in a real network environment.

Issue #325, opened 2022-03-08
Implement a dockerized deployment of a celestia testnet bridged to an ethereuem testnet by the qgb


evan-forbes:
In order to help test the qgb, we should create a docker compose script that will initiate a celestia and ethereuem testnet, bridged together by the qgb.

PR #408, opened 2022-05-13
Qgb integration test environment

SweeXordious:
Closing https://github.com/celestiaorg/celestia-app/issues/398 https://github.com/celestiaorg/celestia-app/issues/325


SweeXordious:
depends on https://github.com/celestiaorg/celestia-app/pull/415

x/qgb/keeper/query_data_commitment.go
```diff
@@ -8,6 +8,7 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+// FIXME should we return an error if it doesn't exist? for this one and the others?
```
SweeXordious, line 11: Question :D

SweeXordious:
This is  ready for review

e2e/qgb/celestia-app/deploy_qgb_contract.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# This script deploys the QGB contract and outputs the address to stdout.
```
evan-forbes, line 3: nice, I can see this script being useful even outside of this test

x/qgb/keeper/query_data_commitment.go
evan-forbes: which thing doesn't exist?

evan-forbes (commented):
wow, this is so cleeaann, will try test later today

SweeXordious:
@evan-forbes If  you wanna test it locally, try this one https://github.com/celestiaorg/celestia-app/pull/423 as this PR only contains the environment, not the actual test scenarios

x/qgb/keeper/query_data_commitment.go
SweeXordious: The element we're looking for. For example,  if the data commitment doesn't exist, it will just return `nil` and not throw an exception...

SweeXordious:
Will replace the:
```
        if os.Getenv("QGB_INTEGRATION_TEST") != TRUE {
		t.Skip("Skipping QGB integration tests")
	}

	network, err := NewQGBNetwork()
	HandleNetworkError(t, network, err, false)

	// preferably, run this also when ctrl+c
	defer network.DeleteAll() //nolint:errcheck
```
with an `IntegrationTestSuite` as of https://github.com/celestiaorg/celestia-app/issues/428 in a subsequent PR

x/qgb/keeper/query_data_commitment.go
evan-forbes: ahh, I see. Intuitively it makes sense to throw and error, but I'm not 110% confident in that decision.

e2e/qgb/qgb_network.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+}
+
+func (network QGBNetwork) GetLatestDeployedQGBContract(ctx context.Context) (*wrapper.QuantumGravityBridge, error) {
+	return network.GetLatestDeployedQGBContractWithCustomTimeout(ctx, 5*time.Minute)
+}
+
+func (network QGBNetwork) GetLatestDeployedQGBContractWithCustomTimeout(
+	ctx context.Context,
+	timeout time.Duration,
```
evan-forbes, line 327: I definitely don't think we should change this in this PR, as its not even a nit pick just a different option, but we could simply rely on this context to cancel instead of a timeout channel. :shrug:

x/qgb/keeper/query_data_commitment.go
SweeXordious: Will create an issue to think about it later

evan-forbes (approved):
this is about as minimal as a setup as I could imagine for a dockerized e2e test. Well done!

utACK

e2e/qgb/qgb_network.go
evan-forbes: I might have already commented about this, if so pls ignore

x/qgb/keeper/query_data_commitment.go
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/437

e2e/qgb/qgb_network.go
SweeXordious: Yes, we can actually rely on both, the context and the timeout, I am planning to implement it later as the current way this is done gives some issues when ctrl+c a running test... Will create an issue

PR #415, opened 2022-05-17
Create repository official Dockerfile

SweeXordious:
This PR simplifies the main Dockerfile for Celestia-app.
It creates an image containing the Celestia-app binary and doesn't take into account  the other development related to it. For example, under `docker/` directory, we had a `Dockerfile` that was shaped to suite the ephemeral cluster needs. However, when tagged using GitHub actions, it was acting as the main Celestia-app docker image. Thus, if someone wants to use that image, they will have to follow a complex way of specifying the home directory and existing files. In the meanwhile, the official repository Dockerfile shouldn't be as opinionated as possible.

Also, I renamed `docker/Dockerfile` to `docker/Dockerfile_ephemeral`  in case we need to keep that image for the ephemeral cluster. However, I suggest we add a new GitHub action that creates that image and tags it with a tag that reflect that it will be used for the ephemeral cluster, example : `ghcr.io/celestiaorg/celestia-app-ephemeral:latest` (Let me know if we want to add that action so that I add it to this PR).

This comes from the need of having a base Celestia-app docker image that will be built during integration tests.


SweeXordious:
The issue with this approach is `priv_validator_state.json`: 
```
ERR CONSENSUS FAILURE!!! err="rename /opt/data/write-file-atomic-08895712196458512116 /opt/data/priv_validator_state.json: device or resource busy" module=consensus s
tack="goroutine 216 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:24 +0x65\ngithub.com/tendermint/tendermint/consensus.(*State).receiveRoutine.func2()\n\t/go/
pkg/mod/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:726 +0x4c\npanic({0x1978200, 0xc0022989c0})\n\t/usr/local/go/src/runtime/panic.go:1038 +0x215\ngithub.com/t
endermint/tendermint/privval.(*FilePVLastSignState).Save(0x78)\n\t/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/privval/file.go:139 +0x90\ngithub.com/tendermint/tenderm
int/privval.(*FilePV).saveSigned(0xc0000fc600, 0xc0027bb700, 0x4106b4, 0x0, {0xc0027c0280, 0x80, 0x203000}, {0xc0022ac700, 0x40, 0x40})\n\t/go/pkg/mod/github.com/celestiaorg/celestia-core@v1
.0.1-tm-v0.34.16/privval/file.go:395 +0x88\ngithub.com/tendermint/tendermint/privval.(*FilePV).signProposal(0xc008daec80, {0xc00906ded7, 0x7}, 0xc0000fc680)\n\t/go/pkg/mod/github.com/celesti
aorg/celestia-core@v1.0.1-tm-v0.34.16/privval/file.go:381 +0x2ba\ngithub.com/tendermint/tendermint/privval.(*FilePV).SignProposal(0xc0000fc600, {0xc00906ded7, 0x0}, 0x0)\n\t/go/pkg/mod/githu
b.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/privval/file.go:265 +0x1e\ngithub.com/tendermint/tendermint/consensus.(*State).defaultDecideProposal(0xc009272700, 0x1, 0x0)\n\t/go/pkg/mod
/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:1132 +0x2e2\ngithub.com/tendermint/tendermint/consensus.(*State).enterPropose(0xc009272700, 0x1, 0x0)\n\t/go/pkg/m
od/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:1096 +0x6a3\ngithub.com/tendermint/tendermint/consensus.(*State).enterNewRound(0xc009272700, 0x1, 0x0)\n\t/go/pk
g/mod/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:1019 +0xa67\ngithub.com/tendermint/tendermint/consensus.(*State).handleTimeout(0xc009272700, {0xc000d923f0, 0
xc000d923f0, 0xd923f0, 0xc0}, {0x1, 0x0, 0x1, {0x181e0134, 0xeda15924b, ...}, ...})\n\t/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:892 +0x4a5\ngith
ub.com/tendermint/tendermint/consensus.(*State).receiveRoutine(0xc009272700, 0x0)\n\t/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:791 +0x6bf\ncreate
d by github.com/tendermint/tendermint/consensus.(*State).OnStart\n\t/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.0.1-tm-v0.34.16/consensus/state.go:378 +0x12f\n"
```
Even when mounted with the correct permissions, it still panics. Thus, forcing users to either:
- Run a custom script when starting the container. This means we should change `entrypoint` to `cmd`
- Or, create a new image which creates that file somewhere.

Will keep on investigating this.

Bidon15:
Noice. I think creating an action for epheremal cluster makes sense and will be cool if this can be done within this PR
Afterwards, we can edit the path in ephemeral-cluster repo to reflect latest changes made in this PR

Passing to @jbowen93 if he is ok with this approach

Bidon15 (dismissed):
Final say to @jbowen93

SweeXordious:
@Bidon15 Awesome. Will add the GitHub action for ephemeral cluster if Josh agrees to these changes and think they make sense.

docker/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# This script creates the necessary files before starting Celestia-appd
+
+# only create the priv_validator_state.json if it doesn't exist and the command is start
+if [[ $3 == "start" && ! -f $CELESTIA_HOME/data/priv_validator_state.json ]]
```
adlerjohn, line 6: Maybe nit, but there's no guarantee that `$CELESTIA_HOME` doesn't contain spaces or special characters. You can surround it in quotes to avoid things breaking if it does.

SweeXordious: good catch :+1:

adlerjohn:
Is this ready for review?

SweeXordious:
@adlerjohn Waiting for Josh input to see whether to add a github action to this PR for the ephemeral cluster docker image or not.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+COPY --from=builder /celestia-app/build/celestia-appd /bin/celestia-appd
+COPY  docker/entrypoint.sh /opt/entrypoint.sh
+
+# p2p, rpc and prometheus port
+EXPOSE 26656 26657 1317 9090
+
+ENV CELESTIA_HOME /opt
+
+ENTRYPOINT [ "/bin/bash", "/opt/entrypoint.sh" ]
```
jbowen93, line 20: I'd prefer
```
ENTRYPOINT ["/opt/entrypoint.sh"]
CMD ["/bin/celestia-appd"]
```
and the suggested changes in entrypoint.sh

docker/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+    cat <<EOF > ${CELESTIA_HOME}/data/priv_validator_state.json
+{
+  "height": "0",
+  "round": 0,
+  "step": 0
+}
+EOF
+fi
+
+/bin/celestia-appd --home ${CELESTIA_HOME} $@
```
jbowen93, line 18: I'd prefer something like this 
```
if [[ $1 == "celestia-appd" && $2 == "start" && ! -f ${CELESTIA_HOME}/data/priv_validator_state.json ]]. then
    # init
    exec "$@" "--"
elif [ "$1" = 'celestia-appd' ]; then
    exec "$@" "--"
fi
exec "$@"
```
alongside the suggested Dockerfile changes. 

see: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint

This results in a default command of `/bin/celestia-appd`and correctly captures flags. 

See [example usage by ephemeral cluster for ethermint](https://github.com/celestiaorg/ephemeral-cluster/blob/main/docker/ethermint/ethermint-docker-compose.yml#L7)

jbowen93 (changes requested):
Apologies for the delay in review. Left some comments. I'm fine adding a gh action for ephemeral to a later PR.

Dockerfile
SweeXordious: This would make the entrypoint to the docker image be bash. Do we want that ? I guess this image should has the entrypoint defined as celestia-appd. And maybe add a command `--help` after that for a default behaviour (even if it's not needed as by default, celestia-appd without arguments will return the help, it is more elegant, except that it adds an extra layer on top).
If I understand right, with this approach, we will be calling something like:
```ssh
$ docker run celestia-app celestia-appd <args>
```
which is redundant, opposed to the proposed approach:
```ssh
$ docker run celestia-app <args>
```
This is more like an executable container rather than a general purpose one that spawns a shell by default.
Example, geth are doing the same providing their binary as the entrypoint https://github.com/ethereum/go-ethereum/blob/master/Dockerfile.

Let me know what you think

SweeXordious: For the provided link https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint
As in the example, postgres uses an entrypoint script to do some initializations before  executing the command while allowing the users to run whatever they want. This comes from the fact that postgres has multiple binaries inside the image: `pg_virtualenv`, `pg_createcluster`, `pg_dropcluster` and many others.  Thus, allowing users to execute anything is a must.
In our case, it is only a single binary that we want to use in all cases...

jbowen93: I sometimes want to `docker exec -it <container-name> /bin/bash` into the box and poke around. But I'm not strongly opposed to have a binary style docker image if it simplifies things for your use case.

jbowen93 (approved):

jbowen93:
Resolved comments and approved

PR #418, opened 2022-05-18
x/payment: Sanity check `nID` length in `BuildPayForData`

renaynay:
This PR implements a sanity check on the length of the given namespace ID inside of the `BuildPayForData` method such that if `len(nID) != types.NamespaceIDSize`, it will fail out.


tzdybal (approved):
LGTM

x/payment/payfordata.go
```diff
@@ -61,6 +62,10 @@ func BuildPayForData(
 	message []byte,
 	gasLim uint64,
 ) (*types.MsgWirePayForData, error) {
+	// sanity check nID size
+	if len(nID) != types.NamespaceIDSize {
+		return nil, fmt.Errorf("expected namespace ID of size %d, got %d", types.NamespaceIDSize, len(nID))
+	}
```
evan-forbes, line 68: this is a good check, but do you think it makes sense to add it to the `NewWirePayForData` function? Perhaps that would be less explicit while reading code, but then it performs the same check each time we create a wirePFD.

better yet, we could just call `ValidateBasic` before returning from `NewWirePayForData`, which performs this check along with a few others. This function has to pass if the wirePFD will make it into the mempool, so its probably good to run ahead of time anywho.

evan-forbes (commented):
left an idea for a more thorough check performed each time we create a wirePFD, but would approve this PR as is if we think otherwise

x/payment/payfordata.go
renaynay: https://github.com/celestiaorg/celestia-app/pull/418/commits/9ed560c46e791aec558ef66d4e44610b4dc3f2b1

adlerjohn (commented):
CI failing

evan-forbes:
we can just do the nid check in `NewWirePayForData` for now lol

evan-forbes (approved):

adlerjohn (approved):

Issue #410, opened 2022-05-16
Integration test for max block size


evan-forbes:
Our current integration test is setup to test for singular cli interactions. We need an integration test that allows us to submit many transactions so that we can properly test our `PrepareProposal` method. Specifically, we want to test

- reaping the maximum transactions from the mempool
- ensuring that blocks are not proposed or processed that are over not within the max and min square sizes
- test transaction prioritization

related to #400

Issue #400, opened 2022-05-12
Test with a larger message in the integration test


evan-forbes:
We should test with varying sizes of message in the integration test. 

This is currently blocked by #379

Issue #389, opened 2022-05-06
Ensure fees were deducted when submitting a PayForData in the cli integration test


evan-forbes:
With #380, we started deducting fees based on the size of the message submitted via `MsgWirePayForData`. While we have unit tests to ensure that funds were actually deducted, we should also add the [cli integration test](https://github.com/celestiaorg/celestia-app/blob/b2617e419280061e33feaa12ff4dcdd9cc612113/x/payment/client/testutil/integration_test.go#L152-L158) by querying for the balance of the user after the tx has been successfully submitted.

**blocked** by #379

evan-forbes:
this can be closed as we're testing for this in the block size integration test

rootulp:
Where are we testing for gas consumption in? https://github.com/celestiaorg/celestia-app/blob/8b9c4c9d13fe0ccb6ea936cc26dee3f52b6f6129/app/test/block_size_test.go#L83

evan-forbes:
:eyes: oh shoot! we aren't! I could swear we were. opened #998 as a simple test

Issue #333, opened 2022-04-15
Test padding for `WirePayForData`s


evan-forbes:
We don't have a unit test that ensure that a `WirePayForData` are padded correctly so that no additional padding needs to be added while preparing the proposal. 

This is related to non-interactive defaults, and using the same code to generate share commitments as core

PR #419, opened 2022-05-19
Integration test for max block size

evan-forbes:
## Description

This PR fixes a quite a few bugs in order to get a new integration test to pass consistently. The test involves 2 validators chewing through 1) 20 ~1Mb PFDs and 2) 100 randomly sized PFDs, along with performing a few simple check on the size of the square for those blocks. This integration test also serves as a great starting point for future tests.

Notably, it also removes the user pre-padding their PFDs by default, which has some significant trade offs that still need to be discussed, likely in a different issue. 

closes: #410, #333, #389, #400, #379


go.mod
```diff
@@ -158,5 +158,5 @@ replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.1.0-sdk-v0.46.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint v0.35.4 => github.com/celestiaorg/celestia-core v1.2.1-tm-v0.35.4
+	github.com/tendermint/tendermint v0.35.4 => github.com/celestiaorg/celestia-core v1.2.1-tm-v0.35.4.0.20220518201820-1dc64a37f296
```
evan-forbes, line 161: still need to replace this with an actual release after #754

x/payment/client/testutil/integration_test.go
```diff
@@ -82,20 +82,7 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForData() {
...
-		{
-			"valid transaction list of square sizes",
-			[]string{
-				hexNS,
-				hexMsg,
-				fmt.Sprintf("--from=%s", username),
-				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
-				fmt.Sprintf("--%s=%s", flags.FlagFees, sdk.NewCoins(sdk.NewCoin(s.cfg.BondDenom, sdk.NewInt(2))).String()),
-				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
-				fmt.Sprintf("--%s=%s", paycli.FlagSquareSizes, "2,4,8,16,32"),
```
evan-forbes, line 98: this test was essentially the same as the other test, so it was deleted.

x/payment/client/testutil/integration_test.go
```diff
@@ -124,39 +111,43 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForData() {
 			out, err := clitestutil.ExecTestCLICmd(clientCtx, cmd, tc.args)
 			if tc.expectErr {
 				require.Error(err)
-			} else {
```
evan-forbes, line 127: this else statement was superfluous

x/payment/types/builder.go
```diff
@@ -112,14 +137,15 @@ func (k *KeyringSigner) BuildSignedTx(builder sdkclient.TxBuilder, msg sdktypes.
...
+
 	// Generate the bytes to be signed.
 	bytesToSign, err := k.encCfg.TxConfig.SignModeHandler().GetSignBytes(
 		signing.SignMode_SIGN_MODE_DIRECT,
-		authsigning.SignerData{
-			ChainID:       k.chainID,
-			AccountNumber: accountNumber,
-			Sequence:      sequence,
-		},
+		signerData,
```
evan-forbes, line 148: this was another bug where we weren't adding the public key and address to the `SignerData`, so the sign bytes were sometimes being generated incorrectly.

evan-forbes (commented):
I'm keeping this PR as a draft for now, as we have yet to merge #754, but it is ready for review.

running the race detector with the integration test that runs 2 validator nodes and 100Mb worth of data causes the test to be super super slow, and seems to timeout no matter how long the timeout is set to, so I'm not sure what we want to do about that. It's probably okay to either not run that test while also running the race detector, as we can still run it during the cli integration test, which is significantly smaller, and should pick up any low handing data races.

evan-forbes:
Marking ready for review while I figure out what to do about the race detector inhibiting the big block integration test to the point of failure

app/test/block_size_test.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func NewIntegrationTestSuite(cfg cosmosnet.Config) *IntegrationTestSuite {
+	return &IntegrationTestSuite{cfg: cfg}
+}
+
+func (s *IntegrationTestSuite) SetupSuite() {
+	s.T().Log("setting up integration test suite")
+
+	if testing.Short() {
```
SweeXordious, line 45: Is using this better than checking an environment variable?

app/test/block_size_test.go
```diff
@@ -0,0 +1,290 @@
...
+		})
+		require.NoError(s.network.WaitForNextBlock())
+	}
+
+}
+
+func TestIntegrationTestSuite(t *testing.T) {
+	cfg := network.DefaultConfig()
+	cfg.EnableTMLogging = false
+	cfg.MinGasPrices = "0uceles"
```
SweeXordious, line 165: shouldn't this follow the new denomination ?

x/payment/types/builder.go
```diff
@@ -64,17 +65,41 @@ func (k *KeyringSigner) QueryAccountNumber(ctx context.Context, conn *grpc.Clien
 	return nil
 }
 
+func (k *KeyringSigner) UpdateAccountFromClient(clientCtx client.Context) error {
+	// clientCtx.AccountRetriever.
```
SweeXordious, line 69: signaling in case you forgot to  delete this comment :D

SweeXordious (commented):
I like the `IntegrationTestSuite` way :+1:

go.mod
adlerjohn: 754 doesn't exist in this repo

adlerjohn (commented):
CI failing?

evan-forbes:
> CI failing?

yeah, still the same issue we were experiencing before and with other large tests in other repos. Running the race detector on large tests just doesn't work. If we use a single validator and small blocks, it does work (see the cli integration test), but if we use large blocks or two validators it doesn't.

In the past, using a build tag and simply not running the test if the race detector was running, as the build tags are visible at runtime. see https://github.com/golang/go/issues/7007

 I haven't found a solution that I'm happy with atm, we might have to go with the very ugly solution we used before in core

adlerjohn (commented):
no comments

app/prepare_proposal.go
```diff
@@ -68,9 +68,8 @@ func (app *App) estimateSquareSize(data *core.Data) uint64 {
 	msgShareEstimate := estimateMsgShares(app.txConfig, data.Txs)
 
 	totalShareEstimate := txShareEstimate + evdShareEstimate + msgShareEstimate
-
-	estimatedSize := types.NextPowerOf2(uint64(math.Sqrt(float64(totalShareEstimate))))
-
+	sr := math.Sqrt(float64(totalShareEstimate))
+	estimatedSize := types.NextHighestPowerOf2(uint64(sr))
```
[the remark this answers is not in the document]
liamsi: Maybe we should add a comment here too.

x/payment/client/testutil/integration_test.go
```diff
@@ -40,10 +40,6 @@ func NewIntegrationTestSuite(cfg cosmosnet.Config) *IntegrationTestSuite {
 func (s *IntegrationTestSuite) SetupSuite() {
 	s.T().Log("setting up integration test suite")
 
-	if testing.Short() {
-		s.T().Skip("skipping test in unit-tests mode.")
-	}
-
```
liamsi, line 46: Isn't skipping somewhere still necessary?

app/split_shares.go
```diff
@@ -23,10 +23,9 @@ import (
 // block data are only used to avoid dereferening, not because we need the block
 // data to be mutable.
 func SplitShares(txConf client.TxConfig, squareSize uint64, data *core.Data) ([][]byte, *core.Data) {
-	var (
-		processedTxs [][]byte
-		messages     core.Messages
-	)
+	processedTxs := make([][]byte, 0)
+	messages := core.Messages{}
```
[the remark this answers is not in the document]
liamsi: add comment in code

x/payment/types/builder.go
evan-forbes: thanks!

x/payment/types/payfordata.go
```diff
@@ -112,16 +114,24 @@ func BuildPayForDataTxFromWireTx(
 // https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#message-layout-rationale
 func CreateCommitment(k uint64, namespace, message []byte) ([]byte, error) {
 	// add padding to the message if necessary
-	message = padMessage(message)
```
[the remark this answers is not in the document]
liamsi: link to issue?

app/test/block_size_test.go
evan-forbes: good find! https://github.com/celestiaorg/celestia-app/pull/419/commits/ec45ceb29983e2472a773908f3162d2b1ac3815e

x/payment/client/testutil/integration_test.go
evan-forbes: not for this test, as its small enough for the detector to handle

liamsi (dismissed):
Approving. Any outstanding comments should be addressed in followup issues/PRs.

app/test/block_size_test.go
evan-forbes: imo, yes, as its builtin. Open to other solutions tho

app/split_shares.go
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/419/commits/8928fccb4c60537c56eeff697ba94ff0cc33066c

sorry, didn't see that you already approved before pushing

x/payment/types/payfordata.go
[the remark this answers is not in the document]
liamsi: Nope just as you did

liamsi (approved):

PR #427, opened 2022-05-20
Register the legacy amino codec

evan-forbes:
## Description

We weren't registering the legacy amino codec, which stops us from being able to unmarshal some things from tendermint in some sub commands of our cli.

closes: #XXXX


Bidon15 (approved):
🚢

SweeXordious (approved):

Issue #245, opened 2022-03-14
Test for a specific size instead of multiple of the share size 


evan-forbes:
in the current [cli integration tes](https://github.com/celestiaorg/celestia-app/blob/master/x/payment/client/testutil/integration_test.go)t, we are checking to see that emitted events contain a multiple of the share size instead of checking for a direct size. We should update that test to check for a specific size.

evan-forbes:
can close as this was completed https://github.com/celestiaorg/celestia-app/blob/6b86e91eea1063c27c1ae461eddbe505f778df08/x/payment/client/testutil/integration_test.go#L118

PR #429, opened 2022-05-21
Updated CLI integration test to check for specific size instead of multiple of the share size

nicoelzer:
## Description

Update CLI integration test to require specific size instead of multiple of the share size

closes: #245

### DIFF
--- a/x/payment/client/testutil/integration_test.go
+++ b/x/payment/client/testutil/integration_test.go
@@ -141,7 +141,7 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForData() {
 						signer := e.GetAttributes()[0].GetValue()
 						_, err = sdk.AccAddressFromBech32(signer)
 						require.NoError(err)
-						msgSize, err := strconv.ParseUint(e.GetAttributes()[1].GetValue(), 10, 64)
+						msgSize, err := strconv.ParseInt(e.GetAttributes()[1].GetValue(), 10, 64)
 						require.NoError(err)
 						msg, err := hex.DecodeString(tc.args[1])
 						require.NoError(err)
--- a/x/payment/client/testutil/integration_test.go
+++ b/x/payment/client/testutil/integration_test.go
@@ -1,6 +1,7 @@
 package testutil
 
 import (
+	"encoding/hex"
 	"fmt"
 	"strconv"
 	"testing"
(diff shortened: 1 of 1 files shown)


evan-forbes (changes requested):
Thanks for digging into this @nicoelzer ! pls see comment and let me know if you have any further questions.

its also worth noting that the way we are adding the default padding for messages is sort of in flux, as we are removing it in #419, and then will be adding a more proper implementation in #431 #432

x/payment/client/testutil/integration_test.go
```diff
@@ -142,7 +142,7 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForData() {
 						require.NoError(err)
 						msgSize, err := strconv.ParseUint(e.GetAttributes()[1].GetValue(), 10, 64)
 						require.NoError(err)
-						s.Equal(uint64(0), msgSize%consts.ShareSize, "Message length should be multiples of const.ShareSize=%v", consts.ShareSize)
+						require.Equal(uint64(consts.ShareSize), msgSize, "Message length should be %v", consts.ShareSize)
```
evan-forbes, line 145: while this specific message is the size of the `consts.ShareSize`, not all messages will be this size. The only reason that it is this size is because it gets padded. Ideally, we would have a function that calculated the padded length of the message being used in the test, and then use that to compare with the result here. I think it would also be okay to simply add a field in the test case to communicate the expected size.

nicoelzer: Hey @evan-forbes, thanks for the review, appreciate it!

I added the padMessage function, which adds the padding to the provided message and uses the result for comparison on the test.

evan-forbes:
hey @nicoelzer, sorry for all the delays, we've been super busy with mamaki! in this rush, we since changed this in master https://github.com/celestiaorg/celestia-app/blob/1c357cb4f7211b3e103ba2d45ee7d76e5d79f7cc/x/payment/client/testutil/integration_test.go#L116-L119

as part of a solution to some bugs in #419. This PR was still very helpful in getting us to think about #431 #432, and I hate to toss perfectly good work, but this change isn't needed any longer since our implementation changed. Again, super sorry that this contribution isn't being recognized fully by accepting this PR, it was made precisely at the time where we were already in the process of changing it. :slightly_frowning_face: 

it is good work though :slightly_smiling_face:

nicoelzer:
Hey @evan-forbes,

no worries at all and thank you for the feedback. I will pick up some other open issues and looking forward to further contribute! And of course, congrats on the launch of the testnet!!

PR #441, opened 2022-05-24
Sign over all applicable square sizes by default

evan-forbes:
## Description

this could still use some cleaning up, but this PR changes things so that we sign over all applicable square sizes when creating a wirePFD

cc @renaynay 

closes: #239

### DIFF
--- a/x/payment/types/square_sizes.go
+++ b/x/payment/types/square_sizes.go
@@ -1,0 +1,54 @@
+package types
+
+import (
+	"github.com/tendermint/tendermint/pkg/consts"
+)
+
+// https://github.com/celestiaorg/celestia-app/issues/236
+// https://github.com/celestiaorg/celestia-app/issues/239
+
+var allSquareSizes = generateAllSquareSizes()
+
+// generateAllSquareSizes generates and returns all of the possible square sizes
+// using the maximum and minimum square sizes
+func generateAllSquareSizes() []int {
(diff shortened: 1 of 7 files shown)


SweeXordious (commented):
Where is the whole signature happening ? It seems that this PR only moves functionality from a place to another?  I guess I am missing some change

renaynay (approved):
lgtm

x/payment/types/square_sizes.go
```diff
@@ -0,0 +1,54 @@
+package types
+
+import (
+	"github.com/tendermint/tendermint/pkg/consts"
+)
+
+// https://github.com/celestiaorg/celestia-app/issues/236
+// https://github.com/celestiaorg/celestia-app/issues/239
```
liamsi, line 8: It's unclear why these are here 🤔

liamsi (approved):
Looks good. One minor nit

SweeXordious (approved):

x/payment/types/square_sizes.go
evan-forbes: whoops, yeah this code was originally just for test code, but those issues describe signing over all applicable square sizes
#236 #239 

This still needs a refactor and we should perform this calculation inside of `NewWirePayForMessage`.

Will do in a follow up PR later today

PR #454, opened 2022-05-28
Change P2P Network version and overide configs

evan-forbes:
## Description

This PR bumps celestia-core to use an incremented network version. :exclamation:  IMPORTANT :exclamation: : this will sever the p2p handshake between nodes that use an older binary. It also intercepts the configs and forces `timeout-commit = 25s` and `use-legacy = true`, along with changing those defaults.

NOTE: we still need to test that this actually works.

closes: #XXXX

### DIFF
--- a/cmd/celestia-appd/root.go
+++ b/cmd/celestia-appd/root.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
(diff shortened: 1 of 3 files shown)


evan-forbes:
wait, how is the linter failing all of a sudden? did it auto update or something? wth 
```
Warning: Unexpected input(s) 'skip-go-installation', valid inputs are ['version', 'args', 'working-directory', 'github-token', 'only-new-issues', 'skip-cache', 'skip-pkg-cache', 'skip-build-cache']
```

evan-forbes:
closing as we will not be doing this

Issue #438, opened 2022-05-23
QGB E2E use context and timeouts for stopping the network


SweeXordious:
I definitely don't think we should change this in this PR, as its not even a nit pick just a different option, but we could simply rely on this context to cancel instead of a timeout channel. :shrug:

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/408#discussion_r879313830_

We could rely also on timeouts to have multiple ways to stop the tests

PR #459, opened 2022-05-30
better context handling + timeouts for QGB E2E network

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/438

### DIFF
--- a/e2e/qgb/Makefile
+++ b/e2e/qgb/Makefile
@@ -5,4 +5,4 @@ PACKAGES=$(shell go list ./...)
 all: test
 
 test:
-	@QGB_INTEGRATION_TEST=true go test -mod=readonly -test.timeout 30m -v $(PACKAGES)
+	@QGB_INTEGRATION_TEST=true go test -mod=readonly -failfast -test.timeout 30m -v $(PACKAGES)
--- a/e2e/qgb/qgb_network.go
+++ b/e2e/qgb/qgb_network.go
@@ -52,6 +52,9 @@ func NewQGBNetwork(_ctx context.Context) (*QGBNetwork, error) {
 }
 
 // registerGracefulExit traps SIGINT or waits for ctx.Done() to release the docker resources before exiting
+// it is not calling `DeleteAll()` here as it is being called inside the tests. No need to call it two times.
+// this comes from the fact that we're sticking with unit tests style tests to be able to run individual tests
+// https://github.com/celestiaorg/celestia-app/issues/428
 func registerGracefulExit(cancel context.CancelFunc, network *QGBNetwork) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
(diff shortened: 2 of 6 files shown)


e2e/qgb/Makefile
```diff
@@ -5,4 +5,4 @@ PACKAGES=$(shell go list ./...)
 all: test
 
 test:
-	@QGB_INTEGRATION_TEST=true go test -mod=readonly -test.timeout 30m -v $(PACKAGES)
+	@QGB_INTEGRATION_TEST=true go test -mod=readonly -failfast -test.timeout 30m -v $(PACKAGES)
```
SweeXordious, line 8: I reintroduced this as it makes it easier to just everything when ctrl+c

e2e/qgb/qgb_network.go
```diff
@@ -86,6 +118,17 @@ func (network QGBNetwork) DeleteAll() error {
 	return nil
 }
 
+// KillAll kills all the containers.
+func (network QGBNetwork) KillAll() error {
```
SweeXordious, line 122: not currently used, but might be useful in the future

e2e/qgb/qgb_network.go
```diff
@@ -250,12 +293,15 @@ func (network QGBNetwork) StartBase() error {
 	return nil
 }
 
-func (network QGBNetwork) WaitForNodeToStart(rpcAddr string) error {
-	timeoutChan := time.After(5 * time.Minute)
+func (network QGBNetwork) WaitForNodeToStart(_ctx context.Context, rpcAddr string) error {
+	ctx, _ := context.WithTimeout(_ctx, 5*time.Minute) //nolint:govet
```
SweeXordious, line 297: no need for a cancel function, we only need the timeout. If the happy path executes, then it will just return.

e2e/qgb/qgb_network.go
```diff
@@ -26,19 +29,48 @@ type QGBNetwork struct {
...
+	}
+
+	// trap Ctrl+C or ctx.Done()
+	// helps release the docker resources without having to do it manually
+	registerGracefulExit(cancel, network)
+
+	return network, nil
+}
+
+// registerGracefulExit traps SIGINT or waits for ctx.Done() to release the docker resources before exiting
```
SweeXordious, line 54:
```suggestion
// registerGracefulExit traps SIGINT or waits for ctx.Done() to release the docker resources before exiting
// it is not calling `DeleteAll()` here as it is being called inside the tests. No need to call it two times.
// this comes from the fact that we're sticking with unit tests style tests to be able to run individual tests
// https://github.com/celestiaorg/celestia-app/issues/428
```

evan-forbes (commented):
iirc, the orignal comment was just to use context.Context as a timeout and not actually pass times to all of the functions. This way ctx.Done() combines both the timeout and the cancels. However I could definitely see this working too! I'm curious to hear more about what you had in mind, and we can discuss it in a sync sometime tmrw or whenever you have time

evan-forbes (approved):
apologies for any delay with this

Issue #331, opened 2022-04-15
restructure modules to be canononical and for readability


liamsi:
Currently, we use the file and folder structure from startport. What we should be doing is rather using the recommended folder structure here: https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md#recommended-folder-structure

Where it helps readability we should deviate from that. I think Osmosis did a great job structuring their modules: https://github.com/osmosis-labs/osmosis/tree/main/x

A good opportunity to tackle this, would be when we upgrade to tendermint 0.35 and 0.46 of the SDK (related to https://github.com/celestiaorg/celestia-core/issues/630#issuecomment-1090035895).

evan-forbes:
leaving a comment to remind us to also fix https://github.com/celestiaorg/celestia-app/pull/472#discussion_r938398480 when we address this issue

rootulp:
Closing this because we have stand-alone modules in the `x/` directory

PR #472, opened 2022-06-07
update to sdk v0.46.0 and downgrade tendermint to v0.34.20

evan-forbes:
## Description

This PR pull the latest from tendermint and the cosmos-sdk. The changes introduced in the sdk are consensus breaking and will need to be included in the upcoming hardfork. 

~~NOTE: the cosmos-sdk uses generics, so we also have to update to go 1.18. We can do that in a different PR instead of also doing it here.~~ see #573


liamsi (commented):
> We can do that in a different PR instead of also doing it here.

I'm in favour of this.

evan-forbes:
>> We can do that in a different PR instead of also doing it here.

> I'm in favour of this.

sorry, forget to edit the description after #573. which still needs to be merged in here (and will fix the last failing test)

cc @liamsi

go.mod
```diff
@@ -10,29 +10,34 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/spf13/cast v1.5.0
-	github.com/spf13/cobra v1.4.0
+	github.com/spf13/cobra v1.5.0
 	github.com/stretchr/testify v1.8.0
-	github.com/tendermint/tendermint v0.35.4
+	github.com/tendermint/tendermint v0.35.2
```
rootulp, line 15: [question] this line may not be used b/c the `replace` later in this file but out of curiosity, why v0.35.2 here instead of v0.34.20

```suggestion
	github.com/tendermint/tendermint v0.34.20
```

app/app.go
```diff
@@ -619,6 +629,17 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(clientCtx, app.BaseApp.GRPCQueryRouter(), app.interfaceRegistry, nil)
 }
 
+func (app *App) setPostHanders() {
+	postHandler, err := posthandler.NewPostHandler(
```
SweeXordious, line 633: Is there a reason why we are creating a new function for setting the `postHandler` and not for setting the `anteHandler`?

SweeXordious (approved):
:shipit:

go.mod
```diff
@@ -10,29 +10,34 @@ require (
...
-	golang.org/x/net v0.0.0-20220520000938-2e3eb7b945c2 // indirect
-	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
-	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211 // indirect
-	google.golang.org/genproto v0.0.0-20220519153652-3a47de7e79bd
-	google.golang.org/grpc v1.47.0
+	golang.org/x/crypto v0.0.0-20220525230936-793ad666bf5e // indirect
+	golang.org/x/net v0.0.0-20220617184016-355a448f1bc9 // indirect
+	golang.org/x/sys v0.0.0-20220702020025-31831981b65f // indirect
+	golang.org/x/term v0.0.0-20220526004731-065cf7ba2467 // indirect
+	google.golang.org/genproto v0.0.0-20220617124728-180714bec0ad
```
rootulp, line 21: [no change needed] I was curious about why there are three require blocks in this go.mod file and it seems like these `// indirect` deps can be moved to the last `require` block if we run `go mod tidy -go=1.17`. Separating direct deps from transitive deps is visual convenience so no change needed.

See: https://stackoverflow.com/questions/69432877/why-there-are-two-require-blocks-in-go-mod-since-go-1-17/69433550

I'm still unsure on why the second require block which contains direct dependencies is distinct from the first one. Perhaps because it includes `-beta` versions?

app/app.go
evan-forbes: I _think_ we are, is this what you mean? https://github.com/celestiaorg/celestia-app/blob/82393670fd39e2a01a13811ceaf9ffae200651f7/app/app.go#L506-L507

we don't define a method as baseapp already has one iirc

rootulp (dismissed):

go.mod
SweeXordious: I can confirm that running `go mod tidy` locally makes more changes to `go.mod` and `go.sum`.
@rootulp Btw, why do you need  to run `go mod tidy -go=1.17`  ? I guess  `go mod tidy` is enough no?

app/app.go
SweeXordious: Ah okey, got you :+1:

SweeXordious (dismissed):

go.mod
SweeXordious: btw removing that line doesn't seem to affect the project. Did it locally and was able to run everything correctly.

go.mod
rootulp: > Btw, why do you need to run `go mod tidy -go=1.17`

I didn't try this on the PR (yet) but this was based on the StackOverflow post which says:

> and/or run go mod tidy -go=1.17

More docs:
```shell
go help mod tidy
...
The -go flag causes tidy to update the 'go' directive in the go.mod
file to the given version, which may change which module dependencies
are retained as explicit requirements in the go.mod file.
(Go versions 1.17 and higher retain more requirements in order to
support lazy module loading.)
```

so it seems like Go 1.17+ will format `go.mod` slightly differently. Given the Go version installed on our machines + in this package's `go.mod` is already Go 1.17, the `-go=1.17` flag likely isn't necessary so apologies for the confusion

SweeXordious: I think the repo is currently using `1.18`

go.mod
evan-forbes: yeah, its still working since we are blindly forcing all uses of tendermint to celestia-core instead of only instance of v0.34.20

good catch tho

go.mod
rootulp: You're right b/c https://github.com/celestiaorg/celestia-app/pull/573
Based on my understanding of the above docs, 1.17+ will format go.mod `require` sections in separate blocks.

rootulp (approved):

SweeXordious (approved):

app/app.go
```diff
@@ -619,6 +629,17 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
...
+		posthandler.HandlerOptions{},
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	app.SetPostHandler(postHandler)
+}
```
liamsi, line 641: Why is this snippet necessary?

liamsi (approved):
Nothing sus. LGTM

app/app.go
evan-forbes: that's a good point, I don't think it is, and we could put it this is normal typical definition of the app. tbh, I don't remember where I was going with this as the base of this PR is pretty old.

not changing for now simply so I don't dismiss the current approvals, but this should be fixed when we do #331

PR #475, opened 2022-06-07
bump core to v1.2.4-tm-v0.35.6

evan-forbes:
## Description

pull all the latest changes from tendermint + our fixes

closes: #XXXX

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -157,5 +157,5 @@ replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.1.1-sdk-v0.46.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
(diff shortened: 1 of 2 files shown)


SweeXordious (dismissed):
LGTM. The CI apparently needs to be restarted.

evan-forbes:
> LGTM. The CI apparently needs to be restarted.

this is actually a bug that I've been digging into the past few hours. I'm not sure exactly what is going on or if its relevant to tendermint, but it doesn't pop up in the previous v0.35. For some reason, when two validators are used in the big block test, they have trouble coming to consensus with the latest update. There are quite a few rounds, and proposal block parts are gossiped incredibly slowly *sometimes* for large blocks. Its really weird!

This bug pops up on the version that I and a few others are already running on mamaki, and this isn't a problem atm. Really really weird, but I don't think we should merge until this is figured out, so converting to a draft for now.

evan-forbes:
This is blocked by https://github.com/celestiaorg/celestia-core/pull/783

evan-forbes:
> This is blocked by https://github.com/celestiaorg/celestia-core/pull/783

this isn't the case any longer as we instead changed the default peer gossip timeout, which fixed the failing test mentioned earlier

SweeXordious (approved):

adlerjohn (approved):

liamsi (approved):

Issue #268, opened 2022-04-01
Add DataCommitmentRequest to the qgb module


evan-forbes:
Currently, there is no way for the protocol or users to request that a data commitment be made and signed over. The validators are expected to create and sign over a commitment that covers the block data of some number of blocks based on the data commitment window defined in the parameters. 

In order to eventually accommodate slashing and variable sized data commitment windows, we will have to we will have to add something similar to `ValsetRequest`s, where some formal request is stored in the state, and validators can listen and sign over that commitment.

SweeXordious:
This struct should be used as a replacement for: `ExtendedDataCommitment` and also should be used to get the latest nonce when relaying data commitments

SweeXordious:
@evan-forbes I will start with this one before implementing the universal nonce. It would help to have the state machine decide on the nonces that will be sent not to have inconsistencies.

PR #478, opened 2022-06-09
Adds data commitment request

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/268


x/qgb/abci.go
```diff
@@ -9,6 +9,18 @@ import (
 
 // EndBlocker is called at the end of every block
 func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
+	handleValset(ctx, k)
+	handleDataCommitment(ctx, k)
```
evan-forbes, line 13: similar nit as above, I think these should be named a bit differently. while this works fine, they are creating data commitment and valset requests, not really handling a valset or data commitment

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -244,63 +258,57 @@ func (oc *orchestratorClient) addOldDataCommitmentAttestations(
 ) {
 	oc.logger.Info("Started adding old Data Commitments attestation to queue")
 	defer oc.logger.Info("Finished adding old Data Commitments attestation to queue")
+	// FIXME I guess we can remove this until we have slashing. The reason being, without slashing
+	// it would be good to just sign everything. So that all attestations are signed.
```
evan-forbes, line 262: that could lead to a lot of waste, especially for new validators who don't need to sign. imo we should definitely keep this. Do we check if we've already signed old ones somewhere?

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -194,43 +193,58 @@ func (oc *orchestratorClient) SubscribeDataCommitment(ctx context.Context) (<-ch
...
+				dc, err := oc.querier.QueryDataCommitmentByNonce(ctx, latestDCNonce)
+				if err != nil {
+					oc.logger.Error(err.Error())
+					time.Sleep(5 * time.Second)
+					continue
+				}
+				if dc == nil {
+					time.Sleep(5 * time.Second)
+					continue
+				}
```
evan-forbes, line 214: will this just loop and query every 5 seconds?

I'm not the biggest fan of hardcoding these timeouts, but I'm okay with keeping them since we will certainly change them with the worker refactor

evan-forbes (commented):
some quick questions and minor nits. There are some things that I would normally request being changed, but as previously discussed, its fine since we will be refactoring to the worker routine and this will certainly be removed

evan-forbes (approved):
pre-approving

make the changes you see fit based on the comments and context, then we can :shipit: :slightly_smiling_face:

x/qgb/orchestrator/orchestrator_client.go
```diff
@@ -244,63 +258,57 @@ func (oc *orchestratorClient) addOldDataCommitmentAttestations(
...
-	currentHeight, err := oc.querier.QueryHeight(ctx)
+	latestDCNonce, err := oc.querier.QueryLatestDataCommitmentNonce(ctx)
 	if err != nil {
 		oc.logger.Error(err.Error())
 		return
 	}
 
-	var previousBeginBlock uint64
-	var previousEndBlock uint64
+	for nonce := latestDCNonce; nonce >= 0; nonce-- { //nolint:staticcheck
```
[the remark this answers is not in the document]
SweeXordious: It won't loop forever, but I will switch to the substraction way. Thanks

proto/qgb/types.proto
```diff
@@ -23,3 +23,19 @@ message Valset {
   // Current chain height
   uint64 height = 3;
 }
+
+// DataCommitment is the data commitment request message that will be signed
+// using orchestrators.
+// It does not contain a `commitment` field as this message will be created
+// inside the state machine and it doesn't make sense to ask tendermint for the
+// commitment there.
+message DataCommitment {
```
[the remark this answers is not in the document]
SweeXordious: Makes sense, just wanted to  keep the same convention as valsets... We don't have a valsetRequest, but we only have a Valset containing the necessary information.
What do you think ? should I keep it or  change ?

x/qgb/types/data_commitment.go
```diff
@@ -0,0 +1,29 @@
...
+	return len(dc)
+}
+
+func (dc DataCommitments) Less(i, j int) bool {
+	return dc[i].Nonce > dc[j].Nonce
+}
+
+func (dc DataCommitments) Swap(i, j int) {
+	dc[i], dc[j] = dc[j], dc[i]
+}
```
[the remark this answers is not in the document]
SweeXordious: I think this will be removed in the worker pool PR as it's not much needed really.

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: yes, even if we want to sign again, it will raise an exception. What I mean is that since we don't have slashing, all orchestrators should sign the whole history, especially for testnets...  But this is to be discussed really, doesn't need to be included in this PR

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: yes, will change, that's why I am adding it  for now.

proto/qgb/types.proto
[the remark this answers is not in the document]
evan-forbes: I see, I was mistaken in that I thought we were using requests like the gravity module. We shouldn't change anything then and it should be in a different PR if we care enough

x/qgb/orchestrator/orchestrator_client.go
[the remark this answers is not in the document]
SweeXordious: It will, my bad. Fixing

x/qgb/orchestrator/orchestrator_client.go
SweeXordious: Done https://github.com/celestiaorg/celestia-app/issues/479

PR #481, opened 2022-06-10
Fixes set latest data commitment nonce

SweeXordious:
A fix for a wrong check when setting the latest data commitment nonce


evan-forbes (approved):
:eyes:

Issue #486, opened 2022-06-16
Correct PayForData signatures are overwritten with signatures of a larger square size


evan-forbes:
and weirder still, some pay for messages are failing due to an incorrect signature, despite this message passing on integration and unit tests, but it indicates that it is using gas

https://celestia.explorers.guru/transaction/0D98FA208FB10AE404B6D3FA3643C8A69E8A6DB0EECD6592BB859826D3B1BE44

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/issues/452#issuecomment-1140052535_

Due to these lines of code 
https://github.com/celestiaorg/celestia-app/blob/b52788be591a9d93c26b46004dc39f67a80eecbe/x/payment/types/wirepayfordata.go#L197-L202

we are overwriting the correct signatures with the signatures of larger square sizes. This is getting past our current tests because they are either testing only for large block size (in the case of our integration tests) or are not going through the entire malleation process (unit tests).

Issue #487, opened 2022-06-16
The canonical client code will produce incorrect PfD signatures if fees are used.


evan-forbes:
Using the canonical client code and adding fees to the tx will result in incorrect signatures for PFD transactions. This can be easily fixed by passing the TxBuilderOptions (whose purpose is to avoid this :sweat_smile:) to the function that build the WirePfDs. More importantly, we need to add integration tests for the canonical client code

PR #488, opened 2022-06-16
fix: Patch all PFD signature bugs 

evan-forbes:
## Description

patches all of the bugs that stop incorrect signatures for PFDs from making it into the block.

closes: #486 #487


x/payment/types/verify_sigs.go
```diff
@@ -0,0 +1,69 @@
...
+		return false, errorSinglSignerExpected
+	}
+
+	return signerData.PubKey.VerifySignature(signBytes, rawSig.Signature), nil
+}
+
+// VerifyPFDSigs checks that all of the signatures for a transaction that
+// contains a MsgWirePayForData message by going through the entire malleation
+// process.
+func VerifyPFDSigs(signerData authsigning.SignerData, txConfig client.TxConfig, wirePFDTx authsigning.Tx) (bool, error) {
```
evan-forbes, line 43: we might even want to call this function when generating any wirePfD using any client code, either the canonical or the cli

x/payment/types/wirepayfordata.go
```diff
@@ -221,3 +215,33 @@ func ProcessWirePayForData(msg *MsgWirePayForData, squareSize uint64) (*tmproto.
...
+
+	// only support malleated transactions that contain a single sdk.Msg
+	if len(tx.GetMsgs()) != 1 {
+		return nil, errors.New("sdk.Txs with a single MsgWirePayForData are currently supported")
+	}
+
+	msg := tx.GetMsgs()[0]
+	wireMsg, ok := msg.(*MsgWirePayForData)
+	if !ok {
+		return nil, noWirePFDError
```
adlerjohn, line 243: This looks like it should be a panic, or a different error. The calling function of this is only prepared to handle the first error on line 232.

x/payment/types/wirepayfordata.go
```diff
@@ -221,3 +215,33 @@ func ProcessWirePayForData(msg *MsgWirePayForData, squareSize uint64) (*tmproto.
...
+
+func ExtractMsgWirePayForData(tx sdk.Tx) (*MsgWirePayForData, error) {
+	noWirePFDError := errors.New("sdk.Tx does not contain MsgWirePayForData sdk.Msg")
+	if !HasWirePayForData(tx) {
+		return nil, noWirePFDError
+	}
+
+	// only support malleated transactions that contain a single sdk.Msg
+	if len(tx.GetMsgs()) != 1 {
+		return nil, errors.New("sdk.Txs with a single MsgWirePayForData are currently supported")
```
adlerjohn, line 237: Ditto

x/payment/types/wirepayfordata.go
evan-forbes: > The calling function of this is only prepared to handle the first error on line 232.

I think I'm missing something here, why is the calling function only equipped to handle the first error but not this one?

adlerjohn: This function is only used when preparing a proposal, which assumes that transactions have either 0 or 1 data. If it has 0, then it can be included in a block via some other means.

The issue is that this function also handles cases where there could be more than 1 data. Such transactions will never be included in blocks and continue to congest the mempool forever. If they have high priority fees (whenever that lands in the Cosmos SDK), then they'll never be evicted but also never included. This applies more to the above error from checking `if len(tx.GetMsgs()) != 1`.

As for this error, shouldn't it be impossible given the precondition that `HasWirePayForData(tx)` is true?

evan-forbes: > This function is only used when preparing a proposal, which assumes that transactions have either 0 or 1 data. If it has 0, then it can be included in a block via some other means.

it's only used during estimating the number of shares that a message takes up and when verifying signatures of PfD during tests, not when determining if a tx is included in the block or not.

> As for this error, shouldn't it be impossible given the precondition that HasWirePayForData(tx) is true?

the precondition only quickly checks that the proto message's URL is `/payment.WirePayForData` which could be faked or from some old version of wirePfD. Something else could go wrong as well and we don't want to panic so we have to handle/bubble this error

evan-forbes: https://github.com/celestiaorg/celestia-app/blob/823063062789c6a2786ce1d06ea758d4c46bbd28/app/prepare_proposal.go#L97-L103

even with these comments, I can see this being confusing. This might be a premature abstraction that trades too much comprehension of the code for simply not copying the code in two places. I'm happy to revert it if we think that's a good idea

adlerjohn: Reverting probably isn't necessary, but could you take a stab at maybe improving the comments?

evan-forbes: yeah, definitely 37beabb42599d972394fd937697decf1f3d28737

adlerjohn (approved):

PR #512, opened 2022-06-30
Orchestrator refactor

SweeXordious:

PR #519, opened 2022-07-04
Go fumpt the qgb-integration branch

SweeXordious:
Applies https://github.com/mvdan/gofumpt on the branch


SweeXordious:
Depends on https://github.com/celestiaorg/celestia-app/pull/512

evan-forbes (approved):
approving pending the e2e test passing

adlerjohn (approved):

adlerjohn:
Can you add a CI job to check this?

evan-forbes:
Yeah I agree with John about using CI. gofumpt is backwards compatible and doesn't do any harm, but whatever linter we choose, we should be consistent instead of one off fixes. 

I'm also in favor of of adding stricter restrictions to golangci, as that's what tendermint and the sdk (and the app currently) use, so that's one less tool contributors have to install.

SweeXordious:
Issue created. Merging.

Issue #406, opened 2022-05-13
QGB Orchestrator should stop signing if no longer in the valset


SweeXordious:
Currently, the QGB orchestrator keeps signing attestations even if it  is not part of the valset. These signatures will  be eventually rejected. Thus, the orchestrator should check if it is in the valset before signing attestations.

evan-forbes:
one point to note, is that a validator could potentially fall into and out of the validator set every other block, and ideally the orchestrator would be able to handle this as well, although this might be a premature feature.

SweeXordious:
100% it's still too early for this. Except that I wanted to keep track of all ideas so that after QGB 1.0,  we have a list of improvements we can work on.
Btw, once the IT tests are set, let's go for a call to decide on the set of features we want for QGB 1.0.

Issue #482, opened 2022-06-10
Improve QGB testing coverage


SweeXordious:
Currently, multiple unit tests are not added since the project is still in early phases, and designs change frequently. However,  once we're confident, we will need to increase test coverage.

SweeXordious:
Also, go more for table driven tests.

evan-forbes:
per sync discussion, this issue can be closed due to moving to v2

PR #523, opened 2022-07-05
QGB Orchestrator only signs if part of valset

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/406


SweeXordious:
depends on https://github.com/celestiaorg/celestia-app/pull/519

x/qgb/keeper/msg_server.go
```diff
@@ -142,6 +153,28 @@ func (k msgServer) DataCommitmentConfirm(
 		return nil, sdkerrors.Wrap(types.ErrInvalid, "submitted eth address does not match delegate eth address")
 	}
 
+	// Verify if validator is part of the previous valset
+	var previousValset *types.Valset
+	// TODO add test for case nonce == 1.
+	if msg.Nonce == 1 {
+		// if the msg.Nonce == 1, the current valset should sign the first valset
```
evan-forbes, line 160: we might want to add to this comment as to why we do this.

evan-forbes (approved):
great work! :1st_place_medal:

x/qgb/handler_test.go
```diff
@@ -20,27 +21,44 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TODO add test for all the possible scenarios defined in msg_server.go
+// TODO add test for all the possible scenarios defined in msg_server.go.
```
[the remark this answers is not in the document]
SweeXordious: yes, it's possible. Actually, a lot of tests in the QGB module contain duplicate code that can be replaced with tables. Just added it to this issue: https://github.com/celestiaorg/celestia-app/issues/482#issuecomment-1181970244

x/qgb/keeper/msg_server.go
```diff
@@ -59,6 +51,25 @@ func (k msgServer) ValsetConfirm(
 		return nil, sdkerrors.Wrapf(err, "discovered invalid validator address for orchestrator %v", orchaddr)
 	}
 
+	// Verify if validator is part of the previous valset
+	var previousValset *types.Valset
+	// TODO add test for case nonce == 1.
+	if msg.Nonce == 1 {
+		// if the msg.Nonce == 1, the current valset should sign the first valset
```
SweeXordious, line 58:
```suggestion
		// if the msg.Nonce == 1, the current valset should sign the first valset. Because, it's the first attestation, and there is no prior validator set defined that should sign this change.
```

x/qgb/keeper/msg_server.go
SweeXordious, line 160:
```suggestion
		// if the msg.Nonce == 1, the current valset should sign the first valset. Because, it's the first attestation, and there is no prior validator set defined that should sign this change.
```

Issue #381, opened 2021-10-19
Create API to check for message inclusion


evan-forbes:
## Summary

Create an API to check for messages inclusion.

## Details

The first step has already been done by writing code to find subtree root paths. The next step should be to create an API that allows us to easily check for message inclusion given the original data square, the position of the message in question, the size of that message, along with the commitment provided in the `SignedPayForMessage`.

### Action Items
- [ ] define API for checking inclusion
- [ ] compare commitments and subtree roots the `SignedPayForData` and the actual data in the square given the location and message size.
- [ ] write integration tests for subtree root code

### References

depends on [#49](https://github.com/celestiaorg/nmt/pull/49)

PR #549, opened 2022-07-15
feat: Cache and traverse nmt sub tree roots

evan-forbes:
## Description

spinning this out into its own PR. I think its mostly finished, so we can begin the review process imo, but I'm keeping as a draft for the moment as it might need to change and I don't really want to merge anything to master until we have fully functional non-interactive defaults.

a big part of #381


liamsi:
> closes https://github.com/celestiaorg/celestia-app/pull/216

216 is a merged PR?

pkg/inclusion/sub_tree_root.go
```diff
@@ -0,0 +1,125 @@
...
+	case 1:
+		return
+	default:
+		panic("unexpected visit")
+	}
+}
+
+// walk recursively traverses the subTreeRootCacher's internal tree by using the
+// provided sub tree root and path. The provided path should be a []bool, false
+// indicating that the first child node (right most node) should be used to find
```
rootulp, line 41: I think these are flipped. The implementation walks left if false

```suggestion
// indicating that the first child node (left most node) should be used to find
```

pkg/inclusion/sub_tree_root.go
```diff
@@ -0,0 +1,125 @@
...
+// the next path, and the true indicating that the second (left) should be used.
```
rootulp, line 42:
```suggestion
// the next path, and the true indicating that the second (right) should be used.
```

pkg/inclusion/sub_tree_root.go
```diff
@@ -0,0 +1,125 @@
...
+	// walk right
+	case true:
+		return strc.walk([]byte(children[1]), path[1:])
+	default:
+		// this is unreachable code, but the compiler doesn't recognize this somehow
+		panic("bool other than true or false, computers were a mistake, everything is a lie, math is fake.")
+	}
+}
+
+// EDSSubTreeRootCacher caches the the inner nodes for each row so that we can
```
rootulp, line 70:
```suggestion
// EDSSubTreeRootCacher caches the inner nodes for each row so that we can
```

pkg/inclusion/sub_tree_root.go
```diff
@@ -0,0 +1,125 @@
...
+
+func NewCachedSubtreeCacher(squareSize uint64) *EDSSubTreeRootCacher {
+	return &EDSSubTreeRootCacher{
+		caches:     []*subTreeRootCacher{},
+		squareSize: squareSize,
+	}
+}
+
+// Constructor fullfills the rsmt2d.TreeCreatorFn by keeping a pointer to the
+// cache and embedding it as a nmt.NodeVisitor into a new wrapped nmt. I only
```
rootulp, line 94: I think the next line(s) of this doc comment were accidentally removed

pkg/inclusion/sub_tree_root.go
```diff
@@ -0,0 +1,125 @@
...
+		newTree = wrapper.NewErasuredNamespacedMerkleTree(stc.squareSize, nmt.NodeVisitor(strc.Visit))
+	default:
+		newTree = wrapper.NewErasuredNamespacedMerkleTree(stc.squareSize)
+	}
+
+	stc.counter++
+	return &newTree
+}
+
+// GetSubTreeRoot traverses the nmt of of the selected row and returns the
```
rootulp, line 113:
```suggestion
// GetSubTreeRoot traverses the nmt of the selected row and returns the
```

adlerjohn (commented):
Is this one ready for review @evan-forbes?

evan-forbes:
> Is this one ready for review @evan-forbes?

kinda but not quite. I would like to have a working well tested version as to avoid changing things, but this will very likely look very similar to the final product so its probably safe to look at and review if you see anything. 

I think I renamed the files on a different branch that's further along, but that's it

evan-forbes:
will break this this branch up a bit now that we have a better idea of what works and is needed

evan-forbes:
marking this ready for review again. It is unchanged from the last review, other than I spun off #621 and moved everything else the non-interactive-defaults feature branch

pkg/inclusion/nmt_caching.go
```diff
@@ -0,0 +1,125 @@
...
+// inner nodes.
+type subTreeRootCacher struct {
+	cache map[string][2]string
+}
+
+func newSubTreeRootCacher() *subTreeRootCacher {
+	return &subTreeRootCacher{cache: make(map[string][2]string)}
+}
+
+// Visit fullfills the nmt.VisitorNode function definition. It stores each inner
```
rootulp, line 25: Ref: https://github.com/celestiaorg/nmt/blob/master/nmt.go#L22
```suggestion
// Visit fullfills the nmt.NodeVisitorFn function definition. It stores each inner
```

pkg/inclusion/nmt_caching.go
```diff
@@ -0,0 +1,125 @@
...
+
+func NewCachedSubtreeCacher(squareSize uint64) *EDSSubTreeRootCacher {
+	return &EDSSubTreeRootCacher{
+		caches:     []*subTreeRootCacher{},
+		squareSize: squareSize,
+	}
+}
+
+// Constructor fullfills the rsmt2d.TreeCreatorFn by keeping a pointer to the
+// cache and embedding it as a nmt.NodeVisitor into a new wrapped nmt. I only
```
rootulp, line 94: [blocking]
the rest of this comment got truncated

pkg/inclusion/nmt_caching.go
```diff
@@ -0,0 +1,125 @@
...
+// subtree root. An error is thrown if the subtree cannot be found.
+func (stc *EDSSubTreeRootCacher) GetSubTreeRoot(dah da.DataAvailabilityHeader, row int, path []bool) ([]byte, error) {
+	const unexpectedDAHErr = "data availability header has unexpected number of row roots: expected %d got %d"
+	if len(stc.caches) != len(dah.RowsRoots) {
+		return nil, fmt.Errorf(unexpectedDAHErr, len(stc.caches), len(dah.RowsRoots))
+	}
+	const rowOutOfBoundsErr = "row exceeds range of cache: max %d got %d"
+	if row >= len(stc.caches) {
+		return nil, fmt.Errorf(rowOutOfBoundsErr, len(stc.caches), row)
+	}
```
rootulp, line 123: [question][non-blocking] given these error messages are only used once, what is the benefit to extracting constants over inlining the message like:

```suggestion
	if len(stc.caches) != len(dah.RowsRoots) {
		return nil, fmt.Errorf("data availability header has unexpected number of row roots: expected %d got %d", len(stc.caches), len(dah.RowsRoots))
	}
	if row >= len(stc.caches) {
		return nil, fmt.Errorf("row exceeds range of cache: max %d got %d", len(stc.caches), row)
	}
```

pkg/inclusion/nmt_caching_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// create a small sub tree
+	subTree1 := wrapper.NewErasuredNamespacedMerkleTree(oss)
+	for i := 0; i < 2; i++ {
+		subTree1.Push(d, rsmt2d.SquareIndex{
+			Axis: uint(rsmt2d.Row),
+			Cell: uint(i),
+		})
+	}
+	shortSTR := subTree1.Root()
```
rootulp, line 38: [nit][optional] consistent naming 

```suggestion
	// create a short sub tree
	shortSubTree := wrapper.NewErasuredNamespacedMerkleTree(oss)
	for i := 0; i < 2; i++ {
		shortSubTree.Push(d, rsmt2d.SquareIndex{
			Axis: uint(rsmt2d.Row),
			Cell: uint(i),
		})
	}
	shortSTR := shortSubTree.Root()
```

pkg/inclusion/nmt_caching_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// create a larger sub tree root
+	subTree2 := wrapper.NewErasuredNamespacedMerkleTree(oss)
+	for i := 0; i < 4; i++ {
+		subTree2.Push(d, rsmt2d.SquareIndex{
+			Axis: uint(rsmt2d.Row),
+			Cell: uint(i),
+		})
+	}
+	tallSTR := subTree2.Root()
```
rootulp, line 48: [nit][optional] consistent naming

```suggestion
	// create a tall sub tree root
	tallSubTree := wrapper.NewErasuredNamespacedMerkleTree(oss)
	for i := 0; i < 4; i++ {
		tallSubTree.Push(d, rsmt2d.SquareIndex{
			Axis: uint(rsmt2d.Row),
			Cell: uint(i),
		})
	}
	tallSTR := tallSubTree.Root()
```

pkg/inclusion/nmt_caching_test.go
```diff
@@ -0,0 +1,178 @@
...
+			Cell: uint(i),
+		})
+	}
+	tallSTR := tallSubTree.Root()
+
+	type test struct {
+		name        string
+		path        []bool
+		subTreeRoot []byte
+		error       string
```
SweeXordious, line 54: Why is `error`  a `string`?  Also, it seems to be always empty

pkg/inclusion/nmt_caching_test.go
```diff
@@ -0,0 +1,178 @@
...
+			[]bool{false},
+			tallSTR,
+			"",
+		},
+		{
+			"right most tall sub tree",
+			[]bool{true},
+			tallSTR,
+			"",
+		},
```
SweeXordious, line 93: Shouldn't we add also cases where this would fail? Like for an empty path and a path longer than the depth of the tree?

SweeXordious (dismissed):
LGTM :rocket:. Only left some questions

pkg/inclusion/nmt_caching_test.go
evan-forbes: good idea! added a case here https://github.com/celestiaorg/celestia-app/pull/549/commits/b6745f6f6acdfbd724898610096bf32c7cb3e088

pkg/inclusion/nmt_caching_test.go
evan-forbes: Its just the expected error, so it checks for more than just if there was an error

pkg/inclusion/nmt_caching.go
```diff
@@ -0,0 +1,123 @@
...
+		newTree = wrapper.NewErasuredNamespacedMerkleTree(stc.squareSize)
+	}
+
+	stc.counter++
+	return &newTree
+}
+
+// GetSubTreeRoot traverses the nmt of the selected row and returns the
+// subtree root. An error is thrown if the subtree cannot be found.
+func (stc *EDSSubTreeRootCacher) GetSubTreeRoot(dah da.DataAvailabilityHeader, row int, path []bool) ([]byte, error) {
```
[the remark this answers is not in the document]
evan-forbes: there is a check on rows, its just kinda hidden

```go
	if len(stc.caches) != len(dah.RowsRoots) {
		return nil, fmt.Errorf("data availability header has unexpected number of row roots: expected %d got %d", len(stc.caches), len(dah.RowsRoots))
	}
	if row >= len(stc.caches) {
		return nil, fmt.Errorf("row exceeds range of cache: max %d got %d", len(stc.caches), row)
	}
```

only relying on the error from walking to do the check instead of checking for max length. There probably could be, but since we're the only ones using this functions, it seems a bit overkill to compute the square root of the length of the dah roots

we might be able to not export this entire struct actually, come to think of it. will wait until we add the get commitment function used in the implement. If we're not using any of these functions we can unexport them

evan-forbes:
wow, I thought I was replying to things here, but apparently github will treat it as a review if you have pending comments and just continue to add them as pending comments!!

pkg/inclusion/nmt_caching.go
evan-forbes: sorry, this was fixed two weeks ago, but I didn't realize github wasn't posting the comments

pkg/inclusion/nmt_caching.go
```diff
@@ -0,0 +1,123 @@
...
+		panic("unexpected visit")
+	}
+}
+
+// walk recursively traverses the subTreeRootCacher's internal tree by using the
+// provided sub tree root and path. The provided path should be a []bool, false
+// indicating that the first child node (left most node) should be used to find
+// the next path, and the true indicating that the second (right) should be used.
+// walk throws an error if the sub tree cannot be found.
+func (strc subTreeRootCacher) walk(root []byte, path []bool) ([]byte, error) {
```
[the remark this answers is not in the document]
evan-forbes: yeah, I like this idea, and added it to https://github.com/celestiaorg/celestia-app/pull/549/commits/b6745f6f6acdfbd724898610096bf32c7cb3e088

rootulp (approved):
🚀

SweeXordious (approved):

Issue #437, opened 2022-05-23
QGB keepers throwing exception when element not found


SweeXordious:
As of the discussion under https://github.com/celestiaorg/celestia-app/pull/408#discussion_r878159254 , we will need to investigate whether to throw an exception when a query doesn't find an element or not.
We should also take a look on the other chains and see how they're handling  this.

PR #550, opened 2022-07-20
QGB State machine error handling

SweeXordious:
Improves error handling in state machine. It follows the following thinking:
- If the state machine is automatically setting something, example, attestation requests, it should panic if something goes wrong.
- If the state machine is getting something from store:
    - either panic for the state that necessarily needs to exist, such as `LastAttestationNonce`.
    - return `(value, found, error)` in other cases to know if element doesn't exist or some error occurred.
- If the state machine is receiving requests from users getting/setting stuff, like setting a confirm:
    - The keepers should return `(value, found, error)` to know if an element doesn't exist or some error occurred.
    - The queries would return `RequestResponse{ element: nil }, nil` if the element doesn't exit and uses will need to do their checks.

Closes: https://github.com/celestiaorg/celestia-app/issues/437


x/qgb/keeper/keeper_data_commitment_confirm.go
```diff
@@ -11,28 +11,30 @@ import (
...
 // GetDataCommitmentConfirm Returns a data commitment confirm by nonce and validator address
 // nonce = endBlock % data window in decimal base
+// Returns (nil, false, nil) if element not found.
 func (k Keeper) GetDataCommitmentConfirm(
 	ctx sdk.Context,
 	endBlock uint64,
 	beginBlock uint64,
 	validator sdk.AccAddress,
-) *types.MsgDataCommitmentConfirm {
+) (*types.MsgDataCommitmentConfirm, bool, error) {
```
evan-forbes, line 20: I like the added explicitness here

x/qgb/keeper/keeper_valset.go
```diff
@@ -11,13 +11,24 @@ import (
...
 	for i := uint64(0); i <= nonce; i++ {
-		at := k.GetAttestationByNonce(ctx, nonce-i)
+		at, found, err := k.GetAttestationByNonce(ctx, nonce-i)
+		if err != nil {
+			return nil, err
+		}
+		if !found {
+			panic(sdkerrors.Wrap(
+				types.ErrNilAttestation,
+				fmt.Sprintf("stumbeled upon nil attestation for nonce %d", i),
```
evan-forbes, line 29:
```suggestion
				fmt.Sprintf("stumbled upon nil attestation for nonce %d", i),
```

evan-forbes, line 29:
```suggestion
				fmt.Sprintf("found nil attestation for nonce %d", i),
```
alternatively this

x/qgb/keeper/keeper_valset.go
```diff
@@ -140,18 +160,23 @@ func (k Keeper) GetLastValsetBeforeNonce(ctx sdk.Context, nonce uint64) (*types.
 }
 
 // TODO add query for this method and make the orchestrator Querier use it.
-func (k Keeper) GetValsetByNonce(ctx sdk.Context, nonce uint64) (*types.Valset, error) {
-	at := k.GetAttestationByNonce(ctx, nonce)
-	if at == nil {
-		return nil, types.ErrAttestationNotFound
+// GetValsetByNonce returns the valset in store with nonce nonce.
```
evan-forbes, line 163:
```suggestion
// GetValsetByNonce returns the stored valset associated with the provided nonce.
```

evan-forbes (approved):
nice! I like this a lot. Really good work as per usual

proto/qgb/query.proto
```diff
@@ -36,22 +37,26 @@ service Query {
...
       returns (QueryDataCommitmentConfirmResponse) {
     option (google.api.http).get =
         "/qgb/data_commitment/confirm/begin/"
         "{begin_block}/end/{end_block}/address/{address}";
   }
   // DataCommitmentConfirms Queries data commitment confirms by the actual
-  // commitment
+  // commitment.
+  // Too heavy, shouldn't mainly be used.
+  // Returns empty array if no element is found.
```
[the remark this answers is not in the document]
SweeXordious: yes, definitely, it has to do with indexing. If we start double indexing data using multiple keys, we can define iterators with more specific store ranges, and it would be way faster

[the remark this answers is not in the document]
SweeXordious: We should investigate however the storage cost if it is worth it compared to the gained performance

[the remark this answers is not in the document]
SweeXordious: opened issue: https://github.com/celestiaorg/celestia-app/issues/552

PR #558, opened 2022-07-24
QGB handle nonce 1 case 

SweeXordious:
adds comments for valset nonce == 1 case and removes unnecessary msg_Server check on data commitments


evan-forbes (approved):
pre-approving :+1: I appreciate the docs

PR #573, opened 2022-07-27
chore: update to go 1.18

### LINKED ISSUE #572: Update to go 1.18
the cosmos-sdk v0.46.0 uses generics, so we need to update to go 1.18 in order to upgrade. This will also mean updating our linter as the current version only supports v1.17

blocking #472

evan-forbes:
## Description

This PR updates this repo to use go 1.18. It also requires an update to golangci-linter in order for that to keep working. 

closes #572 
unblocks #472

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,7 +25,6 @@ jobs:
             go.sum
       - uses: golangci/golangci-lint-action@v3.2.0
         with:
-          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: v1.47.2
           args: --timeout 10m
           github-token: ${{ secrets.github_token }}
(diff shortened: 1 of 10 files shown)


.github/workflows/lint.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v3.2.0
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
```
rootulp, line 28: [optional][nit] this comment no longer appears relevant for golangci-lint-action 3.0.0+ based on
```
# Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
    version: v1.29
```
from https://github.com/golangci/golangci-lint-action 

so proposal to remove the comment because we do specify a patch version
```suggestion

```

rootulp (approved):

SweeXordious (dismissed):
LGTM. CI apparently only needs to be restarted to be green.

rootulp (approved):

Issue #571, opened 2022-07-27
Consider adding explicit block validity checks to ensure messages with reserved namespaces do not make it in the block


evan-forbes:
> why doesn't this function throw an error if NamespaceID is lower. than the max reserved namespace ID?

Nice catch! this makes sense to add imo, as this would stop an honest user from making a mistake and is explicit. Feel free to add this to this PR or we can elsewhere. 

As a note, we might also want to add checks via a validate basic method on [this struct](https://github.com/celestiaorg/celestia-core/blob/ff54db5d24f005696c17051d8e737c4285eba24c/types/block.go#L1171-L1182) as well (and proceed to actually call that method). The reason being that validator could include an invalid PFD with message that includes a reserve namespace, and we might not catch it.

This makes things more complicated too, as weird things could happen if someone does manage to include a reserved namespace. It might cause bugs for nodes that would rely on block reconstruction (meaning they eventually call `DataFromSquare`, which I don't think we have yet) where a message with a reserved namespace makes it in the square, but then doens't make it in the block data when call `DataFromSquare`.
https://github.com/celestiaorg/celestia-core/blob/ff54db5d24f005696c17051d8e737c4285eba24c/types/share_merging.go#L43-L45
where honest nodes will disregard

The original thinking of why it wasn't added is non-intuitive. Until we have immediate execution, checking that the PFD is valid doesn't really do anything. Any validator can include invalid PFD (ie a bad nonce, or an account with zero funds to pay fees) without consequence other than people yelling at them on Twitter. The only way to actually check for these things is by adding block validity rule via process proposal, where we'd have to throw away the entire block should this occur. If we do that, then we also want to add a check to PrepareProposal to ensure that this type of tx doesn't accidently get in the block, which would cause it to be invalid. This might already be occurring implicitly too. If someone tries to add a message to an in use important namespace, such as the TxNamespacedID ([0,0,0,0,0,0,0,1]), then this will cause an share splitting/merging error, which would cause ProcessProposal to reject the block.

We might want to do this and add block validity mechanisms!! More thinking on this is definitely worth while.

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/570#discussion_r930870919_

PR #592, opened 2022-08-04
feat: Add namespace id checks for messages

SweeXordious:
Closes: https://github.com/celestiaorg/celestia-app/issues/571 and https://github.com/celestiaorg/celestia-app/issues/596

### DIFF
--- a/app/split_shares.go
+++ b/app/split_shares.go
@@ -60,14 +60,15 @@ func SplitShares(txConf client.TxConfig, squareSize uint64, data *core.Data) ([]
 		}
 
 		msg := authTx.GetMsgs()[0]
-		// run basic validation on the message
-		err = msg.ValidateBasic()
-		if err != nil {
-			continue
-		}
 
 		wireMsg, ok := msg.(*types.MsgWirePayForData)
 		if !ok {
+			continue
+		}
+
+		// run basic validation on the message
+		err = wireMsg.ValidateBasic()
+		if err != nil {
 			continue
 		}
 
--- a/app/split_shares.go
+++ b/app/split_shares.go
@@ -60,6 +60,12 @@ func SplitShares(txConf client.TxConfig, squareSize uint64, data *core.Data) ([]
 		}
 
 		msg := authTx.GetMsgs()[0]
+		// run basic validation on the message
+		err = msg.ValidateBasic()
+		if err != nil {
+			continue
+		}
+
 		wireMsg, ok := msg.(*types.MsgWirePayForData)
 		if !ok {
 			continue
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -121,7 +121,7 @@ func TestPrepareMessagesWithReservedNamespaces(t *testing.T) {
 		{"evidence namespace id for message", consts.EvidenceNamespaceID, 0},
 		{"tail padding namespace id for message", consts.TailPaddingNamespaceID, 0},
 		{"parity shares namespace id for message", consts.ParitySharesNamespaceID, 0},
-		{"namespace id 200 for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, 0},
+		{"reserved namespace id for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, 0},
 		{"correct namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, 1},
 	}
 
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -122,7 +122,7 @@ func TestPrepareMessagesWithReservedNamespaces(t *testing.T) {
 		{"tail padding namespace id for message", consts.TailPaddingNamespaceID, 0},
 		{"parity shares namespace id for message", consts.ParitySharesNamespaceID, 0},
 		{"reserved namespace id for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, 0},
-		{"correct namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, 1},
+		{"valid namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, 1},
 	}
 
 	for _, tt := range tests {
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -8,12 +8,14 @@ import (
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/testutil"
 	"github.com/celestiaorg/celestia-app/x/payment/types"
+	"github.com/celestiaorg/nmt/namespace"
 	"github.com/cosmos/cosmos-sdk/client"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/pkg/consts"
 	core "github.com/tendermint/tendermint/proto/tendermint/types"
 )
 
@@ -99,6 +101,39 @@ func TestPrepareProposal(t *testing.T) {
 			)
 			assert.NoError(t, err)
 		}
+	}
+}
+
+func TestPrepareMessagesWithReservedNamespaces(t *testing.T) {
+	testApp := testutil.SetupTestAppWithGenesisValSet(t)
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+
+	signer := testutil.GenerateKeyringSigner(t, testAccName)
+
+	type test struct {
+		name             string
+		namespace        namespace.ID
+		expectedMessages int
+	}
+
+	tests := []test{
+		{"transaction namespace id for message", consts.TxNamespaceID, 0},
+		{"evidence namespace id for message", consts.EvidenceNamespaceID, 0},
+		{"tail padding namespace id for message", consts.TailPaddingNamespaceID, 0},
+		{"parity shares namespace id for message", consts.ParitySharesNamespaceID, 0},
+		{"namespace id 200 for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, 0},
+		{"correct namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, 1},
+	}
+
+	for _, tt := range tests {
+		tx := generateRawTx(t, encCfg.TxConfig, tt.namespace, []byte{1}, signer, 2, 4, 8, 16)
+		input := abci.RequestPrepareProposal{
+			BlockData: &core.Data{
+				Txs: [][]byte{tx},
+			},
+		}
+		res := testApp.PrepareProposal(input)
+		assert.Equal(t, tt.expectedMessages, len(res.BlockData.Messages.MessagesList))
 	}
 }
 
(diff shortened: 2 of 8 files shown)


SweeXordious:
@evan-forbes Are these changes enough for prepare/process proposal? So that I can start adding tests for that.

app/split_shares.go
```diff
@@ -60,11 +60,18 @@ func SplitShares(txConf client.TxConfig, squareSize uint64, data *core.Data) ([]
 		}
 
 		msg := authTx.GetMsgs()[0]
+
```
SweeXordious, line 63:
```suggestion
```

app/test/process_proposal_test.go
```diff
@@ -164,13 +165,76 @@ func TestMessageInclusionCheck(t *testing.T) {
...
+		name           string
+		namespace      namespace.ID
+		expectedResult abci.ResponseProcessProposal_Result
+	}
+
+	tests := []test{
+		{"transaction namespace id for message", consts.TxNamespaceID, abci.ResponseProcessProposal_REJECT},
+		{"evidence namespace id for message", consts.EvidenceNamespaceID, abci.ResponseProcessProposal_REJECT},
+		{"tail padding namespace id for message", consts.TailPaddingNamespaceID, abci.ResponseProcessProposal_REJECT},
+		//{"parity shares namespace id for message", consts.ParitySharesNamespaceID, abci.ResponseProcessProposal_REJECT},
```
SweeXordious, line 184: Issue:

app/test/process_proposal_test.go
```diff
@@ -164,13 +165,76 @@ func TestMessageInclusionCheck(t *testing.T) {
...
+		// this case is commented because the shares are ordered. Then, when adding a message with namespace consts.ParitySharesNamespaceID,
+		// it panics as it is higher than the maximum allowed namespace ID.
+		// issue:
```
SweeXordious, line 187:
```suggestion
		// issue: https://github.com/celestiaorg/celestia-app/issues/596
```

app/test/process_proposal_test.go
```diff
@@ -164,13 +165,76 @@ func TestMessageInclusionCheck(t *testing.T) {
...
+		res := testApp.ProcessProposal(input)
+		assert.Equal(t, tt.expectedResult, res.Result)
+	}
+}
+
 func genRandMsgPayForData(t *testing.T, signer *types.KeyringSigner, squareSize uint64) (*types.MsgPayForData, []byte) {
 	ns := make([]byte, consts.NamespaceSize)
 	_, err := rand.Read(ns)
 	require.NoError(t, err)
+	return genRandMsgPayForDataForNamespace(t, signer, squareSize, ns)
```
rootulp, line 232: 👍 great idea extracting this. The function name increases readability

app/test/prepare_proposal_test.go
```diff
@@ -102,6 +104,39 @@ func TestPrepareProposal(t *testing.T) {
...
+		namespace        namespace.ID
+		expectedMessages int
+	}
+
+	tests := []test{
+		{"transaction namespace id for message", consts.TxNamespaceID, 0},
+		{"evidence namespace id for message", consts.EvidenceNamespaceID, 0},
+		{"tail padding namespace id for message", consts.TailPaddingNamespaceID, 0},
+		{"parity shares namespace id for message", consts.ParitySharesNamespaceID, 0},
+		{"namespace id 200 for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, 0},
```
rootulp, line 124: [question] why does namespace id 200 have 0 messages? Is it because 200 is less than reserved namespace ID? If so, then maybe:

```suggestion
		{"reserved namespace id for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, 0},
```

app/test/prepare_proposal_test.go
```diff
@@ -102,6 +104,39 @@ func TestPrepareProposal(t *testing.T) {
...
+		{"correct namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, 1},
```
rootulp, line 125: [out of curiosity] how was this namespace ID determined? Is it just an arbitrary namespace ID that is considered a valid namespace ID because it's the right length and doesn't use any reserved namespaces?

[nit]

```suggestion
		{"valid namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, 1},
```

rootulp (approved):
LGTM. I proposed a few renames based on my understanding but all are optional. Defer to your judgement on what's clearer

app/test/process_proposal_test.go
evan-forbes: can we still check for that without actually causing the panic? I _think_ its possible is the message is placed perfectly as the last share in the original square.

SweeXordious: yes, it is actually possible. But  would make the test a bit ugly I guess.
Currently, the panic happens when calculating the data availability root for the input block. If we skip that step, it will catch the wrong namespace id. 
To do that, we can add a parameter to the test struct called `calculateDARoot` of type `bool`. 
Something like this:

```golang
	type test struct {
		name            string
		namespace       namespace.ID
		calculateDARoot bool
		expectedResult  abci.ResponseProcessProposal_Result
	}

	tests := []test{
		{"transaction namespace id for message", consts.TxNamespaceID, true, abci.ResponseProcessProposal_REJECT},
		{"evidence namespace id for message", consts.EvidenceNamespaceID, true, abci.ResponseProcessProposal_REJECT},
		{"tail padding namespace id for message", consts.TailPaddingNamespaceID, true, abci.ResponseProcessProposal_REJECT},
		{"parity shares namespace id for message", consts.ParitySharesNamespaceID, false, abci.ResponseProcessProposal_REJECT},
		// this case is commented because the shares are ordered. Then, when adding a message with namespace consts.ParitySharesNamespaceID,
		// it panics as it is higher than the maximum allowed namespace ID.
		// issue:
		{"namespace id 200 for message", namespace.ID{0, 0, 0, 0, 0, 0, 0, 200}, true, abci.ResponseProcessProposal_REJECT},
		{"correct namespace id for message", namespace.ID{3, 3, 2, 2, 2, 1, 1, 1}, true, abci.ResponseProcessProposal_ACCEPT},
	}

	for _, tt := range tests {
		pfd, msg := genRandMsgPayForDataForNamespace(t, signer, 8, tt.namespace)
		input := abci.RequestProcessProposal{
			BlockData: &core.Data{
				Txs: [][]byte{
					buildTx(t, signer, encConf.TxConfig, pfd),
				},
				Messages: core.Messages{
					MessagesList: []*core.Message{
						{
							NamespaceId: pfd.GetMessageNamespaceId(),
							Data:        msg,
						},
					},
				},
				OriginalSquareSize: 8,
			},
		}
		if tt.calculateDARoot { <= here
			data, err := coretypes.DataFromProto(input.BlockData)
			require.NoError(t, err)

			shares, _, err := data.ComputeShares(input.BlockData.OriginalSquareSize)
			require.NoError(t, err)

			rawShares := shares.RawShares()

			require.NoError(t, err)
			eds, err := da.ExtendShares(input.BlockData.OriginalSquareSize, rawShares)
			require.NoError(t, err)
			dah := da.NewDataAvailabilityHeader(eds)
			input.Header.DataHash = dah.Hash()
		}
		res := testApp.ProcessProposal(input)
		assert.Equal(t, tt.expectedResult, res.Result)
	}
```

SweeXordious: if you like it this way, I can update the test

app/test/prepare_proposal_test.go
SweeXordious: @rootulp yes 100%. We could also fuzz all reserved namespaces instead of providing only an example, etc. But, I guess it is too much for this. 
However, if it makes sense, we can create an issue to improve the test.

app/test/process_proposal_test.go
evan-forbes: nah, as you mentioned in https://github.com/celestiaorg/celestia-app/issues/596#issuecomment-1205785617, this specific case isn't reachable

SweeXordious: But now that I think about it, I guess it's worth it. So that if someday, for some reason, we're no longer checking for parity shares namespace, we have this test to know exactly what's the issue?
I am okey with both, whichever you prefer.

x/payment/types/wirepayfordata.go
```diff
@@ -139,6 +121,43 @@ func (msg *MsgWirePayForData) ValidateBasic() error {
...
+
+	// ensure that TxNamespaceID is not used
+	if bytes.Equal(ns, consts.TxNamespaceID) {
+		return ErrTxNamespace
+	}
+
+	// ensure that EvidenceNamespaceID is not used
+	if bytes.Equal(ns, consts.EvidenceNamespaceID) {
+		return ErrEvidenceNamespace
+	}
```
evan-forbes, line 156: these checks are redundant to the one on line 134
```suggestion
```

x/payment/types/wirepayfordata.go
```diff
@@ -84,12 +85,8 @@ func (msg *MsgWirePayForData) Route() string { return RouterKey }
...
 // interface.
 func (msg *MsgWirePayForData) ValidateBasic() error {
-	// ensure that the namespace id is of length == NamespaceIDSize
-	if nsLen := len(msg.GetMessageNameSpaceId()); nsLen != NamespaceIDSize {
-		return ErrInvalidNamespaceLen.Wrapf("got: %d want: %d",
-			nsLen,
-			NamespaceIDSize,
-		)
+	if err := ValidateMessageNamespaceID(msg.GetMessageNameSpaceId()); err != nil {
+		return err
```
evan-forbes, line 89: :+1:

evan-forbes (approved):
found a redundant check, but pre-approving

app/test/process_proposal_test.go
evan-forbes: > But now that I think about it, I guess it's worth it. So that if someday, for some reason, we're no longer checking for parity shares namespace, we have this test to know exactly what's the issue?

that's another great point. before reading your comment in #596, I was going to suggest that we split it off into its own test for this exact reason. Its a super explicit reminder of this edge case for the future generations to remember.

evan-forbes: then we can leave the original test nice and clleeeaaaannnn

SweeXordious: sounds awesome. Will do.

x/payment/types/wirepayfordata.go
```diff
@@ -139,6 +121,34 @@ func (msg *MsgWirePayForData) ValidateBasic() error {
...
+	if bytes.Equal(ns, consts.ParitySharesNamespaceID) {
+		return ErrParitySharesNamespace
+	}
+
+	// ensure that TailPaddingNamespaceID is not used
+	if bytes.Equal(ns, consts.TailPaddingNamespaceID) {
+		return ErrTailPaddingNamespace
+	}
+
+
```
SweeXordious, line 148:
```suggestion
```

app/test/prepare_proposal_test.go
```diff
@@ -102,6 +104,39 @@ func TestPrepareProposal(t *testing.T) {
 	}
 }
 
+func TestPrepareMessagesWithReservedNamespaces(t *testing.T) {
```
evan-forbes, line 107: thanks for remembering this, I forgot lol

evan-forbes (approved):
:100: :1st_place_medal: :fire:

evan-forbes:
just noticed that we are running `ValidateBasic` for the wirePFD twice during PrepareProposal, once for the message itself, and another time when we call `ValidateBasic` for the entire tx directly after as well. I like the explicitness of calling it directly, as we are relient upon the sdk always calling `ValidateBasic` for the msg for us, but :shrug:

SweeXordious:
Just checked it, they are not the same `ValidateBasic`. One is the one we added, and the other one is this one:

https://github.com/celestiaorg/cosmos-sdk/blob/c20a3b1d3d68497f8e8fe8c1e20a12190246939f/types/tx/types.go#L34-L96

So, I guess it is good, no? or I am missing something.

Issue #499, opened 2022-06-23
QGB use app config to register modules 


SweeXordious:
As discussed under _https://github.com/celestiaorg/celestia-app/pull/497#discussion_r905569617_, we  should switch to using the existing encoding infrastructure to register the modules.  
This can be done when we merge the qgb integration branch to master.

Issue #517, opened 2022-07-04
QGB resolve merge conflicts with master


SweeXordious:
The qgb-integration branch currently has merge conflicts with the master branch. These will need to be resolved before shipping.

PR #600, opened 2022-08-07
Update the QGB branch to use cosmos-sdk v0.46.x and latest celestia-app master 

SweeXordious:
Update the QGB branch to use cosmos-sdk v0.46.x and latest celestia-app master.

Closes: https://github.com/celestiaorg/celestia-app/issues/499, https://github.com/celestiaorg/celestia-app/issues/517


evan-forbes (commented):
mind moving things like the cosmetics to a different PR? normally I think its fine, but in this case there are already a ton of changes. da6835d93efeebd0cb1f1b5bb22ca3d0e35094b9

are you wanting to preserve the git history of the main branch to on the feature branch (this will effect how we are able to merge it in the future), if so, we need to not squash merge this. if we can't squash merge because we have to preserve git history, do you mind cleaning up the changes you made ontop of the commits merged?

as a side note, we need to come up with a plan to merge the qgb integration branch, purely from a git perspective. We want to keep track of all the work we've (mainly you lol) done, while keeping things semi concise, while also not disturbing the git history of the main branch in a confusing way. Ideally, we would be able to incorporate all or most of the commits we've made on the feature branch alongside the commits that were made on the main branch. This also depends on if we want to/can spin the orchestrator and relayer into their own repo.

SweeXordious:
@evan-forbes Sure, but the thing is, if I remove the cosmetics and the fixes, the CI will not pass.
So, I guess we have the following options:
- Squash all  the last 4 commits into 1 to have a clean history
- Create a PR  targeting this branch having only those CI fixes so you can review them. Then, proceed  to review  this one.
- Merge this one without the CI passing and the project not building.

For the merge, we need to sit down and see what we want to do with the QGB. Whether we want a separate repo for orchestrator and relayer or keep them here.

evan-forbes:
Ahh I see I didn't know that it wouldn't pass. Hmm, ok then it's probably fine. Maybe just clean them up slightly like rename them or something. 

I'm definitely fine with keeping them (or squashing them)

SweeXordious:
@evan-forbes I just separated them into 2 chained PRs:
- This one should be reviewed first, contains QGB changes: https://github.com/celestiaorg/celestia-app/pull/602
- This one contains the changes from main: https://github.com/celestiaorg/celestia-app/pull/601

PR #601, opened 2022-08-08
Update the QGB branch to use cosmos-sdk v0.46.x and latest celestia-app master

SweeXordious:
Update the QGB branch to use cosmos-sdk v0.46.x and latest celestia-app master.

Opened to separate the PRs as of: https://github.com/celestiaorg/celestia-app/pull/600 and https://github.com/celestiaorg/celestia-app/pull/600#pullrequestreview-1064491249

Closes: https://github.com/celestiaorg/celestia-app/issues/517

PR #602, opened 2022-08-08
Update QGB implementation to support cosmos-sdk  v0.46.x

SweeXordious:
Fix the implementation in the branch.

Closes: https://github.com/celestiaorg/celestia-app/issues/499


app/app.go
```diff
@@ -367,7 +367,7 @@ func New(
 	app.PaymentKeeper = *paymentmodulekeeper.NewKeeper(
 		appCodec,
 	)
-	paymentmodule := paymentmodule.NewAppModule(appCodec, app.PaymentKeeper)
+	paymentmod := paymentmodule.NewAppModule(appCodec, app.PaymentKeeper)
```
evan-forbes, line 370: this appears to be because we're overriding the import, which makes a lot of sense to change. Do we have extra CI here that we don't have on the main branch?

SweeXordious: it's better for readability I guess and I have extra linters on my IDE. I can revert if not needed.

evan-forbes: no we should change this upstream too, overriding the import is bad (harmless in this case, but bad).

evan-forbes (approved):

SweeXordious:
Won't merge this one until we also review this: https://github.com/celestiaorg/celestia-app/pull/601

Issue #604, opened 2022-08-09
Investigate the QGB relayer signatures issue  


SweeXordious:
### Summary of Bug

When running the QGB, sometimes, the relayer fails to relay attestations and reverts here:
https://github.com/celestiaorg/quantum-gravity-bridge/blob/37b613843c7a0681fe2a8ae82d3422e1f1cfb311/src/QuantumGravityBridge.sol#L220

This should be investigated. A possible problem could be that the signatures are not ordered correctly when creating the sigs array in the relayer. 

### Version

https://github.com/celestiaorg/celestia-app/commit/2fef6fd9fe4bd34e1bb9edd6bc76c47c1d7e13fc

### Steps to Reproduce

Run the QGB e2e tests long enough

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SweeXordious:
Update: The commitments queried from core are not deterministic. Will start investigating that.

SweeXordious:
The issue was a race condition.
Sometimes the orchestrator is getting the commitment for a range where the last block is still not indexed.
So, we only get the commitment over the ones that are indexed.
Example:
- Height 30 is committed to
- orchestrator asks for commitment [10, 30]
- Height 30 is indexed
- The commitment received will be on [10, 29]
https://github.com/celestiaorg/celestia-core/pull/838 fixes it.

PR #605, opened 2022-08-10
QGB Use Ecrecover to validate Ethereum signatures

SweeXordious:
Updates the implementation of the Ethereum signer.

Contributes to https://github.com/celestiaorg/celestia-app/issues/604


SweeXordious:
reverting after 179 attestation. Converting to draft.

evan-forbes:
we still might want to merge this even if it isn't the (sole) source of the issue described in #604. there's no rush tho, we can wait until we figure it out.

SweeXordious:
The next issue found is that the commitments queried from Celestia-core are not deterministic. Will investigate that separately.
Aside from that, we need the changes we have in this PR. So, we can merge if approved.

x/qgb/types/ethereum_signer.go
```diff
@@ -18,40 +17,24 @@ func NewEthereumSignature(hash []byte, privateKey *ecdsa.PrivateKey) ([]byte, er
...
+		return common.Address{}, sdkerrors.Wrap(err, "ec recover failed")
 	}
-
-	protectedHash := crypto.Keccak256Hash(append([]uint8(signaturePrefix), hash...))
-
-	pubkey, err := crypto.SigToPub(protectedHash.Bytes(), signature)
+	pubKey, err := crypto.UnmarshalPubkey(sigPublicKey)
 	if err != nil {
-		return common.Address{}, sdkerrors.Wrap(err, "signature to public key")
+		return common.Address{}, sdkerrors.Wrap(err, "unmarshall signture public key failed")
```
SweeXordious, line 35:
```suggestion
		return common.Address{}, sdkerrors.Wrap(err, "unmarshalling signature' public key failed")
```

x/qgb/types/ethereum_signer.go
```diff
@@ -18,40 +17,24 @@ func NewEthereumSignature(hash []byte, privateKey *ecdsa.PrivateKey) ([]byte, er
...
-	// as 27 or 28, internally though it should be a 0-3 value due to changed formats.
-	// It seems that go-ethereum expects this to be done before sigs actually reach it's
-	// internal validation functions. In order to comply with this requirement we check
-	// the sig an dif it's in standard format we correct it. If it's in go-ethereum's expected
-	// format already we make no changes.
-	//
-	// We could attempt to break or otherwise exit early on obviously invalid values for this
-	// byte, but that's a task best left to go-ethereum
-	if signature[64] == 27 || signature[64] == 28 {
-		signature[64] -= 27
```
rootulp, line 44: Out of curiosity, why is this no longer necessary?

rootulp (commented):
For my learning: what was wrong with the previous implementation? 
Related: Do we want a unit test that prevents a future `EthAddressFromSignature` regression?

x/qgb/types/ethereum_signer.go
SweeXordious: https://twitter.com/alexberegszaszi/status/1534461421454606336?t=6hUf4_GjhEuFBVdj_Kqzew&s=19

SweeXordious:
@rootulp Nothing is wrong with the implementation, it was alright. I was suspecting it to be the reason why the relayer was failing. But, turned out to be the commitments.
However, I wanted to be the same as what we are using on the smart contract side:
https://github.com/celestiaorg/quantum-gravity-bridge/blob/37b613843c7a0681fe2a8ae82d3422e1f1cfb311/src/QuantumGravityBridge.sol#L140-L148

x/qgb/types/ethereum_signer.go
rootulp: wow so a bunch of smart contracts include this check even though it's unnecessary

rootulp:
> Nothing is wrong with the implementation, it was alright

Thanks for explaining! So no changes were strictly needed in `x/qgb/types/ethereum_signer_test.go`? On first pass, I thought something about the implementation changed b/c values used in the test were updated.

x/qgb/types/ethereum_signer.go
SweeXordious: yes. And also in our case we were adding this check while it's not needed.

SweeXordious:
@rootulp The different values in tests are because we removed the signatures whos `v` is either `27` or `28`.
The reason is that we no longer need to support that legacy change:
https://ethereum.stackexchange.com/a/113505/65649

adlerjohn (approved):
utACK

evan-forbes (approved):
:+1:

SweeXordious:
Will not merge until I get to the bottom of the commitments issue.

PR #613, opened 2022-08-12
QGB Relayer add commitments check when relaying confirms

SweeXordious:
Adds a check on the confirms' commitments before relaying them

### DIFF
--- a/x/qgb/orchestrator/relayer.go
+++ b/x/qgb/orchestrator/relayer.go
@@ -147,22 +147,6 @@ func (r *Relayer) submitDataRootTupleRoot(
 	commitment string,
 	confirms []types.MsgDataCommitmentConfirm,
 ) error {
-	// sanitizing the confirms commitments
-	for _, conf := range confirms {
-		if conf.Commitment != commitment {
-			r.logger.Error("confirms don't have the same commitment",
-				"expected_commitment",
-				commitment,
-				"confirm_commitment",
-				conf.Commitment,
-				"nonce",
-				conf.Nonce,
-				"signer_eth_address",
-				conf.EthAddress)
-			return ErrInvalidCommitmentInConfirm
-		}
-	}
-
 	sigsMap := make(map[string]string)
 	// to fetch the signatures easilly by eth address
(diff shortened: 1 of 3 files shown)


adlerjohn (approved):
utACK

evan-forbes (approved):
:+1: :+1:

SweeXordious:
Will not merge until I get to the bottom of the commitments issue.

x/qgb/orchestrator/relayer.go
```diff
@@ -147,6 +147,22 @@ func (r *Relayer) submitDataRootTupleRoot(
...
+			r.logger.Error("confirms don't have the same commitment",
+				"expected_commitment",
+				commitment,
+				"confirm_commitment",
+				conf.Commitment,
+				"nonce",
+				conf.Nonce,
+				"signer_eth_address",
+				conf.EthAddress)
+			return ErrInvalidCommitmentInConfirm
```
SweeXordious, line 162: @evan-forbes What do you think makes more sense:
- not relay attestations where one or more data commitments have an invalid commitment?
- relay if we have enough data commitments that are good (without considering the faulty ones).

I guess we can go for second option for now. 
What do you think?

evan-forbes: Agree that option 2 is probably best. We want to ignore potentially malicious validators. In the future, we will slash for this, as they sign an incorrect data commitment.

SweeXordious: Awesome. Thanks a lot.

SweeXordious:
We could even increase the validations we do on the commitments before adding them to make the relayer smarter and not just keep sending transactions to the EVM chain that we know won't pass.
What do you think? Should we create an issue for that?

evan-forbes:
> We could even increase the validations we do on the commitments before adding them to make the relayer smarter and not just keep sending transactions to the EVM chain that we know won't pass.
What do you think? Should we create an issue for that?

hey sorry I missed this. Yeah, this sounds good! Perhaps this should be covered in ADR 005?

Issue #617, opened 2022-08-15
Use nmt wrapper when creating commitments for PayForData transactions


evan-forbes:
currently, we're not using the nmt wrapper when generating commitments for pfd transactions. This is problematic because the wrapper is used when creating the data commitments, so when we modify the message inclusion check to actually use subtree roots, it won't be able to find them.

Also, we need the payment module to be more or less stable so that the audit process can begin.

distractedm1nd:
An ErrInvalidShareCommit (11116) is thrown occasionally when a node running an earlier app version (v0.7.0-rc-1) submits a pfd to an app with #618.

PR #618, opened 2022-08-15
fix: use wrapped nmt when generating commitments

evan-forbes:
## Description

swap to using the wrapped erasure merkle tree when generating commitments for PFD transactions

note: After this, we should try to avoid any other changes to payment module for the time being, as it will be under audit.

closes: #617


x/payment/types/payfordata_test.go
```diff
@@ -157,7 +157,7 @@ func TestCreateCommitment(t *testing.T) {
 			k:         4,
 			namespace: bytes.Repeat([]byte{0xFF}, 8),
 			message:   bytes.Repeat([]byte{0xFF}, 11*ShareSize),
-			expected:  []byte{0xf2, 0xd4, 0xfc, 0x39, 0x4e, 0xf3, 0x97, 0x9d, 0xf4, 0x4c, 0x99, 0x87, 0x36, 0x7d, 0x7d, 0x4, 0xf2, 0xa7, 0x89, 0x26, 0x6d, 0xf5, 0x78, 0xe1, 0xff, 0x72, 0xb4, 0x75, 0x12, 0x1e, 0x71, 0xc3},
+			expected:  []byte{0xe3, 0xaa, 0xdb, 0x1c, 0x2e, 0xc3, 0x77, 0x4a, 0x46, 0xb6, 0xb2, 0xf0, 0x4f, 0x6b, 0x3d, 0x4f, 0x17, 0x99, 0x82, 0x6b, 0xd6, 0xd4, 0x40, 0x5e, 0xbe, 0x1b, 0x1f, 0x29, 0x5a, 0x53, 0x3f, 0xbb},
```
rootulp, line 160: [no change needed] it's difficult to determine if this change is desirable but we already have a [todo](https://github.com/celestiaorg/celestia-app/blob/320ab5be792199b294845ee57afc01d84659b955/x/payment/types/payfordata_test.go#L146) to fix.

x/payment/types/payfordata.go
```diff
@@ -155,16 +155,13 @@ func CreateCommitment(k uint64, namespace, message []byte) ([]byte, error) {
...
+		tree := wrapper.NewErasuredNamespacedMerkleTree(consts.MaxSquareSize)
 		for _, leaf := range set {
 			nsLeaf := append(make([]byte, 0), append(namespace, leaf...)...)
-			err := tree.Push(nsLeaf)
-			if err != nil {
-				return nil, err
-			}
+			// here we hardcode pushing as axis 0 cell 0 because we never want
+			// to add the parity namespace to our shares when we create roots.
+			tree.Push(nsLeaf, rsmt2d.SquareIndex{Axis: 0, Cell: 0})
```
rootulp, line 163: [question] it's not clear to me how this is related to https://github.com/celestiaorg/celestia-core/blob/3c6122ae98f3ea98d4885a986d3e8d5ec1077d45/pkg/consts/consts.go#L69

Can you please elaborate?

x/payment/types/payfordata_test.go
evan-forbes: we should expect the commitment to change since we are adding bytes the leafs and inner nodes. This test is basically just to make sure that we don't accidentally do that. So its desirable that something happened, but I obvi didn't manually calculate this myself, so I'm unsure if its 100% what we want.

x/payment/types/payfordata.go
evan-forbes: yeah, so we have to ensure that the namespace is included in erasure data. To do that, we have to include the namespace in the share. This means that the erasure data will not have a namspace, so efffectively this wrapper adds the correct namespace to the data so that nmt knows how to interpret it. 

here we add the parity namespace to a share, depending on its position
https://github.com/celestiaorg/celestia-core/blob/08c0c038e63a7848f2761beb1013ee102883be98/pkg/wrapper/nmt_wrapper.go#L58

x/payment/types/payfordata_test.go
evan-forbes: by manually calculate, I mean like, literally hand calculate the values. I did generate the commitment by hand double check the work.

x/payment/types/payfordata.go
evan-forbes: by hardcoding 0,0, we're forcing it to use the namespace we want (the namespace of the share) instead of potentially going over the square size and then accidentally using an incorrect namespace.

rootulp: thanks for explaining! It seems like this line is tightly coupled to an implementation detail of https://github.com/celestiaorg/celestia-core/blob/08c0c038e63a7848f2761beb1013ee102883be98/pkg/wrapper/nmt_wrapper.go#L50

To make this more explicit, I wonder if we may add a method in a subsequent refactor:

```golang
// PushWithNamespace adds the provided data to the underlying
// NamespaceMerkleTree and preserves the first DefaultNamespaceIDLen bytes as
// the namespace for this data. Fulfills the rsmt.Tree interface. NOTE: panics
// if an error is encountered while pushing or if the tree size is exceeded.
func (w *ErasuredNamespacedMerkleTree) PushWithNamespace(data []byte) {
	nidAndData := make([]byte, consts.NamespaceSize+len(data))
	copy(nidAndData[consts.NamespaceSize:], data)
	copy(nidAndData[:consts.NamespaceSize], data[:consts.NamespaceSize])
	// push to the underlying tree
	err := w.tree.Push(nidAndData)
	// panic on error
	if err != nil {
		panic(err)
	}
}
```

x/payment/types/payfordata_test.go
rootulp: > I did generate the commitment by hand double check the work.

can you share how I would do this? If the expected bytes are indeed what we want, then approving to unblock

evan-forbes: yeah, we can be super lazy and just run this test since its creating the commitment for us, or we can write a quick script that does the same

x/payment/types/payfordata.go
evan-forbes: yeah, I like this idea. I'd also be down for refactoring the wrapped nmt into something else entirely. It relies on rsmt2d to call it, which we don't actually have to do. During nmt cacheing to check for message inclusion, this makes things even more confusing and suboptimal as we have to make assumtions over when and how rsmt2d actually calls the tree and generates commitments.

a while ago we considered not passing some nmt (or wrapped nmt) object to rsmt, and instead simply reading the data we need as we need it. https://github.com/celestiaorg/rsmt2d/pull/9#issuecomment-736805729 . We don't actually need to change anything in rsmt2d for this, we can just code it up.

evan-forbes:
merging w/o further feedback as this is a smaller change and we need a stable version of the payment module. We can still continue conversations here and open new issues PRs if neccessary :slightly_smiling_face:

PR #621, opened 2022-08-17
feat: add path generation for nmt subroots

evan-forbes:
## Description

This PR adds code that generates paths to nmt sub tree roots. While we also have code for this funtionality [in the nmt repo](https://github.com/celestiaorg/nmt/blob/master/subrootpaths.go), that code returns the paths that could include multiple rows, with out providing info as to which paths belonged to which row (afaict). Changing that was difficult, and requires encoding rsmt2d logic (rows) into nmt, so I decided to rewrite the functionality here given the simplicity of it.

part of #381


pkg/inclusion/paths.go
```diff
@@ -0,0 +1,113 @@
+package inclusion
+
+// genSubTreeRootPath calculates the path to a given subtree root of a node, given the
+// depth and position of the node. note: the root of the tree is depth 0.
+// nolint
```
rootulp, line 5: > Use //nolint instead of // nolint because machine-readable comments should have no space by Go convention.

From: https://golangci-lint.run/usage/false-positives/

Proposal to disable the explicit linters that are in violation rather than all linters so that we get linter warnings if a separate violation (that we don't expect) is flagged in this code block.

[Optional] to include a note on why so that maintainers know when this `//nolint` can be removed

```suggestion
// The following nolint can be removed after this function is used.
//nolint:unused,deadcode
```

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,113 @@
...
+		position: c.position / 2,
+	}
+}
+
+// canClimbRight uses the current position to calculate the direction of the next
+// climb. Returns true if the next climb is right (if the position (index) is
+// even).
+func (c coord) canClimbRight() bool {
+	return c.position%2 == 0
+}
```
rootulp, line 41: [no change needed] one alternative way of describe this relationship is parent / child 

```golang
// getParent returns the coordinate for the parent of the current coordinate.
func (c coord) getParent() coord {
	return coord{
		depth:    c.depth - 1,
		position: c.position / 2,
	}
}

// isLeftChild returns ture if the coordinate are the left child of a parent.
func (c coord) isLeftChild() bool {
	return c.position%2 == 0
}
```

but I think the climb terminology makes more sense given it's usage below.

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,113 @@
...
+
+// calculateSubTreeRootCoordinates generates the sub tree root coordinates of a
+// set of shares for a balanced binary tree of a given depth. It assumes that
+// end does not exceed the range of a tree of the provided depth, and that end
+// >= start. This function works by starting at the first index of the msg and
+// working our way right.
+func calculateSubTreeRootCoordinates(maxDepth, start, end uint64) []coord {
+	cds := []coord{}
+	// leafCursor keeps track current leaf that we are starting with when
```
rootulp, line 50:
```suggestion
	// leafCursor keeps track of the current leaf that we are starting with when
```

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,113 @@
...
+	// finding the subtree root for some set. When leafCursor == end, we are
+	// finished calculating sub tree roots
+	leafCursor := start
+	// nodeCursor keeps track of the current node tree node when finding sub
```
rootulp, line 54:
```suggestion
	// nodeCursor keeps track of the current tree node when finding sub
```

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,113 @@
...
+	// tree roots
+	nodeCursor := coord{
+		depth:    maxDepth,
+		position: start,
+	}
+	// lastNodeCursor keeps track of the last node cursor so that when we climb
+	// to high, we can use this node as a sub tree root
```
rootulp, line 61:
```suggestion
	// too high, we can use this node as a sub tree root
```

pkg/inclusion/paths_test.go
```diff
@@ -0,0 +1,229 @@
+package inclusion
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_calculateSubTreeRootCoordinates(t *testing.T) {
```
rootulp, line 9: 👍

rootulp (approved):

pkg/inclusion/paths.go
evan-forbes: >but I think the climb terminology makes more sense given it's usage below

yeah, this is precisely the reason why I initially went with climb.

rootulp (dismissed):

pkg/inclusion/paths_test.go
rootulp: [for other reviewers] I had a tough time verifying the 8 leaf tests but drawing it out helped me so leaving here in case helpful

03CD7E57-681D-498F-8959-A2D46D2FA9A0_1_102_a

evan-forbes: yes! thank you!!

need an ascii version of this in the docs

rootulp: yes ser

```ascii
Depth       Position
0              0
              / \
             /   \
1           0     1
           /\     /\
2         0  1   2  3
         /\  /\ /\  /\ 
3       0 1 2 3 4 5 6 7
```

SweeXordious (commented):
first pass, looks good. But, I will still need to  give  it  more time to understand.

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,123 @@
...
+// climb is a state transition function to simulate climbing a balanced binary
+// tree, using the current node as input and the next highest node as output.
+func (c coord) climb() coord {
+	return coord{
+		depth:    c.depth - 1,
+		position: c.position / 2,
+	}
+}
+
+// canClimbRight uses the current position to calculate the direction of the next
```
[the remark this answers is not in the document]
rootulp: Like you mentioned, with climb terminology, one could think a child node climbs **up** to a parent node (i.e. only one direction). But another way of thinking about this method: `isLeftChild()`.

Maybe this comment helps? https://github.com/celestiaorg/celestia-app/pull/621/files/d2d645b024464bcbcfd7114b42e9e154f77729cf#r948000052

rootulp (approved):

SweeXordious (dismissed):
LGTM. Great work :rocket:

evan-forbes:
thanks everyone for the great feedback on this! I'll try to incorporate all of them today, I've still been debugging the last e2e on non-interactives defaults

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,123 @@
+package inclusion
+
+// genSubTreeRootPath calculates the path to a given subtree root of a node, given the
+// depth and position of the node. note: the root of the tree is depth 0.
+// The following nolint can be removed after this function is used.
+//nolint:unused,deadcode
+func genSubTreeRootPath(depth int, pos uint) []bool {
+	path := make([]bool, depth)
+	for i := depth; i >= 0; i-- {
```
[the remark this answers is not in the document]
evan-forbes: good catch! the implementation changed as this was one of the bugs
https://github.com/celestiaorg/celestia-app/pull/621/commits/9cf9f545b5fb6ce6543a9e6f41b93f073b881805

there's a test specifically for this now too

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,123 @@
+package inclusion
+
+// genSubTreeRootPath calculates the path to a given subtree root of a node, given the
+// depth and position of the node. note: the root of the tree is depth 0.
+// The following nolint can be removed after this function is used.
+//nolint:unused,deadcode
```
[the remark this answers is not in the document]
evan-forbes: yes, we should! cherry picked this from the implementation https://github.com/celestiaorg/celestia-app/pull/621/commits/9cf9f545b5fb6ce6543a9e6f41b93f073b881805

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,123 @@
...
+type coord struct {
+	// depth is the typical depth of a tree, 0 being the root
+	depth uint64
+	// position is the index of a node of a given depth, 0 being the left most
+	// node
+	position uint64
+}
+
+// climb is a state transition function to simulate climbing a balanced binary
+// tree, using the current node as input and the next highest node as output.
```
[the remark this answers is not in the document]
evan-forbes: I don't think being able to climb right is a technically a precondition? or perhaps I'm just not clear on what you mean by this? the implementation that uses this code only climbs if it can, but is that different?

pkg/inclusion/paths.go
```diff
@@ -0,0 +1,123 @@
...
+		depth:    c.depth - 1,
+		position: c.position / 2,
+	}
+}
+
+// canClimbRight uses the current position to calculate the direction of the next
+// climb. Returns true if the next climb is right (if the position (index) is
+// even).
+func (c coord) canClimbRight() bool {
+	return c.position%2 == 0
```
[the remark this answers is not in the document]
evan-forbes: you're right, we probably should just to avoid potential leaky abstractions (although I hope others do not use coord for anything :sweat_smile:) as we're relying on the implementation to do this for us atm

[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/621/commits/32918de7e0959e5f647c2639ee114adfd15f1217

rootulp (approved):
LGTM! Great work

Issue #624, opened 2022-08-17
Move scripts from contrib


rootulp:
### Summary

Move scripts from contrib

### Problem Definition

This repo contains a `contrib/` directory that contains scripts. This location seems non-intuitive because there is also a root-level `scripts/` directory

### Proposal

1. Move all scripts to `scripts/`
2. (Optional) Delete all unneeded scripts

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #625, opened 2022-08-17
chore: move scripts from contrib to scripts/

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/624

I moved scripts from `contrib/` to `scripts/` and updated references to these scripts

## Testing
`make test-cover` runs the script


.gitignore
```diff
@@ -6,3 +6,4 @@ build
 coverage.txt
 tools-stamp
 .vscode
+profile.out
```
rootulp, line 9: This is a temp file generated by `test_cover.sh`

docker/Dockerfile.test
```diff
@@ -15,7 +15,7 @@ COPY . .
```
rootulp, line 15: `contrib/` seems like the wrong location for a Dockerfile considering we have `docker/` so I moved this too.

This file may be entirely removable. Defer to reviewers if that's the case.

evan-forbes (approved):
lgtm

Issue #623, opened 2022-08-17
Move share encoding and decoding logic to the app


evan-forbes:
Since we started performing all data root generation logic to the app, we don't actually need that logic to be in celestia-core anymore. Moving this code to the app will allow us to significantly reduce our diff with upstream tendermint.

NOTE: the non-interactive defaults feature branch has already done this, but it required a slight refactor. Before refactoring that code, we should simply copy it over with as few modifications as possible so that we can better review the refactor of that code.

PR #627, opened 2022-08-17
chore: move share encoding logic from core -> app

evan-forbes:
## Description

This PR moves the share encoding code from core to the app. It makes as few changes as possible, as to make the upcoming refactor for non-interactive defaults actually reviewable.

the code originally came from: 
https://github.com/celestiaorg/celestia-core/blob/5c844685421fcb0ff328bba1752d080420002f41/types/shares.go 
https://github.com/celestiaorg/celestia-core/blob/5c844685421fcb0ff328bba1752d080420002f41/types/shares_test.go
https://github.com/celestiaorg/celestia-core/blob/5c844685421fcb0ff328bba1752d080420002f41/types/share_splitting.go
https://github.com/celestiaorg/celestia-core/blob/5c844685421fcb0ff328bba1752d080420002f41/types/share_merging.go

closes: #623

### DIFF
--- a/pkg/shares/shares_test.go
+++ b/pkg/shares/shares_test.go
@@ -1,0 +1,573 @@
+package types
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"math"
+	"math/rand"
+	"reflect"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/celestiaorg/rsmt2d"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/pkg/consts"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+type Splitter interface {
+	SplitIntoShares() NamespacedShares
+}
+
+// func TestMakeShares(t *testing.T) {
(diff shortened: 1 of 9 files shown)


adlerjohn (approved):
utACK. Can you add to the PR description a permalink to the source of the code?

pkg/shares/shares_test.go
```diff
@@ -0,0 +1,573 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/pkg/consts"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+type Splitter interface {
+	SplitIntoShares() NamespacedShares
+}
+
+// func TestMakeShares(t *testing.T) {
```
SweeXordious, line 25: Why is this test commented?

SweeXordious (approved):
utAck

pkg/shares/shares_test.go
evan-forbes: oh shoot, good point forgot about this. This test requires things like implementations of evidence and a "ShareSplitter" interface which is no longer followed. will create an issue to make sure we do the equivalent and manually checking the shares in the upcoming refactor

evan-forbes: #630

Issue #632, opened 2022-08-18
Test compatibility of SplitMessages with/without share index


rootulp:
### Context
During non-interactive defaults, we added empty message shares that act as padding between messages. In order for transactions to be linked to the share(s) that they pay for, `indexes` were added to [SplitMessages](https://github.com/celestiaorg/celestia-app/blob/a3ef48c582403f6df13d1a65635c5fadbdbfbd3f/pkg/shares/split.go#L96). 

### Problem Definition

Since `indexes` didn't exist in v0.6.0, we need to verify if v0.6.0 is compatible with main.

Ref:
- https://github.com/celestiaorg/celestia-app/blob/a3ef48c582403f6df13d1a65635c5fadbdbfbd3f/pkg/shares/split.go#L69
- https://github.com/celestiaorg/celestia-app/blob/a3ef48c582403f6df13d1a65635c5fadbdbfbd3f/pkg/shares/split.go#L96

### Proposal

Write a unit test or integ test that verifies compatibility of `SplitMessages` with/without share index

evan-forbes:
to expand on this slightly, the current tests do test comparability of the encoding and decoding, as we haven't changed them in #627 and won't be changing them when introducing non-interactive defaults either. 

What we need to add is a mechanism to detect that we are decoding an old block, so that we don't add any padding or care if there is a share index in the wrapped transactions. This could be as simple (and implicit) as reading the first share index from a wrapped transaction. If that index is 0, then we automatically know that we are reading an old block, and can pass `nil` to `SplitMessages(nil, msgs)`, which effectively uses identical share splitting code to v0.6.0

evan-forbes:
ref #608

Issue #382, opened 2022-05-05
Tracking: Non-interactive defaults


evan-forbes:
With our current message inclusion check, we are not checking the location of the messages, only that each message has a corresponding commitment in a `MsgPayForData`. This works for ensuring that each message was included, but this doesn't actually check that the commitment is actually a subtree root to one of the data availability header row or column roots.

In order to do this, we have to have block producers organize the message layout to follow the [non-interactive defaults](https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md). This will involve a refactor during the app's `PrepareProposal` call, along with adding the message [starting location to wrapped txs](https://github.com/celestiaorg/celestia-core/issues/462#issuecomment-1118738645).

Additionally, we need validators and full nodes to check that each share commitment is actually a subtree root to one of the data availability headers. This will require a refactor to the app's `ProcessProposal` call, and is blocked by the functionality described in the above paragraph.

- [x] Add `MessageStartIndex` field to the [malleated transactions](https://github.com/celestiaorg/celestia-core/blob/7e29d6560b7e2c96c60c2c5b2ae5d9e371e382d8/proto/tendermint/types/types.proto#L221-L227) struct https://github.com/celestiaorg/celestia-core/issues/462
- [x] Create a good API for checking for subtree root inclusion #381
- [x] Refactor `PrepareProposal` to organize the message layout according to the non-interactive defaults
- [x] #383

A **huge bonus** would be if we would also implement message inclusion fraud proofs, but this does not have to be completed to close this issue

evan-forbes:
I have this listed as an incentivized testnet feature for now, as I don't think we will have it ready in time for testnet considering our other priorities, BUT we can (and should imho) obviously include this in an upgrade to testnet after we finish.

evan-forbes:
The current plan for introducing features from the non-interactive defaults branch is to first review the library code:

#### Message inclusion
- [ ] https://github.com/celestiaorg/celestia-app/pull/549
- [ ] https://github.com/celestiaorg/celestia-app/pull/621
- [ ] General purpose API for ProcessProposal and celestia-node

#### Encoding
- [ ] #623 
- [ ] General refactor
- [ ] Add non-interactive defaults logic

#### Modify ABCI++ methods
- [ ] PrepareProposal Refactor to fill the square using non-interactive defaults
- [ ] ProcessProposal Refactor to check for message inclusion using sub tree roots

#### Documentation
- [ ] #598

[...]

PR #637, opened 2022-08-18
chore: refactor share splitting and merging

evan-forbes:
## Description

This PR refactors splitting and merging. It incorporates the changes found in https://github.com/celestiaorg/celestia-core/pull/820, **please see the second commit that addresses the remaining feedback** from that PR. The rest of the changes involves splitting up the existing code into different files and adding backwards compatibility with blocks that use and don't use wrapped txs for non-interactive defaults (pointed out below). 

part of: #382 
closes: #632


pkg/shares/share_splitting.go
```diff
@@ -1,412 +1,120 @@
...
+			// attempting to extract share indexes from txs that do not have any
+			// due to them being old. here we return nil to indicate that we are
+			// attempting to extract indexes from a block that doesn't support
+			// it. It's check for 0 because if there is a message in the block,
+			// then there must also be a tx, which will take up at least one
+			// share.
+			if malleatedTx.ShareIndex == 0 {
+				return nil
+			}
+			msgIndexes = append(msgIndexes, malleatedTx.ShareIndex)
```
evan-forbes, line 81: here we make this code compatible with blocks of transactions that don't use wrapped txs or specify the share index. This will be useful if we want to preserve the ability of syncing from scratch on mamaki.

pkg/shares/merge_contiguous_shares.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(ss.shares) == 0 {
+		return nil, nil
+	}
+	err := ss.peel(ss.shares[0][consts.NamespaceSize+consts.ShareReservedBytes:], true)
+	return ss.txs, err
+}
+
+// peel recursively parses each chunk of data (either a transaction,
+// intermediate state root, or evidence) and adds it to the underlying slice of data.
+func (ss *shareStack) peel(share []byte, delimited bool) (err error) {
```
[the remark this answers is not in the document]
rootulp: **UPDATE**: disregard, I see there are tests for `processContiguousShares` in `shares_test.go`. Thoughts on moving them to `merge_contiguous_shares_test.go`?

rootulp (commented):
[optional][proposal] with respect to naming consistency

Currently
- `parseMsgShares` in `merge_message_shares.go`
- `processContiguousShares` in `merge_contiguous_shares.go`

It may help if similar behavior was named consistently. I think `parse` may be a better fit so proposal

- `parseMsgShares` in `parse_message_shares.go`
- `parseContiguousShares` in `parse_contiguous_shares.go`

^ this doesn't necessarily need to be addressed in this PR

rootulp:
this refactor is a herculean effort 💪 great work! @evan-forbes

pkg/shares/shares_test.go
```diff
@@ -438,14 +420,57 @@ func Test_parseMsgShares(t *testing.T) {
 	}
 }
 
+func TestParsePaddedMsg(t *testing.T) {
```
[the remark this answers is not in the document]
evan-forbes: good idea, added msg_shares_test and contiguous_shares_test as I felt there were enough tests to justify the distinction between message and contiguous, but not also between merging/splitting the as well.

pkg/shares/utils.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return e
+}
+
+func powerOf2(v uint64) bool {
+	if v&(v-1) == 0 && v != 0 {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
evan-forbes: :+1: 721e31e2b9409ac5df10ad6ff7730c29c296a67c

app/split_shares.go
```diff
@@ -188,7 +191,7 @@ func (sqwr *shareSplitter) writeMalleatedTx(
 		return false, nil, nil, err
 	}
 
-	wrappedTx, err := coretypes.WrapMalleatedTx(parentHash[:], rawProcessedTx)
+	wrappedTx, err := coretypes.WrapMalleatedTx(parentHash[:], 0, rawProcessedTx)
```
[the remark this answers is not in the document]
evan-forbes: split shares in the app doesn't support non-interactive defaults or using wrapped txs at all, so I just put a zero here to fulfill the api change. This entire file will be removed. added a comment in the mean time 8cc17bf7201409f34f7aac8b70daa7a60bdacdad

pkg/shares/merge_contiguous_shares.go
```diff
@@ -0,0 +1,81 @@
...
+
+	ss := newShareStack(shares)
+	return ss.resolve()
+}
+
+// shareStack hold variables for peel
+type shareStack struct {
+	shares [][]byte
+	txLen  uint64
+	txs    [][]byte
```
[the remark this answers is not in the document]
evan-forbes: good idea, df122c845a8f0ad27aa7246dddb12981d2bb3d95

evan-forbes:
after thinking about this a bit more, unfortunately I think we'll have to this, along with the non-interactive default logic, to core. The idea of keeping all app specific logic is ideal, but as discussed in https://github.com/celestiaorg/celestia-core/issues/842 I don't think we'll be to avoid have at least a copy there. It would be one thing to only have a copy of the contiguous share splitting logic, but we will also have to have a copy of the non-interactive default logic. 

The refactor and move to ./pkg here is still worthwhile, so I think we should merge this first and then we can move the relevant portions of the non-interactive defaults over to the core/pkg directory as well.

pkg/shares/message_shares_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/pkg/consts"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+func Test_parseMsgShares(t *testing.T) {
+	// exactMsgShareSize is the length of message that will fit exactly into a single
+	// share, accounting for namespace id and the length delimiter prepended to
+	// each message
+	const exactMsgShareSize = consts.MsgShareSize - 2
```
rootulp, line 17: [no change needed] because this was present in the test prior to the refactor but since `consts.MsgShareSize` already accounts for the 8 byte namespace id, this assumes that the length delimiter is 2 bytes. Since length delimiter is a varint, isn't it possible that the delimiter is > 2 bytes?

pkg/shares/message_shares_test.go
```diff
@@ -0,0 +1,99 @@
...
+		{"many small msgs", 100, 10},
+		{"single big msg", 1000, 1},
+		{"many big msgs", 1000, 10},
+		{"single exact size msg", exactMsgShareSize, 1},
+		{"many exact size msgs", consts.MsgShareSize, 10},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		// run the tests with identically sized messagses
```
rootulp, line 38: [nit] typo

```suggestion
		// run the tests with identically sized messages
```

pkg/shares/message_shares_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Run(fmt.Sprintf("%s idendically sized ", tc.name), func(t *testing.T) {
```
rootulp, line 39: [nit] typo
```suggestion
		t.Run(fmt.Sprintf("%s identically sized ", tc.name), func(t *testing.T) {
```

pkg/shares/message_shares_test.go
```diff
@@ -0,0 +1,99 @@
...
+			msgs.SortMessages()
+
+			shares, _ := SplitMessages(nil, msgs.MessagesList)
+
+			parsedMsgs, err := parseMsgShares(shares)
+			if err != nil {
+				t.Error(err)
+			}
+
+			// check that the namesapces and data are the same
```
rootulp, line 55: [nit] typo
```suggestion
			// check that the namespaces and data are the same
```

something must be misconfigured with our misspell linter: https://github.com/celestiaorg/celestia-app/blob/721e31e2b9409ac5df10ad6ff7730c29c296a67c/.golangci.yml#L9

because I would expect these to be caught by that

pkg/shares/split_contiguous_shares.go
```diff
@@ -0,0 +1,190 @@
...
+// ContiguousShareSplitter will write raw data contiguously across a progressively
+// increasing set of shares. It is used to lazily split block data such as transactions
+// into shares.
+type ContiguousShareSplitter struct {
+	shares       []NamespacedShare
+	pendingShare NamespacedShare
+	namespace    namespace.ID
+}
+
+// NewContiguousShareSplitter returns a ContigousShareWriter using the provided
```
rootulp, line 22:
```suggestion
// NewContiguousShareSplitter returns a ContiguousShareSplitter using the provided
```

rootulp (approved):
Great work! The typos are non-blocking but indicative that something is wrong with our misspell config, I'll investigate

pkg/shares/message_shares_test.go
evan-forbes: yeah! it could be more, but not for this test? tbh I'd need to go check if there are large enough messages

pkg/shares/merge_contiguous_shares.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (ss *shareStack) resolve() ([][]byte, error) {
+	if len(ss.shares) == 0 {
+		return nil, nil
+	}
+	err := ss.peel(ss.shares[0][consts.NamespaceSize+consts.ShareReservedBytes:], true)
+	return ss.txs, err
+}
+
+// peel recursively parses each chunk of data (either a transaction,
```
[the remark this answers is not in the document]
rootulp: `peel` seems good to me. I think your preference for one syllable unix sounding names conforms to Go conventions for short naming

pkg/shares/shares_test.go
```diff
@@ -275,227 +274,6 @@ func TestFuzz_Merge(t *testing.T) {
 	}
 }
 
-func Test_processContiguousShares(t *testing.T) {
```
rootulp, line 297: [for reviewers] this was moved to pkg/shares/contiguous_shares_test.go

rootulp (approved):

evan-forbes:
merging now, even tho we're planning on moving this to its own repo as per this [comment](https://github.com/celestiaorg/celestia-core/issues/842#issuecomment-1230635653)

PR #641, opened 2022-08-19
chore: add lint and fmt Makefile tasks

rootulp:
Inspired by https://github.com/celestiaorg/celestia-node/blob/main/Makefile#L61-L74

## Testing

I intentionally introduced a markdownlint violation
```
$ make lint
--> Running linter
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649.
README.md:62 MD025/single-title/single-h1 Multiple top-level headings in the same document [Context: "# Carrers"]
make: *** [lint] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,11 @@ proto-format:
 build-docker:
 	$(DOCKER) build -t celestiaorg/celestia-app -f docker/Dockerfile .
 
+lint:
+	@echo "--> Running linter"
+	@golangci-lint run
+	@markdownlint --config .markdownlint.yaml '**/*.md'
+.PHONY: lint
 
 ###############################################################################
 ###                           Tests & Simulation                            ###
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,12 @@ proto-format:
 build-docker:
 	$(DOCKER) build -t celestiaorg/celestia-app -f docker/Dockerfile .
 
+fmt:
+	@go mod tidy -compat=1.17
+	@golangci-lint run --fix
+	@markdownlint --fix --quiet --config .markdownlint.yaml .
+.PHONY: fmt
+
 lint:
 	@echo "--> Running linter"
 	@golangci-lint run
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,6 @@ build-docker:
 	$(DOCKER) build -t celestiaorg/celestia-app -f docker/Dockerfile .
 
 fmt:
-	@go mod tidy -compat=1.17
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):
hell yeah how have we been living without this!!

do we want to add `gofumpt` as well somewhere? I'd be fine with either not adding it, or maybe have some sort of differentiation between the commands that actually modify code and others that just let us know something is wrong.

rootulp:
I agree, running linters with `--fix` would be great and we should differentiate between lint check and lint fix. Proposal to run:
- lint `--fix` on `make fmt`
- lint check on `make lint`
 
Motivation: conforms to what celestia-node does [here](https://github.com/celestiaorg/celestia-node/blob/main/Makefile#L61-L74). 

Note: `gofumpt` will be run by golangci-lint because

https://github.com/celestiaorg/celestia-app/blob/435e557dded38bfd3371ea62d73641e637481ee1/.golangci.yml#L8

Makefile
```diff
@@ -61,6 +61,17 @@ proto-format:
 build-docker:
 	$(DOCKER) build -t celestiaorg/celestia-app -f docker/Dockerfile .
 
+fmt:
+	@go mod tidy -compat=1.17
```
evan-forbes, line 65: should we update this to 1.18?

Makefile
```diff
@@ -61,6 +61,17 @@ proto-format:
...
+	@golangci-lint run --fix
```
evan-forbes, line 66: do we need to call `gofumpt -w` here?

I know it gets run with golangci-lint, but it at least isn't actually fixing the code for me. perhaps I have it configured wrong?

evan-forbes (dismissed):

Makefile
rootulp: 🤔 it's not immediately clear based on docs

> By default, go mod tidy will check that the [selected versions](https://go.dev/ref/mod#glos-selected-version) of modules do not change when the module graph is loaded by the Go version immediately preceding the version indicated in the go directive. The versioned checked for compatibility can also be specified explicitly via the -compat flag.

from https://go.dev/ref/mod#go-mod-tidy

Given we have other Makefile tasks that already use `-compat=1.18` I think we should be consistent here. 

On second thought, we can also just remove this line entirely and rely on `make build` and `make mod` to run `go mod tidy` so will remove.

evan-forbes (approved):

adlerjohn (approved):

PR #643, opened 2022-08-20
fix: invalidate reserved namespace ids in ValidateBasic

rootulp:
Closes https://github.com/informalsystems/audit-celestia/issues/2


evan-forbes (approved):
thanks again for fixing this so quickly :slightly_smiling_face:

PR #644, opened 2022-08-20
fix: comment for NextHighestPowerOf2

rootulp:
Closes https://github.com/informalsystems/audit-celestia/issues/4

### DIFF
--- a/x/payment/types/payfordata.go
+++ b/x/payment/types/payfordata.go
@@ -188,7 +188,11 @@ func powerOf2MountainRange(l, k uint64) []uint64 {
 	return output
 }
 
-// NextHighestPowerOf2 returns the next highest power of 2
+// NextHighestPowerOf2 returns the next highest power of 2.
+// Examples:
+// NextHighestPowerOf2(1) = 2
+// NextHighestPowerOf2(2) = 4
+// NextHighestPowerOf2(5) = 8
 func NextHighestPowerOf2(v uint64) uint64 {
 	// keep track of the value to check if its the same later
 	i := v
--- a/x/payment/types/payfordata.go
+++ b/x/payment/types/payfordata.go
@@ -188,8 +188,7 @@ func powerOf2MountainRange(l, k uint64) []uint64 {
 	return output
 }
 
-// NextHighestPowerOf2 returns the next lowest power of 2 unless the input is a power
-// of two, in which case it returns the input
+// NextHighestPowerOf2 returns the next highest power of 2
 func NextHighestPowerOf2(v uint64) uint64 {
 	// keep track of the value to check if its the same later
 	i := v


x/payment/types/payfordata.go
```diff
@@ -188,8 +188,7 @@ func powerOf2MountainRange(l, k uint64) []uint64 {
 	return output
 }
 
-// NextHighestPowerOf2 returns the next lowest power of 2 unless the input is a power
-// of two, in which case it returns the input
+// NextHighestPowerOf2 returns the next highest power of 2
```
adlerjohn, line 191: Not sure this is the correct resolution. If anything, the function should be renamed to `NextLowestPowerOf2`? Additionally, the linked issue should be resolved by properly documenting behavior when the input is a power of 2.

These are two orthogonal things.

rootulp: 1. This function finds the next highest power of two. Confirmed by checking [these tests](https://github.com/rootulp/celestia-app/blob/11e5355b66c1f38085b5682ef85feab64f72e561/x/payment/types/payfordata_test.go#L75-L101)
2. I think it is more confusing to explicitly document the behavior when the input is a power of 2 because the behavior is the same as when the input is not a power of 2. In both cases, this function returns the next highest power of 2. I think examples may be the easiest way to convey the behavior

```golang
// NextHighestPowerOf2 returns the next highest power of 2. Examples:
// NextHighestPowerOf2(1) = 2
// NextHighestPowerOf2(2) = 4
// NextHighestPowerOf2(11) = 16
func NextHighestPowerOf2(v uint64) uint64 {
```

thoughts?

rootulp: UPDATE: now I see why **highest** is misleading here 🤦‍♂️ this function returns the next lowest power of 2 that is higher than the input. Hmm will re-think comments / function naming here

adlerjohn: `NextHigherPowerOf2` would also work.

rootulp: Will avoid renaming because there exists a `nextLowestPowerOf2` already: https://github.com/celestiaorg/celestia-app/blob/127f990d4b5aab538d4b24dc254967fafb665789/x/payment/types/payfordata.go#L218

evan-forbes (approved):
this is my bad, pls feel free to rename any of the functions as we see fit.

rootulp:
Will merge this but happy to create a follow-up PR if @adlerjohn has any other feedback

adlerjohn:
It's fine as-is, but could you file a follow-up issue @rootulp so we don't forget?

rootulp:
Sure https://github.com/celestiaorg/celestia-app/issues/648

Issue #638, opened 2022-08-19
QGB state machine confirm signed by validator  


SweeXordious:
### Summary of Bug

Currently, the QGB queries the validator by orchestrator to check if the validator was registered.
However, this check only checks the current set of validators. Thus, submitting confirms for old valsets can fail even if they are correct:

https://github.com/celestiaorg/celestia-app/blob/d63b99891023d153ea5937e4f3c1907a784654d8/x/qgb/keeper/msg_server.go#L51

This will need to be fixed.

### Version

d63b99891023d153ea5937e4f3c1907a784654d8

### Steps to Reproduce

not needed.

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

SweeXordious:
Following the decision taken on https://github.com/celestiaorg/celestia-app/issues/631, this check will need to be deleted also as we will no longer iterate over valsets

PR #645, opened 2022-08-22
QGB state machine remove confirm valset check

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/631

Issue #501, opened 2022-06-24
QGB investigate the EndBlocker panics and document them


SweeXordious:
as per normal, we just have to document when this panic can occur. we can handle this either in this PR or in a different issue

same with line 75

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/497#discussion_r906174812_

Issue #270, opened 2022-04-01
Write ADR for QGB Relayer and Orchestrator 1.0


evan-forbes:
The QGB is currently designed around having the ability to slash individual validators on celestia, not unlike the normal non-quantum gravity module. This is not needed, and dramatically changes how the orchestrator and relayer should be designed. We should flesh out this design in an ADR that can be written in parallel with an implementation.

Currently, the orchestrator and relayer are piggybacking on the state required for slashing individual validators. The thinking was that since the state already had to exist for slashing, that the orchestrator would create that state by submitting their signed commitment transactions to Celestia, and then the relayer would relay that state directly to the evm chain of choice.

This is a very reasonable design, assuming that the state has to be there for slashing. However, the state does not have to be there for slashing, as we can rely entirely on hardforking, as it takes ⅔ of the voting power to update the contracts in the first place. If a malicious update occurs, then all the evidence needed for fraud proofs is posted on that evm chain. The one thing that we cannot slash using this model is downtime. If validators choose not to sign or send their txs to a relayer, then we have no way of punishing them other than shaming them twitter. This is not a bad thing imho, as shaming validators has proven effective in the past for other PoS networks, and it puts a lot less strain on the validators ensuring that every signature is perfect. If for some reason they accidentally sign the wrong commitment or double sign, then they don't get slashed.

The good news is that while we will have to ditch a bit of work done for the app-integration, most of the orchestrator and relayer code can be reused.

The new functionality will be in how the relayer and orchestrators communicate. This will have to be done in some p2p fashion, for the 0.1.0 (or whatever release), we can simply have the orchestrators send their signed commitents to some set of relayers via rpc. In the future, we could get fancier.

evan-forbes:
To update this, after two synchronous calls, we decided that we did actually want to slash individual validators for not signing. This requires the additional state that is introduced in the qgb-integration branch, mainly that signatures are posted on celestia. This means that we will not need an additional p2p network. 

@SweeXordious came up with the great idea of removing most, if not all, additional state after the unbonding period, as the signatures are not useful after that period of time.

evan-forbes:
per sync discussion, this issue can be closed due to moving to v2

PR #646, opened 2022-08-22
QGB state machine remove invalid validator set check 

SweeXordious:
closes https://github.com/celestiaorg/celestia-app/issues/638


SweeXordious:
Depends on https://github.com/celestiaorg/celestia-app/pull/645

x/qgb/keeper/msg_server.go
```diff
@@ -47,48 +45,11 @@ func (k msgServer) ValsetConfirm(
 		return nil, sdkerrors.Wrap(types.ErrInvalid, "acc address invalid")
 	}
 
-	// Verify if validator exists
```
rootulp, line 48: [question][non-blocking] do any checks need to be removed from https://github.com/celestiaorg/celestia-app/blob/qgb-integration/docs/architecture/ADR-002-QGB-ValSet.md?plain=1#L158 ? I don't see these checks in that ADR section so maybe not but the contents of the ADR don't match 1:1 with the `ValsetConfirm` implementation here so wanted to confirm.

SweeXordious: Yes, the ADRs are pretty oooold now. We will need to update them once we have a stable QGB 1.0.  The reason we stopped updating them, is we were trying a lot of approaches, and we changed the design multiple times. Thus, it was too much work to keep updating them always.

rootulp: Makes sense. If the ADR is no longer accurate then perhaps we can include a disclaimer at the top stating that it's unmaintained and create a GH issue to update it once we have a stable QGB 1.0? Defer to you on how best to track / update it.

SweeXordious: Makes sense. We have few issues concerning this: 
https://github.com/celestiaorg/celestia-app/issues/217
https://github.com/celestiaorg/celestia-app/issues/501
https://github.com/celestiaorg/celestia-app/issues/270

evan-forbes (approved):
:shipit:

PR #651, opened 2022-08-23
ADR-002: InfoReservedByte

rootulp:
## Description

Moved from https://github.com/celestiaorg/celestia-core/pull/843
ADR with implementation details for https://github.com/celestiaorg/celestia-core/issues/839

### DIFF
--- a/docs/architecture/ADR-002-info-reserved-byte.md
+++ b/docs/architecture/ADR-002-info-reserved-byte.md
@@ -1,0 +1,133 @@
+# ADR 009: InfoReservedByte
+<!-- disable markdownlint MD010 because Go code snippet uses tabs -->
+<!-- markdownlint-disable MD010 -->
+
+## Changelog
+
+- 2022/9/22: inital draft
+
+## Context
+
+The current message share format is as follows:
+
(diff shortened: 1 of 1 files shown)


rootulp:
We had a meeting on info reserved byte and aligned on using it universally rather than just for message shares. See https://github.com/celestiaorg/celestia-app/issues/659

Closing this PR but will use the contents of this ADR in a new one for the universal share encoding format.

evan-forbes:
> Closing this PR but will use the contents of this ADR in a new one for the universal share encoding format.

initially these seem like separate proposals/implementation efforts, but I can see how they might be blocking. Is that why they were merged in https://github.com/celestiaorg/celestia-app/pull/660?

rootulp:
I merged them because I had the wrong interpretation of the previous proposal. I thought the info byte was only applicable to message shares. Given info share byte applies to both types of shares, I renamed and updated the ADR. This PR (for info bye just on message shares) seems no longer necessary.

With regards to implementation: we can split PRs into smaller chunks (i.e. one for updating the message share format and a separate one for updating the reserved namespace share format). I don't think one is strictly blocking the other but I do think it would be easier to tackle this work post merging in-flight share refactor PRs.

Issue #656, opened 2022-08-24
QGB posting confirms to a reserved namespace ADR


SweeXordious:
Currently, we're posting Valset Confirms and DataCommitment Confirms directly in state. This is posing us with limitations such as: state grows and increasingly high gas fees.
Thus, we can think of posting the confirms to a reserved namespace to solve the above issues.
So, we will need to formalize the decision in an ADR.

SweeXordious:
Update, we decided to update the ADR to reducing the QGB module state.

SweeXordious:
no longer relevant as per the new design

PR #657, opened 2022-08-24
QGB ADR-005 Reducing QGB module state usage

SweeXordious:
Closing https://github.com/celestiaorg/celestia-app/issues/656

### DIFF
--- a/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
+++ b/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
@@ -6,7 +6,7 @@
 
 ## Context
 
-Currently, the QGB is using the state solely to save all the attestations and confirms it needs.
+Currently, the QGB is using the state to save all the attestations and confirms it needs.
 We propose to post `Valset Confirms` and `DataCommitment Confirms` in a reserved namespace instead of adding them to the state to achieve the following:
 
 - Reduce the state machine complexity
@@ -25,7 +25,7 @@ The approach that we were planning to take is to cleanup the state after the unb
 This way, we will always have a fixed sized state instead of a gigantic one, issue defining this: [Prune the QGB state after the unbonding period ends #309](https://github.com/celestiaorg/celestia-app/issues/309).
 This would mean that the QGB contracts deployed after genesis will never have the whole history of attestations.
 
-### Remove state definitively
+### Separate P2P network
 
 This would mean gossiping the confirms and attestations in a separate P2P network.
 The pros of this approach are that it will be cheaper and wouldn't involve any state changes.
@@ -36,7 +36,7 @@ However, slashing will not be possible.
 Deploy the QGB as a Rollup that posts its data to Celestia, and, uses a separate settlement layer for slashing.
 This might be the end goal of the QGB, but it will be very involved to build at this stage.
 
-Also, if this ADR got accepted, it will be an important stepping stone into the Rollup direction.
+Also, if this ADR got accepted, it will be an important stepping stone in the Rollup direction.
 
 ## Decision
 
@@ -53,14 +53,14 @@ The following issue lists these checks: [QGB data commitments/valsets state mach
 Also, the state machine for the QGB module became complex and more prone to bugs that might end up halting/forking the chain.
 
 In addition to this, with the gas leak issue discussed in this [comment](https://github.com/celestiaorg/celestia-app/issues/631#issuecomment-1220848130), we ended up removing the sanitizing checks we used to run on the submitted `Valset Confirms` and `DataCommitment Confirms`.
-This was done in the goal of not charging orchestrators increasing gas fees with every posted attestations.
+This was done with the goal of not charging orchestrators increasing gas fees for every posted attestation.
 A simple benchmark showed that the gas usage multiplied 2 times from `~50 000` to `100 000` after submitting 16 attestation.
 Also, even if removing the checks was the most practical solution, it ended up opening new attack vectors on the QGB module state, such as flooding the network with incorrect attestations from users who are not even validators.
 Which would increase the burden on validators to handle all of that state.
 Furthermore, it put the responsibility on the relayer to cherry-pick the right confirms from the invalid ones.
 
 This led us to think that the direction we're taking makes more sense if it moves the confirms from the state to the block space.
-This way, we can remove all the complex logic handling the Valset Confirms and DataCommitment Confirms to the orchestrator side, and also block any opportunity for any malicious party to post data to state.
+This way, we can remove all the complex logic handling the Valset Confirms and DataCommitment Confirms to the orchestrator side, and also block any opportunity for a malicious party to post data to state.
 
 In fact, we will be able to remove the `MsgValsetConfirm` defined in [here](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L24-L49)
 And also, the `MsgDataCommitmentConfirm` defined in [here](
@@ -85,7 +85,7 @@ Proposed
 
 - Reduce significantly the gas fees paid by orchestrators.
 - Reduce significantly the use of Celestia state.
-- Keep the whole QGB confirms history via using block data, instead of being forced to delete it after the unbonding period not to end up with a super gigantic state. This would allow any QGB contract to have the whole history on any EVM chain.
+- Keep the whole QGB confirms history via using block data, instead of being forced to delete it after the unbonding period to not end up with a super gigantic state. This would allow any QGB contract to have the whole history on any EVM chain.
 
 ### Negative
 
--- a/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
+++ b/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
@@ -72,16 +72,16 @@ https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1a
 ```go
 // ValsetConfirm handles MsgValsetConfirm.
 func (k msgServer) ValsetConfirm(
-	c context.Context,
-	msg *types.MsgValsetConfirm,
+    c context.Context,
+    msg *types.MsgValsetConfirm,
 ) (*types.MsgValsetConfirmResponse, error) {
-	// <delete_all_this>
+    // <delete_all_this>
 }
 
 // DataCommitmentConfirm handles MsgDataCommitmentConfirm.
 func (k msgServer) DataCommitmentConfirm(
-	c context.Context,
-	msg *types.MsgDataCommitmentConfirm,
+    c context.Context,
+    msg *types.MsgDataCommitmentConfirm,
 ) (*types.MsgDataCommitmentConfirmResponse, error) {
     // <delete_all_this>
 }
@@ -89,11 +89,11 @@ func (k msgServer) DataCommitmentConfirm(
 
 This would reduce significantly the QGB module state usage, reduce the complexity of the state machine and give us the same benefits of the current design.
 
-As a direct consequence to this, the relayer will need to adapt to this change and start getting the transactions straight from the blocks, parse them and sanitize the commits. Thus, making the relayer implementation more complex. 
+As a direct consequence to this, the relayer will need to adapt to this change and start getting the transactions straight from the blocks, parse them and sanitize the commits. Thus, making the relayer implementation more complex.
 
 However, we can assume that the relayer will have enough computing power to do the latter. Also, only one relayer is necessary to have a working QGB contract. So, the relayer cost is justified.
 
-For the orchestrators, they will also need to parse the history to keep track of any missed signatures. But, same as with the relayers, the cost is justified. 
+For the orchestrators, they will also need to parse the history to keep track of any missed signatures. But, same as with the relayers, the cost is justified.
 
 For posting transactions, we will rely on gas fees as a mechanism to limit malicious parties to flood the network with invalid transactions. Then, eventually slash malicious behavious. However, since posting confirms will be possible for any user of the network. It won't be possible to slash ordinary users, who are not running validators, if they post invalid confirms.
 
--- a/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
+++ b/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
@@ -21,7 +21,8 @@ Keeping the current design would entail using the state extensively.
 This proves bad when the state grows after a few hundred attestations, and performing checks on the `Valset Confirms` and `DataCommitment Confirms` becomes too expensive.
 An example of such issue in here: [QGB data commitments/valsets state machine checks #631](https://github.com/celestiaorg/celestia-app/issues/631) and [Investigate the QGB transactions cost #603](https://github.com/celestiaorg/celestia-app/issues/603).
 
-Also, we will be forced to prune the state after the unbonding period not to end up with a gigantic state, issue defining this: [Prune the QGB state after the unbonding period ends #309](https://github.com/celestiaorg/celestia-app/issues/309).
+The approach that we were planning to take is to cleanup the state after the unbonding period.
+This way, we will always have a fixed sized state instead of a gigantic one, issue defining this: [Prune the QGB state after the unbonding period ends #309](https://github.com/celestiaorg/celestia-app/issues/309).
 This would mean that the QGB contracts deployed after genesis will never have the whole history of attestations.
 
 ### Remove state definitively
--- a/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
+++ b/docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
@@ -8,28 +8,29 @@
 
 Currently, the QGB is using the state solely to save all the attestations and confirms it needs.
 We propose to post `Valset Confirms` and `DataCommitment Confirms` in a reserved namespace instead of adding them to the state to achieve the following:
+
 - Reduce the state machine complexity
 - Reduce the amount of state used by the QGB
 - Prepare for a more Rollup oriented QGB design
 
 ## Alternative Approaches
 
-#### Keep the existing design
+### Keep the existing design
 
 Keeping the current design would entail using the state extensively.
 This proves bad when the state grows after a few hundred attestations, and performing checks on the `Valset Confirms` and `DataCommitment Confirms` becomes so expensive.
-An example of such issue in here: https://github.com/celestiaorg/celestia-app/issues/603 and https://github.com/celestiaorg/celestia-app/issues/631.
+An example of such issue in here: [QGB data commitments/valsets state machine checks #631](https://github.com/celestiaorg/celestia-app/issues/631) and [Investigate the QGB transactions cost #603](https://github.com/celestiaorg/celestia-app/issues/603).
 
-Also, we will be forced to prune the state after the unbonding period not to end up with a gigantic state, issue defining this: https://github.com/celestiaorg/celestia-app/issues/309.
+Also, we will be forced to prune the state after the unbonding period not to end up with a gigantic state, issue defining this: [Prune the QGB state after the unbonding period ends #309](https://github.com/celestiaorg/celestia-app/issues/309).
 This would mean that the QGB contracts deployed after genesis will never have the whole history of attestations.
 
-#### Remove state definitively
+### Remove state definitively
 
 This would mean gossiping the confirms and attestations in a separate P2P network.
 The pros of this approach are that it will be cheaper and wouldn't involve any state changes.
 However, slashing will not be possible.
 
-#### QGB Rollup
+### QGB Rollup
 
 Deploy the QGB as a Rollup that posts its data to Celestia, and, uses a separate settlement layer for slashing.
 This might be the end goal of the QGB, but it will be very involved to build at this stage.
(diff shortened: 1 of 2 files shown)


docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
...
+
+Proposed
+
+## Consequences
+
+### Positive
+
+- Reduce significantly the gas fees paid by orchestrators.
+- Reduce significantly the use of Celestia state.
+- Keep the whole QGB confirms history via using block data, instead of being forced to delete it after the unbonding period not to end up with a super gigantic state. This would allow any QGB contract to have the whole history on any EVM chain.
```
rootulp, line 87: [nit]
```suggestion
- Keep the whole QGB confirms history via using block data, instead of being forced to delete it after the unbonding period to not end up with a super gigantic state. This would allow any QGB contract to have the whole history on any EVM chain.
```

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Negative
+
+- Reducing the checks applied on the confirms: everyone will be able to post commitments, and we will have no way to sanitize them before adding them to a block. This would add an extra overhead for the relayer to choose the right commitments. However, it is alright, since we will not have a lot of Also, we will need to decide if this will be relayers, and we can give them enough computation power to compensate the overhead.
```
rootulp, line 91: > However, it is alright, since we will not have a lot 

the rest of this sentence got lost

rootulp, line 91: > everyone will be able to post commitments,

If this is the case, why do commitments need to be posted to a reserved namespace?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
+# ADR 5: Using block space instead of state to post attestations confirms
+
+## Changelog
+
+- {date}: {changelog}
+
+## Context
+
+Currently, the QGB is using the state solely to save all the attestations and confirms it needs.
```
rootulp, line 9: [nit]
```suggestion
Currently, the QGB is using the state to save all the attestations and confirms it needs.
```

Rationale: Even if confirms are removed from state, there are other things stored in state due to the attestations:

> Then, the only state that the QGB module will keep is the one that is created in [EndBlocker](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/x/qgb/abci.go#L12-L16).
Which are `Attestations`, i.e. `Valset`s and `DataCommitmentRequest`s.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Alternative Approaches
+
+### Keep the existing design
+
+Keeping the current design would entail using the state extensively.
+This proves bad when the state grows after a few hundred attestations, and performing checks on the `Valset Confirms` and `DataCommitment Confirms` becomes so expensive.
+An example of such issue in here: [QGB data commitments/valsets state machine checks #631](https://github.com/celestiaorg/celestia-app/issues/631) and [Investigate the QGB transactions cost #603](https://github.com/celestiaorg/celestia-app/issues/603).
+
+Also, we will be forced to prune the state after the unbonding period not to end up with a gigantic state, issue defining this: [Prune the QGB state after the unbonding period ends #309](https://github.com/celestiaorg/celestia-app/issues/309).
```
rootulp, line 24: [question] why is it problematic to prune the state? I assume that the state that gets pruned is no longer necessary.

rootulp (commented):
Great work articulating the problem! Most of my comments are nits. I'm super curious on what the "QGB Rollup" design entails but it doesn't seem necessary to flesh out now.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
+# ADR 5: Using block space instead of state to post attestations confirms
+
+## Changelog
+
+- {date}: {changelog}
```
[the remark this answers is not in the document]
SweeXordious: I guess we can leave it for now until we see if this ADR gets accepted

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
...
+- Reduce the state machine complexity
+- Reduce the amount of state used by the QGB
+- Prepare for a more Rollup oriented QGB design
+
+## Alternative Approaches
+
+### Keep the existing design
+
+Keeping the current design would entail using the state extensively.
+This proves bad when the state grows after a few hundred attestations, and performing checks on the `Valset Confirms` and `DataCommitment Confirms` becomes so expensive.
```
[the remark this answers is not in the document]
SweeXordious: check now please, added a sentence explaining that.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,96 @@
...
+
+In fact, we will be able to remove the `MsgValsetConfirm` defined in [here](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L24-L49)
+And also, the `MsgDataCommitmentConfirm` defined in [here](
+https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L55-L76).
+Which were the way orchestrators were able to post confirms to the QGB module.
+Then, the only state that the QGB module will keep is the one that is created in [EndBlocker](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/x/qgb/abci.go#L12-L16).
+Which are `Attestations`, i.e. `Valset`s and `DataCommitmentRequest`s.
+We will need to investigate the state's growth if we keep in it only these latter.
+
+In addition to this, we will need to reserve a certain namespace ID to only be reserved for Confirms.
```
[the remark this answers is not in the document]
SweeXordious: The QGB rollup design is to deploy the QGB as a Rollup on top of Celestia: have its separate state and network and post data to Celestia. Then, think of a way to add a minimal settlement layer in Celestia for disputes.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious: check now please

SweeXordious: We can actually post them to non-reserved namespaces. The reason for mentioning reserved here is because, as far as I understand, they are used for non-message data and stuff related to Celestia. @evan-forbes correct me please if I am missing something.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious: No longer necessary is relative. In this case, pruning the state means that newly deployed QGB no  longer necessary contracts will not be able to sync from genesis. Only, from the current unbonding period. If we don't really care about that, then yes, makes sense to  prune.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
rootulp: Yep you're right. Reserved namespaces are for non-message data, specifically transactions, intermediate state roots (eventually), and evidence. Reserved namespaces all use contiguous shares so they are more space efficient but I think the rationale for posting QGB data to a reserved namespace is whether this state needs to be enshrined in the Celestia protocol. If it is the case that the QGB module relies on this state (I think it is) then it's possible that the data does need to posted to a reserved namespace.

If the QGB were an entirely separate application (what you describe in the QGB Rollup section) and nothing in celestia-app depended it, then it seems possible to publish data to a non-reserved namespace.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
rootulp: > No longer necessary is relative.

Sorry I don't follow. Are there scenarios where contracts on Ethereum would need the state from prior to the unbonding period (in other words the state we're considering pruning)?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,97 @@
+# ADR 5: Reduce QGB state
+
+## Changelog
+
+- {date}: {changelog}
+
+## Context
+
+Currently, the QGB is using the state to save all the attestations and confirms it needs.
```
rootulp, line 9: > the state

[optional] it may help non-familiar readers (i.e. me) to clarify what this state is. Is it specifically state stored within the QGB module which would end up bloating the celestia-app state?

rootulp (commented):
Thanks for addressing all my nits! I still don't understand what data needs to be published to a reserved namespace vs. non-reserved namespace but it sounds like you're working on a proposal for that so will re-review after its published

evan-forbes (commented):
just to note some things @SweeXordious and I discussed synchronously:

1. For any new more stateless approach, we need to first have a plausible-potentially-hand-wavy form of slashing. As already mentioned, deciding on _when_ we actually want this formal mechanism for slashing is also important.
2. A more stateless approach could be as simple as keeping identical transaction types, but not storing the confirmations in the state! We'd still have the valset/data commitment requests be handled and stored by the state machine, and the mapping between validators and their orchestrator keys, but the rest would simply be charged gas fees and be stored in block data. Relayers and orchestrators can then reconstruct a local copy of the state using the block data.
3. One hand-wavy-potentially-viable formal slashing mechanism could involve some sort of commit-reveal scheme, where anyone can post a bond and claim that a validator should be slashed for a specific violation. It is then up to any other honest participant to refute that claim. If successful, then they get the bond. This would likely require other additional data be kept in state, but that's still to be investigated. As mentioned, it might be easier to implement fraud proofs (or tendermint like evidence)

note to selves: specifying a formal mechanism for slashing is a perfect thing to do at the onsite.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,97 @@
...
+
+The approach that we were planning to take is to cleanup the state after the unbonding period.
+This way, we will always have a fixed sized state instead of a gigantic one, issue defining this: [Prune the QGB state after the unbonding period ends #309](https://github.com/celestiaorg/celestia-app/issues/309).
+This would mean that the QGB contracts deployed after genesis will never have the whole history of attestations.
+
+### Separate P2P network
+
+This would mean gossiping the confirms and attestations in a separate P2P network.
+The pros of this approach are that it will be cheaper and wouldn't involve any state changes.
+However, slashing will not be possible.
```
evan-forbes, line 32: > slashing will not be possible.

perhaps "very difficult" is more accurate.

the main issue from my point of view with this approach is that it is a different set of liveness guarantees. We have no way of knowing if the orchestrator signed in time as we have no permanent record of when they signed.

evan-forbes (commented):
:rocket:

SweeXordious:
This also will need to be addressed in this ADR: https://github.com/celestiaorg/celestia-app/pull/613#issuecomment-1232870746

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious: Sorry, just seen this message. What I meant by relative is:
- If we want any QGB contract to be able to sync from Genesis, then, we can't prune.
- If we want QGB contracts to only start from the current unbonding period. Then, yes, the old state is not necessary to keep.

rootulp: You explained this well, and I see the difference. I think we need to determine the intended use cases for QGB so that we can make this decision. In other words, we need to understand if QGB contracts would ever want to sync all the state from genesis. Since this is more of a product requirements question, I think it can be a separate issue and doesn't necessarily need to be resolved in this ADR.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious:
```suggestion
Currently, the QGB module is using the Celestia-app state to save all the attestations and confirms it needs.
```
?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,97 @@
...
+
+Proposed
+
+## Consequences
+
+### Positive
+
+- Reduce significantly the gas fees paid by orchestrators.
+- Reduce significantly the use of Celestia state.
+- Keep the whole QGB confirms history via using block data, instead of being forced to delete it after the unbonding period to not end up with a super gigantic state. This would allow any QGB contract to have the whole history on any EVM chain.
```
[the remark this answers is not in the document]
SweeXordious:
```suggestion
- Keep the QGB confirms history via using block data, instead of being forced to delete it after the unbonding period to not end up with a super gigantic state. This would allow any QGB contract to have the whole history on any EVM chain using simple state queries instead of the relayer relying on parsing the blocks to get the past transactions containing the confirms.
```
?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,97 @@
...
+This was done with the goal of not charging orchestrators increasing gas fees for every posted attestation.
+A simple benchmark showed that the gas usage multiplied 2 times from `~50 000` to `100 000` after submitting 16 attestation.
+Also, even if removing the checks was the most practical solution, it ended up opening new attack vectors on the QGB module state, such as flooding the network with incorrect attestations from users who are not even validators.
+Which would increase the burden on validators to handle all of that state.
+Furthermore, it put the responsibility on the relayer to cherry-pick the right confirms from the invalid ones.
+
+This led us to think that the direction we're taking makes more sense if it moves the confirms from the state to the block space.
+This way, we can remove all the complex logic handling the Valset Confirms and DataCommitment Confirms to the orchestrator side, and also block any opportunity for a malicious party to post data to state.
```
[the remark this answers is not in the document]
SweeXordious: I thought the context should be concise? If not, I guess the following can be there:
```
The first design for the QGB was to use the state extensively to store all the QGB related data: Attestations, `Valset Confirms` and `DataCommitment Confirms`.
As a direct consequence of this, we needed to add thorough checks on the state machine level to be sure that any proposed attestation is correct and will eventually be relayed to the target EVM chain.
The following issue lists these checks: [QGB data commitments/valsets state machine checks #631](https://github.com/celestiaorg/celestia-app/issues/631) and here is their [implementation](https://github.com/celestiaorg/celestia-app/blob/d63b99891023d153ea5937e4f3c1907a784654d8/x/qgb/keeper/msg_server.go#L28-L262).
Also, the state machine for the QGB module became complex and more prone to bugs that might end up halting/forking the chain.

In addition to this, with the gas leak issue discussed in this [comment](https://github.com/celestiaorg/celestia-app/issues/631#issuecomment-1220848130), we ended up removing the sanitizing checks we used to run on the submitted `Valset Confirms` and `DataCommitment Confirms`.
This was done with the goal of not charging orchestrators increasing gas fees for every posted attestation.
A simple benchmark showed that the gas usage multiplied 2 times from `~50 000` to `100 000` after submitting 16 attestation.
Also, even if removing the checks was the most practical solution, it ended up opening new attack vectors on the QGB module state, such as flooding the network with incorrect attestations from users who are not even validators.
Which would increase the burden on validators to handle all of that state.
Furthermore, it put the responsibility on the relayer to cherry-pick the right confirms from the invalid ones.

This led us to think that the direction we're taking makes more sense if it moves the confirms from the state to the block space.
```
And do a simple reformulation in the detailed design.
What do you think?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious:
```suggestion
However, slashing will be very difficult especially for liveness, i.e. an orchestrator not signing an attestation then slash them after a certain period.
```

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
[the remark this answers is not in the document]
evan-forbes: I like the changes!

> I thought the context should be concise?

I had a different impression from reading other ADRs, but perhaps I was being too picky. The info being there is the important part so the change is definitely optional

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
evan-forbes: literally perfect!

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,97 @@
...
+- Reduce the state machine complexity
+- Reduce the amount of state used by the QGB
+- Prepare for a more Rollup oriented QGB design
+
+## Alternative Approaches
+
+### Keep the existing design
+
+Keeping the current design would entail using the state extensively.
+This proves bad when the state grows after a few hundred attestations, and performing checks on the `Valset Confirms` and `DataCommitment Confirms` becomes too expensive.
```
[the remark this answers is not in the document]
SweeXordious: check now please?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
```diff
@@ -0,0 +1,97 @@
+# ADR 5: Reduce QGB state
+
+## Changelog
+
+- {date}: {changelog}
+
+## Context
+
+Currently, the QGB is using the state to save all the attestations and confirms it needs.
+We propose to post `Valset Confirms` and `DataCommitment Confirms` in a reserved namespace instead of adding them to the state to achieve the following:
```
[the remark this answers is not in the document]
SweeXordious: check now please?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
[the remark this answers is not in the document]
SweeXordious: check now please?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious: check now please?

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious: Discussion no longer relevant as we will no longer use namespaces. Resolving.

rootulp (commented):
Overall LGTM. I defer to other reviewers for approval on this b/c my approval here doesn't carry much weight

docs/architecture/ADR-005-QGB_reduce_state_usage.md
```diff
@@ -0,0 +1,120 @@
...
+
+This would reduce significantly the QGB module state usage, reduce the complexity of the state machine and give us the same benefits of the current design.
+
+As a direct consequence to this, the relayer will need to adapt to this change and start getting the transactions straight from the blocks, parse them and sanitize the commits. Thus, making the relayer implementation more complex.
+
+However, we can assume that the relayer will have enough computing power to do the latter. Also, only one relayer is necessary to have a working QGB contract. So, the relayer cost is justified.
+
+For the orchestrators, they will also need to parse the history to keep track of any missed signatures. But, same as with the relayers, the cost is justified.
+
+For posting transactions, we will rely on gas fees as a mechanism to limit malicious parties to flood the network with invalid transactions. Then, eventually slash malicious behavious. However, since posting confirms will be possible for any user of the network. It won't be possible to slash ordinary users, who are not running validators, if they post invalid confirms.
```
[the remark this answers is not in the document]
SweeXordious: > If we assume that the QGB data (i.e. Valset Confirms and DataCommitment Confirms) is stored in a non-reserved namespace, I think the problem stated above == https://forum.celestia.org/t/woods-attack-on-celestia/59

We are no longer posting in confirms in namespaces, so woods attack is not relevant, if I understand right.

> If the QGB data is only stored in transaction data and this document proposed adding new transaction types (distinct from MsgPayForData), I think we need a follow-up issue to discuss the metering (i.e. gas accounting) for these new message types.

No need for a new message type, we already have them defined and used. And the gas usage will be minimal since there will be no read/write in state.

evan-forbes (approved):
Nice!! excited about this change.

I don't think its necessary at all, since this ADR is focused on the state machine, but after we flesh out the implementation for the orchestrator and relayer, we might want to add a few more details into how that's done. Very optional in my opinion tho, as this serves the purpose of documenting the decision.

docs/architecture/ADR-005-QGB_posting_confirms_in_reserved_namespace.md
SweeXordious: issue: https://github.com/celestiaorg/celestia-app/issues/683

Issue #659, opened 2022-08-24
EPIC: Universal share prefix


evan-forbes:
We want to switch to more universal share encoding format, similar to what is discussed in  https://github.com/celestiaorg/celestia-core/issues/839. The main benefit of this approach being that parsing shares should be more similar in that we add a varint length delimiter to the total length of contiguous shares. 

> nid (8 bytes) | info byte | message length (varint) | share data for shares at the start of a message;
nid (8 bytes) | info byte | share data for other shares.

This means that the current format for contiguous shares (reserved namespace shares) will look like 

starting shares (last info bit == 1)
```nid (8 bytes) | info byte | message length (varint) | reserved byte | share data``` 
non starting shares (last info bit == 0)
```nid (8 bytes) | info byte | reserved byte | share data``` 

note that depending on what we decide for https://github.com/celestiaorg/celestia-core/issues/842, this issue might be moved to core.

rootulp:
### Currently

#### Contiguous shares 

#### Non-contiguous shares (shares that don't use a reserved namespace. Ex. messages)

### Proposed

#### Contiguous shares (shares that use a reserved namespace. Ex. transactions, ISRs, evidence)

#### Non-contiguous shares (shares that don't use a reserved namespace. Ex. messages)

rootulp:
- [ ] ADR
- [ ] Update reserved namespace shares to use universal share encoding
- [ ] Update non-reserved namespace shares to use universal share encoding
- [ ] Write README in share pkg that describes universal share encoding
- [ ] Update specs

evan-forbes:
I like that idea. our whole naming scheme has kinda been thrown off since we changed `PayForMessage` to `PayForData` (which was a really good change imo). Might just want to refactor the `Message` struct to `ShareData` or something and stick everything in that.

rootulp:
I've been confused on this because I've been using `Message` for message shares and did not think that term applied to shares in the reserved namespace (i.e. transactions, ISRs, evidence).  See https://github.com/celestiaorg/celestia-app/pull/660#discussion_r957457800

If we've been using `Message` to also describe data in reserved namespaces, I totally agree with your suggestion and will create an issue for it.

PR #660, opened 2022-08-24
ADR-002: Universal Share Encoding

rootulp:
ADR version of https://github.com/celestiaorg/celestia-app/issues/659

### DIFF
--- a/docs/architecture/ADR-002-universal-share-encoding.md
+++ b/docs/architecture/ADR-002-universal-share-encoding.md
@@ -1,0 +1,152 @@
+# ADR 009: Universal Share Encoding
+<!-- disable markdownlint MD010 because Go code snippet uses tabs -->
+<!-- markdownlint-disable MD010 -->
+
+## Changelog
+
+- 2022/9/22: inital draft of InfoReservedByte
+- 2022/9/24: update draft to Universal Share Encoding
+
+## Context
+
+The current contiguous (transaction, ISRs, evidence) share format is:
+
+- First share of namespace: `nid (8 bytes) | reserved byte | share data`
+- Contiguous share in namespace: `nid (8 bytes) | share data`
+
+The current non-contigous (message) share format is:
+
+- First share of message: `nid (8 bytes) | message length (varint) | share data`
+- Contiguous share in message: `nid (8 bytes) | share data`
+
+The current share format poses multiple challenges:
+
+1. Clients must have two share parsing implementations (one for contiguous shares and one for non-contiguous shares).
+1. It is difficult to make changes to the share format in a backwards compatible way because clients can't determine which version of the share format an individual share conforms to.
+1. It is not possible for a client that samples a random share to determine if the share is the start of a namespace (for reserved namespaces) / message (for non-reserved namespaces) or a contiguous share for a multi-share namespace / message.
+
+## Proposal
+
+Introduce a universal share encoding that applies to both contiguous and non-contiguous share formats:
+
+- First share of namespace (for reserved namespaces) or message (for non-reserved namespaces): `nid (8 bytes) | info (1 byte)| message length (varint) | data`
+- Contiguous shares in namespace / message: `nid (8 bytes) | info (1 byte)| data`
+
+The contiguous share format has the added constraint:
+
+- First share of namespace: the first byte of `data` is a reserved byte so the format is: `nid (8 bytes) | info (1 byte) | message length (varint) | reserved (1 byte) | data`
+- Contiguous shares in namespace: no additional constraint
+
+Where info byte is a byte with the following structure:
+
+- the first 7 bits are reserved for the version information in big endian form (initially, this will just be 0000000 until further notice);
+- the last bit is a *message start indicator*, that is 1 if the share is at the start of a namespace (for reserved namespaces) / message (for non-reserved namespaces).
+
+Rationale:
+
+1. The first 9 bytes of a share are formatted in a consistent way regardless of the type of share (contiguous or non-contiguous). Clients can therefore parse shares into data via one mechanism rather than two.
+1. The message start indicator allows clients to parse a whole message in the middle of a namespace, without needing to read the whole namespace.
+1. The version bits allow us to upgrade the share format in the future, if we need to do so in such a way that different share formats can be mixed within a block.
+
+## Questions
+
+1. Does the info byte introduce any new attack vectors?
+1. What happens if a block producer publishes a message with a version that isn't in the list of supported versions (initially only `0000000`)?
+
+## Alternative Approaches
+
+// TODO
+
+## Decision
+
+// TODO
+
+## Implementation Details
+
+### Protobuf
+
+1. (Potentially) add `Version` to [`MsgPayForData`](https://github.com/celestiaorg/celestia-app/blob/main/proto/payment/tx.proto#L44)
+
+**NOTE**: Protobuf does not support the byte type (see [Scalar Value Types](https://developers.google.com/protocol-buffers/docs/proto3#scalar)) so a `uint32` will be used for `Version`. Since `Version` is constrained to 2^7 bits (0 to 127 inclusive), a `Version` outside the supported range (i.e. 128) will seriealize / deserialize correctly but be considered invalid by the application. Adding this field increases the size of the message by one byte + protobuf overhead.
+
+### Constants
+
+1. Define a new constant for `InfoReservedBytes = 1`.
+1. Update [`MsgShareSize`](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/pkg/consts/consts.go#L26) to account for one less byte available
+1. Update [`TxShareSize`](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/pkg/consts/consts.go#L24) to account for one less byte available
+
+**NOTE**: Currently constants are defined in celestia-core's [consts.go](https://github.com/celestiaorg/celestia-core/blob/master/pkg/consts/consts.go) but some will be moved to celestia-app's [appconsts.go](https://github.com/celestiaorg/celestia-app/tree/evan/non-interactive-defaults-feature/pkg/appconsts). See [celestia-core#841](https://github.com/celestiaorg/celestia-core/issues/841).
+
+### Types
+
+1. Introduce a new type `InfoReservedByte` to encapsulate the logic around getting the `Version()` or `IsMessageStart()` from a share.
+
+```golang
+// InfoReservedByte is a byte with the following structure: the first 7 bits are
+// reserved for version information in big endian form (initially `0000000`).
+// The last bit is a "message start indicator", that is `1` if the share is at
+// the start of a message and `0` otherwise.
+type InfoReservedByte byte
+
+func NewInfoReservedByte(version uint8, isMessageStart bool) (InfoReservedByte, error) {
+	if version > 127 {
+		return 0, fmt.Errorf("version %d must be less than or equal to 127", version)
+	}
+
+	prefix := version << 1
+	if isMessageStart {
+		return InfoReservedByte(prefix + 1), nil
+	}
+	return InfoReservedByte(prefix), nil
+}
+
+// Version returns the version encoded in this InfoReservedByte.
+// Version is expected to be between 0 and 127 (inclusive).
+func (i InfoReservedByte) Version() uint8 {
+	version := uint8(i) >> 1
+	return version
+}
+
+// IsMessageStart returns whether this share is the start of a message.
+func (i InfoReservedByte) IsMessageStart() bool {
+	return uint(i)%2 == 1
+}
+```
+
+### Logic
+
+#### celestia-core
+
+1. Account for the new `InfoReservedByte` in `./types/share_splitting.go` and `./types/share_merging.go`.
+    - **NOTE**: These files are subject to be deleted soon. See <https://github.com/celestiaorg/celestia-core/issues/842>
+
+#### celestia-app
+
+1. Account for the new `InfoReservedByte` in all share splitting and merging code. There is an in-progress refactor of the relevant files. See <https://github.com/celestiaorg/celestia-app/pull/637>
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+This proposal resolves challenges posed above.
+
+### Negative
+
+This proposal reduces the number of bytes a message share can use for data by one byte.
+
+### Neutral
+
+If 127 versions is larger than required, the share format spec can be updated (in a subsequent version) to reserve fewer bits for the version in order to use some bits for other purposes.
+
+If 127 versions is smaller than required, the share format spec can be updated (in a subsequent version) to occupy multiple bytes for the version. For example if the 7 bits are `1111111` then read an additional byte.
+
+## References
+
+- <https://github.com/celestiaorg/celestia-core/issues/839>
+- <https://github.com/celestiaorg/celestia-core/issues/759>
+- <https://github.com/celestiaorg/celestia-core/issues/757>
+- <https://github.com/celestiaorg/celestia-app/issues/659>


docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,152 @@
...
+Rationale:
+
+1. The first 9 bytes of a share are formatted in a consistent way regardless of the type of share (contiguous or non-contiguous). Clients can therefore parse shares into data via one mechanism rather than two.
+1. The message start indicator allows clients to parse a whole message in the middle of a namespace, without needing to read the whole namespace.
+1. The version bits allow us to upgrade the share format in the future, if we need to do so in such a way that different share formats can be mixed within a block.
+
+## Questions
+
+1. Does the info byte introduce any new attack vectors?
+1. What happens if a block producer publishes a message with a version that isn't in the list of supported versions (initially only `0000000`)?
```
rootulp, line 54: It seems like this could be a `ProcessProposal` validity check. Validators already compute the shares in `ProcessProposal` [here](https://github.com/rootulp/celestia-app/blob/ad050e28678119adae02536db3ef5ce083ea1436/app/process_proposal.go#L104-L110) so we can add a check to verify that every share has a valid version.

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Decision
+
+// TODO
+
+## Implementation Details
+
+### Protobuf
+
+1. (Potentially) add `Version` to [`MsgPayForData`](https://github.com/celestiaorg/celestia-app/blob/main/proto/payment/tx.proto#L44)
```
evan-forbes, line 68: I was wrong on this before! we definitely don't need to now

evan-forbes (commented):
Nice!! I like combining the two features into a single ADR, I might mention in the detailed design part that these are distinct features and can be implemented as such.

As we flesh this out, I look forward to your thoughts over what the new API will look like, and what our approach should be to handle that API change :slightly_smiling_face:

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,152 @@
...
+
+- 2022/9/22: inital draft of InfoReservedByte
+- 2022/9/24: update draft to Universal Share Encoding
+
+## Context
+
+The current contiguous (transaction, ISRs, evidence) share format is:
+
+- First share of namespace: `nid (8 bytes) | reserved byte | share data`
+- Contiguous share in namespace: `nid (8 bytes) | share data`
```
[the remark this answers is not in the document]
rootulp: you're right, thanks for commenting! It looks like the reserved byte always exists but it is set to `0` if there is no transaction that starts in the share. Updated ADR to resolve this in https://github.com/celestiaorg/celestia-app/pull/660/commits/fb338a15aa26e502f10a5cbfd9167d8a5bbf8d65

docs/architecture/ADR-002-universal-share-encoding.md
rootulp: Totally agreed because user's won't set a `Version` when they submit a `PayForData`. Instead, block producers will set a version at data square construction time.  Resolved in https://github.com/celestiaorg/celestia-app/pull/660/commits/fb338a15aa26e502f10a5cbfd9167d8a5bbf8d65

rootulp:
Will update this ADR to replace the contiguous / non-contiguous share lingo in favor of reserved namespace / unreserved namespace shares

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -13,36 +13,36 @@
 **reserved**: is the location of the first transaction, ISR, or evidence in this share if there is one and `0` if there isn't one
 **message length**: is the length of the entire message in bytes
 
-The current contiguous (transaction, ISRs, evidence) share format is:<br>`nid (8 bytes) | reserved (1 byte) | share data`
+The current reserved namespace (transaction, ISRs, evidence) share format is:<br>`nid (8 bytes) | reserved (1 byte) | share data`
```
evan-forbes, line 16: :+1:

evan-forbes: Screenshot from 2022-08-25 16-18-38

evan-forbes: after sleeping on this, while I really liked it at first, contiguous seems more descriptive as opposed to describing an implementation detail (reserved namespace). Also, is there anything stopping us from using a reserved namespace in a non-contiguous way (doesn't need the reserved byte)?

rootulp: > Also, is there anything stopping us from using a reserved namespace in a non-contiguous way

I don't think so besides updating the spec and implementation.

Agreed, "reserved" vs. "unreserved" is an implementation detail. My rationale for renaming from contiguous shares to reserved namespace shares is to disambiguate this reference from the separate but related term contiguous message shares. If possible, I think we should avoid using "contiguous" for both types of things. Message shares can be contiguous in that they are split across multiple shares but all part of one message. One alternative naming ideas for reserved vs. unreserved is non-message shares vs. message shares but that's also an implementation detail.

We want to use a word that describes how:
- one type of share can contain multiple things (transactions, ISRs, evidence) 
- one type of share can only contain 0 or 1 of a thing (messages)

I'm open to ideas

evan-forbes: how about monogamous shares vs polygamous shares? mono-shares poly-shares?

rootulp: - mono-content share vs poly-content shares?
- atomic-content share vs compound-content shares?

I don't like any of these as much as reserved vs non-reserved shares because readers now need to learn that these new terms map to the existing reserved vs non-reserved shares. In the future, if we introduce a reserved namespace that doesn't conform to the proposed share format (i.e. it doesn't support contiguous shares) then we can revise our terminology then, similar to how we introduced the notion of term: "universal share encoding"

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,161 @@
...
+1. Clients must have two share parsing implementations (one for reserved namespace shares and one for unreserved namespace shares).
+1. It is difficult to make changes to the share format in a backwards compatible way because clients can't determine which version of the share format an individual share conforms to.
+1. It is not possible for a client that samples a random share to determine if the share is the first share of that namespace or a contiguous share.
+
+## Proposal
+
+Introduce a universal share encoding that applies to both reserved and unreserved share formats:
+
+- First share of namespace (for reserved namespaces) or message (for unreserved namespaces):<br>`nid (8 bytes) | info (1 byte) | message length (varint 1 to 10 bytes) | data`
+- Contiguous shares in namespace:<br>`nid (8 bytes) | info (1 byte) | data`
```
musalbas, line 34: Should rather be contiguous shares in message?

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,161 @@
...
+
+Where **info** (1 byte) is a byte with the following structure:
+
+- the first 7 bits are reserved for the version information in big endian form (initially, this will be `0000000` for version 0);
+- the last bit is a *message start indicator*, that is `1` if the share is at the start of a namespace or `0` if it is a contiguous share within a namespace.
+
+Rationale:
+
+1. The first 9 bytes of a share are formatted in a consistent way regardless of the type of share (reserved or unreserved namespace). Clients can therefore parse shares into data via one mechanism rather than two.
+1. The message start indicator allows clients to parse a whole share in the middle of a namespace, without needing to read the whole namespace.
```
musalbas, line 46: Whole message, rather than whole share

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,161 @@
...
+The current unreserved namespace (message) share format is:
+
+- First share of message:<br>`nid (8 bytes) | message length (varint 1 to 10 bytes) | share data`
+- Contiguous share in message:<br>`nid (8 bytes) | share data`
+
+The current share format poses multiple challenges:
+
+1. Clients must have two share parsing implementations (one for reserved namespace shares and one for unreserved namespace shares).
+1. It is difficult to make changes to the share format in a backwards compatible way because clients can't determine which version of the share format an individual share conforms to.
+1. It is not possible for a client that samples a random share to determine if the share is the first share of that namespace or a contiguous share.
```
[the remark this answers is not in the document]
musalbas: By the way the use case isn't really about the client sampling a random share. It's more that nodes can prove that specific message in a namespace is "complete" (i.e. the node was provided all the data for that message).

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,161 @@
...
+## Changelog
+
+- 2022/9/22: inital draft of InfoReservedByte
+- 2022/9/24: update draft to Universal Share Encoding
+
+## Context
+
+- **nid** (8 bytes): namespace id
+- **reserved** (1 byte): is the location of the first transaction, ISR, or evidence in this share if there is one and `0` if there isn't one
+- **message length** (varint 1 to 10 bytes): is the length of the entire message in bytes
```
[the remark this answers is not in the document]
rootulp: Yep, the maximum value the message length varint can be is 2^64 which takes up 10 bytes.

```golang
	MaxVarintLen64 = 10
```

Ref:
- https://github.com/rootulp/celestia-app/blob/3ddf2ddcc7c7061b84b2295e4a2e07f55472597c/pkg/shares/shares.go#L42
- https://pkg.go.dev/encoding/binary#pkg-constants

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,161 @@
...
+The current reserved namespace (transaction, ISRs, evidence) share format is:<br>`nid (8 bytes) | reserved (1 byte) | share data`
+
+The current unreserved namespace (message) share format is:
+
+- First share of message:<br>`nid (8 bytes) | message length (varint 1 to 10 bytes) | share data`
+- Contiguous share in message:<br>`nid (8 bytes) | share data`
+
+The current share format poses multiple challenges:
+
+1. Clients must have two share parsing implementations (one for reserved namespace shares and one for unreserved namespace shares).
```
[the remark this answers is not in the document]
rootulp: Clarifying question: in this context does "raw message data" apply to the raw data for message shares and for reserved namespace shares (i.e. transactions)?

In other words, is "message" applicable to message shares and for reserved namespace shares? I was confused by the "message" terminology used in https://github.com/celestiaorg/celestia-core/issues/839 because I've been thinking of the two types of shares as different and "message" only applying to message shares. 

> for shares at the start of a message;

docs/architecture/ADR-002-universal-share-encoding.md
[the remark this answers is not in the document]
rootulp: thanks for clarifying. If I understand correctly, the situation where message start indicator is of interest:
- multiple messages with the same namespace are included in the same block
- a client wants to know if one of the messages in this namespace was included in the block
- (potentially) the share of interest spans multiple shares

## illustrated example

```golang
var ns = []byte{1, 2, 3, 4, 5, 6, 7, 8}
var version = `0000000`
```
| share number            | 10      | 11      | 12              | 13      |
| ----------------------- | ------- | ------- | --------------- | ------- |
| namespace               | ns      | ns      | ns              | ns      |
| version                 | version | version | version         | version |
| message start indicator | `1`     | `1`     | `0`             | `1`     |
| data                    | foo     | bar     | bar (continued) | buzz    |


if a client is provided share 11 (with the universal share format) the client can parse the message length delim in share 11, then they know that the **bar** message will complete after reading N bytes (where N includes shares 11 and 12). This was previously impossible because a client couldn't know whether the message length delim was included in a share without parsing previous shares.

An alternative way to prove the same thing that doesn't rely on parsing the message length delim is to include a continuation share bit that is `1` when there is a continuation share after the current share and `0` when the current share ends a message. This option may be favorable in environments where parsing varints is difficult but that motivation seems disproved by https://github.com/celestiaorg/celestia-core/issues/757#issuecomment-1135592765

docs/architecture/ADR-002-universal-share-encoding.md
rootulp: On second thought, I see why namespace could be confusing in the scenario where multiple messages for the same namespace are included in a block.

Is this clearer?
```suggestion
- Contiguous shares in namespace (for reserved namespaces) or message (for unreserved namespaces):<br>`nid (8 bytes) | info (1 byte) | data`
```

docs/architecture/ADR-002-universal-share-encoding.md
evan-forbes: > We want to use a word that describes how:
one type of share can contain multiple things (transactions, ISRs, evidence)
one type of share can only contain 0 or 1 of a thing (messages)

would prefer something that does this.

personally, I don't like that we're using the same word to represent adjacent but different things (reserved namespaces and reserved byte). I can see that being confusing to new readers. I'd also prefer to avoid having to change names in the future.  Again, I feel like that will only add to the confusion of future readers.

since this is an opinionated thing that we don't have data for and could very well be a premature optimization, its not worth bike-shedding any further imo, and I'll definitely defer to the implementer's preference and other reviewers :slightly_smiling_face:

rootulp: Apologies for the several rounds of feedback on this issue. I think it's important for us to agree on a term to describe this. 

> I think that using the same word to represent adjacent but different things (reserved namespaces and reserved byte) could be confusing to new readers.

Thanks for articulating, I can now also see this being confusing to new readers. I think reserved namespace shares is superior to contiguous shares but still open to alternatives. One more suggestion that comes to mind:

- compact shares vs sparse shares

> compact: joined or packed together; closely and firmly united; [dense](https://www.dictionary.com/browse/dense); [solid](https://www.dictionary.com/browse/solid):

Source: https://www.dictionary.com/browse/compact

I think @adlerjohn is precise with language so I'm curious if he has opinions on this

evan-forbes: compact shares vs sparse shares is good

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/660/commits/c9a2fd52a04df1302a236c7efb613e2d9a6d28fc

docs/architecture/ADR-002-universal-share-encoding.md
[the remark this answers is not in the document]
musalbas: Yes, exactly.

docs/architecture/ADR-002-universal-share-encoding.md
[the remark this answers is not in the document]
musalbas: Yes, by "raw message data" I also mean the bytes in the reserved namespace that fall outside of the "universal share format". Not sure what is a better terminology.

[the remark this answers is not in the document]
rootulp: Proposal: "raw share data"

Related: https://github.com/celestiaorg/celestia-core/issues/848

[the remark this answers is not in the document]
musalbas: Hmm share data might also be confusing, because to me that sounds like it's the entire share, because in erasure coding terminology share means the entire share.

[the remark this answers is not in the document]
rootulp: I see. Given that context, "share data" isn't accurate because it describes all of the share and the term we're looking for only applies to the portion of the share that excludes the universal prefix. 

Alternative proposals: 
- share contents
- share suffix

docs/architecture/ADR-002-universal-share-encoding.md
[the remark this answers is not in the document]
rootulp: Thanks for confirming. I added the example in https://github.com/celestiaorg/celestia-app/pull/660/commits/82191cfa0e5fe5c02e49180aaea2f5726a3955bf

docs/architecture/ADR-002-universal-share-encoding.md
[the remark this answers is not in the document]
rootulp: Resolving in favor of https://github.com/celestiaorg/celestia-core/issues/848

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,141 @@
...
+
+- **nid** (8 bytes): namespace id
+- **reserved** (1 byte): is the location of the first transaction, ISR, or evidence in this share if there is one and `0` if there isn't one
+- **message length** (varint 1 to 10 bytes): is the length of the entire message in bytes
+- **compact share**: a type of share that can accomodate multiple units. Currently, compact shares are used for transactions, ISRs, and evidence to efficiently pack this information into as few shares as possible.
+- **sparse share**: a type of share that can accomodate zero or one unit. Currently, sparse shares are used for messages.
+
+## Context
+
+The current compact share format is:<br>`nid (8 bytes) | reserved (1 byte) | data`
```
rootulp, line 15: This statement isn't congruent with [this implementation](https://github.com/celestiaorg/celestia-app/blob/main/pkg/shares/parse_contiguous_shares.go#L49). In particular, it looks like data length is encoded as a varint into compact shares.

```suggestion
The current compact share format is:

- First share of namespace: <br>`nid (8 bytes) | reserved (1 byte) | message length (varint 1 to 10 bytes) | data`
- Contiguous share in namespace:<br>`nid (8 bytes) | reserved (1 byte) | data`
```

docs/architecture/ADR-002-universal-share-encoding.md
rootulp: This discussion motivated https://github.com/celestiaorg/celestia-app/issues/688

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,144 @@
...
+
+- First share of namespace for compact shares or message for sprase shares:<br>`nid (8 bytes) | info (1 byte) | data length (varint 1 to 10 bytes) | data`
+- Contiguous share in namespace for compact shares or message for sparse shares:<br>`nid (8 bytes) | info (1 byte) | data`
+
+Shares in the reserved namespace have the added constraint: the first byte of `data` is a reserved byte so the format is:<br>`nid (8 bytes) | info (1 byte) | data length (varint 1 to 10 bytes) | reserved (1 byte) | data`
+
+Where **info** (1 byte) is a byte with the following structure:
+
+- the first 7 bits are reserved for the version information in big endian form (initially, this will be `0000000` for version 0);
+- the last bit is a **message start indicator**, that is `1` if the share is at the start of a message or `0` if it is a contiguous share within a message.
```
rootulp, line 43: [naming] The message start indicator will be included in the `InfoReservedByte` that is used for compact shares so it seems appropriate to drop the "message" from this name.

[question] when should the start indicator be `1` for a compact share? Two options:
1. For the first share in the namespace (i.e. the first transaction share in the transaction namespace)
2. For any share in the namespace that contains the start of a unit. (i.e. for the transaction namespace, in any share where the reserved byte is non `0`)

Tagging @evan-forbes if you have insight on the above. The current implementation [ref](https://github.com/celestiaorg/celestia-app/pull/691/files#diff-3da7e413fef3c90c01286110ba9bdfc3d61bbb360355bb32adede4e7b3ebfe5bR27) handles **option 1** but it can be modified if option 2 is preferred.

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,144 @@
...
+1. It is not possible for a client that samples a random share to determine if the share is the first share of a message or a contiguous share in the message.
+
+## Proposal
+
+Introduce a universal share encoding that applies to both compact and sparse shares:
+
+- First share of namespace for compact shares or message for sprase shares:<br>`nid (8 bytes) | info (1 byte) | data length (varint 1 to 10 bytes) | data`
+- Contiguous share in namespace for compact shares or message for sparse shares:<br>`nid (8 bytes) | info (1 byte) | data`
+
+Shares in the reserved namespace have the added constraint: the first byte of `data` is a reserved byte so the format is:<br>`nid (8 bytes) | info (1 byte) | data length (varint 1 to 10 bytes) | reserved (1 byte) | data`
```
rootulp, line 38: [clarification] `data length (varint 1 to 10 bytes)` is the sum of the length of all units in a particular compact share namespace (i.e. all transactions in the transaction namespace). This is distinct the from transaction length varint that is prefixed to an individual transaction.

If so https://github.com/celestiaorg/celestia-app/pull/691 doesn't currently handle this.

rootulp: `data length (varint 1 to 10 bytes)` is confusing b/c for compact shares it is the sum of lengths for all units in a particular namespace but for messages, it is the length of a message which isn't necessarily the sum of lengths of all messages in the namespace.

docs/architecture/ADR-002-universal-share-encoding.md
evan-forbes: The important thing in an implementation would be to treat everything as that struct. So we could not name it different, and then treat everything as a "Message". I don't really have a firm opinion yet, dropping the message makes sense too.

> [question] when should the start indicator be 1 for a compact share? Two options:

yes definitely 1 and not 2. 

I don't think this is occurring, but tbc on wording for other reviewers
> For the first share in the namespace

is only for the current implementation of info byte for _compact_ shares, the implementation of universal shares will obvi not only be for the first share in a namespace

evan-forbes: well shoot, maybe we need to change the wording in the adr to make that clearer? the entire point of universal shares is that we treat the splitting of each namespace identically for the first step, so saying

For the first share in the namespace

could be confusing.

rootulp: > The fact that there is only a single message in reserved namespaces is purely an application detail 

Agreed. 

Since compact shares contain a reserved byte, we need some way of distinguishing between sparse shares and compact shares. In other words, since compact shares are an application detail, parsing them requires separate logic from sparse shares. Although the API to `ParseShare` can be used for both compact and sparse shares, the API still needs to be able to differentiate the type of share being parsed which I assume will be based on the namespace id.

Note: we can enforce an additional requirement that there is only one compact share (i.e. the first) per reserved namespace with the `start indicator=1`.

But the terminology used in this ADR is confusing with respect to "For the first share in the namespace", will try to improve.

musalbas: >the API still needs to be able to differentiate the type of share being parsed which I assume will be based on the namespace id

Why is that?

docs/architecture/ADR-002-universal-share-encoding.md
musalbas: I think the trick is that conceptually we think of all the data in a reserved namespace as a single message.

rootulp: Created https://github.com/celestiaorg/celestia-app/issues/725 as a follow-up but great point @musalbas , will improve wording in this ADR based on that

docs/architecture/ADR-002-universal-share-encoding.md
rootulp: I think to parse the reserved byte that is only encoded into compact shares

musalbas: That reserved byte should be part of the raw share data so there shouldn't be a need to parse it to parse a share

rootulp: Agreed parsing the reserved byte is only necessary after parsing the universal prefix. 

I think it's possible to improve the terminology here but given other readers are familiar with "message" as describing the contents of a share (even if the share contains non-message data, i.e. transactions), I think it's safe to resolve my thread on [naming]

docs/architecture/ADR-002-universal-share-encoding.md
```diff
@@ -0,0 +1,143 @@
+# ADR 009: Universal Share Encoding
```
rootulp, line 1:
```suggestion
# ADR 009: Universal Share Prefix
```

docs/architecture/ADR-002-universal-share-encoding.md
rootulp: Unresolving b/c I think the naming is still confusing. What are reviewers thoughts on renaming the message start indicator to payload start indicator or data start indicator? @evan-forbes @musalbas 

The rename is motivated by an attempt at conveying that the start indicator is applicable to shares without message data (i.e. compact shares).

musalbas:
I don't have any strong opinions. That being said, I would consider compact shares to have one "message".

rootulp:
What is the definition of "message"? I interpreted it as the previous name for data that a user submits to be included in a block (i.e. the second message in `MsgPayForMsg` which was renamed `MsgPayForData`). Maybe in this context, "message" means something else because we're now using it to describe evidence, intermediate state roots, and transactions. 

> the idea that data in reserved namespace follow the same share-level format as any other namespace.

Agreed and I like the share-level format terminology. I think we can split the format into:
1. Share-level
    1. namespace
    2. info reserved byte
    3. [if info reserved byte has `message start indicator=1`] message length varint
2. Application-level
    1. For sparse shares: no additional format imposed
    2. For compact shares: 1 reserved byte for location of first unit in the share and every unit is prefixed with a varint of the unit's length

if this separation adds clarity

docs/architecture/adr-007-universal-share-prefix.md
```diff
@@ -0,0 +1,148 @@
+# ADR 007: Universal Share Prefix
+
+## Terminology
+
+- **compact share**: a type of share that can accomodate multiple units. Currently, compact shares are used for transactions, ISRs, and evidence to efficiently pack this information into as few shares as possible.
```
liamsi, line 5:
```suggestion
- **compact share**: a type of share that can accommodate multiple units. Currently, compact shares are used for transactions, and evidence to efficiently pack this information into as few shares as possible.
```
Typo and I think we got rid of ISRs?

docs/architecture/adr-007-universal-share-prefix.md
```diff
@@ -0,0 +1,148 @@
...
+- **sparse share**: a type of share that can accomodate zero or one unit. Currently, sparse shares are used for messages.
```
liamsi, line 6:
```suggestion
- **sparse share**: a type of share that can accommodate zero or one unit. Currently, sparse shares are used for messages.
```

liamsi (commented):
Well written and accessible ADR. Left some nits. The proposed scheme is clear and it looks good to me.

docs/architecture/adr-007-universal-share-prefix.md
rootulp: The current implementation doesn't contain intermediate state roots. 

https://github.com/rootulp/celestia-app/blob/375093397458a32316a41666264fd2e9d6111219/pkg/appconsts/appconsts.go#L85-L88

Do we expect to include intermediate state roots in the future?

liamsi (approved):
I think this is good to go :)

PR #667, opened 2022-08-26
Update protobuf generator

### LINKED ISSUE #553: Update the protobuf generator to be the same as Cosmos-sdk one
### Summary

The current protobuf generator used for celestia-app is old and needs to be updated

### Problem Definition

We currently rely on an old version of `tendermintdev/sdk-proto-gen` to generate protobuf wrappers.
This gives us issues when using for example the `Any` type implementation from the cosmos-sdk.
Also, it would be good to be using the same version as cosmos-sdk. 

### Proposal

Update the version and make necessary changes

### For Admin Use

- [ ] Not duplicate issue

rahulghangas:
Closes #553

Update protbuf generator to follow the version in cosmos-sdk upstream - published `v0.45.8`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ test:
 .PHONY: test
 
 proto-gen:
-	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.2 sh ./scripts/protocgen.sh
+	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.7 sh ./scripts/protocgen.sh
 
 proto-lint:
 	@$(DOCKER_BUF) lint --error-format=json
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ test:
 .PHONY: test
 
 proto-gen:
-	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.7 sh ./scripts/protocgen.sh
+	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:latest sh ./scripts/protocgen.sh
 
 proto-lint:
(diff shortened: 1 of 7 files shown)


evan-forbes:
@SweeXordious, when you get the chance, can you see if this messes with the qgb-integration branch, or perhaps even fixes our issue?

Makefile
```diff
@@ -49,7 +49,7 @@ test:
 .PHONY: test
 
 proto-gen:
-	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.2 sh ./scripts/protocgen.sh
+	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.7 sh ./scripts/protocgen.sh
```
SweeXordious, line 52: Wouldn't it make more sense if we use latest, instead of specifying a version? To know when we break compatibility again

proto/buf.gen.gogo.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
+plugins:
+  - name: gocosmos
+    out: ..
+    opt: plugins=interfacetype+grpc,Mgoogle/protobuf/any.proto=github.com/celestiaorg/celestia-app/codec/types
+  - name: grpc-gateway
+    out: ..
+    opt: logtostderr=true,allow_colon_final_segments=true
```
SweeXordious, line 8: Was this generated automatically? If not, I would prefer if we end a line at  the end of the file:

```suggestion
    opt: logtostderr=true,allow_colon_final_segments=true
    
```

SweeXordious (commented):
utACK. Great work fixing this, will also test it on the QGB side and see

SweeXordious:
@evan-forbes QGB branch is working fine with this change. However, it  is not fixing the `Any` type problem that we have. We might need to probably tweak the configuration.

evan-forbes (approved):
thanks for checking @SweeXordious :slightly_smiling_face: 

:+1:

SweeXordious (approved):
Awesome stuff :+1:

PR #671, opened 2022-08-29
Implement InfoReservedByte

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/659

### DIFF
--- a/pkg/shares/info_reserved_byte.go
+++ b/pkg/shares/info_reserved_byte.go
@@ -1,0 +1,33 @@
+package types
+
+import "fmt"
+
+// InfoReservedByte is a byte with the following structure: the first 7 bits are
+// reserved for version information in big endian form (initially `0000000`).
+// The last bit is a "message start indicator", that is `1` if the share is at
+// the start of a message and `0` otherwise.
+type InfoReservedByte byte
+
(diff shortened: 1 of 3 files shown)


pkg/shares/info_reserved_byte.go
```diff
@@ -0,0 +1,33 @@
...
+import "fmt"
+
+// InfoReservedByte is a byte with the following structure: the first 7 bits are
+// reserved for version information in big endian form (initially `0000000`).
+// The last bit is a "message start indicator", that is `1` if the share is at
+// the start of a message and `0` otherwise.
+type InfoReservedByte byte
+
+func NewInfoReservedByte(version uint8, isMessageStart bool) (InfoReservedByte, error) {
+	if version > 127 {
```
evan-forbes, line 12: not sure if its necessary, but perhaps we could have a constant for the max version? wdyt?

rootulp: Yes I think that's a great idea. I'll put it in `pkg/appconsts/appconsts.go` unless there is a better location for this

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/671/commits/54dc074b139efb23dc1ccc17b9eb5743b89fb1b4

evan-forbes (approved):
:+1:

PR #673, opened 2022-08-30
ADR 006: Non-interactive Defaults, Wrapped Txs, and Subtree Root Message Inclusion Checks

### LINKED ISSUE #598: ADR: Wrapped txs, non-interactive defaults, and sub tree root based message inclusion checks
>Can we add a very short ADR for this as well. I'm also wondering what the best way to document this. Maybe the ADR could simply reference and slightly augment the existing spec when describing the decision here?

(issue derived from this [comment](https://github.com/celestiaorg/celestia-core/issues/462#issuecomment-1119101621) by @liamsi)

While the implementation specifically for wrapped txs is very simple, we should write an ADR for how/why wrapped txs are used in the app. Specifically their usage when performing message inclusion checks and their involvement in non-interactive defaults.

evan-forbes:
# ADR 006
#### Non-interactive Defaults, Wrapped Txs, and Subtree Root Message Inclusion Checks

[rendered](https://github.com/celestiaorg/celestia-app/blob/63b0ff3671e20e397b951f3ff82a93d3e36e84c6/docs/architecture/ADR-003-Non-interactive-defaults.md)

closes #598


docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+- Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
+- If the largest power of 2 of a given message spans multiple rows it must begin at the start of a row (this can occur if a message is longer than k shares or if the block producer decides to start a message partway through a row and it cannot fit).
+
+We can always create a commitment to the data that is a subtree root of the data root while only knowing the data in that message. Below illustrates how we can break a message up into two different subtree roots, the first for first four shares, the second consisting of the last two shares.
+
+![before](./assets/subtree-root.png "Subtree Root based commitments")
+
+In practice this means that we end up adding padding between messages (zig-zag hatched share). Padding consists of a the namespace of the message before it, with all zeros for data.
```
rootulp, line 19:
```suggestion
In practice this means that we end up adding padding between messages (zig-zag hatched share). Padding consists of shares with the namespace of the message before it, with all zeros for data.
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,49 @@
...
+![after](./assets/after.png "after")
+![example](./assets/example-full-block.png "example")
+
+
+Not only does doing this allow for easy trust minimized message inclusion checks for specific messages by light clients, but also allows for the creation of message inclusion fraud proofs for all messages in the block. 
+
+
+## Alternative Approaches
+
+Arranging the messages in the block to maximize for fees is an NP-hard problem because each change to the square potentially affects the rest of the messages in the square. There will likely be many different strategies we could use to quickly and efficiently fill the square in a valid way.
```
rootulp, line 31: [no change needed] it does seem NP-hard per https://en.wikipedia.org/wiki/Bin_packing_problem

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,49 @@
...
+
+While this functions as a message inclusion check, the light client has to assume that 2/3's of the voting power is honest in order to be assured that both the messages they are interested in and the rest of the messages paid for in that block are actually included.
+
+To fix this, the spec outlines the “non-interactive default rules”. These involve a few additional message layout rules that allow for commitments to messages to consist entirely of sub tree roots of the data root, and for those sub tree roots to be generated only from the message itself (so that the user can sign something “non-interactively”). NOTE: MODIFIED FROM THE SPEC
+
+
+- Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
+- If the largest power of 2 of a given message spans multiple rows it must begin at the start of a row (this can occur if a message is longer than k shares or if the block producer decides to start a message partway through a row and it cannot fit).
```
rootulp, line 13: [question] I'm confused by this rule's application in `example-full-block.png` for `ns=2`. `ns=2` is a message that spans 7 shares. The largest power of 2 less than the message length (7) or k (8) is 4. The message is aligned with index 4.

> If the largest power of 2 of a given message spans multiple rows

The largest power of 2 for this message (4) doesn't span multiple rows but this message does because the block producer decided to start this message partway through the row and it cannot fit. If `ns=2` is indeed supposed to be split across rows, then should this read:
```suggestion
- If the largest power of 2 of a given message is larger than k then it must begin at the start of a row.
```

If `ns=2` is supposed to start at a new row, I think the current language makes sense.

## example-full-block
example-full-block

rootulp (commented):
I am SO happy about these diagrams b/c they're really helping me develop an intuition for non-interactive defaults. Thanks for creating them!

evan-forbes:
this draft is drafty that I don't think its worth review

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: Still thinking about this, but I think the first rule might be the only one we need. The we can just go into more detail of a what an "aligned power of two" actually means

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+       }
+   }
+}
+```
+
+#### Pruning excess transactions
+
+If there are too many transactions and messages in the square to fit in the max square size, then we have to remove them from the block. This can be complicated, as by default we want to prioritzie transactions that have higher fees, but removing a low fee transaction doesn't always result in using less shares.
+
+The simplest approach, and the one taken in the initial implementation, works by prematurely pruning the txs if we estimate that too many shares are being used. While this does work, and fulfills the constraints discussed earlier to create valid blocks, it is suboptimal. Ideally we would be able to identify the most optimal message and transactions to remove and simply remove only those. As mentioned earlier, technically, a single byte difference could change the entire arrangement of the square. Which makes arranging the square with complete confidence difficult not only because we have to follow all of the constraints, but also because of our frequent reliance on variable length length delimiters, and protofuf changing the amount of bytes used depending on the size of ints/uints.
```
rootulp:
```suggestion
The simplest approach, and the one taken in the initial implementation, works by prematurely pruning the txs if we estimate that too many shares are being used. While this does work, and fulfills the constraints discussed earlier to create valid blocks, it is suboptimal. Ideally we would be able to identify the most optimal message and transactions to remove and simply remove only those. As mentioned earlier, technically, a single byte difference could change the entire arrangement of the square. Which makes arranging the square with complete confidence difficult not only because we have to follow all of the constraints, but also because of our frequent reliance on variable length length delimiters, and protobuf changing the amount of bytes used depending on the size of ints/uints.
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+## Intro
+
+pls see the [original specs](https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md), from which this ADR paraphrases heavily.
+
+Currently, when checking for message inclusion, validators recreate the share commitment from the messages in the block and compare those with what are signed over in the `MsgPayForData` transactions also in that block. If any commitment is not found in one of the PFD transactions, or if there is a commitment that doesn't have a corresponding message, then they reject that block.
+
+While this functions as a message inclusion check, the light client has to assume that 2/3's of the voting power is honest in order to be assured that both the messages they are interested in and the rest of the messages paid for in that block are actually included. In order to have this property, we need a block (or tx? pls halp reviewers) validity rule where:
```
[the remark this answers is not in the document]
evan-forbes: I agree that it should probably be block. 1) that's my interpretation of the specs, 2) it makes msg inclusion fraud proofs waaaay easier 3) its just simpler.

After think about it more, we would need immediate execution in order to make it a transaction validity rule. Along with a block validity rule that state that no invalid PFDs are allowed. We'll need immediate execution and the above for trusted fee burning, and then state fraud proofs to make that block validity rule trust minimized.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+  if count == 0 {
+      return
+  }
+  lastMessage := msw.shares[len(msw.shares)-1]
+  msw.shares = append(msw.shares, namespacedPaddedShares(lastMessage[0].ID, count))
+  msw.count += count
+}
+```
+
+Now we simply combine this new functionality with the `share_index`s described above, and we can properly split and pad messages when needed. Note, the below implementation allows for `nil` to be passed as indexes. This is important, as it allows the same implementation to be used in the cases where we don't want to split messages using wrapped transactions, such as supporting older networks or when users create commitments to sign over for `MsgWirePayForData`
```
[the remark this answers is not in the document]
evan-forbes: yeah I'm cool with this, being explicit is rarely bad. Although, I'm not sure I would want to commit to always doing this in the future.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: it won't let me commit that suggestion, but fixed https://github.com/celestiaorg/celestia-app/pull/673/commits/d370eeb7f7b8c9a3b19a490eb2a3fbf88a3bd991

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: also, I removed this comment

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+> **Messages must begin at a location aligned with the largest power of 2 that is not larger than the message length or k.**
+
+![Subtree root commitment](./assets/subtree-root.png "Subtree Root based commitments")
+
+> **If the messages are larger than k, then they must start on a new row.**
+
+![Multi row share commitment](./assets/multirow.png "Subtree Root based commitments")
+
+If we can follow these rules, we can always create a commitment to the data that is a subtree root of the data root while only knowing the data in that message. Below illustrates how we can break a message up into two different subtree roots, the first for first four shares, the second consisting of the last two shares.
```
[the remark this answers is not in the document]
evan-forbes: this was remedied in dd0ac52de7aeb1fe26ba8ee7e72848878eac3cf1

good catch tho

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+The main issue with that requirement is that it requires the users to know the relevant subtree roots before they sign, which is problematic considering that is the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.
+
+To fix this, the spec outlines the “non-interactive default rules”. These involve a few additional message layout rules that allow for commitments to messages to consist entirely of sub tree roots of the data root, and for those sub tree roots to be generated only from the message itself (so that the user can sign something “non-interactively”). NOTE: MODIFIED FROM THE SPEC
+
+> **Messages must begin at a location aligned with the largest power of 2 that is not larger than the message length or k.**
+
+![Subtree root commitment](./assets/subtree-root.png "Subtree Root based commitments")
+
+> **If the messages are larger than k, then they must start on a new row.**
```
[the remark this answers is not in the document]
evan-forbes: hmmm idk, @adlerjohn is good with this sort of concise wording. I tried to stick to as close to what is written in the specs

> Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.

the rule was mainly just modified to be first, as I felt it was more important since we're only following a portion of the other rule now

> Messages that span multiple rows must begin at the start of a row (this can occur if a message is longer than k shares or if the block producer decides to start a message partway through a row and it cannot fit).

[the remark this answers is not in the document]
evan-forbes: but we should probably be more explicit about that in the ADR come to think of it

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+Currently, when checking for message inclusion, validators recreate the share commitment from the messages in the block and compare those with what are signed over in the `MsgPayForData` transactions also in that block. If any commitment is not found in one of the PFD transactions, or if there is a commitment that doesn't have a corresponding message, then they reject that block.
+
+While this functions as a message inclusion check, the light client has to assume that 2/3's of the voting power is honest in order to be assured that both the messages they are interested in and the rest of the messages paid for in that block are actually included. In order to have this property, we need a block (or tx? pls halp reviewers) validity rule where:
+
+> **All share commitments included in `MsgPayForData` must consist only of subtree roots of the data square.**
+
+The main issue with that requirement is that it requires the users to know the relevant subtree roots before they sign, which is problematic considering that is the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.
```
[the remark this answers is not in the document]
evan-forbes: again, it wouldn't let me commit the suggestion due to changes, but fixed here https://github.com/celestiaorg/celestia-app/pull/673/commits/a161fbbfccf371df1f246cbcd0c8b299014849c7

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/673/commits/27ee1dac22be7498ebfb87be85c7cd4718a4d56e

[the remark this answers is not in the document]
evan-forbes: > I'm not sure I would want to commit to always doing this in the future.

and by this I mean that I can see there being a case where nil makes sense, but also just in general always following that specific book is unrealistic I think. _unless_ they have a magic linter that catches everything there lol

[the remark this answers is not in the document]
evan-forbes: > Although, I'm not sure I would want to commit to always doing this in the future.

meaning that I could see a few use cases where accepting nil is unavoidable or against the grain of an existing api (specifically a few in tendermint). for our new apis, this makes sense.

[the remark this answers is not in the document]
evan-forbes: I'm also not against sticking to an agreed upon basic set of go opinions that someone else has written for things that linters can't catch

rootulp (approved):
Incredible ADR, thanks for writing this! Again, the diagrams are insanely helpful.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+### Add metadata to wrapped transactions [#819](https://github.com/celestiaorg/celestia-core/pull/819)
+
+In order to check for message inclusion, create message inclusion fraud proofs, split the block data into squares, and not force non-interactive defaults for every square, we have to connect a `MsgPayForData` transaction to its corresponding message by adding the index of the share that the message starts on as metadata. Since users cannot know this ahead of time, block producers have to add this metadata before the transaction gets included in the block.
+
+We are already wrapping/unwrapping malleated transactions, so including the `share_index` as metadata using the current code is essentially as simple as adding the `share_index` to the struct. Transactions are unwrapped selectively by using a [`MalleatedTxDecoder`](https://github.com/celestiaorg/celestia-app/blob/5ac236fb1dab6628e98a505269f295c18e150b27/app/encoding/malleated_tx_decoder.go#L8-L15) or by using the [`UnwrapMalleatedTx`](https://github.com/celestiaorg/celestia-core/blob/212901fcfc0f5a095683b1836ea9e890cc952dc7/types/tx.go#L214-L237) function.
```
SweeXordious, line 87: [Non Blocking][Proposition]
```suggestion
We are already wrapping/unwrapping malleated transactions, so including the `share_index` as metadata using the current code is essentially as simple as adding the `share_index` to the struct. Transactions are unwrapped selectively by using a [`MalleatedTxDecoder(...)`](https://github.com/celestiaorg/celestia-app/blob/5ac236fb1dab6628e98a505269f295c18e150b27/app/encoding/malleated_tx_decoder.go#L8-L15) or by using the [`UnwrapMalleatedTx(...)`](https://github.com/celestiaorg/celestia-core/blob/212901fcfc0f5a095683b1836ea9e890cc952dc7/types/tx.go#L214-L237) function.
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+Currently, when checking for message inclusion, validators recreate the share commitment from the messages in the block and compare those with what are signed over in the `MsgPayForData` transactions also in that block. If any commitment is not found in one of the PFD transactions, or if there is a commitment that doesn't have a corresponding message, then they reject that block.
+
+While this functions as a message inclusion check, the light client has to assume that 2/3's of the voting power is honest in order to be assured that both the messages they are interested in and the rest of the messages paid for in that block are actually included. In order to have this property, we need a block validity rule where:
+
+> **All share commitments included in `MsgPayForData` must consist only of subtree roots of the data square.**
+
+The main issue with that requirement is that users must know the relevant subtree roots before they sign, which is problematic considering that if the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.
```
SweeXordious, line 17: [Non Blocking][Question]
which subtree roots? The root of the message shares?

SweeXordious (commented):
Awesome stuff :rocket: :+1:. Mainly adding questions that I have.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+Our encoding scheme now has to actually use the metadata added to wrapped transactions.
+
+Note: In order to properly test the new encoding scheme, we have to perform identical if not very similar application logic to that in `PrepareProposal`, for this reason, we initially wanted to move the share encoding/decoding logic to the app instead of core. There were some other quality of life improvements that were also added during this refactor that are technically unrelated to these changes.
+
+We currently utilize a struct to store the state needed for lazily writing message shares. Here we add a method to it that allows for us to write namespaced padded shares.
+
+```go
+// MessageShareSplitter lazily splits messages into shares that will eventually be
+// included in a data square. It also has methods to help progressively count
+// how many shares the messages written take up.
```
SweeXordious, line 108: [Non Blocking][Question]
The count also reflects how many padded shares has been written:  Line 126, no?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+type MessageShareSplitter struct {
+  shares [][]NamespacedShare
+  count  int
+}
+...
+// WriteNamespacedPaddedShares adds empty shares using the namespace of the last written share.
+// This is useful to follow the message layout rules. It assumes that at least
+// one share has already been written, if not it panics.
+func (msw *MessageShareSplitter) WriteNamespacedPaddedShares(count int) {
```
SweeXordious, line 117: [Non Blocking][Cosmetic]
Renaming `count` to `paddingsCound` or similar for readability and not to be confused with `mss.count`.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+  msw.shares = append(msw.shares, namespacedPaddedShares(lastMessage[0].ID, count))
+  msw.count += count
+}
+```
+
+Now we simply combine this new functionality with the `share_index`s described above, and we can properly split and pad messages when needed. Note, the below implementation allows for `nil` to be passed as indexes. This is important, as it allows the same implementation to be used in the cases where we don't want to split messages using wrapped transactions, such as supporting older networks or when users create commitments to sign over for `MsgWirePayForData`
+
+```go
+func SplitMessages(cursor int, indexes []uint32, msgs []coretypes.Message, useShareIndexes bool) ([][]byte, error) {
+  if len(indexes) != len(msgs) {
```
SweeXordious, line 134: [Non Blocking][Question]
According to this:
> Note, the below implementation allows for `nil` to be passed as indexes.

If `indexes` is `nil` or `0`, it will error.  Is this intended?
```suggestion
  if len(indexes) != 0 && len(indexes) != len(msgs) {
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+```go
+// NextAlignedPowerOfTwo calculates the next index in a row that is an aligned
+// power of two and returns false if the entire the msg cannot fit on the given
+// row at the next aligned power of two. An aligned power of two means that the
+// largest power of two that fits entirely in the msg or the square size. pls
+// see specs for further details. Assumes that cursor < k, all args are non
+// negative, and that k is a power of two.
+// https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
+func NextAlignedPowerOfTwo(cursor, msgLen, k int) (int, bool) {
```
SweeXordious, line 167: [Non Blocking][Cosmetics]
Change `k` to a more describing name like: `rowLen`, `rowSize` or similar, if of course it means the `rowLen` and I am not mistaking it for something else.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+       }
+   }
+}
+```
+
+#### Pruning excess transactions
+
+If there are too many transactions and messages in the square to fit in the max square size, then we have to remove them from the block. This can be complicated, as by default we want to prioritzie transactions that have higher fees, but removing a low fee transaction doesn't always result in using less shares.
+
+The simplest approach, and the one taken in the initial implementation, works by prematurely pruning the txs if we estimate that too many shares are being used. While this does work, and fulfills the constraints discussed earlier to create valid blocks, it is suboptimal. Ideally we would be able to identify the most optimal message and transactions to remove and then simply remove only those. As mentioned earlier, technically, a single byte difference could change the entire arrangement of the square. This makes arranging the square with complete confidence difficult not only because we have to follow all of the constraints, but also because of our frequent reliance on variable length length delimiters, and protobuf changing the amount of bytes used depending on the size of ints/uints.
```
SweeXordious, line 366: [Non Blocking][Question]

Doesn't it make sense to create an issue stating that we can improve the txs picking algorithm to something more efficient and with higher optimized value? So that we always remember this?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+}
+```
+
+## Future Improvements
+
+The current implementation performs many different estimation and calculation steps. It might be possible to amortize these calculations to each transaction, which would make it a lot easier to confidently arrange an optimal block.
+
+## Status
+
+{Deprecated|Proposed|Accepted|Declined}
```
SweeXordious, line 675: [Non Blocking][Suggestion]

This is probably accepted as it is already implemented

```suggestion
Accepted
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+## Consequences
+
+### Positive
+
+- Trust minimized message inclusion checks
+- Ability to create message inclusion fraud proofs
```
SweeXordious, line 682:
```suggestion
- Ability to create message inclusion fraud proofs
- Ability to sign the right commitments to messages without interacting with the block producer
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+```
+
+Now we can fulfill the second constraint:
+
+- The commitments signed over in each `MsgPayForData` must consist only of subtree roots of the data square.
+
+```go
+func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponseProcessProposal {
+   ...
+   // iterate over all of the MsgPayForData transactions and ensure that they
```
SweeXordious, line 591: [Non Blocking][Suggestion]

```suggestion
   // iterate over all of the MsgPayForData transactions and ensure that their
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+// EDSSubTreeRootCacher caches the inner nodes for each row so that we can
+// traverse it later to check for message inclusion. NOTE: Currently this has to
+// use a leaky abstraction (see docs on counter field below), and is not
+// threadsafe, but with a future refactor, we could simply read from rsmt2d and
+// not use the tree constructor which would fix both of these issues.
+type EDSSubTreeRootCacher struct {
+   caches     []*subTreeRootCacher
+   squareSize uint64
+   // counter is used to ignore columns NOTE: this is a leaky abstraction that
```
SweeXordious, line 536: [Non Blocking][Suggestion]

```suggestion
   // counter is used to ignore columns. NOTE: this is a leaky abstraction that
```

SweeXordious (dismissed):
:rocket: :+1: :tada: :tada: Awesome work. This should also be called: Demystifying non-interactive defaults.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: > which is problematic considering that if the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.

this is for messages that don't use the non-interactive defaults, there's no way to know which subtree roots to sign over ahead of time. The signers of the PFD would essentially have to know the arrangement of the square ahead of time.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+To fix this, the spec outlines the “non-interactive default rules”. These involve a few additional **default but optional** message layout rules that enables the user to follow the above block validity rule, while also not interacting with a block producer. Commitments to messages can consist entirely of sub tree roots of the data hash, and for those sub tree roots to be generated only from the message itself (so that the user can sign something “non-interactively”). NOTE: THE STATEMENT BELOW IS MODIFIED FROM THE SPEC
+
+> **Messages must begin at a location aligned with the largest power of 2 that is not larger than the message length or k.**
+
+Below illustrates how we can break a message up into two different subtree roots, root `A` for first four shares, and root `B` consisting of the last two shares. We can then create a commitment out of any number of subtree roots by creating a merkle root of those commitments.
```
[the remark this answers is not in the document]
evan-forbes: yes!! super great addition

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: good point, I think I forgot to rename them during the refactor, where they used to be MessageShareWriter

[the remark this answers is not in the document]
evan-forbes: changed here https://github.com/celestiaorg/celestia-app/pull/673/commits/7dc68b252da5ca9807f6c8861baf5a35036f7e44

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: yeah

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: yeah, good idea, we can do that

evan-forbes: yeah, good point #682

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: > Ability to sign the right commitments to messages without interacting with the block producer

I didn't list this originally because this is currently what is happening. The PFD signers are still signing over correct commitments in that the validators can check for message inclusion using them, just not in a trust minimized way

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+
+> **All share commitments included in `MsgPayForData` must consist only of subtree roots of the data square.**
+
+The main issue with that requirement is that users must know the relevant subtree roots before they sign, which is problematic considering that if the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.
+
+To fix this, the spec outlines the “non-interactive default rules”. These involve a few additional **default but optional** message layout rules that enables the user to follow the above block validity rule, while also not interacting with a block producer. Commitments to messages can consist entirely of sub tree roots of the data hash, and for those sub tree roots to be generated only from the message itself (so that the user can sign something “non-interactively”). NOTE: THE STATEMENT BELOW IS MODIFIED FROM THE SPEC
```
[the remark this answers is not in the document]
evan-forbes: yeah, this is why we have `WirePayForData` transactions, which get malleated into `PayForData` transactions.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,691 @@
...
+We are already wrapping/unwrapping malleated transactions, so including the `share_index` as metadata using the current code is essentially as simple as adding the `share_index` to the struct. Transactions are unwrapped selectively by using a [`MalleatedTxDecoder`](https://github.com/celestiaorg/celestia-app/blob/5ac236fb1dab6628e98a505269f295c18e150b27/app/encoding/malleated_tx_decoder.go#L8-L15) or by using the [`UnwrapMalleatedTx`](https://github.com/celestiaorg/celestia-core/blob/212901fcfc0f5a095683b1836ea9e890cc952dc7/types/tx.go#L214-L237) function.
+
+```proto
+message MalleatedTx {
+bytes original_tx_hash = 1;
+bytes tx               = 2;
+uint32 share_index     = 3;
+}
+```
+
```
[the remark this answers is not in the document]
evan-forbes: they can either use the PFD.MessageSize and calculate it themselves, or they can read the first shares and get the length delimiter

while this certinaly is an important point, I actually feel that it distracts the reader with implementation details.

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
SweeXordious: And, we're signing the subtree roots right? these can change if the block size changes, so how can we sign the right thing?

docs/architecture/ADR-003-Non-interactive-defaults.md
SweeXordious: Does it make sense for the comment to reflect that? 
```suggestion
// how many shares and padding the messages written take up.
```
Or, we expect shares to be messages + paddings?

docs/architecture/ADR-003-Non-interactive-defaults.md
SweeXordious: Makes sense, thanks a lot

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: cause we create commitments and sign over all possible square sizes when creating the wire message

https://github.com/celestiaorg/celestia-app/blob/feebea0433a399bc10f704b204bb88a4275e5a0f/x/payment/types/wirepayfordata.go#L39-L48

https://github.com/celestiaorg/celestia-app/blob/feebea0433a399bc10f704b204bb88a4275e5a0f/x/payment/types/wirepayfordata.go#L69-L77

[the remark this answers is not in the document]
SweeXordious: Ooooh right, got  it. Thanks a lot.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: I think it makes more sense without specifying padding.

If we specify padding, it kind of sounds like there are two different numbers, one for shares and one for padding. The implementation treats padding and info shares exactly the same, and I actually think this reflects that in a clearer way.

perhaps in the future we could refactor this to be clearer by only have a single writing method, and then have a function that creates padded shares given a namespace and count. Then the user keeps track of the last namespace.

SweeXordious: Sounds good, thanks a lot

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: good point 98f10e16573589a6d2d42276b24593367991c7c4

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: that makes sense 70f62c6c3ee519e0fa25a12719915f41362b5e2b

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: this comment, along with the other comment https://github.com/celestiaorg/celestia-app/pull/673#discussion_r962160480 confuses me, as they nullify eachother lol

we can sync on this super super quick if this still doesn't make sense. Perhaps I can fix some wording somewhere

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: > Whether the power is strictly less than the length, or, less or equal.

> that is not larger than the message length or k.

I interpret this to mean not strictly less than, but less than or equal to

does that clarify this?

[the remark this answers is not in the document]
SweeXordious: Aaaah, now I get it, when calculating the largest power of 2 that is less or equal than message length:

1. We start the count from the share after the last share of the previous namespace
2. We start the new namespace at the last position of the count. For example, if message length is `7`, the largest power of 2 that is less than message length is 4. Then, we start the new namespace at the fourth share after the last share of the previous namespace. 

Thanks a lot for your help, now I get this more.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+![after](./assets/after.png "After requiring that all commits consist of subtree roots")
+
+Below is an example block that has been filled using the non-interactive default rules to ensure that all commitments consist only of subtree roots.
+
+![example](./assets/example-full-block.png "example")
+
+- `ns = 1`: The first namespace is put in the beginning of the messages space.
+- `ns = 2`: Starting at the largest power of 2 that is not  (strictly) larger than the message length or `k`.
+- `ns = 3`: Message spanning multiple rows. So, it will be at the beginning of a new row.
+- `ns  = 4`: Starting at the largest power of 2 that is not  (strictly) larger than the message length or `k`.
```
SweeXordious, line 47:
```suggestion
- `ns = 2`: Starting at the largest power of 2 that is not larger than the message length or `k`.
- `ns = 3`: Message spanning multiple rows. So, it will be at the beginning of a new row.
- `ns  = 4`: Starting at the largest power of 2 that is not larger than the message length or `k`.
```

SweeXordious (approved):
Fixing a wrong suggestion before I understand correctly how it works.
Thanks a lot @evan-forbes for your help.

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: > Then, we start the new namespace at the fourth share after the last share of the previous namespace.

just tbc, it doesn't really have that much to do with where the last namespace ended. we want to start at the next aligned power of two, because we want a single subroot to contain all of the set of 4 shares. In a square size of 8, there are exactly two spots for this, share 0, and share 4. If the largest power of two a message cannot fit on the next aligned power of two, then it must always start on the next row.

docs/architecture/ADR-003-Non-interactive-defaults.md
SweeXordious: My bad, there are some nuances that I am still not grasping when it comes to generating the state roots. I will resolve this conversations and give them some more time then come back at you with questions if there are any.
Thanks a lot for your time, this ADR was extremely helpful, great work :rocket:

evan-forbes: thanks for all the great feedback! and sure thing. it took me like four times of learning and forgetting it before I got it lol

happy to answer any questions in our 1:1 or something similar.

HoytRen:
Hello everyone, this is a great effort that implement proof of inclusion, however, I believe the original design of NMT in the thesis is much simpler than this because it doesn't require that arrange namespaces as subtree for verification. Generally speaking, if we allow share have 2 namespaces (min and max), we can make it work. More generally, I believe things become much easier if we split share and message to different logic level but not bind them, and we may have fewer bugs, just like we split the monolithic blockchain. Here are earlier discussion between Mustafa and me, please check the attachments (2 images). Tell me if you are interested in and need more information. I suggest implement share at the level in red of the 2rd. image. In fact I wrote a long article about the whole design in Chinese ( maybe, a little shortter than this ADR003 ), but I just lazy to translate it to English, tell me if you really need it ( because I just re-write the original thesis in fact, then it may not be necessary ).
info
NMT

HoytRen:
The message even don't need to be aligned with share. We could simply treat it as N messages when we build dataroot if a message is splited into multiple shares.

evan-forbes:
@HoytRen thanks for commenting

The reason we must use known subtree roots is so that the user can deterministically create a commitment to sign over that can be used to check that a **message is paid for**. We split the messages and the transactions that pay for them up, so its entirely possible for a block producer to include one and not the other. We need to be able to create fraud proofs in order for light clients to not need to trust validators while also not having to download the entire message from the network again.

do you mind providing more info after taking another look at the ADR? I cannot see how what you posted addresses that issue in a different way, but I am certainly interested if it does :slightly_smiling_face:

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
SweeXordious: Other questions, please, when you have time:

1. In your previous response:
> In a square size of 8, there are exactly two spots for this, share 0, and share 4

You mean share 0 and share 3, right? Because the message will be aligned with the power of 2?

2) If `ns = 1` had 7 shares. Where would it start?
- The beginning of a new row
- The 4th position in a new row
- Where it is now?

3) if the `k = 8`, then the largest power of 2 that are possible are: 1, 2, 4, and 8. 
If a message length is 4. Then, the possible start of message are 1, 2 and 4. Then, the block producer checks if that collides with a previous namespace, then it does that on a new row, right?
In the above diagram, `ns = 4`  is still a mystery to me.

evan-forbes:
bump @adlerjohn, not in a super big rush since we've already merged a lot of the implementation and this is just an ADR, but I believe the only thing blocking this PR is your review of the intro, specifically the block validity rule wording, and reducing the message layout rules to a single rule.

HoytRen:
I don't want to stop the merge, but I'd like to have further discussion about this issue. Based on my current understanding of ADR003, I draw following diagram about workflow of celestia. Please have a look and tell me if I misunderstand. If here isn't big mistake in the diagram, then we have same start point, and I still feel the approach of subtree is an overkill and heavily impact advantage of celestia. I will start working on my proposal tomorrow.

Celestia work flow

HoytRen:
After reading #382, I still don't understand the importance of the following statement.
 
”This works for ensuring that each message was included, but this doesn't actually check that the commitment is actually a subtree root to one of the data availability header row or column roots.“

If we include the commitment into somewhere ( block header for example ), then It's even possible ( however have some additional work to do ) for a light node to validate the message, then we don't need the commitment to be a row/column root? As long as the message is in a share ( or some continuous shares ) and the row/column root ( or continuous roots ) includes these shares ( I know this is another addition work ), then it should be ok?

You mentioned the fraud proof at the end of the post. Then, why don't rely on it ( and may remove the necessary that include commitments for light node and download of the message ), but it's just a 'good to have? The fraud proof seems important for celestia' basic assumption, then we could trust this way?

evan-forbes:
@HoytRen I think the confusion here lies with term "Message Inclusion". As you stated, we could provide merkle proofs to prove that a message was included in the square. That's not the issue though. 

The issue is that we have to prove that each payment transaction (PFD) has a message included in the same block.

evan-forbes:
@HoytRen I'm not sure I see how this is related to this ADR. This ADR is strictly related to the mechanism that allows users to sign over a commitment that links to subtree roots of the message without knowing the square layout ahead of time.

I could be wrong, but I'm not sure this specific PR is the best place to answer your questions. Perhaps it would be easier via a different medium?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
+# ADR 003: Non-interactive Defaults, Wrapped Transactions, and Subtree Root Message Inclusion Checks
+
+NOTE: Unlike normal tendermint/cosmos ADRs, this ADR isn't for deciding on whether or not we will implement non-interactive defaults. The goal of this document is to help reviewers and future readers understand what non-interactive defaults are, the considerations that went into the initial implementation, and how it differs from the original specs.
+
+The exact approach taken by the initial implementation, however, is certainly up for scrutiny.
+
+## Intro
+
+pls see the [original specs](https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md), from which this ADR paraphrases heavily.
```
adlerjohn, line 9: Professionalize.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
+# ADR 003: Non-interactive Defaults, Wrapped Transactions, and Subtree Root Message Inclusion Checks
+
+NOTE: Unlike normal tendermint/cosmos ADRs, this ADR isn't for deciding on whether or not we will implement non-interactive defaults. The goal of this document is to help reviewers and future readers understand what non-interactive defaults are, the considerations that went into the initial implementation, and how it differs from the original specs.
```
adlerjohn, line 3: Since this is going on GitHub, you could use the cute

```md
> **Note**
> lorem ipsum
```

ref: https://github.com/community/community/discussions/16925

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+Below illustrates how we can break a message up into two different subtree roots, following a [Merkle Mountain Range](https://docs.grin.mw/wiki/chain-state/merkle-mountain-range/) structure, i.e. root `A` for first four shares, and root `B` consisting of the last two shares. We can then create a commitment out of any number of subtree roots by creating a merkle root of those commitments.
+
+![Subtree root commitment](./assets/subtree-root.png "Subtree Root based commitments")
+
+> **If the messages are larger than k, then they must start on a new row.**
+
+![Multi row share commitment](./assets/multirow.png "Subtree Root based commitments")
+
+If we can follow these rules, we can always create a commitment to the data that is a subtree root of the data root while only knowing the data in that message.
```
adlerjohn, line 33: Hmmm not sure this is worded correctly. It should be something like "a commitment to subtree roots of the data root, that commit to the data"?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+In practice this usually means that we end up adding padding between messages (zig-zag hatched share) to ensure that each message is starting at an "aligned power of two". Padding consists of shares with the namespace of the message before it, with all zeros for data.
```
adlerjohn, line 35:
```suggestion
In practice this usually means that we end up adding padding between messages (zig-zag hatched shares in the figure below) to ensure that each message is starting at an "aligned power of two". Padding consists of shares with the namespace of the message before it, with all zeros for data.
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+However, the default implementation of non-interative defaults is within the scope of this ADR. Whatever design we use ultimately needs to support not using the non-interactive defaults. Meaning we should be able to encode block data into a square even if the messages are not arranged according to the non-interactive defaults. Again, this does not change the requirement that all share commitments signed over in PFDs consist only of subtree roots.
+
+## Detailed Design
+
+To recap the default constraints of arranging a square:
+
+- All messages must be ordered lexicographically by namespace.
+- The commitments signed over in each `MsgPayForData` must consist only of subtree roots of the data square.
+- If a `MsgPayForData` is added to the square, then its corresponding message must also be included.
+- There must not be a message without a `MsgPayForData` (does this need to be a rule? cc @adlerjohn).
```
adlerjohn, line 70: I think having this rule at the very least makes things simpler, even if not strictly required (have to think of if it's required). So, yes.

Ditto for the repeat of this line further down.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+- Transactions with higher fees should be prioritized by default.
+- Ideally, the square should be filled as optimally as possible.
```
adlerjohn, line 72: If you're using "should" instead of "must," then "ideally" is redundant.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+For squares that are smaller than the max squaresize, the exact approach is much less important. This is because if we can't fit all of the transactions in a square, then by default we shouldn't be using that square size in the first place.
```
adlerjohn, line 74:
```suggestion
For squares that are smaller than the max square size, the exact approach is much less important. This is because if we can't fit all of the transactions in a square, then by default we shouldn't be using that square size in the first place.
```

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+Arranging the messages in the block to maximize for fees and optimize square space is an NP-hard problem. While the actual computation of the number of shares could be cached to an extent, each change to the square potentially affects the rest of the messages in the square. The only way to know for certain is to calculate the number of shares used. Calculating the exact number of bytes used is further complicated by the order of the steps due and our rampant use of varints, both in our encoding scheme and protobuf's. The example below shows how removing a single share (from the transactions in this case) could change the rest of the square and allow for a message that otherwise would not fit.
```
adlerjohn, line 76: > is an NP-hard problem

`[citation needed]`

docs/architecture/ADR-003-Non-interactive-defaults.md
liamsi: yes bls

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+> **All share commitments included in `MsgPayForData` must consist only of subtree roots of the data square.**
+
+The main issue with that requirement is that users must know the relevant subtree roots before they sign, which is problematic considering that if the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.
+
+To fix this, the spec outlines the “non-interactive default rules”. These involve a few additional **default but optional** message layout rules that enables the user to follow the above block validity rule, while also not interacting with a block producer. Commitments to messages can consist entirely of sub tree roots of the data hash, and for those sub tree roots to be generated only from the message itself (so that the user can sign something “non-interactively”).
+
+NOTE: THE STATEMENT BELOW IS MODIFIED FROM THE SPEC
```
liamsi, line 21: nit: I do not think we need to capitalize this (given that we deprecated the spec and it has a huge warning in the readme). But don't feel strongly about it either.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+While this functions as a message inclusion check, the light client has to assume that 2/3's of the voting power is honest in order to be assured that both the messages they are interested in and the rest of the messages paid for in that block are actually included. In order to have this property, we need a block validity rule where:
+
+> **All share commitments included in `MsgPayForData` must consist only of subtree roots of the data square.**
+
+The main issue with that requirement is that users must know the relevant subtree roots before they sign, which is problematic considering that if the block is not organized perfectly, the subtree roots will include data unknown to the user at the time of signing.
+
+To fix this, the spec outlines the “non-interactive default rules”. These involve a few additional **default but optional** message layout rules that enables the user to follow the above block validity rule, while also not interacting with a block producer. Commitments to messages can consist entirely of sub tree roots of the data hash, and for those sub tree roots to be generated only from the message itself (so that the user can sign something “non-interactively”).
```
liamsi, line 19: > These involve a few additional **default but optional** message layout rules 

This is something that always confuses me. The way I understood this on calls in the past is that this messages layout will be verified during ProcessProposal. How is it only a default and even more so: how is it **optional**?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+![fit the extra message by removing a tx](./assets/fit-extra-msg-byremoving-tx.png "extra message")
+
+To meet the above constraints, there are multiple refactors required.
+
+- Add metadata to wrapped transactions, connecting that transaction with a message that it pays for.
+- Refactor share splitting and merging to use the above metadata when decoding and encoding the square.
+- Implement the non-interactive default logic.
+- Implement the ability to traverse nmt tree to find subtree roots.
+- Refactor `PrepareProposal` to arrange the shares such that each message has the appropriate subtree roots, and so that the metadata connection between transactions and messages is correct.
+- Refactor `ProcessProposal` to check for message inclusion using only subtree roots to row roots.
```
liamsi, line 88: What happens if the defaults are not followed during ProcessProposal? Does the block get rejected?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+![extra message that can't fit](./assets/extra-message-that-doesnt-fit.png "extra message")
+![fit the extra message by removing a tx](./assets/fit-extra-msg-byremoving-tx.png "extra message")
+
+To meet the above constraints, there are multiple refactors required.
+
+- Add metadata to wrapped transactions, connecting that transaction with a message that it pays for.
+- Refactor share splitting and merging to use the above metadata when decoding and encoding the square.
+- Implement the non-interactive default logic.
+- Implement the ability to traverse nmt tree to find subtree roots.
```
liamsi, line 86: Is there an issue about this?

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+```
+
+### ProcessProposal
+
+Fortunately, most of the work necessary for non-interactive defaults is encapsulated by `PrepareProposal`. Our goal in `ProcessProposal` is to enforce the constraints that we set during `PrepareProposal`. Note that we cannot actually check the last two constraints, so we don't.
+
+- All messages must be ordered lexicographically by namespace.
+- The commitments signed over in each `MsgPayForData` must consist only of subtree roots of the data square.
+- If a `MsgPayForData` is added to the square, then its corresponding message must also be included.
+- There must not be a message without a `MsgPayForData` (does this need to be a rule? cc @adlerjohn).
```
liamsi, line 466: @adlerjohn can you reply here? We should not ask questions to each other in ADRs IMO.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+#### Square Estimation
+
+Using some of the non-interactive defaults code above, we can quickly calculate the number of shares required by each type of block data. As discussed in the docs of this function, the goal here is not necessarily to get a perfect count of the shares that are being used, but we do need to know roughly what square size is needed. When estimating, we should round up in square size.
+
+```go
+// FitsInSquare uses the non interactive default rules to see if messages of
+// some lengths will fit in a square of size origSquareSize starting at share
+// index cursor. See non-interactive default rules
+// https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
```
liamsi, line 320: Do we really want to link to spec here and in the code? If yes, then please use permalinks.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+
+The key to arranging the square into non-interactive defaults is calculating then next "aligned power of 2". We do that here statelessly with two simple functions.
+
+```go
+// NextAlignedPowerOfTwo calculates the next index in a row that is an aligned
+// power of two and returns false if the entire the msg cannot fit on the given
+// row at the next aligned power of two. An aligned power of two means that the
+// largest power of two that fits entirely in the msg or the square size. pls
+// see specs for further details. Assumes that cursor < k, all args are non
+// negative, and that k is a power of two.
```
liamsi, line 172: bls https://github.com/celestiaorg/celestia-app/pull/673/files#r978232971

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+// https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
```
liamsi, line 173: Do we really want to link to spec here and in the code? If yes, then please use permalinks.

docs/architecture/ADR-003-Non-interactive-defaults.md
```diff
@@ -0,0 +1,698 @@
...
+   // txs (and their messages if they're pfd txs).
+   if totalSharesUsed > int(squareSize*squareSize) {
+       parsedTxs = prune(app.txConfig, parsedTxs, totalSharesUsed, int(squareSize))
+   }
+   ...
+```
+
+#### Malleation
+
+Due to the use of the `parsedTxs` data structure, we can now abstract the malleation process entirely, whereas before it was hard coded into the share proprietary proposal share splitting logic.
```
liamsi, line 389: > before it was hard coded into the share proprietary proposal share splitting logic

This sentence is hard to follow.

liamsi (commented):
Not fully ready with the review but submitting my first comments.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: It's optional because when other validators are verifying the square, they don't pick where the message starts. That is encoded in the wrapped transactions as a share index. 

This way, its _possible_ for the block validity rule to still be followed provided that the user signs over a share commitment that matches the subtree roots of a message at an arbitrary starting point in the square.
> All share commitments included in MsgPayForData must consist only of subtree roots of the data square.

thinking about this, we don't actually test for this scenario, so I'll add an issue.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: there's no way to do that atm, and it would be quite complicated. So a simpler answer would probably just be, yes.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: good point, I'll add links

we have some listed here https://github.com/celestiaorg/celestia-app/issues/382#issuecomment-1218372206

docs/architecture/ADR-003-Non-interactive-defaults.md
adlerjohn: Responded in my review above.

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
SweeXordious: I have other questions, please, when you have time :smile_cat:  :

1. In your previous response:
> In a square size of 8, there are exactly two spots for this, share 0, and share 4

You mean share 0 and share 3, right? Because the message will be aligned with the power of 2?

2) If `ns = 1` had 7 shares. Where would it start?
- The beginning of a new row
- The 4th position in a new row
- The same position it is starting now

3) if the `k = 8`, then the largest power of 2 that are possible are: 1, 2, 4, and 8. 
If a message length is 4. Then, the possible start of message are 1, 2 and 4. Then, the block producer checks if that collides with a previous namespace, then it does that on a new row, right?
In the above diagram, `ns = 4`  is still a mystery to me.

And, for the clients, if I understand right, all they need to do is decompose the message to a  merkle mountain range structure, that is bound by the row size (which is always a power of 2), and sign all the subtree roots commitments, for all row sizes.

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: :+1: e3f9831e099b1510a711c7ddfb42dca2f150731e

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: :+1: ccea1f55104789af1bdf5573e503cc060183e069

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: good point e8842c8457397e36bd74e78230c5c79724ed89db

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: :+1: c162833a9b0b5dc4478aad4a7b6970505f1d8c35

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: good point: 69826cce25e69b231a2952692635acb41f537c06

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: :+1: 

removed this specific reference, but used permalinks to the others 185e1574c1f746d8bd2067f165170e1275f0d364

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: 703384a67328574b50c5cd7e4e0dd88367106390

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: removed, e706becb357c3042f8ed3971b17cccf26def1dd7 as this was mainly a comment for reviewers

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: changed here, but created an issue to change the rest of the code https://github.com/celestiaorg/celestia-app/issues/780

docs/architecture/ADR-003-Non-interactive-defaults.md
evan-forbes: :+1: https://github.com/celestiaorg/celestia-app/commit/c162833a9b0b5dc4478aad4a7b6970505f1d8c35

docs/architecture/ADR-003-Non-interactive-defaults.md
adlerjohn: @evan-forbes can you remove uses of `@adlerjohn` in this page?

evan-forbes: yeah sure thing https://github.com/celestiaorg/celestia-app/pull/673/commits/eb745c89db1feca139b77ba017303b34587b4ca2
good catch, forgot about this one

adlerjohn (approved):

docs/architecture/ADR-003-Non-interactive-defaults.md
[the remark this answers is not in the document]
evan-forbes: > You mean share 0 and share 3, right? Because the message will be aligned with the power of 2?

we actually want 4 in this case, as we want to start at the next aligned power of two, and we would want to start on the next row (5th position).

> f ns = 1 had 7 shares. Where would it start?

if there was room on the row for 4 shares, the largest power of two less than or equal to the size of _the message_, then it would start there. If not, then it would start on the next row.

> if the k = 8, then the largest power of 2 that are possible are: 1, 2, 4, and 8.

yeah!

> If a message length is 4. Then, the possible start of message are 1, 2 and 4.

then we have to use the largest, 4. it has to start at a share index that is divisible by 4

evan-forbes:
will likely merge this Monday w/o further feedback @liamsi, just so we can officially close #382 and the implementation has already been merged.

PR #675, opened 2022-08-30
Delete precommit

rootulp:
The `precommit` file isn't used but this `pre-commit` file is used: https://github.com/celestiaorg/celestia-app/blob/main/contrib/githooks/pre-commit


evan-forbes (approved):
:axe:

Issue #677, opened 2022-08-31
Unit test AllSquareSizes


rootulp:
## Context
- https://github.com/celestiaorg/celestia-app/pull/666#discussion_r959676873

## Problem
We want to use https://github.com/rootulp/celestia-app/blob/13076a62c567f624b2508752b7e129467cd71bf9/x/payment/types/square_sizes.go#L26 in more locations (i.e tests) but no unit tests exist for this behavior

## Proposal
Add unit tests for the behavior of `AllSquareSizes`

PR #679, opened 2022-09-02
test: AllSquareSizes

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/677

### DIFF
--- a/x/payment/types/square_sizes_test.go
+++ b/x/payment/types/square_sizes_test.go
@@ -1,0 +1,44 @@
+package types
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestAllSquareSizes(t *testing.T) {
+	type testCase struct {
+		msgSize int
+		want    []uint64
+	}
+
+	tests := []testCase{
+		{0, []uint64{2, 4, 8, 16, 32, 64, 128}}, // should this function return an error for a message size of 0?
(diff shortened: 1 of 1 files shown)


x/payment/types/square_sizes_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func TestAllSquareSizes(t *testing.T) {
+	type testCase struct {
+		msgSize int
+		want    []uint64
+	}
+
+	tests := []testCase{
+		{0, []uint64{2, 4, 8, 16, 32, 64, 128}}, // should this function return an error for a message size of 0?
```
evan-forbes, line 15: this is a really good point! I'm not sure. Technically, iirc, messages of length zero are invalid, but handling that error each time we use this function seems cumbersome. Perhaps we should just handle that in validate basic (if we're not already) and then assume that its fine here.

evan-forbes (approved):

evan-forbes:
really great comments on the test btw

x/payment/types/square_sizes_test.go
rootulp: I pulled the trigger on merge before seeing this but agreed seems cumbersome to handle errors from this function and would prefer a validate basic check on message length.

PR #690, opened 2022-09-07
chore: move malleated transaction decoder to the encoding package

evan-forbes:
I don't think it makes sense that we are keeping the custom decoder in the app package, since it is strictly related to encoding. 

Moved this in the non-interactive defaults feature branch, but it should probably be its own PR.


rootulp (approved):
:shipit:

PR #691, opened 2022-09-08
Add `InfoReservedByte` to compact and sparse shares

rootulp:
Implementation for https://github.com/celestiaorg/celestia-app/issues/659

Note this may be easier for reviewers to review after https://github.com/celestiaorg/celestia-app/issues/718 is completed. In the interim https://github.com/celestiaorg/celestia-app/pull/660 may be helpful.


evan-forbes (commented):
nice work! I need to take a second deeper look, but at first glance this looks like it's as minimal addition as possible, which is impressive given you had to grok all the encoding code lol

pkg/shares/compact_shares_test.go
```diff
@@ -121,3 +121,43 @@ func Test_processCompactShares(t *testing.T) {
...
+	assert.Condition(t, func() bool { return len(shares) > 1 })
+
+	infoByte := shares[1][appconsts.NamespaceSize : appconsts.NamespaceSize+appconsts.ShareInfoBytes][0]
+
+	isMessageStart := false
+	want, err := NewInfoReservedByte(appconsts.ShareVersion, isMessageStart)
+
+	require.NoError(t, err)
+	assert.Equal(t, byte(want), infoByte)
+}
```
evan-forbes, line 163: [optional, prefer different PR due to timing]
I think this should be done in a different PR (if at all) since we want to optimize to merge into #692 asap

but we might refactor these tests into a table drive test that tests evidence and the rest of the shares that are exported. another super minor [nit], would be to use a formula to calculate the size of the random compact shares so that we can incorporate appconsts. (we don't do this other places, but should so that can probably be a separate PR too)

evan-forbes (commented):
[blocking question]
I didn't see anywhere that we were throwing an error if the first share was not the start of the message. If that's missing, then I think we should add that.

[comment on future work]
This does an **excellent** job of adding the info byte, but we have yet to have a universal share API. celestia-node should be able to call a function on any set of shares and get a message out.

As discussed sync and in other issues (I think?), this might entail a larger refactor of share encoding and decoding in general. Things such as changing the name of Messages to data, maybe refactoring the encoding logic to better fit the new API, and then refactoring the Merge/Split functions to use that new API. I think we still have to write a few issues for this, and should definitely be done in different PRs. I don't think we can close existing Universal share encoding issues until that is completed, and we should update the ADR to reflect the changes to the API there.

evan-forbes:
this does break #692, and I consider it to be blocking, since that PR should handle adding the info byte. We might also want to consider changing the estimation API and fixing tx inclusion before we merge that too

rootulp:
> I didn't see anywhere that we were throwing an error if the first share was not the start of the message. If that's missing, then I think we should add that.

Is this question specifically for `parse_sparse_shares.go`? If so I'll amend this PR to handle, good call!

There is some ambiguity on when the start indicator should be `1` for compact shares, see https://github.com/celestiaorg/celestia-app/pull/660#discussion_r973256796

> Perhaps this was just confusion from the name of this PR?

Agreed. Renamed PR to clarify.

> celestia-node should be able to call a function on any set of shares and get a message out.

Clarifying question: celestia-node should be able to call a function on any set of shares and get data out (i.e. transactions if the shares provided are transaction shares), right? 

I think celestia-node is relying on celestia-core's share parsing logic (see [here](https://github.com/celestiaorg/celestia-node/blob/d051a94bce1cf8a6b72be03a9484abdb9a6983a3/service/share/share_test.go#L264)). So I think upgrading them to a release of celestia-app with the existing (refactored) share logic is a prerequisite to converting them to using a new API exposed by celestia-app. Created  https://github.com/celestiaorg/celestia-app/issues/723

> this does break https://github.com/celestiaorg/celestia-app/pull/692, and I consider it to be blocking, since that PR should handle adding the info byte

Sorry I don't quite understand. For clarity: this PR is blocking #692 so you want to expedite merging this PR - correct?

pkg/shares/compact_shares_test.go
rootulp: can you please elaborate on 

> would be to use a formula to calculate the size of the random compact shares so that we can incorporate appconsts

evan-forbes:
> Is this question specifically for parse_sparse_shares.go? If so, good call! Done in https://github.com/celestiaorg/celestia-app/commit/5c4fef735d43378540b704350e3b88a585a7a839

also for compact shares, as I think that makes sense, wdyt?

> Clarifying question: celestia-node should be able to call a function on any set of shares and get data out (i.e. transactions if the shares provided are transaction shares), right?

yeah

> So I think upgrading them to a release of celestia-app with the existing (refactored) share logic is a prerequisite to converting them to using a new API exposed by celestia-app.

not sure I follow entirely, as what we do in #723 will change all of this? The existing logic being this PR and the naming changes? Is this just to split up the changes necessary?

pkg/shares/compact_shares_test.go
evan-forbes: yeah sure,

instead of just using the number 100 or 1000, we could say `appconsts.TxShareSize / 2` or `appconsts.TxShareSize * 4`

rootulp:
> also for compact shares, as I think that makes sense, wdyt?

Makes sense, thanks for clarifying that only the first share of a reserved namespace should have the `start indicator = 1`. Will address in an upcoming commit.

> Is this just to split up the changes necessary?

Yep exactly. I'm proposing we split this process into multiple steps to reduce the size of PRs which may make it easier to identify regressions:

1. Cut a release of celestia-app that includes the refactored share logic
2. Upgrade celestia-node to ^ this version. Replace `core` on [this line](https://github.com/celestiaorg/celestia-node/blob/d051a94bce1cf8a6b72be03a9484abdb9a6983a3/service/share/share_test.go#L264) with an import from celestia-app (defined [here](https://github.com/celestiaorg/celestia-app/blob/566b3d41d2bf097ac49f1a925cb56a3abeabadc8/pkg/shares/share_merging.go#L125))
3. Add `ParseShare` API in celestia-app
4. Cut a new release of celestia-app
5. Upgrade celestia-node to ^ this version

> but also want to limit the number of changes they need to make

Agreed. Maybe we make the PRs to celestia-node because we're issuing the breaking changes and should know how to use the old / new APIs.

evan-forbes:
agree that we need to put more thought into which releases we cut, as we also need to account for non-interactive defaults, qgb state machine (this will change how to initalize a network) and the other breaking changes we have. We can isolate the encoding changes to their own release, tho yeah. lets discuss this sync outside of this PR

evan-forbes (approved):
we still have a lot things going on with universal shares, but this is ready and isolated. Great work! happy someone else had to grok the encoding logic lololol

:shipit:

PR #692, opened 2022-09-08
Refactor `PrepareProposal` to produce blocks using the non-interactive defaults

evan-forbes:
This PR refactors `PrepareProposal` to produce blocks according to the non-interactive defaults. It could probably be split up into a few PRs, and we might have to do that to make it easier to review. 

a big part of #382


rootulp (commented):
One blocking comment. Sorry my review isn't particularly helpful b/c I don't understand most of the changes in this PR.

app/estimate_square_size.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	evdCount, available := evdSplitter.Count()
+	if consts.TxShareSize-available > 0 {
+		evdCount++
+	}
+	return txCount + evdCount
+}
+
+// estimateSquareSize uses the provided block data to estimate the square size
+// assuming that all malleated txs follow the non interactive default rules.
```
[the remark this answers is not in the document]
evan-forbes: yeah, good point. I tried to also use return values, but it causes us to declare an unused uint64 or have a bunch of type conversions

app/estimate_square_size.go
```diff
@@ -0,0 +1,242 @@
...
+	for _, tx := range txs {
+		rawTx := tx.rawTx
+		if tx.malleatedTx != nil {
+			rawTx, err = coretypes.WrapMalleatedTx(tx.originalHash(), uint32(msgSharesCursor), tx.malleatedTx)
+			// we should never get to this point, but just in case we do, we
+			// catch the error here on purpose as we want to ignore txs that are
+			// invalid (cannot be wrapped)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
evan-forbes: that's a really good point :thinking: 

the original reason was if someone somehow found how, putting that tx in the mempool would cause all block producers to panic when attempting to create a block with it. In theory this should get caught since we have already decoded the transaction and are just wrapping it in a type ourselves, but since protobuf optimizes for speed over safety (wasn't designed to unmarshal things from byzantine actors) I'm paranoid. Ignoring it would be the safer thing, but I can see how that would be difficult to track down.

perhaps we should log it?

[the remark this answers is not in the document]
evan-forbes: Actually, just decided to panic here. I think you're correct that this is best thing to do in this case. Not only just because it will be clear as day, but I think weirder bugs would occur if this did happen, like blocks getting rejected becuase a non-wrapped PFD could make it in the block.

app/estimate_square_size.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	return txShares + 2, evdShares, msgShares
+}
+
+// overEstimateMalleatedTxSize estimates the size of a malleated tx. The formula it uses will always over estimate.
+func overEstimateMalleatedTxSize(txLen, msgLen, sharesCommitments int) int {
+	// the malleated tx uses meta data from the original tx, but removes the
+	// message and extra share commitments. Only a single share commitment will
+	// make it on chain, and the square size (uint64) is removed.
+	malleatedTxLen := txLen - msgLen - ((sharesCommitments - 1) * 128) - 8
```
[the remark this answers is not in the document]
evan-forbes: yeah, we can do a better job of that. tried to with the comment above it, but refactored it to hopefully be clearer. 324966d5fefa043563685f71d0be05a38cd4db3e

app/estimate_square_size.go
```diff
@@ -0,0 +1,242 @@
...
+	sort.Slice(msgSummaries, func(i, j int) bool {
+		return bytes.Compare(msgSummaries[i].namespace, msgSummaries[j].namespace) < 0
+	})
+
+	// isolate the sizes as we no longer need the namespaces
+	msgShares := make([]int, len(msgSummaries))
+	for i, summary := range msgSummaries {
+		msgShares[i] = summary.size
+	}
+	return txShares + 2, evdShares, msgShares
```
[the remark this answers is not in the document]
evan-forbes: whoooooopppsss :grimacing: I literally don't remember, was probably some super late night debugging, but great catch!! removed here 9c0aa5887c936752e8a35f84cf9b2d950047ca6d

app/app.go
```diff
@@ -240,7 +240,7 @@ func New(
 	cdc := encodingConfig.Amino
 	interfaceRegistry := encodingConfig.InterfaceRegistry
 
-	bApp := baseapp.NewBaseApp(Name, logger, db, MalleatedTxDecoder(encodingConfig.TxConfig.TxDecoder()), baseAppOptions...)
+	bApp := baseapp.NewBaseApp(Name, logger, db, encoding.MalleatedTxDecoder(encodingConfig.TxConfig.TxDecoder()), baseAppOptions...)
```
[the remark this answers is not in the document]
evan-forbes: unfortunately, I'm not sure we can without force pushing to drop the commit here since the hashes are different

app/estimate_square_size.go
```diff
@@ -0,0 +1,242 @@
...
+		case !fits:
+			// increment the square size
+			squareSize = int(nextPowerOfTwo(squareSize + 1))
+		}
+	}
+}
+
+// rawShareCount calculates the number of shares taken by all of the included
+// txs, evidence, and each msg.
+func rawShareCount(txs []*parsedTx, evd core.EvidenceList) (txShares, evdShares int, msgLens []int) {
```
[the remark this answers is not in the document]
evan-forbes: yeah, added another comment here to try and improve 50baf90206e0f482c4fbeaa0a7ef93ea501a3799

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	msgShares := 0
+	for _, msgLen := range msgLens {
+		msgShares += msgLen
+	}
+
+	// calculate the smallest possible square size that could contain all the
+	// messages
+	squareSize := nextPowerOfTwo(int(math.Ceil(math.Sqrt(float64(txShares + evdShares + msgShares)))))
+
+	// the starting square size should be the minimum
```
rootulp, line 124:
```suggestion
	// the starting square size should at least be the minimum
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	if squareSize < consts.MinSquareSize {
+		squareSize = consts.MinSquareSize
+	}
+
+	var fits bool
+	for {
+		// assume that all the msgs in the square use the non-interactive
+		// default rules and see if we can fit them in the smallest starting
+		// square size. We start the cusor (share index) at the begginning of
```
rootulp, line 133:
```suggestion
		// square size. We start the cursor (share index) at the beginning of
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+		switch {
+		// stop estimating if we know we can reach the max square size
+		case squareSize >= consts.MaxSquareSize:
+			return consts.MaxSquareSize, txShares + evdShares + msgShares
+		// return if we've found a square size that fits all of the txs
+		case fits:
+			return uint64(squareSize), txShares + evdShares + msgShares
+		// try the next largest square size if we can't fit all the txs
+		case !fits:
+			// increment the square size
```
rootulp, line 146:
```suggestion
			// double the square size
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+}
+
+// rawShareCount calculates the number of shares taken by all of the included
+// txs, evidence, and each msg. msgLens is a slice of the number of shares used
+// by each message without accounting for the non-interactive message layout
+// rules.
+func rawShareCount(txs []*parsedTx, evd core.EvidenceList) (txShares, evdShares int, msgLens []int) {
+	// msgSummary is used to keep track of the size and the namespace so that we
+	// can sort the namespaces before returning.
```
rootulp, line 158:
```suggestion
	// msgSummary is used to keep track of the size and the namespace so that we
	// can sort the messages by namespace before returning.
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	txBytes, evdBytes := 0, 0
+	for _, pTx := range txs {
+		// if there is no wire message in this tx, then we can simply add the
+		// bytes and move on.
+		if pTx.msg == nil {
+			txBytes += len(pTx.rawTx)
+			continue
+		}
+
+		// if the there is a malleated tx, then we want to also account for the
```
rootulp, line 178:
```suggestion
		// if there is a malleated tx, then we want to also account for the
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+		// txs that gets included onchain. The formula used here over
```
rootulp, line 179:
```suggestion
		// txs that get included on-chain. The formula used here over
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+
+// rawShareCount calculates the number of shares taken by all of the included
+// txs, evidence, and each msg. msgLens is a slice of the number of shares used
+// by each message without accounting for the non-interactive message layout
+// rules.
+func rawShareCount(txs []*parsedTx, evd core.EvidenceList) (txShares, evdShares int, msgLens []int) {
+	// msgSummary is used to keep track of the size and the namespace so that we
+	// can sort the namespaces before returning.
+	type msgSummary struct {
+		size      int
```
rootulp, line 160:
```suggestion
                // size is the number of shares used by this message
		size      int
```

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	// one byte over the number of shares in the square size. This will also cause
+	// blocks that are one square size too big instead of being perfectly snug.
+	// The estimation must be perfect or greater than what the square actually
+	// ends up being.
+	if txShares > 0 {
+		txShares++
+	}
+
+	for _, e := range evd.Evidence {
+		evdBytes += e.Size() + shares.DelimLen(uint64(e.Size()))
```
rootulp, line 204: [no change needed b/c I don't have a concrete proposal] I'm concerned this code may not be updated when compact share encoding code is modified.

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	evdShares = evdBytes / consts.TxShareSize
+	if evdBytes > 0 {
+		evdShares++ // add one to round up
+	}
+
+	// sort the msgSummaries in order to order properly. This is okay to do here
```
rootulp, line 212:
```suggestion
	// sort the msgSummaries by namespace to order them properly. This is okay to do here
```

rootulp (commented):
Re-reviewed a subset of the files after in-person meeting and wanted to share feedback but still have more files to review.

> It could probably be split up into a few PRs, and we might have to do that to make it easier to review.

This would definitely help me! Additional benefits of smaller PRs:
1. It helps unfamiliar reviewers see which changes were relevant to closing which issues
1. It may be easier to track down when a bug was introduced
1. It may be easier to revert if a regression is discovered

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+	removedContiguousShares := 0
+	contigBytesCursor := 0
+	removedMessageShares := 0
+	removedTxs := 0
+
+	// adjustContigCursor checks if enough contiguous bytes have been removed
+	// inorder to tally total contiguous shares removed
+	adjustContigCursor := func(l int) {
+		contigBytesCursor += l + shares.DelimLen(uint64(l))
+		if contigBytesCursor >= consts.TxShareSize {
```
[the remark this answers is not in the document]
evan-forbes: hmm good point, I'm unsure. We will need to make a canonical API for this and all other estimation code that we keep in the shares package in order to avoid this.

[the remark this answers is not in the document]
evan-forbes: Left an issue #700 for this, not sure how long it will take, but I think its a great idea.

app/estimate_square_size.go
evan-forbes: referenced in #700

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+		rawTx := tx.rawTx
+		if tx.malleatedTx != nil {
+			rawTx, err = coretypes.WrapMalleatedTx(tx.originalHash(), uint32(msgSharesCursor), tx.malleatedTx)
+			// we should never get to this point, but just in case we do, we
+			// catch the error here on purpose as we want to ignore txs that are
+			// invalid (cannot be wrapped)
+			if err != nil {
+				panic(err)
+			}
+			used, _ := shares.MsgSharesUsedNIDefaults(msgSharesCursor, squareSize, tx.msg.Size())
```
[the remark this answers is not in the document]
evan-forbes: github is still showing the old code for some reason, but it should be fixed in 750e31ce146c536454b4d7631ed53e3c62f63086

evan-forbes:
> It could probably be split up into a few PRs, and we might have to do that to make it easier to review.

we can try and do this. It might mean duplicate or unused code in some instances, or that we have to spend more time debugging getting that chunk of code working with the old implementation, but those things might be unavoidable in order to get a proper review. 

We also have roughly 2.5 weeks until this needs to be finished and merged now that the audit has been postponed a bit, so we should have the necessary time.

app/estimate_square_size.go
```diff
@@ -0,0 +1,246 @@
...
+func overEstimateMalleatedTxSize(txLen, msgLen, sharesCommitments int) int {
+	// the malleated tx uses the original txLen to account for meta data from
+	// the original tx, but removes the message and extra share commitments that
+	// are in the wire message by subtracting msgLen and all extra share
+	// commitments.
+	malleatedTxLen := txLen - msgLen - ((sharesCommitments - 1) * 128) - 8
+	// we need to ensure that the returned number is at least larger than or
+	// equal to the actual number, which is difficult to calculate without
+	// actually malleating the tx
+	return appconsts.MalleatedTxBytes + 100 + malleatedTxLen
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/692/commits/7279a0d6e1a66f91cbca12c52e125fd49d496c09

also see #701

app/estimate_square_size.go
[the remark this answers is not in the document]
evan-forbes: :+1: 

https://github.com/celestiaorg/celestia-app/pull/692/commits/7279a0d6e1a66f91cbca12c52e125fd49d496c09

and see #701

rootulp:
Blocked on https://github.com/celestiaorg/celestia-app/pull/691

evan-forbes:
I haven't made many changes other than to make compatible with main. Reopening after attempting to break into a few PRs. While possible to break into smaller chunks, its impossible to add tests to those chunks until all portions are functional.

app/estimate_square_size.go
```diff
@@ -0,0 +1,244 @@
...
+// TODO: include the padding used by each msg when counting removed shares
+func prune(txConf client.TxConfig, txs []*parsedTx, currentShareCount, squareSize int) parsedTxs {
+	maxShares := squareSize * squareSize
+	if maxShares >= currentShareCount {
+		return txs
+	}
+	goal := currentShareCount - maxShares
+
+	removedContiguousShares := 0
+	contigBytesCursor := 0
```
rootulp, line 29: [follow-up PR] we may replace the `contiguous` terminology with `compact`

go.mod
```diff
@@ -160,5 +160,5 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.2.0-sdk-v0.46.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.5.0-tm-v0.34.20
+	github.com/tendermint/tendermint v0.34.20 => github.com/celestiaorg/celestia-core v1.5.0-tm-v0.34.20
```
rootulp, line 163: [question][non-blocking] out of curiosity, why is the exact tendermint version specified?

rootulp (approved):

go.mod
evan-forbes: whoops, good catch, this actually wasn't on purpose. While it technically shouldn't matter we might want to switch it back. I think the only time it did matter was when used to change the imports from `tendermint/tendermint` to `celestiaorg/celestia-core`

evan-forbes: merging w/o as this is optional, and I don't want to look at this PR anymore lol

rootulp: Agreed. :shipit: and we can always remove later

Issue #693, opened 2022-09-08
Get rid of unused tests


evan-forbes:
There are simulation tests that aren't testing anything, and are remnants of starport. While we certainly want equivalent tests, we will not be able to rely on the current simulation package or the testapp given that most of the application is stateless and those tools were not designed for that.

https://github.com/celestiaorg/celestia-app/blob/main/contrib/devtools/sims.mk
https://github.com/celestiaorg/celestia-app/blob/main/.github/workflows/release-sims.yml
https://github.com/celestiaorg/celestia-app/blob/main/.github/workflows/sims.yml

I'm also curious as to why these tests didn't start failing until recently

evan-forbes:
>sims failed on my branch for a known reason: https://github.com/celestiaorg/celestia-app/actions/runs/3012237262

what was this reason?

> sims passes on recent runs main: https://github.com/celestiaorg/celestia-app/actions/workflows/sims.yml?query=branch%3Amain

yeah! the thing that urks me is that I don't know why its passing in the first place lolol

> sims failed on https://github.com/celestiaorg/celestia-app/runs/8242496184?check_suite_focus=true
the -Enabled flag looks suspicious [ref](https://github.com/celestiaorg/celestia-app/blob/b75f6beaed31fff81e1a646abb4094037021f981/contrib/devtools/sims.mk#L10)
likely not worth investigating / fixing if we don't need sims.mk

yeah, I tried to remove flags, and it just kept giving the same error but the next flag/command

rootulp:
> what was this reason?

`go build` fails because [this draft PR](https://github.com/celestiaorg/celestia-app/pull/691) depends on celestia-core changes. The draft PR doesn't modify `go.mod` because the celestia-core changes in [this PR](https://github.com/celestiaorg/celestia-core/pull/849) haven't been merged or released.

rootulp:
oof I just noticed this is closed by https://github.com/celestiaorg/celestia-app/pull/692

rootulp:
Research into sims

- osmosis has https://github.com/osmosis-labs/osmosis/blob/main/.github/workflows/sim.yml
- ignite CLI has docs on running simulations for scaffoled modules https://docs.ignite.com/kb/simapp#scaffold-a-simulation

PR #697, opened 2022-09-12
chore: remove sims

rootulp:
Resolves https://github.com/celestiaorg/celestia-app/issues/693


rootulp:
Will close in favor of https://github.com/celestiaorg/celestia-app/pull/692

evan-forbes (commented):
do we also need to remove something from the root makefile?

I think we can reopen this since it can be isolated.

rootulp:
> do we also need to remove something from the root makefile?



Yes. Likely https://github.com/celestiaorg/celestia-app/blob/main/Makefile#L76-L81

Makefile
```diff
@@ -73,14 +73,7 @@ lint:
...
 .PHONY: lint
 
-###############################################################################
-###                           Tests & Simulation                            ###
-###############################################################################
-# The below include contains the tools target.
-include contrib/devtools/Makefile
-include contrib/devtools/sims.mk
-
-test-all: check test-race test-cover
```
rootulp, line 83: `check` was removed b/c it is not defined in any of the Makefiles

evan-forbes (approved):
:+1:

Issue #687, opened 2022-09-06
Use a single name for each variable that references square size


evan-forbes:
This repo uses three names for the same thing: `origSquareSize`, `squareSize`, `k`. Is it safe to assume they are all interchangeable? If so, would it aid readability to standardize on one? My impression is that readers who are familiar with https://celestiaorg.github.io/celestia-specs/latest/rationale/message_block_layout.html#non-interactive-default-rules are able to infer `k` == `squareSize` so this may not be confusing for other readers but may be helpful for new readers who aren't as familiar with the specs.

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/680#discussion_r964048463_

We should use a single reference to square size for the entire repo.

rootulp:
If we opt to use `SquareSize` and `squareSize`, this implies a breaking change to protobuf types: https://github.com/rootulp/celestia-core/blob/114375ab4f4e2976ff3f6dde1caeddb430a14f4e/types/block.go#L1034

rootulp:
I found another instance in https://github.com/celestiaorg/celestia-app/blob/27632ddc804fbf7655ef5ba941eb4f346c539be9/proto/payment/tx.proto#L34

PR #705, opened 2022-09-13
Use `squareSize` instead of `k` or `origSquareSize`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/687

Note this doesn't address https://github.com/celestiaorg/celestia-core/issues/853


evan-forbes (approved):
:+1:

PR #708, opened 2022-09-14
feat: Implement and install new ABCI Query for transaction inclusion proofs 

### LINKED ISSUE #698: Create a custom query route for tendermint to use to create transaction inclusion proofs
Currently, when creating a tx inclusion proof in celestia-core, we have to encode a portion of the block data into rows. This is necessary to create the nmt tree(s) used to create the proof(s). It's problematic not only because we don't want to have any application logic in core, but in order to fully test that code we have to have a lot of other application specific logic (malleated txs, non-interactive defaults/other square layout rules).

To fix this, we'd like to move all share encoding logic to the app, but this means we can't do tx inclusion proofs in core like normal tendermint. We can get around this by creating a [custom query route](https://github.com/celestiaorg/cosmos-sdk/blob/368049e57edb435ffc58a7e7afedab7e54d83776/baseapp/abci.go#L443-L445), and calling that route each time we want to create a tx inclusion proof in tendermint.

This will involve:
- [x] #699
- [ ] #708
- [x] https://github.com/celestiaorg/celestia-core/issues/851

evan-forbes:
This PR introduces a new custom ABCI query that generates the ABCI inclusion proof for a transaction of a specific index. 

Note that this PR does not fix the tx inclusion proof logic to follow the non-interactive defaults, as we haven't merged those yet. It does however alter the integration test to require that transactions include proofs, so we are testing the new ABCI query method.

blocked by https://github.com/celestiaorg/celestia-core/pull/859
closes #698

### DIFF
--- a/app/split_shares.go
+++ b/app/split_shares.go
@@ -113,8 +113,8 @@ func SplitShares(txConf client.TxConfig, squareSize uint64, data *core.Data) ([]
 // that message and their corresponding txs get written to the square
 // atomically.
 type shareSplitter struct {
-	txWriter  *coretypes.ContiguousShareWriter
-	msgWriter *coretypes.MessageShareWriter
+	txWriter  *shares.ContiguousShareSplitter
+	msgWriter *shares.MessageShareSplitter
 
 	// Since evidence will always be included in a block, we do not need to
 	// generate these share lazily. Therefore instead of a ContiguousShareWriter
@@ -143,8 +143,8 @@ func newShareSplitter(txConf client.TxConfig, squareSize uint64, data *core.Data
 		panic(err)
 	}
 
-	sqwr.txWriter = coretypes.NewContiguousShareWriter(consts.TxNamespaceID)
-	sqwr.msgWriter = coretypes.NewMessageShareWriter()
+	sqwr.txWriter = shares.NewContiguousShareSplitter(consts.TxNamespaceID)
+	sqwr.msgWriter = shares.NewMessageShareSplitter()
 
 	return &sqwr
 }
@@ -152,7 +152,7 @@ func newShareSplitter(txConf client.TxConfig, squareSize uint64, data *core.Data
 // writeTx marshals the tx and lazily writes it to the square. Returns true if
 // the write was successful, false if there was not enough room in the square.
 func (sqwr *shareSplitter) writeTx(tx []byte) (ok bool, err error) {
-	delimTx, err := coretypes.Tx(tx).MarshalDelimited()
+	delimTx, err := shares.MarshalDelimitedTx(tx)
 	if err != nil {
 		return false, err
 	}
@@ -161,7 +161,7 @@ func (sqwr *shareSplitter) writeTx(tx []byte) (ok bool, err error) {
 		return false, nil
 	}
 
-	sqwr.txWriter.Write(delimTx)
+	sqwr.txWriter.WriteBytes(delimTx)
 	return true, nil
 }
 
@@ -205,12 +205,12 @@ func (sqwr *shareSplitter) writeMalleatedTx(
 		return false, nil, nil, nil
 	}
 
-	delimTx, err := wrappedTx.MarshalDelimited()
+	delimTx, err := shares.MarshalDelimitedTx(wrappedTx)
 	if err != nil {
 		return false, nil, nil, err
 	}
 
-	sqwr.txWriter.Write(delimTx)
+	sqwr.txWriter.WriteBytes(delimTx)
 	sqwr.msgWriter.Write(coretypes.Message{
 		NamespaceID: coreMsg.NamespaceId,
 		Data:        coreMsg.Data,
@@ -256,33 +256,36 @@ func (sqwr *shareSplitter) export() [][]byte {
 	if availableBytes < consts.TxShareSize {
 		count++
 	}
-	shares := make([][]byte, sqwr.maxShareCount)
+	rawShares := make([][]byte, sqwr.maxShareCount)
 
 	txShares := sqwr.txWriter.Export().RawShares()
 	txShareCount := len(txShares)
-	copy(shares, txShares)
+	copy(rawShares, txShares)
 
 	evdShareCount := len(sqwr.evdShares)
 	for i, evdShare := range sqwr.evdShares {
-		shares[i+txShareCount] = evdShare
+		rawShares[i+txShareCount] = evdShare
 	}
 
 	msgShares := sqwr.msgWriter.Export()
+	sort.SliceStable(msgShares, func(i, j int) bool {
+		return msgShares[i].ID.Less(msgShares[j].ID)
+	})
 	msgShareCount := len(msgShares)
 	for i, msgShare := range msgShares {
-		shares[i+txShareCount+evdShareCount] = msgShare.Share
+		rawShares[i+txShareCount+evdShareCount] = msgShare.Share
 	}
 
-	tailShares := coretypes.TailPaddingShares(sqwr.maxShareCount - count).RawShares()
+	tailShares := shares.TailPaddingShares(sqwr.maxShareCount - count).RawShares()
 
 	for i, tShare := range tailShares {
 		d := i + txShareCount + evdShareCount + msgShareCount
-		shares[d] = tShare
+		rawShares[d] = tShare
 	}
 
-	if len(shares[0]) == 0 {
-		shares = coretypes.TailPaddingShares(consts.MinSharecount).RawShares()
+	if len(rawShares[0]) == 0 {
+		rawShares = shares.TailPaddingShares(consts.MinSharecount).RawShares()
 	}
 
-	return shares
+	return rawShares
 }
--- a/app/split_shares.go
+++ b/app/split_shares.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/x/auth/signing"
-	"github.com/tendermint/tendermint/pkg/consts"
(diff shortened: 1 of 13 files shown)


evan-forbes:
NOTE: we probably should not merge this until we cut a new release of tendermint with the new query usuage and replace that tag in the go.mod

app/test/block_size_test.go
```diff
@@ -129,7 +129,7 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
 			for _, hash := range hashes {
 				// TODO: once we are able to query txs that span more than two
 				// shares, we should switch to proving txs existence in the block
```
rootulp, line 131: [question] is this comment still necessary after we pass `prove=true` below?

```suggestion
```

app/split_shares.go
```diff
@@ -113,8 +113,8 @@ func SplitShares(txConf client.TxConfig, squareSize uint64, data *core.Data) ([]
 // that message and their corresponding txs get written to the square
 // atomically.
 type shareSplitter struct {
-	txWriter  *coretypes.ContiguousShareWriter
-	msgWriter *coretypes.MessageShareWriter
+	txWriter  *shares.ContiguousShareSplitter
+	msgWriter *shares.MessageShareSplitter
```
evan-forbes, line 117: quite a bit of code had to be changed in the shareSplitter in order to update it to work with the local copies of the share encoding code, as that was never done before as I assumed we'd merge non-interactive defaults code first :sweat_smile: 

would prefer not create a separate PR for this, but I understand if its something we have to do

app/test/block_size_test.go
```diff
@@ -309,5 +309,5 @@ func queryWithOutProof(clientCtx client.Context, hashHexStr string) (*rpctypes.R
 		return nil, err
 	}
 
-	return node.Tx(context.Background(), hash, false)
+	return node.Tx(context.Background(), hash, prove)
```
evan-forbes, line 312: this change requires that we include a tx inclusion proof when querying for a tx

pkg/prove/querier.go
```diff
@@ -0,0 +1,56 @@
...
+
+const TxInclusionQueryPath = "txInclusionProof"
+
+// Querier defines the logic performed when the ABCI client using the Query
+// method with the custom prove.QueryPath. The index of the transaction being
+// proved must be appended to the path. The marshalled bytes of the transction
+// proof (tmproto.TxProof) are returned.
+//
+// example path for proving the third transaction in that block:
+// custom/tx-inclusion-proof/3
```
rootulp, line 22: [question] which fragment of the path is correct `tx-inclusion-proof` or `txInclusionProof`?

app/test/block_size_test.go
```diff
@@ -129,12 +129,13 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
...
 				// TODO: once we are able to query txs that span more than two
 				// shares, we should switch to proving txs existence in the block
-				resp, err := queryWithOutProof(val.ClientCtx, hash)
+				resp, err := queryTx(val.ClientCtx, hash, true)
 				assert.NoError(err)
 				assert.Equal(abci.CodeTypeOK, resp.TxResult.Code)
 				if resp.TxResult.Code == abci.CodeTypeOK {
 					heights[resp.Height]++
 				}
+				require.True(resp.Proof.VerifyProof())
```
evan-forbes, line 138: Now we're requiring that all transactions queried for during the integration test have valid inclusion proofs.

rootulp (approved):

pkg/prove/querier.go
evan-forbes: good catch! 28da36536c35c8e488f108ee77b72dcb9b5d1b1c

app/test/block_size_test.go
evan-forbes: good catch d1f615293bfad076cda6913880450b160842c383

rootulp (approved):
:shipit:

evan-forbes:
still blocked by celestia-core, and need to update to the new release before we can merge, so converting to a draft until then

pkg/prove/querier.go
```diff
@@ -0,0 +1,56 @@
...
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/pkg/consts"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	"github.com/tendermint/tendermint/types"
+)
+
+const TxInclusionQueryPath = "txInclusionProof"
```
rootulp, line 14: [question] does this not need the `custom/` prefix that is used in https://github.com/celestiaorg/celestia-core/pull/859/files#diff-6572dddc535b4d2c98746d71dde97840af1328797e193db28376939a320be8f8R48 ?

rootulp (approved):

pkg/prove/querier.go
evan-forbes: per comment in core, this one needs to not include the path as it will be removed/added by the cosmos-sdk

rootulp (approved):

pkg/prove/querier.go
```diff
@@ -0,0 +1,56 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	"github.com/tendermint/tendermint/types"
+)
+
+const TxInclusionQueryPath = "txInclusionProof"
+
+// Querier defines the logic performed when the ABCI client using the Query
+// method with the custom prove.QueryPath. The index of the transaction being
+// proved must be appended to the path. The marshalled bytes of the transction
```
SweeXordious, line 18: [typo]
```suggestion
// proved must be appended to the path. The marshalled bytes of the transaction
```

pkg/prove/querier.go
```diff
@@ -0,0 +1,56 @@
...
+// proof (tmproto.TxProof) are returned.
+//
+// example path for proving the third transaction in that block:
+// custom/txInclusionProof/3
+func QueryTxInclusionProof(_ sdk.Context, path []string, req abci.RequestQuery) ([]byte, error) {
+	// parse the index from the path
+	if len(path) != 1 {
+		return nil, fmt.Errorf("unexpected query path length %d", len(path))
```
SweeXordious, line 26: [Non Blocking][Cosmetics]
```suggestion
		return nil, fmt.Errorf("expected query path length: 1 actual: %d ", len(path))
```

pkg/prove/querier.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading block: %w", err)
+	}
+	data, err := types.DataFromProto(&pbb.Data)
+	if err != nil {
+		panic(fmt.Errorf("error from proto block: %w", err))
+	}
+
+	// create and marshal the tx inclusion proof, which we return in the form of []byte
+	txProof, err := TxInclusion(appconsts.DefaultCodec(), data, uint64(index))
```
SweeXordious, line 45: [Unrelated][Question]
In the docs of `TxInclusion(...)`:
```
// TxInclusion uses the provided block data to progressively generate rows
// of a data square, and then using those shares to creates nmt inclusion proofs
// It is possible that a transaction spans more than one row. In that case, we
// have to return two proofs.
func TxInclusion(codec rsmt2d.Codec, data types.Data, txIndex uint64) (types.TxProof, error) {
```
I checked the implementation, and if I understand right, it will generate a proof for each row a transaction spans on?
If so, why does it state in the docs:
```
In that case, we have to return two proofs.
```
and also:
```
var proofs []*tmproto.NMTProof  //nolint:prealloc // rarely will this contain more than a single proof
```

SweeXordious (approved):
Awesome stuff :rocket:
Only questions to understand more and cosmetics.

pkg/prove/querier.go
evan-forbes: They seem to match to me, but perhaps we can dig a bit deeper and clarify them. not sure I'm understanding the question properly, mind elaborating? which part doesn't match or make sense?

SweeXordious: The part I am not getting is the **two** proofs. Why not any number of proofs? Will the number of proofs be the same as the number of rows?

evan-forbes: true! yeah, we can change that, but all normal txs will only ever result in two proofs. You're correct tho that the implementation can handle more than two if that ever does happen (submitting a tx with a giiiiaaaannnnnntttttt memo maybe?)

evan-forbes: https://github.com/celestiaorg/celestia-app/pull/708/commits/26b8baa00ee372f609e8de7260b010f65c94b696

SweeXordious: Awesome,  thanks a lot

SweeXordious (approved):
:shipit:

pkg/prove/proof.go
```diff
@@ -16,7 +16,7 @@ import (
 // TxInclusion uses the provided block data to progressively generate rows
 // of a data square, and then using those shares to creates nmt inclusion proofs
 // It is possible that a transaction spans more than one row. In that case, we
-// have to return two proofs.
+// have to return more than one proofs.
```
rootulp, line 19: [grammar] honestly may not be worth accepting this suggestion b/c it will dismiss approvals

```suggestion
// have to return more than one proof.
```

rootulp (approved):

pkg/prove/proof.go
evan-forbes: :grimacing: degen mergingggggggg

evan-forbes: I will sneak this into a different PR tho lol

PR #709, opened 2022-09-14
feat: move share consts from celestia-core

rootulp:
Part of https://github.com/celestiaorg/celestia-core/issues/841
Follow up to https://github.com/celestiaorg/celestia-core/pull/861

### DIFF
--- a/pkg/appconsts/appconsts.go
+++ b/pkg/appconsts/appconsts.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/celestiaorg/nmt/namespace"
 	"github.com/celestiaorg/rsmt2d"
+	"github.com/tendermint/tendermint/pkg/consts"
 )
 
 // These constants are sourced from:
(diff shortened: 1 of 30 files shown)


pkg/appconsts/appconsts.go
```diff
@@ -3,10 +3,72 @@ package appconsts
...
+	// MaxShareVersion is the maximum value a share version can be.
+	MaxShareVersion = 127
+)
+
+var (
+	// IntermediateStateRootsNamespaceID is the namespace reserved for
+	// intermediate state root data
+	// TODO(liamsi): code commented out but kept intentionally.
+	// IntermediateStateRootsNamespaceID = namespace.ID{0, 0, 0, 0, 0, 0, 0, 2}
+
```
evan-forbes, line 54: [nit][blocking question]
wdyt about forwarding a link here to the missing constants? This way people don't have to remember which ones are where

like 
```go
TxNamespaceID = consts.TxNamespaceID
```

evan-forbes: if not, I'm also ok with approving as is. I think we're the only ones that use them

rootulp: Great idea, will do for all consts still in celestia-core

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/709/commits/6dde9c0f9d6ae56eda9eed6f5c28b3a1cb44e586

evan-forbes (approved):

PR #710, opened 2022-09-15
Rename appconsts to use `Compact` and `Sparse` terminology

rootulp:
- Rename `ShareReservedBytes` => `CompactShareReservedBytes`
- Rename `TxShareSize` => `CompactShareContentSize`
- Rename `MsgShareSize` => `SparseShareContentSize`
- Improve doc comments

### DIFF
--- a/pkg/appconsts/appconsts.go
+++ b/pkg/appconsts/appconsts.go
@@ -17,23 +17,16 @@ const (
 	// NamespaceSize is the namespace size in bytes.
 	NamespaceSize = 8
 
-	// ShareInfoBytes is the number of bytes reserved for information. The info
-	// byte contains the share version and a start idicator.
-	ShareInfoBytes = 1
-
-	// ShareVersion is the current version of the share format
-	ShareVersion = uint8(0)
-
 	// CompactShareReservedBytes is the number of bytes reserved for the location of
 	// the first unit (transaction, ISR, evidence) in a compact share.
 	CompactShareReservedBytes = 1
 
 	// CompactShareContentSize is the number of bytes usable for data in a compact
 	// (i.e. transactions, ISRs, evidence) share.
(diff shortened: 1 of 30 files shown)


pkg/appconsts/appconsts.go
```diff
@@ -8,23 +8,25 @@ import (
 	"github.com/tendermint/tendermint/pkg/consts"
 )
 
-// These constants are sourced from:
+// These constants were originally sourced from:
```
SweeXordious, line 11: [Non Blocking][Suggestion]
Probably would make sense to put where this was decided. Something like:
```
// And, then, updated following the discussion under:
// https://github.com/celestiaorg/celestia-app/pull/660#discussion_r958603307
```

SweeXordious (commented):
:rocket:

evan-forbes:
I would approve, but I want to ensure that Rachid's question or suggestion has a chance to be addressed before automerging, even if its non-blocking/orthogonal

pkg/appconsts/appconsts.go
rootulp: Lines 11-12 link to celestia-specs because its relevant for several consts in the block below.

I think a `README.md` in the `share/` package that explains the difference between `CompactShare` vs `SparseShare` may be better than a link to the motivating GH discussion but we can include the link you suggest immediately above compact and sparse share consts and then remove once we have such a README.

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/710/commits/a81d232dacfc7610b53a564a6a56c84927753c23

SweeXordious: Sounds good, thanks

SweeXordious (approved):
:shipit:

PR #717, opened 2022-09-16
chore: define `OriginalSquareSize` inside `generateRandomBlockData`

rootulp:
Spun out of https://github.com/celestiaorg/celestia-app/pull/691

## Motivation
While iterating on the `TestMerge` test, I encountered an unexpected error
```
            	            	square size is not a power of two: 0
```

because I failed to declare the square size on the data returned from `generateRandomBlockData`. I moved a variable definition to avoid this scenario.

Very minor improvement. Can be closed if not helpful.

### DIFF
--- a/pkg/shares/shares_test.go
+++ b/pkg/shares/shares_test.go
@@ -224,7 +224,6 @@ func TestMerge(t *testing.T) {
 				tc.msgCount,
 				tc.maxSize,
 			)
-			data.OriginalSquareSize = appconsts.MaxSquareSize
 
 			rawShares, err := Split(data)
 			require.NoError(t, err)
@@ -279,7 +278,7 @@ func generateRandomBlockData(txCount, evdCount, msgCount, maxSize int) (data cor
 	data.Txs = generateRandomlySizedCompactShares(txCount, maxSize)
 	data.Evidence = generateIdenticalEvidence(evdCount)
 	data.Messages = generateRandomlySizedMessages(msgCount, maxSize)
-	data.OriginalSquareSize = 16
+	data.OriginalSquareSize = appconsts.MaxSquareSize
 	return data
 }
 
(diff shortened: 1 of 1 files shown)


pkg/shares/shares_test.go
```diff
@@ -275,12 +274,12 @@ func TestFuzz_Merge(t *testing.T) {
...
-	var out coretypes.Data
-	out.Txs = generateRandomlySizedCompactShares(txCount, maxSize)
-	out.Evidence = generateIdenticalEvidence(evdCount)
-	out.Messages = generateRandomlySizedMessages(msgCount, maxSize)
-	return out
+func generateRandomBlockData(txCount, evdCount, msgCount, maxSize int) (data coretypes.Data) {
+	data.Txs = generateRandomlySizedCompactShares(txCount, maxSize)
+	data.Evidence = generateIdenticalEvidence(evdCount)
+	data.Messages = generateRandomlySizedMessages(msgCount, maxSize)
+	data.OriginalSquareSize = appconsts.MaxSquareSize
```
evan-forbes, line 281: ideally we should calculate the square size here instead of just picking the max size, but this is going to be replaced in #692 since calculating the square size essentially requires calling `PrepareProposal`

evan-forbes (commented):
unsure about this one per comment. is this needed to fix a test?

appreciate the fix tho, and we can merge even it only fixes tests temporarily

evan-forbes:
will investigate this more tmrw

rootulp:
> we can merge even it only fixes tests temporarily

I don't need this change for tests to pass so this is strictly optional and can be closed if undesired

rootulp:
@evan-forbes do we still want this? If not, I'll close.

evan-forbes (approved):

evan-forbes:
no reason not to I don't think

evan-forbes:
we might have to revert or override the square size if we don't want the max each time, but that's nbd

PR #720, opened 2022-09-16
Remove valset/data commitment confirm logic from state machine

SweeXordious:
Remove valset/data commitment confirm logic from state machine in preparation for merging the branch and implementing **ADR 5: Reduce QGB state usage**


go.mod
```diff
@@ -30,6 +28,8 @@ require (
 require (
 	cosmossdk.io/errors v1.0.0-beta.7
 	cosmossdk.io/math v1.0.0-beta.2
+	github.com/cosmos/cosmos-sdk v0.0.0-00010101000000-000000000000
```
rootulp, line 31: [question] this version number seems odd, is it intended?

rootulp (approved):
Approved w/ a question

x/qgb/keeper/msg_server.go
```diff
@@ -25,186 +18,16 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
-
 	return &types.MsgValsetConfirmResponse{}, nil
 }
 
 // DataCommitmentConfirm handles MsgDataCommitmentConfirm.
 func (k msgServer) DataCommitmentConfirm(
-	c context.Context,
-	msg *types.MsgDataCommitmentConfirm,
+	_ context.Context,
+	_ *types.MsgDataCommitmentConfirm,
```
evan-forbes, line 30: [blocking]
perhaps we can leave a comment pointing to ADR005 here and in all of the other empty methods to describe why we're keeping empty ones.

go.mod
SweeXordious: it's certainly not. I am still investigating it

x/qgb/types/valset_confirm.go
```diff
@@ -9,21 +9,6 @@ import (
...
-	validator sdk.AccAddress,
-	signature string,
-) *MsgValsetConfirm {
-	return &MsgValsetConfirm{
-		Nonce:        nonce,
-		Orchestrator: validator.String(),
-		EthAddress:   ethAddress.Hex(),
-		Signature:    signature,
-	}
-}
```
evan-forbes, line 25: [blocking][question]
We'll still want to create these tx types, no?

x/qgb/keeper/msg_server.go
evan-forbes: also, I think it would still be useful to emit an event that some confirm was submitted, we just don't store it in the state.

evan-forbes (commented):
overall this is definitely headed in the correct direction, as it removes the state storing and querying, but I we might have removed a bit too much. Left some comments/questions

we still need to make sure we can create and submit the txs, and we might even want to emit an event (maybe not, this depends on the relayer design and if things like block explorers want to keep track of these things). (we could also potentially do these things in a different PR, but will leave it up to the author)

The only thing that needs removing is storing in the state

x/qgb/keeper/msg_server.go
SweeXordious: Cool, will do, thanks

x/qgb/types/valset_confirm.go
SweeXordious: I  thought maybe we want to take this to the orchestrator/relayer side and remove  anything that is not explicitly used in the state machine. What do you think?

evan-forbes: Hmmm... perhaps!! if we do that, then we have to entirely remove methods to msgServer that handle those types.

evan-forbes: yeah, I agree after thinking about it for a few minutes lol

x/qgb/keeper/msg_server.go
SweeXordious: Check  now please.
Also, I am leaving the handler as we might need it  down the line, not to have to hard fork or... Will remove it after implementing the orchestrator and relayer.

SweeXordious:
@evan-forbes 
> we still need to make sure we can create and submit the txs, and we might even want to emit an event (maybe not, this depends on the relayer design and if things like block explorers want to keep track of these things). (we could also potentially do these things in a different PR, but will leave it up to the author)

Orchestrators and relayers will be parsing the blocks. So, no need for events.
And for block explorer, I guess since we're not adding any checks on the submitted confirms, even if an explorer shows that 100 confirm was published, they might all be invalid. 
Thus, I  don't see a need for events.

However, I am happy to re-introduce and not delete until later.

evan-forbes:
currently blocked by https://github.com/celestiaorg/cosmos-sdk/pull/276 which is blocked by https://github.com/celestiaorg/cosmos-sdk/issues/279

evan-forbes:
I think you're right, since it would require running something that looks like a relayer (but doesn't actually relay) to know for sure, we can delete them.

Presumably tho, given that delegators can get slashed if a validator is not signing properly, they will be interested in that info. We should definitely want to make it easy to monitor, so as long as we have some mechanism for that. (perhaps a third node, relayer, orchestrator, and a monitor)

SweeXordious:
I don't see another way of monitoring besides parsing the blocks and checking if confirms are correct. And if we're parsing blocks, events won't be useful.
Anyway, I defer to you for the decision. I'm fine with both.

evan-forbes:
> I don't see another way of monitoring besides parsing the blocks and checking if confirms are correct. And if we're parsing blocks, events won't be useful.
Anyway, I defer to you for the decision. I'm fine with both.

yeah, same, I think you're right and we can delete them. We just have to keep this fact in mind and provide a way for people to monitor them is all

SweeXordious:
I will create an issue for this :+1:

evan-forbes:
> currently blocked by https://github.com/celestiaorg/cosmos-sdk/pull/276 which is blocked by https://github.com/celestiaorg/cosmos-sdk/issues/279

ahh sorry, I was wrong on this, thought this was for merging the qgb branch into main lolol :facepalm:

evan-forbes (approved):

go.mod
SweeXordious: this is probably coming from the conflicting cosmos-sdk and celestia-core versions.
I guess it will be fixed when we merge the cosmos-sdk QGB branch to the default one and we rebase on master. 
Will merge for now and take a look on it later

PR #747, opened 2022-09-20
feat(ProcessProposal)!: Check for message inclusion using subtree roots

### LINKED ISSUE #382: Tracking: Non-interactive defaults

### LINKED ISSUE #383: Refactor`ProcessProposal` to check that subtree roots are actually part of data root

evan-forbes:
This PR updates `ProcessProposal` to use subtree roots to check for message inclusion, instead of only relying on comparing the commitments. 

This is last non-interactive defaults PR :tada: 

closes #383


evan-forbes:
will rebase to fix the git history

x/payment/types/payfordata.go
```diff
@@ -150,13 +150,16 @@ func CreateCommitment(squareSize uint64, namespace, message []byte) ([]byte, err
...
+		tree := nmt.New(sha256.New())
 		for _, leaf := range set {
 			nsLeaf := append(make([]byte, 0), append(namespace, leaf...)...)
-			// here we hardcode pushing as axis 0 cell 0 because we never want
-			// to add the parity namespace to our shares when we create roots.
-			tree.Push(nsLeaf, rsmt2d.SquareIndex{Axis: 0, Cell: 0})
+			err := tree.Push(nsLeaf)
+			if err != nil {
+				return nil, err
+			}
```
evan-forbes, line 160: had to revert this change from #618 to get the commitments to match, which was really weird. I'm still unsure why this is and requires further debugging. We are using the nmt wrapper to create row and col roots, so it only makes sense to use them when we create commitment so that they match.

pkg/shares/utils.go
```diff
@@ -1,16 +1,17 @@
...
 	core "github.com/tendermint/tendermint/proto/tendermint/types"
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
 // DelimLen calculates the length of the delimiter for a given message size
-func DelimLen(x uint64) int {
-	return 8 - bits.LeadingZeros64(x)%8
+func DelimLen(size uint64) int {
+	lenBuf := make([]byte, binary.MaxVarintLen64)
+	return binary.PutUvarint(lenBuf, size)
```
evan-forbes, line 14: reverted this change to calculate the delim len how we used to, as the new way wasn't doing what we thought. We still need to add a unit test to make sure we don't make the same mistake again. This is implicitly tested in the new fuzzer added in this PR, and message inclusion checks.

app/process_proposal.go
```diff
@@ -122,7 +61,70 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
 
 	if !bytes.Equal(dah.Hash(), req.Header.DataHash) {
-		logInvalidPropBlockError(app.Logger(), req.Header, "proposed data root differs from calculated data root", err)
+		logInvalidPropBlock(app.Logger(), req.Header, "proposed data root differs from calculated data root")
+		return abci.ResponseProcessProposal{
+			Result: abci.ResponseProcessProposal_REJECT,
+		}
+	}
+
+	// iterate over all of the MsgPayForData transactions and ensure that they
```
rootulp, line 70:
```suggestion
	// iterate over all of the MsgPayForData transactions and ensure that their
```

rootulp (dismissed):

rootulp (approved):

PR #753, opened 2022-09-21
QGB nits + docs for the merge 2 

SweeXordious:


x/qgb/keeper/keeper_attestation.go
```diff
@@ -44,10 +44,14 @@ func (k Keeper) StoreAttestation(ctx sdk.Context, at types.AttestationRequestI)
 // SetLatestAttestationNonce sets the latest attestation request nonce, since it's
 // expected that this value will only increase by one and it panics otherwise.
 func (k Keeper) SetLatestAttestationNonce(ctx sdk.Context, nonce uint64) {
-	if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != nonce {
-		panic("not incrementing latest attestation nonce correctly")
-	}
+	if k.CheckLatestAttestationNonce(ctx) {
```
rootulp, line 47: Sorry my proposal didn't include the negation but `CheckLatestAttestationNonce` returns `false` when the latest attestation nonce isn't set so we need to negate this check

```suggestion
	if !k.CheckLatestAttestationNonce(ctx) {
```

rootulp (changes requested):
Will approve after:
1. File is go formatted
2. Negate check

SweeXordious:
@rootulp Sorry, wasn't focusing earlier. The best we can do to refactor is:
```golang
	store := ctx.KVStore(k.storeKey)
	if k.CheckLatestAttestationNonce(ctx) {
		if k.GetLatestAttestationNonce(ctx)+1 != nonce {
			panic("not incrementing latest attestation nonce correctly")
		}
		store.Set([]byte(types.LatestAttestationtNonce), types.UInt64Bytes(nonce))
	}
	store.Set([]byte(types.LatestAttestationtNonce), types.UInt64Bytes(nonce))
```
But I don't think it is worth it.
In fact, we should never panic if the latest nonce doesn't exist in store. Because, if it doesn't, we just proceed and initialize it.
We could add a check `nonce == 0` to know that we're initializing state for the first time. However, we might need in the future, after a hard fork, to manually set the first nonce, to a random value that is different from 0, and initialize the store using it.
Let me know what you think.

SweeXordious:
If you don't mind, I can rename this PR to contain extra nits that are taken from the Merge PR.

evan-forbes (approved):
nice additions

PR #762, opened 2022-09-21
fix: Update Dockerfiles and fix CI

SweeXordious:
Currently, the CI is failing because of not found `gcc` and standard C libraries when building the ephemeral cluster docker image. This PR fixes that.
Also, removes unnecessary `git` install.


evan-forbes (approved):
LGTM makes sense utACK

Issue #777, opened 2022-09-23
CompactShareSplitter `Count()` doesn't implement its doc comment


rootulp:
## Problem
CompactShareSplitter claims to return the current number of shares that will be made if exporting but it actually returns one less than the number of shares that would be created if the pending share has any bytes written to it.

## Link

https://github.com/celestiaorg/celestia-app/blob/907da5f587cee1cbabae4549db7ada0df35e122b/pkg/shares/split_compact_shares.go#L156-L165

## Proposal

1. Fix either the doc comment or the implementation depending on which one is incorrect
3. [Optional but preferred] write unit tests for the behavior of this function

evan-forbes:
Originally this was used for the mechanism to arrange a square during prepare proposal, although its no longer used for that. In that implementation, we needed the counter and the available bytes, but now we don't so we could change this function to follow what the comment says.

PR #778, opened 2022-09-23
fix: compact share `Count`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/777

I choose to fix the implementation to match the doc comment because it seems like a leaky abstraction for this function to also return the # of available bytes in the pending share. By removing the second return value, the compact share `Count` more closely matches the sparse share [`Count`](https://github.com/rootulp/celestia-app/blob/5eab874a214861a5de9af2a662c9bcbe352c5608/pkg/shares/split_sparse_shares.go#L87-L90) function signature.

### DIFF
--- a/app/estimate_square_size.go
+++ b/app/estimate_square_size.go
@@ -92,14 +92,8 @@ func calculateCompactShareCount(txs []*parsedTx, evd core.EvidenceList, squareSi
 			panic(err)
 		}
 	}
-	txCount, available := txSplitter.Count()
-	if appconsts.CompactShareContentSize-available > 0 {
-		txCount++
-	}
-	evdCount, available := evdSplitter.Count()
-	if appconsts.CompactShareContentSize-available > 0 {
-		evdCount++
-	}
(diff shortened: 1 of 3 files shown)


app/estimate_square_size.go
```diff
@@ -92,14 +92,8 @@ func calculateCompactShareCount(txs []*parsedTx, evd core.EvidenceList, squareSi
...
-	txCount, available := txSplitter.Count()
-	if appconsts.CompactShareContentSize-available > 0 {
-		txCount++
-	}
-	evdCount, available := evdSplitter.Count()
-	if appconsts.CompactShareContentSize-available > 0 {
-		evdCount++
-	}
+	txCount := txSplitter.Count()
+	evdCount := evdSplitter.Count()
```
evan-forbes, line 96: [optional]
combine with return statement

evan-forbes (dismissed):

evan-forbes (approved):

PR #788, opened 2022-09-26
fix: update default `timeout_commit` when running `sed` on genesis in `single-node.sh` script

SweeXordious:
the default is `25s` now instead of `5s`


evan-forbes:
I'm actually against this change, as this script is exclusively used for testing celestia-app

evan-forbes:
I'm not sure who uses this script anymore, so there's no need to change it back

evan-forbes:
wait nvm, I'm dumb

SweeXordious:
what just happened

Issue #789, opened 2022-09-26
`joinByteSlices` in proof_test contains a bug


rootulp:
## Problem
`joinByteSlices` in `proof_test.go` invokes `ParseDelimiter` with an entire share when the input should be stripped of namespace id and info reserved byte. This hasn't resulted in failing tests because `ParseDelimiter` won't modify the share at all if a length delimiter isn't encountered.

## Motivation
I broke this test in a separate PR and debugging it proved confusing given the existing test seems incorrect. 

## Proposal
Fix the `joinByteSlices` implementation.

## Link
https://github.com/celestiaorg/celestia-app/blob/ab55fc4164407883ddd97f9d607f4a9e34fbcfd1/pkg/prove/proof_test.go#L191-L198

evan-forbes:
> won't modify the share at all if a length delimiter isn't encountered.

just add a bit of detail here, the length of the delimiter it is encountering is zero because the bytes that its thinks are the delimiter are the tx namespace id.

PR #791, opened 2022-09-26
fix: `joinByteSlice` implementation

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/789

### DIFF
--- a/pkg/prove/proof_test.go
+++ b/pkg/prove/proof_test.go
@@ -3,7 +3,6 @@ package prove
 import (
 	"bytes"
 	"math/rand"
-	"strings"
 	"testing"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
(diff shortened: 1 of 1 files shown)


pkg/prove/proof_test.go
```diff
@@ -188,13 +185,14 @@ func TestTxSharePosition(t *testing.T) {
...
-	out := make([]string, len(s))
-	for i, sl := range s {
-		sl, _, _ := shares.ParseDelimiter(sl)
-		out[i] = string(sl[appconsts.NamespaceSize+appconsts.ShareInfoBytes:])
+// stripCompactShares strips the universal prefix (namespace, info byte) and
+// reserved byte from a list of compact shares and joins them into a single byte
+// slice.
+func stripCompactShares(compactShares [][]byte, start uint64, end uint64) (result []byte) {
+	for i := start; i <= end; i++ {
+		result = append(result, compactShares[i][appconsts.NamespaceSize+appconsts.ShareInfoBytes+appconsts.CompactShareReservedBytes:]...)
```
rootulp, line 193: [for reviewers] note we don't need to parse the unit length delimiter that is prefixed to a transaction because the test checks that this share range contains the raw tx bytes. Therefore, it is safe to keep the unit length delimiter in the stripped compact shares returned from this function

evan-forbes (approved):
nice catch

Issue #650, opened 2022-08-23
EPIC: Add an official spec for the app


evan-forbes:
While archived, the [specs](https://github.com/celestiaorg/celestia-specs) contain a lot of really important and critical designs that are effectively cannon. We should move as much as we can over from them, while refactoring, modifying, and updating to match the current implementation.

some special love is needed for our the encoding section, as its much more difficult to decipher the share splitting/merging implementation

- [x] #688
- [x] #727
- [x] #728
- [x] #729
- [x] #730
- [x] #732
- [x] #733
- [x] #792
- [x] #866

# Pure specs

- [ ] #735
- [ ] #736
- [ ] #737
- [x] #743
- [x] #751
- [x] #784
- [x] #883
- [x] #995
- [ ] #1007
- [x] #1063
- [ ] #1066
- [x] #1067
- [x] #1349
- [x] #1435
- [x] #1437
- [x] #1440 
- [x] #1453
- [ ] #1459
- [x] #1488
- [x] #1532

# Needs implementation

- [x] #431
- [x] #731
- [x] #734
- [x] #739
- [ ] #795 (blocked by #835)
- [x] #940
- [x] #1487

rootulp:
One alternative approach is to unarchive the specs repo and remove the outdated specs.

adlerjohn:
Per-repo specs will be easier to maintain, as each team can update spec in tandem with implementation changes.

evan-forbes:
I'm going to use this comment to continually add reminders as I think of them. Others should feel free to edit this comment or post their own comment as well.

we introduced a new validity rule here for `MsgPayForData` and `MsgWirePayForData`.
https://github.com/celestiaorg/celestia-app/pull/666

rootulp:
Relevant: https://github.com/celestiaorg/celestia-specs/issues/213

evan-forbes:
nice!! I didn't see that

adlerjohn:
Can we turn this into a tracking issues @evan-forbes

rootulp:
The **Pure Specs** issues are blocking the merge of `specs-staging` into `main`.

evan-forbes:
referencing a comment with changes that are likely a good idea to include above 

https://github.com/celestiaorg/celestia-app/pull/796#issuecomment-1467871463

evan-forbes:
here's an list from a recent sync on overhauling the spec. Some of these are duplicates from the main comment in this issue :pray: 

- [ ] #1067
[...]

PR #796, opened 2022-09-27
Spec migration

adlerjohn:
Ref: #650

Staging branch for specs migration. Draft PR so it can be tracked.


rootulp:
hmm I can't merge `main` into this branch

Screen Shot 2022-09-28 at 2 55 15 PM


I wanted to merge `main` to pick up the new CODEOWNERS file

musalbas:
What's the status of this?

rootulp:
@musalbas apologies for delayed response. @adlerjohn and I triaged issues in https://github.com/celestiaorg/celestia-app/issues/650 today. 

I think that the `specs-staging` branch is already a more accurate representation of the specs than https://celestiaorg.github.io/celestia-specs/latest/index.html so I'd like to merge `specs-staging` to `main` ASAP. But if we want to block this merge on updating specs to match the existing implementation, we can block on all the issues in the "Pure specs" section.

musalbas:
What's the latest with this and generally adding specs to celestia-app/core now?

rootulp:
This branch is the most up to date version of the specs but it's not publicly visible yet. IMO we should merge this branch to `main`, make it publicly visible, and take down the existing specs site at https://celestiaorg.github.io/celestia-specs/latest/index.html

The remaining specs issues are tracked under https://github.com/celestiaorg/celestia-app/issues/650 . We intend on prioritizing specs and documentation work after the core/app feature freeze for incentivized testnet.

cmwaters:
I have some general comments that can possible be followed up at a later point:

- Can we break up data structures into smaller more self-explanatory files so that parts can be easily accessed in the table of contents. I also think it's less intuitive to explain a spec by just dumping all the data structures that are included in it. Would prefer more of a hierarchical function-orientated tree.
- Can we add a section explaining how the "celestia app" fits in with the rest of the system and include links so people can be easily redirected if they're interested with something in "celestia node" for example. Furthermore, I think calling it "celestia app" can be better worded. As this isn't the specification of the entire system, rather a component of it, we should look to distinguish it functionally. Call it "celestia data publication specification" (to complement "celestia data availability specfication") if we want to be deliberate about having two networks. 
- I don't think `Rationale` should be a separate section but should be embedded right next to the part in the spec it refers to. I should read what a certain protocol is and directly after why it is so. 
- It's evident that the specification includes not just novel work specific to Celestia but the components that it is built on top of. There is a lot of work on Tendermint's consensus algorithm (such as block validity, leader selection) and how the validator set is controlled. What about other components like governance, IBC, vesting, upgrades that still seem to be missing. It seems that not all state transition messages have been documented.
- Is QGB spec somewhere else? 
- Can we have some preamble/context in the top level pages to each of the subpages.
- Are we sure we want a copy of many of the proto files in the spec as well? These are not the canonical ones that get used in generation and will constantly become outdated.

evan-forbes (commented):
are the markdown linter complaints valid?

```
  ERROR: 2 dead links found!
  [✖] .consensus.md#constants → Status: 400
  [✖] ../rationale/distributing_rewards.md → Status: 400
 ```
 
 other than that I think we're okay to merge (and actually merge, not squash!)
 
 we should also add @cmwaters feedback into the spec epic as @rootulp mentioned because I think its really good

**Comment by rootulp** [comment]
Will merge this by EOD tomorrow if no more feedback.

File: .github/workflows/gh-pages.yml [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+      - name: Setup mdBook
+        uses: peaceiris/actions-mdbook@v1
+        with:
+          mdbook-version: "0.4.21"
+
+      - name: Build book
+        run: mdbook build specs
+
+      - name: Deploy main
+        if: github.ref == 'refs/heads/main' && github.event_name == 'push'
```
MSevey, line 24: [nit] I'm pretty sure you just need to check the `github.event_name` since we are already filtering to only triggering on the `main` branch for `push` events up top. But if you'd prefer to keep this here as an additional level of protection that is fine. 

```suggestion
        if: github.event_name == 'push'
```

evan-forbes:
its weird that the docker builds are still failing here despite passing in other PRs, we didn't change anything related here afaict

.github/workflows/gh-pages.yml
rootulp: good point! Will adopt your suggestion

.github/workflows/gh-pages.yml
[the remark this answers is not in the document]
rootulp: good idea! Created https://github.com/celestiaorg/celestia-app/issues/1495

MSevey (approved):

Issue #711, opened 2022-04-15
Increase share size to 512 bytes


adlerjohn:
I like big blocks and I cannot lie.

Change `ShareSize` to `512`: https://github.com/celestiaorg/celestia-core/blob/c193de698f889a3c78565ac4a73899326a23d544/pkg/consts/consts.go#L15

This will also require a change to the reserved bytes to `2`: https://github.com/celestiaorg/celestia-core/blob/c193de698f889a3c78565ac4a73899326a23d544/pkg/consts/consts.go#L21

rootulp:
We want to do this as a fast follow after https://github.com/celestiaorg/celestia-app/issues/659

rootulp:
Based on synchronous discussion, this issue needs investigation to determine what the share size should be.

evan-forbes:
to help figure out what breaks, we are doing https://github.com/celestiaorg/celestia-app/issues/825 first, which could potentially be merged and close this issue

musalbas:
| ODS | 256 byte shares | 512 byte shares | 1024 byte shares |
| - | - | - | - |
| 4 | 53 MB/s | 95 MB/s | 169 MB/s
| 8 | 116 MB/s | 203 MB/s | 297 MB/s
| 16 | 208 MB/s | 367 MB/s | 540 MB/s
| 32 | 265 MB/s | 478 MB/s | 697 MB/s
| 64 | 300 MB/s | 540 MB/s | 745 MB/s
| 128 | 305 MB/s | 534 MB/s | 695 MB/s
| 256 | 293 MB/s | 496 MB/s | 634 MB/s
| 512 | 276 MB/s | 458 MB/s | 612 MB/s

rootulp:
https://github.com/celestiaorg/celestia-app/pull/850 merged so this issue seems close-able. In the spirit of:

> I like big blocks and I cannot lie.

if we can increase share size to 1024 bytes then can we please create a new issue to tackle that?

PR #801, opened 2022-09-27
chore: prefer appconsts for # of reserved bytes

rootulp:
## Motivation

We will increase the # of reserved bytes from 1 to 2 when we do https://github.com/celestiaorg/celestia-app/issues/711. The current implementation hardcodes only 1 reserved byte rather than relying on `appconsts.CompactShareReservedBytes`.

### DIFF
--- a/pkg/shares/split_compact_shares.go
+++ b/pkg/shares/split_compact_shares.go
@@ -57,10 +57,11 @@ func (css *CompactShareSplitter) WriteEvidence(evd coretypes.Evidence) error {
 
 // WriteBytes adds the delimited data to the underlying compact shares.
 func (css *CompactShareSplitter) WriteBytes(rawData []byte) {
-	// if this is the first time writing to a pending share, we must add the
-	// reserved bytes
+	// if this is the first time writing to a pending share, we must add a
+	// placeholder for the reserved bytes
 	if len(css.pendingShare.Share) == appconsts.NamespaceSize+appconsts.ShareInfoBytes {
-		css.pendingShare.Share = append(css.pendingShare.Share, 0)
+		reservedBytesPlaceholder := make([]byte, appconsts.CompactShareReservedBytes)
+		css.pendingShare.Share = append(css.pendingShare.Share, reservedBytesPlaceholder...)
 	}
 
 	txCursor := len(rawData)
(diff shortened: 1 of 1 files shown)


pkg/shares/split_compact_shares.go
```diff
@@ -57,10 +57,11 @@ func (css *CompactShareSplitter) WriteEvidence(evd coretypes.Evidence) error {
 
 // WriteBytes adds the delimited data to the underlying compact shares.
 func (css *CompactShareSplitter) WriteBytes(rawData []byte) {
-	// if this is the first time writing to a pending share, we must add the
-	// reserved bytes
+	// if this is the first time writing to a pending share, we must add a
+	// placeholder for the reserved bytes
```
evan-forbes, line 61: [question]
It's been a minute since I've looked at this code, is this a placeholder that we overwrite or is this an edge case where we actually want zeros?

evan-forbes (commented):
using the constants instead of hardcoded values is :fire:, I did have two questions tho

should we also handle is this https://github.com/celestiaorg/celestia-app/blob/a33563d67a1794594c371e0faf641ca2e79afbf7/pkg/shares/split_compact_shares.go#L92-L98

pkg/shares/split_compact_shares.go
rootulp: it seems like a placeholder but it doesn't look like it's overwritten for [this case](https://github.com/celestiaorg/celestia-app/pull/801/files#diff-3da7e413fef3c90c01286110ba9bdfc3d61bbb360355bb32adede4e7b3ebfe5bR74-R77) so we need unit tests to identify if there's a bug there

rootulp:
> should we also handle is this

Good idea. Handled in https://github.com/celestiaorg/celestia-app/pull/801/commits/7eafb902479e7628d3fff9724674c7d3ca2c8cbe

pkg/shares/split_compact_shares.go
evan-forbes: agree, I don't think we have an explicit test for reserved bytes and we just ignore them when merging.

evan-forbes (dismissed):
[optional]
its super nit picky :grimacing: so I'm leaving as optional. imo, we should remove the placeholder terminology, as I'm not sure that's not what is currently happening. we can also just handle this in #802 as well after further testing.

evan-forbes (approved):
:pray: thanks!

PR #803, opened 2022-09-27
chore: rename to `ContinuationCompactShareContentSize`

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/725

Rename `CompactShareContentSize` to `ContinuationCompactShareContentSize` because we will soon introduce a new appconst for `FirstCompactShareContentSize`.

Issue #725, opened 2022-09-16
Encode data length varint into first compact share


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/660 introduces a `data length (varint 1 to 10 bytes)` to the first compact share in a reserved namespace where `data length (varint 1 to 10 bytes)` is the sum of lengths of all units in that namespace. 

For example, in the transaction namespace `data length` is the sum of all transaction lengths. Note this is distinct from the varint prefix to each individual transaction.

## Proposal

Modify the files below to account for the new varint. Note the varint is dependent on all data in the namespace so it is only determinable in the [`Export`](https://github.com/celestiaorg/celestia-app/blob/main/pkg/shares/split_compact_shares.go#L115-L116) function. However, it must be prefixed to the first share which will modify the subsequent share lengths and therefore force a rewrite of a subsequent shares.
- https://github.com/celestiaorg/celestia-app/blob/main/pkg/shares/parse_compact_shares.go
- https://github.com/celestiaorg/celestia-app/blob/main/pkg/shares/split_compact_shares.go

evan-forbes:
yeah, I've ran into similar issues a lot with the non-interactive defaults. 

perhaps we can simply put the varint at the end instead of the beginning. This would allow us to continue to lazily write the compactShareSplitter, while also not having to fill in all of the reserved bytes afterwords.

rootulp:
> perhaps we can create a second version to use in the info byte, and continue to not use a varint for compact shares

Is you proposal as follows?

**version 0**:
- compact share does not contain data length varint
- sparse share does contain data length varint

**version 1**: 
- compact share does contain data length varint
- sparse share does contain data length varint

where we implement **version 0** now and punt **version 1** to a later date?

rootulp:
We want to implement this ASAP because it is share format breaking. Two clarifying notes:
1. The data length for the transaction reserved namespace includes the length all transactions, and the length delimiters prefix to each transaction. It does not include the universal prefix (namespace, info byte), or reserved byte in each share or the total length in the first share. 
2. We may be able to reserve N bytes in the first compact share for the data length varint and write it to in the export function without rewriting all subsequent shares if we overestimate the # of bytes needed for the data length.

PR #804, opened 2022-09-27
feat: introduce `FirstCompactShareContentSize`

rootulp:
~~Blocked on https://github.com/celestiaorg/celestia-app/pull/803~~
Part of https://github.com/celestiaorg/celestia-app/issues/725

### DIFF
--- a/pkg/appconsts/appconsts.go
+++ b/pkg/appconsts/appconsts.go
@@ -31,9 +31,11 @@ const (
 	// the first unit (transaction, ISR, evidence) in a compact share.
 	CompactShareReservedBytes = 1
 
-	// CompactShareContentSize is the number of bytes usable for data in a compact
-	// (i.e. transactions, ISRs, evidence) share.
-	CompactShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes - CompactShareReservedBytes
+	// ContinuationCompactShareContentSize is the number of bytes usable for
+	// data in a continuation compact share. A continuation share is any
+	// share in a reserved namespace that isn't the first share in that
+	// namespace.
+	ContinuationCompactShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes - CompactShareReservedBytes
 
 	// SparseShareContentSize is the number of bytes usable for data in a sparse (i.e.
 	// message) share.
--- a/pkg/appconsts/appconsts.go
+++ b/pkg/appconsts/appconsts.go
@@ -2,6 +2,7 @@ package appconsts
 
 import (
 	"bytes"
+	"encoding/binary"
 
 	"github.com/celestiaorg/nmt/namespace"
 	"github.com/celestiaorg/rsmt2d"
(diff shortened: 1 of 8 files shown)


pkg/appconsts/appconsts.go
```diff
@@ -115,4 +116,29 @@ var (
...
+	// contains less space for data than a ContinuationCompactShare because the
+	// first compact share includes a total data length varint.
+	FirstCompactShareContentSize = ContinuationCompactShareContentSize - FirstCompactShareDataLengthBytes
 )
+
+// NumberOfBytesVarint calculates the number of bytes needed to write a varint of n
+func NumberOfBytesVarint(n uint64) (numberOfBytes int) {
+	buf := make([]byte, binary.MaxVarintLen64)
+	return binary.PutUvarint(buf, n)
+}
```
evan-forbes, line 144: it makes sense why we need this here, but I'd prefer to somehow get around having two distinct copies of the same logic (unless we have plans to change the below that I'm unaware of). Could we delete the other code and only use this one, or if we don't want to frequently import a function from a consts package, then we could wrap this function there. wdyt?

https://github.com/celestiaorg/celestia-app/blob/595fd3bc1bdbf9f0ab01a8ef6ce2a2156a7e0cec/pkg/shares/utils.go#L11-L15

pkg/appconsts/appconsts.go
```diff
@@ -115,4 +116,29 @@ var (
...
+	// namespace. This value is the maximum number of bytes required to store
+	// the data length of a block that only contains compact shares of one type.
+	// For example, if a block contains only evidence then it could contain:
+	// MaxSquareSize * MaxSquareSize * ShareSize bytes of evidence.
+	//
+	// Assuming MaxSquareSize is 128 and ShareSize is 256, this is 4194304 bytes
+	// of evidence. It takes 4 bytes to store a varint of 4194304.
+	//
+	// https://go.dev/play/p/MynwcDHQ_me
+	FirstCompactShareDataLengthBytes = NumberOfBytesVarint(MaxSquareSize * MaxSquareSize * ShareSize)
```
evan-forbes, line 131: :+1:

pkg/appconsts/appconsts.go
rootulp: I opted to duplicate because there is precedence in this repo for duplicating functions when used in distinct package 

Ref:
- https://github.com/rootulp/celestia-app/blob/47a88feadb079c03334f78c6b3e3cf554e4fc1d1/pkg/shares/non_interactive_defaults.go#L80
- https://github.com/rootulp/celestia-app/blob/47a88feadb079c03334f78c6b3e3cf554e4fc1d1/app/estimate_square_size.go#L230

But I'm open to consolidating on one function definition. Using the function defined in the `shares` package introduces an import cycle:

Screen Shot 2022-09-27 at 8 12 12 PM

so will fallback to the other option proposed: delete the definition in the shares package and use the one in `appconsts`. IMO this isn't ideal because `appconsts` will export a util when I would only expect constants to be defined in it

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/804/commits/285d1a221119d1bca4b7525992bed70e21fb4f37

evan-forbes: > Using the function defined in the shares package introduces an import cycle:

yeah understood that much :sweat_smile:, I meant to wrap the import in the other direction. or just not export the code in the consts package if we don't want people using it, not unlike the functions linked in the above comment.

evan-forbes (approved):
disabled auto merge to provide time to pursue another option if you want, but I am satisfied, so I'm approving

:pray:

pkg/appconsts/appconsts.go
evan-forbes: fwiw, this was likely subconsciously driven by this exact function causing me a full day of debugging raw bytes and hashes

https://github.com/celestiaorg/celestia-app/pull/747/commits/0749b62ecc33a6920c050b90a34bc1ab1f60b60d#r975882452

rootulp: TBH I don't understand the implementation of `DelimLen` that uses `%8` but we have another instance of it here: https://github.com/celestiaorg/celestia-app/blob/595fd3bc1bdbf9f0ab01a8ef6ce2a2156a7e0cec/x/payment/types/payfordata.go#L237-L239

> sorry I meant to wrap the import in the other direction

Sorry I don't understand this approach. Is it what was done in https://github.com/celestiaorg/celestia-app/commit/285d1a221119d1bca4b7525992bed70e21fb4f37 or something else?

> or just not export the code
I'm onboard with this option. Concretely:
1. Revert https://github.com/celestiaorg/celestia-app/commit/285d1a221119d1bca4b7525992bed70e21fb4f37
1. Rename `NumberOfBytesVarint` to `numberOfBytesVarint` in appconsts

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/804/commits/6dc3066a5d7d7816c238d3b7dc6b6c941e0a590b and https://github.com/celestiaorg/celestia-app/pull/804/commits/038e423cccb0c8f38eeb88bdf3cbf2506c818e5d

evan-forbes (approved):

pkg/appconsts/appconsts.go
evan-forbes: > TBH I don't understand the implementation of DelimLen that uses %8 but we have another instance of it here:

yeah, I don't get it either. thank for finding this!!!! BURN IT WITH FIRE #809

PR #781, opened 2022-09-23
feat: write data length to first compact share

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/725

PR #810, opened 2022-09-28
feat: return `bytesOfPadding`

rootulp:
Spun out of https://github.com/celestiaorg/celestia-app/pull/781

In https://github.com/celestiaorg/celestia-app/pull/781 , there is a need to calculate the total data length written to a compact share splitter so that a varint of this data length can be written to the first share. It becomes easier to perform this calculation if the utility function `zeroPadIfNecessary` returns the `bytesOfPadding` because the export command can subtract the `bytesOfPadding` from the number of bytes used by all shares in the compact share splitter.

Without this change, it may be possible to calculate the # of bytes written by summing the # of bytes written to shares and the number of bytes in the pending share _prior_ to adding padding to the pending share but this option includes more edge cases.

### DIFF
--- a/pkg/shares/shares_test.go
+++ b/pkg/shares/shares_test.go
@@ -176,19 +176,24 @@ func Test_zeroPadIfNecessary(t *testing.T) {
 		width int
 	}
 	tests := []struct {
-		name string
-		args args
-		want []byte
+		name               string
+		args               args
+		wantPadded         []byte
+		wantBytesOfPadding int
 	}{
-		{"pad", args{[]byte{1, 2, 3}, 6}, []byte{1, 2, 3, 0, 0, 0}},
-		{"not necessary (equal to shareSize)", args{[]byte{1, 2, 3}, 3}, []byte{1, 2, 3}},
-		{"not necessary (greater shareSize)", args{[]byte{1, 2, 3}, 2}, []byte{1, 2, 3}},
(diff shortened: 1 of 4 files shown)


pkg/shares/shares_test.go
```diff
@@ -176,19 +176,24 @@ func Test_zeroPadIfNecessary(t *testing.T) {
 		width int
 	}
 	tests := []struct {
```
rootulp, line 178: [for reviewers] I found it a little confusing that the `zeroPadIfNecessary` function is defined in `parse_sparse_shares.go` but tested in `shares_test.go` when I would've expected it to be tested in `parse_sparse_shares_test.go`.

Additionally this function is used by sparse shares so proposal to move the function to `utils.go` (and this test to `utils_test.go`)

will do this as a follow-up if reviewers agree

evan-forbes (approved):
seems reasonable. @rootulp perhaps we can add more info to this PR in a comment or the description as to why this is needed for future reference (and my own understanding lol)

pkg/shares/shares_test.go
evan-forbes: yeah why not. my guess was this was just an oversight when moving things around the past few weeks

rootulp:
Sorry for not including a proper PR description initially. I updated the PR description above to provide more context.

Issue #802, opened 2022-09-27
Implement API for parsing out of context shares with explicit tests for reserved bytes


evan-forbes:
We don't actually use the reserved bytes for anything at the moment, and we don't have a way to use them to parse out of context shares. We should create an api that parses compact shares without the context of starting at the beginning of the namespace/message. We might even want to refactor the existing ParseTxs (or whatever compact shares) to use this implementation under the hood. Having this API should help us test reserved bytes to ensure their integrity.

evan-forbes:
we will likely have to implement this in a different repo, because if optimint is to use it, then it cannot require including celestia-core/celestia-app/our fork of the sdk in the dependency tree.

rootulp:
We can implement this API inside celestia-app (potentially the shares package). This issue includes corresponding unit tests.

Issue #630, opened 2022-08-18
Recreate the manual testing of share encoding


evan-forbes:
Why is this test commented?

_Originally posted by @SweeXordious in https://github.com/celestiaorg/celestia-app/pull/627#discussion_r948585018_

When moving share encoding and decoding logic to the app, we had to comment out a test because it was designed around some tendermint logic, like creating votes and using methods instead of functions to encode shares. We should recreate this test by changing those requirements.

evan-forbes:
we have since added quite a few manual tests, and I'm not sure of the value that we would get from this anymore. I guess the manual tests would act as a forcing function to notify reviewers if anything changes the encoding, but that's about it.

Leaving this as a stretch issue for the epic, but I am also okay with closing this as unplanned. Will defer to @rootulp

rootulp:
This issue refers to the `TestMakeShares` test which was commented out and eventually removed. It appears to have tested

```
// 			got := tt.args.data.SplitIntoShares()
```

which no longer exists but it resembles https://github.com/celestiaorg/celestia-app/blob/0b85a0fabf07ff02bf03e9205638c6cd5462afba/pkg/shares/share_splitting.go#L25 which doesn't have unit tests so I think we can update this issue to add unit tests to that function.

rootulp:
That file has been migrated to the go-square repo: https://github.com/celestiaorg/go-square/blob/main/shares/share_splitting.go

SplitTxs has tests but SplitBlobs doesn't: https://github.com/celestiaorg/go-square/blob/main/shares/share_splitting_test.go

PR #813, opened 2022-09-28
test: transaction splitting

rootulp:
Define unit tests for transaction splitting. These should be helpful when we tackle https://github.com/celestiaorg/celestia-app/issues/802

### DIFF
--- a/pkg/shares/share_splitting_test.go
+++ b/pkg/shares/share_splitting_test.go
@@ -27,7 +27,7 @@ func TestSplitTxs(t *testing.T) {
 				append([]uint8{
 					0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, // namespace id
 					0x1, // info byte
-					0x0, // BUG: reserved byte should be non-zero
+					0x0, // reserved byte
 					0x1, // unit length of first transaction
 					0xa, // data of first transaction
 				}, bytes.Repeat([]byte{0}, 244)...), // padding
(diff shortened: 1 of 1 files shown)


evan-forbes:
ref: part of (and an improvement of!) #630 I think

pkg/shares/share_splitting_test.go
```diff
@@ -0,0 +1,98 @@
...
+			name: "one small tx",
+			txs:  coretypes.Txs{coretypes.Tx{0xa}},
+			want: [][]uint8{
+				append([]uint8{
+					0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, // namespace id
+					0x1, // info byte
+					0x0, // reserved byte
+					0x1, // unit length of first transaction
+					0xa, // data of first transaction
+				}, bytes.Repeat([]byte{0}, 244)...), // padding
```
evan-forbes, line 33: [optional]
these will technically break when we increase share size, I think we could subtract from the constants here, but considering we are rarely doing that I'm 100% fine with leaving it.

having a test that breaks on purpose if the share size is changed might even be desirable.

evan-forbes (commented):
dope :pray: very good tests as documentation.

evan-forbes (approved):

pkg/shares/share_splitting_test.go
rootulp: Agreed, this test will definitely break when we increase share size because the bytes of padding will be incorrect and the reserved byte will likely be two bytes instead of one.

I'm on-board with having these tests break when we introduce share breaking changes.

Issue #779, opened 2022-09-23
Stop forcing the last reserved byte to 0 for compact shares


rootulp:
## Question

Why is reserved byte of last compact share forced to zero? This behavior seems undesirable if the last compact share contains a new transaction.

## Context
- https://github.com/celestiaorg/celestia-core/pull/637#issuecomment-1091461481
- https://github.com/celestiaorg/celestia-core/pull/637#issuecomment-1092278815
- https://github.com/celestiaorg/celestia-core/pull/637#issuecomment-1092585704

## Link

https://github.com/celestiaorg/celestia-app/blob/907da5f587cee1cbabae4549db7ada0df35e122b/pkg/shares/split_compact_shares.go#L141-L145

evan-forbes:
I'm confused by this as well.

> As such, the reserved bytes should use the well-defined zero value to indicate a termination of the transaction stream.

@adlerjohn ?

perhaps we should rethink this now that we are adding the length delimiter to compact shares (formerly contiguous shares)

rootulp:
Based on synchronous discussion, it seems like the last transaction share _shouldn't_ force the reserve byte to zero.

Issue #817, opened 2022-09-29
Reserved byte in compact shares aren't set correctly


rootulp:
The reserved byte in a compact share be set to the location in the share where the length of the first unit in that share starts. However, the reserved byte is `0` in cases where a unit (i.e. transaction) does start in the share. See unit tests in `pkg/shares/share_splitting_test.go` for scenarios where this occurs.

Note: this is related to https://github.com/celestiaorg/celestia-app/issues/779 but occurs in cases where the share isn't the last compact share in a reserved namespace so this is a separate issue to fix the failure scenarios described in `pkg/shares/share_splitting_test.go`

PR #819, opened 2022-09-29
feat!: stop forcing reserved byte to zero

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/779

### DIFF
--- a/pkg/shares/share_splitting_test.go
+++ b/pkg/shares/share_splitting_test.go
@@ -64,7 +64,7 @@ func TestSplitTxs(t *testing.T) {
 				append([]uint8{
 					0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, // namespace id
 					0x0, // info byte
-					0x0, // reserved byte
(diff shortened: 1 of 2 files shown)


pkg/shares/share_splitting_test.go
```diff
@@ -64,7 +64,7 @@ func TestSplitTxs(t *testing.T) {
 				append([]uint8{
 					0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, // namespace id
 					0x0, // info byte
-					0x0, // reserved byte
+					0xb, // BUG: reserved byte should be zero
```
evan-forbes, line 67: [question]

for my own understanding,  this change to a different bug by removing the forced 0, and then we will fix this in an upcoming PR for #817

evan-forbes (approved):

pkg/shares/share_splitting_test.go
rootulp: Yep, exactly!

PR #821, opened 2022-09-30
fix: reserved byte for existing test cases

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/817

### DIFF
--- a/pkg/shares/split_compact_shares.go
+++ b/pkg/shares/split_compact_shares.go
@@ -30,9 +30,12 @@ func NewCompactShareSplitter(ns namespace.ID, version uint8) *CompactShareSplitt
 		panic(err)
 	}
 	placeholderDataLength := make([]byte, appconsts.FirstCompactShareDataLengthBytes)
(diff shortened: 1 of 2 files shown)


pkg/shares/split_compact_shares.go
```diff
@@ -183,6 +167,40 @@ func (css *CompactShareSplitter) writeDataLengthVarintToFirstShare(dataLengthVar
 	css.shares[0] = newFirstShare
 }
 
+// maybeWriteReservedByteToPendingShare will be a no-op if the reserved byte has
+// already been populated. If the reserved byte is empty, it will write the
+// location of the next unit of data to the reserved byte.
+func (css *CompactShareSplitter) maybeWriteReservedByteToPendingShare() {
```
evan-forbes, line 173: admittedly the naming for this maybe weird, but I also don't have any better options. hopefully whenever we have more time to refactor this we can and just have a method to write the reserved byte

evan-forbes (approved):
nice, this unblocks #802

Issue #721, opened 2022-09-16
Remove type `NamespacedShares` from pkg/shares


rootulp:
[comment] tbh, we should get rid of this struct as its pointless middleware. This is remnant of the first implementation done two years ago

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/716#discussion_r973181450_

rootulp:
The current implementation:

```golang
// Share contains the raw share data (including namespace ID).
type Share []byte

// NamespacedShare extends a Share with the corresponding namespace.
type NamespacedShare struct {
	Share
	ID namespace.ID
}

func (n NamespacedShare) NamespaceID() namespace.ID {
	return n.ID
}

func (n NamespacedShare) Data() []byte {
	return n.Share
}

// NamespacedShares is just a list of NamespacedShare elements.
// It can be used to extract the raw shares.
type NamespacedShares []NamespacedShare
```

**Pros**
- Provides getters for `NamespaceID` and `Data` so that the struct type doesn't leak implementation details (i.e. # of bytes a namespace ID occupies) to clients
- It's possible to provide docs on this named type

**Cons**
- Fighting with the type system to convert `[][]byte` -> to `[]NamespacedShare`

**Proposal**
- We may be able to move the getters to the `Share` type like:

```golang
func (s Share) NamespaceID() namespace.ID {
	return namespace.ID(s[:8])
}
```

PR #822, opened 2022-09-30
chore!: remove `NamespacedShares`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/721

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -57,13 +57,9 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 	if err != nil {
 		panic(err)
 	}
-	rawShares := make([][]byte, len(dataSquare))
-	for i, share := range dataSquare {
-		rawShares[i] = []byte(share)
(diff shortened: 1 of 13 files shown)


evan-forbes (approved):
nice, in the future, we might want to consider implementing the `Share` type as an alias, just like celestia-node does. Ideally, they should be able to use our implementation or we use a common implementation in a different repo (as discussed in the comments there). This might also be related to #802, in that that API will have to not include celestia-core or celestia-app in the dependency tree so that optimint can use it.

https://github.com/celestiaorg/celestia-node/blob/6455b1720aecc12f1bed0f823b30b76b04de24d1/ipld/share.go#L27-L30

rootulp:
Thanks for sharing this context! I don't yet understand the benefit to using a type alias rather than a type definition. I do think it could be useful if the `Share` type were common amongst all these repos so if not all Celestia repos can depend on celestia-app, then it seems reasonable to extract to a standalone package.

Ref:
- https://yourbasic.org/golang/type-alias/

PR #823, opened 2022-09-30
chore: resolve local lint issues

rootulp:
Resolve the following issues observed on `make lint`

```
--> Running linter
pkg/inclusion/paths.go:49: File is not `gofumpt`-ed (gofumpt)
// The following nolint can be removed after this function is used.
pkg/inclusion/paths.go:68: File is not `gofumpt`-ed (gofumpt)
//               / \
//              /   \
pkg/inclusion/paths.go:71: File is not `gofumpt`-ed (gofumpt)
//            /\     /\
x/payment/keeper/keeper.go:31: File is not `gofumpt`-ed (gofumpt)
//  MsgPayForData moves a user's coins to the module address and burns them.
app/export.go:49: File is not `gofumpt`-ed (gofumpt)
//      in favour of export at a block height
make: *** [lint] Error 1
```

These lines haven't been modified recently. It's not clear to me why these issues haven't been flagged by lint on CI.

### DIFF
--- a/app/export.go
+++ b/app/export.go
@@ -44,9 +44,9 @@ func (app *App) ExportAppStateAndValidators(
 	}, err
 }
 
-// prepare for fresh start at zero height
-// NOTE zero height genesis is a temporary feature which will be deprecated
-//      in favour of export at a block height
+// prepForZeroHeightGenesis preps for fresh start at zero height. Zero height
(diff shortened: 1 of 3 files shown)


evan-forbes (approved):
:shrug:

PR #824, opened 2022-09-30
feat: Implement an in-process test core node with celestia-app and custom state 

### LINKED ISSUE #770: Export a public testutil function for use in celestia-node swamp tests
## Problem

celestia-node currently uses a kvstore application in swamp tests. The kvstore doesn't adhere to assumptions of celestia-app (i.e. `squareSize` is block data is a power of two). It would be preferable if celestia-node swamp tests used a celestia-app application but it is cumbersome for them to use [SetupTestAppWithGenesisValSet](https://github.com/celestiaorg/celestia-app/blob/d53f66b67946d20af5a24487c2f79da4ebb1b171/testutil/test_app.go#L71) because the test Tendermint node will hit

```error
Did you reset Tendermint without resetting your application's data? [recovered]
	panic: state.AppHash does not match AppHash after replay. Got
0A98F8A6D292AD8D517FD57F77617DC27ECC7637BD82973795A7A473917F00A9, expected .
```

## Proposal

Export a public testutil function from celestia-app that can be used in celestia-node swamp tests. The implementation of the testutil function will involve spinning up a Tendermint node with a genesis state that celestia-app and the Tendermint node agree upon.

evan-forbes:
This PR is based of the [network](https://github.com/cosmos/cosmos-sdk/tree/main/testutil/network) tests in the cosmos-sdk. Those tests are focused on running multiple nodes, while this test only runs a single node validator. This node also has the full access to change the consensus parameters and the full tendermint config. This is needed to to have block times shorter than 1 second. There is also discussion to modify the sdk code to only be a mock, where we need to run a full in process node to include any changes to tendermint and to not have to apply those same changes to the new upcoming mock node in the sdk.

While this does have extra features and a built in API, it does have a lot of duplicate features to our own wrapper around the linked above network tests. Eventually we should replace the usage of that code in the integrations tests with this code. 

closes #770


testutil/testnode/node_init.go
```diff
@@ -0,0 +1,254 @@
...
+	bankGenState.Balances = append(bankGenState.Balances, genBalances...)
+	state[banktypes.ModuleName] = codec.MustMarshalJSON(&bankGenState)
+
+	appGenStateJSON, err := json.MarshalIndent(state, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	cparams := types.DefaultConsensusParams()
+	cparams.Block.TimeIotaMs = 1
```
SweeXordious, line 132: [Non Blocking][Question]
Doesn't it make sense to take the consensus params as a parameter for more flexibility?
Especially the block time, I guess it makes sense to give the user the abilitiy to change it.
Then, we can for example check if the provided consensus params are null, we initialize using the default ones.

testutil/testnode/node_interaction_api.go
```diff
@@ -0,0 +1,143 @@
...
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	timeout := time.NewTimer(t)
+	defer timeout.Stop()
+
+	var latestHeight int64
+	for {
+		select {
+		case <-timeout.C:
```
SweeXordious, line 48: [Non Blocking][Suggestion]
Doesn't it make sense also to check if the root context was cancelled or done?

testutil/testnode/node_interaction_api.go
```diff
@@ -0,0 +1,143 @@
...
+	return status.SyncInfo.LatestBlockHeight, nil
+}
+
+// WaitForHeightWithTimeout is the same as WaitForHeight except the caller can
+// provide a custom timeout.
+func (c *Context) WaitForHeightWithTimeout(h int64, t time.Duration) (int64, error) {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	timeout := time.NewTimer(t)
```
SweeXordious, line 42: [Non Blocking][Suggestion]
Doesn't it make sense to use the context with timeout? to also propagate timeouts if something happens?

testutil/testnode/node_interaction_api.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	signer.SetAccountNumber(acc)
+	signer.SetSequence(seq)
+
+	// create a random msg per row
+	pfd, err := payment.BuildPayForData(
+		context.TODO(),
```
SweeXordious, line 113: [Non Blocking][Question]
Why use the TODO context when we have a context passed `func (c *Context) PostData...`?

testutil/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/celestiaorg/nmt/namespace"
+	"github.com/cosmos/cosmos-sdk/client"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+	rpctypes "github.com/tendermint/tendermint/rpc/core/types"
+)
+
+func RandomValidNamespace() namespace.ID {
+	for {
+		s := tmrand.Bytes(8)
+		if bytes.Compare(s, appconsts.MaxReservedNamespace) > 0 {
```
SweeXordious, line 18: [Non Blocking][Question]
This one is used to create a PFD. Shouldn't we also check if the namespace is not the ns of evidence or?

testutil/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	node, err := clientCtx.GetNode()
+	if err != nil {
+		return nil, err
+	}
+
+	return node.Tx(context.Background(), hash, false)
```
SweeXordious, line 35: [Question][Non Blocking]
Why using the background context when we have the `clientCtx`?

SweeXordious (commented):
utAck. Awesome stuff :rocket: Would be really helpful for QGB tests :fire:

testutil/testnode/full_node.go
```diff
@@ -0,0 +1,124 @@
...
+
+// New creates a ready to use tendermint node that operates a single validator
+// celestia-app network. The passed account names in fundedAccounts are used to
+// generate new private keys, which are included as funded accounts in the
+// genesis file. These keys are stored in the keyring that is returned in the
+// client.Context. NOTE: the forced delay between blocks, TimeIotaMs in the
+// consensus parameters, is set to the lowest possible value (1ms).
+//
+// note: the passed application config is currently unused atm, but we plan to
+// add support.
```
rootulp, line 33: [question] which parameter corresponds to the application config?

If present, proposal to explicitly include the parameter name in the note (because I don't think it's `tmCfg`).
If no longer present, proposal to remove this note.

testutil/testnode/rpc_client.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return Context{}, func() {}, err
+	}
+
+	cctx.Context = cctx.WithGRPCClient(conn)
+
+	return cctx, grpcSrv.Stop, nil
+}
+
+// DefaultAppConfig wraps the default config described in the ser
```
rootulp, line 65:
```suggestion
// DefaultAppConfig wraps the default config described in the server
```

testutil/utils.go
rootulp: `MaxReservedNamespace` is greater than the  namespace used for transactions, intermediate state roots, or evidence so this conditional will evaluate to `false` if `s` matches any of those namespaces.

In other words, the returned `s` should only be valid message namespaces, `TailPaddingNamespaceID` or `ParitySharesNamespaceID`

https://github.com/celestiaorg/celestia-app/blob/main/pkg/appconsts/appconsts.go#L90-L96

rootulp (approved):

testutil/utils.go
SweeXordious: Sorry yes, I meant tail padding or parity. Is that alright?

evan-forbes: nice catch! 6f50ff6dfbf028c7170d4b2f4eef962d16f3b732

testutil/testnode/node_interaction_api.go
evan-forbes: nice catch! wrote this when we didn't yet have contexts 95529af7aae854787e47f8a541a858d22fe8943a

testutil/testnode/node_interaction_api.go
evan-forbes: :+1: 96cec2a433a259762810e6c0427f8034bc7810ba

testutil/testnode/node_init.go
evan-forbes: good idea 903bc0b96e40d319aff87ebf044dcecf164a4586

testutil/testnode/full_node.go
evan-forbes: good catch 14aca810e103b3fbf803f0acadecbc3f234a999c

testutil/utils.go
evan-forbes: the clientCtx doesn't actually have a context.Context in it, and I didn't want to force the testnode package as a dependency. This is good point tho

evan-forbes:
had to skip the network test while the race detector is on, exactly like what we do for the other larger network test. https://github.com/celestiaorg/celestia-app/pull/824/commits/a376539ccadbb7cf9002cdcadc08c6f41754d03b

testutil/utils.go
```diff
@@ -14,10 +14,14 @@ import (
...
-		if bytes.Compare(s, appconsts.MaxReservedNamespace) > 0 {
-			return s
+		ns := tmrand.Bytes(8)
+		isReservedNS := bytes.Compare(ns, appconsts.MaxReservedNamespace) <= 0
+		isParityNS := bytes.Equal(ns, appconsts.ParitySharesNamespaceID)
+		isTailPaddingNS := bytes.Equal(ns, appconsts.TailPaddingNamespaceID)
+		if isReservedNS || isParityNS || isTailPaddingNS {
+			continue
 		}
+		return ns
```
rootulp, line 24: [question] Nice! There are multiple implementations of this function that may benefit from similar parity / tail padding namespace checks:

1. https://github.com/rootulp/celestia-app/blob/2f26c385831048808d99f4626aa3569df18c187c/app/test_util.go#L178-L185
2. https://github.com/rootulp/celestia-app/blob/2f26c385831048808d99f4626aa3569df18c187c/testutil/payment/testutil.go#L134-L141
3. https://github.com/rootulp/celestia-app/blob/2f26c385831048808d99f4626aa3569df18c187c/app/test/block_size_test.go#L299-L306
4. https://github.com/rootulp/celestia-app/blob/2f26c385831048808d99f4626aa3569df18c187c/pkg/prove/proof_test.go#L249-L257

It's not clear to me if these should be refactored into one function definition. Were they intentionally duplicated to avoid cyclic imports / importing across package boundaries?

rootulp (approved):

Issue #723, opened 2022-09-16
Define a `ParseShares` API


rootulp:
## Context

celestia-node currently parses shares depending on the type of share (see [here](https://github.com/celestiaorg/celestia-node/blob/d051a94bce1cf8a6b72be03a9484abdb9a6983a3/service/share/share_test.go#L264)). Universal share encoding should enable us to expose one API where any type of share can be parsed. As a result, it may be possible to stop exporting 
https://github.com/celestiaorg/celestia-app/blob/f67bb6ba30e1d4886c5248efef12df5a7b6d5173/pkg/shares/share_merging.go#L125

https://github.com/celestiaorg/celestia-app/blob/f67bb6ba30e1d4886c5248efef12df5a7b6d5173/pkg/shares/share_merging.go#L78

https://github.com/celestiaorg/celestia-app/blob/f67bb6ba30e1d4886c5248efef12df5a7b6d5173/pkg/shares/share_merging.go#L95

evan-forbes:
I still think we will need to export these functions, but also expose a `ParseShares` api. Perhaps something like this.

```go
type ShareSet [][]byte

func ParseShares(shares [][]byte) []ShareSet
```

then we can parse depending on namespace

rootulp:
1. Does this proposal imply an additional mechanism to convert from `[]ShareSet` => `coretypes.Messages`, `coretypes.Txs`, or `coretypes.EvidenceData`? 
2. Is the type `ShareSet` meant to describe that shares in this set are of a particular type. In other words, all share in this ShareSet are `coretypes.Txs`?

One alternative proposal that comes to mind:

```golang
func ParseShares(shares [][]byte) (coretypes.Messages, coretypes.Txs, coretypes.EvidenceData, error)
```

However this has the disadvantage that the return signature must be updated if/when we introduce more types.

evan-forbes:
ParseShares should perform only the universal shares parsing logic, which means splitting up some set of data using only the namespace and the message start byte. 

Now we have some universal and purposefully ambiguous (potentially called a `ShareSet`) data that we can do something with. If we're celestia-node, and we just want to serve some complete set of shares, then we can do that, if we actually want to parse the shares into a data structure, then we can do that using the namespace and the version portion of the info byte.

evan-forbes:
If we are wanting to parse the entire square into block data, then we already have the `Merge` function to do that.

rootulp:
[...]

PR #828, opened 2022-10-03
feat: new `ParseShare` API

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/723
Opens https://github.com/celestiaorg/celestia-app/issues/839

### DIFF
--- a/pkg/shares/share_merging.go
+++ b/pkg/shares/share_merging.go
@@ -163,8 +163,7 @@ func ParseShares(rawShares [][]byte) (result []ShareSequence, err error) {
 				Shares:      []Share{share},
 				NamespaceID: share.NamespaceID(),
 			}
-		}
-		if !infoByte.IsMessageStart() {
+		} else {
 			if !bytes.Equal(currentSequence.NamespaceID, share.NamespaceID()) {
 				return result, fmt.Errorf("share sequence %v has inconsistent namespace IDs with share %v", currentSequence, share)
 			}
--- a/pkg/shares/share_merging.go
+++ b/pkg/shares/share_merging.go
@@ -2,8 +2,10 @@ package shares
 
 import (
 	"bytes"
+	"fmt"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/nmt/namespace"
 	"github.com/celestiaorg/rsmt2d"
 	"github.com/gogo/protobuf/proto"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
@@ -131,4 +133,44 @@ func ParseMsgs(shares [][]byte) (coretypes.Messages, error) {
 	return coretypes.Messages{
 		MessagesList: msgList,
 	}, nil
+}
+
+// ShareSequence represents a contiguous sequence of shares that are part of the
+// same namespace and message. For compact shares, one share sequence exists per
+// reserved namespace. For sparse shares, one share sequence exists per message.
+type ShareSequence struct {
+	NamespaceID namespace.ID
+	Shares      []Share
+}
+
+func ParseShares(rawShares [][]byte) (result []ShareSequence, err error) {
+	currentSequence := ShareSequence{}
+
+	for _, rawShare := range rawShares {
+		share, err := NewShare(rawShare)
+		if err != nil {
+			return result, err
+		}
+		infoByte, err := share.InfoByte()
+		if err != nil {
+			return result, err
+		}
+		if infoByte.IsMessageStart() {
+			if len(currentSequence.Shares) > 0 {
+				result = append(result, currentSequence)
+			}
+			currentSequence = ShareSequence{
+				Shares:      []Share{share},
+				NamespaceID: share.NamespaceID(),
+			}
+		}
+		if !infoByte.IsMessageStart() {
+			if !bytes.Equal(currentSequence.NamespaceID, share.NamespaceID()) {
+				return result, fmt.Errorf("share sequence %v has inconsistent namespace IDs with share %v", currentSequence, share)
+			}
+			currentSequence.Shares = append(currentSequence.Shares, share)
+		}
+	}
+
+	return result, nil
 }
--- a/pkg/shares/share_merging_test.go
+++ b/pkg/shares/share_merging_test.go
@@ -1,0 +1,146 @@
+package shares
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+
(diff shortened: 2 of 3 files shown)


pkg/shares/share_merging.go
```diff
@@ -132,3 +134,43 @@ func ParseMsgs(shares [][]byte) (coretypes.Messages, error) {
...
+			if len(currentSequence.Shares) > 0 {
+				result = append(result, currentSequence)
+			}
+			currentSequence = ShareSequence{
+				Shares:      []Share{share},
+				NamespaceID: share.NamespaceID(),
+			}
+		}
+		if !infoByte.IsMessageStart() {
+			if !bytes.Equal(currentSequence.NamespaceID, share.NamespaceID()) {
```
evan-forbes, line 168: [comment for other reviewers/posterity]
this isn't actually consensus breaking I don't think, even tho we weren't previously checking for this, because is order for this to be hit, the incorrect info byte has to be used which won't happen due to validators inserting it themselves and comparing the data root.

evan-forbes (commented):
[question]
should we also be checking that the length delimiter matches the sequence length?

pkg/shares/share_merging.go
evan-forbes: on second thought, once we start using this, it will be consensus breaking with a dishonest majority

rootulp:
> should we also be checking that the length delimiter matches the sequence length?

It's possible to add this check to the contents of `ParseShares()`. In order to do so, it must trim padding from the last share of the sequence to calculate the true total length. If this approach is recommended:

1. How do we differentiate padding bytes (`0`) from intentionally added zero bytes (i.e. a user included 3 trailing `0` bytes in their message)?
2. Should this API return an error if the true total data length doesn't match the total data length varint?

evan-forbes:
> In order to do so, it must trim padding from the last share of the sequence to calculate the true total length.

do we need to trim if we're only checking sequence length, not the exact length of bytes?

rootulp:
> do we need to trim if we're only checking sequence length, not the exact length of bytes?

We need to trim the last share of padding if we want to compare the total data length varint to the exact sequence length. We don't need to trim if we want this check to compare the total data length varint to an approximate sequence length (i.e. for message shares)

```golang
(len(shareSequence.Shares) - 1) * appconsts. SparseShareContentSize <= varint <= len(shareSequence.Shares) * appconsts. SparseShareContentSize
```

pkg/shares/share_merging.go
rootulp: > once we start using this

I thought `ParseShares` was for celestia-node consumption. Do we expect the state machine to invoke this API?  I didn't expect this API to be consensus breaking because I assumed the state machine would not call it.

evan-forbes: even if light clients are the only ones rejecting a block, then that is still consensus breaking (just like a change to fraud proofs)

evan-forbes: at least, I think its consensus breaking... :sweat_smile: 

imo, it makes sense to eventually use this code when we merge shares. It keeps things consistent and can possibly even inform other implementations on how to use universal shares.

pkg/shares/share_merging_test.go
```diff
@@ -0,0 +1,146 @@
...
+	evidenceShareContinuation := generateRawShare(appconsts.EvidenceNamespaceID, continuation)
+
+	messageOneStart := generateRawShare(messageOneNamespace, start)
+	messageOneContinuation := generateRawShare(messageOneNamespace, continuation)
+
+	messageTwoStart := generateRawShare(messageTwoNamespace, start)
+	messageTwoContinuation := generateRawShare(messageTwoNamespace, continuation)
+
+	invalidShare := generateRawShare(messageOneNamespace, start)
+	invalidShare = append(invalidShare, []byte{0}...)
```
evan-forbes, line 38: [blocking]
perhaps we could modify this test to use the share splitting api so that we know if we break this in the future?

pkg/shares/share_merging.go
```diff
@@ -132,3 +134,43 @@ func ParseMsgs(shares [][]byte) (coretypes.Messages, error) {
...
+		if infoByte.IsMessageStart() {
+			if len(currentSequence.Shares) > 0 {
+				result = append(result, currentSequence)
+			}
+			currentSequence = ShareSequence{
+				Shares:      []Share{share},
+				NamespaceID: share.NamespaceID(),
+			}
+		}
+		if !infoByte.IsMessageStart() {
```
evan-forbes, line 167: [nit][optional]
could we get rid of this second check or using an else?

pkg/shares/share_merging_test.go
```diff
@@ -0,0 +1,146 @@
...
+		})
+	}
+}
+
+func generateRawShare(namespace namespace.ID, isMessageStart bool) (rawShare []byte) {
+	infoByte, _ := NewInfoByte(appconsts.ShareVersion, isMessageStart)
+
+	rawShare = append(rawShare, namespace...)
+	rawShare = append(rawShare, byte(infoByte))
+	rawShare = append(rawShare, bytes.Repeat([]byte{0}, appconsts.ShareSize-len(rawShare))...)
```
evan-forbes, line 144: [optional/related to the above comment]
I know it shouldn't matter, but testing with data other than zero data seems like a good idea

evan-forbes (commented):
left one remaining change request for the testing we're using. 

While I still think we should check for the length of the sequence `(len(shareSequence.Shares))`, that would require a large enough change to the existing code that doing this in a different PR makes sense

pkg/shares/share_merging.go
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/828/commits/16f8d623230f56a396ae00dd02bae9b16f50f0b5

pkg/shares/share_merging_test.go
rootulp: Good idea, resolved in https://github.com/celestiaorg/celestia-app/pull/828/commits/2dad6e56ceb12e8de1423aa9cab2ad1fc52b5e1f

pkg/shares/share_merging_test.go
rootulp: Good idea. Resolved in https://github.com/celestiaorg/celestia-app/pull/828/commits/c184932f0926d119de14ce30bc2611d75f2b1b3e

evan-forbes (approved):
dope! thanks! preapprooovvviinnngggggg

Issue #834, opened 2022-10-04
Create ADR for blocksize independent commitments


nashqueue:
### Summary

Currently, commitments in Celestia are dependent on the size of the block. It would be nice to have commitments that are independent of size. 

### Problem Definition

See ADR #835 

### Proposal

See ADR #835 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #827, opened 2022-10-03
CreateCommitment should follow Non-Interactive Default Rules as intended


nashqueue:
### Summary of Bug

This [check](https://github.com/celestiaorg/celestia-app/blob/0c81704939cd743937aac2859f3cb5ae6368f174/x/payment/types/payfordata.go#L132-L139) accepts messages that not follow the non-interactive default rules. 
```go
	// if the number of shares is larger than that in the square, throw an error
	// note, we use (squareSize*squareSize)-1 here because at least a single
	// share will be reserved for the transaction paying for the message,
	// therefore the max number of shares a message can be is number of shares
	// in square - 1.
if uint64(len(shares)) > (squareSize*squareSize)-1 {
		return nil, fmt.Errorf("message size exceeds max shares for square size %d: max %d taken %d", squareSize, (squareSize*squareSize)-1, len(shares))
	}
```
It accepts such a block: 
![image](https://user-images.githubusercontent.com/99758629/193522146-ecc03a60-75dd-4c5c-bf0b-2cfc13cd9304.png)

This should not happend following the non-interactive default [rules](https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules): 
1. Messages that span multiple rows must begin at the start of a row (this can occur if a message is longer than `k` shares _or_ if the block producer decides to start a message partway through a row and it cannot fit).

The maximum message following this rule should looks like this:
![image](https://user-images.githubusercontent.com/99758629/193523283-6d5970b2-2c9d-4ca1-bbb9-8a386dfb7d78.png)

The code should be changed like this: 
```go
	// if the number of shares is larger than that in the square, throw an error
	// note, we use (squareSize*(squareSize-1) here because at least a single
	// share will be reserved for the transaction paying for the message,
	// therefore following the non-interactive default rules a message that 
        // spans multiple rows always starts in the new row. 
if uint64(len(shares)) > (squareSize*(squareSize-1)) {
		return nil, fmt.Errorf("message size exceeds max shares for square size %d: max %d taken %d", squareSize, (squareSize*(squareSize-1), len(shares))
	}
```

### Version

v0.6.0 8302c9fea54cff53d0551cf14188491547a99405

### Steps to Reproduce

Create a commit with message length 14 and a block size 4. It should throw an error with everything larger than 12. 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
[...]

PR #835, opened 2022-10-04
ADR square size independent message commitments

nashqueue:
Resolvses #834

[rendered](https://github.com/celestiaorg/celestia-app/blob/fbfbf111bcaa056e53b0bc54d327587dee11a945/docs/architecture/adr-008-blocksize-independent-commitment.md)


docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+
+![Old Commitment](./assets/size-dependent-commitment.png)
+
+To have a commitment independent of the block size, you would need to create a merkle tree over subroots that are included in every possible block size.
+If we agree on producing the commitment in the `minimalBlocksize` the resulting commitment will be the same no matter the block size. *This is because all the subroots in the `minimalBlocksize` are also included in bigger block sizes*.
+You can see this in the updated version of the diagram above. Now the merkle root will stay the same. This property results from the block width and length being a power of 2.  
+
+![New Commitment](./assets/size-independent-commitment.png)
+
+You can see in the Diagramm below that no matter how the message is split up into rows, the hashes of the subroots stay the same.
```
SweeXordious, line 23:
```suggestion
You can see in the Diagram below that no matter how the message is split up into rows, the hashes of the subroots stay the same.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+  ![Interactive Commitment 2](./assets/complexity.png)
+
+2. With more subroots the amount of merkle proofs will increase. With deeper subroots the size of the merkle proofs will increase. So instead of having a merkle proof from the `DataRoot` to the `originalSubRoot` you will need the merkle proof from `DataRoot` to `2^k` amount of `miniSubRoots` with `k` being the height difference of `originalSubRoot` and `miniSubRoots`. You can optimize this by having one merkle proof from `DataRoot` to `originalSubRoot` and then `k` merkle proofs from `originalSubRoot` to the `miniSubRoots`. Because the merkle proof is created after the block is published, we know the block size and, therefore, if this `originalSubRoot` exists or not, like in smaller block sizes.
+
+### Positive
+
+1. A Rollup can include the commitment in the block header *before* posting on Celestia because it is size-independent and does not have to wait for Celestia to confirm the block size. In general, the rollup needs access to this commitment in some form to verify a message inclusion proof guaranteeing data availability, which Rollmint currently does not have access to.
```
SweeXordious, line 78:
```suggestion
1. A Rollup can include the commitment in the block header *before* posting to Celestia because it is size-independent and does not have to wait for Celestia to confirm the block size. In general, the rollup needs access to this commitment in some form to verify a message inclusion proof guaranteeing data availability, which Rollmint currently does not have access to.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+3. Here is one scheme on how a Rollup might use this new commitment in the block header. Let's assume a Rollup that looks like this:  
+  BH1 <-- BH2 <-- BH3 <-- BH4  
+  The Messages that are submitted to Celestia could look like this:  
+  Message 1: B1  
+  The Commitment of B1 is saved into BH1.  
+  Message 2: (BH1+B2)  
+  The Commitment of BH1+B2 is saved into BH2.  
+  Message 3: (BH2+B3)  
+  The Commitment of (BH2+B3) is saved into BH3, and so on.  
+4. Verifying a message inclusion proof could be done with merkle proofs of the subroots to the `DataRoot` and then recalculating the commitment and comparing to what's in the rollup block header. If you trust the validator set of Celestia/Rollup (aka Rollmint Light Client) it could be as simple as submitting a proof over the pfd transaction that included the message and then check if the commitment is the same as in the pfd transaction.
```
SweeXordious, line 89:
```suggestion
4. Verifying a message inclusion proof could be done with merkle proofs of the subroots to the `DataRoot` and then recalculating the commitment and comparing to what's in the rollup block header. If you trust the validator set of Celestia/Rollup (aka Rollmint Light Client) it could be as simple as submitting a proof over the PFD transaction that included the message and then check if the commitment is the same as in the PFD transaction.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+5. So far, a full node in Rollmint downloads new blocks from the DA Layer after each Celestia block, coupled tightly for syncing. With this approach, we can send blocks over the p2p Layer giving a soft-commit to full nodes. Then, they would receive the hard-commit after verifying a message inclusion proof without the need to download the blocks anymore. **P2P Blocksync**
+    1. P2P Blocksync allows a Rollmint full node to run a Celestia light node and not a Celestia full node.
+    2. It allows the Rollup node to continue running after Celestia halts, relying on soft commits with no data availability.
+    3. It gives the Rollup the option to run asynchronously to Celestia because you don't have to wait for new Celestia blocks/commitments of the messages.
+6. Combining P2P Blocksync and the scheme in 3. we could have multiple rollup blocks in one Celestia block. It could look like this:
```
SweeXordious, line 94:
```suggestion
6. Combining P2P Blocksync and the scheme in 3, we could have multiple rollup blocks in one Celestia block. It could look like this:  
```

SweeXordious (commented):
Nits

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+![New Commitment](./assets/size-independent-commitment.png)
+
+You can see in the Diagramm below that no matter how the message is split up into rows, the hashes of the subroots stay the same.
+
+![Row Size Comparison](./assets/row-size-comparison.png)
+
+This scheme works for interactive commitments as well as long as the index of when a message starts is the same in every block size.
+For example, in the following diagram, you see the message starting at the second share and still having the same commitment. I marked the skipped blocks pink to show that both messages have the same starting index.
+
+![Interactive Commitment](./assets/interactive-commitment.png)
```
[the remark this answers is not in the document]
nashqueue: If we use the first 2 elements of the hash then we would create a subroot that does not connect to the `DataRoot`of Celestia. We cannot use or that else we would need the whole message to verify the commitment defeating the purpose.

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+
+If the message starting point index is larger than the row of `minimalBlocksize`, you take the index mod `minimalBlocksize`. The `minimalBlocksize` is shown in green and the index in pink. In both cases, the commitment stays the same.
```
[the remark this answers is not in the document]
nashqueue: The amount of commitments you have to calculate depends on the amount of messages in a block. This does not change in this proposal. The proposal just shows how to calculate the commitments it does not change the amount.

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Negative
+
+1. The amount of subtroots per commitment increases from O(log(n)) to O(sqrt(n)) while n is the number of shares of a message. The worst case for the most amount of shares is depicted in the diagram below - an entire block missing one share.
```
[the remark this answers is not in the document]
rootulp: [question] why can't the last share of the last row be occupied by the message? In other words, for a square size of 8 (same as diagram) why is the maximum message length `(8 * 7) - 1 = 55` instead of `(8 * 7) = 56`?

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+## Status
+
+Proposed
+
+## Consequences
+
+### Negative
+
+1. The amount of subtroots per commitment increases from O(log(n)) to O(sqrt(n)) while n is the number of shares of a message. The worst case for the most amount of shares is depicted in the diagram below - an entire block missing one share.  
+  ![Interactive Commitment 2](./assets/complexity.png)
```
rootulp, line 72: [question] in this diagram, I think the y-axis `O(sqrt(n))` is the number of subtree roots. What is the x-axis `O(log(sqrt(n))`?

rootulp (commented):
Only non-blocking questions for my understanding. This is a really well written ADR, great work!

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,98 @@
...
+In [`CreateCommit`](https://github.com/celestiaorg/celestia-app/blob/0c81704939cd743937aac2859f3cb5ae6368f174/x/payment/types/payfordata.go#L112-166), we calculate the commitment of the subroots with [`HashFromByteSlices`](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/crypto/merkle/tree.go#L7-L21). This resulting merkleroot is `Com`.
+In a square of 8x8, it results in the following subtree roots: `H1 C D`
+[`HashFromByteSlices`](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/crypto/merkle/tree.go#L7-L21) results now in the merkleroot `Com'`, which is not equal to `Com`
+
+![Old Commitment](./assets/size-dependent-commitment.png)
+
+To have a commitment independent of the block size, you would need to create a merkle tree over subroots that are included in every possible block size.
+If we agree on producing the commitment in the `minimalBlocksize` the resulting commitment will be the same no matter the block size. *This is because all the subroots in the `minimalBlocksize` are also included in bigger block sizes*.
```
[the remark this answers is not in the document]
nashqueue: Second Question: Close, as you can see in the implementation at the bottom it still needs to fit regarding the non-interactive-default. A message can only start in the second row. I mentioned that in #827 with additional visualizations.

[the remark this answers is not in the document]
nashqueue: First Question: Maybe "We agree on producing the commitment as if the message was put into the `minimalBlocksize` . Therefore the subtree roots will be the same and  the resulting commitment as well no matter the blocksize." explains it better.

docs/architecture/adr-008-blocksize-independent-commitment.md
nashqueue: The x-axis is also the number of subtree roots. Here we go after the mountainRange Merkle Tree. So in this case the complexity is logarithmic to the size of the row (sqrt(n)). In this example it would be a maximum of 3 subtree roots.

docs/architecture/adr-008-blocksize-independent-commitment.md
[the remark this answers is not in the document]
rootulp: Oh thanks for clarifying! 
```suggestion
1. The amount of subtree roots per commitment increases from O(log(n)) to O(sqrt(n)) while n is the number of shares of a message. The worst case for number of subtree roots is depicted in the diagram below - an entire block missing one share.  
```

docs/architecture/adr-008-blocksize-independent-commitment.md
rootulp: Sorry I don't follow. What does the Merkle mountain range look like for this example and why are there only 3 subtree roots in it?

docs/architecture/adr-008-blocksize-independent-commitment.md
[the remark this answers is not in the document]
SweeXordious: got you, thanks a lot

docs/architecture/adr-008-blocksize-independent-commitment.md
nashqueue: I added the diagram above as they are the same example. The 3 subtree roots are A B C in this case.

MSevey (commented):
I think this is very well written and the pictures are a great add :-) 

Not sure I have enough context yet to really approve this yet, so will wait on @liamsi and/or @adlerjohn

MSevey (approved):

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+3. Here is one scheme on how a Rollup might use this new commitment in the block header. Let's assume a Rollup that looks like this:  
+  BH1 <-- BH2 <-- BH3 <-- BH4  
+  The Messages that are submitted to Celestia could look like this:  
+  Message 1: B1  
+  The Commitment of B1 is saved into BH1.  
+  Message 2: (BH1+B2)  
+  The Commitment of BH1+B2 is saved into BH2.  
+  Message 3: (BH2+B3)  
+  The Commitment of (BH2+B3) is saved into BH3, and so on.  
+4. Verifying a message inclusion proof could be done with merkle proofs of the subtree roots to the `DataRoot`, recalculating the commitment, and comparing to what's in the rollup block header. If you trust the validator set of Celestia/Rollup (aka Rollmint Light Client) it could be as simple as submitting a proof over the PFD transaction that included the message and then checking if the commitment is the same as in the PFD transaction.
```
evan-forbes, line 91: I think you mentioned why this wasn't the case before but I forgot... Why can't this be done with the current approach? Don't we have to wait for it to be included in a block before we can create a proof for it?

> If you trust the validator set of Celestia/Rollup (aka Rollmint Light Client)

side note, but just to document that while this is currently the case, this will not be after state fraud proofs.

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+        return blocksize
+    } else {
+        return blocksize << 1
+    }
+}
+```
+
+## Status
+
+Proposed
```
evan-forbes, line 65: imo, making a decision on the status of this should not be a blocker for merging. We should merge it regardless.

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
+# ADR 008: Blocksize independent Commitments for Messages
+
+## Changelog
+
+- 03.08.2022: Initial Draft
+
+## Context
+
+Currently, commitments in Celestia are dependent on the size of the block. The following diagram visualizes how this can happen. The visualizations in this ADR are without parity shares/erasure encoding.
```
evan-forbes, line 9: just to note that removing the neccessity to malleate transactions depending on square size is a massive amount of work, considering that this is a major responsibility of the application. That's **not** to say that we shouldn't!! this is purely a comment

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+![Interactive Commitment](./assets/interactive-commitment.png)
+
+If the message starting point index is larger than the row of `minimalBlocksize`, you take the index mod `minimalBlocksize`. The `minimalBlocksize` is shown in green, and the skipped block is in pink. In both cases, the commitment stays the same.
+
+![Interactive Commitment 2](./assets/interactive-commitment2.png)
+
+## Alternative Approaches
+
+TODO
```
evan-forbes, line 38: can we link ADR007 from rollmint here?

evan-forbes (commented):
Nice! as commented below, regardless of the decision made for pursueing an implementation or not, I think we should merge this.

Left a few questions and comments. Thanks for writing this up so well, and including a lot of helpful diagrams. We should definitely continue thinking about this idea, such as even increasing the minimum square size to something quite large, which would help considerably. Since light clients don't have to sample rows that are completely tail padding, increasing the minimum square size could have very few downsides.

After very we get consistently very large blocks, most rollups will not take up an entire row\. In which case they will only ever have a single commitment anyway.

I think we should also note in this somewhere that this doesn't actually have to be done by the celestia-app logic, the rollup can do this themselves if they only ever want a single commitment.

What are the benefits from having celestia-app perform this logic rather than having rollups choose to do this?

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+  Message 1: B1  
+  The Commitment of B1 is saved into BH1.  
+  Message 2: (BH1+B2)  
+  The Commitment of BH1+B2 is saved into BH2.  
+  Message 3: (BH2+B3)  
+  The Commitment of (BH2+B3) is saved into BH3, and so on.  
+4. Verifying a message inclusion proof could be done with merkle proofs of the subtree roots to the `DataRoot`, recalculating the commitment, and comparing to what's in the rollup block header. If you trust the validator set of Celestia/Rollup (aka Rollmint Light Client) it could be as simple as submitting a proof over the PFD transaction that included the message and then checking if the commitment is the same as in the PFD transaction.
+5. So far, a full node in Rollmint downloads new blocks from the DA Layer after each Celestia block, coupled tightly for syncing. With this approach, we can send blocks over the p2p Layer giving a soft-commit to full nodes. Then, they would receive the hard-commit after verifying a message inclusion proof without the need to download the blocks anymore. **P2P Blocksync**
+    1. P2P Blocksync allows a Rollmint full node to run a Celestia light node and not a Celestia full node.
```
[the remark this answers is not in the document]
nashqueue: That would mean that the blockheader would include as many commitments as possible blocksizes. Which is possible and still deterministic. Validation would take longer, because the subroots to verify the commitments would change, as you need to verify all subroot combinations. You are right you dont _need_ a single commitment.

docs/architecture/adr-008-blocksize-independent-commitment.md
nashqueue: Yes. You need to wait for the rollup message to be included in celestia to create a proof that it was included.
 For the verification you need to have committed to a commitment before it was posted to celestia. If you include all possible commitments in the rollup blockheader than verifying the commitment via the pfd transaction is possible as well. Currently we don't save any commitment to the header of the rollmint block.

nashqueue: > side note, but just to document that while this is currently the case, this will not be after state fraud proofs.
wdym by that.

docs/architecture/adr-008-blocksize-independent-commitment.md
nashqueue: @musalbas

docs/architecture/adr-008-blocksize-independent-commitment.md
nashqueue: Should i add this as a neutral consequence, because it will take a lot of work but could simplify the design. ?

docs/architecture/adr-008-blocksize-independent-commitment.md
evan-forbes: just that rollmint light clients shouldn't have to trust the validator set after we have fraud proofs for everything

evan-forbes:
>  I think we should also note in this somewhere that this doesn't actually have to be done by the celestia-app logic, the rollup can do this themselves if they only ever want a single commitment.
What are the benefits from having celestia-app perform this logic rather than having rollups choose to do this?

I think this is the only remaining thing from my end

evan-forbes:
Its also important to note that while this ADR has great info how this new scheme would work, it does not included details on how this would actually be implemented. Per https://github.com/celestiaorg/celestia-app/pull/835#discussion_r996154742, removing malleation and non-interactive defaults in favor of this proposal would essentially be a re-write of the application

docs/architecture/adr-008-blocksize-independent-commitment.md
rootulp, line 9:
```suggestion
Currently, commitments in Celestia are dependent on the size of the square. The following diagram visualizes how this can happen. The visualizations in this ADR are without parity shares/erasure encoding.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+Currently, commitments in Celestia are dependent on the size of the block. The following diagram visualizes how this can happen. The visualizations in this ADR are without parity shares/erasure encoding.
+The yellow block in the diagram is a message of length 11 shares. A square of 4x4 results in the following subtree roots: `A B C D`.
+In [`CreateCommit`](https://github.com/celestiaorg/celestia-app/blob/0c81704939cd743937aac2859f3cb5ae6368f174/x/payment/types/payfordata.go#L112-166), we calculate the commitment of the subtree roots with [`HashFromByteSlices`](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/crypto/merkle/tree.go#L7-L21). This resulting merkleroot is `Com`.
+A square of 8x8 results in the following subtree roots: `H1 C D`
+[`HashFromByteSlices`](https://github.com/celestiaorg/celestia-core/blob/v0.34.x-celestia/crypto/merkle/tree.go#L7-L21) results now in the merkleroot `Com'`, which is not equal to `Com`.
+
+![Old Commitment](./assets/size-dependent-commitment.png)
+
+To have a commitment independent of the block size, you would need to create a merkle tree over subtree roots that are included in every possible block size.
```
rootulp, line 17:
```suggestion
To have a commitment independent of the square size, you would need to create a Merkle tree over subtree roots that are included in every possible square size.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+We agree on producing the commitment as if the message was put into the `minimalBlocksize`. Therefore the subtree roots will be the same and the resulting commitment as well, no matter the block size. *This is because all the subtree roots in the `minimalBlocksize` are also included in bigger block sizes*.
```
rootulp, line 18:
```suggestion
We agree on producing the commitment as if the message was put into the `minimumSquareSize`. Therefore the subtree roots will be the same and the resulting commitment as well, no matter the square size. *This is because all the subtree roots in the `minimumSquareSize` are also included in bigger square sizes*.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+You can see this in the updated version of the diagram above. Now the merkle root will stay the same. This property results from the block width and length being a power of 2.  
+
+![New Commitment](./assets/size-independent-commitment.png)
+
+You can see in the diagramm below that no matter how the message is split up into rows, the hashes of the subtree roots stay the same.
+
+![Row Size Comparison](./assets/row-size-comparison.png)
+
+This scheme works for interactive commitments as well as long as the index of when a message starts is the same in every block size. In other words, the leaf nodes and leaf node pairs are independent of the block size.
```
rootulp, line 27: [nit]

```suggestion
This scheme also works for interactive commitments as long as the index of when a message starts is the same in every block size. In other words, the leaf nodes and leaf node pairs are independent of the square size.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+For example, in the following diagram, you see the message starting at the second share and still having the same commitment. I marked the skipped blocks pink to show that both messages have the same starting index.
+
+![Interactive Commitment](./assets/interactive-commitment.png)
+
+If the message starting point index is larger than the row of `minimalBlocksize`, you take the index mod `minimalBlocksize`. The `minimalBlocksize` is shown in green, and the skipped block is in pink. In both cases, the commitment stays the same.
```
rootulp, line 32:
```suggestion
If the message starting point index is larger than the row of `minimumSquareSize`, you take the index mod `minimumSquareSize`. The `minimumSquareSize` is shown in green, and the skipped block is in pink. In both cases, the commitment stays the same.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
rootulp: [optional] to resolve @evan-forbes 's comment

```suggestion
## Alternative Approaches

- [rollmint/adr-007](https://github.com/celestiaorg/rollmint/blob/cb5c7440a8e879778e71097e254c3dd692c39d14/docs/lazy-adr/adr-007-header-commit-to-shares.md)
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+## Decision
+
+TODO
+
+## Detailed Design
+
+To implement this decision, you need to change [`CreateCommit`](https://github.com/celestiaorg/celestia-app/blob/0c81704939cd743937aac2859f3cb5ae6368f174/x/payment/types/payfordata.go#L112-166).
+In Detail, [`powerOf2MountainRange`](https://github.com/celestiaorg/celestia-app/blob/0c81704939cd743937aac2859f3cb5ae6368f174/x/payment/types/payfordata.go#L142) should take `minimalBlocksize` as an argument instead of `squareSize`.
+
+Following the non-interactive default rules, `minimalBlocksize` can be calculated like this:
```
rootulp, line 49:
```suggestion
Following the non-interactive default rules, `minimumSquareSize` can be calculated like this:
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+```go
+func minimalBlocksize(messageLength uint64) uint64 {
+    blocksize := NextHigherPowerOf2(messageLength)
+    //Check if message fits with non-interactive default rules
+    if messageLength <= (blocksize * (blocksize -1)) {
+        return blocksize
+    } else {
+        return blocksize << 1
+    }
+}
```
rootulp, line 60: [question][proposal]

If we proceed with implementing this ADR, do we simultaneously want to increase https://github.com/celestiaorg/celestia-app/blob/76a275fd110ffb8990deca626a843780fd117733/pkg/appconsts/appconsts.go#L55 

and make this function default to a minimum square size of `appconsts.MinSquareSize`? @evan-forbes mentioned a higher minimum square size is desirable to keep message inclusion proofs small but we probably want to evaluate the table below to determine a reasonable minimum square size

minSquareSize | message size (bytes) | size of message inclusion proof (bytes)
--- | --- | ---
1 | |
2 | |
4 | |
8 | |
16 | |
32 | | 
64 | | 
128 | |
256 | |

Since a message inclusion proof needs to include all subtree roots for a message, if we don't increase the minimum square size, then I think small messages will result in large message inclusion proofs (to be confirmed).

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+Proposed
+
+## Consequences
+
+### Negative
+
+1. The amount of subtree roots per commitment is O(sqrt(n)), while n is the number of message shares. The worst case for the number of subtree roots is depicted in the diagram below - an entire block missing one share.  
+  ![Interactive Commitment 2](./assets/complexity.png)  
+The worst case for the current implementation depends on the block size. If it is the worst block size, as in `minimalBlocksize`, it is O(sqrt(n)) as well. On the other hand, if the message is only in one row, then it is O(log(n)).
```
rootulp, line 73:
```suggestion
The worst case for the current implementation depends on the block size. If it is the worst block size, as in `minimumSquareSize`, it is O(sqrt(n)) as well. On the other hand, if the message is only in one row, then it is O(log(n)).
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+Therefore the height of the tree over the subtree roots is in this implementation O(log(sqrt(n))), while n is the number of message shares. In the current implementation, it varies from O(log(sqrt(n))) to O(log(log(n))) depending on the block size.
```
rootulp, line 74:
```suggestion
Therefore the height of the tree over the subtree roots is in this implementation O(log(sqrt(n))), while n is the number of message shares. In the current implementation, it varies from O(log(sqrt(n))) to O(log(log(n))) depending on the square size.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+2. With more subtree roots, the number of merkle proofs will increase. With deeper subtree roots, the size of the merkle proofs will increase. So instead of having a merkle proof from the `DataRoot` to the `originalSubRoot` you will need the merkle proof from `DataRoot` to `2^k` amount of `miniSubRoots` with `k` being the height difference of `originalSubRoot` and `miniSubRoots`. You can optimize this by having only **one** merkle proof from `DataRoot` to `originalSubRoot` and then calculate the `originalSubRoot` from the `k` `miniSubRoots`. Because the merkle proof is created after the block is published, we know the block size and, therefore, if this `originalSubRoot` exists or not.
```
rootulp, line 76:
```suggestion
2. With more subtree roots, the number of merkle proofs will increase. With deeper subtree roots, the size of the merkle proofs will increase. So instead of having a merkle proof from the `DataRoot` to the `originalSubRoot` you will need the merkle proof from `DataRoot` to `2^k` amount of `miniSubRoots` with `k` being the height difference of `originalSubRoot` and `miniSubRoots`. You can optimize this by having only **one** merkle proof from `DataRoot` to `originalSubRoot` and then calculate the `originalSubRoot` from the `k` `miniSubRoots`. Because the merkle proof is created after the block is published, we know the square size and, therefore, if this `originalSubRoot` exists or not.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Positive
+
+1. A Rollup can include the commitment in the block header *before* posting to Celestia because it is size-independent and does not have to wait for Celestia to confirm the block size. In general, the rollup needs access to this commitment in some form to verify a message inclusion proof guaranteeing data availability, which Rollmint currently does not have access to.
```
rootulp, line 80:
```suggestion
1. A Rollup can include the commitment in the block header *before* posting to Celestia because it is size-independent and does not have to wait for Celestia to confirm the square size. In general, the rollup needs access to this commitment in some form to verify a message inclusion proof guaranteeing data availability, which Rollmint currently does not have access to.
```

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+2. In turn, this would serve as an alternative to [ADR 007](https://github.com/celestiaorg/optimint/blob/main/docs/lazy-adr/adr-007-header-commit-to-shares.md)
```
rootulp, line 81: [optional] to disambiguate with adr-007 in this repo

```suggestion
2. In turn, this would serve as an alternative to [rollmint/adr-007](https://github.com/celestiaorg/optimint/blob/main/docs/lazy-adr/adr-007-header-commit-to-shares.md)
```

rootulp (commented):
Sorry for the pedantic proposal to rename block size to square size. Given we already use square size to describe the dimensions of the data square, I think this rename may make it slightly easier to discuss the proposal. Additionally, the [block](https://celestiaorg.github.io/celestia-specs/latest/specs/data_structures.html#block) includes information outside the data square.

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Alternative Approaches
+
+TODO
+
+## Decision
+
+TODO
+
+## Detailed Design
```
rootulp, line 44: [no change needed] we don't need to tackle in this PR but wanted to note this: we may consider including the `minimumSquareSize` in the block header so that we can increase this value in the future (potentially via governance proposal). 

If we don't include it in the block header, it would be difficult (perhaps impossible) to determine the `minimumSquareSize` for historical blocks.

docs/architecture/adr-008-blocksize-independent-commitment.md
rootulp: [optional] it may help future readers if this section states that the diagram below assumes `minimumSquareSize=4`

nashqueue:
> Instead of following the current non-interactive defaults, where we start each message at the next aligned power of two, we would start the message at the next index that is divisible by the minimum block size. This would likely require less padding for most square arrangements.

Should i make a diagram for that ?

docs/architecture/adr-008-blocksize-independent-commitment.md
nashqueue: The table could have number of subroots instead of size of message inclusion proof as this is directly proportional right? I don`t think small messages will result in large message inclusion proofs

docs/architecture/adr-008-blocksize-independent-commitment.md
```diff
@@ -0,0 +1,100 @@
...
+
+This scheme works for interactive commitments as well as long as the index of when a message starts is the same in every block size. In other words, the leaf nodes and leaf node pairs are independent of the block size.
+For example, in the following diagram, you see the message starting at the second share and still having the same commitment. I marked the skipped blocks pink to show that both messages have the same starting index.
+
+![Interactive Commitment](./assets/interactive-commitment.png)
+
+If the message starting point index is larger than the row of `minimalBlocksize`, you take the index mod `minimalBlocksize`. The `minimalBlocksize` is shown in green, and the skipped block is in pink. In both cases, the commitment stays the same.
+
+![Interactive Commitment 2](./assets/interactive-commitment2.png)
+
```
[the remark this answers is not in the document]
nashqueue: I was thinking of using old non-interactive default rules. I did not realize we could actually save padding using the index that is a multiple of the minimumSquareSize.

rootulp:
> Should i make a diagram for that ?

Sorry for repeating but my comment isn't blocking. In other words, I think a section on non-interactive defaults would help but doesn't block this PR. If you share the software you used to generate the diagrams, I can try drafting one

docs/architecture/adr-008-blocksize-independent-commitment.md
rootulp: > and make this function default to a minimum square size of appconsts.MinSquareSize?

Disregard, I learned we don't want to do this and the proposed minimalBlocksize dependent on message size seems preferable.

> The table could have number of subroots instead of size of message inclusion proof as this is directly proportional right?

Yes, I think so.

docs/architecture/adr-008-blocksize-independent-commitment.md
rootulp: This is resolvable since we'll use a minimumSquareSize that is message dependent. Perhaps `msgMinimumSquareSize` is a more appropriate name to disambiguate from `appconsts.MinSquareSize`

evan-forbes (approved):
> Should i make a diagram for that ?

I think its fine. We should remember to update the status of this ADR after we complete the implementation, and potentially update the naming after the implementation as well.

evan-forbes:
w/o further feedback, imo we should merge this on monday

rootulp:
> and potentially update the naming

Specifically the remaining renames for block size => square size are the PR title and Markdown heading

docs/architecture/adr-008-square-size-independent-message-commitments.md
```diff
@@ -0,0 +1,123 @@
...
+We should note that Rollups can decide to do this scheme without changing core-app apart from Number 4.
+
+## Positive Core App changes
+
+Core app simplifications after refactoring core-app:
+
+- simplifies malleation process
+- simplifies arranging the square
+- simplifies the creation of PFDs
+- **Padding Savings**
```
rootulp, line 109: 🙌

docs/architecture/adr-008-square-size-independent-message-commitments.md
```diff
@@ -0,0 +1,123 @@
...
+func msgMinSquareSize(messageLength uint64) uint64 {
+    squareSize := NextHigherPowerOf2(messageLength)
+    //Check if message fits with non-interactive default rules
+    if messageLength <= (squareSize * (squareSize -1)) {
+        return squareSize
+    } else {
+        return squareSize << 1
+    }
+}
+```
```
rootulp, line 61: [optional] I think we should calculate msgMinSquareSize without consideration for the current non-interactive defaults. This is safe to do because msgMinSquareSize that ignores non-interactive defaults is less than or equal to a msgMinSquareSize that conforms to current or future non-interactive defaults.

```suggestion
`msgMinSquareSize` can be calculated like this:

```go
// MsgMinSquareSize returns the minimum square size that msgSize can be included
// in. The returned square size does not account for the associated transaction
// shares or non-interactive defaults so it is a minimum.
func MsgMinSquareSize(msgSize uint64) uint64 {
	shareCount := uint64(shares.MsgSharesUsed(int(msgSize)))
	return MinSquareSize(shareCount)
}

// MinSquareSize returns the minimum square size that can contain shareCount
// number of shares.
func MinSquareSize(shareCount uint64) uint64 {
	return shares.RoundUpPowerOfTwo(uint64(math.Ceil(math.Sqrt(float64(shareCount)))))
}
```
```

File: docs/architecture/adr-008-square-size-independent-message-commitments.md [hunk 30]
```diff
@@ -0,0 +1,123 @@
...
+> - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
+
+To this:
+*Messages start at an index that is a multiple of its `msgMinSquareSize`.*
+
+As an example, we have this diagram. Message 1 is three shares long and is followed by message 2, which is 11 shares long, so the `msgMinSquareSize` of the second message is equal to four. Therefore we have a padding of 5 shares shown in light blue. Furthermore, with the new non-interactive default rule set, a message of size 11 can start in this block at index zero and index three because they are multiples of four. Therefore, we save four shares of padding while retaining the same commitment.
+
+![Padding Savings](./assets/padding-savings.png)
+
+### Neutral
```
rootulp, line 123: [optional] this heading from the template doesn't appear necessary if we don't populate its contents

```suggestion
```

rootulp (approved):

rootulp (approved):

Issue #830, opened 2022-10-03
All shares must contain a share version that belongs to a list of supported versions


rootulp:
### Problem
https://github.com/celestiaorg/celestia-app/issues/659 introduced an info byte which contains a share version. Currently a block proposer could construct a share with a version that hasn't yet been defined (ex. share version `5`). The block that includes this share would be considered valid even though it should be considered invalid.

### Proposal
Introduce a new validity rule that ensures all shares contain a share version that belongs to a list of supported versions. Initially the list of supported versions should contain share version: `0` because https://github.com/celestiaorg/celestia-app/blob/e69283fa6afdce9a96e6fe264ae007a8c74b18f0/pkg/appconsts/appconsts.go#L26

rootulp:
Upon further investigation, this validity rule doesn't seem practical in [`ProcessProposal` ](https://github.com/celestiaorg/celestia-app/blob/6b86e91eea1063c27c1ae461eddbe505f778df08/app/process_proposal.go#L23) because the share version is not included in block [data](https://github.com/celestiaorg/celestia-core/blob/094cd2c631a5b970ce89e04a81f404c99c38bdf8/types/block.go#L1005-L1032) so a validator has no way of knowing what share version to use for the shares that are split from the block data. The current implementation splits block data into shares and parses shares back into block data using the only valid share format (`0`). If we want to enforce a block validity rule for share versions, then we need a mechanism of persisting share versions into the block data.

Alternatively, if we don't want this to be a `ProcessProposal` validity rule, we can enforce this type of check when we parse the data square back into block data (e.g. [here](https://github.com/celestiaorg/celestia-app/blob/6b86e91eea1063c27c1ae461eddbe505f778df08/pkg/shares/share_merging.go#L14))

rootulp:
> then we need a mechanism of persisting share versions into the block data.

I think we should consider doing this. If all blocks write shares with one share version, then we can add this as a field to [Data](https://github.com/celestiaorg/celestia-core/blob/9002f1bf6be303d2a6f46865ead288812fc9a49a/proto/tendermint/types/types.pb.go#L419-L429) and verify it in `ProcessProposal`. It also seems necessary for when we increase share versions in the future. This needs further investigation if we want to support the scenario where one block can include shares with different share versions but I don't see a use case for that currently. I'd appreciate your input @evan-forbes

evan-forbes:
what you've discussed here makes sense

[...]

PR #837, opened 2022-10-04
feat: throw error when parsing unsupported share version

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/830 specifically:

> Alternatively, if we don't want this to be a ProcessProposal validity rule, we can enforce this type of check when we parse the data square back into block data (e.g. [here](https://github.com/celestiaorg/celestia-app/blob/6b86e91eea1063c27c1ae461eddbe505f778df08/pkg/shares/parse_compact_shares.go#L14))

### DIFF
--- a/pkg/appconsts/appconsts.go
+++ b/pkg/appconsts/appconsts.go
@@ -135,6 +135,9 @@ var (
 	// contains less space for data than a ContinuationCompactShare because the
 	// first compact share includes a total data length varint.
 	FirstCompactShareContentSize = ContinuationCompactShareContentSize - FirstCompactShareDataLengthBytes
+
+	// SupportedShareVersions is a list of supported share versions
+	SupportedShareVersions = []uint8{ShareVersion}
 )
 
 // numberOfBytesVarint calculates the number of bytes needed to write a varint of n
--- a/pkg/appconsts/appconsts.go
+++ b/pkg/appconsts/appconsts.go
@@ -136,7 +136,7 @@ var (
 	// first compact share includes a total data length varint.
 	FirstCompactShareContentSize = ContinuationCompactShareContentSize - FirstCompactShareDataLengthBytes
 
-	// SupportedShareVersions is a list of supported share versions
+	// SupportedShareVersions is a list of supported share versions.
 	SupportedShareVersions = []uint8{ShareVersion}
 )
 
--- a/pkg/shares/parse_compact_shares.go
+++ b/pkg/shares/parse_compact_shares.go
@@ -9,11 +9,13 @@ import (
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 )
 
-// parseCompactShares takes rawShares and extracts out transactions,
-// intermediate state roots, or evidence. The returned [][]byte do not have
-// namespaces, info bytes, data length delimiter, or unit length
-// delimiters and are ready to be unmarshalled
-func parseCompactShares(rawShares [][]byte) (data [][]byte, err error) {
+// parseCompactShares returns data (transactions, intermediate state roots, or
+// evidence) based on the contents of rawShares and supportedShareVersions. If
+// rawShares contains a share with a version that isn't present in
+// supportedShareVersions, an error is returned. The returned data [][]byte does
+// not have namespaces, info bytes, data length delimiter, or unit length
+// delimiters and are ready to be unmarshalled.
+func parseCompactShares(rawShares [][]byte, supportedShareVersions []uint8) (data [][]byte, err error) {
 	if len(rawShares) == 0 {
 		return nil, nil
 	}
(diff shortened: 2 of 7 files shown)


pkg/shares/utils.go
```diff
@@ -93,3 +93,13 @@ func zeroPadIfNecessary(share []byte, width int) (padded []byte, bytesOfPadding
 	share = append(share, padding...)
 	return share, missingBytes
 }
+
+// contains returns true if slice contains element.
+func contains(slice []byte, element byte) bool {
```
SweeXordious, line 98: [Non Blocking][Suggestion]
Why not use https://pkg.go.dev/bytes#Contains ?

pkg/shares/parse_compact_shares.go
```diff
@@ -3,20 +3,31 @@ package shares
...
 )
 
-// parseCompactShares takes raw shares and extracts out transactions,
+// parseCompactShares takes rawShares and extracts out transactions,
 // intermediate state roots, or evidence. The returned [][]byte do not have
 // namespaces, info bytes, data length delimiter, or unit length
 // delimiters and are ready to be unmarshalled
-func parseCompactShares(shares [][]byte) (data [][]byte, err error) {
-	if len(shares) == 0 {
+func parseCompactShares(rawShares [][]byte) (data [][]byte, err error) {
```
SweeXordious, line 15: [Non Blocking][Question]
I like the idea of checking the share version when parsing, since inside `ProcessProposal`, we don't have this information. However, since this might be used in multiple contexts, doesn't it make more sense to pass the supported share versions as a parameter? While providing a default one.
This way, we can parse shares for any version.

pkg/appconsts/appconsts.go
```diff
@@ -135,6 +135,9 @@ var (
 	// contains less space for data than a ContinuationCompactShare because the
 	// first compact share includes a total data length varint.
 	FirstCompactShareContentSize = ContinuationCompactShareContentSize - FirstCompactShareDataLengthBytes
+
+	// SupportedShareVersions is a list of supported share versions
```
SweeXordious, line 139:
```suggestion
	// SupportedShareVersions is a list of supported share versions.
```

pkg/shares/parse_compact_shares.go
```diff
@@ -3,20 +3,31 @@ package shares
...
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 )
 
-// parseCompactShares takes raw shares and extracts out transactions,
+// parseCompactShares takes rawShares and extracts out transactions,
 // intermediate state roots, or evidence. The returned [][]byte do not have
 // namespaces, info bytes, data length delimiter, or unit length
 // delimiters and are ready to be unmarshalled
-func parseCompactShares(shares [][]byte) (data [][]byte, err error) {
```
SweeXordious, line 14: [Non Blocking][Question]
Where is this actually called? I found it being used in `func Merge(eds...)`, but I don't see this latter being used anywhere

SweeXordious (commented):
utAck, LGTM, some questions to understand more

pkg/shares/parse_compact_shares.go
rootulp: > Where is this actually called?

1. https://github.com/rootulp/celestia-app/blob/24ddfce6c6b523f4e0dc5de6dd780bc3ad5ed272/pkg/shares/share_merging.go#L80
2. https://github.com/rootulp/celestia-app/blob/24ddfce6c6b523f4e0dc5de6dd780bc3ad5ed272/pkg/shares/share_merging.go#L80

which get called from `func Merge(eds...)` like you said. I also don't see `Merge` used anywhere outside tests

pkg/shares/utils.go
rootulp: Good idea. Done in https://github.com/celestiaorg/celestia-app/pull/837/commits/9a0d60e2d6be8fd82a45086f3b3a08a480be971c

pkg/shares/parse_compact_shares.go
rootulp: good point. I made `supportedShareVersions` a param in https://github.com/celestiaorg/celestia-app/pull/837/commits/ff75095f6eef1d4f7e4f3c16168801e141ce6ecd

I only lifted this as far as the public `Parse*` methods because those are public and may be called from external consumers. It doesn't seem necessary for an external consumer (i.e. celestia-node) to pass in a different list of supportedShareVersions because that list is defined in this repo. In the future, when we have share version `1`, logic in this repo's `share` package will change to conditionally parse different share versions (i.e. `1` and `0` differently) and at the same time, we can add `1` to https://github.com/rootulp/celestia-app/blob/ff75095f6eef1d4f7e4f3c16168801e141ce6ecd/pkg/appconsts/appconsts.go#L140

pkg/shares/parse_compact_shares.go
SweeXordious: As far as I understand, this was moved from core to app as it was mainly used in app. And, this was done prior to non-interactive defaults as intended to be used by that. 
I don't want to mention Evan as he's on vacation and if I mention him, he will respond :D

rootulp: Agreed. It seems non-blocking to find usage of this `Merge(eds...)` function so I think this can merge irrespective of investigating that further

SweeXordious: Should we create an issue for that so that we remmeber it? or just keep it in mind until Evan is back?
I am fine with any. Approooving

SweeXordious (approved):

pkg/shares/parse_compact_shares.go
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/840

PR #838, opened 2022-10-05
Remove duplicated random valid namespace

raphaelts3:
Closes https://github.com/celestiaorg/celestia-app/issues/836

Hey, @rootulp I'm not an expert go, so I don't know if the move I've made is the best possible, but it did its job. That said if you have any suggestions to improve that I'm open to adjusting it.

PS: Since the repo doesn't contain the tag `hacktoberfest`, would be possible to label this PR with `hacktoberfest-accepted` so it can be validated there by the rule below?
> PR/MRs that also have the “hacktoberfest-accepted” label cannot be marked as spammy via a label.

### DIFF
--- a/pkg/nsgenerator/random_ns_generator.go
+++ b/pkg/nsgenerator/random_ns_generator.go
@@ -1,4 +1,4 @@
-package utils
+package nsgenerator
 
 import (
 	"bytes"
@@ -8,7 +8,7 @@ import (
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 )
 
-func RandomValidNamespace() namespace.ID {
+func RandomMessageNamespace() namespace.ID {
 	for {
 		ns := tmrand.Bytes(8)
 		isReservedNS := bytes.Compare(ns, appconsts.MaxReservedNamespace) <= 0
--- a/pkg/utils/util.go
+++ b/pkg/utils/util.go
@@ -1,0 +1,22 @@
+package utils
+
+import (
+	"bytes"
+
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/nmt/namespace"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func RandomValidNamespace() namespace.ID {
+	for {
+		ns := tmrand.Bytes(8)
+		isReservedNS := bytes.Compare(ns, appconsts.MaxReservedNamespace) <= 0
+		isParityNS := bytes.Equal(ns, appconsts.ParitySharesNamespaceID)
+		isTailPaddingNS := bytes.Equal(ns, appconsts.TailPaddingNamespaceID)
+		if isReservedNS || isParityNS || isTailPaddingNS {
+			continue
+		}
+		return ns
+	}
+}
--- a/app/estimate_square_size_test.go
+++ b/app/estimate_square_size_test.go
@@ -5,7 +5,7 @@ import (
 
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
-	"github.com/celestiaorg/celestia-app/pkg/nsgenerator"
+	"github.com/celestiaorg/celestia-app/pkg/namespace"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/celestiaorg/celestia-app/x/payment/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -128,7 +128,7 @@ func Test_overEstimateMalleatedTxSize(t *testing.T) {
 		wpfdTx := generateRawWirePFDTx(
 			t,
 			encConf.TxConfig,
-			nsgenerator.RandomMessageNamespace(),
+			namespace.RandomMessageNamespace(),
 			tmrand.Bytes(tt.size),
 			signer,
 			tt.opts...,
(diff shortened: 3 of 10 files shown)


pkg/utils/util.go
```diff
@@ -0,0 +1,22 @@
+package utils
```
SweeXordious, line 1: [Non Blocking][Suggestion]
Might make sense to rename the file to something like `namespace` or similar

pkg/utils/util.go
```diff
@@ -0,0 +1,22 @@
...
+	"bytes"
+
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/nmt/namespace"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func RandomValidNamespace() namespace.ID {
+	for {
+		ns := tmrand.Bytes(8)
```
SweeXordious, line 13: [Blocking][Question]
This function seems to create a valid namespace that is a message. However, some of the other deleted functions:
```golang
func randomValidNamespace() namespace.ID {
	for {
		s := tmrand.Bytes(8)
		if bytes.Compare(s, appconsts.MaxReservedNamespace) > 0 {
			return s
		}
	}
}
```
Seem to define a valid namespace as smaller than the max reserved namespace.
Is this alright? 
Do we need to generate messages namespaces in some tests while in others we need any namespace that is `<= MaxReservedNamespace`?

SweeXordious (commented):
Thanks a lot for helping out with this. LGTM. Left a blocking question

pkg/utils/util.go
raphaelts3: That's a much better name, thank you!

pkg/utils/util.go
raphaelts3: The one that was kept also validated that `bytes.Compare(ns, appconsts.MaxReservedNamespace) <= 0` and I believe in a better way.
Also, the idea of keeping this specific function, came from the original issue https://github.com/celestiaorg/celestia-app/issues/836 where @rootulp mentioned:
> The last function definition is the most preferred because it accounts for the parity and tail padding namespaces. 

So, this one only should be fine.

rootulp (dismissed):
Thanks for the contribution @raphaelts3 ! Excited to hear you're participating in [hacktoberfest](https://hacktoberfest.com/).

I think the label you mentioned is sufficient to pass their rule based on:
Screen Shot 2022-10-05 at 2 36 16 PM

pkg/utils/util.go
SweeXordious: Ah, sounds good :+1:

pkg/utils/util.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"bytes"
+
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/nmt/namespace"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func RandomValidNamespace() namespace.ID {
```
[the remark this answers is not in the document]
raphaelts3: Accepted and renamed.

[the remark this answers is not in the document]
raphaelts3: The new name still doesn't feel like the best one but seems less troublesome than the previous one.

Thanks for [this](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common) article, very good points there.

SweeXordious (approved):
:rocket:

pkg/nsgenerator/random_ns_generator.go
```diff
@@ -0,0 +1,22 @@
+package nsgenerator
```
rootulp, line 1: [blocking]

The prefix `ns` may not be known by new contributors so proposal to rename this back to `utils` or `namespace` with a slight preference for `namespace`.


If we use `namespace`, we  may need to use an [import alias](https://scene-si.org/2018/01/25/go-tips-and-tricks-almost-everything-about-imports/) if the file already uses nmt's namespace.

rootulp (changes requested):

pkg/nsgenerator/random_ns_generator.go
raphaelts3: `namespace` it's!

evan-forbes:
[optional][should be done in a different PR]
I don't want to block this, since I haven't participated in reviews so far. So this is just a nit comment that we should address in a different PR

It makes sense to me tho that the new namespace package is moved to the testutil package, given that this is exclusively a testutil function

pkg/utils/util.go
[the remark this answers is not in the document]
evan-forbes: `RandomMessageNamespace` :+1: makes sense to me

pkg/utils/util.go
evan-forbes: > utils isn't an ideal package name for the reasons stated [here](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common).
namespace may cause import conflicts in files that use this utility and https://github.com/celestiaorg/nmt/blob/master/namespace/doc.go

second this comment.

rootulp:
> It makes sense to me tho that the new namespace package is moved to the testutil package, given that this is exclusively a testutil function

I took a stab at this yesterday but it caused cyclic imports. Agreed that there may be a more desirable long-term location for this util

evan-forbes:
>  but it caused cyclic imports.

> the new namespace package is moved to the testutil package

while this doesn't resolve the annoying name conflict, it should fix the cyclic imports, no?

rootulp:
oh sorry, I meant that I tried moving the `func RandomValidNamespace` to the `testutil` package but I think your proposal is to move the namespace package inside the testutil directory.

PR #842, opened 2022-10-06
docs: move shares README to godoc

rootulp:
Motivated by celestia-node's [package-specific-documentation](https://github.com/celestiaorg/celestia-node#package-specific-documentation), I opted to move the shares package `README.md` to godoc. Some changes were necessary because godoc doesn't support all Markdown features. Also, I updated "message start" terminology to use "sequence start" because it applies to non-message shares.

## Screenshot

Screen Shot 2022-10-05 at 10 43 10 PM


SweeXordious (commented):
LGTM. Thanks a lot for taking care of this. But, I would defer to Evan, as I never used `godoc` before.

SweeXordious:
Interesting, thanks, didn't know we were publishing docs

evan-forbes (approved):
I'm unsure of this suggestion, so please feel free to ignore or modify:

From my own personal workflow, I appreciate seeing markdown files in github/text editors. Markdown files are also easier to edit than go comments. Using godocs is also super legit. Do we know of a way to do both, have a markdown file that gets pushed to go docs? if not nbd, we can just use godocs or maybe include a readme that links to the godocs?

rootulp:
I'd rather read and edit Markdown than Godoc comments so I agree with your concern. It seems like a README.md is preferable for developers contributing to this repo but Godoc comments are preferable for developers using the share package as a dependency because the Godoc comment will be more discoverable than a README.md in the share package. 

I'm not opposed to a README.md in the share package that links to the Godoc comments but I'm more inclined to copy celestia-node and include a section like https://github.com/celestiaorg/celestia-node#package-specific-documentation in our root README.md because that's more discoverable

PR #845, opened 2022-10-06
fix(pkg/da): ensure DAH.hash is always calculated during ValidateBasic

Wondertan:
This is useful for headers downloaded from the network. Each such header goes through validation during which the hash should be calculated. Otherwise, the hash is calcualated at much later point, which is prone to races. In celestia-node we observed such races when the DAH.Hahs is read from multiple routines, where both end up calculating the hash and setting the it to the cache value making the race detector complain. Such precompute fixes this.

### DIFF
--- a/pkg/da/data_availability_header.go
+++ b/pkg/da/data_availability_header.go
@@ -150,7 +150,7 @@ func (dah *DataAvailabilityHeader) ValidateBasic() error {
 			len(dah.ColumnRoots),
 		)
 	}
-	if err := validateHash(dah.hash); err != nil {
+	if err := validateHash(dah.Hash()); err != nil {
 		return fmt.Errorf("wrong hash: %v", err)
 	}
 


evan-forbes (approved):
nice find, :+1: :+1:

evan-forbes:
is this needed for https://github.com/celestiaorg/celestia-node/pull/1160?

rootulp (approved):

Wondertan:
> is this needed for https://github.com/celestiaorg/celestia-node/pull/1160?

I observed the described race while playing with the code there and enabling race detector for tests it was disabled before 1.19. Don't think it's blocking the merge of #1160 as it's a mostly unrelated, but still should be included in the upcoming release.

rootulp:
Merging based on https://github.com/celestiaorg/celestia-node/pull/1160?notification_referrer_id=NT_kwDOADhxZ7I0NDY0MzU0MDEyOjM2OTkwNDc#issuecomment-1273054501

PR #848, opened 2022-10-07
Update qgb orchestrator relayer branch to main

SweeXordious:
Updates the QGB orchestrator relayer branch to latest main


SweeXordious:
The problem is that I had some commits I shouldn't have included in the PR, so I force pushed to remove those commits. However, the CI is still running on the old wrong commit

PR #853, opened 2022-10-10
fix: FitsInSquare for case with 0 messages

rootulp:
Closes https://github.com/informalsystems/audit-celestia/issues/16


evan-forbes (approved):
thanks for fixing this so quick

Issue #858, opened 2022-10-11
Rename token ticker from `TIA` to `CLS`


rootulp:
Currently the token ticker uses `TIA` and `utia`. We'd like to rename to `CLS` and `ucls` respectively.

https://github.com/celestiaorg/celestia-app/blob/7ca6eca76824fc54816243da9a695f56669cc27f/app/app.go#L94-L99

PR #859, opened 2022-10-11
chore!: rename `TIA` to `CLS`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/858
Opens https://github.com/celestiaorg/celestia-app/issues/862

### DIFF
--- a/x/payment/types/builder_options.go
+++ b/x/payment/types/builder_options.go
@@ -65,7 +65,7 @@ func InheritTxConfig(builder sdkclient.TxBuilder, tx authsigning.Tx) sdkclient.T
 		builder.SetGasLimit(gas)
 	}
 
-	if feeAmmount := tx.GetFee(); !feeAmmount.AmountOf("utia").Equal(sdk.NewInt(0)) {
+	if feeAmmount := tx.GetFee(); !feeAmmount.AmountOf("ucls").Equal(sdk.NewInt(0)) {
 		builder.SetFeeAmount(tx.GetFee())
 	}
 
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ celestia-appd --help
 # Create your own single node devnet
 celestia-appd init test --chain-id test
 celestia-appd keys add user1
-celestia-appd add-genesis-account <address from above command> 10000000utia,1000token
-celestia-appd gentx user1 1000000utia --chain-id test
+celestia-appd add-genesis-account <address from above command> 10000000ucls,1000token
+celestia-appd gentx user1 1000000ucls --chain-id test
 celestia-appd collect-gentxs
 celestia-appd start
 
--- a/app/app.go
+++ b/app/app.go
@@ -92,11 +92,11 @@ const (
 	AccountAddressPrefix = "celestia"
 	Name                 = "celestia-app"
 	// BondDenom defines the native staking token denomination.
-	BondDenom = "utia"
+	BondDenom = "ucls"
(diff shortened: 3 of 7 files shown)


x/payment/types/builder_options.go
```diff
@@ -65,7 +65,7 @@ func InheritTxConfig(builder sdkclient.TxBuilder, tx authsigning.Tx) sdkclient.T
 		builder.SetGasLimit(gas)
 	}
 
-	if feeAmmount := tx.GetFee(); !feeAmmount.AmountOf("utia").Equal(sdk.NewInt(0)) {
+	if feeAmmount := tx.GetFee(); !feeAmmount.AmountOf("ucls").Equal(sdk.NewInt(0)) {
```
rootulp, line 68: [note for reviewers] we can't replace this with `app.BondDenom` because it will introduce an import cycle.

We may consider extracting [these constants](https://github.com/rootulp/celestia-app/blob/b92d17d709c688f8f9174f7b868829b70154ed73/app/app.go#L94-L99) to the `appconsts` package to mitigate

SweeXordious (commented):
[Blocking][Question]
Doesn't it make sense to also update the following:
- `scripts/single-node.sh` line 8 and 13
- README line 51 and 52
- `app/test/block_size_test.go` line 33
And, also others, in case I missed any?

SweeXordious:
Thanks a lot for tackling this. For the following:

> This change implies breaking changes in multiple downstream repos: https://github.com/search?p=1&q=org%3Acelestiaorg+tia&type=Code

Might be a good idea to write a tracker issue that concern all repos to do the update. What do you think?

rootulp:
> Doesn't it make sense to also update the following:

Good catch. I missed these b/c VSCode "Match Whole Word" was enabled 🤦‍♂️ 

Screen Shot 2022-10-11 at 9 17 07 PM

rootulp:
> Might be a good idea to write a tracker issue that concern all repos to do the update. What do you think?

Good idea! I created an issue and updated the PR description.

SweeXordious (approved):
Awesome :tada:

evan-forbes (approved):
Dope! I'm also unsure on when to include this in an upgrade, as this will require special migration logic for any existing networks

evan-forbes:
we can merge this now imo

evan-forbes:
we really need to switch to a more normal v0.x.x release branch esque style or something to avoid issues like this. It seems silly to hold of on merge already completed work :sweat_smile:

rootulp:
One possible approach is to adopt a more frequent release cutting process from `main`. We can invest in tooling to expedite this process or even do it automatically based on conventional commits. If we cut releases often, then I think we'll have fewer scenarios where PRs are blocked on a previous release being created. This has the additional benefit that celestia-app will have many small releases that downstream repos can upgrade to incrementally so we get feedback earlier than if we bundle many changes into one larger release.

evan-forbes:
yes, good points, we should be cutting releases more frequently. We also have to maintain different versions of core across multiple different testnets/mainet, so I'm not sure yet if simply cutting more releases is a full replacement to having release branches.

do you have specific tooling in mind?

SweeXordious:
This guy looks awesome https://github.com/semantic-release/semantic-release :fire: 

We might also want to use something similar to snapshot releases: https://www.baeldung.com/maven-snapshot-release-repository#snapshot-repository
I was using them when I was coding in Java/Kotlin, and they made life easier because they always give you access to latest master and unreleased changes, in different repos, so you can build upon them

PR #863, opened 2022-10-12
Add a package-specific documentation section

rootulp:
Inspired by https://github.com/celestiaorg/celestia-node#package-specific-documentation

Addresses https://github.com/celestiaorg/celestia-app/pull/842#pullrequestreview-1139027323

We can keep the share docs in Godoc form or move them back to README.md form because both will render in https://pkg.go.dev/github.com/celestiaorg/celestia-app/pkg/shares

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ make fmt
 make proto-gen
 ```
 
+### Package-specific documentation
+
+- [Shares](https://pkg.go.dev/github.com/celestiaorg/celestia-app/pkg/shares)
+
 ## Careers
 
 We are hiring Go engineers! Join us in building the future of blockchain scaling and interoperability. [Apply here](https://jobs.lever.co/celestia).


README.md
```diff
@@ -82,6 +82,10 @@ make fmt
 make proto-gen
 ```
 
+### Package-specific documentation
+
+- [Shares](https://pkg.go.dev/github.com/celestiaorg/celestia-app/pkg/shares)
```
SweeXordious, line 87: [Non Blocking][Question]
Does this line help in publishing the docs to go doc, instead of writing the actual godoc? or it's for users?

SweeXordious (commented):
What do you think of the following tool ?
https://github.com/posener/goreadme

Might make sense to add it and make our lives even easier

README.md
rootulp: This line is just for users

rootulp:
> What do you think of the following tool ?
https://github.com/posener/goreadme

Good find. Looks like a cool tool and on cursory investigation, I think we may be able to use it to generate package specific READMEs by overriding:
- https://github.com/posener/goreadme/blob/23812f8369444ad3655570992c32b2c8f3bac433/action.yml#L6
- https://github.com/posener/goreadme/blob/23812f8369444ad3655570992c32b2c8f3bac433/action.yml#L26

SweeXordious (approved):

Issue #839, opened 2022-10-05
`ParseShare` should return an error if total sequence length varint doesn't match actual


rootulp:
### Context
Motivated by https://github.com/celestiaorg/celestia-app/pull/828#pullrequestreview-1128996407

### Problem
The prefix of the first share of a sequence contains a total data length varint. Currently the `ParseShare` API ignores this varint, however it can throw an error if the total data length varint doesn't match the actual data length of the share sequence.

### Proposal
With the assumption that we want `ParseShare` to return an error ([ref](https://github.com/celestiaorg/celestia-app/pull/828#pullrequestreview-1130893811)), we must introduce a new check.

Note, we can't determine the actual data length because `0` padding is added to the end of a share and it isn't possible to diffrentiate between padding `0` bytes and data `0` bytes without relying on the total data length varint. Therefore, this check will likely compare based on # of shares and amount of data per share:

```
(len(shareSequence.Shares) - 1) * appconsts. SparseShareContentSize <= varint <= len(shareSequence.Shares) * appconsts. SparseShareContentSize
```

PR #865, opened 2022-10-12
feat: throw error if incorrect sequenceLength

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/839


pkg/shares/share_merging.go
```diff
@@ -176,5 +176,58 @@ func ParseShares(rawShares [][]byte) ([]ShareSequence, error) {
...
 }
+
+// validSequenceLength extracts the sequenceLength written to the first share
+// and returns an error if the number of shares needed to store a sequence of
+// length sequenceLength doesn't match the number of shares in this share
+// sequence. Returns nil if there is no error.
+func (s ShareSequence) validSequenceLength() error {
+	if len(s.Shares) == 0 {
+		return nil
+	}
```
evan-forbes, line 195: [question]
is any sequence of length 0 valid? I was always under the impression that it was explicitly invalid

pkg/shares/share_merging.go
```diff
@@ -176,5 +176,58 @@ func ParseShares(rawShares [][]byte) ([]ShareSequence, error) {
...
+}
+
+// CompactSharesUsed returns the number of compact shares used to store a
+// sequence of length sequenceLength.
+func CompactSharesUsed(sequenceLength int) (sharesUsed int) {
+	if sequenceLength == 0 {
+		return 0
+	}
+
+	if sequenceLength < appconsts.FirstCompactShareContentSize {
```
evan-forbes, line 222: [question]
by sequence, are we referring to a sequence of shares, a sequence of compact units, or a sequence of bytes?

evan-forbes (commented):
[comment]
switching from message to sequence is good and makes more sense for universal shares imo

had a few preliminary questions even tho this is still a draft. so far LGTM!

pkg/shares/share_merging.go
rootulp: In this context: `sequenceLength` is the number of bytes of message, transaction, ISR, or evidence data in a sequence of shares.

A sequence of shares is applicable to both sparse and compact shares.

pkg/shares/share_merging.go
rootulp: Good question, I think you're right that it should always be explicitly invalid. I'll add a unit test for this edge case.

pkg/shares/share_merging.go
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/865/commits/84df6ae7c8575e8c49fc9b20c52879f8a0440037

pkg/shares/share_merging.go
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/865/commits/02905fc521eda72bcb152ce95a28384270c783ee

Note, a unit test wasn't added for this case b/c its impossible to reach this code path without finagling with internals of `ParseShares` because `validSequenceLength` won't be called if no sequences exist

evan-forbes (approved):

PR #874, opened 2022-10-16
docs: update code block in ADR-006

rootulp:
Addresses https://github.com/informalsystems/audit-celestia/issues/20#issuecomment-1278896978

Question for reviewers: do we treat ADRs as a snapshot in time or should we update them to reflect changes that take place after the ADR was reviewed?

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,3 +1,5 @@
 "default": true # Default state for all rules
+"MD010":
+  "code_blocks": false # Disable rule for hard tabs in code blocks
 "MD013": false # Disable rule for line length
 "MD033": false # Disable rule banning inline HTML
--- a/docs/architecture/adr-006-non-interactive-defaults.md
+++ b/docs/architecture/adr-006-non-interactive-defaults.md
@@ -108,8 +108,6 @@ Note: In order to properly test the new encoding scheme, we have to perform iden
 
 We currently utilize a struct to store the state needed for lazily writing message shares. Here we add a method to it that allows for us to write namespaced padded shares.
 
-<!-- markdownlint-disable MD010 -->
-
 ```go
 // SparseShareSplitter lazily splits messages into shares that will eventually be
 // included in a data square. It also has methods to help progressively count
--- a/docs/architecture/adr-006-non-interactive-defaults.md
+++ b/docs/architecture/adr-006-non-interactive-defaults.md
@@ -108,47 +108,49 @@ Note: In order to properly test the new encoding scheme, we have to perform iden
 
 We currently utilize a struct to store the state needed for lazily writing message shares. Here we add a method to it that allows for us to write namespaced padded shares.
 
+<!-- markdownlint-disable MD010 -->
+
 ```go
-// MessageShareSplitter lazily splits messages into shares that will eventually be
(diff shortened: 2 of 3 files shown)


docs/architecture/adr-006-non-interactive-defaults.md
```diff
@@ -108,47 +108,49 @@ Note: In order to properly test the new encoding scheme, we have to perform iden
 
 We currently utilize a struct to store the state needed for lazily writing message shares. Here we add a method to it that allows for us to write namespaced padded shares.
 
+<!-- markdownlint-disable MD010 -->
```
SweeXordious, line 111: [Non Blocking][Question]
Why disabling this? All other ADRs are using spaces instead of tabs. Is there a specific reason for this? 
I guess if we no longer want spaces instead of tabs, we could disable the markdown rule, better. What do you think?

evan-forbes: yeah, second this

evan-forbes (commented):
> Question for reviewers: do we treat ADRs as a snapshot in time or should we update them to reflect changes that take place after the ADR was reviewed?

we can do both, as that's what tendermint and cosmos do. Updating can be tedious, and we should only do it if it makes sense (or is requested) or leave them as good beginner issues imo

docs/architecture/adr-006-non-interactive-defaults.md
rootulp: > Why disabling this?

Because Go fmt uses tabs.

**Indentation**
We use tabs for indentation and gofmt emits them by default. Use spaces only if you must.

Ref: https://go.dev/doc/effective_go#formatting

If this rule isn't disabled, markdownlint will complain for correctly formatted Go code blocks and `markdownlint --fix` will reformat tabs to spaces (in other words change properly formatted Go code into unformatted Go code).

> All other ADRs, as far as I know, are using spaces instead of tabs.

True. This isn't for an ADR but there are other instances of disabling this rule prior to Go code blocks: 
- https://github.com/celestiaorg/celestia-app/blob/8e35ab2d06e38e7c71059b4f26c45b1944162ba7/x/payment/spec/docs.md?plain=1#L26
- https://github.com/celestiaorg/celestia-app/blob/8e35ab2d06e38e7c71059b4f26c45b1944162ba7/x/payment/spec/docs.md?plain=1#L118

> I guess if we no longer want spaces instead of tabs, we could disable the markdown rule from the linter directly.

I think we want this rule for all Markdown text _except_ code blocks and fortunately there is a config param for this. 

Ref: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md010---hard-tabs

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/874/commits/54777cefffb620d7aa0cd8b5f67d795d09c1b1f3

.markdownlint.yaml
```diff
@@ -1,3 +1,5 @@
 "default": true # Default state for all rules
+"MD010":
+  "code_blocks": false # Disable rule for hard tabs in code blocks
```
evan-forbes, line 3: :fire: :fire: :+1:

evan-forbes (approved):

PR #875, opened 2022-10-16
shares: extract powers of two

rootulp:
Resolves https://github.com/informalsystems/audit-celestia/issues/15

## Description
We have a few implementations of utility functions that handle powers of two. This PR consolidates them in the `shares` package, renames them, adds unit tests.


pkg/utils/powers_of_two.go
```diff
@@ -0,0 +1,70 @@
...
+	return c / 2
+}
+
+// RoundUpPowerOfTwoU rounds input up to the next power of two.
+// Examples:
+//
+//	RoundUpPowerOfTwoU(1) = 2
+//	RoundUpPowerOfTwoU(2) = 4
+//	RoundUpPowerOfTwoU(5) = 8
+func RoundUpPowerOfTwoU(input uint64) uint64 {
```
SweeXordious, line 27: [Non Blocking][Question]
Whats the difference between this and:

```go
func RoundUpPowerOfTwo(input int) int {
	k := 1
	for k < input {
		k = k << 1
	}
	return k
}
```

aside from the sign?

rootulp: good question. Unsigned vs signed is the rationale for two separate function definitions. As for why the implementations are different 🤷‍♂️ 

The unsigned implementation isn't particularly readable to me so I assume that it intentionally traded-off readability for performance.

Ref:
1. signed: https://www.tutorialspoint.com/golang-program-to-round-up-the-next-highest-power-of-2
2. unsigned (approach 4): https://www.techiedelight.com/round-next-highest-power-2/

I'm inclined to update the less readable unsigned implementation to the signed implementation but would need to run benchmarks to make sure performance isn't significantly different.

SweeXordious: agreed, if it's not giving us much performance, probably better sacrifice it for readability :+1: 
@evan-forbes When you have time, please let us know what you think :cat:

evan-forbes, line 27: [suggestion]
instead of using U, I think we could use generics.

evan-forbes (commented):
two suggestions

pkg/utils/powers_of_two.go
evan-forbes: yeah, we can rewrite it, I don't think it was about performance, and even it is its probably not that much

go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/celestiaorg/rsmt2d v0.6.0
 	github.com/regen-network/cosmos-proto v0.3.1
 	github.com/tendermint/tendermint v0.34.20
+	golang.org/x/exp v0.0.0-20221012211006-4de253d81b95
```
rootulp, line 34: this was added to import the type constraint https://pkg.go.dev/golang.org/x/exp@v0.0.0-20221012211006-4de253d81b95/constraints#Integer

pkg/utils/powers_of_two_test.go
```diff
@@ -0,0 +1,96 @@
+package utils
```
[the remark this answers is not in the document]
rootulp: Good idea! Resolved by https://github.com/celestiaorg/celestia-app/pull/875/commits/d44a8308b0a8516c1333e7b53de5041df8154191

pkg/utils/powers_of_two.go
rootulp: Tried it out in https://github.com/celestiaorg/celestia-app/pull/875/commits/c407ce28cc931200fd5e11630be4dcec60aedbe2 

Pro: results in less code
Con: less readable especially for developers unfamiliar with generics

SweeXordious (approved):
Great work :fire: 
I'm fine with the experimental import, even if it disappears, something equivalent will pop up.  So, I guess it's cool to use it for now.

evan-forbes (approved):
:+1:

Issue #684, opened 2022-09-05
QGB implement the confirms in transactions design


SweeXordious:
Implementation of the https://github.com/celestiaorg/celestia-app/pull/657

evan-forbes:
per sync discussion, this issue can be closed due to moving to v2

PR #878, opened 2022-10-17
feat: implement parser for data commitment/valset confirm messages

SweeXordious:
Implements a parser for tendermint and sdk messages for the QGB transactions.
Adds to: https://github.com/celestiaorg/celestia-app/issues/684

### DIFF
--- a/x/qgb/orchestrator/parser.go
+++ b/x/qgb/orchestrator/parser.go
@@ -1,0 +1,93 @@
+package orchestrator
+
+import (
+	"fmt"
+	"github.com/celestiaorg/celestia-app/app"
+	qgbtypes "github.com/celestiaorg/celestia-app/x/qgb/types"
+	sdkcodec "github.com/cosmos/cosmos-sdk/codec"
+	sdkcodectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/cosmos/cosmos-sdk/std"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
+	"github.com/gogo/protobuf/proto"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+var _ QGBParserI = QGBParser{}
+
+type QGBParserI interface {
+	IsDataCommitmentConfirm(msg sdktypes.Msg) (bool, error)
+	ParseDataCommitmentConfirm(msg sdktypes.Msg) (qgbtypes.MsgDataCommitmentConfirm, error)
+	IsValsetConfirm(msg sdktypes.Msg) (bool, error)
+	ParseValsetConfirm(msg sdktypes.Msg) (qgbtypes.MsgValsetConfirm, error)
+	ParseCoreTx(tx coretypes.Tx) (sdktx.Tx, error)
+	ParseSdkTx(any *sdkcodectypes.Any) (sdktypes.Msg, error)
+}
+
+type QGBParser struct {
+	codec sdkcodec.Codec
+}
+
+func NewQGBParser(codec sdkcodec.Codec) *QGBParser {
+	return &QGBParser{codec: codec}
+}
+
+func (parser QGBParser) IsValsetConfirm(msg sdktypes.Msg) (bool, error) {
+	switch msg.(type) {
+	case *qgbtypes.MsgValsetConfirm:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+func (parser QGBParser) IsDataCommitmentConfirm(msg sdktypes.Msg) (bool, error) {
+	switch msg.(type) {
+	case *qgbtypes.MsgDataCommitmentConfirm:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+func (parser QGBParser) ParseValsetConfirm(msg sdktypes.Msg) (qgbtypes.MsgValsetConfirm, error) {
+	vs, ok := msg.(*qgbtypes.MsgValsetConfirm)
+	if !ok {
+		return qgbtypes.MsgValsetConfirm{}, fmt.Errorf("not good")
+	}
+	return *vs, nil
+}
+
+func (parser QGBParser) ParseDataCommitmentConfirm(msg sdktypes.Msg) (qgbtypes.MsgDataCommitmentConfirm, error) {
+	dcc, ok := msg.(*qgbtypes.MsgDataCommitmentConfirm)
+	if !ok {
+		return qgbtypes.MsgDataCommitmentConfirm{}, fmt.Errorf("not good")
+	}
+	return *dcc, nil
+}
+
+func (parser QGBParser) ParseCoreTx(tx coretypes.Tx) (sdktx.Tx, error) {
+	var sdkTx sdktx.Tx
+	err := proto.Unmarshal(tx, &sdkTx)
+	if err != nil {
+		return sdktx.Tx{}, fmt.Errorf("error while unmarshalling core transaction: %s", err)
+	}
+	return sdkTx, nil
+}
+
+func (parser QGBParser) ParseSdkTx(any *sdkcodectypes.Any) (sdktypes.Msg, error) {
+	var stdMsg sdktypes.Msg
+	err := parser.codec.UnpackAny(any, &stdMsg)
+	if err != nil {
+		return nil, fmt.Errorf("error while unpacking message: %s", err)
+	}
+	return stdMsg, nil
+}
+
+func MakeDefaultAppCodec() sdkcodec.Codec {
+	interfaceRegistry := sdkcodectypes.NewInterfaceRegistry()
+	std.RegisterInterfaces(interfaceRegistry)
+	app.ModuleBasics.RegisterInterfaces(interfaceRegistry)
+	qgbtypes.RegisterInterfaces(interfaceRegistry)
+	return sdkcodec.NewProtoCodec(interfaceRegistry)
+}
--- a/x/qgb/orchestrator/parser.go
+++ b/x/qgb/orchestrator/parser.go
@@ -16,6 +16,7 @@ import (
 
 var _ QGBParserI = QGBParser{}
 
+// FIXME remove this interface after writing tests and see if it is needed or not
 type QGBParserI interface {
 	ParseDataCommitmentConfirm(msg sdktypes.Msg) (qgbtypes.MsgDataCommitmentConfirm, error)
 	ParseValsetConfirm(msg sdktypes.Msg) (qgbtypes.MsgValsetConfirm, error)
--- a/x/qgb/orchestrator/parser.go
+++ b/x/qgb/orchestrator/parser.go
@@ -54,7 +54,8 @@ func (parser QGBParser) IsDataCommitmentConfirm(msg sdktypes.Msg) (bool, error) 
 func (parser QGBParser) ParseValsetConfirm(msg sdktypes.Msg) (qgbtypes.MsgValsetConfirm, error) {
 	vs, ok := msg.(*qgbtypes.MsgValsetConfirm)
 	if !ok {
-		return qgbtypes.MsgValsetConfirm{}, fmt.Errorf("not good")
+		return qgbtypes.MsgValsetConfirm{},
+			fmt.Errorf("unable to cast sdk message to msg valset confirm")
 	}
 	return *vs, nil
 }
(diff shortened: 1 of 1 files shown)


rootulp (approved):
LGTM with one non-blocking question

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,96 @@
...
+func (parser QGBParser) ParseCoreTx(tx coretypes.Tx) (sdktx.Tx, error) {
+	var sdkTx sdktx.Tx
+	err := proto.Unmarshal(tx, &sdkTx)
+	if err != nil {
+		return sdktx.Tx{}, fmt.Errorf("error while unmarshalling core transaction: %s", err)
+	}
+	return sdkTx, nil
+}
+
+func (parser QGBParser) ParseSdkTx(any *sdkcodectypes.Any) (sdktypes.Msg, error) {
```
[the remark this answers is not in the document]
SweeXordious: Messages and transactions are different. A transaction is the following:
https://github.com/celestiaorg/cosmos-sdk/blob/482d8986c7842b93f40239a210fac8b3975db39a/types/tx/tx.pb.go#L33

which contains a body:
https://github.com/celestiaorg/cosmos-sdk/blob/482d8986c7842b93f40239a210fac8b3975db39a/types/tx/tx.pb.go#L351

containing messages of type `Any`:
https://github.com/celestiaorg/cosmos-sdk/blob/482d8986c7842b93f40239a210fac8b3975db39a/types/tx/tx.pb.go#L359

Also, for more information, a cosmos-sdk transaction is different from a tendermint transaction. The latter is basically `[]byte` while the former contains a lot of fields.

The reason I am using the terminology  `ParseSdkTx` is to reflect that this will be related to parsing what's inside a cosmos-sdk sdk transaction, and not a tendermint one.

You are right, for the naming I will create an issue to update it later. If I make this change in here, and I built a lot on this method name, I will go through a nightmare of conflicts haha 

Issue: https://github.com/celestiaorg/celestia-app/issues/879

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/cosmos/cosmos-sdk/std"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
+	"github.com/gogo/protobuf/proto"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+var _ QGBParserI = QGBParser{}
+
+type QGBParserI interface {
```
evan-forbes, line 19: do we need an interface here? are we planning for multiple different implementations?

evan-forbes (commented):
Had a few blocking questions and comments. At first glance, I'm a bit confused about the purpose of the parser.

What is the goal of the parser? Is it purely to type check or to sort? Should this just be a classic type switch statement?

x/qgb/orchestrator/parser.go
SweeXordious: I added an interface in case we need to mock this parser in unit tests. For example, if we want to test the ingestion service with random data without worrying if it is random

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (parser QGBParser) IsDataCommitmentConfirm(msg sdktypes.Msg) (bool, error) {
+	switch msg.(type) {
+	case *qgbtypes.MsgDataCommitmentConfirm:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
```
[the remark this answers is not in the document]
SweeXordious: You're right, they will always be used after parsing. So, probably a switch makes more sense. Should I remove these methods and make the change?

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (parser QGBParser) IsValsetConfirm(msg sdktypes.Msg) (bool, error) {
+	switch msg.(type) {
+	case *qgbtypes.MsgValsetConfirm:
+		return true, nil
+	default:
+		return false, nil
+	}
+}
```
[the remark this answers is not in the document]
SweeXordious: Good find, no need for an error in this case. Thanks

x/qgb/orchestrator/parser.go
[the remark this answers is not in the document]
evan-forbes: imo, yeah

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,96 @@
...
+	return stdMsg, nil
+}
+
+func MakeDefaultAppCodec() sdkcodec.Codec {
+	interfaceRegistry := sdkcodectypes.NewInterfaceRegistry()
+	std.RegisterInterfaces(interfaceRegistry)
+	app.ModuleBasics.RegisterInterfaces(interfaceRegistry)
+	qgbtypes.RegisterInterfaces(interfaceRegistry)
+	return sdkcodec.NewProtoCodec(interfaceRegistry)
+}
```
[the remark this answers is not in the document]
evan-forbes: the only time we don't is during testing for packages that are imported by the app package (or that is the only time that we should be ideally, like in the [payment](https://github.com/celestiaorg/celestia-app/blob/c9dce391bc360bb5f068f43d4eb1681a9f35b710/x/payment/types/codec.go#L51-L57) module). If this one is just for testing and we need the orchestrator package imported by the app package, then we should unexport this one.

x/qgb/orchestrator/parser.go
SweeXordious: I see. For me, I was thinking more of the ingestion service tests. Since it will need to extract data from a node (or a mock of a node that will return random data) and will need to index that data into the right place and be able to load it correctly.
What do you think about this 
 testing scenario?

x/qgb/orchestrator/parser.go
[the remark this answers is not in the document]
SweeXordious: This one is not for testing. It is used to decode messages that the ingestion receives.
Is there a place where I can get the app encoding without creating it myself here?
If so, that would be even better.

x/qgb/orchestrator/parser.go
evan-forbes: How will it index random data? Unless its hashing the random data, won't it need some metadata to use as an index? I could be missing something, but it doesn't seem that difficult to mock create mock data. When we used mocks before, it was because it was very difficult to spin up an entire chain to do unit tests, so the mocks make sense. I could be wrong tho, and if it is difficult, then we should keep it.

x/qgb/orchestrator/parser.go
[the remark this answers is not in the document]
evan-forbes: We create the encoding config each time to avoid requiring a global or access to the app instance itself, so unfortunately we just have to create it using the link in the comment above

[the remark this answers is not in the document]
SweeXordious: So, I create something like this on the state machine side:
```go
type localEncoder struct{}

func (localEncoder) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
	RegisterCodec(cdc)
}

func (localEncoder) RegisterInterfaces(r codectypes.InterfaceRegistry) {
	RegisterInterfaces(r)
}

func MakeQGBEncodingConfig() encoding.Config {
	e := localEncoder{}
	return encoding.MakeConfig(e)
}
```
and use it on the ingestion side?

x/qgb/orchestrator/parser.go
SweeXordious: I will add a TODO, and when implementing tests, we see if it makes sense to keep it or not. How does that sound?

x/qgb/orchestrator/parser.go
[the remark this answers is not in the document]
SweeXordious: fixed [93dffaa](https://github.com/celestiaorg/celestia-app/pull/878/commits/93dffaa3b4f36a1ed12369ae4041fc4246d74dd6)

x/qgb/orchestrator/parser.go
evan-forbes: sure, that works

SweeXordious: [68aad0d](https://github.com/celestiaorg/celestia-app/pull/878/commits/68aad0d9d7fecfbcc23ccb6971bf8fafbe1f54f4)

x/qgb/orchestrator/parser.go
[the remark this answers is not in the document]
SweeXordious: 8652ce9

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (parser QGBParser) ParseCoreTx(tx coretypes.Tx) (sdktx.Tx, error) {
+	var sdkTx sdktx.Tx
+	err := proto.Unmarshal(tx, &sdkTx)
+	if err != nil {
+		return sdktx.Tx{}, fmt.Errorf("error while unmarshalling core transaction: %s", err)
+	}
+	return sdkTx, nil
+}
```
evan-forbes, line 58: [not blocking][for future reference]
I'm fine with this, as it doesn't matter, but fwiw this is non-standard and not what we do in the rest of this repo

typically, we refer to the interface 

https://github.com/cosmos/cosmos-sdk/blob/c9b446029fc5e73560a3be966f06453ab33e3e83/types/tx_msg.go#L42-L50

x/qgb/orchestrator/parser.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (parser QGBParser) ParseSdkMsg(any *sdkcodectypes.Any) (sdktypes.Msg, error) {
+	var stdMsg sdktypes.Msg
+	err := parser.codec.UnpackAny(any, &stdMsg)
+	if err != nil {
+		return nil, fmt.Errorf("error while unpacking message: %s", err)
+	}
+	return stdMsg, nil
+}
```
evan-forbes, line 67: [question] 
are we still using `Any`s anymore?

evan-forbes (approved):

x/qgb/orchestrator/parser.go
SweeXordious: yes, for attestations

x/qgb/orchestrator/parser.go
evan-forbes, line 58:
```suggestion
func (parser QGBParser) ParseCoreTx(tx coretypes.Tx) (sdktx.Tx, error) {
	var sdkTx sdktx.Tx
	err := proto.Unmarshal(tx, &sdkTx)
	return sdkTx, err
}
```

SweeXordious: If I understand right, that is unmarshalling a tendermint transaction to an sdk one. How do you propose to go from one to another?
The other way I think of it is getting the hash from the tendermint transaction and querying it from the app, but I wanted to avoid that double download of transactions (was done is juno block explorer and was so slow to ingest).
If there is a better way to do, I am really open for it

evan-forbes: we don't have to use, as I don't see specific benefits other than following the docs, but the standard way that is used by most cosmos-sdk chains is to use the encoding config
```go
func decode(b coretypes.Block) error {
	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)

	for _, rawTx := range b.Data.Txs {
		sdkTx, err := encCfg.TxConfig.TxDecoder()(tx)
		if err != nil {
			return err
		}
		for _, msg := range sdkTx.GetMsgs() {
			url := sdk.MsgTypeURL(msg)
			if !qgbtypes.IsConfirmURL(url) {
				continue
			}
			switch msg.(type) {
			case *qgbtypes.MsgValsetConfirm:

			}
		}
	}
}
```

evan-forbes: the encoding config is threadsafe iirc

SweeXordious: That I didn't know is possible, I will update it to this and see if it still works as expected :+1: 
Thanks a lot for your help

evan-forbes: checking for protobuf string url is also faster than type checking, so it acts as a quick check before we do the heavy type check

evan-forbes: thanks, the main reason we should imo is there is quite a bit of other logic going on in the [default decoder](https://github.com/cosmos/cosmos-sdk/blob/c9b446029fc5e73560a3be966f06453ab33e3e83/x/auth/tx/decoder.go#L15-L76). Just using protobuf _should_ work, but sticking to what the sdk useses will likely save us from weird bugs in the future. It also should work with not just protubuf encoded txs (so people can sign things with ledgers (not that anyone should be doing that with an orchestrator lolol))

SweeXordious: it's way better :+1: , I will add the changes shortly

SweeXordious:
After reflecting upon:
https://github.com/celestiaorg/celestia-app/pull/878#discussion_r998492347
I guess it is not worth it to have a parser, we can do everything more efficiently in place.
Thus, I will close this PR and update the relevant code locally before pushing a new iteration.
Thanks a lot guys for all the help and feedback :pray:

PR #881, opened 2022-10-18
create a new docker image on release branch

SweeXordious:
Do we want also to release the ephemeral cluster image?


.github/workflows/docker-build.yml
```diff
@@ -4,8 +4,13 @@ on:
   push:
     branches:
       - "main"
+      - "release/*"
```
evan-forbes, line 7: currently, we're using v0.X.X, but I'm okay with switching to release

can we do tags as well?

evan-forbes:
> Do we want also to release the ephemeral cluster image?

not a requirement atm, but yeah we can

.github/workflows/docker-build.yml
evan-forbes, line 7:
```suggestion
      - "release/*"
      tags:
      - "v[0-9]+.[0-9]+.[0-9]+"
      - "v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
      - "v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+"
```

evan-forbes: the formatting above might not be correct, but something similar

.github/workflows/docker-build.yml
```diff
@@ -31,11 +41,23 @@ jobs:
...
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Release name
+        id: release_name
+        uses: actions/github-script@v4
+        with:
+          result-encoding: string
+          script: |
+            const branch_name = process.env.GITHUB_REF
+            const release_version = branch_name.substring(branch_name.lastIndexOf('/') +1 ,branch_name.length  )
```
SweeXordious, line 51: Will need to update this script to also get the version from the tag

.github/workflows/docker-build.yml
```diff
@@ -4,8 +4,18 @@ on:
   push:
     branches:
       - "main"
+      - "release/*"
```
SweeXordious, line 7:
```suggestion
      - "release/v0.[0-9]+.[0-9]+*"
```
@evan-forbes How does this look like?

evan-forbes (approved):
let's test it out

.github/workflows/docker-build.yml
evan-forbes: yeah, makes sense

.github/workflows/docker-build.yml
evan-forbes: how do we do that? just using tags and not release branches is also fine I think if that makes it easier

.github/workflows/docker-build.yml
```diff
@@ -31,11 +41,36 @@ jobs:
...
+      - name: tag version
+        uses: nowsprinting/check-version-format-action@v3
+        id: tag_version
+        with:
+          prefix: 'v'
+
+      - name: tag version extract
+        run: |
+          process.env.VERSION = ${{ steps.version.outputs.full }}
+        if: steps.tag_version.outputs.is_valid == 'true'
```
SweeXordious, line 54: If the `nowsprinting/check-version-format-action@v3` was able to get the tag version and is valid, then we will use that one. If not, we will get it from the branch name.
@evan-forbes what do you think? should we test this?

.github/workflows/docker-build.yml
SweeXordious: We  can remove the branch release then, [da72c7b](https://github.com/celestiaorg/celestia-app/pull/881/commits/da72c7bd8dd39d2cb3219d21e039b34473494195) does them both, but  I guess it's a bit complicated.  What do you think? should I remove getting the version from the branch name?

evan-forbes (approved):
lets test

PR #884, opened 2022-10-18
.github: combine workflows, use refactored actions, align execution

MSevey:
This PR restructures the CI to be more unified, with (more or less) one single file that captures the execution. 

With this format, it drive the process of ensuring all linting and testing is passing before key actions like releases or docker builds are generated.


.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          version-bump: ${{inputs.version}}
+
+  # TODO: include docker build, Josh currently working on updates.
+  #
+  # The below if statement is a recommendation of how to only build the docker
+  # image on manual workflow triggers and push events for the tag
+  #
+  # docker-build:
```
MSevey, line 74: @jbowen93 leaving a call out here for your docker build updates, just fyi

.github/workflows/clean-artifacts.yml
```diff
@@ -1,17 +1,18 @@
+# Pretty sure this isn't needed, see attached screen shot of repo settings
```
MSevey, line 1: Is this file doing something different than this repo setting? if not we can remove this file and just update the repo setting. 

Screen Shot 2022-10-18 at 2 58 49 PM

.github/workflows/linkchecker.yml
```diff
@@ -1,6 +1,8 @@
+# TODO (matt): refactor out
```
MSevey, line 1: see https://github.com/celestiaorg/.github/pull/3

.github/workflows/stale.yml
```diff
@@ -1,3 +1,4 @@
+# TODO: Refactor to common workflow
```
MSevey, line 1: see https://github.com/celestiaorg/.github/pull/3

.github/workflows/tag.yml
```diff
@@ -7,6 +7,8 @@ on:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
 
+# TODO: Is this action doing anything beside auto generating the Github release?
```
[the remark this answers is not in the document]
evan-forbes: yeah, I think we can remove it

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -7,9 +7,10 @@ body:
   - type: markdown
     attributes:
       value: |
-        IMPORTANT: Prior to opening a bug report, check if it affects one of the core modules
-        and if its eligible for a bug bounty on `SECURITY.md`. Bugs that are not submitted
-        through the appropriate channels won't receive any bounty.
+        IMPORTANT: Prior to opening a bug report, check if it affects one of the
+        core modules and if its eligible for a bug bounty on `SECURITY.md`. Bugs
```
SweeXordious, line 11:
```suggestion
        core modules and if it's eligible for a bug bounty on `SECURITY.md`. Bugs
```

.github/workflows/test.yml
```diff
@@ -142,6 +140,7 @@ jobs:
           cat ./*profile.out | grep -v "mode: atomic" >> coverage.txt
         if: env.GIT_DIFF
       - name: filter out DONTCOVER
+        # yamllint disable
```
SweeXordious, line 143: [Non Blocking][Suggestion]
Same as above for formatting correctly the embedded shell script

.github/workflows/proto.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: "Check protobuf generated code matches committed code"
+        # yamllint disable
```
SweeXordious, line 22: [Non-Blocking][Suggestion]
Instead of disabling yaml lint, probably fix the indentation like mentionned here: https://stackoverflow.com/a/69825610/11529598

.github/workflows/test.yml
```diff
@@ -1,19 +1,19 @@
...
+
 jobs:
   cleanup-runs:
     runs-on: ubuntu-latest
     steps:
       - uses: rokroskar/workflow-run-cleanup-action@master
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
+    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'" # yamllint disable-line rule:line-length
```
SweeXordious, line 16: [Non-Blocking][Suggestion]
Doesn't it make sense to increase the line length in the configuration since multiple yaml files are having this issue, instead of having to disable it everytime?

SweeXordious (commented):
LGTM, but I will defer to the other reviewers, I don't have enough knowledge on this.
Left some non blocking questions/suggestions

.github/workflows/ci_release.yml
evan-forbes: [non blocking][questions]
can we create an issue for this? I don't think the ephemeral cluster is using the docker builds as we just added in #881 and #882, should they be using the same? ref https://github.com/celestiaorg/.github/pull/4

The ones we just added should build when pushing matching tags, so I don't think we have to do anything there. 

possibly stupid question, will we be using this https://github.com/celestiaorg/.github/blob/main/.github/actions/docker-publish/action.yml ?

evan-forbes (commented):
can we also delete https://github.com/celestiaorg/celestia-app/blob/sevey/ci-release/.github/workflows/stale.yml?

.github/workflows/test.yml
MSevey: The issue is line length, and I don't believe I can break up the lines that are violating the rule, which is why I did it this way.

.github/workflows/proto.yml
MSevey: That is a different issue, that stackoverflow is asking about invalid workflow, which is Github linting the workflow, not `yamllint`.

Yamllint was failing for line length. I choose to disable the lint for these blocks because I didn't want to mess with the code block for something basic like line length.

MSevey: same response.

.github/workflows/test.yml
MSevey: not imo, the line length is nice for readability.  Where possible I simply fixed the line to comply. There were a few places that I didn't want to touch the code to avoid opening a can of worms.

MSevey:
> possibly stupid question, will we be using this https://github.com/celestiaorg/.github/blob/main/.github/actions/docker-publish/action.yml ?

Not a stupid question :-) 

That was a docker action I've used in the past, but it is set to push to DockerHub.  Looking at the docker build workflows in the various Celestia repos, it looks like we are pushing to the github docker image repository. So we could use it, but we would need to update it, so it is probably easiest to just replace it with the ones we are using. 

I'll link to this response in the issue I created as well.

.github/workflows/clean-artifacts.yml
MSevey: @evan-forbes thought on removing the workflow and just using this repo setting?  Do you remember why this workflow was added?

Dockerfile
```diff
@@ -1,13 +1,13 @@
 # stage 1 Generate celestia-appd Binary
 FROM golang:1.18-alpine as builder
-RUN apk update && apk --no-cache add make gcc musl-dev
+RUN apk update && apk --no-cache add gcc=latest musl-dev=latest
```
SweeXordious, line 3: [Non Blocking][Question]
> By default apk tries to select a set of packages so that all dependencies are satisfied. This means that some packages can be held back in an older version if there is direct or implicit dependencies to the older version.
https://man.archlinux.org/man/apk-add.8.en

I guess it's better not to force `latest` as it might not be compatible with the already installed packages in the base image. What do you think?

SweeXordious (commented):
Left questions concerning the dockerfile versions

evan-forbes (approved):
pre-approving, but I think the questions Rachid asked are worth answering before merging (so I disabled auto merge this time)

Dockerfile
MSevey: So these were `WARNINGS` about not specifying a version. In general it is best practice to use specific versions but `apk` is a nice tool.

If we merge this https://github.com/celestiaorg/.github/pull/8 I can revert these changes.

Dockerfile
```diff
@@ -1,13 +1,13 @@
...
 COPY . /celestia-app
 WORKDIR /celestia-app
 RUN make build
 
 # stage 2
-FROM alpine
-RUN apk update && apk --no-cache add bash
+FROM alpine:3.16
```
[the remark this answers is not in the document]
MSevey: same answer as above.

Dockerfile
SweeXordious: Understandable. I am fine with whatever you prefer :cat:

SweeXordious (dismissed):

Dockerfile
MSevey: Went with ignoring this rule inline in favor of `apk` compatibility.

Dockerfile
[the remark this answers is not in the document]
MSevey: Added dependabot docker to handle this.

.github/workflows/clean-artifacts.yml
evan-forbes: yeah, I think we can delete this

evan-forbes: It was likely copied over from the sdk or default from startport

evan-forbes (approved):

SweeXordious (approved):

PR #892, opened 2022-10-20
chore: default rpc params

### LINKED ISSUE #256: Increase timeouts in cli to match long block times
Currently, when using the `--broadcast-mode block` flag, the block times on a live network are too long, resulting the cli almost always timing out. We need to communicate this to the user in some way, along with increasing timeouts to be useful

### LINKED ISSUE #591: config/rpc: define default parameters
Introduction ✌🏻
---
Here is the current params we have for rpc part (I've removed some params that is indeed highly specific for each operator team)
```toml

# Maximum number of simultaneous connections.
# Does not include RPC (HTTP&WebSocket) connections. See max_open_connections
# If you want to accept a larger number than the default, make sure
# you increase your OS limits.
# 0 - unlimited.
# Should be < {ulimit -Sn} - {MaxNumInboundPeers} - {MaxNumOutboundPeers} - {N of wal, db and other open files}
# 1024 - 40 - 10 - 50 = 924 = ~900
grpc_max_open_connections = 900

# Activate unsafe RPC commands like /dial_seeds and /unsafe_flush_mempool
unsafe = false

# Maximum number of simultaneous connections (including WebSocket).
# Does not include gRPC connections. See grpc_max_open_connections
# If you want to accept a larger number than the default, make sure
# you increase your OS limits.
# 0 - unlimited.
# Should be < {ulimit -Sn} - {MaxNumInboundPeers} - {MaxNumOutboundPeers} - {N of wal, db and other open files}
# 1024 - 40 - 10 - 50 = 924 = ~900
max_open_connections = 900

# Maximum number of unique clientIDs that can /subscribe
# If you're using /broadcast_tx_commit, set to the estimated maximum number
# of broadcast_tx_commit calls per block.
max_subscription_clients = 100

# Maximum number of unique queries a given client can /subscribe to
# If you're using GRPC (or Local RPC client) and /broadcast_tx_commit, set to
# the estimated # maximum number of broadcast_tx_commit calls per block.
max_subscriptions_per_client = 5

# How long to wait for a tx to be committed during /broadcast_tx_commit.
# WARNING: Using a value larger than 10s will result in increasing the
# global HTTP write timeout, which applies to all connections and endpoints.
# See https://github.com/tendermint/tendermint/issues/3435
timeout_broadcast_tx_commit = "10s"

# Maximum size of request body, in bytes
max_body_bytes = 1000000

# Maximum size of request header, in bytes
max_header_bytes = 1048576
```

New Definition 📜 
---

- We should change `timeout_broadcast_tx_commit` to 40 or 50 seconds as the `timeout_commit` is 30 seconds
- We have `mempool.max_tx_bytes` equals to 1048576 and the question here is for `max_body_bytes` and `max_header_bytes`
- Are we fine with those 2 figures as they are rn? 

Notes 📝 
---
Ref: #585

rahulghangas:
Configure default rpc params to factor in large txns and long block times

- [x] Closes #256 
- [x] Closes #591


cmd/celestia-appd/cmd/root.go
```diff
@@ -77,7 +78,13 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
+			// change tendermint config for celestia app
+			// set broadcast timeout to be 50 seconds in order to factor in long block times
+			// set max body bytes for transaction to be 8 Mb
+			// TODO: maybe we should make this timeout configurable?
 			tmCfg := tmcfg.DefaultConfig()
+			tmCfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
```
SweeXordious, line 86: [Non Blocking][Question for reviewers]
Even in the latest networks, we're seeing inconsistent block times. Why not increase this value to more? Like 5 min or so?
Or, even pass it as a parameter to the command, with the default being long enough.

SweeXordious (approved):
LGTM :+1: 
Left non blocking question

cmd/celestia-appd/cmd/root.go
```diff
@@ -77,7 +78,13 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
+			// change tendermint config for celestia app
+			// set broadcast timeout to be 50 seconds in order to factor in long block times
+			// set max body bytes for transaction to be 8 Mb
+			// TODO: maybe we should make this timeout configurable?
```
rootulp, line 84: [follow-up issue]

> If my understanding is correct, timeouts aren't configurable through the cli

I think you're right based on no help message for it in:

```bash
celestia-appd help
```

Screen Shot 2022-10-20 at 9 39 30 AM


```bash
$ celestia-appd start --rpc.max_body_bytes 300
Error: unknown flag: --rpc.max_body_bytes
```

but I think this would be a good follow-up issue especially if (as @SweeXordious points out) we have inconsistent block times so the 50 seconds may need to be modified.

rootulp (commented):
One blocking question

cmd/celestia-appd/cmd/root.go
evan-forbes: ++ yeah this is new `good first issue`

cmd/celestia-appd/cmd/root.go
```diff
@@ -77,7 +78,13 @@ func NewRootCmd() *cobra.Command {
...
 			tmCfg := tmcfg.DefaultConfig()
+			tmCfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
+			tmCfg.RPC.MaxBodyBytes = int64(8000000)
```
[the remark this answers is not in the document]
evan-forbes: I'm unsure if we want the default to be the max, but if that's the case, perhaps we can use existing constants

evan-forbes (approved):
[optional]
perhaps we can put these defaults somewhere a bit more explicit. not exactly sure where

evan-forbes:
what's left here? just he above reviewer feedback?

evan-forbes:
bump @rahulghangas

rahulghangas:
I've left the timeout value as is for now, creating a new issue to make it configurable

cmd/celestia-appd/cmd/root.go
```diff
@@ -77,7 +78,13 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
+			// change tendermint config for celestia app
+			// set broadcast timeout to be 50 seconds in order to avoid timeouts for long block times
+			// set max body bytes for transaction to be 8 Mb
```
rootulp, line 83: Note this comment is still inaccurate ` 8 Mb` => `8 MiB`. On re-review, the comment doesn't provide any information that isn't evident in the of code so proposal to remove it entirely

```
			tmCfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
```

PR #893, opened 2022-10-20
feat: add payment module params

### LINKED ISSUE #183: Add the `Max` and `MinSquareSize` as a parameter to the payment module
Currently, the `MaxSquareSize` constant has to be in celestia-core, but ideally, it should be a parameter in the app. This way, it doesn't require a hardfork to change, we only need 2/3 of the voting power to vote for a proposal using the gov module.

As celestia-core might need that constant and we cannot have import cycles, we can either manually mirror that parameter in celestia-core, or after we begin using ABCI++. It is possible that celestia-core will no longer need to know that constant, because the filling of the data square could happen entirely on the application side.

releated to #168

rahulghangas:
Adds `minSquareSize` and `maxSquareSize` as params to the payment module. 
Defines relevant stores and queries

Relevant changes in 
`proto/payment/*`
`x/payment/keeper*`
`x/payment/types/*`

- [x] closes #183 

Note: The constants that currently define min/max square size have not been deprecated, will do so in another PR

### DIFF
--- a/x/blob/types/params.go
+++ b/x/blob/types/params.go
@@ -104,7 +104,7 @@ func validateMaxSquareSize(v interface{}) error {
 
 func validateMinMaxSquareSizeOrder(minSquareSize, maxSquareSize uint32) error {
 	if minSquareSize > maxSquareSize {
-		return fmt.Errorf("min square size cannot be less than max square size")
+		return fmt.Errorf("max square size cannot be less than min square size")
 	}
 	return nil
 }


proto/payment/params.proto
```diff
@@ -0,0 +1,14 @@
...
+
+option go_package = "github.com/celestiaorg/celestia-app/x/payment/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  int32 minSquareSize = 1 [(gogoproto.moretags) = "yaml:\"min_square_size\""];
+  int32 maxSqaureSize = 2 [(gogoproto.moretags) = "yaml:\"max_sqaure_size\""];
+}
```
SweeXordious, line 14:
```suggestion
}

```

proto/payment/params.proto
```diff
@@ -0,0 +1,14 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/payment/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  int32 minSquareSize = 1 [(gogoproto.moretags) = "yaml:\"min_square_size\""];
+  int32 maxSqaureSize = 2 [(gogoproto.moretags) = "yaml:\"max_sqaure_size\""];
```
SweeXordious, line 13:
```suggestion
  int32 maxSquareSize = 2 [(gogoproto.moretags) = "yaml:\"max_square_size\""];
```
probably this change will require regenerating proto files and doing corresponding updates

proto/payment/query.proto
```diff
@@ -1,12 +1,24 @@
...
 option go_package = "github.com/celestiaorg/celestia-app/x/payment/types";
 
 // Query defines the gRPC querier service.
 service Query {
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/payment/params";
+  }
 }
 
+// QueryParamsRequest is request type for the Query/Params RPC method.
```
SweeXordious, line 18:
```suggestion
// QueryParamsRequest is the request type for the Query/Params RPC method.
```

proto/payment/query.proto
```diff
@@ -1,12 +1,24 @@
...
+message QueryParamsRequest {}
+
+// QueryParamsResponse is response type for the Query/Params RPC method.
```
SweeXordious, line 21:
```suggestion
// QueryParamsResponse is the response type for the Query/Params RPC method.
```

x/payment/keeper/params.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/celestiaorg/celestia-app/x/payment/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GetParams get all parameters as types.Params
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	return types.NewParams(
+		k.MinSquareSize(ctx),
+		k.MaxSqaureSize(ctx),
```
SweeXordious, line 12:
```suggestion
		k.MaxSquareSize(ctx),
```

x/payment/keeper/params.go
```diff
@@ -0,0 +1,31 @@
...
+
+// MinSquareSize returns the MinSquareSize param
+func (k Keeper) MinSquareSize(ctx sdk.Context) (res int32) {
+	k.paramstore.Get(ctx, types.KeyMinSquareSize, &res)
+	return
+}
+
+// MaxSqaureSize returns the MaxSqaureSize param
+func (k Keeper) MaxSqaureSize(ctx sdk.Context) (res int32) {
+	k.paramstore.Get(ctx, types.KeyMaxSqaureSize, &res)
```
SweeXordious, line 29:
```suggestion
	k.paramstore.Get(ctx, types.KeyMaxSquareSize, &res)
```

x/payment/keeper/params.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// MinSquareSize returns the MinSquareSize param
+func (k Keeper) MinSquareSize(ctx sdk.Context) (res int32) {
+	k.paramstore.Get(ctx, types.KeyMinSquareSize, &res)
+	return
+}
+
+// MaxSqaureSize returns the MaxSqaureSize param
+func (k Keeper) MaxSqaureSize(ctx sdk.Context) (res int32) {
```
SweeXordious, line 28:
```suggestion
func (k Keeper) MaxSquareSize(ctx sdk.Context) (res int32) {
```

x/payment/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+var _ paramtypes.ParamSet = (*Params)(nil)
+
+var (
+	KeyMinSquareSize = []byte("MinSquareSize")
+	// TODO: Determine the default value
+	DefaultMinSquareSize int32 = 0
+)
+
+var (
+	KeyMaxSqaureSize = []byte("MaxSqaureSize")
```
SweeXordious, line 19:
```suggestion
	KeyMaxSquareSize = []byte("MaxSquareSize")
```

x/payment/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// NewParams creates a new Params instance
+func NewParams(
+	minSquareSize int32,
+	maxSqaureSize int32,
+) Params {
+	return Params{
+		MinSquareSize: minSquareSize,
+		MaxSqaureSize: maxSqaureSize,
```
SweeXordious, line 36: [Blocking][typo]
All `MaxSqaureSize` => `MaxSquareSize`

x/payment/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// validateMinSquareSize validates the MinSquareSize param
+func validateMinSquareSize(v interface{}) error {
+	minSquareSize, ok := v.(int32)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", v)
+	}
+
+	// TODO implement validation
```
SweeXordious, line 82: [Blocking][Question]
Is this intended to be defined in this PR?

x/payment/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// validateMaxSqaureSize validates the MaxSqaureSize param
+func validateMaxSqaureSize(v interface{}) error {
+	maxSqaureSize, ok := v.(int32)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", v)
+	}
+
+	// TODO implement validation
```
SweeXordious, line 95: [Blocking][Question]
Same

SweeXordious (commented):
Good start :+1: 
Did a first pass and left some blocking comments.
Also, doesn't it make sense also to make the corresponding genesis changes?
I once set some parameters for the QGB, and I updated the following when  `init` and `export`:
https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/genesis.go

x/payment/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+	DefaultMinSquareSize int32 = 0
+)
+
+var (
+	KeyMaxSquareSize = []byte("MaxSquareSize")
+	// TODO: Determine the default value
+	DefaultMaxSquareSize int32 = 0
+)
+
+// ParamKeyTable the param key table for launch module
```
rootulp, line 24: [question] what is launch in this context? Perhaps:

```suggestion
// ParamKeyTable the param key table for the payment module
```

x/payment/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+
+// DefaultParams returns a default set of parameters
+func DefaultParams() Params {
+	return NewParams(
+		DefaultMinSquareSize,
+		DefaultMaxSquareSize,
+	)
+}
+
+// ParamSetPairs get the params.ParamSet
```
rootulp, line 48:
```suggestion
// ParamSetPairs gets the ParamSetPairs
```

rootulp (changes requested):
- golangci-lint is failing [here](https://github.com/celestiaorg/celestia-app/actions/runs/3290219599/jobs/5422767195#step:5:34) (likely fixable via `make fmt`)
- proto-gen is failing [here](https://github.com/celestiaorg/celestia-app/actions/runs/3290219607/jobs/5422767428#step:3:85) (likely fixable via `make proto-gen`)

x/payment/keeper/params.go
```diff
@@ -0,0 +1,31 @@
...
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	return types.NewParams(
+		k.MinSquareSize(ctx),
+		k.MaxSquareSize(ctx),
+	)
+}
+
+// SetParams set the params
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
+	k.paramstore.SetParamSet(ctx, &params)
```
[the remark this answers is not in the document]
rootulp: ^ sorry this suggestion doesn't make sense because the type is defined in cosmos-sdk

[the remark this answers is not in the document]
rahulghangas: The subspace method is named `SetParamSet`, not in my control

proto/payment/params.proto
```diff
@@ -0,0 +1,14 @@
...
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/payment/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  int32 min_square_size = 1 [(gogoproto.moretags) = "yaml:\"min_square_size\""];
```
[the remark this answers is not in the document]
rahulghangas: I thought the same, but the current consts are integers, so I use `int` instead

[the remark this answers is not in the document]
rahulghangas: changed to uint32

x/payment/client/cli/query.go
```diff
@@ -24,5 +24,7 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 		RunE:                       client.ValidateCmd,
 	}
 
+	cmd.AddCommand(CmdQueryParams())
```
evan-forbes, line 27: nice, thanks for handling this in this PR

rootulp (commented):
Thanks for addressing feedback so quickly!

I'm observing the following panic when attempting to test the CLI query command. 

```
$ celestia-appd query payment params
Error: rpc error: code = Unknown desc = parameter MinSquareSize not registered: panic
```

To reproduce this error, I checked out this branch locally and installed the binary (i.e `make install`)

proto/payment/params.proto
```diff
@@ -0,0 +1,14 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/payment/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  uint32 min_square_size = 1 [(gogoproto.moretags) = "yaml:\"min_square_size\""];
+  uint32 max_square_size = 2 [(gogoproto.moretags) = "yaml:\"max_square_size\""];
```
evan-forbes, line 13: possibly stupid question, why are we using the yaml here instead of something like jsontag? this seems to be uncommon in the sdk afaiu

proto/payment/params.proto
```diff
@@ -0,0 +1,14 @@
...
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/payment/types";
+
+// Params defines the parameters for the module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  uint32 min_square_size = 1 [(gogoproto.moretags) = "yaml:\"min_square_size\""];
```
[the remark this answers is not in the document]
rahulghangas: It's standard practice from what I've seen. Also, it makes it very easy to define genesis parameters through a `config.yml` file

For eg. https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/lockup/params.proto

rahulghangas:
> I'm observing the following panic when attempting to test the CLI query command.
> 
> ```
> $ celestia-appd query payment params
> Error: rpc error: code = Unknown desc = parameter MinSquareSize not registered: panic
> ```
> 
> To reproduce this error, I checked out this branch locally and installed the binary (i.e `make install`)

Whoops, fixing this

x/payment/keeper/grpc_query_params.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+func (k Keeper) Params(c context.Context, req *types.QueryParamsRequest) (*types.QueryParamsResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+	ctx := sdk.UnwrapSDKContext(c)
+
+	return &types.QueryParamsResponse{Params: k.GetParams(ctx)}, nil
+}
```
evan-forbes, line 19: [doesn't have to be done in this PR][new good first issue]
Its might be a bit extravagant to have this level of test coverage, but we can even include a test for querying these params in cli testutil

evan-forbes (commented):
Nice!! these will be really useful in the future.

> The constants that currently define min/max square size have not been deprecated, will do so in another PR

IMO, we should actually keep these, but just change them to be the defaults. This way other projects import really similar defaults without booting up an application.

x/payment/keeper/grpc_query_params.go
evan-forbes: could also have a test to update these params, but again, that's kinda extravagant.

x/payment/types/params.go
```diff
@@ -0,0 +1,91 @@
...
+
+var (
+	KeyMinSquareSize            = []byte("MinSquareSize")
+	DefaultMinSquareSize uint32 = 1
+)
+
+var (
+	KeyMaxSquareSize            = []byte("MaxSquareSize")
+	DefaultMaxSquareSize uint32 = 128
+)
```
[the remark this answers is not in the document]
rahulghangas: That's exactly what I had in mind, but in another PR. Convert the consatnts into defaults, and use `appconsts.xyz` instead of hardcoded values here

proto/payment/params.proto
rahulghangas: > It's standard practice from what I've seen. Also, it makes it very easy to define genesis parameters through a `config.yml` file
>
>For eg. https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/lockup/params.proto

rahulghangas: Can go with jsontag if that's what we prefer

proto/payment/params.proto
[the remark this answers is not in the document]
rootulp: I can't find a `config.yml` in https://github.com/osmosis-labs/osmosis so I'm not sure what you mean. Can you please elaborate or include a link to the file?

proto/payment/params.proto
evan-forbes: yml makes sense too! I don't think it matters, as long as the genesis state is working

x/payment/keeper/grpc_query_params.go
SweeXordious: [Blocking]

> but we can even include a test for querying these params in cli testutil

:+1: 

I guess it is really important to test it either using a test of confirming that the params are queried correctly from the CLI.
I had some fun errors when querying params before for the QGB, even if the implementation looked right. Thus, I would prefer if we can get a confirmation on this.

rahulghangas: Done, please re-review

testutil/nullify/nullify.go
```diff
@@ -0,0 +1,57 @@
+// Package nullify provides methods to init nil values structs for test assertion.
```
rootulp, line 1: [nit]
```suggestion
// Package nullify provides methods to init nil value structs for test assertion.
```

testutil/nullify/nullify.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+var (
+	coinType  = reflect.TypeOf(sdk.Coin{})
+	coinsType = reflect.TypeOf(sdk.Coins{})
+)
+
+// Fill analyze all struct fields and slices with
+// reflection and initialize the nil and empty slices,
+// structs, and pointers.
```
rootulp, line 18: [nit] is it possible to improve the grammar here? I'm not sure what this meant to say so can't provide a meaningful suggestion

x/payment/genesis_test.go
```diff
@@ -0,0 +1,25 @@
...
+		Params: types.DefaultParams(),
+	}
+
+	k, ctx := keepertest.PaymentKeeper(t)
+	payment.InitGenesis(ctx, *k, genesisState)
+	got := payment.ExportGenesis(ctx, *k)
+	require.NotNil(t, got)
+
+	nullify.Fill(&genesisState)
+	nullify.Fill(got)
```
rootulp, line 24: [question] There aren't any assertions made after these lines and the test passes without them so I'm curious why they are necessary?

Since the nullify package introduces confusing reflection and unsafe code, is it possible to delete entirely?

x/payment/keeper/keeper.go
```diff
@@ -8,19 +8,36 @@ import (
...
 
 const payForDataGasDescriptor = "pay for data"
 
 // Keeper handles all the state changes for the celestia-app module.
 type Keeper struct {
-	cdc codec.BinaryCodec
+	cdc        codec.BinaryCodec
+	storeKey   storetypes.StoreKey
+	memKey     storetypes.StoreKey
+	paramstore paramtypes.Subspace
```
rootulp, line 23: [uber nit][optional] proposal to camelCase this variable globally because the variable is composed of two words. The proposal conforms to the naming convention used for the other fields in this struct (i.e. `storeKey` and `memKey`)

```suggestion
	paramStore paramtypes.Subspace
```

x/payment/keeper/params.go
```diff
@@ -0,0 +1,31 @@
+package keeper
+
+import (
+	"github.com/celestiaorg/celestia-app/x/payment/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GetParams get all parameters as types.Params
```
rootulp, line 8:
```suggestion
// GetParams gets all parameters as types.Params
```

x/payment/types/params.go
```diff
@@ -0,0 +1,91 @@
...
+
+// ParamKeyTable the param key table for the payment module
```
rootulp, line 22:
```suggestion
// ParamKeyTable returns the param key table for the payment module
```

x/payment/types/params.go
```diff
@@ -0,0 +1,91 @@
...
+
+// DefaultParams returns a default set of parameters
+func DefaultParams() Params {
+	return NewParams(
+		DefaultMinSquareSize,
+		DefaultMaxSquareSize,
+	)
+}
+
+// ParamSetPairs gets the list of param key-pair values
```
rootulp, line 46:
```suggestion
// ParamSetPairs gets the list of param key-value pairs
```

rootulp (commented):
In one terminal, I ran

```shell
make install
celestia-appd start
```

and in a separate terminal, I'm observing a new error now:

```
$ celestia-appd query payment params
Error: rpc error: code = Unknown desc = UnmarshalJSON cannot decode empty bytes: panic
```
@rahulghangas are you able to reproduce this?

evan-forbes:
@rootulp 

> In one terminal, I ran
make install
celestia-appd start
and in a separate terminal, I'm observing a new error now:

did you initialize the chain from scratch? the only way to set genesis values after upgrading binaries is to include it in an upgrade handler or perhaps vote on it. In a normal situation, it will be included in the genesis.

rootulp:
> did you initialize the chain from scratch?

Nope. Thanks for identifying my mistake. I just did and this command now works as expected

```
$ celestia-appd query payment params
params:
  max_square_size: 128
  min_square_size: 1
```

rootulp:
[optional][can be a follow-up PR] we likely want to update https://github.com/celestiaorg/celestia-app/blob/main/x/payment/spec/docs.md#parameters

x/payment/genesis_test.go
rahulghangas: It's a check to make sure that they are copied, iirc

x/payment/types/params.go
```diff
@@ -0,0 +1,91 @@
...
+
+// String implements the Stringer interface.
+func (p Params) String() string {
+	out, _ := yaml.Marshal(p)
+	return string(out)
+}
+
+// validateMinSquareSize validates the MinSquareSize param
+func validateMinSquareSize(v interface{}) error {
+	_, ok := v.(uint32)
```
SweeXordious, line 75: [Blocking]
Doesn't it make sense to check min square size not to be 0? or >= DefaultMinSquareSize ?

testutil/nullify/nullify.go
```diff
@@ -0,0 +1,57 @@
...
+
+var (
+	coinType  = reflect.TypeOf(sdk.Coin{})
+	coinsType = reflect.TypeOf(sdk.Coins{})
+)
+
+// Fill analyzes all struct fields and slices with
+// reflection and initializes the nil and empty slices,
+// structs, and pointers.
+func Fill(x interface{}) interface{} {
```
SweeXordious, line 19: [Non Blocking]
I don't like having unsafe code on the state machine side. Isn't there a way not to use this?

SweeXordious (changes requested):

x/payment/genesis_test.go
rootulp: Sorry but I don't understand. Can you please elaborate? To make sure what is copied? Happy to hop on a call if you'd prefer that

x/payment/types/params.go
evan-forbes: Good catch! 0 is invalid

testutil/nullify/nullify.go
rootulp: Agreed. The only usage of `Fill` appears to be in a test and I think the test can remove its usage ref: https://github.com/celestiaorg/celestia-app/pull/893#discussion_r1004230693 so if possible, I think we should remove this file entirely.

x/payment/genesis_test.go
SweeXordious: Following the comment in here https://github.com/celestiaorg/celestia-app/pull/893#discussion_r1010285425 (don't want to discuss the test in two separate places). 
I guess what this test is doing is checking whether the exported genesis is correct.
And considering the parameters that we're passing to it:
```go
        genesisState := types.GenesisState{
		Params: types.DefaultParams(),
	}

	k, ctx := keepertest.PaymentKeeper(t)
}
```
And the expected behavior for the genesis export is to have the default genesis (which is already tested in other modules we're importing) + payment params that we just set.
I guess, we will only need the following assertion:
```go
require.Equal(t, defaultParams, got.Params)
```
And, we can get rid of the `Fill` usage

SweeXordious (changes requested):

testutil/nullify/nullify.go
rahulghangas: Done

x/payment/types/params.go
rahulghangas: Done

x/payment/genesis_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+func TestGenesis(t *testing.T) {
+	genesisState := types.GenesisState{
+		Params: types.DefaultParams(),
+	}
+
+	k, ctx := keepertest.PaymentKeeper(t)
+	payment.InitGenesis(ctx, *k, genesisState)
+	got := payment.ExportGenesis(ctx, *k)
+	require.NotNil(t, got)
```
SweeXordious, line 20:
```suggestion
	require.NotNil(t, got)
	require.Equal(t, types.DefaultParams(), got.Params)
```

SweeXordious (approved):
LGTM, would defer to other reviewers approval.

x/payment/genesis_test.go
rahulghangas: Sorry missed this

SweeXordious:
@rahulghangas The failing test if probably because of this:
https://github.com/rahulghangas/celestia-app/blob/6fb4e2fff5c3598e572401d92d62b80946c910b6/x/payment/types/genesis_test.go#L23
The minimum square size needs to be set

SweeXordious (approved):
LGTM, would defer to others for final decision

SweeXordious (dismissed):
:shipit:

evan-forbes:
@rootulp is there any remaining things here or can we merge?

SweeXordious:
@evan-forbes probably just the merge conflict, regeneration of the proto files, then would be ready to shit

rahulghangas:
Proto files have already been regenerated, I'll double check. I'll also resolve the merge conflict

rootulp (approved):

x/blob/types/params.go
```diff
@@ -0,0 +1,99 @@
...
+// Validate validates the set of params
+func (p Params) Validate() error {
+	if err := validateMinSquareSize(p.MinSquareSize); err != nil {
+		return err
+	}
+
+	if err := validateMaxSquareSize(p.MaxSquareSize); err != nil {
+		return err
+	}
+
```
SweeXordious, line 63: Probably also add a `minSquareSize > maxSquareSize` check and error in that case. Not sure if this is covered somewhere else tho.

testutil/keeper/payment.go
```diff
@@ -0,0 +1,53 @@
+package keeper
```
SweeXordious, line 1: probably rename this file to `blob.go` or  `keeper.go`

SweeXordious (commented):
LGTM, just some final comments

evan-forbes (approved):
lets mergeeeeeeee

Issue #897, opened 2022-10-20
single-node.sh fails with `sed: -I or -i may not be used with stdin`


rootulp:
## Problem
I'm observing an error when running the single node devnet script:

```sh
$ ./scripts/single-node.sh
...
sed: -I or -i may not be used with stdin
```

I'm on macOS Monterey Version 12.6. and I haven't configured a different sed

```
$ which sed
/usr/bin/sed
```

## Proposal
I found two options to resolve
1. Install gnu-sed via Homebrew per the steps on this [gist comment](https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272)
2. Add a zero-length extension per this [StackOverflow](https://stackoverflow.com/a/42667816/2813210) (breaks the command on Linux)

PR #898, opened 2022-10-20
fix: link to gnu-sed in single-node script

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/897
This is pretty janky but may save a future dev time.


rootulp:
Broke on Linux
```
sed: can't read s#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g: No such file or directory
```

SweeXordious (approved):

PR #904, opened 2022-10-21
hardcode share index to 0 when calculating compact share count

rootulp:
Closes: https://github.com/informalsystems/audit-celestia/issues/25

FWIW this PR becomes unnecessary if we adopt ADR 008 because then we can likely delete square size estimation (WIP [commit](https://github.com/celestiaorg/celestia-app/commit/c0cf6af3ab715dd514bdf1958dd6f3e1dfcf7549)). I'm on the fence about merging this PR because I don't have full confidence this works in all circumstances. Unit tests pass but I'm concerned about edge cases.

### DIFF
--- a/app/estimate_square_size.go
+++ b/app/estimate_square_size.go
@@ -72,12 +72,12 @@ func calculateCompactShareCount(txs []*parsedTx, evd core.EvidenceList, squareSi
 	for _, tx := range txs {
 		rawTx := tx.rawTx
 		if tx.malleatedTx != nil {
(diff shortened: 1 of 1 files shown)


rootulp:
FWIW this PR becomes unnecessary if we adopt ADR 008 because then we can likely delete square size estimation (WIP [commit](https://github.com/celestiaorg/celestia-app/commit/c0cf6af3ab715dd514bdf1958dd6f3e1dfcf7549))

PR #909, opened 2022-10-24
chore: rename `height` to `treeSizes` in Merkle mountain range

rootulp:
This PR is a pure refactor that renames `heights` => `treeSizes` because the previous name was slightly misleading. This function doesn't return the heights of the trees in the Merkle mountain range, it returns the number of leaves in each tree. For example, for the unit test that was added:

``` golang
			totalSize:  19,
			squareSize: 8,
			expected:   []uint64{8, 8, 2, 1},
```

the Merkle mountain range looks like

```ascii
		// Height
		// 3              x                               x
		//              /    \                         /    \
		//             /      \                       /      \
		//            /        \                     /        \
		//           /          \                   /          \
		// 2        x            x                 x            x
		//        /   \        /   \             /   \        /   \
		// 1     x     x      x     x           x     x      x     x         x
		//      / \   / \    / \   / \         / \   / \    / \   / \      /   \
		// 0   0   1 2   3  4   5 6   7       8   9 10  11 12 13 14  15   16   17    18

```

and if this function returned the heights, then expected would be:

```diff
-expected:   []uint64{8, 8, 2, 1},
+expected:   []uint64{3, 3, 1, 0},
```

but it actually returns `{8, 8, 2, 1}` (the number of leaf nodes in each tree). The remaining renames are optional but may aid clarity if we adopt the rename proposed above

### DIFF
--- a/x/payment/types/payfordata.go
+++ b/x/payment/types/payfordata.go
@@ -138,15 +138,15 @@ func CreateCommitment(squareSize uint64, namespace, message []byte) ([]byte, err
 	}
 
 	// organize shares for merkle mountain range
-	heights := powerOf2MountainRange(uint64(len(shares)), squareSize)
-	leafSets := make([][][]byte, len(heights))
+	treeSizes := merkleMountainRangeSizes(uint64(len(shares)), squareSize)
+	leafSets := make([][][]byte, len(treeSizes))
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):
LGTM makes sense

PR #915, opened 2022-10-25
Update docker-build workflow to fix version tags

jbowen93:
Currently the docker-build workflow attempts to tag releases with 
```
...${{ steps.version.outputs.full }}
```
but the correct string is
```
...${{ steps.tag_version.outputs.full }}
``` 
to match the previous step id of `tag_version`


SweeXordious (approved):
Thanks for addressing this. Was my mistake :cat:

evan-forbes (approved):

PR #917, opened 2022-10-26
feat!: square size independent commitments

rootulp:
### Description
Hackathon implementation of square size independent commitments. Draft PR because still work in progress but wanted to make visible for collaboration. Post hackathon, this PR will likely be deprecated in favor of smaller, reviewable PRs that we can merge into `main`.


rootulp:
This PR is bloated and should be split into:
- Move `TestProcessWirePayForData` to `wirepayfordata_test.go`
- Resolve `// TODO this error message is incorrect. Rename signers => signatures.`

I misinterpreted the proposal. It isn't necessary to use `appconsts.MinSquareSize`, it's possible to determine a message's minimalSquareSize based on the message length. Since a lot of files were modified in the pursuit of propagating that const, will close this PR in favor of a new one.

PR #918, opened 2022-10-26
chore!: rename `SubmitPayForData` => `SubmitWirePayForData`

rootulp:
## Description

Previously, there were functions named `*PayForData` that actually dealt with `WirePayForData`s. Since a PayForData != WirePayForData, I found the previous naming confusing. This change renames functions to more accurately reflect the types they deal with.

This change is breaking because `SubmitPayForData` is part of the public API and is used by celestia-node.

### DIFF
--- a/x/payment/payfordata.go
+++ b/x/payment/payfordata.go
@@ -13,9 +13,9 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 )
 
-// SubmitPayForData constructs, signs and synchronously submits a PayForData
+// SubmitWirePayForData constructs, signs, and synchronously submits a WirePayForData
 // transaction, returning a sdk.TxResponse upon submission.
-func SubmitPayForData(
+func SubmitWirePayForData(
 	ctx context.Context,
 	signer *types.KeyringSigner,
 	conn *grpc.ClientConn,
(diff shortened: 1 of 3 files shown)


x/payment/payfordata.go
```diff
@@ -48,16 +48,16 @@ func SubmitPayForData(
 	return txResp.TxResponse, nil
 }
 
-// BuildPayForData constructs a PayForData transaction.
```
rootulp, line 51: this function doesn't return a transaction, it returns a message

evan-forbes:
This does make sense, but as this is the main user facing function, I'm unsure we want to surface implementation details in the name.

For example, what if we end up getting rid of WirePayForData?

rootulp:
Fair enough. We don't have to merge this as-is considering we may get rid of `WirePayForData` entirely which would make these names considerably less confusing.

If we don't get rid of `WirePayForData` in the medium term, and we don't want to surface implementation details to end users then proposal to rename to some alternative so that we don't keep misnamed functions (e.g. `SubmitData`)

rootulp:
We don't need to change the function name but I noticed we're already leaking this implementation detail to users:

> payForData  Creates a new MsgWirePayForData

from

```shell
$ celestia-appd tx payment
payment transactions subcommands

Usage:
  celestia-appd tx payment [flags]
  celestia-appd tx payment [command]

Available Commands:
  payForData  Creates a new MsgWirePayForData
```

rootulp:
Closing this b/c it will conflict with https://github.com/celestiaorg/celestia-app/pull/978/files and we're not aligned on exposing `MsgWirePayForData`

PR #928, opened 2022-10-28
added a sed compatibility tweak for macOS

S1nus:


rootulp (approved):
Thanks for the fix! I confirmed this works with the `sed` installed by default on MacOS and `gnu-sed` installed via https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272

PR #937, opened 2022-10-30
feat!: square size independent message commitments

rootulp:
Implementation for square size independent message commitments. Note this targets `feat/square-size-independent-commitments` per https://github.com/celestiaorg/celestia-app/issues/941 so that we can unblock review.


.gitignore
```diff
@@ -7,5 +7,6 @@ coverage.txt
 .idea
 .vscode
 tools-stamp
-.vscode
```
SweeXordious, line 10: [Non Blocking]
Probably this should be kept as it will be created when creating debug configs for examples and we don't want those to be committed to

app/estimate_square_size.go
```diff
@@ -170,7 +170,7 @@ func rawShareCount(txs []*parsedTx, evd core.EvidenceList) (txShares, evdShares
 		// compensates for the actual size of the message, and in some cases can
 		// result in some wasted square space or picking a square size that is
 		// too large. TODO: improve by making a more accurate estimation formula
-		txBytes += overEstimateMalleatedTxSize(len(pTx.rawTx), len(pTx.msg.Message), len(pTx.msg.MessageShareCommitment))
+		txBytes += overEstimateMalleatedTxSize(len(pTx.rawTx), len(pTx.msg.Message), 1)
```
SweeXordious, line 173: [Blocking]
```
// overEstimateMalleatedTxSize estimates the size of a malleated tx. The formula it uses will always over estimate.
func overEstimateMalleatedTxSize(txLen, msgLen, sharesCommitments int) int {
	// the malleated tx uses the original txLen to account for meta data from
	// the original tx, but removes the message
	malleatedTxLen := txLen - msgLen
	// we need to ensure that the returned number is at least larger than or
	// equal to the actual number, which is difficult to calculate without
	// actually malleating the tx
	return appconsts.MalleatedTxBytes + appconsts.MalleatedTxEstimateBuffer + malleatedTxLen
}
```
The third argument doesn't seem to be used, why keep it?

app/test/fuzz_abci_test.go
```diff
@@ -25,7 +25,7 @@ func TestFuzzPrepareProcessProposal(t *testing.T) {
 	encConf := encoding.MakeConfig(app.ModuleEncodingRegisters...)
 	signer := testutil.GenerateKeyringSigner(t, testAccName)
 	testApp := testutil.SetupTestAppWithGenesisValSet(t)
-	timer := time.After(time.Second * 30)
+	timer := time.After(time.Second * 10)
```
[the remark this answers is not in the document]
SweeXordious: I don't know the reasoning behind the 30sec value inside the tests, but if it is necessary, you can update the default timeout in vs code: https://stackoverflow.com/questions/73033725/how-to-edit-default-go-test-timeout-in-visual-studio-code
Would defer to Evan if 30s is needed or not.

SweeXordious (commented):
Awesome work :fire: 
Added couple questions, would defer to Evan for approval.

.gitignore
rootulp: Thanks for flagging!

I intended on adding `.__debug_bin` to this file because VSCode creates it when running tests in debug mode. 

However, removing the `.` prefix from `.vscode` was unintentional. I think my editor removed it because there is already a `.vscode` entry on line 8 of this file. I'll remove both all modifications to this file from this PR b/c the `.__debug_bin` modification can be added separately.

app/estimate_square_size.go
rootulp: Great catch, will remove!

rootulp: resolved in https://github.com/celestiaorg/celestia-app/pull/937/commits/b7aa8936b6a1925a662130c8387dd4868c938c9c

.gitignore
rootulp: resolved in https://github.com/celestiaorg/celestia-app/pull/937/commits/4324bd2909f941dcaab758921833a3263ac9b5b5

rootulp:
Sorry I can't comment on https://github.com/celestiaorg/celestia-app/pull/937#discussion_r1010064147 but thanks for the tip! I reverted this test timeout decrease and changed my local VSCode go test timeout config.

SweeXordious (approved):
:fire:

evan-forbes (approved):
had one non-blocking question and a non-blocking comment. I'm really impressed, I don't think this could have been broken up into a smaller chunk! really great work!!

pkg/inclusion/paths.go
```diff
@@ -32,8 +33,11 @@ func calculateCommitPaths(squareSize, start, msgShareLen int) []path {
 		if i == endRow {
 			end = normalizedEndIndex
 		}
-		coord := calculateSubTreeRootCoordinates(maxDepth, start, end)
-		for _, c := range coord {
+
+		subTreeRootMaxHeight := int(math.Log2(float64(types.MinSquareSize(uint64(msgShareLen)))))
```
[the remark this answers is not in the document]
rootulp: Agreed. We may be able to avoid one type conversions if `MinSquareSize` accepts an `int` arg which can be done by adopting generics so will do shortly. I don't immediately see an alternative to avoid the remaining type conversions but open to ideas.

x/payment/types/wirepayfordata.go
```diff
@@ -34,20 +35,19 @@ func NewWirePayForData(namespace, message []byte, sizes ...uint64) (*MsgWirePayF
...
-		commit, err := CreateCommitment(size, namespace, message)
-		if err != nil {
-			return nil, err
-		}
-		out.MessageShareCommitment[i] = ShareCommitAndSignature{SquareSize: size, ShareCommitment: commit}
+	// generate the share commitment
+	squareSize := MsgMinSquareSize(uint64(len(message)))
+	if !shares.IsPowerOfTwo(squareSize) {
+		return nil, fmt.Errorf("invalid square size, the size must be power of 2: %d", squareSize)
+	}
```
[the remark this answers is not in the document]
rootulp: Good catch. Will remove this check and add a unit test to handle the edge case of an empty message b/c I'm concerned that may result in a square size of 0 which may break things

pkg/inclusion/paths.go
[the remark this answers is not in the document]
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/937/commits/671bc800acecfdeef426db580811fc0baeeda037

evan-forbes (approved):

musalbas:
Does this PR preserve the current non-interactive default rules? As mentioned in https://github.com/celestiaorg/celestia-app/issues/941#issuecomment-1305410774, the current non-interactive default rules would be required to ensure that message inclusion proofs remain O(log(n)).

evan-forbes:
@musalbas good question and thanks for checking on this. This PR does not change the existing non-interactive defaults.

This function and its usage remains unmodified https://github.com/rootulp/celestia-app/blob/671bc800acecfdeef426db580811fc0baeeda037/pkg/shares/non_interactive_defaults.go#L39-L80

only code relating to which subtree roots are used to calculate the commitment has been changed.

rootulp:
Will merge this to the feature branch `feat/square-size-independent-commitments` so that I can update docs in a separate PR.

Issue #431, opened 2022-05-21
Charge gas per message share


evan-forbes:
Padding was removed from messages entirely in #419 as it was causing complexity. We can however still pay for padding by increasing the declared size of the message in the `MsgPayForData` and `MsgWirePayForData`. While this will slightly complicate `ValidateBasic`, it is definitely possible.

It's worth noting that paying for padding isn't enforceable. As previously discussed conversations around fee burning, due to tendermint's [deferred execution](https://github.com/tendermint/tendermint/issues/7898) model, enforcing that a message declare the correct padded size is not possible atm. It would be trivial for a block producer to include a `MsgPayForData` transaction that declares the size to be 0 or whatever size they want. This isn't really an issue, as its the same as a block producer including a message with a zero fee. However, adding a check in `ValidateBasic` would stop messages that are not paying for padding from being propagated by nodes running unmodified software.

evan-forbes:
to update from a sync: we still want to include the in message padding when calculating gas costs. This effectively means also including the padding in the last share when calculating gas costs. This means that we will be charging per share used, instead of per byte. 

We will not be including padding from the non-interactive defaults into gas costs.

rootulp:
> we will be charging per share used, instead of per byte

Clarification: I infer this as paying for each byte used by the message shares associated with a PFD (including namespace IDs, info bytes, message length delimiter, and padding). I also infer this as not charging for the amount of transaction shares associated with the PFD.

PR #946, opened 2022-11-02
feat: charge gas per message shares

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/431

### DIFF
--- a/x/payment/keeper/gas_test.go
+++ b/x/payment/keeper/gas_test.go
@@ -35,8 +35,8 @@ func TestPayForDataGas(t *testing.T) {
 		},
 	}
 
+	app := simapp.Setup(t, false)
 	for _, tc := range testCases {
-		app := simapp.Setup(t, false)
 		ctx := app.BaseApp.NewContext(false, tmproto.Header{})
 		k := Keeper{}
 		_, err := k.PayForData(sdk.WrapSDKContext(ctx), &tc.msg)
--- a/x/payment/keeper/gas_test.go
+++ b/x/payment/keeper/gas_test.go
@@ -11,17 +11,38 @@ import (
 )
 
 func TestPayForDataGas(t *testing.T) {
-	app := simapp.Setup(t, false)
-	ctx := app.BaseApp.NewContext(false, tmproto.Header{})
-	k := Keeper{}
-
-	messageSize := uint64(100)
-
-	msg := types.MsgPayForData{
-		MessageSize: messageSize,
+	type testCase struct {
+		name            string
+		msg             types.MsgPayForData
+		wantGasConsumed uint64
 	}
 
-	_, err := k.PayForData(sdk.WrapSDKContext(ctx), &msg)
-	require.NoError(t, err)
-	require.Equal(t, messageSize, ctx.GasMeter().GasConsumed())
+	testCases := []testCase{
+		{
+			name:            "1 byte message", // occupies 1 share
+			msg:             types.MsgPayForData{MessageSize: 1},
+			wantGasConsumed: uint64(4096), // 1 share * 512 bytes per share * 8 gas per byte = 4096 gas
+		},
+		{
+			name:            "100 byte message", // occupies 1 share
+			msg:             types.MsgPayForData{MessageSize: 100},
+			wantGasConsumed: uint64(4096),
+		},
+		{
+			name:            "1024 byte message", // occupies 3 shares because share prefix (namespace, info byte, e.g.)
+			msg:             types.MsgPayForData{MessageSize: 1024},
+			wantGasConsumed: uint64(12288), // 3 shares * 512 bytes per share * 8 gas per byte = 12288 gas
+		},
+	}
+
+	for _, tc := range testCases {
+		app := simapp.Setup(t, false)
+		ctx := app.BaseApp.NewContext(false, tmproto.Header{})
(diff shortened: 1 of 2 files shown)


x/payment/keeper/keeper.go
```diff
@@ -6,12 +6,21 @@ import (
...
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-const payForDataGasDescriptor = "pay for data"
+const (
+	payForDataGasDescriptor = "pay for data"
+
+	// GasPerMsgByte is the amount of gas to charge per byte of message data.
+	// TODO: extract GasPerMsgByte as a parameter to this module.
```
rootulp, line 20: Opens https://github.com/celestiaorg/celestia-app/issues/949

x/payment/keeper/keeper.go
```diff
@@ -32,7 +41,8 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) PayForData(goCtx context.Context, msg *types.MsgPayForData) (*types.MsgPayForDataResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	ctx.GasMeter().ConsumeGas(msg.MessageSize, payForDataGasDescriptor)
+	gasToConsume := uint64(shares.MsgSharesUsed(int(msg.MessageSize)) * GasPerMsgShare)
```
SweeXordious, line 44: [Non Blocking][Question]
From the gas docs:
> The AnteHandler is run for every transaction during CheckTx and DeliverTx, before a Protobuf Msg service method for each sdk.Msg in the transaction.

So, users will pay `gasToConsume` + `checkTx|DeliverTx gas` ?  But, since we're not touching any state, the `checkTx|DeliverTx gas` will be 0 ?

x/payment/keeper/gas_test.go
```diff
@@ -11,17 +11,38 @@ import (
...
+			msg:             types.MsgPayForData{MessageSize: 1},
+			wantGasConsumed: uint64(4096), // 1 share * 512 bytes per share * 8 gas per byte = 4096 gas
+		},
+		{
+			name:            "100 byte message", // occupies 1 share
+			msg:             types.MsgPayForData{MessageSize: 100},
+			wantGasConsumed: uint64(4096),
+		},
+		{
+			name:            "1024 byte message", // occupies 3 shares because share prefix (namespace, info byte, e.g.)
```
SweeXordious, line 32: [Non Blocking]
```suggestion
			name:            "1024 byte message", // occupies 3 shares because share prefix (e.g. namespace, info byte)
```

SweeXordious (commented):
LGTM, left a question to understand more.

x/payment/keeper/keeper.go
rootulp: > But, since we're not touching any state, the checkTx|DeliverTx gas will be 0 ?

I think so

x/payment/keeper/gas_test.go
```diff
@@ -11,17 +11,38 @@ import (
...
+			msg:             types.MsgPayForData{MessageSize: 1024},
+			wantGasConsumed: uint64(12288), // 3 shares * 512 bytes per share * 8 gas per byte = 12288 gas
+		},
 	}
 
-	_, err := k.PayForData(sdk.WrapSDKContext(ctx), &msg)
-	require.NoError(t, err)
-	require.Equal(t, messageSize, ctx.GasMeter().GasConsumed())
+	for _, tc := range testCases {
+		app := simapp.Setup(t, false)
```
evan-forbes, line 39: [non-blocking question/nit]
did the simapp have to be moved here? its probably negligible, but spinning up a new app each time seems a bit heavy.

x/payment/keeper/keeper.go
evan-forbes: the gas for checktx and delivertx will still have the normal sig verification and standard 10 gas/byte of tx costs as well

evan-forbes, line 44: [comment]
we should use generics or something idk, these type conversion gymnastics are everywhere

evan-forbes (dismissed):

x/payment/keeper/gas_test.go
rootulp: this line doesn't have to occur inside the `for` loop. I moved it inside for greater test isolation between test cases but it can be moved outside. I sampled test run time with this line inside / outside the for loop:

1. Inside for loop
```
ok  	github.com/celestiaorg/celestia-app/x/payment/keeper	0.430s
ok  	github.com/celestiaorg/celestia-app/x/payment/keeper	0.429s
```
2. Outside for loop
```
ok  	github.com/celestiaorg/celestia-app/x/payment/keeper	0.400s
ok  	github.com/celestiaorg/celestia-app/x/payment/keeper	0.415s
```

So it looks like tests run _slightly_ faster with the line outside the for loop. Will move outside

rootulp: resolved by https://github.com/celestiaorg/celestia-app/pull/946/commits/486a15a5fb3d17d1e68be2a0d604e6e46f577030

evan-forbes (approved):

PR #947, opened 2022-11-02
gitignore `__debug_bin`

rootulp:
Add `__debug_bin` to gitignore because this file is generated when you run tests in Debug mode inside VSCode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,5 @@ coverage.txt
 .idea
 .vscode
 tools-stamp
-.vscode
+__debug_bin
 profile.out


.gitignore
```diff
@@ -7,5 +7,5 @@ coverage.txt
 .idea
 .vscode
 tools-stamp
-.vscode
```
rootulp, line 10: `.vscode` is removed because it is duplicative of the entry two lines above

SweeXordious (approved):
Thanks for taking care of this

PR #966, opened 2022-11-07
fix: test packages and coverage generation

### LINKED ISSUE #924: Fix test coverage script
Currently, the CI test coverage script only cover the payment module, and it includes generated files such as protobuf files. We should fix it to cover go files in most of the repo (but not the things that we aren't testing such as generated files).

rahulghangas:
- [x] Closes #924


.github/workflows/test.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Create a file with all the pkgs
         run: go list ./... > pkgs.txt
       - name: Split pkgs into 4 files
-        run: split -d -n l/4 pkgs.txt pkgs.txt.part.
+        run: split -d -n l/4 -e pkgs.txt pkgs.txt.part.
```
rootulp, line 67: [proposal] what does the `-e` flag do? Does it require a specific `split` version because on my machine the man page doesn't describe a `-e` flag.

```
SPLIT(1)                                                  General Commands Manual                                                 SPLIT(1)

NAME
     split – split a file into pieces

SYNOPSIS
     split -d [-l line_count] [-a suffix_length] [file [prefix]]
     split -d -b byte_count[K|k|M|m|G|g] [-a suffix_length] [file [prefix]]
     split -d -n chunk_count [-a suffix_length] [file [prefix]]
     split -d -p pattern [-a suffix_length] [file [prefix]]

DESCRIPTION
     The split utility reads the given file and breaks it up into files of 1000 lines each (if no options are specified), leaving the file
     unchanged.  If file is a single dash (‘-’) or absent, split reads from the standard input.

     The options are as follows:

     -a suffix_length
             Use suffix_length letters to form the suffix of the file name.

     -b byte_count[K|k|M|m|G|g]
             Create split files byte_count bytes in length.  If k or K is appended to the number, the file is split into byte_count
             kilobyte pieces.  If m or M is appended to the number, the file is split into byte_count megabyte pieces.  If g or G is
             appended to the number, the file is split into byte_count gigabyte pieces.

     -d      Use a numeric suffix instead of a alphabetic suffix.

     -l line_count
             Create split files line_count lines in length.

     -n chunk_count
             Split file into chunk_count smaller files.  The first n - 1 files will be of size (size of file / chunk_count ) and the last
             file will contain the remaining bytes.

     -p pattern
             The file is split whenever an input line matches pattern, which is interpreted as an extended regular expression.  The
             matching line will be the first line of the next output file.  This option is incompatible with the -b and -l options.

     If additional arguments are specified, the first is used as the name of the input file which is to be split.  If a second additional
```

Update: https://man7.org/linux/man-pages/man1/split.1.html describes

>  -e, --elide-empty-files
>              do not generate empty output files with '-n'

if that's accurate then proposal to use

```suggestion
        run: split -d -n l/4 --elide-empty-files pkgs.txt pkgs.txt.part.
```

Motivation: https://changelog.com/posts/use-long-flags-when-scripting

rootulp (commented):
[blocking][question] sorry for the naive question but how do we know that code coverage is fixed with this PR? Is there a PR on your fork that demonstrates what the comment for coverage looks like after this PR is merged? 

https://github.com/celestiaorg/celestia-app/pull/966#issuecomment-1305032571 doesn't appear exhaustive of all packages in this repo.

.github/workflows/test.yml
```diff
@@ -150,12 +150,12 @@ jobs:
...
-          excludelist+=" $(find ./ -type f -name '*.pb.go')"
-          excludelist+=" $(find ./ -type f -name '*.pb.gw.go')"
-          excludelist+=" $(find ./ -type f -path './tests/mocks/*.go')"
+          excludelist="$(find . -type f -name '*.go' | xargs grep -l 'DONTCOVER')\n"
+          excludelist+="$(find . -type f -name '*.pb.go')\n"
+          excludelist+="$(find . -type f -name '*.pb.gw.go')\n"
+          excludelist+="$(find . -type f -path './tests/mocks/*.go')"
           for filename in ${excludelist}; do
-            filename=$(echo $filename | sed 's/^./github.com\/cosmos\/cosmos-sdk/g')
+            filename=$(echo $filename | sed 's/^./github.com\/celestiaorg\/celestia-app/g')
```
SweeXordious, line 158: [Blocking]
If I understand right, the purpose of this PR is not to upload the proto generated files, etc, to test coverage.
I created a dummy branch to check which files exist at this step:
https://github.com/celestiaorg/celestia-app/actions/runs/3413868856/jobs/5681165683

As you can see, the tree doesn't contain any folders named `github.com`. 

Also, the `excludelist` contains the right files, as  per:
```
          excludelist="$(find . -type f -name '*.go' | xargs grep -l 'DONTCOVER')\n"
          excludelist+="$(find . -type f -name '*.pb.go')\n"
          excludelist+="$(find . -type f -name '*.pb.gw.go')\n"
          excludelist+="$(find . -type f -path './tests/mocks/*.go')"
```
Thus, I guess  this could be updated to the following:

```
          for filename in ${excludelist}; do
            echo "Excluding ${filename} from coverage report..."
            sed -i.bak "/$(echo $filename | sed 's/\//\\\//g')/d" coverage.txt
          done
```
Let me know if I am missing something please.

SweeXordious (changes requested):
Thanks a lot for taking care of this :+1:  :+1: 
I left a blocking comment, but since I am not very familiar with uploading tests, I might be wrong :D

rahulghangas:
@rootulp https://app.codecov.io/github/celestiaorg/celestia-app/tree/chore%2Ffix-coverage

.github/workflows/test.yml
rahulghangas: I think the coverage file does contain files with github prefix, but it shouldn't matter I think, have removed it

rahulghangas: I'll double check by triggering the CI once more

rahulghangas: Works

SweeXordious (dismissed):

.github/workflows/test.yml
rootulp: @rahulghangas why was this resolved with `-e` still present? I think it was commited and then reverted.

.github/workflows/test.yml
```diff
@@ -150,12 +150,10 @@ jobs:
...
         run: |
-          excludelist="$(find ./ -type f -name '*.go' | xargs grep -l 'DONTCOVER')"
-          excludelist+=" $(find ./ -type f -name '*.pb.go')"
-          excludelist+=" $(find ./ -type f -name '*.pb.gw.go')"
-          excludelist+=" $(find ./ -type f -path './tests/mocks/*.go')"
+          excludelist="$(find . -type f -name '*.go' | xargs grep -l 'DONTCOVER')\n"
+          excludelist+="$(find . -type f -name '*.pb.*')\n"
+          excludelist+="$(find . -type f -path './tests/mocks/*.go')"
           for filename in ${excludelist}; do
-            filename=$(echo $filename | sed 's/^./github.com\/cosmos\/cosmos-sdk/g')
```
rootulp, line 158: it's not clear to me what the fix is. Can you please elaborate in the PR description or a code comment?

rootulp (commented):
[blocking]
Code coverage on main: https://app.codecov.io/github/celestiaorg/celestia-app/tree/main/ has 8199 lines covered
Code coverage on this branch: https://app.codecov.io/github/celestiaorg/celestia-app/tree/chore%2Ffix-coverage/ has 4845 lines covered

Why does this branch have fewer lines covered? Can it be rebased on top of current main to see what the difference is? If this PR fixes the issue, I would expect more lines to be covered, not less.

rootulp:
>  Probably, because the proto generated files are not included.

https://github.com/celestiaorg/celestia-app/pull/966/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R153-R155 removed 

```
          excludelist+=" $(find ./ -type f -name '*.pb.gw.go')"
```

from the exclude list so I'd expect more proto files to be included in the coverage.txt

SweeXordious:
@rootulp I guess it's these files that are excluded:
```
          excludelist="$(find . -type f -name '*.go' | xargs grep -l 'DONTCOVER')\n"
          excludelist+="$(find . -type f -name '*.pb.*')\n"
          excludelist+="$(find . -type f -path './tests/mocks/*.go')"
```

As per:
```
            sed -i.bak "/$(echo $filename | sed 's/\//\\\//g')/d" coverage.txt
```

Or, I am missing something

rootulp:
It looks like before and after this PR, `.pb.go` files are being covered:
- `main`: https://app.codecov.io/github/celestiaorg/celestia-app/tree/main/?search=.pb.go
- this branch: https://app.codecov.io/github/celestiaorg/celestia-app/tree/chore%2Ffix-coverage/?search=.pb.go&displayType=list

Fewer `.pb.go` files show up for this branch but I don't think any are intended.

No `.pb.gw.go` files are being covered in this branch (which seems desirable):
- `main`: https://app.codecov.io/github/celestiaorg/celestia-app/tree/main/?search=.pb.gw.go
- this branch: https://app.codecov.io/github/celestiaorg/celestia-app/tree/chore%2Ffix-coverage/?search=.pb.gw.go&displayType=list

rahulghangas:
I don't get why it's picking up those files, or why changing it would help since the command above returns all the `*pb.go` files successfully when I run that command locally

.github/workflows/test.yml
rahulghangas: Sorry, forgot to push the change I think

evan-forbes (approved):
LGTM utACK

thx for this @rahulghangas !

rahulghangas:
How do I go about triggering a code coverage report?

SweeXordious:
If I understand right, codecov will automatically push a comment if there were some changes, for the first time the PR is open. Then, it will start updating the comment, it posted the first time, every time a commit does a change to the patterns:
```
            **/**.go
            go.mod
            go.sum
```
However, we can always access the coverage from here: https://app.codecov.io/github/celestiaorg/celestia-app/tree/chore%2Ffix-coverage/

To trigger a new coverage report, changes need to be done to the above files. Probably, a dummy commit and revert would do the job. Or, someone with admin access (don't know which permissions they need to have).

rahulghangas:
https://app.codecov.io/github/celestiaorg/celestia-app/tree/chore%2Ffix-coverage/
All proto files are exlcuded correctly now

ps - I hate bash!

SweeXordious (approved):
@rahulghangas Thanks a lot for handling this :+1: :+1: :tada: :tada:

PR #972, opened 2022-11-08
chore!: rename payment module -> blob

### LINKED ISSUE #961: Consider renaming `x/payment`
## Context

During an onsite meeting, @liamsi proposed renaming the `x/payment` module to something else prior to the v1.0.0 release because this type of rename would be a breaking change. We should align on if we want a rename and what the new name should be.

## Proposals

- `x/da`
- `x/blob`

evan-forbes:
speak now or forever hold your peace. 

NOTE: this PR does not rename MsgPayForData or MsgWirePayForData, as this was already too big of a diff.

closes #961


SweeXordious (approved):
:shipit:

x/qgb/keeper/keeper.go
```diff
@@ -58,7 +58,7 @@ func (k Keeper) DeserializeValidatorIterator(vals []byte) stakingtypes.ValAddres
 	return validators
 }
 
-// StakingKeeper restricts the functionality of the bank keeper used in the payment keeper
+// StakingKeeper restricts the functionality of the bank keeper used in the qgb keeper
```
rootulp, line 61: 👍

rootulp (approved):

rootulp:
`blobkeeper` 😆

SweeXordious:
@evan-forbes The test `TestMaxBlockSize` has started to fail quite a lot lately. Should we investigate?

evan-forbes:
yeah I've started, https://github.com/celestiaorg/celestia-app/pull/988

it might be anecdotal, but it seems to fail a lot more often when we are running the CI in different PRs at once. It also started to when we decreased the block time.

PR #985, opened 2022-11-09
feat: remove QGB valset signing code

SweeXordious:
remove unnecessary QGB code


rahulghangas (approved):
LGTM

x/qgb/types/valset.go
```diff
@@ -26,54 +20,6 @@ func NewValset(nonce, height uint64, members InternalBridgeValidators) (*Valset,
...
-		panic(fmt.Sprintf("Error packing checkpoint! %s/n", err))
-	}
-
-	hash := crypto.Keccak256Hash(bytes[4:])
-	return hash, nil
-}
-
-// Hash mimics the 'computeValsetHash' function used by the qgb contracts by using
-// a Valset to compute the hash of the abi encoded validator set.
-func (v *Valset) Hash() (ethcmn.Hash, error) {
```
evan-forbes, line 60: [blocking question]
If we're keeping the `Valset` struct, why do we want to delete the hash function here?

evan-forbes (commented):
LGTM except for one blocking question

x/qgb/types/valset.go
SweeXordious: Yes, I always keep forgetting this could be used for slashing eventually. Also, could be used as a source of truth for the signatures.
Closing PR. 
Thanks a lot for reminding me

PR #986, opened 2022-11-09
chore: improve Makefile

rootulp:
- Reorder commands into groups:
  1. Install
  2. Build
  3. Protobuf
  4. Docker
  5. Lint
  6. Test
- Add `.PHONY` targets to commands that don't generate files
- Add `echo` statements to describe what command is doing


evan-forbes (approved):
utACK

PR #987, opened 2022-11-09
chore!: use different default codec

evan-forbes:
After updating rsmt2d, we never actually switched the default codec! This still needs to be tested w/ celestia-node before cutting the v0.10.0


pkg/wrapper/nmt_wrapper_test.go
```diff
@@ -95,7 +95,7 @@ func TestExtendedDataSquare(t *testing.T) {
 	raw := generateRandNamespacedRawData(
 		squareSize*squareSize,
 		appconsts.NamespaceSize,
-		appconsts.SparseShareContentSize,
+		appconsts.SparseShareContentSize+1, // we +1 here to account for the version byte
```
SweeXordious, line 98:
```
SparseShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes
```
Doesn't it make sense to increase the `ShareInfoBytes` throughout the repo?

pkg/wrapper/nmt_wrapper_test.go
```diff
@@ -26,7 +26,7 @@ func TestPushErasuredNamespacedMerkleTree(t *testing.T) {
 		tree := NewErasuredNamespacedMerkleTree(uint64(tc.squareSize), 0)
 
 		// push test data to the tree
-		for _, d := range generateErasuredData(t, tc.squareSize, appconsts.DefaultCodec()) {
+		for _, d := range generateErasuredData(t, tc.squareSize, rsmt2d.NewLeoRSCodec()) {
```
SweeXordious, line 29: will this change also need to be tested with QGB messages?

pkg/appconsts/appconsts.go
```diff
@@ -104,7 +104,7 @@ var (
 	// DefaultCodec is the default codec creator used for data erasure
 	// TODO(ismail): for better efficiency and a larger number shares
 	// we should switch to the rsmt2d.LeopardFF16 codec:
```
rootulp, line 106: [question] should we remove this TODO because we're switching codecs in this PR? 

```suggestion
```

pkg/wrapper/nmt_wrapper_test.go
rootulp: [blocking][question]

`SparseShareContentSize` already accounts for the one info byte and I don't think we need to increase it from 1 to 2 so it's not clear to me what the `+1` is for. Any ideas @evan-forbes ?

pkg/wrapper/nmt_wrapper_test.go
evan-forbes: it shouldn't, but this still reminded me to switch that back to the default https://github.com/celestiaorg/celestia-app/pull/987/commits/97feb2ca4faec88b7a37827d20bc6db9523395f0

evan-forbes: it shouldn't because this only has an effect of the erasure data.

pkg/wrapper/nmt_wrapper_test.go
SweeXordious: which version? is it this?
```
Where `info (1 byte)` is a byte with the following structure:

- the first 7 bits are reserved for the version information in big endian form (initially, this will be `0000000` for version 0);
- the last bit is a **sequence start indicator**, that is `1` if the share is at the start of a sequence or `0` if it is a continuation share.
```
?

evan-forbes: yeah, sorry, the info byte. We need to add a byte to replace it

SweeXordious: Sorry, still don't get the reasoning behind. Now with the new rsmt2d, will we increase the version byte for all protocol? 
If so, then this should be reflected in the `appconsts`, not just the tests.
I am missing some context apparently... Please bare with me :D

evan-forbes: lol its fine

the old codec handled [][]byte of sizes that aren't powers of two, this one does not, so we just have to make sure that its 512. The current constant is fine and what it should be, we just need to tell the test code to account for the info byte, which was added after the test generation code. 

we don't need to change anything anywhere, only the test code

SweeXordious: aaah, the leaf size.  Okey, got you :+1:

SweeXordious (approved):
Thanks a lot for explanations
:shipit:

evan-forbes:
This is blocking the v0.10.0-rc1 release, so I'm merging. @rootulp pls feel free to fix docs in a different PR is the new ones don't make sense

rootulp:
>  @rootulp pls feel free to fix docs in a different PR is the new ones don't make sense

No need b/c this doc comment makes more sense to me:

```
// we +1 here to keep the generated data to be 512 bytes in len
```

Issue #829, opened 2022-10-03
Consolidate network testing code


evan-forbes:
With the introduction of our own network tests, we should replace our usage of the old cosmos-sdk network tests for our integration tests. 

This will involve:
- [ ] Adding cli execution functionality for the new testnode
- [ ] Refactor existing integration tests to use new testnode

we should also expect a decrease in integration test times, since the new testnode can produce sub 1s blocks.

rootulp:
Resolving this issue will likely also resolve https://github.com/celestiaorg/celestia-app/issues/574

rahulghangas:
Picking this up

evan-forbes:
in #1474, we are introducing the ibctesting framework for testing the ibc token filter. This seems like a good solution imo considering the amount of work we'd have to do replicate the same functionality. However, I'll ref it here to have a more exhaustive list

rootulp:
I think the original post refers to the two test networks:
1. https://github.com/celestiaorg/celestia-app/blob/be3acf0a79849af0dafda5eecb0644cfc08c58b0/test/util/network/network.go#L27
2. https://github.com/celestiaorg/celestia-app/blob/be3acf0a79849af0dafda5eecb0644cfc08c58b0/test/util/testnode/full_node.go#L69

where 1 should be deleted and usage of 1 should be replaced with 2.

PR #988, opened 2022-11-09
chore: increase retries and block times in integration test for CI

evan-forbes:
trying some things out to reduce flakiness of CI as its getting extremely annoying


evan-forbes:
one of the things that is causing issues is that the mempool is full when we submit txs. We don't have access to the full tendermint config needed to increase it until we switch to our own proprietary integration test code 

#829

SweeXordious:
Aah makes sense. Thanks :+1:

rahulghangas:
I'll create the relevant PR to close #829 today

evan-forbes:
closing this, but we might want to incorporate something similar when we are refactoring the testing in https://github.com/celestiaorg/celestia-app/issues/829

PR #1003, opened 2022-11-14
ADR009: Non-Interactive Default Rules for Reduced Padding

nashqueue:
This PR is regarding this [Question](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-008-square-size-independent-message-commitments.md#positive-celestia-app-changes).

[rendered](https://github.com/celestiaorg/celestia-app/blob/8d0335b99324c4cfec9c7f2affbae044708ef356/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md)


docs/architecture/adr-008-square-size-independent-message-commitments.md
```diff
@@ -114,13 +114,4 @@ We should note that Rollups can decide to do this scheme without changing core-a
...
-    > - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
-
-    To this: Messages start at an index that is a multiple of its `msgMinSquareSize`.
-
-    As an example, we have this diagram. Message 1 is three shares long and is followed by message 2, which is 11 shares long, so the `msgMinSquareSize` of the second message is equal to four. Therefore we have a padding of 5 shares shown in light blue. Furthermore, with the new non-interactive default rule set, a message of size 11 can start in this block at index zero and index three because they are multiples of four. Therefore, we save four shares of padding while retaining the same commitment.
-
-    ![Padding Savings](./assets/padding-savings.png)
+- Inter-message padding can be reduced in the worst case by 50% if we can change the non-interactive default rules. An in-depth analysis is performed at [ADR 009: New Non-Interactive Default Rules for Reduced Padding](Link).
```
rootulp, line 117: It looks like this link is causing markdownt lint link checking to fail.

```suggestion
- Inter-message padding can be reduced in the worst case by 50% if we can change the non-interactive default rules. An in-depth analysis is performed at [ADR 009: New Non-Interactive Default Rules for Reduced Padding](./adr-009-proof-size-analysis.md).
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,151 @@
+# ADR 009: New Non-Interactive Default Rules for Reduced Padding
+
+## Changelog
+
+- 14.11.2022: Initial Draft
+
+## Context
+
+ Is it worth it to change non-interactive default rules as a follup up of ADR 008 [Link??????] to decrease the padding? The downside is that the message inclusion proof size will not be as efficient in big square sizes.
```
rootulp, line 9:
```suggestion
 Is it worth it to change non-interactive default rules as a follup up of [ADR 008](./adr-008-square-size-independent-message-commitments.md) to decrease the padding? The downside is that the message inclusion proof size will not be as efficient in big square sizes.
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+- Inter-message padding can be reduced if we can change the non-interactive default rules from this:
+
+    > - Messages that span multiple rows must begin at the start of a row (this can occur if a message is longer than k shares or if the block producer decides to start a message partway through a row and it cannot fit).
+    > - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
+
+    To this: **Messages start at an index that is a multiple of its `msgMinSquareSize`.**
+
+    As an example, we have this diagram. Message 1 is three shares long and is followed by message 2, which is 11 shares long, so the `msgMinSquareSize` of the second message is equal to four. Therefore we have a padding of 5 shares shown in light blue. Furthermore, with the new non-interactive default rule set, a message of size 11 can start in this block at index zero and index three because they are multiples of four. Therefore, we save four shares of padding while retaining the same commitment.
+
+    ![Padding Savings](./assets/padding-savings.png)
```
rootulp, line 22: [optional] I noticed a few small styling things in this section and rather than suggesting changes on individual lines, I suggested a revised section. The suggestions are based on:
- reduce one layer of indentation
- visually format the current and proposed non-interactive default rules so that they are easier to compare
- link to issue #1004 
- describe upside and downside in the same paragraph
- reword the padding savings explanation and separate the current vs proposed into distinct paragraphs

```suggestion
## Context

[ADR 008](./adr-008-square-size-independent-message-commitments.md) makes it possible for the current non-interactive default rules to be modified slightly:

Current

> - Messages that span multiple rows must begin at the start of a row (this can occur if a message is longer than k shares or if the block producer decides to start a message partway through a row and it cannot fit).
> - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.

Proposed

> - Messages start at an index that is a multiple of its `msgMinSquareSize`.

The upside of this proposal is that it reduces the inter-message padding. The downside is that a message inclusion proof will not be as efficient for large square sizes so the proof will be larger.

> **Note**
> This analysis assumes the implementation of [celestia-app#1004](https://github.com/celestiaorg/celestia-app/issues/1004). If the tree over the subtree roots is not a Namespace Merkle Tree then both methods have the same proof size.**

As an example, take the diagram below. Message 1 is 3 shares long and message 2 is 11 shares long.

With the current non-interactive default rules, message 2 must start at a location aligned with the largest power of 2 that is not larger than 11 (message length) or 8 (square size). Therefore, message 2 must start at a location aligned with 8 which is index 16 in the example below. This arragenemnt results in 5 shares of padding.

With the proposed non-interactive default rules, message 2 must start at an index that is a multiple of `msgMinSquareSize`. A message of 11 shares can fit in a square size of 4 (since 4 * 4 = 16 available shares which is < 11) so `msgMinSquareSize` is 4. Therefore, message 2 can start at index 12. This arrangement results in 1 share of padding.

![Padding Savings](./assets/padding-savings.png)
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Defining variables for this Analysis
+
+- n:= Message length in number of shares
+- k:= Square size
+- r:= Number of rows a message fills in a Celestia square.
+
+### The following Questions will be answered
+
+- Why has a message in a bigger square size even with square size independent commitments O(log(n)) proof size?
```
rootulp, line 32: [question] can this be rephrased to something like

```suggestion
- Given square size independent commitments, why does a message in a larger square size result in a O(log(n)) message inclusion proof size?
```

Note: accepting this suggestion implies updating the heading below

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+- What is the worst-case constructible message in a square to have the biggest impact from switching from O(log(n)) to O(sqrt(n))?
```
rootulp, line 33: [optional] rephrased for easier readability 

```suggestion
- Assuming message inclusion proof sizes change from O(log(n)) to O(sqrt(n)), what is the worst-case constructible message?
```

Note: accepting this suggestion implies updating the heading below

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+- Why can we not use the same trick that we used in Question 1 in a single row for more efficient proofs over the row roots?
+- How big is the proof size for this message?
+- What is the worst constructible block with the most amount of padding with old and new non-interactive defaults?
+- Quantify padding and proof size cost.
```
rootulp, line 37: [optional] Since the rest of these list items are questions and the heading is question, proposal to rephrase to a question

```suggestion
- What is the quantified padding and proof size cost?
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+If you use the old NI-Rules then the message begins at a location aligned with the largest power of 2 that is not larger than the message length or k. Because the subtree roots are aligned you can skip some subtree roots and calculate their parents.
+In the example below instead of proving H1, H2, H3, and H4 to the DataRoot you can prove H10. **H10 is part of the commitment generation and part of the Merkle tree to the DataRoot.** That is why you can use it for more efficient proofs. In smaller square sizes, you cannot do this, because H10 does not exist. The nodes in **red** are the subtree nodes that you need to provide for the message inclusion proof. The nodes in **blue** are the additional nodes for the Merkle proof.
+
+![Efficient Merkle Proofs with ADR008](./assets/effizicient-proof-size-ADR008.png)
+
+So why can you not do it with the new NI-Rules? This is because H10 is not generated. In the diagram below the first 8 shares are in the row before and therefore the tree structure changes. The previous subtree root H10 is now H23 and cannot be used for the efficiency boost.
```
rootulp, line 46: [optional] I realize this document uses NI-Rules in many places so it may not be desirable to use the longer form "non-interactive default rules". However, the longer form avoids any ambiguity around abbreviations which we've historically had with NID (sometimes confused with namespaceID). If we want to keep the abbreviation, we may clarify its first usage (I think this instance) by including the longer form in parenthesis after: NI-Rules (non-interactive rules)

```suggestion
So why can you not do it with the new non-interactive default rules? This is because H10 is not generated. In the diagram below the first 8 shares are in the row before and therefore the tree structure changes. The previous subtree root H10 is now H23 and cannot be used for the efficiency boost.
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+### New Non-Interactive Default Rules
+
+Each row consists of O(sqrt(n)/log(n)) subtree roots. Which makes in total sqrt(n) subtree roots. The rest is here the same as before.
+
+Proofsize = sqrt(n) + log(n) + log(k) + 2*log(k)
+
+## 5. What is the worst constructible block with the most amount of padding with old and new non-interactive default rules?
+
+For the old NI-Rules, when you have a square size of k you can alternate messages of size k/2 +1 and k/4 +1 to have the most amount of padding. This forces you to always use a new row.
+Padding = (k/2 -1) \* (k/2 -1) + (3k/4 -1) \* (k/2 -1)
```
rootulp, line 122: [optional] to visually separate the equation from the text above

```suggestion
For the old NI-Rules, when you have a square size of k you can alternate messages of size k/2 +1 and k/4 +1 to have the most amount of padding. This forces you to always use a new row.

Padding = (k/2 -1) \* (k/2 -1) + (3k/4 -1) \* (k/2 -1)
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+To have the most amount of padding for the new NI-Rules you use repeated messages of size 5 which will result in a padding of 3 in between.
+Padding = 3 \* (k-1) \* k/8
```
rootulp, line 125: [optional]

```suggestion
To have the most amount of padding for the new NI-Rules you use repeated messages of size 5 which will result in a padding of 3 in between.

Padding = 3 \* (k-1) \* k/8
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+The Padding decreases a lot.
```
rootulp, line 147:
```suggestion
The padding decreases a lot.
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Negative
+
+The Proof size increases a bit.
```
rootulp, line 151:
```suggestion
The proof size increases a bit.
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+## 4. How big is the proof size for this message?
+
+We differentiate the size of the proof between the Old NI-Rules and the new NI-Rules.
+
+### Old Non-Interactive Default Rules
+
+Each row consists of one subtree root, which means if you have log(n) rows you will have in total log(n) subtree roots. The last row has log(n) subtree roots. The last row will also need log(k) blue nodes. Blue nodes are additional nodes that you need for the Merkle proof, which have been used in the previous diagrams. After having now r row roots we need to a Merkle proof of them to the `DataRoot`.In the worst case, the message lies in the middle of the block. Therefore we will need 2* log(k) blue nodes for the proof.
+
+Proofsize = log(n) + log(n) + log(k) + 2*log(k)
```
rootulp, line 111:
```suggestion
Proof size = log(n) + log(n) + log(k) + 2*log(k)
```

docs/architecture/adr-009-proof-size-analysis.md
```diff
@@ -0,0 +1,155 @@
...
+
+### New Non-Interactive Default Rules
+
+Each row consists of O(sqrt(n)/log(n)) subtree roots. Which makes in total sqrt(n) subtree roots. The rest is here the same as before.
+
+Proofsize = sqrt(n) + log(n) + log(k) + 2*log(k)
```
rootulp, line 117:
```suggestion
Proof size = sqrt(n) + log(n) + log(k) + 2*log(k)
```

rootulp (approved):
I don't have any blocking feedback, it's all optional to make the document read easier.

I plan on doing another review later this week to inspect the math closer but don't want to block the PR on my re-review.

docs/architecture/adr-009-proof-size-analysis.md
nashqueue: I will switch to the longer form. Maybe it's hinders readability to add new abbreviations. I did not want to use NID, because of the misunderstanding you mentioned.

docs/architecture/adr-008-square-size-independent-message-commitments.md
```diff
@@ -114,13 +114,4 @@ We should note that Rollups can decide to do this scheme without changing core-a
...
-    > - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
-
-    To this: Messages start at an index that is a multiple of its `msgMinSquareSize`.
-
-    As an example, we have this diagram. Message 1 is three shares long and is followed by message 2, which is 11 shares long, so the `msgMinSquareSize` of the second message is equal to four. Therefore we have a padding of 5 shares shown in light blue. Furthermore, with the new non-interactive default rule set, a message of size 11 can start in this block at index zero and index three because they are multiples of four. Therefore, we save four shares of padding while retaining the same commitment.
-
-    ![Padding Savings](./assets/padding-savings.png)
+- Inter-message padding can be reduced in the worst case by 50% if we can change the non-interactive default rules. An in-depth analysis is performed at [ADR 009: New Non-Interactive Default Rules for Reduced Padding](./adr-009-proof-size-analysis.md).
```
rootulp, line 117: dang, sorry but markdownlint now complains because `./adr-009-proof-size-analysis.md` doesn't exist

https://github.com/celestiaorg/celestia-app/actions/runs/3486753915/jobs/5833589081#step:4:223

I wonder if this will fix:

```suggestion
- Inter-message padding can be reduced in the worst case by 50% if we can change the non-interactive default rules. An in-depth analysis is performed at [ADR 009: New Non-Interactive Default Rules for Reduced Padding](./adr-009-non-interactive-default-rules-for-reduced-padding.md).
```

nashqueue:
Thinks i need to adjust the calculation: 

- [ ] Add parity shares as blue nodes to the calculation
- [ ] Lower the size of the nodes above the Row roots as it is not a nmt

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+> - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
+
+Proposed
+
+> - Messages start at an index that is a multiple of its `msgMinSquareSize`.
+
+The upside of this proposal is that it reduces the inter-message padding. The downside is that a message inclusion proof will not be as efficient for large square sizes so the proof will be larger.
+
+> **Note**  
+> **This analysis assumes the implementation of [celestia-app#1004](https://github.com/celestiaorg/celestia-app/issues/1004). If the tree over the subtree roots is not a Namespace Merkle Tree then both methods have the same proof size.**
```
rootulp, line 23: [optional] bold can be effective at adding emphasis to a few words but loses its significance if used for large chunks of text. In this case, the note syntax already adds emphasis to the following text so bold doesn't seem necessary

```suggestion
> This analysis assumes the implementation of [celestia-app#1004](https://github.com/celestiaorg/celestia-app/issues/1004). If the tree over the subtree roots is not a Namespace Merkle Tree then both methods have the same proof size.
```

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+| 2048 | 30509 | 15 |
+
+ Reminder: We did this calculation because we need O(log(n)) rows.
+
+## 3. Why can we not use the same trick that we used in Question 1 in a single row for more efficient proofs over the row roots?
+
+The node needs to be part of the commitment generation **and** part of the Merkle tree to the DataRoot for the trick to work. The diagram shows a Celestia square that is erasure coded and those parity shares are marked in green.
+H12 is part of the commitment generation and part of the Merkle tree to the DataRoot.
+It is only generated in the bigger square and not in the smaller square because in the smaller square you have to take into account the nodes over the parity shares.
+As H12 only exists in the bigger square the more efficient proofs only work in those squares.
```
rootulp, line 108: [question] It seems like the bottom most square in `./assets/rowroots-might-not-be-subtreeroot.png` also needs to be erasure coded so I don't follow why erasure coding the small square breaks the trick but erasure coding the larger square doesn't break the trick.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+| 2048 | 30509 | 15 |
+
+The worst case constructible message in a square to have the biggest impact from switching from O(log(n)) to O(sqrt(n)) with n being the size of the message, has r rows in a square size of k. If r is larger than k/4 we need to take k/4 as the number of rows instead because of the first point in this message. With adopting this rule the messages look as follows:
+
+|   k  |   n   |  r |
+|:----:|:-----:|:--:|
+|    2 |     **2** |  **1** |
+|    4 |     **4** |  **1** |
+|    8 |    **16** |  **2** |
+|   16 |    **64** |  **4** |
```
rootulp, line 92: [question] are these bold for any particular reason? If not, proposal to remove

```suggestion
|    2 |     2 |  1 |
|    4 |     4 |  1 |
|    8 |    16 |  2 |
|   16 |    64 |  4 |
```

rootulp (commented):
No blocking feedback, still LGTM

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
nashqueue: I changed the picture so it should be clearer :) H12 does not exist in the smaller square but in the larger square it does.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
rootulp: Thanks for clarifying, I think we can resolve this comment.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
rootulp: Ahh I see, thanks for updating the diagram.

Small correction in the bottom square: `ROW3` => `ROW2` and `ROW2` => `ROW3`

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Consequences
+
+### Positive
+
+The padding decreases a lot.
+
+### Negative
+
+The proof size increases a bit.
```
evan-forbes, line 173: [very optional]
I could see it as being orthogonal, but we might want to go into more of the indirect consequences, such as zk prover costs or possibly decreasing bandwith for trust minimized light clients by decreasing the average DAH size (assuming the same amount of data posted).

evan-forbes (commented):
Nice! Really great analysis :+1: Left a few suggestions and questions. I'd defer to others before merging, but **iff** this ADR is going stale after a month or so, I am fine merging provided we keep it as **`Proposed`** status. 

I still need to think about the tradeoffs a bit more, but fwiw I'm leaning towards being in favor of this proposal. 

If we are only thinking about bandwidth costs of super light clients, and we are only thinking about 134MB blocks (square size 512) for the foreseeable future. Then in the worst possible case we would only increase super light clients bandwidth by about 11 GB **per year**. (roughly 2,102,400 blocks per year (4 per minute), max block size each block, and worst size of rollup block each block)

However, if we went with this proposal have a more efficient padding, then we would either have a smaller average square size or be able to post more data. If we have a smaller average square size, then we reduce the bandwidth costs of minimized light clients because they gossip a smaller average DAH.

nashqueue:
> However, if we went with this proposal have a more efficient padding, then we would either have a smaller average square size or be able to post more data. If we have a smaller average square size, then we reduce the bandwidth costs of minimized light clients because they gossip a smaller average DAH.

Unfortunately this might be false considering the [Jevons_paradox](https://en.wikipedia.org/wiki/Jevons_paradox). It would be the same as making Blockspace bigger to decrease congestion, but more block space _could_ in turn create more demand creating the same congestion.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+
+Given a square size k, the biggest message that you can construct that is affected by the proposed non-interactive default rules has a size (k/2)². If you construct a message that is bigger than (k/2)² the `minSquareSize` will be k. If the minSquareSize is k in a square of size k then the current non-interactive default rules are equivalent to the proposed non-interactive default rules, because the message starts always at the beginning of a row. In other words, if you have k² shares in a message the worst constructible message is a quarter of that k²/4, because that is the size of the next smaller square.
+
+If you choose k²/4 as the worst constructible message it would still have O(sqrt(n)) subtree roots. This is because the size of the message is k²/4 with a width of k and a length of k/4. This means the number of rows the message fills approaches O(sqrt(n)). Therefore we need to find a message where the number of rows is log(n) of the size of the message.
+
+With k being the square size and n being the number of shares and r being the number of rows, we want to find a message so that:
+k * r = n & log(n) = r => k = n/log(n)
```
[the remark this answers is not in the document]
nashqueue: So we want to find a message of size n where the rows of the message are log(n) so we still have O(log(n)) proof size. This means that r = log(n) . k is the square size , so a rollup message is roughly k (width) *r(height) = n (Area). I plugged the 2 equations together and i got a relationship between square size and message size to resolve this. As k is fixed 2, 4, 8 etc we can calculate n and in turn with n we can calculate r for k = 2,4,8 etc. 
I must say though that we could have had for example any c for r = c * log(n) as it would still be O(log(n)). But as we don't increase blocksize to infinity, we need to find a compromise between O-Notation and quantitative solution and mixing them here might create some confusion. Maybe r = k/4 is a better candidate for analysis and that is still up for debate. Happy for input @musalbas

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+- 14.11.2022: Initial Draft
+
+## Context
+
+[ADR 008](./adr-008-square-size-independent-message-commitments.md) makes it possible for the current non-interactive default rules to be modified slightly:
+
+Current
+
+> - Messages that span multiple rows must begin at the start of a row (this can occur if a message is longer than k shares or if the block producer decides to start a message partway through a row and it cannot fit).
+> - Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or k.
```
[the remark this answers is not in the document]
nashqueue: I guess that makes sense the first rule implicit :) Does that mean a message of size 5 in block of square size 8 can start at a index 3 ? I think i made a mistake in padding regrading this, as I assumed a message of size 5 needs to start at index 8. I will fix this and adapt the padding analysis.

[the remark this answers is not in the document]
nashqueue: You still need to inlclude the first sentence as you did it in https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-006-non-interactive-defaults.md?plain=1#L28

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+
+Proposed
+
+> - Messages start at an index that is a multiple of its `msgMinSquareSize`.
+
+The upside of this proposal is that it reduces the inter-message padding. The downside is that a message inclusion proof will not be as efficient for large square sizes so the proof will be larger.
+
+> **Note**  
+> This analysis assumes the implementation of [celestia-app#1004](https://github.com/celestiaorg/celestia-app/issues/1004). If the tree over the subtree roots is not a Namespace Merkle Tree then both methods have the same proof size.
```
musalbas, line 23: Even if it's an NMT wouldn't it still have the same size, because the row roots also contain the erasure coded data? Which means you need to "open" each row roots anyways in order to get the non-parity shares.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
```diff
@@ -0,0 +1,177 @@
...
+### The following questions will be answered
+
+- Given square size independent commitments, why does a message in a larger square size result in an O(log(n)) message inclusion proof size?
+- Assuming message inclusion proof sizes change from O(log(n)) to O(sqrt(n)), what is the worst-case constructible message?
+- Why can we not use the same trick that we used in Question 1 in a single row for more efficient proofs over the row roots?
+- How big is the proof size for this message?
+- What is the worst constructible block with the most amount of padding with old and new non-interactive defaults?
+- What is the quantified padding and proof size cost?
+
+## 1. Given square size independent commitments, why does a message in a larger square size result in an O(log(n)) message inclusion proof size?
```
musalbas, line 48: As above; is this still still true when we consider that rows also have erasure coded data?

docs/architecture/adr-008-square-size-independent-message-commitments.md
```diff
@@ -72,7 +72,7 @@ Implemented
 ### Negative
 
 1. The amount of subtree roots per commitment is O(sqrt(n)), while n is the number of message shares. The worst case for the number of subtree roots is depicted in the diagram below - an entire block missing one share.
-  ![Interactive Commitment 2](./assets/complexity.png)
+  ![Interactive Commitment 2](./assets/complexity.png)
```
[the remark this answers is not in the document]
nashqueue: 2 spaces are the same as an enter in markdown. If we do not have that then the text will continue after the picture and not in a new line. It looks nicer with the 2 spaces.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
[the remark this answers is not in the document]
nashqueue: Yes makes sense it's different and a padding save in itself!

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
nashqueue: The parity shares only happen at the top of the tree so at least I need to download each row root. If we assume that we have log(n) rows than we still need sqrt(n) subtree roots vs log(n) roots, becasue i am  talking about proving subroots within the same row.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
nashqueue: see reply above

docs/architecture/adr-008-square-size-independent-message-commitments.md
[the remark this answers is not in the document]
adlerjohn: Oh, then
```suggestion

![Interactive Commitment 2](./assets/complexity.png)  

```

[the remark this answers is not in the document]
nashqueue: That does not work because then the enumeration breaks in which the picture is in. Line breaks inside enumerations is a heated topic. https://meta.stackexchange.com/a/192898 I did not get Option 1 (from stack overflow) to work, because of bad indentation and makrdownlint complaining.

docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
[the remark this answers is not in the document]
rootulp: > we have yet to document it in the specs

created https://github.com/celestiaorg/celestia-app/issues/1063

evan-forbes (dismissed):

evan-forbes:
what are the todos left on this @nashqueue before we merge?

rootulp (approved):

evan-forbes (approved):

Issue #590, opened 2022-08-03
config/mempool: define default parameters


Bidon15:
Introduction ✌🏻
---
Here is the current params we have for mempool part 
```toml
# Mempool version to use:
#   1) "v0" - (default) FIFO mempool.
#   2) "v1" - prioritized mempool.
version = "v0"

recheck = true
broadcast = true
wal_dir = ""

# Maximum number of transactions in the mempool
size = 5000

# Limit the total size of all txs in the mempool.
# This only accounts for raw transactions (e.g. given 1MB transactions and
# max_txs_bytes=5MB, mempool will only accept 5 transactions).
max_txs_bytes = 1073741824

# Size of the cache (used to filter transactions we saw earlier) in transactions
cache_size = 10000

# Do not remove invalid transactions from the cache (default: false)
# Set to true if it's not possible for any invalid transaction to become valid
# again in the future.
keep-invalid-txs-in-cache = true

# Maximum size of a single transaction.
# NOTE: the max size of a tx transmitted over the network is {max_tx_bytes}.
max_tx_bytes = 1048576

# Maximum size of a batch of transactions to send to a peer
# Including space needed by encoding (one varint per transaction).
# XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
max_batch_bytes = 0

# ttl-duration, if non-zero, defines the maximum amount of time a transaction
# can exist for in the mempool.
#
# Note, if ttl-num-blocks is also defined, a transaction will be removed if it
# has existed in the mempool at least ttl-num-blocks number of blocks or if it's
# insertion time into the mempool is beyond ttl-duration.
ttl-duration = "0s"

# ttl-num-blocks, if non-zero, defines the maximum number of blocks a transaction
# can exist for in the mempool.
#
# Note, if ttl-duration is also defined, a transaction will be removed if it
# has existed in the mempool at least ttl-num-blocks number of blocks or if
# it's insertion time into the mempool is beyond ttl-duration.
ttl-num-blocks = 15
```

New Definition 📜 
---
- Are we using as default prioritized mempool or FIFO? 
   `version = "v0"`
- Here I don't understand why we have 1024 MB for `max_txs_bytes` ? 
   - Why do we need such figure considering that we won't have 1GB blocks out of the gate? 
- Are we fine with `ttl-*` figures? 

Notes 📝 
---
Ref: #585

rootulp:
## Questions

> Are we using as default prioritized mempool or FIFO?

[...]

PR #1008, opened 2022-11-15
feat!: override default mempool params

rootulp:
Supersedes https://github.com/celestiaorg/celestia-core/pull/885

Closes https://github.com/celestiaorg/celestia-core/issues/812 
Closes https://github.com/celestiaorg/celestia-core/issues/867
Closes https://github.com/celestiaorg/celestia-app/issues/590

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -78,6 +78,9 @@ func NewRootCmd() *cobra.Command {
 			}
 
 			tmCfg := tmcfg.DefaultConfig()
+			tmCfg.Mempool.TTLNumBlocks = 10
+			tmCfg.Mempool.MaxTxBytes = 2 * 1024 * 1024 // 2 MiB
+			tmCfg.Mempool.Version = "v1"               // prioritized mempool
 
 			customAppTemplate, customAppConfig := initAppConfig()
 			return server.InterceptConfigsPreRunHandler(cmd, customAppTemplate, customAppConfig, tmCfg)


evan-forbes (approved):
thanks!! tested locally and it works as expected

evan-forbes (approved):
thanks!! tested this locally and it worked as expected

rootulp:
One thing I didn't consider until after merging is that it should be possible for validators to override this default param by configuring `~/.celestia-app/config/config.toml`. I'll test shortly.

evan-forbes:
> One thing I didn't consider until after merging is that it should be possible for validators to override this default param by configuring

thanks for testing, I didn't test, but I did check it out the code in the sdk and it looked like we should be good https://github.com/celestiaorg/cosmos-sdk/blob/a78f3e1b2d09a853695c1b85da72a37445d7124f/server/util.go#L201-L225

rootulp:
Tested by adding the following log after [this](https://github.com/rootulp/celestia-core/blob/97c3cce834c0fd0568bebf44c6c2fde6b06626e5/node/node.go#L334-L340):

```go
	logger.Info("config.Mempool.Version", "config.Mempool.Version", config.Mempool.Version)
```

and running
```shell
rm -rf ~/.celestia-app
./scripts/single-node.sh
cat ~/.celestia-app/config.toml // verified that mempool version is "v1"
```

the mempool version configured in `config.toml` is respected so after stopping the node above, overriding the version to `"0"`, and restarting the node, logs indicate that the v0 mempool is in fact being used. E.g.

```
2:48PM INF config.Mempool.Version config.Mempool.Version=v0
```

evan-forbes:
@rootulp was this breaking?

rootulp:
I think so because configuration changes can be considered breaking changes

evan-forbes:
there's no harm in leaving it, but I actually think this is one of the few changes that we can make that is not breaking given that it is strictly a default change (anyone can change this at any time) and its part of the mempool (not consensus critical)

rootulp:
I think it's subject to interpretation but assume we don't mark this change as breaking and a validator upgrades from a celestia-app release before this change to a release after this change. Assume the validator intends to use the FIFO mempool but hasn't overriden any config because the default config previously did what they want. The validator operator  upgrades blindly because the release notes don't mark this as a breaking change so they don't expect any visible behavior change. After the upgrade, the validator (contrary to their desire) is using the prioritized mempool. 

I see your point that it's not consensus critical and is overridable so I think this change could be interpreted as non-breaking but if there's ambiguity on whether a change should be marked breaking vs. non-breaking, I think it's safer to mark as breaking to explicitly call out to consumers behavior changes that they should be aware of when upgrading. In the future (i.e. post mainnet launch), I think we'll have to be much more careful about releasing / labeling breaking changes and make an effort to minimize the number of such changes.

evan-forbes:
> I think we'll have to be much more careful about releasing / labeling breaking changes and make an effort to minimize the number of such changes.

Yes I agree :slightly_smiling_face:. As you and I have discussed in the past, almost everything we do is breaking (except changes to the default config imho :sweat_smile:). 

> upgrades from a celestia-app release before this change to a release after this change

Doesn't this require resetting their config? If they only upgrade binaries (most situations) or use their own custom config (all validator setups and most power users) thier mempool will be the same, that's why I don't see this as breaking.

> I think it's safer to mark as breaking...

okay let's keep it breaking :slightly_smiling_face:.

rootulp:
Oh very good point, the situation I described would involve resetting their config which wouldn't happen if strictly upgrading binaries.

PR #1019, opened 2022-11-16
chore: add support for QGB in mergify

SweeXordious:
## Overview

- [x] mergify changes
- [ ] `qgb-integration` branch rename to `qgb-v1` (I don't have enough permissions)

## Checklist

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -31,3 +31,11 @@ pull_request_rules:
       backport:
         branches:
           - v0.9.x
+  - name: backport commits to qgb-v1 branch
+    conditions:
+      - base=main
+      - label=S:backport-to-qgb-v1
+    actions:
+      backport:
+        branches:
+          - qgb-v1


evan-forbes:
I like the idea of automating this, but wouldn't essentially open a second PR for every PR?

SweeXordious:
@evan-forbes Any other ideas? I agree, it is too much, but it is the easiest I could think of, aside from cherry picking. I'm fine cherry picking btw

evan-forbes:
I don't have a better solution in mind unfortunately. Is there a way that we could batch mergify? perhaps update the qgb-integration branch once a week or something? We shouldn't be using it for too long hopefully, so I'm also ok with just manually merging on occasion

SweeXordious:
Will close then, and update the branch manually occasionally, every week is good.
For the QGB branch rename, I think it would be better to rename so that anyone knows that it's QGB V1.  What do you think?

Issue #1022, opened 2022-11-16
Upgrade to latest celestia-core release


rootulp:
## Context

https://github.com/celestiaorg/celestia-core/releases/tag/v1.6.0-tm-v0.34.20 was released but this repo still uses `v1.5.0-tm-v0.34.20`. See https://github.com/celestiaorg/celestia-app/blob/778a18e19338832848607c62e895a8910ac33197/go.mod#L185

## Proposal

Upgrade to the latest celestia-core release

rootulp:
Would prefer if https://github.com/celestiaorg/celestia-core/pull/888 merged and was included in a release so that we can upgrade directly to that

Issue #995, opened 2022-11-10
Rename `message` => `blob` where applicable


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/978#discussion_r1016812915

## Problem

The codebase currently uses the previous term `message` in a number of locations where `blob` is more accurate after https://github.com/celestiaorg/celestia-app/pull/978 merged.

## Proposal

On a case by base basis, rename `message`, `msg`, or `data` => `blob`. Note: Cosmos SDK messages should still be called messages. We should only rename instances that refer to the data a user submits to be published to the Celestia blockchain.

- [x] Update code: https://github.com/celestiaorg/celestia-app/pull/1060
- [ ] Update ADRs: https://github.com/celestiaorg/celestia-app/issues/1531
- [ ] Update specs: https://github.com/celestiaorg/celestia-app/issues/1532

rootulp:
Also instances of `data` may need to be renamed to `blob`. Ex:

https://github.com/celestiaorg/celestia-app/blob/b81723b114282f98007cb7a9066b0e69efebe98b/app/process_proposal.go#L26-L27

rootulp:
Will do this soon because it's leaking into other unrelated PRs. Unfortunately it may cause merge conflicts will several in-progress open PRs but the merge conflicts should be easy to resolve

rootulp:
https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-001-abci%2B%2B-adoption.md contains code blocks which were helpful at the time of reviewing the ADR but the code blocks have become stale so renaming portions of it to resolve this issue seems futile. We can either:
1. remove all the code blocks that have diverged from their real implementations in `main`
2. update all the code blocks

rootulp:
Sure, will leave ADRs as-is for now.

rootulp:
Spun out two new issues so closing this

PR #1023, opened 2022-11-16
chore!: upgrade to celestia-core `v1.8.0-tm-v0.34.20`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1022

Does not rename all instances of `message` to `blob` see https://github.com/celestiaorg/celestia-app/issues/995

### DIFF
--- a/app/malleate_txs.go
+++ b/app/malleate_txs.go
@@ -16,11 +16,11 @@ func malleateTxs(
 	squareSize uint64,
 	txs parsedTxs,
 	evd core.EvidenceList,
-) ([][]byte, []*core.Message, error) {
+) ([][]byte, []*core.Blob, error) {
 	// trackedMessage keeps track of the pfb from which it was malleated from so
 	// that we can wrap that pfb with appropriate share index
 	type trackedMessage struct {
-		message     *core.Message
+		message     *core.Blob
 		parsedIndex int
 	}
 
@@ -47,7 +47,7 @@ func malleateTxs(
 	})
 
 	// split the tracked messagse apart now that we know the order of the indexes
-	msgs := make([]*core.Message, len(trackedMsgs))
+	msgs := make([]*core.Blob, len(trackedMsgs))
 	parsedTxReverseIndexes := make([]int, len(trackedMsgs))
 	for i, tMsg := range trackedMsgs {
 		msgs[i] = tMsg.message
--- a/app/malleate_txs.go
+++ b/app/malleate_txs.go
@@ -16,7 +16,7 @@ func malleateTxs(
 	squareSize uint64,
 	txs parsedTxs,
 	evd core.EvidenceList,
-) ([][]byte, []*core.Blob, error) {
+) ([][]byte, []core.Blob, error) {
 	// trackedMessage keeps track of the pfb from which it was malleated from so
 	// that we can wrap that pfb with appropriate share index
 	type trackedMessage struct {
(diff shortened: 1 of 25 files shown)


app/malleate_txs.go
```diff
@@ -47,7 +47,7 @@ func malleateTxs(
 	})
 
 	// split the tracked messagse apart now that we know the order of the indexes
-	msgs := make([]*core.Message, len(trackedMsgs))
+	msgs := make([]*core.Blob, len(trackedMsgs))
```
evan-forbes, line 50: [optional suggestion]
do you think using non pointers here (and ~20 lines above) would simplify things at all? for example, perhaps we could remove the extra conversion later in `PrepareProposal`

https://github.com/rootulp/celestia-app/blob/f888edab22c5553bb7b2a7d7c1806a567590016e/app/prepare_proposal.go#L44-L50

app/prepare_proposal.go
```diff
@@ -33,19 +33,27 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
 		panic(err)
 	}
 
+	blobs := make([]core.Blob, len(blobPointers))
+	for i, blob := range blobPointers {
+		blobs[i] = core.Blob{
+			NamespaceId: blob.GetNamespaceId(),
+			Data:        blob.GetData(),
+		}
+	}
```
evan-forbes, line 50: [optional]

see other comment on how we might be able to remove this extra conversion. if not, do you think we would get any benefit out of dereferencing instead of creating an entire new struct?

it doesn't really matter, as we are going to change this soon. pls move along if its not a quick change :pray:

app/test/process_proposal_test.go
```diff
@@ -165,7 +165,7 @@ func TestMessageInclusionCheck(t *testing.T) {
 // 		require.NoError(t, err)
 
 // 		require.NoError(t, err)
-// 		eds, err := da.ExtendShares(input.BlockData.OriginalSquareSize, shares)
+// 		eds, err := da.ExtendShares(input.BlockData.SquareSize, shares)
```
rootulp, line 168: I think we should be deleting these tests rather than leaving them commented out. Created https://github.com/celestiaorg/celestia-app/issues/1026

app/test/prepare_proposal_test.go
```diff
@@ -28,22 +28,22 @@ func TestPrepareProposal(t *testing.T) {
...
-		input            abci.RequestPrepareProposal
-		expectedMessages []*core.Message
-		expectedTxs      int
+		input         abci.RequestPrepareProposal
+		expectedBlobs []core.Blob
+		expectedTxs   int
 	}
 
-	firstNS := []byte{2, 2, 2, 2, 2, 2, 2, 2}
+	firstNamespace := []byte{2, 2, 2, 2, 2, 2, 2, 2}
```
evan-forbes, line 36: [non-blocking][for future reference]
is this change related at all? its already completed, so **pls keep**, but if its 100% not related, in the future I would appreciate it if we try to not arbitrarily change variable names and put them into unrelated PRs. I don't mind this specific change, and I don't wan't to discourage renaming things, and I've been guilty of this too, but if any of us did it would be silly to block on an otherwise perfect PR.

thx :slightly_smiling_face:

evan-forbes (dismissed):
left one possible change, but we will likely change it soon anyway, so pls feel free to ignore if it takes too long

app/test/prepare_proposal_test.go
rootulp: Agreed completely. This change isn't necessary in this PR. While working on this PR, I noticed that we use two names to refer to the same thing `ns` and `namespace`. I started correcting in this PR which was a mistake, should've been a new issue and PR.

app/malleate_txs.go
rootulp: Agreed. I took a stab at this before but I wasn't sure why `malleateTxs` returned a slice of pointers to messages instead of a slice of messages. Without unit tests on `malleateTxs` I'm hesitant to make changes but will try again.

app/prepare_proposal.go
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/1023/commits/3543e6d9c92760542b0396ac6a918c69b0e4be22

evan-forbes (approved):

Issue #1026, opened 2022-11-17
Remove commented tests


rootulp:
## Problem
This repo contains a few instances instances of commented out test code. The commented code doesn't benefit from Go refactoring tools (like VSCode rename symbol) that gets applied to uncommented Go code. As a result, the commented test code becomes even less maintained and more difficult to uncomment / re-apply.

## Proposal
I think we should remove the commented code and optionally cut individual issues to track adding uncommented versions of the test back if desired. Motivation for deleting commented code: https://kentcdodds.com/blog/please-dont-commit-commented-out-code

Candidates for removal:
1. https://github.com/celestiaorg/celestia-app/blob/778a18e19338832848607c62e895a8910ac33197/app/test/process_proposal_test.go#L121-L175
2. https://github.com/celestiaorg/celestia-app/blob/778a18e19338832848607c62e895a8910ac33197/pkg/prove/proof_test.go#L173-L226
3. https://github.com/celestiaorg/celestia-app/blob/778a18e19338832848607c62e895a8910ac33197/pkg/shares/shares_test.go#L20-L170

PR #1027, opened 2022-11-17
chore: delete commented tests

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1026


pkg/prove/proof_test.go
```diff
@@ -170,61 +170,6 @@ func TestTxShareIndex(t *testing.T) {
 	}
 }
 
-// TODO: Uncomment/fix this test after we've adjusted tx inclusion proofs to
```
SweeXordious, line 173: I think we should create an issue to fix this one as it seems important

SweeXordious (approved):

pkg/prove/proof_test.go
MSevey: an alternative to just deleting would be to just replace the test code with `t.Skip("this test needs to be re-implemented")` 

but if the goal is to eventually redo these tests, we should definitely have issues for them

evan-forbes: [blocking] 
if we don't have an issue for this, lets create one and merge

evan-forbes (approved):

pkg/prove/proof_test.go
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/1049

PR #1029, opened 2022-11-17
feat: Gas consumption script

evan-forbes:
## Overview

This PR adds a test suite that we can use to measure the gas consumed by standard txs and modifies the network tests so that we can use custom genesis parameters. We still have to add more tx types and some visualization would be nice.

## Checklist

### DIFF
--- a/testutil/testnode/full_node.go
+++ b/testutil/testnode/full_node.go
@@ -43,7 +43,8 @@ func New(
 	supressLog bool,
 	genState map[string]json.RawMessage,
 	gasMon *gasmonitor.Decorator,
-	kr keyring.Keyring) (*node.Node, srvtypes.Application, Context, error) {
+	kr keyring.Keyring,
+) (*node.Node, srvtypes.Application, Context, error) {
 	var logger log.Logger
 	if supressLog {
 		logger = log.NewNopLogger()
--- a/testutil/testnode/full_node.go
+++ b/testutil/testnode/full_node.go
@@ -1,6 +1,7 @@
 package testnode
 
 import (
+	"encoding/json"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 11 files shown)


testutil/testnode/full_node.go
```diff
@@ -135,3 +139,63 @@ func DefaultTendermintConfig() *config.Config {
...
+
+// DefaultNetwork creates an in-process single validator celestia-app network
+// using test friendly defaults. These defaults include fast block times and
+// funded accounts. The returned client.Context has a keyring with all of the
+// funded keys stored in it.
+func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func(), accounts []string, cctx Context) {
+	// we create an arbitrary number of funded accounts
+	for i := 0; i < 300; i++ {
+		accounts = append(accounts, tmrand.Str(9))
+	}
```
MSevey, line 180: Append re-allocates memory. So it is more memory performant to initialize the slice and then fill it.
```go
size := 300
accounts = make([]string, 0, size)
for i:=0; i<size; i++ {
  accounts[i] = tmrand.Str(9)
}
```

evan-forbes: thx for the reminder, we actually have a linter for this but I'm not sure why it didn't get hit

evan-forbes:
this is still a draft, and is not worth reviewing atm

evan-forbes:
created a few plots using data collected using the test here
[gas_investigation.pdf](https://github.com/celestiaorg/celestia-app/files/10039380/gas_investigation.pdf)

evan-forbes:
We very likely won't run this test to trace the gas consumption very often, and there is quite a bit of code, so we should just keep them in their own repo.

The only thing this PR should do is enable the investigations to be ran by including the gas consumption monitoring logic imo

evan-forbes:
closing this, as we don't have plans on merging

PR #1030, opened 2022-11-17
Feat/emit typed event

### LINKED ISSUE #968: Switch usage from `EmitEvent` to `EmitTypedEvent`
### Summary

Change `EmitEvent` in favour of `EmitTypedEvent`

### Problem Definition

`EmitEvent` has been deprecated

### Proposal

We should switch to `EmitTypedEvent` and define relavnt type(s) in protobuf. Changes need to be done both in `payment` and `qgb` module, primarily `x/payment/keeper/keeper.go` and `x/qgb/keeper/keeper_attestation.go`

### For Admin Use

- [ ] Not duplicate issue

rahulghangas:
- [x] Closes #968


x/blob/keeper/keeper.go
```diff
@@ -61,9 +61,13 @@ func (k Keeper) PayForBlob(goCtx context.Context, msg *types.MsgPayForBlob) (*ty
 	gasToConsume := uint64(shares.MsgSharesUsed(int(msg.BlobSize)) * GasPerMsgShare)
 	ctx.GasMeter().ConsumeGas(gasToConsume, payForBlobGasDescriptor)
 
-	ctx.EventManager().EmitEvent(
-		types.NewPayForBlobEvent(sdk.AccAddress(msg.Signer).String(), msg.GetBlobSize()),
-	)
```
vidhanarya, line 66: Can remove this (`types.NewPayForBlobEvent`) method as it's not being used, otherwise can delegate
```go
&types.EventPayForBlob{
	Signer:   sdk.AccAddress(msg.Signer).String(),
	BlobSize: msg.GetBlobSize(),
}
```
to above mentioned method and change the signature.

proto/blob/event.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package blob;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/blob/types";
+
+// EventPayForBlob defines an event that is emitted with the signer and blob size
+// at the end of a pay for blob txn
+message EventPayForBlob {
```
vidhanarya, line 10: So we are changing event.type from `payforblob` to `blob.EventPayForBlob` or am I missing something here? If we are changing event type then relevant test (`TestSubmitWirePayForBlob`) in `integration_test.go` can be updated.

rahulghangas: Yes, there's a pending task in this PR for the same

rahulghangas:
@evan-forbes since we're changing the event name, in order to follow convention from the typed events ADR in cosmos-sdk, this will be a breaking change

proto/blob/event.proto
vidhanarya: My bad, I overlooked the draft status and checklist (used to GitLab PRs 😅 ). My apologies 🙏

proto/blob/events.proto
```diff
@@ -0,0 +1,13 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/blob/types";
+
+// EventPayForBlob defines an event that is emitted with the signer and blob size
+// at the end of a pay for blob txn
+message EventPayForBlob {
+  string signer = 1;
+  uint64 blob_size = 2;
+}
```
SweeXordious, line 13:
```suggestion
}

```

x/blob/keeper/keeper.go
```diff
@@ -61,9 +61,12 @@ func (k Keeper) PayForBlob(goCtx context.Context, msg *types.MsgPayForBlob) (*ty
 	gasToConsume := uint64(shares.MsgSharesUsed(int(msg.BlobSize)) * GasPerMsgShare)
 	ctx.GasMeter().ConsumeGas(gasToConsume, payForBlobGasDescriptor)
 
-	ctx.EventManager().EmitEvent(
+	err := ctx.EventManager().EmitTypedEvent(
 		types.NewPayForBlobEvent(sdk.AccAddress(msg.Signer).String(), msg.GetBlobSize()),
 	)
+	if err != nil {
+		return &types.MsgPayForBlobResponse{}, nil
```
SweeXordious, line 68:
```suggestion
		return &types.MsgPayForBlobResponse{}, err
```

SweeXordious (dismissed):
Good catch :+1: the older way was deprecated.

x/blob/keeper/keeper.go
rahulghangas: Whoops, thanks for catching this. I'll change it right away

x/blob/keeper/keeper.go
```diff
@@ -61,9 +61,12 @@ func (k Keeper) PayForBlob(goCtx context.Context, msg *types.MsgPayForBlob) (*ty
 	gasToConsume := uint64(shares.MsgSharesUsed(int(msg.BlobSize)) * GasPerMsgShare)
 	ctx.GasMeter().ConsumeGas(gasToConsume, payForBlobGasDescriptor)
 
-	ctx.EventManager().EmitEvent(
+	err := ctx.EventManager().EmitTypedEvent(
 		types.NewPayForBlobEvent(sdk.AccAddress(msg.Signer).String(), msg.GetBlobSize()),
```
SweeXordious, line 65:
```suggestion
		types.NewPayForBlobEvent(msg.Signer, msg.GetBlobSize()),
```
I guess the `msg.Signer` already contains the account address in Bech32 format. What do you think?

evan-forbes: second this, we should be able to just use the signer. We can test manually by looking at the events in one of the integration tests

evan-forbes:
what's left for this? just reviewer feedback?

x/blob/keeper/keeper.go
rahulghangas: It doesn't look like that the conversion is as is, I'll confirm via a test

rahulghangas:
> Where does this happen? It doesn't appear applicable to this PR.

It doesn't, I've removed it from the checklist. Initially the idea was to change the test because the attribute key was hardcoded, but I've made it such that it infers the type name from the struct itself

x/blob/keeper/keeper.go
SweeXordious: @rahulghangas Did you manage to check if it is the same?

proto/blob/events.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
rootulp, line 1: [blocking] Based on the following, I would expect the filename to be `event.proto` and not `events.proto`:

> Typed Events are Protobuf-defined [messages](https://docs.cosmos.network/main/architecture/adr-032-typed-events) used by the Cosmos SDK for emitting and querying Events. They are defined in a event.proto file, on a per-module basis and are read as proto.Message. Legacy Events are defined on a per-module basis in the module's /types/events.go file. They are triggered from the module's Protobuf [Msg service](https://docs.cosmos.network/main/building-modules/msg-services) by using the [EventManager](https://docs.cosmos.network/main/core/events#eventmanager).

Ref: https://docs.cosmos.network/main/core/events

rootulp (commented):
[not blocking] can we please update this section https://github.com/celestiaorg/celestia-app/blob/29d47afbfa4b05ef81e82d0d3412e034a09b3098/x/blob/README.md?plain=1#L30-L32 to remove the permalink to the outdated event? Proposal:

```markdown
## Events

- `PayForBlobEvent` is emitted with the signer's address and size of the message that is paid for.
```

Alternatively, we can also copy something like this: https://github.com/cosmos/cosmos-sdk/blob/16c772d900f530d16518491b19f60cfef9d713af/x/bank/README.md?plain=1#L339-L353

x/blob/keeper/keeper.go
rahulghangas: @SweeXordious `msg.Signer` is the raw form while the `.String()` method encodes it in bech32. Removing the above logic will be breaking change

SweeXordious: I tried an example and apparently the `msg.Signer` contains:
```
cosmos1n5c5wmz5s2n0d8zpr5ldcl45nkjclaqktls7gd
```

while `sdk.AccAddress(msg.Signer).String()` returns:
```
cosmos1vdhhxmt0wvckudtrx4mk6734wvexuvry8pa8qu34d3jxxmp5x4hxk6nrd3shz6m5d3enwemycqe9zf
```

Which one do we want?

x/blob/README.md
```diff
@@ -29,14 +29,23 @@ The malleation process occurs during the PrepareProposal step.
 
 ## Events
 
-- [`NewPayForBlobEvent`](https://github.com/celestiaorg/celestia-app/pull/213/files#diff-1ce55bda42cf160deca2e5ea1f4382b65f3b689c7e00c88085d7ce219e77303dR17-R21) is emitted with the signer's address and size of the blob that is paid for.
+The bank module emits the following events:
```
rootulp, line 32:
```suggestion
The blob module emits the following events:
```

x/blob/types/events.go
```diff
@@ -1,23 +1,13 @@
...
-)
-
-// NewPayForBlobEvent constructs a new payforblob sdk.Event
-func NewPayForBlobEvent(signer string, size uint64) sdk.Event {
-	return sdk.NewEvent(
-		EventTypePayForBlob,
-		sdk.NewAttribute(AttributeKeySigner, signer),
-		sdk.NewAttribute(AttributeKeySize, strconv.FormatUint(size, 10)),
-	)
+// NewPayForBlobEvent constructs a new pay for blob sdk.Event
```
rootulp, line 7: [blocking] `sdk.Event` appears incorrect here

```suggestion
// NewPayForBlobEvent returns a new EventPayForBlob
```

x/blob/keeper/keeper.go
rootulp: After we align on one, should we add a unit test that verifies an event is emitted with the appropriate signer?

rahulghangas: I'm leaning towards first, since it's more compact and helps us reduce on-chain storage

PR #1031, opened 2022-11-17
chore: move default const values to appconsts module

### LINKED ISSUE #993: Move default values for blob params to appconsts module
### Summary

Default values for blob params are currently hardcoded and should be moved to `appconsts` module

rahulghangas:
Moves default values for min/max swaure size and gas per blob byte to appconsts module. 
Renames `MinSquareSize` and `MaxSquareSize` in appconsts module to `DefaultMinSquareSize` and `DefaultMaxSquareSize`

- [x] closes #993


rootulp (commented):
Sorry but I don't understand the motivation behind this change (see https://github.com/celestiaorg/celestia-app/issues/993).

pkg/appconsts/appconsts.go
```diff
@@ -40,22 +40,22 @@ const (
...
 	// Corresponds to AVAILABLE_DATA_ORIGINAL_SQUARE_MAX in the spec.
 	// 128*128*256 = 4 Megabytes
 	// TODO(ismail): settle on a proper max square
 	// if the square size is larger than this, the block producer will panic
-	MaxSquareSize = 128
+	DefaultMaxSquareSize = 128
 	// MaxShareCount is the maximum number of shares allowed in the original data square.
 	// if there are more shares than this, the block producer will panic.
-	MaxShareCount = MaxSquareSize * MaxSquareSize
+	MaxShareCount = DefaultMaxSquareSize * DefaultMaxSquareSize
```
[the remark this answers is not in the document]
rootulp: Disregard, I just learned of https://github.com/celestiaorg/celestia-app/issues/1032

SweeXordious (commented):
:fire: waiting for the transient store :rocket: 
Left couple blocking comments. Correct me if I am missing something please :cat:

x/blob/types/params.go
```diff
@@ -97,7 +98,7 @@ func validateMinSquareSize(v interface{}) error {
 	return nil
 }
 
-// validateMaxSquareSize validates the MaxSquareSize param
+// validateMaxSquareSize validates the DefaultMaxSquareSize param
```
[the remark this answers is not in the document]
rahulghangas: any square size

x/blob/types/params.go
```diff
@@ -111,6 +112,7 @@ func validateMaxSquareSize(v interface{}) error {
 	return nil
 }
 
+// validateMinMaxSquareSizeOrder validates that minSquareSize if less than or equal to maxSquareSize
```
rootulp, line 115:
```suggestion
// validateMinMaxSquareSizeOrder validates that minSquareSize is less than or equal to maxSquareSize
```

x/blob/types/params.go
```diff
@@ -111,6 +112,7 @@ func validateMaxSquareSize(v interface{}) error {
...
 func validateMinMaxSquareSizeOrder(minSquareSize, maxSquareSize uint32) error {
 	if minSquareSize > maxSquareSize {
 		return fmt.Errorf("max square size cannot be less than min square size")
```
rootulp, line 118: [no change needed] because not modified in this PR but proposal for the error message to match the ordering in the conditional so that it's easier to review.

```suggestion
	if minSquareSize > maxSquareSize {
		return fmt.Errorf("min square size cannot be greater than max square size")
```

rahulghangas:
> I still don't understand the motivation behind this change but I don't have any blocking feedback. I can approve after merge conflicts are resolved.

min/max square size is used throughout (tests, standalone functions, etc.). I'm moving min/max square size to be app level params, so that they can be passed in everywhere easily. It's easier to move defaults values to `appconsts` and refactor the params specific changes from there

rootulp (approved):

pkg/appconsts/appconsts.go
```diff
@@ -40,22 +40,22 @@ const (
...
 
-	// MinSquareSize depicts the smallest original square width. A square size smaller than this will
+	// DefaultMinSquareSize depicts the smallest original square width. A square size smaller than this will
 	// cause block producer to panic
-	MinSquareSize = 1
+	DefaultMinSquareSize = 1
 	// MinshareCount is the minimum shares required in an original data square.
-	MinShareCount = MinSquareSize * MinSquareSize
+	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
```
evan-forbes, line 58: [do in a different PR]

we could even put these in their own blob file as this one is getting crowded

evan-forbes (approved):

PR #1033, opened 2022-11-17
chore: rename gas descriptor to use `blob` instead of `data`

rahulghangas:
As title


rootulp (approved):

evan-forbes:
related to #995 

this might actually be a breaking change if anything were to rely on this event for anything. adding the tag just in case

evan-forbes (approved):

Issue #658, opened 2022-08-24
Investigate and finalize resource pricing 


evan-forbes:
We need to investigate and decide on resource pricing. While we should easily be able to change the default gas configs found in tendermint and the sdk, its unclear how specific we can get as to adding multipliers to specific types of sdk txs.

liamsi:
My understanding is that for PayForData we want to measure account for the bytes but ideally we also want to add a constant cost per PfD s.t. batching is encouraged in the future. I do think this is possible and not difficult in the SDK.

For the other Txs types (sdk messages) I guess we rather want clarity on how exactly they are priced and if we want to change any of that default behaviour? One message type that comes to mind are the IBC related txs (so this issue/comment is kinda related here: https://github.com/celestiaorg/celestia-app/issues/235#issuecomment-1190439900).

evan-forbes:
The goal of this issue to is to document the formula used for each transaction type, so that we have a concrete understanding of how gas is deducted.

evan-forbes:
in #1029 , we're creating a script to run standard cosmos-sdk transactions on an inprocess network that has the ability to change genesis params and monitor gas. This should help us investigate gas params.

evan-forbes:
listing this as incentivized testnet stretch milestone, as we can change parameters and add docs at any time after the feature freeze without breaking anything. We also have a good idea of how much gas roughly a PFB takes up relative to other tx types, and we have code to trace the gas consumption of any transaction in any scenario should we need more precise data for a given situation.

evan-forbes:
referencing a comment where we discuss the gas costs of IBC. Notably, following the defaults of IBC, we are not charging gas per signature verified when updating the client.

https://github.com/celestiaorg/celestia-app/issues/1125#issuecomment-1354812142

evan-forbes:
can close imo since we have the above analysis and have set the default min gas price to .1utia

PR #1038, opened 2022-11-18
chore!: refactor the testnode to expose the genesis state 

### LINKED ISSUE #1037: Expose genesis state for the testnode
We want to be able to modify the genesis state when running the testnode. Currently we can't do that do to it always using the default genesis state. We should refactor it so that we can modify that state.

blocking #658

evan-forbes:
## Overview

We need to be able to modify the genesis state when using the testnode. This PR refactors the testnode to allow for that. It also includes a helper function that handles creation of this genesis state and simplify the process of starting an in process node. 

closes #1037 
part of/blocking #658 
spun out of and blocking #1029 

## Checklist

### DIFF
--- a/testutil/testnode/full_node.go
+++ b/testutil/testnode/full_node.go
@@ -1,6 +1,7 @@
 package testnode
 
 import (
+	"encoding/json"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


testutil/testnode/full_node.go
```diff
@@ -135,3 +140,64 @@ func DefaultTendermintConfig() *config.Config {
 	tmCfg.Mempool.MaxTxBytes = 22020096 // 21MB
 	return tmCfg
 }
+
+// DefaultGenesisState will create
```
SweeXordious, line 144:
```suggestion
// DefaultGenesisState will create a default genesis state, make the `fundedAccounts` validators, and fund them.
```

SweeXordious (dismissed):
:rocket:

testutil/testnode/full_node.go
rootulp:
```suggestion
// DefaultGenesisState returns a default genesis state and a keyring with
// accounts that have coins. The keyring accounts are based on the
// fundedAccounts parameter.
```

slightly different doc comment proposal because I don't see how this function makes the `fundedAccounts` validators

rootulp (approved):

testutil/testnode/full_node.go
evan-forbes: good catches, ended up going with the second one as they are not validators

evan-forbes:
for posterity, the testnode code could still use some serious refactors to make it easier to understand and more useful. There's only so much we can do to make the process of starting an in process node simpler, but yeah, this is only a step in the right direction instead of a final solution.

rootulp (approved):

SweeXordious (approved):

rahulghangas (approved):

PR #1039, opened 2022-11-18
doc: document NMT wrapper in prepare_proposal

SweeXordious:
Adds a small note to specify that the NMT wrapper is used to extend the shares


pkg/da/data_availability_header.go
```diff
@@ -6,7 +6,7 @@ import (
 	"fmt"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
-	"github.com/celestiaorg/celestia-app/pkg/wrapper"
+	nmtwrapper "github.com/celestiaorg/celestia-app/pkg/wrapper"
```
evan-forbes, line 9: [not blocking comment]
I like this change, but we use wrapper everywhere else. Not going to block on this, but perhaps in the future we should rename the package to force this across all repos

pkg/da/data_availability_header.go
```diff
@@ -68,7 +68,8 @@ func ExtendShares(squareSize uint64, shares [][]byte) (*rsmt2d.ExtendedDataSquar
 		)
 	}
 	// here we construct a tree
-	return rsmt2d.ComputeExtendedDataSquare(shares, appconsts.DefaultCodec(), wrapper.NewConstructor(squareSize))
+	// Note: uses the nmt wrapper to construct the tree.
+	return rsmt2d.ComputeExtendedDataSquare(shares, appconsts.DefaultCodec(), nmtwrapper.NewConstructor(squareSize))
```
evan-forbes, line 72: [nit non-blocking]
perhaps we could merge this we the above outdated comment (that doesn't really make sense by itself)?

`// we construct the data root using the nmt wrapper`

app/prepare_proposal.go
```diff
@@ -59,6 +59,8 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 	}
 
 	// erasure the data square which we use to create the data root.
+	// Note: uses the nmt wrapper to construct the tree.
+	// checkout pkg/wrapper/nmt_wrapper.go for more information.
```
evan-forbes, line 63: [I'll do this in a different PR]
oh woah we should definitely add docs to https://github.com/celestiaorg/celestia-app/blob/24626ff8d00854d7e0027abcfdf50de1a2aee6a3/pkg/da/data_availability_header.go#L52

evan-forbes (approved):

pkg/da/data_availability_header.go
SweeXordious: `// Note: uses the nmt wrapper to construct the tree.` => `// we construct the data root using the nmt wrapper`?

pkg/da/data_availability_header.go
SweeXordious: I'll create an issue for this: https://github.com/celestiaorg/celestia-app/issues/1069

app/prepare_proposal.go
SweeXordious: Thanks :+1:

evan-forbes:
sure why not, but lets stay consistent with the naming https://github.com/celestiaorg/celestia-app/issues/1069#issuecomment-1327702847

pkg/da/data_availability_header.go
evan-forbes: yep, makes sense!

pkg/da/data_availability_header.go
evan-forbes: since #1069 is unplanned, do you mind reverting this?

evan-forbes (commented):
apologies for letting this slip through the cracks @SweeXordious 

left one comment to use the same name for the wrapper across the board, then should finally merge

SweeXordious:
done, all we need is an approve :D

pkg/da/data_availability_header.go
SweeXordious: done

Issue #1045, opened 2022-11-21
QGB remove Bridge ID


SweeXordious:
As per https://github.com/celestiaorg/quantum-gravity-bridge/issues/71, we will also make the change on Celestia-app side to remove the Bridge ID

PR #1046, opened 2022-11-21
chore: remove unnecessary QGB BridgeID

SweeXordious:
## Overview
Closes https://github.com/celestiaorg/celestia-app/issues/1045

## Checklist


x/qgb/types/abi_consts.go
```diff
@@ -132,7 +132,6 @@ var (
 
 	VsDomainSeparator ethcmn.Hash
 	DcDomainSeparator ethcmn.Hash
-	BridgeID          = ethcmn.HexToHash("Evm_Celestia_Bridge") //  TODO to be removed afterwards
```
evan-forbes, line 135: [question]
do we use the bridge id when creating the hashes when setting the attestation requests, or do the orchestrator's handle that?

evan-forbes (dismissed):

x/qgb/types/abi_consts.go
SweeXordious: Good point, I was just making the changes on the QGB repo, and I noticed that it is being passed as param. I will open a PR there removing it, then, after it is merged, finish the updates here.

SweeXordious:
depends on https://github.com/celestiaorg/quantum-gravity-bridge/pull/130 and https://github.com/celestiaorg/celestia-app/issues/1047

SweeXordious:
Reminder: also update the internal ABI not to contain the bridge id.

x/qgb/types/abi_consts.go
evan-forbes: wait, do we need it there?

SweeXordious: It is used to create the sign bytes of the valset for example. However, we don't need it. Since, we don't care about replaying transactions. It's actually good for us to replay transaction from an evm chain to another

SweeXordious: Example of use: 

https://github.com/celestiaorg/celestia-app/blob/8604be960b0565df66a6c4da79371f713ca62798/x/qgb/types/valset.go#L29-L56

evan-forbes: > It's actually good for us to replay transaction from an evm chain to another

to prevent replayability? I think we should eventually remove it entirely, as validators will not be using the same key to sign things for multiple chains anyway

SweeXordious: Ah, sorry, I meant sending the same byte code. Since, the txs byte code will be the same for all EVM chains

evan-forbes (approved):
super :fire:

PR #1051, opened 2022-11-22
chore!(testnode): return errors in cleanup functions and wait for cleanup

evan-forbes:
## Overview

@renaynay found that we weren't actually waiting for the tendermint node to shutdown during cleanup. This might be causing an issue when running multiple different nodes in CI (not locally). We also were not returning errors for the cleanup functions, which could be silencing useful errors. 

## Checklist

### DIFF
--- a/testutil/testnode/full_node_test.go
+++ b/testutil/testnode/full_node_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/celestiaorg/celestia-app/testutil"
 	"github.com/celestiaorg/celestia-app/testutil/namespace"
 	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	abci "github.com/tendermint/tendermint/abci/types"
 	tmrand "github.com/tendermint/tendermint/libs/rand"
@@ -18,7 +19,7 @@ import (
 type IntegrationTestSuite struct {
 	suite.Suite
 
-	cleanups []func()
+	cleanups []func() error
 	accounts []string
 	cctx     Context
 }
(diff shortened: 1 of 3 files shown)


testutil/testnode/full_node_test.go
```diff
@@ -56,7 +57,8 @@ func (s *IntegrationTestSuite) SetupSuite() {
 func (s *IntegrationTestSuite) TearDownSuite() {
 	s.T().Log("tearing down integration test suite")
 	for _, c := range s.cleanups {
-		c()
+		err := c()
+		require.NoError(s.T(), err)
```
SweeXordious, line 61: [Blocking]
Since we have multiple cleanups, I guess on this level, we shouldn't really care about cleanups errors.
In most cases, one cleanup will succeed, but the other ones might fail because one of them succeeded and already closed the connections (can't close an already closed connection).
So, I guess this require is not needed

rootulp: > In most cases, one cleanup will succeed, but the other ones might fail because one of them succeeded and already closed the connections

It looks like the only cleanups that are possible in `s.cleanups` are `stopNode` and `cleanupGRPC`. `cleanupGRPC` never returns an error so this line can only fail when `stopNode` returns an error. Based on the PR description, I think it is intended to throw an error when that happens

rootulp (approved):

testutil/testnode/full_node_test.go
SweeXordious: I agree that the errors should be returned as per the PR goal. However, unless we have specific code to handle those errors and act upon them, which the node team would, we shouldn't be checking: `require.NoError(s.T(), err)`. Because, errors will happen.
Let me know if I am missing something or I'm not being clear.

evan-forbes: I think the errors that were occurring were a bug. I think we should check to ensure that there aren't any errors. https://github.com/celestiaorg/celestia-app/pull/1051/commits/4cde987c90b95873bf4811251c61a6d59aa5e543

wdyt?

SweeXordious: Aaah, thanks :+1: I had a different reasoning.

SweeXordious (approved):

rootulp (approved):

PR #1052, opened 2022-11-22
docs: add qgb module readme

SweeXordious:
## Overview
Closes https://github.com/celestiaorg/celestia-app/issues/501

## Checklist

### DIFF
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -1,21 +1,12 @@
 # `x/qgb`
 
 ## Abstract
+
 This module contains the [Quantum Gravity Bridge](https://blog.celestia.org/celestiums/) (QGB) state machine implementation.
 
 ## State machine
 
-The QGB state machine handles the creation of the attestations requests:
-
-https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/x/qgb/types/attestation.go#L15-L23
-
-These latter is either data commitments:
-
-https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/proto/qgb/types.proto#L31-L47
-
-Or, valsets:
-
-https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/proto/qgb/types.proto#L17-L29
+The QGB state machine handles the creation of [`AttestationRequestI`](https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/x/qgb/types/attestation.go#L15-L23) which can be either a [`DataCommitment`](https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/proto/qgb/types.proto#L31-L47) or a [`Valset`](https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/proto/qgb/types.proto#L17-L29).
 
 During their creation, the state machine might panic due to an unexpected behavior or event. These panics will be discussed below.
 
@@ -25,9 +16,13 @@ During their creation, the state machine might panic due to an unexpected behavi
 
 A new valset is created in the following situations:
 
-https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/x/qgb/abci.go#L74
+```golang
+if (latestValset == nil) || (lastUnbondingHeight == uint64(ctx.BlockHeight())) || significantPowerDiff {
+    ...
+}
+```
 
-- No valset exist in store, so a new valset will be created. This happens mostly after genesis, or after a hard fork.
+- No valset exists in store, so a new valset will be created. This happens mostly after genesis, or after a hard fork.
 - The current block height is the last unbonding height, i.e. when a validator is leaving the validator set. A new valset will need to be created to accommodate that change.
 - A significant power difference happened since the last valset. This could happen if a validator has way more staking power or the opposite. The significant power difference threshold is defined by the constant `SignificantPowerDifferenceThreshold`, and is set to 5% currently.
 
@@ -35,7 +30,11 @@ https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb6
 
 A new data commitment is created in the following situation:
 
-https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/x/qgb/abci.go#L23
+```golang
+if ctx.BlockHeight() != 0 && ctx.BlockHeight()%int64(k.GetDataCommitmentWindowParam(ctx)) == 0 {
+	...
+}
+```
 
 I.e. the current block height is not 0, and we're at a data commitment window height.
 
@@ -52,7 +51,8 @@ During EndBlock, if the block height corresponds to a `DataCommitmentWindow`, it
 - An unexpected behavior happened while getting the current data commitment:
 
 ```golang
-dataCommitment, err := k.GetCurrentDataCommitment(ctx)  
+dataCommitment
+}, err := k.GetCurrentDataCommitment(ctx)  
 if err != nil {  
    panic(sdkerrors.Wrap(err, "coudln't get current data commitment"))  
 }
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -1,6 +1,7 @@
 # Quantum gravity bridge state machine
 
-This module contains the [Quantum gravity bridge](https://blog.celestia.org/celestiums/) state machine implementation.
+## Abstract
+This module contains the [Quantum Gravity Bridge](https://blog.celestia.org/celestiums/) (QGB) state machine implementation.
 
 ## State machine
 
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -102,7 +102,7 @@ if err != nil {
 
 #### Attestations panics
 
-When storing a new attestation, which is either a data commitment or a valset, the state machine can panic for in the following cases:
+When storing a new attestation, which is either a data commitment or a valset, the state machine can panic in the following cases:
 
 - The attestation request created from the data commitment is a duplicate of an existing attestation:
 
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -90,7 +90,7 @@ if err != nil {
 
 When storing a new attestation, the state machine can panic if it finds invalid state. This latter can happen in the following cases:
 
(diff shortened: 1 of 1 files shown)


x/qgb/README.md
```diff
@@ -0,0 +1,139 @@
+# Quantum gravity bridge state machine
+
+This module contains the [Quantum gravity bridge](https://blog.celestia.org/celestiums/) state machine implementation.
+
+## State machine
+
+The QGB state machine handles the creation of the attestations requests:
+
+https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/x/qgb/types/attestation.go#L15-L23
```
SweeXordious, line 9: [Question for reviewers]
Currently, markdown linter complains about bare URLs:
```
MD034/no-bare-urls Bare URL used [Context: "https://github.com/celestiaorg..."]
```
However, I guess in the case where we're providing snippets links, it should be okay. Since, using links to code makes it easier to scroll, syntax highlighting and navigation.

Do you think we should remove the rule? or I switch to copy-paste?

x/qgb/README.md
```diff
@@ -0,0 +1,140 @@
+# Quantum gravity bridge state machine
+
+This module contains the [Quantum gravity bridge](https://blog.celestia.org/celestiums/) state machine implementation.
```
rootulp, line 3:
```suggestion
## Abstract
This module contains the [Quantum Gravity Bridge](https://blog.celestia.org/celestiums/) (QGB) state machine implementation.
```

x/qgb/README.md
```diff
@@ -0,0 +1,140 @@
+# Quantum gravity bridge state machine
```
rootulp, line 1: [no change needed][discussion] since this README contains docs for a Cosmos SDK module, thoughts on conforming to the structure of https://github.com/cosmos/cosmos-sdk/tree/main/x/auth. For this line:

```suggestion
# `x/qgb`
```

if we do adopt the Cosmos SDK module docs structure, we may also adopt the headings from https://github.com/cosmos/cosmos-sdk/tree/main/x/auth for other sections

x/qgb/README.md
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+## Useful links
+
+The smart contract implementation is in [quantum-gravity-bridge](https://github.com/celestiaorg/quantum-gravity-bridge/).
+
+The orchestrator and relayer implementations are in [orchestrator-relayer](https://github.com/celestiaorg/orchestrator-relayer/).
+
+QGB v1 implementation, including the orchestrator and relayer, is in the [qgb-integration](https://github.com/celestiaorg/celestia-app/tree/qgb-integration) branch.
```
rootulp, line 140: 👍

x/qgb/README.md
rootulp: I think we should keep the linter rule and provide a link name like so:

```markdown
The QGB state machine handles the creation of [`AttestationRequestI`](https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/x/qgb/types/attestation.go#L15-L23) which can be a [`DataCommitment`](https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/proto/qgb/types.proto#L31-L47) or [`Valset`](https://github.com/celestiaorg/celestia-app/blob/801a0d412631989ce97748badbd7bb676982db16/proto/qgb/types.proto#L17-L29).
```

x/qgb/README.md
```diff
@@ -0,0 +1,140 @@
...
+```golang
+intLatestMembers, err := types.BridgeValidators(latestValset.Members).ToInternal()  
+if err != nil {  
+   panic(sdkerrors.Wrap(err, "invalid latest valset members"))  
+}
+```
+
+#### Attestations panics
+
+When storing a new attestation, which is either a data commitment or a valset, the state machine can panic for in the following cases:
```
rootulp, line 104:
```suggestion
When storing a new attestation, which is either a data commitment or a valset, the state machine can panic in the following cases:
```

rootulp (commented):
No blocking feedback but I think the links in the doc could use link text to make the markdownlint rule pass

x/qgb/README.md
```diff
@@ -0,0 +1,141 @@
...
+```golang
+dataCommitment, err := k.GetCurrentDataCommitment(ctx)  
+if err != nil {  
+   panic(sdkerrors.Wrap(err, "coudln't get current data commitment"))  
+}
+```
+
+#### Valset panics
+
+Similar to data commitments, when checking if the state machine needs to generate a new valset, it might panic in the following cases:
```
evan-forbes, line 63: Might want to clarify a bit that it only panics when invalid state is created

x/qgb/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+The data commitment window is defined as a governance parameter that can be changed. Currently, it is set to 400.
+
+### Panics
+
+During EndBlock step, the state machine generates new attestations if needed. During this generation, the state machine could panic.
+
+#### Data commitment panics
+
+During EndBlock, if the block height corresponds to a `DataCommitmentWindow`, it will generate a new data commitment, during which, the state machine can panic in the following case:
```
evan-forbes, line 50: [nit]
same as comment below

x/qgb/README.md
```diff
@@ -0,0 +1,141 @@
...
+```golang
+intLatestMembers, err := types.BridgeValidators(latestValset.Members).ToInternal()  
+if err != nil {  
+   panic(sdkerrors.Wrap(err, "invalid latest valset members"))  
+}
+```
+
+#### Attestations panics
+
+When storing a new attestation, which is either a data commitment or a valset, the state machine can panic in the following cases:
```
evan-forbes, line 105: [nit]
same as below

x/qgb/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+- The universal nonce was not incremented correctly by 1:
+
+```golang
+if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != nonce {  
+   panic("not incrementing latest attestation nonce correctly")  
+}
+```
+
+## Useful links
```
evan-forbes, line 135: could make link the ADRs

evan-forbes (commented):
I like keeping what used to be in the spec folder just in the main readme a ton!!

x/qgb/README.md
SweeXordious: [13e6376](https://github.com/celestiaorg/celestia-app/pull/1052/commits/13e6376199508e32ec8c5524c7d96869b69a028e)

x/qgb/README.md
SweeXordious: [13e6376](https://github.com/celestiaorg/celestia-app/pull/1052/commits/13e6376199508e32ec8c5524c7d96869b69a028e)

x/qgb/README.md
SweeXordious: [13e6376](https://github.com/celestiaorg/celestia-app/pull/1052/commits/13e6376199508e32ec8c5524c7d96869b69a028e)

x/qgb/README.md
SweeXordious: [7f28cb9](https://github.com/celestiaorg/celestia-app/pull/1052/commits/7f28cb9584fe4cd7eba0eaf2dc9f4b75757c87f0)

rootulp (dismissed):
:shipit:

x/qgb/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+The data commitment window is defined as a governance parameter that can be changed. Currently, it is set to 400.
+
+### Panics
+
+During EndBlock step, the state machine generates new attestations if needed. During this generation, the state machine could panic.
+
+#### Data commitment panics
+
+During EndBlock, if the block height corresponds to a `DataCommitmentWindow`, it will generate a new data commitment, during which, the state machine can panic, if it finds invalid state, in the following case:
```
evan-forbes, line 49:
```suggestion
During EndBlock, if the block height corresponds to a `DataCommitmentWindow`, it will generate a new data commitment. Note that the state machine can panic if invalid state is introduced.
```

I would also then either remove
`- An unexpected behavior happened while getting the current data commitment:`
or clarify with more precise details of what could cause that. Like if the last nonce's key is missing or something not sure. Also, if we only have a single entry, I'm not sure we need a list

x/qgb/README.md
```diff
@@ -0,0 +1,143 @@
...
+dataCommitment
+}, err := k.GetCurrentDataCommitment(ctx)  
+if err != nil {  
+   panic(sdkerrors.Wrap(err, "coudln't get current data commitment"))  
+}
+```
+
+#### Valset panics
+
+Similar to data commitments, when checking if the state machine needs to generate a new valset, it might panic, if it finds invalid state, in the following cases:
```
evan-forbes, line 63:
```suggestion
Note that, similar to data commitments, the state machine can panic if the required valset is not stored in the state db.
```

x/qgb/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+- When checking that a previous valset has been emitted, but it is unable to get it:
```
evan-forbes, line 65:
```suggestion
```

evan-forbes: might also need to do some formatting if the above change is adopted

x/qgb/README.md
```diff
@@ -0,0 +1,143 @@
...
+if err != nil {  
+   panic(sdkerrors.Wrap(err, "invalid latest valset members"))  
+}
+```
+
+#### Attestations panics
+
+When storing a new attestation, which is either a data commitment or a valset, the state machine can panic, if it finds invalid state, in the following cases:
+
+- The attestation request created from the data commitment is a duplicate of an existing attestation:
```
evan-forbes, line 107:
```suggestion
When storing a new attestation, which is either a data commitment or a valset, the state machine can panic if the attestation request is a duplicate of an existing attestation.
```

evan-forbes (changes requested):
Thanks for incorporating the suggested changes earlier, I still have some nit picks tho that I think we should handle before merging. We don't have to use only the suggestions listed here, but I do think we can clean up a few minor things.

x/qgb/README.md
SweeXordious: I'll remove the data commitment panic section because currently, the `GetDataCommitment` will never return an error => no panics.

x/qgb/README.md
SweeXordious: doesn't it make sense to keep the list? we have three cases where the state machine can panic...

evan-forbes (approved):
my mistake! I didn't see the other list items between the code until I rendered the file :see_no_evil:

PR #1060, opened 2022-11-24
chore: rename `message` => `blob`

rootulp:
https://github.com/celestiaorg/celestia-app/issues/995

Renames the following on a case by case basis:
- `Message` => `Blob`
- `message`, `msg`, `data` => `blob`

Does not address:
- ADRs
- specs

It's likely that I didn't catch every instance due to human error


evan-forbes (commented):
omg thanks for doing this! such a pain (we're never renaming it again). 

let's fix the any conflicts and merge ASAP to avoid other conflicts here. Apologies that we didn't merge sooner

rootulp:
> Apologies that we didn't merge sooner

No worries, the merge conflicts were my fault and not difficult to resolve. This is ready for re-review.

evan-forbes (approved):
merge merge merge before we get more conflicts

PR #1070, opened 2022-11-25
feat: enable blob simulation

### LINKED ISSUE #24: Implement the simulation package for lazyledgerapp
Modules in the sdk all have a package called simulation that interfaces with the `simapp` framework for fuzzing. Implement the simulation package for lazyledgerapp so that proper simulation fuzzing can be done as described [here](https://docs.cosmos.network/master/building-modules/simulator.html)

rahulghangas:
- [x] Closes #24

### DIFF
--- a/app/test/simulation_test.go
+++ b/app/test/simulation_test.go
@@ -1,12 +1,12 @@
 package app_test
 
 import (
-	"github.com/celestiaorg/celestia-app/app/encoding"
 	"os"
 	"testing"
 	"time"
 
 	"github.com/celestiaorg/celestia-app/app"
+	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/simapp"
--- a/app/test/simulation_test.go
+++ b/app/test/simulation_test.go
@@ -1,6 +1,9 @@
 package app_test
 
 import (
+	"os"
+	"testing"
+
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/cosmos/cosmos-sdk/baseapp"
@@ -12,8 +15,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/simulation"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
-	"os"
-	"testing"
 )
 
 func init() {
--- a/app/test/simulation_test.go
+++ b/app/test/simulation_test.go
@@ -1,10 +1,6 @@
 package app_test
 
 import (
-	"os"
-	"testing"
-	"time"
-
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/cosmos/cosmos-sdk/baseapp"
@@ -16,8 +12,8 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/simulation"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
-	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
-	tmtypes "github.com/tendermint/tendermint/types"
+	"os"
+	"testing"
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


x/blob/module_simulation.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/cosmos/cosmos-sdk/x/simulation"
+)
+
+// avoid unused import issue
+var (
+	_ = sample.AccAddress
+	_ = bloblsimulation.FindAccount
+	_ = simappparams.StakePerAccount
+	_ = simulation.MsgEntryKind
+	_ = baseapp.Paramspace
```
SweeXordious, line 24: [Question]
Question please, if these are not used, why do we need to import them?

x/blob/module_simulation.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	simState.GenState[types.ModuleName] = simState.Cdc.MustMarshalJSON(&blobGenesis)
+}
+
+// ProposalContents doesn't return any content functions for governance proposals
+func (AppModule) ProposalContents(_ module.SimulationState) []simtypes.WeightedProposalContent {
+	return nil
+}
+
+// RandomizedParams creates randomized  param changes for the simulator
```
SweeXordious, line 44:
```suggestion
// RandomizedParams creates randomized param changes for the simulator
```

x/blob/module_simulation.go
```diff
@@ -0,0 +1,67 @@
...
+		simulation.NewSimParamChange(types.ModuleName, string(types.KeyGasPerBlobByte), func(r *rand.Rand) string {
+			return string(codec.Amino.MustMarshalJSON(blobParams.GasPerBlobByte))
+		}),
+	}
+}
+
+// RegisterStoreDecoder registers a decoder
+func (am AppModule) RegisterStoreDecoder(_ sdk.StoreDecoderRegistry) {}
+
+// WeightedOperations returns the all the gov module operations with their respective weights.
```
SweeXordious, line 63:
```suggestion
// WeightedOperations returns all the gov module operations with their respective weights.
```

app/test/simulation_test.go
```diff
@@ -0,0 +1,108 @@
...
+//		MaxAgeNumBlocks: 302400,
+//		MaxAgeDuration:  504 * time.Hour, // 3 weeks is the max duration
+//		MaxBytes:        10000,
+//	},
+//	Validator: &tmproto.ValidatorParams{
+//		PubKeyTypes: []string{
+//			tmtypes.ABCIPubKeyTypeEd25519,
+//		},
+//	},
+//}
```
evan-forbes, line 51: [blocking question]
can we delete these? if we need these, perhaps we can use the testnode.DefaultParams func

app/test/simulation_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+// BenchmarkSimulation run the chain simulation
+// Running using starport command:
+// `starport chain simulate -v --numBlocks 200 --blockSize 50`
+// Running as go benchmark test:
+// `go test -benchmem -run=^$ -bench ^BenchmarkSimulation ./app -NumBlocks=200 -BlockSize 50 -Commit=true -Verbose=true -Enabled=true`
```
evan-forbes, line 57: [non blocking question]
Do you know if simapp is using ABCI++? if not, the results of this could be misleading and not as useful as we might hope.

How can we see these results? do we just run them locally? What is blockSize 50?

x/blob/module_simulation.go
```diff
@@ -0,0 +1,67 @@
...
+			return string(codec.Amino.MustMarshalJSON(blobParams.MinSquareSize))
+		}),
+		simulation.NewSimParamChange(types.ModuleName, string(types.KeyMaxSquareSize), func(r *rand.Rand) string {
+			return string(codec.Amino.MustMarshalJSON(blobParams.MaxSquareSize))
+		}),
+		simulation.NewSimParamChange(types.ModuleName, string(types.KeyGasPerBlobByte), func(r *rand.Rand) string {
+			return string(codec.Amino.MustMarshalJSON(blobParams.GasPerBlobByte))
+		}),
+	}
+}
```
evan-forbes, line 58: [blocking question]
I'm guessing that this uses a random parameter when booting this up. How does know to pick a power of two for the square size? if its not picking a power of two, shouldn't this be failing?

evan-forbes (commented):
mainly just questions, and only two blocking iirc. this is cool!

evan-forbes:
closing this for now since due to it being stale, but if anyone wants to complete this, we can definitely reopen it!

Issue #1071, opened 2022-11-25
Implement shares inclusion integration test


SweeXordious:
### Summary

Add integration test to shares inclusion proofs

### Problem Definition

After adding support to the shares inclusion proofs, we should implement an integration test for that on Celestia-app.

### Proposal

.

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1042, opened 2022-11-19
feat: add shares inclusion proofs 

SweeXordious:
Adds support for shares inclusion proofs. 
Depends on https://github.com/celestiaorg/celestia-core/pull/878

PR #1072, opened 2022-11-25
Implement shares inclusion proofs e2e

SweeXordious:
## Overview
Closes: https://github.com/celestiaorg/celestia-app/issues/1071

## Checklist

### DIFF
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/prove"
-	"github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/celestiaorg/celestia-app/testutil/namespace"
 	"github.com/celestiaorg/celestia-app/testutil/network"
 	"github.com/celestiaorg/celestia-app/x/blob"
@@ -211,39 +210,28 @@ func (s *IntegrationTestSuite) TestSharesInclusionProof() {
 		beginTxShare, endTxShare, err := prove.TxSharePosition(blockRes.Block.Txs, uint64(txResp.Index))
 		require.NoError(err)
 
-		// split the block to shares
-		rawShares, err := shares.Split(blockRes.Block.Data, true)
-		require.NoError(err)
-
-		// verify the transaction shares proof
-		txProof, err := prove.SharesInclusion(
-			rawShares,
-			blockRes.Block.SquareSize,
-			appconsts.TxNamespaceID,
+		txProof, err := node.ProveShares(
+			context.Background(),
+			uint64(txResp.Height),
 			beginTxShare,
 			endTxShare,
 		)
 		require.NoError(err)
-		require.NoError(txProof.Validate())
+		require.NoError(txProof.Validate(blockRes.Block.DataHash))
 
 		// get the message shares
 		beginMsgShare, endMsgShare, err := prove.MsgSharesPosition(blockRes.Block.Txs[txResp.Index])
 		require.NoError(err)
 
-		// parse the message namespace
-		nID, err := prove.ParseNamespaceID(rawShares, int64(beginMsgShare), int64(endMsgShare))
-		require.NoError(err)
-
 		// verify the message shares proof
-		msgProof, err := prove.SharesInclusion(
-			rawShares,
-			blockRes.Block.SquareSize,
-			nID,
+		msgProof, err := node.ProveShares(
+			context.Background(),
+			uint64(txResp.Height),
 			beginMsgShare,
 			endMsgShare,
 		)
 		require.NoError(err)
-		require.NoError(msgProof.Validate())
+		require.NoError(msgProof.Validate(blockRes.Block.DataHash))
 	}
 }
 
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -8,14 +8,15 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	cosmosnet "github.com/cosmos/cosmos-sdk/testutil/network"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/stretchr/testify/suite"
-
-	cosmosnet "github.com/cosmos/cosmos-sdk/testutil/network"
 
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/pkg/prove"
+	"github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/celestiaorg/celestia-app/testutil/namespace"
 	"github.com/celestiaorg/celestia-app/testutil/network"
 	"github.com/celestiaorg/celestia-app/x/blob"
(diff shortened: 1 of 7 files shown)


SweeXordious (commented):
Depends on https://github.com/celestiaorg/celestia-app/pull/1042. 
After merging it, I will fix the git history for this PR so that we merge it afterwards.

app/test/integration_test.go
```diff
@@ -172,6 +173,80 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
...
+	for i := 0; i < 50; i++ {
+		require.NoError(s.network.WaitForNextBlock())
+	}
+
+	for _, hash := range hashes {
+		txResp, err := queryTx(val.ClientCtx, hash, true)
+		require.NoError(err)
+		require.Equal(abci.CodeTypeOK, txResp.TxResult.Code)
+
+		// verify that the transaction inclusion txProof is valid
```
SweeXordious, line 203:
```suggestion
		// verify that the transaction inclusion proof is valid
```

app/test/integration_test.go
```diff
@@ -172,6 +173,80 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
...
+
+	for i, tx := range txs {
+		res, err := val.ClientCtx.BroadcastTxSync(tx)
+		require.NoError(err)
+		require.Equal(abci.CodeTypeOK, res.Code)
+		hashes[i] = res.TxHash
+	}
+
+	// wait a few blocks to clear the txs
+	for i := 0; i < 50; i++ {
```
evan-forbes, line 194: [blocking]
50 is way too large and will significantly increase the time of the test, I think we should lower this

app/test/integration_test.go
```diff
@@ -172,6 +173,80 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
...
+		// verify the transaction shares proof
+		txProof, err := prove.SharesInclusion(
+			rawShares,
+			blockRes.Block.SquareSize,
+			appconsts.TxNamespaceID,
+			beginTxShare,
+			endTxShare,
+		)
+		require.NoError(err)
+		require.NoError(txProof.Validate())
```
evan-forbes, line 227: [blocking]
can we actually use the rpc here?

evan-forbes (commented):
can we point this at the other PR so its easier to see the diff?

also, is it just my browser, or is the syntax highlighting not working on the integration test files?

app/test/integration_test.go
evan-forbes: worst case would be about 10-20 iirc, but we should just use fewer or smaller PFBs imo

SweeXordious:
> can we point this at the other PR so its easier to see the diff?

I wanted to do so, but since both PRs are on my fork, then we will be reviewing a PR on my fork if I change the base

app/test/integration_test.go
SweeXordious: I did 20 and I ran it multiple times, and sometimes some  transactions don't get included... will reduce the number of PFBs then. Thanks.

app/test/integration_test.go
SweeXordious: This is not an `rpc` call, this is executing the `Validate` function that is defined for the `TxProof` struct.

SweeXordious: Now, I have a question.  Concerning the proof struct:

```
message SharesProof {
  // data raw shares that are proven.
  repeated bytes data = 1;
  // shares_proof NMT proofs for the raw shares. Could be more than one
  // if the shares span more than one row.
  repeated NMTProof shares_proof = 2;
  // namespace_id namespaceID of the shares.
  bytes namespace_id = 3;
  // rows_proof binary merkle proofs, to the data root, of the rows
  // containing the raw shares.
  RowsProof rows_proof = 4;
```

Currently, it is self-sufficient. It contains everything it needs to verify.
What I mean by that, is that you don't need to specify a data root as parameter, to check the proof against. The data root is already included there.

Also, it contains the raw shares that need to be proved.

Thus, calling `Validate()`, will check if it is a valid proof.

Which is similar to how the `TxProof` is implemented:

```
// TxProof represents a Merkle proof of the presence of a transaction in the Merkle tree.
type TxProof struct {
	RowRoots []tmbytes.HexBytes  `json:"root_hash"`
	Data     [][]byte            `json:"data"`
	Proofs   []*tmproto.NMTProof `json:"proof"`
}
```

Contrarily to, for example, the tendermint implementation of a merkle proof:

```

// Verify that the Proof proves the root hash.
// Check sp.Index/sp.Total manually if needed
func (sp *Proof) Verify(rootHash []byte, leaf []byte) error {
	leafHash := leafHash(leaf)
	if sp.Total < 0 {
		return errors.New("proof total must be positive")
	}
	if sp.Index < 0 {
		return errors.New("proof index cannot be negative")
	}
	if !bytes.Equal(sp.LeafHash, leafHash) {
		return fmt.Errorf("invalid leaf hash: wanted %X got %X", leafHash, sp.LeafHash)
	}
	computedHash := sp.ComputeRootHash()
	if !bytes.Equal(computedHash, rootHash) {
		return fmt.Errorf("invalid root hash: wanted %X got %X", rootHash, computedHash)
	}
	return nil
}
```
Where, you need to specify the leaf and the root hash to verify it.

Thus, do the proofs that we generate need to be self-verifiable, or they should need extra information from outside?

I personally think that the former makes more sense, since it is a multiproof. But, I am not sure.

app/test/integration_test.go
SweeXordious: [fe0b7b2](https://github.com/celestiaorg/celestia-app/pull/1072/commits/fe0b7b2d9889a6dbe6079c07ff403c724615f80e)

SweeXordious:
@evan-forbes Can I start merging the proofs PRs?

app/test/integration_test.go
evan-forbes: > This is not an rpc call, this is executing the Validate function that is defined for the TxProof struct.

sorry, I meant that I think the important thing to test here is that the abci query is working via RPC. The rest seems like it could be a unit test.

evan-forbes: and I forget why we don't pass the root when validating. We at least need the user to verify that the hashes match or its not safe

SweeXordious: > sorry, I meant that I think the important thing to test here is that the abci query is working via RPC. The rest seems like it could be a unit test.

This is testing that the proofs are generated correctly for real blocks.

What do you mean by the ABCI query? if we're testing the RPC, then this test should be on Celestia-core side, because that's where we're receiving the RPC call. Otherwise, on Celestia-app side, we're just taking that request, parsing it and then returning the proof.

SweeXordious: [fa1cbcf](https://github.com/celestiaorg/celestia-app/pull/1072/commits/fa1cbcf0ea8363720293ebd3830d26db7823f5cf) @evan-forbes Please take a look when you have time

pkg/prove/proof.go
```diff
@@ -107,6 +117,49 @@ func txSharePosition(txs types.Txs, txIndex uint64) (startSharePos, endSharePos
 	return startSharePos, endSharePos, nil
 }
 
+// MsgSharesPosition returns the start and end positions for the shares that
```
SweeXordious, line 120:
```suggestion
// MsgSharesPosition returns the start and end positions for the shares
```

go.mod
```diff
@@ -204,5 +204,5 @@ replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.4.0-sdk-v0.46.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint v0.34.20 => github.com/celestiaorg/celestia-core v1.8.0-tm-v0.34.20
+	github.com/tendermint/tendermint v0.34.20 => github.com/sweexordious/celestia-core v0.34.14-celestia.0.20221201135919-1692b16b341e
```
evan-forbes, line 207: just triple checking that this is the only blocker here?

SweeXordious: Yes, once the core PR is merged, we can start merging everything

SweeXordious:
Closed for https://github.com/celestiaorg/celestia-app/pull/1233

PR #1077, opened 2022-11-28
refactor: remove duplciate `validateHash`

rootulp:
## Description
It looks like the helper function `validateHash` was intentionally duplicated in the `da` package to avoid import issues when this package lived in celestia-core. Now that `da` lives in celestia-app, it seems possible to import celestia-core [`ValidateHash`](https://github.com/celestiaorg/celestia-core/blob/2ec23f804691afc196d0104616e6c880d4c1ca41/types/validation.go#L30-L41) directly rather than duplicate it.

## Motivation
Observed while reading `da` package to answer an unrelated question.


evan-forbes (approved):
nice catch! this must have been a remnant of when this code was in celestia-core and we needed to import it into the types package.

omfg, do we have a flakey markdown linter??!

rootulp:
> omfg, do we have a flakey markdown linter??!

It seems like it. Both of the links that error [here](https://github.com/celestiaorg/celestia-app/actions/runs/3566186239/jobs/5994986049#step:4:319) work for me.

Issue #1068, opened 2022-11-24
Remove evidence from `pkg/shares`


rootulp:
## Context

Part of https://github.com/celestiaorg/celestia-app/issues/950

## Proposal

Remove evidence from `pkg/shares`. Candidates for removal https://github.com/celestiaorg/celestia-app/blob/24626ff8d00854d7e0027abcfdf50de1a2aee6a3/pkg/appconsts/appconsts.go#L83-L84

https://github.com/celestiaorg/celestia-app/blob/d32f68742cdfcfd5787f1c892da8401f3d0ebb72/pkg/shares/share_splitting.go#L117-L126

and more (see [search](https://github.com/celestiaorg/celestia-app/search?p=1&q=Evidence)).

rootulp:
On second thought, I don't think we should remove https://github.com/celestiaorg/celestia-app/blob/24626ff8d00854d7e0027abcfdf50de1a2aee6a3/pkg/appconsts/appconsts.go#L83-L84 at this time because that namespace remains reserved due to https://github.com/celestiaorg/celestia-app/blob/24626ff8d00854d7e0027abcfdf50de1a2aee6a3/pkg/appconsts/appconsts.go#L89

It can be removed from the codebase / specs at a later date if desired.

PR #1079, opened 2022-11-29
feat!: remove evidence from data square

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1068


evan-forbes (approved):
LGTM, I'm merging as this as it's a pretty straight forward change, and is blocking me since we merged the evidence change in the last update to core. If we need to make a followup PR, we can do so soon.

We still need decide what to do with the EvidenceNamespaceID and ofc update the specs

rootulp:
> We still need decide what to do with the EvidenceNamespaceID and ofc update the specs

Spun out https://github.com/celestiaorg/celestia-app/issues/1081 to track

PR #1080, opened 2022-11-30
refactor: register msg service

rootulp:
Closes https://github.com/informalsystems/audit-celestia/issues/1

## Questions
1. Why don't we need an rpc route for WirePayForBlob in [tx.proto](https://github.com/rootulp/celestia-app/blob/b2c271a00f0b9aa03d3df908835191289630fa8a/proto/blob/tx.proto#L12-L14)?
2. I feel like we need additional testing to have confidence in this change. Where can I find reference examples for tests on this type of Cosmos SDK msg service set-up (likely [cosmos-sdk](https://github.com/cosmos/cosmos-sdk) / [osmosis](https://github.com/osmosis-labs/osmosis))


x/blob/types/msgs.go
```diff
@@ -2,23 +2,198 @@ package types
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"errors"
 	"fmt"
-	"math"
+	math "math"
```
rootulp, line 8: why gofmt?

rootulp: resolved via https://github.com/celestiaorg/celestia-app/pull/1080/commits/b2c271a00f0b9aa03d3df908835191289630fa8a

evan-forbes (commented):
> Why don't we need an rpc route for WirePayForBlob in [tx.proto](https://github.com/rootulp/celestia-app/blob/b2c271a00f0b9aa03d3df908835191289630fa8a/proto/blob/tx.proto#L12-L14)?

the state machine shouldn't actually know how to handle the wire messages as they should never be included in the block, and if they are they need to be ignored.

this is super seceded by removal of the wire msg, so not sure this is worth continuing to pursue

> I feel like we need additional testing to have confidence in this change. Where can I find reference examples for tests on this type of Cosmos SDK msg service set-up

setting up a state mahcine to execute transactions could be done, but to test the entire routing and execution essentially an integration test is needed.

rootulp:
> the state machine shouldn't actually know how to handle the wire messages as they should never be included in the block

makes sense

> this is super seceded by removal of the wire msg, so not sure this is worth continuing to pursue

To clarify: I won't pursue adding a route for wire msg considering the in-flight removal of wire msg. But I should continue to pursue registering the msg service (i.e. the remainder of this PR) - right?

x/blob/keeper/keeper.go
```diff
@@ -53,17 +51,3 @@ func NewKeeper(
...
-
-	gasToConsume := uint64(shares.BlobSharesUsed(int(msg.BlobSize)) * GasPerBlobShare)
-	ctx.GasMeter().ConsumeGas(gasToConsume, payForBlobGasDescriptor)
-
-	ctx.EventManager().EmitEvent(
-		types.NewPayForBlobEvent(sdk.AccAddress(msg.Signer).String(), msg.GetBlobSize()),
-	)
-
-	return &types.MsgPayForBlobResponse{}, nil
-}
```
evan-forbes, line 69: why move this?

was this because of this comment? 

> move msgServer methods to x/payment/keeper/msg_server.go, i.e., PayForData method

x/blob/keeper/msg_server.go
```diff
@@ -15,3 +19,17 @@ type msgServer struct {
...
+
+	gasToConsume := uint64(shares.BlobSharesUsed(int(msg.BlobSize)) * GasPerBlobShare)
+	ctx.GasMeter().ConsumeGas(gasToConsume, payForBlobGasDescriptor)
+
+	ctx.EventManager().EmitEvent(
+		types.NewPayForBlobEvent(sdk.AccAddress(msg.Signer).String(), msg.GetBlobSize()),
+	)
+
+	return &types.MsgPayForBlobResponse{}, nil
+}
```
evan-forbes, line 35: [blocking discussion]
imo if we want to move this here, then we should change the method receiver from `Keeper` to `msgServer` and add more boiler plate to wrap the keeper method.

evan-forbes (commented):
> But I should continue to pursue registering the msg service (i.e. the remainder of this PR) - right?

ahh whoops yeah, that we can still do. I do have some questions and personal disagreements with following the frequently changing "best practices" for naming thing in the sdk.

we only have a single sdk.Msg, so I'll defer to you if you want to keep it there, but no one else does this atm (not even the sdk), and I think its bad practice to keep giant files filled with all of our messages. tbh I just flat out disagree with the sdk that this should be done. This only makes sense if your message types are super simple and effectively boiler plate, which ours are definitely not.

maybe if we split out the create commitment function out of the pfb file, but I even disagree with that to a decent extent.

x/blob/keeper/keeper.go
evan-forbes: see below comment

rootulp:
> with following the frequently changing "best practices" for naming thing in the sdk.

Yea fair enough. I haven't used Cosmos SDK long enough to know when it's appropriate to violate "best practices" so I opted to adopt what was suggested. If we aren't aligned on adopting the contents of https://github.com/informalsystems/audit-celestia/issues/1 proposal to move the discussion there? Apologies for working on this issue without gathering alignment on what's desirable.

rootulp:
Pausing / closing until we have alignment on what we want to adopt from https://github.com/informalsystems/audit-celestia/issues/1

x/blob/keeper/msg_server.go
rootulp: Thanks for flagging. I'll refer to https://github.com/cosmos/cosmos-sdk/blob/main/x/gov/keeper/msg_server.go and can do something similar.

PR #1084, opened 2022-11-30
refactor: optimize blob shares used

rootulp:
Closes https://github.com/informalsystems/audit-celestia/issues/11

```bash
$ go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/celestiaorg/celestia-app/pkg/shares
BenchmarkBlobSharesUsed-10             	480383745	         2.488 ns/op
BenchmarkBlobSharesUsedOptimized-10    	482551569	         2.493 ns/op
PASS
ok  	github.com/celestiaorg/celestia-app/pkg/shares	3.648s
```

### DIFF
--- a/pkg/shares/utils.go
+++ b/pkg/shares/utils.go
@@ -31,6 +31,13 @@ func BlobSharesUsed(blobSize int) int {
 	return shareCount
 }
 
+func BlobSharesUsedOptimized(blobSize int) int {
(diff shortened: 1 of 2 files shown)


rootulp:
Closing because the optimization is slower based on this Go benchmark and in my opinion seems less readable but that argument is subjective.

Issue #913, opened 2022-10-25
TestFuzzPrepareProcessProposal: makeslice: len out of range


rootulp:
### Summary of Bug

We're observing instances where a fuzz test has uncovered a bug. See CI run: https://github.com/celestiaorg/celestia-app/actions/runs/3318414871/jobs/5482354309#step:6:149

### Version

https://github.com/celestiaorg/celestia-app/commit/26eeca4d35588991ccec7dcaf2defd4d63854e82 and `main`

### Steps to Reproduce

@evan-forbes ran fuzz test locally long enough to reproduce the error: https://gist.github.com/rootulp/86c4a7a80e95f64bafdd4f968c0f131a

evan-forbes:
Hopefully we can fix this when we refactor/remove estimation to use square size independent commitments

cc @rootulp

PR #1088, opened 2022-12-01
fix: throw error if too many shares for square size

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/913

This is an attempt at resolving the above issue. If the issue repros after this merges, I'll investigate further (and re-open the issue).

### DIFF
--- a/pkg/shares/share_splitting.go
+++ b/pkg/shares/share_splitting.go
@@ -58,6 +58,9 @@ func Split(data coretypes.Data, useShareIndexes bool) ([]Share, error) {
 		return nil, err
 	}
 	currentShareCount += len(blobShares)
+	if currentShareCount > wantShareCount {
+		return nil, fmt.Errorf("too many shares: %d > %d", currentShareCount, wantShareCount)
+	}
 	tailShares := TailPaddingShares(wantShareCount - currentShareCount)
 
 	// todo: optimize using a predefined slice
--- a/pkg/shares/share_splitting_test.go
+++ b/pkg/shares/share_splitting_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/stretchr/testify/require"
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/shares/share_splitting.go
```diff
@@ -58,6 +58,9 @@ func Split(data coretypes.Data, useShareIndexes bool) ([]Share, error) {
 		return nil, err
 	}
 	currentShareCount += len(blobShares)
+	if currentShareCount > wantShareCount {
+		return nil, fmt.Errorf("too many shares: %d > %d", currentShareCount, wantShareCount)
+	}
```
evan-forbes, line 63: what happens if this is reached during PrepareProposal?

evan-forbes (commented):
I really like adding this error, but I'm not sure this counts as fixing #913 as it doesn't actually fix the source of the bug, but does stop the block producer from creating a block. Tendermint doesn't currently have a way of handling this, so our only option would be to panic (what we currently do implicitly) or produce an empty block, both of which are not very good options.

fwiw, I think the error was in the estimation logic, and _should_ be handled when we simplify the estimation logic after removing the wirePFB (kinda separate issues, but both done in #1089).

pkg/shares/share_splitting.go
rootulp: > what happens if this is reached during PrepareProposal?

that would indicate an error in https://github.com/celestiaorg/celestia-app/blob/1af66978dac3fa3133dbd6f9be4c60d5642d846e/app/prepare_proposal.go#L25

we could try to reproduce such an error by fuzz testing `estimateSquareSize`

evan-forbes: yeah, but how do we handle that error?

rootulp:
> Tendermint doesn't currently have a way of handling this

Sorry I don't understand. What feature do we want from Tendermint? 

> so our only option would be to panic (what we currently do implicitly)

Do we want this scenario to panic or throw an error? I assume we should throw an error in the `shares` package but may consider catching that error and panicing in prepare proposal / process proposal

evan-forbes:
> Do we want this scenario to panic or throw an error? I assume we should throw an error in the shares package but may consider catching that error and panicing in prepare proposal / process proposal

yeah, bubbling and panicking that error in an explicit way is better than implicit, and I think we should do that.

> Sorry I don't understand. What feature do we want from Tendermint?

tendermint assumes that calling `PrepareProposal` will never throw and error, it doesn't know what to do in that scenario. fixing the bug that causes #913 means that we find the source of that bug and never hit this error so tendermint can continue to make that assumption

pkg/shares/share_splitting.go
rootulp: something like https://github.com/rootulp/celestia-app/blob/57126cc55babd643eb73be3dca7a962d8e4e0d25/app/prepare_proposal.go#L59-L68

```golang
	// estimate the square size. This estimation errors on the side of larger
	// squares but can only return values within the min and max square size.
	squareSize, totalSharesUsed, err := estimateSquareSize(parsedTxs)
	if err != nil {
		app.Logger().Error(
			"failure to estimate square size while creating a proposal block",
			"error",
			err.Error(),
		)
		panic(err)
	}
```

rootulp:
@evan-forbes I confirmed this is an issue in estimate square size. After the most recent commit, the fuzz test fails with

```
--- FAIL: TestFuzzPrepareProcessProposal (0.58s)
    --- FAIL: TestFuzzPrepareProcessProposal/randomized_inputs_to_Prepare_and_Process_Proposal (0.54s)
panic: shares used: 61287 exceeds number of shares in square size: 128  [recovered]
	panic: shares used: 61287 exceeds number of shares in square size: 128

goroutine 3948 [running]:
```

Given https://github.com/celestiaorg/celestia-app/pull/1089 is already in progress, I'm inclined to close this PR because we're reasonably confident that https://github.com/celestiaorg/celestia-app/pull/1089 will close https://github.com/celestiaorg/celestia-app/issues/913

PR #1096, opened 2022-12-08
feat: implement `share.RawData()`

rootulp:
This is the celestia-app side of a fix for https://github.com/celestiaorg/celestia-node/issues/1448. It introduces `share.RawData()` for usage in celestia-node

### DIFF
--- a/pkg/shares/shares.go
+++ b/pkg/shares/shares.go
@@ -28,7 +28,7 @@ func (s Share) NamespaceID() namespace.ID {
 
 func (s Share) InfoByte() (InfoByte, error) {
 	if len(s) < appconsts.NamespaceSize+appconsts.ShareInfoBytes {
-		panic(fmt.Sprintf("share %s is too short to contain an info byte", s))
+		return 0, fmt.Errorf("share %s is too short to contain an info byte", s)
 	}
 	// the info byte is the first byte after the namespace ID
 	unparsed := s[appconsts.NamespaceSize]
--- a/pkg/shares/shares.go
+++ b/pkg/shares/shares.go
@@ -35,19 +35,53 @@ func (s Share) InfoByte() (InfoByte, error) {
 	return ParseInfoByte(unparsed)
 }
 
-func (s Share) SequenceLength() (uint64, error) {
+// SequenceLen returns the value of the sequence length varint, the number of
+// bytes occupied by the sequence length varint, and optionally an error. It
+// returns 0, 0, nil if this is a continuation share (i.e. doesn't contain a
+// sequence length).
+func (s Share) SequenceLen() (len uint64, numBytes int, err error) {
+	isSequenceStart, err := s.isSequenceStart()
+	if err != nil {
+		return 0, 0, err
+	}
+	if !isSequenceStart {
+		return 0, 0, nil
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/shares/shares_test.go
```diff
@@ -98,3 +98,181 @@ func generateRandomBlobOfShareCount(count int) coretypes.Blob {
...
+		1, 1, 1, 1, 1, 1, 1, 1, // namespace
+		0,                             // info byte
+		1, 2, 3, 4, 5, 6, 7, 8, 9, 10, // data
+	}
+	firstCompactShare := []byte{
+		0, 0, 0, 0, 0, 0, 0, 1, // namespace
+		1,           // info byte
+		10, 0, 0, 0, // sequence len
+		1, 2, 3, 4, 5, 6, 7, 8, 9, 10, // data
+	}
```
cmwaters, line 220: More of a side question: aren't there supposed to be two reserved bytes as well?

cmwaters (approved):
LGTM

pkg/shares/shares_test.go
rootulp: Great question and yes! This unit test still passes because the two reserved bytes are treated as part of the data (i.e. `1, 2` are the reserved bytes in this test) but this may make it clearer:

```golang
	firstCompactShare := []byte{
		0, 0, 0, 0, 0, 0, 0, 1, // namespace
		1,           // info byte
		10, 0, 0, 0, // sequence len
                15, 0, // reserved bytes
		1, 2, 3, 4, 5, 6, 7, 8, 9, 10, // data
	}
```

rootulp: Explicitly added in https://github.com/celestiaorg/celestia-app/pull/1096/commits/a2b374d04cf52e412c8d77618b033bf1ccf682f4

FWIW I don't have signal on whether consumers of `share.RawData()` want the reserved bytes included or not. I elected to include them now and we can revisit if we get signal otherwise.

pkg/shares/shares.go
```diff
@@ -70,3 +105,9 @@ func FromBytes(bytes [][]byte) (shares []Share) {
 	}
 	return shares
 }
+
+// numberOfBytesVarint calculates the number of bytes needed to write a varint of n
+func numberOfBytesVarint(n uint64) (numberOfBytes int) {
+	buf := make([]byte, binary.MaxVarintLen64)
+	return binary.PutUvarint(buf, n)
+}
```
evan-forbes, line 113: can we use this instead? https://github.com/celestiaorg/celestia-app/blob/8c4982c9b111377b9ab3566f3619cff23f472101/pkg/shares/utils.go#L14-L18 ?

evan-forbes (commented):
one question, but otherwise LGTM :+1:

pkg/shares/shares_test.go
cmwaters: Who are the consumers of `share.RawData()`? I assume they'll want all the information they need to parse the txs within the share so perhaps everything minus the namespace itself

evan-forbes (approved):

cmwaters (approved):

pkg/shares/shares.go
rootulp: Great question and yes we can. I'll cut a new issue to consolidate these `DelimLen` and `numberOfBytesVarint` functions in a subsequent PR.

PR #1104, opened 2022-12-09
feat: use `private` chain ID for single node

rootulp:
If you try connecting a local celestia-node instance to celestia-app that is running via this script, it fails with:

```error
$ celestia bridge init --p2p.network "test"
Error: invalid network specified: test
```

Ref:
https://github.com/celestiaorg/celestia-node/blob/52048ce466d58c1688ba48de0a2656b27306634d/nodebuilder/p2p/network.go#L12-L19

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -2,7 +2,7 @@
 
 set -o errexit -o nounset
 
-CHAINID="test"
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):
:+1: nice

Issue #1110, opened 2022-12-13
Refactor the share splitting logic


evan-forbes:
Currently, the share splitting logic supports super legacy code for splitting and merging shares that don'e use indexes. It makes the code more confusing than it has to be, and we have no reason to support this, so we should remove this functionality.

rootulp:
> and we have no reason to support this

I thought the conditional logic for using share indexes was originally added to support blocks without share indexes. I assume early blocks on Mamaki don't contain share indexes. Why is it safe to remove backwards compatibility now? Are we guaranteed that future celestia-app versions won't be used for testnets with blocks that contain transactions without share indexes?

> It makes the code more confusing than it has to be

Agreed! If it is safe to remove, then I'm on-board

evan-forbes:
> Why is it safe to remove backwards compatibility now? 

mamaki's days are numbered! maybe people are running their own networks else where, but since we haven't hit 1.0 I think its fine. speaking of which, we should actually try to get rid of this before 1.0, so we don't have to keep it for any longer than we have to :grimacing:

evan-forbes:
I guess we don't have to hit 1.0 on the feature freeze, and can spend some time cleaning up apis such as this before actually cut

evan-forbes:
as I think was made pretty clear by #1108, is that our share tests also need unit tests for splitting shares using indexes. Currently, we rely too much on the prepare/process fuzzer, which does test that, it just relies on randomness to test for edge cases.

rootulp:
IIUC the goal of this issue is to deprecate the `useShareIndexes` parameter because all block data (in tests and in networks that we care about moving forward) use blocks that contain transactions with share indexes.

rootulp:
Related: https://github.com/celestiaorg/celestia-node/issues/1231

cmwaters:
Closing this as it is no longer relevant given https://github.com/celestiaorg/celestia-app/pull/1690

PR #1111, opened 2022-12-13
fix!: Share splitting bug

### LINKED ISSUE #1108: Bug: v0.10.0 Share splitting with use the incorrect index when many small blobs are submitted
Currently, with each blob submitted, we overestimate the bytes takes up by the actual transaction itself. In some cases, if we have enough blob transactions make it in the same block, then the amount of overestimated shares will be greater than the amount of padding added between the end of the compact shares and start of the sparse shares. This results in using an incorrect share index, and thus proposing an invalid block, validators will validate the share commitments in each PFB using the incorrect starting share index for the first blob. 

This can be fixed by forcing the share splitting logic to use the first share index, instead of calculating it itself.

evan-forbes:
## Overview

fixes the bug described in #1108

the modified tests fails on main. Also see #1110, which we should really get to as soon as we don't have more important things as this code is ancient

closes #1108

## Checklist


pkg/shares/share_splitting.go
```diff
@@ -45,6 +45,11 @@ func Split(data coretypes.Data, useShareIndexes bool) ([]Share, error) {
 			BlobSharesUsed(len(data.Blobs[0].Data)),
 			int(data.SquareSize),
 		)
+		// force msgIndexes to be the first share index
+		if len(blobIndexes) != 0 && useShareIndexes {
+			blobShareStart = int(blobIndexes[0])
+		}
```
cmwaters, line 51: Are we able to specifically add a unit test for this?

pkg/shares/share_splitting.go
```diff
@@ -45,6 +45,11 @@ func Split(data coretypes.Data, useShareIndexes bool) ([]Share, error) {
 			BlobSharesUsed(len(data.Blobs[0].Data)),
 			int(data.SquareSize),
 		)
+		// force msgIndexes to be the first share index
```
rootulp, line 48:
```suggestion
		// force blobSharesStart to be the first share index
```

rootulp (dismissed):

rahulghangas (dismissed):

pkg/shares/share_splitting.go
evan-forbes: :+1: yeah defo needed, thanks. fails on main
https://github.com/celestiaorg/celestia-app/pull/1111/commits/c67976551ac4cebc636bef35ea73ff1ffca65bb2

also https://github.com/celestiaorg/celestia-app/issues/1110#issuecomment-1352389547

pkg/shares/shares_test.go
```diff
@@ -62,6 +64,32 @@ func TestMerge(t *testing.T) {
 	}
 }
 
+// TestPadFirstIndexedBlob ensures that we are adding padding to the first share
+// instead of calculating the value.
```
rootulp, line 68: [non-blocking] sorry I don't understand what "calculating the value" means in this context. Do you mind elaborating? It doesn't strictly have to go into the comment b/c I don't want to block this PR's merge.

rootulp (approved):

pkg/shares/shares_test.go
evan-forbes: calculate using the non-interactive defaults, previously this is what it was doing. The estimation process was accurate enough that this was the case the vast majority of the time, but not for all cases.

Issue #1120, opened 2022-12-14
specs: update sequence length and reserved bytes


rootulp:
Currently, the specs state that the sequence length is encoded as a varint and that the reserved bytes are encoded as a two byte padded varint. We recently decided to change both to big endian 4 byte unsigned integers.

PR #1124, opened 2022-12-14
specs: fixed sequence length

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1120


specs/src/specs/figures/sparse_start_share.dot
```diff
@@ -7,7 +7,7 @@ digraph G {
             <td align="left" border="0" cellpadding="0">0</td>
             <td align="left" border="0" cellpadding="0">8</td>
             <td align="left" border="0" cellpadding="0">9</td>
-            <td align="left" border="0" cellpadding="0">12</td>
+            <td align="left" border="0" cellpadding="0">13</td>
```
evan-forbes, line 10: unrelated comment

if we pad this to 16, zk rollups could save quite a few cycles, correct? could they do that themselves?

specs/src/specs/consensus.md
```diff
@@ -55,8 +55,9 @@
 | `MAX_VALIDATORS`                        | `uint16` | `64`         |         | Maximum number of active validators.                                                                                                                                |
 | `NAMESPACE_ID_BYTES`                    | `uint64` | `8`          | `byte`  | Size of namespace ID, in bytes.                                                                                                                                     |
 | `NAMESPACE_ID_MAX_RESERVED`             | `uint64` | `255`        |         | Value of maximum reserved namespace ID (inclusive). 1 byte worth of IDs.                                                                                            |
+| `SEQUENCE_BYTES`                        | `uint64` | `4`          | `byte`  | The number of bytes used to store the sequence length in the first share of a sequence                                                                              |
```
evan-forbes, line 58: is this a uint32? same question for the change below
```suggestion
| `SEQUENCE_BYTES`                        | `uint32` | `4`          | `byte`  | The number of bytes used to store the sequence length in the first share of a sequence                                                                              |
```

rootulp: I think the type for this constant only needs to contain the number of sequence bytes (`4`) so hypothetically this could be anything in the range from `uint8` to `uint64`. I'm not strongly opinionated on which type we use but want to point out that this type doesn't strictly need to be `uint32` even though the value stored in these four bytes will be ` uint32`. Does that make sense?  

On an related note, appconsts.go doesn't explicitly type integer constants so the specs diverge from the current implementation. https://github.com/celestiaorg/celestia-app/issues/1130

evan-forbes (dismissed):
LGTM, but would defer to others

specs/src/specs/figures/sparse_start_share.dot
rootulp: @preston-evans98 commented that aligning to 4 or 8 byte boundaries would help their use-case. I assume this is for performance reasons but if that's the case, we may want benchmarks to make an informed decision on how best to pad / align the universal share prefix.

evan-forbes: cool, 👍

specs/src/specs/data_structures.md
```diff
@@ -497,31 +497,31 @@ For shares **with a reserved namespace ID through [`NAMESPACE_ID_MAX_RESERVED`](
 
 > **Note** The first [`NAMESPACE_ID_BYTES`](./consensus.md#constants) of a share's raw data `rawData` is the namespace ID of that share, `namespaceID`. The next [`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are for share information.
 
-- If this is the first share of a sequence, the next 1 to 10 bytes contain a [varint](https://developers.google.com/protocol-buffers/docs/encoding) of the `uint64` length of the sequence that follows, in bytes.
+- If this is the first share of a sequence, the next [`SEQUENCE_BYTES`](./consensus.md#constants) contain a big endian `uint32` that represents the length of the sequence that follows in bytes.
```
MSevey, line 500:
```suggestion
- If this is the first share of a sequence, the next [`SEQUENCE_BYTES`](./consensus.md#constants) contain a big endian `uint64` that represents the length of the sequence that follows in bytes.
```

Building on Evan's comment above, I think this needs to be updated to match right? Seems like the table references `uint64` but all the explanations here reference `uint32`

rootulp: Sorry this is a bit confusing. The two types are for two different fields. 
1. The constant `SEQUENCE_BYTES` is an a hard-coded number: `4`. The type of this constant can be `uint8`, `uint16`, `uint32`, or `uint64` because all four types can contain the number `4`. In practice, I choose `uint64` to match some of the other constants in that table.
2. The value written to the `4` sequence bytes is a `uint32` and it is variable (i.e. the sequence length is based on the sequence it is describing). Since a `uint64` would occupy `8` bytes, it can't be written to the sequence bytes. We aligned on a `uint32` for the sequence length in https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-012-sequence-length-encoding.md

Maybe this comment helps: https://github.com/celestiaorg/celestia-app/pull/1124#discussion_r1049838170

specs/src/specs/data_structures.md
```diff
@@ -497,31 +497,31 @@ For shares **with a reserved namespace ID through [`NAMESPACE_ID_MAX_RESERVED`](
...
+- The remaining [`SHARE_SIZE`](./consensus.md#constants)`-`[`NAMESPACE_ID_BYTES`](./consensus.md#constants)`-`[`SHARE_INFO_BYTES`](./consensus.md#constants) `-` [`SEQUENCE_BYTES`](.consensus.md#constants) bytes (only if this is the first share of a sequence) `-` [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes are transactions, intermediate state roots, or evidence data depending on the namespace of ths share. Each transaction, intermediate state root, or evidence is prefixed with a [varint](https://developers.google.com/protocol-buffers/docs/encoding) of the length of that unit.
 - If there is insufficient transaction, intermediate state root, or evidence data to fill the share, the remaining bytes are filled with `0`.
 
 First share in a sequence:
 
 ![fig: compact start share.](./figures/compact_start_share.svg)
 
-where reserved bytes would be `14` encoded as an unsigned varint (`[0b00001110, 0b00000000]`).
+where reserved bytes would be `17` binary big endian `uint32` (`[0 0 0 17]`).
```
adlerjohn, line 509: That's a decimal encoding, no? Binary would be
```suggestion
where reserved bytes would be `17` as a binary big endian `uint32` (`[0b00000000, 0b00000000, 0b00000000, 0b00010001]`).
```

specs/src/specs/data_structures.md
```diff
@@ -497,31 +497,31 @@ For shares **with a reserved namespace ID through [`NAMESPACE_ID_MAX_RESERVED`](
...
 
 Continuation share in a sequence:
 
 ![fig: compact continuation share.](./figures/compact_continuation_share.svg)
 
-where reserved bytes would be `80` encoded as an unsigned varint (`[0b01010000, 0b00000000]`).
+where reserved bytes would be `80` binary big endian `uint32` (`[0 0 0 80]`).
```
adlerjohn, line 515: Ditto
```suggestion
where reserved bytes would be `80` as a binary big endian `uint32` (`[0b00000000, 0b00000000, 0b00000000, 0b01010000]`).
```

specs/src/specs/consensus.md
```diff
@@ -55,8 +55,9 @@
...
 | `SHARE_INFO_BYTES`                      | `uint64` | `1`          | `byte`  | The number of bytes used for [share](data_structures.md#share) information                                                                                          |
-| `SHARE_RESERVED_BYTES`                  | `uint64` | `2`          | `byte`  | Bytes reserved at the beginning of each [share](data_structures.md#share). Must be sufficient to represent `SHARE_SIZE`.                                            |
+| `SHARE_RESERVED_BYTES`                  | `uint64` | `4`          | `byte`  | The number of bytes used to store the location of the first unit in a compact share. Must be able to represent any index in `SHARE_SIZE`.                           |
```
adlerjohn, line 60: This is a bit confusing because `SHARE_SIZE` is an integer, not a list.

```suggestion
| `SHARE_RESERVED_BYTES`                  | `uint64` | `4`          | `byte`  | The number of bytes used to store the location of the first unit in a compact share. Must be able to represent any integer up to and including `SHARE_SIZE - 1`.                           |
```

adlerjohn (approved):

PR #1126, opened 2022-12-15
feat: use square size params

rahulghangas:
Removes dependence of `appconsts.DefaultMinSquareSize` and `appconsts.DefaultMaxSquareSize` and instead fetches the relevant params from the blob keeper during process proposal

- [x] remove usage of default values
- [x] testing


cmwaters (commented):
Looks good. Just have a few trivial comments.

A slightly adjacent question I have with these upper and lower square size bounds is how are they initially calculated and how do we know when it's time to change them?

pkg/da/data_availability_header.go
```diff
@@ -124,11 +119,13 @@ func DataAvailabilityHeaderFromProto(dahp *daproto.DataAvailabilityHeader) (dah
...
 
-	return dah, dah.ValidateBasic()
+	return dah, dah.ValidateBasic(minSquareSize, maxSquareSize)
 }
 
 // ValidateBasic runs stateless checks on the DataAvailabilityHeader.
-func (dah *DataAvailabilityHeader) ValidateBasic() error {
+func (dah *DataAvailabilityHeader) ValidateBasic(minSquareSize, maxSquareSize int) error {
+	maxExtendedSquareWidth := minSquareSize * 2
+	minExtendedSquareWidth := maxSquareSize * 2
```
rootulp, line 128: [blocking] there is a subtle bug here.
```suggestion
	minExtendedSquareWidth := minSquareSize * 2
	maxExtendedSquareWidth := maxSquareSize * 2
```

evan-forbes: wow, good catch!!! This would have broken celestia-node for sure. We don't call validate basic iirc, since the nodes are the ones generating the DAH.

We should think about how this will get used there as well, and making this function state-full might not be the best approach. I think we can stick to using the default max and mins. If we want to change the max and mins beyond those values in governance, then we can force a hardfork. wdyt?

app/estimate_square_size.go
```diff
@@ -202,3 +202,8 @@ func overEstimateMalleatedTxSize(txLen, blobSize int) int {
 	// actually malleating the tx
 	return appconsts.MalleatedTxBytes + appconsts.MalleatedTxEstimateBuffer + malleatedTxLen
 }
+
+func isValidSquareSize(squareSize, minSquareSize, maxSquareSize int) bool {
+	// Check that square size is with range
```
rootulp, line 207: [typo] 
```suggestion
	// Check that square size is within range
```

app/estimate_square_size.go
```diff
@@ -202,3 +202,8 @@ func overEstimateMalleatedTxSize(txLen, blobSize int) int {
...
+	return squareSize < minSquareSize || squareSize > maxSquareSize
+}
```
rootulp, line 209: [blocking] this function returns `true` if the squareSize is invalid so the function name is inaccurate. Either the function name or the return statement should be updated.

Proposal:

```suggestion
func isValidSquareSize(squareSize, minSquareSize, maxSquareSize int) bool {
	return squareSize >= minSquareSize && squareSize <= maxSquareSize
}
```

app/process_proposal.go
```diff
@@ -35,6 +40,21 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
 		}
 	}
 
+	squareSize := data.SquareSize
+
+	// validate proposed square size
+	if !isValidSquareSize(int(squareSize), int(minSquareSize), int(maxSquareSize)) {
```
rootulp, line 46: [question] given this line assumes `isValidSquareSize` returns true for a valid square, why didn't the unit test that was added in this PR identify the logic error inside `isValidSquareSize`?

rootulp (commented):
[question] why aren't unit tests running on CI for this PR? I checked it out locally and I'm observing a test failure

```
--- FAIL: TestBlobInclusionCheck (0.03s)
    /Users/rootulp/git/rootulp/celestia/celestia-app/app/test/process_proposal_test.go:124: 
        	Error Trace:	/Users/rootulp/git/rootulp/celestia/celestia-app/app/test/process_proposal_test.go:124
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestBlobInclusionCheck
        	Messages:   	valid untouched data
FAIL
FAIL	github.com/celestiaorg/celestia-app/app/test	0.534s
FAIL
```

app/test/process_proposal_test.go
```diff
@@ -96,6 +96,14 @@ func TestBlobInclusionCheck(t *testing.T) {
...
 			expectedResult: abci.ResponseProcessProposal_REJECT,
 		},
+		{
+			name:  "square size greater than max square size",
+			input: validData(),
+			mutator: func(d *core.Data) {
+				d.SquareSize = appconsts.DefaultMaxSquareSize * 2
+			},
+			expectedResult: abci.ResponseProcessProposal_REJECT,
+		},
```
rootulp, line 106: [optional] an extra test case

```go
		{
			name:  "square size of zero",
			input: validData(),
			mutator: func(d *core.Data) {
				d.SquareSize = 0
			},
			expectedResult: abci.ResponseProcessProposal_REJECT,
		},
```

x/blob/keeper/params.go
```diff
@@ -21,12 +22,18 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
 
 // MinSquareSize returns the MinSquareSize param
 func (k Keeper) MinSquareSize(ctx sdk.Context) (res uint32) {
+	if ctx.BlockHeader().Height < 1 {
+		return appconsts.DefaultMinSquareSize
+	}
```
rootulp, line 27: [non blocking question] sorry I'm not familiar with defining genesis params. Is there a way to define the DefaultMinSquareSize param at chain genesis time so that we don't need this conditional?

If it isn't possible, why is a similar conditional not needed for the `GasPerBlobByte` param later in this file?

rootulp (commented):
I don't have any blocking feedback but there is one unresolved blocking question on this PR: https://github.com/celestiaorg/celestia-app/pull/1126/files#r1049690184 so holding off on approval

app/prepare_proposal.go
```diff
@@ -22,7 +22,10 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 
 	// estimate the square size. This estimation errors on the side of larger
 	// squares but can only return values within the min and max square size.
-	squareSize, totalSharesUsed := estimateSquareSize(parsedTxs)
+	ctx := app.NewContext(true, core.Header{Height: app.LastBlockHeight()})
```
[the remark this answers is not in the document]
rahulghangas: That was discussed asynchronously, I will mark it resolved

x/blob/keeper/params.go
evan-forbes: I think it would be good to document this in the code

app/prepare_proposal.go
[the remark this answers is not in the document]
rootulp: Out of curiosity what was the outcome of the discussion? Is a gas meter not needed because it wouldn't impact the usage of this context in this scope?

x/blob/keeper/params.go
rootulp: Is it possible for someone to include a payForBlob in the first block of the chain? It seems unlikely but if yes, would `GasPerBlobByte` also need to be initialized?

evan-forbes: that's a good point, I don't think so, as the params are used in many places, but I'm not sure of what is done to make that the case! perhaps we can do that process a bit earlier, so that it works for processproposal, which would almost certainly be a better solution

evan-forbes, line 27: I'm okay with merging as soon as we add a comment or two here, thanks!
```suggestion
	// use the default size for the first block so that we return a value on the
	// first block in PrepareProposal
	if ctx.BlockHeader().Height < 1 {
		return appconsts.DefaultMinSquareSize
	}
```

rahulghangas:
I'm going to create an issue on the cosmos-sdk repository to get their opinion

x/blob/keeper/params.go
cmwaters: > Apparently the genesis params aren't set until the first block is committed

This seems odd. You should be able to set params in genesis that can be used for the first block.

x/blob/keeper/params.go
```diff
@@ -21,12 +22,22 @@ func (k Keeper) SetParams(ctx sdk.Context, params types.Params) {
 
 // MinSquareSize returns the MinSquareSize param
 func (k Keeper) MinSquareSize(ctx sdk.Context) (res uint32) {
+	// use the default size for the first block so that we return a value on the
+	// first block in PrepareProposal
+	if ctx.BlockHeader().Height < 1 {
```
cmwaters, line 27: I'm pretty sure the first block is at height 1 not 0. Also we should be using something like `InitialHeight` as it's possible with Tendermint chains to start at any height. 

Most importantly, I'm still not sure if this logic is necessary. At least I haven't seen this in other SDK modules.

rahulghangas: I'm not sure either. However, with this logic removed, prepare proposal (where the params are fetched for the first time) blocks and using the cli to query the params panics with an error

rahulghangas: chain logs
```
5:47PM INF Timed out dur=10000 height=1 module=consensus round=0 step=3
5:47PM INF Ensure peers module=pex numDialing=0 numInPeers=0 numOutPeers=0 numToDial=10
5:47PM INF No addresses to dial. Falling back to seeds module=pex
5:48PM INF Ensure peers module=pex numDialing=0 numInPeers=0 numOutPeers=0 numToDial=10
5:48PM INF No addresses to dial. Falling back to seeds module=pex
```

using `celestia-appd query blob params` results in
```
Error: rpc error: code = Unknown desc = UnmarshalJSON cannot decode empty bytes: panic
```

cmwaters: Interesting. Maybe you're right. I'd have thought the genesis params would be set at `InitChain` which comes before `PrepareProposal` but maybe I'm wrong.

What you could also do is add a log for when the params first get set in the Keeper and we can see if that comes before the first block gets proposed. Also wouldn't this panic you're seeing also happen in `PrepareProposal`?

evan-forbes:
it looks like others have ran into this issue as well
https://github.com/cosmos/cosmos-sdk/issues/14446

facundomedica:
This might be a fix for this: https://github.com/cosmos/cosmos-sdk/pull/14505 would love to get some more eyes on it.
- The height in PrepareProposal was always `currentHeight-1`. Now that's fixed, PrepareProposal can't be called with height < 1.
- The state from InitChain/Genesis wasn't accesible by PrepareProposal/ProcessProposal. That's also fixed, so you should be able to get params safely.

cmwaters:
Is there any update on this PR? 

Can we either use the hack where we use the default size for the first block or we cherry-pick the commit that facundo did.

evan-forbes:
I think the decision was to not pursue this change, so I think we can close this issue and PR. 

IIRC, the reasoning was that while we want something like BlockSize (determined in MB), we don't want to expose square sizes as those are more complex internal params. We'd also still have a "HardMaxSquareSize", that we'd have to hardfork to change, so it seems a bit extra to have a blocksize param along with a max square size param along with a hard max square size

cmwaters:
In that case can we look to either increase the current max square size from 128 to 256 i.e. 32MB square size or reduce DefaultMaxBlockBytes to 4MB. What I'm trying to get at is that the ceiling in terms of size should be something we control (i.e. MaxBlockBytes) and not something we don't (i.e. MaxSquareSize)

evan-forbes:
closing since we're simply using the existing block size limits in tendermint

Issue #979, opened 2022-11-08
Proposal: Add validity checks for PFB transactions during ProcessProposal


evan-forbes:
Due to tendermint's deferred execution model, the transaction in a block are not verified during consensus, so it is possible for a block producer to include invalid transactions (which are ignored). It has been proposed synchronously that we add additional verification steps to `ProcessProposal` for all transactions that have `MsgPayForBlob` `sdk.Msg`s. This would ensure that all `MsgPayForBlob`'s that make it into a block are valid provided an honest majority, which allows for certain types of zk rollups to avoid proving those same checks in their circuit. 

In practice this could mean that we only add an additional signature check, but we could also check that there are enough funds to pay gas, perform stateless checks via `ValidateBasic`, and, if the transaction contains other `sdk.Msg`s, execute those as well.

PR #1128, opened 2022-12-15
feat: add additonal validity checks during process proposal

rahulghangas:
Adds additional validity checks during process proposal to make sure that all transactions that include `MsgPayForBlob` messages should be valid. See #979 for details

- [x] Check validity of signer and signer length for `MsgPayForBlob` messages
- [x] do basic validation check using `ValidateBasic`
- [x] Execute all messages in txs with `MsgPAyForBlob` messages
- [x]  Make sure there is enough gas to pay for transaction


evan-forbes:
> Execute all messages in txs with MsgPAyForBlob messages

we actually don't have to do this atm, as we are currently only supporting sdk.Tx with a single sdk.Msg, that being a PFB

app/process_proposal.go
```diff
@@ -79,25 +79,63 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
+					app.Logger().Error("Tx including MsgPayForBlob is invalid")
+					return abci.ResponseProcessProposal{
+						Result: abci.ResponseProcessProposal_REJECT,
+					}
+				}
+
+				response := app.BaseApp.CheckTx(abci.RequestCheckTx{
+					Tx:   rawTx,
+					Type: abci.CheckTxType_New,
+				})
```
evan-forbes, line 106: [blocking]
We can not use `CheckTx` here, as it uses its own copy of the state, not the actual state! I think its very probable that this would not be deterministic, since the check tx state depends on which txs that validator has added to their mempool.

we also only need to check the signature, and therefore also looking up the account.

app/process_proposal.go
```diff
@@ -79,25 +79,63 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
+
+				if response.GasUsed > response.GasWanted {
+					return abci.ResponseProcessProposal{
+						Result: abci.ResponseProcessProposal_REJECT,
+					}
+				}
```
evan-forbes, line 112: let's only check the signature for now as state in the target issue. As a side note, I think manually checking for gas would be more explicit

app/process_proposal.go
```diff
@@ -79,25 +79,63 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
 			pfb, ok := msg.(*types.MsgPayForBlob)
 			if !ok {
 				app.Logger().Error("Msg type does not match MsgPayForBlob URL")
 				continue
 			}
 
-			if err = pfb.ValidateBasic(); err != nil {
-				logInvalidPropBlockError(app.Logger(), req.Header, "invalid MsgPayForBlob", err)
+			signers := msg.GetSigners()
+			if len(signers) != 1 {
```
evan-forbes, line 124: why are we rejecting txs that have more than one signer?

app/process_proposal.go
```diff
@@ -79,25 +79,63 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
+				}
+			}
+
+			signer, err := sdk.AccAddressFromBech32(pfb.Signer)
+			if err != nil {
+				// this panic should be unreachable
+				panic("signer address validation should not fail after basic validation has already been done")
+			}
+
+			if !bytes.Equal(signer, signers[0]) {
```
evan-forbes, line 137: why are we checking the signer here?

evan-forbes (commented):
had a few blocking questions and comments

app/process_proposal.go
rahulghangas: Ah makes sense, I will revert to previous changes

evan-forbes:
closing in favor of #1300

Issue #405, opened 2022-05-13
Use the `file` keyring backend as a default


evan-forbes:
In order to simplify keyring use, celestia-node will be using the `file` keyring backend as a default. In order to stay consistent and avoid possible confusion for users, we should investigate exactly what this entails, and then make a decision on switching the backend here as well.

rootulp:
Is this issue coupled to the celestia-node implementation of the same? The celestia-node side isn't implemented yet: https://github.com/celestiaorg/celestia-node/issues/603 and https://github.com/celestiaorg/celestia-node/blob/0c2669574fcbc8478cf9572ae5d36c98901916e8/nodebuilder/state/keyring.go#L25

cmwaters:
Is there a justification anywhere about why `file` over `os`? - is it just to simplify things?

evan-forbes:
cc @renaynay do you remember the exact reason for switching?

renaynay:
`os` is wonky on macos

PR #1141, opened 2022-12-20
chore: switch default keyring backend

rahulghangas:
- [x] closes https://github.com/celestiaorg/celestia-app/issues/405
- [x] manually test that default value is switched

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -151,12 +151,16 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig encoding.Config) {
 
 	server.AddCommands(rootCmd, app.DefaultNodeHome, NewAppServer, createAppAndExport, addModuleInitFlags)
 
+	// set the default keyring backend to `file`
+	keybase := keys.Commands(app.DefaultNodeHome)
+	keybase.Flag(flags.FlagKeyringBackend).DefValue = "file"
+
 	// add keybase, auxiliary RPC, query, and tx child commands
 	rootCmd.AddCommand(
 		rpc.StatusCommand(),
 		queryCommand(),
 		txCommand(),
-		keys.Commands(app.DefaultNodeHome),
+		keybase,
 	)
 }
 


cmwaters (approved):
LGTM.

I also built it locally and can see it has changed:
```
Flags:
  -h, --help                     help for keys
      --keyring-backend string   Select keyring's backend (os|file|test) (default "file")
      --keyring-dir string       The client Keyring directory; if omitted, the default 'home' directory will be used
      --output string            Output format (text|json) (default "text")
```

cmd/celestia-appd/cmd/root.go
```diff
@@ -151,12 +151,16 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig encoding.Config) {
 
 	server.AddCommands(rootCmd, app.DefaultNodeHome, NewAppServer, createAppAndExport, addModuleInitFlags)
 
+	// set the default keyring backend to `file`
+	keybase := keys.Commands(app.DefaultNodeHome)
+	keybase.Flag(flags.FlagKeyringBackend).DefValue = "file"
```
evan-forbes, line 156: [blocking question]
do we know if this default will get overwritten by the client.toml setting? Will this default override that setting? will changing that have a different effect?

evan-forbes: I know this is merged, but can we follow up with these questions as well @rahulghangas ?

rootulp: My client.toml has `keyring-backend = "os"` so it seems like in increasing order of precedence:

1. client.toml
2. the default value on this line
3. CLI flag: `--keyring-backend`

In other words, `client.toml` isn't helpful for end-users because the value specified on this line will override it

rootulp: ^ disregard, the help message is misleading. I did some investigating on [this branch](https://github.com/rootulp/cosmos-sdk/tree/rp/keys-debug) and learned that with this change, the help message claims the default is `file` but the actual keyring backend that gets used is whatever is in client.toml. If a value isn't specified in client.toml, then I encounter:

```error
$ celestia-appd keys debug
Error: couldn't get key ring: unknown keyring backend
```

even though that's when I would expect the default specified on this line to be applied.

rootulp: Can resolve. See https://github.com/celestiaorg/celestia-app/issues/1201

PR #1095, opened 2022-12-08
feat: use kvstore to fetch gas per blob byte

rahulghangas:
Replaces gas per blob byte constant with stored param

- [x] closes #1001

-- Note - To be merged after #1031

PR #1145, opened 2022-12-21
chore: address rootulp feedback on PR 1095

rootulp:
Addresses my late feedback on https://github.com/celestiaorg/celestia-app/pull/1095


evan-forbes (approved):
thx! good feedback imo

(we should probably require two reviews to merge now, which is only implicit atm)

PR #1146, opened 2022-12-21
feat: implement `ShareSequence.RawData()`

rootulp:
Part of https://github.com/celestiaorg/celestia-node/issues/1448

PR #1147, opened 2022-12-22
fix: remove inaccurate `BlobSharesUsed`

rootulp:
Follow up to https://github.com/celestiaorg/celestia-app/pull/1137 that fixes Godocs and removes an inaccurate helper function.

While investigating https://github.com/informalsystems/audit-celestia/issues/14 I noticed that the `BlobSharesUsed` helper still exists. It should be removed because it assumes blobs are prefixed with varints (no longer the case) and doesn't differentiate between the first and continuation shares in a sparse share sequence. The helper `SparseSharesNeeded` should be used instead.


evan-forbes:
whooops wait could this be the reason for the flake in #1146?

rootulp:
TBH unclear if it's the cause of the flake b/c minimal debug logging [here](https://github.com/celestiaorg/celestia-app/actions/runs/3753129913/jobs/6376098053)

app/process_proposal.go
```diff
@@ -103,7 +103,7 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
 				}
 			}
 
-			commitment, err := inclusion.GetCommit(cacher, dah, int(malleatedTx.ShareIndex), shares.BlobSharesUsed(int(pfb.BlobSize)))
+			commitment, err := inclusion.GetCommit(cacher, dah, int(malleatedTx.ShareIndex), shares.SparseSharesNeeded(uint32(pfb.BlobSize)))
```
rootulp, line 106: FWIW it seems very hacky to convert a `uint64` => ` uint32` like this because overflows. We should consider changing https://github.com/rootulp/celestia-app/blob/e049c4eb97c5935728b7bf8fe6f63a42706a6772/proto/blob/tx.proto#L30 to `uint32` or being more defensive here

evan-forbes (approved):

PR #1152, opened 2022-12-22
feat: implement non-interactive default rules for reduced padding

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1149

### DIFF
--- a/app/write_square_test.go
+++ b/app/write_square_test.go
@@ -75,9 +75,22 @@ func Test_finalizeLayout(t *testing.T) {
 			nonreserveStart: 32,
 			ptxs: generateParsedTxsWithNIDs(
 				[][]byte{ns1, ns1, ns1},
-				[]int{1000, 10000, 100000},
+				[]int{1000, 10000, 100000}, // blob share lengths of 2, 20, 199 respectively
 			),
-			expectedIndexes: []uint32{32, 48, 128},
+			expectedIndexes: []uint32{
+				// BlobMinSquareSize(2) = 2 so the first blob has to start at the
+				// next multiple of 2 > 32 which is 32. This blob occupies
+				// shares 32 to 34.
+				32,
+				// BlobMinSquareSize(20) = 8 so the second blob has to start at
+				// the next multiple of 8 > 34 which is 40. This blob occupies
+				// shares 40 to 48.
+				40,
+				// BlobMinSquareSize(199) = 16 so the third blob has to start at
+				// the next multiple of 16 > 48 which is 64. This blob occupies
+				// shares 64 to 263.
+				64,
+			},
 		},
 		{
 			squareSize:      32,
--- a/pkg/shares/non_interactive_defaults.go
+++ b/pkg/shares/non_interactive_defaults.go
@@ -1,10 +1,15 @@
 package shares
 
+import (
+	"math"
+)
+
 // FitsInSquare uses the non interactive default rules to see if blobs of
 // some lengths will fit in a square of squareSize starting at share index
 // cursor. Returns whether the blobs fit in the square and the number of
 // shares used by blobs. See non-interactive default rules
 // https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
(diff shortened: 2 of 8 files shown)


rootulp:
💀 https://github.com/celestiaorg/celestia-app/actions/runs/3761251730/attempts/1

evan-forbes:
yeah, we should run the fuzzers locally for a while on main and then on #1147 and see if we get failures on each. naively it makes sense that we would get failures without #1147 after #1137

rootulp:
> naively it makes sense that we would get failures without https://github.com/celestiaorg/celestia-app/pull/1147 after https://github.com/celestiaorg/celestia-app/pull/1137

+1, hoping it gets resolved via #1137

evan-forbes:
I ran tests for a over an hour with no failures locally in #1154 after merging #1147, so I think we're back to no flakes :crossed_fingers:

pkg/shares/non_interactive_defaults.go
```diff
@@ -80,3 +82,16 @@ func roundUpBy(cursor, v int) int {
 		return ((cursor / v) + 1) * v
 	}
 }
+
+// MinSquareSize returns the minimum square size that can contain shareCount
+// number of shares.
+//
+// NOTE: this function was duplicated to avoid import cycles.
```
evan-forbes, line 89: [non-blocking question]
where's the other one?

evan-forbes (commented):
will take a second look tmrw, but LGTM so far

pkg/shares/non_interactive_defaults.go
rootulp: https://github.com/celestiaorg/celestia-app/blob/b65353f456335ed195c67522d4ec546dc3a583fb/x/blob/types/payforblob.go#L178-L182

evan-forbes (approved):
this LGTM, but might want to have @nashqueue take a look just to have a second pair of eyes

I'm also okay with merging this now, and @nashqueue can inspect after, as it would be nice to know if it breaks #1154

rootulp:
sending it 🚀 can address any other PR feedback in a follow-up PR.

app/write_square_test.go
```diff
@@ -75,9 +75,22 @@ func Test_finalizeLayout(t *testing.T) {
...
+				32,
+				// BlobMinSquareSize(20) = 8 so the second blob has to start at
+				// the next multiple of 8 > 34 which is 40. This blob occupies
+				// shares 40 to 48.
+				40,
+				// BlobMinSquareSize(199) = 16 so the third blob has to start at
+				// the next multiple of 16 > 48 which is 64. This blob occupies
+				// shares 64 to 263.
+				64,
+			},
```
nashqueue, line 93: [non-blocking][nit]
```suggestion
			expectedIndexes: []uint32{
				// BlobMinSquareSize(2) = 2 so the first blob has to start at the
				// next multiple of 2 > 32 which is 32. This blob occupies
				// shares 32 to 33.
				32,
				// BlobMinSquareSize(20) = 8 so the second blob has to start at
				// the next multiple of 8 > 34 which is 40. This blob occupies
				// shares 40 to 59
				40,
				// BlobMinSquareSize(199) = 16 so the third blob has to start at
				// the next multiple of 16 > 48 which is 64. This blob occupies
				// shares 64 to 262.
				64,
			},
```
If the blob occupies shares 2 to 4 it would occupy 3 shares in my understanding. Also depends on how you start counting 0 or 1. I started counting at 0 for this example. The last share of the square would be 263 in this case. The last square of the blob would be 262.

nashqueue (commented):
LGTM

app/write_square_test.go
rootulp: Agreed, thanks for pointing this out! This example assumes 0-based indexing. The last index in the ranges are all off by one so I'll update to your suggestion in a follow-up PR.

nashqueue: heads up the second one is off by 11

rootulp: Agreed, it wasn't just off by one. Fixed in https://github.com/celestiaorg/celestia-app/pull/1185

Issue #1138, opened 2022-12-19
Rename wirepayfordata.go file


evan-forbes:
I think this was review feedback, but we forgot to incorporate it when refactoring to remove the malleation step and `WirePayForBlob`. This [file](https://github.com/celestiaorg/celestia-app/blob/main/x/blob/client/cli/wirepayfordata.go) is still name `wirepayforblob`. We should rename it to something else.

rootulp:
Proposal to rename: x/blob/client/cli/wirepayfordata.go => x/blob/client/cli/payforblob.go

This has the downside that it introduces a naming collision so it's harder to <kbd>CMD</kbd>+<kbd>P</kbd> to the file of based on filename:
- x/blob/client/cli/payforblob.go
- x/blob/payforblob.go
- x/blob/types/payforblob.go

evan-forbes:
that works for me, or if we don't want to add to the collisions, then we can simply name it pfb_command, payforblob_command, or I'm even okay with just command. whatever works

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/1288

Issue #1115, opened 2022-12-13
Proposal: make the namespace in PFBs optional


evan-forbes:
The namespace selected by the user is already committed to in share commitments, because it is a commitment over encoded shares. The namespace is also already included in the blob transaction. This means that we don't actually need to include it in the PFB itself. However, teams have already expressed interested in having it included. 

Therefore, we might be able to make it optional. While this wouldn't have that much improvement in various different metrics of efficiency, it is compounded per blob posted. In the future, we could have a single PFB post many different blobs, so removing the namespace might actually be appealing depending on how many different blobs are posted per PFB.

evan-forbes:
see relevant comment to the downside of doing this https://github.com/celestiaorg/celestia-app/pull/1154#discussion_r1063318450

adlerjohn:
If the namespace ID isn't in the transaction then it would require an additional proof attached to each tx to prove that it pays for a blob in a particular namespace.

evan-forbes:
> If the namespace ID isn't in the transaction then it would require an additional proof attached to each tx to prove that it pays for a blob in a particular namespace.

do you mind elaborating more on this? specifically on how this new proof is different from the message inclusion proof

adlerjohn:
It wouldn't be different. But e.g. a partial node only downloads transactions, not blobs.

PR #1154, opened 2022-12-23
feat!: multiple blobs per PFB

### LINKED ISSUE #388: ADR: Pay for multiple blobs with a single `MsgPayForBlob`
Currently, each `MsgPayForData` transaction can only pay for a single message. Ideally, users would be able to pay for multiple messages using a single payment transaction. Not only that, but because this reduces the number of state transitions each full node would have to execute, we would want to incentivize this behavior, by adding some base fee to `PayForData`

### LINKED ISSUE #890: Add the ability to include multiple `sdk.Msg`s in a `sdk.Tx` that contains a PFD
Currently, we limit the number of `sdk.Msg` that are allowed in a viable `sdk.Tx`. This is due in a large part because of the complexity that it would add to the creation of `MsgWirePayForData`s and the malleation process itself. 

Signature validation occurs over the entire `sdk.Tx`, so in order for this feature to work with malleation, we have to replace that signature with one that is valid for the malleated transaction that contains a `MsgPayForData`. 

There could be a few different approaches implementation wise:
- During the malleation process in `PrepareProposal`, split the PFD into its own `sdk.Tx`
- Account for the other `sdk.Msg`s in the transaction when signing commitments in a `MsgWirePayForData`
- Eliminate the malleation process entirely by refactoring commitment verification similar to #835

evan-forbes:
## Overview

This PR implements the ability to add an arbitrary number of blobs to a single PFB. Leaving as a draft until we merge the blocking [PR](https://github.com/celestiaorg/celestia-core/pull/922) in core, and probably tidy up a bit, rebase, or add a few more unit tests.

In part thanks to us planning ahead earlier this year, there's actually not that many required changes to get multiple blobs per PFB. Mostly just adding tests, using a **new mechanism to calculate the share commitments**, making a few things a slice instead of a single thing, and minor adjustments to make square estimation/square layout work with multiple blobs.

closes #388 

## Checklist


evan-forbes:
we should be able to split one or two minor refactors out of this PR

proto/blob/event.proto
```diff
@@ -5,9 +5,9 @@ import "gogoproto/gogo.proto";
 
 option go_package = "github.com/celestiaorg/celestia-app/x/blob/types";
 
-// EventPayForBlob defines an event that is emitted with the signer and blob size
-// at the end of a pay for blob txn
+// EventPayForBlob defines an event that is emitted with the signer and blob
+// size at the end of a pay for blob txn
```
rootulp, line 9:
```suggestion
// EventPayForBlob defines an event that is emitted with the signer and the
// total shares used at the end of a pay for blob txn
```

proto/blob/tx.proto
```diff
@@ -26,16 +26,16 @@ message ShareCommitAndSignature {
...
+  repeated bytes namespace_ids = 2;
+  repeated uint64 blob_sizes = 3;
   // share_commitment is the share_commitment from
   // ShareCommitAndSignature that will be included in a block
   bytes share_commitment = 4;
-  // share_version is the version of the share format that the blob associated
+  // share_versions are the version of the share format that the blob associated
   // with this message should use when included in a block. The share_version
   // specified must match the share_version used to generate the
   // share_commitment in this message.
```
rootulp, line 37:
```suggestion
  // share_versions are the versions of the share format that the blobs associated
  // with this message should use when included in a block. The share_versions
  // specified must match the share_versions used to generate the
  // share_commitment in this message.
```

testutil/blobfactory/payforblob_factory.go
```diff
@@ -223,6 +253,114 @@ func RandBlobTxsWithNamespaces(enc sdk.TxEncoder, nIds [][]byte, sizes []int) []
 	return RandBlobTxsWithNamespacesAndSigner(enc, signer, nIds, sizes)
 }
 
+// ManyMultiBlobBlobTxSameSigner generates and returns many blob transactions with the
```
rootulp, line 256:
```suggestion
// ManyMultiBlobTxSameSigner generates and returns many blob transactions with the
```

x/blob/client/cli/wirepayfordata.go
```diff
@@ -37,14 +37,20 @@ func CmdWirePayForBlob() *cobra.Command {
 			}
 
 			// decode the blob
-			blob, err := hex.DecodeString(args[1])
+			rawBlob, err := hex.DecodeString(args[1])
 			if err != nil {
 				return fmt.Errorf("failure to decode hex blob: %w", err)
 			}
 
+			// TODO: allow for more than one blob to be sumbmitted via the cliw
```
rootulp, line 45:
```suggestion
			// TODO: allow for more than one blob to be submitted via the cli
```

x/blob/types/payforblob.go
```diff
@@ -140,6 +172,57 @@ func CreateCommitment(namespace []byte, blobData []byte, shareVersion uint8) ([]
...
+	}
+
+	for _, blob := range blobs {
+		if len(blob) == 0 {
+			return ErrZeroBlobSize
+		}
+	}
+
+	for _, v := range shareVersions {
+		if uint8(v) != appconsts.ShareVersionZero {
```
rootulp, line 218: [optional] we may want to cast the appconst instead of the shareVersion to avoid an overflow in case a shareVersion > the size of uint8 makes it here

```suggestion
		if v != uint32(appconsts.ShareVersionZero) {
```

rootulp (commented):
Overall looks really good to me! Hyped for this feature

x/blob/client/cli/wirepayfordata.go
cmwaters: Also should we change the name of this file?

cmwaters (commented):
Just some high level comments first

app/process_proposal.go
```diff
@@ -85,51 +83,42 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
 	}
 
-	// compare the number of MPFBs and blobs, if they aren't
-	// identical, then we already know this block is invalid
-	if commitmentCounter != len(req.BlockData.Blobs) {
-		logInvalidPropBlock(app.Logger(), req.Header, "varying number of MsgPayForBlob and blobs in the same block")
-		return abci.ResponseProcessProposal{
-			Result: abci.ResponseProcessProposal_REJECT,
-		}
-	}
```
cmwaters, line 132: Why are we removing this? We should still be counting the amount of blobs that PFBs are paying for.

app/process_proposal.go
```diff
@@ -85,51 +83,42 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
-		for _, msg := range tx.GetMsgs() {
-			if sdk.MsgTypeURL(msg) != types.URLMsgPayForBlob {
-				continue
-			}
+		msgs := tx.GetMsgs()
+		// ignore transactions with more than a single sdk.Msg since we do not
+		// yet support arbitrary sdk.Tx
+		if len(msgs) != 1 {
+			continue
+		}
```
cmwaters, line 91: If it's not possible for a correct node to generate this in `PrepareProposal` then we should outright reject then entire proposal. Same for line 96

x/blob/client/cli/wirepayfordata.go
rootulp: https://github.com/celestiaorg/celestia-app/issues/1138

app/estimate_square_size_test.go
```diff
@@ -41,6 +45,67 @@ func Test_estimateSquareSize(t *testing.T) {
...
+			},
+			8, 7,
+		},
+		{
+			"10 multiblob 4 share transactions",
+			func() [][]int {
+				return blobfactory.Repeat([]int{2000}, 10)
+			},
+			16, 7,
+		},
```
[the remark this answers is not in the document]
evan-forbes: yeah

app/process_proposal.go
evan-forbes: good point, I agree and don't see why we shouldn't change this now. e9a8014d39f49cb1b832d3272ca4f5d8ba46bdc9

I tried to test this, but it would take a ton of code to hit this test, as we'd basically have to write a custom share splitting method just to create an other wise valid block.

pkg/inclusion/get_commit.go
```diff
@@ -4,9 +4,26 @@ import (
...
+	}
+	commitments := make([][]byte, len(startIndexes))
+	for i := range startIndexes {
+		commitment, err := GetCommit(cacher, dah, int(startIndexes[i]), shareCounts[i])
+		if err != nil {
+			return nil, err
+		}
+		commitments[i] = commitment
+	}
+	return merkle.HashFromByteSlices(commitments), nil
```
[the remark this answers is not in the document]
evan-forbes: > I have a question regarding ordering: how does a user know which order to sort their blobs prior to generating this merkle root. Blobs posted to different namespaces are guaranteed to be sorted by namespace, but what about blobs that are posted to the same namespace?

good question, the user can put them in an arbitrary order, and the proposer will match the starting index to the size in pfb.sizes

proto/blob/tx.proto
```diff
@@ -26,16 +26,16 @@ message ShareCommitAndSignature {
 // MsgPayForBlob pays for the inclusion of a blob in the block.
 message MsgPayForBlob {
   string signer = 1;
-  bytes namespace_id = 2;
-  uint64 blob_size = 3;
+  repeated bytes namespace_ids = 2;
```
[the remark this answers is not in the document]
evan-forbes: I actually think there will be a lot of use cases where people post PFBs with many different namespaces, but better yet, we have https://github.com/celestiaorg/celestia-app/issues/1115

proto/blob/tx.proto
```diff
@@ -26,16 +26,16 @@ message ShareCommitAndSignature {
...
   // share_commitment is the share_commitment from
   // ShareCommitAndSignature that will be included in a block
   bytes share_commitment = 4;
-  // share_version is the version of the share format that the blob associated
+  // share_versions are the version of the share format that the blob associated
   // with this message should use when included in a block. The share_version
   // specified must match the share_version used to generate the
   // share_commitment in this message.
-  uint32 share_version = 8;
+  repeated uint32 share_versions = 8;
```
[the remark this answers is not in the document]
evan-forbes: there could maybe be a scenario with multiple share_versions, but we could enforce that if we wanted to. Its probably a good trade-off

x/blob/types/blob_tx.go
```diff
@@ -4,15 +4,15 @@ import (
...
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	shares "github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/celestiaorg/nmt/namespace"
 	"github.com/cosmos/cosmos-sdk/client"
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 )
 
 // ProcessedBlobTx caches the unmarshalled result of the BlobTx
 type ProcessedBlobTx struct {
```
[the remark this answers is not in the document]
evan-forbes: adding that method in that exact way might not work since BlobTx is in core, but we still could replace it with BlobTx I think

evan-forbes (commented):
addressed some feedback here, but I still plan on breaking off one refactor in a meager attempt to reduce the size of the PR.

proto/blob/tx.proto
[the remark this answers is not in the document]
cmwaters: Ok perhaps a better question: What changes would cause us to change the share version?

app/process_proposal.go
cmwaters: > If there is an extra PFB, then the starting positions will also be off.

Are you sure? Don't PFBs use compact shares, couldn't you have multiple PFBs in a single share which means the starting position of the blobs would not be offset?

pkg/inclusion/get_commit.go
cmwaters, line 24: What's the advantage of merkelizing all the commitments over plain hashing

proto/blob/tx.proto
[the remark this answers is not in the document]
evan-forbes: One use case is to add the sender of the PFB, or other metadata, before the blob in the square. 

Other usecases would be to allow posters to not use varints or some other encoding should it be useful. 

I'm not sure if people will ever want to mix those tho. Even if they do, then only in that specific case do we have more linear increase in state transitions per different share version.

With the benefit of removing a lot of redundant bytes. Which sounds like a really good tradeoff.

app/process_proposal.go
evan-forbes: ahh, I see. I didn't word that well, by starting positions I meant the starting index of the blob that wraps the PFB.

If there's an extra PFB, then we check that the share commitment is included using the starting position that wraps it. there's no way to fake that without actually including the blob, so we're gucci

pkg/inclusion/get_commit.go
evan-forbes: good question, so if we have many blobs being paid for by a PFB, then there might be some scenario where you want to prove that someone paid for a specific blob without downloading the entirely blob. other than that, I'm not really sure. Using a hash would probably work fine

proto/blob/tx.proto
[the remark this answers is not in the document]
cmwaters: Yeah I don't think that constraining the share versions for blobs within a single PFB is going to be a problem for users. Worst case they just have two PFBs (for each version) and pay the extra 2 cents in gas.

proto/blob/tx.proto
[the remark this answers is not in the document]
cmwaters: The only problem with not encoding namespace is it means we have events without namespaces, which could be annoying for indexers. Say I want to query the most popular namespaces - this now becomes super difficult.

x/blob/types/blob_tx.go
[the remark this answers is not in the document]
cmwaters: Ah yeah then simply `func ValidateBlobTx(tx tmproto.BlobTx, txcfg client.TxEncodingConfig) error`

pkg/inclusion/get_commit.go
cmwaters: > want to prove that someone paid for a specific blob without downloading the entirely blob

Won't you still need the blob or at least the commitment for that blob? 

(I'm generally a bit hesitant about merkelizing something without a strong use case. For example, tendermint merkelizes the header which I always felt was pointless. The headers is like 400 bytes. Just download the entire thing)

app/estimate_square_size_test.go
```diff
@@ -41,6 +45,67 @@ func Test_estimateSquareSize(t *testing.T) {
...
+		name                       string
+		getBlobSizes               func() [][]int
+		expectedSquareSize         uint64
+		expectedStartingShareIndex int
+	}
+	tests := []test{
+		{
+			"single multiblob transaction",
+			func() [][]int { return [][]int{{4}} },
+			2, 1,
```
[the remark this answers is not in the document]
evan-forbes: ahh I see, this threw me off after you mentioned it too. This only takes a single share as its blobsize, not the number of shares that are used. so we have single single share tx, and therefore

```
|C|B|
|_|_|
```

go.mod
```diff
@@ -186,5 +186,5 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.5.1-sdk-v0.46.6
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.12.1-tm-v0.34.23
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.12.1-tm-v0.34.23.0.20230102172645-b0a2ff008326 // v1.12.1-tm-v0.34.23
```
evan-forbes, line 189: still need to update after we create a new release of core

proto/blob/tx.proto
```diff
@@ -26,16 +26,16 @@ message ShareCommitAndSignature {
 // MsgPayForBlob pays for the inclusion of a blob in the block.
 message MsgPayForBlob {
```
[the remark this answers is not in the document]
evan-forbes: Definitely!! started to change this in this PR, but since this PR is already large, decided to spin it off as a follow up. https://github.com/celestiaorg/celestia-app/issues/1221

proto/blob/tx.proto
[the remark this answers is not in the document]
evan-forbes: left a comment in #1115 to link to this comment as its a good point

proto/blob/tx.proto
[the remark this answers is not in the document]
evan-forbes: left an issue in https://github.com/celestiaorg/celestia-app/issues/1223

x/blob/types/blob_tx.go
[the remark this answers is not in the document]
evan-forbes: created a followup issue for this https://github.com/celestiaorg/celestia-app/issues/1224

evan-forbes (commented):
FINALLY ready for review, but still technically blocked by a new release of core

pkg/inclusion/get_commit.go
evan-forbes: another good point.

it would be good to get @nashqueue 's point of view on this as well for inclusion fraud proofs. We can discuss synchronously if necessary, as we should probably make a decision soon. I'm definitely fine with changing this to simply include all of the share commitments in the PFB instead of using a secondary commitment over all of them.

evan-forbes: I'm fairly certain that we would have to end up downloading all of the share commitments for a single PFB if we're attempting to prove inclusion for a single one in both of the other options, those being taking a hash over all of the commitments, or including all commitments in the PFB.

evan-forbes: If there is ever a scenario where a single PFB pays for 1000 blobs, then I could see the secondary commitment being useful, until then, probably not.

nashqueue: I think you make a tradeoff between fraud-proof size and DA bytes. In a way, you can see it if you have multiple commitments, you only need to show that one of them is broken. If you have multiple, you need all blobs to compute the merkle hash and compare that it is wrong. If there is no limit to PFBs per Tx then the worst case fraud proof size for a single commitment fraud-proof would be O(n), with n being the number of shares in a square. This is the case if you have 1 share-sized blob per Tx. So, therefore, I would say saving all commitments is the better option here.

evan-forbes: sounds good, that's what we'll do then

app/estimate_square_size_test.go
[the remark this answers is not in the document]
rootulp: ohh I see, blobSizes is the # of bytes in the blob. [No change needed] then I expect this to behave similarly and indeed it does

```suggestion
			func() [][]int { return [][]int{{1}} },
			2, 1,
```

app/test/block_size_test.go
```diff
@@ -154,8 +177,11 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
 			for _, hash := range hashes {
 				// TODO: reenable fetching and verifying proofs
 				resp, err := queryTx(val.ClientCtx, hash, false)
-				require.NoError(err)
-				require.NotNil(resp)
+				assert.NoError(err)
+				assert.NotNil(resp)
+				if resp == nil {
+					continue
+				}
```
rootulp, line 184: [question] is this conditional ever execution? If `resp == nil` were true, I would expect the assertion on the line above to fail.

proto/blob/event.proto
```diff
@@ -5,9 +5,9 @@ import "gogoproto/gogo.proto";
...
 message EventPayForBlob {
```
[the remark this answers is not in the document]
rootulp: I think this conversation is resolved by https://github.com/celestiaorg/celestia-app/issues/1169

x/blob/keeper/gas_test.go
```diff
@@ -54,21 +55,33 @@ func TestPayForBlobGas(t *testing.T) {
...
+		},
+		{
+			name:            "3 blobs, 1 share each",
+			msg:             types.MsgPayForBlob{BlobSizes: []uint32{1, 1, 1}},
+			wantGasConsumed: uint64(3*appconsts.ShareSize*types.DefaultGasPerBlobByte + paramLookUpCost), // 3 shares * 512 bytes per share * 8 gas per byte + 1060 gas for fetching param = 13348 gas
+		},
+		{
+			name:            "3 blobs, 6 shares total",
+			msg:             types.MsgPayForBlob{BlobSizes: []uint32{1024, 1000, 100}},
+			wantGasConsumed: uint64(6*appconsts.ShareSize*types.DefaultGasPerBlobByte + paramLookUpCost), // 6 shares * 512 bytes per share * 8 gas per byte + 1060 gas for fetching param = 25636 gas
```
rootulp, line 84: 👍

x/blob/types/blob_tx.go
```diff
@@ -58,53 +59,60 @@ func ProcessBlobTx(txcfg client.TxEncodingConfig, bTx tmproto.BlobTx) (Processed
...
 	}
 
-	if pfb.BlobSize != uint32(len(blob.Data)) {
-		return ProcessedBlobTx{}, ErrDeclaredActualDataSizeMismatch.Wrapf(
-			"declared: %d vs actual: %d",
-			pfb.BlobSize,
-			len(blob.Data),
-		)
+	for i := range pfb.NamespaceIds {
+		// check that the meta data matches
```
rootulp, line 79: [nit]
```suggestion
		// check that the metadata matches
```

x/blob/types/payforblob.go
```diff
@@ -156,6 +179,41 @@ func CreateMultiShareCommitment(blobs ...*Blob) ([]byte, error) {
...
+		if len(blob) == 0 {
+			return ErrZeroBlobSize
+		}
+	}
+
+	for _, v := range shareVersions {
+		if v != uint32(appconsts.ShareVersionZero) {
+			return ErrUnsupportedShareVersion
+		}
+	}
```
rootulp, line 212: [super optional] future proofing for when we add a new share version

```suggestion
	for _, shareVersion := range shareVersions {
		if !slices.Contains(appconsts.SupportedShareVersions, uint8(shareVersion)) {
			return ErrUnsupportedShareVersion
		}
	}```

**Review by rootulp — DISMISSED** [review]
Great work! Let's :shipit:

File: x/blob/types/payforblob.go (hunk 20 above)
Comment (evan-forbes) [reply x/blob/types/payforblob.go:212]: noice

File: app/test/block_size_test.go (hunk 16 above)
Comment (evan-forbes) [reply app/test/block_size_test.go:184]: assert fails, but doesn't stop the test, so when we try to access the resp below it will panic

File: pkg/inclusion/get_commit.go (hunk 9 above)
Comment (evan-forbes) [reply pkg/inclusion/get_commit.go]: left an issue to handle after this is merged https://github.com/celestiaorg/celestia-app/issues/1231

**Comment by evan-forbes** [comment]
there are still a lot of follow ups to this PR, but if there is no more feedback, I'll try to merge this PR monday

File: x/blob/types/payforblob_test.go [hunk 21]
```diff
@@ -336,3 +336,27 @@ func TestBlobMinSquareSize(t *testing.T) {
...
+		name        string
+		blob        *Blob
+		expectError bool
+	}
+
+	tests := []test{
+		{name: "valid blob", blob: &Blob{Data: []byte{1}, NamespaceId: []byte{1, 2, 3, 4, 5, 6, 7, 8}, ShareVersion: uint32(appconsts.DefaultShareVersion)}, expectError: false},
+		{name: "invalid share version", blob: &Blob{Data: []byte{1}, NamespaceId: []byte{1, 2, 3, 4, 5, 6, 7, 8}, ShareVersion: uint32(10000)}, expectError: true},
+		{name: "empty blob", blob: &Blob{Data: []byte{}, NamespaceId: []byte{1, 2, 3, 4, 5, 6, 7, 8}, ShareVersion: uint32(appconsts.DefaultShareVersion)}, expectError: true},
+		{name: "valid blob", blob: &Blob{Data: []byte{1}, NamespaceId: appconsts.TxNamespaceID, ShareVersion: uint32(appconsts.DefaultShareVersion)}, expectError: true},
```
rootulp, line 351:
```suggestion
		{name: "invalid namespace", blob: &Blob{Data: []byte{1}, NamespaceId: appconsts.TxNamespaceID, ShareVersion: uint32(appconsts.DefaultShareVersion)}, expectError: true},
```

rootulp (approved):

rootulp (approved):

Issue #1151, opened 2022-12-22
specs: non-interactive default rules for reduced padding


rootulp:
[ADR 009: Non-Interactive Default Rules for Reduced Padding](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md) reached accepted status so we are implementing it. We should update the specs-staging branch with the new non-interactive default rules.

PR #1156, opened 2022-12-23
specs: non-interactive default rules for reduced padding

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1151

### DIFF
--- a/specs/src/rationale/message_block_layout.md
+++ b/specs/src/rationale/message_block_layout.md
@@ -37,7 +37,7 @@ This, however, requires the block producer to interact with the transaction send
 
 As a non-consensus-critical default, we can impose one additional rule on message placement to make the possible starting locations of messages sufficiently predictable and constrained such that users can deterministically compute subtree roots without interaction:
 
-> Messages start at an index that is a multiple of the message minimum square size. The message minimum square size is the smallest square that can contain the message.
+> Messages start at an index that is a multiple of the message minimum square size. The message minimum square size is the smallest square that can contain the message in isolation (i.e. a square with only this message and no other transactions or messages).
 
 With the above constraint, we can compute subtree roots deterministically. In order to compute the subtree roots, split the message into chunks that are of maximum size: message minimum square size. As an example, a message of length `11` has a minimum square size of `4` because `11` is not greater than `4 * 4 = 16` total shares. Split the message into chunks of length `4, 4, 2, 1`. The resulting slices are the leaves of subtrees whose roots can be computed. These subtree roots will be present as internal nodes in the NMT of _some_ row(s).
 
--- a/specs/src/rationale/message_block_layout.md
+++ b/specs/src/rationale/message_block_layout.md
@@ -39,7 +39,7 @@ As a non-consensus-critical default, we can impose one additional rule on messag
 
 > Messages start at an index that is a multiple of the message minimum square size. The message minimum square size is the smallest square that can contain the message.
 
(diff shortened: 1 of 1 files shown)


MSevey (approved):

nashqueue (dismissed):

specs/src/rationale/message_block_layout.md
```diff
@@ -35,14 +35,13 @@ This, however, requires the block producer to interact with the transaction send
...
 
-With the above constraints, we can compute subtree roots deterministically easily: simply slice off either the largest power of 2 that isn't larger than the remaining message length, or `k`, whichever is smaller. This is done recursively. As an example, with `k = 4` and message length of `11`, the message would be sliced with lengths `4, 4, 2, 1`. The resulting slices are the leaves of subtrees whose roots can be computed. Due to the rules above, the subtrees are guaranteed to be aligned to powers of 2, and thus the subtree roots will be present as internal nodes in the NMT of _some_ row(s).
+With the above constraint, we can compute subtree roots deterministically. In order to compute the subtree roots, split the message into chunks that are of maximum size: message minimum square size. As an example, a message of length `11` has a minimum square size of `4` because `11` is less than `4 * 4 = 16` total shares. Split the message into chunks of length `4, 4, 2, 1`. The resulting slices are the leaves of subtrees whose roots can be computed. These subtree roots will be present as internal nodes in the NMT of _some_ row(s).
```
adlerjohn, line 42:
```suggestion
With the above constraint, we can compute subtree roots deterministically. In order to compute the subtree roots, split the message into chunks that are of maximum size: message minimum square size. As an example, a message of length `11` has a minimum square size of `4` because `11` is not larger than `4 * 4 = 16` total shares. Split the message into chunks of length `4, 4, 2, 1`. The resulting slices are the leaves of subtrees whose roots can be computed. These subtree roots will be present as internal nodes in the NMT of _some_ row(s).
```

specs/src/rationale/message_block_layout.md
```diff
@@ -35,14 +35,13 @@ This, however, requires the block producer to interact with the transaction send
 
 ### Non-Interactive Default Rules
 
-As a non-consensus-critical default, we can impose some additional rules on message placement to make the possible starting locations of messages sufficiently predictable and constrained such that users can deterministically compute subtree roots without interaction:
+As a non-consensus-critical default, we can impose one additional rule on message placement to make the possible starting locations of messages sufficiently predictable and constrained such that users can deterministically compute subtree roots without interaction:
 
-1. Messages that span multiple rows must begin at the start of a row (this can occur if a message is longer than `k` shares _or_ if the block producer decides to start a message partway through a row and it cannot fit).
-1. Messages begin at a location aligned with the largest power of 2 that is not larger than the message length or `k`.
+> Messages start at an index that is a multiple of the message minimum square size. The message minimum square size is the smallest square that can contain the message.
```
adlerjohn, line 40: Can you add a line that "contain" here ignores other transactions and messages, and only considers the message in isolation? Something like that.

rootulp: Yes absolutely. How about

> Messages start at an index that is a multiple of the message minimum square size. The message minimum square size is the smallest square that can contain the message in isolation (i.e. a square with only this message and no other transactions or messages).

adlerjohn: Yeah sure.

rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/1156/commits/8a970ab656b740b8cf67878fff9a1d6940efa33d

nashqueue (approved):

adlerjohn (approved):

PR #1159, opened 2023-01-03
fix!: pass the baseapp msg service router

evan-forbes:
## Overview

as noticed by @cmwaters, we accidentally removed setting the app's msg service router. This is why this functionality was present in v0.6.0, but not in v0.10.0. The only usage of the msg service router is for the gov keeper, so it doesn't have an effect on other modules. 

this fix is tested in a followup PR #1160 

closes #1158

PR #1160, opened 2023-01-03
chore!: Test std modules

evan-forbes:
## Overview

Adds a smoke test for some of the std cosmos-sdk modules. This test does not test that the state was changed as expected, only that the msgs included in the transactions were applied successfully. 

blocked by #1159, as the test for submitting a governance proposal currently fails until that is merged

## Checklist

### DIFF
--- a/app/test/std_sdk_test.go
+++ b/app/test/std_sdk_test.go
@@ -26,6 +26,9 @@ import (
 )
 
 func TestStandardSDKIntegrationTestSuite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in unit-tests or race-detector mode.")
+	}
 	suite.Run(t, new(StandardSDKIntegrationTestSuite))
 }
 
@@ -43,12 +46,7 @@ type StandardSDKIntegrationTestSuite struct {
 
 func (s *StandardSDKIntegrationTestSuite) SetupSuite() {
 	t := s.T()
-	if testing.Short() {
-		t.Skip("skipping test in unit-tests or race-detector mode.")
-	}
-
 	t.Log("setting up integration test suite")
-
 	cleanup, accounts, cctx := testnode.DefaultNetwork(t, time.Millisecond*400)
 	s.cleanup = cleanup
 	s.accounts = accounts
@@ -91,8 +89,6 @@ func (s *StandardSDKIntegrationTestSuite) TestStandardSDK() {
 			},
 		},
 		{
-			// demonstrate that a send transactions use different amounts of gas
-			// depending on the number of funds
 			name: "send 1,000,000 TIA",
 			msgFunc: func() (msg sdk.Msg, signer string) {
 				account1, account2 := s.unusedAccount(), s.unusedAccount()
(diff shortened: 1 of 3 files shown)


x/blob/types/builder.go
```diff
@@ -202,6 +202,12 @@ func (k *KeyringSigner) SetKeyringAccName(name string) {
 	k.keyringAccName = name
 }
 
+// SetEncodingConfig manually overides the encoding config of the underlying
+// signer
+func (k *KeyringSigner) SetEncodingConfig(encCfg encoding.Config) {
+	k.encCfg = encCfg
+}
```
evan-forbes, line 209: kinda hacky. the entire keyring signer can be refactored or removed entirely now that we don't have to malleate txs, as its simply a script to programatically sign things. We still use it a lot in tests, and I think celestia-node uses it when creating a PFB, but we could probably replace those with a simple function instead.

https://github.com/celestiaorg/celestia-app/issues/1162

SweeXordious (approved):

testutil/testnode/sign.go
```diff
@@ -0,0 +1,56 @@
...
+// inside the client.Context, then broadcasts it synchronously.
+func SignAndBroadcastTx(encCfg encoding.Config, c client.Context, account string, msg ...sdk.Msg) (res *sdk.TxResponse, err error) {
+	opts := []types.TxBuilderOption{
+		types.SetGasLimit(veryLargeGasLim),
+		types.SetFeeAmount(sdk.NewCoins(
+			sdk.NewCoin(app.BondDenom, sdk.NewInt(1)),
+		)),
+	}
+
+	// use the key for accounts[i] to create a singer used for a single PFB
```
rootulp, line 26:
```suggestion
	// use the key for accounts[i] to create a signer used for a single PFB
```

app/test/std_sdk_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+func TestStandardSDKIntegrationTestSuite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in unit-tests or race-detector mode.")
```
rootulp, line 29: [no change needed]

[`testing.Short()`](https://pkg.go.dev/testing#Short) will only return true when the `-short` flag is provided so this conditional won't be executed when `-race` is provided. However, I think this line was written with  awareness that https://github.com/celestiaorg/celestia-app/blob/ec5498c8cf3dc075b54182a986db72cdfd32eecb/Makefile#L89 only uses the `-race` flag when `-short` is also provided so this skip description seems fine. An alternative description may be:

```suggestion
		t.Skip("skipping SDK integration test in short mode.")
```

rootulp (dismissed):

rootulp (approved):

PR #1161, opened 2023-01-03
inital draft

nashqueue:
## Checklist


docs/architecture/adr012-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,56 @@
+# ADR 012: Non-Interactive Default Rules for Zero Padding
```
rootulp, line 1:
```suggestion
# ADR 013: Non-Interactive Default Rules for Zero Padding
```

because 012 has been taken. Ref: https://github.com/celestiaorg/celestia-app/tree/main/docs/architecture

docs/architecture/adr012-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Changelog
+
+- 03.01.12: initial draft
```
rootulp, line 5: I'm not sure which date format this is but I think the year is wrong.

```suggestion
- 03.01.2023: initial draft
```

docs/architecture/adr012-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,56 @@
...
+Assuming there is a threshold, the number of subtree roots in a proof, where the proof size of a blob is acceptable we can use this threshold to determine the index of the blob in the square. This would give us zero padding for blobs that are smaller than the threshold and non-zero padding for blobs that are larger than the threshold but still smaller than before.
+
+Let's assume a good threshold assumption is that the number of subtree roots in a blob inclusion proof is acceptable if it is smaller than the `MaxSquareSize`. This would mean that the blobs smaller than the square size can use the index of one to get zero padding. Blobs that are larger than `MaxSquareSize` but smaller than `MaxSquareSize * 2` can use the index of two to get a maximum of 1 padding square. Blobs that are larger than `MaxSquareSize * 2` but smaller than `MaxSquareSize * 4` can use the index of 4 to get a maximum of 2 padding squares and so on.
```
rootulp, line 14: [typo]

```suggestion
Let's assume a good threshold assumption is that the number of subtree roots in a blob inclusion proof is acceptable if it is smaller than the `MaxSquareSize`. This would mean that the blobs smaller than the square size can use the index of one to get zero padding. Blobs that are larger than `MaxSquareSize` but smaller than `MaxSquareSize * 2` can use the index of two to get a maximum of 1 padding square. Blobs that are larger than `MaxSquareSize * 2` but smaller than `MaxSquareSize * 4` can use the index of 4 to get a maximum of 2 padding shares and so on.
```

[question] if a blob must be aligned to an index of 4, I think the maximum number of padding shares is 3.

```ascii
| B1 | P | P | P | B2 | B2 | B2 | B2 | B2 | B2 | ...

B1: blob one
P: padding
B2: blob two
```

Can you please elaborate on why there is a maximum of 2 padding shares?

docs/architecture/adr012-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,56 @@
...
+
+Partial nodes in this context are Celestia-node light nodes that may download all of the data in the reserved namespace. They check that the data behind the PFB was included in the `DataRoot`, via blob inclusion proofs.
+
+The sum of the size of all blob inclusion proofs will be larger than the sum of all blob inclusion proofs with the previous non-interactive default rules.
+
+In the current worst case a Celestia block is full of blobs of size 1, making the total amount of subtree roots that need to be downloaded O(n) where n is the number of shares in a block. With the new non-interactive default rules, the worst case stays the same, but the average case goes up. If the block is filled with blobs that are smaller than the threshold then the partial node will still need to download O(n) subtree roots.
+
+### Positive
+
+Most Blocks will have close to zero padding.
```
rootulp, line 48: [nit]

```suggestion
Most blocks will have close to zero padding.
```

docs/architecture/adr012-non-interactive-default-rules-for-zero-padding.md
nashqueue: Yes you a right that was a typo :)

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Context
+
+When laying out blobs in the square we create padding to follow the non-interactive default rules. With ADR 009 we decreased padding by aligning the blob in the square to the index of the `MinSquareSize` of the given blob. This is a good improvement but we can do better.
+
+Looking at different ranges of blob sizes we can see that the ratio of blob size to padding is not constant. Insight:  **The ratio of blob size to padding is smaller for smaller blobs and larger for larger blobs.** The bigger the ratio the better.
+
+![Worst Case Padding In Blob Size Range](./assets/worst-case-padding-in-blob-size-range.png)
+
+This means that you get more padding for fever data in small blobs. This is not ideal as we want to have as little padding as possible. As padding is not being paid for there is no incentive to use larger blobs.
```
rootulp, line 15: [typo]: "for fever data"

Apologies for no suggestion. I'm not sure what it meant to say.

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+
+In the naive approach if you align the blob at an index of one you get zero padding as each blob can follow the next. This would make the hash of each share a subtree root. In a square with n shares, you would always get n subtree roots. But having a blob inclusion proof of size n for large blobs is too much and unfeasible.
```
rootulp, line 17: [nit] capitalize N for readability

```suggestion
In the naive approach if you align the blob at an index of one you get zero padding as each blob can follow the next. This would make the hash of each share a subtree root. In a square with N shares, you would always get N subtree roots. But having a blob inclusion proof of size N for large blobs is too much and unfeasible.
```

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+Let's assume a good threshold assumption is that the number of subtree roots in a blob inclusion proof is acceptable if it is smaller than the `MaxSquareSize`. This would mean that the blobs smaller than the square size can use the index of one to get zero padding. Blobs that are larger than `MaxSquareSize` but smaller than `MaxSquareSize * 2` can use the index of two to get a maximum of 1 padding square. Blobs that are larger than `MaxSquareSize * 2` but smaller than `MaxSquareSize * 4` can use the index of 4 to get a maximum of 3 padding shares and so on.
+
+The new non-interactive default rules would be:
+
+Blobs start at an index that is equal to the blob length/`MaxSquareSize` rounded up.
+
+If the blob length is smaller than `MaxSquareSize` then the blob starts at index 1.
+`MaxSquareSize` can be changed to another threshold. The smaller the threshold the more padding we will have.
+
+The Picture below shows the difference between the old and new non-interactive default rules in a square of size 8 and a thereshold of 8.
```
rootulp, line 32:
```suggestion
The picture below shows the difference between the old and new non-interactive default rules in a square of size 8 and a threshold of 8.
```

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+
+![Blob Alignment Comparison](./assets/blob-alignment-comparison.png)
+
+## Analysis
+
+### Light-Nodes
+
+The Proof size is bounded by the number of subtree roots in the blob inclusion proof. If the new bound is the `MaxSquareSize` then the worst case for the number of subtree roots in a blob inclusion proof will be `MaxSquareSize`.
+
+If Light-Nodes can process this proof size without a problem then we can use this bound. If not we can use a smaller bound. The smaller the bound the more padding we will have.
```
rootulp, line 42:
```suggestion
If Light-Nodes can process this proof size without a problem then we can use this bound. If not, we can use a smaller bound. The smaller the bound the more padding we will have.
```

rootulp (commented):
mostly nits, thanks for the diagrams!

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
nashqueue: I rephrased

nashqueue:
Not sure why this housekeeping workflow is breaking

MSevey:
> Not sure why this housekeeping workflow is breaking

seems like a bug with forks. I'm working on it. https://github.com/celestiaorg/.github/issues/19

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+
+Assuming there is a threshold, **the number of subtree roots in a proof**, where the proof size of a blob is acceptable, we can use this threshold to determine the index of the blob in the square. This would give us zero padding for blobs that are smaller than the threshold and non-zero padding for blobs that are larger than the threshold but still smaller than before.
+
+Let's assume a good threshold assumption is that the number of subtree roots in a blob inclusion proof is acceptable if it is smaller than the `MaxSquareSize`. This would mean that the blobs smaller than the square size can use the index of one to get zero padding. Blobs that are larger than `MaxSquareSize` but smaller than `MaxSquareSize * 2` can use the index of two to get a maximum of 1 padding square. Blobs that are larger than `MaxSquareSize * 2` but smaller than `MaxSquareSize * 4` can use the index of 4 to get a maximum of 3 padding shares and so on.
+
+The new non-interactive default rules would be:
+
+Blobs start at an index that is equal to the blob length/`MaxSquareSize` rounded up.
```
cmwaters, line 27: What is the index? Is it the offset from where the previous blob was at?

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+
+Small blob sizes have the lowest ratio but also small inclusion proofs. Therefore increasing the proof size is not a problem until some threshold. It would increase the ratio of blob size to padding for small blobs which have the worst ratio.
+
+Assuming there is a threshold, **the number of subtree roots in a proof**, where the proof size of a blob is acceptable, we can use this threshold to determine the index of the blob in the square. This would give us zero padding for blobs that are smaller than the threshold and non-zero padding for blobs that are larger than the threshold but still smaller than before.
+
+Let's assume a good threshold assumption is that the number of subtree roots in a blob inclusion proof is acceptable if it is smaller than the `MaxSquareSize`. This would mean that the blobs smaller than the square size can use the index of one to get zero padding. Blobs that are larger than `MaxSquareSize` but smaller than `MaxSquareSize * 2` can use the index of two to get a maximum of 1 padding square. Blobs that are larger than `MaxSquareSize * 2` but smaller than `MaxSquareSize * 4` can use the index of 4 to get a maximum of 3 padding shares and so on.
```
cmwaters, line 23: Just to confirm, `MaxSquareSize` in the default case is 128, right?

evan-forbes (commented):
Overall I really like this proposal!! It makes a lot of sense.

Soon we will have multiple blobs per a single PFB. We might want to also want to include some discussion of how this proposal would affect those as well. If there are many small blobs, then might need to use the same threshold mechanism used for large blocks.

[kinda orthogonal discussion]
I believe there was some discussion during the onsite of incorporating resource pricing into a similar proposal, and it might work here as well. Provided we have precise insight into the actual cost to the network each of these options bring, then we can change the cost of gas accordingly. For example, if your blob can have a suitably small inclusion proof while also not taking any padding, then we might want to decrease the amount of gas used for that blob.

nashqueue:
> Soon we will have multiple blobs per a single PFB.

What are the specs/ideas for ordering those blobs? What does the commitment of this PFB look like?

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
nashqueue: No, index, in this case, is as we looked at it from the previous adrs and in the code. There should be a `multiple of` in this sentence.

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
nashqueue: yes I think currently MaxSquareSize is 128

evan-forbes:
> What does the commitment of this PFB look like?

Currently, the PR has a merkle tree over all of the commitment, but as discussed in #1154 and https://github.com/celestiaorg/celestia-app/issues/1231, we will revert that and simply include each share commitment in the PFB. Originally we decided to move writing the ADR until after the feature freeze, but we might have to start it before that, as this topic requires more thought and buy-in

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
```diff
@@ -0,0 +1,86 @@
...
+The new non-interactive default rules would be:
+
+Blobs start at an index that is equal to a multiple of the blob length divided by `MaxSquareSize` rounded up.
+
+If the blob length is smaller than `MaxSquareSize` then the blob starts at index 1.
+`MaxSquareSize` can be changed to another threshold. The smaller the threshold the more padding we will have.
+
+The picture below shows the difference between the old and new non-interactive default rules in a square of size 8 and a threshold of 8.
+
+![Blob Alignment Comparison](./assets/blob-alignment-comparison.png)
```
MSevey, line 34: I might be missing something but can you elaborate on how this works with the commitments that are independent of square size?  

For example, the orange blob would appear to me to have different commitments in the two examples, and so I'd assume it would have a different commitment based on what other data is include in the block. 

Or is the intent of this ADR only to focus on reducing padding to the absolute minimum, and not necessarily have padding independent commitments.

evan-forbes (approved):
continuing from https://github.com/celestiaorg/celestia-app/pull/1161#pullrequestreview-1241337684

I think the remaining question is that if there are use cases where we need to verify many blob inclusion proofs, which afaiu is the only case where this would have an effect. State fraud proof producing partial nodes might be that case, where we are increasing the average size for all of the proofs that they have to download, but perhaps they could rely on other nodes to produce fraud message inclusion fraud proofs.

Despite that, I think this would make a great addition and makes a lot of sense! I will defer to other reviewers though, and am looking forward to other's opinions

docs/architecture/adr013-non-interactive-default-rules-for-zero-padding.md
nashqueue: It is conseues breaking as it changes the non-interactive default rules. The 2 pictures have 2 different NI-Rules. One before and one after. If you decide on one NI rule it can be either independent of square size or not. In this case both are independent of square size. My goal was to reduce padding and it just happened to almost eliminate padding.

nashqueue:
Can we merge this as proposed? been idle for a while

evan-forbes:
@nashqueue  yeah we can imo. I'd still like to get an approved tbh, as I think its a really good optimization

evan-forbes (approved):

rootulp (approved):
Great work @nashqueue ! Really informative ADR

musalbas:
Btw I think the `status` field in these ADRs should be at the top so people can immediately see if they're proposed or accepted etc instead of only realizing it when they get to the bottom.

evan-forbes:
merging as proposed

evan-forbes:
merging as proposed, but we will have a sync discussion over this during the onsite

Issue #1166, opened 2023-01-04
Consider refactoring our CheckTx overwrite to be an antehandler


evan-forbes:
We could write our wrapper around CheckTx as an antehandler, which would likely be more of a standard thing to do.

rootulp:
Here's a link to the [CheckTx](https://github.com/celestiaorg/celestia-app/blob/4f5dc9151b38d800b6e1cbf5838c1d54ff676dc2/app/check_tx.go#L12-L15) referenced in the OP.

PR #1167, opened 2023-01-04
fix: reject txs from the mempool that have a PFB but not a blob

### LINKED ISSUE #1165: Reject PFB transactions from the mempool without blobs attached
We don't actually test for or enforce that a PFB transaction must have a blob attached when being submitted to the mempool. This will result in a block producer including a PFB without a blob!

related to #1164

evan-forbes:
## Overview

fixes two bugs, one where we were creating PFD transactions without actually wrapping them with a blob :face_with_spiral_eyes:, and then another bug where we should have been checking for that during checkTx. 

closes #1165 
blocking, and related to #1168 

## Checklist

### DIFF
--- a/app/check_tx.go
+++ b/app/check_tx.go
@@ -3,7 +3,6 @@ package app
 import (
 	"fmt"
 
-	"github.com/celestiaorg/celestia-app/x/blob/types"
 	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	abci "github.com/tendermint/tendermint/abci/types"
@@ -37,7 +36,7 @@ func (app *App) CheckTx(req abci.RequestCheckTx) abci.ResponseCheckTx {
 	switch req.Type {
 	// new transactions must be checked in their entirety
 	case abci.CheckTxType_New:
-		pBTx, err := types.ProcessBlobTx(app.txConfig, btx)
+		pBTx, err := blobtypes.ProcessBlobTx(app.txConfig, btx)
 		if err != nil {
 			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false)
 		}
--- a/app/check_tx.go
+++ b/app/check_tx.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/celestiaorg/celestia-app/x/blob/types"
+	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	abci "github.com/tendermint/tendermint/abci/types"
 	coretypes "github.com/tendermint/tendermint/types"
@@ -17,8 +18,19 @@ func (app *App) CheckTx(req abci.RequestCheckTx) abci.ResponseCheckTx {
 	// check if the transaction contains blobs
 	btx, isBlob := coretypes.UnmarshalBlobTx(tx)
 
-	// don't do anything special if we have a normal transactions
 	if !isBlob {
+		sdkTx, err := app.txConfig.TxDecoder()(tx)
+		if err != nil {
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false)
+		}
+		// reject transactions that have PFBs, but no blobs attached
+		for _, msg := range sdkTx.GetMsgs() {
+			if _, ok := msg.(*blobtypes.MsgPayForBlob); !ok {
+				continue
+			}
+			return sdkerrors.ResponseCheckTxWithEvents(blobtypes.ErrBloblessPFB, 0, 0, []abci.Event{}, false)
+		}
+		// don't do anything special if we have a normal transaction
 		return app.BaseApp.CheckTx(req)
 	}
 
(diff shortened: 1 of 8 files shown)


app/test/block_size_test.go
```diff
@@ -247,6 +249,28 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForBlob() {
 	}
 }
 
+func (s *IntegrationTestSuite) TestUnmalleatedPFBRejection() {
```
evan-forbes, line 252: will change name in a second

evan-forbes (commented):
we should probably turn this entire thing into an antehandler #1166

x/blob/client/cli/wirepayfordata.go
```diff
@@ -54,11 +58,89 @@ func CmdWirePayForBlob() *cobra.Command {
...
 	return cmd
 }
+
+// writeTx attempts to generate and sign a transaction using the normal
+// cosmos-sdk cli argument parsing code with the given set of messages. It will also simulate gas
+// requirements if necessary. It will return an error upon failure.
+//
+// NOTE: Copy paste forked from the cosmos-sdk so that we can wrap the PFB with
+// a blob while still using all of the normal cli parsing code
+func writeTx(clientCtx client.Context, txf sdktx.Factory, msgs ...sdk.Msg) ([]byte, error) {
```
evan-forbes, line 92: had to copy and paste this from the cosmos-sdk so that we can wrap the tx with a blob before broadcasting it.

x/blob/client/testutil/integration_test.go
```diff
@@ -123,12 +124,9 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForBlob() {
...
 
 			// attempt to query for the malleated transaction using the original tx's hash
-			qTxCmd := authcmd.QueryTxCmd()
-			out, err = clitestutil.ExecTestCLICmd(clientCtx, qTxCmd, []string{txResp.TxHash, "--output=json"})
+			res, err := testfactory.QueryWithoutProof(clientCtx, txResp.TxHash)
 			require.NoError(err)
-
-			var result sdk.TxResponse
-			s.Require().NoError(val.ClientCtx.Codec.UnmarshalJSON(out.Bytes(), &result))
+			require.Equal(abci.CodeTypeOK, res.TxResult.Code)
```
evan-forbes, line 129: we have to use the normal query mechanism, and not the this one because the cosmos-sdk mocks the inputs and outputs, which doesn't know how to decode a blob transaction.

x/blob/types/errors.go
```diff
@@ -27,4 +27,5 @@ var (
 	ErrNamespaceMismatch              = sdkerrors.Register(ModuleName, 11127, "namespace of blob and its respective MsgPayForBlob differ")
 	ErrProtoParsing                   = sdkerrors.Register(ModuleName, 11128, "failure to parse a transaction from its protobuf representation")
 	ErrMultipleMsgsInBlobTx           = sdkerrors.Register(ModuleName, 11129, "not yet supported: multiple sdk.Msgs found in BlobTx")
+	ErrBloblessPFB                    = sdkerrors.Register(ModuleName, 11130, "MsgPayForBlob transactions must be paird with a blob")
```
cmwaters, line 30:
```suggestion
	ErrBloblessPFB                    = sdkerrors.Register(ModuleName, 11130, "MsgPayForBlob transactions must be paired with a blob")
```

cmwaters (approved):
Good catch!

rahulghangas (commented):
Fist pass looks good. Small nit. Will do a second pass before approving

app/check_tx.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/celestiaorg/celestia-app/x/blob/types"
+	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
```
rootulp, line 7: [nit] we can remove the duplicate import statement if we always use the import alias in this file

```suggestion
	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
```

x/blob/types/errors.go
rootulp: [optional] alternative error message

```suggestion
	ErrBloblessPFB                    = sdkerrors.Register(ModuleName, 11130, "transaction that contains a MsgPayForBlob must also contain a blob")
```

rootulp (dismissed):

rahulghangas (approved):

app/test/block_size_test.go
```diff
@@ -247,6 +250,28 @@ func (s *IntegrationTestSuite) TestSubmitWirePayForBlob() {
...
+		false,
+		s.cfg.ChainID,
+		s.accounts[:1],
+	)
+
+	btx, isBlob := coretypes.UnmarshalBlobTx(blobTx[0])
+	require.True(t, isBlob)
+
+	res, err := val.ClientCtx.BroadcastTxSync(btx.Tx)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
evan-forbes: yeah, good point, we should definitely be more consistent. I have been moving towards just using t, mainly because its less of a change between test suite tests and normal tests. also its one less thing we have to declare each test.

app/check_tx.go
evan-forbes: damn I though the linter would pick up on this!

evan-forbes: damn, we were doing this in a few places 1895fa48486c54a38acc50ada27c336e5f2c215b

evan-forbes: good catch!

rahulghangas (approved):

PR #1174, opened 2023-01-04
chore: un-export namespaced padded share bytes

rootulp:
This constant is an implementation detail and shouldn't be exported from celestia-app


evan-forbes (approved):
This brings up a great question, of whether or not this is an implementation detail! tbh, idk, only because what we use as padding feels important

I totally understand not exporting this specific thing, its just a zeroed slice of bytes

PR #1177, opened 2023-01-04
chore: improve appconsts godoc

rootulp:


pkg/appconsts/appconsts.go
```diff
@@ -49,21 +49,25 @@ const (
...
 	ContinuationSparseShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes
 
-	// DefaultMaxSquareSize is the maximum number of
-	// rows/columns of the original data shares in square layout.
-	// Corresponds to AVAILABLE_DATA_ORIGINAL_SQUARE_MAX in the spec.
-	// 128*128*256 = 4 Megabytes
-	// TODO(ismail): settle on a proper max square
-	// if the square size is larger than this, the block producer will panic
+	// DefaultMaxSquareSize is the maximum original square width. A square size
+	// larger than this will cause the block producer to panic.
```
cmwaters, line 53: How does the block producer panic?

pkg/appconsts/appconsts.go
```diff
@@ -49,21 +49,25 @@ const (
...
-	// cause block producer to panic
+	// DefaultMinSquareSize is the smallest original square width. A square size
+	// smaller than this will cause the block producer to panic.
 	DefaultMinSquareSize = 1
-	// MinshareCount is the minimum shares required in an original data square.
+
+	// MinshareCount is the minimum number of shares allowed in the original
+	// data square. If there are fewer shares than this, the block producer will
+	// panic.
 	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
```
cmwaters, line 71: Why do we have a minimum?

pkg/appconsts/appconsts.go
evan-forbes: that comment was from before we used prepare and process proposal, https://github.com/celestiaorg/celestia-core/blob/f4cf3df6e222160a1c4a3c66841dbd0ec17eb408/pkg/consts/consts.go#L33, so we can remove it imo

pkg/appconsts/appconsts.go
evan-forbes: to generate the minimum data availability header, but tbh, we can probably just remove and calculate the value there

rootulp: Will create a follow-up issue.

1. `MaxShareCount` appears unused so it can likely be removed
2. `MinShareCount` can be removed from appconsts if we inline it https://github.com/rootulp/celestia-app/blob/51e826de5358542a797890ad9b26bbfe0e6b7c8f/pkg/da/data_availability_header.go#L178

pkg/appconsts/appconsts.go
rootulp: Removed references to panic in https://github.com/celestiaorg/celestia-app/pull/1177/commits/51e826de5358542a797890ad9b26bbfe0e6b7c8f

pkg/appconsts/appconsts.go
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/1183

cmwaters: What about empty blocks?

cmwaters (approved):

PR #1193, opened 2023-01-06
.github: fix permissions issue on pr review request on forks

MSevey:
## Overview

The housingkeeping workflow was failing due to forks not having write access during workflow runs. This is for security reasons. However there are many non-security critical tasks that require write access, such as requesting reviewers or making comments on a PR. For these, the `pull_request_target` trigger can be used, as it allows write access to forks for the select workflows. 

I tested this on a personal github org and it worked for both native branches and forks. 

**NOTE**:  the reason the workflow is not running on this PR is because of the chicken and egg issue of updating workflows, in that the update needs to be on main in order for it to run. 

Closes https://github.com/celestiaorg/.github/issues/19

## Checklist

### DIFF
--- a/.github/workflows/housekeeping.yml
+++ b/.github/workflows/housekeeping.yml
@@ -1,7 +1,12 @@
 name: Housekeeping
 
 on:
-  pull_request:
+  # pull_request_target is used to allow forks write permissions when running
+  # this workflow. With the pull_request trigger, forks do not have any write
+  # access for security reasons, however write access is needed in order to
+  # request reviews. Since this workflow is simply requesting reviewers, it is
+  # safe to allow forks write access. 
+  pull_request_target:
 
 jobs:
   pr-assignment:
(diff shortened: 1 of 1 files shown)


.github/workflows/housekeeping.yml
```diff
@@ -1,7 +1,12 @@
 name: Housekeeping
 
 on:
-  pull_request:
+  # pull_request_target is used to allow forks write permissions when running
+  # this workflow. With the pull_request trigger, forks do not have any write
+  # access for security reasons, however write access is needed in order to
+  # request reviews. Since this workflow is simply requesting reviewers, it is
+  # safe to allow forks write access.
```
rootulp, line 8: This may fix yamllint

```suggestion
  # safe to allow forks write access.
```

rootulp (approved):
🙏

.github/workflows/housekeeping.yml
MSevey: that's what I get for doing it quickly in vim haha

rootulp (approved):

PR #1196, opened 2023-01-09
feat!: use MultiShareCommitments instead of a single ShareCommitment

evan-forbes:
## Overview

part of #388, and split apart from #1154

this PR uses a new mechanism to generate the share commit that can create a commitment over multiple blobs. We don't yet support multiple blobs per PFB, so this is currently effectively hashing the normal share commitment.

## Checklist

### DIFF
--- a/pkg/inclusion/get_commit.go
+++ b/pkg/inclusion/get_commit.go
@@ -4,8 +4,25 @@ import (
 	"errors"
 
 	"github.com/celestiaorg/celestia-app/pkg/da"
+	"github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/tendermint/tendermint/crypto/merkle"
 )
+
+func GetMultiCommit(cacher *EDSSubTreeRootCacher, dah da.DataAvailabilityHeader, startIndexes, lengths []uint32) ([]byte, error) {
+	shareCounts := make([]int, len(lengths))
+	for i := range lengths {
+		shareCounts[i] = shares.SparseSharesNeeded(uint32(lengths[i]))
+	}
+	commitments := make([][]byte, len(startIndexes))
+	for i := range startIndexes {
+		commitment, err := GetCommit(cacher, dah, int(startIndexes[i]), shareCounts[i])
+		if err != nil {
+			return nil, err
+		}
+		commitments[i] = commitment
+	}
+	return merkle.HashFromByteSlices(commitments), nil
+}
 
 func GetCommit(cacher *EDSSubTreeRootCacher, dah da.DataAvailabilityHeader, start, blobShareLen int) ([]byte, error) {
 	squareSize := len(dah.RowsRoots) / 2
--- a/x/blob/types/payforblob.go
+++ b/x/blob/types/payforblob.go
@@ -26,7 +26,7 @@ const (
 var _ sdk.Msg = &MsgPayForBlob{}
 
 func NewMsgPayForBlob(signer string, nid namespace.ID, blob []byte) (*MsgPayForBlob, error) {
-	commitment, err := CreateCommitment(nid, blob, appconsts.ShareVersionZero)
+	commitment, err := CreateMultiShareCommitment([][]byte{nid}, [][]byte{blob}, []uint32{uint32(appconsts.ShareVersionZero)})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


x/blob/types/payforblob.go
```diff
@@ -140,6 +140,22 @@ func CreateCommitment(namespace []byte, blobData []byte, shareVersion uint8) ([]
 	return merkle.HashFromByteSlices(subTreeRoots), nil
 }
 
+// CreateMultiShareCommitment generates a commitment over multiple blobs at
+// arbitrary points in the square. It uses the normal commitment creation
+// function per blob, and then creates a merkle root of those commitments.
+func CreateMultiShareCommitment(nsids [][]byte, blobs [][]byte, shareVersions []uint32) ([]byte, error) {
```
cmwaters, line 146: Why uint32 here for shareVersions?

pkg/inclusion/get_commit.go
```diff
@@ -4,9 +4,26 @@ import (
...
+	}
+	commitments := make([][]byte, len(startIndexes))
+	for i := range startIndexes {
+		commitment, err := GetCommit(cacher, dah, int(startIndexes[i]), shareCounts[i])
+		if err != nil {
+			return nil, err
+		}
+		commitments[i] = commitment
+	}
+	return merkle.HashFromByteSlices(commitments), nil
```
cmwaters, line 24: Q: Is the merkle hash of a single value just the sha256 hash?

cmwaters (approved):

pkg/inclusion/get_commit.go
evan-forbes: yeah, until we have multiple blobs, or for txs with a single blob, we're effectively taking the hash twice

x/blob/types/payforblob.go
evan-forbes: protobuf smallest supported uint is uint32, so that's what we end up using in many places

pkg/inclusion/get_commit.go
```diff
@@ -4,9 +4,26 @@ import (
 	"errors"
 
 	"github.com/celestiaorg/celestia-app/pkg/da"
+	"github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/tendermint/tendermint/crypto/merkle"
 )
 
+func GetMultiCommit(cacher *EDSSubTreeRootCacher, dah da.DataAvailabilityHeader, startIndexes, lengths []uint32) ([]byte, error) {
```
MSevey, line 11: [non-blocking] nit: Is there a rhyme or reason to the use of `commit` vs `commitment`?  I'd hate for us to run into some dumb bug that came from not finding all the right `commit` locations because we searched for `commitment`....  

I'd recommend a `chore` task to unify the naming to avoid any confusion in the future.

x/blob/types/payforblob_test.go
```diff
@@ -225,8 +225,9 @@ func validMsgPayForBlob(t *testing.T) *MsgPayForBlob {
 func TestNewMsgPayForBlob(t *testing.T) {
 	type test struct {
 		signer      string
```
rootulp, line 227: [optional] it may be helpful to include a name in the test struct so that the test cases that expect an error can describe why an error was expected:

```diff
	type test struct {
+              name         string
		signer      string
		nids        [][]byte
		blobs       [][]byte
		versions    []uint8
		expectedErr bool
	}
```

x/blob/types/payforblob_test.go
```diff
@@ -239,47 +240,52 @@ func TestNewMsgPayForBlob(t *testing.T) {
 	tests := []test{
 		{
 			signer:      addr.String(),
-			nid:         []byte{1, 2, 3, 4, 5, 6, 7, 8},
-			blob:        []byte{1},
+			nids:        [][]byte{{1, 2, 3, 4, 5, 6, 7, 8}},
+			blobs:       [][]byte{{1}},
+			versions:    make([]uint8, 1),
```
rootulp, line 245: [blocking][discussion] I would expect this test case to fail because it uses an unsupported share version. Do you think `CreateMultiShareCommitment` should perform validation on the share versions provided?

evan-forbes: > it uses an unsupported share version

wait, I'm confused, why is this? isn't it zero? perhaps we should be more explicit and use the const.

> should perform validation on the share versions provided?

imo no, because we don't want to add state to an otherwise stateless function. All verification of share version should occur during the creation process of a PFB

pkg/inclusion/get_commit.go
evan-forbes: The previous function was named commit, so this one was as well, but there is no reason we shouldn't be consistent across the board and not just per package

x/blob/types/payforblob_test.go
rootulp: > wait, I'm confused, why is this? isn't it zero? perhaps we should be more explicit and use the const.

Sorry I misread. On initial read I thought this was attempting to use the share version 1. You're right it is zero and I agree the const would be clearer.

> All verification of share version should occur during the creation process of a PFB

Fair enough

rootulp (approved):

pkg/inclusion/get_commit.go
evan-forbes: opened #1202 :slightly_smiling_face:

Issue #1206, opened 2023-01-10
Which namespace should padding between the last transaction and first blob use?


rootulp:
The specs describe a namespace that should be used for padding between the last transaction and the first message (a.k.a. blob)

https://github.com/celestiaorg/celestia-app/blob/b610a0b3052a2451871e16b097b71816681b6b8a/specs/src/specs/consensus.md?plain=1#L73

However, the current implementation uses the transactions namespace for such padding:

https://github.com/celestiaorg/celestia-app/blob/a9b97e67a41c6eff9d02b1b934428d79b35a5b1a/pkg/shares/share_splitting.go#L51

With the introduction of a new reserved namespace for PFB transactions, it is important to align on a desired namespace for such padding. Such padding will go between the last PFB transaction and the first blob. If no PFB transactions exist in the block, the padding will go between the last transaction and the first blob. Therefore, the namespace used for such padding should be > than the transactions namespace and the PFB transaction namespace but < than the first available blob namespace. The namespace in the specs matches this description.

rootulp:
> Does each reserved namespace result in a new share

Yes. 

> can you have multiple namespaces in a single compact share?

Each share has one and only one namespace so multiple distinct namespaces can't occupy the same share.

> If there are no PFBs then technically there are no blobs. In that case do we just use tail padding?

Good point, if there are no PFBs then there are no blobs. In that case, I expect the data square to have tail padding (with the namespace https://github.com/celestiaorg/celestia-app/blob/b610a0b3052a2451871e16b097b71816681b6b8a/specs/src/specs/consensus.md?plain=1#L74 between the last ordinary transaction and the end of the data square.

evan-forbes:
I'm also okay with doing what we do now, which is just normal namespace padding, which uses the last namespace added

rootulp:
> Do we get a specific benefit from use one particular namespace?

It currently isn't possible to differentiate namespace padding from a share with `0` bytes (see https://github.com/celestiaorg/celestia-app/issues/1136) so clients must download preceding shares to determine if the namespace padding share is significant. If we use a dedicated namespace then clients can determine if a share is tx padding by inspecting the namespace (i.e. no need to download other shares).

PR #1207, opened 2023-01-10
feat!: use tail transaction padding namespace

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1206

Marked breaking because although this is an implementation detail, it has end-user visible changes for users that parse shares directly.


rootulp:
Marking as ready for review because this is a prerequisite to the dedicated PFB tx namespace work.

pkg/appconsts/appconsts.go
```diff
@@ -87,6 +87,10 @@ var (
 	// EvidenceNamespaceID is the namespace reserved for evidence.
 	EvidenceNamespaceID = namespace.ID{0, 0, 0, 0, 0, 0, 0, 3}
 
+	// TailTransactionPaddingNamespaceId is the namespace used for padding after
+	// all transactions (ordinary and PFBs) but before blobs.
+	TailTransactionPaddingNamespaceID = namespace.ID{0, 0, 0, 0, 0, 0, 0, 255}
```
cmwaters, line 92: Q: We don't want it to just increment from the previous one i.e. 4?

cmwaters (approved):

pkg/appconsts/appconsts.go
rootulp: I think it would be preferable if TailTransactionPaddingNamespaceID was the maximum reserved namespace (i.e. `0, 0, 0, 0, 0, 0, 0, 255`) because it leaves us the option of introducing new reserved namespaces in the 4-255 range in the future.

evan-forbes, line 92: [non-blocking naming question]
Will we always have only transactions in the reserved namespaces? For example, what happens when we add ISRs? should we change this name then?

evan-forbes (approved):

pkg/appconsts/appconsts.go
rootulp: Agreed. Proposed alternative names:
- `TailReservedNamespacePadding`
- `ReservedNamespacePadding`

rootulp: Created a separate issue b/c this entails two renames:
1. the implementation
2. the specs

PR #1211, opened 2023-01-11
chore: stop overestimating the square size 

### LINKED ISSUE #1200: Reduce estimated padding
### Summary

Currently, we estimate the padding by 2 fold. 

https://github.com/celestiaorg/celestia-app/blob/5f8b39b4cb7ab0a9083b97f2a20cae7a2cc1a1c5/app/estimate_square_size.go#L32-L35

We can reduce this estimation to 0,375 as this is the maximum amount of padding that can be created without tail padding following the current non-interactive default rules. This was shown in my calculations of [ADR 009 ](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md#5-what-is-the-worst-constructible-block-with-the-most-amount-of-padding-with-old-and-new-non-interactive-default-rules). 

### Proposal

	totalSharesUsed := float64(txSharesUsed + blobSharesUsed)
	totalSharesUsed *= 1.375
	minSize := uint64(math.Sqrt(totalSharesUsed))
	squareSize = shares.RoundUpPowerOfTwo(minSize)

### For Admin Use

- [ ] Not duplicate issue

evan-forbes:
## Overview

closes #1200 by reducing the overestimation of the square size

## Checklist


testutil/testnode/full_node_test.go
```diff
@@ -96,7 +96,7 @@ func TestIntegrationTestSuite(t *testing.T) {
 func (s *IntegrationTestSuite) Test_FillBlock() {
 	require := s.Require()
 
-	for squareSize := 2; squareSize < appconsts.DefaultMaxSquareSize; squareSize *= 2 {
+	for squareSize := 4; squareSize < appconsts.DefaultMaxSquareSize; squareSize *= 2 {
```
cmwaters, line 99: Why did you change this? Just to make the tests quicker?

cmwaters (approved):

nashqueue:
Thank you for the quick fix! 

It should be `txSharesUsed + 1.375 * blobSharesUsed + 3`

So first, it was 37.5% and not 35% in the analysis.
I also realised that txSharesUsed could be seen as a big first blob in the analysis, so we can disregard that as padding is already included in that estimation. If we disregard it, then we need to add three extra padding in the worst case, like in the example below. 
image

app/estimate_square_size.go
```diff
@@ -26,14 +26,13 @@ func estimateSquareSize(txs []parsedTx) (squareSize uint64, nonreserveStart int)
...
-	// estimate as much
-	totalSharesUsed := uint64(txSharesUsed + blobSharesUsed)
-	totalSharesUsed *= 2
-	minSize := uint64(math.Sqrt(float64(totalSharesUsed)))
+	totalSharesUsed := float64(txSharesUsed + blobSharesUsed)
+	if totalSharesUsed <= 1 {
+		return appconsts.DefaultMinSquareSize, txSharesUsed
+	}
+	// increase the total shares used by the worst case padding ratio
+	totalSharesUsed *= 1.35
```
rootulp, line 34: See https://github.com/celestiaorg/celestia-app/issues/1200#issuecomment-1379404250

rootulp (commented):
I briefly explored adding a fuzz test to identify if reducing the overestimation would cause a invalid square size to be estimated and it didn't identify any issues. Source code if you're curious:

```golang
func FuzzEstimateSquareSize(f *testing.F) {
	f.Add([]byte{0xff}, uint32(1), []byte{0xff}, []byte{0xff})
	f.Fuzz(func(t *testing.T, normalTx []byte, shareIndex uint32, data []byte, tx []byte) {
		parsedTx := parsedTx{
			normalTx:   normalTx,
			shareIndex: shareIndex,
			blobTx: types.ProcessedBlobTx{
				Blobs: []tmproto.Blob{
					{
						NamespaceId:  namespace.RandomBlobNamespace(),
						Data:         data,
						ShareVersion: uint32(appconsts.ShareVersionZero),
					},
				},
				Tx: tx,
			},
		}

		parsedTxs := Repeat(parsedTx, 1000)
		squareSize, _ := estimateSquareSize(parsedTxs)
		assert.LessOrEqual(t, int(squareSize), appconsts.DefaultMaxSquareSize)
		assert.GreaterOrEqual(t, int(squareSize), appconsts.DefaultMinSquareSize)
	})
}

func Repeat[T any](s T, count int) []T {
	ss := make([]T, count)
	for i := 0; i < count; i++ {
		ss[i] = s
	}
	return ss
}
```

evan-forbes:
apologies for the delay on this

The fuzzer is a really good idea, but unfortunately I don't think that its telling us what we want because of these lines
https://github.com/celestiaorg/celestia-app/blob/222bb1e0a6013efe44e910d7328c6d141fa99d1a/app/estimate_square_size.go#L40-L45

the estimation function cannot return values outside of the min and the max, and it cannot return values that are not powers of two, therefore, that fuzzer should always pass unless I'm missing something

testutil/testnode/full_node_test.go
evan-forbes: originally, yeah just to simplify the test, and I don't think anyone is using this to fill the square to size 2. Still, in retrospect it was the wrong move, and I should have just adjusted the fill block test to account for this...

However now we can't actually have square sizes of 2 anymore after the adjustment as we're adding 3 shares of padding to all estimates.

app/estimate_square_size.go
```diff
@@ -28,14 +40,14 @@ func estimateSquareSize(txs []parsedTx) (squareSize uint64, nonreserveStart int)
...
-	// estimate as much
-	totalSharesUsed := uint64(txSharesUsed + blobSharesUsed)
-	totalSharesUsed *= 2
-	minSize := uint64(math.Sqrt(float64(totalSharesUsed)))
+	totalSharesUsed := float64(txSharesUsed + blobSharesUsed)
+	if totalSharesUsed <= 1 {
+		return appconsts.DefaultMinSquareSize, txSharesUsed
+	}
+	// increase the total shares used by the worst case padding ratio
+	totalSharesUsed *= worstCasePaddingCoefficient
```
nashqueue, line 48: Because we intorduced the `worstCasePaddingBase` we don't need to multiply `totalSharesUsed` by the coefficient. Only blobSharesUsed.

evan-forbes:
just to update this, I'm still planning on completing this, but it is simply a lower priority than all the other incentivized testnet things

evan-forbes:
closing in favor of pursuing deterministic square creation

Issue #235, opened 2022-03-08
Investigate limiting IBC connections


liamsi:
## Summary

While we do want IBC connections to a few selected chains (https://github.com/celestiaorg/celestia-app/issues/734), we do not want that anyone can establish a connection to arbitrary zones. Reason is that we want to keep the state machine as minimal as possible as well as the base layer as focused as possible.

## Problem Definition

Need to limit the chains that can connect to Celestia to a predefined allow-list. That allow-list might grow over time. Either it can be modified via governance (#168), or, modifications to the list must require a hard-fork.

## Proposal

TODO write proposal in more details:
 - create ibc clients for target chains on genesis
 - reject CreateClient txs
 - allow governance to whitelist and create new clients?
 - ...

musalbas:
>Reason is that we want to keep the state machine as minimal as possible as well as the base layer as focused as possible.

Can we expand on this a bit? What extra complexity on computation is introduced if arbitrary IBC connections are allowed? Is it that relayers can cause Celestia validators to validate arbitrary block headers and signatures from other chain? I think, if the transaction fee for that is priced correctly it might not be necessarily bad.

liamsi:
> Is it that relayers can cause Celestia validators to validate arbitrary block headers and signatures from other chain?

Yes, though as the validator sets (or at least the +2/3 of stake) are relatively stable usually, this is not too bad as IBC defaults to skipping verification and does not require a [MsgUpdateClient](https://github.com/cosmos/ibc-go/blob/f5962c3324ee7e69eeaa9918b65eb1b089da6095/modules/core/02-client/types/tx.pb.go#L115) too often to keep the connection alive.

> I think, if the transaction fee for that is priced correctly it might not be necessarily bad, but not entirely sure yet.

I am not sure if the Txs are priced in a particularly special way. The more sigs, the larger the Tx, the more gas it would consume (and the higher the fee). But not sure if that matches "priced correctly".

liamsi:
An alternative we could consider here is to instead "white-list" the token denoms that go in and out. Basically any Tx that is not related to the celestia native token should be dropped.

musalbas:
Are we sure we need a white-list anyways? Is adding new IBC connections execution/state heavy?

PR #1219, opened 2023-01-11
feat! implement token filter IBC middleware

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/235

This PR creates and wires up a new IBC middleware that acts as a firewall, rejecting all `FungibleTokenPacketData` (i.e. transfer packets) that have a denom which did not originally came from the celestia chain. 

This simple implemenation will mean that the chain state only consists of the native celestia token.


go.mod
```diff
@@ -184,5 +186,5 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.5.1-sdk-v0.46.6
```
rootulp, line 187: should we cut https://github.com/celestiaorg/cosmos-sdk v1.6.0-sdk-v0.46.7 and bump to that here given the diff above

```diff
-	github.com/cosmos/cosmos-sdk v0.46.6
+	github.com/cosmos/cosmos-sdk v0.46.7
```

go.mod
```diff
@@ -184,5 +186,5 @@ require (
...
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint v0.34.23 => github.com/celestiaorg/celestia-core v1.12.1-tm-v0.34.23
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.12.1-tm-v0.34.23
```
rootulp, line 189: should we cut https://github.com/celestiaorg/tendermint/ v1.13.0-tm-v0.34.24 and bump to that here given the diff above

```diff
-	github.com/tendermint/tendermint v0.34.23
+	github.com/tendermint/tendermint v0.34.24
```

go.mod
cmwaters: Already did. Just haven't updated it here. But will do in a separate PR

x/tokenfilter/ibc_middleware.go
```diff
@@ -0,0 +1,81 @@
...
+// tokenFilterMiddleware directly inherits the IBCModule and ICS4Wrapper interfaces.
+// Only with OnRecvPacket, does it wrap the underlying implementation with additional
+// stateless logic for rejecting the inbound transfer of non-native tokens. This
+// middleware is unilateral and no handshake is required. If using this middleware
+// on an existing chain, tokens that have been routed through this chain will still
+// be allowed to unwrap.
+type tokenFilterMiddleware struct {
+	porttypes.IBCModule
+	porttypes.ICS4Wrapper
+}
```
damiannolan, line 25: Kinda nice trick with the embedding to avoid code boilerplate. Was this the intention here?

I do like the idea of not having to have a `keeper` package, given its a stateless middleware. 
But I would've expected to see the transfer keeper composed with something like `app.TokenFilterKeeper`.

```diff
app.TransferKeeper = ibctransferkeeper.NewKeeper(
	appCodec, keys[ibctransfertypes.StoreKey], app.GetSubspace(ibctransfertypes.ModuleName),
-	app.IBCKeeper.ChannelKeeper, app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
+       app.TokenFilterKeeper, app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
	app.AccountKeeper, app.BankKeeper, scopedTransferKeeper,
)
```
At the moment its completely bypassing this tokenfilter middleware for `SendPacket`. With the above, the `TokenFilterKeeper` would just be a simple struct composed with the ibc channel keeper.

E.g. for packets being received the flow is: `ibc core -> token filter -> transfer` where token filter essentially inherits transfers callbacks but overrides `OnRecvPacket`. When transfer sends a packet, the current flow is: `transfer -> ibc core`. 

While its technically fine for this right now, and token filter doesn't care about outbound packets I think the configuration might be a little bit error prone, for example, the `ICS4Wrapper` in this struct is never actually used. It would definitely become apparent especially if you try to add more middlewares to the transfer application stack in future. 

Does this make sense?

damiannolan (commented):
Nice work! The general logic looks good for `OnRecvPacket`. 
I like the approach for avoiding boilerplate code but I think there is some concerns for maintainability as its really subtle how things are wired here. My main concern is that transfer sends completely bypass the middleware

x/tokenfilter/ibc_middleware.go
damiannolan: > So because I just reference the ChannelKeeper directly, any middleware below this would get bypassed?

Yes, because `ChannelKeeper` is passed directly as the `ICS4Wrapper` arg to `NewTransferKeeper` then `SendPacket` would bypass all middlewares in the application stack. 

> Also, I'm not sure if I really care about SendPacket in the case of the TokenFilter and I would like to avoid having a keeper if possible.

Yeah I understand wanting to avoid a keeper, and like you say, `SendPacket` is essentially going to be a passthrough anyways and not implement any logic. 
I guess its just a general concern that you should be aware of with the current way its wired up, if you decide to make any changes or add additional middlewares in future which will want to implement logic on `SendPacket`. 

I would probably advise for taking the approach where the token filter is invoked bi-directionally, even if its a no-op. 
i.e.

```
recv: ibc core -> token filter -> transfer
send: transfer -> token filter -> ibc core
```

This way you avoid any future errors made my devs who are unaware of the call flow. 

All that being said, I understand your reasoning for avoiding boilerplate and the need for a keeper package. As long as you're aware of the risks then feel free to go ahead with this.

damiannolan: fwiw you could define an `ICS4Wrapper` struct within this package, and use that as the arg in `NewTransferKeeper`. You'd avoiding having to add a keeper package that way and still stick to convention and adhere to the correct flows

cmwaters: Ok I've given what you suggested a go just for conventions sake. 

It's does strike me however a bit strange that `Middleware` is both `IBCModule` and `ICS4Wrapper`. This is kind of impossible to implement within the same struct else you'd end with circular dependencies. In this case transfer keeper needs to both import token filter and be imported by token filter.

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 require (
 	cosmossdk.io/errors v1.0.0-beta.7
 	cosmossdk.io/math v1.0.0-beta.3
-	github.com/celestiaorg/rsmt2d v0.7.0
+	github.com/celestiaorg/rsmt2d v0.8.0
```
rootulp, line 30: [question][optional] does this PR depend on anything from rsmt2d v0.8.0? I don't think it does so this dependency upgrade could be extracted to a stand-alone PR to minimize the diff and isolate independent changes.

rootulp (approved):
Overall LGTM 🚀

go.mod
cmwaters: Must have bumped when I ran go mod tidy. No it doesn't depend on rsmt2d. I can extract it out

cmwaters: Ah it's in `main`. I just haven't updated

rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1219/commits/eff692744563c57e659ce8d08ac91f80056ce0a6

evan-forbes:
will finish reviewing this by the end of the day. I was attempting to get use ibctest to test this Friday, but I think we can push that into a different PR later as this change makes sense to me afaiu it

cmwaters:
> I was attempting to test using ibctest to test this Friday, but we can definitely push that into a different PR later. 

Yeah some integration tests were definitely on my mind as well but I don't think they're that trivial when it comes to IBC so I was going to address that in a different PR

evan-forbes (approved):
nice! I don't think it could be implemented any simpler :+1: :+1: 

created #1249 for us to create some test between now an mainnet after this gets included in a docker container on ghcr. We'll also be testing this thoroughly an all testnets between now and then.

evan-forbes:
also, huge shoutout to @damiannolan for the expertise! :slightly_smiling_face:

damiannolan:
> > I was attempting to test using ibctest to test this Friday, but we can definitely push that into a different PR later.
> 
> Yeah some integration tests were definitely on my mind as well but I don't think they're that trivial when it comes to IBC so I was going to address that in a different PR

We have a pretty nice e2e module which uses the ibctest docker framework, its a thin wrapper around it to suit the needs of ibc-go. We maintain it as a separate go module which you could import. We've found it extremely valuable and its caught many small issues, even some within the sdk etc. Essentially you build a test struct which embeds `testsuite.E2ETestSuite` and you can leverage a whole bunch of functionality (spin up chains, relayer, generic broadcast msg, query clients..etc). 

Might be worth checking out some of tests in here if you're interested https://github.com/cosmos/ibc-go/blob/main/e2e

damiannolan (approved):
Very clean PR @cmwaters! 

Thanks for the feedback around the ibc middleware wiring. We'd like to improve the dev ux wrt wiring application stacks and I think this PR brought some things to light especially for stateless middlewares that don't necessarily require keepers.

cmwaters:
> Might be worth checking out some of tests in here if you're interested https://github.com/cosmos/ibc-go/blob/main/e2e

Will look into it thanks

PR #1235, opened 2023-01-13
fix: add size check in RandBlobTxsWithAccounts

SweeXordious:
## Overview

The parameter `size` in `RandBlobTxsWithAccounts` accepts positive values as well as negative.
However, passing a negative value will make the random generator panic, thus making the integration tests fail without any clear error message. An example of this: https://github.com/celestiaorg/celestia-app/actions/runs/3911542927/jobs/6685657620

Thus, adding a check on `size` would fix this and never allow anyone to pass a negative value.

## Checklist


evan-forbes (approved):
:+1:

Issue #1184, opened 2023-01-06
QGB create an inclusion proof verification command 


SweeXordious:
As part of the inclusion proofs generation, we will want to have a simple way to verify that a certain share has been committed to in a QGB contract.

To do so, we will implement a `verify` command as part of the QGB V1 that takes a share and a QGB contract and verifies if the former has been committed to.

PR #1233, opened 2023-01-13
feat: add shares inclusion proofs

SweeXordious:
## Overview
This PR contains the changes proposed by https://github.com/celestiaorg/celestia-app/pull/1042 and https://github.com/celestiaorg/celestia-app/pull/1072.

The reason for creating a new PR  is to target main branch instead of qgb-integration without having to cherry pick all the changes.

Closes: https://github.com/celestiaorg/celestia-app/issues/1071 and https://github.com/celestiaorg/celestia-app/issues/1044.

The other PRs will be closed.

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #1236, opened 2023-01-13
feat: Add QGB verification command

SweeXordious:
## Overview

This command verifies that some shares were attested to in the QGB smart contract. 

Blocked by https://github.com/celestiaorg/celestia-app/pull/1233

Closes https://github.com/celestiaorg/celestia-app/issues/1184

## Checklist


x/qgb/cmd/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return VerifyConfig{
+		CelestiaChainID: chainID,
+		EVMChainID:      evmChainID,
+		CelesGRPC:       celesGRPC,
+		TendermintRPC:   tendermintRPC,
+		EVMRPC:          evmRPC,
+		ContractAddr:    address,
+	}, nil
+}
```
evan-forbes, line 78: alternatively, if we wanted to, we could get some of these flags from the client context passed around in the cobra context.

cmd/celestia-appd/cmd/root.go
```diff
@@ -165,6 +167,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig encoding.Config) {
 		queryCommand(),
 		txCommand(),
 		keybase,
+		qgbcmd.VerifyCmd(),
```
evan-forbes, line 170: something we might want to think about is adding this command to the query command, since this is a query

evan-forbes (commented):
@SweeXordious I'm assuming that this is required for the inclusion proofs? Let's also create an issue to add more usage documentation over how to use this command, what we are receving in return etc. Also, how to access the same things via RPC would also be needed.

will defer to @rahulghangas, let's try to merge this by thursday to give us some time for the rest of the things needed for the qgb mocked launch

SweeXordious:
> Let's also create an issue to add more usage documentation over how to use this command, what we are receving in return etc. Also, how to access the same things via RPC would also be needed.

https://github.com/celestiaorg/celestia-app/issues/1254

cmd/celestia-appd/cmd/root.go
SweeXordious: that's a good idea. The reason I didn't do it  is that the query commands  somehow are  consistent: they use the RPC/GRPC  endpoints to query things from state, like everything needed is gotten from core/app. However, in this case, we're also connecting to an EVM chain and interacting with a  smart contract.

I am okey with either tbh, let me know if I should change

evan-forbes (dismissed):
took another look, and LGTM

would appreciate a second set of eyes, but we also need to push this through and focus on the other things needed for the mocked QGB v1 launch

x/qgb/cmd/config.go
SweeXordious: Concerning this, I made it using `flags.AddQueryFlagsToCmd(command)` locally , however, it adds a special height flag:
```
Use a specific height to query state at (this can error if the node is pruning state)
```
which I don't support in my query.

what do you think?

I guess, the better thing to do is to keep the same explicitly declared flags. However, change their description to be consistent with the query flags defined in the cosmos-sdk. What do you think? 

For example `celes-http-rpc` => `node` for tendermint RPC.

SweeXordious:
Applied feedback. Can I get an approve to merge this?

Issue #1224, opened 2023-01-12
Remove the ProcessedTx struct


evan-forbes:
Question: why do we have this when it's essentially no different from `BlobTx` (with the exception of the TypeID field)? Can't we just have something like `func (b BlobTx) Validate(txcfg client.TxEncodingConfig) error`

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/pull/1154#discussion_r1060476229_

PR #1240, opened 2023-01-16
chore: consolidate processedBlobTx with BlobTx

cmwaters:
Closes: #1224

### DIFF
--- a/x/blob/types/blob_tx.go
+++ b/x/blob/types/blob_tx.go
@@ -9,12 +9,6 @@ import (
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 )
 
-// ProcessedBlobTx caches the unmarshalled result of the BlobTx
-type ProcessedBlobTx struct {
-	Blobs []Blob
-	Tx    []byte
-}
-
 // Blob wraps the tendermint type so that users can simply import this one.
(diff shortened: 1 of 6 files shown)


evan-forbes (dismissed):

x/blob/types/blob_tx.go
```diff
@@ -37,46 +31,46 @@ func NewBlob(ns namespace.ID, blob []byte) (*Blob, error) {
 	}, nil
 }
 
-// ProcessBlobTx performs stateless checks on the BlobTx to ensure that the
+// ValidateBlobTx performs stateless checks on the BlobTx to ensure that the
 // blobs attached to the transaction are valid. During this process, it
 // separates the blobs from the MsgPayForBlob, which are returned in the
 // ProcessedBlobTx.
```
rootulp, line 37: This function no longer returns a `ProcessedBlobTx`.

```suggestion
// ValidateBlobTx performs stateless checks on the BlobTx to ensure that the
// blobs attached to the transaction are valid.
```

SweeXordious (approved):

x/blob/types/blob_tx.go
cmwaters: Thanks

rootulp (approved):

Issue #1175, opened 2023-01-04
specs: tail padding shares


rootulp:
## Context

Similar to https://github.com/celestiaorg/celestia-app/issues/1136 but for tail padding shares

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/1244

PR #1244, opened 2023-01-16
docs: spec tail padding share

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1175 

See the implementation [here](https://github.com/celestiaorg/celestia-app/blob/d7ac0b84765db3f649cd82326e3ead066d164637/pkg/shares/tail_padding.go#L9-L21).

## Screenshot

Screenshot 2023-01-17 at 11 04 14 AM

### DIFF
--- a/specs/src/specs/data_structures.md
+++ b/specs/src/specs/data_structures.md
@@ -546,7 +546,7 @@ For shares **with a namespace ID equal to [`PARITY_SHARE_NAMESPACE_ID`](./consen
 
 For shares **with a namespace ID equal to [`TAIL_PADDING_NAMESPACE_ID`](./consensus.md#constants)** (i.e. tail padding shares):
 
-The first [`NAMESPACE_ID_BYTES`](./consensus.md#constants) of a share's raw data `rawData` is the namespace ID of that share, `namespaceID`. The next [`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are for share information. The sequence start indicator is always `0`. The version bits are filled with the the share version. The remaining [`SHARE_SIZE`](./consensus.md#constants)`-`[`NAMESPACE_ID_BYTES`](./consensus.md#constants)`-`[`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are filled with `0`.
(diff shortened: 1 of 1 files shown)


specs/src/specs/data_structures.md
```diff
@@ -538,6 +542,12 @@ For shares **with a namespace ID equal to [`PARITY_SHARE_NAMESPACE_ID`](./consen
 
 - Bytes carry no special meaning.
 
+#### Tail Padding Share
+
+For shares **with a namespace ID equal to [`TAIL_PADDING_NAMESPACE_ID`](./consensus.md#constants)** (i.e. tail padding shares):
+
+The first [`NAMESPACE_ID_BYTES`](./consensus.md#constants) of a share's raw data `rawData` is the namespace ID of that share, `namespaceID`. The next [`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are for share information. The sequence start indicator is always `0`. The version bits are filled with the the share version. The remaining [`SHARE_SIZE`](./consensus.md#constants)`-`[`NAMESPACE_ID_BYTES`](./consensus.md#constants)`-`[`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are filled with `0`.
```
rootulp, line 549:
```suggestion
The first [`NAMESPACE_ID_BYTES`](./consensus.md#constants) of a share's raw data `rawData` is the namespace ID of that share, `namespaceID`. The next [`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are for share information. The sequence start indicator is always `0`. The version bits are filled with the share version. The remaining [`SHARE_SIZE`](./consensus.md#constants)`-`[`NAMESPACE_ID_BYTES`](./consensus.md#constants)`-`[`SHARE_INFO_BYTES`](./consensus.md#constants) bytes are filled with `0`.
```

cmwaters (dismissed):
LGTM. 

Does it make sense to add here how tail padding is used rather than just what it is?

rootulp:
I think it would help. How about?

> Tail padding shares are placed after the last blob in the data square so that the number of shares in the data square is a perfect square. Clients can safely ignore the contents of these shares because they don't contain any significant data.

Although this just made me wonder: what if a block proposer includes tail padding shares with the tail padding namespace but actual data populated in the `rawData`? Do we want to check for and explicitly reject blocks with such shares?

cmwaters (approved):
Thank you!

Issue #1245, opened 2023-01-16
namespaced padded shares contain `isSequenceStart=true`


rootulp:
## Context

We haven't reached alignment on how to specify namespaced padding shares (see https://github.com/celestiaorg/celestia-app/issues/1136). The current implementation doesn't conform to either of the proposed options in the above issue. Instead, it uses a hacky mechanism where `isSequenceStart=true` but the sequence length is set to `0` [here](https://github.com/celestiaorg/celestia-app/blob/5144692c0ef1cd7ca3dd22e83dcbeec00abad89e/pkg/shares/split_sparse_shares.go#L168-L174). This fact is used later when we parse shares [here](https://github.com/celestiaorg/celestia-app/blob/db4f16179cb41732b0e10267f5742850cc1f1ba0/pkg/shares/parse_sparse_shares.go#L15-L19). 

## Proposal
We don't need to rely on the hacky mechanism. We can assume that the blob prior to the namespaced padding share has written a valid sequence length and so the namespaced padding `0` bytes will be trimmed when we parse shares [here](https://github.com/celestiaorg/celestia-app/blob/db4f16179cb41732b0e10267f5742850cc1f1ba0/pkg/shares/parse_sparse_shares.go#L76-L77).

PR #1246, opened 2023-01-16
chore: namespaced padding share `isSequenceStart=false`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1245

- Extract `NamespacedPaddedShare` to a stand-alone file
- Modify namespaced padded share to set `isSequenceStart=false`
- Resolve a TODO by moving tests to `parse_sparse_shares_test.go`


pkg/shares/namespaced_padding.go
```diff
@@ -0,0 +1,34 @@
...
+	"bytes"
+
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/nmt/namespace"
+)
+
+// NamespacedPaddedShare returns a share that acts as padding. Namespaced
+// padding shares follow a blob so that the next blob may start at an index that
+// conforms to non-interactive default rules.
+func NamespacedPaddedShare(ns namespace.ID) Share {
```
cmwaters, line 13: Is the namespace here always the same as the blob before it?

cmwaters (approved):
Nice work. I just have one question

pkg/shares/namespaced_padding.go
rootulp: Yes. I should make that explicit in the godoc. How about?

```go
// NamespacedPaddedShare returns a share that acts as padding. Namespaced
// padding shares follow a blob so that the next blob may start at an index that
// conforms to non-interactive default rules. The ns parameter provided should
// be the namespace of the blob that precedes this padding in the data square.
```

MSevey:
I don't think I did.... something weird seems to be happening with the github action maybe.
Screen Shot 2023-01-17 at 3 58 07 PM

pkg/shares/namespaced_padding.go
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1246/commits/ef002ec4cf2869c3c061e07ae24c689193f9a2ec

cmwaters (approved):

cmwaters:
I think that's just how the bot works. Also whenever I change is made to a PR, all prior reviews are dismissed so the request for reviews happens again

PR #1255, opened 2023-01-18
fix: stop saving snapshot files to wrong directory during tests

evan-forbes:
## Overview

by not setting the testing directory during integration tests, we end up saving the LOCK and LOG files from the snapshot directory to permanent directories instead of the temp ones created during the test.

this can be testing by running the integration tests locally without this change, and then with this change.

## Checklist


rootulp (approved):

rootulp:
Thanks! I was wondering why I had a dirty git dir at `testutil/testnode/data/` after running `make test` locally.

Issue #1133, opened 2022-12-18
Proposal: Remove the gas limit arg from the canonical `SubmitPayForBlob` api


evan-forbes:
We are currently including the gas limit as an argument to the `SubmitPayForBlob` function. While specifying the gas limit is important, it differs from how we modify other equally important transaction options. I'm suggesting that we use the same mechanism for changing a transaction option for everything, which is to use the `types.TxBuilderOption` variadic arg in the same function.

https://github.com/celestiaorg/celestia-app/blob/c16d6e7fd984184d2aba44c1e8c4a5aadfe36aef/x/blob/payforblob.go#L16-L27

kinda related: in the release that includes the upcoming refactor to include multiple blobs per PFB, we are going to have to change the canonical interface for submitting data on chain. While we're at it, I figured we might want to include other breaking changes.

cc @renaynay @distractedm1nd

rootulp:
We already have a [TxBuilderOption](https://github.com/celestiaorg/celestia-app/blob/b6e265133788dc45e94284dcf72be4b0ca4b04c0/x/blob/types/builder_options.go#L12) so it seems safe to remove the [`gasLim`](https://github.com/celestiaorg/celestia-app/blob/b6e265133788dc45e94284dcf72be4b0ca4b04c0/x/blob/payforblob.go#L22) parameter.

PR #1262, opened 2023-01-18
chore!: remove `gasLim` param in favor of `TxBuilderOption`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1133

### DIFF
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -237,6 +237,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 			mustNewBlob([]byte{1, 2, 3, 4, 5, 6, 7, 8}, tmrand.Bytes(3000)),
 			[]blobtypes.TxBuilderOption{
 				blobtypes.SetFeeAmount(sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))),
+				blobtypes.SetGasLimit(1_000_000_000),
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


cmwaters (approved):
LGTM.

What happens if you don't set a gas limit. Is it just that the tx is run with infinite gas (or until the tx feepayer runs out)?

rootulp:
Good question! I'm observing that a default gas limit of `210000` gets used. This appears to originate from cosmos-sdk [here](https://github.com/rootulp/cosmos-sdk/blob/0a209fbf6ac3e3239dcca955f30b8860b7c8b97b/client/flags/flags.go#L18). 

## Example

```go
$ celestia-appd tx blob payForBlob 0303030303030303 73616d706c6520626c6f62 --from celestia192vxtd9y5frcjvse5mjagmekwd50m8svxfep5z --keyring-backend test
auth_info:
  fee:
    amount: []
    gas_limit: "210000"
    granter: ""
    payer: ""
  signer_infos: []
  tip: null
body:
  extension_options: []
  memo: ""
  messages:
  - '@type': /blob.MsgPayForBlob
    blob_sizes:
    - 11
    namespace_ids:
    - AwMDAwMDAwM=
    share_commitment: eOqL62hbmSx+2Y4k7lE3VcfFNRORZo59iNAgYtsZkbs=
    share_versions:
    - 0
    signer: celestia192vxtd9y5frcjvse5mjagmekwd50m8svxfep5z
  non_critical_extension_options: []
  timeout_height: "0"
signatures: []
confirm transaction before signing and broadcasting [y/N]: y
```

evan-forbes (approved):

Issue #1253, opened 2023-01-17
Remove references to "malleated"


rootulp:
## Context

celestia-app no longer performs transaction malleation. There are a few residual filenames, variable names, comments that reference "malleated".

## Proposal

Remove all references to "malleated" and "malleation"

PR #1264, opened 2023-01-18
chore: remove references to "malleated"

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1253

### DIFF
--- a/app/encoding/index_wrapper_decoder.go
+++ b/app/encoding/index_wrapper_decoder.go
@@ -5,10 +5,10 @@ import (
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
-func WrappedTxDecoder(dec sdk.TxDecoder) sdk.TxDecoder {
+func IndexWrapperDecoder(dec sdk.TxDecoder) sdk.TxDecoder {
 	return func(txBytes []byte) (sdk.Tx, error) {
-		if malleatedTx, has := coretypes.UnmarshalIndexWrapper(txBytes); has {
-			return dec(malleatedTx.Tx)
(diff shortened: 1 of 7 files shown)


x/blob/README.md
```diff
@@ -23,10 +23,6 @@ When a `MsgPayForBlob` is processed, it consumes gas based on the blob size.
 
 - [`MsgPayForBlob`](https://github.com/celestiaorg/celestia-app/blob/8b9c4c9d13fe0ccb6ea936cc26dee3f52b6f6129/proto/blob/tx.proto#L39-L44) pays for the blob to be included in the block.
 
-## PrepareProposal
-
-The malleation process occurs during the PrepareProposal step.
```
cmwaters, line 28: I would still have a section here talking about how the PFB gets put in a wrapper that also contains the index(es) of the blob(s) that it references

rootulp: proposal to tackle this in a stand-alone PR b/c there are several things in this file that need updating and they're not all related to malleation: https://github.com/celestiaorg/celestia-app/issues/1268

app/encoding/index_wrapper_decoder.go
```diff
@@ -5,10 +5,10 @@ import (
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
-func WrappedTxDecoder(dec sdk.TxDecoder) sdk.TxDecoder {
+func IndexWrapperDecoder(dec sdk.TxDecoder) sdk.TxDecoder {
```
evan-forbes, line 8: [super minor no blocking comment]
I know no one uses this function, but does this mean this is technically API breaking? (so chore! ?)

rootulp: oof yea it's technically exported public API so this should've been `chore!`

Issue #1268, opened 2023-01-19
Update `x/blob/README.md`


rootulp:
## Context

The `x/blob` module contains docs at [x/blob/README.md](https://github.com/celestiaorg/celestia-app/blob/main/x/blob/README.md)

## Problem

The implementation has changed so some things in the docs are inaccurate.

## Proposal

1. `MsgPayForBlob` transactions use a dedicated reserved namespace
2. A MsgPayForBlob can support multiple blobs
3. There is no more transaction malleation, instead PrepareProposal does index wrapping
4. Explain the different types of transactions (blob tx vs. index wrapped tx)
5. [optional] include a diagram of transaction flow from end user to in a block
6. ~~links to code need to be updated. E.g. https://github.com/celestiaorg/celestia-app/blob/76153bf7f3263734f31e7afd84f1e48a2f573599/x/blob/README.md?plain=1#L23~~

evan-forbes:
this can supersede #293 and probably #1113

PR #1272, opened 2023-01-19
docs: update x/blob/README.md

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1268


evan-forbes (approved):
thanks! and good idea yeah, in the future we should definitely get in the habit of updating the docs when we change the code

Issue #1230, opened 2023-01-12
Order PFBs after ordinary transactions in block data


rootulp:
## Context

In core/app retro, @cmwaters raised this question: is it desirable to reproduce the block data from the data square? We lose the ability to order the transactions in the block data after we implement https://github.com/celestiaorg/celestia-app/issues/1173 because transactions are split across two separate namespaces so it isn't deterministic how to order them if only given the data square.

## Proposal

If reproducing the block data from the data square is desirable, we can enforce that block producers order all ordinary transactions before all PFB txs in the block data. By enforcing this ordering, a client can reproduce the block data transactions by sequentially reading all transactions (ordinary then PFB txs) in the data square.

PR #1273, opened 2023-01-20
feat: enforce that PFBs are always ordered after normal transactions

cmwaters:
Closes: #1230


app/write_square_test.go
```diff
@@ -171,28 +168,25 @@ func Test_finalizeLayout(t *testing.T) {
 		},
 	}
 	for i, tt := range tests {
-		res, blobs := finalizeLayout(tt.squareSize, tt.nonreserveStart, tt.ptxs)
-		require.Equal(t, len(tt.expectedIndexes), len(res), i)
-		for j, ptx := range res {
-			assert.Equal(t, tt.expectedIndexes[j], ptx.shareIndexes, i)
-		}
+		t.Run(fmt.Sprintf("case%d", i), func(t *testing.T) {
```
evan-forbes, line 171: [super minor non-blocking question]
possibly stupid question, but what is the reasoning behind using t.Run here, but removing it elsewhere?

evan-forbes (approved):
nice! the separation is a lot cleaner

app/write_square_test.go
cmwaters: I don't think I removed it anywhere. I don't think there's any major benefit but out of habit I've always broken these table tests into sub tests. In this case, adding the `fmt.Sprintf("case%d", i)` lets me work out faster what case failed.

app/process_proposal.go
```diff
@@ -81,11 +81,18 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
 		tx := rawTx
 		wrappedTx, isWrapped := coretypes.UnmarshalIndexWrapper(rawTx)
 		if isWrapped {
+			seenFirstPFB = true
 			tx = wrappedTx.Tx
+		} else if seenFirstPFB {
+			logInvalidPropBlock(app.Logger(), req.Header, "Wrapped PFBs must come at the end of the block")
+			return abci.ResponseProcessProposal{
+				Result: abci.ResponseProcessProposal_REJECT,
+			}
```
rootulp, line 95: [no change needed][optional] This works as-is but it adds complexity to a `for` loop that is performing a different check. Since the check for PFB tx ordering seems distinct, it may be extracted like:


```go
	if !arePFBsOrderedAfterTxs(req.BlockData.Txs) {
		logInvalidPropBlock(app.Logger(), req.Header, "Wrapped PFBs must come at the end of the block")
		return abci.ResponseProcessProposal{
			Result: abci.ResponseProcessProposal_REJECT,
		}
	}
```


```go
func arePFBsOrderedAfterTxs(txs [][]byte) bool {
	seenFirstPFB := false
	for _, rawTx := range txs {
		_, isWrapped := coretypes.UnmarshalIndexWrapper(rawTx)
		if isWrapped {
			seenFirstPFB = true
		} else if seenFirstPFB {
			return false
		}
	}
	return true
}

```

rootulp (dismissed):

app/process_proposal.go
cmwaters: Yeah I actually wrote something similar to this at first. The reason why I piggybacked off the for loop is because we're already identifying whether a tx is wrapped or not and I wanted to avoid needlessly unmarshalling the same proto bytes multiple times. 

If I recall correctly, we're already using that for loop for multiple validity checks

rootulp: Agreed, the proposal may be a performance penalty. Without measuring the impact it seems preferable to opt for code clarity over potential performance hit. If needed in the future, we can benchmark and optimize for performance. At that time, it will be easier to optimize the clearer code.

evan-forbes: apologies, I read the diff wrong in one of the test files, I thought we had it and removed it but that's not the case.

cmwaters: Sure I don't mind doing that.

evan-forbes (approved):

Issue #1231, opened 2023-01-13
Revert MultiShareCommitment and include each ShareCommitment in the PFB


evan-forbes:
per review feedback [here](https://github.com/celestiaorg/celestia-app/pull/1154#discussion_r1067993094), we should revert the multi commitment #1196, and simply include each share commitment in the PFB

PR #1275, opened 2023-01-20
revert!: multi share commitment

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1231


x/blob/types/payforblob.go
```diff
@@ -61,10 +61,10 @@ func (msg *MsgPayForBlob) Type() string {
 // ValidateBasic fulfills the sdk.Msg interface by performing stateless
 // validity checks on the msg that also don't require having the actual blob
 func (msg *MsgPayForBlob) ValidateBasic() error {
-	if len(msg.NamespaceIds) != len(msg.ShareVersions) || len(msg.NamespaceIds) != len(msg.BlobSizes) {
+	if len(msg.NamespaceIds) != len(msg.ShareVersions) || len(msg.NamespaceIds) != len(msg.BlobSizes) || len(msg.NamespaceIds) != len(msg.ShareCommitments) {
```
cmwaters, line 64: Do we also want to add a check that this length is not zero?

cmwaters (approved):
lgtm

app/estimate_square_size_test.go
```diff
@@ -88,7 +88,7 @@ func Test_estimateSquareSize_MultiBlob(t *testing.T) {
 			"100 multiblob single share transaction", func() [][]int {
 				return [][]int{blobfactory.Repeat(int(100), 100)}
 			},
-			16, 5,
+			16, 11,
```
evan-forbes, line 91: [simple comment]
perhaps we can eventually include some some option to use a multiblob commit and no namespaces to be more efficient in the future

evan-forbes (approved):
makes sense, LGTM :+1: :+1: other than Callum's comment

x/blob/types/payforblob.go
rootulp: By this length are you referring to msg.ShareCommitments or msg.NamespaceIds? So something like this:

```go
	if len(msg.NamespaceIds) == 0 {
		return ErrNoNamespaceIds
	}
```

or 

```go
	if len(msg.ShareCommitments) == 0 {
		return ErrNoShareCommitments
	}
```

rootulp: IMO both would work. We could include checks that all of these fields are non-zero: NamespaceIds, ShareVersions, BlobSizes, and ShareCommitments. Will do in a follow-up PR.

Issue #1226, opened 2023-01-12
Should `genOrigRowShares` consider padding between the last transaction and the first blob?


rootulp:
Currently, it looks like `pkg/proof/proof.go` doesn't account for padding between the last transaction and the first blob. Since the first blob must start at an index that is aligned with its BlobMinSquareSize, I expect there to be padding in-between the last transaction and the first blob.

https://github.com/celestiaorg/celestia-app/blob/7ad3df7248b406ffb5d22a36628c3455d2f9b78a/pkg/prove/proof.go#L161-L173

evan-forbes:
yes, it needs to, but I'm not sure we will use that implementation in the future. ref https://github.com/celestiaorg/celestia-app/issues/703

Issue #1049, opened 2022-11-22
test: `genRowShares` and `genOrigRowShares`


rootulp:
## Context

There exist two functions in `pkg/prove/proof.go` that do not contain unit tests. https://github.com/celestiaorg/celestia-app/blob/8604be960b0565df66a6c4da79371f713ca62798/pkg/prove/proof.go#L127-L128

https://github.com/celestiaorg/celestia-app/blob/8604be960b0565df66a6c4da79371f713ca62798/pkg/prove/proof.go#L154-L157

There were previously unit tests that can be used for inspiration when working on this issue. The relevant tests were commented out in https://github.com/celestiaorg/celestia-app/pull/692 and removed in https://github.com/celestiaorg/celestia-app/pull/1027

## Proposal

Add unit tests for the functions above

## Related

https://github.com/celestiaorg/celestia-app/issues/703

PR #1276, opened 2023-01-20
feat!: adapt tx inclusion proofs to use share inclusion proofs

rootulp:
Closes https://github.com/celestiaorg/celestia-core/issues/920
Closes https://github.com/celestiaorg/celestia-app/issues/1226 b/c deletes that code
Closes https://github.com/celestiaorg/celestia-app/issues/1049 b/c deletes that code

## Description

Previously the logic to generate tx proofs was flawed because it did not accurately identify which shares a transaction belongs to. Additionally the struct `TxProof` (originally defined by Tendermint but later modified in celestia-core) was   incomplete because it did not contain proofs for the inclusion of a set of rows to the data square root. This PR modifies the TxInclusion logic to return a share inclusion proof instead of a transaction inclusion proof. The share inclusion proof does contain proofs for the rows to the data square root. In order for txs to use share inclusion proofs, `SplitTxs` was updated to keep track of the range of shares that a tx occupies.


rootulp:
Question: should `NewTxInclusionProof` return a proof of type `TxProof`? I read the issue as dropping support for `TxProof` in favor of `SharesProof` but while exploring follow-ups, I realized that celestia-core and tendermint support [`TxProof`](https://github.com/celestiaorg/celestia-core/blob/d0f0650c217916b7fc4ae7c89349dcad0eb42e5c/proto/tendermint/types/types.proto#L185-L190) so we may want celestia-app to support creating them.

If we want celestia-app to continue supporting them: update this PR to retrofit a SharesProof into a TxProof.
If we don't want celestia-app to continue supporting them: update celestia-core to handle SharesProof and delete TxProof.

pkg/prove/proof.go
```diff
@@ -158,101 +115,10 @@ func BlobShareRange(tx types.Tx) (beginShare uint64, endShare uint64, err error)
...
-	rowCount := uint64(len(s)) / squareSize
-	rows := make([][]shares.Share, rowCount)
-	for i := uint64(0); i < rowCount; i++ {
-		rows[i] = s[i*squareSize : (i+1)*squareSize]
-	}
-	return rows
-}
-
-// GenerateSharesInclusionProof generates an nmt inclusion proof for a set of shares to the data root.
+// NewShareInclusionProof returns a an NMT inclusion proof for a set of shares to the data root.
```
SweeXordious, line 118:
```suggestion
// NewShareInclusionProof returns an NMT inclusion proof for a set of shares to the data root.
```

pkg/shares/split_compact_shares.go
```diff
@@ -9,14 +9,26 @@ import (
...
 // increasing set of shares. It is used to lazily split block data such as
 // transactions or intermediate state roots into shares.
 type CompactShareSplitter struct {
 	shares       []Share
 	pendingShare Share
 	namespace    namespace.ID
-	version      uint8
+	shareVersion uint8
+	// shareRanges is a map from a transaction key to the range of shares it
+	// occupies. The range assumpes this compact share splitter is the only
```
SweeXordious, line 28:
```suggestion
	// occupies. The range assumes this compact share splitter is the only
```

SweeXordious (commented):
LGTM :+1: :+1: :+1: :tada:, would defer to other reviewers for approval.

Concerning `TxProof` vs `SharesProof`. `TxProof` currently contains a proof only to the row roots. However, `SharesProof` has shares proof to rows, and rows to data root. 

The thing is, `TxProof`, the way it is right now, is not a full proof, since in traditional chains that tx proof should be to the block root. 

I guess `SharesProof` is more general and it also contains `TxProof`. And making the change wouldn't diverge Celestia-core from tendermint by much (only few files changes).

So, I'm in favor of `TxProof` -> `SharesProof`.

pkg/shares/split_compact_shares.go
```diff
@@ -97,9 +125,9 @@ func (css *CompactShareSplitter) stackPending() {
 }
 
 // Export finalizes and returns the underlying compact shares.
-func (css *CompactShareSplitter) Export() []Share {
+func (css *CompactShareSplitter) Export() (shares []Share, shareRanges map[coretypes.TxKey]ShareRange) {
```
cmwaters, line 128: nit: Does this need to have named returns? They're not really used and the return types are relatively self explanatory

pkg/prove/proof.go
```diff
@@ -22,99 +22,56 @@ import (
...
-		RowRoots: rowRoots,
-		Data:     rawShares,
-		Proofs:   proofs,
+		RowRoots: proof.RowsProof.RowsRoots,
+		Data:     proof.Data,
+		Proofs:   proof.SharesProofs,
 	}, nil
 }
 
+func getTxNamespace(tx types.Tx) (ns namespace.ID, err error) {
```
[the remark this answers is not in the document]
rootulp: good point. Line 56 doesn't return an err b/c the function signature is:
```
func UnmarshalIndexWrapper(tx Tx) (indexWrapper tmproto.IndexWrapper, isIndexWrapper bool) {
```
so I will remove the error return parameter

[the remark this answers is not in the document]
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/1276/commits/c413d7bc3e3eeffc74f1dc212b1cb3f09be0fd09

pkg/prove/proof.go
```diff
@@ -22,99 +22,52 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
-// TxInclusion uses the provided block data to progressively generate rows
-// of a data square, and then using those shares to creates nmt inclusion proofs.
-// It is possible that a transaction spans more than one row. In that case, we
-// have to return more than one proof.
```
cmwaters, line 28: Is this no longer possible?

pkg/shares/compact_shares_test.go
```diff
@@ -20,9 +20,9 @@ func TestCompactShareWriter(t *testing.T) {
 	txs := testfactory.GenerateRandomTxs(33, 200)
 	for _, tx := range txs {
 		rawTx, _ := MarshalDelimitedTx(tx)
-		w.WriteBytes(rawTx)
+		w.write(rawTx)
```
cmwaters, line 23: You could just combine the two to `w.WriteTx(tx)`

pkg/prove/proof.go
```diff
@@ -22,99 +22,52 @@ import (
...
-	prevTxTotalLen := 0
-	for i := uint64(0); i < txIndex; i++ {
-		txLen := len(txs[i])
-		prevTxTotalLen += (shares.DelimLen(uint64(txLen)) + txLen)
-	}
-
-	currentTxLen := len(txs[txIndex])
-	currentTxTotalLen := shares.DelimLen(uint64(currentTxLen)) + currentTxLen
-	endOfCurrentTxLen := prevTxTotalLen + currentTxTotalLen
+	_, _, shareRanges := shares.SplitTxs(data.Txs)
```
cmwaters, line 67: Is there ever the case where we want the shares and the share ranges i.e. what is the case for having both of these values returned by the same method and not having different methods for each? (other than the extra lines of code)

cmwaters (approved):
Few non blocking comments but otherwise LGTM. 

If I understand correctly the main change is to have tx inclusion proofs be the same as the proof of the inclusion of a range of shares that correspond to the tx share position.

evan-forbes (approved):
LGTM


I also appreciate the idea of separating the mapping functionality, but I can see why we keep it all in the same functionality.

pkg/shares/split_compact_shares.go
rootulp: This function doesn't need to have named returns. I like named returns because they serve as additional function signature documentation but can remove and put this information in godoc. TBH godoc should probably have been updated regardless. Do you prefer?

**Option A**
```go
// Export finalizes and returns the underlying compact shares and a map of shareRanges.
func (css *CompactShareSplitter) Export() ([]Share, map[coretypes.TxKey]ShareRange) {
```

**Option B**
```go
// Export finalizes and returns the underlying compact shares and a map of shareRanges.
func (css *CompactShareSplitter) Export() (shares []Share, shareRanges map[coretypes.TxKey]ShareRange) {
```

pkg/prove/proof.go
rootulp: I think it is still possible. I found the godoc confusing because only one `types.TxProof` is returned but this godoc is alluding to the fact that more than one proof may be present at `types.TxProof.Proofs`. I'll make an attempt at clarifying the godoc after we resolve https://github.com/celestiaorg/celestia-core/issues/948

pkg/prove/proof.go
rootulp: Good point. I don't see a scenario where they both need to be returned at the same time. I'll consider extracting to a stand-alone function

pkg/shares/compact_shares_test.go
rootulp: Resolved in https://github.com/celestiaorg/celestia-app/pull/1276/commits/9378d296e66a09619b1a63a8ecca63e739134cc9

pkg/shares/split_compact_shares.go
rootulp: Went with **Option A** in 4e557a0e016abbca3ee28ec984b5437ae2e0f176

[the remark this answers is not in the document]
rootulp: Great idea! Resolved in https://github.com/celestiaorg/celestia-app/pull/1276/commits/a2816b328c4b6ea94d1449d8ce65d77789441370

pkg/prove/proof.go
rootulp: No longer applicable b/c this function returns a singular share proof.

rootulp:
golangci-lint is not happy 😞

pkg/proof/proof.go
```diff
@@ -325,19 +178,19 @@ func GenerateSharesInclusionProof(
...
 
-	return types.SharesProof{
-		RowsProof: types.RowsProof{
-			RowsRoots: rowsRoots,
-			Proofs:    rowsProofs,
-			StartRow:  uint32(startRow),
-			EndRow:    uint32(endRow),
+	return types.ShareProof{
+		RowProof: types.RowProof{
+			RowRoots: rowsRoots,
```
staheri14, line 187: [Nitpick] Considering that the field names are changed to singular format, what do you think about changing the argument names to singular as well i.e., `rowsRoots` -> `rowRoots`. Likewise for `sharesProofs` -> `shareProofs`. Btw, I guess that `Roots` and `Proofs` can also be changed to `Root` and `Proof`?.

pkg/shares/share_splitting.go
```diff
@@ -126,3 +132,10 @@ func SplitBlobs(cursor int, indexes []uint32, blobs []coretypes.Blob, useShareIn
 	}
 	return writer.Export(), nil
 }
+
+func mergeMaps(mapOne, mapTwo map[coretypes.TxKey]ShareRange) map[coretypes.TxKey]ShareRange {
+	merged := make(map[coretypes.TxKey]ShareRange)
```
staheri14, line 137: [non-blocking] we can benefit from increased-performance if we could specify the map capacity at this point i.e., 
```suggestion
	merged := make(map[coretypes.TxKey]ShareRange, len(mapOne)+len(mapTwo))
```

pkg/shares/share_splitting.go
```diff
@@ -126,3 +132,10 @@ func SplitBlobs(cursor int, indexes []uint32, blobs []coretypes.Blob, useShareIn
...
+	maps.Copy(merged, mapOne)
+	maps.Copy(merged, mapTwo)
+	return merged
```
staheri14, line 140: What happens if the two maps have common keys? which one would take precedence? I think it is worth mentioning in the function description.

staheri14 (commented):
Looks good to me! also made some nonblocking suggestions.

pkg/proof/proof.go
rootulp: Good catch. I agree with:
- `rowsRoots` -> `rowRoots`
- `sharesProofs` -> `shareProofs`

Since rowRoots and shareProofs are slices, I'm inclined to keep them plural.

rootulp: resolved in https://github.com/celestiaorg/celestia-app/pull/1276/commits/76ee56b0b2d4789ff7a6ac8eddc0a300ef97a191

pkg/shares/share_splitting.go
rootulp: Good point. It looks like the second map's duplicate key would take precedence. Resolved in https://github.com/celestiaorg/celestia-app/pull/1276/commits/24f3d420859875283f6b31853700bd499c90c5d5

pkg/prove/proof.go
rootulp: I took a stab at extracting a separate function like this:

```go
func TxShareRanges(txs coretypes.Txs) (shareRanges map[coretypes.TxKey]ShareRange) {
	txWriter := NewCompactShareSplitter(appconsts.TxNamespaceID, appconsts.ShareVersionZero)
	pfbTxWriter := NewCompactShareSplitter(appconsts.PayForBlobNamespaceID, appconsts.ShareVersionZero)

	for _, tx := range txs {
		if _, isIndexWrapper := coretypes.UnmarshalIndexWrapper(tx); isIndexWrapper {
			pfbTxWriter.WriteTx(tx)
		} else {
			txWriter.WriteTx(tx)
		}
	}

	txShares, txMap := txWriter.Export(0)
	_, pfbMap := pfbTxWriter.Export(len(txShares))

	return mergeMaps(txMap, pfbMap)
}
```

It shares a lot of the same implementation with SplitTxs but makes it so that the SplitTxs function signature can be simpler:

```diff
-func SplitTxs(txs coretypes.Txs) (txShares []Share, pfbShares []Share, shareRanges map[coretypes.TxKey]ShareRange) {
+func SplitTxs(txs coretypes.Txs) (txShares []Share, pfbShares []Share, shareRanges) {
```

the downside to this approach is that we have duplicate logic in the bodies of these functions and I'm nervous that the implementation of `SplitTxs` evolves without a corresponding update to `TxShareRanges` - does that concern make sense?

rootulp:
> If I understand correctly the main change is to have tx inclusion proofs be the same as the proof of the inclusion of a range of shares that correspond to the tx share position.

Exactly @cmwaters

pkg/shares/split_compact_shares_test.go
```diff
@@ -111,10 +111,145 @@ func TestExport(t *testing.T) {
...
+				css.write(bytes)
 			}
-			got := css.Export()
+			got, _ := css.Export(0)
 			assert.Equal(t, tc.want, got)
 		})
 	}
 }
+
+func TestExport(t *testing.T) {
```
evan-forbes, line 122: thanks for this!!

evan-forbes (approved):
nice! thanks for finally getting rid of the old code, it feels so much better w/o it

[only kinda related comment]
I'm still thinking about if we want to support proving multiple blobs at once. presumably just wrapping many ShareProofs, but perhaps we just shouldn't support that, and people can do that themselves if they need it.

rootulp:
Given we support multiple blobs in a single transaction, I think celestia-app should also consider supporting a single proof for the multiple blobs associated with a single tx. I think this would involve some form of multiple ShareProofs (one per share range that a blob occupies).

Issue #1280, opened 2023-01-23
`IsCompactShare` should return true for shares in the PFB tx namespace


rootulp:
https://github.com/celestiaorg/celestia-app/blob/1aec7233a34f05254043616f3c1a6daa8e556512/pkg/shares/shares.go#L54-L57

PR #1281, opened 2023-01-23
fix: IsCompactShare for PFB tx shares

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1280


cmwaters (approved):
Looks good.

Do we also want to modify the `merge` function in `shares/share_merging.go` to accommodate for the PFB namespace? I think that's the only other place that's missing

evan-forbes (approved):

Issue #1285, opened 2023-01-24
`merge` doesn't handle PayForBlobNamespaceID


rootulp:
Looks good.

Do we also want to modify the `merge` function in `shares/share_merging.go` to accommodate for the PFB namespace? I think that's the only other place that's missing

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/pull/1281#pullrequestreview-1267217197_

rootulp:
`merge` doesn't appear used anywhere outside of `shares_test` and it isn't exported so it seems safe to delete entirely. 

https://github.com/celestiaorg/celestia-app/blob/570dc411d481f05967fbc147f071a316ba687215/pkg/shares/share_merging.go#L14

evan-forbes:
> merge doesn't appear used anywhere outside of shares_test and it isn't exported so it seems safe to delete entirely

I feel like we've discussed this a few times before but haven't dug up our past decisions. IIRC, while its not currently used in core/app, we do want to preserve this functionality because being able to go from the square to block data will be needed in the future, specicically for different types of partial nodes, and we should already have all of the pieces to do that for each reserved namespaces and blobs, so ideally there shouldn't be that much new necessary logic.

While the current roundtrip unit test/fuzzer for merging has not been maintained, it is a very convenient way to fuzz the share encoding. We create random block data, split it, then merge it and compare the input and output for equality.

I can see some reasoning for removing it, but I don't think us not using something now is water tight. If we do want to remove it, then I think we need to write down why we are removing it and in what cases we would add it. Not wanting to maintain the round trip test, or provide any guarantees that we even can do so, until we need the functionality is a perfectly good reason imo. We just need to agree  that is the reasoning, and document that decision.

PR #1286, opened 2023-01-24
chore: delete share `merge`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1285

If CI passes, this seems safe to me

### DIFF
--- a/pkg/shares/share_merging.go
+++ b/pkg/shares/share_merging.go
@@ -6,63 +6,8 @@ import (
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/nmt/namespace"
-	"github.com/celestiaorg/rsmt2d"
 	coretypes "github.com/tendermint/tendermint/types"
 )
-
-// merge extracts block data from an extended data square.
-func merge(eds *rsmt2d.ExtendedDataSquare) (coretypes.Data, error) {
-	squareSize := eds.Width() / 2
-
-	// sort block data shares by namespace
-	var (
-		sortedTxShares   [][]byte
-		sortedBlobShares [][]byte
-	)
-
-	// iterate over each row index
-	for x := uint(0); x < squareSize; x++ {
-		// iterate over each share in the original data square
-		row := eds.Row(x)
-
-		for _, share := range row[:squareSize] {
-			// sort the data of that share types via namespace
-			nid := share[:appconsts.NamespaceSize]
-			switch {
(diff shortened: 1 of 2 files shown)


pkg/shares/share_merging.go
```diff
@@ -6,64 +6,9 @@ import (
```
cmwaters, line 6: Should we rename this file?

pkg/shares/shares_test.go
```diff
@@ -90,21 +38,6 @@ func TestPadFirstIndexedBlob(t *testing.T) {
...
-		case <-ctx.Done():
-			return
-		default:
-			TestMerge(t)
-		}
-	}
-}
-
 // generateRandomBlockData returns randomly generated block data for testing purposes
 func generateRandomBlockData(txCount, blobCount, maxSize int) (data coretypes.Data) {
```
cmwaters, line 42: Looks like this is also unused now

cmwaters (commented):
Need to appease the linter but otherwise looks good

evan-forbes (changes requested):
cross posting https://github.com/celestiaorg/celestia-app/issues/1285#issuecomment-1403500853

> merge doesn't appear used anywhere outside of shares_test and it isn't exported so it seems safe to delete entirely

I feel like we've discussed this a few times before but haven't dug up our past decisions. IIRC, while its not currently used in core/app, we do want to preserve this functionality because being able to go from the square to block data will be needed in the future, specicically for different types of partial nodes, and we should already have all of the pieces to do that for each reserved namespaces and blobs, so ideally there shouldn't be that much new necessary logic.

While the current roundtrip unit test/fuzzer for merging has not been maintained, it is a very convenient way to fuzz the share encoding. We create random block data, split it, then merge it and compare the input and output for equality.

I can see some reasoning for removing it, but I don't think us not using something now is water tight in this very specific scenario. If we do want to remove it, then I think we need to write down why we are removing it and in what cases we would add it. Not wanting to maintain the round trip test, or provide any guarantees that we even can do so, until we need the functionality is a perfectly good reason imo. We just need to agree that is the reasoning, and document that decision.

rootulp:
I synced w/ @evan-forbes offline:
- If we maintain the `merge` function, it should be fixed, tested, and publicly exported
- If we don't want to maintain `merge` function, remove it and associated utils. 

We're leaning towards preserving it at the moment because the fuzz tester that uses it is still helpful.

Issue #1254, opened 2023-01-17
Document QGB verify command usage


SweeXordious:
As discussed under https://github.com/celestiaorg/celestia-app/pull/1236#pullrequestreview-1252389704, we should write some documentation over the usage of the verification command.

PR #1287, opened 2023-01-24
docs: add QGB verify command docs

SweeXordious:
## Overview
Closes https://github.com/celestiaorg/celestia-app/issues/1254

We could add more documentation after executing the command to give an example output, but I guess it's better to do it after we have mocked QGB deployed so that we have the same output as users.

## Checklist

### DIFF
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -150,7 +150,7 @@ Use "celestia-appd verify [command] --help" for more information about a command
 
 It currently supports three sub-commands:
 
-- `blob`: Takes a transaction hash, in hex format, and verifies that the blob that has been posted using it has been committed to, by the QGB contract. It only supports one message for now.
+- `blob`: Takes a transaction hash, in hex format, and verifies that the blob that has been posted using it has been committed to, by the QGB contract. It only supports one blob for now.
 - `shares`: Takes a shares range and a height, and verifies that these former have been committed to by the QGB contract.
 - `tx`: Takes a transaction hash, in hex format, and verifies that it has been committed to by the QGB contract.
 
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -150,7 +150,7 @@ Use "celestia-appd verify [command] --help" for more information about a command
 
 It currently supports three sub-commands:
 
-- `blob`: Takes a transaction hash, in hex format, and verifies that the blob that has been posted using it has been committed to, by the QGB contract. It only supports one blob for now.
+- `blob`: Takes a transaction hash, in hex format, and verifies that the blob paid for by the transaction has been committed to by the QGB contract. It only supports one blob for now.
 - `shares`: Takes a shares range and a height, and verifies that these former have been committed to by the QGB contract.
 - `tx`: Takes a transaction hash, in hex format, and verifies that it has been committed to by the QGB contract.
 
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -120,7 +120,7 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
 
 ## Verification command
 
-The QGB verification command is part of the `Celestia-appd` binary. It allows to verify that a set of shares has been posted to a specific QGB contract.
+The QGB verification command is part of the `celestia-appd` binary. It allows the user to verify that a set of shares has been posted to a specific QGB contract.
 
 ```shell
 $ celestia-appd verify --help          
--- a/x/qgb/README.md
+++ b/x/qgb/README.md
@@ -125,13 +125,13 @@ The QGB verification command is part of the `celestia-appd` binary. It allows th
 ```shell
 $ celestia-appd verify --help          
                               
-Verifies that a transaction hash, a range of shares, or a message referenced by its transaction hash were committed to by the QGB contract
+Verifies that a transaction hash, a range of shares, or a blob referenced by its transaction hash were committed to by the QGB contract
 
 Usage:
   celestia-appd verify [command]
 
 Available Commands:
-  msg         Verifies that a message, referenced by its transaction hash, in hex format, has been committed to by the QGB contract. Only supports one message for now
+  blob        Verifies that a blob, referenced by its transaction hash, in hex format, has been committed to by the QGB contract. Only supports one blob for now
   shares      Verifies that a range of shares has been committed to by the QGB contract
   tx          Verifies that a transaction hash, in hex format, has been committed to by the QGB contract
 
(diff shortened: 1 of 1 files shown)


x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
 }
 ```
 
+## Verification command
+
+The QGB verification command is part of the `Celestia-appd` binary. It allows to verify that a set of shares has been posted to a specific QGB contract.
```
rootulp, line 123:
```suggestion
The QGB verification command is part of the `celestia-appd` binary. It allows the user to verify that a set of shares has been posted to a specific QGB contract.
```

x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
...
+$ celestia-appd verify --help          
+                              
+Verifies that a transaction hash, a range of shares, or a message referenced by its transaction hash were committed to by the QGB contract
+
+Usage:
+  celestia-appd verify [command]
+
+Available Commands:
+  msg         Verifies that a message, referenced by its transaction hash, in hex format, has been committed to by the QGB contract. Only supports one message for now
+  shares      Verifies that a shares range has been committed to by the QGB contract
```
rootulp, line 135:
```suggestion
  shares      Verifies that a range of shares has been committed to by the QGB contract
```

x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
...
+      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
+      --trace               print out full stack trace on errors
+
+Use "celestia-appd verify [command] --help" for more information about a command.
+
+```
+
+It currently supports three sub-commands:
+
+- `msg`: Takes a transaction hash, in hex format, and verifies that the message that has been posted using it has been committed to, by the QGB contract. It only supports one message for now.
```
rootulp, line 153: [question] how is this different from the `tx` subcommand? Does it verify that the blob shares associated with a BlobTx were committed to? If so should this use blob instead of msg?

SweeXordious: the `tx` verifies that the transaction carrying the blob was committed to, in the tx namespace.

> If so should this use blob instead of msg

100%, thanks a lot for spotting these. I prepared these PRs before the renaming and I missed renaming some places, will change now.

SweeXordious: https://github.com/celestiaorg/celestia-app/pull/1289 updating the command too

x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
...
+      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
+      --trace               print out full stack trace on errors
+
+Use "celestia-appd verify [command] --help" for more information about a command.
+
+```
+
+It currently supports three sub-commands:
+
+- `blob`: Takes a transaction hash, in hex format, and verifies that the blob that has been posted using it has been committed to, by the QGB contract. It only supports one blob for now.
```
rootulp, line 153:
```suggestion
- `blob`: Takes a transaction hash, in hex format, and verifies that the blob paid for by the transaction has been committed to by the QGB contract. It only supports one blob for now.
```

x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
...
+- `shares`: Takes a shares range and a height, and verifies that these former have been committed to by the QGB contract.
```
rootulp, line 154:
```suggestion
- `shares`: Takes a range of shares and a height, and verifies that these shares have been committed to by the QGB contract.
```

rootulp (dismissed):

x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
 }
 ```
 
+## Verification command
+
+The QGB verification command is part of the `celestia-appd` binary. It allows the user to verify that a set of shares has been posted to a specific QGB contract.
+
+```shell
+$ celestia-appd verify --help
```
rootulp, line 126: [non-blocking] This code snippet may become out of date with the true `--help` docs so instead of copy + pasting the entire help output, I think it may be preferable to include the text that isn't the code snippet and at the bottom say something like: 

See the following command for more details


```shell
celestia-appd verify --help
```

and then ensure that these docs make it into the appropriate `--help` docs

rootulp (approved):

x/qgb/README.md
SweeXordious: What do you think is better for docs, providing the commands but without an output (that might be outdated after some time), or the commands + their output?

x/qgb/README.md
```diff
@@ -118,6 +118,42 @@ if k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx)+1 != n
...
+      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
+      --trace               print out full stack trace on errors
+
+Use "celestia-appd verify [command] --help" for more information about a command.
+
+```
+
+It currently supports three sub-commands:
+
+- `blob`: Takes a transaction hash, in hex format, and verifies that the blob paid for by the transaction has been committed to by the QGB contract. It only supports one blob for now.
```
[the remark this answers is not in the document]
SweeXordious: Which prefix?  A transaction hash in hex format is something like: `7826D5310CD007F82B918359D74A853B4ECB6633E9F231C5477B7F21D68D642D` if I am not missing something

x/qgb/README.md
rootulp: > What do you think is better for docs, providing the commands but without an output, or the commands + their output (that might be outdated after some time)?

In this scenario: providing the commands but without an output. I think this because the goal of the README is to give developers a high level overview of the QGB module. Developers aren't necessarily users of the QGB module CLI. Users of the QGB CLI should be able to get information on CLI flags via `--help` and I expect them to try that before finding this README.

Also (repeated from above): the command output becomes a maintenance burden but I defer to you @SweeXordious

SweeXordious: I guess we can leave it as it is for now, with the help output. And, if it becomes a burden later, we can remove it.
Thanks guys for your help :+1: :+1:

x/qgb/README.md
evan-forbes, line 153: > It only supports one blob for now.

does this mean that it only supports proving to one blob at a time, or only being supports PFBs that pay for a single blob?

evan-forbes (approved):
nice, :+1:

might also want to link to nmt/rsmt2d/specs, and eventually a diagram would likely be useful to show precising what inclusion proofs are being used for which proof since we're using two or three different proofs

x/qgb/README.md
SweeXordious: It means that if a PFB has one blob, it will prove it. If it has multiple blobs, then it will prove the first one.
Extending it to prove multiple blobs should be no biggy, I guess there is already an issue for that.

SweeXordious: > might also want to link to nmt/rsmt2d/specs, and eventually a diagram would likely be useful to show precising what inclusion proofs are being used for which proof since we're using two or three different proofs

https://github.com/celestiaorg/celestia-app/issues/1313

SweeXordious:
:shipit:

PR #1298, opened 2023-01-25
chore: handle PayForBlobNamespaceID in `merge`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1285

## Description
1. Handle PayForBlobNamespaceID in `merge`
1. Extract `merge` to a new file
1. Extract `merge` to a new file
1. Add a unit test that uses sample block data


pkg/shares/testdata/sample-block.json
```diff
@@ -0,0 +1 @@
...F0cEiKT4ZVvPErTGckYfEpAuTd1JbNekrTAT3Gdy9E97LKxBGAJiP7dcfRuZKaVtt62lSkOM9dK61YYY7TlRRsG31YV04Q4eoF1Zxh81xRyV19OjIjngHzYuG7KUnmt6uMc/aQjplIEfQXud7PYm8CP4DBaqd6cAgclmsW0XkLmHk04JTe7Vd6xaPAjhK+H/2Z59tiDAYOxUqybAPy/Un8p3kf9WfkqLVzQZviOJYoIEXsQBtAA3P1f3Cs0V8NZQZcZmHCL/cCCJ6V+V3sEmhUl1nblrTgBmpfdlZRcASB20XgFXT6LQDimSYv9ycrMwpJXRxHLEgnbCo72mrBh4p2opBbfZBLkhAnkqfV/mjS/sxAZ4kKmASHM/Wbqqia2mo6FIvk9OeyVwU1v15v5iGwDd77uGB95MfcA/7qXqgPrUxaZUl9fQ8R6kcqvWIXOu8+KbSm1IB58VRxnXYFFlcRoBr4lTWHPrBivtArOSIZ6fZ/HY9Dw9oj0lACaXxZEtkLl/vDNxPEx5B8MN+onhOzgLccK5q8RBGf8hA9ebClWvRmLvYT/+yTcvOmwPWgSs6nOYprckWc47tia2350CXU6EPCnzECSE/X3AcNnwgli2Z1T+vUbYrBr3jrUjmQpujFOsuJZV9oaAfXXPlN+Lsz4I7bil4kyomQ/w3q3gTQfWmmkU="}],"square_size":4,"hash":"uPWH3TVoZNb3NcyXN/vbX5N1mDZ8dfXMh9bhne04CpM="},"evidence":{"evidence":[]},"last_commit":{"height":3,"block_id":{"hash":"CnTXMWDvJXaHOjecWXb2nGvLsJ23GmgUbAYIZmteP2M=","part_set_header":{"total":1,"hash":"PgAxB6DYr8huhDZJWMnFSgYDCXwuF3jI1GNbnOskv9U="}},"signatures":[{"block_id_flag":2,"validator_address":"XbENo47oRQwd9YSP9vr+10SttTc=","timestamp":"2023-01-30T18:55:27.534092Z","signature":"CPafxD85F5yNxKP/9hm2ew4WLPRIo472AcVNMgOZh239V4AnKsRpL8FUwS14X48DpsZuzVCqLgUKY2+agupxBQ=="}]}}
```
evan-forbes, line 1: how was this testdata generated? how do we regenerate it? 

admittedly, I'm not the biggest fan of using test data because the tests that rely on this could pass indefinitely, even if something in the block changes. imo, we should at least have a function to regenerate it a la https://github.com/celestiaorg/celestia-node/blob/3a1353c0d122e64b991983cf889530e2c7747d10/state/integration_test.go#L120-L147 and even then we'd still want some CI or something to notify us that it changed :grimacing: 

if we want to avoid import cycles, then I would prefer that we have those tests in a test package, but this is just a preference and doesn't have to be done in this PR

pkg/shares/reconstruct.go
```diff
@@ -0,0 +1,72 @@
...
+	"bytes"
+
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/rsmt2d"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+// reconstruct reconstructs block data from an extended data square.
+// TODO: export this function
+func reconstruct(eds *rsmt2d.ExtendedDataSquare) (coretypes.Data, error) {
```
evan-forbes, line 13: [not blocking]
I understand changing this name, but it might be a tad confusing since celestia-node using the term reconstruction for a very different thing. 

Also, it deviates from the existing splitting merging terminology. Not going to block on it though, so will defer to you on that.

evan-forbes (commented):
mind pointing out any changes involved in the fix? its a bit difficult to distinguish since we're moving code around.

if we wanted to get a little fancy, we could also modify an existing "fuzzing" integration test that is already generating random block data, to also test reconstruction.

pkg/shares/reconstruct.go
```diff
@@ -0,0 +1,72 @@
...
+		row := eds.Row(x)
+
+		for _, share := range row[:squareSize] {
+			// sort the data of that share types via namespace
+			nid := share[:appconsts.NamespaceSize]
+			switch {
+			case bytes.Equal(appconsts.TxNamespaceID, nid):
+				sortedTxShares = append(sortedTxShares, share)
+			case bytes.Equal(appconsts.PayForBlobNamespaceID, nid):
+				sortedPfbTxShares = append(sortedPfbTxShares, share)
```
rootulp, line 35: [for reviewers] this is the fix

pkg/shares/reconstruct.go
rootulp: > Also, it deviates from the existing splitting merging terminology.

That was my intention because splitting and merging sound like inverse operations when they aren't
- **Split** converts block data into shares
- **Parse** converts shares into txs, blobs, or share sequences
- **Merge** or **Reconstruct** converts an extended data square into block data 

I can revert the rename if reviewers don't have alternative proposals

pkg/shares/testdata/sample-block.json
rootulp: > how was this testdata generated? how do we regenerate it?

https://github.com/celestiaorg/celestia-node/blob/3a1353c0d122e64b991983cf889530e2c7747d10/state/integration_test.go#L120-L147

> how do we regenerate it?

I can add that to celestia-app

pkg/shares/reconstruct.go
rootulp: Resolved by reverting the rename.

rootulp:
> if we wanted to get a little fancy, we could also modify an existing "fuzzing" integration test that is already generating random block data, to also test reconstruction.

I'm concerned the existing "fuzzing" test that tests reconstruction doesn't actually generate pay for blob txs [here](https://github.com/rootulp/celestia-app/blob/a42b83dac1627f428201ca98cea5914205ca5bfe/pkg/shares/merge_test.go#L101)

evan-forbes:
> I'm concerned the existing "fuzzing" test that tests reconstruction doesn't actually generate pay for blob txs [here](https://github.com/rootulp/celestia-app/blob/a42b83dac1627f428201ca98cea5914205ca5bfe/pkg/shares/merge_test.go#L101)I'm concerned the existing "fuzzing" test that tests reconstruction doesn't actually generate pay for blob txs [here](https://github.com/rootulp/celestia-app/blob/a42b83dac1627f428201ca98cea5914205ca5bfe/pkg/shares/merge_test.go#L101)


whoops, I should have pointed to the test I was thinking of, yeah I think we should test this using blocks that are using more realistic data, like with this test.

https://github.com/celestiaorg/celestia-app/blob/main/app/test/fuzz_abci_test.go

evan-forbes (approved):

evan-forbes:
created a follow up for this PR https://github.com/celestiaorg/celestia-app/issues/1322, in which case I think we're clear to merge

rootulp:
Thanks for creating the follow-up issue. Will merge this as-is because it contains a fix.

Issue #1114, opened 2022-12-13
Consolidate and refactor testing code


evan-forbes:
We still have a lot of work needed around consolidating our test utility code. 

one instance is https://github.com/celestiaorg/celestia-app/pull/1089#discussion_r1045911931

similar to #829

evan-forbes:
we might be able to consolidate some utility code to create the gensis file in a few testing utility packages

see in testnode, test, and in #1914

Issue #233, opened 2022-03-08
Implement end to end tests that attempt to propose invalid blocks


evan-forbes:
With the introduction of ABCI++ #231, we are adding checks to proposal blocks. while we have tests to ensure that invalid proposals get rejected, we don't have an end to end test for this specific case.

evan-forbes:
one strategy to accomplish this in both an integration or e2e test is to simply wrap the application and overwrite the prepare proposal method

evan-forbes:
can close as completed with the addition of the `malicious` package, which adds the ability to propose and sign invalid blocks.

ensuring that `ProcessProposal` is always called is a different issue, and the rest of the test cases are covered by existing correctness testing

PR #1300, opened 2023-01-26
feat!: Add a block validity rule that all PFB transactions must have valid signatures

evan-forbes:
## Overview

This PR adds a signature check in ProcessProposal for PFB transactions.

## Checklist

### DIFF
--- a/app/test/fuzz_abci_test.go
+++ b/app/test/fuzz_abci_test.go
@@ -8,8 +8,10 @@ import (
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/testutil"
 	"github.com/celestiaorg/celestia-app/testutil/blobfactory"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
 	core "github.com/tendermint/tendermint/proto/tendermint/types"
 	coretypes "github.com/tendermint/tendermint/types"
 )
@@ -25,7 +27,12 @@ func TestPrepareProposalConsistency(t *testing.T) {
 		t.Skip("skipping TestPrepareProposalConsistency in short mode.")
 	}
 	encConf := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-	testApp, _ := testutil.SetupTestAppWithGenesisValSet()
+	accounts := make([]string, 10000)
+	for i := range accounts {
+		accounts[i] = tmrand.Str(20)
+	}
+
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(accounts...)
 
 	type test struct {
 		name                   string
@@ -45,7 +52,7 @@ func TestPrepareProposalConsistency(t *testing.T) {
 				case <-timer:
 					return
 				default:
-					ProcessRandomProposal(t, tt.count, tt.size, tt.blobCount, encConf, testApp)
+					ProcessRandomProposal(t, tt.count, tt.size, tt.blobCount, encConf, testApp, accounts, kr, "")
 				}
 			}
 		})
@@ -59,9 +66,27 @@ func ProcessRandomProposal(
 	maxBlobCount int,
 	cfg encoding.Config,
 	capp *app.App,
+	accounts []string,
+	kr keyring.Keyring,
+	chainid string,
 ) {
-	txs := blobfactory.RandBlobTxsRandomlySized(cfg.TxConfig.TxEncoder(), count, maxSize, maxBlobCount)
-	sendTxs := blobfactory.GenerateManyRawSendTxs(cfg.TxConfig, count)
+	txs := testutil.RandBlobTxsWithAccounts(
+		t,
+		capp,
+		cfg.TxConfig.TxEncoder(),
+		kr,
+		maxSize,
+		maxBlobCount,
+		true,
+		chainid,
+		accounts[:count],
+	)
+	sendTxs := blobfactory.GenerateManyRawSendTxsWithAccounts(
+		cfg.TxConfig,
+		kr,
+		accounts[:count],
+		chainid,
+	)
 	txs = append(txs, sendTxs...)
 	resp := capp.PrepareProposal(abci.RequestPrepareProposal{
 		BlockData: &core.Data{
--- a/app/test/process_proposal_test.go
+++ b/app/test/process_proposal_test.go
@@ -21,20 +21,47 @@ import (
 )
 
 func TestProcessProposal(t *testing.T) {
-	testApp, _ := testutil.SetupTestAppWithGenesisValSet()
+	accounts := make([]string, 100)
+	for i := range accounts {
+		accounts[i] = tmrand.Str(20)
+	}
+
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(accounts...)
 	encConf := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+
+	testTxs := testutil.RandBlobTxsWithAccounts(
+		t,
+		testApp,
+		encConf.TxConfig.TxEncoder(),
+		kr,
+		1000,
+		4,
+		false,
+		"",
+		accounts,
+	)
 
 	// block with all blobs included
 	validData := func() *core.Data {
 		return &core.Data{
-			Txs: coretypes.Txs(blobfactory.RandBlobTxs(encConf.TxConfig.TxEncoder(), 4, 1000)).ToSliceOfBytes(),
+			Txs: coretypes.Txs(testTxs).ToSliceOfBytes(),
 		}
 	}
 
 	// create block data with a PFB that is not indexed and has no blob
 	unindexedData := validData()
-	blobtx := blobfactory.RandBlobTxs(encConf.TxConfig.TxEncoder(), 1, 1000)[0]
-	btx, _ := coretypes.UnmarshalBlobTx(blobtx)
+	blobtx := testutil.RandBlobTxsWithAccounts(
+		t,
+		testApp,
+		encConf.TxConfig.TxEncoder(),
+		kr,
+		1000,
+		2,
+		false,
+		"",
+		accounts[:1],
+	)
+	btx, _ := coretypes.UnmarshalBlobTx(blobtx[0])
 	unindexedData.Txs = append(unindexedData.Txs, btx.Tx)
 
 	// create block data with a tx that is random data, and therefore cannot be
@@ -125,17 +152,19 @@ func TestProcessProposal(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		resp := testApp.PrepareProposal(abci.RequestPrepareProposal{
-			BlockData: tt.input,
+		t.Run(tt.name, func(t *testing.T) {
+			resp := testApp.PrepareProposal(abci.RequestPrepareProposal{
+				BlockData: tt.input,
+			})
+			tt.mutator(resp.BlockData)
+			res := testApp.ProcessProposal(abci.RequestProcessProposal{
+				BlockData: resp.BlockData,
+				Header: core.Header{
+					DataHash: resp.BlockData.Hash,
+				},
+			})
+			assert.Equal(t, tt.expectedResult, res.Result, tt.name)
 		})
-		tt.mutator(resp.BlockData)
-		res := testApp.ProcessProposal(abci.RequestProcessProposal{
-			BlockData: resp.BlockData,
-			Header: core.Header{
-				DataHash: resp.BlockData.Hash,
-			},
-		})
-		assert.Equal(t, tt.expectedResult, res.Result, tt.name)
 	}
 }
 
--- a/app/test/process_proposal_test.go
+++ b/app/test/process_proposal_test.go
@@ -60,8 +60,8 @@ func TestProcessProposal(t *testing.T) {
 		false,
 		"",
 		accounts[:1],
-	)
-	btx, _ := coretypes.UnmarshalBlobTx(blobtx[0])
+	)[0]
+	btx, _ := coretypes.UnmarshalBlobTx(blobtx)
 	unindexedData.Txs = append(unindexedData.Txs, btx.Tx)
 
 	// create block data with a tx that is random data, and therefore cannot be
--- a/go.mod
+++ b/go.mod
@@ -184,7 +184,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.6.0-sdk-v0.46.7
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.6.0-sdk-v0.46.7.0.20230124110235-fd313811a3ef // v1.6.0-sdk-v0.46.7
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.13.0-tm-v0.34.23
 )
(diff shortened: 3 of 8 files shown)


app/test/fuzz_abci_test.go
```diff
@@ -25,7 +27,12 @@ func TestPrepareProposalConsistency(t *testing.T) {
 		t.Skip("skipping TestPrepareProposalConsistency in short mode.")
 	}
 	encConf := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-	testApp, _ := testutil.SetupTestAppWithGenesisValSet()
+	accounts := make([]string, 10000)
```
evan-forbes, line 30: we might be able to decrease this back to something smaller, which is desired since using this many different accounts makes this test about 10-20 seconds slower locally

go.mod
```diff
@@ -184,7 +184,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.6.0-sdk-v0.46.7
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.6.0-sdk-v0.46.7.0.20230126051657-d4a340f4a10b
```
evan-forbes, line 187: still need to update this after we merge and bumb the cosmos-sdk PR

testutil/manual_blobtx_gen.go
```diff
@@ -0,0 +1,177 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+// RandBlobTxsWithAccounts will create random blob transactions using the
+// provided configuration. The account info is queried directly from the
+// application. One blob transaction is generated per account provided.
+func RandBlobTxsWithAccounts(
```
evan-forbes, line 20: adding to our testing technical debt a bit by continuing to copy paste minor alterations of this function. #1114 

we have to keep this in the testutil pkg instead of blob factory because we're importing the app to query directly, and don't want to cause an import cycle in blobfactory

cmwaters (dismissed):
Everything here makes sense to me.

Valid signatures don't mean valid PFBs. I'm not sure how important the distinction is for the users that requested this. But just want to make sure that they're aware that although all PFBs have valid signatures they may be missing other things.

evan-forbes:
> Valid signatures don't mean valid PFBs. I'm not sure how important the distinction is for the users that requested this. But just want to make sure that they're aware that although all PFBs have valid signatures they may be missing other things.

yes good point this is definitely true. In this case, we only want to be able to make an honest majority assumption that the signatures are valid, because it dramatically reduces the proving time of some zk proofs. 

This is only the minimum to close #979, 
> In practice this could mean that we only add an additional signature check, but we could also check that there are enough funds to pay gas... 

if we want other features such as fee burning, that will essentially require executing the transactions in full before we progress

cmwaters:
I believe the default implementation of the SDK v0.47 (i.e. with the introduction of PrepareProposal and ProcessProposal) by default calls something similar to CheckTx on all transactions in `ProcessProposal`

evan-forbes:
to clarify things a bit, we are also already enforcing the validate basic checks, that only a single MsgPFB is in a PFB transaction, and that the share commitment is valid. Therefore, I believe the only other way an invalid PFB can be introduced is if the account does not have enough funds to pay for gas.

cmwaters:
> Therefore, I believe the only other way an invalid PFB can be introduced is if the account does not have enough funds to pay for gas

Or because of an incorrect nonce

evan-forbes:
> Or because of an incorrect nonce

I believe we are checking for that in this PR since that is required to verify the signature via the SigVerifyAnteDecorator

cmwaters:
> I believe we are checking for that in this PR since that is required to verify the signature via the SigVerifyAnteDecorator

I see, but is the cached state tracked and updated? 

Take the following scenarios:
- A user sends a vote message and a PFB with the same nonce number (5 for example). Upon execution only the first message will be executed. The second will be invalidated by the execution of the first. However both would pass CheckTx.
- A user sends a vote message and a PFB message with incrementing nonce numbers (i.e. 5 and 6). In that order both should be able to be executed in the same block. But if we check from a snapshot of state, the second message with a nonce of 6 will be invalid.

evan-forbes:
that's a good point! its not, but needs to be!

we might be able to get away with added the sequence increment decorator. I'll go add a test and try

cmwaters:
We could potentially add a small antedecorator after everything that calls something like `acc.SetSequence(sig.Sequence + 1)`. The context is already a branch of state so state mutations won't impact the actual state machine

evan-forbes:
> We could potentially add a small antedecorator after everything that calls something like acc.SetSequence(sig.Sequence + 1). The context is already a branch of state so state mutations won't impact the actual state machine

yeah, I was thinking about just adding the increment nonce decorator on everything, since I don't think there should ever be a case where the nonce doesn't increase, but would need to double check. 

> However both would pass CheckTx

going back to this point quickly, assuming we're running standard code, we should be updating the checktxState with any remaining block data. Does this mean that this shouldn't happen? I know its not good enough for what we need, but do we need to check for this in PrepareProposal as well?

cmwaters:
> going back to this point quickly, assuming we're running standard code, we should be updating the checktxState with any remaining block data. Does this mean that this shouldn't happen? I know its not good enough for what we need, but do we need to check for this in PrepareProposal as well?

I'm a bit confused by what you mean here. What remaining block data are you talking about.

evan-forbes:
sorry, didn't get good sleep :sweat_smile: 

not remaining block data, and txs that are in the mempool that didn't get included in the block

cmwaters:
> not remaining block data, but txs that are in the mempool that didn't get included in the block

checkTxState will get updated when `Commit` get's called. If `Recheck` is enabled all remaining txs in the mempool will get `CheckTx`ed. If they're nonce is now invalid they will be removed. Does that answer your question or am I going off on something else :)

cmwaters:
Yes unfortunately `CheckTx` is not sufficient for `PrepareProposal`

app/prepare_proposal.go
```diff
@@ -25,6 +25,22 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
+	// increment the sequences of the standard cosmos-sdk transactions. Panics
+	// from the anteHandler are caught and logged.
+	isHandler := incrementSequenceAnteHandler(&app.AccountKeeper)
+	normalTxs, sdkCtx = filterStdTxs(app.Logger(), app.txConfig.TxDecoder(), sdkCtx, isHandler, normalTxs)
+
+	// check the signatures and increment the sequences of the blob transations,
+	// and filter out any that fail. Panics from the anteHandler are caught and
+	// logged.
+	svHandler := sigVerifyAnteHandler(&app.AccountKeeper, app.txConfig)
+	blobTxs, _ = filterBlobTxs(app.Logger(), app.txConfig.TxDecoder(), sdkCtx, svHandler, blobTxs)
```
evan-forbes, line 42: here is where we add stateful checks to prepare proposal. notice that we are incrementing the nonce of the normal txs first before we are checking the signatures of the PFB txs.

app/test/fuzz_abci_test.go
```diff
@@ -25,54 +25,68 @@ func TestPrepareProposalConsistency(t *testing.T) {
...
+
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(accounts...)
 
 	type test struct {
 		name                   string
 		count, blobCount, size int
 	}
 	tests := []test{
-		{"many small single share single blob transactions", 10000, 1, 400},
+		{"many small single share single blob transactions", 1000, 1, 400},
```
evan-forbes, line 40: reduced the length of this first test because it takes so long that sometimes we don't even run it once. changing this means that, at least in this test, are no longer testing for the weird things that can happen when using a bunch of small PFBs occasionally. We might still want to do that in the future, but I'm less worried now that we are limiting the number of txs to 5000.

app/test/fuzz_abci_test.go
```diff
@@ -25,54 +25,68 @@ func TestPrepareProposalConsistency(t *testing.T) {
...
+					sendTxs := testutil.SendTxsWithAccounts(
+						t,
+						testApp,
+						encConf.TxConfig.TxEncoder(),
+						kr,
+						1000,
+						accounts[0],
+						accounts[len(accounts)-100:],
+						"",
+					)
```
evan-forbes, line 74: this refactor was needed cause now we need access to state to query for account info before we can create the txs

app/test/prepare_proposal_test.go
```diff
@@ -80,42 +100,14 @@ func TestPrepareProposalBlobSorting(t *testing.T) {
 		res := testApp.PrepareProposal(tt.input)
 		assert.Equal(t, tt.expectedBlobs, res.BlockData.Blobs)
 		assert.Equal(t, tt.expectedTxs, len(res.BlockData.Txs))
-
-		// verify the signatures of the prepared txs
-		sdata, err := signer.GetSignerData()
-		require.NoError(t, err)
-
-		dec := encoding.IndexWrapperDecoder(encCfg.TxConfig.TxDecoder())
-		for _, tx := range res.BlockData.Txs {
```
evan-forbes, line 89: this portion of this test was leftover from when we were malleating txs for multiple square sizes, so we can remove. It no longer works since this portion of the test only works when we don't need access to state.

app/test/prepare_proposal_test.go
```diff
@@ -168,15 +162,40 @@ func TestPrepareProposalOverflow(t *testing.T) {
 
 func TestPrepareProposalPutsPFBsAtEnd(t *testing.T) {
 	numBlobTxs, numNormalTxs := 3, 3
+	accnts := testfactory.GenerateAccounts(numBlobTxs + numNormalTxs)
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(accnts...)
```
evan-forbes, line 166: all of our prepare/process proposal tests need to sign txs with correct nonces and account numbers, therefore we have to change most of the tests to account for this

app/parse_txs.go
```diff
@@ -20,3 +23,66 @@ func separateTxs(txcfg client.TxConfig, rawTxs [][]byte) ([][]byte, []core.BlobT
 	}
 	return normalTxs, blobTxs
 }
+
+// filterStdTxs applies the provided antehandler to each transaction and removes
+// transactions that return an error. Panics are caught by the checkTxValidity
+// function used to apply the ante handler.
+func filterStdTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handler sdk.AnteHandler, txs [][]byte) ([][]byte, sdk.Context) {
```
evan-forbes, line 30: I didn't add unit tests for these, since those tests are essentially identical to the `TestPrepareProposalFiltering`, in both what we test and structure of the test itself

app/test/process_proposal_test.go
```diff
@@ -154,59 +270,6 @@ func TestProcessProposal(t *testing.T) {
 	}
 }
 
-func TestProcessProposalWithParityShareNamespace(t *testing.T) {
```
evan-forbes, line 157: moved this test up to the other test, see comment there to see why we probably don't even need this test unless we have some code that let's us bypass these checks in prepare proposal #233

app/test/process_proposal_test.go
```diff
@@ -154,59 +270,6 @@ func TestProcessProposal(t *testing.T) {
...
-	resp.BlockData.Blobs[0].NamespaceId = appconsts.ParitySharesNamespaceID
-
-	input := abci.RequestProcessProposal{
-		BlockData: resp.BlockData,
-	}
-	res := testApp.ProcessProposal(input)
-	require.Equal(t, abci.ResponseProcessProposal_REJECT, res.Result)
-}
-
-func TestProcessProposalWithTamperedSequenceStart(t *testing.T) {
```
evan-forbes, line 179: consolidated this test in the above test

app/test/process_proposal_test.go
```diff
@@ -216,3 +279,11 @@ func flipSequenceStart(share shares.Share) shares.Share {
 	share[infoByteIndex] = share[infoByteIndex] ^ 0x01
 	return share
 }
+
+func deref[T any](s []*T) []T {
```
evan-forbes, line 283: go really needs simple generic funcs like these in the standard lib or something

app/verify_txs.go
```diff
@@ -0,0 +1,41 @@
...
+
+// recoverHandler will simply wrap the caught panic in an error containing the
+// stack trace.
+func recoverHandler(recoveryObj interface{}) error {
+	return sdkerrors.Wrap(
+		sdkerrors.ErrPanic, fmt.Sprintf(
+			"recovered: %v\nstack:\n%v", recoveryObj, string(debug.Stack()),
+		),
+	)
+}
```
evan-forbes, line 41: the nonce increment decorator panics if the account isn't present, so we have to add this. It should never get hit if block producers are honest and are running check tx

testutil/direct_tx_gen.go
```diff
@@ -0,0 +1,272 @@
...
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+// RandBlobTxsWithAccounts will create random blob transactions using the
+// provided configuration. The account info is queried directly from the
+// application. One blob transaction is generated per account provided.
+func RandBlobTxsWithAccounts(
+	t *testing.T,
+	capp *app.App,
```
evan-forbes, line 24: We need to sign txs with valid sequences and account numbers, and have to query for those w/o grpc if we want to use the testapp, so I created even more testing functions (that we should consolidate in #1114) to query the app directly

evan-forbes (commented):
most of the changes introduced here are testing related since we added stateful requirements to both prepare and process proposal, and many of the test txs we were using before didn't have valid signatures or nonce usage.

rootulp (approved):
I don't have any blocking feedback. I do have a question: IIUC the motivation for this PR was to verify the validity of signatures on PFB transactions so that once a transaction is included in a block, we know that its signature was valid.

Now that prepare proposal and process proposal are verifying the sequence number and signature of PFB transactions, should we consider (at a later date) extending that to apply to all types of  transactions?

app/prepare_proposal.go
```diff
@@ -25,6 +25,22 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
 
+	sdkCtx, err := app.NewProcessProposalQueryContext()
+	if err != nil {
+		panic(err)
+	}
+
+	// increment the sequences of the standard cosmos-sdk transactions. Panics
+	// from the anteHandler are caught and logged.
+	isHandler := incrementSequenceAnteHandler(&app.AccountKeeper)
+	normalTxs, sdkCtx = filterStdTxs(app.Logger(), app.txConfig.TxDecoder(), sdkCtx, isHandler, normalTxs)
```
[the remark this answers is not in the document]
evan-forbes: yeah, but I like your suggestion for seq better

app/test/prepare_proposal_test.go
```diff
@@ -189,3 +211,142 @@ func TestPrepareProposalPutsPFBsAtEnd(t *testing.T) {
...
+	validTxs := func() [][]byte {
+		txs := make([][]byte, 0, len(sendTxs)+len(blobTxs))
+		txs = append(txs, blobTxs...)
+		txs = append(txs, sendTxs...)
+		return txs
+	}
+
+	// create 3 MsgSend transactions that are using the same sequence as the
+	// first three blob transactions above
+	invalidSendTxs := coretypes.Txs(testutil.SendTxsWithAccounts(
```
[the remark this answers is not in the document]
evan-forbes: good idea https://github.com/celestiaorg/celestia-app/pull/1300/commits/d508e9efa7d9c619cd8bc9370b8999a3b88cb59d

rootulp (approved):

rootulp (approved):

PR #1301, opened 2023-01-26
PoC: iterative and deterministic square construction

cmwaters:
This is a proof of concept PR meant to help illustrate an architectural change. This passes the fuzzy tester but needs more tests and benchmarks and many of the internal methods need to be cleaned up.

This PR proposes the addition of a square package responsible for square construction. There are two main functions `Construct` and `Reconstruct` called by `PrepareProposal` and `ProcessProposal` respectively. Rather than estimating the square size, the `Constructor` iteratively allocates space based on the maximum worst case padding for each individual blob (as `MinSquareSize(blobShareLen) - 1`).

Rather than dropping the blobs with the last namespace, it drops based on the lowest priority PFBs. There are no "padding gaps" from earlier connected blobs dropped in the same PFB. 

The functions require only the raw ordered tendermint transactions to construct, increasing the effectiveness of compact blocks. The deterministic construction means that square size doesn't necessarily need to be added (but could still be for the sake of light clients). It also theoretically means that our ABCI methods PrepareProposal and ProcessProposal are no different from the canonical version and we could look at removing the SDK fork after v0.37.

Downsides: Blob order within a namespace is dictated by the priority of the PFBS associated with it. Validators don't have the ability to individually order blobs within a namespace. However, in the worst case, Blobs can include metadata that indicates the order with which they should be interpreted.

cc: @nashqueue @evan-forbes


nashqueue:
I would love for you to comment on the possible code paths and the if/else, what kind of assumptions you make at each code path as I think some assumptions are wrong, for example, the canFit() Function.

cmwaters:
Closing this PR. It has been superceded by https://github.com/celestiaorg/celestia-app/pull/1690, https://github.com/celestiaorg/celestia-app/pull/1659 and https://github.com/celestiaorg/celestia-app/pull/1657

Issue #1221, opened 2023-01-12
Rename `MsgPayForBlob` to `MsgPayForBlobs`


evan-forbes:
In #1154, we added the functionality to pay for multiple blobs using a single PFB, so we should change the name of `MsgPayForBlob` to `MsgPayForBlobs`. We didn't want to handle this in #1154 as that PR was already quite beefy.

PR #1304, opened 2023-01-26
chore!: rename MsgPayForBlob to MsgPayForBlobs

cmwaters:
Closes: #1221

This breaks a lot of the APIs. If it's important to us that we don't do that, then I'd suggest instead having both types of messages i.e. `MsgPayForBlob` and `MsgPayForBlobs`


proto/blob/tx.proto
```diff
@@ -9,13 +9,13 @@ option go_package = "github.com/celestiaorg/celestia-app/x/blob/types";
 // Msg defines the blob Msg service.
 service Msg {
   // PayForBlob allows the user to pay for the inclusion of a blob
```
rootulp, line 11:
```suggestion
  // PayForBlobs allows the user to pay for the inclusion of one or more blobs
```

proto/blob/tx.proto
```diff
@@ -30,4 +30,4 @@ message MsgPayForBlob {
 
 // MsgPayForBlobResponse describes the response returned after the submission
 // of a PayForBlob
```
rootulp, line 32:
```suggestion
// MsgPayForBlobsResponse describes the response returned after the submission
// of a PayForBlobs
```

x/blob/keeper/keeper.go
```diff
@@ -49,7 +49,7 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 }
 
 // PayForBlob consumes gas based on the blob size.
```
rootulp, line 51:
```suggestion
// PayForBlobs consumes gas based on the blob sizes in the MsgPayForBlobs.
```

x/blob/types/payforblob.go
```diff
@@ -51,16 +51,16 @@ func NewMsgPayForBlob(signer string, blobs ...*Blob) (*MsgPayForBlob, error) {
...
 
 // Type fulfills the sdk.Msg interface
-func (msg *MsgPayForBlob) Type() string {
-	return URLMsgPayForBlob
+func (msg *MsgPayForBlobs) Type() string {
+	return URLMsgPayForBlobs
 }
 
 // ValidateBasic fulfills the sdk.Msg interface by performing stateless
 // validity checks on the msg that also don't require having the actual blob
```
rootulp, line 62:
```suggestion
// ValidateBasic fulfills the sdk.Msg interface by performing stateless
// validity checks on the msg that also don't require having the actual blob(s)
```

rootulp (commented):
Overall LGTM. A few godocs need to be updated

evan-forbes (commented):
nice! this gets the vast majority of them.

some of these might be redundant to @rootulp's suggestions, but we have a few things I think we can update either before merging, or we can leave open an issue for

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/x/blob/README.md#L28

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/app/prepare_proposal.go#L33-L35 

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/app/write_square.go#L23

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/app/prepare_proposal.go#L34-L35

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/app/write_square.go#L23

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/x/blob/README.md#L24-L25

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/x/blob/README.md#L28

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/x/blob/types/errors.go#L25-L30

https://github.com/celestiaorg/lazyledger-app/blob/988e5b00689fa4075d01659f4da8761b646f7014/x/blob/types/errors.go#L13

cmwaters:
Ok I've gone through a second pass, trying to pick up on all the cases that @evan-forbes and @rootulp have outlined. There is one struct I haven't touched which is the `EventPayForBlob`. Should I also rename this?

evan-forbes (approved):
will defer to you about the event, we can also just change in a different PR

cmwaters:
Okay I went with adding the events. It seemed to make sense

evan-forbes (approved):

Issue #1305, opened 2023-01-27
Use versioned proto packages


cmwaters:
### Summary

Add v1 suffixes to the proto packages for `blob` and `qgb` modules so we can have versioned messages. Whenever a message undergoes a modification that changes the way the state machine executes it, we should then bump that version. 

This is alignment with the way SDK messages are versioned. 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1306, opened 2023-01-27
chore!: use versioned proto packages

cmwaters:
Closes: #1305

I'm not sure if we also want to version the `da` package

## Checklist


proto/celestia/da/data_availability_header.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
-package da;
+package celestia.da;
 
-option go_package = "github.com/celestiaorg/celestia-app/proto/da";
+option go_package = "github.com/celestiaorg/celestia-app/proto/celestia/da";
```
SweeXordious, line 4: [Question]
Why did we add the `/celestia/` to the `go_package` for DA and to for others?
IMO, the `celestia` is a bit redundant in the `da` package as we're already in the Celestia repo:
```
daproto "github.com/celestiaorg/celestia-app/proto/celestia/da"
```

SweeXordious (commented):
:+1: :+1:

proto/celestia/da/data_availability_header.proto
cmwaters: Because in the other packages we generate to the types directory but here we just generate the go code to the same directory

SweeXordious (dismissed):
LGTM :+1: :+1: 

Would defer to other reviewers for final approval

evan-forbes (changes requested):
we just need to update https://github.com/celestiaorg/celestia-app/blob/f0488e3c9e2d734d622b6d959ad7b1e0470d4c00/x/blob/types/payforblob.go#L22

and we should be good to go

x/blob/types/payforblob.go
```diff
@@ -18,8 +18,7 @@ import (
 )
 
 const (
-	URLBlobTx         = "/blob.BlobTx"
```
cmwaters, line 21: This isn't used anywhere so I removed it (cc @evan-forbes)

x/blob/types/payforblob_test.go
```diff
@@ -110,6 +111,10 @@ func TestCreateCommitment(t *testing.T) {
 	}
 }
 
+func TestMsgTypeURLParity(t *testing.T) {
+	require.Equal(t, sdk.MsgTypeURL(&MsgPayForBlobs{}), URLMsgPayForBlobs)
+}
```
cmwaters, line 116: Added a little test to make sure these two values are in parity. I think in v0.47 of the SDK we no longer will need to do this.

evan-forbes (approved):
only the nit from the linter is needed now

(we really need to add gofumpt to `make lint`)

MSevey (approved):

evan-forbes (approved):

evan-forbes:
merging for the feature freeze release candidate

PR #1309, opened 2023-01-29
chore: update to golangci-lint 1.50.1

rootulp:
Update to the latest release of golangci-lint: https://github.com/golangci/golangci-lint/releases while I try debugging an issue with golangci-lint failures on a different PR


SweeXordious (approved):

rootulp:
This wasn't the root cause of the issue but I don't see a downside to upgrading.

PR #1325, opened 2023-02-01
feat: add `LastValsetRequestBeforeNonce` rpc call

SweeXordious:
## Overview

This RPC call was left out somewhere when adapting the state for V2. This PR introduces it back.

## Checklist


proto/celestia/qgb/v1/query.proto
```diff
@@ -31,6 +31,15 @@ service Query {
       returns (QueryLatestAttestationNonceResponse) {
     option (google.api.http).get = "/qgb/v1/attestations/nonce/latest";
   }
+  // LastValsetRequestBeforeNonce Queries last Valset request before nonce.
```
staheri14, line 34: What does `Valset` stand for? is it validator set? can we please add the full term in the description as well

staheri14 (commented):
Looks good to me! I made a few minor suggestions for improvement. But, since I don't have all the details, I'll leave it for others to give the final approval.

proto/celestia/qgb/v1/query.proto
SweeXordious: It's a type that describes validator sets:
https://github.com/celestiaorg/celestia-app/blob/4f9deb07b72e34f5eeeda87ef2a70f45f05c1c3f/proto/celestia/qgb/v1/types.proto#L17-L29

proto/celestia/qgb/v1/query.proto
```diff
@@ -31,6 +31,15 @@ service Query {
...
+  // And, even if the current nonce is a valset, if will return the previous
```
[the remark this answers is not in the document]
SweeXordious: good catch :+1: 
[749bbdb](https://github.com/celestiaorg/celestia-app/pull/1325/commits/749bbdb047a3a63e935eb70c6b15e2cf24264c41)

x/qgb/keeper/keeper_valset.go
```diff
@@ -130,7 +130,7 @@ func normalizeValidatorPower(rawPower uint64, totalValidatorPower cosmosmath.Int
 
 // GetLastValsetBeforeNonce returns the previous valset before the provided `nonce`.
 // the `nonce` can be a valset, but this method will return the valset before it.
-// If the provided nonce is 1. It will return an error. Because, there is no valset before nonce 1.
+// If the provided nonce is 1, it will return an error, because, there is no valset before nonce 1.
 func (k Keeper) GetLastValsetBeforeNonce(ctx sdk.Context, nonce uint64) (*types.Valset, error) {
```
cmwaters, line 134: Shouldn't the second argument be `*QueryLastValsetRequestBeforeNonceRequest`? Or where does the actual query get called to

SweeXordious: That's the keeper. It's called in the querier here: https://github.com/celestiaorg/celestia-app/blob/4f9deb07b72e34f5eeeda87ef2a70f45f05c1c3f/x/qgb/keeper/query_valset.go#L13-L16

That change is just a docs change that Sanaz spotted and I fixed it in this PR (instead of opening a new one)

evan-forbes (approved):
so if I'm understanding correctly, the querier was already defined, we just missed the proto definitions somewhere along the line? 

do we need some basic unit tests to ensure basic functionality? like, perhaps we could include this in this [test](https://github.com/celestiaorg/celestia-app/blob/main/app/test/std_sdk_test.go), or in a similar one, where we spin up a node and query using grpc

SweeXordious:
> so if I'm understanding correctly, the querier was already defined, we just missed the proto definitions somewhere along the line?

Yes 100%

> do we need some basic unit tests to ensure basic functionality? like, perhaps we could include this in this [test](https://github.com/celestiaorg/celestia-app/blob/main/app/test/std_sdk_test.go), or in a similar one, where we spin up a node and query using grpc

Good idea, https://github.com/celestiaorg/celestia-app/issues/1343

PR #1333, opened 2023-02-03
chore!: bump `nmt` from v0.13 to v0.14 and remove the codec from `NewTxInclusionProof`

Bidon15:
## Overview

Latest v0.12.0-rc5 is not compliant with what node will have to upcoming v0.7.0-rc`x`, as celestia-node is using nmt `v0.14.0` and dropped v0.13.0

This PR fixes this, hence I propose to make a `v0.12.0-rc6` afterwards 🙏 
In addition, this PR cleans up codec from the signature that is not used in `NewTxInclusionProof ` anymore

## Checklist

### DIFF
--- a/pkg/proof/proof.go
+++ b/pkg/proof/proof.go
@@ -15,7 +15,6 @@ import (
 	blobmodule "github.com/celestiaorg/celestia-app/x/blob"
 	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
 	"github.com/celestiaorg/nmt/namespace"
-	"github.com/celestiaorg/rsmt2d"
 	"github.com/tendermint/tendermint/crypto/merkle"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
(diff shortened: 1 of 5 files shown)


cmwaters (dismissed):
LGTM but I think you need to appease the linter first

pkg/proof/proof.go
```diff
@@ -24,7 +23,7 @@ import (
 
 // NewTxInclusionProof returns a new share inclusion proof for the given
 // transaction index.
-func NewTxInclusionProof(codec rsmt2d.Codec, data types.Data, txIndex uint64) (types.ShareProof, error) {
+func NewTxInclusionProof(data types.Data, txIndex uint64) (types.ShareProof, error) {
```
evan-forbes, line 26: thanks for removing this since we weren't using this!

out of curiosity, was there any reason in node to remove this as well?

evan-forbes (approved):

pkg/proof/proof.go
Bidon15: To be honest, I have no idea even when node removed this 😓

PR #1336, opened 2023-02-04
feat: override consensus configs

### LINKED ISSUE #1326: Override configured consensus timeouts
as discussed elsewhere and synchronously, having the consensus timeouts in the config is an easy way for validators and network participants to make mistakes. We can avoid some of this by overriding the configured timeouts, and using hardcoded ones by default. This way, full nodes won't accidentally use inconsistent timeouts by reusing an old config.

evan-forbes:
## Overview

This PR introduces a function that overrides configs when starting the in process tendermint node from the cli. This approach has the benefit of keeping the override logic in the application but still allows for all CI and tests to run as fast as we want.

close #1326 

## Checklist

### DIFF
--- a/cmd/celestia-appd/cmd/overrides.go
+++ b/cmd/celestia-appd/cmd/overrides.go
@@ -1,0 +1,15 @@
+package cmd
+
+import (
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/cosmos/cosmos-sdk/server"
+	"github.com/spf13/cobra"
+)
+
+func overrideServerConfig(command *cobra.Command) error {
+	ctx := server.GetServerContextFromCmd(command)
+	ctx.Config.Consensus.TimeoutPropose = appconsts.TimeoutPropose
+	ctx.Config.Consensus.TimeoutCommit = appconsts.TimeoutCommit
+	ctx.Config.Consensus.SkipTimeoutCommit = false
+	return server.SetCmdServerContext(command, ctx)
+}
--- a/cmd/celestia-appd/cmd/overrides.go
+++ b/cmd/celestia-appd/cmd/overrides.go
@@ -6,6 +6,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// overrideServerConfig applies overrides to the embedded server context's
+// configurations.
 func overrideServerConfig(command *cobra.Command) error {
 	ctx := server.GetServerContextFromCmd(command)
 	ctx.Config.Consensus.TimeoutPropose = appconsts.TimeoutPropose
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	qgbcmd "github.com/celestiaorg/celestia-app/x/qgb/client"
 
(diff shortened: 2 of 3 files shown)


rootulp (approved):

cmd/celestia-appd/cmd/overrides.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/spf13/cobra"
+)
+
+// overrideServerConfig applies overrides to the embedded server context's
+// configurations.
+func overrideServerConfig(command *cobra.Command) error {
+	ctx := server.GetServerContextFromCmd(command)
+	ctx.Config.Consensus.TimeoutPropose = appconsts.TimeoutPropose
+	ctx.Config.Consensus.TimeoutCommit = appconsts.TimeoutCommit
+	ctx.Config.Consensus.SkipTimeoutCommit = false
```
cmwaters, line 15: Isn't this default false anyway? (not that I have anything against being explicit)

cmd/celestia-appd/cmd/root.go
```diff
@@ -92,13 +93,19 @@ func NewRootCmd() *cobra.Command {
...
 
 			customAppTemplate, customAppConfig := initAppConfig()
-			return server.InterceptConfigsPreRunHandler(cmd, customAppTemplate, customAppConfig, tmCfg)
+
+			err = server.InterceptConfigsPreRunHandler(cmd, customAppTemplate, customAppConfig, tmCfg)
+			if err != nil {
+				return err
+			}
+
+			return overrideServerConfig(cmd)
```
cmwaters, line 108: How come we're setting the consensus config stuff twice i.e. here and on lines 96 and 97

evan-forbes: here we're overriding whatever the values configured by the user are, but the first time is a simple nice to have so that the config is accurate.

cmd/celestia-appd/cmd/overrides.go
evan-forbes: yeah I believe so. Also, this was just copied from before, but as you said, being explicit isn't the worst thing.

cmwaters:
What would happen if we wanted to change those values i.e. one of them was too low and we were going into multiple rounds. Would we issue a new version just with the changes?

evan-forbes:
> Would we issue a new version just with the changes?

atm, yeah. I feel like block times should be though, as everyone should be using the same values. either that, or perhaps we could make them consensus parameters, and remove them from the config, not unlike block size?

I could be wrong though, do you know of a reason for users to configure this value?

cmwaters:
I agree about ensuring that everyone has the same values. It's just that leaving it configurable means we could make an announcement to validators to set the value to a new value. Consensus Params is one option. Also the chain wouldn't have halted if we incorrectly set these times. It just means it might take 2 or 3 rounds to reach consensus (which is more of a performance impact)

evan-forbes:
> It's just that leaving it configurable means we could make an announcement to validators to set the value to a new value

I was considering this as well, and initially thought that having to deploy a new patch would be worth the trade off of being more confident that people are using the correct timeouts. That's my opinion atm though, if we don't agree on that, then we definitely don't have to merge this :slightly_smiling_face:

cmwaters:
No I actually tend to agree. It's easier to announce hey upgrade to v2 then to say, everyone go to your `config.toml`, update this param, and then restart your node

cmwaters (approved):

Issue #1136, opened 2022-12-19
How to determine if a share is namespaced padding?


rootulp:
## Context

Unlike tail padding or transaction tail padding shares, a client can't use the namespace of a namespaced padding share to determine that the share is padding. In order for a client to determine if a namespaced padding share contains significant data, it must download and parse all of the preceding shares in the data square until it reaches the first share of the preceeding blob. The client can then use the sequence length in that share to determine which bytes of data are part of the blob and which are part of namespaced padding and can therefore be ignored.

## Proposal 

If we want clients to be able to download and parse a namespaced padding share without the need for downloading additional shares, we may consider: reserving one bit in the info byte to indicate if this is a namespaced padded share. If the bit is set to `1`, all remaining bytes are guaranteed to be `0`. The client can then treat the share as padding. For completeness: the sequence start indicator in the info byte would be set to `0`. The sequence length would be zeroed out.

rootulp:
Tagging @evan-forbes @adlerjohn @musalbas for feedback

rootulp:
I'm going to spin out an issue to update the specs with the implemented design of namespaced padding shares and modify this issue to contain the proposal a revised design of the namespaced padding share format.

musalbas:
Can't the client also know that a share is a padding share by pre-computing the hash of padding shares within a namespace?
cc @adlerjohn

musalbas:
>The client can't unless it knows which shares are not-padding, since there's nothing in the padding shares to indicate they're padding. This requires information that's not in the padding shares in question.

Can't they know which shares are not-padding if the share hash does not match the pre-computed root for padding in that namespace?

musalbas:
>If we want to support rollup nodes that don't look at PFB transactions at all, then yes I think you'll need to add metadata at the beginning of the padding to say how many shares are padding.

I don't think you do. Consider a model where a full storage node does not send padding shares to a rollup light node over the wire, but simply tells them "hey, shares between index x to y are padding, and here's the sub-tree root of those shares". The light node can then recompute the sub-tree root of those padding shares itself, to authenticate that statement. Do you agree?

Issue #783, opened 2022-09-24
Consider treating namespaced padding as blobs or sequences


evan-forbes:
With the non-interactive defaults, we introduced padding between messages. Per the original specs [`Caveats`](https://github.com/celestiaorg/celestia-specs/blob/e59efd63a2165866584833e91e1cb8a6ed8c8203/src/rationale/message_block_layout.md#caveats) section of Message Layout Rational, these are described as 

> [Namespaced padded shares] have a value of zero and a namespace ID equal to the preceding message's. Since their value is known, they can be omitted from NMT proofs of all shares of a given namespace ID.

We should consider updating the implementation and specs to treat the each section of padding as individual messages. This will prepend an info byte along with a length delimiter.

edit: ~~While the added bytes would break the second property,~~
> Since their value is known, they can be omitted from NMT proofs of all shares of a given namespace ID.

this might mean that light clients could skip these shares entirely when parsing a namespace.

musalbas:
Afaik this would only break the second property for the first padding share, as it has different message lengths.

rootulp:
> this might mean that light clients could skip these shares entirely when parsing a namespace.

In order for this to be the case, additional metadata must be added to the share so that a client can differentiate between a sequence of namespaced padding shares and a sequence of shares with `0` data. See https://github.com/celestiaorg/celestia-app/issues/1136

musalbas:
Assuming that all padding shares have a start indicator of 1, and data len of 0, then shouldn't you be able to differentiate padding shares from non-padding shares, because only padding shares can have a start indicator of 1 and data len of 0?

rootulp:
Good point. celestia-app denies blob sizes of `0` [here](https://github.com/celestiaorg/celestia-app/blob/0b447adf776689cc8f14823c6ef925b8f8249b5a/x/blob/types/payforblob.go#L208-L210) so it is possible to treat a share with a sequence start indicator of `1` and a sequence length of `0` as a namespaced padding share. 

Pros:
- an additional reserved bit is not needed

Cons: 
- prevents us from writing a sequence length for the number of padding bytes that follow to the first namespaced padded share. I briefly considered that option so that clients could read the first namespaced padded share and learn how many padding shares followed and potentially skip downloading them. An alternative optimization with similar benefit may  be possible because a namespaced padded share has a variable namespace id and constant contents.

musalbas:
[...]

PR #1341, opened 2023-02-06
feat!: differentiate namespaced padded shares

rootulp:
Implements the proposal in https://github.com/celestiaorg/celestia-app/issues/783#issuecomment-1419007954. 

Closes https://github.com/celestiaorg/celestia-app/issues/1136
Closes https://github.com/celestiaorg/celestia-app/issues/783
Opens https://github.com/celestiaorg/celestia-app/issues/1344

### DIFF
--- a/pkg/shares/namespaced_padding.go
+++ b/pkg/shares/namespaced_padding.go
@@ -2,6 +2,7 @@ package shares
 
 import (
 	"bytes"
+	"encoding/binary"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/nmt/namespace"
@@ -12,15 +13,20 @@ import (
 // conforms to non-interactive default rules. The ns parameter provided should
 // be the namespace of the blob that precedes this padding in the data square.
 func NamespacedPaddedShare(ns namespace.ID) Share {
-	infoByte, err := NewInfoByte(appconsts.ShareVersionZero, false)
+	infoByte, err := NewInfoByte(appconsts.ShareVersionZero, true)
 	if err != nil {
 		panic(err)
 	}
-	padding := bytes.Repeat([]byte{0}, appconsts.ShareSize-appconsts.NamespaceSize-appconsts.ShareInfoBytes)
+
+	sequenceLen := make([]byte, appconsts.SequenceLenBytes)
+	binary.LittleEndian.AppendUint32(sequenceLen, uint32(0))
+
+	padding := bytes.Repeat([]byte{0}, appconsts.ShareSize-appconsts.NamespaceSize-appconsts.ShareInfoBytes-appconsts.SequenceLenBytes)
 
 	share := make([]byte, 0, appconsts.ShareSize)
 	share = append(share, ns...)
 	share = append(share, byte(infoByte))
+	share = append(share, sequenceLen...)
 	share = append(share, padding...)
 	return share
 }
@@ -32,4 +38,17 @@ func NamespacedPaddedShares(ns namespace.ID, n int) []Share {
 		shares[i] = NamespacedPaddedShare(ns)
 	}
 	return shares
+}
+
+func IsNamespacedPadded(s Share) (bool, error) {
+	isSequenceStart, err := s.IsSequenceStart()
+	if err != nil {
+		return false, err
+	}
+	sequenceLen, err := s.SequenceLen()
+	if err != nil {
+		return false, err
+	}
+
+	return isSequenceStart && sequenceLen == 0, nil
 }
--- a/pkg/shares/namespaced_padding.go
+++ b/pkg/shares/namespaced_padding.go
@@ -19,7 +19,7 @@ func NamespacedPaddedShare(ns namespace.ID) Share {
 	}
 
 	sequenceLen := make([]byte, appconsts.SequenceLenBytes)
-	binary.LittleEndian.AppendUint32(sequenceLen, uint32(0))
+	binary.BigEndian.PutUint32(sequenceLen, uint32(0))
 
 	padding := bytes.Repeat([]byte{0}, appconsts.ShareSize-appconsts.NamespaceSize-appconsts.ShareInfoBytes-appconsts.SequenceLenBytes)
 
--- a/pkg/shares/namespaced_padding.go
+++ b/pkg/shares/namespaced_padding.go
@@ -21,7 +21,7 @@ func NamespacedPaddedShare(ns namespace.ID) Share {
 	sequenceLen := make([]byte, appconsts.SequenceLenBytes)
 	binary.BigEndian.PutUint32(sequenceLen, uint32(0))
 
-	padding := bytes.Repeat([]byte{0}, appconsts.ShareSize-appconsts.NamespaceSize-appconsts.ShareInfoBytes-appconsts.SequenceLenBytes)
+	padding := bytes.Repeat([]byte{0}, appconsts.FirstSparseShareContentSize)
 
 	share := make([]byte, 0, appconsts.ShareSize)
 	share = append(share, ns...)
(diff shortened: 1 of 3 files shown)


cmwaters (commented):
LGTM mostly a few nits.

I wonder why both sparse shares and compact shares are the same version. To me, the versioning indicates a different way of interpreting an array of 512 bytes

pkg/shares/parse_sparse_shares.go
```diff
@@ -31,6 +31,14 @@ func parseSparseShares(rawShares [][]byte, supportedShareVersions []uint8) (blob
 			return nil, fmt.Errorf("unsupported share version %v is not present in supported share versions %v", version, supportedShareVersions)
 		}
 
+		isNamespacedPadded, err := IsNamespacedPadded(share)
+		if err != nil {
+			return nil, err
+		}
+		if isNamespacedPadded {
+			continue
+		}
```
evan-forbes, line 40: [question]
what is the current mechanism for skipping namespace padding shares? if there is an old mechanism, do we need to remove it? is there a way to test that this is actually getting hit? do we even need such a test?

evan-forbes (commented):
thanks for doing this so fast!!

~~doesn't have to be in this PR, but we should keep a reminder somewhere to update the spec if we don't have one already.~~

edit: shoot, sorry should have read the intro comment :facepalm:

pkg/shares/namespaced_padding.go
```diff
@@ -33,3 +39,16 @@ func NamespacedPaddedShares(ns namespace.ID, n int) []Share {
 	}
 	return shares
 }
+
+func IsNamespacedPadded(s Share) (bool, error) {
```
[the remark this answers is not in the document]
rootulp: Yes great idea 🤦‍♂️ for not including in initial PR

pkg/shares/namespaced_padding.go
```diff
@@ -12,15 +13,20 @@ import (
 // conforms to non-interactive default rules. The ns parameter provided should
 // be the namespace of the blob that precedes this padding in the data square.
 func NamespacedPaddedShare(ns namespace.ID) Share {
```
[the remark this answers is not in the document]
rootulp: > How does this work, if for example you wanted 3 shares of namespaced padding after a blob. Are they all considered separate sparse shares or would the latter two be continuation sparse shares

If this PR is accepted, then 3 shares of namespaced padding after a blob would be 3 separate sequences. All 3 shares would be sparse shares.

> Does it make sense to add a quantity argument to make multiple shares

We have a separate function for that: https://github.com/celestiaorg/celestia-app/blob/a138c19a2215747e11c48bd3ef810bbca29bd93a/pkg/shares/namespaced_padding.go#L28-L29

rootulp:
> I wonder why both sparse shares and compact shares are the same version. To me, the versioning indicates a different way of interpreting an array of 512 bytes

They don't strictly need to be the same version but at the moment, it's possible to differentiate compact shares and sparse shares via the share's namespace ID so they don't need to be separate versions. I think it's a little clearer to not version them separately because consider the following scenario:

### Time `t=0` 
- share version `0` is for compact shares 
- share version `1` is for sparse shares

### Time `t=1` 
We want to make a breaking change to compact shares so we introduce
- share version `2` is for compact shares 
- share version `1` is for sparse shares

### Time `t=2`
We want to introduce a breaking change for sparse shares so 
- share version `2` is for compact shares 
- share version `3` is for sparse shares

Notice how share versions are flipped back and forth so it's not immediately clear which share version is applicable to which share type. If we version them together:

### Time `t=0` 
- share version `0` is for compact shares 
- share version `0` is for sparse shares

### Time `t=1` 
We want to make a breaking change to compact shares so we introduce
- share version `1` is for compact shares with new format
- share version `1` is for sparse shares

### Time `t=2`
We want to introduce a breaking change for sparse shares so 
- share version `2` is for compact shares with new format
- share version `2` is for sparse shares with new format

I'm open to versioning them separately if there are benefits I'm not accounting for.

pkg/shares/parse_sparse_shares.go
rootulp: Great question. We do have such a unit test here: https://github.com/rootulp/celestia-app/blob/bf8026570fada9b2a638f38a81d1f012bad45c2a/pkg/shares/parse_sparse_shares_test.go#L131 which continues to pass.

Previously, namespaced padded shares were added to the blob preceeding them and then trimmed from the previous blob [here](https://github.com/rootulp/celestia-app/blob/bf8026570fada9b2a638f38a81d1f012bad45c2a/pkg/shares/parse_sparse_shares.go#L78-L80). We don't want to remove those lines because a partial blob share will still have padding at the end of the share and those lines are responsible for trimming the trailing `0s`.

pkg/shares/namespaced_padding.go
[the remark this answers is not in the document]
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1341/commits/bf8026570fada9b2a638f38a81d1f012bad45c2a

cmwaters:
This is true but you're also viewing it from an angle where there are only ever two main types of share encoding/decoding (compact and sparse). These terms themselves are only named relative to themselves. Say we introduce an `ExtraCompactShare` where some transactions use that and others use `CompactShare`s. Using the namespace to dictate which protocol we use to interpret the bytes puts us in a bind. If we wanted to change so that certain transactions use `SparseShare`s and perhaps that certain Blobs could use a more compact encoding then we wouldn't be able to do this without making it quite messy. 

You could also set constants like `CurrentCompactShareVersion = 7` or `CurrentSparseShareVersion = 8`

Btw, I'm not implying that we have to do all this, I'm just exploring out loud the design space.

pkg/shares/parse_sparse_shares.go
evan-forbes: nice!

rootulp:
> Btw, I'm not implying that we have to do all this, I'm just exploring out loud the design space.

Agreed and thanks for articulating the use case with `ExtraCompactShare`, I can now see it being valuable. Proposal to continue the discussion in https://github.com/celestiaorg/celestia-app/issues/1347 because the discussion doesn't block this PR.

pkg/shares/namespaced_padding.go
```diff
@@ -33,3 +39,16 @@ func NamespacedPaddedShares(ns namespace.ID, n int) []Share {
...
+	isSequenceStart, err := s.IsSequenceStart()
+	if err != nil {
+		return false, err
+	}
+	sequenceLen, err := s.SequenceLen()
+	if err != nil {
+		return false, err
+	}
+
+	return isSequenceStart && sequenceLen == 0, nil
```
SweeXordious, line 53: [Question][Non-blocking]
Is there another case where this condition is true but it's not a namespaced padded? For example posting an empty blob?

SweeXordious (commented):
LGTM, left a question to understand more. Would defer approval to other reviewers.

pkg/shares/namespaced_padding.go
evan-forbes: > For example posting an empty blob?

empty blobs are invalid, so this is something that we have to (and are) checking for https://github.com/celestiaorg/celestia-app/blob/580d1864f7179689e1aa9994dbb526115f1a175e/x/blob/types/payforblob.go#L208-L210

that func is called when checking a tx before its accepted in the mempool, and before creating a PFB

evan-forbes (approved):
:shipit: LGTM

pkg/shares/namespaced_padding.go
SweeXordious: Cool, thanks :+1: :+1:

Issue #1299, opened 2023-01-26
Only allow a single testnode network to be spun up at once


evan-forbes:
Currently, it is possible to spin up two testnode networks at once on the same machine, resulting in the second network always failing. This could be one reason why we occasionally see flakey integration tests in app and node. We should add a local mutex that prevents this from occurring.

cmwaters:
Do we know what's causing the collision? Is it the p2p/rpc ports?

SweeXordious:
yes. Also, it's not possible to run 2 nodes because the node home is the same and the db complains + the config gets sealed

SweeXordious:
This issue, I guess, can be closed after the discussion under https://github.com/celestiaorg/celestia-app/pull/1342#issuecomment-1425944402

evan-forbes:
was this fixed @SweeXordious since we changed which ports are used?

SweeXordious:
I just did a small test running multiple `DefaultNetwork(...)` and the nodes started alright. I don't think they will be connected to each other, or they will be validating the same network. But, if used in separate tests, they shouldn't complain as it's running 2 independent networks.

PR #1342, opened 2023-02-06
feat: add mutex to testnode not to run multiple nodes

SweeXordious:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/1299

## Checklist


testutil/testnode/full_node.go
```diff
@@ -195,13 +200,17 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func() error
 	tmNode, app, cctx, err := New(t, DefaultParams(), tmCfg, false, genState, kr)
 	require.NoError(t, err)
 
+	// locking the mutex not to be able to spin up multiple nodes at the same time.
+	mut.Lock()
```
cmwaters, line 204:
```suggestion
	mut.Lock()
	t.Cleanup(mut.Unlock)
```
I would personally do it this way. Perhaps even remove the custom cleanup function since the testing fixture already provides it. There are two cases that I can see where the code exits before calling `Unlock`. Also I generally prefer to not pass responsibility for cleanup to a function outside the scope where the resources were created (if possible).

cmwaters (commented):
Have you tested just to make sure that this resolves these flaky tests?

testutil/testnode/full_node.go
SweeXordious: > Perhaps even remove the custom cleanup function since the testing fixture already provides it

Understandable, we could also define a certain `TestNode` struct that contains an exported `Cleanup` function to be used in tests. However, I am not against the way it is right now as it is really easy to start and stop.

> There are two cases that I can see where the code exits before calling Unlock

Yes, but those are not returning errors, they're `require`ing no errors. This means that, if I understand right, if we hit that case, the whole test suite will tear down and there will be no need to unlock nor cleanup.

This brings a question to my mind, will starting a separate test in a separate package, after failing a require, be affected by the locked mutex? I assume not, but not sure.

evan-forbes, line 204: [blocking suggestion]
we should probably put this in the `New` function instead of the default, as the default is only a helper function, and sometimes we need to specify things outside of the default

evan-forbes: if we can stick everything in the testing.T.Cleanup, then we totally should. @SweeXordious if you think this is out of scope of this PR, then I can handle it in a follow up

SweeXordious: The `New` is used to prepare the node, however the `DefaultNetwork` starts it. I guess that we want the lock when we want to start the node. If a user uses `New` to create the node structure (without starting it), then it's up to him to handle the locks, no?

SweeXordious: I can open a follow up PR doing that. Do we want to  merge this one?

SweeXordious:
> Have you tested just to make sure that this resolves these flaky tests?

@cmwaters I didn't try the flaky tests specifically, but this should give us an extra layer of security to avoid misusing the default node. So, even if it doesn't fix the flakiness, it's a step in that path

testutil/testnode/full_node.go
evan-forbes: we have other scenarios where we won't call default node, for example, each time we want to changes something in the config, consensus params, or genesis state. I think adding it it either the new or start function will accomplish the same thing, so which ever is fine imo

SweeXordious: [35c4042](https://github.com/celestiaorg/celestia-app/pull/1342/commits/35c4042788f56111ba1d1ca449f8045cf73f889f)  something like this?

evan-forbes (dismissed):
@SweeXordious when you get the chance, can you try this with the qgb tests to see if it fixes the issue? thx!

SweeXordious:
@evan-forbes Yes definitly, tomorrow. Can I merge this one?

evan-forbes:
yeah sure. i'm probably just stating the obvious, but if for some reason this doesn't help, or we missed something, then lets make a mental note to remove or fix this change :slightly_smiling_face:

testutil/testnode/rpc_client.go
```diff
@@ -13,12 +14,20 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+// mut used to lock the testnode after running a node.
+// This is because some tests are flaky as they try to start a node while another is running.
```
rootulp, line 18:
```suggestion
// mut is used to ensure that only one testnode is running at a time.
// mut was added to prevent simultaneous test executions from impacting each other and causing flaky failures.
```

testutil/testnode/rpc_client.go
```diff
@@ -13,12 +14,20 @@ import (
...
+var mut sync.Mutex
+
 // StartNode starts the tendermint node along with a local core rpc client. The
 // rpc is returned via the client.Context. The function returned should be
 // called during cleanup to teardown the node, core client, along with canceling
 // the internal context.Context in the returned Context.
 func StartNode(tmNode *node.Node, cctx Context) (Context, func() error, error) {
+	// locking the mutex not to be able to spin up multiple nodes at the same time.
```
rootulp, line 26:
```suggestion
	// lock the mutex so only one node is running at a time.
```

testutil/testnode/rpc_client.go
```diff
@@ -13,12 +14,20 @@ import (
...
+	mut.Lock()
 	if err := tmNode.Start(); err != nil {
+		// unlocking the mutex after cleanup finishes.
```
rootulp, line 29:
```suggestion
		// unlock the mutex after cleanup finishes.
```

testutil/testnode/rpc_client.go
```diff
@@ -28,6 +37,8 @@ func StartNode(tmNode *node.Node, cctx Context) (Context, func() error, error) {
 	goCtx, cancel := context.WithCancel(context.Background())
 	cctx.rootCtx = goCtx
 	cleanup := func() error {
+		// unlocking the mutex after cleanup finishes.
```
rootulp, line 40:
```suggestion
		// unlock the mutex after cleanup finishes.
```

rootulp (commented):
Optional comment edits to improve clarity. No blocking feedback

SweeXordious:
@evan-forbes Just checked, doesn't fix my issue 
```
failed to listen on 127.0.0.1:26657: listen tcp 127.0.0.1:26657: bind: address already in use
```
Should we still merge? or keep it up until we figure out why it's failing this way

MSevey:
I don't think this is the correct fix. How is this different from just ensuring that no tests run in parallel? i.e. not using `t.Parallel()`?  Which the repo doesn't appear to be using, so it makes sense that this doesn't fix the issue.

So it isn't a test running in parallel issue, but something else.  Since the tests aren't running in parallel, then ports don't seem to be getting freed efficiently, or tests just need to handle this and find a free port.

Node had this issue, they implemented a `getFreePort` helper https://github.com/celestiaorg/celestia-node/blob/main/core/testing.go#L192

Where in the test code are we defining ports?

Additionally looking at the linked issue, the db concerned should be solvable by simply creating unique test directories for each node, unless I'm misunderstanding that bug.

MSevey:
> @evan-forbes Just checked, doesn't fix my issue

is this deterministic on your machine? I just ran the testsuite on my Mac and it didn't fail.

SweeXordious:
> Node had this issue, they implemented a getFreePort helper https://github.com/celestiaorg/celestia-node/blob/main/core/testing.go#L192

That's one way to solve it. But I wanted in this PR to help address this issue not to have to assign ports dynamically in every repo using this code.

> Where in the test code are we defining ports?

They're coming from the default config from Celestia-core: 
https://github.com/celestiaorg/celestia-core/blob/07b8dbcb153563db8f2429b94a632b23d2621519/config/config.go#L83-L96

> Additionally looking at the linked issue, the db concerned should be solvable by simply creating unique test directories for each node, unless I'm misunderstanding that bug.

Yes, definitely the DB issue could be solved by that.

In fact, we could fix the ports issue by assigning them dynamically when initializing the config, the db by specifying different home directories, but still we will be stuck with the configuration being sealed when starting the node. Thus, unless we do a big refactor not to seal the config somehow during startup, we will never be able to run multiple nodes in parallel with different configs.

Thus, comes this PR to add an extra layer of security by ensuring that no node is being run in parallel.
The original goal is to fix the port allocation issue which I'm still investigating.

> is this deterministic on your machine? I just ran the testsuite on my Mac and it didn't fail.
The tests running into these issues are not the ones in this repo, but ones I am working on for the [orchestrator-relayer](https://github.com/celestiaorg/orchestrator-relayer) repo. Sorry for not providing much context.

Correct me please if I'm wrong @evan-forbes

MSevey (changes requested):
I don't think having a global lock to bottleneck tests to a single node is the right approach. We should be able to run tests in parallel.  Additionally since the issue persists there is no benefit to merging this in.

I think we need to clarify the actually problem statement we are trying to solve. The issue cites the port collision and the overlap in directory, which this PR doesn't address. 

Along with the detailed problem statement, we should also align on the ideal end state. 

To me being able to run tests and nodes in parallel would be ideal, so we should find a solution that does that.

SweeXordious:
@MSevey Makes sense, I'll see if I can solve the tests issue locally and update this PR with the changes. And then, we open an issue that concerns running the tests in parallel.

SweeXordious:
Created an issue for the above problem + a repo that is able to reproduce it: https://github.com/celestiaorg/orchestrator-relayer/issues/104

SweeXordious:
Just found out what the issue is about. I didn't know that `go test ./...` runs the packages tests in parallel.
In fact, this can be controlled with the flag `-p`: `go test ./... -p 1` which makes everything work.
So, there is no leak in resources. 

As a solution, what do you think of following:

1. Update the configuration of the testnode in here as follows:
```
	tmCfg.RPC.ListenAddress = "tcp://localhost:0"
	tmCfg.P2P.ListenAddress = "tcp://localhost:0"
	tmCfg.RPC.GRPCListenAddress = "tcp://localhost:0"
	appConf.GRPC.Address = "localhost:0"
	appConf.API.Address = "tcp://localhost:0"
```
So ports by default are allocated dynamically.

2. Keep the mutex that this PR adds so that if anyone wants to run 2 testnodes in parallel (inside the same package), the tests are forced to run sequentially.

I am happy to open a fast PR for `1`. And for 2, we can just merge this one.

Let me know what you think. 

cc @evan-forbes @MSevey

MSevey:
> Just found out what the issue is about. I didn't know that `go test ./...` runs the packages tests in parallel. In fact, this can be controlled with the flag `-p`: `go test ./... -p 1` which makes everything work. So, there is no leak in resources.


interesting, I also did not know that it would run things in parallel by default. So this runs directories in parallel? In order to run tests in parallel, within a directory, I imagine you still need the `t.Parallel()` flag. 

I'm in favor of option 1. 

I think option 2 is a little bit of an anti pattern. If  someone is trying to run things in parallel, and our test code is forcing it to be sequential, that could be giving developers a false positive.

SweeXordious:
> In order to run tests in parallel, within a directory, I imagine you still need the t.Parallel() flag.

Yes, 100%

Cool, will open a new PR making the change and close this one. 

Thanks a lot everyone for the help :+1: :+1:

evan-forbes:
I'd be in favor for 1 as well, we should also change the default home dir tendermint uses to the t.TempDir if we're not doing that already somewhere

SweeXordious:
> we should also change the default home dir tendermint uses to the t.TempDir if we're not doing that already somewhere

I have seen some `tempDir` used somewhere, but not sure where.
Should we create an issue to investigate the `data` folder that gets sometimes created when running the tests?

SweeXordious:
> Just so I follow, two separate packages are running tests that spin up a testnode and these testnodes conflict? Which packages are they? I see app_test [here](https://github.com/rootulp/celestia-app/blob/14d0c13c1bdca8b9cd02341b7c525ddff6a1f060/app/test/std_sdk_test.go#L49) but I don't see another invocation of the testutil/testnode node. I do see usage of the cosmos-sdk network.New node in app_test and testutil so maybe they conflict with the testutil/testnode?

Sorry for not specifying correctly. This is happening in a separate repo on a personal branch. You can reproduce using this: https://github.com/sweexordious/orchestrator-relayer/tree/not_working_test_poc

 > Maybe silly question but is it possible to make the testnodes independent so that multiple can be run simultaneously? If that's too difficult then it may not be worth it. 

Do you mean parallel tests inside a single package or parallel between packages? 

If it's the latter, then with this change https://github.com/celestiaorg/celestia-app/pull/1368, you can do it.

If you mean parallel tests inside the same package. Then, that can be achieved (I believe, I tried it locally, but might be missing something), with this https://github.com/celestiaorg/celestia-app/pull/1368, but only for testnodes having the same configuration. However, if we want to test something that change something in the `Bech32` prefixes or the `baseapp` configuration, we won't be able to:

https://github.com/celestiaorg/celestia-app/blob/6163e271b5128f554a6c19b6eddf415109224833/app/config.go#L12

https://github.com/celestiaorg/cosmos-sdk/blob/0aef7ba8d215a3978de8c298dc40d9fa65041695/baseapp/baseapp.go#L80-L81

Also, there are some data races happening even inside a single testnode instance. So, running multiple ones might lead to undefined behavior https://github.com/celestiaorg/celestia-app/issues/1369

rootulp:
> Do you mean parallel tests inside a single package or parallel between packages?

I meant parallel between packages so the default `go test ./...` behavior. Cool, looking forward to #1368 !

Issue #1102, opened 2022-12-09
Test that we are only removing txs from the block at the max square size


evan-forbes:
ideally we should only be hitting this if we're at the max squaresize. that might be something we want to add a test for in the future

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/1089#discussion_r1042916298_

this test could be as simple adding an additional check to an existing test that creates square not the max square size, and ensure that each transaction is getting included in the block.

cmwaters:
Closing this as I believe it has been solved through this test: https://github.com/celestiaorg/celestia-app/blob/c121c285005c0b5d47542fd955fc5be688905e1a/pkg/square/builder_test.go#L86

PR #1345, opened 2023-02-06
testing: add extra tests to ensure we remove txs correctly

cmwaters:
Addresses: https://github.com/celestiaorg/celestia-app/issues/1102

> this test could be as simple adding an additional check to an existing test that creates square not the max square size, and ensure that each transaction is getting included in the block.

`Test_finalizeLayout` already includes tests where a non max square size square is created and none of the transactions are dropped so I'm a little confused about this. I've added tests to include that we are dropping transactions if the blobs exceed the square size.

### DIFF
--- a/app/write_square_test.go
+++ b/app/write_square_test.go
@@ -147,6 +147,50 @@ func Test_finalizeLayout(t *testing.T) {
 			expectedIndexes: [][]uint32{{4}, {6, 8}, {5}},
 		},
 		{
+			// no blob txs should make it in the square
+			squareSize:      2,
+			nonreserveStart: 4,
+			blobTxs: generateBlobTxsWithNIDs(
+				t,
+				[][]byte{ns1, ns2, ns3},
+				[][]int{{1000}, {1000}, {1000}},
+			),
+			expectedIndexes: [][]uint32{},
+		},
+		{
+			// only two blob txs should make it in the square
+			squareSize:      4,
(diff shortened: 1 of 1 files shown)


evan-forbes:
> Test_finalizeLayout already includes tests where a non max square size square is created and none of the transactions are dropped so I'm a little confused about this.

hmm good point, I guess that is tested!

evan-forbes (approved):

cmwaters:
Shall I close https://github.com/celestiaorg/celestia-app/issues/1102?

rootulp:
I think the test described in #1102 tests something slightly different. Since the tests in this PR specify a square size, I don't think they are exactly applicable. These tests are contrived (but still useful) because for example in this scenario:

```
		{
			// only two blob txs should make it in the square
			squareSize:      4,
			nonreserveStart: 4,
			blobTxs: generateBlobTxsWithNIDs(
				t,
				[][]byte{ns1, ns2, ns3},
				[][]int{{2000}, {2000}, {6000}},
			),
			expectedIndexes: [][]uint32{{4}, {8}},
		},
```

[this](https://github.com/rootulp/celestia-app/blob/39bdc8fb69089bd4c0f5c052c20f8f639d4f8c47/app/estimate_square_size.go#L135-L136) would keep doubling the square size so that all three blob txs would fit. I think the test for #1102 may want to test `PrepareProposal` as a whole.

Issue #1331, opened 2023-02-02
Seal global cosmos config outside of the cli


evan-forbes:
to avoid breaking all the tests and increase diff from upstream in this repo, I think we should attempt a different approach where we only call the config/seal functions once in the app in some init function, then import something from that file in celestia-node. I think that will force us to only seal a single time, instead of once in the cli for app and node each.

_Originally posted by @evan-forbes in https://github.com/celestiaorg/cosmos-sdk/pull/297#pullrequestreview-1274063484_

Wondertan:
Nice! We did almost [the same](https://github.com/celestiaorg/celestia-node/pull/1583) recently. The PR name, though does not reflect what was in the final form, but I moved this init to one place in `core` pkg. Once you move this into the app, release and update the node with a PR, it would be nice to include the removal of init on our side.

evan-forbes:
@vgonkivs currently we are sealing here 

https://github.com/celestiaorg/celestia-app/blob/0b447adf776689cc8f14823c6ef925b8f8249b5a/cmd/celestia-appd/cmd/root.go#L51-L55

I haven't tried this, but we can probably just include that in an init function in the app package. either directly in this [file](https://github.com/celestiaorg/celestia-app/blob/main/app/app.go), or in its own

PR #1348, opened 2023-02-07
improvement: move sealing to init function

vgonkivs:
## Overview
Closes https://github.com/celestiaorg/celestia-app/issues/1331

## Checklist


cmd/celestia-appd/cmd/root.go
```diff
@@ -25,34 +18,42 @@ import (
...
 )
 
 const EnvPrefix = "CELESTIA"
 
-// NewRootCmd creates a new root command for celestia-appd. It is called once in the
-// main function.
-func NewRootCmd() *cobra.Command {
-	encodingConfig := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-
+func init() {
```
evan-forbes, line 45: to ensure that this init function is called, I think we will need to include this in a package that gets imported. We could do this a few different ways. I think including the init in the app package or we could include it in its own package, provided that we import it here and in node

evan-forbes (dismissed):
dope! thanks for taking the time to put into the app pkg :slightly_smiling_face:

app/config.go
```diff
@@ -0,0 +1,13 @@
+package app
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func init() {
```
rootulp, line 7: [for reviewers] `init()` is called once regardless of how many times this package is imported per https://tutorialedge.net/golang/the-go-init-function/

IMO this seems less clear than an explicit `New()` or `NewConfig()` invocation but maybe that's because I haven't seen it before. If it's clearer for celestia-node then :shipit:

rootulp (approved):

evan-forbes (approved):

rootulp (approved):

Issue #1271, opened 2023-01-19
Consider supporting multiple blob sizes in an `EventPayForBlob`


rootulp:
https://github.com/celestiaorg/celestia-app/blob/61a91faff6c6bd53729e322da9b41b8a321004b4/proto/blob/event.proto#L8-L14

How does this event support multiple namespace IDs but only a singular blob size? After the introduction of multiple blobs in a single PFB tx, I think this needs to be updated to support multiple blob sizes

PR #1354, opened 2023-02-07
fix!: emit event with blob sizes

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1271


evan-forbes (approved):
:shipit: 

thanks for noticing and fixing this before incentivized testnet!

Issue #1241, opened 2023-01-16
Rename `TxSharePosition` to `TxShareRange`


SweeXordious:
As discussed under https://github.com/celestiaorg/celestia-app/pull/1233#discussion_r1069999682, we could rename `TxSharePosition` to `TxShareRange` because the returned values form a range instead of a position.

rootulp:
Will do this after https://github.com/celestiaorg/celestia-app/pull/1276

PR #1357, opened 2023-02-08
chore!: rename `TxSharePosition` to `TxShareRange`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1241

The new name `TxShareRange` is consistent with https://github.com/celestiaorg/celestia-app/blob/fd6382278f6fff4ab3d272e5f3034203c9378126/pkg/proof/proof.go#L66

Technically API breaking because this is an exported function but I doubt there are external users of this function.


SweeXordious (approved):
LGTM :+1: :+1:

Yes, probably nobody is using this for now, so we can update no issue. Evan can correct me if I'm wrong.

PR #1366, opened 2023-02-10
chore: Add EditorConfig

jrmanes:
## Overview

Hello team!

This is a tiny one, I've added the `.editorconfig` file to the project.

Let me know if you prefer any other config 😄 

Best! 🚀


rootulp (commented):
Out of curiosity, why was this closed? If we choose to adopt editorconfig, should we subsequently explore formatting all existing files in the repo according to these rules and then enable a CI tool to catch regressions? https://github.com/editorconfig/editorconfig/wiki/FAQ#are-there-any-tools-to-check-orand-reformat-existing-files-against-editorconfig-rules contains a list but the first 3 don't appear to be actively maintained.

jrmanes:
> Out of curiosity, why was this closed? If we choose to adopt editorconfig, should we subsequently explore formatting all existing files in the repo according to these rules and then enable a CI tool to catch regressions? https://github.com/editorconfig/editorconfig/wiki/FAQ#are-there-any-tools-to-check-orand-reformat-existing-files-against-editorconfig-rules contains a list but the first 3 don't appear to be actively maintained.

Hello @rootulp 
Yes, I closed because it's not really needed, the linters should take care about the code format, and yes, as you mentioned, it will take more effort to update all the code already generated

rootulp:
Thanks for the info. For what it's worth, I'm not opposed to editorconfig or additional formatting rules if we can auto-fix violations and add a CI check for it.

Issue #1343, opened 2023-02-06
Add LastValsetBeforeNonce and other QGB RPC methods tests


SweeXordious:
We could add a test for the `LastValsetBeforeNonce` RPC call and also the other QGB RPC calls to make sure we always have them.

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/1325#pullrequestreview-1285973449_

PR #1374, opened 2023-02-13
test: add QGB queries test

SweeXordious:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/1343

## Checklist

### DIFF
--- a/app/test/qgb_rpc_test.go
+++ b/app/test/qgb_rpc_test.go
@@ -1,0 +1,95 @@
+package app_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/celestiaorg/celestia-app/testutil/testnode"
+	"github.com/celestiaorg/celestia-app/x/qgb/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestQGBRPCQueries(t *testing.T) {
+	_, cctx := testnode.DefaultNetwork(t, time.Millisecond)
+	h, err := cctx.WaitForHeightWithTimeout(401, time.Minute)
+	require.NoError(t, err)
+	require.Greater(t, h, int64(401))
+
+	queryClient := types.NewQueryClient(cctx.GRPCClient)
+
+	type test struct {
+		name string
+		req  func() error
+	}
+	tests := []test{
+		{
+			name: "attestation request by nonce",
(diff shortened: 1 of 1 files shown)


rootulp (commented):
LGTM with one small test name fix

app/test/qgb_rpc_test.go
```diff
@@ -0,0 +1,98 @@
...
+			req: func() error {
+				_, err := queryClient.LastUnbondingHeight(
+					context.Background(),
+					&types.QueryLastUnbondingHeightRequest{},
+				)
+				return err
+			},
+		},
+		{
+			name: "attestation request by nonce",
```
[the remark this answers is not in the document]
SweeXordious: Thanks a lot :+1: :+1:

rootulp (approved):

app/test/qgb_rpc_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestQGBRPCQueries(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping QGB integration test in short mode.")
+	}
+	_, cctx := testnode.DefaultNetwork(t, time.Millisecond)
+	h, err := cctx.WaitForHeightWithTimeout(401, time.Minute)
```
evan-forbes, line 19: [definitely optional]
if we don't want to wait a minute here, we can change the genesis state to include a significantly shorter data window param

evan-forbes (approved):

app/test/qgb_rpc_test.go
SweeXordious: We won't wait a minute. It's just a worst case scenario. Generally, with a block time of 1ms, 401 blocks will be created in less than half a second.
Yes, we can do that, but ,with fast blocktime, i dont see why tinker with the genesis state and do more changes.

Im okey with any tbh, whichever you prefer :cat:

evan-forbes: its fine as is, just an optimization :+1:

evan-forbes: block times will likely take longer than a millisecond, just fyi, that's kinda deceiving, as its really just the timeout commit

Issue #1282, opened 2023-01-23
Should we reserve byte(s) in the namespace IDs for versioned namespaces?


musalbas:
Right now we have a reserved info byte in shares with version information, in case we want to change the share format in the future.

Shall we also add a reserved info byte in namespace IDs for version information, in case we ever want to have version-specific namespaces? This would be relevant for example if we wanted to change the padding scheme in namespaces, without breaking backwards compatibility with existing namespaces.

Related to https://github.com/celestiaorg/celestia-app/issues/1136

evan-forbes:
This sounds like a good idea to me. I also don't think it would super difficult to add, at least to celestia-app, but we still likely won't be able to finish this for feature freeze. Finishing it soon after is definitely an option though.

one slight modification could be to combine the info bytes into a single one at the beginning of the share.

musalbas:
>one slight modification could be to combine the info bytes into a single one at the beginning of the share.
This would prevent some use cases, such as reading a blob in the middle of a namespace, because the bit for the start share indicator would no longer be useful.

>+1 to the idea of combining the version information in the current info byte with the proposed version info for the namespace. I think the version in the current info byte can be used to version changes to the share format with the exception of a change to the length of the namespace.
To be clear, the idea is not to combine any version information. I'm proposing a totally separate, additive, feature of reserving bytes in namespace ids.

>If such a change doesn't involve a change to the namespace length, I think this could be handled by the current info byte. Example:

I'm not following here. How would this work for non-reserved namespaces? If you change the padding scheme for an existing non-reserved namespace, it would be backwards-incompatible.

rootulp:
> If you change the padding scheme for an existing non-reserved namespace

I think this refers to padding in-between blob shares (a.k.a namespaced padding) and not padding at the end of a share but can you please confirm? 

If we change the scheme of namespaced padding I expect that change to be backwards-incompatible and therefore necessitate a hard fork. I don't see how a namespaced padding scheme change can be communicated in the version byte inside a share because it applies across a data square and there is no guarantee that an entire data square is composed of shares with one particular version (unless that is also part of the hard fork). In other words, can the rules related to padding scheme of the square be determined based on the block height rather than the version byte inside a share?

musalbas:
>I think this refers to padding in-between blob shares (a.k.a namespaced padding) and not padding at the end of a share but can you please confirm?

Yes.

[...]

PR #1382, opened 2023-02-14
docs: ADR 14 Versioned Namespaces

rootulp:
ADR for https://github.com/celestiaorg/celestia-app/issues/1282


docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,192 @@
...
+    2. Option 2: when there are changes to any part of the remaining data in a share
+3. When do we expect to increment the namespace version?
+    1. During a backwards incompatable non-interactive default rule change
+    2. If we change the format of a padding share (e.g. a namespace padding share) instead of `0` bytes, pad with something else like. We may need to preserve backwards compatibility for padding shares that use old namespaces. Note this scenario likely implies a namespace version and share version increase.
+    3. Change the format of PFB tx serialization. This scenario likely implies duplicating the PFB txs in a data square, one with the old namespace version and one with the new namespace version.
+4. Inspired by [type-length-value](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value), should we consider prefixing optional fields (sequence length and reserved bytes) with a type and a length? This would enable us to modify those fields without introducing new share versions.
+5. [Requires investigation] what changes need to be made to NMT in order to support namespaces of a different length (e.g. 16 bytes)?
```
evan-forbes, line 178: nmt should already have support, we simply create an nmt with the `nmt.Option` to set the size to 16.

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,192 @@
...
+6. [Requires investigation] what changes need to be made to support variable length namespace IDs? In other words, roll-ups may use a namespace ID of 8 - 32 bytes. Are variable length namespace IDs something we are considering supporting?
```
evan-forbes, line 179: is this something that we are still wanting to investigate? I don't think we were explicit about it, but my understanding was that it would be too messy, although possible

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,192 @@
...
+
+Notes
+
+- With this option, the share version is prefixed to the namespace prior to pushing to the NMT. When a user queries a namespace, they must also specify a share version.
+
+## Implementation Details
+
+Regardless of the option chosen, end-users must explicitly specify the namespace ID version (and/or) share version they intend on using in their PFB transaction.
+
+Since the NMT needs to be aware of the version byte, Option A is equivalent to increasing the namespace ID size to `9` bytes and then constraining the namespaces available for use to namespaces that have a leading `0` byte. In other words:
```
evan-forbes, line 158: :+1:

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,192 @@
...
+
+- MinReservedNamespace: []byte{0, 0, 0, 0, 0, 0, 0, 0, 1}
+- MaxReservedNamespace: []byte{0, 0, 0, 0, 0, 0, 0, 0, 255}
+- MinBlobNamespace: []byte{0, 0, 0, 0, 0, 0, 0, 1, 0}
+- MaxBlobNamespace: []byte{0, 255, 255, 255, 255, 255, 255, 255, 255}
+
+When a user creates a PFB, concatenate the namespace version with the namespace ID to derive the namespace that is pushed to the NMT. Option C is similar to Option A, if we constrain the available share versions at mainnet to `0`.
+
+## Open Questions
```
evan-forbes, line 167: what version byte should we use for tail padding shares and parity shares? I assume we need to use the largest version?

evan-forbes (commented):
awesome idea essentially pre-writing the ADR before the decision! I'm glad we're recording all of this!

had one or two questions that we might want to discuss quickly, but I'm tempted to approve already

docs/architecture/adr-014-versioned-namespaces.md
rootulp: Good question. Given we adopt **Option A**.

### Option A1 
Use the maximum available namespace version. For example, at mainnet launch if we constrained the namespace versions to only `0`, then the tail padding share would be:

- namespace version: `0`
- namespace ID: `namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE}`

At a later date when we support namespace versions `0` and `1`, the tail padding share would be:

- namespace version: `1`
- namespace ID: `namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE}`

### Option A2
Use the maximum namespace version regardless of supported namespace versions. So a tail padding share would be:

- namespace version: `255`
- namespace ID: `namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE}`

docs/architecture/adr-014-versioned-namespaces.md
rootulp: related question: do we want to make the NMT's default namespace ID size 16 (something like https://github.com/celestiaorg/nmt/pull/100)

docs/architecture/adr-014-versioned-namespaces.md
evan-forbes: A2 makes the most sense to me, since we won't have to continuously update it

evan-forbes: its kinda weird since the only reason we have to do this is because nmt doesn't know about the version byte

docs/architecture/adr-014-versioned-namespaces.md
evan-forbes: imo yes, and we should probably not rely on that like we do now, simply for the sake of being explicit. so each time we create an nmt, we should use the option as if the default was wrong.

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,192 @@
...
+An approach that addresses these issues is to prefix the namespace ID with version metadata.
+
+### Option A: `Namespace Version | Namespace ID`
+
+| Field | Number of bytes | Description |
+| --- | --- | --- |
+| Namespace Version | 1 | the version of the namespace ID |
+| Namespace ID | 8 if Namespace Version=0, 32 if Namespace Version=1 | namespace ID of the share |
+
+For example, consider the scenario where at mainnet launch namespace IDs have a length of 8 bytes. In this scenario, the only supported namespace ID is `0`. At some point in the future,  we want to increase the length of namespace IDs from 8 bytes to 32 bytes. We can expand the range of available namespaces to include namespaces that start with a leading `0` or `1` byte.
```
liamsi, line 50: Q: I might have misunderstood that on the call but didn't we say that despite the namespace version increasing the namespace size would be a breaking/non-backwards compatible change where the version alone will not help (because it would also impact the roots etc). If that is the case (and I understood it correctly) I'm in favour with using a proper use-case here instead.

rootulp: You understood correctly. I'll update this section with a different use case.

docs/architecture/adr-014-versioned-namespaces.md
rootulp: Agreed. Created https://github.com/celestiaorg/celestia-app/issues/1384

evan-forbes (commented):
only remaining thing imo, otherwise LGTM and I think we should merge
```
ERROR: 2 dead links found!
  [✖] https://proto.school/anatomy-of-a-cid → Status: 0
  [✖] https://cid.ipfs.tech/#QmcRD4wkPPi6dig81r5sLj9Zm1gDCL4zgpEj9CfuRrGbzF → Status: 0
```

docs/architecture/adr-014-versioned-namespaces.md
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1382/commits/e579db3ac970fe7a7097d48be51c43f6e1e3d2bd

docs/architecture/adr-014-versioned-namespaces.md
rootulp: In this case, I expect we'll prefix the namespace ID with the namespace version prior to pushing leaves onto the NMT so the NMT would be made aware of the version.

I agree option A2 makes sense. Added to doc in https://github.com/celestiaorg/celestia-app/pull/1382/commits/4a1489dde48a0c8a4c285ee13f679c27754fd022

evan-forbes (dismissed):

docs/architecture/adr-014-versioned-namespaces.md
rootulp: I'm going to investigate 

> 5. [Requires investigation] what changes need to be made to NMT in order to support namespaces of a different length (e.g. 16 bytes)?

in another branch / PR with the assumption that we increase namespace ID size to 16 bytes (or higher). Will punt investigating variable length namespace IDs for now

rootulp: @evan-forbes is it safe to resolve this convo?

evan-forbes (approved):

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,193 @@
...
+          namespaceID := namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE}
+
+          namespaceVersion := 1
+          namespaceID := namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE}
+          ```
+
+        but they differ in the location where they would be placed in the NMT.
+    1. Option 6.2: `255`
+      - Pros: The format of a tail padding share or parity share doesn't need to change when a future namespace version bump occurs.
+
```
rootulp, line 181:
```suggestion
7: How can we increase the namespace ID post mainnet?
    - Increment the namespace version. Construct two data squares and two NMTs. Data square 1 uses NMT 1 with namespace version 0 (namespace ID size 8 bytes). Data square 2 uses NMT 2 with namespace version 1 (namespace ID size of 32 bytes). Would celestia-nodes sample two separate data squares or is there a clever way to combine both data squares?

```

evan-forbes (dismissed):

evan-forbes (approved):

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,195 @@
...
+  0, 0, 0, 3, // sequence length
+  1, 2, 3, // blob data
+  // 0 padding until share is full
+ }
+```
+
+The current schema poses a challenge for the following scenarios:
+
+1. Changes to the non-interactive default rules that don't break backwards compatibility with existing namespaces [celestia-app#1282](https://github.com/celestiaorg/celestia-app/issues/1282), [celestia-app#1161](https://github.com/celestiaorg/celestia-app/pull/1161)
+    - After mainnet launch, if we want to change the non-interactive default rules but retain the previous non-interactive default rules for backwards compatibility, it isn't possible to differentiate the namespaces that want to use the old rules vs the new rules.
```
cmwaters, line 37: Just to regurgitate this point for my own understanding: We want the rules that define the construction of a square (i.e. potentially the non-interactive default rules or any other algorithm) to be on a per namespace basis. Not per square (i.e. a version number that reflects the construction algorithm for the entire square - implying homogeneity for all txs in that square) and not per share (i.e. multiple shares for the same namespace could theoretically have different versions for interpreting where they should be placed in the square - this assumes that all messages are always tied to a namespace and are always ordered (ascending)).

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,195 @@
...
+1. Changes to the format of a padding share.
+1. Changes to the format of PFB transaction serialization.
+
+## Proposal
+
+An approach that addresses these issues is to prefix the namespace ID with version metadata.
+
+### Option A: `Namespace Version | Namespace ID`
```
cmwaters, line 45: How are blobs ordered now? Do they include the namespace version or just the ID (and if it is including the version is it at the start or end)

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,195 @@
...
+Cons
+
+- 2 bytes of metadata per share
+
+Notes
+
+- Option B is implementable at a later time if we adopt Option A.
+- It isn't immediately clear how NMT can support variable length namespaces.
+
+### Option C: `Share Version | Namespace ID`
```
cmwaters, line 111: I'm still for this option. (I'm aware we've already decided. I just haven't yet been convinced that there is a specific need for namespace versioning that isn't covered by share versioning and square versioning).

I'm also a big fan for putting the version number in a fixed indexed position (like the first or last byte) as opposed to the 9th byte or 17th byte

docs/architecture/adr-014-versioned-namespaces.md
rootulp: Yes I believe so.

> this assumes that all messages are always tied to a namespace and are always ordered

messages => blobs ? If so, I agree.

docs/architecture/adr-014-versioned-namespaces.md
```diff
@@ -0,0 +1,195 @@
...
+  1, 2, 3, // blob data
+  // 0 padding until share is full
+ }
+```
+
+The current schema poses a challenge for the following scenarios:
+
+1. Changes to the non-interactive default rules that don't break backwards compatibility with existing namespaces [celestia-app#1282](https://github.com/celestiaorg/celestia-app/issues/1282), [celestia-app#1161](https://github.com/celestiaorg/celestia-app/pull/1161)
+    - After mainnet launch, if we want to change the non-interactive default rules but retain the previous non-interactive default rules for backwards compatibility, it isn't possible to differentiate the namespaces that want to use the old rules vs the new rules.
+1. Changes to the format of a padding share.
```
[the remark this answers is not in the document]
rootulp: > Why couldn't we use the version number of the share to reflect a change in the format. That would seem most intuitive for me

TBH I think we could.

docs/architecture/adr-014-versioned-namespaces.md
rootulp: > How are blobs ordered now?

In the current implementation? Shares are ordered by namespace ID. There is no namespace version in the current implementation. There is a share version in the current implementation and that is after the namespace ID. 

Ref: https://github.com/celestiaorg/celestia-app/blob/c09843d07d4c3842753138de96b304b4866e8f5d/specs/src/specs/data_structures.md?plain=1#L448

docs/architecture/adr-014-versioned-namespaces.md
cmwaters: Well messages could also be an IBC transaction or a governance vote (they still have namespaces)

docs/architecture/adr-014-versioned-namespaces.md
rootulp: > I'm also a big fan for putting the version number in a fixed indexed position (like the first or last byte) as opposed to the 9th byte or 17th byte

FWIW the accepted proposal has a namespace version at index 0 and a share version at index 33

PR #1386, opened 2023-02-16
fix: remove changing unnecessary parameters in single-node.sh

SweeXordious:
## Checklist

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -20,11 +20,9 @@ celestia-appd collect-gentxs
 # Set proper defaults and change ports
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
-sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_commit = "25s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_propose = "3s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/index_all_keys = false/index_all_keys = true/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/mode = "full"/mode = "validator"/g' ~/.celestia-app/config/config.toml
+sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/client.toml
+sed -i'.bak' 's/timeout_commit = "10s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
+sed -i'.bak' 's/timeout_propose = "10s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
 
 # Start the celestia-app
 celestia-appd start
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -20,7 +20,7 @@ celestia-appd collect-gentxs
 # Set proper defaults and change ports
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
-sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
+sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
 
 # Start the celestia-app
 celestia-appd start
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -21,8 +21,6 @@ celestia-appd collect-gentxs
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
 sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/client.toml
-sed -i'.bak' 's/timeout_commit = "10s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_propose = "10s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
 
 # Start the celestia-app
 celestia-appd start
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -20,7 +20,7 @@ celestia-appd collect-gentxs
 # Set proper defaults and change ports
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
-sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/client.toml
+sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
 
 # Start the celestia-app
 celestia-appd start


scripts/single-node.sh
```diff
@@ -20,11 +20,9 @@ celestia-appd collect-gentxs
 # Set proper defaults and change ports
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
-sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_commit = "25s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_propose = "3s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/index_all_keys = false/index_all_keys = true/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/mode = "full"/mode = "validator"/g' ~/.celestia-app/config/config.toml
+sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/client.toml
+sed -i'.bak' 's/timeout_commit = "10s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
```
SweeXordious, line 24: [not for reviewers]
the defaults are now `10s`

scripts/single-node.sh
```diff
@@ -20,11 +20,9 @@ celestia-appd collect-gentxs
 # Set proper defaults and change ports
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
-sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_commit = "25s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_propose = "3s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/index_all_keys = false/index_all_keys = true/g' ~/.celestia-app/config/config.toml
```
SweeXordious, line 26: [not for reviewers]
these fields are no longer existent

scripts/single-node.sh
```diff
@@ -20,11 +20,9 @@ celestia-appd collect-gentxs
 # Set proper defaults and change ports
 # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
 # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
-sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
```
SweeXordious, line 23: [not for reviewers]
this was moved to the `client.toml` file

SweeXordious:
Also, by default, we have this:
```
[tx_index]

# What indexer to use for transactions
#
# The application will set which txs to index. In some cases a node operator will be able
# to decide which txs to index based on configuration set in the application.
#
# Options:
#   1) "null"
#   2) "kv" (default) - the simplest possible indexer, backed by key-value storage (defaults to levelDB; see DBBackend).
#               - When "kv" is chosen "tx.height" and "tx.hash" will always be indexed.
#   3) "psql" - the indexer services backed by PostgreSQL.
# When "kv" or "psql" is chosen "tx.height" and "tx.hash" will always be indexed.
indexer = "null"
```

Why is the indexer disabled by default?

evan-forbes (commented):
one question

scripts/single-node.sh
SweeXordious: Apparently sometime between `v0.12.0-rc7`  and latest master. I checked the history and it seems that only this https://github.com/celestiaorg/celestia-app/commit/990c7201c395551e56652b086056e05fe4150763 touches on those. Want me to investigate?

evan-forbes:
> Are we hard-coding the 10s and not allowing to change?

yeah, unfortunately. if we use the cli, then consensus timeouts are overridden. Ideally we make then params and not configs.

evan-forbes:
> Why is the indexer disabled by default even if the docs state "kv" (default) -... ?

that's another good point, we changed the defaults here in the app, but those docs are in tendermint

scripts/single-node.sh
evan-forbes: forgot to mention that https://github.com/celestiaorg/celestia-app/pull/1386#issuecomment-1433109353 changing these won't do anything. The only way to spin up a faster chain is programmatically.

SweeXordious: I see, thanks :+1:

SweeXordious: so we can just delete these `sed` commands?

SweeXordious:
> that's another good point, we changed the defaults here in the app, but those docs are in tendermint

should we do the change? or no need?

evan-forbes:
> should we do the change? or no need?

probably just a low priority change in tendermint

scripts/single-node.sh
SweeXordious: What about the config files that still contain those parameters, probably they should be deleted to avoid any confusion for validators, no?

SweeXordious: > so we can just delete these sed commands?

[0c8806d](https://github.com/celestiaorg/celestia-app/pull/1386/commits/0c8806d9ed2932e30f6f24af07c00cba14da5175)

scripts/single-node.sh
rootulp: [question] can you clarify what you mean by no longer existent? I see them in 

config.toml
```toml
#######################################################
###         Consensus Configuration Options         ###
#######################################################
[consensus]

wal_file = "data/cs.wal/wal"

# How long we wait for a proposal block before prevoting nil
timeout_propose = "10s"
# How much timeout_propose increases with each round
timeout_propose_delta = "500ms"
# How long we wait after receiving +2/3 prevotes for “anything” (ie. not a single block or nil)
timeout_prevote = "1s"
# How much the timeout_prevote increases with each round
timeout_prevote_delta = "500ms"
# How long we wait after receiving +2/3 precommits for “anything” (ie. not a single block or nil)
timeout_precommit = "1s"
# How much the timeout_precommit increases with each round
timeout_precommit_delta = "500ms"
# How long we wait after committing a block, before starting on the new
# height (this gives us a chance to receive some more precommits, even
# though we already have +2/3).
timeout_commit = "10s"
```

Do you mean we should remove them from this script because the values in config.toml aren't respected?

rootulp (commented):
I don't understand what the bug is or how this PR addresses it. @SweeXordious do you mind creating an issue with repro steps and expected behavior?

scripts/single-node.sh
evan-forbes: I compiled main and reintiated, and for me I'm still seeing the line on 91 of config.toml

SweeXordious: @rootulp Yes, just checked, we have both. Which refer to what? I guess they're the same no?

scripts/single-node.sh
SweeXordious: I meant:
```
sed -i'.bak' 's/index_all_keys = false/index_all_keys = true/g' ~/.celestia-app/config/config.toml
``` 
and 
```
sed -i'.bak' 's/mode = "full"/mode = "validator"/g' ~/.celestia-app/config/config.toml
```

SweeXordious:
> I don't understand what the bug is or how this PR addresses it. @SweeXordious do you mind creating an issue with repro steps and expected behavior?

I added the `bug` label because I thought there is a bug as `timeout_commit` was not picked up, before I knew it's expected because those values are hardcoded.

I will update the title of the PR to something more meaningful now as the discussion can be switched to correcting the `single-node.sh` script to have correct commands (and delete commands that are not making any change to the resulting chain)

scripts/single-node.sh
rootulp: Oh got it! I confirmed that these fields don't appear present in

config.toml

```
index_all_keys = false
mode = "full"
```

and then it seems like

```
timeout_commit
timeout_propose
```

are overriden [here](https://github.com/celestiaorg/celestia-app/blob/b40a6908954828b90aebd530c5d30592626834f7/cmd/celestia-appd/cmd/overrides.go#L13-L14) so this convo can be resolved

scripts/single-node.sh
SweeXordious: The `client.toml` contains the default configuration for running commands, https://github.com/cosmos/cosmos-sdk/pull/8953.
So, I will revert the change: [10b99a6](https://github.com/celestiaorg/celestia-app/pull/1386/commits/10b99a639ac260f9be1bcc5cd0ea2edda2acc34b)

SweeXordious:
PR updated, the new purpose of this PR  is to remove  these commands that are not making any change:

```
sed -i'.bak' 's/timeout_commit = "25s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
sed -i'.bak' 's/timeout_propose = "3s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
sed -i'.bak' 's/index_all_keys = false/index_all_keys = true/g' ~/.celestia-app/config/config.toml
sed -i'.bak' 's/mode = "full"/mode = "validator"/g' ~/.celestia-app/config/config.toml
```

I guess all the  conversations are now resolved. Let me know if I should address something else

evan-forbes (approved):

rootulp (changes requested):

scripts/single-node.sh
SweeXordious: Good catch, messed the address

scripts/single-node.sh
```diff
@@ -20,11 +20,7 @@ celestia-appd collect-gentxs
...
-sed -i'.bak' 's/timeout_commit = "25s"/timeout_commit = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/timeout_propose = "3s"/timeout_propose = "1s"/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/index_all_keys = false/index_all_keys = true/g' ~/.celestia-app/config/config.toml
-sed -i'.bak' 's/mode = "full"/mode = "validator"/g' ~/.celestia-app/config/config.toml
+sed -i'.bak' 's#"tcp://localhost:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
```
SweeXordious, line 23:
```suggestion
sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.celestia-app/config/config.toml
```

rootulp (approved):

evan-forbes (approved):

Issue #1396, opened 2023-02-16
Remove unused build tags


rootulp:
## Context

The tests workflow on Github CI currently runs with several unused build tags.

https://github.com/celestiaorg/celestia-app/blob/b40a6908954828b90aebd530c5d30592626834f7/.github/workflows/test.yml#L113

https://github.com/celestiaorg/celestia-app/blob/b40a6908954828b90aebd530c5d30592626834f7/.github/workflows/test.yml#L195

The tags were initially added in https://github.com/celestiaorg/celestia-app/commit/c3b312eb68ba8e87f4f26b8c11cfe8ce919ad58b

## Proposal

Since the repo doesn't use any of these build tags: `norace`, `ledger`,  `test_ledger_mock`, `cgo`, proposal to delete them from these commands

PR #1397, opened 2023-02-16
ci: remove unused build tags from go test

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1396


SweeXordious (commented):
I guess these were added when the project was initially generated. I would defer to Evan to confirm if they're needed

rootulp:
I'm pretty confident they're not needed b/c I couldn't find any instances of the build tags in test files so specifying the tags in `go test` doesn't change which tests are run

evan-forbes (approved):

evan-forbes:
yeah, they're not needed

PR #1400, opened 2023-02-16
ci: consolidate linters in lint workflow

rootulp:
## Motivation

I found the **Before** workflow diagram a bit confusing. I would expect all linters to run as part of the lint workflow and I wouldn't expect lint to block splitting up test files.

Before | After
--- | ---
Screenshot 2023-02-16 at 5 01 40 PM| Screenshot 2023-02-16 at 5 06 11 PM

TBH I don't know why **test / cleanup-runs** shows up in the lint block but I think we can get rid of that job soon via a different PR

### DIFF
--- a/.github/workflows/ci_release.yml
+++ b/.github/workflows/ci_release.yml
@@ -27,32 +27,16 @@ on:
           - major
 
 jobs:
-  # Dockerfile Linting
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):
noice

Issue #1308, opened 2023-01-27
Consider increasing the namespace id size


rootulp:
## Context

In Slack, we discussed that it would be difficult to change the namespace id size post mainnet launch because such a change has implications on the NMT, share encoding, etc.

## Proposal

Consider increasing the namespace id size from 8 bytes to something larger.

## Tasks
- [x] https://github.com/celestiaorg/celestia-app/pull/1405
- [x] https://github.com/celestiaorg/celestia-app/pull/1419
- [x] https://github.com/celestiaorg/celestia-app/issues/1453

cc: @musalbas

rootulp:
Napkin math:

256 ^ 8 = 1.84467441e19
256 ^ 10 = 1.20892582e24
256 ^ 16 = 3.40282367e38

rootulp:
When implementing this, inform @staheri14 that it will involve a change to NMT and the spec

rootulp:
| Namespace ID size (in bytes) | % of 512 byte share |
|------------------------------|---------------------|
| 8                            | 1.5%                |
| 16                           | 3.1%                |
| 20                           | 3.9%                |
| 32                           | 6.2%                |

## Notes

- [SHA256](https://en.wikipedia.org/wiki/SHA-2) has a digest size of 32 bytes so using a namespace ID size of 32 bytes would enable users to generate stable namespace IDs (e.g. `sha256('sov-labs')`) or unique namespace IDs (e.g. `sha256(blob)`)
- [IPv6](https://en.wikipedia.org/wiki/IPv6) has an address space of 16 bytes [source](https://en.wikipedia.org/wiki/IPv6#Addressing)

    > the address space is deemed large enough for the foreseeable future

- The size of Ethereum's private key space is 2^256 (32 bytes) [source](https://github.com/ethereumbook/ethereumbook/blob/develop/04keys-addresses.asciidoc)
- The size of Ethereum's address space is 2^160 (20 bytes) [source](https://github.com/ethereumbook/ethereumbook/blob/05f0dfe6c41635ac85527a60c06ac5389d8006e7/04keys-addresses.asciidoc)
- The size of Bitcoin's address space is 2^160 (20 bytes) [source](https://www.coinhouse.com/insights/news/what-if-my-wallet-generated-an-existing-bitcoin-address/)
- The size of Fuel's contract ID is 32 bytes [source](https://fuellabs.github.io/fuel-specs/master/protocol/id/contract.html)
[...]

PR #1405, opened 2023-02-17
docs: ADR 15 Namespace ID Size

rootulp:
ADR for https://github.com/celestiaorg/celestia-app/issues/1308


docs/architecture/adr-014-namespace-id-size.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Context
+
+Namespace ID is currently an 8 byte slice. 8 bytes provides a maximum of 2^64 possible namespace IDs. In practice some namespace IDs are reserved for protocol use so the number of namespace IDs available to users is 2^56 - 2. Modifying the size of a namespace ID post-launch is a breaking change and has implications for the NMT, share encoding, etc. so we'd like to carefully consider the size of the namespace ID pre-launch.
+
+Desirandum:
+
+1. Ability to randomly generate a new namespace that hasn't been used before
+
+// what other desirandum do we have?
```
rootulp, line 15: @musalbas @liamsi do ya'll have any input here?

docs/architecture/adr-014-namespace-id-size.md
```diff
@@ -0,0 +1,57 @@
+# ADR 14: Namespace ID Size
+
+## Changelog
+
+- 2023/2/17: initial draft
+
+## Context
+
+Namespace ID is currently an 8 byte slice. 8 bytes provides a maximum of 2^64 possible namespace IDs. In practice some namespace IDs are reserved for protocol use so the number of namespace IDs available to users is 2^56 - 2. Modifying the size of a namespace ID post-launch is a breaking change and has implications for the NMT, share encoding, etc. so we'd like to carefully consider the size of the namespace ID pre-launch.
```
musalbas, line 9: Do we want to consider increasing the number of reserved namespace IDs to 65536 or larger, if we also increase namespace ID size?

rootulp: Good question. Currently the reserved namespace range occupies:

```go
MinReservedNamespace := namespace.ID{0, 0, 0, 0, 0, 0, 0, 0}
MaxReservedNamespace = namespace.ID{0, 0, 0, 0, 0, 0, 0, 255}
```

and it currently uses 3 of the 256 range for transactions, PFB transactions, and max reserved namespace for reserved namespace padding. Given such low utilization (3/256 is 1% utilization) and without any proposals that increase utilization, I think it's safe to preserve the 256 range and expand at a later date if needed. Thoughts?

For completeness, the protocol also reserves these two:

```
TailPaddingNamespaceID = namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE}
ParitySharesNamespaceID = namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}
```

musalbas: I think this reserved namespace range is only for namespace version 0 right (assuming this adr is merged)? So if we add extra namespace versions we have to explicitly define the reserved namespace range for those

rootulp: Exactly. I expect the strategy you proposed to work if we wanted to increase the reserved namespace range via a namespace id version bump.

rootulp: With notable exceptions for tail padding and parity shares namespaces: https://github.com/celestiaorg/celestia-app/pull/1382#discussion_r1112247526

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,64 @@
...
+1. Ability to randomly generate a new namespace that hasn't been used before
+
+// what other desirandum do we have?
+
+## Notes
+
+- The namespace ID must provide at least 72 bits of randomness ([Eager](https://eager.io/blog/how-long-does-an-id-need-to-be/)) to satisfy desirandum 1. Since an 8 byte namespace ID can only provide 64 bits of randomness, it fail to meet this desirandum.
+- [SHA256](https://en.wikipedia.org/wiki/SHA-2) has a digest size of 32 bytes so using a namespace ID size of 32 bytes would enable users to generate stable namespace IDs (e.g. `sha256('sov-labs')`) or unique namespace IDs (e.g. `sha256(blob)`) with the SHA256 hash function.
+- [IPv6](https://en.wikipedia.org/wiki/IPv6) has an address space of 16 bytes and "the address space is deemed large enough for the foreseeable future" ([Wikipedia](https://en.wikipedia.org/wiki/IPv6#Addressing)).
+- [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) have slightly less than 16 bytes of randomnees  and are considered "unique enough for practical purposes" ([Towards Data Science](https://towardsdatascience.com/are-uuids-really-unique-57eb80fc2a87)).
```
evan-forbes, line 26: random kneeeeees!
```suggestion
- [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) have slightly less than 16 bytes of randomness  and are considered "unique enough for practical purposes" ([Towards Data Science](https://towardsdatascience.com/are-uuids-really-unique-57eb80fc2a87)).
```

evan-forbes (approved):
nice, makes sense to me! afaiu, nmt, celestia-app, and celestia-node should be able to have larger namespaces with only minor changes here and celestia-node, but I could be wrong.

musalbas:
Should we not agree on the actual new namespace size before merging this or marking as ready for review? The decision section is empty.

rootulp:
> Should we not agree on the actual new namespace size before merging this or marking as ready for review?

I was hoping we could reach a decision via feedback on this PR. If you think we need a meeting to decide, I can set one up.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,64 @@
...
+Desirandum:
+
+1. Ability to randomly generate a new namespace that hasn't been used before
+
+// what other desirandum do we have?
+
+## Notes
+
+- The namespace ID must provide at least 72 bits of randomness ([Eager](https://eager.io/blog/how-long-does-an-id-need-to-be/)) to satisfy desirandum 1. Since an 8 byte namespace ID can only provide 64 bits of randomness, it fail to meet this desirandum.
+- [SHA256](https://en.wikipedia.org/wiki/SHA-2) has a digest size of 32 bytes so using a namespace ID size of 32 bytes would enable users to generate stable namespace IDs (e.g. `sha256('sov-labs')`) or unique namespace IDs (e.g. `sha256(blob)`) with the SHA256 hash function.
```
staheri14, line 24: > unique namespace IDs (e.g. `sha256(blob)`) with the SHA256 hash function.

It would be helpful to clarify that even in the case of sha256(blob), the namespace ID uniqueness still depends on the blob uniqueness. For example, if two blobs are identical, their hashes will also be identical.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,64 @@
...
+- [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) have slightly less than 16 bytes of randomness  and are considered "unique enough for practical purposes" ([Towards Data Science](https://towardsdatascience.com/are-uuids-really-unique-57eb80fc2a87)).
+- The size of the Ethereum and Bitcoin address space is 2^160 (20 bytes) ([Mastering Ethereum](https://github.com/ethereumbook/ethereumbook/blob/05f0dfe6c41635ac85527a60c06ac5389d8006e7/04keys-addresses.asciidoc) and [Coinhouse](https://www.coinhouse.com/insights/news/what-if-my-wallet-generated-an-existing-bitcoin-address/)).
+- The size of Fuel's contract ID is 32 bytes ([Fuel specs](https://fuellabs.github.io/fuel-specs/master/protocol/id/contract.html)). The size of Fuel's address space is either 20 bytes ([Fuel docs](https://docs.fuel.sh/v1.1.0/Concepts/Data%20Structures/Addresses.html#addresses)) or 32 bytes ([fuel-docs#75](https://github.com/FuelLabs/fuel-docs/issues/75)).
+
+## Options
+
+| Namespace ID size (in bytes) | % of 512 byte share | Desirandum 1 |
```
staheri14, line 32: Can you please add a short description of what column titled `% of 512 byte share` represents.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,64 @@
...
+
+Q: What are the negative consequences of having a large namespace ID size?
+
+A1: The namespace ID is prefixed to each NMT leaf. Two namespace IDs are prefixed to each NMT non-leaf hash. Therefore, the nodes of an NMT will be larger based on the namespace ID size.
+
+A2: The namespace ID is prefixed to each share. Since a share is a fixed 512 bytes, a share's capacity for blob data decreases as the namespace ID increases.
+
+Q: What are the performance implications on celestia-node for a larger namespace ID size?
+
+Q: How can we increase the namespace ID post mainnet?
```
staheri14, line 51: Does the current ADR align with the concept of versioned namespace ID? If so, could you please provide more details regarding the last question, which asks about increasing the namespace ID after the mainnet launch? Wouldn't this involve incrementing the version number?

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,64 @@
...
+
+Desirandum:
+
+1. Ability to randomly generate a new namespace that hasn't been used before
+
+// what other desirandum do we have?
+
+## Notes
+
+- The namespace ID must provide at least 72 bits of randomness ([Eager](https://eager.io/blog/how-long-does-an-id-need-to-be/)) to satisfy desirandum 1. Since an 8 byte namespace ID can only provide 64 bits of randomness, it fail to meet this desirandum.
```
staheri14, line 23: Can you please provide a link to the `Desirandum` definition? or add a short explanation in the text on what `satisfy desirandum 1` means.

rootulp: I spelled it incorrectly but I meant to use: https://www.dictionary.com/browse/desideratum
I can replace with something more explicit like "Desirable criteria"

docs/architecture/adr-015-namespace-id-size.md
rootulp: Will do! This column describes how much of the share is occupied by the namespace ID. In other words 8/512 or 32/512. The values are represented as a percentage

docs/architecture/adr-015-namespace-id-size.md
rootulp: > Does the current ADR align with the concept of versioned namespace ID?

ADR 15 was created after ADR 14 because we aligned on ADR 14 with the assumption that we'll also increase the namespace ID size (hence this ADR). So yes both ADRs are related and aligned.

> If so, could you please provide more details regarding the last question, which asks about increasing the namespace ID after the mainnet launch? Wouldn't this involve incrementing the version number?

Yes I expect it to involve incrementing a version number. Given that, I'll move this question back to ADR 14 and elaborate more there.

rootulp: > Does the current ADR align with the concept of versioned namespace ID?

ADR 15 was created after ADR 14 because we aligned on ADR 14 with the assumption that we'll also increase the namespace ID size (hence this ADR). So yes both ADRs are related and aligned.

> If so, could you please provide more details regarding the last question, which asks about increasing the namespace ID after the mainnet launch? Wouldn't this involve incrementing the version number?

Yes I expect it to involve incrementing a version number. Given that, I'll move this question back to ADR 14 and elaborate more there.

docs/architecture/adr-015-namespace-id-size.md
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1405/commits/1241dd04e85cf6097ed7d5c167cb5e7d8d9f2895

docs/architecture/adr-015-namespace-id-size.md
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1405/commits/1241dd04e85cf6097ed7d5c167cb5e7d8d9f2895

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,58 @@
...
+
+- 2023/2/17: initial draft
+
+## Context
+
+Namespace ID is currently an 8 byte slice. 8 bytes provides a maximum of 2^64 possible namespace IDs. In practice some namespace IDs are reserved for protocol use so the number of namespace IDs available to users is 2^56 - 2. Modifying the size of a namespace ID post-launch is a breaking change and has implications for the NMT, share encoding, etc. so we'd like to carefully consider the size of the namespace ID pre-launch.
+
+Desirable criteria:
+
+1. Ability to randomly generate a namespace that hasn't been used before
```
liamsi, line 17:
```suggestion
1. Ability to randomly generate a namespace that hasn't been used before (with large enough probability)
```

Other cirteria:

- Potentially large enough for all rollups that will (ever) exist 
- use a truncated secure hash function (like sha256) for generating the namespace 

The 2nd though is basically the same as "Ability to randomly generate a namespace that hasn't been used before".

rootulp: > Potentially large enough for all rollups that will (ever) exist

Will add.

> The 2nd though is basically the same as "Ability to randomly generate a namespace that hasn't been used before".

Agreed so there is a bullet point for it in the Notes section but I didn't include it as an explicit desirable criteria.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,110 @@
...
+- For the non-adversarial use-case, we want to avoid users having to check if a random namespace has already been used.
+- Use case for larger namespace ID size: rollups may have multiple namespaces (e.g. Twitter) where a roll-up may give each user a namespace within a namespace range.
+- Is it possible to make the namespace ID a parameter, so that the namespace ID is a parameter to proof verification for roll-ups?
+  - Assumes that a block height may have a different namespace ID
+- There are talks in the Ethereum community about a potential address range increase.
+- 20 bytes gives us Ethereum address compatability so Ethereum addresses could be mapped to a Celestia namespace ID.
+- Currently namespace ID size = 8 bytes. Each intermediate node in the NMT is 8 (namespace ID) + 8 (namespace ID) + 32 (SHA256)= 48 bytes
+- If namespace ID size = 16 bytes. Each intermediate nodes in the NMT become: 16 (namespace ID) + 16 (namespace ID) + 32 (SHA256) = 64 bytes
+- Other option: increase size to 32 bytes with an optimization that reserves the first N bytes. The first N bytes wouldn't be sent over the wire.
```
SweeXordious, line 97: [Question]
I missed this point, what do you mean by:
 
> wouldn't be sent over the wire

They will be deterministic in some sense?

rootulp: The optimization would reserve the first N bytes to be unused initially. For example: if the total namespace ID size was 32 bytes, reserve the first 16 bytes as unusable initially. The first 16 bytes would be set to `0`. The last 16 bytes would contain a user's desired namespace.

When the share is sent across the network, a client could receive the first byte (the namespace version) and read that its of namespace version `0` so it can interpret the next 16 bytes as `0` and then read the next `16` bytes as the real namespace. In this optimization, the first 16 bytes of `0` don't need to be sent over the network. Does that make sense?

SweeXordious: Defintly, thanks :+1: :+1:

liamsi: The 0s also wouldn't be stored, right?

rootulp: If they aren't stored, then I don't understand where the namespace ID size actually represents 32 bytes. With this optimization, I expect the leading 16 bytes of `0`s to be stored in the share. 

Can you clarify where you don't think they would be stored @liamsi ?

musalbas: I don't think they need to be stored on disk, for example.

The only place where the 0s would be recognized is when computing the hash of intermediate nodes.

That being said, for this to solve the backwards-compatibility issue we're talking about, rollup software and NMT implementations would have to not assume that all intermediate nodes will have 0s at the start, in case we do ever increase the namespace size as use those preceding 0s...

musalbas: For example, assuming hashes are big endian, then when serializing the hashes over the wire or on disk, preceding 0s can be omitted, just like how Ethereum addresses that start up with 0s are more gas efficient

rootulp: > The only place where the 0s would be recognized is when computing the hash of intermediate nodes.

I think the 0s (or at least the share version) would also need to be recognized when creating a share commitment so that these two shares don't end up resulting in the same share commitment:

```go
shareA := []byte {
	0x0 // share version
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // 16 bytes of unrecognized namespace
	0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, // 16 bytes of user specified namespace
	// remainder of share ommited for brevity
}
shareB := []byte {
	0x1 // share version
	0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, // 16 bytes of user specified namespace
	0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, 0x1, // another 16 bytes of user specified namespace
	// remainder of share ommited for brevity
}
```

if that's the case, we're special casing the share version `0` to be interpreted as 16 bytes of `0`s in most instances where the namespace ID is used. In order to evaluate this optimization, I think we need to consider the code complexity and investigate how much bandwidth this decreases (presumably 16 bytes every time a namespace ID is sent over the network) but idk how much that is relative to total bandwidth requirements.

musalbas: Oh I see, yeah you can also make the 16 0s enforced by the share version, and if we increase the namespace size, introduce a new share version (which is opt-in by rollups thus it doesn't break backwards compatibility)

That being said, if a Merkle proof in an old share version contains a sibling for a new share version that has a longer namespace, that could break backwards compatibility, so you may need a bunch of padding shares ... (or I think the non-interactive defaults already solve this issue ?)

rootulp: > That being said, if a Merkle proof in an old share version contains a sibling for a new share version that has a longer namespace, that could break backwards compatibility

Agreed this does break backwards compatability. I don't expect non-interactive defaults to solve the issue. An NMT proof doesn't appear to contain the raw share of a sibling based on [these docs](https://github.com/celestiaorg/nmt/blob/1bc0bb0099e01b30e37ddb56642734ae875917cd/nmt.go#L150-L155) but `proof.nodes` contains namespaced hashes so if shareA and shareB are in the same data square and correspondingly same NMT, then inner nodes of the NMT for `namespaceVersion=0`  would be of size: 2 * 16 + 32 = 64 and inner nodes of the NMT for `namespaceVersion=1` would be of size: 2 * 32 + 32 = 96.

rootulp: In order for NMT proof verification to work, the original NMT proof verification logic would need to interpret `namespaceVersion=0` with 16 leading bytes of `0`s and to not perform such interpretation for `namespaceVersion=1` and above. It seems contrived to implement NMT proof verification in this way and it may in fact break if we made a different change when introducing `namespaceVersion=1`.

Preserving backwards compatibility of proofs for an NMT constructed with different size namespaces seems brittle in this scenario. Thoughts? @musalbas

musalbas: Yeah, you wouldn't have backwards compatibility if you have different namespace sizes for different versions anyways, because at least the root node might have a max or min node of a different size than older clients expect.

I think the only backwards compatible way is if all clients don't make any assumption about how many leading 0s there are, but have a consensus rule that requires leading 0s, like the Ethereum address optimization I mentioned here

>For example, assuming hashes are big endian, then when serializing the hashes over the wire or on disk, preceding 0s can be omitted, just like how Ethereum addresses that start with 0s are more gas efficient

rootulp:
CI markdownlint doesn't like the footnotes: https://github.com/celestiaorg/celestia-app/actions/runs/4318011765/jobs/7535787254#step:4:32 but local markdownlint doesn't complain

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,200 @@
...
+  - Disagreement on serialization
+  - Implementation complexity of parsing a varint
+- Desirable property: first 40 fixed bytes to be metadata
+  - IPV6 packet header is fixed
+
+## Decision
+
+Increase the namespace ID size to 32 bytes.
+
+- Prefix the namespace ID with 1 byte for the namespace version. See [ADR 14](./adr-014-versioned-namespaces.md).
```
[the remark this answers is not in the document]
rootulp: Correct. The namespace ID size in the NMT will be 33 bytes. 

Users will specify a namespace version (1 byte / uint8) and a namespace ID (32 bytes) in their PFB. Additionally we should strive to make it clear to users that these two will be interpreted as distinct namespace IDs:

```go
namespaceVersion := 0
namespaceId := []byte{1, 1, 1, ...} // 32 bytes

namespaceVersion := 1
namespaceId := []byte{1, 1, 1, ...} // 32 bytes
```

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,200 @@
...
+- Solution to woods attack
+  - Rollups can't assume that all blobs in a namespace are honest
+  - Rollups shouldn't scan a namespace directly. Instead they should gossip block headers and light clients should only request blobs of interest.
+- Why no dynamic namespace ID length?
+  - Disagreement on serialization
+  - Implementation complexity of parsing a varint
+- Desirable property: first 40 fixed bytes to be metadata
+  - IPV6 packet header is fixed
+
+## Decision
```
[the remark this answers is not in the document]
rootulp: Can do.

docs/architecture/adr-015-namespace-id-size.md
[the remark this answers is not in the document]
rootulp: Clarified in https://github.com/celestiaorg/celestia-app/pull/1405/commits/4a91bf6ace87bd91820be2b7794d9276ae2c9eea

docs/architecture/adr-015-namespace-id-size.md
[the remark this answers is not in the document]
rootulp: Done in https://github.com/celestiaorg/celestia-app/pull/1405/commits/4a91bf6ace87bd91820be2b7794d9276ae2c9eea

evan-forbes (approved):

rootulp:
Will merge by EOD if there is no more feedback

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Blob Inclusion Proof Size
+
+Blob inclusion proofs haven't yet been implemented so this proposal can't precisely determine the impact on blob inclusion proofs. A naive implementation of blob inclusion proofs may return NMT proofs for all shares that a blob occupies, in other words one NMT proof per row that a blob spans. Assuming shares are 512 bytes, square size is 128, and a blob is less than 128 shares, a blob would occupy a maximum of 2 rows. Therefore, the namespace ID size's impact on blob inclusion proofs would be approximately 2 * the impact on NMT proofs. A [blob size independent inclusion proof](https://github.com/celestiaorg/celestia-app/blob/6d27b78aa64a749a808e84ea682352b8b551fbd7/docs/architecture/adr-011-optimistic-blob-size-independent-inclusion-proofs-and-pfb-fraud-proofs.md?plain=1#L19) is likely smaller than this naive implementation because it depends on the number of shares that a PFB transaction spans (likely significantly fewer than 2 rows).
+
+### Share Size
+
+Another tradeoff to consider is the size of the namespace ID in the share. Since a share is a fixed 512 bytes, a share's capacity for blob data decreases as the namespace ID increases.
```
staheri14, line 131: [Sharing thoughts] The total size of a blob is also affected by the ID size. For example, if a 32-byte ID is used instead of an 8-byte ID, the total size of a blob is reduced by 128 * (32-8) (128 refers to the max number of shares per blob as mentioned earlier).

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+| 16                        | 3.1%                                    |
+| 20                        | 3.9%                                    |
+| 32                        | 6.2%                                    |
+
+## Open Questions
+
+1. What are the performance implications on celestia-node for a larger namespace ID size?
+1. Is it possible to mitigate some tradeoffs when adopting a large namespace ID size?
+    1. It may be possible to decrease the bandwidth requirements for NMT proofs by using lossless compression (proposed by @evan-forbes) and explored above.
+    1. It may be possible to avoid writing the namespace ID to continuation blob shares (proposed by @nashqueue)
```
staheri14, line 145: [Question] I'm uncertain how we could incorporate this idea into NMT, since the block shares' corresponding leaves in NMT must have distinct namespaces. Consequently, we may not be able to apply this optimization in the final NMT representation.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+        1. Note this introduces complexity for erasure reconstruction. A share in row B may have its namespace in row A so to reconstruct a data square, we must refactor the process to two steps:
+            1. Reconstruct all shares from the erasure coding
+            1. Reconstruct the NMT
```
staheri14, line 148:
```suggestion
1. What are the performance implications on celestia-node for a larger namespace ID size?
2. Is it possible to mitigate some tradeoffs when adopting a large namespace ID size?
    1. It may be possible to decrease the bandwidth requirements for NMT proofs by using lossless compression (proposed by @evan-forbes) and explored above.
    2. It may be possible to avoid writing the namespace ID to continuation blob shares (proposed by @nashqueue)
        1. Note this introduces complexity for erasure reconstruction. A share in row B may have its namespace in row A so to reconstruct a data square, we must refactor the process to two steps:
            1. Reconstruct all shares from the erasure coding
            2. Reconstruct the NMT
```

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+|---------------------------|-----------------------------------------|
+| 8                         | 1.5%                                    |
+| 16                        | 3.1%                                    |
+| 20                        | 3.9%                                    |
+| 32                        | 6.2%                                    |
+
+## Open Questions
+
+1. What are the performance implications on celestia-node for a larger namespace ID size?
+1. Is it possible to mitigate some tradeoffs when adopting a large namespace ID size?
```
staheri14, line 143: [nit] All the items have been assigned the same number "1", it needs to be fixed.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+- 2023/3/1: blob inclusion proof size
+- 2023/3/2: accepted
+
+## Context
+
+Namespace ID is currently an 8 byte slice. 8 bytes provides a maximum of 2^64 possible namespace IDs. In practice some namespace IDs are reserved for protocol use so the number of namespace IDs available to users is 2^56 - 2. Modifying the size of a namespace ID post-launch is a breaking change and has implications for the NMT, share encoding, etc. so we'd like to carefully consider the size of the namespace ID pre-launch.
+
+## Decision
+
+Increase the namespace ID size to 32 bytes.
```
staheri14, line 22: [Question] Can we have shares with different namespace versions in the same NMT?

staheri14 (approved):
Nice! left some non-blocking comments.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+
+- Prefix the namespace ID with 1 byte for the namespace version. See [ADR 14](./adr-014-versioned-namespaces.md).
+- For `namespaceVersion=0`, the namespace ID is 32 bytes where:
+  - The first 22 bytes are 0s
```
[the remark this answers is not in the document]
rootulp: resolved by cc14b43c87fe2b8069272483ba6e394ada6793bb

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+  - The last 10 bytes are unreserved namespace bytes. In other words, a user can specify a 10 byte namespace ID
+- Add a consensus rule that the leading 22 bytes of a namespace ID are 0s.
```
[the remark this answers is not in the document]
rootulp: Yes. Such a rule would be consensus-critical (likely enforced in process_proposal.go).

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+
+The motivation for reserving the first 22 bytes of the namespace ID as 0s is to enable future bandwidth optimizations. In particular, the namespace ID may be run length encoded to reduce the size of NMT proofs.
+
+The namespace ID size is 32 bytes so that a future namespace version can be introduced to expand the namespace ID address space without a backwards incompatible change from the perspective of NMT.
+
+The namespace version will be prefixed to the namespace ID prior to pushing data to the NMT so NMT should be constructed with a namespaceID size of 33 bytes.
+
+Users will specify a namespace version (1 byte / uint8) and a namespace ID (32 bytes) in their PFB. Additionally we should strive to make it clear to users that these two will be interpreted as distinct namespace IDs:
```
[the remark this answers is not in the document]
rootulp: these two refers to the two namespace IDs below. I'll re-word the section below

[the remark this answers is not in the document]
rootulp: resolved by 3cbd49eea640d20d28e052ddaf6b79f5bfb4f384

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+namespaceId := []byte{1, 1, 1, ...} // 32 bytes
+
+namespaceVersion := 1
+namespaceId := []byte{1, 1, 1, ...} // 32 bytes
+```
+
+## Desirable Criteria
+
+1. A user should be able to randomly generate a namespace that hasn't been used before[^1]
+1. There should exist a large enough namespace ID space for all rollups that may exist in the forseeable future (e.g. 100 years)
```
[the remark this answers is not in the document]
rootulp: [informational] Markdown lets one use the `1.` prefix for multiple entires in an [ordered list](https://www.markdownguide.org/basic-syntax/#ordered-lists). I prefer the `1.` prefix because if you re-order items in the list, you don't need to manually change each number. Will accept suggestion so clearer for readers.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+Namespace ID size (bytes) | NMT data leaf size (bytes) | NMT inner node size (bytes)
+--------------------------|----------------------------|----------------------------
+8                         | 8 + 512 = 520              | 2*8 + 32 = 48
+16                        | 16 + 512 = 528             | 2*16 + 32 = 64
+20                        | 20 + 512 = 532             | 2*20 + 32 = 72
+32                        | 32 + 512 = 544             | 2*32 + 32 = 96
+
+### NMT Proof Size
+
+Increasing the size of NMT nodes will increase the size of the NMT proof. Assuming shares are 512 bytes, square size is 128, and the NMT proof is for a single leaf:
```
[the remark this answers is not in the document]
rootulp: resolved by 5e136bac34e044d799bc56e7025b27a2a8239213

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -0,0 +1,212 @@
...
+
+Namespace ID size (bytes) | Unencoded NMT proof size (bytes) | Protobuf encoded NMT proof size (bytes) | Protobuf encoded NMT proof with [gzip](https://pkg.go.dev/compress/gzip) (bytes)
```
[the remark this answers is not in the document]
rootulp: thanks, will add a note below this section.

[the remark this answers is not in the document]
rootulp: resolved by f94c7f249e4a592e46622c3bad84c41ef2f3961b

docs/architecture/adr-015-namespace-id-size.md
rootulp: Yea, this optimization would happen before writing to the leaf. There's a Slack thread exploring this topic here: https://celestia-team.slack.com/archives/C02PFFCGMNW/p1677197644501209

The rough idea is that shares wouldn't contain the namespace ID but the entire data square (or at least preceding rows) would need to be reconstructed prior to constructing the NMT so that leaves pushed to the NMT would contain the namespace ID.

docs/architecture/adr-015-namespace-id-size.md
rootulp: resolving b/c https://www.markdownguide.org/basic-syntax/#ordered-lists

docs/architecture/adr-015-namespace-id-size.md
rootulp: Depends on what change the new namespace version is introduced for but if a new namespace version is introduced to change something like non-interactive default rules then yes b/c both namespace version `0` and `1` will be of size 32 bytes (ID) + 1 byte (version) = 33 bytes.

docs/architecture/adr-015-namespace-id-size.md
rootulp: Yea good point. Will add a section for it.

rootulp: Resolved by 70fb1308288b7bcb2a4bfbe152496b16aea89a98

evan-forbes (approved):

staheri14 (approved):
Thanks for addressing the comments, LGTM!

PR #1407, opened 2023-02-20
docs: hoist ADR status

rootulp:
Addresses https://github.com/celestiaorg/celestia-app/pull/1161#issuecomment-1437293103

## Questions
1. Do we want to make the same change across other celestiaorg repos? 
2. Should we consider an org wide ADR template?
    1. https://github.com/celestiaorg/celestia-app/issues/1415

### DIFF
--- a/docs/architecture/adr-001-abci++-adoption.md
+++ b/docs/architecture/adr-001-abci++-adoption.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-The proposed and initial implementation is complete.
+Implemented
 
 ## Changelog
 
@@ -35,10 +35,6 @@ While the adoption of ABCI++ is inevitable given the already made decision by up
 
 - [Alternatives for Message Inclusion.](https://github.com/celestiaorg/celestia-app/blob/92341dd68ee6e555ec6c0bb780afa3a1c8243a93/adrs/adr008:adopt-ABC%2B%2B-early.md#alternative-approaches)
 - [Alternatives for Picking a square size.](https://github.com/celestiaorg/celestia-core/issues/454)
-
-## Decision
-
-Proposed and initial implementation is complete.
 
 ## Detailed Design
 
--- a/docs/architecture/adr-001-abci++-adoption.md
+++ b/docs/architecture/adr-001-abci++-adoption.md
@@ -1,5 +1,9 @@
 # ADR 001: ABCI++ Adoption
 
+## Status
+
+The proposed and initial implementation is complete.
+
 ## Changelog
 
 - 2022-03-03: Initial Commit
@@ -369,10 +373,6 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
    }
 }
 ```
-
-## Status
-
-The proposed and initial implementation is complete.
 
 ## Consequences
 
--- a/docs/architecture/adr-005-qgb-reduce-state-usage.md
+++ b/docs/architecture/adr-005-qgb-reduce-state-usage.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-Proposed
+Deprecated in favor of [orchestrator-relayer#65](https://github.com/celestiaorg/orchestrator-relayer/pull/66)
 
 ## Context
 
--- a/docs/architecture/adr-005-qgb-reduce-state-usage.md
+++ b/docs/architecture/adr-005-qgb-reduce-state-usage.md
@@ -1,5 +1,9 @@
 # ADR 005: QGB Reduce State Usage
 
+## Status
+
+Proposed
+
 ## Context
 
 The first design for the QGB was to use the state extensively to store all the QGB-related data: Attestations, `Valset Confirms` and `DataCommitment Confirms`.
@@ -40,7 +44,7 @@ However, slashing will be very difficult, especially for liveness, i.e. an orche
 
 Remove the `MsgValsetConfirm` defined in [here](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L24-L49)
 And also, the `MsgDataCommitmentConfirm` defined in [here](
-https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L55-L76).
+<https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L55-L76>).
 Which were the way orchestrators were able to post confirms to the QGB module.
 Then, keep only the state that is created in [EndBlocker](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/x/qgb/abci.go#L12-L16).
 Which are `Attestations`, i.e. `Valset`s and `DataCommitmentRequest`s.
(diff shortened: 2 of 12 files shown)


docs/architecture/adr-001-abci++-adoption.md
```diff
@@ -1,5 +1,9 @@
 # ADR 001: ABCI++ Adoption
 
+## Status
+
+The proposed and initial implementation is complete.
```
musalbas, line 5: What does that mean? Is it accepted or only proposed?

docs/architecture/adr-007-universal-share-prefix.md
```diff
@@ -1,5 +1,9 @@
 # ADR 007: Universal Share Prefix
 
+## Status
+
+Implemented
```
musalbas, line 5: Implemented is not a real status according to ADR template. Do we want to add implemented as a status type, and if so, should we update all the implemented Accepted statuses to Implemented?

docs/architecture/adr-011-optimistic-blob-size-independent-inclusion-proofs-and-pfb-fraud-proofs.md
```diff
@@ -1,5 +1,11 @@
 # ADR 011: Optimistic Blob Size Independent Inclusion Proofs and PFB Fraud Proofs
 
+## Status
+
+Accepted -> Does not affect the Celestia Core Specification
+
+Optimization 1 & 2 **Declined** as it is currently not worth it to introduce extra complexity for reducing the PFB proof size by 512-1024 bytes.
```
musalbas, line 7: Bit messy imo to have half-accepted ADRs..

docs/architecture/adr-template.md
```diff
@@ -1,5 +1,11 @@
 # ADR {ADR-NUMBER}: {TITLE}
 
+## Status
+
+> A decision may be "proposed" if it hasn't been agreed upon yet, or "accepted" once it is agreed upon. Once the ADR has been implemented mark the ADR as "implemented". If a later ADR changes or reverses a decision, it may be marked as "deprecated" or "superseded" with a reference to its replacement.
+
+{Deprecated|Proposed|Accepted|Declined}
```
musalbas, line 7: Implemented is not here - should it be?

musalbas:
Left some general comments about inconsistent usage of status message but it doesn't have to block this PR which just moves the status field up.

docs/architecture/adr-007-universal-share-prefix.md
rootulp: > Do we want to add implemented as a status type

Sure.

> should we update all the implemented Accepted statuses to Implemented?

Will do in this PR.

docs/architecture/adr-template.md
rootulp: It should be because the text above it:

> Once the ADR has been implemented mark the ADR as "implemented"

rootulp: Resolved by 7d564a1ff4c571744d63825869b246cc543ce01f

rootulp:
I looked through Github repos in celestiaorg but couldn't find an adequate place for an ADR template. @MSevey or @liamsi do you have a suggestion for where such an ADR template should live?

docs/architecture/adr-001-abci++-adoption.md
rootulp: Resolved by 34ed8282e7571fe3c105799bdb7840785af39f4e

docs/architecture/adr-005-qgb-reduce-state-usage.md
```diff
@@ -1,5 +1,9 @@
 # ADR 005: QGB Reduce State Usage
 
+## Status
+
+Proposed
```
rootulp, line 5: @SweeXordious is it safe to update this to Accepted or Implemented? I'm not confident because the `## Decision` section has incomplete checkboxes

docs/architecture/adr-template.md
```diff
@@ -1,5 +1,11 @@
 # ADR {ADR-NUMBER}: {TITLE}
 
+## Status
+
+> A decision may be "proposed" if it hasn't been agreed upon yet, or "accepted" once it is agreed upon. Once the ADR has been implemented mark the ADR as "implemented". If a later ADR changes or reverses a decision, it may be marked as "deprecated" or "superseded" with a reference to its replacement.
+
+{Deprecated|Proposed|Accepted|Declined|Implemented}
```
rootulp, line 7: this list doesn't contain "Deprecated" or "Superseded" either.

docs/architecture/adr-005-qgb-reduce-state-usage.md
SweeXordious: `This one is huum, I implemented an MVP for it in [here](https://github.com/sweexordious/celestia-app/tree/qgb_working_draft_of_new_ingestion_2.0) but we never really had a deployed version. Instead, we stopped working on it and switched to [this](https://github.com/celestiaorg/orchestrator-relayer/pull/66). 

Can we put in the status something like `replaced by deciding to use a P2P network` or similar?

docs/architecture/adr-template.md
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1407/commits/be4369590deba49f6fab881aaa7988b22edfd2fe

docs/architecture/adr-007-universal-share-prefix.md
rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1407/commits/ce7c3bebcd9fd762e9d635d437a15e07f6ed8710

docs/architecture/adr-005-qgb-reduce-state-usage.md
rootulp: Yea, we have a Deprecated option. Is there a doc I can link to for the P2P network?
```suggestion
Deprecated
```

MSevey:
> I looked through Github repos in celestiaorg but couldn't find an adequate place for an ADR template. @MSevey or @liamsi do you have a suggestion for where such an ADR template should live?

Since the ADRs are files, I don't know if any GitHub features for the templates, like issue templates for example. 

The simplest thing would be to just have an adr_template.md file in each repo. It's mostly a one time cost to make the template in all the repos that need it.

docs/architecture/adr-005-qgb-reduce-state-usage.md
SweeXordious: It's this PR: https://github.com/celestiaorg/orchestrator-relayer/pull/66 that contains the ADr, which is still not merged

rootulp: Resolved by f34bc88914e437c58aafb1f2039d2da11c308071

rootulp:
We already have an ADR template per repo so changes to the ADR template (like the one in this PR) would need to be re-applied across all repos. I was proposing a celestiaorg wide repo that contains one ADR template. It may also be useful for things like linter rules where we establish an org wide standard.

docs/architecture/adr-005-qgb-reduce-state-usage.md
```diff
@@ -40,7 +44,7 @@ However, slashing will be very difficult, especially for liveness, i.e. an orche
 
 Remove the `MsgValsetConfirm` defined in [here](https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L24-L49)
 And also, the `MsgDataCommitmentConfirm` defined in [here](
-https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L55-L76).
+<https://github.com/celestiaorg/celestia-app/blob/a965914b8a467f0384b17d9a8a0bb1ac62f384db/proto/qgb/msgs.proto#L55-L76>).
```
SweeXordious, line 47: [Question][Non-blocking]

What's the difference between `[...](<link>)` and `[...](link)`?

rootulp: there isn't a difference. My editor auto surrounded the link with `<` and `>` likely to conform to https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md but in this case, it's not a bare URL because of the preceding `[here]`

nashqueue:
We should maybe include authors / Co-authors in ADRs  so they can be pinged in future references

evan-forbes (approved):

evan-forbes:
org wide template would be cool

SweeXordious (approved):
LGTM, would  defer to other reviewers for final approval

Issue #1422, opened 2023-02-23
`nmt_wrapper_test.go` depends on magic constant


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/e6bbc302194376dbfd909cfabbf82e16e8fd5e7f/pkg/wrapper/nmt_wrapper_test.go#L98

## Problem

Based on [these lines](https://github.com/celestiaorg/celestia-app/blob/e6bbc302194376dbfd909cfabbf82e16e8fd5e7f/pkg/wrapper/nmt_wrapper.go#L75-L80) I expect the entire share (i.e. 512 bytes) to be pushed to the erasured NMT. `appconsts.ContinuationSparseShareContentSize` + 1 may not always == the share size - the len of the namespace

PR #1423, opened 2023-02-23
chore: improve nmt wrapper tests

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1422

Motivation: I was reading through nmt_wrapper.go to prepare an ADR and got confused by the usage of ` appconsts.ContinuationSparseShareContentSize+1`. I made a few changes to the test file for clarity & maintainability.

### DIFF
--- a/pkg/wrapper/nmt_wrapper_test.go
+++ b/pkg/wrapper/nmt_wrapper_test.go
@@ -46,7 +46,7 @@ func TestPushErasuredNamespacedMerkleTree(t *testing.T) {
 // to the second half of the tree.
 func TestRootErasuredNamespacedMerkleTree(t *testing.T) {
 	size := 8
-	data := generateRandNamespacedRawData(size, appconsts.NamespaceSize, appconsts.ShareSize)
+	data := generateRandNamespacedRawData(size, appconsts.NamespaceSize, appconsts.ShareSize-appconsts.NamespaceSize)
 	tree := NewErasuredNamespacedMerkleTree(uint64(size), 0)
 	nmtTree := nmt.New(sha256.New())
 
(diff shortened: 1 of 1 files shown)


rootulp:
This change broke for unexpected reasons. The error I observed is 
```
    nmt_wrapper_test.go:140: shard sizes do not match
```
The root cause was klauspost expects shards to have a length that are a multiple of 64 bytes

pkg/wrapper/nmt_wrapper_test.go
```diff
@@ -132,7 +142,8 @@ func generateErasuredData(t *testing.T, numLeaves int, codec rsmt2d.Codec) [][]b
 	return append(raw, erasuredData...)
 }
 
-// this code is copy pasted from the plugin, and should likely be exported in the plugin instead
+// generateRandNamespacedRawData returns random data of length total. Each chunk
+// of random data is of size nidSize + leafSize.
 func generateRandNamespacedRawData(total int, nidSize int, leafSize int) [][]byte {
```
rootulp, line 147: FWIW this could be simplified to:

```go
func generateRandNamespacedRawData(numShares int, shareSize int) (data [][]byte) {
	for i := 0; i < numShares; i++ {
		share := make([]byte, shareSize)
		_, err := rand.Read(share)
		if err != nil {
			panic(err)
		}
		data = append(data, share)
	}
	sortByteArrays(data)
	return data
}
```

evan-forbes (approved):

pkg/wrapper/nmt_wrapper_test.go
staheri14: There is just a slight difference and that is the final `data` is sorted both based on namespace ID and their data whereas in the original implementation it is not the case.

staheri14 (approved):

PR #1438, opened 2023-03-02
feat!: removes the `Tree` method from the NMT wrapper and introduces the `ProveRange` API

staheri14:
## Overview
While reviewing the NMT wrapper implementation and attempting to write its specs, came across a contradiction in the wrapper's design. On one hand, the wrapper is intended to encapsulate an instance of NMT and re-define/modify some of its APIs, specifically the [`Push`](https://github.com/celestiaorg/celestia-app/blob/6d27b78aa64a749a808e84ea682352b8b551fbd7/pkg/wrapper/nmt_wrapper.go#L80) method. On the other hand, the wrapper also exposes the [`Tree()`](https://github.com/celestiaorg/celestia-app/blob/6d27b78aa64a749a808e84ea682352b8b551fbd7/pkg/wrapper/nmt_wrapper.go#L112) method, which returns a pointer to the underlying NMT tree and thus allows direct access to the tree and enables arbitrary manipulation of the underlying tree outside of the intended control of the wrapper. For instance, one can bypass the NMT wrapper `Push` method and directly call the `Push` method of the underlying tree using `.Tree().Push()`. This not only widens the error surface but also violates the initial objective of the wrapper.

To address this issue, the `Tree()` API has been removed, and all necessary methods of the underlying NMT are now exposed through the wrapper directly. 

## Checklist

### DIFF
--- a/pkg/proof/proof.go
+++ b/pkg/proof/proof.go
@@ -162,7 +162,7 @@ func NewShareInclusionProof(
 		}
 
 		rawShares = append(rawShares, shares.ToBytes(row[startLeafPos:endLeafPos+1])...)
-		proof, err := tree.Tree().ProveRange(int(startLeafPos), int(endLeafPos+1))
+		proof, err := tree.ProveRange(int(startLeafPos), int(endLeafPos+1))
 		if err != nil {
 			return types.ShareProof{}, err
 		}
--- a/pkg/wrapper/nmt_wrapper.go
+++ b/pkg/wrapper/nmt_wrapper.go
@@ -104,10 +104,12 @@ func (w *ErasuredNamespacedMerkleTree) Root() []byte {
 	return w.tree.Root()
 }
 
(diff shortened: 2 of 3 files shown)


rootulp (approved):
LGTM!

evan-forbes (approved):
hmmm I couldn't find anywhere where this is used, and I don't remember why this was added in the first place, so it makes sense to remove it to me! 

good find!

staheri14:
> hmmm I couldn't find anywhere where this is used, and I don't remember why this was added in the first place, so it makes sense to remove it to me!
> 


Indeed! there was no significant use case for it. It was only utilized [here](https://github.com/celestiaorg/celestia-app/blob/62eae8774efeb46036fe4d7fce3fd571a89a5b77/pkg/proof/proof.go#L165) to invoke the `ProveRange` method of the underlying NMT. However, it could have been defined directly as a method for the NMT wrapper, which has been implemented as part of this pull request.

PR #1441, opened 2023-03-02
specs: auto generate table of contents

rootulp:
## Motivation 
The table of contents at the top of each file can be a maintenance burden. We can auto-generate it with [mdbook-toc](https://github.com/badboy/mdbook-toc). 

## Screenshot

Before | After
--- | ---
before | after


evan-forbes (approved):
utACK! :+1:

Issue #1296, opened 2023-01-25
EPIC: Specify data root construction using rsmt2d, nmt, and the nmt wrapper


evan-forbes:
Having a spec for the construction of the data root, particularly one that is consistent with the current implementation, would be very valuable. It would help on boarding others (internally such as new engineers, and externally via auditors and research enthusiasts) on how celestia functions at a low level. Ideally, someone should be able to write a fully compatible separate implementation using such a spec.

Currently, a lot of this information is disparate or implied. We need to compile the various sources of information, make sure they follow the implementation, and then write the missing pieces.

Notably, some of the missing pieces include (incomplete list) todo: flesh out
- [x] Spec out the NMT library https://github.com/celestiaorg/nmt/pull/101
- [x] Spec out the functionality of the nmt wrapper https://github.com/celestiaorg/celestia-app/pull/1442

Other issues that are addressed or identified as part of this EPIC:
- https://github.com/celestiaorg/nmt/issues/95
- https://github.com/celestiaorg/nmt/issues/112
- https://github.com/celestiaorg/nmt/issues/99
- https://github.com/celestiaorg/nmt/issues/108
- https://github.com/celestiaorg/nmt/issues/109
- https://github.com/celestiaorg/nmt/issues/110
- https://github.com/celestiaorg/nmt/issues/111
- https://github.com/celestiaorg/nmt/issues/115
- https://github.com/celestiaorg/nmt/issues/118
- https://github.com/celestiaorg/nmt/issues/125
-  and https://github.com/celestiaorg/nmt/pull/126
- https://github.com/celestiaorg/nmt/pull/94
- https://github.com/celestiaorg/nmt/pull/92
- https://github.com/celestiaorg/nmt/pull/96
- https://github.com/celestiaorg/nmt/pull/102
- https://github.com/celestiaorg/nmt/pull/103
- https://github.com/celestiaorg/nmt/pull/113
- https://github.com/celestiaorg/nmt/pull/114
- https://github.com/celestiaorg/nmt/pull/116
- https://github.com/celestiaorg/nmt/pull/119
- https://github.com/celestiaorg/nmt/pull/120
- https://github.com/celestiaorg/celestia-app/pull/1438
- https://github.com/celestiaorg/celestia-app/pull/1420

staheri14:
Adding some resources relevant to this EPIC: 
[...]

PR #1442, opened 2023-03-02
chore: adds NMT Wrapper specifications

staheri14:
## Overview
This PR addresses issue #1296 by adding the specification for the NMT wrapper. The specification is based on the state of implementation after the inclusion of the changes made in pull request #1438.

## Checklist


docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,84 @@
+# Namespaced Merkle Tree Wrapper
+## Abstract
+In Celestia, block transactions are grouped into identical-size shares and arranged in a `k` by `k` matrix, called [original data square](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/data_structures.md#arranging-available-data-into-shares).
+The rows and columns of the original data square is then extended using 2D Reed-Solomon coding scheme.
```
evan-forbes, line 4:
```suggestion
The rows and columns of the original data square are then extended using a 2D Reed-Solomon coding scheme.
```

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,84 @@
...
+ -------   -------
+|        |        |
+|   Q0 → |   Q1   | k
+|   ↓    |        |
+-------  -------
+|        |        |
+|   Q2 → |   Q3   | k
+|        |        |
+ -------   -------
+```
```
evan-forbes, line 21: [non-blocking comment]

is there another place where we describe the square? perhaps we can link there as well, but

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,84 @@
...
+The namespace ID of a share depends on the quadrant it belongs to.
+
+Shares in the original data square `Q0` already carry their namespace IDs, which are located in the first `NamespaceIDSize` bytes of the share. 
+Thus, the namespace ID of a share in `Q0` can be extracted from the first `NamespaceIDSize` bytes of the share.
+
+However, shares in `Q1`, `Q2`, and `Q3` (that are erasure-coded versions of `Q0` and known as parity shares) do not have any namespace IDs by default. 
+These shares must be assigned a reserved namespace ID, which is called `ParitySharesNamespaceID`.
+`ParitySharesNamespaceID` corresponds to the maximum value representable by `NamespaceIDSize` bytes.
+
+In Celestia, as of the writing of this specification, the `NamespaceIDSize` (which maps to the [`NAMESPACE_ID_BYTES`](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/consensus.md#constants) constant) is `8` bytes, and therefore, [`ParitySharesNamespaceID`](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/consensus.md#reserved-namespace-ids)  is represented by `8` bytes of `0xFF`.
```
evan-forbes, line 55: [non-blocking]

Instead of specifying 8 here, perhaps we could simply say that the parity namespace is the lexicographically last namespace. Either that, or update to 33 now cc @rootulp

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,84 @@
...
+If the added share is not within `Q0` i.e, it is a parity share, then it is prepended with the `ParitySharesNamespaceID` before getting pushed into the tree.
+
+
+**Some insightful observations can be made from the NMT wrapper description**:
+- In the NMT wrapper, the shares are extended in size by `NamespaceIDSize` bytes before insertion to the tree. 
+- For every row and column that overlaps with `Q0`, it is the case that the shares in the first half of the tree leaves  belong to `Q0`, whereas the second half of the leaves are the erasure coded version of the first half.
+ This means, the second half of the tree leaves all have identical namespace IDs, i.e., `ParitySharesNamespaceID`.
+- Each leaf in the NMT wrapper that corresponds to shares in `Q0` has a doubly namespaced structure. 
+Specifically, the underlying data of the leaf contains the namespace ID of the share twice.
```
evan-forbes, line 77: :+1:

evan-forbes (commented):
only left one suggestion that I _think_ is a typo, but if not we can ignore

docs/spec/nmt-wrapper.md
staheri14: I have rephrased this part to eliminate the use of a specific number for the namespace ID size in Celestia. By doing this, we can ensure that this specification does not need to be updated every time there is a change to the namespace size in Celestia. Please see https://github.com/celestiaorg/celestia-app/pull/1442/commits/d0519e7661a14c510b47de30a361ca7a23c55075.

docs/spec/nmt-wrapper.md
staheri14: Addressed https://github.com/celestiaorg/celestia-app/pull/1442/commits/d0519e7661a14c510b47de30a361ca7a23c55075.

docs/spec/nmt-wrapper.md
staheri14: Addressed https://github.com/celestiaorg/celestia-app/pull/1442/commits/d0519e7661a14c510b47de30a361ca7a23c55075.

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,83 @@
...
+NMTs require the data items they represent to be namespaced, which means that the shares within each row or column of the extended data square must be namespaced before being added to the NMT.
+This is where the [Namespaced MerkleTree Wrapper](https://github.com/celestiaorg/celestia-app/blob/main/pkg/wrapper/nmt_wrapper.go)  comes into play.
+It is a data structure that wraps around the [Namespaced Merkle Tree](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md) and ensures the proper namespaces are prepended to the shares  before they are added to their respective row or column NMT.
+In this specification, we elaborate on the design and structure of the Namespace Merkle Tree wrapper.
+
+
+
+## Namespaced Merkle Tree Wrapper
+Namespaced Merkle Tree Wrapper is used to represent a row or column of Celestia extended data square.
+At its core, it is an [NMT](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md) and is defined by a similar set of parameters.
```
MSevey, line 34: [nit] if you reference the same link multiple times you can use an id to avoid needing to edit the link in multiple places. 

```markdown
[Namespace Merkle Tree][nmtlink]
[NMT][nmtlink]

...

[nmtlink] : https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md
```

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,83 @@
+# Namespaced Merkle Tree Wrapper
+## Abstract
+In Celestia, block transactions are grouped into identical-size shares and arranged in a `k` by `k` matrix, called [original data square](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/data_structures.md#arranging-available-data-into-shares).
```
rootulp, line 3: [optional] I expect this link to break when we merge the specs-staging branch to main and delete the branch. To avoid the broken link we can use permalinks. 

```suggestion
In Celestia, block transactions are grouped into identical-size shares and arranged in a `k` by `k` matrix, called [original data square](https://github.com/celestiaorg/celestia-app/blob/c09843d07d4c3842753138de96b304b4866e8f5d/specs/src/specs/data_structures.md?plain=1#L494).
```

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,83 @@
...
+|        |        |
+|   Q2 → |   Q3   | k
+|        |        |
+ -------   -------
+```
+Figure 1.
+
+Each row and column of the extended data square is modeled by a [Namespace Merkle Tree](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md).
+NMTs require the data items they represent to be namespaced, which means that the shares within each row or column of the extended data square must be namespaced before being added to the NMT.
+This is where the [Namespaced MerkleTree Wrapper](https://github.com/celestiaorg/celestia-app/blob/main/pkg/wrapper/nmt_wrapper.go)  comes into play.
```
rootulp, line 26: [question] the linked file only has a `ErasuredNamespacedMerkleTree`, is that what this is referring to? If so, is this alright:

```suggestion
This is where the [ErasuredNamespacedMerkleTree](https://github.com/celestiaorg/celestia-app/blob/main/pkg/wrapper/nmt_wrapper.go) comes into play.
```

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,83 @@
...
+
+
+
+## Namespaced Merkle Tree Wrapper
+Namespaced Merkle Tree Wrapper is used to represent a row or column of Celestia extended data square.
+At its core, it is an [NMT](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md) and is defined by a similar set of parameters.
+Namely, the [namespace ID size `NamespaceIDSize`](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md#nmt-data-structure), 
+the underlying hash function for the digest calculation of the [namespaced hashes](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md#namespaced-hash), 
+and the [`IgnoreMaxNamespace` flag](https://github.com/celestiaorg/nmt/blob/master/spec/nmt.md#ignore-max-namespace) which dictates how namespace range of non-leaf nodes are calculated.
+In addition, the NMT wrapper is configured with the data square size `SqaureSize` (`k` in the above example), and the index of the row or column it represents `AxisIndex` which is a value in `[0, SquareSize)`.
```
rootulp, line 38: [silly question] I see that `axisIndex` is the row or column index but I don't see something in `ErasuredNamespacedMerkleTree` that indicates if it is a row or column. I don't understand how the erasured NMT knows which quadrant a leaf is being pushed to. 

For example if `squareSize=4`, `axisIndex=3`, and the first first leaf is being pushed to an NMT. How does the NMT know if it is pushing to Q1 (if this is column 3 of EDS) or Q2 (if this is row 3 of EDS)? Perhaps it doesn't matter because in either case this is parity data but I'm curious if it matters. If the NMT doesn't know which quadrant is being pushed, then I think this [godoc](https://github.com/celestiaorg/celestia-app/blob/514c8dd37ac2c4dcee0c653d20f1d9020f34075a/pkg/wrapper/nmt_wrapper.go#L28-L29) is misleading.

[nit] since the fields are unexported, proposal to use the lowercase versions so that the variable here matches the variable in the file. Also typo in squareSize:

```suggestion
In addition, the NMT wrapper is configured with the data square size `squareSize` (`k` in the above example), and the index of the row or column it represents `axisIndex` which is a value in `[0, SquareSize)`.
```

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,83 @@
...
+To understand the NMT wrapper data structure, it's important to first understand how shares are namespaced in Celestia.
+The namespace ID of a share depends on the quadrant it belongs to.
+
+Shares in the original data square `Q0` already carry their namespace IDs, which are located in the first `NamespaceIDSize` bytes of the share. 
+Thus, the namespace ID of a share in `Q0` can be extracted from the first `NamespaceIDSize` bytes of the share.
+
+However, shares in `Q1`, `Q2`, and `Q3` (that are erasure-coded versions of `Q0` and known as parity shares) do not have any namespace IDs by default. 
+These shares must be assigned a reserved namespace ID, which is called `ParitySharesNamespaceID`.
+The `ParitySharesNamespaceID` corresponds to the lexicographically last namespace ID that can be represented by `NamespaceIDSize` bytes.
+"If `NamespaceIDSize` is `8`, then the value of `ParitySharesNamespaceID` is `2^64-1`, which is equivalent to `0xFFFFFFFFFFFFFFFF`.
```
rootulp, line 54:
```suggestion
If `NamespaceIDSize` is `8`, then the value of `ParitySharesNamespaceID` is `2^64-1`, which is equivalent to `0xFFFFFFFFFFFFFFFF`.
```

docs/spec/nmt-wrapper.md
rootulp: I think my silly question can be disregarded b/c later in this document it states

> During the insertion, the wrapper first checks whether the share is within the original data square or not.

so the wrapper doesn't actually know which quadrant the leaf is being pushed to.

rootulp (approved):
No blocking feedback so LGTM.

The comment about permalinks applies to more than one link in the doc but defer to the author on resolving that now or updating the links at a later point if/when they break

docs/spec/nmt-wrapper.md
MSevey, line 3: > identical-size shares

Do we want to call out which layer owns defining that size? 

A related question that may be for just for my understanding, what, if any, are the notable impacts to changing the share size?

docs/spec/nmt-wrapper.md
```diff
@@ -0,0 +1,83 @@
+# Namespaced Merkle Tree Wrapper
+## Abstract
+In Celestia, block transactions are grouped into identical-size shares and arranged in a `k` by `k` matrix, called [original data square](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/data_structures.md#arranging-available-data-into-shares).
+The rows and columns of the original data square are then extended using a 2D Reed-Solomon coding scheme.
+The [extended version](https://github.com/celestiaorg/celestia-app/blob/specs-staging/specs/src/specs/data_structures.md#2d-reed-solomon-encoding-scheme) is called extended data square, that is a `2k` by `2k` square consisting of `4` quadrants namely, `Q0`, `Q1`, `Q2`, and `Q3`.
+Figure 1 provides an illustration of a sample data square and its extended version.
+`Q0` corresponds to the original data square.
+`Q1` and `Q2` represent the horizontal and vertical extensions of `Q0`, respectively. 
+`Q3` is the extension of `Q2` in the horizontal direction.
```
MSevey, line 9: Should we talk about if there is a reason Q2 is extending horizontally to form Q3 vs Q1 being extended vertically to make Q3?

docs/spec/nmt-wrapper.md
rootulp: > what, if any, are the notable impacts to changing the share size?

Relevant context: https://github.com/celestiaorg/celestia-app/issues/711

Shares are the unit that gets downloaded by light clients during DAS so:
- Increasing the share size increases the amount of data a light client must download per sample
- Decreasing the share size decreases the amount of data a light client must download per sample

Each share has a fixed size prefix (e.g. namespace ID, info byte) and some variable prefix depending on the type of the share so:
- Increasing the share size decreases the % of overhead occupied by the prefix. 
- Decreasing the share size increases the % of overhead occupied by the prefix.

Additionally if all other variables remain constant (i.e. max square size):
- Increasing the share size increases the capacity for data in the data square
- Decreasing the share size decreases the capacity for data in the data square

docs/spec/nmt-wrapper.md
staheri14: Good question and correct answer! The NMT wrapper is designed to only identify if shares fall within Q0. If it were to identify shares in other quadrants, it would require a row or column indicator to distinguish shares in Q1 from those in Q3.

docs/spec/nmt-wrapper.md
staheri14: Fixed, please see https://github.com/celestiaorg/celestia-app/pull/1442/commits/22e737c961c551e85166cfd9ecb455d248c5d8c4.

docs/spec/nmt-wrapper.md
staheri14: Yes, the NMT wrapper refers to the `ErasuredNamespacedMerkleTree` struct, I revised the text to make it clear.  Please see https://github.com/celestiaorg/celestia-app/pull/1442/commits/22e737c961c551e85166cfd9ecb455d248c5d8c4.

docs/spec/nmt-wrapper.md
staheri14: Good point, I fixed all the links accordingly. Please see https://github.com/celestiaorg/celestia-app/pull/1442/commits/22e737c961c551e85166cfd9ecb455d248c5d8c4.

docs/spec/nmt-wrapper.md
staheri14: Thanks for the tips, I now defined link references for all the urls, please see https://github.com/celestiaorg/celestia-app/pull/1442/commits/22e737c961c551e85166cfd9ecb455d248c5d8c4

docs/spec/nmt-wrapper.md
staheri14: That is a good question! The mechanics of erasure coding are not within the scope of the NMT wrapper, and this logic is implemented in the caller application.  I recommend elaborating on this topic in the following document, which explains the erasure coding of the original data square: https://github.com/celestiaorg/celestia-app/blob/c09843d07d4c3842753138de96b304b4866e8f5d/specs/src/specs/data_structures.md#2d-reed-solomon-encoding-scheme.
cc: @evan-forbes @rootulp what do you think?

staheri14: > Should we talk about if there is a reason Q2 is extending horizontally to form Q3 vs Q1 being extended vertically to make Q3?

One more point: The description of the extended data square construction is provided in the abstract for additional context, if you think this causes confusion, I can either revise or remove that part.

docs/spec/nmt-wrapper.md
staheri14: > Do we want to call out which layer owns defining that size?

AFAIK, the maximum value of `k` is governed by the [`AVAILABLE_DATA_ORIGINAL_SQUARE_MAX`](https://github.com/celestiaorg/celestia-app/blob/c09843d07d4c3842753138de96b304b4866e8f5d/specs/src/specs/consensus.md#constants) constant among other Celestia consensus rule constants. However, the actual value of `k` depends on the number of shares in each block and is the closest power of 2 to a square enclosing all the shares.
Quoted from [Header description](https://github.com/celestiaorg/celestia-app/blob/c09843d07d4c3842753138de96b304b4866e8f5d/specs/src/specs/data_structures.md#header)
> The size of the [original data square](https://github.com/celestiaorg/celestia-app/blob/c09843d07d4c3842753138de96b304b4866e8f5d/specs/src/specs/data_structures.md#arranging-available-data-into-shares), availableDataOriginalSquareSize, isn't explicitly declared in the block header. Instead, it is implicitly computed as the smallest power of 2 whose square is at least availableDataOriginalSharesUsed (in other words, the smallest power of 4 that is at least availableDataOriginalSharesUsed).

staheri14: > Do we want to call out which layer owns defining that size?

Apologies for the confusion. In my previous response, I misunderstood your question and provided information about the `k` value. To clarify, the share size in Celestia is also determined by a constant called `SHARE_SIZE`, which is defined as part of the Celestia consensus rule (AFAIK).

staheri14:
I would appreciate it if you could kindly provide any further feedback you may have on the specification. If there are no further comments or changes needed, I plan to merge it tomorrow.

docs/spec/nmt-wrapper.md
staheri14: Added clarification about the share size and `k` value in this update https://github.com/celestiaorg/celestia-app/pull/1442/commits/0ad159a7d60d0fdcf7b560d3796b4ed6f68cf9e8.

docs/spec/nmt-wrapper.md
staheri14: > since the fields are unexported, proposal to use the lowercase versions so that the variable here matches the variable in the file.

Thank you for the suggestion. My intention was to describe the wrapper as a general data structure without being specific to its implementation, which is why the capitalization did not match the implementation (this was intentional).

docs/spec/nmt-wrapper.md
evan-forbes: > I recommend elaborating on this topic in the following document, which explains the erasure coding of the original data square:

100% agree

rootulp: +1 agreed, seems out of scope for NMT wrapper but in-scope for specs. 

> Should we talk about if there is a reason Q2 is extending horizontally to form Q3 vs Q1 being extended vertically to make Q3?

My impression from @adlerjohn 's [whiteboard series w/ NEAR](https://youtu.be/jM-om3AqH94?t=2970) is that it doesn't matter if Q3 is extended from Q1 or Q2.

rootulp (dismissed):
Great work!

docs/spec/nmt-wrapper.md
MSevey: sounds good, I thought it might apply somewhere else.

if you've captured the follow up then you can resolve this 👍

docs/spec/nmt-wrapper.md
MSevey: sounds like this similarly applies more elsewhere, so I'll resolve.

MSevey (dismissed):
LGTM 👍

docs/spec/nmt-wrapper.md
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/1454

evan-forbes (approved):

Issue #1403, opened 2023-02-17
Run all ci linters on `make lint`


rootulp:
CI runs four linters in the [lint workflow](https://github.com/celestiaorg/celestia-app/blob/f19c2ff1c223d446cd86dd7fc0d2fcb9e32b645b/.github/workflows/lint.yml):
1. golangci-lint
2. markdownlint
3. hadolint (for Dockerfile)
4. yamllint

[`make lint`](https://github.com/celestiaorg/celestia-app/blob/f19c2ff1c223d446cd86dd7fc0d2fcb9e32b645b/Makefile#L82-L83) only runs the first two but not the last two. We should consider adding them to `make lint` so that developers can get faster feedback if their changes violate any linters. Note this change may expect yamllint and hadolint binaries installed on the developer's machine so we likely want to include install steps in the README [here](https://github.com/celestiaorg/celestia-app#tools).

PR #1447, opened 2023-03-07
chore: run hadolint and yamllint during `make lint`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1403

yamllint config copied from https://github.com/celestiaorg/.github/blob/main/.yamllint.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ See <https://docs.celestia.org/category/celestia-app> for more information
 
 1. Install [golangci-lint](https://golangci-lint.run/usage/install/)
 1. Install [markdownlint](https://github.com/DavidAnson/markdownlint)
+1. Install [hadolint](https://github.com/hadolint/hadolint)
+1. Install [yamllint](https://yamllint.readthedocs.io/en/stable/quickstart.html)
 
 ### Helpful Commands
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -72,6 +72,8 @@ See <https://docs.celestia.org/category/celestia-app> for more information
 
 1. Install [golangci-lint](https://golangci-lint.run/usage/install/)
 1. Install [markdownlint](https://github.com/DavidAnson/markdownlint)
+1. Install [hadolint](https://github.com/hadolint/hadolint)
+1. Install [yamllint](https://yamllint.readthedocs.io/en/stable/quickstart.html)
```
cmwaters, line 76: nit: shouldn't the numbers be incrementing (or why are they there)

cmwaters: We could also do something like https://github.com/cometbft/cometbft/blob/main/tools/tools.go which, IIUC, will install these tools so developers don't need to manually install them

cmwaters (approved):
LGTM

README.md
rootulp: Markdown will render an ordered list with all leading `1.` with the correct number. This make it easy to re-arrange the lines in the list without manually updating each number.

Ref: https://www.markdownguide.org/basic-syntax/#ordered-lists

The idea from cometbft sounds good to me. Unfortunately in the list above, only golangci-lint is written in Go.

cmwaters: > Markdown will render an ordered list with all leading 1. with the correct number. This make it easy to re-arrange the lines in the list without manually updating each number.

TIL

Issue #1455, opened 2023-03-08
Auto specify project when creating new issues in celestia-app/core


rootulp:
In our grooming meeting today, we discussed how it would be helpful if Github automatically added new issues that were created in celestia-core / celestia-app to the core/app GH project.

rootulp:
related: https://github.com/celestiaorg/celestia-node/pull/1877

PR #1456, opened 2023-03-08
chore: issue label and project automation

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1455

The issue label automation was inspired by https://github.com/celestiaorg/celestia-node/pull/1877
The project automation uses [add-to-project](https://github.com/actions/add-to-project)

### DIFF
--- a/.github/workflows/issue-label-automation.yml
+++ b/.github/workflows/issue-label-automation.yml
@@ -1,6 +1,6 @@
 name: issue-label-automation
 on:
-  pull_request:
+  pull_request_target:
     types: [opened]
   issues:
     types: [opened]
--- a/.github/workflows/issue-label-automation.yml
+++ b/.github/workflows/issue-label-automation.yml
@@ -1,0 +1,38 @@
+name: issue-label-automation
+on:
+  pull_request:
+    types: [opened]
+  issues:
+    types: [opened]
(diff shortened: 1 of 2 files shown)


rootulp:
I don't see either of the automation CI checks running on this PR so I don't have confidence that they work as expected. It could be they only run after this merges but I'm suspicious

MSevey:
> I don't see either of the automation CI checks running on this PR so I don't have confidence that they work as expected. It could be they only run after this merges but I'm suspicious

Yea I haven't fully figured out when actions PRs run vs needing to be merged in first. 

Also a second reason why it might not have run is because we are intentionally using the `open` trigger. So PR updates aren't going to trigger it.  iirc the reason for that is for labels we don't want the action to re-add labels we removed for example. Plus there is no reason it needs to run more than once.

evan-forbes (approved):
utACK and makes sense afaiu 

>I don't see either of the automation CI checks running on this PR so I don't have confidence that they work as expected. It could be they only run after this merges but I'm suspicious

same, unfortunately there's only one way to test this lol and we can adjust with future PRs. we can also try on other repos (not forks, since github treats forks very differently)

rootulp:
I'm going to test in prod 😆

Issue #1443, opened 2023-03-03
Consider refactoring `Share` type


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/e6bbc302194376dbfd909cfabbf82e16e8fd5e7f/pkg/shares/shares.go#L11-L19

## Problem

Currently `Share` is a byte slice. Since the byte slice doesn't guarantee a length, methods like `share.NamespaceID()` and `share.InfoByte()` must perform checks on the length of the byte slice prior to returning information. In other words, since the byte slice can't enforce invariants like `len(share) == appconsts.ShareSize`, all methods that deal with shares must do additional error handling.

## Proposal

We may be able to refactor the `Share` type to be a struct that enforces invariatns (e.g. share length). On a related note, we may be able to leverage a [builder pattern](https://refactoring.guru/design-patterns/builder/go/example) to encapsulate the logic for constructing shares. Currently a large portion of the share package must operate on byte slices so when we make changes to the share format, several files need to be updated. We could constrain these modifications to a single file (i.e. `share.go`) if the `Share` type didn't allow direct byte manipulation and instead clients were forced to operate on create / read contents of the share through public methods.

rootulp:
@mojtaba-esk was this closed by https://github.com/celestiaorg/celestia-app/pull/1462?

PR #1462, opened 2023-03-09
chore: share encoding refactor

mojtaba-esk:
context in the issue #1443 

## Checklist


pkg/shares/parse_compact_shares.go
```diff
@@ -30,79 +26,44 @@ func parseCompactShares(rawShares [][]byte, supportedShareVersions []uint8) (dat
...
 }
 
-func newShareStack(shares [][]byte) *shareStack {
-	return &shareStack{shares: shares}
-}
+// peel parses each unit of data (either a transaction or
+// intermediate state root) and adds it to the underlying slice of data.
+// data may be transactions or intermediate state roots depending
+// on the namespace ID for this share
+func peel(shares []Share) (data [][]byte, err error) {
```
mojtaba-esk, line 36: @rootulp @evan-forbes please review this function carefully to see if there are some edge cases that we need to consider in it.

pkg/shares/share_builder.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// writeSequenceLen writes the sequence length to the first share.
+func (b *Builder) WriteSequenceLen(sequenceLen uint32) error {
+	if !b.isFirstShare {
+		return errors.New("not the first share")
+	}
+	if b == nil {
+		return errors.New("the builder object is not initialized (is nil)")
+	}
```
rootulp, line 145: [question] when will this error be thrown? If the builder object isn't initialized, the error above will be thrown.

```go
func TestWriteSequenceLen(t *testing.T) {
	b := Builder{}
	got := b.WriteSequenceLen(10)
	fmt.Printf("got: %v", got) // got: not the first share
	assert.Error(t, got)
}
```

rootulp (commented):
Overall I like this approach! I think the builder would benefit from unit tests.

In a separate PR we may consider enforcing invariants on the share type (i.e. is a byte slice of length 512) and on a related note enforce that the share type can only be constructed through this share builder. Thoughts?

pkg/shares/share_builder.go
```diff
@@ -0,0 +1,205 @@
...
+
+type Builder struct {
+	namespace      namespace.ID
+	shareVersion   uint8
+	isFirstShare   bool
+	isCompactShare bool
+	rawShareData   []byte
+}
+
+func NewBuilder(ns namespace.ID, shareVersion uint8, isCompactShare bool, isFirstShare bool) *Builder {
```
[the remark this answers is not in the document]
mojtaba-esk: Sounds a good idea, I am always a fan of simplicity :)

pkg/shares/share_builder.go
```diff
@@ -0,0 +1,205 @@
...
+func (b *Builder) Build() (*Share, error) {
+	return NewShare(b.rawShareData)
+}
+
+// IsEmptyShare returns true if no data has been written to the share
+func (b *Builder) IsEmptyShare() bool {
+	if b.isFirstShare {
+		return len(b.rawShareData) == appconsts.NamespaceSize+appconsts.ShareInfoBytes+appconsts.SequenceLenBytes+appconsts.CompactShareReservedBytes
+	}
+	return len(b.rawShareData) == appconsts.NamespaceSize+appconsts.ShareInfoBytes+appconsts.CompactShareReservedBytes
```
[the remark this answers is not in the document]
mojtaba-esk: Yes you are right. I will fix it and the builder is impolitely tested in some other unit tests, but I think having a separate unit test for it, gives us opportunity to test more scenarios.

pkg/shares/parse_compact_shares.go
rootulp: [optional] I found it difficult to comprehend the recursive `peel` functionality so I wrote a non-recursive implementation. It hasn't been unit tested but was inspired by the implementation in this PR

```
package shares

import (
	"bytes"
	"fmt"
)

// parseCompactShares returns data (transactions or intermediate state roots
// based on the contents of rawShares and supportedShareVersions. If rawShares
// contains a share with a version that isn't present in supportedShareVersions,
// an error is returned. The returned data [][]byte does not have namespaces,
// info bytes, data length delimiter, or unit length delimiters and are ready to
// be unmarshalled.
func parseCompactShares(shares []Share, supportedShareVersions []uint8) (data [][]byte, err error) {
	if len(shares) == 0 {
		return nil, nil
	}
	err = validateShareVersions(shares, supportedShareVersions)
	if err != nil {
		return nil, err
	}

	rawData, err := extractRawData(shares)
	if err != nil {
		return nil, err
	}

	sequenceLen, err := shares[0].SequenceLen()
	if err != nil {
		return nil, err
	}

	if int(sequenceLen) != len(rawData) {
		return nil, fmt.Errorf("sequence length %v does not match the sum of the lengths of the raw data in the shares %v", sequenceLen, len(rawData))
	}

	data, err = parseRawData(rawData)
	if err != nil {
		return nil, err
	}
	return data, nil
}

// validateShareVersions returns an error if the shares contain a share with an
// unsupported share version. Returns nil if all shares contain supported share
// versions.
func validateShareVersions(shares []Share, supportedShareVersions []uint8) error {
	for _, share := range shares {
		infoByte, err := share.InfoByte()
		if err != nil {
			return err
		}
		if !bytes.Contains(supportedShareVersions, []byte{infoByte.Version()}) {
			return fmt.Errorf("unsupported share version %v is not present in the list of supported share versions %v", infoByte.Version(), supportedShareVersions)
		}
	}
	return nil
}

// parseRawData returns the units (transactions, PFB transactions, intermediate
// state roots) contained in raw data by parsing the unit length delimiter
// prefixed to each unit.
func parseRawData(rawData []byte) (units [][]byte, err error) {
	units = make([][]byte, 0)
	for {
		actualData, unitLen, err := ParseDelimiter(rawData)
		if err != nil {
			return nil, err
		}
		if unitLen == 0 {
			return units, nil
		}
		rawData = actualData[unitLen:]
		units = append(units, actualData[:unitLen])
	}
}

// extractRawData returns the raw data contain in the shares. The raw data does
// not contain the namespace ID, info byte, sequence length, or reserved bytes.
func extractRawData(shares []Share) (rawData []byte, err error) {
	for _, share := range shares {
		raw, err := share.RawData()
		if err != nil {
			return nil, err
		}
		rawData = append(rawData, raw...)
	}
	return rawData, nil
}
```

evan-forbes:
will review this by eod!!

pkg/shares/parse_compact_shares.go
rootulp: > Do you want me to replace my peel function with this 

Sorry for the lack of clarity. I defer to you on which one is clearer but I agree unit tests would increase our confidence on merging either implementation so will let you add tests if you're comfortable doing so.

pkg/shares/split_compact_shares.go
```diff
@@ -78,59 +67,42 @@ func (css *CompactShareSplitter) WriteTx(tx coretypes.Tx) {
...
+		if !css.shareBuilder.IsEmptyShare() {
 			css.shares = css.shares[:len(css.shares)-1]
 		}
 		css.done = false
 	}
 
-	css.maybeWriteReservedBytesToPendingShare()
+	if err := css.shareBuilder.MaybeWriteReservedBytes(); err != nil {
+		panic(err)
+	}
```
evan-forbes, line 78: [optional]

leaving as optional since we should only ever reach this point if there's an invalid state, but it does limit the safe uses of this library. 

I think there are a few places where we could bubble up these errors with the only consequence being something that's slightly less aesthetic, since all of there uses are in functions that can or already are returning an error

pkg/shares/share_builder.go
```diff
@@ -0,0 +1,220 @@
...
+// indexOfInfoBytes returns the index of the InfoBytes.
+func (b *Builder) indexOfInfoBytes() int {
+	// the info byte is immediately after the namespace
+	return appconsts.NamespaceSize
+}
+
+// MaybeWriteReservedBytes will be a no-op if the reserved bytes
+// have already been populated. If the reserved bytes are empty, it will write
+// the location of the next unit of data to the reserved bytes.
+func (b *Builder) MaybeWriteReservedBytes() error {
```
evan-forbes, line 126: if we're going to throw an error and panic on that error, we should document when we can error.

evan-forbes (commented):
overall this is a really really good refactor!! well done @mojtaba-esk and @rootulp 

I do think there are some instances where we are panicking, and if we want this to be a library that can be used by others outside of celestia-app, then I think those need to be removed. Otherwise its way too easy to accidently write a bug where someone can shutdown your light node/whatever. I also think most can be bubbled easily or with little consequence to other current apis, but even if that's not the case, we should avoid panics at all costs for all user facing things. We can **definitely** do this in a different PR though!

since this is a refactor and we're not in a particular rush, we could take advantage of the new design to add unit tests where previously we couldn't. With any encoding code changes, we almost need fuzzers, as there are quite a few edge cases, even after this refactor removed some.

app/test/process_proposal_test.go
```diff
@@ -272,12 +272,14 @@ func TestProcessProposal(t *testing.T) {
...
-	return share
+	// input params do not matter as we import the entire share in raw format
+	b := shares.NewEmptyBuilder().ImportRawShare(share.ToBytes())
+	b.FlipSequenceStart()
+	updatedShare, err := b.Build()
+	if err != nil {
+		panic(err)
+	}
+	return *updatedShare
 }
```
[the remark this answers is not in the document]
mojtaba-esk: yes you're right, I will apply it

pkg/shares/split_compact_shares.go
mojtaba-esk: good idea. 
I am going to get rid of all panics in a separate PR.

evan-forbes (dismissed):
:shipit: !

pkg/shares/parse_compact_shares.go
```diff
@@ -1,39 +1,17 @@
...
-)
+import "errors"
 
 // parseCompactShares returns data (transactions or intermediate state roots
 // based on the contents of rawShares and supportedShareVersions. If rawShares
 // contains a share with a version that isn't present in supportedShareVersions,
 // an error is returned. The returned data [][]byte does not have namespaces,
 // info bytes, data length delimiter, or unit length delimiters and are ready to
-// be unmarshalled.
+// be unmarshaled.
```
rootulp, line 10: [typo]

```suggestion
// be unmarshalled.
```

pkg/shares/parse_compact_shares.go
```diff
@@ -43,27 +21,63 @@ func peel(shares []Share) (data [][]byte, err error) {
...
 		if unitLen == 0 {
-			return data, nil
+			return units, nil
+		}
+		rawData = actualData[unitLen:]
+		units = append(units, actualData[:unitLen])
+	}
+}
+
+// extractRawData returns the raw data contain in the shares. The raw data does
```
rootulp, line 72:
```suggestion
// extractRawData returns the raw data contained in the shares. The raw data does
```

pkg/shares/share_builder_test.go
```diff
@@ -0,0 +1,311 @@
...
+		{
+			name:    "small share",
+			builder: NewBuilder([]byte{1, 1, 1, 1, 1, 1, 1, 1}, appconsts.ShareVersionZero, true),
+			data:    []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			want:    nil,
+		},
+		{
+			name:    "exact fit first compact share",
+			builder: NewBuilder(appconsts.TxNamespaceID, appconsts.ShareVersionZero, true),
+			data:    bytes.Repeat([]byte{1}, appconsts.ShareSize-appconsts.NamespaceSize-appconsts.CompactShareReservedBytes-appconsts.SequenceLenBytes-1 /*1 = info byte*/),
```
rootulp, line 165: [nit] I think the tests in this file can use the const `appconsts.ShareInfoBytes` instead of `-1 /*1 = info byte*/`.

```suggestion
			data:    bytes.Repeat([]byte{1}, appconsts.ShareSize-appconsts.NamespaceSize-appconsts.ShareInfoBytes-appconsts.CompactShareReservedBytes-appconsts.SequenceLenBytes),
```

rootulp (dismissed):

evan-forbes (approved):

Issue #1398, opened 2023-02-16
Remove usage of deprecated `rokroskar/workflow-run-cleanup-action`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/b40a6908954828b90aebd530c5d30592626834f7/.github/workflows/test.yml#L13-L18

## Problem

https://github.com/rokroskar/workflow-run-cleanup-action appears deprecated and we may achieve similar functionality with the `concurrency` command

## Proposal

Remove usage of the deprecated workflow

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/1468

PR #1468, opened 2023-03-11
chore: Remove usage of deprecated rokroskar/workflow-run-cleanup-action

amityadav0:
Remove usage of the deprecated workflow

#1398

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,8 +7,9 @@ on:
   workflow_call:
 
 concurrency:
+  # do not cancel jobs from earlier commits for tags or the main branch
   group: ${{ github.workflow }}-${{ !startsWith(github.ref, 'refs/tags/') &&
-      github.ref != 'refs/heads/master' }}
+      github.ref != 'refs/heads/main' }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,17 +6,12 @@ name: Tests / Code Coverage
 on:
   workflow_call:
 
-jobs:
-  cleanup-runs:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: rokroskar/workflow-run-cleanup-action@master
-        env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-    if: |
-      !startsWith(github.ref, 'refs/tags/') &&
-      github.ref != 'refs/heads/master'
+concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -6,17 +6,12 @@ name: Tests / Code Coverage
...
-    steps:
-      - uses: rokroskar/workflow-run-cleanup-action@master
-        env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-    if: |
-      !startsWith(github.ref, 'refs/tags/') &&
-      github.ref != 'refs/heads/master'
+concurrency:
+  group: ${{ github.workflow }}-${{ !startsWith(github.ref, 'refs/tags/') &&
+      github.ref != 'refs/heads/master' }}
```
rootulp, line 11: Thanks for working on this!

[question] why is the [example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) here insufficient? Is `github.ref != 'refs/heads/master'` specified so that CI jobs on earlier commits on `master` aren't cancelled if new commits are added to master? This repo uses `main` instead of `master` as its canonical branch but preference to use the example if we don't need the `!startsWith(...` logic:

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

amityadav0: Yes, this logic ensures that `master` and `tags` branches are exempted from canceling  
 the earlier commits.

amityadav0: Do you want me to update `master` to `main` or use the example as it is ?

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

rootulp: Thanks for explaining. I think we can should `master` to `main`. Can you please also include a comment above these lines with a short explanation on the rationale?

```yaml
  # do not cancel jobs from earlier commits for tags or the main branch
  group: ${{ github.workflow }}-${{ !startsWith(github.ref, 'refs/tags/') &&
      github.ref != 'refs/heads/master' }}
```

Then I think this is ready to merge 👍

amityadav0: Got it. Thanks for the review.

amityadav0: Updated

rootulp: awesome thanks!

rootulp (approved):

evan-forbes (approved):
makes sense to me! utACK thanks @amityadav0

PR #1469, opened 2023-03-12
feat!: use `utia` for gov module genesis

rootulp:
Fixes an issue identified in https://github.com/celestiaorg/itn-doc/issues/168

## Testing

Verified genesis.json contains `utia` in the gov module in [this diff](https://www.diffchecker.com/Q14DBeHB/)

## FLUPs

- [ ] Re-apply this on `main`

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -35,7 +35,6 @@ import (
 	crisiskeeper "github.com/cosmos/cosmos-sdk/x/crisis/keeper"
 	crisistypes "github.com/cosmos/cosmos-sdk/x/crisis/types"
 	distr "github.com/cosmos/cosmos-sdk/x/distribution"
-	distrclient "github.com/cosmos/cosmos-sdk/x/distribution/client"
 	distrkeeper "github.com/cosmos/cosmos-sdk/x/distribution/keeper"
 	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
 	"github.com/cosmos/cosmos-sdk/x/evidence"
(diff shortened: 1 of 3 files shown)


cmwaters:
Can't we just modify the `validate-genesis` CLI to ensure that the denom in all areas is the same instead of hardcoding this?

rootulp:
> Can't we just modify the validate-genesis CLI to ensure that the denom in all areas is the same instead of hardcoding this?

Are you proposing this as a replacement for the unit test in this PR? AFAIK we still need the implementation change in this PR because the `validate-genesis` command will only validate a genesis.json and wouldn't modify it.

rootulp:
Per [this comment](https://github.com/celestiaorg/cosmos-sdk/blob/0aef7ba8d215a3978de8c298dc40d9fa65041695/x/genutil/client/cli/validate_genesis.go#L57-L59) validate-genesis only appears to validate the Tendermint portion of genesis.json

evan-forbes (approved):

evan-forbes:
we might want to begin thinking about how to do this in a uniform and clean way across all of the modules that we want to changes the defaults of

rootulp:
Given you manually fixed the blockspace race genesis.json, we don't technically need this on the 0.12.x release branch so I'll re-target main.

I think it may add more confusion if we merge this to 0.12.x and cut a 0.12.1 release that isn't really used or necessary.

rootulp:
oof nvm that didn't work as expected b/c it introduces a bunch of other commits. Will close this and open a new PR just against `main`.

rootulp:
Deprecated in favor of https://github.com/celestiaorg/celestia-app/pull/1484

Issue #377, opened 2022-05-02
app/init: app is always looking for a user in HOME path through etc/passwd


Bidon15:
H/t @SweeXordious for helping finding the root cause

If we initialize a celestia-app instance in testground, the app is panicking in not finding the home directory, due to these lines

https://github.com/celestiaorg/celestia-app/blob/d7bb3934e0c13259bf3d666378ea67fda852b6e5/app/app.go#L156-L163

https://github.com/celestiaorg/cosmos-sdk/blob/eee26c3b6763edf8817d47b4e6695aa04c49bf9f/simapp/app.go#L179-L186

Here is the log output of the failure in testground: 
```
May  2 16:36:56.293013	INFO	starting run	{"run_id": "c9o0h7tvqc7pa50kffmg", "plan": "celestia-app", "case": "capp-1", "runner": "local:exec", "instances": 1}
May  2 16:36:56.293516	INFO	starting test case instance	{"plan": "celestia-app", "group": "single", "number": 0, "total": 1}
May  2 16:36:56.855331	INFO	0.5622s      ERROR << single[000] >> panic: $HOME is not defined
May  2 16:36:56.855438	INFO	0.5624s      ERROR << single[000] >>
May  2 16:36:56.855470	INFO	0.5624s      ERROR << single[000] >> goroutine 1 [running]:
May  2 16:36:56.855519	INFO	0.5625s      ERROR << single[000] >> github.com/celestiaorg/celestia-app/app.init.0()
```

Resource: [link](https://unix.stackexchange.com/questions/259791/how-to-change-the-home-directory-of-the-currently-logged-in-user#:~:text=You%20need%20to%20edit%20the,home%20directory%20of%20the%20user)

As a hot-fix/dirty-hack for testground case, we should create a branch where those init funcs are commented out in order for the app to be intialized properly.

evan-forbes:
We might be able to pass a flag that overwrites whatever that initialized `DefaultNodeHome` value is.

SweeXordious:
It would be better if we have a way to initialize celestia-app without having to call:

```golang
os.UserHomeDir()
```

Currently, we can pass `homePath` to:

https://github.com/celestiaorg/celestia-app/blob/d7bb3934e0c13259bf3d666378ea67fda852b6e5/app/app.go#L213-L224 

However, it is still evaluating:

```golang
os.UserHomeDir()
```
Which ends up `panic`ing in the context of a user who doesn't have an explicitly defined home directory.

[...]

PR #1473, opened 2023-03-13
docker: Images Security

smuu:
This is part of: https://github.com/celestiaorg/test-infra/issues/160.

- Use a non-root user named celestia (UID/GID = 10001)
- Dependabot now also scans `Dockerfiles` in the `docker` folder
- Change to folder /home/celestia

## Checklist

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -172,17 +172,17 @@ var (
 var _ servertypes.Application = (*App)(nil)
 
 func init() {
-	homeDir := os.Getenv("CELESTIA_HOME")
+	userHomeDir := os.Getenv("CELESTIA_HOME")
 
-	if homeDir == "" {
-		userHomeDir, err := os.UserHomeDir()
+	if userHomeDir == "" {
+		var err error
+		userHomeDir, err = os.UserHomeDir()
 		if err != nil {
 			panic(err)
 		}
-		homeDir = userHomeDir
 	}
 
-	DefaultNodeHome = filepath.Join(homeDir, "."+Name)
+	DefaultNodeHome = filepath.Join(userHomeDir, "."+Name)
 }
 
 // App extends an ABCI application, but with most of its parameters exported.


Dockerfile
```diff
@@ -19,4 +24,9 @@ EXPOSE 26656 26657 1317 9090
 
 ENV CELESTIA_HOME /opt
 
+RUN adduser -D -u 1000 celestia \
+    && chown -R celestia:celestia /opt
```
SweeXordious, line 28: [Question]
What are the permissions given to the `/opt` folder? Does the user `celestia` has write permissions to `/opt/...`?
Because it will write to it being the default home

evan-forbes (dismissed):
thanks!!

docker/Dockerfile_ephemeral
```diff
@@ -27,10 +27,16 @@ WORKDIR $HOME
...
 EXPOSE 26656 26657 1317 9090
 
-# This allows us to always set the --home directory using an env 
+# This allows us to always set the --home directory using an env
 # var while still capturing all arguments passed at runtime
 ENTRYPOINT [ "/bin/bash", "-c", "exec ./celestia-appd \
             --home ${HOME} \
             \"${@}\"", "--" ]
+
+RUN adduser -D -u 1000 celestia \
```
SweeXordious, line 36: same

Dockerfile
smuu: Yes, a user has `rwx` permission to the folders he owns. I tested starting the container locally and had no issues starting the binary.
I also tested it with a custom directory, and in this case you need to ensure that the folder has the correct permissions (either owned by user `1000` or correct `rwx` permission) as it has not access to all folders as `root` has.

This should be documented so that people running the container don't run into issues.

SweeXordious: Cool :+1: Guess we should create an issue to document this somewhere, approving

SweeXordious (dismissed):

Dockerfile
smuu: https://github.com/celestiaorg/docs/issues/508

jrmanes: hey @smuu 
good point to bring ;) 

let me add some suggestions:

```suggestion

# User data
ENV UID=10001
ENV GID=10001
ENV USR_NAME=celestia
ENV CELESTIA_HOME /opt/${USR_NAME}

# Create the group and the user, provide GID & UID
RUN groupadd -g ${GID} ${USR_NAME} \
    && adduser -u ${UID} \
       --gid ${GID} \
       --disabled-login \
       --disabled-password \
       --gecos "" \
       --home ${CELESTIA_HOME}
       
USER ${USR_NAME}    
```

why don't use UID/GID below 10000?
*UIDs below 10,000 are a security risk on several systems, because if someone does manage to escalate privileges outside the Docker container their Docker container UID may overlap with a more privileged system user's UID granting them additional permissions. For best security, always run your processes as a UID/GID above 10,000.*

docker/Dockerfile.test
```diff
@@ -31,5 +31,10 @@ COPY ./scripts/single-node.sh .
 
 EXPOSE 26657
 
+RUN adduser -D -u 1000 celestia \
+    && chown -R celestia:celestia /opt
+
+USER celestia
```
jrmanes, line 37: I'll do the same approach as the other

Dockerfile
evan-forbes: we don't actually check 

> ENV CELESTIA_HOME /opt/${USR_NAME}

atm, but imo we should keep this and update the code to do so. I forget where, but I think we have this in some issue somewhere

jrmanes: in this case it's just for using the env var later when creating the user `home` dir

```
RUN groupadd -g ${GID} ${USR_NAME} \
	    && adduser -u ${UID} \
	       --gid ${GID} \
	       --disabled-login \
	       --disabled-password \
	       --gecos "" \
	       --home ${CELESTIA_HOME} # here
```

regarding the other issue, it was because we have to specify the `--home` when executing the CLI every time, by default it tries to use the user home dir (currently root)
```
"/root/.celestia-app"
```

we'll have to test it, tho I'd say that with this change, running the container as `celestia` user, the `--home` flag should go to the user home dir `(/opt/${USR_NAME})`

smuu: Good suggestion @jrmanes. Is there a reason, that you're using `ENV` instead of `ARG` here?

smuu: > we don't actually check
> 
> > ENV CELESTIA_HOME /opt/${USR_NAME}
> 
> atm, but imo we should keep this and update the code to do so. I forget where, but I think we have this in some issue somewhere

@evan-forbes You mean this will not work with the current code and we would need to change the code to support this? (With this I mean `/opt/{USR_NAME}`.

jrmanes: > Good suggestion @jrmanes. Is there a reason, that you're using `ENV` instead of `ARG` here?

nothing special, we can use `ARG` for all of them  👌

evan-forbes: @smuu yeah, we just have to add a line above this to look for the env var first https://github.com/celestiaorg/celestia-app/blob/da2e54b0ca48c7e85a2626c9b10a985706ae39fe/app/app.go#L178-L184

MSevey: > @evan-forbes Please check this: [b41b2e4](https://github.com/celestiaorg/celestia-app/commit/b41b2e4dc43ab9014df0e5ee3f81bca83aee5778)

left a comment/question for evan.

Dockerfile
```diff
@@ -1,22 +1,41 @@
 # stage 1 Generate celestia-appd Binary
-FROM golang:1.18-alpine as builder
+FROM docker.io/golang:1.18.10-alpine3.17 as builder
```
MSevey, line 2: question: what is the reason for updating to `docker.io`?

Dockerfile
```diff
@@ -1,22 +1,41 @@
...
+    musl-dev
 COPY . /celestia-app
 WORKDIR /celestia-app
 RUN make build
 
 # stage 2
-FROM alpine:3.17.2
+FROM docker.io/alpine:3.17.2
+
+ARG UID=10001
```
MSevey, line 16: nit: might be worth referencing @jrmanes comment about why it is over 10,000 for future reference

README.md
```diff
@@ -47,6 +47,12 @@ node            |  |                               |  |
 celestia-appd --help
 ```
 
+### Environment variables
+
+| Variable        | Explanation                        | Default value | Required |
+| --------------- | ---------------------------------- | ------------- | -------- |
+| `CELESITA_HOME` | Home directory for the application |               | Optional |
```
MSevey, line 54: We should include the default value. I think we can just reference `os.UserHomeDir()` and link to the docs https://pkg.go.dev/os#UserHomeDir

app/app.go
```diff
@@ -180,10 +180,20 @@ func init() {
...
 	}
+	// Try to get CELESTIA_HOME env variable, if not set, use user home dir
+	userHomeDir = getEnv("CELESTIA_HOME", userHomeDir)
 
 	DefaultNodeHome = filepath.Join(userHomeDir, "."+Name)
 }
 
+// Get environment variable or fallback to default value
+func getEnv(key, fallback string) string {
+	if value, ok := os.LookupEnv(key); ok {
```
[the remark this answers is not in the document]
smuu: As far as I know, an empty string is the same as an environment variable that is not set. So how can we catch this?

MSevey (changes requested):
we should close the loop on unit testing the new function and adding the link to the go docs about the default value for the user home directory

app/app.go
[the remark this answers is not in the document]
smuu: > also new method requires new unit test :-)
> 
> We should right the unit test in a way that it is set up to be extended for all future env vars.

Wo can assist with that? I don't have a lot of experience with Go.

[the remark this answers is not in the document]
MSevey: plus then you can skip any new unit tests :-)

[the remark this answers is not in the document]
smuu: Great idea! Thanks :)
I needed to adapt your solution a little because I had an error: https://github.com/celestiaorg/celestia-app/pull/1473/commits/d0273abd439fbf3787a3fdebd4630982de3d1b54

evan-forbes (dismissed):
really nice improvements, thanks for this 🙂

MSevey (approved):
👏

MSevey:
Does this resolve #377 now? Or is that a slightly different issue?

Since we are now loading the env var.

evan-forbes:
yeah, we should be able to close #377

Issue #1249, opened 2023-01-16
Integration test for IBC connections


evan-forbes:
Ideally, we would be able to test that IBC still works as intended with a few commonly connected chains, such as osmosis and the cosmos hub. Not only could we test functionality such as the token filter introduced in #1219, but we could also have more confidence that we don't accidentally break compatability. We might be able to do this with [ibctest](https://github.com/strangelove-ventures/ibctest) after each docker release.

PR #1474, opened 2023-03-13
test: add integration test for the ibc tokenfilter

cmwaters:
Closes: #1249

I first explored using [interchaintest](https://github.com/strangelove-ventures/interchaintest) which similar to testground spun up all necessary infra using docker containers and then exposed some go APIs for executing IBC messages and making assertions. I unfortunately, came across several problems owing to the fact that celestia has made several changes to the cosmos-sdk that interchain test bases assumptions on.

I switched tack, and instead looked at using ibc's own testing framework. There were a few similar complications before (i.e. it's difficult to inject a custom app) but I managed to overcome them and have put together a test suite under `testing/tokenfilter` that does the following:
- ensures that native celestia tokens can be sent out from celestia and can return
- ensures that other chains can't send tokens to celestia chains and that their balance is unaffected.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,3 +16,4 @@
 docs @liamsi @adlerjohn @MSevey @evan-forbes
 x/qgb @SweeXordious @evan-forbes
 pkg/shares @rootulp @evan-forbes
+pkg/wrapper @staheri14 @evan-forbes
--- a/.github/workflows/issue-label-automation.yml
+++ b/.github/workflows/issue-label-automation.yml
@@ -12,6 +12,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Check if issue or PR was created by external contributor
+        if: ${{ github.actor != 'dependabot[bot]' }}
         uses: tspascoal/get-user-teams-membership@v2
         id: teamCheck
         with:
@@ -31,8 +32,16 @@ jobs:
       # If an issue or PR was created by someone that isn't part of
       # celestia-core, add the `external` label.
       - name: Maybe label issue or PR with `external`
-        if: ${{ steps.teamCheck.outputs.isTeamMember == 'false' }}
+        if: ${{ github.actor != 'dependabot[bot]' && steps.teamCheck.outputs.isTeamMember == 'false' }}
         uses: andymckay/labeler@master
         with:
           add-labels: "external"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # If a PR was created by dependabot add the `bot` label.
+      - name: Maybe label PR with `bot`
+        if: ${{ github.actor == 'dependabot[bot]' }}
+        uses: andymckay/labeler@master
+        with:
+          add-labels: "bot"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,17 +6,13 @@ name: Tests / Code Coverage
 on:
   workflow_call:
 
-jobs:
-  cleanup-runs:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: rokroskar/workflow-run-cleanup-action@master
-        env:
-          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-    if: |
-      !startsWith(github.ref, 'refs/tags/') &&
-      github.ref != 'refs/heads/master'
+concurrency:
+  # do not cancel jobs from earlier commits for tags or the main branch
+  group: ${{ github.workflow }}-${{ !startsWith(github.ref, 'refs/tags/') &&
+      github.ref != 'refs/heads/main' }}
+  cancel-in-progress: true
 
+jobs:
   install-tparse:
     runs-on: ubuntu-latest
     steps:
@@ -29,7 +25,7 @@ jobs:
         run: >
           export GO111MODULE="on" &&
           go install github.com/mfridman/tparse@v0.8.3
-      - uses: actions/cache@v3.3.0
+      - uses: actions/cache@v3.3.1
         with:
           path: ~/go/bin
           key: ${{ runner.os }}-go-tparse-binary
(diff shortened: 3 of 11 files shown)


testing/tokenfilter/setup.go
```diff
@@ -0,0 +1,230 @@
...
+	// create current header and call begin block
+	header := tmproto.Header{
+		ChainID: chainID,
+		Height:  1,
+		Time:    coord.CurrentTime.UTC(),
+	}
+
+	txConfig := app.GetTxConfig()
+
+	// create an account to send transactions from
```
rootulp, line 88: [question] I'm not sure how this comment is applicable to the line below. Should the comment be moved?

testing/tokenfilter/setup.go
```diff
@@ -0,0 +1,230 @@
...
+		SenderAccounts: senderAccs,
+	}
+
+	coord.CommitBlock(chain)
+
+	return chain
+}
+
+// NewTestChain initializes a new test chain with a default of 4 validators
+// Use this function if the tests do not need custom control over the validator set
```
rootulp, line 112:
```suggestion
// NewTestChain initializes a new test chain with a default of 4 validators.
// Use this function if the tests do not need custom control over the validator set.
```

testing/tokenfilter/setup.go
```diff
@@ -0,0 +1,230 @@
...
+func NewTestChain(t *testing.T, coord *ibctesting.Coordinator, chainID string) *ibctesting.TestChain {
+	// generate validators private/public key
```
rootulp, line 114: [nit] can this comment please be moved to the location immediately above where it is relevant?

```go
	        // generate validators private/public key
		privVal := mock.NewPV()
		pubKey, err := privVal.GetPubKey()
```

testing/tokenfilter/tokenfilter_test.go
```diff
@@ -0,0 +1,132 @@
...
+	// relay send
+	err = path.RelayPacket(packet)
+	suite.Require().NoError(err) // relay committed
+
+	// check that the token exists on chain B
+	voucherDenomTrace := types.ParseDenomTrace(types.GetPrefixedDenom(packet.GetDestPort(), packet.GetDestChannel(), sdk.DefaultBondDenom))
+	balance := suite.otherChain.GetSimApp().BankKeeper.GetBalance(suite.otherChain.GetContext(), suite.otherChain.SenderAccount.GetAddress(), voucherDenomTrace.IBCDenom())
+	coinSentFromAToB := types.GetTransferCoin(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, sdk.DefaultBondDenom, amount)
+	suite.Require().Equal(coinSentFromAToB, balance)
+
```
rootulp, line 82: [optional]

```go
	// check that the account on celestiaChain has less tokens than the original starting balance
	celestiaBalance := celestiaApp.BankKeeper.GetBalance(suite.celestiaChain.GetContext(), suite.celestiaChain.SenderAccount.GetAddress(), sdk.DefaultBondDenom)
	suite.Require().True(celestiaBalance.Amount.LT(originalBalance.Amount))
```

[question] Does this test intentionally use the `stake` denomination instead of `utia`?

```
celestiaBalance: 5000000000000000000stake
originalBalance: 10000000000000000000stake
```

 If not, proposal to use `utia`. On second thought, maybe it intentionally doesn't use `utia` to avoid depending on the `app` package.

rootulp (approved):
Overall LGTM. No blocking feedback

testing/tokenfilter/setup.go
```diff
@@ -0,0 +1,230 @@
...
+	// Note that the validators are sorted by voting power
+	// or, if equal, by address lexical order
+	valSet := tmtypes.NewValidatorSet(validators)
+
+	return NewTestChainWithValSet(t, coord, chainID, valSet, signersByAddress)
+}
+
+// SetupWithGenesisValSet initializes a new SimApp with a validator set and genesis accounts
+// that also act as delegators. For simplicity, each validator is bonded with a delegation
+// of one consensus engine unit (10^6) in the default token of the simapp from first genesis
```
evan-forbes, line 139: [non blocking question]

> For simplicity, each validator is bonded with a delegation
// of one consensus engine unit (10^6) 

does this depend on the power reduction passed?

evan-forbes (approved):
we have a lot of things in testuil, but I think the testing package makes more sense name wise for some of the things in testutil, so we might want to move some things over to `testing` after merging this.

ibc's testing suite is really cool!! mentioned this PR in #829, as there might be something that we can consolidate in the future with the `testnode` package, but I would need to take a much deeper look.

testing/tokenfilter/setup.go
cmwaters: Yeah I can cut it out. This entire setup.go page is copy pasted from IBC's testing framework just FYI. Predominantly because our `staking.Validator` requires an `EvmAddress`.

testing/tokenfilter/tokenfilter_test.go
cmwaters: > suite.Require().True(celestiaBalance.Amount.LT(originalBalance.Amount))

I would prefer to be exact.

> [question] Does this test intentionally use the stake denomination instead of utia?

No intentions around the denomination. I just used whatever the defaults are

rootulp: > I would prefer to be exact.

An exact assertion also works so something like

```go
	// check that the account on celestiaChain has fewer tokens than it started with
	celestiaBalance := celestiaApp.BankKeeper.GetBalance(suite.celestiaChain.GetContext(), suite.celestiaChain.SenderAccount.GetAddress(), sdk.DefaultBondDenom)
	want := originalBalance.Amount.Sub(coinToSendToB.Amount)
	suite.Require().Equal(want, celestiaBalance.Amount)
```

cmwaters: Done: [add extra assertion](https://github.com/celestiaorg/celestia-app/pull/1474/commits/d37325c46b1eb19148de19c4764ff0d7a48a37be)

PR #1476, opened 2023-03-13
feat(pkg/da): Extract min share creation into separate function for convenience

renaynay:
This PR extracts min share creation into an exported function so that celestia-node can depend on it canonically without duplicating code and risking error.


evan-forbes (approved):

cmwaters (approved):

rootulp (approved):

Wondertan (commented):
Let's also make MinEDS then

PR #1481, opened 2023-03-13
ci: automate-labels for dependabot PRs

amityadav0:
Automate-labels for dependabot PRs

Closes #1464

PR #1482, opened 2023-03-13
chore: add Sanaz as NMT wrapper code owner

rootulp:
Inspired by https://github.com/celestiaorg/celestia-app/pull/1481#pullrequestreview-1337892356 😆


evan-forbes (approved):
hell yeah! I'm still of the opinion that we should try to add more folks. 3 active contributors + leadership should be the minimum imo

rootulp:
It would be nice if each directory had at least 2 owners to reduce bus factor. Will merge this as-is and ppl can add themselves as owners to directories as they work on them.

staheri14:
Thanks @rootulp for this! 👍

PR #1484, opened 2023-03-13
feat!: specify `utia` as denom for gov module genesis

rootulp:
Fixes an issue identified in https://github.com/celestiaorg/itn-doc/issues/168

## Testing

Verified genesis.json contains `utia` in the gov module in [this diff](https://www.diffchecker.com/Q14DBeHB/)

### DIFF
--- a/app/module_test.go
+++ b/app/module_test.go
@@ -1,0 +1,34 @@
+package app
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/celestiaorg/celestia-app/app/encoding"
+	"github.com/cosmos/cosmos-sdk/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+// Test_newGovModule tests that the default genesis state for the gov module
+// uses the utia denominiation.
+func Test_newGovModule(t *testing.T) {
+	encCfg := encoding.MakeConfig(ModuleEncodingRegisters...)
+
+	govModule := newGovModule()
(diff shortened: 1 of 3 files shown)


app/module_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	govModule := newGovModule()
+	raw := govModule.DefaultGenesis(encCfg.Codec)
+	govGenesisState := govtypes.GenesisState{}
+
+	// HACKHACK explicitly ignore the error returned from json.Unmarshal because
+	// the error is a failure to unmarshal the string StartingProposalId as a
+	// uint which is unrelated to the test here.
+	//nolint:errcheck
+	json.Unmarshal(raw, &govGenesisState)
```
evan-forbes, line 26: this comment is good but also why not explicitly ignore the error and get rid of the no lint?

```suggestion
	// HACKHACK explicitly ignore the error returned from json.Unmarshal because
	// the error is a failure to unmarshal the string StartingProposalId as a
	// uint which is unrelated to the test here.
	_ = json.Unmarshal(raw, &govGenesisState)
```

evan-forbes (approved):

app/module_test.go
rootulp: Not sure how.

Screenshot 2023-03-13 at 6 44 03 PM


Screenshot 2023-03-13 at 6 43 48 PM

evan-forbes: we just need to get rid of the `:` in the first image since we're not declaring anything

rootulp: oof how did I miss that. Can do in a follow-up PR.

Issue #1014, opened 2022-11-16
EPIC: Social Upgrades


evan-forbes:
We've discussed the desire for validators and consensus fullnodes to be able to update binaries before the actual upgrade height. This differs from the standard cosmos-sdk upgrade module, which typically relies on switching binaries at the upgrade height. 

#### Tracking
- [x] #1491 
- [x] https://github.com/celestiaorg/celestia-app/issues/1571
- [ ] ADR for social upgrades #1562 
- [ ] https://github.com/celestiaorg/celestia-app/issues/2464

evan-forbes:
related to #1009 in that we also want to remove most text based governance proposals, which includes upgrade proposals

evan-forbes:
After implementing auto switching upgrades, there's an interesting addition proposed by @cmwaters where we can add a signalling component. Each node would gossip when each node is ready to upgrade. Nodes can then vote to upgrade after they have received 2/3s. This would also need to be communicated with light clients, which would also need to upgrade at that time.

MicahZoltu:
It seems like it would be trivial to Sybil attack such a signaling mechanism, and in fact I struggle to imagine a future where such a mechanism *isn't* Sybil attacked eventually.

evan-forbes:
> It seems like it would be trivial to Sybil attack such a signaling mechanism

apologies, I wasn't clear enough, by 2/3's I mean 2/3's of the voting power of validators, not just full nodes. Would this still be subject to a sybil attack? It seems like it would be easy to check if a message was sent from from a validator and remove a peer if they're not, but I definitely could be missing something.

MicahZoltu:
Ah, if it is 2/3 of validators (not nodes) then it isn't susceptible to Sybil.

This is still susceptible to governance attacks though.  In a well designed PoS network validators are tightly constrained on the damage they can do, usually limited to censorship and double spending within finality window.  It is important to be careful not to give validators additional powers like the ability to push out a rule change onto users (which effectively is ultimately unconstrained power).

Thinking more on this, I am somewhat against the signalling mechanism you have proposed.  Imagine if an honest minority wants to execute a fork away from a dishonest majority.  By having them wait for the signal from the validators they would be functionally blocked from forking away by the dishonest majority.

If a user has decided that they want to apply a rule chain, they should be allowed to do so freely regardless of what other network participants think/want.

cmwaters:
>  Imagine if an honest minority wants to execute a fork away from a dishonest majority. 

[...]

PR #1491, opened 2023-03-16
ADR 016: Removing Token Voting for Upgrades

evan-forbes:
## Overview

still a WIP and not yet ready for review

## Checklist

### DIFF
--- a/docs/architecture/adr-016-social-upgrades.md
+++ b/docs/architecture/adr-016-social-upgrades.md
@@ -1,0 +1,128 @@
+# ADR 016: Forcing Social Upgrades
+
+## Status
+
+Proposed
+
+## Changelog
+
+- 2023/03/15: initial draft
+
+## Context
+
+Standard cosmos-sdk based chains are soverign and have the ability to hardfork, but the current implementation makes heavy use of onchain token voting. After a proposal is submitted and crosses some threshold of aggreement, the halt height is set in the state machine for all nodes, and validators are expected to switch binaries after the state machine halts the node. This mechanism does work, but it bypasses social consensus, one of the core principles of Celestia. The ultimate security, ruleset, and result of that ruleset rests soley on the shoulders of node operators, not validators or token holders. Therefore we are seeking a solution that engraves social consensus into the implementation itself.
+
+The most pertinent issue at the moment is that we are launching mainnet soon, and don't have a fully functional upgrade mechanism that empowers social consensus in place. We need to first decide on a short term solution that will work for mainnet, and later decide on and implement a long term solution. The latter discussion on how to actually implement upgrades that fit all of our desired properties is out of scope for this ADR, and will be discussed in a separate ADR.
+
+It should also be noted that this ADR is not discussion parameter changes, as limiting token voting on those is also a separate discussion.
+
+## Alternative Approaches
+
+### Option 1: Temporarily use the current token voting mechansism
+
+Given that we don't have a lot of time until mainnet, we could leave the current implementation in place which gives us the option to use it if needed. Ideally, we would work on the longer term upgrade mechanism that respects social consensus and finish it before the first upgrade. While this option does allow for maximum flexibility, it is also very risky because if the mechanism is ever needed or used, then it could set a precedent for future upgrades.
+
+### Option 2: Predetermined halt height and removal of token voting for upgrades (aka "difficulty bomb")
+
+The goal of this approach is to set a deadline to reach social consensus. It involves removing the upgrade mechanism from the state machine, and replaces it with a predefined shut down height. If this height is reached without first upgrading, then the chain will halt. Without an onchain voting mechanism, the only way to upgrade the chain is to reach social consensus. If we don't collectively do that, then the chain will halt.
+
+## Decision
+
+> This section records the decision that was made.
+> It is best to record as much info as possible from the discussion that happened. This aids in not having to go back to the Pull Request to get the needed information.
+
+## Detailed Design
+
+### Implementing Option 1
+
+No changes are needed.
+
+### Implementing Option 2
+
+- Remove the upgrade module from the state machine
+
+This involves a fairly standard removal of a module. Which is summarized (but not all encompassing!) in the following diff:
+
+```diff
+
+// ModuleBasics defines the module BasicManager is in charge of setting up basic,
+// non-dependant module elements, such as codec registration
+// and genesis verification.
+ModuleBasics = module.NewBasicManager(
+...
+ibc.AppModuleBasic{},
+-upgrade.AppModuleBasic{},
+evidence.AppModuleBasic{},
+...
+)
+
+// New returns a reference to an initialized celestia app.
+func New(
+	logger log.Logger,
+	db dbm.DB,
+	traceStore io.Writer,
+	loadLatest bool,
+	skipUpgradeHeights map[int64]bool,
+	homePath string,
+	invCheckPeriod uint,
+	encodingConfig encoding.Config,
+	appOpts servertypes.AppOptions,
+	baseAppOptions ...func(*baseapp.BaseApp),
+) *App {
+    ...
+-	app.UpgradeKeeper = upgradekeeper.NewKeeper(skipUpgradeHeights, keys[upgradetypes.StoreKey], appCodec, homePath, app.BaseApp, authtypes.NewModuleAddress(govtypes.ModuleName).String())
+    ...
+
+	// register the proposal types
+	govRouter := oldgovtypes.NewRouter()
+	govRouter.AddRoute(govtypes.RouterKey, oldgovtypes.ProposalHandler).
+		AddRoute(paramproposal.RouterKey, params.NewParamChangeProposalHandler(app.ParamsKeeper)).
+		AddRoute(distrtypes.RouterKey, distr.NewCommunityPoolSpendProposalHandler(app.DistrKeeper)).
+-		AddRoute(upgradetypes.RouterKey, upgrade.NewSoftwareUpgradeProposalHandler(app.UpgradeKeeper)).
+		AddRoute(ibcclienttypes.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper))
+```
+
+Removing the [upgrade module](https://github.com/celestiaorg/cosmos-sdk/tree/v1.8.0-sdk-v0.46.7/x/upgrade) will also remove the `sdk.Msg` that is used by the (non-legacy) governance module, along with any capacity to halt. After this, both the legacy and the current governance modules will no longer have the ability to call the below function:
+
+```go
+// SoftwareUpgrade implements the Msg/SoftwareUpgrade Msg service.
+func (k msgServer) SoftwareUpgrade(goCtx context.Context, req *types.MsgSoftwareUpgrade) (*types.MsgSoftwareUpgradeResponse, error) {
+	if k.authority != req.Authority {
+		return nil, errors.Wrapf(gov.ErrInvalidSigner, "expected %s got %s", k.authority, req.Authority)
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	err := k.ScheduleUpgrade(ctx, req.Plan)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,170 @@
...
+```go
+func BeginBlocker(k keeper.Keeper, ctx sdk.Context, _ abci.RequestBeginBlock) {
+   if ctx.BlockHeader().Height >= BombHeight {
+       panic("social consensus must be reached to pass the bomb height")
+   }
+   return
+}
+```
+
+We could also impose other similar restrictions elsewhere instead, such as voting to reject the block if the deadline height is `BombHeight` is reached.
```
rootulp, line 117:
```suggestion
We could also impose other similar restrictions elsewhere instead, such as voting to reject the block if the deadline height in `BombHeight` is reached.
```

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,170 @@
...
+
+It should also be noted that this ADR is not discussion parameter changes, as limiting token voting on those is also a separate discussion.
+
+## Alternative Approaches
+
+### Option 1: Temporarily use the current token voting mechanism
+
+Given that we don't have a lot of time until mainnet, we could leave the current implementation in place which gives us the option to use it if needed. Ideally, we would work on the longer term upgrade mechanism that respects social consensus and finish it before the first upgrade. While this option does allow for maximum flexibility, it is also very risky because if the mechanism is ever needed or used, then it could set a precedent for future upgrades.
+
+### Option 2: Predetermined halt height and removal of token voting for upgrades (aka "difficulty bomb")
```
rootulp, line 25: [question] is there an Option 3 that removes token voting but doesn't introduce the difficulty bomb?

Reasons why a difficulty bomb may be unnecessary:
1. There isn't a need to introduce an upgrade mechanism prior to the next (currently unplanned) upgrade. Given the next upgrade isn't planned, we don't have a block height for it. Since no block height exists, we can't implement a difficulty bomb that coincides with when such an upgrade were to take place. Implementing a difficulty bomb with an estimated block height may result in several subsequent patches to delay the bomb.
2. The protocol may not need the difficulty bomb as an incentive to implement an upgrade mechanism. If the next upgrade brings improvements to the protocol, protocol participants may be sufficiently incentivized to introduce an upgrade mechanism to benefit from the proposed changes.

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,170 @@
...
+# a node will gracefully halt and shutdown that can be used to assist upgrades
+# and testing.
+#
+# Note: Commitment of state will be attempted on the corresponding block.
+halt-time = 0
+```
+
+## Consequences
+
+If we adopt Option 2, then we will be able to remove token voting from the statemachine sooner rather than later. This is riskier in that we will not have the battle tested mechanism, but it will force future upgrades that respect social consensus.
```
rootulp, line 168:
```suggestion
If we adopt Option 2, then we will be able to remove token voting from the state machine sooner rather than later. This is riskier in that we will not have the battle tested mechanism, but it will force future upgrades that respect social consensus.
```

rootulp (approved):
Overall LGTM

docs/architecture/adr-016-social-upgrades.md
evan-forbes: we could yeah, but here I didn't want to specify all of the different combinations, as this was just a quick ADR to cover something that was discussed synchronously. 

> There isn't a need to introduce an upgrade mechanism prior to the next (currently unplanned) upgrade. Given the next upgrade isn't planned, we don't have a block height for it. Since no block height exists, we can't implement a difficulty bomb that coincides with when such an upgrade were to take place. Implementing a difficulty bomb with an estimated block height may result in several subsequent patches to delay the bomb.

Adding the difficulty bomb is inherently planning for an upgrade within some period of time. It's explicitly telling the validators that they have to at least make one change before that determined height.

> The protocol may not need the difficulty bomb as an incentive to implement an upgrade mechanism. If the next upgrade brings improvements to the protocol, protocol participants may be sufficiently incentivized to introduce an upgrade mechanism to benefit from the proposed changes.

The attack scenario is that the validators choose not to upgrade but social consensus does. This is just an explicit form of that. Participants could want to upgrade, but that does not mean validators will want to. I can add a line making this more explicit though!

evan-forbes: I'll add the third option though

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -19,12 +19,13 @@ It should also be noted that this ADR is not discussion parameter changes, as li
 ## Alternative Approaches
 
 ### Option 1: Temporarily use the current token voting mechanism
-
 Given that we don't have a lot of time until mainnet, we could leave the current implementation in place which gives us the option to use it if needed. Ideally, we would work on the longer term upgrade mechanism that respects social consensus and finish it before the first upgrade. While this option does allow for maximum flexibility, it is also very risky because if the mechanism is ever needed or used, then it could set a precedent for future upgrades.
 
-### Option 2: Predetermined halt height and removal of token voting for upgrades (aka "difficulty bomb")
+### Option 2: Removal of token voting for upgrades
+The goal of this approach is to force social consensus to reach an upgrade instead of relying on token voting. It works by simply removing the currently upgrade module. This way, the only way to upgrade the chain is to agree on the upgrade logic and the upgrade height.
```
rootulp, line 25:
```suggestion
The goal of this approach is to force social consensus to reach an upgrade instead of relying on token voting. It works by simply removing the current upgrade module. This way, the only way to upgrade the chain is to agree on the upgrade logic and the upgrade height.
```

rootulp (approved):

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Option 2: Removal of token voting for upgrades
+The goal of this approach is to force social consensus to reach an upgrade instead of relying on token voting. It works by simply removing the current upgrade module. This way, the only way to upgrade the chain is to agree on the upgrade logic and the upgrade height.
+
+### Option 3: Adding a predetermined halt height (aka "difficulty bomb")
+This option is not mutually exclusive to option 2. Its goal is to explicitly state that, without changing binaries, light clients will halt at a given height, despite what logic validators are running. This acts as an explicit statement to large token holders that they either come to some sort of agreement with the rest of the network, or chain will halt. Not coming to agreement is not a viable option.
+
+## Decision
+
+TODO
```
liamsi, line 32: I think it's safe to say we settled on Option 2

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,147 @@
...
+
+The most pertinent issue at the moment is that we are launching mainnet soon, and don't have a fully functional upgrade mechanism that empowers social consensus in place. We need to first decide how to remove token voting in a way that supports our future efforts to change the upgrade mechanism. The latter discussion on how to actually implement upgrades that fit all of our desired properties is out of scope for this ADR, and will be discussed separately.
+
+It should also be noted that this ADR is not discussion parameter changes, as limiting token voting on those is also a separate discussion.
```
cmwaters, line 17:
```suggestion
It should also be noted that this ADR is not discussing parameter changes, as limiting token voting on those is also a separate discussion.
```

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,147 @@
...
+#
+# Note: Commitment of state will be attempted on the corresponding block.
+halt-time = 0
+```
+
+## Consequences
+
+If we adopt Option 2, then we will be able to remove token voting from the state machine sooner rather than later. This is riskier in that we will not have the battle tested mechanism, but it will force future upgrades that respect social consensus.
+
+## References
```
cmwaters, line 147: Would be nice to reference the ADR018 as to a plan to upgrading

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,147 @@
...
+
+It should also be noted that this ADR is not discussion parameter changes, as limiting token voting on those is also a separate discussion.
+
+## Alternative Approaches
+
+### Option 1: Temporarily use the current token voting mechanism
+Given that we don't have a lot of time until mainnet, we could leave the current implementation in place which gives us the option to use it if needed. Ideally, we would work on the longer term upgrade mechanism that respects social consensus and finish it before the first upgrade. While this option does allow for maximum flexibility, it is also very risky because if the mechanism is ever needed or used, then it could set a precedent for future upgrades.
+
+### Option 2: Removal of token voting for upgrades
+The goal of this approach is to force social consensus to reach an upgrade instead of relying on token voting. It works by simply removing the current upgrade module. This way, the only way to upgrade the chain is to agree on the upgrade logic and the upgrade height.
```
cmwaters, line 25: Might be helpful to expand on what the upgrade mechanism would be. Given that we've decided to use a single binary sync which basically enables rolling upgrades, operators can update their binaries ahead of time and a height is embedded in the binary which indicates when to switch to the next state machine.

docs/architecture/adr-016-social-upgrades.md
```diff
@@ -0,0 +1,147 @@
...
+
+Accepted
+
+## Changelog
+
+- 2023/03/15: initial draft
+
+## Context
+
+Standard cosmos-sdk based chains are sovereign and have the ability to hardfork, but the current implementation makes heavy use of onchain token voting. This bypasses social consensus, one of the core principles of Celestia. After a proposal is submitted and crosses some threshold of agreement, the halt height is set in the state machine for all nodes, and validators are expected to switch binaries after the state machine halts the node. The ultimate security, ruleset, and result of that ruleset rests solely on the shoulders of node operators, not validators or token holders. Therefore we are seeking a solution that engraves social consensus into the implementation itself.
```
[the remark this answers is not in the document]
evan-forbes: You are correct that people can always use which ever binary they choose, which is what the first part of the sentence was attempting to address. 
> Standard cosmos-sdk based chains are sovereign and have the ability to hardfork,

but the upgrade module does put a loose blocker to this by and the point of the onchain mechanism is to use token voting to decide. The social contract is to do whatever token voting determines, which is not the social contract that celestia wants to adhere to.

MSevey (dismissed):
Are there pending open questions/decisions that need to happen offline to move forward with this?  Leaving my approval so we can merge if not.

evan-forbes:
I have some offline updates that I still want to push, just been busy or ooo, apologies for any delay

docs/architecture/adr-016-social-upgrades.md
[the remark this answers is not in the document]
evan-forbes: updated, and I tried to focus on difference in social contract more

evan-forbes:
before merging, we should also understand how ibc uses the upgrade module. we might only want to remove the ability for the gov module to use the upgrade module, and then incorporate the existing upgrade management into social upgrades #1014

cmwaters:
So the idea now is that the upgrade module still remains but we remove the proposal handler and the messages (do we even need to remove the messages if only governance can call this?). 

I still think it's important that we clarify how, after these changes, a chain would expect to upgrade else this ADR feels incomplete.

evan-forbes:
> I still think it's important that we clarify how, after these changes, a chain would expect to upgrade else this ADR feels incomplete.

That's fair, we can postpone merging this one until ADR018 is finished, as I think that topic requires an entire separate ADR. Or perhaps we could just move the relevant portions of this ADR into that one and only merge that ADR.

cmwaters:
> That's fair, we can postpone merging this one until ADR018 is finished, as I think that topic requires an entire separate ADR. Or perhaps we could just move the relevant portions of this ADR into that one and only merge that ADR.

My personal view is that these shouldn't be considered two separate decisions but one

evan-forbes:
that makes a lot of sense!

iirc per sync discussion this ADR was initially separate because we just wanted to explore what a difficulty bomb and removing the upgrade module would look like before we discussed it during the onsite, but wanted to wait on upgrades. 

I'll close this ADR for now, and we can update ADR018 to ADR016

PR #1502, opened 2023-03-17
fix: signal handling for container

smuu:
Kubernetes and other container schedulers and runtimes send Unix signals to PID 1 in the container.
Currently, PID 1 is the call of `entrypoint.sh`, which starts the celestia-app with another PID. Therefore celestia-app never receives Unix signals (eg. SIGTERM) that where sent.
This PR changes the `entrypoint.sh` in such a way that celestia-app replaces the process itself. This way celestia-app will be PID 1 and then receives signals inside the container.

Interesting read:
- https://hynek.me/articles/docker-signals/

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -19,4 +19,4 @@ echo "Starting celestia-appd with command:"
 echo "/bin/celestia-appd $@"
 echo ""
 
-/bin/celestia-appd $@
+exec /bin/celestia-appd $@


SweeXordious:
In addition to this, PID 1 needs to be able to adopt orphaned processes and reap zombie processes. Is celestia-app binary able to do so? Or this will he handled by the shell we're using to run the binary?

SweeXordious:
In addition to this, PID 1 needs to be able to adopt orphaned processes and reap zombie processes. Is celestia-app binary able to do so? Or this will he handled by the shell we're using to run the binary?

evan-forbes (approved):

smuu:
> In addition to this, PID 1 needs to be able to adopt orphaned processes and reap zombie processes. Is celestia-app binary able to do so? Or this will he handled by the shell we're using to run the binary?

After `exec` is executed, there is no shell process anymore. Only the process of celestia-app with PID 1 exists. That's what you usually want in a container: your application is PID.
I think this is unnecessary without me knowing any details about handling orphaned and zombie processes. But please provide me with more information.
When PID 1 in a container exits (either with exit code zero or non-zero), the container stops successfully or fails. Then it is the task of the container orchestrator like Kubernetes to handle that.

Open questions on my side:
- Does celestia-app have child processes, or does it spawn new processes?
- Is it possible that celestia-app is in a dead-lock?

smuu:
> I'm also not sure if celestia-app spawns any child processes. More information on this https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem . This could be solved, for example, in docker-compose using the `init` parameter. However, in the case of k8s, don't know if it's the same.
> 
> Anyway, this is not blocking this PR, just a thought :D
> 
> approving

Interesting read, thank you!
I checked the running processes in a container, and there was only one process. So unlesss not an issue for us unless we start using child processes.
We should bring up that discussion again once we introduce child processes and how we want to handle that topic.

PR #1512, opened 2023-03-21
chore: got rid of most panics in share pkg

### LINKED ISSUE #1490: chore: error propagation refactore for the share package
### Summary

The share package is used by node and can be used by others in future. Currently there are a number of panics in the code, we need to manage the error propagation properly in order to make it more usable as a lib.

### Problem Definition

Currently there are a number of panics in the code and that creates the application halt for those who use it as a lib.

### Proposal

We need to make the errors bubble up to the higher levels of the calls and handle them properly. In general avoid panics.

### For Admin Use

- [ ] Not duplicate issue

mojtaba-esk:
Context here: #1490 

closes #1490 

## Checklist

### DIFF
--- a/pkg/da/data_availability_header.go
+++ b/pkg/da/data_availability_header.go
@@ -185,6 +185,8 @@ func MinDataAvailabilityHeader() DataAvailabilityHeader {
 }
 
 // MinShares returns one tail-padded share.
+// An error returned when there is an issue with building the share
+// for example if the shareVersion is wrong, or WriteSequenceLen fails, or the share size is incorrect
 func MinShares() ([][]byte, error) {
 	tpShares, err := shares.TailPaddingShares(appconsts.MinShareCount)
 	if err != nil {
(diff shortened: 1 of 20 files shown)


rootulp (approved):

pkg/da/data_availability_header.go
```diff
@@ -182,6 +185,10 @@ func MinDataAvailabilityHeader() DataAvailabilityHeader {
 }
 
 // MinShares returns one tail-padded share.
-func MinShares() [][]byte {
-	return shares.ToBytes(shares.TailPaddingShares(appconsts.MinShareCount))
+func MinShares() ([][]byte, error) {
```
evan-forbes, line 188: [nice to have, but optional]
we could add some documentation here to when an error is thrown to avoid having to dig through the tailpadding shares function.

evan-forbes (approved):
nice!

pkg/da/data_availability_header.go
mojtaba-esk: Sure will add it

mojtaba-esk: done

rootulp: [no change needed] on re-review it seems strange to return an error from this function because there are no function parameters. I expect `MinShares()` to never return an error. It could `panic()` if `shares.TailPaddingShares(appconsts.MinShareCount)` returns an error but I don't think such a change is blocking.

rootulp (approved):

PR #1521, opened 2023-03-22
docs: ADR-017 single binary sync

mojtaba-esk:
## Checklist

### DIFF
--- a/docs/architecture/adr-017-single-binary-upgrade.md
+++ b/docs/architecture/adr-017-single-binary-upgrade.md
@@ -1,0 +1,86 @@
+# ADR 017: Single binary upgrade
+
+## Status
+
+Proposed
+
+## Changelog
+
+- 2023/03/22: initial draft
+
+## Context
+
+It important to sync from genesis with one binary.
+It is needed for wallets, block explorers, indexers, analytics (e.g. Nansen).
+It makes it possible to sync headers from chain and download blobs from third party service.
+
+## Alternative Approaches
+
+Instead of having a single binary, we can have multiple binaries that are managed by another supervisory binary e.g. [Cosmovisor](https://docs.desmos.network/fullnode/cosmovisor/).
+Osmosis does it through having a significant number of binaries gathered in one docker container and switch them in particular heights.
+This approach has a lot of maintenance burden and is fragile to sync from genesis.
+
+## Decision
+
+TBD
+
+## Detailed Design
+
+- Cosmos blocks contain a version.app (version of state machine) that can be used to indicate which version of `celestia-app` should be used to process a block.
+  - Example: Osmosis is currently on [version.app: 15](https://www.mintscan.io/osmosis/blocks/8804111) and release [v15.0.0](https://github.com/osmosis-labs/osmosis/releases/tag/v15.0.0).
+
+  - Have to maintain old versions of binaries. Need to update dependencies for security issues.
+
+- Genesis has an original version.app. Does celestia-app need to explicitly define this?
+
+- Bitcoin / Ethereum leverage if statement upgrade logic. Downside to if statement logic in Cosmos ecosystem is if statements may not be relevant for new chains
+
+- UX issue to improve experience for node operators to upgrade
+
+- Maintenance issue for maintaining historical releases of binaries
+
+- What granularity should conditional logic be?
+  - App struct level
+  - Module level
+  - if/else on a line by line basis
+
+- Why app version and not block height?
+  - App version is bumped via upgrade module
+  - Depending on app version is universal across multiple networks.
+  - Block height will differ across testnets / chain-ids so depends on a `.toml` config file
+
+- Are we planning on upstream this?
+  - [CometBFT](https://github.com/cometbft/cometbft) is working on this for consensus.
+  - We’re working on Cosmos SDK portion
+
+- `celestia-node` light nodes need to version erasure coding block or celestia-node should import this from celestia-app
+
+- Will celestia-node need to pass a block height to celestia-app? Header contains version.app (state machine version) and version.block (consensus version)
+
+- Is it important to support rolling back versions? Involves upgrading (again) through the faulty version
+
+- Since if conditional logic uses one Go version, all code will be compiled with this one Go version. If behavior changes across Go versions, we have to debug why.
+
+- Option: depend on multiple major versions of a dependency. What is the relationship between if/else conditional statements and Cosmos SDK upstream changes?
+
+- How to do this in CometBFT?
+  - Always use latest p2p version
+  - Version.block in header should be used to version changes to Tendermint / CometBFT
+
+- Block sync needs to be aware of changes across block heights
+
+## Consequences
+
+> This section describes the consequences, after applying the decision. All consequences should be summarized here, not just the "positive" ones.
+
+### Positive
+
+### Negative
+
+### Neutral
+
+## References
+
+[The meeting notes](https://docs.google.com/document/d/1UuiM9sKQ4g30OBoZLI5pwYBwoicvgC65xm9yzqmbh0g/edit#)
+
+- {reference link}
--- a/docs/architecture/adr-017-single-binary-upgrade.md
+++ b/docs/architecture/adr-017-single-binary-upgrade.md
@@ -1,4 +1,4 @@
-# ADR 017: Single binary upgrade
+# ADR 017: Single binary sync
 
 ## Status
 
--- a/docs/architecture/adr-017-single-binary-upgrade.md
+++ b/docs/architecture/adr-017-single-binary-upgrade.md
@@ -17,7 +17,7 @@ It makes it possible to sync headers from chain and download blobs from third pa
 ## Alternative Approaches
 
 Instead of having a single binary, we can have multiple binaries that are managed by another supervisory binary e.g. [Cosmovisor](https://docs.desmos.network/fullnode/cosmovisor/).
-Osmosis does it through having a significant number of binaries gathered in one docker container and switch them in particular heights.
+Osmosis does it through having a significant number of binaries gathered in one docker container and switching them at particular heights.
 This approach has a lot of maintenance burden and is fragile to sync from genesis.
 
 ## Decision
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Changelog
+
+- 2023/03/22: initial draft
+
+## Context
+
+It important to sync from genesis with one binary.
+It is needed for wallets, block explorers, indexers, analytics (e.g. Nansen).
+It makes it possible to sync headers from chain and download blobs from third party service.
```
rootulp, line 15: I think we need to clarify this. In the meeting the point was made that the Celestia network doesn't need to retain all historical blob data and that blobs could eventually be pruned and fetched from a third party service.

I don't see how single binary sync inherently makes this possible. Maybe this can state that an engineering requirement is that a single celestia-app binary can fetch all historical block headers from the chain (?)

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Alternative Approaches
+
+Instead of having a single binary, we can have multiple binaries that are managed by another supervisory binary e.g. [Cosmovisor](https://docs.desmos.network/fullnode/cosmovisor/).
+Osmosis does it through having a significant number of binaries gathered in one docker container and switch them in particular heights.
+This approach has a lot of maintenance burden and is fragile to sync from genesis.
```
rootulp, line 21: > This approach has a lot of maintenance burden

I'm not convinced that the Osmosis approach has any more maintenance burden than what we're proposing with this ADR.

> and is fragile to sync from genesis

Are there scenarios where Osmosis sync from genesis fails? Should we include data to support this point?

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+## Decision
+
+TBD
+
+## Detailed Design
+
+- Cosmos blocks contain a version.app (version of state machine) that can be used to indicate which version of `celestia-app` should be used to process a block.
+  - Example: Osmosis is currently on [version.app: 15](https://www.mintscan.io/osmosis/blocks/8804111) and release [v15.0.0](https://github.com/osmosis-labs/osmosis/releases/tag/v15.0.0).
+
+  - Have to maintain old versions of binaries. Need to update dependencies for security issues.
```
rootulp, line 32: I think this should be moved up as a supporting point for what the maintenance burden is we had multiple binaries required for syncing.

```suggestion
```

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Genesis has an original version.app. Does celestia-app need to explicitly define this?
```
rootulp, line 34: The link got lost in copy + paste but we may not need this line at all

```suggestion
- Genesis has an original version.app. Does celestia-app need to explicitly define [this](https://github.com/celestiaorg/networks/blob/309f961dcdc7e7353e8143cc59c1797e4c0fc326/blockspacerace/genesis.json#L21)?
```

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Bitcoin / Ethereum leverage if statement upgrade logic. Downside to if statement logic in Cosmos ecosystem is if statements may not be relevant for new chains
```
rootulp, line 36:
```suggestion
- Bitcoin and Ethereum leverage if statement upgrade logic. The downside to `if` statement logic in the Cosmos ecosystem is that `if` statements may not be relevant for new chains. For example if a Cosmos chain is hard-spooned to a fresh chain that starts at block height 0, then it does not need to retain the historical `if` statements for irrelevant block heights.
```

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+- UX issue to improve experience for node operators to upgrade
+
+- Maintenance issue for maintaining historical releases of binaries
```
rootulp, line 40: Maybe we can articulate these as the two compelling points driving this ADR. So proposal to life them to the top as part of a problem statement section

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Since if conditional logic uses one Go version, all code will be compiled with this one Go version. If behavior changes across Go versions, we have to debug why.
+
+- Option: depend on multiple major versions of a dependency. What is the relationship between if/else conditional statements and Cosmos SDK upstream changes?
+
+- How to do this in CometBFT?
+  - Always use latest p2p version
+  - Version.block in header should be used to version changes to Tendermint / CometBFT
+
+- Block sync needs to be aware of changes across block heights
```
rootulp, line 70: I'm not sure if these raw notes make sense to include in the ADR but defer to your judgement

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+      maps.Copy(merged, mapOne)
+
+    default:
+        panic(version.ErrVersionNotFound)
+  }
+  return merged
+}
+```
+
+This option has an advantage over the option #1 as we do not need to pass an extra parameter everywhere in the code and it has less maintenance burden to modify the version logic.
```
rootulp, line 124: A counterpoint is that option 2 makes it more difficult to unit test because pure functions are easier to test than impure functions.

Implementations in option 2 rely on a global state. It is likely still possible to unit test by mocking the version module with a test version module but I think it makes sense to call this out as a downside.

Related question: how do we ensure all necessary packages that want to read the global version will invoke it _after_ it has been set?

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Negative
+
+### Neutral
+
+## References
+
+[The meeting notes](https://docs.google.com/document/d/1UuiM9sKQ4g30OBoZLI5pwYBwoicvgC65xm9yzqmbh0g/edit#)
+
+- {reference link}
```
rootulp, line 141: I think we can drop the Google docs link b/c it's a private google doc so external contributors won't be able to view it. It doesn't seem necessary b/c you've copied relevant context into this doc.

```suggestion
```

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,141 @@
...
+  - Version.block in header should be used to version changes to Tendermint / CometBFT
+
+- Block sync needs to be aware of changes across block heights
+
+### Implementation proposal
+
+There are multiple ways of doing it. The app version can be found in the SDK context.
+So, in the `PrepareProposal` method, we can get the app version [from the context](https://github.com/celestiaorg/celestia-app/blob/main/app/prepare_proposal.go#L28) and use it as the version indicator.
+
+#### Option 1: Passing the app version throughout all the required functions
```
rootulp, line 77: 👍

evan-forbes (commented):
great start!!

the initial notes are a really good start, but we should take that information and construct it into a more cohesive ADR document.

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,157 @@
...
+
+In this case we need to pass either sdk context or the app version itself to all the functions and objects that we think they might be affected by a version change in future.
+So for those functions that do not have the ctx already, we need to add an extra parameter to pass it.
```
evan-forbes, line 80: which functions do not have access to the app version? I believe just PrepareProposal, as the header is passed to ProcessProposal

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,157 @@
...
+}
+
+```
+
+Then we call the setter function in `ProcessProposal` where it reads it from the block header.
+The getter function will be called everywhere e.g. [mergeMaps](https://github.com/celestiaorg/celestia-app/blob/f1dec1014a7159c0f0b213182aff4793163e9732/pkg/shares/share_splitting.go#L159)
+
+```go
+func mergeMaps(mapOne, mapTwo map[coretypes.TxKey]ShareRange) map[coretypes.TxKey]ShareRange {
+	merged := make(map[coretypes.TxKey]ShareRange, len(mapOne)+len(mapTwo))
```
evan-forbes, line 105: what is ShareRange here?

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Alternative Approaches
+
+Instead of having a single binary, we can have multiple binaries that are managed by another supervisory binary e.g. [Cosmovisor](https://docs.desmos.network/fullnode/cosmovisor/).
+Osmosis does it through having a significant number of binaries gathered in one docker container and switching them at particular heights.
+This approach has a lot of maintenance burden and is fragile to sync from genesis.
+
+## Decision
+
+TBD
```
evan-forbes, line 25: let's update this to using the app version and arbitrary switch statements

docs/architecture/adr-017-single-binary-upgrade.md
evan-forbes: bumping this :slightly_smiling_face:

docs/architecture/adr-017-single-binary-upgrade.md
```diff
@@ -0,0 +1,157 @@
...
+  - if/else on a line by line basis
+
+- Why app version and not block height?
+  - App version is bumped via upgrade module
+  - Depending on app version is universal across multiple networks.
+  - Block height will differ across testnets / chain-ids so depends on a `.toml` config file
+
+- Are we planning on upstream this?
+  - [CometBFT](https://github.com/cometbft/cometbft) is working on this for consensus.
+  - We’re working on Cosmos SDK portion
```
evan-forbes, line 54: are these just notes from the call? if so, perhaps we should clean this up and focus on the content of this ADR before merging

evan-forbes (commented):
we still need to flesh this out a bit by providing code examples of using the app version and discussing a few of the edge cases, but I think we're on the right track. 

we should also convert this to an ADR and remove any notes that are not relevant to the discussion/decision when we get the chance :slightly_smiling_face:

evan-forbes:
closing for now in favor of #1562

PR #1524, opened 2023-03-22
chore: panic on too short data

rootulp:
Closes https://github.com/celestiaorg/nmt/issues/144

### DIFF
--- a/pkg/wrapper/nmt_wrapper.go
+++ b/pkg/wrapper/nmt_wrapper.go
@@ -81,6 +81,11 @@ func (w *ErasuredNamespacedMerkleTree) Push(data []byte) {
 	if w.axisIndex+1 > 2*w.squareSize || w.shareIndex+1 > 2*w.squareSize {
 		panic(fmt.Sprintf("pushed past predetermined square size: boundary at %d index at %d %d", 2*w.squareSize, w.axisIndex, w.shareIndex))
 	}
+	if len(data) < appconsts.NamespaceSize {
+		// TODO: consider adding an error return parameter to the rsmt.Tree interface for Push
+		// https://github.com/celestiaorg/rsmt2d/issues/156
+		panic(fmt.Sprintf("data is too short to contain namespace ID"))
+	}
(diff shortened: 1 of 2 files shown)


staheri14 (approved):

pkg/wrapper/nmt_wrapper.go
```diff
@@ -81,6 +81,11 @@ func (w *ErasuredNamespacedMerkleTree) Push(data []byte) {
 	if w.axisIndex+1 > 2*w.squareSize || w.shareIndex+1 > 2*w.squareSize {
 		panic(fmt.Sprintf("pushed past predetermined square size: boundary at %d index at %d %d", 2*w.squareSize, w.axisIndex, w.shareIndex))
 	}
+	if len(data) < appconsts.NamespaceSize {
+		// TODO: consider adding an error return parameter to the rsmt.Tree interface for Push
+		// https://github.com/celestiaorg/rsmt2d/issues/156
+		panic("data is too short to contain namespace ID")
+	}
```
evan-forbes, line 88: where does this get checked just to triple check?

I think its when we create a PFB

evan-forbes (approved):

pkg/wrapper/nmt_wrapper.go
rootulp: We push in the following places:
1. https://github.com/rootulp/celestia-app/blob/d993624c6e223c00750fedd3bdb42fff972f9992/pkg/proof/proof.go#L146-L148 and shares should be of length share size. In a future refactor we should make the `Share` type a struct and not a byte slice so we can enforce invariants.
2. https://github.com/rootulp/celestia-app/blob/d993624c6e223c00750fedd3bdb42fff972f9992/x/blob/types/payforblob.go#L214-L218 which contains a namespace and leaf data

FWIW the `Push` function would panic before / after this change because of the line below: 
```
	copy(nidAndData[:appconsts.NamespaceSize], data[:appconsts.NamespaceSize])
```

so this PR just adds an explicit message to the panic.

Issue #1489, opened 2023-03-14
Enforce that a portion of the namespace is not used during consensus


evan-forbes:
after #1419, we need to enforce that only a portion of the namespace is used. This could likely be done several different ways, including but not limited to adding a check to the `ValidateBasic` method of PFBs or explicitly checking during process proposal.

evan-forbes:
cool that makes sense to me, feel free to close if you think its redundant

Issue #1434, opened 2023-03-01
Implement namespace version byte


rootulp:
https://github.com/celestiaorg/celestia-app/pull/1382 reached the accepted status so this issue is for tracking the implementation of a namespace version byte.

evan-forbes:
I'm still not sure how we should track the implementation issues such as this, since they weren't originally part of the q1 goals, but they defintiely feel like part of some larger block encoding refactor/epic thing... still chewing on this though

wdyt @rootulp ?

rootulp:
I can create an epic with all the issues in [label: share encoding](https://github.com/celestiaorg/celestia-app/labels/share%20encoding) if that helps in communicating / tracking our work.

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/1529

Issue #1322, opened 2023-01-31
Test share encoding using random block data using PrepareProposal


evan-forbes:
> I'm concerned the existing "fuzzing" test that tests reconstruction doesn't actually generate pay for blob txs [here](https://github.com/rootulp/celestia-app/blob/a42b83dac1627f428201ca98cea5914205ca5bfe/pkg/shares/merge_test.go#L101)

whoops, I should have pointed to the test I was thinking of, yeah I think we should test this using blocks that are using more realistic data, like with this test.

https://github.com/celestiaorg/celestia-app/blob/main/app/test/fuzz_abci_test.go

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/issues/1298#issuecomment-1409591659_

this will hopefully provide us with enough confidence to no longer use the test data introduced in #1298, and we could then remove it

~~blocked by #1298~~

rootulp:
This issue may not be necessary after transaction fuzzing https://github.com/celestiaorg/celestia-app/pull/1613 and/or https://github.com/celestiaorg/celestia-app/pull/1745

PR #1529, opened 2023-03-23
feat!: versioned namespaces

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/1434
Closes https://github.com/celestiaorg/celestia-app/issues/1489


evan-forbes:
Mind elaborating on how this is apphash breaking to speed up reviews?

mojtaba-esk (approved):
LGTM.
There are a few linter complains as well.

rootulp:
Good question. AppHash is a hash of a block when serialized into a byte array. I assumed the AppHash would be different because the Block contains Data. Data contains Txs. Txs contain BlobTxs. BlobTxs contain the Cosmos SDK txs with a MsgPayForBlobs with namespace IDs that are larger. So I think this change may only indirectly modify AppHash.

go.mod
```diff
@@ -185,5 +185,6 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.8.0-sdk-v0.46.7
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.15.0-tm-v0.34.23.0.20230308212354-07aba19d0bda
+	// TODO remove this replace once we merge dependent PR and cut a celestia-core release
+	github.com/tendermint/tendermint => ../celestia-core
```
evan-forbes, line 189: can we point to a commit here so we can see failing tests here and in our own local repos?

pkg/appconsts/appconsts.go
```diff
@@ -1,16 +1,15 @@
...
 	"github.com/tendermint/tendermint/pkg/consts"
 )
 
 // These constants were originally sourced from:
 // https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/consensus.md#constants
 const (
-	// NamespaceSize is the namespace size in bytes.
-	NamespaceSize = consts.NamespaceSize
+	// NamespaceSize is the size of a namespace (version + ID) in bytes.
+	NamespaceSize = 33
```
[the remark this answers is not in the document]
rootulp: I can move consts from pkg/namespace here if that's what you're asking. I don't want this package to depend on pkg/namespace because that could lead to cyclic imports.

pkg/namespace/consts.go
```diff
@@ -0,0 +1,63 @@
...
+	// MaxReservedNamespace is lexicographically the largest namespace that is
+	// reserved for protocol use.
+	MaxReservedNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 255})
+
+	// TailPaddingNamespaceID is the namespace reserved for tail padding. All data
+	// with this namespace will be ignored.
+	TailPaddingNamespaceID = MustNewV0([]byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE})
+
+	// ParitySharesNamespaceID is the namespace reserved for erasure coded data.
+	ParitySharesNamespaceID = MustNewV0([]byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF})
```
rootulp, line 62: [optional] we may consider removing the `ID` suffix from these constants because they are of type `Namespace`

rootulp: done in https://github.com/celestiaorg/celestia-app/pull/1529/commits/a5a6ade324326e0153a80ac6bbd5848a7d03e221

proto/celestia/blob/v1/event.proto
```diff
@@ -10,5 +10,8 @@ option go_package = "github.com/celestiaorg/celestia-app/x/blob/types";
 message EventPayForBlobs {
   string signer = 1;
   repeated uint32 blob_sizes = 2;
-  repeated bytes namespace_ids = 3;
+  // namespaces is a list of namespaces that the blobs in blob_sizes belong to.
+  // A namespace has length of 33 bytes where the first byte is the
+  // namespaceVersion and the subsequent 32 bytes are the namespaceID.
+  repeated bytes namespaces = 3;
```
evan-forbes, line 16: is this change required?

x/blob/types/payforblob.go
```diff
@@ -195,14 +240,21 @@ func CreateCommitments(blobs []*Blob) ([][]byte, error) {
...
 	for _, blob := range blobs {
-		err := ValidateBlobNamespaceID(blob.NamespaceId)
+		if blob.NamespaceVersion > math.MaxUint8 {
+			return fmt.Errorf("namespace version %d is too large", blob.NamespaceVersion)
+		}
+		ns, err := appns.New(uint8(blob.NamespaceVersion), blob.NamespaceId)
+		if err != nil {
+			return err
+		}
+		err = ns.ValidateBlobNamespace()
```
evan-forbes, line 257: [optional comment/question]
could we combine the validation the the New method? If so, do we gain anything from that?

evan-forbes (commented):
It's difficult to see the exact consensus breaking logic due to the number of changes in this PR, mind pointing them out to make sure I didn't miss anything? I had one blocking question, but I think this makes sense. I will review again after having some time to think over the important pieces here, nice work!

pkg/shares/testdata/sample-block.json
```diff
@@ -3,38 +3,38 @@
 		"version": {
 			"block": 11
 		},
-		"chain_id": "fjOJHN",
+		"chain_id": "chain-id",
```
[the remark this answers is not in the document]
rootulp: Context: https://github.com/celestiaorg/celestia-app/pull/1298 and https://github.com/celestiaorg/celestia-app/issues/1322

I think if we have test data that generates PFB txs then we can test `merge` using such data and can drop the sample-block.json

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,130 @@
...
+func New(version uint8, id []byte) (Namespace, error) {
+	err := validateVersion(version)
+	if err != nil {
+		return Namespace{}, err
+	}
+
+	err = validateID(version, id)
+	if err != nil {
+		return Namespace{}, err
+	}
```
evan-forbes, line 23: should this validation logic go in the validation method?

proto/celestia/blob/v1/event.proto
rootulp: it's a name change so not strictly required but I think it's confusing to have a field called `namespace_ids` that contains the namespace version and namespace ID

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,130 @@
...
+	if n.IsParityShares() {
+		return fmt.Errorf("invalid blob namespace: %v cannot use parity shares namespace ID", n.Bytes())
+	}
+
+	if n.IsTailPadding() {
+		return fmt.Errorf("invalid blob namespace: %v cannot use tail padding namespace ID", n.Bytes())
+	}
+
+	return nil
+}
```
evan-forbes, line 85: shouldn't we also be checking that only a portion of the namespace is used? is that in a different PR?

x/blob/types/blob_tx.go
```diff
@@ -71,10 +72,19 @@ func ValidateBlobTx(txcfg client.TxEncodingConfig, bTx tmproto.BlobTx) error {
...
 
-	for i := range pfb.NamespaceIds {
-		// check that the metadata matches
-		if !bytes.Equal(bTx.Blobs[i].NamespaceId, pfb.NamespaceIds[i]) {
-			return ErrNamespaceMismatch.Wrapf("%v %v", bTx.Blobs[i].NamespaceId, pfb.NamespaceIds[i])
+	for i, ns := range pfb.Namespaces {
+		pfbNamespace, err := appns.From(ns)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
rootulp: 1. https://github.com/rootulp/celestia-app/blob/a5a6ade324326e0153a80ac6bbd5848a7d03e221/pkg/namespace/namespace.go#L55-L56
2. pfbNamespace is probably a confusing variable name. Is `namespace` clearer?

x/blob/types/payforblob.go
rootulp: we could if we created a new function:
```go
NewBlobNamespace
```

I don't think it makes sense to enforce this validation on all usage of `New` because all reserved namespaces will fail the `ValidateBlobNamespace` check

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,130 @@
...
+// validateID returns an error if the provided id does not meet the requirements
+// for the provided version.
+func validateID(version uint8, id []byte) error {
+	if len(id) != NamespaceIDSize {
+		return fmt.Errorf("unsupported namespace id length: id %v must be %v bytes but it was %v bytes", id, NamespaceIDSize, len(id))
+	}
+
+	if version == NamespaceVersionZero && !bytes.HasPrefix(id, NamespaceVersionZeroPrefix) {
+		return fmt.Errorf("unsupported namespace id with version %v. ID %v must start with %v leading zeros", version, id, len(NamespaceVersionZeroPrefix))
+	}
```
evan-forbes, line 104: ahh I see now

pkg/namespace/namespace.go
rootulp: It can but why should the `Namespace` type support creating a new namespace that violates the rules of version / ID? Then consumers need to call `validateBasic()` prior to using a Namespace. If we enforce these variants at construction time (i.e. `New()`) then users don't need to validate the namespace at each usage.

mojtaba-esk (approved):

proto/celestia/blob/v1/event.proto
evan-forbes: I see! yeah, the change makes sense

x/blob/types/payforblob.go
evan-forbes: I see thanks for the clarification

app/process_proposal.go
```diff
@@ -45,8 +45,15 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
-			logInvalidPropBlock(app.Logger(), req.Header, fmt.Sprintf("invalid blob namespace %v", blob.NamespaceID))
+		namespace, err := appns.New(blob.NamespaceVersion, blob.NamespaceID)
+		if err != nil {
+			return abci.ResponseProcessProposal{
+				Result: abci.ResponseProcessProposal_REJECT,
+			}
+		}
+		err = namespace.ValidateBlobNamespace()
+		if err != nil {
+			logInvalidPropBlock(app.Logger(), req.Header, fmt.Sprintf("invalid blob namespace %v", namespace.Bytes()))
```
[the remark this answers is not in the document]
evan-forbes: ahh I see! I just missed that :sweat_smile:

x/blob/types/blob_tx.go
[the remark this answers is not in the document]
rootulp: Attempted to clarify variable name in https://github.com/celestiaorg/celestia-app/pull/1529/commits/41c541386b43d993bbd32e92f37cf60d2c94f20e

pkg/namespace/namespace.go
evan-forbes: thanks, I think I was just confused by why we had a separation of validation logic, but

> I don't think it makes sense to enforce this validation on all usage of New because all reserved namespaces will fail the ValidateBlobNamespace check

explains it

pkg/namespace/namespace.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/1529/files#r1150771472

x/blob/types/payforblob.go
evan-forbes: out of curiosity, where are we using the new function to create reserved namespaces?

rootulp: 1. https://github.com/rootulp/celestia-app/blob/41c541386b43d993bbd32e92f37cf60d2c94f20e/app/process_proposal.go#L48
2. https://github.com/rootulp/celestia-app/blob/41c541386b43d993bbd32e92f37cf60d2c94f20e/pkg/shares/split_sparse_shares.go#L32
3. https://github.com/rootulp/celestia-app/blob/41c541386b43d993bbd32e92f37cf60d2c94f20e/x/blob/client/cli/payforblob.go#L39
+ usage in 3 more files

app/process_proposal.go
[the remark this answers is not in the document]
rootulp: I'm still working on a ProcessProposal test for this

pkg/namespace/consts.go
```diff
@@ -0,0 +1,63 @@
...
+	// NamespaceVersionZero is the first namespace version.
+	NamespaceVersionZero = uint8(0)
+
+	// NamespaceZeroPrefixSize is the number of `0` bytes that are prefixed to
+	// namespace IDs for version 0.
+	NamespaceVersionZeroPrefixSize = 22
+
+	// NamespaceVersionZeroIDSize is the number of bytes available for
+	// user-specified namespace ID in a namespace ID for version 0.
+	NamespaceVersionZeroIDSize = NamespaceIDSize - NamespaceVersionZeroPrefixSize
```
staheri14, line 28: [suggestion-optional] what do you think about defining a type for Namespace version `0` e.g., an array of 10 bytes? it may make type conversions and boundary checks easier.

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,130 @@
...
+func MustNew(version uint8, id []byte) Namespace {
+	ns, err := New(version, id)
+	if err != nil {
+		panic(err)
+	}
+	return ns
+}
+
+// MustNewV0 returns a new namespace with version 0 and the provided id. This
+// function panics if the provided id is not exactly 10 bytes.
```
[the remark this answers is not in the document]
staheri14:
```suggestion
// function panics if the provided id is not exactly NamespaceVersionZeroIDSize bytes.
```

staheri14:
Great work! added some quick suggestions above, will continue my review.

app/estimate_square_size_test.go
```diff
@@ -159,7 +159,7 @@ func Test_estimateTxSharesUsed(t *testing.T) {
 		{"one tx", generateNormalTxs(1), 1},             // 1 tx is approximately 316 bytes which fits in 1 share
 		{"two txs", generateNormalTxs(2), 2},            // 2 txs is approximately 632 bytes which fits in 2 shares
 		{"ten txs", generateNormalTxs(10), 7},           // 10 txs is approximately 3160 bytes which fits in 7 shares
-		{"one hundred txs", generateNormalTxs(100), 67}, // 100 txs is approximately 31600 bytes which fits in 67 share
+		{"one hundred txs", generateNormalTxs(100), 68}, // 100 txs is approximately 31600 bytes which fits in 68 share
```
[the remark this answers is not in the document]
rootulp: good catch, will update.

pkg/namespace/consts.go
```diff
@@ -0,0 +1,63 @@
...
+	NamespaceVersionSize = appconsts.NamespaceVersionSize
+
+	// NamespaceIDSize is the size of a namespace ID in bytes.
+	NamespaceIDSize = appconsts.NamespaceIDSize
+
+	// NamespaceSize is the size of a namespace (version + ID) in bytes.
+	NamespaceSize = appconsts.NamespaceSize
+
+	// NamespaceVersionZero is the first namespace version.
+	NamespaceVersionZero = uint8(0)
```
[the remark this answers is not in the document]
rootulp: I'm not opposed. What is the benefit of using a type alias instead of using `uint8` directly?

Ref: https://yourbasic.org/golang/type-alias/ doesn't point out any strong benefits.

pkg/namespace/consts.go
rootulp: I'm not opposed to using a fixed length array instead of a byte slice but I don't expect it to make type conversions easier. Can you elaborate on why a fixed length array is a better choice than a slice here?

app/estimate_square_size_test.go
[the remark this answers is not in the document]
rootulp: resolved by https://github.com/celestiaorg/celestia-app/pull/1529/commits/c9fbfbbe3a290c8a96613d7431943827379f4160

app/process_proposal.go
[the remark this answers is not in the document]
rootulp: Spun out https://github.com/celestiaorg/celestia-app/issues/1558 b/c we want to merge this to unblock testing an unrelated PR

pkg/appconsts/appconsts.go
```diff
@@ -1,16 +1,21 @@
...
-	// NamespaceSize is the namespace size in bytes.
-	NamespaceSize = consts.NamespaceSize
+	// NamespaveVersionSize is the size of a namespace version in bytes.
+	NamespaceVersionSize = 1
+
+	// NamespaceIDSize is the size of a namespace ID in bytes.
+	NamespaceIDSize = 32
+
+	// NamespaceSize is the size of a namespace (version + ID) in bytes.
+	NamespaceSize = NamespaceVersionSize + NamespaceIDSize
```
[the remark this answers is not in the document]
staheri14: I see, The way that it has been used in the `nmt` package is different, the `namespace ID` and `namespace` words are used interchangeably. For example, https://github.com/celestiaorg/nmt/blob/eded06696df88e15fcd8625047653957f548e383/nmt.go#L15. 
We don't have to use my suggested name i.e., `NamespaceIDVersionSize`,any name that resolves the potential confusion would work.

pkg/namespace/consts.go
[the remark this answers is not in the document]
staheri14: Good question! For instance, if we were to allocate 16 bits for the version in the future, we could easily do so by updating the alias type. This change would immediately clarify all other places where the alias is used, making refactoring a simple and straightforward process.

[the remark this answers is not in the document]
staheri14: At present, the reason for defining `NamespaceVersionZero` (or any other variable holding a version type as its value) as uint8 is unclear. It is unclear whether this is intended for mathematical operations with another uint8 value or if the version is supposed to be 8 bits long by default. This question may also arise in other parts of the code.

pkg/namespace/consts.go
staheri14: Sure, the main advantage is that we can avoid having to check the input size in every function call to ensure that it matches the expected length. By defining and initializing the variable as an array, we can safely pass it around the code without worrying about its size.

rootulp: The namespace struct enforces an invariant for the length of a namespace so code that passes around namespaces don't need to check for length.

pkg/appconsts/appconsts.go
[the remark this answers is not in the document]
rootulp: I think the crux is that a namespace in celestia-app is being used as a namespace ID in the NMT repo so one option is to rename instances of namespace ID to namespace in the nmt repo.

pkg/namespace/consts.go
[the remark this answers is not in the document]
rootulp: The three articles I read on it just now make it seem like the primary benefit is refactoring. So I agree with your point on refactoring but I think it comes at the cost of readability (subjective argument) because IMO this

```go
func NewCompactShareSplitter(ns appns.Namespace, shareVersion uint8) *CompactShareSplitter {
```

is slightly easier to read than this:

```go
func NewCompactShareSplitter(ns appns.Namespace, shareVersion appconsts.ShareVersion) *CompactShareSplitter {
```

because the latter forces you to hover or jump to another file to learn what underlying type a `ShareVersion` is. I'm a little hesitant to introduce them because:

> Type aliases are not meant for everyday use.

Ref: https://yourbasic.org/golang/type-alias/

> Type alias declaration wasn’t designed with everyday use in mind. 

Ref: https://gosolve.io/golang-type-alias/

but will open a follow-up issue if you have strong opinions about adopting it.

PR #1540, opened 2023-03-24
feat(app)!: add authz to the state-machine

liamsi:
## Checklist


cmwaters (approved):
LGTM - I don't think you've missed any places where authz needs to be declared

evan-forbes (approved):
optionally, we could add a test case to the [standard sdk smoke test](https://github.com/celestiaorg/celestia-app/blob/142ff04832f4ee72833be5acb67f1673fb33f962/app/test/std_sdk_test.go#L62-L206), but I'll defer to the author as its purely bonus

LGTM

Issue #1440, opened 2023-03-02
specs: public keys


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/329dc50c51ee4865439ef3ef2d466c0ab61bcc28/specs/src/specs/data_structures.md?plain=1#L245-L265

## Problem

Public keys use Cosmos SDK public keys which appear to be 20 bytes not 32 bytes. Ref:
1. https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys
2. https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-028-public-key-addresses.md#legacy-public-key-addresses-dont-change

## Proposal

Update the specs to match the implementation

rootulp:
>  [Public keys](#publickey) are encoded in uncompressed form, as the concatenation of the `x` and `y` values. No prefix is needed to distinguish between encoding schemes as this is the only encoding supported. 

I can't find an instance in celestia-app that deals with uncompressed public keys. I don't think we've made modifications in celestia-core specific to public keys so I would expect this section to describe public keys and their usage in [CometBFT](https://github.com/celestiaorg/celestia-core/blob/e3bad3d08ce3372fb7473b9e13be00cdaaa4fa91/spec/core/encoding.md#public-key-cryptography)

rootulp:
I'm trying to verify the accuracy of the signature section. I don't see where the `s` value is modified to include the `v` value. Signatures do appear to be 64 bytes.

Ref:
- https://github.com/cosmos/cosmos-sdk/blob/d29e8eb4f6ac9c26de7b6f275eb5552f7c862f20/crypto/keys/secp256k1/secp256k1_nocgo.go#L15-L17
- https://github.com/btcsuite/btcd/blob/2f508b3f86ed9ef87bcf3426b87b6c0dc0d3632c/btcec/privkey.go#L57-L61
- https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki
- https://github.com/tendermint/btcd/blob/ec996c56d1b44508d4b0b511445129b21c058fed/btcec/signature.go#L33-L50

rootulp:
Update: I don't think the `v` component exists in our signatures b/c 

<img width="795" alt="Screenshot 2023-04-06 at 3 45 34 PM" src="https://user-images.githubusercontent.com/3699047/230479370-8f75b872-a235-459a-a2c5-3dcca420e409.png">

PR #1553, opened 2023-03-28
docs: public key crypto specs

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1440

### DIFF
--- a/specs/src/specs/data_structures.md
+++ b/specs/src/specs/data_structures.md
@@ -171,13 +171,7 @@ Public keys are serialized in a compressed format described [here](https://docs.
 
 Deterministic signatures ([RFC-6979](https://tools.ietf.org/rfc/rfc6979.txt)) should be used when signing, but this is not enforced at the protocol level as it cannot be.
 
-[Signatures](#signature) are represented as the `r` and `s` (each 32 bytes) values of the signature. `r` and `s` take on their usual meaning (see: [SEC 1, 4.1.3 Signing Operation](https://www.secg.org/sec1-v2.pdf)). The encoding for signatures is:
-
-<!-- markdownlint-disable-next-line MD040 -->
-```
-|    32 bytes   ||    32 bytes   |
-[256-bit r value][256-bit s value]
-```
+[Signatures](#signature) are represented as the `r` and `s` (each 32 bytes) values of the signature. `r` and `s` take on their usual meaning (see: [SEC 1, 4.1.3 Signing Operation](https://www.secg.org/sec1-v2.pdf)). Signatures are encoded with protobuf as described [here](https://docs.cosmos.network/v0.46/core/encoding.html#transaction-encoding).
 
 ## Merkle Trees
 
--- a/specs/src/specs/data_structures.md
+++ b/specs/src/specs/data_structures.md
@@ -10,7 +10,6 @@
 
 | name                        | type                        |
 |-----------------------------|-----------------------------|
-| [`Address`](#address)       | `byte[32]`                  |
 | `Amount`                    | `uint64`                    |
 | `Graffiti`                  | `byte[MAX_GRAFFITI_BYTES]`  |
 | [`HashDigest`](#hashdigest) | `byte[32]`                  |
@@ -112,11 +111,7 @@ Abstraction over transaction fees.
 
 ### Address
 
-Address is a [type alias](#type-aliases).
-
-Addresses are the [hash](#hashing) [digest](#hashdigest) of the [public key](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys).
-
-Addresses have a length of 32 bytes.
+Celestia supports [secp256k1](https://en.bitcoin.it/wiki/Secp256k1) keys where [addresses](https://docs.cosmos.network/v0.46/basics/accounts.html#addresses) are 20 bytes in length. Addresses are prefixed with the [Bech32](https://en.bitcoin.it/wiki/Bech32) prefix `celestia`. For example, a valid address is `celestia1kj39jkzqlr073t42am9d8pd40tgudc3e2kj9yf`.
 
 ### CommitSig
 
@@ -172,7 +167,7 @@ Unless otherwise indicated explicitly, objects are first [serialized](#serializa
 
 Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1-v2.pdf), with the curve [secp256k1](https://en.bitcoin.it/wiki/Secp256k1). A highly-optimized library is available in C (<https://github.com/bitcoin-core/secp256k1>), with wrappers in Go (<https://pkg.go.dev/github.com/ethereum/go-ethereum/crypto/secp256k1>) and Rust (<https://docs.rs/crate/secp256k1>).
 
-[Public keys](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys) are encoded in uncompressed form, as the concatenation of the `x` and `y` values. No prefix is needed to distinguish between encoding schemes as this is the only encoding supported.
+[Public keys](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys) are serialized in a compressed format.
 
 Deterministic signatures ([RFC-6979](https://tools.ietf.org/rfc/rfc6979.txt)) should be used when signing, but this is not enforced at the protocol level as it cannot be.
 
--- a/specs/src/specs/data_structures.md
+++ b/specs/src/specs/data_structures.md
@@ -171,19 +171,13 @@ Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1
 
 Deterministic signatures ([RFC-6979](https://tools.ietf.org/rfc/rfc6979.txt)) should be used when signing, but this is not enforced at the protocol level as it cannot be.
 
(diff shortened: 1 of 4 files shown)


specs/src/specs/data_structures.md
```diff
@@ -172,24 +167,18 @@ Unless otherwise indicated explicitly, objects are first [serialized](#serializa
...
-
-`v` represents the parity of the `Y` component of the point, `0` for even and `1` for odd. The `X` component of the point is assumed to always be low, since [the possibility of it being high is negligible](https://bitcoin.stackexchange.com/a/38909).
-
-Putting it all together, the encoding for signatures is:
+[Signatures](#signature) are represented as the `r` and `s` (each 32 bytes) values of the signature. `r` and `s` take on their usual meaning (see: [SEC 1, 4.1.3 Signing Operation](https://www.secg.org/sec1-v2.pdf)). The encoding for signatures is:
```
rootulp, line 174: @adlerjohn I'd appreciate your review on this section

specs/src/specs/data_structures.md
```diff
@@ -112,11 +111,7 @@ Abstraction over transaction fees.
 
 ### Address
 
-Address is a [type alias](#type-aliases).
-
-Addresses are the [hash](#hashing) [digest](#hashdigest) of the [public key](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys).
-
-Addresses have a length of 32 bytes.
+Celestia supports [secp256k1](https://en.bitcoin.it/wiki/Secp256k1) keys where [addresses](https://docs.cosmos.network/v0.46/basics/accounts.html#addresses) are 20 bytes in length. Addresses are prefixed with the [Bech32](https://en.bitcoin.it/wiki/Bech32) prefix `celestia`. For example, a valid address is `celestia1kj39jkzqlr073t42am9d8pd40tgudc3e2kj9yf`.
```
adlerjohn, line 114: The HRP of the bech32 encoding (and indeed the encoding itself) is orthogonal to the actual address value. I think the specs should have a section for the value, and a section for the human-readable encoding used by frontends.

specs/src/specs/data_structures.md
```diff
@@ -172,24 +167,18 @@ Unless otherwise indicated explicitly, objects are first [serialized](#serializa
 
 Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1-v2.pdf), with the curve [secp256k1](https://en.bitcoin.it/wiki/Secp256k1). A highly-optimized library is available in C (<https://github.com/bitcoin-core/secp256k1>), with wrappers in Go (<https://pkg.go.dev/github.com/ethereum/go-ethereum/crypto/secp256k1>) and Rust (<https://docs.rs/crate/secp256k1>).
 
-[Public keys](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys) are encoded in uncompressed form, as the concatenation of the `x` and `y` values. No prefix is needed to distinguish between encoding schemes as this is the only encoding supported.
+[Public keys](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys) are serialized in a compressed format.
```
adlerjohn, line 170:
```suggestion
Public keys are serialized in a compressed format described [here](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys).
```

specs/src/specs/data_structures.md
adlerjohn: Are signatures _encoded_ as described here, or encoded with protobuf? Also, there's no `v` here?

rootulp: I don't think `v` is supposed to be here. Ref: https://github.com/celestiaorg/celestia-app/issues/1440#issuecomment-1486806399

rootulp: > Are signatures encoded as described here, or encoded with protobuf?

Good point. They appear to be encoded using protobuf per

https://github.com/celestiaorg/cosmos-sdk/blob/f19e20459b8bcf195167ad256de5a6ca6edc41f5/x/auth/tx/encoder.go#L13-L29

specs/src/specs/data_structures.md
rootulp: I can extract a new section for the human-readable encoding but it seems like it doesn't belong on this page (data_structures.md). Proposal to extract a new page for `Public-Key Cryptography` which contains the section from this page along with a new section for human-readable Address encoding.

rootulp: Done in https://github.com/celestiaorg/celestia-app/pull/1553/commits/cb52073e9cd3f9f69ef432d9d408db97c7767046

specs/src/specs/data_structures.md
```diff
@@ -112,11 +111,11 @@ Abstraction over transaction fees.
...
 
-Address is a [type alias](#type-aliases).
+Celestia supports [secp256k1](https://en.bitcoin.it/wiki/Secp256k1) keys where [addresses](https://docs.cosmos.network/v0.46/basics/accounts.html#addresses) are 20 bytes in length.
 
-Addresses are the [hash](#hashing) [digest](#hashdigest) of the [public key](https://docs.cosmos.network/v0.46/basics/accounts.html#public-keys).
-
-Addresses have a length of 32 bytes.
+| name         | type       | description                                                             |
+|--------------|------------|-------------------------------------------------------------------------|
+| `AccAddress` | `[20]byte` | AccAddress a wrapper around bytes meant to represent an account address |
```
rootulp, line 118: [for reviewers] inspired by https://github.com/celestiaorg/cosmos-sdk/blob/f19e20459b8bcf195167ad256de5a6ca6edc41f5/types/address.go#L130-L132

specs/src/specs/public_key_cryptography.md
```diff
@@ -0,0 +1,23 @@
+# Public-Key Cryptography
+
+<!-- toc -->
+
+Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1-v2.pdf), with the curve [secp256k1](https://en.bitcoin.it/wiki/Secp256k1). A highly-optimized library is available in C (<https://github.com/bitcoin-core/secp256k1>), with wrappers in Go (<https://pkg.go.dev/github.com/ethereum/go-ethereum/crypto/secp256k1>) and Rust (<https://docs.rs/crate/secp256k1>).
```
liamsi, line 5: Just noting that this file and section makes it sounds as if ecdsa/secp was the only pub-key crypto we were using. Validators still use ed25519 during consensus (which is consensus critical data).

adlerjohn: Both signatures schemes should be described then.

specs/src/specs/data_structures.md
adlerjohn: In that case I would distinguish between the _value_ and the _encoding_ of that value.

rootulp: Agreed. I attempted to in https://github.com/celestiaorg/celestia-app/pull/1553/commits/0d691af1d3c42427dc73ef936b7e055ecfec0524 and https://github.com/celestiaorg/celestia-app/pull/1553/commits/cb52073e9cd3f9f69ef432d9d408db97c7767046

specs/src/specs/public_key_cryptography.md
rootulp: Agreed will add a section for ed25519 based on https://github.com/cometbft/cometbft/blob/be6f4272d4ebedf264e3ce29475ffd31ecadff47/spec/core/encoding.md?plain=1#L42-L66

rootulp: Added a section in https://github.com/celestiaorg/celestia-app/pull/1553/commits/64164c41784b508da778bd630c2f3c3db1362695

specs/src/specs/public_key_cryptography.md
```diff
@@ -0,0 +1,66 @@
+# Public-Key Cryptography
+
+<!-- toc -->
+
+Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1-v2.pdf) with the curves: [Secp256k1](https://en.bitcoin.it/wiki/Secp256k1) or [Ed25519](https://en.wikipedia.org/wiki/EdDSA#Ed25519).
+
+## Secp256k1
+
+The Secp256k1 key type is used by accounts that submit transactions to be included in Celestia.
```
cmwaters, line 9: Doesn't the SDK support multiple key types? Not just Secp256k1

cmwaters (approved):
LGTM

specs/src/specs/public_key_cryptography.md
rootulp: Yes it looks like it does. I can't seem to create one via celestia-appd though:

```go
$ ./build/celestia-appd keys add test --algo ed25519
Error: provided algorithm "ed25519" is not supported
```

so still investigating.

Ref:
1. https://docs.cosmos.network/v0.46/run-node/keyring.html#adding-keys-to-the-keyring

rootulp: Relevant: https://github.com/cosmos/cosmos-sdk/discussions/8543

rootulp: Resolving for now b/c I don't think we can actually use ed25519 keys in Cosmos SDK.

evan-forbes (approved):
:+1:

PR #1554, opened 2023-03-28
feat: inflation rate change

mojtaba-esk:
This PR aims to implement the inflation schedule that is currently being discussed.
The initial parameters are determined as follows:

| Initial Inflation |     8.00% | 
|------------------------|----------------------------|
| Disinflation rate p.a	|    -10.00% | 
| Target Inflation (floor)	 |  1.50% | 

It reduces the inflation per year until it reaches the target inflation.

The method that is modified mostly is the `NextInflationRate` in `x/mint/types/minter.go`.

**Note:** if we stick to `10%` decrease per year, we will have something like this:

| year | inflation (%) |
|------|------|
|0  | 8.00 |
|1  | 7.20 |
|2  | 6.48 |
|3  | 5.832 |
|4  | 5.2488 |
|5  | 4.72392 |
|6  | 4.251528 |
|7  | 3.8263752 |
|8  | 3.44373768 |
|9  | 3.099363912 |
|10 | 2.7894275208 |
|11 | 2.51048476872 |
|12 | 2.259436291848 |
|13 | 2.0334926626632 |
|14 | 1.83014339639688 |
|15 | 1.647129056757192 |
|16 | 1.50 |
|17 | 1.50 |
|18 | 1.50 |
|19 | 1.50 |
|20 | 1.50 |

@liamsi please have a look at the table above. If we do not round the inflation rates, it will look like that.

## Checklist


liamsi (commented):
We haven't made a final decision yet btw. Though the outcome will be like the proposed inflation schedule. Note that I definitely want to review this PR once done and also would appreciate an ADR (I can write that too if necessary).

evan-forbes (commented):
This is a good exploration into the mint module @mojtaba-esk, but we should clarify if we want this PR to ever be merged.

There's definitely a lot that we can remove here, and this isn't actually switching from the mint module in the [application](https://github.com/celestiaorg/celestia-app/blob/053cbad425e0008aedb56e3c15d684a2be96b37c/app/app.go#L53-L55). IIRC, this was a good exercise to see how the mint module is incorporated into the app, but we can, and should imo, do something significantly simpler. Our mint module will not need any logic that is pertaining to moving the inflation rate dynamically, since we're following a strict schedule. 

as a side note I think the important thing to document here is to understand the flow of funds for the existing module, and how that change that fit the simpler new schedule.

evan-forbes (changes requested):
adding a block here just to be explicit, but the above comment explains some of the necessary changes

mojtaba-esk:
Closing due to some gh issues in file changes. Another PR is opened to make the review easier.

PR #1559, opened 2023-03-29
chore: move NMT spec

rootulp:
This PR moves the NMT spec to the place I expected it to be. I was looking for it in `pkg/wrapper/*` because other packages have docs (or specs) in the root of the directory (e.g. https://github.com/celestiaorg/celestia-app/tree/main/x/blob and https://github.com/celestiaorg/celestia-app/tree/main/x/blob).


staheri14 (approved):
The proposed change seems to align with how other documentation is organized in this repository. However, (as a general comment) it is a bit confusing to locate specific documentation since we also have a `specs` folder. It would be helpful to provide guidance in the repository's main README file on how docs/specs are organized.

Issue #1570, opened 2023-03-30
EPIC: Strict inflation schedule


evan-forbes:
We will be using a simplified inflation schedule that is strict. We start at 8%, decrease 10% each year, and end at 1.5%

rootulp:
- ADR https://github.com/celestiaorg/celestia-app/pull/1584
- Implementation https://github.com/celestiaorg/celestia-app/pull/1578
- Testing https://github.com/celestiaorg/celestia-app/issues/1758

rootulp:
Closing this epic b/c the ADR and implementation merged.

PR #1578, opened 2023-03-31
feat: strict inflation schedule

mojtaba-esk:
### Description

Part of https://github.com/celestiaorg/celestia-app/issues/1570
Implement the strict inflation schedule specified in https://github.com/celestiaorg/celestia-app/pull/1584

### Remaining tasks
- [x] implement inflation rate change based on block timestamp rather than block height
- [x] set genesis time in x/mint state
- [x] move `MintDenom` from param to `x/mint` state
- [x] remove `TestRandomizedGenState`
- [x] investigate why `TestQueryGRPC` / other integration tests are never run
- [x] write a test that 1. starts a new chain, 2. verifies state of x/mint module for first few blocks, 3. skips ahead to a block one year later, 4. verifies inflation rate went down a bit
- [x] remove `Params` entirely
- [x] annual provisions for a year should be based on total supply at the beginning of that year
- [x] for time-based inflation, calculate block provision as current block timestamp - previous block timestamp = seconds elapsed / seconds in a year * annual provisions for a year
- [x] remove `BlocksPerYear`
- [x] elaborate in README.md

### FLUPs

- https://github.com/celestiaorg/celestia-app/issues/1755
- https://github.com/celestiaorg/celestia-app/issues/1757
- https://github.com/celestiaorg/celestia-app/issues/1758
- https://github.com/celestiaorg/celestia-app/issues/1756


liamsi:
> please have a look at the table above. If we do not round the inflation rates, it will look like that.

table looks fine IMO.

x/mint/types/params.go
```diff
@@ -0,0 +1,94 @@
...
+func ParamKeyTable() paramtypes.KeyTable {
+	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
+}
+
+func NewParams(
+	mintDenom string, inflationRateChange, inflationMax, inflationMin, goalBonded sdk.Dec, blocksPerYear uint64,
+) Params {
+	return Params{
+		MintDenom:     mintDenom,
+		BlocksPerYear: blocksPerYear,
```
liamsi, line 30: Need to think about that one. On one hand, I think the gov-surface should be as small as possible and that would also apply here. On the other hand, the block times are not very stable and it might be good to leave this changeable without requiring a software upgrade.

x/mint/types/params.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+// default minting module parameters
+func DefaultParams() Params {
+	return Params{
+		MintDenom:     sdk.DefaultBondDenom,
+		BlocksPerYear: uint64(60 * 60 * 8766 / 5), // assuming 5 second block times
```
liamsi, line 38: This should be an exported and documented constant btw (so the go-docs also document the assumptions around block times). I know this is still in draft status but I think we are going to have slower block times than 5s.

mojtaba-esk: Done

x/mint/types/params.go
```diff
@@ -0,0 +1,94 @@
...
+// ParamTable for minting module.
+func ParamKeyTable() paramtypes.KeyTable {
+	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
+}
+
+func NewParams(
+	mintDenom string, inflationRateChange, inflationMax, inflationMin, goalBonded sdk.Dec, blocksPerYear uint64,
+) Params {
+	return Params{
+		MintDenom:     mintDenom,
```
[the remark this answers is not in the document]
rootulp: Removed this param.

x/mint/types/params.go
rootulp: Removed this param.

app/test/fuzz_abci_test.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
-	"github.com/celestiaorg/celestia-app/test/util"
+	testutil "github.com/celestiaorg/celestia-app/test/util"
```
rootulp, line 9: not sure why this import alias didn't already exist

proto/celestia/mint/v1/mint.proto
```diff
@@ -0,0 +1,38 @@
...
+  // AnnualProvisions is the total number of tokens to be minted in the current
+  // year due to inflation.
+  string annual_provisions = 2 [
+    (cosmos_proto.scalar)  = "cosmos.Dec",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+
+  // GenesisTime is the timestamp of the genesis block.
+  google.protobuf.Timestamp genesis_time = 3 [(gogoproto.stdtime) = true];
```
rootulp, line 29: [TODO] @rootulp it's a bit confusing that this timestamp gets populated with the timestamp of block height = 1 and not genesis time. Explore clarifying this (either rename) or see how other chains name block 1

rootulp: nvm there won't be a block at height=0 per https://github.com/celestiaorg/networks/blob/309f961dcdc7e7353e8143cc59c1797e4c0fc326/blockspacerace/genesis.json#L4 so this variable name seems fine

proto/celestia/mint/v1/mint.proto
```diff
@@ -0,0 +1,38 @@
...
+  // current year. For example if InflationRate=0.1, then 10% of the total
+  // supply will be minted over the course of the year.
+  string inflation_rate = 1 [
+    (cosmos_proto.scalar)  = "cosmos.Dec",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+  // AnnualProvisions is the total number of tokens to be minted in the current
+  // year due to inflation.
+  string annual_provisions = 2 [
```
cmwaters, line 22: Just for my understanding, is this effectively the total supply multiplied by the inflation rate

x/mint/README.md
```diff
@@ -0,0 +1,33 @@
+# `x/mint`
+
+`x/mint` is a fork of the Cosmos SDK `x/mint` module that makes some changes to the inflation mechanism. See ADR-019 for the motivation behind these changes.
```
cmwaters, line 3: I think it would be better if there was a brief description to what these changes are. Most people looking for information around minting will come here first (unless we want to house this information under the `spec` directory)

x/mint/abci_test.go
```diff
@@ -0,0 +1,171 @@
...
+			name: "inflation rate is 0.0648 for year two",
+			ctx:  ctx.WithBlockTime(yearTwo),
+			want: sdk.NewDecWithPrec(648, 4), // 0.0648
+		},
+		{
+			name: "inflation rate is 0.015 for year twenty",
+			ctx:  ctx.WithBlockTime(yearTwenty),
+			want: sdk.NewDecWithPrec(15, 3), // 0.015
+		},
+	}
```
cmwaters, line 101: Could you add a test case where it's one year minus one minute (so should sill be 0.08)

x/mint/abci_test.go
```diff
@@ -0,0 +1,171 @@
...
+			ctx:  ctx.WithBlockTime(yearZeroAndSixMonths),
+			want: sdk.NewDec(80_000), // 1,000,000 (total supply) * 0.08 (inflation rate)
+		},
+		{
+			name: "annual provisions is 72,000 for year one",
+			ctx:  ctx.WithBlockTime(yearOne),
+			want: sdk.NewDec(72_000), // 1,000,000 (total supply) * 0.072 (inflation rate)
+		},
+		{
+			name: "annual provisions is 72,000 for year two",
```
cmwaters, line 149:
```suggestion
			name: "annual provisions is 64,800 for year two",
```

x/mint/abci_test.go
```diff
@@ -0,0 +1,171 @@
...
+			ctx:  ctx.WithBlockTime(yearTwo),
+			want: sdk.NewDec(64_800), // 1,000,000 (total supply) * 0.0648 (inflation rate)
+		},
+		{
+			name: "annual provisions is 72,000 for year twenty",
```
cmwaters, line 154:
```suggestion
			name: "annual provisions is 15,000 for year twenty",
```

x/mint/abci_test.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+		{
+			name: "annual provisions is 80,000 for year zero and six months",
+			ctx:  ctx.WithBlockTime(yearZeroAndSixMonths),
+			want: sdk.NewDec(80_000), // 1,000,000 (total supply) * 0.08 (inflation rate)
+		},
+		{
+			name: "annual provisions is 72,000 for year one",
+			ctx:  ctx.WithBlockTime(yearOne),
+			want: sdk.NewDec(72_000), // 1,000,000 (total supply) * 0.072 (inflation rate)
```
cmwaters, line 146: I'm guessing this doesn't take into consideration the compounding effects of inflation?

x/mint/types/constants.go
```diff
@@ -0,0 +1,28 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+const (
+	SecondsPerMinute = 60
+	MinutesPerHour   = 60
+	HoursPerDay      = 24
+	DaysPerYear      = 365
+	SecondsPerYear   = SecondsPerMinute * MinutesPerHour * HoursPerDay * DaysPerYear // 31,536,000
```
cmwaters, line 10: I'm not sure if you want to go down the rabbit hole of time but using a Gregorian calendar there is actually 365.2425 days in a year (because of the leap year) which means theres actually 31556952 seconds in a year. YMMV

cmwaters (approved):
Logically this all looks sound. I just have a few minor comments.

If I understand it correctly, the inflation rate we use is dependent on the block time i.e. the time that has passed since genesis, but the actual minting of coins depends on block height and a hardcoded estimation at the amount of blocks per year.

rootulp:
> but the actual minting of coins depends on block height and a hardcoded estimation at the amount of blocks per year.

Clarification: the actual minting of coins depends on a hardcoded estimation for the amount of blocks per year but it doesn't depend on block height. @cmwaters

proto/celestia/mint/v1/mint.proto
```diff
@@ -0,0 +1,38 @@
...
+  google.protobuf.Timestamp genesis_time = 3 [(gogoproto.stdtime) = true];
+
+  // BondDenom is the denomination of the token that should be minted.
+  string bond_denom = 4;
+}
+
+// Params holds parameters for the mint module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+}
```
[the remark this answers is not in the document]
rootulp: We don't need it. I can explore removing it and that would lead to fewer lines maintained.

proto/celestia/mint/v1/mint.proto
rootulp: Exactly.

x/mint/README.md
rootulp: Good idea, will list the changes here.

x/mint/abci_test.go
rootulp: Great idea!

x/mint/types/constants.go
rootulp: Related: https://github.com/cosmos/cosmos-sdk/issues/16009

I choose to use `365` because I value readability and don't think a high degree of precision is necessary for this calculation but I can update for higher precision with more documentation.

x/mint/abci_test.go
rootulp: 9621325a0c9a1e4c6f11c07b773482e6e49cc4fc

x/mint/types/constants.go
rootulp: fd0274ee7cda88cb9835619ffe0e272f903c87da

evan-forbes (approved):
Nice!! this is a really great starting point, and I couldn't find anything that we would want to remove. 

This PR is already quite large due to all the tests, so I think we can merge and iterate, but will defer to the author. Left some comments to start a few discussions around :slightly_smiling_face: 

what happens if the blocks per year is not accurate here? I didn't see one, so apologies if there already is one, but could we have a test for that somehow to document that a certain behavior is expected?

x/mint/types/minter.go
```diff
@@ -0,0 +1,71 @@
...
+func (m Minter) CalculateAnnualProvisions(totalSupply math.Int) sdk.Dec {
+	return m.InflationRate.MulInt(totalSupply)
+}
+
+// CalculateBlockProvision returns the total number of coins that should be
+// minted due to inflation for the current block).
+func (m Minter) CalculateBlockProvision() sdk.Coin {
+	blockProvision := m.AnnualProvisions.QuoInt(blocksPerYear)
+	return sdk.NewCoin(m.BondDenom, blockProvision.TruncateInt())
+}
```
evan-forbes, line 62: my understanding of this [thread](https://github.com/celestiaorg/celestia-app/pull/1584#discussion_r1160877702) was that we wanted to use the time in the header to calculate this value to avoid having to hardcode blocks per year. Is that possible? Where we know the time elapsed thanks to the headers, and can then use a percentage of the annual provisions purely based on that

x/mint/module.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+// Name returns the mint module's name.
+func (AppModule) Name() string {
+	return types.ModuleName
+}
+
+// RegisterInvariants registers the mint module invariants.
+func (am AppModule) RegisterInvariants(_ sdk.InvariantRegistry) {}
```
evan-forbes, line 102: not entirely sure what this would look like, but perhaps one potential useful invarient to add would be a check that the funds minted in a given year never exceed the annual provisions. 

like, maybe we keep track of the total funds minted in a year, and then compare here

x/mint/types/minter.go
evan-forbes: ahhhhh just saw ahhh just saw https://github.com/celestiaorg/celestia-app/issues/1730

evan-forbes:
> what happens if the blocks per year is not accurate here?

nvm just saw #1730

proto/celestia/mint/v1/mint.proto
rootulp: Good point https://github.com/celestiaorg/networks/blob/309f961dcdc7e7353e8143cc59c1797e4c0fc326/blockspacerace/genesis.json#L2 exists and it matches https://testnet.mintscan.io/celestia-incentivized-testnet/blocks/1 so I need to explore if it's accessible from this module

proto/celestia/mint/v1/mint.proto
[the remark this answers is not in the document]
rootulp: Resolved via https://github.com/celestiaorg/celestia-app/pull/1578/commits/6dd1fc34927bc37f19f2b1c67b2cf8a594c7b829

x/mint/module.go
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/1755

rootulp:
CI fails on a test that wasn't modified https://github.com/celestiaorg/celestia-app/actions/runs/4962401229/jobs/8880507945#step:6:19

app/test/std_sdk_test.go
```diff
@@ -163,7 +163,16 @@ func (s *StandardSDKIntegrationTestSuite) TestStandardSDK() {
...
+				// HACKHACK this test implicitly depends on other test cases
+				// being run before it because it attempts to spend tokens from
+				// the community pool. Tokens get minted to the community pool
+				// due to inflation so this test may fail if executed in
+				// isolation or the number of tokens that are included in the
+				// spend proposal exceeds the number of tokens minted to the
+				// community pool. To avoid depending on the number of coins
+				// present in the community pool, the spend proposal attempts to
+				// spend 0 coins.
+				coins := sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(0)))
```
cmwaters, line 175: lol :)

rootulp, line 175: @evan-forbes would appreciate your opinion on this

x/mint/types/constants.go
```diff
@@ -0,0 +1,24 @@
...
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+const (
+	SecondsPerMinute = 60
+	MinutesPerHour   = 60
+	HoursPerDay      = 24
+	// DaysPerYear is the mean length of the Gregorian calendar year. Note this
+	// value isn't 365 because 97 out of 400 years are leap years. See
+	// https://en.wikipedia.org/wiki/Year
```
cmwaters, line 11: https://www.britannica.com/story/ten-days-that-vanished-the-switch-to-the-gregorian-calendar If you're curious about how the transition went

x/mint/types/minter.go
```diff
@@ -0,0 +1,76 @@
...
+
+// CalculateBlockProvision returns the total number of coins that should be
+// minted due to inflation for the current block.
+func (m Minter) CalculateBlockProvision(current time.Time, previous time.Time) sdk.Coin {
+	timeElapsed := current.Sub(previous).Seconds()
+	portionOfYear := sdk.NewDec(int64(timeElapsed)).Quo(sdk.NewDec(int64(SecondsPerYear)))
+	blockProvision := m.AnnualProvisions.Mul(portionOfYear)
+	return sdk.NewCoin(m.BondDenom, blockProvision.TruncateInt())
+}
```
cmwaters, line 67: Do you know what the error of this function is? i.e. the difference between the actual `AnnualProvision` and multiplying this over x many blocks with ranging intervals around 15 seconds.

cmwaters (approved):
Everything seems to look good to me

x/mint/types/minter.go
rootulp: it looks pretty close:

```go
func TestCalculateBlockProvisionError(t *testing.T) {
	minter := DefaultMinter()
	current := time.Date(2023, 1, 2, 0, 0, 0, 0, time.UTC)
	end := time.Date(2024, 1, 2, 0, 0, 0, 0, time.UTC)

	totalSupply := sdk.NewDec(1_000_000_000_000)             // 1 trillion utia
	annualProvisions := totalSupply.Mul(initalInflationRate) // 80 billion utia

	minter.AnnualProvisions = annualProvisions
	totalBlockProvisions := sdk.NewDec(0)
	for current.Before(end) {
		blockInterval := randomBlockInterval()
		previous := current
		current = current.Add(blockInterval)
		got := minter.CalculateBlockProvision(current, previous)
		totalBlockProvisions = totalBlockProvisions.Add(sdk.NewDecFromInt(got.Amount))
	}
	fmt.Printf("totalBlockProvisions %v\n", totalBlockProvisions)
	fmt.Printf("annualProvisions %v\n", annualProvisions)
}

func randomBlockInterval() time.Duration {
	return time.Duration(randInRange(14, 16)) * time.Second
}

// randInRange returns a random number in the range (min, max) inclusive.
func randInRange(min int, max int) int {
	return rand.Intn(max-min) + min
}
```

gives the output

```
totalBlockProvisions 79945957690.000000000000000000
annualProvisions 80000000000.000000000000000000
```

cmwaters: Ok one last thing. Can you try the same but instead of 

```go
func randomBlockInterval() time.Duration {
	return time.Duration(randInRange(14, 16)) * time.Second
}
```
use
```
func randomBlockInterval() time.Duration {
	return time.Duration(randInRange(14000, 16000)) * time.Millisecond
}
```

app/test/std_sdk_test.go
evan-forbes: :eyes: woah weird! 

I believe this analysis is partly correct, but I think its actually due to how the mint module is calculating how many funds get minted.

because we round down and the blocks are fast, the mint module will never actually mint anything, no matter how long we wait. One would expect this test to pass if we wait for long enough and only send a single utia, but that's not the case. However, if we set the set the block time to 1s and the spend amount to 1utia then the test passes!

I suggest we either:
 - move this test to the mint module integration test
 - set the block time to 1s with a comment + lower spend amount to 1utia
 - keep fast block time and set amount to 0 like above, but also test in the mint module integration test
 - change the way we are calculating inflation per block to not round down nearly as much (see comment below, also this is my preferred option)

x/mint/types/minter.go
```diff
@@ -0,0 +1,76 @@
...
+// minted due to inflation for the current year.
+func (m Minter) CalculateAnnualProvisions(totalSupply math.Int) sdk.Dec {
+	return m.InflationRate.MulInt(totalSupply)
+}
+
+// CalculateBlockProvision returns the total number of coins that should be
+// minted due to inflation for the current block.
+func (m Minter) CalculateBlockProvision(current time.Time, previous time.Time) sdk.Coin {
+	timeElapsed := current.Sub(previous).Seconds()
+	portionOfYear := sdk.NewDec(int64(timeElapsed)).Quo(sdk.NewDec(int64(SecondsPerYear)))
```
evan-forbes, line 64: note that time elapsed is a `float64` and when we convert it to an int64 here, we're round down significantly based on the actual amount of inflation. If a block is 15.9 seconds, we're only giving out 15s worth of inflation, and thus stiffing the community their divine remuneration (:trollface:). Since we have so many blocks per, this adds up pretty quick

after 40 3 second blocks the community pool balance is:

| Using Time Unit       | Value                           |
|-----------------------|---------------------------------|
| seconds        | 143456203394.140000000000000000utia |
| milliseconds   | 173977278048.860000000000000000utia |
| nanoseconds | 174005731475.800000000000000000utia |


We already have nano second precision from the header time, so I actually think we should use that. This way, we get almost exactly the desired amount of inflation per year.

x/mint/types/minter.go
```diff
@@ -0,0 +1,76 @@
...
+	return sdk.NewCoin(m.BondDenom, blockProvision.TruncateInt())
+}
+
+// yearsSinceGenesis returns the number of years that have passed between
+// genesis and current (rounded down).
+func yearsSinceGenesis(genesis time.Time, current time.Time) (years uint64) {
+	if current.Before(genesis) {
+		return 0
+	}
+	return uint64(current.Sub(genesis).Seconds() / float64(SecondsPerYear))
```
evan-forbes, line 75: we could also switch this to nano seconds, although it will certainly have less of an impact

evan-forbes (commented):
okay, took a dive and found some interesting stuff, but overall I think this is almost ready to go! have a question and a quick discussion then I think we're ready to go.

I'm _definitely_ fine with merging and iterating. 

you just coded some really important logic! :nerd_face:

x/mint/types/minter.go
rootulp:
```
totalBlockProvisions 77283450905.000000000000000000
annualProvisions 80000000000.000000000000000000
```

x/mint/types/minter.go
rootulp: Great catch! Resolved by using nanoseconds

rootulp:
> change the way we are calculating inflation per block to not round down nearly as much (see comment below, also this is my preferred option)

After doing this, the test passes for a proposal to spend 1utia 🎊

app/test/std_sdk_test.go
rootulp: > change the way we are calculating inflation per block to not round down nearly as much (see comment below, also this is my preferred option)

After doing this, the test passes for a proposal to spend 1utia 🎊

evan-forbes (dismissed):
nice work!! ship ittttt :shipit:

my only complaint is that we didn't get to an even 4200 lines, but I guess we can't get everything perfect 
Screenshot from 2023-05-16 11-39-52

x/mint/abci.go
```diff
@@ -0,0 +1,93 @@
...
+
+// maybeSetGenesisTime sets the genesis time if the current block height is 1.
+func maybeSetGenesisTime(ctx sdk.Context, k keeper.Keeper) {
+	if ctx.BlockHeight() == 1 {
+		genesisTime := ctx.BlockTime()
+		minter := k.GetMinter(ctx)
+		minter.GenesisTime = &genesisTime
+		k.SetMinter(ctx, minter)
+	}
+}
```
cmwaters, line 31: I have an alternative proposal to this which you can do in a follow up PR if you choose to. Rather than this, add a line in:

https://github.com/celestiaorg/celestia-app/blob/1e60076cd52de601b6e403bc783389d17faa46ef/app/app.go#L571-L578

Something like:

```go
app.MintKeeper.SetGenesisTime(req.Time)
```
Which will set the genesis time. This also avoids the problem of `PreviousBlockTime` being nil at height 1

cmwaters (approved):
Nice work on this 👍

evan-forbes (approved):
githaaaaaaa

SweeXordious (approved):

SweeXordious (approved):

PR #1584, opened 2023-03-31
docs: ADR 019 draft for inflation rate

mojtaba-esk:
## Checklist

### DIFF
--- a/docs/architecture/adr-019-inflation-rate.md
+++ b/docs/architecture/adr-019-inflation-rate.md
@@ -1,0 +1,113 @@
+# ADR 019: Inflation rate change
+
+## Status
+
+Proposed
+
+## Changelog
+
+- 2023/3/31: initial draft
+
+## Context
+
+The inflation rate is decided to have a fixed rate with an initial value and a target value.
+Table below shows that how it looks like according to the made decision.
+
+| Initial Inflation |     8.00% |
+|------------------------|----------------------------|
+| Disinflation rate p.a |    10.00% |
+| Target Inflation (floor)  |  1.50% |
+
+Once we approach the target inflation, we stick to it forever.
+Table below depicts the values throughout the coming years:
+
+| year | inflation (%) |
+|------|------|
+|0  | 8.00 |
+|1  | 7.20 |
+|2  | 6.48 |
+|3  | 5.832 |
+|4  | 5.2488 |
+|5  | 4.72392 |
+|6  | 4.251528 |
+|7  | 3.8263752 |
+|8  | 3.44373768 |
+|9  | 3.099363912 |
+|10 | 2.7894275208 |
+|11 | 2.51048476872 |
+|12 | 2.259436291848 |
+|13 | 2.0334926626632 |
+|14 | 1.83014339639688 |
+|15 | 1.647129056757192 |
+|16 | 1.50 |
+|17 | 1.50 |
+|18 | 1.50 |
+|19 | 1.50 |
+|20 | 1.50 |
+
+## Alternative Approaches
+
+In order to do so, we need to have a simple mint module. So we remove the dynamic inflation rate calculation and simply map the current block height to the respective year and calculate the inflation rate according to that.
+
+The logic is implemented in the `NextInflationRate` function and is called in the `BeginBlock` of the `mint` module.
+
+If no custom inflation calculator function is passed to the Minter, the default function calculates the inflation rate that this ADR desires to achieve.
+
+## Decision
+
+The decision is already made and the ADR is accepted. We are waiting for a final inspection though.
+
+## Detailed Design
+
+Since the rates are fixed, we hardcoded them in a set of constants.
+
+```go
+const (
+	InitialInflationRate    = 0.08
+	TargetInflationRate     = 0.015 // floor
+	DisinflationRatePerYear = 0.1
+)
+```
+
+The `NextInflationRate` function calculated the `year` based on the current block height that comes from the `sdk.Context` and the number of `BlocksPerYear` that comes from the params.
+Then it computes the inflation rate according to the determined fixed rate per year.
+
+```go
+NextInflationRate(ctx sdk.Context, params Params) sdk.Dec {
+
+	year := uint64(ctx.BlockHeader().Height) / params.BlocksPerYear
+
+	initInflationRate := sdk.NewDecWithPrec(InitialInflationRate*1000, 3 /* since we used 1000 */)
+	targetInflationRate := sdk.NewDecWithPrec(TargetInflationRate*1000, 3 /* since we used 1000 */)
+
+	// initInflationRate * ((1 - DisinflationRate) ^ year)
+	newInflationRate := initInflationRate.Mul(
+		sdk.OneDec().Sub(
+			sdk.NewDecWithPrec(DisinflationRatePerYear*100, 2 /* since we used 100 */)).
+			Power(year))
+
+	if newInflationRate.LT(targetInflationRate) {
+		newInflationRate = targetInflationRate
+	} else {
+		if newInflationRate.GT(initInflationRate) {
+			newInflationRate = initInflationRate
+		}
+	}
+
+	return newInflationRate
+}
+```
+
+## Consequences
+
+> This section describes the consequences, after applying the decision. All consequences should be summarized here, not just the "positive" ones.
+
+### Positive
+
+### Negative
+
+### Neutral
+
+## References
+
+- {reference link}


docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,113 @@
+# ADR 019: Inflation rate change
```
rootulp, line 1: [nit] if possible can the PR title == the filename == the header in the file

```suggestion
# ADR 019: Inflation rate
```

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Context
+
+The inflation rate is decided to have a fixed rate with an initial value and a target value.
+Table below shows that how it looks like according to the made decision.
+
+| Initial Inflation |     8.00% |
+|------------------------|----------------------------|
+| Disinflation rate p.a |    10.00% |
+| Target Inflation (floor)  |  1.50% |
```
rootulp, line 19:
```suggestion
| Name                     | Value  |
|--------------------------|--------|
| Initial Inflation        | 8.00%  |
| Disinflation rate p.a    | 10.00% |
| Target Inflation (floor) | 1.50%  |
```

[optional] https://marketplace.visualstudio.com/items?itemName=darkriszty.markdown-table-prettify
[proposal] to include a table header

rootulp (commented):
No blocking feedback

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,113 @@
...
+- 2023/3/31: initial draft
+
+## Context
+
+The inflation rate is decided to have a fixed rate with an initial value and a target value.
+Table below shows that how it looks like according to the made decision.
+
+| Initial Inflation |     8.00% |
+|------------------------|----------------------------|
+| Disinflation rate p.a |    10.00% |
```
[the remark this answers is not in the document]
mojtaba-esk: yes that means per year, sure I will change that

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,113 @@
...
+
+Proposed
+
+## Changelog
+
+- 2023/3/31: initial draft
+
+## Context
+
+The inflation rate is decided to have a fixed rate with an initial value and a target value.
```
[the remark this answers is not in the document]
mojtaba-esk: Not that I know. Maybe @evan-forbes has something to share.

docs/architecture/adr-019-inflation-rate.md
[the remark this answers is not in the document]
mojtaba-esk: Done

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,113 @@
...
+|13 | 2.0334926626632 |
+|14 | 1.83014339639688 |
+|15 | 1.647129056757192 |
+|16 | 1.50 |
+|17 | 1.50 |
+|18 | 1.50 |
+|19 | 1.50 |
+|20 | 1.50 |
+
+## Alternative Approaches
```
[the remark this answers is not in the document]
mojtaba-esk: Done

docs/architecture/adr-019-inflation-rate.md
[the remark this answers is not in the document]
liamsi: I do not think we need to go into detail here. We might add some rationale in a blog post though. Agreed @musalbas?

MSevey (approved):

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,112 @@
...
+
+> This section describes the consequences, after applying the decision. All consequences should be summarized here, not just the "positive" ones.
+
+### Positive
+
+### Negative
+
+### Neutral
+
+## References
```
evan-forbes, line 110: still todo or remove

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Detailed Design
+
+In order to do so, we need to have a simple mint module. So we remove the dynamic inflation rate calculation and simply map the current block height to the respective year and calculate the inflation rate according to that.
+
+The logic is implemented in the `NextInflationRate` function and is called in the `BeginBlock` of the `mint` module.
+
+If no custom inflation calculator function is passed to the Minter, the default function calculates the inflation rate that this ADR desires to achieve.
+
+Since the rates are fixed, we hardcoded them in a set of constants.
```
evan-forbes, line 61: could we add like one or two sentences talking about the flow of funds from a state perspective? ie funds are minted using the bank module to the mint module address etc

evan-forbes (commented):
almost there imo!! then we can :shipit:

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,112 @@
...
+	if newInflationRate.LT(targetInflationRate) {
+		newInflationRate = targetInflationRate
+	} else {
+		if newInflationRate.GT(initInflationRate) {
+			newInflationRate = initInflationRate
+		}
+	}
+
+	return newInflationRate
+}
```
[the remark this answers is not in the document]
musalbas: if it is time based would it be deterministic? what level of clock drift does tendermint tolerate?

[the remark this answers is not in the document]
evan-forbes: It would be deterministic if the time in the header is used, which is determined by [the median](https://github.com/celestiaorg/celestia-core/blob/ed981bbc97f8068b02b7fc52a05c4ae550b8cc20/state/state.go#L244-L250) of each time provided by the validators in the last commit.

> what level of clock drift does tendermint tolerate?

I could be missing something, but it looks like the only [check](https://github.com/celestiaorg/celestia-core/blob/ed981bbc97f8068b02b7fc52a05c4ae550b8cc20/state/validation.go#L116-L129) is that the time is increasing, and that it actually matches the median time.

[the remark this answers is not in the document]
liamsi: It might be simpler to reason about a block-based instead of time-based schedule. Then it should be communicated as that: every X blocks (which roughly corresponds to one year) the disinflation param will be applied. 

That said, I still wonder if the current flow could not be simplified. Roughly to: if a year has passed (in blocks/or via timestamp), calculate and return next inflation, else simply return current inflation.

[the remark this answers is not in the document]
musalbas: I do think time based might be better, because block interval is not a consensus critical parameter, so block producers could print more money by reducing their block interval.

It would also be possible for validators to manipulate the time, but I think would be harder given it's based on median time. Plus if they change the time to the future that would be non-ambitiously malicious. Whereas if a small minority of validators reduce their block interval, that could have an impact on emissions.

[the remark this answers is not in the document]
evan-forbes: I also agree that time seems to be the best approach here. It seems like it could help simplify the approach because we don't have to think about historic expected blocks per year, and also this is what osmosis does to schedule/[trigger epochs.](https://github.com/osmosis-labs/osmosis/blob/087a97b2976b989f94ab52000a64377c0bcf1a79/x/epochs/keeper/abci.go#L19-L22)

> It would also be possible for validators to manipulate the time, but I think would be harder given it's based on median time. Plus if they change the time to the future that would be non-ambitiously malicious. 

+1, after digging a bit deeper, at least offchain [light client verification](https://github.com/celestiaorg/celestia-core/blob/ed981bbc97f8068b02b7fc52a05c4ae550b8cc20/light/verifier.go#L176-L181) would fail

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,112 @@
...
+
+	if newInflationRate.LT(targetInflationRate) {
+		newInflationRate = targetInflationRate
+	} else {
+		if newInflationRate.GT(initInflationRate) {
+			newInflationRate = initInflationRate
+		}
+	}
+
+	return newInflationRate
```
evan-forbes, line 96: we should update this before merging, remove entirely, or ensure we do so when we have the actual implemenation #1578. 

Also, while we could calculate the inflation rate with each block, we could also store it in the state with very restricted access, and change it once per year.

docs/architecture/adr-019-inflation-rate.md
evan-forbes: bump

evan-forbes (commented):
we should update to reflect the decision to use the header time as well as the below comments

docs/architecture/adr-019-inflation-rate.md
rootulp: Yes added a `// TODO`

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,99 @@
...
+
+A: Likely calculate an estimate for the # of `BlocksPerYear` in BeginBlock by looking at the # of blocks in the year so far or the # of seconds between the last block and the current block.
+
+Q: Do any other Cosmos chains use a time-based inflation schedule?
+
+// TODO
+
+Q: What is the skew between `BlocksPerYear` on popular Cosmos SDK chains and the actual number of blocks per year?
+
+// TODO
```
rootulp, line 95: @evan-forbes feel free to add any more questions here

docs/architecture/adr-019-inflation-rate.md
```diff
@@ -0,0 +1,112 @@
...
+
+```go
+const (
+	InitialInflationRate    = 0.08
+	TargetInflationRate     = 0.015 // floor
+	DisinflationRatePerYear = 0.1
+)
+```
+
+The `NextInflationRate` function calculates the `year` based on the current block height that comes from the `sdk.Context` and the number of `BlocksPerYear` that comes from the params.
```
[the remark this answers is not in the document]
rootulp: Resolving b/c no longer using `BlocksPerYear`

rootulp:
Pinging @evan-forbes @cmwaters @liamsi for review

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Detailed Design
+
+1. Remove all parameters from the module
```
cmwaters, line 66: Can we extend this to specify that the inflation model is fixed and could only be changed by a major upgrade

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,115 @@
...
... to make a one significant modification to the existing `x/mint` module. Instead of relying on a block height (i.e. `header.height`) based inflation schedule, we are considering a time-based (i.e. `header.time`) inflation schedule. Using a time-based schedule has one key advantage: since the block interval isn't fixed, individual validators are able to produce blocks at a rate (i.e. one block every 10 seconds) that differs from the network target (i.e. one block every 15 seconds). An inflation schedule based on block height needs to make an assumption on the number of blocks created per year so a large discrepancy between the expected and actual block interval will result in a discrepancy between target and actual inflation rate. A time-based schedule is potentially more robust against this attack because "the timestamp is equal to the weighted median of validators present in the last commit." [cometbft/specs](https://github.com/cometbft/cometbft/blob/c58597d656d5c816334aff9ea8e600bdbc534817/spec/core/data_structures.md?plain=1#L127). On the other hand, a time-based inflation schedule isn't as battle-tested as the Cosmos SDK `x/mint` module and may expose additional attack vectors.
```
cmwaters, line 62: The weighted median timestamp has the same 1/3+ security model as the rest of Tendermint I.e. it takes 1/3+ to arbitrarily modify the time (monotonically increasing).

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Detailed Design
+
+1. Remove all parameters from the module
+1. Calculate inflation rate based on the number of years since genesis
```
cmwaters, line 67: As well as calculate the provision for that year

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,115 @@
...
+1. Calculate block provision based on the number of seconds elapsed between the current block and the previous block
```
cmwaters, line 68: To be exact, nanoseconds. Also we multiply this by the amount of new coins per nanosecond given the hard codes amount of nanoseconds per year.

Just as a meta comment, I think it's okay to be a little more detailed here because readers will likely not have anywhere the same context as we do

cmwaters (commented):
I like the FAQ at the end :) - just a few wording comments

docs/architecture/adr-019-strict-inflation-schedule.md
liamsi, line 62: There is a typo here (make a one significant change)

liamsi (dismissed):
Excellent ADR 🙏

docs/architecture/adr-019-strict-inflation-schedule.md
rootulp: b4c2322334c5251a4ccf73824afd0e09f37fa02e

rootulp: Resolved by 8f7c5e9ea4e75ecce3c01d607d2fda744706fadf

docs/architecture/adr-019-strict-inflation-schedule.md
rootulp: Resolved by b1db8bdffa57e9ca0ce8bb9b408992b6402a6c34

docs/architecture/adr-019-strict-inflation-schedule.md
rootulp: Resolved by b1db8bdffa57e9ca0ce8bb9b408992b6402a6c34

cmwaters (approved):
Nice write up 👍

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,126 @@
...
+
+An inflation schedule based on block height needs to make an assumption on the number of blocks created per year so a large discrepancy between the expected and actual block interval will result in a discrepancy between target and actual inflation rate.
+
+A time-based inflation schedule depends on block timestamps and "the timestamp is equal to the weighted median of validators present in the last commit." [cometbft/specs](https://github.com/cometbft/cometbft/blob/c58597d656d5c816334aff9ea8e600bdbc534817/spec/core/data_structures.md?plain=1#L127). In other the weighted median timestamp has the same 1/3+ security model as the rest of Tendermint and it takes 1/3+ to arbitrarily modify the time (monotonically increasing).
```
evan-forbes, line 66: we could also mention that offchain light clients reject blocks that are too far in the future, providing strong guarantees around this value.

docs/architecture/adr-019-strict-inflation-schedule.md
```diff
@@ -0,0 +1,126 @@
...
+
+Note: a time-based inflation schedule implemented isn't as battle-tested as the Cosmos SDK `x/mint` module and may expose additional attack vectors.
```
evan-forbes, line 68: could we maybe clarify the thing that is making it less tested? the reason I mention this is because using the time in the header has been tested extensively in the standard sdk and in osmosis, but testing this strict inflation schedule has not. 

Like, does using blocks per year estimates vs using time have any more or less risks?

MSevey (approved):

evan-forbes (commented):
just two non-blocking comments, glad we wrote this down!!

docs/architecture/adr-019-strict-inflation-schedule.md
rootulp: > could we maybe clarify the thing that is making it less tested? 

Discussed offline and I think the primary concern is implementation risk.

docs/architecture/adr-019-strict-inflation-schedule.md
rootulp: Discussed offline and decided to not do this.

PR #1585, opened 2023-03-31
feat: Introduce `--namespace-version` CLI flag

### LINKED ISSUE #1528: Introduce `--namespace-version` CLI flag
Similar to https://github.com/celestiaorg/celestia-app/issues/1041 but for the namespace version introduced as part of https://github.com/celestiaorg/celestia-app/issues/1434

amityadav0:
## Overview

Introduce --namespace-version CLI flag 
- Closes #1528 

I have introduced namespace flag. Currently this supports namespace version zero only.
I think data structure for storing namespace versions can be updated. Maybe a map with version details.
For example:
```
Key: Version[0], Value: { NamespaceVersionPrefixSize: ... , NamespaceVersionIDSize: ... }
Key: Version[1], Value: { NamespaceVersionPrefixSize: ... , NamespaceVersionIDSize: ... }
```
instead of constants
```
	NamespaceVersionZero = uint8(0)
	NamespaceVersionZeroPrefixSize = 22
	NamespaceVersionZeroIDSize = NamespaceIDSize - NamespaceVersionZeroPrefixSize
```

Then it would be possible to switch versions easily from cli. same for share version.

Please let me know if this makes sense or am I missing something.


x/blob/client/cli/payforblob.go
```diff
@@ -57,6 +75,7 @@ func CmdPayForBlob() *cobra.Command {
 	}
 
 	flags.AddTxFlagsToCmd(cmd)
+	cmd.PersistentFlags().String("namespace-version", "", "User can use this to specify the namespace version when they submit a pay for blob.")
```
rootulp, line 78: once we extract a const for the flag name, we can use it here

rootulp (commented):
Sorry for the delay in reviewing @amityadav0 but thanks for working on this!

x/blob/client/cli/payforblob.go
amityadav0: Updated

x/blob/client/cli/payforblob.go
```diff
@@ -34,11 +35,28 @@ func CmdPayForBlob() *cobra.Command {
 				return fmt.Errorf("failure to decode hex namespace ID: %w", err)
 			}
 
-			// TODO: allow the user to override the namespace version via a new flag
-			// See https://github.com/celestiaorg/celestia-app/issues/1528
-			namespace, err := appns.New(appns.NamespaceVersionZero, append(appns.NamespaceVersionZeroPrefix, namespaceID...))
-			if err != nil {
-				return fmt.Errorf("failure to create namespace: %w", err)
+			nameSpaceFlag, _ := cmd.Flags().GetString("--namespace-version")
```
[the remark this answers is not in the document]
amityadav0: Thanks for the review.
Done, I have updated the PR.

amityadav0:
Thanks for the review @rootulp. I have updated the PR.

rootulp:
Thanks for updating the PR! This overall LGTM. I had two optional suggestions that I applied on top of this PR. I made the changes myself b/c I forgot to include them in the previous round of PR feedback and I didn't want to block the PR on addressing them.

rootulp (approved):

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +60,24 @@ func CmdPayForBlob() *cobra.Command {
 	}
 
 	flags.AddTxFlagsToCmd(cmd)
+	cmd.PersistentFlags().String(FlagNamespaceVersion, "0", "Specify the namespace version (default is 0)")
```
evan-forbes, line 63: why use a string here? if we use a int, would it be easier to convert to uint8?

x/blob/client/testutil/integration_test.go
```diff
@@ -62,8 +64,8 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	require := s.Require()
 	val := s.network.Validators[0]
 	hexNamespace := hex.EncodeToString(appns.RandomBlobNamespaceID())
+	invalidNamespaceID := hex.EncodeToString(bytes.Repeat([]byte{0}, 8)) // invalid because ID is expected to be 10 bytes
```
evan-forbes, line 67: > // invalid because ID is expected to be 10 bytes

@rootulp is this comment true?

x/blob/client/cli/payforblob.go
evan-forbes, line 63: we shouldn't write defaults, since they're filled in for us by cobra
```suggestion
	cmd.PersistentFlags().String(FlagNamespaceVersion, "0", "Specify the namespace version")
```

evan-forbes (changes requested):
Left a few minor blocking comments, thanks @amityadav0 !!

x/blob/client/cli/payforblob.go
amityadav0: yes, changed to uint8. Thanks

x/blob/client/testutil/integration_test.go
rootulp: Yep! As a convenience, the CLI accepts a 10 byte namespace ID b/c that's the only usable portion for namespace version `0`.

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +59,20 @@ func CmdPayForBlob() *cobra.Command {
 	}
 
 	flags.AddTxFlagsToCmd(cmd)
+	cmd.PersistentFlags().Uint8(FlagNamespaceVersion, 0, "Specify the namespace version")
 
 	return cmd
 }
 
+func getNamespace(namespaceID []byte, namespaceVersion uint8) (appns.Namespace, error) {
+	switch uint8(namespaceVersion) {
```
rootulp, line 68: this type cast is no longer necessary

```suggestion
	switch namespaceVersion {
```

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +59,20 @@ func CmdPayForBlob() *cobra.Command {
...
+	case appns.NamespaceVersionZero:
+		return appns.New(uint8(namespaceVersion), append(appns.NamespaceVersionZeroPrefix, namespaceID...))
```
rootulp, line 70: this type cast is no longer necessary

```suggestion
		return appns.New(namespaceVersion, append(appns.NamespaceVersionZeroPrefix, namespaceID...))
```

rootulp (changes requested):
2 small nits based on the last change to `uint8` flag

x/blob/client/cli/payforblob.go
rootulp: @evan-forbes where are they filled in? After this suggestion the help doc looks like:

```
$ ./build/celestia-appd tx blob PayForBlobs --help
...
      --namespace-version uint8   Specify the namespace version
```

rootulp (dismissed):

x/blob/client/cli/payforblob.go
evan-forbes: hmmm I'm not sure that's a good question. the rest get filled in for example https://github.com/celestiaorg/cosmos-sdk/blob/f19e20459b8bcf195167ad256de5a6ca6edc41f5/server/start.go#L159-L161

```
--address string                                  Listen address (default "tcp://0.0.0.0:26658")
--home string                                     The application home directory (default "/home/evan/.celestia-app")
```

x/blob/client/cli/payforblob.go
```diff
@@ -63,11 +65,20 @@ func CmdPayForBlob() *cobra.Command {
...
+	cmd.PersistentFlags().Uint8(FlagNamespaceVersion, 0, "Specify the namespace version")
 	cmd.PersistentFlags().Uint8(FlagShareVersion, 0, "Specify the share version")
-
 	return cmd
 }
 
+func getNamespace(namespaceID []byte, namespaceVersion uint8) (appns.Namespace, error) {
+	switch namespaceVersion {
+	case appns.NamespaceVersionZero:
+		return appns.New(namespaceVersion, append(appns.NamespaceVersionZeroPrefix, namespaceID...))
```
evan-forbes, line 76: we should never append directly to a global, as it could cause tons of bugs

we need to first make an empty slice of appropriate capacity, and then append

evan-forbes (changes requested):
Missed this first time around! sorry about that @amityadav0 thanks again for your patience

x/blob/client/cli/payforblob.go
amityadav0: I don't think we are appending to global value. Append function will return a value here which is passed to `appns.New()`
z = append(x, y)
This line won't change x or y.

rootulp: My takeaway after reading that article is that 

```go
z = append(x, y)
```

is safe because the length or capacity of x will not be impacted by the above statement. After execution the global `x` remains unmodified. My understanding is that the copy is only necessary if a function is invoked on the global `x` that modifies elements of the original slice. If `append` is dangerous in this usage, I would expect there to be a linter to warn us about incorrect usage but I couldn't find one on https://golangci-lint.run/usage/linters/

@evan-forbes how do we enforce that `append` isn't misused? Additionally are there examples of just the `append` causing bugs:

> we should never append directly to a global, as it could cause tons of bugs

evan-forbes: I think we have some linter, but otherwise this is just good practice imo since it always fixes the issue and eliminates the reader having to think "is this an issue". Also, I still think its an issue but don't have the brain _capacity_ at the moment to figure it out lol :laughing: 

examples https://github.com/celestiaorg/celestia-core/issues/336, https://github.com/celestiaorg/celestia-core/issues/250, https://github.com/celestiaorg/nmt/pull/30, there was also one in node that fixed during the barcelona onsite but I couldn't find it in a cursory search

**all** were caused by 
z = append(x, y)

so now we should never ever ever do that no matter what because evan is too scarred

x/blob/client/cli/payforblob.go
```diff
@@ -63,11 +65,21 @@ func CmdPayForBlob() *cobra.Command {
...
 	cmd.PersistentFlags().Uint8(FlagShareVersion, 0, "Specify the share version")
-
 	return cmd
 }
 
+func getNamespace(namespaceID []byte, namespaceVersion uint8) (appns.Namespace, error) {
+	switch namespaceVersion {
+	case appns.NamespaceVersionZero:
+		namespaceVersionZeroPrefix := appns.NamespaceVersionZeroPrefix
+		return appns.New(namespaceVersion, append(namespaceVersionZeroPrefix, namespaceID...))
```
evan-forbes, line 77: this doesn't actually avoid the issue, since slices can be treated similar to pointers. Its true that we're passing by value here, but we're passing a pointer by value, which unfortunatley still won't solve our problem

in order to avoid the underlying issue, we need to make an explicit copy of the slice
similar to 
```go
ns := make([]byte, 0, appconsts.NamespaceSize)
ns = append(ns, append(prefix, namespaceID)...)
```

amityadav0: I guess now I get it.
with above example prefix may still be changed
```
nsp := make([]byte, 0, appns.NamespaceVersionZeroPrefixSize)
nsp = append(nsp, appns.NamespaceVersionZeroPrefix...)
return appns.New(namespaceVersion, append(nsp, namespaceID...))
```
I have updated the code to this

evan-forbes (dismissed):
made super minor change to push this over the line

thanks again @amityadav0 !! apologies this took so long

x/blob/client/cli/payforblob.go
```diff
@@ -37,12 +43,10 @@ func CmdPayForBlob() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("failure to decode hex namespace ID: %w", err)
 			}
-
-			// TODO: allow the user to override the namespace version via a new flag
-			// See https://github.com/celestiaorg/celestia-app/issues/1528
-			namespace, err := appns.New(appns.NamespaceVersionZero, append(appns.NamespaceVersionZeroPrefix, namespaceID...))
+			namespaceVersion, _ := cmd.Flags().GetUint8(FlagNamespaceVersion)
```
rootulp, line 46: [question] why not check for `err` here?

x/blob/client/cli/payforblob.go
```diff
@@ -63,11 +67,22 @@ func CmdPayForBlob() *cobra.Command {
...
+	cmd.PersistentFlags().Uint8(FlagNamespaceVersion, 0, "Specify the namespace version")
 	cmd.PersistentFlags().Uint8(FlagShareVersion, 0, "Specify the share version")
-
 	return cmd
 }
 
+func getNamespace(namespaceID []byte, namespaceVersion uint8) (appns.Namespace, error) {
+	switch namespaceVersion {
+	case appns.NamespaceVersionZero:
+		nsp := make([]byte, 0, appns.NamespaceSize)
```
rootulp, line 78: [nit] 
```suggestion
		nsp := make([]byte, 0, appns.NamespaceIDSize)
```

rootulp (approved):

x/blob/client/cli/payforblob.go
evan-forbes: why not namespace size here? aren't we using the full 33?

rootulp: The ID is 32 bytes. The version is 1 byte. The total namespace is 33 bytes. 

Since this variable is just the ID, only 32 bytes are necessary.

evan-forbes (approved):

x/blob/client/cli/payforblob.go
evan-forbes: merged to clear out his PR, but we should still fix this imo if need be,

I'm confused a bit

so we should not be appending the version at all here since its added when we call Bytes()? therefore we're adding it twice?

Issue #1256, opened 2023-01-18
EPIC: Post feature freeze testing


evan-forbes:
We have been pushing many breaking changes lately to fit them in before feature freeze at the end of the month. After feature freeze, we will want to spend more time building out robust tests for mainnet. This includes but is not limited to:

- [x] #1102 
- [ ] #832 
- [ ] #829 
- [x] #695
- [x] #1249
- [x] https://github.com/celestiaorg/celestia-app/issues/313
- [ ] #1324
- [x] https://github.com/celestiaorg/celestia-core/issues/939
- [x] https://github.com/celestiaorg/celestia-app/issues/1516
- [x] https://github.com/celestiaorg/celestia-app/issues/1541

### Stretch
- [x] #1258
- [x] #233
- [x] #513
- [ ] https://github.com/celestiaorg/celestia-core/issues/945
- [x] #630
- [x] https://github.com/celestiaorg/celestia-app/issues/1343
- [ ] https://github.com/celestiaorg/celestia-app/issues/1556

MSevey:
Something that might be good to add to the stretch goals is thinking about a framework for testing version upgrades and compatibility. 

This could start by just ensuring we have the ability to launch nodes on different versions, and upgrade or downgrade a node without restarting the test.

evan-forbes:
yes! good idea @MSevey, I totally forgot about #313, added to stretch goals

MSevey:
nice yea, honestly i think that test #313  might make sense to bump up to a required and tie it into release. Basically launch the tests with half the nodes on the current path version and half on the new patch version as a sanity check that nothing breaks. 

Then a future test could be more involved in terms of more patch versions, minor versions, and major versions to really stress test the network and understand compatibility issues.  This one would of course be more of a debugging test. There might be some specific upgrade scenarios that we would want to build into the pipeline, but this and those would definitely be stretchy stretch goals.

Bidon15:
Thanks for the detail overview of the epic. It's Epic 😄 
I would appreciate if we take our time to discuss how we approach this epic both from app and devops/testing teams 

We can break down for example #1258 or compact blocks https://github.com/celestiaorg/celestia-core/issues/883 to decide who does what so at the end the tests bring the most value and insights to core/app team 
As well as we can learn how we can collaborate better for future joint issues

cmwaters:
[...]

PR #1586, opened 2023-03-31
feat: minimal e2e framework

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/issues/1256

This PR puts together the base pieces for an e2e testing suite. It includes a CLI that can:

- Setup the file directories for multiple nodes, including generating a genesis with several accounts
- Start the testnet (with different versions and at different start heights)
- Stop the network and cleanup the used resources.

Further work aims to support:
- compatibility testing
- upgrade testing
- sync testing
- fuzz testing
   - for non-determinism
   - invariant checking
   - as a generalized tool for other teams


testing/e2e/pkg/setup.go
```diff
@@ -0,0 +1,351 @@
...
+	(privval.NewFilePV(node.SignerKey,
+		filepath.Join(nodeDir, "config", "priv_validator_key.json"),
+		filepath.Join(nodeDir, "data", "priv_validator_state.json"),
+	)).Save()
+
+	return nil
+}
+
+func WriteDockerCompose(testnet *Testnet, file string) error {
+	// Must use version 2 Docker Compose format, to support IPv6.
```
evan-forbes, line 284: aren't we using IPv4 in the compose files?

testing/e2e/pkg/setup.go
```diff
@@ -0,0 +1,351 @@
...
+services:
+{{- range .Nodes }}
+  {{ .Name }}:
+    labels:
+      e2e: true
+    container_name: {{ .Name }}
+    image: ghcr.io/celestiaorg/celestia-app:{{ index .Versions 0 }}
+    entrypoint: ["/bin/celestia-appd"]
+    command: ["start"]
+    init: true
```
evan-forbes, line 306: my local docker setup might be wonk, but I had to add 
`    user: root` in order to get the test to work. If not, the test will appear to be waiting, but none of the nodes actually start because they do not have permission to write to the volume.

evan-forbes (commented):
this is dope! I appreciate the similarities to the tendermint e2e test, and see how we can add the desired tests. 

Are we wanting to eventually enable celestia-node to use this as well? I could see them being able to reuse a lot of the code here.

mainly just a had a few questions, but other than that I think it LGTM, and we should merge and start iterating

testing/e2e/cmd/e2e/main.go
```diff
@@ -0,0 +1,124 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			if err := e2e.Cleanup(ctx, cli.testnet); err != nil {
+				return fmt.Errorf("preparing testnet: %w", err)
+			}
+
+			if err := e2e.Setup(ctx, cli.testnet); err != nil {
+				return fmt.Errorf("setting up testnet: %w", err)
+			}
+			// defer e2e.Cleanup(ctx, cli.testnet)
```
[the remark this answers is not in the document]
cmwaters: Yup. I should add this back.

testing/e2e/pkg/setup.go
cmwaters: We're just using IPv4 atm. I don't know if you want to test with IPv6 as well?

cmwaters:
> Are we wanting to eventually enable celestia-node to use this as well? I could see them being able to reuse a lot of the code here.

Yeah Potentially. Matt mentioned having a common repository for e2e testing of various node types in a system. If we are to do this I'd definitely recommend reusing parts of this code.

testing/e2e/pkg/exec.go
```diff
@@ -0,0 +1,33 @@
...
+		return fmt.Errorf("failed to run %q:\n%v", args, string(out))
+	default:
+		return err
+	}
+}
+
+// execCompose runs a Docker Compose command for a testnet.
+func execCompose(dir string, args ...string) error {
+	return exec(append(
+		[]string{"docker-compose", "-f", filepath.Join(dir, "docker-compose.yml")},
```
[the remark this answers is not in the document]
cmwaters: Yeah probably better that way

testing/e2e/pkg/setup.go
evan-forbes: no we don't have to I was just confused by the comment, can resolve

evan-forbes (approved):
:+1: :shipit:

SweeXordious (commented):
Awesome stuff :tada: 

When I run this locally, the validators start correctly, but when I want to stop them using `ctrl + c`, everything hangs until I kill the terminal session.
When this happens, I check the running docker images and I see none, so probably this is not related to docker runtime.

Is that the correct way to stop the framework?
Is anyone having the same issue or it's just me?

MSevey:
cc @celestiaorg/testing for visibility

Bidon15:
Few questions from my side

### why we opted for in-house framework instead of using interchaintest? 

[state-sync example](https://github.com/strangelove-ventures/interchaintest/blob/main/examples/cosmos/state_sync_test.go)
[chain-upgrade example](https://github.com/strangelove-ventures/interchaintest/blob/main/examples/cosmos/chain_upgrade_test.go)

#### why not just re-use their implementation of chain creation using docker-compose and networking? 
https://github.com/strangelove-ventures/interchaintest/blob/main/chain/cosmos/cosmos_chain.go

It has the same capabilities as we have rn. 

Same with network pruning/setup 
https://github.com/strangelove-ventures/interchaintest/blob/main/internal/dockerutil/setup.go

imho, it looks like they have took care of docker networks cleanup which is clumsy


### if we want more control over implementation - why not reuse [compose-spec](https://github.com/libp2p/test-plans/pull/97) like libp2p does? 

Seeing the matrix versioning schema that libp2p does - we can def reuse the hard work they pushed. 


### Pros of above

Both have CI/CD implementation already meaning that we can speed up dev time and contribute back if something we really need

Bidon15:
My main concern over this PR - we are signing up on maintaining long-term, while there are tools that can help us achieve the same without maintaining it (interchaintest is constantly being updated, same is for libp2p/test-plan)
cc: @MSevey

cmwaters:
Hey,

> why we opted for in-house framework instead of using interchaintest?

The simple answer is we have a modified fork of the SDK which is not compatible with `interchaintest`. There are changes that I've spoken in the past that we can (and I think we should probably make) that means we don't have this compatibility issue and can eliminate the need to maintain the SDK fork. 

Secondly, interchaintest is very IBC orientated. This means some of the initial goals of upgrade testing (we will have a different upgrade mechanism to most cosmos chains), compatibility testing and sync testing don't seem to natively fit their framework.
 
Lastly, for an open source library, I think there documentation and tutorials around using their library are lacking (although the inline commenting is relatively helpful). As well as maintenance we have to also consider onboarding costs to understanding a new (and large) codebase (when something simpler might suffice). 

> My main concern over this PR - we are signing up on maintaining long-term

I'm sympathetic to this. Ideally we reach our testing goals with little duplication and overlap. When we spoke prior to this it seemed like testground was not suitable. Rene tells me that robusta uses a different more light-weight framework that may be ideal for the outlined purposes. I'm not sure about the other (libp2p) library you mention but I definitely think that what's done here is quite generic: It's simply managing a series of binaries within a network and using the RPC endpoints to make assertions. 

What I'm more concerned about is that it seems like these testing goals are perhaps the same across the other teams (node have their "swamp" testing) which seems, at least to me, to heavily overlap with what core/app is doing. If we have a "testing" team then perhaps efforts from individual teams can be consolidated

testing/e2e/pkg/rpc.go
```diff
@@ -0,0 +1,130 @@
...
+			continue
+		}
+
+		heights = append(heights, result.SyncInfo.LatestBlockHeight)
+	}
+	if len(heights) == 0 {
+		return nil, errors.New("network is not running")
+	}
+
+	// return heights in ascending order
```
rootulp, line 124: [nit] this line claims to sort heights in ascending order but the function godoc comment claims to return heights in descending order.

```
// GetHeights loops through all running nodes and returns an array of heights
// in order of highest to lowest.
func GetHeights(ctx context.Context, testnet *Testnet) ([]int64, error) {
```

rootulp (commented):
Overall LGTM. I'm encountering an error while trying to run locally

```shell
$ ./build/e2e -f networks/simple.toml
Setting up network simple-53126
Spinning up testnet
Starting validator01 at height 0 on http://localhost:4202
Error: starting network: exec: "docker compose": executable file not found in $PATH
```

I have docker compose installed in my $PATH:

```shell
$ docker-compose --version
Docker Compose version v2.15.1
$ which docker-compose
/usr/local/bin/docker-compose
```

1. Should we list that installing docker compose is a prerequisite (potentially in README.md)?
2. Am I missing a different prerequisite step?

Bidon15:
There are 2 topics that got confused and we clarified them here. 

The original idea behind this PR is for teams to do nightly tests using minimal e2e framework. I thought that this is only for core/app and integration ones executed on every PR. 

Going forward, we separated into 2 topics: 

1. Nightly Tests. 
2. Integration Tests that are executed per PR.


### Nightly tests

@celestiaorg/devops will implement infra features for robusta-nightlies after @cmwaters shares the notion doc with us. 
Robusta-nightly is a good point as the telemetry is already there and a fuzzing client can be plugged in easily. 

### Integration tests

This PR should be used on Integration level per PR run. In addition, we need to migrate it to a standalone repo, so @celestiaorg/celestia-node and @rollkit can use to fulfil their testing strategies 

In addition, we will reuse the compose-spec that [libp2p/test-plan](https://github.com/libp2p/test-plans/pull/97) has to make it CI/CD ready out-of-the-box 

Thanks for the input and clarifications 🚀

testing/e2e/pkg/testnet.go
```diff
@@ -0,0 +1,315 @@
...
+	SelfDelegation int64
+}
+
+type Account struct {
+	Name   string
+	Tokens int64
+	Key    crypto.PrivKey
+}
+
+func LoadTestnet(manifest Manifest, file string) (*Testnet, error) {
```
staheri14, line 57: [nit-optional] I think `filePath` is closer to what this parameter means.
```suggestion
func LoadTestnet(manifest Manifest, filePath string) (*Testnet, error) {
```

testing/e2e/pkg/testnet.go
```diff
@@ -0,0 +1,315 @@
...
+	addr := fmt.Sprintf("%v:26656", n.IP.String())
+	if withID {
+		addr = fmt.Sprintf("%x@%v", n.NetworkKey.PubKey().Address().Bytes(), addr)
+	}
+	return addr
+}
+
+// Address returns an RPC endpoint address for the node.
+func (n Node) AddressRPC() string {
+	return fmt.Sprintf("%v:26657", n.IP.String())
```
staheri14, line 222: [nit-optional]: Introducing a constant variable to represent the port number `26657` would enhance the code's future maintainability and readability.

testing/e2e/pkg/testnet.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func (n Node) IsValidator() bool {
+	return n.SelfDelegation != 0
```
staheri14, line 226: [optional] Since a non-zero `SelfDelegation` is the only requirement for being recognized as a validator, why not use a boolean type for `SelfDelegation` instead?

staheri14 (approved):
Nice work on this PR. 👍

testing/e2e/pkg/testnet.go
cmwaters: Because we may want to set the self delegation to an actual number which will reflect the initial voting power of the validator

evan-forbes (dismissed):

evan-forbes:
morge?

Bidon15 (approved):

cmwaters:
> I have docker compose installed in my $PATH:

Sorry for missing this @rootulp. `docker compose` isn't the same binary as `docker-compose`, it's a command from the `docker` binary that was recently added. If you haven't got it then I think you need to upgrade your docker version

testing/e2e/pkg/rpc.go
cmwaters: Good catch

evan-forbes (approved):

rootulp (approved):

Issue #1587, opened 2023-03-31
Observations from bumping celestia-node


rootulp:
A couple of observations while bumping celestia-node to the latest celestia-app release

1. It's possible to construct a `Share` that doesn't adhere to the length invariant via https://github.com/celestiaorg/celestia-app/blob/d916dfe162fb00a6ef1d3062dafc61849bfb1061/pkg/shares/shares.go#L206

2. Since ParseShares accepts a Share type, we need external code to be able to create new Shares (i.e. export `newShare`) or change ParseShares to accept a byte slice https://github.com/celestiaorg/celestia-app/blob/d916dfe162fb00a6ef1d3062dafc61849bfb1061/pkg/shares/parse.go#L38

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/1590

PR #1590, opened 2023-03-31
feat!: max namespace version for parity namespace

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1587

Motivated by an effort to bump celestia-node's dependency on celestia-app. Will cut a new v0.14.x RC after this merges and also cherry-pick this to main. Can split this into multiple PRs if requested.

1. Export `NewShare`
2. Modify parity namespace to use the max namespace version so that the NMT feature `ignoreMaxNamespace` works as expected

Follow-ups:
- https://github.com/celestiaorg/celestia-app/issues/1610 (not blocking the upgrade so spinning out)


x/blob/types/payforblob_test.go
```diff
@@ -174,9 +175,9 @@ func TestValidateBasic(t *testing.T) {
 			wantErr: nil,
 		},
 		{
-			name:    "parity shares namespace id",
-			msg:     paritySharesMsg,
-			wantErr: ErrParitySharesNamespace,
+			name:        "parity shares namespace id",
+			msg:         paritySharesMsg,
+			wantErrBool: true,
```
evan-forbes, line 180: quick question: do you mind pointing out the change that forces this? thanks!

evan-forbes (approved):
LGTM :+1: 

only had a non-blocking question

pkg/namespace/consts.go
```diff
@@ -59,5 +60,8 @@ var (
 	TailPaddingNamespace = MustNewV0([]byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE})
 
 	// ParitySharesNamespace is the namespace reserved for erasure coded data.
-	ParitySharesNamespace = MustNewV0([]byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF})
+	ParitySharesNamespace = Namespace{
+		Version: math.MaxUint8,
+		ID:      bytes.Repeat([]byte{0xFF}, NamespaceIDSize),
+	}
```
cmwaters, line 66: This seems adjacent to the issue you've outlined and I'm wondering the motivation for this being the max version

rootulp: The motivation is that a parity namespace with version 0 isn't the largest possible 33 byte namespace. A parity namespace with max version is the largest possible 33 byte namespace so it is ignored by the [IgnoreMaxNamespace](https://github.com/celestiaorg/nmt/blob/89434babb8f93c49f87859de909ab63123abdcef/docs/nmt-lib.md?plain=1#L18) option. celestia-node assumes that each row root is an NMT root with this option enabled.

Additional Context in Slack: https://celestia-team.slack.com/archives/C02PFFCGMNW/p1680900937688359

x/blob/types/payforblob_test.go
rootulp: If the test case were instead:
```go
		{
			name: "parity shares namespace id",
			msg:  paritySharesMsg,
			// wantErrBool: true,
			wantErr: ErrParitySharesNamespace,
		},
```

we would encounter a test failure:

```
--- FAIL: TestValidateBasic (0.03s)
    --- FAIL: TestValidateBasic/parity_shares_namespace_id (0.00s)
        /Users/rootulp/git/rootulp/celestia/celestia-app/x/blob/types/payforblob_test.go:243: 
            	Error Trace:	/Users/rootulp/git/rootulp/celestia/celestia-app/x/blob/types/payforblob_test.go:243
            	Error:      	"unsupported namespace version 255" does not contain "cannot use parity shares namespace ID"
            	Test:       	TestValidateBasic/parity_shares_namespace_id
```

because this error is hit [here](https://github.com/rootulp/celestia-app/blob/c13409907a234fbd5e12231fb64edba140109af0/x/blob/types/payforblob.go#L105-L108) before the error [here](https://github.com/rootulp/celestia-app/blob/c13409907a234fbd5e12231fb64edba140109af0/x/blob/types/payforblob.go#L109-L112)

evan-forbes: certainly not blocking on this, but since this test only checks for inclusion of a specific error message, could we also either create an skderror for invalid namespace (would have the benefit of helping users see specific error codes), or change the test to accept `error` instead of `sdkerror.Error`, then use
```go
{
			name: "parity shares namespace id",
			msg:  paritySharesMsg,
			wantErr: errors.New("unsupported namespace version"),
},
```

evan-forbes (dismissed):

x/blob/types/payforblob_test.go
rootulp: Good idea, resolved via https://github.com/celestiaorg/celestia-app/pull/1590/commits/93f74344b7394173d264151628b5fe076c57c9a0

evan-forbes (approved):

pkg/namespace/consts.go
cmwaters: Yeah that makes sense. Was trying to reason if the parity share would have any need to change the version

x/blob/types/payforblob_test.go
```diff
@@ -176,7 +176,7 @@ func TestValidateBasic(t *testing.T) {
 		{
 			name:    "parity shares namespace id",
 			msg:     paritySharesMsg,
-			wantErr: ErrParitySharesNamespace,
```
cmwaters, line 179: Can this error still occur. Do we have tests to ensure it errors as expected?

cmwaters (approved):
Just some minor questions

x/blob/types/payforblob_test.go
rootulp: It can no longer occur b/c `ErrInvalidNamespace` will be hit before `ErrParitySharesNamespace` is hit. I choose not to remove b/c a check exists for it [here](https://github.com/celestiaorg/celestia-app/blob/e03617e561fb3c01e33bf03859f53a05645a469d/x/blob/types/payforblob.go#L141-L143)

Issue #1041, opened 2022-11-18
Introduce `--share-version` CLI flag


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/issues/936 introduces the share version to `MsgWirePayForBlob` however there is no way for a user to configure the share version when they construct a pay for blob transaction.

## Proposal

Introduce a CLI flag (e.g. `--share-version`) that a user can use to specify the share version when they submit a pay for blob. For improved usability, the flag should be optional and if unspecified, share version should default to the latest supported share version.

```bash

# if `--share-version` isn't specified defaults to latest supported share version (currently 0)
celestia-appd tx blob payForData 0101010101010101 0101010101010101 --from validator --keyring-backend test 

# the above command behaves identically to the following command
celestia-appd tx blob payForData 0101010101010101 0101010101010101 --from validator --keyring-backend test --share-version 0

# throws error if unsupported share version is used (e.g. `1`)
celestia-appd tx blob payForData 0101010101010101 0101010101010101 --from validator --keyring-backend test --share-version 1


```

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/1603

PR #1603, opened 2023-04-06
feat: Introduce `--share-version` CLI flag

amityadav0:
## Overview

Introduce `--share-version` CLI flag 
-  #1041  

## Checklist

- [.] New and updated code has appropriate documentation
- [.] New and updated code has new and/or updated testing
- [.] Required CI checks are passing
- [] Visual proof for any user facing features like CLI or documentation updates
- [.] Linked issues closed with keywords

### DIFF
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -225,7 +225,6 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	assert := s.Assert()
 	val := s.network.Validators[0]
 	ns1 := appns.MustNewV0(bytes.Repeat([]byte{1}, appns.NamespaceVersionZeroIDSize))
-	share1 := uint8(0)
 
 	mustNewBlob := func(ns appns.Namespace, data []byte, share uint8) *blobtypes.Blob {
 		b, err := blobtypes.NewBlob(ns, data, share)
@@ -242,7 +241,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	tests := []test{
 		{
 			"small random typical",
-			mustNewBlob(ns1, tmrand.Bytes(3000), share1),
+			mustNewBlob(ns1, tmrand.Bytes(3000), appconsts.ShareVersionZero),
 			[]blobtypes.TxBuilderOption{
 				blobtypes.SetFeeAmount(sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))),
 				blobtypes.SetGasLimit(1_000_000_000),
@@ -250,7 +249,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 		},
 		{
 			"large random typical",
-			mustNewBlob(ns1, tmrand.Bytes(350000), share1),
+			mustNewBlob(ns1, tmrand.Bytes(350000), appconsts.ShareVersionZero),
 			[]types.TxBuilderOption{
 				blobtypes.SetFeeAmount(sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(10)))),
 				blobtypes.SetGasLimit(1_000_000_000),
@@ -258,7 +257,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 		},
 		{
 			"medium random with memo",
-			mustNewBlob(ns1, tmrand.Bytes(100000), share1),
+			mustNewBlob(ns1, tmrand.Bytes(100000), appconsts.ShareVersionZero),
 			[]blobtypes.TxBuilderOption{
 				blobtypes.SetMemo("lol I could stick the rollup block here if I wanted to"),
 				blobtypes.SetGasLimit(1_000_000_000),
@@ -266,7 +265,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 		},
 		{
 			"medium random with timeout height",
-			mustNewBlob(ns1, tmrand.Bytes(100000), share1),
+			mustNewBlob(ns1, tmrand.Bytes(100000), appconsts.ShareVersionZero),
 			[]blobtypes.TxBuilderOption{
 				blobtypes.SetTimeoutHeight(1000),
 				blobtypes.SetGasLimit(1_000_000_000),
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -226,8 +226,8 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	val := s.network.Validators[0]
 	ns1 := appns.MustNewV0(bytes.Repeat([]byte{1}, appns.NamespaceVersionZeroIDSize))
 
-	mustNewBlob := func(ns appns.Namespace, data []byte, share uint8) *blobtypes.Blob {
-		b, err := blobtypes.NewBlob(ns, data, share)
+	mustNewBlob := func(ns appns.Namespace, data []byte, shareVersion uint8) *blobtypes.Blob {
+		b, err := blobtypes.NewBlob(ns, data, shareVersion)
 		require.NoError(err)
 		return b
 	}
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -225,9 +225,10 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	assert := s.Assert()
 	val := s.network.Validators[0]
 	ns1 := appns.MustNewV0(bytes.Repeat([]byte{1}, appns.NamespaceVersionZeroIDSize))
+	share1 := uint8(0)
 
-	mustNewBlob := func(ns appns.Namespace, data []byte) *blobtypes.Blob {
-		b, err := blobtypes.NewBlob(ns, data)
+	mustNewBlob := func(ns appns.Namespace, data []byte, share uint8) *blobtypes.Blob {
+		b, err := blobtypes.NewBlob(ns, data, share)
 		require.NoError(err)
 		return b
 	}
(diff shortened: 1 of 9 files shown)


x/blob/types/blob_tx.go
```diff
@@ -15,7 +15,7 @@ type Blob = tmproto.Blob
 
 // NewBlob creates a new coretypes.Blob from the provided data after performing
 // basic stateless checks over it.
-func NewBlob(ns appns.Namespace, blob []byte) (*Blob, error) {
+func NewBlob(ns appns.Namespace, blob []byte, share uint8) (*Blob, error) {
```
rootulp, line 18: [nit]
```suggestion
func NewBlob(ns appns.Namespace, blob []byte, shareVersion uint8) (*Blob, error) {
```

[blocking] since this function accepts a new parameter, I would accept it to use the `shareVersion` below:

```diff
	return &tmproto.Blob{
		NamespaceId:      ns.ID,
		Data:             blob,
-		ShareVersion:     uint32(appconsts.DefaultShareVersion),
+               ShareVersion:     uint32(shareVersion),
		NamespaceVersion: uint32(ns.Version),
	}, nil
```

Additionally this function should verify that the shareVersion argument is present in https://github.com/celestiaorg/celestia-app/blob/ab24a7b4d5190a53cf4bb0a9e989f4c824759f8e/pkg/appconsts/appconsts.go#L103 prior to returning the Blob

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +69,24 @@ func CmdPayForBlob() *cobra.Command {
 	}
 
 	flags.AddTxFlagsToCmd(cmd)
+	cmd.PersistentFlags().String(FlagShareVersion, "0", "Specify the share version (default is 0)")
```
evan-forbes, line 72: if we define this flag as something other than a string, like an int for instance, will we be able to avoid having to do a conversion? Is there a different reason we need to use a string here?

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +69,24 @@ func CmdPayForBlob() *cobra.Command {
...
+	if err != nil {
+		return uint8(0), fmt.Errorf("failed to convert share version from string to uint8: %w", err)
+	}
+	switch uint8(version) {
+	case appconsts.ShareVersionZero:
+		return uint8(version), nil
+	default:
+		return uint8(0), fmt.Errorf("share version %d is not supported", version)
+	}
+}
```
evan-forbes, line 88: see above comment about using a differently typed flag

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +69,24 @@ func CmdPayForBlob() *cobra.Command {
 	}
 
 	flags.AddTxFlagsToCmd(cmd)
+	cmd.PersistentFlags().String(FlagShareVersion, "0", "Specify the share version (default is 0)")
 
 	return cmd
 }
 
+func getShare(shareVersion string) (uint8, error) {
```
rootulp, line 77: [nit]

```suggestion
func getShareVersion(shareVersion string) (uint8, error) {
```

x/blob/client/testutil/integration_test.go
```diff
@@ -87,6 +88,21 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
...
+			args: []string{
+				hexNamespace,
+				hexBlob,
+				fmt.Sprintf("--from=%s", username),
+				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
+				fmt.Sprintf("--%s=%s", flags.FlagFees, sdk.NewCoins(sdk.NewCoin(s.cfg.BondDenom, sdk.NewInt(2))).String()),
+				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
+				fmt.Sprintf("--%s=1", cli.FlagShareVersion),
+			},
+			expectErr:    true,
```
evan-forbes, line 102: nice

x/blob/client/cli/payforblob.go
```diff
@@ -41,13 +47,19 @@ func CmdPayForBlob() *cobra.Command {
 				return fmt.Errorf("failure to create namespace: %w", err)
 			}
 
+			shareVersion, _ := cmd.Flags().GetString(FlagShareVersion)
+			share, err := getShare(shareVersion)
```
rootulp, line 51: [nit] Proposal to rename `share` to `shareVersion` because the share type is a distinct concept in celestia-app. See https://github.com/celestiaorg/celestia-app/blob/ab24a7b4d5190a53cf4bb0a9e989f4c824759f8e/pkg/shares/shares.go#L12-L15

```suggestion
			shareVersion, err := getShareVersion(cmd.Flags().GetString(FlagShareVersion))
```

x/blob/types/blob_tx.go
evan-forbes, line 18: the name `share` here for the share version is a bit confusing imo. Also, we should pass it to the blob below

app/test/integration_test.go
```diff
@@ -225,9 +225,10 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	assert := s.Assert()
 	val := s.network.Validators[0]
 	ns1 := appns.MustNewV0(bytes.Repeat([]byte{1}, appns.NamespaceVersionZeroIDSize))
+	share1 := uint8(0)
```
rootulp, line 228: Usages of `share1` in this file can be replaced with `appconsts.ShareVersionZero`

```suggestion
```

evan-forbes (commented):
thanks for taking this on @amityadav0 :slightly_smiling_face: 

I left a few comments on things we should likely fix before merging, but I think we're 90% there!!

rootulp (changes requested):
Thanks for the contribution @amityadav0 ! I have one blocking comment but this PR is almost good to go.

x/blob/types/blob_tx.go
amityadav0: got it.

x/blob/client/cli/payforblob.go
amityadav0: done

x/blob/client/cli/payforblob.go
amityadav0: Thanks, changed to uint8.

x/blob/types/blob_tx.go
amityadav0: changed to shareVersion, done.

app/test/integration_test.go
amityadav0: done

app/test/integration_test.go
```diff
@@ -226,8 +226,8 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	val := s.network.Validators[0]
 	ns1 := appns.MustNewV0(bytes.Repeat([]byte{1}, appns.NamespaceVersionZeroIDSize))
 
-	mustNewBlob := func(ns appns.Namespace, data []byte) *blobtypes.Blob {
-		b, err := blobtypes.NewBlob(ns, data)
+	mustNewBlob := func(ns appns.Namespace, data []byte, share uint8) *blobtypes.Blob {
+		b, err := blobtypes.NewBlob(ns, data, share)
```
rootulp, line 230: [nit] rename `share` => `shareVersion`

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +68,27 @@ func CmdPayForBlob() *cobra.Command {
...
+			break
+		}
+	}
+
+	if result {
+		return shareVersion, nil
+	}
+	return uint8(0), fmt.Errorf("share version %d is not supported", shareVersion)
+}
+
```
rootulp, line 91: this may be simplified

```go
func validateShareVersion(shareVersion uint8) error {
	if !slices.Contains(appconsts.SupportedShareVersions, shareVersion) {
		return fmt.Errorf("share version %d is not supported", shareVersion)
	}
	return nil
}
```

x/blob/client/cli/payforblob.go
```diff
@@ -41,13 +46,19 @@ func CmdPayForBlob() *cobra.Command {
 				return fmt.Errorf("failure to create namespace: %w", err)
 			}
 
+			shareVersionFlag, _ := cmd.Flags().GetUint8(FlagShareVersion)
+			shareVersion, err := getShareVersion(shareVersionFlag)
+			if err != nil {
+				return err
+			}
```
rootulp, line 53:
```suggestion
			shareVersion, _ := cmd.Flags().GetUint8(FlagShareVersion)
			err = validateShareVersion(shareVersion)
			if err != nil {
				return err
			}
```

x/blob/client/cli/payforblob.go
rootulp: in a future refactor we can consider consolidating all the code that validates share versions into one exported function in the `shares` package

amityadav0: Thanks, Updated

x/blob/client/cli/payforblob.go
amityadav0: updated

x/blob/client/testutil/integration_test.go
amityadav0: do you mean in `integration_test.go` file or create another file for unit test ?

rootulp (dismissed):

x/blob/client/cli/payforblob.go
```diff
@@ -57,10 +69,19 @@ func CmdPayForBlob() *cobra.Command {
...
 	return cmd
 }
 
+func validateShareVersion(shareVersion uint8) error {
+	// Check if share version is present in SupportedShareVersions
+	if !slices.Contains(appconsts.SupportedShareVersions, shareVersion) {
+		return fmt.Errorf("share version %d is not supported", shareVersion)
+	}
+	return nil
+}
```
evan-forbes, line 83: after simplifying this, we're performing these exact checks when we validate the blobs during the creation of the PFB, so we can actually remove this function entirely

https://github.com/celestiaorg/celestia-app/blob/ab24a7b4d5190a53cf4bb0a9e989f4c824759f8e/x/blob/types/payforblob.go#L29-L30

x/blob/client/testutil/integration_test.go
evan-forbes: nvm, I think this is more than fine now that we can remove any additional validation :slightly_smiling_face:

evan-forbes (changes requested):
one last thing, thanks for your patience @amityadav0 :slightly_smiling_face:

rootulp (approved):

evan-forbes (approved):
thank you @amityadav0 !!!

Issue #1214, opened 2023-01-11
Adopt a deterministic square construction mechanism


cmwaters:
### Summary

Have some function `square.Construct(...` that takes an ordered list of transactions and builds a square (`[]Share`) such that all processes performing that function with the same ordered list of transactions would produce the same square.

### Problem Definition

In the current design, the proposer is at liberty to design a square in whatever way they like and all other validators simply interpret it from the block. This gives the proposer the ability to create unnecessarily large squares which may only contain 1 or 2 transactions and puts unnecessary strain on light clients. This is a part of the protocol that should be behind the curtain and that block proposers don't need to be aware of. They should only be concerned with transaction inclusion and ordering. 

### Proposal

The block data over the wire should just be an ordered list of transactions. Both the proposer and all other nodes deterministically build the same square using a method like `square.Construct(txs)` and then compare using the data root. This means that both `ProcessProposal` and `PrepareProposal` use this function and validators will prevote nil if the block does not follow the square construction rules.

This does not effect how blocks are persisted to disk nor the wrapping process for PFB transactions. That remains the same. 

This also has the added benefit that it makes compact blocks far easier and more effective because now a block can be communicated as simply an array of hashes. 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

evan-forbes:
very related to https://github.com/celestiaorg/celestia-app/issues/935 (also [specific comment](https://github.com/celestiaorg/celestia-app/issues/935#issuecomment-1377823587) discussing some of the reasoning behind the original design)

Issue #935, opened 2022-10-30
Mandatory non-interactive defaults


rootulp:
## Discussion
[Non-interactive default rules](https://celestiaorg.github.io/celestia-specs/latest/rationale/message_block_layout.html#non-interactive-default-rules) are currently framed as optional.

However, parts of the current implementation assume that non-interactive defaults are mandatory (e.g. [inclusion/paths.go](https://github.com/celestiaorg/celestia-app/blob/29e0a2751182499f7dc03598eabfc8d049ae62cb/pkg/inclusion/paths.go#L17-L20)). If desired, we can modify the current implementation to support optional non-interactive defaults. Before we do so, I'm wondering why non-interactive defaults are optional rather than mandatory (i.e. a block validity rule)?

evan-forbes:
The original reasoning behind allowing for non-deterministic square layouts was to allow for tighter packing of the square via allowing some form of interaction between the block producer and the users submitting blobs. With recent changes specified in [ADR009](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md), and the potential to reduce padding even further with other mechanisms #1161, this might not be needed any more.

This would mean that if we wanted to increase the packing efficiency, then it is up to us, and the community, to build such tooling. It also has the benefit of forcing all validators to use an efficient packing algorithm, instead of only the sophisticated ones.

Deterministic block layout would allow us to simplify the implementation by not having to wrap PFB transactions with the index when we gossip them. As pointed out by @cmwaters, doing so would significantly simplify the compact blocks implementation, because we would be able to reference the transactions gossiped via its hash. Not only that, but it would also stop block producers from setting an arbitrarily setting the max square size.

Notably, deterministic block layouts might limit other features that we have discussed, such as adding priority metadata to specific blobs within the same namespace, or allowing for custom ordering of blobs in the same namespace.

nashqueue:
I would be in favour of mandatory non-interactive default rules, but I do not know the consequences and possible limitations. 

Custom ordering within the same namespace could still be possible if the deterministic block layout has an **ordered** set of transactions as an input. The order can be specified in the input. The only downside is that it would follow the non-interactive default rules if that is one.

cmwaters:
I would like to elaborate on some of these points:
[...]

PR #1604, opened 2023-04-06
feat!: pack squares densely by implementing ADR013

### LINKED ISSUE #1538: EPIC: Implement ADR 13
## Context

Per an onsite meeting, we decided to move forward with [ADR 13](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md).

## Next steps

- [ ] Determine a threshold for the size of a blob inclusion proof. This likely requires discussion with @celestiaorg/celestia-node 
- [ ] Assuming a threshold where the # of subtree roots in a blob inclusion proof is bounded by the max square size, does padding change if the max square size increases in a subsequent hard-fork but the threshold remains the same?
- [x] https://github.com/celestiaorg/celestia-app/issues/1572
- [x] #1606
- [x] #1607

Any other follow-ups @nashqueue?

### LINKED ISSUE #1572: Implement dense squares (ADR013)
In #1161, we decided in to remove the vast majority of padding from the square by changing the formula that we use to determine which subtree roots are used. We still need to implement and incorporate that new formula.

evan-forbes:
## Overview

implements ADR013

closes #1572 
part of #1538
## Checklist


app/write_square_test.go
```diff
@@ -114,29 +101,9 @@ func Test_finalizeLayout(t *testing.T) {
...
-		{
-			squareSize:      4,
-			nonreserveStart: 3,
-			blobTxs: generateBlobTxsWithNamespaces(
-				t,
-				[]appns.Namespace{ns1, ns3, ns2},
-				[][]int{{100}, {1000}, {420}},
-			),
-			expectedIndexes: [][]uint32{{3}, {6}, {4}},
+			expectedIndexes: [][]uint32{{32}, {35}, {33}},
```
evan-forbes, line 106: tried to delete redundant tests if I noticed them

app/write_square_test.go
```diff
@@ -198,19 +154,19 @@ func Test_finalizeLayout(t *testing.T) {
...
+			expectedIndexes: [][]uint32{{4}, {8, 11, 5, 6}, {7}},
 		},
 		{
 			squareSize:      4,
 			nonreserveStart: 4,
 			blobTxs: generateBlobTxsWithNamespaces(
 				t,
 				[]appns.Namespace{ns1, ns3, ns3, ns1, ns2, ns2},
-				[][]int{{100}, {1000, 1400, 200, 200}, {420}},
+				[][]int{{100}, {900, 1400, 200, 200}, {420}},
```
evan-forbes, line 167: switched to 900 instead of 1000, because 1000 used to be 2 shares before the namespace increase

pkg/inclusion/paths.go
```diff
@@ -14,11 +14,7 @@ type path struct {
 // calculateCommitmentPaths calculates all of the paths to subtree roots needed to
 // create the commitment for a given blob.
 func calculateCommitmentPaths(squareSize, start, blobShareLen int) []path {
-	// TODO: make the non-interactive defaults optional. by calculating the
-	// NextMultipleOfBlobMinSquareSize, we are forcing use of the
-	// non-interactive defaults. If we want to make this optional in the future,
-	// we have to move this next line out of this function.
-	start, _ = shares.NextMultipleOfBlobMinSquareSize(start, blobShareLen, squareSize)
+	start, _ = shares.NextShareIndex(start, blobShareLen, squareSize)
```
evan-forbes, line 17: ref #935 and #1214

pkg/inclusion/paths_test.go
```diff
@@ -338,97 +340,144 @@ func Test_genSubTreeRootPath(t *testing.T) {
...
+		start               int
+		blobLen             int
+		expectedPath        []path
+		expectedPathIndexes []int
 	}
+	// note that calculateCommitPaths assumes ODS, so we prepend a WalkLeft to
+	// everything elsewhere
 	tests := []test{
-		{2, 0, 1, []path{{instructions: []WalkInstruction{WalkLeft}, row: 0}}},
-		{2, 2, 2, []path{{instructions: []WalkInstruction{}, row: 1}}},
```
evan-forbes, line 348: had to re-write these tests cause now there are way too many damn paths to type by hand. Now this test only looks for specific paths of the result.

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,10 +83,27 @@ func roundUpBy(cursor, v int) int {
...
+
+	// use the next power of two, as subtree roots inherently require this
+	s = RoundUpPowerOfTwo(s)
+
+	// per ADR013, use the min square size should that ever raise
+	if s < appconsts.DefaultMinSquareSize {
+		return appconsts.DefaultMinSquareSize
+	}
+
+	return s
```
evan-forbes, line 106: This is the only necessary meaningful logic change!!! the rest is just fixing tests lol

evan-forbes (commented):
note that there's only one real logic change required here, the rest is just fixing tests

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,10 +83,31 @@ func roundUpBy(cursor, v int) int {
...
+	// root threshold
+	s = RoundUpPowerOfTwo(s)
+
+	// using a value below the min square size is wasteful, see ADR013 for more
+	// details
+	if s < appconsts.DefaultMinSquareSize {
+		return appconsts.DefaultMinSquareSize
+	}
+
+	return s
```
evan-forbes, line 110: this is the only consensus logic we have to change!! the rest is just fixing tests

rootulp:
I'm still reviewing but @evan-forbes what are you thoughts on updating the specs (i.e. [here](https://celestiaorg.github.io/celestia-app/rationale/message_block_layout.html#non-interactive-default-rules)) at the same time as merging the implementation?

evan-forbes:
> I'm still reviewing but @evan-forbes what are you thoughts on updating the specs (i.e. [here](https://celestiaorg.github.io/celestia-app/rationale/message_block_layout.html#non-interactive-default-rules)) at the same time as merging the implementation?

that's a good point! I don't want to block the merging of this by codeowners of ADRs and docs, so will do in separate PRs, I added #1606 and #1607 as requirements to the EPIC though

pkg/inclusion/paths_test.go
```diff
@@ -338,97 +340,144 @@ func Test_genSubTreeRootPath(t *testing.T) {
...
+				},
+			},
+			[]int{0, 1},
+		},
+		{
+			"single share in the middle of a 128x128", 128, 8252, 1,
+			[]path{
+				{
+					row:          64,
+					instructions: []WalkInstruction{WalkLeft, WalkRight, WalkRight, WalkRight, WalkRight, WalkLeft, WalkLeft},
```
rootulp, line 400: [for reviewers]

We're looking for index 60 of 128 on row 64

Initially the Merkle root spans leaf nodes: 0-128
1. Left: 0-64
2. Right: 32-64
3. Right: 48-64
4. Right: 56-64
5. Right: 60-64
6. Left: 60-62
7. Left: 60-61

pkg/inclusion/paths_test.go
```diff
@@ -338,97 +340,144 @@ func Test_genSubTreeRootPath(t *testing.T) {
...
+				},
+			},
+			[]int{0},
+		},
+		{
+			"the 32nd path for the smallest blob with a subtree width of 128", 128, 0, 8193,
```
rootulp, line 406: [for reviewers] A blob of length 8193 spans a little over 64 rows of a square with width 128. I think this test is asserting that an arbitrary row in the range 0-64 uses the root of the NMT for that row.

pkg/inclusion/paths_test.go
```diff
@@ -338,97 +340,144 @@ func Test_genSubTreeRootPath(t *testing.T) {
...
+				},
+			},
+			[]int{31},
+		},
+		{
+			"the 32nd path for the largest blob with a subtree width of 1", 128, 0, appconsts.SubtreeRootHeightThreshold,
+			[]path{
+				{
+					row:          0,
+					instructions: []WalkInstruction{WalkLeft, WalkLeft, WalkRight, WalkRight, WalkRight, WalkRight, WalkRight},
```
rootulp, line 430: [for reviewers]

Original Merkle tree spans 0-128
1. Left: 0 - 64
2. Left: 0 - 32
3. Right: 16-32
4. Right: 24-32
5. Right: 28-32
6. Right: 30-32
7. Right: 31-32

pkg/shares/non_interactive_defaults_test.go
```diff
@@ -236,7 +230,7 @@ func TestNextMultipleOfBlobMinSquareSize(t *testing.T) {
 			blobLen:       2,
 			squareSize:    8,
 			fits:          true,
-			expectedIndex: 12,
+			expectedIndex: 11,
 		},
 		{
 			// inspired by the diagram at https://github.com/celestiaorg/celestia-app/blob/1b80b94a62c8c292f569e2fc576e26299985681a/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md?plain=1#L30
```
rootulp, line 236: proposal to delete this comment b/c the diagram from ADR9 and the test case now diverge.

x/blob/types/payforblob.go
```diff
@@ -185,8 +184,8 @@ func CreateCommitment(blob *Blob) ([]byte, error) {
 	// the commitment is the root of a merkle mountain range with max tree size
 	// equal to the minimum square size the blob can be included in. See
 	// https://github.com/celestiaorg/celestia-app/blob/fbfbf111bcaa056e53b0bc54d327587dee11a945/docs/architecture/adr-008-blocksize-independent-commitment.md
```
rootulp, line 186: [blocking] this comment needs to be updated

x/blob/types/payforblob.go
```diff
@@ -293,14 +292,6 @@ func extractBlobComponents(pblobs []*tmproto.Blob) (namespaceVersions []uint32,
 	return namespaceVersions, namespaceIds, sizes, shareVersions
 }
 
-// BlobMinSquareSize returns the minimum square size that blobSize can be included
-// in. The returned square size does not account for the associated transaction
-// shares or non-interactive defaults, so it is a minimum.
-func BlobMinSquareSize[T constraints.Integer](blobSize T) T {
```
rootulp, line 299: 👍

rootulp (commented):
Overall mostly LGTM. One blocking comment. Defer to you if we want to update specs in this PR or after this PR.

pkg/appconsts/appconsts.go
```diff
@@ -72,6 +72,13 @@ const (
 	// data square.
 	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
 
+	// SubtreeRootHeightThreshold dictates the threashold for increasing the
+	// sub tree root height for blobs. It is used for calculating the amount of
+	// padding between blobs, and the number of shares included in each subtree
+	// root for a blob. The rationale for this value is described in more detail
+	// in ADR013.
+	SubtreeRootHeightThreshold = 128
```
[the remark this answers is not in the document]
evan-forbes: if I'm understanding this correctly, we would use this constant to calculate the threshold, then proceed as normal?

If the limit is actually 128 that would be a clever way of framing it, but I think we can have blobs with more than 128 even when following ADR013. For instance, a blob with 16143 shares that starts on the second row would have one per 126 rows, then for last row we could have 15 shares, which results in 1 sub tree root for the first 8, then another for 4, 2, 1. so 130 roots total

x/blob/types/payforblob.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/1604/commits/6b0353f84554a6c4a711d37b6d2e33ca40db957c

pkg/appconsts/appconsts.go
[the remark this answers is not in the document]
evan-forbes: tried to improve using this feedback in https://github.com/celestiaorg/celestia-app/pull/1604/commits/6b0353f84554a6c4a711d37b6d2e33ca40db957c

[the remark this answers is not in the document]
rootulp: > For instance, a blob with 16143 shares that starts on the second row would have one per 126 rows, then for last row we could have 15 shares, which results in 1 sub tree root for the first 8, then another for 4, 2, 1. so 130 roots total

Very good point.

rootulp (dismissed):

staheri14 (commented):
Left a minor comment so far. LGTM in general.
Though, I have a question about the ADR itself.

How do we calculate the final commitment of the blob shares based on this new arrangement, take the [picture at the bottom](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/assets/blob-alignment-comparison.png) as an example (the picture is borrowed from [the original ADR 13](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md).

In the previous arrangement (the picture at the top), for the yellow Blob which had a total of 13 shares, we would calculate the minimum square size which is 4, then split the shares into groups of 4, 4, 4, and 1, and then calculate the roots for each group accordingly. This calculation was not impacted by the actual data square size (which seems to be one of the original design principles).

However, in the new arrangement, the first two shares of the yellow blob are placed at the end of the third row. Therefore, it's unclear how this change will impact the share commitment calculation. Could you provide more clarification on this? Thanks. **Update**: I think it is going to be 2,2,2,2,2,2,1 i.e., the output of `merkleMountainRangeSizes(13, 2)`

pkg/appconsts/appconsts.go
```diff
@@ -72,6 +72,13 @@ const (
 	// data square.
 	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
 
+	// SubtreeRootHeightThreshold dictates the threashold for increasing the sub
+	// tree root height for blobs. If the number of subtree roots used to create
+	// a share commitment surpasses this threshold, then the height is
+	// increased. The rationale for this value is described in more detail in
+	// ADR013 (./docs/architecture/adr-013).
+	SubtreeRootHeightThreshold = 128
```
staheri14, line 80: [question] Does `height` mean the actual tree height or does it mean the total number of sub-tree roots? The term `height` is not used in the ADR 13.

evan-forbes: height is the height of the subtree, so if 128 subtree roots are crossed, then the height of the subtree is increased

you are definitely correct that the ADR does not mention height! One thing that I want to do in https://github.com/celestiaorg/celestia-app/issues/1607 is to make a clearer reference to the algorithm, as I don't believe the part where we increase the height (aka round the sub tree width up to a power of two) is clear enough

per the seventh paragraph in the Context section
> ...Blobs that are larger than MaxSquareSize * 2 but smaller than MaxSquareSize * 4 can use the index of 4 to get a maximum of 3 padding shares and so on.

I interpreted this to mean that we are using a higher sub tree root

pkg/shares/non_interactive_defaults.go
```diff
@@ -34,32 +34,32 @@ func BlobSharesUsedNonInteractiveDefaults(cursor, squareSize int, blobShareLens
...
+// https://github.com/celestiaorg/celestia-app/blob/0334749a9e9b989fa0a42b7f011f4a79af8f61aa/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md
+func NextShareIndex(cursor, blobShareLen, squareSize int) (index int, fitsInRow bool) {
 	// if we're starting at the beginning of the row, then return as there are
 	// no cases where we don't start at 0.
 	if isStartOfRow(cursor, squareSize) {
 		return cursor, true
 	}
 
-	blobMinSquareSize := MinSquareSize(blobLen)
+	blobMinSquareSize := SubTreeWidth(blobShareLen)
```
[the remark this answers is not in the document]
evan-forbes: good catch! f74b6eb5202bd835b4dca0820c89c65817fc93ad

pkg/shares/non_interactive_defaults.go
```diff
@@ -34,35 +34,35 @@ func BlobSharesUsedNonInteractiveDefaults(cursor, squareSize int, blobShareLens
...
-// non negative, and that squareSize is a power of two.
+// NextShareIndex determines the next index in a square that can be used. It
+// follows the non-interactive default rules defined in ADR013. This function
+// returns false if the entire the blob cannot fit on the given row. Assumes
+// that all args are non negative, and that squareSize is a power of two.
 // https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
-// https://github.com/celestiaorg/celestia-app/blob/1b80b94a62c8c292f569e2fc576e26299985681a/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
-func NextMultipleOfBlobMinSquareSize(cursor, blobLen, squareSize int) (index int, fitsInRow bool) {
+// https://github.com/celestiaorg/celestia-app/blob/0334749a9e9b989fa0a42b7f011f4a79af8f61aa/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md
+func NextShareIndex(cursor, blobShareLen, squareSize int) (index int, fitsInRow bool) {
```
cmwaters, line 50: Is `fitsInRow` still used by anything?  (that is not a test)

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,10 +83,31 @@ func roundUpBy(cursor, v int) int {
...
+
+	// use a power of two equal to or larger than the multiple of the subtree
+	// root threshold
+	s = RoundUpPowerOfTwo(s)
+
+	// using a value below the min square size is wasteful, see ADR013 for more
+	// details
+	if s < appconsts.DefaultMinSquareSize {
+		return appconsts.DefaultMinSquareSize
+	}
```
cmwaters, line 108: I thought we were removing the notion of a `MinSquareSize`?

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,10 +83,31 @@ func roundUpBy(cursor, v int) int {
...
 }
 
-// MinSquareSize returns the minimum square size that can contain shareCount
-// number of shares.
-func MinSquareSize(shareCount int) int {
-	return RoundUpPowerOfTwo(int(math.Ceil(math.Sqrt(float64(shareCount)))))
+// SubTreeWidth determines how many shares should be included when creating
+// subtree. This is used for determining which subtree roots are used in the
+// share commitment for PFBs, and therefore also the padding between blobs. The
+// reasoning behind this algorithm is discussed in depth in ADR013.
```
cmwaters, line 89: Would include a link here as well if possible

pkg/shares/non_interactive_defaults.go
```diff
@@ -34,35 +34,35 @@ func BlobSharesUsedNonInteractiveDefaults(cursor, squareSize int, blobShareLens
...
-// the entire the blob cannot fit on the given row. Assumes that all args are
-// non negative, and that squareSize is a power of two.
+// NextShareIndex determines the next index in a square that can be used. It
+// follows the non-interactive default rules defined in ADR013. This function
+// returns false if the entire the blob cannot fit on the given row. Assumes
+// that all args are non negative, and that squareSize is a power of two.
 // https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
-// https://github.com/celestiaorg/celestia-app/blob/1b80b94a62c8c292f569e2fc576e26299985681a/docs/architecture/adr-009-non-interactive-default-rules-for-reduced-padding.md
-func NextMultipleOfBlobMinSquareSize(cursor, blobLen, squareSize int) (index int, fitsInRow bool) {
+// https://github.com/celestiaorg/celestia-app/blob/0334749a9e9b989fa0a42b7f011f4a79af8f61aa/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md
```
cmwaters, line 49: Ideally, this links to the spec instead of ADR013 as this ADR builds upon quite a bit of existing context and won't be as complete as an explanation for how the square is constructed than the spec should be.

pkg/shares/non_interactive_defaults.go
evan-forbes: since ADR009, we've had two minsquaresizes, the above is a constant depicting the smallest viable ODS, where as the ADR009 min square size is a var that changes pers the blob for the minimum square size that the blob can fit in

pkg/shares/non_interactive_defaults.go
evan-forbes: no we haven't used it for a while, we just haven't removed it

pkg/shares/non_interactive_defaults.go
cmwaters: But isn't this always 1x1? When would it not be 1x1?

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,10 +83,31 @@ func roundUpBy(cursor, v int) int {
 	}
 }
 
-// MinSquareSize returns the minimum square size that can contain shareCount
-// number of shares.
-func MinSquareSize(shareCount int) int {
-	return RoundUpPowerOfTwo(int(math.Ceil(math.Sqrt(float64(shareCount)))))
+// SubTreeWidth determines how many shares should be included when creating
```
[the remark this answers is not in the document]
cmwaters: Also, don't we call this the `MaxSubTreeWidth` since a share commitment could be made of subtree roots of a smaller width (where width is the amount of leaf nodes in the subtree right?)

pkg/appconsts/appconsts.go
```diff
@@ -72,6 +72,13 @@ const (
 	// data square.
 	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
 
+	// SubtreeRootHeightThreshold dictates the threashold for increasing the sub
```
evan-forbes, line 75:
```suggestion
	// SubtreeRootHeightThreshold dictates the threshold for increasing the sub
```

nashqueue:
> Could you provide more clarification on this? Thanks. Update: I think it is going to be 2,2,2,2,2,2,1 i.e., the output of merkleMountainRangeSizes(13, 2)

It is exactly how you described it. The merkleMountainRangeSizes takes in the max height of the subTree, which is 2 in this case. 
If you move in powers of 2, it will stay blocksize independent because the trees we need will exist in every blocksize that the blobs fit in unless they don't see explanation 2. 

@evan-forbes One thing that we have to watch out for is that we should always take the min(MinSquareSize(),SubTreeWidth). I am not sure if I mentioned that in ADR13. This is because, depending on the Threshold, MinSquareSize() could be more desirable. For this implementation, this is not a problem because we chose 128 as the threshold, and SubTreeWitdh is always <= MinSquareSize(). If we continue to adjust the Threshold to be MaxSquareSize, then SubTreeWitdh <= MinSquareSize() always holds. 

Let's say we have a small threshold of (3), and a blob of the size of 51. 
The MinSquareSize(51) is 16 so it would be 16,16,16, 2, 1
The SubTreeWidth(51) is 32 =  51/3 -> 17 -> 32
This means the tree would look like 32, 16, 2, 1

Here we have 2 Problems: 
1. 32, 16, 2, 1 are 4 subtree roots which is not what we signed up for (3). Therefore the Threshold has to be >= log(MaxSquareSize*MaxSquareSize). The Intuition here is to ask yourself how many binary digits you need to represent a number (the largest possible blob). This will relate directly to how many subtree roots I need in a maximum to represent a Merkle mountain range. This is log_2(2^x * 2^x) = 2x. So we need at minimum 2* log(MaxSquareSize) for the threshold to hold that the number of subtree roots <= the Threshold. 

2. 32 is the largest subtree root. This means that the padding in the worst case for this blob is 31 instead of using MinSquareSize() and having 15 as the worst-case padding for this blob. If the threshold is above 2* log(MaxSquareSize) but still small, then I believe we can run into this issue again. Therefore we need to use min(MinSquareSize(),SubTreeWidth) if the threshold is < MaxSquareSize . This will preserve the padding savings and optimally use it the best. Another thing why we have to do this is because let's say blob 51 is in a 16*16 square. The resulting subtree root of 32 does not exist in this square, so we have to take the MinSquareSize() instead.

pkg/appconsts/appconsts.go
cmwaters: I tend to agree with omitting height from the name

cmwaters:
> 32, 16, 2, 1 are 4 subtree roots which is not what we signed up for (3). Therefore the Threshold has to be >= log(MaxSquareSize * MaxSquareSize).

This makes sense that we need to impose a lower bound to the threshold (although it was assumed that the threshold would always be a number close to the MaxSquareSize in any case)

pkg/appconsts/appconsts.go
staheri14: Thanks for clarification on this. I have a followup question for @nashqueue:
Quoting from ADR 13, where there is an argument on the proof size and its relation with the number of subtree roots for a given blob, i.e., 
>Small blob sizes have the lowest ratio but also small inclusion proofs. Therefore increasing the proof size is not a problem until some threshold. It would increase the ratio of blob size to padding for small blobs which have the worst ratio.
 Assuming there is a threshold, the number of subtree roots in a proof, where the proof size of a blob is acceptable, we can use this threshold to determine the index of the blob in the square. This would give us zero padding for blobs that are smaller than the threshold and non-zero padding for blobs that are larger than the threshold but still smaller than before.

My understanding is that the `SubtreeRootThreshold` in the code corresponds to the threshold mentioned in the paragraph above. So, my question is that what is this threshold mean to enforce, if any? Does this threshold is to enforce a hard cap on the number of subtree roots per blob (as the current algorithm stands, the number of subtree roots per blob is not capped, take a blob share size of 72 and a threshold of 8 as a counterexample).

pkg/shares/non_interactive_defaults.go
evan-forbes: It might simply be a remnant of malleation, where we might have wanted to limit the number of possible square sizes to sign over.

pkg/shares/non_interactive_defaults.go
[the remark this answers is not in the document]
evan-forbes: > By this do you mean how many shares should be in a subtree root that gets used in the share commitment

yeah exactly, given the size of the blob in shares, how many leaves (width) should be in the subtree. 

>So if it's a leaf node the answer is 1 share and if it's an intermediary node it could be 2, 4, 8 or more shares that constitute the subtree root.

the input should be size of the blob determined in shares, tried to clarify in 643a18e2dd68a9e651e42ff359b961b0ce61d959

The width determined by this function is the width that must be used, correct? If that's the case, then I don't think max is appropriate.

pkg/shares/non_interactive_defaults.go
evan-forbes: for now, I have only included `celestia-app/docs/architecture/adr-013...` instead of a link, because I don't want to use an outdated or relative link

however, I did leave a note in #1607 to update links here after that or the changes to the specs are added

evan-forbes:
> One thing that we have to watch out for is that we should always take the min(MinSquareSize(),SubTreeWidth).

> The resulting subtree root of 32 does not exist in this square, so we have to take the MinSquareSize() instead.

@staheri14 and myself were discussing this Tuesday, and was the reason for me hardcoding to the maxsquare size in https://github.com/celestiaorg/celestia-app/pull/1604/commits/1d4f815e85165bae1f7fbc35999d5441e16d2aed . The issues you mention are true and using the min(MinSquareSize(),SubTreeWidth) should fix them I think. 

> 32, 16, 2, 1 are 4 subtree roots which is not what we signed up for (3)

I think there was some confusion over what exactly this threshold determines, and it can't (in every single case) determine the max number of subtree roots. see https://github.com/celestiaorg/celestia-app/pull/1604#discussion_r1160946823

we definitely need some limit to the lower bound, log(MaxSquareSize * MaxSquareSize) makes sense, for now, hardcoding to the max square size seems like the best option. 

we also need to write down all of this in the specs #1606 and update the ADR #1607 before the next audit

pkg/shares/non_interactive_defaults.go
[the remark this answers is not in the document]
cmwaters: Not quite. Say there is a blob of 129 shares (and a threshold of 128). The result is 2 however the set of subroots that make the share commitment (AFAIU), is going to be [2, 1, 1, 1, 1, 1, ....] i.e. 1x2 and 127x1. Thus here 2 is the max and the rest is 1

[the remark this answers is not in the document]
cmwaters: Actually it might be [2,2,2,2,2,2....,1] (but still 2 is the max width as there are subroots there of a smaller width

nashqueue:
> Isn't the MinSquareSize(51) = 8 as 8*8 = 64 is greater than 51, hence can accommodate all the shares?

Yes my bad you are correct. so the sequence would look like 8 , 8 , 8 , 8 , 8 , 8 , 2 , 1 

> I am not sure what ` explanation 2` refers to? can you please clarify pray

Explanation 2 refers to point 2 I made in the comment. To reiterate the comment I made with the corrected sequence:  Another thing why we have to do this is because let's say blob 51 is in an **8 * 8** square. The resulting subtree of 32 does not exist in this square, so we have to take the MinSquareSize() instead.

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,12 +86,54 @@ func roundUpBy(cursor, v int) int {
...
+	s = RoundUpPowerOfTwo(s)
+
+	// use the minimum of the subtree width and the min square size, this
+	// gurarantees that a valid value is returned
+	s = min(s, BlobMinSquareSize(shareCount))
+
+	// using a value below the min square size is wasteful, see ADR013 for more
+	// details
+	return max(s, appconsts.DefaultMinSquareSize)
+}
```
evan-forbes, line 121: note the update to the algo per the call this morning:

the blob min square size was re-added, then we're using the min of that value.

staheri14:
@evan-forbes Could you please let me know what are the next steps for the current PR? Are we going to:

1. Discuss the algorithm's details in a separate specification PR before finalizing the implementation PR review?
2. Proceed with the implementation PR as it is and have the specification PR come later? If any changes are decided upon due to the specification PR, the implementation change will happen afterwards.

evan-forbes:
@staheri14 I think we can go with option 2

I opened #1630 for the spec portion, but have yet to complete it, I think this PR is ready to merge. It has been updated to reflect the most recent comments and discussions

pkg/appconsts/appconsts.go
```diff
@@ -72,6 +72,19 @@ const (
...
+	// tree root height for blobs. If the number of subtree roots used to create
+	// a share commitment surpasses this threshold, then the height is
+	// increased. The rationale for this value is described in more detail in
+	// ADR013 (./docs/architecture/adr-013). Note that this value should not
+	// drop below the max square size. That could result in the caculating a sub
+	// tree width for a blob that is larger than whatever the actual suquare
+	// size for that block. Having a value that is larger than the square is
+	// simply wasteful in that the proofs for large blobs are unnecessarily
+	// large.
+	// ADR013 https://github.com/celestiaorg/celestia-app/blob/e905143e8fe138ce6085ae9a5c1af950a2d87638/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md //nolint: lll
```
rootulp, line 85: [typo] suquare => square

```suggestion
	// SubtreeRootThreshold dictates the threashold for increasing the sub tree
	// root height for blobs. If the number of subtree roots used to create a
	// share commitment surpasses this threshold, then the height is increased.
	// The rationale for this value is described in more detail in ADR013
	// (./docs/architecture/adr-013). Note that this value should not drop below
	// the max square size because that would result in sub tree widths that are
	// larger than the actual suquare size for that block. Having a value that
	// is larger than the square is simply wasteful in that the proofs for large
	// blobs are unnecessarily large.
	// ADR013 https://github.com/celestiaorg/celestia-app/blob/e905143e8fe138ce6085ae9a5c1af950a2d87638/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md //nolint: lll
```

rootulp (approved):

pkg/appconsts/appconsts.go
evan-forbes: wait, you don't know what a suquare is?

pkg/appconsts/appconsts.go
```diff
@@ -72,16 +72,15 @@ const (
...
-	// a share commitment surpasses this threshold, then the height is
-	// increased. The rationale for this value is described in more detail in
-	// ADR013 (./docs/architecture/adr-013). Note that this value should not
-	// drop below the max square size. That could result in the caculating a sub
-	// tree width for a blob that is larger than whatever the actual suquare
-	// size for that block. Having a value that is larger than the square is
-	// simply wasteful in that the proofs for large blobs are unnecessarily
-	// large.
+	// SubtreeRootThreshold dictates the threashold for increasing the sub tree
+	// root height for blobs. If the number of subtree roots used to create a
```
staheri14, line 76: [question] Is the term `height` still applicable?

staheri14 (approved):
LGTM! added some comments.

app/write_square_test.go
```diff
@@ -159,25 +126,14 @@ func Test_finalizeLayout(t *testing.T) {
...
-			// only two blob txs should make it in the square
-			squareSize:      4,
-			nonreserveStart: 4,
-			blobTxs: generateBlobTxsWithNamespaces(
-				t,
-				[]appns.Namespace{ns1, ns2, ns3},
-				[][]int{{1800}, {1800}, {6000}},
-			),
-			expectedIndexes: [][]uint32{{4}, {8}},
-		},
```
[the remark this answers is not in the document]
evan-forbes: just because a lot of these unit tests are redundant

pkg/appconsts/appconsts.go
```diff
@@ -72,6 +72,19 @@ const (
 	// data square.
 	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
 
+	// SubtreeRootThreshold dictates the threashold for increasing the sub
+	// tree root height for blobs. If the number of subtree roots used to create
+	// a share commitment surpasses this threshold, then the height is
+	// increased. The rationale for this value is described in more detail in
+	// ADR013 (./docs/architecture/adr-013). Note that this value should not
+	// drop below the max square size. That could result in the caculating a sub
```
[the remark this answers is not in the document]
evan-forbes: yeah, you are correct, good catch those docs were from the last iteration. fixed in https://github.com/celestiaorg/celestia-app/pull/1604/commits/01ae46da55e36e4e7248cf37ee6e56af029082ba

pkg/appconsts/appconsts.go
```diff
@@ -72,6 +72,19 @@ const (
 	// data square.
 	MinShareCount = DefaultMinSquareSize * DefaultMinSquareSize
 
+	// SubtreeRootThreshold dictates the threashold for increasing the sub
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/1604/commits/dd8a14f1727d00970ce8224d9c915a86eb9c68a7

pkg/inclusion/paths.go
```diff
@@ -34,9 +30,9 @@ func calculateCommitmentPaths(squareSize, start, blobShareLen int) []path {
 		}
 
 		// subTreeRootMaxHeight is the maximum height of a subtree root that was
-		// used to generate the commitment. The height is based on the minimum
-		// square size the blob can fit into. See ADR-008 for more details.
-		subTreeRootMaxHeight := int(math.Log2(float64(shares.MinSquareSize(blobShareLen))))
+		// used to generate the commitment. The height is based on the
+		// SubtreeRootThreshold. See ADR-013 for more details.
+		subTreeRootMaxHeight := int(math.Log2(float64(shares.SubTreeWidth(blobShareLen))))
 		minDepth := maxDepth - subTreeRootMaxHeight
```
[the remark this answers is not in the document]
evan-forbes: :+1: good point 3615971f393721f068f2702b4ab481af95b57196

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,12 +86,54 @@ func roundUpBy(cursor, v int) int {
...
+	s := (shareCount / appconsts.SubtreeRootThreshold)
+
+	// round up if the width is not an exact multiple of the threshold
+	if shareCount%appconsts.SubtreeRootThreshold != 0 {
+		s++
+	}
+
+	// use a power of two equal to or larger than the multiple of the subtree
+	// root threshold
+	s = RoundUpPowerOfTwo(s)
```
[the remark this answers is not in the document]
evan-forbes: yeah, that's correct. Roughly, the new algo aims for a max threshold of subtree roots, but that value is not guaranteed

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,12 +86,54 @@ func roundUpBy(cursor, v int) int {
...
+
+	// use the minimum of the subtree width and the min square size, this
+	// gurarantees that a valid value is returned
+	s = min(s, BlobMinSquareSize(shareCount))
+
+	// using a value below the min square size is wasteful, see ADR013 for more
+	// details
+	return max(s, appconsts.DefaultMinSquareSize)
```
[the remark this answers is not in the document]
evan-forbes: this is good point, I might have misunderstood what @nashqueue meant by "min square size", it seems like this was a mistake after adding it, so it has been removed.

@nashqueue do we need to take the max here? If the DefaultMinSquareSize got moved up to 32 or something large for some reason, it seems like we would still want to use min( `BlobMinSquareSize`, `new threshold based value`)

pkg/shares/non_interactive_defaults.go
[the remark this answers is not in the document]
staheri14: Thanks for clarification 👍

pkg/inclusion/paths.go
```diff
@@ -33,11 +29,11 @@ func calculateCommitmentPaths(squareSize, start, blobShareLen int) []path {
 			end = normalizedEndIndex
 		}
 
-		// subTreeRootMaxHeight is the maximum height of a subtree root that was
-		// used to generate the commitment. The height is based on the minimum
-		// square size the blob can fit into. See ADR-008 for more details.
-		subTreeRootMaxHeight := int(math.Log2(float64(shares.MinSquareSize(blobShareLen))))
-		minDepth := maxDepth - subTreeRootMaxHeight
+		// subTreeRootMaxDepth is the maximum height of a subtree root that was
```
staheri14, line 32: [nit] 
```suggestion
		// subTreeRootMaxDepth is the maximum depth of a subtree root that was
```

pkg/shares/non_interactive_defaults.go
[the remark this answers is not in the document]
evan-forbes: We do not want to increase it, but we do currently use it to force a square size on an empty block, so I'm not sure if we should remove it entirely yet either. We should definitely be able to unexport and rename it though.

created #1636

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,12 +86,43 @@ func roundUpBy(cursor, v int) int {
...
-// MinSquareSize returns the minimum square size that can contain shareCount
+// BlobMinSquareSize returns the minimum square size that can contain shareCount
 // number of shares.
-func MinSquareSize(shareCount int) int {
+func BlobMinSquareSize(shareCount int) int {
 	return RoundUpPowerOfTwo(int(math.Ceil(math.Sqrt(float64(shareCount)))))
 }
 
+// SubTreeWidth determines the maximum number of leaves per subtree in the share
+// commitment over a given blob. The input should be the total number of shares
```
cmwaters, line 96: 👍 Like this description

cmwaters (approved):
This LGTM. Where do we update the `ShareCommitment` generation rules (do they need to be updated). Do we also need to modify the logic for proving blobs?

staheri14 (approved):
Thanks for being so responsive and attentive to the comments 👍 LGTM!

pkg/shares/non_interactive_defaults_test.go
```diff
@@ -361,8 +410,72 @@ func TestMinSquareSize(t *testing.T) {
...
+		{
+			shareCount: (appconsts.SubtreeRootThreshold * appconsts.DefaultMaxSquareSize) - 1,
+			want:       128,
+		},
+	}
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("shareCount %d", tc.shareCount), func(t *testing.T) {
+			got := SubTreeWidth(tc.shareCount)
+			assert.Equal(t, tc.want, got, i)
+			assert.GreaterOrEqual(t, got, appconsts.DefaultMinSquareSize)
```
cmwaters, line 478: This isn't relevant anymore. Perhaps you want to assert:
```suggestion
			assert.GreaterOrEqual(t, got, BlobMinSquareSize(tc.shareCount))
```

pkg/shares/non_interactive_defaults.go
```diff
@@ -83,12 +86,43 @@ func roundUpBy(cursor, v int) int {
 	}
 }
 
-// MinSquareSize returns the minimum square size that can contain shareCount
+// BlobMinSquareSize returns the minimum square size that can contain shareCount
 // number of shares.
-func MinSquareSize(shareCount int) int {
+func BlobMinSquareSize(shareCount int) int {
```
cmwaters, line 91: Do we want to write a couple of unit tests for this. I don't see any here

pkg/shares/non_interactive_defaults.go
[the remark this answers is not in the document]
nashqueue: I think if we have a "min square size" of 16 but s will turn into 8 we should still use 8. If it is the other way and "min square size" is 8 but s is 16 then we probably use 16 and the square will be at least 16*16. So default minSquareSIze does not affect the result in any direction if I am seeing this correctly.

pkg/shares/non_interactive_defaults.go
evan-forbes: ahh I forgot to rename the existing ones, this is the same func we've been using since ADR009, so those unit tests are still there I believe. Do these suffice?

https://github.com/celestiaorg/celestia-app/pull/1604/commits/0b2802e24c529e189ec743322139ec925be1e4e0

https://github.com/celestiaorg/celestia-app/blob/74c4b248a430183cef85ed0de6d06e9c37b2b255/pkg/shares/non_interactive_defaults_test.go#L372-L417

pkg/shares/non_interactive_defaults_test.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/1604/commits/74c4b248a430183cef85ed0de6d06e9c37b2b255

evan-forbes: good catch!

cmwaters (approved):

pkg/shares/non_interactive_defaults_test.go
cmwaters: As long as we're covering for it in our test cases then I'm good

pkg/appconsts/appconsts.go
```diff
@@ -72,16 +72,11 @@ const (
...
-	// increased. The rationale for this value is described in more detail in
-	// ADR013 (./docs/architecture/adr-013). Note that this value should not
-	// drop below the max square size. That could result in the caculating a sub
-	// tree width for a blob that is larger than whatever the actual suquare
-	// size for that block. Having a value that is larger than the square is
-	// simply wasteful in that the proofs for large blobs are unnecessarily
-	// large.
+	// SubtreeRootThreshold works as a target value for the number of subtree roots in the
+	// share commitment. If a blob contains more shares than this number, than the height
+	// of the subtree roots will gradually increases to so that the amount remains within that limit.
```
rootulp, line 77:
```suggestion
	// share commitment. If a blob contains more shares than this number, then the height
	// of the subtree roots will increase so that the number of subtree roots remains below the threshold.
```

rootulp (approved):

Issue #1162, opened 2023-01-03
Remove or refactor the KeyringSigner


evan-forbes:
Originally, the `KeyringSigner` was needed for malleated transactions, where we had to programmatically sign a single transaction multiple times. Now that we are not performing the malleation step, we should refactor or remove that code, since its only purpose now is to programmatically sign normal txs, and could probably be done simpler.

Issue #832, opened 2022-10-03
Consolidate utility testing functions where possible


evan-forbes:
As we've discussed synchronously many times, there's quite a bit of redundant testing utility functions. We should consolidate where possible. If its not possible due to cyclic imports, then we should consider putting those tests in their own package. 

related to #829

amoghrajesh:
@evan-forbes I can work on this issue. Can you point me to the files to get started?

rootulp:
A candidate sub-task for this issue is to delete the duplicate definitions of `randomValidNamespace`:
1. https://github.com/celestiaorg/celestia-app/blob/27632ddc804fbf7655ef5ba941eb4f346c539be9/testutil/payment/testutil.go#L134
2. https://github.com/celestiaorg/celestia-app/blob/e18de882c3de4ee92e4632af21d4007c1b642adb/app/test_util.go#L178
3. https://github.com/celestiaorg/celestia-app/blob/8f4fd811b19b8539bfd9e77e83f23518ec702c88/app/test/block_size_test.go#L299
4. https://github.com/celestiaorg/celestia-app/blob/906fdf74f871f13d3fe0c632dfda95db494521ee/pkg/prove/proof_test.go#L290
5. https://github.com/celestiaorg/celestia-app/blob/07b5bb4a6dd905166c33b3434895cd292501f100/testutil/utils.go#L15

The last function definition is the most preferred because it accounts for the parity and tail padding namespaces. Note: the duplicate functions are defined in different packages so the last function may need to be extracted to a package that can be imported by all usages of the duplicate definitions.

rootulp:
Spun out https://github.com/celestiaorg/celestia-app/issues/836 in-case @evan-forbes has other ideas for what this issue should include.

evan-forbes:
another instance per this https://github.com/celestiaorg/celestia-app/pull/1991#discussion_r1243461687

rootulp:
Closing this b/c it's not actionable. Please re-open or create a new issue with more specific examples of the redundant testing utilities that need consolidation.

PR #1613, opened 2023-04-11
feat: tx simulator

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/issues/1256

This PR introduces a new package `txsim` for contolled fuzz testing at a transaction level. It's purpose is to simulate a wide range of possible user interactions while also being able to apply a considerable load to the network.

### How it works

The package is designed in a way to make it easy to compose new fuzzing scenarios, removing the complexity around account nonces and balances, signing, encoding and broadcasting. The basic element is a `Sequence`. This is an interface that represents a pattern of recursive messages (one could be I send 5 dollars to you and you send 5 dollars back and so forth). Each sequence is provisioned a set of accounts that are responsible for formulating the messages. The accounts are isolated, allowing for a single sequence to be replicated hundreds of times. Sequences are also provided access to the current state as well as a deterministic source of randomness for constructing their output messages.

Remaining work:
- Create a command line version of the tool
- Look to brainstorm useful sequences such as goverance proposals and authz.

### DIFF
--- a/testing/txsim/account.go
+++ b/testing/txsim/account.go
@@ -1,0 +1,442 @@
+package txsim
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/celestiaorg/celestia-app/app"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/tx/signing"
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+	auth "github.com/cosmos/cosmos-sdk/x/auth/types"
+	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/cosmos/cosmos-sdk/x/feegrant"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	gasLimit  = 1000000
+	feeAmount = 1000
+)
+
+type AccountManager struct {
+	mtx           sync.Mutex
+	keys          keyring.Keyring
+	masterAccount *Account
+	accounts      map[string]*Account
+	pending       []*Account
+	tx            *TxClient
+	query         *QueryClient
+}
+
+type Account struct {
+	Address       types.AccAddress
+	PubKey        cryptotypes.PubKey
+	Sequence      uint64
+	AccountNumber uint64
+	Balance       int64
+}
+
+func NewAccountManager(ctx context.Context, keys keyring.Keyring, txClient *TxClient, queryClient *QueryClient) (*AccountManager, error) {
+	records, err := keys.List()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(records) == 0 {
+		return nil, fmt.Errorf("no accounts found in keyring")
+	}
+
+	am := &AccountManager{
+		keys:     keys,
+		accounts: make(map[string]*Account),
+		pending:  make([]*Account, 0),
+		tx:       txClient,
+		query:    queryClient,
+	}
+
+	if err := am.setupMasterAccount(ctx); err != nil {
+		return nil, err
+	}
+
+	log.Info().
+		Str("address", am.masterAccount.Address.String()).
+		Int64("balance", am.masterAccount.Balance).
+		Msg("set master account")
+	am.accounts[am.masterAccount.Address.String()] = am.masterAccount
+
+	return am, nil
+}
+
+// setupMasterAccount loops through all accounts in the keyring and picks out the one with
+// the highest balance as the master account. Accounts that don't yet exist on chain are
+// ignored.
+func (am *AccountManager) setupMasterAccount(ctx context.Context) error {
+	records, err := am.keys.List()
+	if err != nil {
+		return err
+	}
+
+	for _, record := range records {
+		address, err := record.GetAddress()
+		if err != nil {
+			return fmt.Errorf("error getting address for account %s: %w", record.Name, err)
+		}
+
+		// search for the account on chain
+		balance, err := am.getBalance(ctx, address)
+		if err != nil {
+			log.Err(err).Str("account", record.Name).Msg("error getting initial account balance")
+			continue
+		}
+
+		// the master account is the account with the highest balance
+		if am.masterAccount == nil || balance > am.masterAccount.Balance {
+			accountNumber, sequence, err := am.getAccountDetails(ctx, address)
+			if err != nil {
+				log.Err(err).Str("account", record.Name).Msg("error getting initial account details")
+				continue
+			}
+			pk, err := record.GetPubKey()
+			if err != nil {
+				return fmt.Errorf("error getting public key for account %s: %w", record.Name, err)
+			}
+			am.masterAccount = &Account{
+				Address:       address,
+				PubKey:        pk,
+				Sequence:      sequence,
+				AccountNumber: accountNumber,
+				Balance:       balance,
+			}
+		}
+	}
+
+	if am.masterAccount == nil {
+		return fmt.Errorf("no suitable account found")
+	}
+
+	return nil
+}
+
+// AllocateAccounts is used by sequences to specify the number of accounts
+// and the balance of each of those accounts. Not concurrently safe.
+func (am *AccountManager) AllocateAccounts(n, balance int) []types.AccAddress {
+	if n < 1 {
+		panic("n must be greater than 0")
+	}
+	if balance < 1 {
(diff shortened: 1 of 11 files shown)


cmwaters:
Looks like there's some data race occurring but I'm not sure if it's actually triggered from the new code I wrote

testing/txsim/run_test.go
```diff
@@ -0,0 +1,152 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	blob "github.com/celestiaorg/celestia-app/x/blob/types"
+	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
+	distribution "github.com/cosmos/cosmos-sdk/x/distribution/types"
+	staking "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTxSimulator(t *testing.T) {
```
SweeXordious, line 22: probably just add here:
```
if t.Short() {
		t.Skip("skipping tx simulator test in short mode.")
	}
```
so the CI passes

testing/txsim/run_test.go
```diff
@@ -0,0 +1,152 @@
...
+		testnode.DefaultParams(),
+		tmCfg,
+		true,
+		genesis,
+		keyring,
+		"testnet",
+	)
+	require.NoError(t, err)
+
+	cctx, stopNode, err := testnode.StartNode(node, cctx)
```
SweeXordious, line 135: This is where the race is happening most likely,

testing/txsim/client.go
```diff
@@ -0,0 +1,277 @@
...
+
+		return resp, nil
+	}
+	return nil, errTimedOutWaitingForTx
+}
+
+// Client multiplexes the RPC clients
+func (tc *TxClient) Client() *http.HTTP {
+	tc.mtx.Lock()
+	defer tc.mtx.Unlock()
```
evan-forbes, line 191: definitely not blocking, but perhaps if we use read mutexes here, we can be more consistent when mutex usage when modifying local state, such as in the next method

testing/txsim/account.go
```diff
@@ -0,0 +1,441 @@
...
+	// create the tx builder and add the messages
+	builder := am.tx.Tx()
+	err := builder.SetMsgs(op.Msgs...)
+	if err != nil {
+		return fmt.Errorf("error setting messages: %w", err)
+	}
+
+	builder.SetFeeAmount(types.NewCoins(types.NewInt64Coin(app.BondDenom, feeAmount)))
+	// the master account is responsible for paying the fees
+	builder.SetFeeGranter(am.masterAccount.Address)
```
evan-forbes, line 179: is there a benefit to using fee granter for every tx instead of funding the accounts in some other mechansim (either by including them in the genesis via the existing tooling in testnode or just sending them funds)?

testing/txsim/run_test.go
```diff
@@ -0,0 +1,156 @@
...
+					t.Errorf("missing %d messages of type %s (blocks: %d)", count, msg, len(blocks))
+				}
+			}
+		})
+	}
+}
+
+func Setup(t *testing.T) (keyring.Keyring, string, string) {
+	t.Helper()
+	genesis, keyring, err := testnode.DefaultGenesisState()
```
evan-forbes, line 120: [not blocking]
the funding of the master account is very implicit atm, imo we should document that here or fund an explicit account ahead of time

evan-forbes (commented):
mostly non blocking comments and questions

ideally we can use the testing framework here to replace the [integration test](https://github.com/celestiaorg/celestia-app/blob/main/app/test/integration_test.go) (and the [smoke test](https://github.com/celestiaorg/celestia-app/blob/main/app/test/std_sdk_test.go))

I think it would also help to simply clarify the different testing directories, such as testutil, testing, and the test directory in `app`. 

related to 
#832 #1162 #829

testing/txsim/run_test.go
cmwaters: Good Idea. I will write something up

testing/txsim/client.go
```diff
@@ -0,0 +1,277 @@
...
+	defer tc.next()
```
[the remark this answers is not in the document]
evan-forbes: oh cool wow TIL

evan-forbes (approved):
LGTM! :+1: :+1: can't wait to write some sequences!

testing/txsim/send.go
```diff
@@ -0,0 +1,55 @@
...
+var _ Sequence = &SendSequence{}
+
+// SendSequence sets up an endless sequence of send transactions, moving tokens
+// between a set of accounts
+type SendSequence struct {
+	numAccounts    int
+	sendAmount     int
+	maxHeightDelay int
+	accounts       []types.AccAddress
+	sequence       int
```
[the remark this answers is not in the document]
rootulp: or rename variable to disambiguate

testing/txsim/blob.go
```diff
@@ -0,0 +1,102 @@
...
+	return Operation{
+		Msgs:  []types.Msg{msg},
+		Blobs: blobs,
+	}, nil
+}
+
+type Range struct {
+	Min int
+	Max int
+}
```
rootulp, line 91: 👍

testing/txsim/blob.go
```diff
@@ -0,0 +1,102 @@
...
+		if s.namespace.ID != nil {
+			namespaces[i] = s.namespace
+		} else {
+			// generate a random namespace for the blob
+			namespace := make([]byte, ns.NamespaceVersionZeroIDSize)
+			_, err := rand.Read(namespace)
+			if err != nil {
+				return Operation{}, fmt.Errorf("generating random namespace: %w", err)
+			}
+			namespaces[i] = ns.MustNewV0(namespace)
```
rootulp, line 71: [optional] can use https://github.com/celestiaorg/celestia-app/blob/dd19531cb4909f7df0058be5300ba8e4eb969a63/pkg/namespace/random_blob.go#L11

testing/txsim/blob.go
```diff
@@ -0,0 +1,102 @@
...
+type Range struct {
+	Min int
+	Max int
+}
+
+func NewRange(min, max int) Range {
+	return Range{Min: min, Max: max}
+}
+
+func (r Range) Rand(rand *rand.Rand) int {
```
rootulp, line 97: [very optional]

```suggestion
// Rand returns a random number between min (inclusive) and max (exclusive).
func (r Range) Rand(rand *rand.Rand) int {
```

verified empirically in https://go.dev/play/p/8vBYgpJULux

testing/txsim/client.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/tendermint/tendermint/types"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// how often to poll the network for the latest height
+const (
+	DefaultPollTime = 3 * time.Second
+	maxRetries      = 20
+)
```
rootulp, line 28: [nit] if more variables are added to the const block, it may not be immediately clear which const this comment refers to

```suggestion
const (
	// DefaultPollTime is how often to poll the network for the latest height
	DefaultPollTime = 3 * time.Second
	maxRetries      = 20
)
```

rootulp (approved):
Overall LGTM, really cool testing tool

testing/txsim/account.go
cmwaters: It's a valid point. Perhaps it does make the most sense if we leak that complexity to the sequence developers so that they have to over fund their accounts and manage the gas limits (we can hardcode the gas price to the minimum)

testing/txsim/blob.go
cmwaters: Thanks but I want to use the random source provided in the method so that by having a specific seed it should always reproduce the same sequence

evan-forbes (approved):

testutil/testnode/node_interaction_api.go
```diff
@@ -177,7 +177,7 @@ func (c *Context) FillBlock(squareSize int, accounts []string, broadcastMode str
 	// in order to get the square size that we want, we need to fill half the
 	// square minus a few for the tx (see the square estimation logic in
 	// app/estimate_square_size.go)
-	shareCount := (squareSize * squareSize / 2) - 1
+	shareCount := (squareSize * squareSize / 2) - 2
```
rootulp, line 180: [out of curiosity] why did the size of one transaction occupy change from 1 to 2 shares? Did this PR fix a bug?

rootulp (approved):

testutil/testnode/node_interaction_api.go
cmwaters: > Did this PR fix a bug?

Yes it added math.Ceil to this line because our uint64 conversion was rounding down the number meaning occasionally we estimated too small of a square: https://github.com/celestiaorg/celestia-app/blob/11374a6953db1f15454621c4c195db197138927f/app/estimate_square_size.go#L37

I was thinking of extracting it out into a separate PR but we're about to overwrite this method entirely with ADR020

Issue #1081, opened 2022-11-30
Consider removing `EvidenceNamespace`


rootulp:
## Context

See https://github.com/celestiaorg/celestia-app/issues/1068#issuecomment-1331330996

Evidence was removed from the data square so we no longer need a reserved namespace for it. If we choose to remove the const in the `appconsts` package, we also need to update the specs.

evan-forbes:
we should definitely still be able to do this, and we should before mainnet imo

PR #1619, opened 2023-04-11
chore: remove evidence namespace

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1081


rootulp:
idk why some CI tasks was cancelled

evan-forbes (approved):

cmwaters:
Is it likely that the evidence namespace will come back?

rootulp:
> Is it likely that the evidence namespace will come back?

My impression is that evidence was originally moved from the `Block` struct to the data square to enable light clients to sample it rather than download it entirely. From in-person conversations during the Paris on-site, it seems like Evidence is small so we assume that light clients will be able to download it entirely. I think it's unlikely we move it back from the `Block` to the data square unless we observe that it is too large for light clients to download.

PR #1627, opened 2023-04-13
feat!: max namespace version for parity namespace

rootulp:
Re-apply https://github.com/celestiaorg/celestia-app/pull/1590 against `main`


x/blob/types/payforblob_test.go
```diff
@@ -172,7 +172,7 @@ func TestValidateBasic(t *testing.T) {
 		{
 			name:    "parity shares namespace id",
 			msg:     paritySharesMsg,
-			wantErr: ErrParitySharesNamespace,
+			wantErr: ErrInvalidNamespace,
```
cmwaters, line 175: Do we have a test still to check that `ErrPartityShareNamespace` is returned when expected?

cmwaters (approved):
Other than my comment and the failing linter, everything else looks good

evan-forbes (approved):

x/blob/types/payforblob_test.go
rootulp: It should no longer be returned https://github.com/celestiaorg/celestia-app/pull/1590#discussion_r1165765160

PR #1629, opened 2023-04-14
feat!: remove token voting upgrades

### LINKED ISSUE #1571: Remove the upgrade module
per #1491 , we will be removing token voting by removing the upgrade module.

evan-forbes:
## Overview

wraps the sdk's standard upgrade module to remove the ability to submit upgrade proposals. The idea being that this maintains compatibility with the IBC module (which requires access to the upgrade module's keeper), while still removing the ability for the upgrade module to work by not registering the upgrade module's msg server or begin block. Also, when we implement #1014 we can progressively flesh this module out with our own logic.

closes #1571 

## Checklist


x/upgrade/module.go
```diff
@@ -0,0 +1,96 @@
...
+func (am AppModule) RegisterServices(cfg module.Configurator) {
+}
+
+// ConsensusVersion implements AppModule/ConsensusVersion.
+func (AppModule) ConsensusVersion() uint64 { return consensusVersion }
+
+// BeginBlock overwrites the default upgrade module's BeginBlock method with a
+// noop.
+func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
+}
```
evan-forbes, line 96: Here's the most important state machine changes. By wrapping the upgrade module with our own and then overwriting these methods, we're removing any of the external functionality that's in the msg server/begin block, while preserving keeper access to the upgrade module to maintain 100% functionality with the IBC module.

x/upgrade/test/integration_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/tendermint/tendermint/config"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func TestUpgrade(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping SDK integration test in short mode.")
+	}
+	suite.Run(t, new(UpgradeTestSuite))
+}
```
evan-forbes, line 33: This integration test suite is probably overkill, but at least we have a very very high degree of confidence that submitting governance proposals to upgrade won't work.

there might be another test we could add where we manually force an upgrade into the state and check that it doesn't get picked up by begin block, but I opted to avoid that for now. If we decide that its worth it, I'm happy to add it!

x/upgrade/test/integration_test.go
```diff
@@ -0,0 +1,202 @@
...
+	})
+
+	msg, err := v1beta1.NewMsgSubmitProposal(sftwr, dep, accAddr)
+	require.NoError(t, err)
+
+	// submit the transaction and wait a block for it to be included
+	res, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, acc, msg)
+	require.NoError(t, err)
+
+	// require.Equal(t, abci.CodeTypeOK, res.Code) // we're only submitting the tx, so we expect everything to work
```
SweeXordious, line 140: Why is this commented? Shouldn't this also pass?

x/upgrade/module.go
SweeXordious: Don't we need to also overwrite `endBlock`?

evan-forbes: I thought the same, but I don't think the upgrade module has an endBlock

x/upgrade/test/integration_test.go
evan-forbes: good catch! https://github.com/celestiaorg/celestia-app/pull/1629/commits/62f6a5fb1c579139c81970363dc8e48f8508e893

I think it was just a casuality of the debug process

x/upgrade/module.go
SweeXordious: my bad, confused it with `gov` module :D

SweeXordious (approved):
LGTM

liamsi:
> The idea being that this maintains compatibility with the IBC module (which requires access to the upgrade module's keeper)

Can you link to that code or spec? Just curious what the IBC module does with the upgrade module's keeper. I assume it is to upgrade ibc clients?

app/app.go
```diff
@@ -218,7 +218,7 @@ type App struct {
 	DistrKeeper      distrkeeper.Keeper
 	GovKeeper        govkeeper.Keeper
 	CrisisKeeper     crisiskeeper.Keeper
-	UpgradeKeeper    upgradekeeper.Keeper
+	UpgradeKeeper    sdkupgradekeeper.Keeper
```
evan-forbes, line 221: I should also note that we are still selectively using components of the sdk's upgrade module, hence the change in naming to hopefully help differentiate the two. For example, see line 152, where we are using the appupgrade module instead of the sdk's

evan-forbes:
cc @liamsi 
>Can you link to that code or spec? Just curious what the IBC module does with the upgrade module's keeper. I assume it is to upgrade ibc clients?

most of the usage afaik is here in the beginblock
https://github.com/cosmos/ibc-go/blob/d7da1b7c5e86648d8142c948ba3d6dd9194d3bf9/modules/core/02-client/abci.go#L11-L42

my current thinking is that we can reuse a lot of the upgrade module as we flesh out #1014. Like fulfilling this same api, and things like `upgradetypes.Plan`/migrations.

x/upgrade/cli.go
```diff
@@ -0,0 +1,32 @@
...
+
+// GetTxCmd returns the transaction commands for this module
+func GetTxCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   upgradetypes.ModuleName,
+		Short: "Upgrade transaction subcommands",
+	}
+
+	return cmd
+}
```
cmwaters, line 18: Do we need this?

cmwaters:
> most of the usage afaik is here in the beginblock
https://github.com/cosmos/ibc-go/blob/d7da1b7c5e86648d8142c948ba3d6dd9194d3bf9/modules/core/02-client/abci.go#L11-L42

What about here: https://github.com/cosmos/ibc-go/blob/9e4b0d2f99c18db80db17697ec0875c48d84f328/modules/core/02-client/keeper/proposal.go#L73

Am I correct in thinking that IBC also has a handler that can kick off an upgrade?

evan-forbes:
> Am I correct in thinking that IBC also has a handler that can kick off an upgrade?

yeah it appears that that is used when other chains upgrade. For some reason, some of the ibc client state is stored in the upgrade keeper, which is a bit confusing since the ibc module is the only one that writes to that state.

x/upgrade/module.go
evan-forbes: > Could we just keep the keeper and not the module?

this is a good idea! I think this would be the simplest approach, but I think we still need some way to register some of the upgrade types.

for example when the governance can [upgrade an IBC clients](https://github.com/celestiaorg/celestia-app/blob/c5ae8ac2d7f2d290f336faf170c7547bfd0929a2/app/app.go#L358)
```go
// NewSoftwareUpgradeProposalHandler creates a governance handler to manage new proposal types.
// It enables SoftwareUpgradeProposal to propose an Upgrade, and CancelSoftwareUpgradeProposal
// to abort a previously voted upgrade.
//
//nolint:staticcheck // we are intentionally using a deprecated proposal here.
func NewSoftwareUpgradeProposalHandler(k *keeper.Keeper) govtypes.Handler {
	return func(ctx sdk.Context, content govtypes.Content) error {
		switch c := content.(type) {
		case *types.SoftwareUpgradeProposal: <--
			return handleSoftwareUpgradeProposal(ctx, k, c)

		case *types.CancelSoftwareUpgradeProposal: <--
			return handleCancelSoftwareUpgradeProposal(ctx, k, c)

		default:
			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "unrecognized software upgrade proposal content type: %T", c)
		}
	}
}
```

x/upgrade/cli.go
evan-forbes: only if we decide that we do need the app module, since the [`AppModuleBasic` interface returns a pointer](https://github.com/celestiaorg/cosmos-sdk/blob/f19e20459b8bcf195167ad256de5a6ca6edc41f5/types/module/module.go#L58)

damiannolan:
> > Am I correct in thinking that IBC also has a handler that can kick off an upgrade?
> 
> yeah it appears that that is used when other chains upgrade. For some reason, some of the ibc client state is stored in the upgrade keeper, which is a bit confusing since the ibc module is the only one that writes to that state.

This is currently the only upgrade path for ibc client upgrades. Historically the sdk upgrade plan used to contain a [proto Any field](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/upgrade/v1beta1/upgrade.proto#L40-L43) which has since been deprecated when x/ibc was migrated to the ibc-go repo. 

The proposal handler in ibc-go essentially uses the upgrade keeper to schedule the upgrade plan and store the upgraded client state under an "upgrade path". The path currently being used as the standard/default is under the x/upgrade module at `upgrade/UpgradedIBCState/{upgradeHeight}/upgradedClient`. Afaik this can be changed but may be a little cumbersome. Essentially its a bit of technical debt from when ibc existed within the sdk. 

cc. @AdityaSripal

app/app.go
```diff
@@ -159,7 +158,7 @@ var (
 
 	// ModuleEncodingRegisters keeps track of all the module methods needed to
 	// register interfaces and specific type to encoding config
-	ModuleEncodingRegisters = moduleMapToSlice(ModuleBasics)
+	ModuleEncodingRegisters = extractRegisters(ModuleBasics, appupgrade.TypeRegister{})
```
evan-forbes, line 161: followed through with @cmwaters 's idea to remove the module all together, and now we're just manually registering the upgrade module's types. 

This is necessary for the IBC module (https://github.com/celestiaorg/celestia-app/pull/1629#discussion_r1170264709)

x/upgrade/test/integration_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/config"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func TestUpgrade(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping SDK integration test in short mode.")
```
rootulp, line 30: [question] do we want to differentiate this log line from https://github.com/rootulp/celestia-app/blob/516ac26aae4db6bbb71e2559297622bf04571c4c/app/test/std_sdk_test.go#L29
```suggestion
		t.Skip("skipping x/upgrade SDK integration test in short mode.")
```

rootulp (dismissed):
utACK. Do we want a `x/upgrade/README.md` that describes the diff between this repo's upgrade module and the standard SDK upgrade module?

evan-forbes:
@rootulp 

good point, added a short one, since this module doesn't do much. Perhaps we can add to if we end up doing anything else with it.

testutil/testnode/query.go
```diff
@@ -0,0 +1,23 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	node, err := clientCtx.GetNode()
+	if err != nil {
+		return nil, err
+	}
+
+	return node.Tx(context.Background(), hash, prove)
```
cmwaters, line 22: Does `clientCtx` have an underlying go `Context`?

x/upgrade/READ_ME.md
```diff
@@ -0,0 +1,8 @@
+# `x/upgrade`
```
cmwaters, line 1: I think usually `READ_ME.md` is `README`

cmwaters (dismissed):
LGTM

testutil/testnode/query.go
evan-forbes: hmm I thought so too, but myself and my IDE could not find it? maybe some other Contetx does and we should refactor to that?

https://github.com/celestiaorg/cosmos-sdk/blob/50f4c49034ee97424a43b32e77715f9c8413bb5d/client/context.go#L25-L63

cmwaters (approved):

x/upgrade/README.md
```diff
@@ -0,0 +1,8 @@
+# `x/upgrade`
+
+## Abstract
+
+The upgrade module removes the entrypoints to the standard upgrade module by not
+registering a message server. It registers the standard upgrade module types to
+preserve the ability to marshal them. Note that the keeper of the standard
+upgrade module is still added to the application.
```
rootulp, line 8: 👍

rootulp (approved):

PR #1633, opened 2023-04-18
docs: ADR020 Deterministic Square Construction

cmwaters:
This ADR records the decision to make square construction follow a set of mandatory rules upheld by consensus. These mandatory rules mean that the square layout is deterministic based on an ordered array of transactions. The ADR also includes improvements to square size estimation. 

[ADR name also up for naming suggestions :)]

[Rendered](https://github.com/celestiaorg/celestia-app/blob/cal/adr/docs/architecture/adr-020-deterministic-square-construction.md)


docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+
+Proposed
+
+## Changelog
+
+- 2023/4/13: initial draft
+
+## Context
+
+The current protocol around the construction of a square (ODS) is based around a set of constraints that are enforced during consensus through validation (See `ProcessProposal`). Block proposers are at liberty to choosing not only what transactions are included and in what order but can effectively decide on the amount of padding (i.e. where each blob is located in the square) and the size of the square. This degree of control leaks needless complexity to users with little upside and allows for adverse behaviour.
```
rootulp, line 13:
```suggestion
The current protocol around the construction of an original data square (ODS) is based around a set of constraints that are enforced during consensus through validation (See `ProcessProposal`). Block proposers are at liberty to choosing not only what transactions are included and in what order but can effectively decide on the amount of padding (i.e. where each blob is located in the square) and the size of the square. This degree of control leaks needless complexity to users with little upside and allows for adverse behaviour.
```

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+Square construction is thus to be reduced to the simple deterministic function:
+
+```go
+func ConstructSquare(txs []Tx) []Share
+```
+
+and it's couterpart
+
+```go
+func DeconstructSquare(shares []Share) []Tx
```
rootulp, line 30: Note: the return type here `[]Tx` isn't identical to the `[]Tx` parameter in `ConstructSquare(txs []Tx)` because of index wrapping and separation of blobs.

Should this proposed API elaborate on the `[]Tx` type in each function?

rootulp: On second thought, the index wrapped transactions don't contain the blobs but the blob transactions do so `DeconstructSquare` may also want to return a slice of Blobs 🤔

rootulp (approved):
No blocking feedback so LGTM. I would appreciate greater clarity on if this ADR proposes any changes to the underlying `[]Tx` types (e.g. blob tx, index wrapped tx). I assume no but then I think the proposed API may need to be modified

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+Block proposers need not know nor care about the internals of square construction. With gradual improvements to construction techniques that best incorporate the tradeoffs between square size (impacting bandwidth and storage) with proof size (impacting light clients), the latest culmination in [ADR013](./adr-013-non-interactive-default-rules-for-zero-padding.md), there is little room for validators to implement optimizations. Such existing control to block proposers also means they are able to spam the network; building full size squares that include a single PFB. Lastly, a deterministic construction mechanism has the benefit of a more compact representation of the data that can freely move between different representations (this also supports a simpler compact blocks protocol):
+
+```go
+[]Tx <=> []Share <=> EDS
```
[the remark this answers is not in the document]
cmwaters: Yes, would you like me to put ODS in brackets to better distinguish?

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+## Detailed Design
+
+A new pkg `square` will be responsible for housing the square construction logic. Any modifications to the algorithm must come as a coordinated upgrade and thus be mapped to the `AppVersion`.
+
+Squares are bound by an upper limit on the square size, ergo not all transactions can fit in a square. Given a prioritised ordering of transactions, each transaction is individually staged and if it fits is added to the pending square else it is dropped. Each addition of a transaction has the potential to shuffle up the order of the squares (since data is ordered by namespace not order of priority). Therefore, the design conservatively estimates how much padding the blob could have and tracks the total offset to understand whether it fits in the square. This is done for both compact shares and sparse shares (which follow different rules):
+
+For compact shares i.e. PFBs and other state transactions, we introduce a `CompactShareCounter` that tracks two fields (`shares` & `remainder`). It is centered around the `Add` method:
+
+```go
+func (c *CompactShareCounter) Add(dataLen int, remainingShareCapacity int) bool
```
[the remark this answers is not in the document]
cmwaters: The idea is that the method only adds the data to the total if it is less than `remainingShareCapacity`. Perhaps this could be better labelled.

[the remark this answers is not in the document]
cmwaters: I could just call it `shareCapacity`

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+If there is sufficient remaining capacity the `CompactShareCounter` will record the share and return true. The `remainder` field is used to record the remaining bytes in the ultimate share.
```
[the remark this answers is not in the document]
cmwaters: These are meant as two different things. The first a field within the struct, the latter an argument in the method

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+
+With these two structs, we can safely guarantee that all staged transactions can be included in the square and from the estimation can formulate the minimum square size that will be used to calculate the share index for all the blobs. The construction of the square itself remains the same. Additionally, as the share index and square size can be deterministically calculated it no longer needs to be gossiped to all nodes in the consensus network. However, for verifiability purposes the square size should still remain in the `Data` struct.
+
+Both `PrepareProposal` and `ProcessProposal` will as a result, call much the same methods. Verification is thus reduced to: did I create the same square as you, rather than is your version of the square valid. The main difference is that `PrepareProposal` will handle overflow of transactions by discarding them, while `ProcessProposal` will handle overflow by rejecting the block.
+
+## Consequences
+
+### Positive
+
+- By staging transactions in order of priority, we prevent the removal of a higher priority transaction that had a higher namespace than a transaction with a lower namespace and lower priority.
```
cmwaters, line 66:
```suggestion
- By staging transactions in order of priority, we prevent the removal of a higher priority transaction that had a higher namespace than a transaction with a lower namespace and lower priority. [#1519](https://github.com/celestiaorg/celestia-app/issues/1519)
```

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+func DeconstructSquare(shares []Share) []Tx
+```
+
+Block proposers need not know nor care about the internals of square construction. With gradual improvements to construction techniques that best incorporate the tradeoffs between square size (impacting bandwidth and storage) with proof size (impacting light clients), the latest culmination in [ADR013](./adr-013-non-interactive-default-rules-for-zero-padding.md), there is little room for validators to implement optimizations. Such existing control to block proposers also means they are able to spam the network; building full size squares that include a single PFB. Lastly, a deterministic construction mechanism has the benefit of a more compact representation of the data that can freely move between different representations (this also supports a simpler compact blocks protocol):
+
+```go
+[]Tx <=> []Share <=> EDS
+```
+
+Validation in `ProcessProposal` is simplified to reconstructing the square given the transaction set and then evaluating whether the computed data root matches the proposed one.
```
evan-forbes, line 39: [not blocking comment]

mind elaborating on this difference a bit more and what is being simplified? are they doing something else atm? does this new approach remove the necessity for any other checks?

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+Square construction is thus to be reduced to the simple deterministic function:
+
+```go
+func ConstructSquare(txs []Tx) []Share
+```
+
+and it's couterpart
+
+```go
+func DeconstructSquare(shares []Share) []Tx
```
evan-forbes, line 30: [optional comment]

here or below might be a good place to compare the old implementation with this one a little bit more detail

from my understanding, this differs from the existing process by not splitting `BlobTxs` and not wrapping the transactions with an index ahead of time. We might want to provode more details here on what is changing from some existing [merge](https://github.com/celestiaorg/celestia-app/blob/952edaec409380d49ec0908f1c07a3134b70c05e/pkg/shares/merge.go#L9-L11)/[split](https://github.com/celestiaorg/celestia-app/blob/952edaec409380d49ec0908f1c07a3134b70c05e/pkg/shares/share_splitting.go#L23-L26)

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+
+```go
+[]Tx <=> []Share <=> EDS
+```
+
+Validation in `ProcessProposal` is simplified to reconstructing the square given the transaction set and then evaluating whether the computed data root matches the proposed one. 
+
+## Detailed Design
+
+A new pkg `square` will be responsible for housing the square construction logic. Any modifications to the algorithm must come as a coordinated upgrade and thus be mapped to the `AppVersion`.
```
evan-forbes, line 43: square pkg described here is a very very good idea imo. In the past, we've struggled with import cycles creating a difference package, since we require the encoding config from the app packagage (plus maybe something I'm forgetting), but if we could separate it out that would be super nice.

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+
+Both `PrepareProposal` and `ProcessProposal` will as a result, call much the same methods. Verification is thus reduced to: did I create the same square as you, rather than is your version of the square valid. The main difference is that `PrepareProposal` will handle overflow of transactions by discarding them, while `ProcessProposal` will handle overflow by rejecting the block.
+
+## Consequences
+
+### Positive
+
+- By staging transactions in order of priority, we prevent the removal of a higher priority transaction that had a higher namespace than a transaction with a lower namespace and lower priority. [#1519](https://github.com/celestiaorg/celestia-app/issues/1519)
+- Block proposers aren't able to spam the network as they no longer have control over the size of the square insofar as the amount of transactions included.
```
evan-forbes, line 67: [not blocking]
is this referring to proposers building large padded empty squares?

> aren't able to spam the network

not blocking, but we might be able to rephrase this to be a tad clearer. They still could spam the network with non-sense 0 fee txs, correct?

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,82 @@
...
+
+With these two structs, we can safely guarantee that all staged transactions can be included in the square and from the estimation can formulate the minimum square size that will be used to calculate the share index for all the blobs. The construction of the square itself remains the same. Additionally, as the share index and square size can be deterministically calculated it no longer needs to be gossiped to all nodes in the consensus network. However, for verifiability purposes the square size should still remain in the `Data` struct.
+
+Both `PrepareProposal` and `ProcessProposal` will as a result, call much the same methods. Verification is thus reduced to: did I create the same square as you, rather than is your version of the square valid. The main difference is that `PrepareProposal` will handle overflow of transactions by discarding them, while `ProcessProposal` will handle overflow by rejecting the block.
```
evan-forbes, line 60: [not blocking question/nit]
> Verification is thus reduced to: did I create the same square as you, rather than is your version of the square valid.

I'm still a bit confused on this distinction. Mainly
> rather than is your version of the square valid

I get that we don't have to check inclusion or share indexes because those things are deterministic, which is really really cool, but we still have to check for other validity conditions, correct? for example, the signatures of PFBs? like, if a set of transactions is provided to constructSquare that is a PFB and has an invalid tx, we have to reject, correct?

evan-forbes (dismissed):
no blocking feedback, since I think I understand what is being described here! However, I left some comments for a few things that we could optionally add some clarification to

docs/architecture/adr-020-deterministic-square-construction.md
evan-forbes: very related to https://github.com/celestiaorg/celestia-app/pull/1633#discussion_r1171347793, can just respond there

cmwaters: Yes, for example, you don't need to check that the blobs are ordered by namespace. You will still need to validate the `BlobTxs` 

[Will write this down]

docs/architecture/adr-020-deterministic-square-construction.md
cmwaters: Yes, I will try better distinguish between what verification is no longer needed and what still remains

cmwaters:
I have updated the ADR. I believe all the comments have been addressed

evan-forbes (approved):

docs/architecture/adr-020-deterministic-square-construction.md
```diff
@@ -0,0 +1,84 @@
...
```
rootulp, line 60: [no change needed] Given we don't plan on removing `SquareSize` from this struct:

```go
// Data contains the set of transactions included in the block
type Data struct {
	// Txs that will be applied by state @ block.Height+1.
	// NOTE: not all txs here are valid.  We're just agreeing on the order first.
	// This means that block.AppHash does not include these txs.
	Txs [][]byte `protobuf:"bytes,1,rep,name=txs,proto3" json:"txs,omitempty"`
	// field number 2 is reserved for intermediate state roots
	// field number 3 was previously used for evidence
	Blobs      []Blob `protobuf:"bytes,4,rep,name=blobs,proto3" json:"blobs"`
	SquareSize uint64 `protobuf:"varint,5,opt,name=square_size,json=squareSize,proto3" json:"square_size,omitempty"`
	Hash       []byte `protobuf:"bytes,6,opt,name=hash,proto3" json:"hash,omitempty"`
}
```

I assume ProcessProposal will need to verify that the proposed `SquareSize` == the validator's created `SquareSize`. I think that's inferred by "did I create the same square as you" but just clarifying

cmwaters: Yes, we will need to check that the `SquareSize` stated in the data struct equals to the one we generated

Issue #1594, opened 2023-04-03
Create a hardfork rolling upgrade for incentivized testnet


evan-forbes:
Part of the incnetivized testnet requires [participating in some form of hardfork](https://github.com/celestiaorg/itn-doc/blob/main/itn-vote-governance.md). We should create a special version of the app for this exact purpose, where we increment the app version at a hardcoded height, and use the app version to determine some state change. This could be a simple parameter change, or involve something larger.

Issue #1568, opened 2023-03-30
EPIC: Single Binary Syncs


evan-forbes:
We want to be able to sync from genesis using a single binary. This means that all past and present consensus logic needs to be maintained forever. The general thinking is that we will use the app version in the header to determine 

more details in the ADR #1521 

## Tasks
- [x] https://github.com/celestiaorg/celestia-app/issues/3105

rootulp:
@cmwaters is this close-able? I'm not sure which in-flight PRs were alluded to in my previous comment but I don't see any current in-flight PRs related to this.

PR #1640, opened 2023-04-21
feat!: Implement the version module

evan-forbes:
## Overview

This PR adds a very minimal version module, which will update the app version in the header using end block based on a config. The configs are just hardcoded maps atm, but its possible for a user to pass their own via app options. This PR also isn't meant to be a final solution for #1014 or #1568, but more of an exploratory demo of one way they could done.

the majority of #1594 

## Checklist


app/end_block.go
```diff
@@ -0,0 +1,14 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+// EndBlock wraps the BaseApp's end block method. This is done to modify the app
+// version if necessary.
+func (app *App) EndBlock(req abci.RequestEndBlock) (res abci.ResponseEndBlock) {
+	res = app.BaseApp.EndBlock(req)
+	ctx := app.GetContextForDeliverTx([]byte{1})
+	return appversion.EndBlocker(ctx, app.VersionKeeper, res)
+}
```
evan-forbes, line 14: we have to wrap end block because the base app doesn't actually have a way to modify the app version. It instead relies on the app version being set in `BaseApp`, and then returned during the `Info` call on boot. To avoid that, we modify the consensus param updates and pass them here.

x/version/abci.go
```diff
@@ -0,0 +1,22 @@
...
+func EndBlocker(ctx sdk.Context, keeper Keeper, resp abci.ResponseEndBlock) abci.ResponseEndBlock {
+	newAppVersion := keeper.GetVersion(ctx)
+	if ctx.BlockHeader().Version.App != newAppVersion {
+		resp.ConsensusParamUpdates.Version = &coretypes.VersionParams{
+			AppVersion: newAppVersion,
+		}
+		// set the version in the application to ensure that tendermint is
+		// passed the correct value upon rebooting
+		keeper.versionSetter.SetProtocolVersion(newAppVersion)
+	}
```
evan-forbes, line 20: we have to set the app version in the application, exactly what the upgrade module does, but we also have to update the consensus params in order to avoid having to reboot.

evan-forbes (commented):
this was meant as an exploration/hacky module for the incentivized testnet fork, but if we end up liking the mechanism, we can cherry-pick this to main as well.

app/end_block.go
cmwaters: I still don't quite get this. Can't you modify the app version in `ResponseEndBlock`?

cmwaters:
A point of clarification. Are we expecting to hardcode the upgrade heights in the binary or get the validators to set them?

evan-forbes:
@cmwaters note that this is targetting the v0.13.x branch, and is only for incentivized testnet atm. While we might want to reuse some of this for main, this PR is meant to explore what we could do and quickly get a rolling hardfork together.

> A point of clarification. Are we expecting to hardcode the upgrade heights in the binary or get the validators to set them?

the validators will eventually set them, although its a good point that we don't yet have that functionality

> Also a design question: why did you choose to have this separate from the modified upgrade module we have

the target branch doesn't have a modified upgrade module, so initially I don't think I wanted to wrap or copy paste fork the current upgrade module. This is a good point and we should just stick it in there, should we go to main.

cmwaters:
> the validators will eventually set them

If the validators set them in their config then why do we need it to be mapped with the chainID. Can't it just be: 

```
upgrades: { 10, 2 }
```

x/version/README.md
```diff
@@ -0,0 +1,103 @@
...
+```go
+vm := map[uint64]int64{
+		1: 0,
+		2: 112093,
+		3: 300442,
+		4: 420420,
+}
+```
+
+The application will convert this mapping into a HeightVersionConfig:
```
rootulp, line 76: ?

```suggestion
The application will convert this mapping into a HeightRange:
```

x/version/README.md
```diff
@@ -0,0 +1,103 @@
...
+custom mappings by passing including a `map[string]ChainVersionConfig` in the
+app options using the `version.CustomVersionConfigKey` key.
+
+Mappings are defined by a `map[uint64]int64` where the key is the app version
+and the value is the height at which the app version should be incremented. For
+example:
+
+```go
+vm := map[uint64]int64{
+		1: 0,
```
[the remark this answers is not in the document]
rootulp: I guess I'm confused on why the app version doesn't start at 0 and then increment to 1 at some future block. Is an example like the following more reflective of what we expect on BSR and mainnet:

```
vm := map[uint64]int64{
		0: 0,
		1: 1000,
}
```

but even this example would increment the app version from `0` to `1` at block height `0` so I don't think it's accurate.

x/version/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Usage
+
+The standard usage of the version module is to load the app version height
+mappings that are embedded into the binary, but it is also possible to load
+custom mappings by passing including a `map[string]ChainVersionConfig` in the
+app options using the `version.CustomVersionConfigKey` key.
+
+Mappings are defined by a `map[uint64]int64` where the key is the app version
+and the value is the height at which the app version should be incremented. For
```
rootulp, line 64: > the key is the app version
and the value is the height at which the app version should be incremented

A few questions on the mapping:
1. Should we enforce that keys of this map are monotonically increasing by 1? The map itself isn't ordered but maybe it should panic if there is a gap in app version keys?
2. Is the app version incremented to one above the key at the value?

rootulp: Disregard, from tests below I think the statement could be updated to

```suggestion
Mappings are defined by a `map[uint64]int64` where the key is the app version
and the value is the first block height for that app version. For
```

x/version/README.md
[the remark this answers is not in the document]
rootulp: Disregard, resolved by looking at tests below.

x/version/chains.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// StandardChainVersions returns a map of chain IDs to their respective
+// ChainVersionConfig. Each hardfork should be added to this map.
+func StandardChainVersions() map[string]ChainVersionConfig {
+	version0Only := NewChainVersionConfig(map[uint64]int64{
+		0: 0,
+	})
+	mainnetVersions := NewChainVersionConfig(map[uint64]int64{
+		1: 0,
```
rootulp, line 17: [question] why doesn't this look something like:

```
0: 0,
1: 1000, // replace 1000 with the first upgrade block height
```

In other words, why start mainnet at `appVersion = 1`?

x/version/version_height_mapping_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	vg := NewChainVersionConfig(input)
+	require.Equal(t, uint64(1), vg.GetVersion(0))
+	require.Equal(t, uint64(1), vg.GetVersion(1))
+	require.Equal(t, uint64(1), vg.GetVersion(9))
+	require.Equal(t, uint64(2), vg.GetVersion(10))
+	require.Equal(t, uint64(2), vg.GetVersion(11))
+	require.Equal(t, uint64(2), vg.GetVersion(19))
+	require.Equal(t, uint64(3), vg.GetVersion(20))
+	require.Equal(t, uint64(3), vg.GetVersion(math.MaxInt64))
```
rootulp, line 24: 👍

rootulp (approved):
[no blocking feedback] I was a bit confused by the version map but the tests helped clarify its intended usage

x/version/chains.go
evan-forbes: the app version can be arbitrary, so I'm assuming that it will match the versions in github as much as possible. Hence starting mainnet at version 1

app/end_block.go
cmwaters: Does it get overwritten somewhere later on?

cmd/celestia-appd/cmd/root.go
```diff
@@ -257,7 +257,6 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts se
 		appOpts,
 		baseapp.SetPruning(pruningOpts),
 		baseapp.SetMinGasPrices(cast.ToString(appOpts.Get(server.FlagMinGasPrices))),
-		baseapp.SetMinRetainBlocks(cast.ToUint64(appOpts.Get(server.FlagMinRetainBlocks))),
```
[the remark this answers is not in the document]
evan-forbes: ahh it was just a duplicate I noticed from line 262, sorry forgot to point this out

x/version/READ_ME.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+## State
+
+The state in the version module is abnormal because it can be modified by the
+party building the binary without breaking the app hash. This is because the
+state is not stored in the iavl tree, but rather as a simple predefined golang
+map. Its important to note that even though this state is not in the state
+store, that is can still cause consensus breaking changes. This is because that
```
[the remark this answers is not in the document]
evan-forbes: it wouldn't let me commit the suggested change cause its outdated, but was included in

e751399692a3d08774c84f894e453956a64b325f

x/version/READ_ME.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Events
+
+TODO: Add events
+
+### Usage
+
+The standard usage of the version module is to load the app version height
+mappings that are embedded into the binary, but it is also possible to load
+custom mappings by passing including a `map[string]ChainVersionConfig` in the
```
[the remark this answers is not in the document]
evan-forbes: it wouldn't let me commit the suggested change cause its outdated, but was included in

https://github.com/celestiaorg/celestia-app/commit/e751399692a3d08774c84f894e453956a64b325f

x/version/READ_ME.md
```diff
@@ -0,0 +1,103 @@
...
+```go
+vm := map[uint64]int64{
+		1: 0,
+		2: 112093,
+		3: 300442,
+		4: 420420,
+}
+```
+
+The application will convert this mapping into a HeightVersionConfig:
```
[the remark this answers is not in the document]
evan-forbes: it wouldn't let me commit the suggested change cause its outdated, but was included in

https://github.com/celestiaorg/celestia-app/commit/e751399692a3d08774c84f894e453956a64b325f

used `ChainVersionConfig` instead of Height range, because that's what it gets converted to, but let me know if that doesn't make sense

x/version/READ_ME.md
```diff
@@ -0,0 +1,103 @@
...
+
+```go
+// HeightRange is a range of heights that a version is valid for. It is an
+// internal struct used to search for the correct version given a height.
+type HeightRange struct {
+	Start   int64
+	End     int64
```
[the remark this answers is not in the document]
evan-forbes: good point, added some more details in https://github.com/celestiaorg/celestia-app/commit/e751399692a3d08774c84f894e453956a64b325f

app/end_block.go
evan-forbes: only the events are collected in [EndBlock](https://github.com/celestiaorg/cosmos-sdk/blob/50f4c49034ee97424a43b32e77715f9c8413bb5d/baseapp/abci.go#L205-L208), and [GetConsensusParams](https://github.com/celestiaorg/cosmos-sdk/blob/50f4c49034ee97424a43b32e77715f9c8413bb5d/baseapp/baseapp.go#L438-L469) never checks for version changes. 

changing the app version using `BaseApp` is entirely reliant upon setting the value via [`SetProtocolVersion`](https://github.com/celestiaorg/cosmos-sdk/blob/50f4c49034ee97424a43b32e77715f9c8413bb5d/baseapp/options.go#L113-L116), which isn't passed back to tendermint until `Info`, despite tendermint having the capability to update the version during `EndBlock`

evan-forbes: so yeah its overwritten. nothing happens because only the result of `GetConsensusParams` is respected, which doesn't check for app vesion

rootulp (approved):

app/end_block.go
cmwaters: Can't we pass the `SetProtocolVersion` function to the version module and just call it in the `EndBlock` :) - Just an idea (It's what the upgrade module does).

BTW, the SDK should really modify the response to hide the fields you can't change to avoid confusion

cmwaters: Ah I think I get what you mean. We have to set the app version in two different areas

x/version/version_height_mapping.go
```diff
@@ -0,0 +1,79 @@
...
+	// set the end of the last range to the max uint64 to cover all heights
+	ranges[len(ranges)-1].End = math.MaxInt64
+
+	return ranges
+}
+
+// getVersion returns the app version for a given block height. It performs a
+// binary search on the ranges slice.
+func getVersion(height int64, ranges []HeightRange) uint64 {
+	// Perform binary search on the ranges slice
```
cmwaters, line 65: lol, good job but I feel this is probably overkill. We might have 10 - 20 versions in our lifetime of this

x/version/README.md
```diff
@@ -0,0 +1,106 @@
...
+	return resp
+}
+```
+
+## State
+
+The state in the version module is abnormal because it can be modified by the
+party building the binary without breaking the app hash. This is because the
+state is not stored in the iavl tree, but rather as a simple predefined golang
+map. It's important to note that even though this state is not in the state
```
liamsi, line 42: Can you add a sentence where this map is persisted if not in the state? (even if that is explained in more detail below)

x/version/chains.go
```diff
@@ -0,0 +1,25 @@
...
+	})
+	return map[string]ChainVersionConfig{
+		MochaChainID:          version0Only,
+		ArabicaChainID:        version0Only,
+		BlockspaceraceChainID: version0Only,
```
cmwaters, line 22: So we should update this when we know the height for the upgrade

x/version/chains.go
```diff
@@ -0,0 +1,25 @@
...
+	ArabicaChainID        = "arabica-4"
+	BlockspaceraceChainID = "blockspacerace-0"
+	MainnetChainID        = "celestia-1"
+)
+
+// StandardChainVersions returns a map of chain IDs to their respective
+// ChainVersionConfig. Each hardfork should be added to this map.
+func StandardChainVersions() map[string]ChainVersionConfig {
+	version0Only := NewChainVersionConfig(map[uint64]int64{
+		0: 0,
```
cmwaters, line 14: This is `0` because the current network is set to `0` right?

cmwaters (approved):
SGTM.

Excited to try this out

liamsi (commented):
This is pretty cool! Would be curious to get some eyes from the core SDK team on this well.

evan-forbes:
merging now to not block, but will have a follow up PR for the extra sentence @liamsi

x/version/chains.go
evan-forbes: yeah, hopefully we can stay 1:1 with our major version for simplicity

x/version/abci.go
```diff
@@ -0,0 +1,22 @@
+package version
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	abci "github.com/tendermint/tendermint/abci/types"
+	coretypes "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+// EndBlocker will modify the app version if the current height is equal to
+// a predefined height at which the app version should be changed.
```
SweeXordious, line 10: Late comment to this.

Will this still be activated if more than 2/3rds of the validator set didn't swap the binary before the upgrade height?

evan-forbes: nodes that did upgrade will, and nodes that didn't won't. So if 2/3s don't, then they can continue producing blocks while the nodes that did upgrade will halt.

SweeXordious: Awesome :+1:

PR #1642, opened 2023-04-21
feat: introduce DefaultMinGasPrice to appconsts

cmwaters:
Uses a constant in th `appconsts` package for the `DefaultMinGasPrice` that users can import.


evan-forbes (approved):
Nice!

rootulp (approved):

PR #1653, opened 2023-04-24
chore: register blob msg server in module

evan-forbes:
## Overview

This is pretty standard to do in the sdk, not sure how we haven't done this yet given over two years of running the state machine and audits lol 

Everything was still occurring via the legacy sdk.Router, but still... 

## Checklist

### DIFF
--- a/x/blob/module.go
+++ b/x/blob/module.go
@@ -127,6 +127,7 @@ func (am AppModule) LegacyQuerierHandler(legacyQuerierCdc *codec.LegacyAmino) sd
 // RegisterServices registers a GRPC query service to respond to the
 // module-specific GRPC queries.
 func (am AppModule) RegisterServices(cfg module.Configurator) {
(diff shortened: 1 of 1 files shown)


SweeXordious (approved):
Nice catch. How did it work without having the querier registered?

rootulp (approved):

cmwaters (approved):

cmwaters:
> Nice catch. How did it work without having the querier registered?

The querier was registered. The msg server is just for transactions

SweeXordious:
> The msg server is just for transactions

Aaah I see, then how are we sending pay for blob transactions?

evan-forbes:
> Aaah I see, then how are we sending pay for blob transactions?

the sdk was using the legacy mechanism, which we're also registering

https://github.com/celestiaorg/cosmos-sdk/blob/50f4c49034ee97424a43b32e77715f9c8413bb5d/baseapp/baseapp.go#L783-L800

PR #1657, opened 2023-04-25
feat: create compact share counter

cmwaters:
As part of ADR 020, we need something to iteratively calculate how many shares both pfbs and txs would take up should they be added. This PR creates a `CompactShareCounter` to perform this job. It would have been also possible to extend the `CompactShareSplitter` to do this task but we would have needed to add the functionality to "undo" transactions (because they were too large) and this seemed like quite some functionality to add.

Ref: https://github.com/celestiaorg/celestia-app/issues/1214

## Checklist


evan-forbes (approved):
makes sense so far! LGTM

pkg/shares/counter.go
```diff
@@ -0,0 +1,82 @@
...
+	// Increment the data len by the varint that will prefix the data.
+	dataLen += DelimLen(uint64(dataLen))
+
+	// save a copy of the previous state
+	c.lastRemainder = c.remainder
+	c.lastShares = c.shares
+
+	// if this is the first share, calculate how much is taken up by dataLen
+	if c.shares == 0 {
+		if dataLen+c.remainder > appconsts.FirstCompactShareContentSize {
```
rootulp, line 32: [blocking] I'm having a tough time understanding why this isn't

```suggestion
		if dataLen >= appconsts.FirstCompactShareContentSize-c.remainder {
```

can you please explain the rationale? I think the suggestion matches line 43 which intuitively makes more sense to me

pkg/shares/counter_test.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+func TestCounterMatchesCompactShareSplitter(t *testing.T) {
+	testCases := []struct {
+		txs []coretypes.Tx
+	}{
+		{txs: []coretypes.Tx{newTx(120)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize - 2)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize - 1)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize), newTx(appconsts.ContinuationCompactShareContentSize - 4)}},
```
rootulp, line 23: [optional] more test cases that pass

```suggestion
		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize)}},
		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize + 1)}},
		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize), newTx(appconsts.ContinuationCompactShareContentSize - 4)}},
```

rootulp (commented):
LGTM after one blocking comment

pkg/shares/counter.go
cmwaters: Yes I think your rationale makes sense (to change from `>` to `>=`) because when it equals a new share you just want to increment the shares. I caught myself at the one below but didn't change it here. Good catch

evan-forbes (approved):

pkg/shares/counter_test.go
```diff
@@ -0,0 +1,74 @@
...
+		txs []coretypes.Tx
+	}{
+		{txs: []coretypes.Tx{newTx(120)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize - 2)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize - 1)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize + 1)}},
+		{txs: []coretypes.Tx{newTx(appconsts.FirstCompactShareContentSize), newTx(appconsts.ContinuationCompactShareContentSize - 4)}},
+		{txs: newTxs(1000, 100)},
+		{txs: newTxs(100, 1000)},
```
rootulp, line 27: [no change needed] the inputs to this test could also be fuzzed given we just need a random `dataLen`

rootulp (approved):

pkg/shares/counter_test.go
cmwaters: Yeah I can explore that in a follow up. I think these cover the edge cases that come to mind

PR #1659, opened 2023-04-26
feat: create square builder class

cmwaters:
Ref: #1214

This creates a builder struct for assembling squares from a (prioritised) list of transactions.

It also includes two functions `Construct` and `Reconstruct` as listed in ADR20


app/estimate_square_size_test.go
```diff
@@ -207,3 +207,11 @@ func Test_maxIndexWrapperOverhead(t *testing.T) {
...
 	}
 }
+
+func txsToBytes(txs coretypes.Txs) [][]byte {
+	e := make([][]byte, len(txs))
+	for i, tx := range txs {
+		e[i] = []byte(tx)
+	}
+	return e
+}
```
cmwaters, line 217: note: this was moved out from the shares package because no one was using it. IIRC there's a similar function in core we could just use

pkg/square/builder.go
```diff
@@ -0,0 +1,238 @@
...
+		// Imagine if, according to the share commitment rules, a transcation took up 11 shares
+		// and had the merkle mountain tree commitment of 4,4,2,1. The first part of the share commitment
+		// would then be something that spans 4 shares and has a depth of 1. The worst case padding
+		// would be if the last transaction had a share at leaf index 0. Thus three padding shares would
+		// be needed to start the transaction at index 4 and be aligned with the first commitment.
+		// Thus the rule is to take the subtreewidh of the share size and subtract 1.
+		//
+		// Note that the padding would actually belong to the namespace of the transaction before it, but
+		// this makes no difference to the total share size.
+		maxPadding: shares.SubTreeWidth(shareSize) - 1,
```
cmwaters, line 224: For the reviewer: Please make sure I have this assumption correct

pkg/shares/utils.go
```diff
@@ -16,39 +13,12 @@ func DelimLen(size uint64) int {
...
-			return nil, fmt.Errorf("share version %d is too large to be a uint8", blob.ShareVersion)
-		}
-		result[i] = coretypes.Blob{
-			NamespaceID:  blob.NamespaceId,
-			Data:         blob.Data,
-			ShareVersion: uint8(blob.ShareVersion),
-		}
-	}
-	return result, nil
-}
```
cmwaters, line 36: Note: these functions are either not used (and probably shoudn't be) or redundant

pkg/shares/utils.go
```diff
@@ -99,3 +69,26 @@ func ParseDelimiter(input []byte) (inputWithoutLenDelimiter []byte, unitLen uint
...
+// AvailableBytesFromSparseShares returns the maximum amount of bytes that could fit in `n` sparse shares
+func AvailableBytesFromSparseShares(n int) int {
+	if n <= 0 {
+		return 0
+	}
+	if n == 1 {
+		return appconsts.FirstSparseShareContentSize
+	}
+	return (n-1)*appconsts.ContinuationSparseShareContentSize + appconsts.FirstSparseShareContentSize
+}
```
cmwaters, line 94: Something that the node team might want to use

pkg/square/builder.go
```diff
@@ -0,0 +1,238 @@
...
+		//	              / \
+		//	             /   \
+		//	1           0     1
+		//	           /\     /\
+		//	2         0  1   2  3
+		//	         /\  /\ /\  /\
+		//	3       0 1 2 3 4 5 6 7
+		//
+		// Imagine if, according to the share commitment rules, a transcation took up 11 shares
+		// and had the merkle mountain tree commitment of 4,4,2,1. The first part of the share commitment
```
rootulp, line 216: I'm having a tough time relating this example to the ascii diagram above b/c it discusses a blob that occupies 11 shares but there are only 8 possible share indexes in the ASCII diagram

pkg/square/builder.go
cmwaters: > Blobs start at an index that is equal to a multiple of the blob length divided by MaxSquareSize rounded up.

Where do you see this in ADR013? I can't find it

cmwaters: Now we use `SubtreeRootThreshold` to work out the height in the tree of the highest merkle mountain and thus the width of the leaves

pkg/square/builder.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/pkg/namespace"
+	"github.com/celestiaorg/celestia-app/pkg/shares"
+	"github.com/tendermint/tendermint/pkg/consts"
+	coretypes "github.com/tendermint/tendermint/proto/tendermint/types"
+	core "github.com/tendermint/tendermint/types"
+)
+
+type Builder struct {
+	// upper bound on the total amount of shares that could fit in the biggest square
```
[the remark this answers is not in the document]
cmwaters: I've seen linters that do. Maybe ours isn't set with that configuration

pkg/square/builder.go
cmwaters: The 8 possible shares indexes are a snapshot in the tree. It's used to demonstrate that if the index of the last transaction ended at 1 (inclusive), and the SubtreeWidth() was 4, then the next one would start at index 4 and we would have 2 shares of padding

pkg/square/builder.go
```diff
@@ -0,0 +1,238 @@
...
+
+func (c *Builder) isEmpty() bool {
+	return c.txCounter.Size() == 0 && c.pfbCounter.Size() == 0
+}
+
+type element struct {
+	blob       core.Blob
+	pfbIndex   int
+	blobIndex  int
+	shareSize  int
```
evan-forbes, line 190: [not blocking naming thing]
is it possible to use a different word besides shareSize that describes what this is? This is the number of shares used correct? the shares size makes me think of https://github.com/celestiaorg/celestia-app/blob/f31ee58890b1f94815057b4bbd77d7cbe62172f4/pkg/appconsts/appconsts.go#L21

pkg/square/builder.go
```diff
@@ -0,0 +1,238 @@
...
+// AppendBlobTx attempts to allocate the blob transaction to the square. It returns false if there is not
+// enough space in the square to fit the transaction.
+func (c *Builder) AppendBlobTx(blobTx coretypes.BlobTx) bool {
+	iw := &coretypes.IndexWrapper{
+		Tx:           blobTx.Tx,
+		TypeId:       consts.ProtoIndexWrapperTypeID,
+		ShareIndexes: worstCaseShareIndexes(len(blobTx.Blobs), c.maxCapacity),
+	}
+	size := iw.Size()
+	pfbShareDiff := c.pfbCounter.Add(size)
```
evan-forbes, line 72: nice, this is a significant improvement to https://github.com/celestiaorg/celestia-app/blob/f31ee58890b1f94815057b4bbd77d7cbe62172f4/app/estimate_square_size.go#L74-L105

evan-forbes (approved):
nice work!! overall this is a really big improvement

mostly non-blocking naming things

pkg/square/builder.go
rootulp: > Where do you see this in ADR013? I can't find it

Sorry I should've linked directly to the line: https://github.com/celestiaorg/celestia-app/blob/f31ee58890b1f94815057b4bbd77d7cbe62172f4/docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md?plain=1#L27

evan-forbes:
I'm working on some changes locally for refactoring limiting the blocksize, and I this change pairs very well

pkg/square/builder.go
cmwaters: I copied this tree across from somewhere else just so I could explain why we have padding and why that formula gives us the maximum possible padding

cmwaters:
Thought I'd post some relatively crude results comparing the square size estimation algorithm before and after (each PFB is 512 bytes)

```
pfbs 1 | square size before 4 | square size after 2
pfbs 2 | square size before 4 | square size after 4
pfbs 4 | square size before 8 | square size after 4
pfbs 8 | square size before 8 | square size after 8
pfbs 16 | square size before 16 | square size after 8
pfbs 32 | square size before 16 | square size after 16
pfbs 64 | square size before 32 | square size after 16
pfbs 128 | square size before 32 | square size after 32
pfbs 256 | square size before 64 | square size after 32
pfbs 512 | square size before 64 | square size after 64
pfbs 1024 | square size before 128 | square size after 64
pfbs 2048 | square size before 128 | square size after 128
```

rootulp (approved):

Issue #1649, opened 2023-04-24
Make `ParseShares` skip padding shares


vgonkivs:
Currently, the function [ParseShares](https://github.com/celestiaorg/celestia-app/blob/8c1fcb5118db90ed6e618dd2d48a3b34335d9a2d/pkg/shares/parse.go#L38) converts raw shares into share sequences. However, when the function encounters a NamespacePaddingShare in the input slice of raw shares, it generates an error. Could you please modify the function to include a condition that skips over any NamespacePaddingShares in the raw share input, rather than generating an error?

PR #1662, opened 2023-04-26
feat!: add ignorePadding option to ParseShares

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1649

### DIFF
--- a/pkg/shares/share_sequence.go
+++ b/pkg/shares/share_sequence.go
@@ -50,6 +50,14 @@ func (s ShareSequence) validSequenceLen() error {
 	if len(s.Shares) == 0 {
 		return fmt.Errorf("invalid sequence length because share sequence %v has no shares", s)
 	}
+	isPadding, err := s.isPadding()
+	if err != nil {
+		return err
+	}
+	if isPadding {
+		return nil
+	}
+
 	firstShare := s.Shares[0]
 	sharesNeeded, err := numberOfSharesNeeded(firstShare)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/shares/share_sequence.go
```diff
@@ -62,6 +70,17 @@ func (s ShareSequence) validSequenceLen() error {
 	return nil
 }
 
+func (s ShareSequence) isPadding() (bool, error) {
+	if len(s.Shares) != 1 {
```
cmwaters, line 74: Is this not supposed to be. Can't a padding sequence be multiple shares? 
```suggestion
	if len(s.Shares) > 0 {
```

rootulp: A padding sequence can only be one share. 

Since there are no bits in the universal share prefix that indicate if a share is namespace padding, we abuse two existing fields to determine if a share is namespace padding. A namespace padding share has:
1. a sequence start indicator of `1` 
2. a sequenceLen of `0`

cmwaters: > A padding sequence can only be one share.

So tail padding is not considered a single sequence even although they would all have the same namespace

rootulp: Yep exactly.

cmwaters: Ok my (admittedly internal) definition of a sequence was a set of shares that all had the same namespace

rootulp: Yea it's definitely a bit confusing and that what you describe is the original idea for a sequence.

TBH I think padding sequences all being of length one is a hack that we should reconsider. I'm in favor of using some other signaling mechanism in the universal share prefix to identify namespace padding shares or padding shares in general.

evan-forbes (approved):

cmwaters (approved):

PR #1664, opened 2023-04-26
chore: upgrade to celestia-core v1.18.0-tm-v0.34.27

rootulp:
Upgrade to https://github.com/celestiaorg/celestia-core/releases/tag/v1.18.0-tm-v0.34.27
Motivated by a desire to pick up https://github.com/celestiaorg/celestia-core/pull/1002


pkg/appconsts/consensus_consts.go
```diff
@@ -4,5 +4,8 @@ import "time"
 
 const (
 	TimeoutPropose = time.Second * 10
-	TimeoutCommit  = time.Second * 10
+	// TargetHeightDuration is the intended block interval duration (i.e. the
+	// time between blocks). Note that this is a target because CometBFT does
+	// not guarantee a fixed block interval.
+	TargetHeightDuration = time.Second * 15
```
rootulp, line 10: @evan-forbes would appreciate your attention here

testing/e2e/pkg/setup.go
```diff
@@ -230,7 +230,7 @@ func MakeConfig(node *Node) (*config.Config, error) {
 	// FIXME: This values get overridden by the timeout consts in the app package.
 	// We should modify this if we want to quicken the time of the blocks.
 	cfg.Consensus.TimeoutPropose = 1000 * time.Millisecond
```
rootulp, line 232: Are these comments still relevant?

pkg/appconsts/consensus_consts.go
cmwaters: Yup this looks good 👍

test/util/testnode/full_node.go
```diff
@@ -141,7 +141,9 @@ func DefaultParams() *tmproto.ConsensusParams {
 
 func DefaultTendermintConfig() *config.Config {
 	tmCfg := config.DefaultConfig()
-	tmCfg.Consensus.TimeoutCommit = time.Millisecond * 300
+	// Reduce the target height duration so that blocks are produced faster
+	// during tests.
+	tmCfg.Consensus.TargetHeightDuration = time.Millisecond * 300
```
cmwaters, line 146: I hope that's not going to be too fast. I guess if consensus can't handle it, it won't be that bad

testing/e2e/pkg/setup.go
cmwaters: Yes

cmwaters (dismissed):

cmwaters (approved):

cmwaters (dismissed):

x/blob/types/blob_tx.go
```diff
@@ -109,6 +111,27 @@ func BlobTxSharesUsed(btx tmproto.BlobTx) int {
...
+		return core.Blob{}
+	}
+
+	if p.ShareVersion > math.MaxUint8 {
+		return core.Blob{}
+	}
+
+	if p.NamespaceVersion > math.MaxUint8 {
+		return core.Blob{}
+	}
```
evan-forbes, line 125: [non blocking question]
do we know the thinking around adding these? was it just copied over, or is it that we're more likely to notice empty blobs than an overflow?

evan-forbes (commented):
quick question

x/blob/types/blob_tx.go
rootulp: Around adding what: the function `BlobFromProto` or the checks for overflow? 

1. `BlobFromProto` was copied over b/c a recent change in celestia-app started using this exported type in celestia-core but a recent change in celestia-core removed it.
2. The checks for overflow are to catch overflows because the fields in the proto type can be larger than uint8

evan-forbes: I see the overflow, but I'm curious if we should instead throw an error, since returning an empty blob for an otherwise valid blob might cause a bug. I don't think we'd ever have issues if we only use this function in the way we are now, but I could see it becoming an issue is usage, by us or otherwise, changes.

rootulp: +100 returning an error seems safer

rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/1664/commits/14d4c7399debcb6f8164bfb0ae1e7ba41bc2b24a

evan-forbes (approved):

cmwaters (approved):

PR #1672, opened 2023-04-27
feat!: implement the BSR hardfork

### LINKED ISSUE #1594: Create a hardfork rolling upgrade for incentivized testnet

evan-forbes:
## Overview

This PR changes the `SignificantPowerDifferenceThreshold` qgb constant depending on the app version. Note that this PR is targeting only the v0.13.x branch.

closes #1594

## Checklist

### DIFF
--- a/testutil/testnode/full_node.go
+++ b/testutil/testnode/full_node.go
@@ -2,6 +2,8 @@ package testnode
 
 import (
 	"encoding/json"
+	"fmt"
+	"net"
 	"os"
 	"testing"
 	"time"
@@ -200,6 +202,9 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func() error
 
 	tmCfg := DefaultTendermintConfig()
 	tmCfg.Consensus.TimeoutCommit = blockTime
+	tmCfg.RPC.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
+	tmCfg.P2P.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
+	tmCfg.RPC.GRPCListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
 
 	genState, kr, err := DefaultGenesisState(accounts...)
 	require.NoError(t, err)
@@ -210,7 +215,11 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func() error
 	cctx, stopNode, err := StartNode(tmNode, cctx)
 	require.NoError(t, err)
 
-	cctx, cleanupGRPC, err := StartGRPCServer(app, DefaultAppConfig(), cctx)
+	appConf := DefaultAppConfig()
+	appConf.GRPC.Address = fmt.Sprintf("127.0.0.1:%d", getFreePort())
+	appConf.API.Address = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
+
+	cctx, cleanupGRPC, err := StartGRPCServer(app, appConf, cctx)
 	require.NoError(t, err)
 
 	return func() error {
@@ -220,4 +229,16 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func() error
 		}
 		return cleanupGRPC()
 	}, accounts, cctx
+}
+
+func getFreePort() int {
+	a, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	if err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close()
+			return l.Addr().(*net.TCPAddr).Port
+		}
+	}
+	panic("while getting free port: " + err.Error())
 }
--- a/testutil/testnode/full_node.go
+++ b/testutil/testnode/full_node.go
@@ -202,9 +202,9 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func() error
 
 	tmCfg := DefaultTendermintConfig()
 	tmCfg.Consensus.TimeoutCommit = blockTime
-	tmCfg.RPC.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
-	tmCfg.P2P.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
-	tmCfg.RPC.GRPCListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", getFreePort())
+	tmCfg.RPC.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", GetFreePort())
+	tmCfg.P2P.ListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", GetFreePort())
+	tmCfg.RPC.GRPCListenAddress = fmt.Sprintf("tcp://127.0.0.1:%d", GetFreePort())
 
 	genState, kr, err := DefaultGenesisState(accounts...)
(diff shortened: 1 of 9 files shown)


x/qgb/v1/consts.go
```diff
@@ -0,0 +1,5 @@
+package v1
+
+// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
+// that would need the creation of a new valset request.
+const SignificantPowerDifferenceThreshold = 0.005
```
evan-forbes, line 5: I'm not really sure how we want to organize all of these changes for main/mainnet. This was just one exploratory option.

x/version/chains.go
```diff
@@ -16,10 +16,14 @@ func StandardChainVersions() map[string]ChainVersionConfig {
 	mainnetVersions := NewChainVersionConfig(map[uint64]int64{
 		1: 0,
 	})
+	bsr := NewChainVersionConfig(map[uint64]int64{
+		0: 0,
+		1: 674000,
+	})
```
evan-forbes, line 22: probably could have been a separate PR, but this includes the upgrade height for BSR

x/qgb/abci.go
```diff
@@ -68,7 +64,10 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 			panic(sdkerrors.Wrap(err, "invalid latest valset members"))
 		}
 
-		significantPowerDiff = intCurrMembers.PowerDiff(*intLatestMembers) > SignificantPowerDifferenceThreshold
+		// change the threshold based on the app version
+		spdt := GetSignificantPowerDiffThreshold(ctx.BlockHeader().Version.App)
+
+		significantPowerDiff = intCurrMembers.PowerDiff(*intLatestMembers) > spdt
```
evan-forbes, line 70: this is the consensus breaking change

SweeXordious (dismissed):
:tada: :tada: Thanks a lot for taking care of this.
Approving given CI is passing

evan-forbes (commented):
we still need to test this using multiple validator nodes, both for upgrading and syncing, but the changes are working as intended

x/qgb/test/versions_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	abci "github.com/tendermint/tendermint/abci/types"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func TestQGBVersionIntegrationTestSuite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping SDK integration test in short mode.")
```
rootulp, line 29:
```suggestion
		t.Skip("skipping QGB version integration test in short mode.")
```

x/qgb/versions.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func GetSignificantPowerDiffThreshold(appVersion uint64) float64 {
+	switch appVersion {
+	case 0:
+		return v1beta1.SignificantPowerDifferenceThreshold
+	case 1:
+		return v1.SignificantPowerDifferenceThreshold
+	default:
+		return v1.SignificantPowerDifferenceThreshold
```
rootulp, line 15: [question] should this panic with something like "unsupported appVersion"

rootulp (approved):

x/qgb/versions.go
evan-forbes: this mechanism could be refactored to be clearer/simpler, we just need to make sure that we use v1 for all app versions greater than zero

x/qgb/v1/consts.go
SweeXordious: Currently, in the QGB state machine, a validator set change is triggered if more than 5% of the total staking power has changed since the last valset.
The issue is that in BSR, all validators have 1% stake, and there aren't any changes to the validator set. Thus, if they change their EVM addresses, which should trigger a new validator set change if their stake is more than 5%, they will need to wait for another 4 validators to do that or else they won't be able to sign attestations

SweeXordious: That change reduces the threshold to 0.5% so that every validator that changes its EVM address (because they lost its private key), will be able to sign right after that

x/qgb/versions.go
rootulp: > we just need to make sure that we use v1 for all app versions greater than zero

I think a v1 validator would use an incorrect significant power different threshold for a block produced with app version 2. Instead of allowing that calculation to take place and produce an incorrect result, why not error / panic here?

rootulp: I guess this isn't blocking b/c it's future proofing but it does mean that a v0 app would handle a v2 block incorrectly rather than panicing.

rootulp: Say we merge this PR as-is and cut a v0.13.0 release. celestia-app v0.13.0:

appVersion 0 => v0 SignificantPowerDifferenceThreshold
appVersion 1 => v1 SignificantPowerDifferenceThreshold
appVersion 2 => v1 SignificantPowerDifferenceThreshold

At some point in the future we introduce a new significant power difference threshold that should apply for appVersion 2. We cut a v1.0.0 release. celestia-app v1.0.0:

appVersion 0 => v0 SignificantPowerDifferenceThreshold
appVersion 1 => v1 SignificantPowerDifferenceThreshold
appVersion 2 => v2 SignificantPowerDifferenceThreshold

Then most of the network validators upgrades to the v1.0.0 binary but a few stragglers remain on the v0.13.0 binary. When a block height is reached where appVersion 2 should be used, the un-upgraded validators incorrectly use the v1 SignificantPowerDifferenceThreshold instead of `panic`. Unless there is a code block that get executed before this where they try to validate a v2 appVersion block and panic immediately but if that's the case, I think we should do the same here for clarity.

evan-forbes: 1) all un-upgraded nodes will reject blocks that use a different appversion then what they are using before they even execute any transactions.

2) if they panic, then for each upgrade, we have to change _every_ if statement switch to support the new version, but that doesn't actually give us any benefit due to point 1. 

if we need to change the constant after some number of app versions then we add
appVersion N => v2 SignificantPowerDifferenceThreshold

rootulp: > 1. all un-upgraded nodes will reject blocks that use a different appversion then what they are using before they even execute any transactions.

Good point. Given this, I don't think it's necessary to do the `panic` thing I was describing.

rootulp (approved):

testutil/testnode/full_node.go
```diff
@@ -221,3 +230,15 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (cleanup func() error
 		return cleanupGRPC()
 	}, accounts, cctx
 }
+
+func GetFreePort() int {
```
cmwaters, line 234: Don't we have this function somewhere already

x/qgb/v1/consts.go
cmwaters: Got it. Thanks for the clarification

cmwaters (approved):
Looks good

testutil/testnode/full_node.go
evan-forbes: yeah on main we do, cherry picked it over from main https://github.com/celestiaorg/celestia-app/pull/1672/commits/4eff397763a333250efeb4146187112ec8b46c88 and then exported to use it in the integration test

SweeXordious (approved):

PR #1675, opened 2023-04-29
Refactor variable name from len to length

Chirag018:
I have made some changes to the code in this pull request. The main change is that I have renamed the variable `len` to `length`. This is because `len` is a built-in function in Go, and using it as a variable name can cause confusion and potential issues down the line. It is best practice to use a different name, and `length` is a clear and descriptive alternative.


rootulp (approved):

evan-forbes (approved):
thanks @Chirag018 :slightly_smiling_face:

PR #1683, opened 2023-05-02
chore: enable tx indexing on dev scripts

rootulp:
While debugging PFB txs, I found it helpful to enable tx indexing for the local node that gets run via the scripts in `./scripts`. Totally optional PR.


evan-forbes (approved):
makes sense to me, I had to do the same

SweeXordious (approved):
+1, I always end up doing this when wanting to test QGB locally

Issue #1685, opened 2023-05-02
Update the data commitment creation mechanism to support window changes


SweeXordious:
As discussed under https://github.com/celestiaorg/celestia-app/issues/1676#issuecomment-1532231143, we will need to update the mechanism we use to create data commitments in order to support changing the data commitment window changes, which wiill be via gov proposals.

Issue #1676, opened 2023-05-01
Investigate changing the data commitment window param


SweeXordious:
Currently, we're exposing the data commitment window as a param that can be changed. However, we didn't test extending/shrinking in a hardfork or via a gov proposal to see the consequences of it.

It would be good to have a test that does that and also the implementation might need to be updated not to miss any blocks when shrinking the window or to have overlapping commitments that might break some query that depend on that.

SweeXordious:
Changing the data commitment window manually in the state machine results in the validators getting:
```
11:16PM ERR Error in validation err="wrong Block.Header.AppHash.  Expected 655856A03936DF8A914425FFFE32CDB3FAC493F4BEEBAD7531986AA0491DAB09, got F9F374A6376C2AF6549D28F8381FEF316FE9833088393F02F1700BFAA1597748" module=blockchain
11:16PM ERR Stopping peer for error err="blockchainReactor validation error: wrong Block.Header.AppHash.  Expected 655856A03936DF8A914425FFFE32CDB3FAC493F4BEEBAD7531986AA0491DAB09, got F9F374A6376C2AF6549D28F8381FEF316FE9833088393F02F1700BFAA1597748" module=p2p peer={"Data":{},"Logger":{}}
```

So, if we want to change the window manually, we will either need to do a gov proposal, or we need to have an upgrade logic that specifies when to update the window.

For the former way, via a gov proposal, changing the window will have the following consequences:

- If we shrink the window, we will miss blocks during the transition
- If we expand the window, we will have blocks that are committed to in two different data commitment attestations.

Thus, we will need to update the window logic to support this.

cmwaters:
> For the former way, via a gov proposal, changing the window will have the following consequences:
>
> If we shrink the window, we will miss blocks during the transition
> If we expand the window, we will have blocks that are committed to in two different data commitment attestations.

Can you explain more why that is?

SweeXordious:
This is for the QGB missing blocks and not related to consensus. And, that is because we're creating the data commitments based on the current height.
For example, if the window is 100, and we're at block 200, the commitment will be: `[200-window, 200] = [100, 200]`.
And if we shrink the window at 300 to 50, then, at height 300, the commitment will be `[300-window, 300] = [250, 300]`.

And the same reasoning for expanding. Thus, I will update the mechanism to get the last commitment and add to it the window: `[lastCommitment.EndBlock +1, lastCommitment.EndBlock + window]`.

[...]

PR #1686, opened 2023-05-02
feat: implement the update logic for data commitment creation

SweeXordious:
## Overview

Implements the update logic for data commitment creation. Because, this is changing the state between the two versions and if a validator starts syncing from genesis (even using the new version), it will end up failing.

Thus, we need to implement the same mechanism we did for the `threshold` so that new binaries are able to sync from genesis.

More information in: https://github.com/celestiaorg/celestia-app/issues/1676#issuecomment-1532231143

Note: this will result in missing a block between the updates. But it's okey for BSR. However, we have an issue to update the data commitment creation to support changes in the window:  https://github.com/celestiaorg/celestia-app/issues/1685

## Checklist

### DIFF
--- a/x/qgb/keeper/keeper_data_commitment.go
+++ b/x/qgb/keeper/keeper_data_commitment.go
@@ -13,10 +13,16 @@ import (
 // GetCurrentDataCommitment creates the latest data commitment at current height according to
 // the data commitment window specified
 func (k Keeper) GetCurrentDataCommitment(ctx sdk.Context) (types.DataCommitment, error) {
-	beginBlock := uint64(ctx.BlockHeight()) - k.GetDataCommitmentWindowParam(ctx) + 1
-	// for a data commitment window of 400, the ranges will be: 1-400;401-800;801-1200
-	endBlock := uint64(ctx.BlockHeight())
-
+	var beginBlock, endBlock uint64
+	if ctx.BlockHeader().Version.App == 0 {
+		beginBlock = uint64(ctx.BlockHeight()) - k.GetDataCommitmentWindowParam(ctx)
+		// for a data commitment window of 400, the ranges will be: 0-399;400-799;800-1199
+		endBlock = uint64(ctx.BlockHeight()) - 1
+	} else {
+		beginBlock = uint64(ctx.BlockHeight()) - k.GetDataCommitmentWindowParam(ctx) + 1
+		// for a data commitment window of 400, the ranges will be: 1-400;401-800;801-1200
+		endBlock = uint64(ctx.BlockHeight())
+	}
 	if !k.CheckLatestAttestationNonce(ctx) {
 		return types.DataCommitment{}, types.ErrLatestAttestationNonceStillNotInitialized
 	}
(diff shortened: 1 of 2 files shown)


SweeXordious:
@evan-forbes Is this worth writing a test for? Or we can just test it manually ourselves?

evan-forbes (approved):
> is this worth testing

under normal circumstances yeah, but since this is just a quick and dirty hardfork, as long as the [existing test hardfork test](https://github.com/SweeXordious/celestia-app/blob/update_logic_for_window/x/qgb/test/versions_integration_test.go) is working (it is) we should be ok. We might want to run it locally long enough to produce a data commitment after the fork to make sure nothing weird happens. we can test locally in general as well.

great catch btw!!!

evan-forbes:
ideally we should update that failing unit test to account for the off by one before merging too :slightly_smiling_face: 

hope to have this up on the new e2e test soon, but still need to debug a few docker things

evan-forbes (approved):

SweeXordious (commented):
Tested this PR using this repo https://github.com/celestiaorg/upgrade-test to simulate an upgrade via switching binaries, while having nodes using different binaries connected. And it works fine

x/qgb/keeper/keeper_data_commitment.go
```diff
@@ -13,10 +13,16 @@ import (
...
-	// for a data commitment window of 400, the ranges will be: 1-400;401-800;801-1200
-	endBlock := uint64(ctx.BlockHeight())
-
+	var beginBlock, endBlock uint64
+	if ctx.BlockHeader().Version.App == 0 {
+		beginBlock = uint64(ctx.BlockHeight()) - k.GetDataCommitmentWindowParam(ctx)
+		// for a data commitment window of 400, the ranges will be: 0-399;400-799;800-1199
+		endBlock = uint64(ctx.BlockHeight()) - 1
+	} else {
+		beginBlock = uint64(ctx.BlockHeight()) - k.GetDataCommitmentWindowParam(ctx) + 1
```
rootulp, line 22: [not blocking] is there a risk of underflow here because this logic performs subtraction with `uint64`?

[unrelated] it would be helpful if https://github.com/SweeXordious/celestia-app/blob/aaa09a5853da9956320775404707b9115e02ee54/x/qgb/types/genesis.pb.go#L28 included a doc comment. Something like

```go
	// DataCommitmentWindow is the number of blocks...
	DataCommitmentWindow uint64 `protobuf:"varint,1,opt,name=data_commitment_window,json=dataCommitmentWindow,proto3" json:"data_commitment_window,omitempty"`
```

rootulp (approved):
[optional] perhaps a more descriptive PR title for the celestia-app auto-generated release notes:

"feat!: shift data commitment window by 1 for appVersion 1"

Note the `!` because IIUC this is a breaking change.

SweeXordious:
> feat!: shift data commitment window by 1 for appVersion 1

Shifting the data commitment window is not done in this PR. This PR  implements the upgrade mechanism so that it doesn't break the state machine

x/qgb/keeper/keeper_data_commitment.go
SweeXordious: If called correctly, there shouldn't be any underflow. But yes, I guess  at some point we should switch to using int64, and taking negative values as bad values

PR #1690, opened 2023-05-03
feat!: swap out square construction mechanism 

### LINKED ISSUE #935: Mandatory non-interactive defaults

### LINKED ISSUE #1186: Remove Estimate Square Size
### Summary

Instead of estimating the square size and then building the square in that square, you could build the square in `MaxSquareSize`and then reduce the square until it is small enough. 

One reduction of square size can happen if the square is only filled to 1/4th of the square. Then you can divide the row in half and iterate. The resulting square should still be a valid square. without changing the shares. The commitments, for example, are square size-independent, so they will not change. The index of the blobs, for example, will stay the same as well.

Here is an example of how a square of 16x16 might reduce to 4x4. The middle step can be skipped computationally and is only here for visualisation purposes. 

image

### Problem Definition

- Code Reduction 
- The Validators do not need to know which square to build as they will deterministically pick the smallest one given an indexed list of txns 

### For Admin Use

- [ ] Not duplicate issue

### LINKED ISSUE #1200: Reduce estimated padding

### LINKED ISSUE #1214: Adopt a deterministic square construction mechanism

### LINKED ISSUE #1216: maxWrappedTxOverhead overestimation
### Summary

The Function `maxWrappedTxOverhead` overestimates, creating a `nonreserveStart` which is far into the block. 

https://github.com/celestiaorg/celestia-app/blob/5f8b39b4cb7ab0a9083b97f2a20cae7a2cc1a1c5/app/estimate_square_size.go#L76-L90

A better estimation will be more complicated. This could also have an effect on https://github.com/celestiaorg/celestia-app/issues/1209

### Problem Definition

While doing napkin math in https://github.com/celestiaorg/celestia-app/issues/1186#issuecomment-1377438879 I looked at `maxWrappedTxOverhead`. Putting in 128 as square size, the function gives us 15 bytes of overhead per transaction. 

This means that using the napkin math above, where I only did 1 byte for the overhead and got 16 shares, we would get 16*15 = 240 shares that we postpone `nonreserveStart` if we have 128 * 64 transactions ( half the block). 

This could overestimate the `nonreserveStart` as that many transactions with `maxWrappedTxOverhead` are clearly not possible. 

On another note, as I was reading the comment, the function might take so long because of the huge slice creation. Maybe a quick table could be enough if it creates performance issues. 

### For Admin Use

- [ ] Not duplicate issue

### LINKED ISSUE #1608: Shared indexes are not checked during ProcessProposal
https://github.com/celestiaorg/celestia-app/blob/ab24a7b4d5190a53cf4bb0a9e989f4c824759f8e/pkg/shares/share_splitting.go#L41-L42

since we're sorting the indexes during the share splitting process, we're not actually checking that each share index is correctly assigned to each PFB. Meaning a validator could swap two share indexes and it would not be caught during process proposal.

unless we add a consensus time check, after blob inclusion fraud proofs are implemented any single validator could halt the chain. 

Note that this might be fixed in #1214

### LINKED ISSUE #1673: Blobs should be ordered by namespace
https://github.com/celestiaorg/celestia-app/blob/318d7853a3bf5a07bb9530a27bb2e5e314bbde2a/app/write_square.go#L52

This should account for ordering due to the entire namespace (namespace version + namespace ID)

### LINKED ISSUE #1711: Unable to get blob share range when there are multiple blobs in a PFB
### Summary

We have hardcoded to only get the range for the first blob listed in the PFB

https://github.com/celestiaorg/celestia-app/blob/c08d1c690945b27c8341d91708419c872f3b69cd/pkg/proof/proof.go#L100-L102

### Problem Definition

Users can't prove blob inclusion when they include multiple blobs in their PFB

### Proposal

Introduce a second argument `blobIndex` that also specifies the index of the blobs in the PFB to get the share range for

### For Admin Use

- [ ] Not duplicate issue

cmwaters:
This PR swaps out the existing square construction mechanism in Prepare and Process proposal for the new one as described in ADR020. This change also fixes up several minor existing bugs

fixes #1711, fixes #1673, fixes #1608, fixes #1214, fixes #935, fixes #1186, fixes #1200, and fixes #1216


rootulp (commented):
Overall LGTM

app/parse_txs.go
```diff
@@ -86,3 +107,15 @@ func checkTxValidity(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, hand
...
 	return handler(ctx, sdkTx, false)
 }
+
+func encodeBlobTxs(blobTxs []tmproto.BlobTx) [][]byte {
+	txs := make([][]byte, len(blobTxs))
+	var err error
+	for i, tx := range blobTxs {
+		txs[i], err = coretypes.MarshalBlobTx(tx.Tx, tx.Blobs...)
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
cmwaters: Just looked through this function and because it's caused by prepare proposal only we could just drop the transaction upon error.

go.mod
```diff
@@ -202,5 +202,5 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.12.0-sdk-v0.46.11
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.18.0-tm-v0.34.27
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.19.0-tm-v0.34.27
```
cmwaters, line 205: note that this bumps celestia-consensus to the version that no longer includes `Blob`s as a separate field in the `Data` struct. This needs to be done in the same PR if we want to avoid breaking everything

pkg/square/square_test.go
```diff
@@ -236,6 +236,77 @@ func TestSquareBlobPostions(t *testing.T) {
...
+	txTwo := types.Tx(bytes.Repeat([]byte{2}, 600))
+	txThree := types.Tx(bytes.Repeat([]byte{3}, 1000))
+
+	testCases := []test{
+		{
+			name:      "txOne occupies shares 0 to 0",
+			txs:       [][]byte{txOne},
+			index:     0,
+			wantStart: 0,
+			wantEnd:   1,
```
SweeXordious, line 259: if the transaction occupies only the first share, then the `wantEnd` should be equal to 0, since the range is inclusive, no?

pkg/square/square.go
```diff
@@ -42,45 +38,62 @@ func Construct(txs [][]byte, maxSquareSize int) (Square, [][]byte, error) {
...
 
+// TxShareRange returns the range of share indexes that the tx, specified by txIndex, occupies.
+// Both ends of the range are inclusive.
+func TxShareRange(txs [][]byte, txIndex int) (shares.ShareRange, error) {
+	builder, err := NewBuilder(appconsts.DefaultMaxSquareSize, txs...)
+	if err != nil {
+		return shares.ShareRange{}, err
+	}
+
+	return builder.FindTxShareRange(txIndex)
```
SweeXordious, line 63: I guess this function `FindTxShareRange` needs to be tested somewhere

pkg/proof/proof_test.go
```diff
@@ -1,66 +1,65 @@
...
-	"github.com/tendermint/tendermint/types"
 )
 
 func TestNewTxInclusionProof(t *testing.T) {
-	blockData := types.Data{
-		Txs:        testfactory.GenerateRandomTxs(50, 500),
-		Blobs:      []types.Blob{},
-		SquareSize: appconsts.DefaultMaxSquareSize,
-	}
+	blockTxs := testfactory.GenerateRandomTxs(50, 500).ToSliceOfBytes()
```
SweeXordious, line 24: Will these contain transactions that have multiple blobs? If not, I guess we should test that also

SweeXordious (commented):
With these changes, I guess we should see if https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/client/verify.go is still working and we will need to update its implementation to support multiple blobs

pkg/proof/proof.go
```diff
@@ -22,52 +24,52 @@ import (
...
-		return 0, 0, err
-	}
-	shareRange := shareRanges[data.Txs[txIndex].Key()]
-
-	return uint64(shareRange.Start), uint64(shareRange.End), nil
-}
-
 // BlobShareRange returns the start and end positions for the shares
 // where a given blob, referenced by its wrapped PFB transaction, was published at.
 // Note: only supports transactions containing a single blob.
```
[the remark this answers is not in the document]
cmwaters: Sorry I realised this part wasn't finished so I've put this back into a draft while I address this

app/process_proposal.go
```diff
@@ -125,76 +55,78 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
+	dataSquare, err := square.Construct(req.BlockData.Txs, appconsts.DefaultMaxSquareSize)
+	if err != nil {
+		logInvalidPropBlockError(app.Logger(), req.Header, "failure to compute data square from transactions:", err)
+		return reject()
+	}
+
+	// Assert that the square size stated by the proposer is correct
+	if dataSquare.Size() != req.BlockData.SquareSize {
+		logInvalidPropBlock(app.Logger(), req.Header, "proposed square size differs from calculated square size")
+		return reject()
```
evan-forbes, line 111: [question]
I think you mentioned it already but I forgot, do we still need the squaresize in the block data?

pkg/appconsts/appconsts.go
```diff
@@ -87,13 +87,6 @@ const (
 	// included in a PayForBlobs txn
 	DefaultGasPerBlobByte = 8
 
-	// TransactionsPerBlockLimit is the maximum number of transactions a block
-	// producer will include in a block.
-	//
-	// NOTE: Currently this value is set at roughly the number of PFBs that
-	// would fill one quarter of the max square size.
-	TransactionsPerBlockLimit = 5090
```
evan-forbes, line 95: soo nice

pkg/shares/merge_test.go
```diff
@@ -1,116 +0,0 @@
...
-			shares, err := Split(data, false)
-			require.NoError(t, err)
-
-			eds, err := rsmt2d.ComputeExtendedDataSquare(ToBytes(shares), appconsts.DefaultCodec(), rsmt2d.NewDefaultTree)
-			assert.NoError(t, err)
-
-			got, err := merge(eds)
-			assert.NoError(t, err)
-			assert.Equal(t, data, got)
-		})
```
evan-forbes, line 69: [blocking]
iirc, I think we are keeping this purely because this is a good way to round trip test share splitting. (correct @rootulp?) perhaps we should add a comment to not delete this test (or merge)

if we decide to delete, then I think we should do so outside the scope of this PR. Does anyone have links to the old discussions deciding this?

evan-forbes (commented):
1) **wow** really nice refactor, and thanks for pushing for this change!!
2) good lesson in that we could have saved ourselves quite a bit of trouble if we would would have just implemented the simpler solution (forced non-interactive defaults) first.

can we delete?
https://github.com/celestiaorg/celestia-app/blob/a4af1000f30c0f498d79cc57a5ed83e248d0d890/app/estimate_square_size.go#L14-L47

I still need to take a second look at process proposal to make sure we're not missing any check, but I'm almost certain all of them are there atm.

otherwise I only had a few minor comments!

nashqueue:
I think this also fixes #1186 , #1200 and #1216

app/parse_txs.go
```diff
@@ -24,6 +25,26 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []core.BlobTx) {
...
+
+	normalTxs, ctx = filterStdTxs(ctx.Logger(), txConfig.TxDecoder(), ctx, seqHandler, normalTxs)
+
+	// check the signatures and increment the sequences of the blob txs,
+	// and filter out any that fail. Panics from the anteHandler are caught and
+	// logged.
+	svHandler := sigVerifyAnteHandler(accountKeeper, txConfig)
+	blobTxs, _ = filterBlobTxs(ctx.Logger(), txConfig.TxDecoder(), ctx, svHandler, blobTxs)
+
+	return append(normalTxs, encodeBlobTxs(blobTxs)...)
```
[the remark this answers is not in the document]
cmwaters: Yeah this is true and I was aware of the redundant encoding/decoding that is being done, however, as you said, I preferred readability over the extra computation cycles this may cause (which I think is very minimal).

I think we could look to change it in a future PR once we have a good gauge on where our bottlenecks are

app/process_proposal.go
```diff
@@ -125,76 +55,78 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
-				return abci.ResponseProcessProposal{
-					Result: abci.ResponseProcessProposal_REJECT,
-				}
-			}
-			if !bytes.Equal(pfb.ShareCommitments[i], commitment) {
-				logInvalidPropBlock(app.Logger(), req.Header, "found commitment does not match user's")
-				return abci.ResponseProcessProposal{
-					Result: abci.ResponseProcessProposal_REJECT,
-				}
-			}
```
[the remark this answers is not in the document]
cmwaters: Yup that makes sense. Let me create an issue to track it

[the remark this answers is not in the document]
cmwaters: https://github.com/celestiaorg/celestia-app/issues/1717

app/process_proposal.go
cmwaters: It seems that QGB requires the square size and since we don't persist it anywhere in `celestia-core`, I've left it in the block data itself. In the future when we use a similar data storage scheme as `node`, I think it will become redundant and we can remove it from the block itself. Alternatively, I'm not sure whether the QGB needs the square size (i.e. why is the data root and height not sufficient)?

cmwaters:
> With these changes, I guess we should see if https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/client/verify.go is still working and we will need to update its implementation to support multiple blobs

I've added a blobIndex argument to the command line

cmwaters:
> I still need to take a second look at process proposal to make sure we're not missing any check, but I'm almost certain all of them are there atm.

Please do! Really don't want to miss anything here

evan-forbes:
also fixes https://github.com/celestiaorg/celestia-app/issues/1110

pkg/square/square.go
```diff
@@ -42,45 +38,62 @@ func Construct(txs [][]byte, maxSquareSize int) (Square, [][]byte, error) {
...
 
 // Size returns the size of the sides of a square
 func (s Square) Size() uint64 {
-	return uint64(math.Sqrt(float64(len(s))))
+	return Size(len(s))
+}
+
+func Size(len int) uint64 {
+	return uint64(math.Sqrt(float64(len)))
 }
```
rootulp, line 97: [not blocking] this implementation hasn't changed but I'm struggling to see why it was like this originally given https://go.dev/play/p/KQDnmVJHpni

I feel like it needs to round up

rootulp (approved):
No blocking feedback. I'm pretty sure all my feedback can be addressed in FLUP issues. Excited for this to merge!

pkg/shares/merge_test.go
```diff
@@ -1,116 +0,0 @@
...
-	}
-}
```
evan-forbes, line 71: we no longer need splitting, and therefore this test doesn't make sense, but a similar test would definitely make sense imo, where we always want to ensure that our parsing code is fully functional. #1729 

therefore, we might want to not delete the merge file, since we're likely going to add it back in #1729, but can delete the splitting function and this test

evan-forbes (approved):
created an issue for the merge test we discussed synchronously

other than that, I triple checked process proposal, and this is ready to :shipit: imo

app/process_proposal.go
evan-forbes: ahh that's right. I believe the square size is added for the qgb's blob inclusion fraud proof mechanism, correct @SweeXordious ?

app/process_proposal.go
```diff
@@ -125,76 +61,78 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) abci.ResponsePr
...
+
+	cacher := inclusion.NewSubtreeCacher(dataSquare.Size())
+	eds, err := rsmt2d.ComputeExtendedDataSquare(shares.ToBytes(dataSquare), appconsts.DefaultCodec(), cacher.Constructor)
+	if err != nil {
+		logInvalidPropBlockError(app.Logger(), req.Header, "failure to erasure the data square", err)
+		return reject()
+	}
+
+	dah := da.NewDataAvailabilityHeader(eds)
+	// by comparing the hashes we know the computed WrappedIndex is correct and that square construction is
```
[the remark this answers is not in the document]
cmwaters: I see that I got WrappedIndex confused with IndexWrapper. Will update

pkg/square/builder.go
```diff
@@ -159,32 +181,148 @@ func (c *Builder) Export() (Square, error) {
...
+		}
+	}
+	index := pfbIndex - len(b.txs)
+	if index < 0 {
+		return 0, fmt.Errorf("pfbIndex %d does not match a pfb", pfbIndex)
+	}
+
+	if index >= len(b.pfbs) {
+		return 0, fmt.Errorf("pfbIndex %d out of range", pfbIndex)
+	}
```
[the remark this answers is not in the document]
cmwaters: Ok tried to streamline them a bit

pkg/square/square.go
cmwaters: I was assuming that the length was always a power of two but that might not be the case. I think you're right that it's safer if we round up

Issue #1717, opened 2023-05-09
Test share commitments match the commitments for the share ranges in the square.


cmwaters:
### Summary

Following https://github.com/celestiaorg/celestia-app/pull/1690 and ADR13, we should add tests to ensure that the pfb share commitments are correct i.e.

```go
for _, pfb := range wrappedPFBs 
    for idx, shareIndex := range pfb.ShareIndexes {
        commitment := inclusion.GetCommitment(cacher, dah, shareIndex, SparseSharesNeeded(pfb.BlobSizes[i]))
        require.Equal(t, pfb.ShareCommitment[i], commitment)
    }
}
```

PR #1726, opened 2023-05-09
test: add benchmarks and share commitment test

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1717


rootulp (dismissed):

pkg/square/square_test.go
```diff
@@ -254,3 +328,71 @@ func generateBlobTxsWithNamespaces(t *testing.T, namespaces []ns.Namespace, blob
...
+
+	_, err = square.BlobShareRange(txs, 10, 0)
+	require.Error(t, err)
+
+	_, err = square.BlobShareRange(txs, 0, 10)
+	require.Error(t, err)
+}
+
+func TestSquareShareCommitments(t *testing.T) {
+	txs := generateOrderedTxs(10, 10, 5)
```
evan-forbes, line 370: in the past, we've found edge case bugs in the share splitting process by running this same test over every square, effectively fuzzing it, so I think it would be valuable to expand on the inputs to this test either here or in the future. I'm also happy to take this on, as not all follow ups to #1690 have to be handled immediatly or in this PR

evan-forbes (commented):
weird that github is showing the total diff before #1690

beyond the typical confict stuff I think this LGTM! its difficult to be certain because of the whole diff issues, but only `TestSquareShareCommitments` and the new benchmark are added here, correct?

cmwaters:
Ok I think I've fixed the merge conflicts here. It is annoying that it can't resolve the diff properly. I think that might be because I made changes to the dependent branch before merging it to main so it doesn't see the common commit

rootulp (approved):

evan-forbes (approved):
thanks for following up on this!

PR #1735, opened 2023-05-10
ci: run `test` workflow even if `*.go` file isn't modified

rootulp:
## Motivation

`test` is now a mandatory CI check for PRs to merge. Since `test` only runs if a `.go` file was modified, PRs like https://github.com/celestiaorg/celestia-app/pull/1712 won't pass all mandatory CI checks. This change always runs the `test` workflow

## Testing

https://github.com/rootulp/celestia-app/pull/7


rootulp:
This may not be necessary

SweeXordious:
I think in general, it makes sense to have this. But, also should be for the docker build I guess

PR #1739, opened 2023-05-11
docs: add section on SHA256 performance to namespace size ADR

rootulp:
## Context

Based on a Slack thread with @preston-evans98

## Remaining tasks
- [x] instrument the Go NMT `HashNode` implementation to see how many times `Sha2::compress` is invoked for different sizes of data
- [x] clarify namespace ID vs namespace

## FLUPs

If this PR is approved, implementation will be tracked in https://github.com/celestiaorg/celestia-app/issues/1753


docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -145,6 +145,31 @@ Another tradeoff to consider is the size of the namespace ID in the share. Since
...
+16                     | 129
+32                     | 193
+33                     | 197
+
+The data size is calculated as follows:
+`SHA256(domain_sep || left_min || left_max || left_data_hash || right_min || right_max || right_data_hash)` = 1 + namespaceSize + namespaceSize + 32 + namespaceSize + namespaceSize + 32 which simplifies to (4 \* namespaceSIze) + (2 \* 32) + 1.
+
+Hashing has a high performance impact in ZK contexts. Since a larger namespace size results in a larger data size for the SHA256 operation, increasing the namespace size has a negative impact impact on the cost to perform the SHA256 operation in a ZK context.
+
+Based on this StackOverflow [post](https://crypto.stackexchange.com/questions/54852/what-happens-if-a-sha-256-input-is-too-long-longer-than-512-bits), when the data provided to SHA256 exceeds 56 bytes, the data must be chunked into 64 byte blocks with a trailing 56 byte block + 8 bytes of metadata. The following table shows the range of namespace sizes acceptable for a single SHA256 compression invocation:
```
rootulp, line 164: @preston-evans98 I'm not totally clear on the relationship between data size and compression invocations. I see that the data needs to be chunked into 64 byte blocks with a trailing 56 byte block + 8 bytes of metadata.

However, I see that the compression invocation takes as input the previous aggregate result + the next chunk. Based on that observation I think it's fair to say that the # of compression invocations == the number of chunks but was hoping you could verify that claim.

docs/architecture/adr-015-namespace-id-size.md
```diff
@@ -145,6 +145,31 @@ Another tradeoff to consider is the size of the namespace ID in the share. Since
 
 If the namespace ID size is increased, the maximum possible blob will decrease. Given the maximum possible blob is bounded by the number of bytes available for blob space in a data square, if a 32 byte namespace ID size is adopted, the maxmimum blob size will decrease by an upper bound of `appconsts.DefaultMaxSquareSize * appconsts.DefaultMaxSquareSize * (32-8)`. Note this is an upper bound because not all shares in the data square can be used for blob data (i.e. at least one share must contain the associated PayForBlob transaction).
 
+### SHA256 Performance
+
+If the namespace size is increased, whenever a SHA256 invocation takes place (e.g. NMT's [HashNode](https://github.com/celestiaorg/nmt/blob/fd00c52175c48bad64d03444689162fb9c6bee41/hasher.go#L265)), more data needs to be SHA256'ed. For example:
+
+Namespace size (bytes) |[HashNode](https://github.com/celestiaorg/nmt/blob/fd00c52175c48bad64d03444689162fb9c6bee41/hasher.go#L302)'s SHA256 max data (bytes)
```
musalbas, line 152: `Namespace size (bytes)` here does not seem to mean the same thing as `Namespace size (bytes)` in the earlier tables right? In this table it seems to include the version byte, whereas in the earlier tables they don't.

docs/architecture/adr-015-namespace-id-size.md
preston-evans98: Yes, that's my understanding as well

docs/architecture/adr-015-namespace-id-size.md
rootulp: Good point. This ADR was written before the namespace version was introduced. I'll clarify in all the tables

musalbas:
Assuming that the analysis of SHA256 compression invocations is correct, I am in favor of decreasing the namespace size to 29 bytes:

* 1 byte for namespace version
* 18 reserved bytes that much be zero
* 10 useful non-reserved bytes that can be non-zero
= 29 bytes

preston-evans98:
> Assuming that the analysis of SHA256 compression invocations is correct, I am in favor of decreasing the namespace size to 29 bytes in order to reduce the number of SHA256 compression invocations from 4 to 3:
> 
> * 1 byte for namespace version
> * 18 reserved bytes that much be zero
> * 10 useful non-reserved bytes that can be non-zero
>   = 29 bytes

This would be fantastic for us. That should translate into roughly a 30% reduction in proof cost for our system. This is probably also a big win for elliptic curve snarks. I think it's pretty common to use a field size of around 255 bits - so a 33 byte namespace would be too big to represent in a single field element. Which means that even just checking namespace equality or ordering would be quite expensive, requiring decomposition into limbs etc.

rootulp:
Thanks to @preston-evans98 for helping me analyze the SHA256 loop invocations!

evan-forbes (approved):
really great additions and thanks for confirming @rootulp!!

liamsi (approved):
👍

Issue #1716, opened 2023-05-09
Consistent usage of ranges for inclusion verification


cmwaters:
### Summary

When verifying the inclusion of blobs or transactions we first find the shares that these data types range over and then verify that the range of shares were included. The code currently has different interpretations of a range - is it end inclusive or end exclusive. We need to agree on a consistent interpretation so that all shares in the range are verified.

### Problem Definition

We currently denote both the start and end of a ShareRange as being inclusive (see https://github.com/celestiaorg/celestia-app/blob/e223493e3bb53a969b18297150a6de20202feaec/pkg/shares/split_compact_shares.go#L66) i.e. a share range of [3, 5] means the tx takes up shares 3, 4, and 5. But the VerifyProof function here https://github.com/celestiaorg/celestia-core/blob/6e590724cde96109ab4732e0a72a92bb5c2dc912/types/share_proof.go#L114 treats end as being exclusive. In the above example it calculates the sharesUsed as 5-3=2, which is incorrect. 3 shares are actually used. Further down, we use that variable to calculate the shares we want to verify as sp.Data[cursor:sharesUsed+cursor] . go slices are end exclusive so this would only check inclusion of the 3rd and 4th and not the 5th.

### Proposal

As mentioned golang is end exclusive. NMT is also end exclusive. End exclusive is less error prone for length calculations as it's just end - start (whereas inclusive you need +1). However, I feel that people naturally think a range is inclusive.

I would opt for having the range exclusive but don't mind either way so long as it's properly documented.

rootulp:
> I would opt for having the range exclusive

+1. I think we should also consider this for QGB data commitment ranges @SweeXordious

SweeXordious:
Sure, sounds good, I will create an issue to start updating the QGB implementation

PR #1742, opened 2023-05-11
fix: make share ranges end exclusive

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1716


pkg/square/builder.go
```diff
@@ -261,22 +261,22 @@ func (b *Builder) BlobShareLength(pfbIndex, blobIndex int) (int, error) {
 
 // FindTxStartingIndex returns the first and last share index that the transaction
 // occupies within the square. The indexes are both inclusive.
```
rootulp, line 263:
```suggestion
// FindTxShareRange returns the range of shares occupied by the tx at txIndex.
```

rootulp (dismissed):

rootulp (approved):

evan-forbes (approved):
LGTM!

this mostly affects the QGB, so would defer to @SweeXordious

pkg/proof/proof.go
```diff
@@ -51,22 +51,21 @@ func getTxNamespace(tx []byte) (ns appns.Namespace) {
...
-	endShare uint64,
+	shareRange shares.Range,
 ) (types.ShareProof, error) {
-	startRow := startShare / squareSize
-	endRow := endShare / squareSize
-	startLeaf := startShare % squareSize
-	endLeaf := endShare % squareSize
+	squareSize := dataSquare.Size()
+	startRow := shareRange.Start / squareSize
+	endRow := (shareRange.End - 1) / squareSize
```
SweeXordious, line 64: why `-1` ? the NMT takes end exclusive ranges too, shouldn't this be `shareRange.End`?

SweeXordious (commented):
LGTM, but shouldn't this change also be reflected in here:

https://github.com/celestiaorg/celestia-app/blob/1b5263ac593c05ee9673bc03c116c1bba94acf13/x/qgb/client/verify.go#L166

Also, this change requires few changes on core for the QGB

pkg/proof/proof.go
cmwaters: Because the endRow might be incorrect if we are using the index of the share that actually comes after the last share

SweeXordious: So we expect Core to send the ranges already end exclusive?

cmwaters: Are you referring to the `ProveShares` rpc endpoint? Core just forwards whatever the users request is, so it will need to be the user that updates and specifies an end exclusive share range.

We should update documentation and comments if we can though

SweeXordious (approved):

rootulp (approved):

evan-forbes (approved):

Issue #1592, opened 2023-04-02
EPIC: Hard capped blocksize


evan-forbes:
We want the ability to configure the square size via token voting, but with hard cap that requires a hardfork to change. In previous discussions, we've decided to simply use the consensus param of max block size to limit the max block size. However, the hardcoded limit is still the max square size (currently 128 as of writing), which gets hit at ~6-7MB.

We should still make the block size limited by the consensus param, and we don't want to go above  a 128x128 square size (ods). This will likely involve setting the default block size consensus param to the maximum value that can fit in a 128x128 square.

related to https://github.com/celestiaorg/celestia-app/issues/1569 in that we are wanting to hardcode a max parameter.

blocked by #1214 because that removes the estimation logic, which would probably be used to limit the square size

PR #1743, opened 2023-05-12
feat!: restrict `MaxBytes` to values at or below that of `MaxSquareSize`

evan-forbes:
## Overview

Leaving as a draft for now until we merge https://github.com/celestiaorg/cosmos-sdk/pull/317 and decide on if this is actually the approach that we want to use to try to limit the square size.

closes https://github.com/celestiaorg/celestia-app/issues/1592 closes https://github.com/celestiaorg/celestia-app/issues/1737

## Checklist

### DIFF
--- a/app/test/limited_block_size_test.go
+++ b/app/test/limited_block_size_test.go
@@ -1,0 +1,59 @@
+package app_test
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/celestiaorg/celestia-app/pkg/square"
+	"github.com/celestiaorg/celestia-app/test/txsim"
+	"github.com/celestiaorg/celestia-app/test/util/testnode"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLimitedBlockSize(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping Limited Block Size integration test in short mode.")
+	}
+
+	desiredSquareSize := uint64(64)
+
+	cparams := testnode.DefaultParams()
+	// limit the max block size to 64 x 64 via the consensus parameter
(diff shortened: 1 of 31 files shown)


x/paramfilter/block_params.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		if v.MaxBytes <= 0 {
+			return fmt.Errorf("block maximum bytes must be positive: %d", v.MaxBytes)
+		}
+
+		maxBlockBytes := square.EstimateMaxBlockBytes(squareSize)
+		if v.MaxBytes > maxBlockBytes {
+			return fmt.Errorf("block maximum bytes must be less than or equal to %d: %d", maxBlockBytes, v.MaxBytes)
+		}
```
evan-forbes, line 74: this is consensus breaking, and tbh I'm not really even sure if we should include it since the only real value it provides is throwing an error if people submit a param proposal to change the block size to something that could get above 128 x 128. By doing so, until https://github.com/celestiaorg/celestia-core/issues/1007, we could easily end up producing less efficient squares in some scenarios.

evan-forbes:
while some of the changes here are ok, this branch mainly introduces a lot of code to simply stop proposals that exceed the max square value. Perhaps its nice to throw an error in that scenario, but its far from necessary and add complexity, so I'm closing this for now. 

if that's something that we decide we want, then we can reopen

app/test/limited_block_size_test.go
```diff
@@ -0,0 +1,59 @@
...
+		[]string{grpcAddr},
+		kr,
+		rand.Int63(),
+		time.Second,
+		seqs...,
+	)
+
+	// check the block sizes
+	blocks, err := testnode.ReadBlockchain(context.Background(), rpcAddr)
+	require.NoError(t, err)
```
evan-forbes, line 52: txsim is cool tho

Issue #1747, opened 2023-05-12
testing should support cases of multiple blobs in a blobTx


cmwaters:
### Summary

Test helper functions such as https://github.com/celestiaorg/celestia-app/blob/49213ba11c7ffbcecfcc3e4d40dc8de9857d4967/test/util/blobfactory/payforblob_factory.go#L217 allow for multiple blobs per Tx. This should then be added to a lot of the integration/fuzz tests

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #1744, opened 2023-05-12
Fuzz the share commitment correctness


evan-forbes:
Previous to #1690, we were recalculating the share commitment in each PFB using only subtree roots of the data square

https://github.com/celestiaorg/celestia-app/blob/54b65f9c0179359f7af7cdf6b00f8e7a12275783/app/process_proposal.go#L170-L184

given the importance of this property and the complexity of the block construction/share encoding, we should replicate this check using some form of fuzzer.

evan-forbes:
also see #1726 which adds a test for the share commitment

cmwaters:
I can pick this up. I had thought about appending it to the existing fuzzy test I had written

evan-forbes:
> I had thought about appending it to the existing fuzzy test I had written

that would be perfect!

PR #1749, opened 2023-05-12
test: fuzz share commitments

cmwaters:
Closes: #1744, closes: #1747


pkg/square/builder.go
```diff
@@ -339,6 +339,14 @@ func (b *Builder) GetWrappedPFB(txIndex int) (*coretypes.IndexWrapper, error) {
 	return b.pfbs[txIndex-len(b.txs)], nil
 }
 
+func (b *Builder) NumPFBs() int {
+	return len(b.pfbs)
+}
+
+func (b *Builder) NumTxs() int {
+	return len(b.txs)
+}
```
rootulp, line 348: [optional] I think it may help if we consistently referred to the non-blob Txs as something like `normalTxs` rather than just `txs`. This proposal is mostly b/c I think of it like: pfbTxs + normalTxs = txs and the way this is written `txs` is ambiguous if it refers to normalTxs or the set of all Txs

pkg/square/testdata/fuzz/FuzzSquare/dd4df2aa7925a6a2
```diff
@@ -0,0 +1,5 @@
+go test fuzz v1
```
rootulp, line 1: [question] should this file be checked into version control?

pkg/square/testdata/fuzz/FuzzSquareShareCommitments/dabfc57957837b0a
```diff
@@ -0,0 +1,5 @@
+go test fuzz v1
```
rootulp, line 1: [question] should this file be checked into version control?

rootulp (dismissed):

pkg/square/testdata/fuzz/FuzzSquare/dd4df2aa7925a6a2
cmwaters: No it must have slipped in. I'll remove it

pkg/square/builder.go
cmwaters: Yeah I can adopt that nomenclature

evan-forbes (approved):
dope!

pkg/square/builder.go
```diff
@@ -344,7 +344,7 @@ func (b *Builder) NumPFBs() int {
 }
 
 func (b *Builder) NumTxs() int {
-	return len(b.txs)
+	return len(b.txs) + len(b.pfbs)
 }
```
rootulp: [optional] this function may warrant a clarifying godoc

```go

// NumTxs returns the total number of txs (PFBs + normalTxs)
func (b *Builder) NumTxs() int {
	return len(b.txs) + len(b.pfbs)
}
```

we may also consider renaming the field `b.txs` => `b.normalTxs`

pkg/square/square_fuzz_test.go
```diff
@@ -52,7 +52,7 @@ func FuzzSquare(f *testing.F) {
 		builder, err := square.NewBuilder(appconsts.DefaultMaxSquareSize, orderedTxs...)
 		require.NoError(t, err)
 		totalPfbs := builder.NumPFBs()
-		totalNormalTxs := builder.NumTxs()
+		totalNormalTxs := builder.NumTxs() - totalPfbs
```
rootulp, line 55: [optional] an exported function on the builder for normal txs: `builder.NumNormalTxs()`

rootulp (approved):

pkg/square/square_fuzz_test.go
cmwaters: I think I'll wait for the need to arise first

pkg/square/builder.go
cmwaters: I'm not 100% convinced about normal txs. I don't like that normality is defined on the basis of whether there is a PFB or not

rootulp: I'm open to other names to help disambiguate. Candidates: `nonPFBTxs`

Issue #1669, opened 2023-04-27
Use unique error codes for `x/qgb` module


evan-forbes:
Currently, we're using error codes that overlap with that of the standard cosmos-sdk in the qgb module. We should do something similar to that of the blob module, where we use unique error codes.

Chirag018:
Hey, @evan-forbes would like to work on this

PR #1752, opened 2023-05-12
feat(qgb module): unique error codes

Chirag018:
Closes: #1669


x/qgb/types/errors.go
```diff
@@ -5,20 +5,20 @@ import (
...
-	ErrNoValsetBeforeNonceOne                    = errors.Register(ModuleName, 28, "there is no valset before attestation nonce 1")
-	ErrDataCommitmentNotGenerated                = errors.Register(ModuleName, 29, "no data commitment has been generated for the provided height")
-	ErrDataCommitmentNotFound                    = errors.Register(ModuleName, 30, "data commitment not found")
-	ErrLatestAttestationNonceStillNotInitialized = errors.Register(ModuleName, 31, "the latest attestation nonce has still not been defined in store")
-	ErrInvalidDataCommitmentWindow               = errors.Register(ModuleName, 32, "invalid data commitment window")
+	ErrDuplicate                                 = errors.Register(ModuleName, 1002, "duplicate")
+	ErrEmpty                                     = errors.Register(ModuleName, 1006, "empty")
+	ErrNoValidators                              = errors.Register(ModuleName, 1012, "no bonded validators in active set")
+	ErrInvalidValAddress                         = errors.Register(ModuleName, 1013, "invalid validator address in current valset %v")
```
SweeXordious, line 11: I think that we can go for sequential error codes because the reason we're jumping from 1002 to 1006 for example, is because we deleted the corresponding error...

Also, @evan-forbes I thought that the error codes are defined per module. I don't see how these overlap with other errors. Or I am missing something

Chirag018: > I think that we can go for sequential error codes because the reason we're jumping from 1002 to 1006 for example, is because we deleted the corresponding error...

Got it, as I never thought of it from this angle, ie, there can be jumping between errors(1002 to 1006),
I thought of keeping it simple so that there can be a flow between the previous and updated one.

SweeXordious: I think that it's okey to do it now since we don't need to worry about backward compatibility as we still didn't launch mainnet. But after mainnet, definitly, the way you did it is the right way.
@evan-forbes What do you think? should we keep the jumps?

evan-forbes: agree, we should probably go in order here.

evan-forbes:
apologies @Chirag018 and @SweeXordious, we can just rely on the codespace returned in the txresult as indicated by @cmwaters and think we should close this as unplanned

PR #1759, opened 2023-05-12
docs: ADR021 Restricted Block Size

evan-forbes:
## Overview

[Rendered](https://github.com/celestiaorg/celestia-app/blob/baef8b8b806d8643e4b4eefe0a8df7eb7e73b193/docs/architecture/adr-021-restricted-block-size.md)

Decided to document a synchronous discussion, and open (a hopefully quick) discussion while I implement two options to choose from (option 2 and 3)

## Checklist


rootulp (commented):
Thanks a ton for documenting the context around this decision! Do you think an additional section for "Goals" or "Desirable criteria" would be helpful?


My opinion is based on the desirable criteria: 

Do we want the square size to be governance modifiable?

- if **yes** we go with option 2 (i.e. `GovMaxSquareSize`)
- if **no** we set `MaxSquareSize=64` and at a later date (i.e. when pruning is implemented) we hard-fork to increase `MaxSquareSize`

evan-forbes:
to quickly update this from the sync, we are moving forward with option 2.

option 3 is implemented and would work, however, it introduces more implicit complexity. Option 2 is more explicit and maximally flexible (we can have large squares and small blocks).

docs/architecture/adr-021-restricted-block-size.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+Currently, block sizes are controlled by three values, the smallest of which will determine what the protocol considers to be a valid block size.
+
+- `MaxBlockBytes` is a hard coded constant to the maximum total size of the protobuf encoded block, and therefore acts as a cap for `MaxBytes`. Currently set to 100 MB.
```
rootulp, line 15: [question] where is `MaxBlockBytes` defined? I can't find it in
- https://github.com/search?q=repo%3Acelestiaorg%2Fcelestia-app+MaxBlockBytes&type=code
- https://github.com/search?q=repo%3Acelestiaorg%2Fcelestia-core+MaxBlockBytes&type=code
- https://github.com/search?q=repo%3Acometbft%2Fcometbft+MaxBlockBytes&type=code

It's not obvious that `MaxBytes` and `MaxBlockBytes` are related and one is governance modifiable. Maybe we consider adopting a similar naming strategy as `MaxSquareSize` and `GovMaxSquareSize` because those are clearly related. In other words should we consider preserving `MaxBlockBytes` and renaming `MaxBytes` => `GovMaxBlockBytes`? My hunch is that the diff it will introduce with cometbft is not worth the extra clarity in naming.

evan-forbes: yeah, the only reason that I didn't change them is because `MaxBytes` is a consensus param, which changing the name might break a few things in the sdk or explorers. We could rename the constant in core

and ahh whoops, the constant is [`MaxBlockSizeBytes`](https://github.com/celestiaorg/celestia-core/blob/1b24b4894042234e7daf68c7924e4f3647b46e30/types/params.go#L15), will update

MSevey:
i wasn't very involved with these discussions so will defer to others.

docs/architecture/adr-021-restricted-block-size.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Context
+
+Currently, block sizes are controlled by three values, the smallest of which
+will determine what the protocol considers to be a valid block size.
+
+| Parameter        | Size   | Unit   | Description  | Value Control |
+|------------------|--------|--------|--------------|---------------|
+| `MaxBlockSizeBytes` | 100 | MiB | Maximum total size of the protobuf encoded block, a hard coded constant acting as a cap for `MaxBytes`. | Hard coded |
+| `MaxBytes`     |  21 | MiB | Determines the valid size of the entire protobuf encoded block. Is a governance modifiable parameter and is capped by `MaxBytes`. Used to regulate the amount of data gossiped in the consensus portion of the network (using the current block gossiping mechanism). | Modifiable |
```
rootulp, line 19: [nit] 

```suggestion
| `MaxBytes`     |  21 | MiB | Determines the maximum size of the entire protobuf encoded block. Is a governance modifiable parameter. Used to regulate the amount of data gossiped in the consensus portion of the network (using the current block gossiping mechanism). | Modifiable |
```

docs/architecture/adr-021-restricted-block-size.md
```diff
@@ -0,0 +1,156 @@
...
+smaller block to require a square size larger than 64. This results in only ever
+confidently achieving a soft block of squares over size 64. Implementation and
+fuzzing test for this option can be found at
+[#1743](https://github.com/celestiaorg/celestia-app/pull/1743).
+
+### Introduce a new Governance Parameter
+
+The second suggestion is to create a new governance parameter,
+`GovMaxSquareSize`, which governance could set. Now the protocol would use the
+lowest of 4 values to determine the size of a block at a given height. This has
```
rootulp, line 48: > Now the protocol would use the
lowest of 4 values to determine the size of a block at a given height

[no change needed] I interpret `MaxBlockSizeBytes` as a maximum value for `MaxBytes`. Similarly, I interpret `MaxSquareSize` as a maximum value for `GovMaxSquareSize`.

When evaluating if a square is valid, I expect that the proposed block only needs to be compared with `MaxSquareSize` and `GovMaxSquareSize` and doesn't need to be compared with the caps for those values because the param proposal that attempts to change `MaxBytes` or `GovMaxSquareSize` above their respective caps should fail.

docs/architecture/adr-021-restricted-block-size.md
```diff
@@ -0,0 +1,156 @@
...
+## Decision
+
+Option 2: Introduce a new parameter, `GovMaxSquareSize`. After implemented, the above chart will look like:
+
+| Parameter        | Size   | Unit   | Description  | Value Control |
+|------------------|--------|--------|--------------|---------------|
+| `MaxBlockSizeBytes` | 100 | MiB | Maximum total size of the protobuf encoded block, a hard coded constant acting as a cap for `MaxBytes`. | Hard coded |
+| `MaxBytes`     |  ~1.8 | MiB | Determines the valid size of the entire protobuf encoded block. Is a governance modifiable parameter and is capped by `MaxBytes`. Used to regulate the amount of data gossiped in the consensus portion of the network (using the current block gossiping mechanism). | Modifiable |
+| `MaxSquareSize` | ~7.5 | MiB | Determines the maximum size of the original data square. Used to regulate storage requirements for celestia-node, and the size of the data availability header. Default set to 128. | Hard coded |
+| `GovMaxSquareSize`     |  ~1.8 | MiB | Governance modifiable parameter that determines valid square sizes. Must be smaller than the `MaxSquareSize`. Default set to 64. | Modifiable |
```
rootulp, line 82:
```suggestion
| `GovMaxSquareSize`     |  ~1.8 | MiB | Governance modifiable parameter that determines the max valid square size. Must be smaller than the `MaxSquareSize`. Default set to 64. | Modifiable |
```

rootulp (approved):
:shipit:

liamsi (approved):
💪🎉👏

docs/architecture/adr-021-restricted-block-size.md
evan-forbes: merging for the sake of getting this through, but this does make sense to change

docs/architecture/adr-021-restricted-block-size.md
evan-forbes: tendermint will check first using both the bytes params no matter what, and then we check using the gov param in prepare/process proposal. So we actually end up using the lowest value of all 4 each height

PR #1754, opened 2023-05-12
chore: bump celestiaorg/cosmos-sdk to v1.13.0-sdk-v0.46.11

evan-forbes:
## Overview

bumps to the most recent version of our fork of the sdk

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #1761, opened 2023-05-13
chore!: set default consensus params and prep for restricted block size

evan-forbes:
## Overview

The base implementation for all ADR021 #1759 options

part of #1592 
blocked by #1754

## Checklist


pkg/appconsts/appconsts.go
```diff
@@ -55,30 +55,38 @@ const (
 	// in a continuation sparse share of a sequence.
 	ContinuationSparseShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes
 
-	// DefaultMaxSquareSize is the maximum original square width.
+	// MaxSquareSize is the maximum original square width.
 	//
 	// Note: 128 shares in a row * 128 shares in a column * 512 bytes in a share
 	// = 8 MiB
-	DefaultMaxSquareSize = 128
+	MaxSquareSize = 128
```
evan-forbes, line 62: most of the diff in this branch is due to this naming change

pkg/square/square.go
```diff
@@ -93,7 +93,7 @@ func (s Square) Size() uint64 {
 }
 
 func Size(len int) uint64 {
-	return uint64(math.Ceil(math.Sqrt(float64(len))))
+	return shares.RoundUpPowerOfTwo(uint64(math.Ceil(math.Sqrt(float64(len)))))
 }
```
evan-forbes, line 97: with its normal use, we don't need to do this because the len is always a power of two due to square construction, but if we want to use this value outside of that context, we need to add this

this change isn't directly releated to setting the defaults, but I think its needed for implementing some restriction on square size

rootulp (approved):

app/default_overrides.go
```diff
@@ -127,3 +130,24 @@ func getLegacyProposalHandlers() (result []govclient.ProposalHandler) {
...
+	}
+}
+
+// DefaultBlockParams returns a default BlockParams with a MaxBytes determined
+// using a goal square size.
+func DefaultBlockParams() tmproto.BlockParams {
+	return tmproto.BlockParams{
+		MaxBytes:   appconsts.DefaultMaxBytes,
+		MaxGas:     -1,
+		TimeIotaMs: 1000, // 1s
```
cmwaters, line 151: Why not 1ms?

It has been hardcoded to 1ms in mainline comet (See https://github.com/cometbft/cometbft/blob/748fe9501226dd3b7a86c514fb668096fe81a1cb/consensus/state.go#L2346)

go.mod
```diff
@@ -200,7 +200,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.12.0-sdk-v0.46.11
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.13.0-sdk-v0.46.11
```
cmwaters, line 203: 👍

test/util/network/network.go
```diff
@@ -83,20 +83,20 @@ func DefaultConfig() network.Config {
...
-		AccountTokens:   sdk.TokensFromConsensusPower(1000, sdk.DefaultPowerReduction),
-		StakingTokens:   sdk.TokensFromConsensusPower(500, sdk.DefaultPowerReduction),
-		BondedTokens:    sdk.TokensFromConsensusPower(100, sdk.DefaultPowerReduction),
-		PruningStrategy: pruningtypes.PruningOptionNothing,
-		CleanupDir:      true,
-		SigningAlgo:     string(hd.Secp256k1Type),
-		KeyringOptions:  []keyring.Option{},
-		PrintMnemonic:   false,
+		GenesisState:         app.ModuleBasics.DefaultGenesis(encCfg.Codec),
+		TargetHeightDuration: 2 * time.Second,
```
cmwaters, line 87: 🙏  bet me to it :)

test/util/testnode/full_node.go
```diff
@@ -226,6 +228,27 @@ func DefaultNetwork(t *testing.T, blockTime time.Duration) (accounts []string, c
...
+// can be accessed using the returned accounts.
+func DefaultNetwork(t *testing.T) (accounts []string, cctx Context) {
+	// we create an arbitrary number of funded accounts
+	accounts = make([]string, 300)
+	for i := 0; i < 300; i++ {
+		accounts[i] = tmrand.Str(9)
+	}
+
+	tmCfg := DefaultTendermintConfig()
+	tmCfg.Consensus.TargetHeightDuration = time.Millisecond * 400
```
cmwaters, line 247: We could probably get away with doing this quicker. It means that certain height dependent tests will operate faster

cmwaters (commented):
Nice work. Mainly the one about IotaMs that I think needs to be addressed

cmwaters (commented):
Nice work. Mainly the one about IotaMs that I think needs to be addressed

test/util/testnode/full_node.go
evan-forbes: iirc, I think we increased it to 400 because of CI flake issues, but I don't remember if this actually fixed anything so we can try them at 1ms as it would certainly be nice

c5c09b3a23e483648c5cbc03501817fbcc03a66e

app/default_overrides.go
evan-forbes: this is just what it used to be, but we can definitely use 1ms. c5c09b3a23e483648c5cbc03501817fbcc03a66e

question tho, if its 1ms, shouldn't we (royal we as in everyone who uses comet) just get rid of it?

pkg/square/square.go
```diff
@@ -93,8 +93,8 @@ func (s Square) Size() int {
 	return Size(len(s))
 }
 
-func Size(len int) uint64 {
-	return shares.RoundUpPowerOfTwo(uint64(math.Ceil(math.Sqrt(float64(len)))))
+func Size(len int) int {
+	return int(math.Ceil(math.Sqrt(float64(len))))
```
rootulp: The merge commit no longer rounds up, do we still want it @evan-forbes ? Related: https://github.com/celestiaorg/celestia-app/pull/1761#discussion_r1192889489

rootulp (commented):
Not approving b/c I don't want this to auto-merge before @evan-forbes sees comment

pkg/square/square.go
evan-forbes: yeah thanks good catch, how this wasn't a conflict idek. yeah we need it in order to get tests to pass here and on the branches based on this one

evan-forbes:
if we merge this one, then we should be able to write pretty easy integration tests for the mint module without changing constants, because we can change the `TimeIota` consensus param to something large. If our changes to the timeout commit don't mess it up, tendermint should automatically add the timeioata to the header despite that time not actually passing. This way, we could make the header time increase by 3 days each block.

pkg/square/square.go
```diff
@@ -94,7 +94,7 @@ func (s Square) Size() int {
 }
 
 func Size(len int) int {
-	return int(math.Ceil(math.Sqrt(float64(len))))
+	return shares.RoundUpPowerOfTwo(int(math.Ceil(math.Sqrt(float64(len)))))
```
rootulp, line 97: 👍

rootulp (approved):

rootulp (approved):

cmwaters (approved):

PR #1765, opened 2023-05-14
feat!: use `consensus.MaxBytes` to determine `GovMaxSquareSize` (option 3 from ADR021)

### LINKED ISSUE #1592: EPIC: Hard capped blocksize

evan-forbes:
## Overview

the implementation of option 3 ADR021 #1759, blocked by first making a decision of which option to use.

closes #1592

## Checklist

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -1,7 +1,6 @@
 package app
 
 import (
-	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/da"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
 	"github.com/celestiaorg/celestia-app/pkg/square"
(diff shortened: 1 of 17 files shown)


evan-forbes:
closing in favor of #1772 per the decision for ADR021 (add new param for gov max square size)

evan-forbes:
closing in favor of #1772

PR #1766, opened 2023-05-14
feat!: reduce subtree root threshold to 64

evan-forbes:
## Overview

per sync discussion/agreement, we are decreasing the subtree root threshold to 64 instead of 128. This has the tradeoff of slightly increasing the % padding for real world usecases while decreasing the size of blob inclusion proofs significantly.

## Checklist


evan-forbes:
omg github. I call bs. let us moorrrggee
Screenshot from 2023-05-15 19-11-16

PR #1767, opened 2023-05-15
docs: adr-010 fix link

SweeXordious:
## Overview

Fixing a typo


SweeXordious:
Also, I guess we should check why the CI didn't catch this... normally we should have an md linter

rootulp (approved):

staheri14 (approved):

PR #1770, opened 2023-05-15
feat: support parsing out of context compact shares

Manav-Aggarwal:
## Overview

Adds logic to handle out of context compact shares and uses it in `parseCompactShares`. Also, adds a corresponding test with different random slices of list of shares to check they are interpretable back into Txs.

Closes: #802 

## Checklist


test/util/testfactory/txs.go
```diff
@@ -31,3 +32,27 @@ func GenerateRandomTxs(count, size int) types.Txs {
 	}
 	return txs
 }
+
+func GetRandomSlice(size int) (start int, length int) {
+	length = rand.Intn(size + 1)
+	start = rand.Intn(size - length + 1)
+	return
```
rootulp, line 39: [uber nit] this codebase strives to avoid [naked returns](https://go.dev/tour/basics/7) because for large function bodies they can be less readable than non-naked returns. This isn't a large function body but proposal to use non-naked return for readability:

```suggestion
	return start, length
```

Maintainers should consider enabling a lint rule that complains if naked returns are used.

test/util/testfactory/txs.go
```diff
@@ -31,3 +32,27 @@ func GenerateRandomTxs(count, size int) types.Txs {
...
+}
+
+// Returns whether subTxList is a subarray of txList
```
rootulp, line 42:
```suggestion
// CheckSubArray returns whether subTxList is a subarray of txList
```

pkg/shares/compact_shares_test.go
```diff
@@ -112,6 +112,23 @@ func Test_processCompactShares(t *testing.T) {
 	}
 }
 
+func TestParseOutOfContextShares(t *testing.T) {
```
rootulp, line 115: [optional] in addition to this test, should we consider adding a test case for https://github.com/rootulp/celestia-app/blob/01282c42a8b1d7ea2cfd325c2ed10aed4c317966/pkg/shares/parse.go#L12 that tests out of context shares?

Motivation: `parseCompactShares` is private and `ParseTxs` is public.

rootulp (approved):
Overall LGTM! Great work

test/util/testfactory/txs.go
evan-forbes: is our linter useless lol

pkg/shares/compact_shares_test.go
Manav-Aggarwal: Since`ParseTxs` uses `parseCompactShares` under the hood so instead of adding another test, modified this test using `ParseTxs`to make it simpler.

rootulp (dismissed):
🚢

pkg/shares/compact_shares_test.go
```diff
@@ -112,6 +112,20 @@ func Test_processCompactShares(t *testing.T) {
...
+	txs := testfactory.GenerateRandomlySizedTxs(1000, 150)
+	txShares, _, _, err := SplitTxs(txs)
+	require.NoError(t, err)
+
+	for i := 0; i < 1000; i++ {
+		start, length := testfactory.GetRandomSlice(len(txShares))
+
+		resTxs, err := ParseTxs(txShares[start : start+length])
+		require.NoError(t, err)
+		assert.True(t, testfactory.CheckSubArray(txs, resTxs))
```
evan-forbes, line 125: [blocking]
this is checking that the transactions returned were included, but I don't think its testing that all of the expected transactions were returned, correct? 

If so, then I think we _also_ need to also check for that. Perhaps we could use the shareRange returned somehow?

for example:

```go
        txShares, _, shareRanges, err := SplitTxs(txs)
	require.NoError(t, err)

	for key, r := range shareRanges {
		resTxs, err := ParseTxs(txShares[r.Start:r.End])
		require.NoError(t, err)
		has := false
		for _, tx := range resTxs {
			if tx.Key() == key {
				has = true
				break
			}
		}
		assert.True(t, has)
	}
}
```

evan-forbes (commented):
thanks for adding this @Manav-Aggarwal!!! :slightly_smiling_face: 

mostly just had a few questions. In future PRs, I think we might be able to test this a bit more thoroughly by incorporating it into some integration tests

pkg/shares/shares.go
```diff
@@ -191,6 +191,53 @@ func (s *Share) rawDataStartIndex() int {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// This means share is the last share and does not have any transaction beginning in it
+	if rawDataStartIndexUsingReserved == 0 {
+		return []byte{}, nil
+	}
+	if len(s.data) < rawDataStartIndexUsingReserved {
+		return rawData, fmt.Errorf("share %s is too short to contain raw data", s)
```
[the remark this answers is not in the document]
Manav-Aggarwal: This error message path should never get hit unless someone tries to create a `Share` without using the share builder.  The share builder populates all the share metadata inside it, namely `namespace`, `sequenceLen`, `infoByte`, and `reservedBytes`. The same error message lives in `rawDataStartIndex`.

pkg/shares/shares.go
```diff
@@ -191,6 +191,53 @@ func (s *Share) rawDataStartIndex() int {
...
+
+func (s *Share) rawDataStartIndexUsingReserved() (int, error) {
+	isStart, err := s.IsSequenceStart()
+	if err != nil {
+		panic(err)
+	}
+	isCompact, err := s.IsCompactShare()
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
Manav-Aggarwal: We also have the same two panics in `rawDataStartIndex`, maybe those should be modified too then? Bubbled it up for this PR.

pkg/shares/shares.go
```diff
@@ -191,6 +191,53 @@ func (s *Share) rawDataStartIndex() int {
...
+func (s *Share) RawDataUsingReserved() (rawData []byte, err error) {
+	rawDataStartIndexUsingReserved, err := s.rawDataStartIndexUsingReserved()
+	if err != nil {
+		return nil, err
+	}
+
+	// This means share is the last share and does not have any transaction beginning in it
+	if rawDataStartIndexUsingReserved == 0 {
+		return []byte{}, nil
+	}
```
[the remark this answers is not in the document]
Manav-Aggarwal: I don't think so, a `tx` cannot start at zero index of a share because a properly constructed valid share has the namespace at index 0. As long as a share has a new transaction beginning in it, reserved bytes should not be 0.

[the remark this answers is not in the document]
rootulp: > this is the first compact share

The first compact share has reserved bytes populated.

> or the tx happens to start at zero

Not possible because the namespace version exists at index 0.

pkg/shares/shares.go
```diff
@@ -191,6 +191,53 @@ func (s *Share) rawDataStartIndex() int {
...
+	if len(s.data) < rawDataStartIndexUsingReserved {
+		return rawData, fmt.Errorf("share %s is too short to contain raw data", s)
+	}
+
+	return s.data[rawDataStartIndexUsingReserved:], nil
+}
+
+func (s *Share) rawDataStartIndexUsingReserved() (int, error) {
```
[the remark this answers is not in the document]
rootulp: Proposed simplification:

```go
func (s *Share) rawDataStartIndexUsingReserved() (int, error) {
	isStart, err := s.IsSequenceStart()
	if err != nil {
		return 0, err
	}
	isCompact, err := s.IsCompactShare()
	if err != nil {
		return 0, err
	}
	if !isCompact {
		return 0, fmt.Errorf("share %s is not a compact share", s)
	}

	index := appconsts.NamespaceSize + appconsts.ShareInfoBytes
	if isStart {
		index += appconsts.SequenceLenBytes
	}

	reservedBytes, err := ParseReservedBytes(s.data[index : index+appconsts.CompactShareReservedBytes])
	if err != nil {
		return 0, err
	}
	return int(reservedBytes), nil
}
```

test/util/testfactory/txs.go
```diff
@@ -31,3 +32,27 @@ func GenerateRandomTxs(count, size int) types.Txs {
 	}
 	return txs
 }
+
+func GetRandomSlice(size int) (start int, length int) {
+	length = rand.Intn(size + 1)
+	start = rand.Intn(size - length + 1)
+	return start, length
+}
```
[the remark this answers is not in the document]
Manav-Aggarwal: Modified the name to use `SubSlice` and added a comment explaining it.

pkg/shares/shares.go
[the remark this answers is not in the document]
evan-forbes: oh shoot okay forever ago the index was relative to that index, this makes more sense thanks

pkg/shares/compact_shares_test.go
Manav-Aggarwal: Yes, this is checking that the sub-list of transactions returned is a subarray of the original list of transactions. Added another test that covers shareRanges and checks a transaction in a shareRange is parsed correctly.

evan-forbes (dismissed):
dope, thanks again @Manav-Aggarwal

evan-forbes:
ohhh boy that's _not_ good lol I can help debug as that could be an issue in other code

```
--- FAIL: TestSquareBlobPostions/case3 (0.00s)
        square_test.go:202: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/pkg/square/square_test.go:202
            	            	share &{�����������������������������} is not a compact share so does not contain reserved bytes
            	Test:       	TestSquareBlobPostions/case3
    --- FAIL: TestSquareBlobPostions/case8 (0.00s)
        square_test.go:202: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/pkg/square/square_test.go:202
            	            	share &{�����������������������������} is not a compact share so does not contain reserved bytes
            	Test:       	TestSquareBlobPostions/case8
            	```

**Comment by evan-forbes** [comment]
ahh nvm, the fact that it was picked up in a square test threw me off, do we need to check that its a compact share 
in https://github.com/Manav-Aggarwal/celestia-app/blob/8d6e812c504dc4fdfa2bb8cc0759592a42ea839f/pkg/shares/shares.go#L218-L224 ?

should we perform that check earlier somehow?

**Comment by Manav-Aggarwal** [comment]
> ahh nvm, the fact that it was picked up in a square test threw me off, fortunately its that bad, do we need to check that its a compact share in https://github.com/Manav-Aggarwal/celestia-app/blob/8d6e812c504dc4fdfa2bb8cc0759592a42ea839f/pkg/shares/shares.go#L218-L224 ?
> 
> should we perform that check earlier somehow?

Removed the commit that introduced that check for now. I don't think we necessarily need to have that check since the assumption that `ParseTxs` is only being used with `compactShares` as input is incorrect. So we can keep the `rawDataStartIndex` method be more versatile and just process non-compact shares without using reserved bytes.

**Review by evan-forbes — DISMISSED** [review]

File: pkg/shares/parse_compact_shares.go [hunk 10]
```diff
@@ -61,7 +51,7 @@ func parseRawData(rawData []byte) (units [][]byte, err error) {
 		if err != nil {
 			return nil, err
 		}
-		if unitLen == 0 {
+		if unitLen == 0 || unitLen > uint64(len(actualData)) {
```
rootulp, line 54: [question] should `unitLen > uint64(len(actualData))` return an error?

pkg/shares/parse_compact_shares.go
```diff
@@ -73,7 +63,12 @@ func parseRawData(rawData []byte) (units [][]byte, err error) {
 // not contain the namespace ID, info byte, sequence length, or reserved bytes.
 func extractRawData(shares []Share) (rawData []byte, err error) {
 	for i := 0; i < len(shares); i++ {
-		raw, err := shares[i].RawData()
+		var raw []byte
+		if i == 0 {
+			raw, err = shares[i].RawDataUsingReserved()
```
rootulp, line 68: [question] the reserved bytes exist for all compact shares. Why does the first share in `shares` uses `.RawDataUsingReserved()` while subsequent shares use `.RawData()`?

I think this function makes assumptions on the shares that are passed as a parameter, potentially that it is only called with share sequences (?) If so, can we document via godoc or better yet enforce via the type of the function argument?

pkg/shares/shares.go
```diff
@@ -191,6 +191,51 @@ func (s *Share) rawDataStartIndex() int {
...
+func (s *Share) RawDataUsingReserved() (rawData []byte, err error) {
+	rawDataStartIndexUsingReserved, err := s.rawDataStartIndexUsingReserved()
+	if err != nil {
+		return nil, err
+	}
+
+	// This means share is the last share and does not have any transaction beginning in it
+	if rawDataStartIndexUsingReserved == 0 {
+		return []byte{}, nil
+	}
```
rootulp, line 204: [question] I expect this conditional to evaluate to `true` for the last share in the transaction namespace. However, for that share, I would expect that `RawDataUsingReserved` returns the raw data from the last share (i.e. the remainder of the last transaction that overflowed into the last share).

```go
func TestRawDataUsingReserved(t *testing.T) {
	type testCase struct {
		name  string
		share Share
		want  []byte
	}
	infoByte, err := NewInfoByte(appconsts.ShareVersionZero, false)
	assert.NoError(t, err)
	reservedBytes := []byte{0, 0, 0, 0}
	lastTxRemainingBytes := []byte{1, 2, 3}

	lastTransactionShareData := []byte{}
	lastTransactionShareData = append(lastTransactionShareData, appns.TxNamespace.Bytes()...)
	lastTransactionShareData = append(lastTransactionShareData, byte(infoByte))
	lastTransactionShareData = append(lastTransactionShareData, reservedBytes...)
	lastTransactionShareData = append(lastTransactionShareData, lastTxRemainingBytes...)

	testCases := []testCase{
		{
			name: "last transaction share",
			share: Share{
				data: lastTransactionShareData,
			},
			want: lastTxRemainingBytes,
		},
	}

		t.Run(tc.name, func(t *testing.T) {
			rawData, err := tc.share.RawDataUsingReserved()
			require.NoError(t, err)
			assert.Equal(t, tc.want, rawData)
		})
	}
}
```

results in

```
            	            	expected: []byte{0x1, 0x2, 0x3}
            	            	actual  : []byte{}
```

rootulp (commented):
Apologies for comments after already approving this. On re-review I have a few questions

pkg/shares/parse_compact_shares.go
Manav-Aggarwal: added a comment explaining it

pkg/shares/parse_compact_shares.go
Manav-Aggarwal: updated the doc string covering the comment.

rootulp (approved):
Discussed in-person. Thanks for adding the comments!

rootulp (approved):
Discussed in-person. Thanks for adding the comments!

pkg/shares/shares.go
Manav-Aggarwal: Since `RawDataUsingReserved` is only called for the first share when parsing out of context shares, in the case when the share only contains remainder of the last transaction that overflowed into this share, there are no complete transactions, so returning an empty byte array, in this case, would make sense.

evan-forbes (approved):

Issue #1753, opened 2023-05-12
Decrease the namespace size from 33 => 29 bytes


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/1739

## Problem

The namespace is currently: version (1 byte) + ID (32 bytes) = namespace (33 bytes)

## Proposal

1. Update the namespace to: version (1 byte) + ID (28 bytes) = namespace (29 bytes)
2. Update docs / specs

Issue #1522, opened 2023-03-22
Consider removing the byte level unit tests for `pkgs/shares`


rootulp:
## Context

[pkg/shares](https://github.com/celestiaorg/celestia-app/tree/main/pkg/shares) currently contains many byte level unit tests for share encoding.

## Problem

The unit tests are brittle and require manual updating every time we change share encoding. 

## Proposal

Consider isolating the byte level unit tests to the share builder. If we unit test the share builder exhaustively, then higher level abstractions (i.e. other files in pkg/shares) can test via the exposed methods on a `Share` rather than the raw bytes.

PR #1771, opened 2023-05-15
feat!: decrease namespace from 33 to 29 bytes

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1753


evan-forbes:
I did two :shipit:, which I rarely do, so interpret that as you will

SweeXordious (approved):
Deferring to other reviewers for final approval

pkg/da/data_availability_header_test.go
```diff
@@ -26,7 +26,7 @@ func TestNilDataAvailabilityHeaderHashDoesntCrash(t *testing.T) {
 
 func TestMinDataAvailabilityHeader(t *testing.T) {
 	dah := MinDataAvailabilityHeader()
-	expectedHash := []byte{0xe9, 0x5, 0x28, 0x49, 0xf, 0x1d, 0x51, 0x67, 0x29, 0x2c, 0x1f, 0x1b, 0x83, 0xe1, 0x74, 0x2a, 0x27, 0x48, 0x17, 0x34, 0x12, 0xc9, 0x1d, 0xf7, 0xdd, 0x1, 0x96, 0x78, 0xa4, 0x62, 0xb9, 0x77}
+	expectedHash := []byte{0x3d, 0x96, 0xb7, 0xd2, 0x38, 0xe7, 0xe0, 0x45, 0x6f, 0x6a, 0xf8, 0xe7, 0xcd, 0xf0, 0xa6, 0x7b, 0xd6, 0xcf, 0x9c, 0x20, 0x89, 0xec, 0xb5, 0x59, 0xc6, 0x59, 0xdc, 0xaa, 0x1f, 0x88, 0x3, 0x53}
```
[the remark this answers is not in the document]
MSevey: good follow up to document this imo.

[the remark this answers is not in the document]
rootulp: > I am wondering how these expected hashes are generated?

I generate them by running the test after a share encoding change and copying the actual output to the expected output. Related doc: https://github.com/celestiaorg/celestia-app/blob/207ec1a233d3a31bddbe913e9da870c032a20e7a/x/blob/types/payforblob_test.go#L64-L66

pkg/da/data_availability_header_test.go
```diff
@@ -42,13 +42,13 @@ func TestNewDataAvailabilityHeader(t *testing.T) {
 	tests := []test{
 		{
 			name:         "typical",
-			expectedHash: []byte{0x5b, 0x27, 0x3e, 0x3a, 0x5d, 0x9e, 0x90, 0x25, 0x58, 0x21, 0xb7, 0xe0, 0x4d, 0x4b, 0xaa, 0xde, 0x37, 0xa6, 0x6f, 0xcc, 0xd, 0x16, 0x6f, 0x9e, 0xe0, 0x7f, 0xbe, 0x8, 0xb4, 0x41, 0xc8, 0xa6},
+			expectedHash: []byte{0xb5, 0x6e, 0x4d, 0x25, 0x1a, 0xc2, 0x66, 0xf4, 0xb9, 0x1c, 0xc5, 0x46, 0x4b, 0x3f, 0xc7, 0xef, 0xcb, 0xdc, 0x88, 0x80, 0x64, 0x64, 0x74, 0x96, 0xd1, 0x31, 0x33, 0xf0, 0xdc, 0x65, 0xac, 0x25},
 			squareSize:   2,
 			shares:       generateShares(4),
```
staheri14, line 47: [nit] Can we use `2*2` instead of `4` in the `generateShares(4)`, it is more clear how 4 is calculated and why.

pkg/shares/share_sequence_test.go
```diff
@@ -113,7 +113,7 @@ func Test_sparseSharesNeeded(t *testing.T) {
 		{appconsts.FirstSparseShareContentSize + appconsts.ContinuationCompactShareContentSize*99, 100},
 		{1000, 3},
 		{10000, 21},
-		{100000, 210},
+		{100000, 208},
```
staheri14, line 116: [Question/suggestion] Is there a specific formula that is used to derive the value of "208"? If such a formula exists, would it be possible to include a document here explaining the details of that formula? makes it easier to understand and maintain the test.

pkg/shares/share_splitting_test.go
```diff
@@ -38,7 +38,7 @@ func TestSplitTxs_forTxShares(t *testing.T) {
 						[]byte{
 							0x1,                // info byte
 							0x0, 0x0, 0x0, 0x2, // 1 byte (unit) + 1 byte (unit length) = 2 bytes sequence length
-							0x0, 0x0, 0x0, 0x2a, // reserved bytes
+							0x0, 0x0, 0x0, 0x26, // reserved bytes
```
staheri14, line 41: [Question] Why `0x2a` is replaced with `0x26`?

staheri14: [Suggestion, non-blocking] It would be incredibly helpful for code readability to have a description of each test function, providing a clear explanation of what is being precisely tested.

pkg/shares/share_splitting_test.go
```diff
@@ -156,10 +156,10 @@ func TestSplitTxs_forTxShares(t *testing.T) {
 						appns.TxNamespace.Bytes(),
 						[]byte{
 							0x0,                 // info byte
-							0x0, 0x0, 0x0, 0x52, // reserved bytes
+							0x0, 0x0, 0x0, 0x4a, // reserved bytes
```
staheri14, line 159: [Suggestion, non-blocking] Could we include visual representations illustrating how the data is organized within their respective shares? by visual representation I mean something similar to the ones in [this example](https://github.com/celestiaorg/celestia-app/blob/a4aec0d531bafc6df608206d376b8a431938dc04/pkg/square/builder.go#L194).

staheri14 (approved):
LGTM! added some suggestions and comments all non-blocking.

staheri14 (approved):
LGTM! added some suggestions, questions, and comments all non-blocking.

pkg/shares/share_splitting_test.go
rootulp: These tests are probably too low level and we're considering removing them. https://github.com/celestiaorg/celestia-app/issues/1522

They were written when we were making a lot of low level byte changes to a share but totally agreed that they're cumbersome to maintain.

rootulp: > [Question] Why 0x2a is replaced with 0x26?

Because the reserved bytes contains the index of the first transaction in the share. Since the namespace size decreased, the index had to be decreased. 

`0x2a` => `42` 
`0x26` => `38`

The namespace size decreased by 4 bytes so the index of the first transaction had to be reduced by 4.

pkg/da/data_availability_header_test.go
rootulp: Sure! https://github.com/celestiaorg/celestia-app/issues/1788

pkg/shares/share_splitting_test.go
rootulp: https://github.com/celestiaorg/celestia-app/issues/1789

PR #1772, opened 2023-05-16
feat!: add GovMaxSquareAddr

### LINKED ISSUE #1737: make default block size limit be 1.8mb

### LINKED ISSUE #1782: txsim.Run occasionally blocks indefinitely
in #1772 we're adding a new test that incorporates txsim to fill blocks, and I've noticed that sometimes `txsim.Run` will block indefinitely causing the tests to timeout both locally and in CI.

I think this might have to do with this loop when broadcasting a tx
https://github.com/celestiaorg/celestia-app/blob/5fc83cf8c822b48dbcfc9f867467ab75f5adf0bf/test/txsim/client.go#L229-L246

since we're not checking if the original context has been cancelled

evan-forbes:
## Overview

Implements ADR 021 option 2 by adding the GovMaxSquareSize param to the blob module. That param should be a consensus param, but I didn't want to modify those due to the hassel of including some change in core, the sdk treating those as different from the rest of the genesis state, and wanted to keep all the logic in the app.

closes #1737 also fixes #1782 

## Checklist


app/test/integration_test.go
```diff
@@ -93,132 +92,6 @@ func (s *IntegrationTestSuite) WaitForBlocks(blockCount int) {
 	}
 }
 
-func (s *IntegrationTestSuite) TestMaxBlockSize() {
```
evan-forbes, line 96: refactored this test to use our latest testing infra and moved to its own test.

app/square_size.go
```diff
@@ -0,0 +1,29 @@
...
+func (app *App) GovMaxSquareSize(ctx sdk.Context) int {
+	// TODO: fix hack that forces the max square size for the first height to
+	// 64. This is due to tendermint not technically supposed to be calling
+	// PrepareProposal when heights are not >= 1. This is remedied in versions
+	// of the sdk and coment that have full support of PreparePropsoal, although
+	// celestia-app does not currently use those. see this PR for more
+	// details https://github.com/cosmos/cosmos-sdk/pull/14505
+	if ctx.BlockHeader().Height == 0 {
+		return int(appconsts.DefaultGovMaxSquareSize)
+	}
```
evan-forbes, line 20: pls review, not sure of the best solution around this, but this seemed the most reasonable. 

I think we should also add a check in PrepareProposal that will simply propose an empty block for the first block (height ==0), this way if there are enough transactions in the first block to reach a govMaxSquareSize, we will likely avoid having to encode this logic in future upgrades forever!!

app/test/fuzz_abci_test.go
```diff
@@ -30,26 +30,64 @@ func TestPrepareProposalConsistency(t *testing.T) {
...
-		{"one hundred normal sized single blob transactions", 100, 1, 400000},
-		{"many single share multi-blob transactions", 1000, 100, 400},
-		{"one hundred normal sized multi-blob transactions", 100, 4, 400000},
+		// running these tests more than once in CI will sometimes timout, so we
+		// have to run them each once per square size. However, we can run these
+		// more locally by increasing the iterations.
+		{"many small single share single blob transactions", 1000, 1, 400, 1},
+		{"one hundred normal sized single blob transactions", 100, 1, 400000, 1},
+		{"many single share multi-blob transactions", 1000, 100, 400, 1},
+		{"one hundred normal sized multi-blob transactions", 100, 4, 400000, 1},
```
evan-forbes, line 45: note to self, we should try to make these tests faster. Generating all the txs or something is taking forever, so we're only running each of these once per size

app/test/square_size_test.go
```diff
@@ -0,0 +1,262 @@
...
+	// times, so we have to retry many times.
+	var newMaxBytes int64
+	for i := 0; i < 20; i++ {
+		cpresp, err := s.cctx.Client.ConsensusParams(s.cctx.GoContext(), nil)
+		if err != nil || cpresp == nil {
+			continue
+		}
+		newMaxBytes = cpresp.ConsensusParams.Block.MaxBytes
+	}
+	require.Equal(t, maxBytes, newMaxBytes)
```
evan-forbes, line 257: this test currently assumes that the proposed values pass, and check them here. in the future, we might want to get higher coverage by also checking that we can't go above the hardcoded values

test/util/testnode/full_node.go
```diff
@@ -147,13 +147,15 @@ func DefaultTendermintConfig() *config.Config {
 	// during tests.
 	tmCfg.Consensus.TargetHeightDuration = 300 * time.Millisecond
 	tmCfg.Consensus.TimeoutPropose = 200 * time.Millisecond
-	tmCfg.Mempool.MaxTxBytes = 22020096 // 21MB
+	tmCfg.Mempool.MaxTxBytes = 220020096 // 210MB
+	tmCfg.RPC.MaxBodyBytes = 220020096   // 210MB
```
evan-forbes, line 151: remove any limit on txsim

evan-forbes:
the new test can be a bit flakey as it appears that txsim.Run will occasionally get blocked for whatever reason with the larger parameter tests, will debug tmrw and opening up for review

pkg/appconsts/appconsts.go
```diff
@@ -55,22 +55,30 @@ const (
...
-	DefaultMaxSquareSize = 128
+	MaxSquareSize = 128
+
+	// DefaultGovMaxSquareSize is the default value for the governance
+	// modifiable max square size.
+	DefaultGovMaxSquareSize = 64
+
+	// DefaultMaxBytes is the default value for the maximum number of bytes
+	// allowed in a valid block.
+	DefaultMaxBytes = DefaultGovMaxSquareSize * DefaultGovMaxSquareSize * ContinuationSparseShareContentSize
```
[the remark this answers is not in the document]
evan-forbes: > out of curiosity, why ContinuationSparseShareContentSize instead of ShareSize?

this is mainly just to not account for namespaces 

> However I don't expect it to be the true upper bound b/c we expect some compact shares for the PFB txs associated with the blobs.

I'm a bit confused by this, how are we not accounting for the PFBs here or what do compact shares have to do with this? max bytes convers BlobTxs, which contain the PFBs no? this is just to count the total number of bytes in a square, including compact shares, but not including any block encoding.

test/util/testnode/genesis_options.go
```diff
@@ -0,0 +1,44 @@
...
+		gs := v1.DefaultGenesisState()
+		gs.DepositParams.MinDeposit = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))
+		gs.TallyParams.Quorum = "0.000001"
+		gs.TallyParams.Threshold = "0.000001"
+		vp := time.Second * 2
+		gs.VotingParams.VotingPeriod = &vp
+		state[govtypes.ModuleName] = codec.MustMarshalJSON(gs)
+		return state
+	}
+}
```
evan-forbes, line 44: had to do some refactoring around some of our integration test helpers to accomodate being able to change the genesis state.

I would prefer the builder pattern here, but left that for a different refactor.

pkg/appconsts/appconsts.go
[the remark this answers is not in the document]
rootulp: > max bytes covers BlobTxs, which contain the PFBs no?

yep!

> what do compact shares have to do with this?

Sorry I should've clarified. `ContinuationSparseShareContentSize` is slightly larger than `ContinuationCompactShareContentSize`

> this is mainly just to not account for namespaces

Makes sense. I wanted to call out that `ContinuationSparseShareContentSize` won't account for the bytes needed for compact shares (i.e. reserved bytes) and since a full square with contain at least one compact share for the PFB tx and at least one starting sparse share, the upper bound is slightly too high. In other words

```
DefaultMaxBytes = DefaultGovMaxSquareSize * DefaultGovMaxSquareSize * ContinuationSparseShareContentSize
DefaultMaxBytes = 64 * 64 * 478
DefaultMaxBytes = 1,957,888
```

But a valid block can't have 1,957,888 bytes because that assumes the entire square is continuation blob shares. Therefore we could have a slightly lower upper bound. Alternatively can we use the more readable `ShareSize` because `MaxSquareSize` will get hit and take affect before `DefaultMaxBytes`?

test/txsim/client.go
```diff
@@ -227,6 +227,9 @@ func (tc *TxClient) Broadcast(ctx context.Context, txBuilder sdkclient.TxBuilder
 	}
 
 	for {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
 		subctx, cancel := context.WithTimeout(ctx, rpcContextTimeout)
 		defer cancel()
```
evan-forbes, line 234: also, this is the change that seems to fix the issue in #1782 since its no longer occuring and this seems to make sense, but would defer to @cmwaters on if we should be doing this

pkg/appconsts/appconsts.go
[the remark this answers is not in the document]
evan-forbes: ahh ok I see, yeah this purely because it seemed like an easy way to get a more accurate estimate. We could switch to share size and it will not have an effect (especifally after https://github.com/celestiaorg/celestia-core/issues/1007 that is)

perhaps we should just use a round number then

test/util/testnode/full_node.go
rootulp, line 151: [blocking]

```suggestion
	tmCfg.Mempool.MaxTxBytes = 220020096 // 21 MB
	tmCfg.RPC.MaxBodyBytes = 220020096   // 21 MB
```

Ref: https://www.flightpedia.org/convert/21-megabytes-to-bytes.html

rootulp (commented):
Great work! Will approve after blocking suggestions

test/util/testnode/full_node.go
evan-forbes: wait 220_020_096 is very close to 210MB no?

rootulp: hmm I'm confused because the comment was updated but the actual number wasn't. Should this be 21 MB or 210MB. Either way, proposal to use underscores in the number to make it easier to read:


21 MB
```go
	tmCfg.Mempool.MaxTxBytes = 22_020_096 // 21 MB
	tmCfg.RPC.MaxBodyBytes = 22_020_096   // 21 MB
```

210 MB

```go
	tmCfg.Mempool.MaxTxBytes = 220_200_960 // 210 MB
	tmCfg.RPC.MaxBodyBytes = 220_200_960   // 210 MB
```

evan-forbes: https://github.com/celestiaorg/celestia-app/pull/1772/commits/0467644818d6e02ab2d7cc45be9139009b2d91c2

test/util/testnode/full_node.go
```diff
@@ -147,8 +147,8 @@ func DefaultTendermintConfig() *config.Config {
 	// during tests.
 	tmCfg.Consensus.TargetHeightDuration = 300 * time.Millisecond
 	tmCfg.Consensus.TimeoutPropose = 200 * time.Millisecond
-	tmCfg.Mempool.MaxTxBytes = 220020096 // 210MB
-	tmCfg.RPC.MaxBodyBytes = 220020096   // 210MB
+	tmCfg.Mempool.MaxTxBytes = 220_000_960 // 210MB
+	tmCfg.RPC.MaxBodyBytes = 220_000_960   // 210MB
```
rootulp, line 151: This is 200_000 bytes less than 210 MB

```suggestion
	tmCfg.Mempool.MaxTxBytes = 220_200_960 // 210MB
	tmCfg.RPC.MaxBodyBytes = 220_200_960   // 210MB
```

Ref: https://www.flightpedia.org/convert/210-megabytes-to-bytes.html

rootulp (approved):

test/util/testnode/full_node.go
evan-forbes: damn I actually tried to copy from the site you linked there to be sure but messed up apparently :sweat_smile: 

as mentioned in a different comment, this value is _purlely_ to remove any barriers for being able to submit transactions to test nodes, there is no rhyme or reason to it besides it being absolutely obtuse. 

therefore I'm just using powers of ten like god intended and the max square size
https://github.com/celestiaorg/celestia-app/pull/1772/commits/7334b9005c81576a848c4e75c66c289eba0e5957

app/prepare_proposal.go
```diff
@@ -25,7 +24,7 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 
 	// build the square from the set of valid and prioritised transactions.
 	// The txs returned are the ones used in the square and block
-	dataSquare, txs, err := square.Build(txs, appconsts.MaxSquareSize)
+	dataSquare, txs, err := square.Build(txs, app.GovMaxSquareSize(sdkCtx))
```
SweeXordious, line 27: wouldn't this have conflicts with: https://github.com/celestiaorg/celestia-app/pull/1768/files#diff-abccb046bddf0559ea4defb5570fe3d6b417f443ddf37c657a7b8f1e485903ea ?

SweeXordious (approved):
Differing to other reviewers for final review

SweeXordious (approved):
Differing to other reviewers for final review

evan-forbes:
as an shameless bribe/hostage situation to encourage reviewers to click the green button, I have increased the scope of this PR and included two changes that should reduce flakiness

https://github.com/celestiaorg/celestia-app/pull/1772/commits/21b271e512546a099e4afc73f650673969225e03

https://github.com/celestiaorg/celestia-app/pull/1772#discussion_r1196668226

test/txsim/client.go
MSevey: is there error check suppose to also catch the context `Done()` event?

if so, shouldn't we just do this:
```go
for {
  select {
    case: <- ctx.Done()
        return nil, ctx.Err()
    default:
  }
  ...
}
```

little bit of a nit, but I think this is more clear that we are managing the for loop based on the ctx being active.

rootulp (dismissed):

app/test/square_size_test.go
MSevey: do we have another test that is checking that the values pass then? or put another way, we are always throttling to the lowest value?

if not that feels like a very baseline required test :-)

test/util/testnode/full_node.go
```diff
@@ -147,13 +148,17 @@ func DefaultTendermintConfig() *config.Config {
 	// during tests.
 	tmCfg.Consensus.TargetHeightDuration = 300 * time.Millisecond
 	tmCfg.Consensus.TimeoutPropose = 200 * time.Millisecond
-	tmCfg.Mempool.MaxTxBytes = 22020096 // 21MB
+	// remove all barriers from the testnode being able to accept very large
+	// transactions (max square size) and respond to very large queries (200MiB)
+	tmCfg.Mempool.MaxTxBytes = appconsts.MaxShareCount * appconsts.ShareSize
+	tmCfg.RPC.MaxBodyBytes = 200_000_000
```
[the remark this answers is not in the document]
evan-forbes: thanks for not blocking, for posterity what do I need in order to say 2 hundred million Bytes in shortform? 2e8 Bytes? ~200MB? unless the precision is actually called for, I would really like to avoid having to use a website or a calculator to write comments for tests :slightly_smiling_face: 

this is why I thought it was common place to implicitly switch to the metric system's "Mega" in almost all casual discussions/values.
Screenshot from 2023-05-18 11-26-07


> to include the godoc for a line immediately above

the change makes sense, so I'll change that in a followup or I'll toss it in if there are other requested changes :slightly_smiling_face:. wait, does this show up on a godoc? I thought this was just a line for anyone curious why we set these values in the tests. 

> I don't understand what is the rationale for testing "very large queries"

Literally it was just because if transaction or repsponses to those transactions exceeds that value it errors, to avoid that in general it was increased. It could be applied to any interation with the node, we mainly use it for tx or getting large blocks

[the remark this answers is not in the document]
evan-forbes: okay went ahead and hopefully fixed it as we should wait for Callum's feedback anyway, pls let me know if we should update this still or there is any clarification needed

6d564db30403e85a14da850497293dccb74eb902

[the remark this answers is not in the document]
rootulp: > what do I need in order to say 2 hundred million Bytes in shortform? 2e8 Bytes? ~200MB?

`200 MB` works so I think this could've been either:

```
tmCfg.RPC.MaxBodyBytes = 200_000_000 // 200 MB
tmCfg.RPC.MaxBodyBytes = 209_715_200 // 200 MiB
```

Note: MiB is [mebibyte](https://www.techtarget.com/searchstorage/definition/mebibyte-MiB) which is different from megabyte.

> wait, does this show up on a godoc?

Sorry, I guess it doesn't show up in a godoc. I just meant that it may help readers if the comment is applicable to the line immediately below it.

test/util/testnode/full_node.go
```diff
@@ -148,10 +148,17 @@ func DefaultTendermintConfig() *config.Config {
...
-	// remove all barriers from the testnode being able to accept very large
-	// transactions (max square size) and respond to very large queries (200MiB)
+
+	// set the mempool's MaxTxBytes to allow the testnode to accept a
+	// transaction that fills the entire square. Any blob transaction larger
+	// than the square size will still fail no matter what.
 	tmCfg.Mempool.MaxTxBytes = appconsts.MaxShareCount * appconsts.ShareSize
+
+	// remove all barriers from the testnode being able to accept very large
+	// transactions and respond to very queries with large responses (~200MB was
```
rootulp, line 158:
```suggestion
	// transactions and respond to queries with large responses (~200MB was
```

rootulp (approved):

test/util/testnode/full_node.go
evan-forbes: beat you to it :zany_face: 

https://github.com/celestiaorg/celestia-app/commit/14b3631ace8d5b2a3179a64d4d090b42850b446b

test/util/testnode/full_node.go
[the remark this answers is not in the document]
evan-forbes: > Note: MiB is [mebibyte](https://www.techtarget.com/searchstorage/definition/mebibyte-MiB) which is different from megabyte.

this is what I thought MB and MiB were

I was confused because of [this](https://github.com/celestiaorg/celestia-app/pull/1772#discussion_r1196854199), which uses MB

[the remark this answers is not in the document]
evan-forbes: so I just switched to MiB cause I thought you were refering to both interchangably

test/util/testnode/full_node.go
evan-forbes: I thought MB was metric! the above suggestion confused me, also is that site incorrect since it gives the same value for mega and mebi?

https://github.com/celestiaorg/celestia-app/pull/1772#discussion_r1198176298

rootulp: oof I think that site is broken. According to [this post](https://www.majordifferences.com/2018/03/differences-between-megabyte-and.html):

> [Home](https://www.majordifferences.com/)[computer](https://www.majordifferences.com/search/label/computer)
Differences between Megabyte and Mebibyte (MB vs MiB)
[3 Comments](https://www.majordifferences.com/2018/03/differences-between-megabyte-and.html#comments)
[Facebook](javascript:;)
[Twitter](javascript:;)
Both Megabyte (MB) and Mebibyte (MiB) are units of information or computer storage. But MB is used in two ways either as equivalent to 1000 Kb or 1024 Kb. This creates some sort of confusion. To avoid this, the International Electrotechnical Commission (IEC) (1998), the leading international organization for worldwide standardization in electrotechnology, approved as an IEC International Standard names and symbols for prefixes for binary multiples for use in the fields of data processing and data transmission. They recommended the use of Mebibyte (a combination of Mega and Binary) MiB, which always refer to 1024 Kb.

Proposal to always use the MiB variant to avoid confusion

test/util/testnode/full_node_test.go
```diff
@@ -27,10 +34,29 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
+		DefaultTendermintConfig(),
+		DefaultAppConfig(),
+		accounts,
+		SetBlobParams(ecfg.Codec, blobGenState.Params),
+	)
+	s.cctx = cctx
+	s.accounts = accounts
 }
 
 func (s *IntegrationTestSuite) Test_Liveness() {
```
cmwaters, line 62: I don't know how this tests liveness. Wouldn't we just want one block to be produced. I also don't understand why this is flaky with fast block times

evan-forbes: we can probably get rid of it and just wait one block. iirc this was just a copy of https://github.com/celestiaorg/cosmos-sdk/blob/57a798a56219a6581a78f94976f592b88ea3aa01/testutil/network/network_test.go#L37-L40

test/util/testnode/genesis_options.go
```diff
@@ -0,0 +1,44 @@
...
+
+// ImmediateProposals sets the thresholds for getting a gov proposal to very low
+// levels.
+func ImmediateProposals(codec codec.Codec) GenesisOption {
+	return func(state map[string]json.RawMessage) map[string]json.RawMessage {
+		gs := v1.DefaultGenesisState()
+		gs.DepositParams.MinDeposit = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))
+		gs.TallyParams.Quorum = "0.000001"
+		gs.TallyParams.Threshold = "0.000001"
+		vp := time.Second * 2
```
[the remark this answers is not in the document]
evan-forbes: yeah, good point that should be a comment. will include in a follow up

app/square_size.go
```diff
@@ -0,0 +1,28 @@
...
+func (app *App) GovMaxSquareSize(ctx sdk.Context) int {
+	// TODO: fix hack that forces the max square size for the first height to
+	// 64. This is due to tendermint not technically supposed to be calling
+	// PrepareProposal when heights are not >= 1. This is remedied in versions
+	// of the sdk and comet that have full support of PreparePropsoal, although
+	// celestia-app does not currently use those. see this PR for more
+	// details https://github.com/cosmos/cosmos-sdk/pull/14505
+	if ctx.BlockHeader().Height == 0 {
+		return int(appconsts.DefaultGovMaxSquareSize)
+	}
```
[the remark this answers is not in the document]
evan-forbes: same, and yeah we should definitely be able to get rid of it in the future.

Issue #1773, opened 2023-05-16
authz's MsgExec isn't supported


cmwaters:
### Summary

It's currently not possible to have a `MsgExec` wrap around `MsgPayForBlobs` within a blobTx. `ValidateBlobTx` will fail, registering it as a blobTx without a PFB. We should be supporting this

### Problem Definition

a

### Proposal

a

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

evan-forbes:
changed this from needs triage to needs discussion and unplanned, since this is consensus breaking and we do not yet have a plan to include this.

that being said, after #2977, we don't need to be as afraid of making the unmarshalling of PFBs more complex for zk rollups

evan-forbes:
I think this is potentially becoming a possibility with authored blobs, but had a few questions

@cmwaters do you know if this is partially or fully unblocked by authored blobs? Do you considering removing the namespace for PFBs part of authored blobs? Do you know if the index wrapper will get in the way of authz here?

@Bidon15 @cmwaters  do you know any specific or could you elaborate on usecases for authz PFBs?

evan-forbes:
this could potentially be included v3 @Bidon15 , do we know of any usecases that could benefit from this?

Bidon15:
So some sample use-cases can be the following: 

1. I want to submit blobs through typical DevX without the keyring backend hosted in my app env. 
2. Any 3rd party services can provide REST blob submissions and keep user's celestia address as the tx sender.

PR #1775, opened 2023-05-16
fix: enable authzs MsgExec to be used in a blobTx

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1773

### DIFF
--- a/app/test/process_proposal_test.go
+++ b/app/test/process_proposal_test.go
@@ -28,10 +28,12 @@ import (
 
 func TestProcessProposal(t *testing.T) {
 	encConf := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-	accounts := testfactory.GenerateAccounts(6)
+	accounts := testfactory.GenerateAccounts(7)
 	testApp, kr := testutil.SetupTestAppWithGenesisValSet(accounts...)
 	infos := queryAccountInfo(testApp, accounts, kr)
 	signer := types.GenerateKeyringSigner(t, accounts[0])
+	signerAddr, err := signer.GetSignerInfo().GetAddress()
+	require.NoError(t, err)
 
 	// create 3 single blob blobTxs that are signed with valid account numbers
 	// and sequences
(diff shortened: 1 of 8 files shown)


app/test/process_proposal_test.go
```diff
@@ -299,10 +312,26 @@ func TestProcessProposal(t *testing.T) {
...
+			},
+			expectedResult: abci.ResponseProcessProposal_ACCEPT,
+		},
 	}
 
-	for _, tt := range tests {
+	for idx, tt := range tests {
+		if idx != 15 {
+			continue
+		}
```
rootulp, line 330: [blocking] I think this is residual from development

```suggestion
```

x/blob/types/blob_tx_test.go
```diff
@@ -117,3 +124,67 @@ func randMsgPayForBlobsWithNamespaceAndSigner(t *testing.T, signer string, ns ap
...
+			msgs:   []sdk.Msg{&invalidMsgExec},
+			blobs:  []*tmproto.Blob{blob},
+			expErr: true,
+		},
+		{
+			name:   "blob tx using invalid msg exec that has multiple messages",
+			msgs:   []sdk.Msg{&msgExecWithTwoMsgs},
+			blobs:  []*tmproto.Blob{blob},
+			expErr: true,
+		},
```
rootulp, line 168: 👍

rootulp (changes requested):
LGTM after one comment

rootulp (approved):

evan-forbes:
per synchronous discussions, I don't think we're moving forward with including authz blob support in the v1.0.0 release candidate, so should we still merge this?

cmwaters:
> per synchronous discussions, I don't think we're moving forward with including authz blob support in the v1.0.0 release candidate, so should we still merge this?

Nope, If you don't mind I'm going to put this back into draft mode and we can reassess it after cutting v1.0.0

cmwaters:
Closing as stale

Issue #1439, opened 2023-03-02
Consider removing split-test-files


rootulp:
## Context

Our current CI test workflow splits test files to run them in parallel. This parallel [example](https://github.com/celestiaorg/celestia-app/actions/runs/4197256126) took 520 seconds to complete test-race. This sequential [example](https://github.com/rootulp/celestia-app/actions/runs/4198117246) took 543 seconds to complete test-race. The sequential test (so not test-race) actually runs faster than the parallel version.

- https://github.com/celestiaorg/celestia-app/blob/6d27b78aa64a749a808e84ea682352b8b551fbd7/.github/workflows/test.yml#L60-L84

## Proposal
Remove the step to split test files from the CI workflow.

Pros:
- simpler to maintain
- `make test` is more representative of how tests are run in CI

Cons:
- in some scenarios the parallel version may be slightly faster

Chirag018:
Hey, @rootulp would like to work on this

rootulp:
@Chirag018 maybe it would help to look at what our current CI 

## Current

https://github.com/celestiaorg/celestia-app/actions/runs/4821189737

<img width="1278" alt="Screenshot 2023-04-27 at 11 23 35 AM" src="https://user-images.githubusercontent.com/3699047/234909843-07dd4e1a-ac6c-4393-9896-b766c8ec2fee.png">

## Proposed

- remove `split-test-files`
- remove the `matrix` from `test/tests`
- remove the `matrix` from `test/test-race`

Chirag018:
Since we are removing the matrix, so even the matrix. part in the following cases should be removed?

rootulp:
We shouldn't need a path here: https://github.com/celestiaorg/celestia-app/blob/4c738244bd6b1429419c3ad9ddb4e780a62e9ab7/.github/workflows/test.yml#L114 and the test.yml can probably be simplified greatly.

You may draw inspiration from https://github.com/celestiaorg/celestia-node/blob/11fea99767f549bebd1355b4c9065115af25ea91/.github/workflows/go-ci.yml#L64-L77

Chirag018:
Hey, @rootulp using the reference below, which I figured out, is it fine?
<img width="1267" alt="Screenshot 2023-04-27 at 11 46 23 PM" src="https://user-images.githubusercontent.com/56735482/234955783-a01a2516-6033-4957-a116-ea689ac5603a.png">

rootulp:
[...]

Issue #1796, opened 2023-05-19
Consistent workflow names


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/5615205760ac3921506e2c457fc2deb924f8aed5/.github/workflows/stale.yml#L2

## Problem

There isn't consistency with respect:
1. The name of the Github workflow file in `.github/workflows/`
2. The name of the workflow which is displayed in 
<img width="314" alt="Screenshot 2023-05-19 at 4 29 48 PM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/1af222af-b8ea-4376-ab8b-b2a9b2681fae">

## Proposal

Rename workflows and filenames so that they match.

PR #1798, opened 2023-05-19
ci: simplify test workflow

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1439

## Description

1. Remove commented code
1. Remove split test files
1. Remove tparse
1. Remove race detector report
1. Remove concurrency and cancelling jobs

## Testing

1. test works: https://github.com/rootulp/celestia-app/actions/runs/5028264091/jobs/9018769458
2. test-race works: https://github.com/rootulp/celestia-app/actions/runs/5028264091/jobs/9018769731
3. test-coverage doesn't work but it looks like that's because https://app.codecov.io/ isn't set up for my fork so testing here. 

## FLUPs

If this is merged, can @MSevey please:
1. remove the required CI checks for the matrix tests (i.e. `(01)`)
2. add required CI checks for test, test-race, test-coverage


.github/workflows/test.yml
```diff
@@ -1,262 +1,48 @@
-name: Tests / Code Coverage
-#  Tests / Code Coverage workflow runs unit tests and uploads a code coverage
-#  report
-#  This workflow is run on pushes to main & every Pull Requests where a .go,
-#  .mod, .sum have been changed
+name: test
```
SweeXordious, line 1:
```suggestion
name: Tests / Code Coverage
```

Because, if I understand right, this workflow is still sending coverage report

scripts/test_cover.sh
SweeXordious, line 1: file could  be deleted?

evan-forbes (approved):
I'm good with trying this out!

if a single test fails, does this mean that we have to retry all the tests? I'm certainly down for hunting flakey CI tests, but I'm not sure if we will ever be able to get to 0.00%, and if we have to re-run all the tests just because github's CI can't hang with consistently running a node or two in process that could get annoying

evan-forbes:
lol I guess we also have to update which tests are required as well?
Screenshot from 2023-05-21 10-40-37

scripts/test_cover.sh
rootulp: this file is used by `make test-coverage`

.github/workflows/test.yml
rootulp: it is sending a coverage report but the rename is motivated by https://github.com/celestiaorg/celestia-app/issues/1796

the workflow name doesn't need to include the words "code coverage" because the job name will include it. For example:

Screenshot 2023-05-22 at 8 40 46 AM

rootulp:
> if a single test fails, does this mean that we have to retry all the tests?

Yes we'll have to retry whichever job failed which will retry all the tests. If it gets annoying we can revert this PR.

> lol I guess we also have to update which tests are required as well?

Good point. We're going to need to change the required rules before this even merges.

scripts/test_cover.sh
SweeXordious: Aaah But its empty no? How is it used?

rootulp: I don't think GH will display the file contents if just the file permissions are modified. So "Empty file." is misleading.

It does have contents: https://github.com/celestiaorg/celestia-app/blob/eca8307e7edb9541bc2630373d7bb0f67ef876c8/scripts/test_cover.sh

MSevey (approved):
lgtm, let's coordinate on slack.

We can do one of the 2:
1. update the require CI then merge 
1. Admin merge, then update the required CI

Issue #309, opened 2022-04-14
Prune the QGB state after the unbonding period ends


SweeXordious:
For the QGB, we only care about the state of the current unbonding period. That is due to the following:
- If no signatures are posted during this period, then the orchestrator will be slashed (upcoming ADR).
- After the unbonding period ends, even if the orchestrator doesn't sign, there is nothing to be done about them.
Thus, we can get rid of the old state and only keep the one related to the current unbonding period.

evan-forbes:
what state is left to prune after ADR005?

SweeXordious:
We have the attestations requests that can be pruned. However, I guess we decided to dump them into block space. So, we will need to write an ADR about that

evan-forbes:
oh shoot, did we? that does sound familiar but I at least didn't write it down anywhere :sweat_smile:. I actually think its ok to keep those, as that will make it quite a bit easier to do slashing, and its really not that much state. 

We can still prune them though, even if its not that much state, it will just continue to grow forever for no reason.

SweeXordious:
Then, we will not be able to sync QGB from genesis.
Okey, makes sense. Thanks a lot

evan-forbes:
> Then, we will not be able to sync QGB from genesis.

ahh yeah! that was it. 

they could using an archive node, but also from another evm chain that has already synced I think

evan-forbes:
related to #683

SweeXordious:
Per sync, we need to prune every unbonding period

SweeXordious:
Final decision: Implement time based pruning

PR #1799, opened 2023-05-20
feat!: implement QGB pruning

SweeXordious:
Closes https://github.com/celestiaorg/celestia-app/issues/309 and https://github.com/celestiaorg/celestia-app/issues/1808

### DIFF
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -491,7 +491,7 @@ func NewTestMsgUnDelegateValidator(address sdk.ValAddress, amt cosmosmath.Int) *
 
 // ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
 // along with the QGB abci.EndBlocker on each one of them.
-// Returns the updated context with the correct height.
+// Returns the updated context with block height advanced to endHeight.
 func ExecuteQGBHeights(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64) sdk.Context {
 	for i := beginHeight; i < endHeight; i++ {
 		ctx = ctx.WithBlockHeight(i)
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -503,12 +503,12 @@ func ExecuteQGBHeights(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int
 // ExecuteQGBHeightsWithTime executes the end exclusive range of heights specified by beginHeight and endHeight
 // along with the QGB abci.EndBlocker on each one of them.
 // Uses the interval to calculate the block header time.
-func ExecuteQGBHeightsWithTime(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64, interval time.Duration) sdk.Context {
+func ExecuteQGBHeightsWithTime(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64, blockInterval time.Duration) sdk.Context {
 	blockTime := ctx.BlockTime()
 	for i := beginHeight; i < endHeight; i++ {
 		ctx = ctx.WithBlockHeight(i).WithBlockTime(blockTime)
 		qgb.EndBlocker(ctx, qgbKeeper)
-		blockTime = blockTime.Add(interval)
+		blockTime = blockTime.Add(blockInterval)
 	}
 	return ctx
 }
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -491,8 +491,7 @@ func NewTestMsgUnDelegateValidator(address sdk.ValAddress, amt cosmosmath.Int) *
 
 // ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
 // along with the QGB abci.EndBlocker on each one of them.
-// Returns the updated context with the correct height. Make sure to use it as the context in argument
-// will not have the last height as the one returned.
+// Returns the updated context with the correct height.
 func ExecuteQGBHeights(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64) sdk.Context {
 	for i := beginHeight; i < endHeight; i++ {
 		ctx = ctx.WithBlockHeight(i)
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -5,6 +5,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/celestiaorg/celestia-app/x/qgb"
+
 	cosmosmath "cosmossdk.io/math"
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/x/qgb/keeper"
@@ -319,6 +321,7 @@ func CreateTestEnvWithoutAttestationNonceInit(t *testing.T) TestInput {
 func CreateTestEnv(t *testing.T) TestInput {
 	input := CreateTestEnvWithoutAttestationNonceInit(t)
 	input.QgbKeeper.SetLatestAttestationNonce(input.Context, 0)
+	input.QgbKeeper.SetEarliestAvailableAttestationNonce(input.Context, 1)
 	return input
 }
 
@@ -484,4 +487,29 @@ func SetupTestChain(t *testing.T, weights []uint64) (TestInput, sdk.Context) {
 func NewTestMsgUnDelegateValidator(address sdk.ValAddress, amt cosmosmath.Int) *stakingtypes.MsgUndelegate {
 	msg := stakingtypes.NewMsgUndelegate(sdk.AccAddress(address), address, sdk.NewCoin("stake", amt))
 	return msg
+}
+
+// ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
+// along with the QGB abci.EndBlocker on each one of them.
+// Returns the updated context with the correct height. Make sure to use it as the context in argument
+// will not have the last height as the one returned.
+func ExecuteQGBHeights(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64) sdk.Context {
+	for i := beginHeight; i < endHeight; i++ {
+		ctx = ctx.WithBlockHeight(i)
+		qgb.EndBlocker(ctx, qgbKeeper)
+	}
+	return ctx
+}
+
+// ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
+// along with the QGB abci.EndBlocker on each one of them.
+// Uses the interval to calculate the block header time.
+func ExecuteQGBHeightsWithTime(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64, interval time.Duration) sdk.Context {
+	blockTime := ctx.BlockTime()
+	for i := beginHeight; i < endHeight; i++ {
+		ctx = ctx.WithBlockHeight(i).WithBlockTime(blockTime)
+		qgb.EndBlocker(ctx, qgbKeeper)
+		blockTime = blockTime.Add(interval)
+	}
+	return ctx
 }
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -500,7 +500,7 @@ func ExecuteQGBHeights(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int
 	return ctx
 }
 
-// ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
+// ExecuteQGBHeightsWithTime executes the end exclusive range of heights specified by beginHeight and endHeight
 // along with the QGB abci.EndBlocker on each one of them.
 // Uses the interval to calculate the block header time.
 func ExecuteQGBHeightsWithTime(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64, interval time.Duration) sdk.Context {
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -146,7 +146,7 @@ func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 	}
 	currentBlockTime := ctx.BlockTime()
 	// if the current time is before the earliest attestation creation time + expiry time
-	// then, all the subsequent attestations are also still valid and no need to prune them.
+	// then, all the subsequent attestations are unexpired so they should not be pruned.
 	if currentBlockTime.Before(earliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
 		return
 	}
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -126,7 +126,7 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 // maybePruneAttestations runs basic checks on saved attestations to see if we need to prune or not.
 // Then, it prunes all expired attestations from state.
 func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
-	// If the attestations nonce hasn't been initialized yet, no pruning is
+	// If the attestation nonce hasn't been initialized yet, no pruning is
 	// required
 	if !k.CheckLatestAttestationNonce(ctx) {
 		return
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -179,7 +179,7 @@ func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 	// persist the new earliest available attestation nonce
 	k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
 	ctx.Logger().Debug(
-		"finished pruning attestations from qgb store",
+		"finished pruning attestations from QGB store",
 		"count",
 		count,
 		"new_earliest_available_nonce",
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -17,7 +17,7 @@ const (
 	SignificantPowerDifferenceThreshold = 0.05
 
 	// AttestationExpiryTime the expiration time of an attestation after which it will be pruned.
-	AttestationExpiryTime = 8 * 7 * 24 * time.Hour
+	AttestationExpiryTime = 3 * 7 * 24 * time.Hour
 )
 
 // EndBlocker is called at the end of every block.
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -146,7 +146,8 @@ func pruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 	}
 	currentBlockTime := ctx.BlockTime()
 	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
-	for newEarliestAvailableNonce := earliestAttestation.GetNonce(); newEarliestAvailableNonce < latestAttestationNonce; newEarliestAvailableNonce++ {
+	var newEarliestAvailableNonce uint64
+	for newEarliestAvailableNonce = earliestAttestation.GetNonce(); newEarliestAvailableNonce < latestAttestationNonce; newEarliestAvailableNonce++ {
 		newEarliestAttestation, found, err := k.GetAttestationByNonce(ctx, newEarliestAvailableNonce)
 		if err != nil {
 			ctx.Logger().Error("error getting attestation for pruning", "nonce", newEarliestAvailableNonce, "err", err.Error())
@@ -160,21 +161,25 @@ func pruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 			ctx.Logger().Error("nil attestation for pruning", "nonce", newEarliestAvailableNonce)
 			return
 		}
-		if currentBlockTime.Before(newEarliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
+		attestationExpirationTime := newEarliestAttestation.BlockTime().Add(AttestationExpiryTime)
+		if attestationExpirationTime.After(currentBlockTime) {
 			// the current attestation is unexpired so subsequent ones are also unexpired
 			// persist the new earliest available attestation nonce
-			k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
-			ctx.Logger().Debug(
-				"pruned attestations from QGB store",
-				"count",
-				newEarliestAvailableNonce-earliestAttestation.GetNonce(),
-				"new_earliest_available_nonce",
-				newEarliestAvailableNonce,
-				"latest_attestation_nonce",
-				latestAttestationNonce,
-			)
-			return
+			break
 		}
 		k.DeleteAttestation(ctx, newEarliestAvailableNonce)
+	}
+	if newEarliestAvailableNonce > earliestAttestation.GetNonce() {
+		// some attestations were pruned and we need to update the state for it
+		k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
+		ctx.Logger().Debug(
+			"pruned attestations from QGB store",
+			"count",
+			newEarliestAvailableNonce-earliestAttestation.GetNonce(),
+			"new_earliest_available_nonce",
+			newEarliestAvailableNonce,
+			"latest_attestation_nonce",
+			latestAttestationNonce,
+		)
 	}
 }
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -124,7 +124,7 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 }
 
 // maybePruneAttestations runs basic checks on saved attestations to see if we need to prune or not.
-// Then, it starts pruning the expired attestations until all attestations in state are still valid.
+// Then, it prunes all expired attestations from state.
 func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 	// If the attestations nonce hasn't been initialized yet, no pruning is
 	// required
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -26,7 +26,7 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
 	// the one responsible for signing from now on.
 	handleValsetRequest(ctx, k)
 	handleDataCommitmentRequest(ctx, k)
-	maybePruneAttestations(ctx, k)
+	pruneAttestations(ctx, k)
 }
 
 func handleDataCommitmentRequest(ctx sdk.Context, k keeper.Keeper) {
@@ -123,9 +123,9 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 	}
 }
 
-// maybePruneAttestations runs basic checks on saved attestations to see if we need to prune or not.
+// pruneAttestations runs basic checks on saved attestations to see if we need to prune or not.
 // Then, it prunes all expired attestations from state.
-func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
+func pruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 	// If the attestation nonce hasn't been initialized yet, no pruning is
 	// required
 	if !k.CheckLatestAttestationNonce(ctx) {
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -2,16 +2,23 @@ package qgb
 
 import (
 	"errors"
+	"time"
 
 	sdkerrors "cosmossdk.io/errors"
+
 	"github.com/celestiaorg/celestia-app/x/qgb/keeper"
 	"github.com/celestiaorg/celestia-app/x/qgb/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
-// that would need the creation of a new valset request.
-const SignificantPowerDifferenceThreshold = 0.05
+const (
+	// SignificantPowerDifferenceThreshold is the threshold of change in the validator set power
+	// that would trigger the creation of a new valset request.
+	SignificantPowerDifferenceThreshold = 0.05
+
+	// AttestationExpiryTime the expiration time of an attestation after which it will be pruned.
+	AttestationExpiryTime = 8 * 7 * 24 * time.Hour
+)
 
 // EndBlocker is called at the end of every block.
 func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
@@ -19,6 +26,7 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
 	// the one responsible for signing from now on.
 	handleValsetRequest(ctx, k)
 	handleDataCommitmentRequest(ctx, k)
+	maybePruneAttestations(ctx, k)
 }
 
 func handleDataCommitmentRequest(ctx sdk.Context, k keeper.Keeper) {
@@ -113,4 +121,70 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 			panic(err)
 		}
 	}
+}
+
+// maybePruneAttestations runs basic checks on saved attestations to see if we need to prune or not.
+// Then, it starts pruning the expired attestations until all attestations in state are still valid.
+func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
+	// If the attestations nonce hasn't been initialized yet, no pruning is
+	// required
+	if !k.CheckLatestAttestationNonce(ctx) {
+		return
+	}
+	earliestAttestation, found, err := k.GetAttestationByNonce(ctx, k.GetEarliestAvailableAttestationNonce(ctx))
+	if err != nil {
+		ctx.Logger().Error("error getting earliest attestation for pruning", "err", err.Error())
+		return
+	}
+	if !found {
+		ctx.Logger().Error("couldn't find earliest attestation for pruning")
+		return
+	}
+	if earliestAttestation == nil {
+		ctx.Logger().Error("nil earliest attestation")
+		return
+	}
+	currentBlockTime := ctx.BlockTime()
+	// if the current time is before the earliest attestation creation time + expiry time
+	// then, all the subsequent attestations are also still valid and no need to prune them.
+	if currentBlockTime.Before(earliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
+		return
+	}
+
+	ctx.Logger().Debug("pruning attestations from qgb store")
+	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
+	count := 0
+	var newEarliestAvailableNonce uint64
+	for newEarliestAvailableNonce = earliestAttestation.GetNonce(); newEarliestAvailableNonce < latestAttestationNonce; newEarliestAvailableNonce++ {
+		newEarliestAttestation, found, err := k.GetAttestationByNonce(ctx, newEarliestAvailableNonce)
+		if err != nil {
+			ctx.Logger().Error("error getting attestation for pruning", "nonce", newEarliestAvailableNonce, "err", err.Error())
+			return
+		}
+		if !found {
+			ctx.Logger().Error("couldn't find attestation for pruning", "nonce", newEarliestAvailableNonce)
+			return
+		}
+		if newEarliestAttestation == nil {
+			ctx.Logger().Error("nil attestation for pruning", "nonce", newEarliestAvailableNonce)
+			return
+		}
+		if currentBlockTime.Before(newEarliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
+			// the earliest attestation is still valid. this means all the subsequent ones are also
+			break
+		}
+		k.DeleteAttestation(ctx, newEarliestAvailableNonce)
+		count++
+	}
+	// persist the new earliest available attestation nonce
+	k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
+	ctx.Logger().Debug(
+		"finished pruning attestations from qgb store",
+		"count",
+		count,
+		"new_earliest_available_nonce",
+		newEarliestAvailableNonce,
+		"latest_attestation_nonce",
+		latestAttestationNonce,
+	)
 }
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -169,7 +169,7 @@ func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 			return
 		}
 		if currentBlockTime.Before(newEarliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
-			// the earliest attestation is still valid. this means all the subsequent ones are also
+			// the current attestation is unexpired so subsequent ones are also unexpired
 			break
 		}
 		k.DeleteAttestation(ctx, newEarliestAvailableNonce)
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -153,7 +153,6 @@ func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 
 	ctx.Logger().Debug("pruning attestations from QGB store")
 	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
-	count := 0
 	var newEarliestAvailableNonce uint64
 	for newEarliestAvailableNonce = earliestAttestation.GetNonce(); newEarliestAvailableNonce < latestAttestationNonce; newEarliestAvailableNonce++ {
 		newEarliestAttestation, found, err := k.GetAttestationByNonce(ctx, newEarliestAvailableNonce)
@@ -174,14 +173,13 @@ func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 			break
 		}
 		k.DeleteAttestation(ctx, newEarliestAvailableNonce)
-		count++
 	}
 	// persist the new earliest available attestation nonce
 	k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
 	ctx.Logger().Debug(
 		"finished pruning attestations from QGB store",
 		"count",
-		count,
+		newEarliestAvailableNonce-earliestAttestation.GetNonce(),
 		"new_earliest_available_nonce",
 		newEarliestAvailableNonce,
 		"latest_attestation_nonce",
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -145,16 +145,8 @@ func pruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 		return
 	}
 	currentBlockTime := ctx.BlockTime()
-	// if the current time is before the earliest attestation creation time + expiry time
-	// then, all the subsequent attestations are unexpired so they should not be pruned.
-	if currentBlockTime.Before(earliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
-		return
-	}
-
-	ctx.Logger().Debug("pruning attestations from QGB store")
 	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
-	var newEarliestAvailableNonce uint64
-	for newEarliestAvailableNonce = earliestAttestation.GetNonce(); newEarliestAvailableNonce < latestAttestationNonce; newEarliestAvailableNonce++ {
+	for newEarliestAvailableNonce := earliestAttestation.GetNonce(); newEarliestAvailableNonce < latestAttestationNonce; newEarliestAvailableNonce++ {
 		newEarliestAttestation, found, err := k.GetAttestationByNonce(ctx, newEarliestAvailableNonce)
 		if err != nil {
 			ctx.Logger().Error("error getting attestation for pruning", "nonce", newEarliestAvailableNonce, "err", err.Error())
@@ -170,19 +162,19 @@ func pruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 		}
 		if currentBlockTime.Before(newEarliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
 			// the current attestation is unexpired so subsequent ones are also unexpired
-			break
+			// persist the new earliest available attestation nonce
+			k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
+			ctx.Logger().Debug(
+				"pruned attestations from QGB store",
+				"count",
+				newEarliestAvailableNonce-earliestAttestation.GetNonce(),
+				"new_earliest_available_nonce",
+				newEarliestAvailableNonce,
+				"latest_attestation_nonce",
+				latestAttestationNonce,
+			)
+			return
 		}
 		k.DeleteAttestation(ctx, newEarliestAvailableNonce)
 	}
-	// persist the new earliest available attestation nonce
-	k.SetEarliestAvailableAttestationNonce(ctx, newEarliestAvailableNonce)
-	ctx.Logger().Debug(
-		"finished pruning attestations from QGB store",
-		"count",
-		newEarliestAvailableNonce-earliestAttestation.GetNonce(),
-		"new_earliest_available_nonce",
-		newEarliestAvailableNonce,
-		"latest_attestation_nonce",
-		latestAttestationNonce,
-	)
 }
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -151,7 +151,7 @@ func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
 		return
 	}
 
-	ctx.Logger().Debug("pruning attestations from qgb store")
+	ctx.Logger().Debug("pruning attestations from QGB store")
 	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
 	count := 0
 	var newEarliestAvailableNonce uint64
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -532,11 +532,11 @@ func TestPruning(t *testing.T) {
 	window := uint64(101)
 	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
 	initialBlockTime := ctx.BlockTime()
-	// make the interval between blocks being one hour
-	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 1345, time.Hour)
+	// make the interval between blocks being 10 minutes
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 1626, 10*time.Minute)
 
 	// check that we created a number of attestations
-	assert.Equal(t, uint64(14), qgbKeeper.GetLatestAttestationNonce(ctx))
+	assert.Equal(t, uint64(17), qgbKeeper.GetLatestAttestationNonce(ctx))
 
 	// check that no pruning occurs if the no attestation is expired
 	for nonce := uint64(1); nonce <= qgbKeeper.GetLatestAttestationNonce(ctx); nonce++ {
@@ -546,10 +546,10 @@ func TestPruning(t *testing.T) {
 	}
 
 	// continue executing heights
-	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1345, 5000, time.Hour)
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1626, 5000, 10*time.Minute)
 
 	earliestAttestationNonce := qgbKeeper.GetEarliestAvailableAttestationNonce(ctx)
-	assert.Equal(t, uint64(38), earliestAttestationNonce)
+	assert.Equal(t, uint64(21), earliestAttestationNonce)
 
 	// check that the first attestations were pruned
 	for nonce := uint64(1); nonce < earliestAttestationNonce; nonce++ {
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -532,7 +532,7 @@ func TestPruning(t *testing.T) {
 	window := uint64(101)
 	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
 	initialBlockTime := ctx.BlockTime()
-	// make the interval between blocks being 3 days
+	// make the interval between blocks being one hour
 	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 5000, time.Hour)
 
 	earliestAttestationNonce := qgbKeeper.GetEarliestAvailableAttestationNonce(ctx)
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -1,9 +1,10 @@
 package qgb_test
 
 import (
-	"github.com/celestiaorg/celestia-app/x/qgb/types"
 	"testing"
 	"time"
+
+	"github.com/celestiaorg/celestia-app/x/qgb/types"
 
 	"github.com/cosmos/cosmos-sdk/x/staking/teststaking"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -521,8 +521,10 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
 	}
 }
 
-// TestPruning tests the pruning mechanism via generating a set of attestations, then checking if the expired
-// ones are pruned
+// TestPruning tests the pruning mechanism by:
+// 1. Generating a set of attestations
+// 2. Running the QGB EndBlocker
+// 3. Verifying that the expired attestations are pruned
 func TestPruning(t *testing.T) {
 	input, ctx := testutil.SetupFiveValChain(t)
 	qgbKeeper := *input.QgbKeeper
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -533,7 +533,20 @@ func TestPruning(t *testing.T) {
 	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
 	initialBlockTime := ctx.BlockTime()
 	// make the interval between blocks being one hour
-	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 5000, time.Hour)
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 1345, time.Hour)
+
+	// check that we created a number of attestations
+	assert.Equal(t, uint64(14), qgbKeeper.GetLatestAttestationNonce(ctx))
+
+	// check that no pruning occurs if the no attestation is expired
+	for nonce := uint64(1); nonce <= qgbKeeper.GetLatestAttestationNonce(ctx); nonce++ {
+		_, found, err := qgbKeeper.GetAttestationByNonce(ctx, nonce)
+		assert.NoError(t, err)
+		assert.True(t, found)
+	}
+
+	// continue executing heights
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1345, 5000, time.Hour)
 
 	earliestAttestationNonce := qgbKeeper.GetEarliestAvailableAttestationNonce(ctx)
 	assert.Equal(t, uint64(38), earliestAttestationNonce)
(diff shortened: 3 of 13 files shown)


x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	case types.ValsetRequestType:
+		vs, ok := lastAttestationInStore.(*types.Valset)
+		if !ok {
+			panic("couldn't cast valset type")
+		}
+		lastNonceHeight = vs.Height
+	}
+	lastUnbondingHeight := k.GetLastUnBondingBlockHeight(ctx)
+	if lastNonceHeight == lastUnbondingHeight {
+		// we don't need to do anything, we wont to keep attestations up to the unbonding height
```
SweeXordious, line 159:
```suggestion
		// we don't need to do anything, we want to keep attestations up to the last unbonding height
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	}
+	if lastNonceHeight > lastUnbondingHeight {
+		// checking if it's the initial case following the startup of the chain
+		if lastNonceHeight == 1 && lastUnbondingHeight == 0 {
+			return
+		}
+		// we should never hit this case, since we will keep the attestations up to the last unbonding height
+		panic("missing attestations up to the unbonding height")
+	}
+	// now we have attestations before the unbonding height, we should check wether we need to prune them
```
SweeXordious, line 170:
```suggestion
	// now we have attestations before the unbonding height, we should check whether we need to prune them
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	// or not yet
+	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
+	if latestAttestationNonce-lastAvailableNonce+1 <= AttestationPruningThreshold {
+		// we don't need to prune as we still have room to store more attestations
+		return
+	}
+	// now we want to prune attestations as we have the following conditions:
+	// - we have attestations up to the last unbonding period
```
SweeXordious, line 178:
```suggestion
	// - we have attestations up to the last unbonding height
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	// - the total number of attestations, including the ones before the unbonding period, are greater
```
SweeXordious, line 179:
```suggestion
	// - the total number of attestations, including the ones before the unbonding height, are greater
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	// than the AttestationPruningThreshold
+	err = pruneAttestations(ctx, k, lastAvailableNonce, latestAttestationNonce)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// pruneAttestations prunes attestations to keep attestations up to the last unbonding period
```
SweeXordious, line 187:
```suggestion
// pruneAttestations prunes attestations to keep attestations up to the last unbonding height
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+// and the total number of attestations lesser than AttestationPruningThreshold.
```
SweeXordious, line 188:
```suggestion
// and the total number of attestations lower than the AttestationPruningThreshold.
```

x/qgb/pruning_test.go
```diff
@@ -0,0 +1,222 @@
...
+	for nonce := uint64(1); nonce <= totalNumberOfAttestations; nonce++ {
+		_, found, err := qgbKeeper.GetAttestationByNonce(ctx, nonce)
+		assert.NoError(t, err)
+		assert.True(t, found)
+	}
+}
+
+// TestPruneMultiple tests the case where the (LatestNonce-LastUnbondingNonce) > AttestationPruningThreshold
+// But still will not prune because we want to keep attestations up to the last unbonding period.
+// Then, sets the last unbonding height to a higher nonce to trigger multiple prunining rounds.
```
SweeXordious, line 101:
```suggestion
// Then, sets the last unbonding height to a higher nonce to trigger multiple pruning rounds.
```

x/qgb/pruning_test.go
```diff
@@ -0,0 +1,222 @@
...
+func TestPruneMultiple(t *testing.T) {
+	input, ctx := testutil.SetupFiveValChain(t)
+	qgbKeeper := *input.QgbKeeper
+	// set the data commitment window
+	window := uint64(101)
+	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
+
+	// test that no prunning occurs if last unbonding height is still 0 (chain startup scenario)
```
SweeXordious, line 109:
```suggestion
	// test that no pruning occurs if last unbonding height is still 0 (chain startup scenario)
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		// However, we might hit it in an upgrade if we don't handle the store values carefully.
+		// So, it's good to have this check in place.
+		return fmt.Errorf("last unbonding nonce not initialized in store")
+	}
+	lastUnbondingNonce := k.GetLastUnbondingNonce(ctx)
+	count := 0
+	newLastAvailableNonce := lastAvailableAttestationNonce
+	for i := lastAvailableAttestationNonce; i < lastUnbondingNonce; i++ {
+		if newLastAvailableNonce == lastUnbondingNonce ||
+			latestAttestationNonce-i+1 == AttestationPruningThreshold {
```
evan-forbes, line 208: pls comment about why we add one here

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+			// we either reached the last unbonding height, and we need to keep all those attestations
+			// or, we reached the minimum number of attestations we want to keep in store
+			newLastAvailableNonce = i
+			break
+		}
+		k.DeleteAttestation(ctx, i)
+		count++
+	}
+	// persist the new last pruned attestation nonce
+	k.SetLastPrunedAttestationNonce(ctx, newLastAvailableNonce-1)
```
evan-forbes, line 218: why subtract one here if we add it back in line 134? (see other comment there as well)

evan-forbes (commented):
thanks for getting this in @SweeXordious!

I'm still going through the tests, but figured I'd submit this first round of feedback now.

overall I think I understand everything, although there are a quite a few if statements and general logic that its difficult to see if all of the logic is being hit somewhere in the unit tests. I would also deeply appreciate an integration test that runs a full blown chain that produces blocks as fast as possible, constantly changes the validator set, and sets the DCWindow to as low as possible. (this should be quite doable using test/util/testnode)

in general, I feel like most of what is left is some simple code clean up and maybe additional testing. becuase there is so much logic here, I think it might make a lot of sense to break it up into unit testable functions.

Also, is there anything we can do to simplify this logic? would it be easier to just always prune after huge number of blocks? for instance, there are ~120960 blocks in the unbonding period, what if we just always prune an attestation if the height is less than (the current height - 2 *blocks in unbonding period) The features that we originally designed around seemed like a good idea at the time, but if it requires this very complex logic, then I'm not sure its worth it

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	lastAvailableNonce := k.GetLastPrunedAttestationNonce(ctx) + 1
+	lastAttestationInStore, found, err := k.GetAttestationByNonce(ctx, lastAvailableNonce)
+	if err != nil {
+		panic(err)
+	}
+	if !found {
+		panic(fmt.Sprintf("couldn't find attestation %s for pruning", lastAttestationInStore))
+	}
+	var lastNonceHeight uint64
+	switch lastAttestationInStore.Type() {
```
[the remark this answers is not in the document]
SweeXordious: I had the same question when implementing this, but didn't wanna investigate it as it's not the purpose for this. I can't remember why we have that method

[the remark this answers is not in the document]
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1804

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,66 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+	store := ctx.KVStore(k.storeKey)
+	has := store.Has([]byte(types.LastUnbondingNonce))
+	return has
+}
+
+// GetLastUnbondingNonce returns the last unbonding nonce.
+// Panics if the last unbonding nonce doesn't exit. Make sure to call `CheckLastUnbondingNonce`
+// before getting the nonce.
+// This value is set on chain startup. However, it won't be written to store until height = 1.
+// Thus, it's mandatory to run `CheckLastUnbondingNonce` before calling this method.
```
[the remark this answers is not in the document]
SweeXordious: Good idea, I am planning to change that throughout the whole codebase, not sure why we had it in the first place. But, I didn't wanna incorporate that change in here not to add more complexity to the PR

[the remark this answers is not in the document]
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1805

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+
+func PruneIfNeeded(ctx sdk.Context, k keeper.Keeper) {
+	if !k.CheckLatestAttestationNonce(ctx) ||
+		k.CheckLatestAttestationNonce(ctx) && k.GetLatestAttestationNonce(ctx) <= AttestationPruningThreshold {
+		// If the attestations nonce hasn't been initialized yet, or the nonce is not greater than the minimum
+		// number of attestations, we don't need to prune anything
+		return
+	}
+
+	lastAvailableNonce := k.GetLastPrunedAttestationNonce(ctx) + 1
```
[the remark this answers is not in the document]
SweeXordious: I think since this is related to pruning, keeping the last pruned nonce makes more sense, no?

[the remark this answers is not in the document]
SweeXordious: I thought the name `lastAvailableNonce` is understandable :D this one is called only once in here, I think a comment is enough no?

```suggestion
	// the last older nonce available in store
	lastAvailableNonce := k.GetLastPrunedAttestationNonce(ctx) + 1
```

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		}
+		lastNonceHeight = dc.EndBlock
+	case types.ValsetRequestType:
+		vs, ok := lastAttestationInStore.(*types.Valset)
+		if !ok {
+			panic("couldn't cast valset type")
+		}
+		lastNonceHeight = vs.Height
+	}
+	lastUnbondingHeight := k.GetLastUnBondingBlockHeight(ctx)
```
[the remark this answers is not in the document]
SweeXordious: there are multiple unbonding heights in the history, this one refers to the last one

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: sorry, by "this section of code" I meant until about line 160. We're getting the height of the oldest attestation, not just a comment for that line

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: I think we inherited this portion from the gravity module, but are we using the validator's unbonding height as a proxy for the real unbonding height (the height from 3 weeks previous)? If so we might have to change that, as the important thing is that 3 weeks value, we don't want to prune after that

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: we don't have to have one, but it would have saved me a bit of time reading for the first time, so just mentioned. definitely not blocking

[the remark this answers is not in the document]
evan-forbes: imo, it makes the most sense to store values that directly useful. If we're always subtracting or adding ones to get the value we want, then I think that makes things more difficult to understand.

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		return
+	}
+	if lastNonceHeight > lastUnbondingHeight {
+		// checking if it's the initial case following the startup of the chain
+		if lastNonceHeight == 1 && lastUnbondingHeight == 0 {
+			return
+		}
+		// we should never hit this case, since we will keep the attestations up to the last unbonding height
+		panic("missing attestations up to the unbonding height")
+	}
```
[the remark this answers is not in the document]
SweeXordious: We should panic as something bad happened. By default, we should always have attestations up to the last unbonding height. If the last available nonce has a height that is higher than the unbonding height, it means that we pruned attestations we shouldn't, so it's better to panic to make it easier to debug what's wrong, and also not to have a state machine behaving differently without us knowing

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	// now we have attestations before the unbonding height, we should check whether we need to prune them
+	// or not yet
+	latestAttestationNonce := k.GetLatestAttestationNonce(ctx)
+	if latestAttestationNonce-lastAvailableNonce+1 <= AttestationPruningThreshold {
+		// we don't need to prune as we still have room to store more attestations
+		return
+	}
```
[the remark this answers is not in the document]
SweeXordious: The one we have at the top is different, since it checks the special case of the time before pruning anything.
The first one checks if the nonces are still under the threshold,  then there is no need to waste gas anymore because we  know we generally want to keep that.

This check here is after the nonces are beyond the threshold and we need to start pruning.

The first check is more like a gas optimization

x/qgb/abci.go
```diff
@@ -113,3 +122,108 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	// now we want to prune attestations as we have the following conditions:
+	// - we have attestations up to the last unbonding height
+	// - the total number of attestations, including the ones before the unbonding height, are greater
+	// than the AttestationPruningThreshold
```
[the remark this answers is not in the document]
SweeXordious: It's already there as a go doc. This one is for readers to know that the at this step, we have these conditions and we can call prune attestations

x/qgb/abci.go
SweeXordious: [2c19389](https://github.com/celestiaorg/celestia-app/pull/1799/commits/2c193899e54ed652ef2339907a9db93d4b6e718c)

x/qgb/abci.go
SweeXordious: This one is to set the value of the last pruned nonce which is: `last pruned nonce == last available nonce - 1`.

The other is to calculate the number of attestations that will be in store

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: on it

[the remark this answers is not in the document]
SweeXordious: I am open to suggestions, but I don't think `oldest` is any different since we will have to use it for other values as well to be consistent in naming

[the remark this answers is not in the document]
SweeXordious: I will extract it in a separate method to reduce the complexity, thanks for letting me know

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: here? https://github.com/SweeXordious/celestia-app/blob/2c193899e54ed652ef2339907a9db93d4b6e718c/x/qgb/abci.go#L126 ?

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: tbh, its difficult to see where in the tests this gets tested for, and seems like it could be hit. my first reaction was "how is this not getting hit" :sweat_smile: .  especially since it appears that the unbonding height is not what I thought it was https://github.com/celestiaorg/celestia-app/pull/1799/files#r1199820231

You're correct that it can make sense to panic! especially if it is not in a critical pathway and lets us not return an error, or if the consequences of that thing happening are of the utmost importance (for example chain invarients in the bank module) but in this specific scenario I don't think it does. We shouldn't be risking a chain halt to ensure that we're pruning state correctly, and the qgb will likely work even if it does. and if it doesn't, then at least the chain continues until the next hardfork that fixes it.

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: ahh I think I see thanks for explaining

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: > but I don't think oldest is any different since we will have to use it for other values as well to be consistent in naming

true, but it would help distinguish between last and latest, which is beneficial "last" is used to mean latest. such as "last time I went to the grocery store", I'm refering to the my latest visit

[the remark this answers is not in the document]
SweeXordious: [03874e2](https://github.com/celestiaorg/celestia-app/pull/1799/commits/03874e2d707c8cbc610079cc415995832ef9dc48)

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: > tbh, its difficult to see where in the tests this gets tested for, and seems like it could be hit

It's not tested because it would require manually changing the state. If you run `abci.EndBlocker`  on every block, that case should never be hit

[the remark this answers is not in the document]
SweeXordious: > and the qgb will likely work even if it does. and if it doesn't, then at least the chain continues until the next hardfork that fixes it.

That's a good point. Maybe an error log message?

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: Last and latest are used interchangeably in the code base, if I remember correctly. You have a point, we should have the same naming convention for all of them, and preferably decided by an English native person :D Personally, I didn't see a difference. 
What do you think we open an issue and change that? Changing the name wouldn't be state/consensus breaking, and only API breaking, so it's okey I guess and can be done after the feature freeze as part of the bug fixes and improvements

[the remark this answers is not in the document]
SweeXordious: [106a952](https://github.com/celestiaorg/celestia-app/pull/1799/commits/106a95261fb737782b435c206f57216da5726361)

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: [63624a7](https://github.com/celestiaorg/celestia-app/pull/1799/commits/63624a72168e7920ea576c9e2f699dc03fc05083)

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: Actually, that value is set in here: https://github.com/celestiaorg/celestia-app/blob/5443e5fca8fb02652dafbca8a64a62c8263bfe96/x/qgb/keeper/hooks.go#L32

Which changes on every validator that starts unbonding. So for example, if validator `A` starts unbonding on height 100, it will take that value. Then, if validator `B` starts unbonding at height 110, it will update its value to 110.

Not sure about the above scenario, if any validator can unbound anytime they want (I assume yes) or not

SweeXordious:
> I think it might make a lot of sense to break it up into unit testable functions.

That was where I wasted a bit of time. I was trying white box testing for the pruning logic, however, we have deep cyclic dependencies with the testing setup used to run the tests... So, it got harder to test in a white box manner and had to export the `PruneIfNeeded` method

SweeXordious:
> I will also write an integration test that runs a full blown chain that produces blocks as fast as possible, constantly changes the validator set, and sets the DCWindow to as low as possible to help with this effort. We might only be able to run it locally, but that's ok atm

I did that with the orchestrator-relayer E2E setup where I setup a chain of 4 validators and let it run for a couple of hours.  Also, I am planning on testing the whole thing using testground, but the priority now was to be in the deadline for feature freeze

SweeXordious:
> Also, is there anything we can do to simplify this logic? would it be easier to just always prune after huge number of blocks? for instance, there are ~120960 blocks in the unbonding period, what if we just always prune an attestation if the height is less than (the current height - 2 *blocks in unbonding period) The features that we originally designed around seemed like a good idea at the time, but if it requires this very complex logic, then I'm not sure its worth it

Good idea, but I think we already have that way implemented. Also, pruning that way might not work all the time if we have a stable network running for multiple days without any validator set change, any validator unbonding, any validator getting jailed, etc. Then, we would end up in a scenario where we don't have all the attestations needed to deploy the QGB

x/qgb/abci.go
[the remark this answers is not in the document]
evan-forbes: that works for me :slightly_smiling_face:

evan-forbes:
opened a PR with an attempt to simplify pruning https://github.com/SweeXordious/celestia-app/pull/39 targeting this branch

evan-forbes:
@SweeXordious, quick question, why are we wanting to preserve the last unbonding validator height here? doesn't that only update when a validator gets jailed or otherwise falls out of the validator set. Why can't we prune past that?

SweeXordious:
> Why can't we prune past that?

That was the first plan, then we decided to also keep a minimum number of attestations in state *just in case* we needed them

x/qgb/abci.go
```diff
@@ -113,3 +122,118 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		// all store values are initialized.
+		// However, we might hit it in an upgrade if we don't handle the store values carefully.
+		// So, it's good to have this check in place.
+		return fmt.Errorf("last unbonding nonce not initialized in store")
+	}
+	lastUnbondingNonce := k.GetLastUnbondingNonce(ctx)
+	count := 0
+	newLastAvailableNonce := lastAvailableAttestationNonce
+	for i := lastAvailableAttestationNonce; i < lastUnbondingNonce; i++ {
+		if newLastAvailableNonce == lastUnbondingNonce ||
```
cmwaters, line 214: I don't see how this clause ever hits. `lastUnbondingNonce` is always less than `i` and thus `newLastAvailableNonce`

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1808

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: [0f22136](https://github.com/celestiaorg/celestia-app/pull/1799/commits/0f2213671a6725c13abc7515faa4e36eaef653cd)

x/qgb/abci.go
```diff
@@ -113,3 +122,118 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		}
+		k.DeleteAttestation(ctx, i)
+		count++
+	}
+	// persist the new last available attestation nonce
+	k.SetLastAvailableAttestationNonce(ctx, newLastAvailableNonce)
+	ctx.Logger().Debug(
+		"finished pruning attestations from qgb store",
+		"count",
+		count,
```
cmwaters, line 232:
```suggestion
        // loop through all the attestations between the last available nonce and the last nonce within the unbonding period
	for newLastAvailableNonce := lastAvailableAttestationNonce; newLastAvailableNonce < lastUnbondingNonce;  {
	        // check how many attestations we have
		if latestAttestationNonce-newLastAvailableNonce+1 == AttestationPruningThreshold {
			// we have reached the minimum number of attestations we want to keep in store
			// we don't want to prune any more, so we break
			break
		}
		k.DeleteAttestation(ctx, newLastAvailableNonce)
		newLastAvailableNonce++
	}
	// persist the new last available attestation nonce
	k.SetLastAvailableAttestationNonce(ctx, newLastAvailableNonce)
	ctx.Logger().Debug(
		"finished pruning attestations from qgb store",
		"count",
		newLastAvailableNonce-lastAvailableAttestationNonce,
```
nit: I think this is a little more readable

cmwaters (commented):
Added some comments. Still not sure about this initialisation problem

cmwaters (commented):
Added some comments. Still not sure about this initialisation problem

cmwaters (commented):
Added some comments. Still not sure about this initialisation problem

x/qgb/abci.go
```diff
@@ -2,23 +2,32 @@ package qgb
...
-// that would need the creation of a new valset request.
-const SignificantPowerDifferenceThreshold = 0.05
+const (
+	// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
+	// that would need the creation of a new valset request.
+	SignificantPowerDifferenceThreshold = 0.05
+
+	// AttestationPruningThreshold the minimum number of recent attestations that will always be present
+	// in state.
+	AttestationPruningThreshold = 100
```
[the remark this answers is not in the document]
SweeXordious: That was the original plan. But we decided to also keep a certain number of attestations if the last unbonding height is so recent as a safety measure.

For example, if at block height 100, we have a new unbonding height, then, at height 101, we will delete all attestations before height 100, leaving only the last valset in state. 
If the orchestrators are running and up to date, then, that would be no issue. However, if they have some high latency, or... they might miss signing an attestation or two before they're pruned.

This won't be problematic in case it only happened for 1 or 2 orchestrators. However, if it happens for 1/3, that means the QGB will halt.

So, we decided to always keep a minimum number of attestations in state to account for this kinda failures on the orchestrators side.

Note: This won't be an issue if orchestrator is connected to an archive node. But we can't make that assumption for all orchestrators

x/qgb/genesis.go
```diff
@@ -10,6 +10,12 @@ import (
 // state.
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
 	k.SetLatestAttestationNonce(ctx, 0)
+	// The reason we're starting the last available nonce at 1 is because at chain startup,
+	// a new valset will be created all the time.
+	// Also, it's easier to set it here to 1 instead of doing it in abci.EndBlocker and do
+	// the check on every iteration
+	k.SetLastAvailableAttestationNonce(ctx, 1)
+	k.SetLastUnbondingNonce(ctx, 0)
```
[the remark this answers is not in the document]
SweeXordious: This is referring to the actual `Last` nonce, and not the earliest.
The earliest would be height 1, the last would be height 100. 
Yes, I agree we need a better naming scheme, but I don't think earliest describes it correctly. Also, I am not a native English speaker, feel free to correct me please :D

evan-forbes:
@SweeXordious but why is that value important? The unbonding period (3 weeks) I get, but I'm unsure about a random validator's unbonding period.

I had difficulty understanding what the gravity module was using this value for as well. It seems they're using it as a proxy for the unbonding period.

SweeXordious:
> Still not sure about this initialisation problem

It's not a problem. It's just that we can do it in endBlocker, but that check would be executed every time. It's more like a gas optimization

x/qgb/abci.go
SweeXordious: 21e373f, also found more simplifications when updating

x/qgb/abci.go
SweeXordious: [dfa4bbe](https://github.com/celestiaorg/celestia-app/pull/1799/commits/dfa4bbeabad638d6572728db36766a04959c7a41)

test/util/common.go
```diff
@@ -485,3 +489,15 @@ func NewTestMsgUnDelegateValidator(address sdk.ValAddress, amt cosmosmath.Int) *
 	msg := stakingtypes.NewMsgUndelegate(sdk.AccAddress(address), address, sdk.NewCoin("stake", amt))
 	return msg
 }
+
+// ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
+// along with the QGB abci.EndBlocker on each one of them.
+// Returns the updated context with the correct height. Make sure to use it as the context in argument
+// will not have the last height as the one returned.
```
rootulp, line 496: Sorry I don't understand the last line here.

> Make sure to use it as the context in argument
> // will not have the last height as the one returned.

x/qgb/abci.go
```diff
@@ -4,21 +4,29 @@ import (
...
 	"github.com/celestiaorg/celestia-app/x/qgb/keeper"
 	"github.com/celestiaorg/celestia-app/x/qgb/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
-// that would need the creation of a new valset request.
-const SignificantPowerDifferenceThreshold = 0.05
+const (
+	// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
```
rootulp, line 14:
```suggestion
	// SignificantPowerDifferenceThreshold is the threshold of change in the validator set power
```

x/qgb/abci.go
```diff
@@ -4,21 +4,29 @@ import (
...
+	// that would need the creation of a new valset request.
```
rootulp, line 15:
```suggestion
	// that would trigger the creation of a new valset request.
```

x/qgb/abci.go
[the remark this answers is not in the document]
rootulp: [nit] if we are retaining this constant, proposal to rename to something like

```suggestion
	// MinimumAvailableAttestations is the minimum number of attestations that will always be present
	// in state. Older attestations will be pruned.
	MinimumAvailableAttestations = 100
```

because it's not clear what happens below or above the "threshold". In other words are attestations pruned below the threshold or above the threshold?

x/qgb/abci.go
```diff
@@ -113,3 +121,89 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 		}
 	}
 }
+
+// PruneIfNeeded runs basic checks on saved attestations to see if we need to prune or not.
+// Also, proposes some gas optimizations by adding more checks to return before running any
+// of the pruning logic.
+func PruneIfNeeded(ctx sdk.Context, k keeper.Keeper) {
```
rootulp, line 128: [question] 

1. Does this function need to be exported?
1. Are the gas optimizations an implementation detail? They don't appear to be "returned" from this function.

```go
// maybePruneAttestations prunes attestations from state if...
func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
```

x/qgb/abci.go
```diff
@@ -113,3 +121,89 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		// checking if it's the initial case following the startup of the chain
+		if lastAvailableNonce == 1 && lastUnbondingNonce == 0 {
+			return
+		}
+		// we should never hit this case, since we will keep the attestations up to the last unbonding nonce
+		ctx.Logger().Error("missing attestations up to the unbonding nonce")
+		return
+	}
+	// now we have attestations before the unbonding height, we should check whether we need to prune them
+	// or not yet
```
rootulp, line 155:
```suggestion
```

x/qgb/abci.go
```diff
@@ -113,3 +121,89 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		return
+	}
+	lastAvailableNonce := k.GetLastAvailableAttestationNonce(ctx)
+	lastUnbondingNonce := k.GetLastUnbondingNonce(ctx)
+	if lastAvailableNonce == lastUnbondingNonce {
+		// we don't need to do anything, we want to keep attestations up to the last unbonding nonce
+		return
+	}
+	if lastAvailableNonce > lastUnbondingNonce {
+		// checking if it's the initial case following the startup of the chain
```
rootulp, line 146:
```suggestion
		// check if it's the initial case following the startup of the chain
```

x/qgb/abci.go
```diff
@@ -113,3 +121,89 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+// pruneAttestations prunes attestations to keep attestations up to the last unbonding height
+// and the total number of attestations lower than the AttestationPruningThreshold.
+func pruneAttestations(
+	ctx sdk.Context,
+	k keeper.Keeper,
+	lastAvailableAttestationNonce uint64,
+	latestAttestationNonce uint64,
+	lastUnbondingNonce uint64,
+) error {
+	ctx.Logger().Debug("pruning attestations from qgb store")
```
rootulp, line 181: [uber nit] for consistency with the log line above.

```go
		ctx.Logger().Error("error while pruning QGB attestations", "error", err.Error())
```

```suggestion
	ctx.Logger().Debug("pruning attestations from QGB store")
```

x/qgb/abci.go
```diff
@@ -113,3 +121,89 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+			// or, we reached the minimum number of attestations we want to keep in store
+			break
+		}
+		k.DeleteAttestation(ctx, newLastAvailableNonce)
+		count++
+	}
+	// persist the new last available attestation nonce
+	k.SetLastAvailableAttestationNonce(ctx, newLastAvailableNonce)
+	ctx.Logger().Debug(
+		"finished pruning attestations from qgb store",
```
rootulp, line 200:
```suggestion
		"finished pruning attestations from QGB store",
```

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,64 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+func (k Keeper) GetLastUnbondingNonce(ctx sdk.Context) uint64 {
+	store := ctx.KVStore(k.storeKey)
+	bytes := store.Get([]byte(types.LastUnbondingNonce))
+	if bytes == nil {
+		panic("nil last unbonding nonce")
+	}
+	return UInt64FromBytes(bytes)
+}
+
+// SetLastUnbondingNonce sets the attestation nonce corresponding the last unbonding height.
```
rootulp, line 133:
```suggestion
// SetLastUnbondingNonce sets the attestation nonce for the last unbonding height.
```

x/qgb/pruning_test.go
```diff
@@ -0,0 +1,237 @@
...
+// TestAttestationsLowerThanPruningThreshold tests the first check for pruning, which is
+// not pruning if the number of attestations is lower than the threshold.
+func TestAttestationsLowerThanPruningThreshold(t *testing.T) {
+	input, ctx := testutil.SetupFiveValChain(t)
+	qgbKeeper := *input.QgbKeeper
+	// set the data commitment window
+	window := uint64(101)
+	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
+
+	// test that no prunning occurs while we're still under the pruning threshold
```
rootulp, line 27:
```suggestion
	// test that no pruning occurs while we're still under the pruning threshold
```

x/qgb/pruning_test.go
```diff
@@ -0,0 +1,237 @@
...
+	for nonce := uint64(1); nonce <= totalNumberOfAttestations; nonce++ {
+		_, found, err := qgbKeeper.GetAttestationByNonce(ctx, nonce)
+		assert.NoError(t, err)
+		assert.True(t, found)
+	}
+
+	// change the last unbonding height to a higher value
+	updateValidatorSet(t, ctx, input)
+	qgbKeeper.SetLastUnBondingBlockHeight(ctx, uint64(ctx.BlockHeight()))
+	qgbKeeper.SetLastUnbondingNonce(ctx, qgbKeeper.GetLatestAttestationNonce(ctx))
```
rootulp, line 102: [nit] consistent casing for `UnBonding` vs `Unbonding`

test/util/common.go
SweeXordious: removed, it was ambigeous

x/qgb/abci.go
SweeXordious: changed

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,36 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+	has := store.Has([]byte(types.LatestAttestationtNonce))
+	return has
+}
+
+// GetEarliestAvailableAttestationNonce returns the earliest available attestation nonce. The
+// nonce is of the last available attestation in store that can be
+// retrieved. Panics if the last available attestation
+// nonce doesn't exit. This value is set on chain startup. However, it won't be
+// written to store until height = 1. Thus, it's mandatory to run
+// `CheckEarliestAvailableAttestationNonce` before calling this method.
```
rootulp, line 94: [uber nit][not blocking] the comments in this PR are wrapped at arbitrary line lengths. https://marketplace.visualstudio.com/items?itemName=stkb.rewrap will rewrap comments to a fixed length so that they appear cleaner.

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,36 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+// CheckEarliestAvailableAttestationNonce returns true if the last available attestation nonce has been initialized
+// in store, and false if not.
+func (k Keeper) CheckEarliestAvailableAttestationNonce(ctx sdk.Context) bool {
+	store := ctx.KVStore(k.storeKey)
+	has := store.Has([]byte(types.LatestAttestationtNonce))
+	return has
+}
+
+// GetEarliestAvailableAttestationNonce returns the earliest available attestation nonce. The
+// nonce is of the last available attestation in store that can be
```
rootulp, line 90:
```suggestion
// nonce is of the earliest available attestation in store that can be
```

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,36 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+// retrieved. Panics if the last available attestation
```
rootulp, line 91:
```suggestion
// retrieved. Panics if the earliest available attestation
```

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,36 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+func (k Keeper) GetEarliestAvailableAttestationNonce(ctx sdk.Context) uint64 {
+	store := ctx.KVStore(k.storeKey)
+	bytes := store.Get([]byte(types.EarliestAvailableAttestationNonce))
+	if bytes == nil {
+		panic("nil earliest available attestation nonce")
+	}
+	return UInt64FromBytes(bytes)
+}
+
+// SetEarliestAvailableAttestationNonce sets the last available attestation nonce.
```
rootulp, line 104:
```suggestion
// SetEarliestAvailableAttestationNonce sets the earliest available attestation nonce.
```

x/qgb/keeper/keeper_attestation.go
```diff
@@ -78,6 +78,36 @@ func (k Keeper) GetLatestAttestationNonce(ctx sdk.Context) uint64 {
...
+// The nonce is of the last available attestation in store that can be retrieved.
```
rootulp, line 105:
```suggestion
// The nonce is of the earliest available attestation in store that can be retrieved.
```

x/qgb/keeper/keeper_data_commitment.go
```diff
@@ -116,8 +124,12 @@ func (k Keeper) HasDataCommitmentInStore(ctx sdk.Context) (bool, error) {
 	if !k.CheckLatestAttestationNonce(ctx) {
 		return false, nil
 	}
+	if !k.CheckEarliestAvailableAttestationNonce(ctx) {
+		return false, types.ErrEarliestAvailableNonceStillNotInitialized
+	}
 	latestNonce := k.GetLatestAttestationNonce(ctx)
-	for i := uint64(1); i <= latestNonce; i++ {
+	lastAvailableNonce := k.GetEarliestAvailableAttestationNonce(ctx)
+	for i := lastAvailableNonce; i <= latestNonce; i++ {
```
rootulp, line 132:
```suggestion
	earliestAvailableNonce := k.GetEarliestAvailableAttestationNonce(ctx)
	for i := earliestAvailableNonce; i <= latestNonce; i++ {
```

x/qgb/keeper/keeper_valset.go
```diff
@@ -17,15 +17,19 @@ import (
...
 	}
-	nonce := k.GetLatestAttestationNonce(ctx)
-	for i := uint64(0); i <= nonce; i++ {
-		at, found, err := k.GetAttestationByNonce(ctx, nonce-i)
+	if !k.CheckEarliestAvailableAttestationNonce(ctx) {
+		return nil, types.ErrEarliestAvailableNonceStillNotInitialized
+	}
+	latestNonce := k.GetLatestAttestationNonce(ctx)
+	lastAvailableNonce := k.GetEarliestAvailableAttestationNonce(ctx)
+	for i := latestNonce; i >= lastAvailableNonce; i-- {
```
rootulp, line 31:
```suggestion
	earliestAvailableNonce := k.GetEarliestAvailableAttestationNonce(ctx)
	for i := latestNonce; i >= earliestAvailableNonce; i-- {
```

x/qgb/keeper/keeper_valset.go
```diff
@@ -143,23 +152,30 @@ func (k Keeper) GetLastValsetBeforeNonce(ctx sdk.Context, nonce uint64) (*types.
...
+	}
 	if nonce > k.GetLatestAttestationNonce(ctx) {
 		return nil, types.ErrNonceHigherThanLatestAttestationNonce
 	}
-	// starting at 1 because the current nonce can be a valset
+	// starting at nonce-1 because the current nonce can be a valset
 	// and we need the previous one.
-	for i := uint64(1); i < nonce; i++ {
-		at, found, err := k.GetAttestationByNonce(ctx, nonce-i)
+	for i := nonce - 1; i >= lastAvailableNonce; i-- {
```
rootulp, line 170:
```suggestion
	for i := nonce - 1; i >= earliestAvailableNonce; i-- {
```

rootulp (changes requested):
[blocking] in a few instances I think `lastAvailableNonce` is used when `earliestAvailableNonce` is meant.

Besides that, I think everything else looks good

x/qgb/abci.go
```diff
@@ -114,3 +122,67 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+		if !found {
+			ctx.Logger().Error("couldn't find attestation for pruning", "nonce", newEarliestAvailableNonce)
+			return
+		}
+		if newEarliestAttestation == nil {
+			ctx.Logger().Error("nil attestation for pruning", "nonce", newEarliestAvailableNonce)
+			return
+		}
+		if currentBlockTime.Before(newEarliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
+			// the earliest attestation is still valid. this means all the subsequent ones are also
```
evan-forbes, line 172: same comment as above, I think adding then comparing before is if just checking if a < b with extra steps

x/qgb/keeper/keeper_data_commitment.go
```diff
@@ -85,22 +89,26 @@ func (k Keeper) GetDataCommitmentForHeight(ctx sdk.Context, height uint64) (type
...
 	}
+	if !k.CheckEarliestAvailableAttestationNonce(ctx) {
+		return types.DataCommitment{}, types.ErrEarliestAvailableNonceStillNotInitialized
+	}
 	latestNonce := k.GetLatestAttestationNonce(ctx)
-	for i := uint64(0); i < latestNonce; i++ {
-		att, found, err := k.GetAttestationByNonce(ctx, latestNonce-i)
+	lastAvailableNonce := k.GetEarliestAvailableAttestationNonce(ctx)
+	for i := latestNonce; i >= lastAvailableNonce; i-- {
+		att, found, err := k.GetAttestationByNonce(ctx, i)
```
evan-forbes, line 106: var names

x/qgb/keeper/keeper_valset.go
```diff
@@ -17,15 +17,19 @@ import (
...
 		return nil, types.ErrLatestAttestationNonceStillNotInitialized
 	}
-	nonce := k.GetLatestAttestationNonce(ctx)
-	for i := uint64(0); i <= nonce; i++ {
-		at, found, err := k.GetAttestationByNonce(ctx, nonce-i)
+	if !k.CheckEarliestAvailableAttestationNonce(ctx) {
+		return nil, types.ErrEarliestAvailableNonceStillNotInitialized
+	}
+	latestNonce := k.GetLatestAttestationNonce(ctx)
+	lastAvailableNonce := k.GetEarliestAvailableAttestationNonce(ctx)
```
evan-forbes, line 30: var naming

x/qgb/keeper/keeper_valset.go
```diff
@@ -43,7 +47,12 @@ func (k Keeper) GetLatestValset(ctx sdk.Context) (*types.Valset, error) {
...
 		}
 	}
-	// should never execute
+	if lastAvailableNonce > 1 {
+		// this means that the latest valset was pruned and we can return the current one as no significant
+		// changes to it happened
+		currentVs, err := k.GetCurrentValset(ctx)
+		return &currentVs, err
+	}
 	panic(errors.Wrap(sdkerrors.ErrNotFound, "couldn't find latest valset"))
```
evan-forbes, line 56: very optional, but this feels like we could instead check `if lastAvailable == 1` panic, then proceed to return the current one.

[question]
if the first nonce always a valset?

evan-forbes (commented):
only one or two blocking comments or discussions that I'll put here now, I still need to go through the new test to make sure we are covering everything, and will do that soon

x/qgb/abci.go
```diff
@@ -114,3 +122,67 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
 		}
 	}
 }
+
+// maybePruneAttestations runs basic checks on saved attestations to see if we need to prune or not.
+// Then, it starts pruning the expired attestations until all attestations in state are still valid.
+func maybePruneAttestations(ctx sdk.Context, k keeper.Keeper) {
```
evan-forbes, line 128: [not blocking]
I think its a bit clearer to just have this named to pruneAttestations since there is only one situation in which we don't prune. Then simply say in the docs that this func prunes all attestations that have surpassed the expiry point, which is a very accurate description of what is occuring

x/qgb/abci_test.go
```diff
@@ -526,3 +520,39 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
...
+	qgbKeeper := *input.QgbKeeper
+	// set the data commitment window
+	window := uint64(101)
+	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
+	initialBlockTime := ctx.BlockTime()
+	// make the interval between blocks being 3 days
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 5000, time.Hour)
+
+	earliestAttestationNonce := qgbKeeper.GetEarliestAvailableAttestationNonce(ctx)
+	assert.LessOrEqual(t, earliestAttestationNonce, qgbKeeper.GetLatestAttestationNonce(ctx))
```
evan-forbes, line 537: what is this testing for? just that earliest and latest are working as expected?

x/qgb/abci_test.go
```diff
@@ -526,3 +520,39 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
...
+
+	// check that the first attestations were pruned
+	for nonce := uint64(1); nonce < earliestAttestationNonce; nonce++ {
+		_, found, err := qgbKeeper.GetAttestationByNonce(ctx, nonce)
+		assert.NoError(t, err)
+		assert.False(t, found)
+	}
```
evan-forbes, line 544: [blocking]
do we need to check that earliest attestation nonce is > 1 here? can we be specific with how many nonce we expect to be pruned?

evan-forbes (commented):
I think we also need to test that nothing is pruned until the expiry period is reached, if all of the valsets have been pruned, we still get the existing one, and we should probably include some valset changes in the nonces that are being pruned

then I think we're good to go!

test/util/common.go
```diff
@@ -485,3 +488,27 @@ func NewTestMsgUnDelegateValidator(address sdk.ValAddress, amt cosmosmath.Int) *
...
+		ctx = ctx.WithBlockHeight(i)
+		qgb.EndBlocker(ctx, qgbKeeper)
+	}
+	return ctx
+}
+
+// ExecuteQGBHeights executes the end exclusive range of heights specified by beginHeight and endHeight
+// along with the QGB abci.EndBlocker on each one of them.
+// Uses the interval to calculate the block header time.
+func ExecuteQGBHeightsWithTime(ctx sdk.Context, qgbKeeper keeper.Keeper, beginHeight int64, endHeight int64, interval time.Duration) sdk.Context {
```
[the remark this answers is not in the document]
SweeXordious: [51cb3eb](https://github.com/celestiaorg/celestia-app/pull/1799/commits/51cb3eba2de0df4f9c838d60e535573be32a5e8a)

x/qgb/abci_test.go
```diff
@@ -526,3 +520,39 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
...
+// ones are pruned
+func TestPruning(t *testing.T) {
+	input, ctx := testutil.SetupFiveValChain(t)
+	qgbKeeper := *input.QgbKeeper
+	// set the data commitment window
+	window := uint64(101)
+	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
+	initialBlockTime := ctx.BlockTime()
+	// make the interval between blocks being 3 days
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 5000, time.Hour)
```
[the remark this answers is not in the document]
SweeXordious: [36742bc](https://github.com/celestiaorg/celestia-app/pull/1799/commits/36742bcdc6b1b0222999aca09396e976e1d6948f) thanks, it should be one hour

x/qgb/keeper/keeper_attestation.go
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1819

x/qgb/abci.go
```diff
@@ -2,23 +2,31 @@ package qgb
...
-// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
-// that would need the creation of a new valset request.
-const SignificantPowerDifferenceThreshold = 0.05
+const (
+	// SignificantPowerDifferenceThreshold is the threshold of change in the validator set power
+	// that would trigger the creation of a new valset request.
+	SignificantPowerDifferenceThreshold = 0.05
+
+	// AttestationExpiryTime the expiration time of an attestation after which it will be pruned.
+	AttestationExpiryTime = 8 * 7 * 24 * time.Hour
```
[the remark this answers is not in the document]
SweeXordious: Forgot to change this after our sync discussion. I am thinking of leaving it for 4 weeks now, to have more attestations, and then we can reduce it if needed via an upgrade. What do you think?

x/qgb/keeper/keeper_valset.go
SweeXordious: > if the first nonce always a valset?

yes

SweeXordious: [c9e926a](https://github.com/celestiaorg/celestia-app/pull/1799/commits/c9e926a06ad6adda74745ce712dff6d46b861241)

x/qgb/abci.go
SweeXordious: [23aac61](https://github.com/celestiaorg/celestia-app/pull/1799/commits/23aac61c090fdab90631f7d7d24488bcb8a0c27b)

x/qgb/keeper/keeper_attestation.go
```diff
@@ -93,3 +123,14 @@ func (k Keeper) GetAttestationByNonce(ctx sdk.Context, nonce uint64) (types.Atte
...
 }
+
+// DeleteAttestation deletes an attestation from state.
+func (k Keeper) DeleteAttestation(ctx sdk.Context, nonce uint64) {
+	key := []byte(types.GetAttestationKey(nonce))
+	store := ctx.KVStore(k.storeKey)
+	if !store.Has(key) {
+		// if the store doesn't have the needed attestation, then no need to do anything.
+		return
+	}
```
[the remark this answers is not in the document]
SweeXordious: [2783d20](https://github.com/celestiaorg/celestia-app/pull/1799/commits/2783d20b1c29c69be90670b41e416394b30eaddf)

x/qgb/abci_test.go
SweeXordious: [302f9a2](https://github.com/celestiaorg/celestia-app/pull/1799/commits/302f9a26d4b097e9f5a5132871f8dcec4afe6e7b)

x/qgb/abci_test.go
SweeXordious: [302f9a2](https://github.com/celestiaorg/celestia-app/pull/1799/commits/302f9a26d4b097e9f5a5132871f8dcec4afe6e7b)

SweeXordious:
> I think we also need to test that nothing is pruned until the expiry period is reached, if all of the valsets have been pruned, we still get the existing one, and we should probably include some valset changes in the nonces that are being pruned

[425b3ff](https://github.com/celestiaorg/celestia-app/pull/1799/commits/425b3ffcd080460fa885988c73163fe879a24189) and [302f9a2](https://github.com/celestiaorg/celestia-app/pull/1799/commits/302f9a26d4b097e9f5a5132871f8dcec4afe6e7b)

x/qgb/abci.go
```diff
@@ -114,3 +122,67 @@ func handleValsetRequest(ctx sdk.Context, k keeper.Keeper) {
...
+	if earliestAttestation == nil {
+		ctx.Logger().Error("nil earliest attestation")
+		return
+	}
+	currentBlockTime := ctx.BlockTime()
+	// if the current time is before the earliest attestation creation time + expiry time
+	// then, all the subsequent attestations are also still valid and no need to prune them.
+	if currentBlockTime.Before(earliestAttestation.BlockTime().Add(AttestationExpiryTime)) {
+		return
+	}
```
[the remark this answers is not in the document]
SweeXordious: [7986e04](https://github.com/celestiaorg/celestia-app/pull/1799/commits/7986e04c63eb7a419646978fe36ebe714c65291f)

x/qgb/abci.go
SweeXordious: [7986e04](https://github.com/celestiaorg/celestia-app/pull/1799/commits/7986e04c63eb7a419646978fe36ebe714c65291f)

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: [05b229e](https://github.com/celestiaorg/celestia-app/pull/1799/commits/05b229ecf731bc31c330d3e0edc7731542433940) something like this?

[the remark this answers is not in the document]
evan-forbes: yeah that works great thanks

x/qgb/abci.go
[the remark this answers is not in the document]
SweeXordious: [00fc81b](https://github.com/celestiaorg/celestia-app/pull/1799/commits/00fc81b2793777d71fa34d798ff95a64012ed50d)

x/qgb/abci_test.go
```diff
@@ -526,3 +520,68 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
...
+		_, found, err := qgbKeeper.GetAttestationByNonce(ctx, nonce)
+		assert.NoError(t, err)
+		assert.True(t, found)
+	}
+
+	// continue executing heights
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1345, 5000, time.Hour)
+
+	earliestAttestationNonce := qgbKeeper.GetEarliestAvailableAttestationNonce(ctx)
+	assert.Equal(t, uint64(38), earliestAttestationNonce)
```
evan-forbes, line 552: should this test have failed after switching to three weeks?

evan-forbes (dismissed):
nice!! thanks for all the efforts and patience @SweeXordious, I feel a lot better about this after the refactor

left one other open question, but approving!

x/qgb/abci_test.go
SweeXordious: Yeeeeeeeeeeeees, and it is failing :D meaning it works haha I will fix now

SweeXordious: [6396271](https://github.com/celestiaorg/celestia-app/pull/1799/commits/6396271f61d2fe02385ddcbe05a9180fd322fbe3)

x/qgb/abci.go
```diff
@@ -2,23 +2,31 @@ package qgb
...
-// SignificantPowerDifferenceThreshold the threshold of change in the validator set power
-// that would need the creation of a new valset request.
-const SignificantPowerDifferenceThreshold = 0.05
+const (
+	// SignificantPowerDifferenceThreshold is the threshold of change in the validator set power
+	// that would trigger the creation of a new valset request.
+	SignificantPowerDifferenceThreshold = 0.05
+
+	// AttestationExpiryTime the expiration time of an attestation after which it will be pruned.
+	AttestationExpiryTime = 3 * 7 * 24 * time.Hour
```
rootulp, line 20: [nit] this can be improved per https://github.com/celestiaorg/celestia-app/pull/1799#discussion_r1202575777

x/qgb/abci_test.go
```diff
@@ -526,3 +520,68 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
...
+// 2. Running the QGB EndBlocker
+// 3. Verifying that the expired attestations are pruned
+func TestPruning(t *testing.T) {
+	input, ctx := testutil.SetupFiveValChain(t)
+	qgbKeeper := *input.QgbKeeper
+	// set the data commitment window
+	window := uint64(101)
+	qgbKeeper.SetParams(ctx, types.Params{DataCommitmentWindow: window})
+	initialBlockTime := ctx.BlockTime()
+	// make the interval between blocks being 10 minutes
```
rootulp, line 535:
```suggestion
	// make the interval between blocks 10 minutes
```

or alternatively remove the comment and document via code:

```go
blockInterval := 10 * time.Minute
ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 1626, blockInterval)
```

x/qgb/abci_test.go
```diff
@@ -526,3 +520,68 @@ func TestDataCommitmentCreationCatchup(t *testing.T) {
...
+	ctx = testutil.ExecuteQGBHeightsWithTime(ctx, qgbKeeper, 1, 1626, 10*time.Minute)
+
+	// check that we created a number of attestations
+	assert.Equal(t, uint64(17), qgbKeeper.GetLatestAttestationNonce(ctx))
+
+	// check that no pruning occurs if the no attestation is expired
```
rootulp, line 541:
```suggestion
	// check that no pruning occurs if no attestations expired
```

x/qgb/keeper/keeper_data_commitment.go
```diff
@@ -85,22 +89,26 @@ func (k Keeper) GetDataCommitmentForHeight(ctx sdk.Context, height uint64) (type
 			return *dcc, nil
 		}
 	}
-	return types.DataCommitment{}, errors.Wrap(types.ErrDataCommitmentNotFound, "data commitment for height not found")
+	return types.DataCommitment{}, errors.Wrap(types.ErrDataCommitmentNotFound, "data commitment for height not found or was pruned")
 }
 
 // GetLastDataCommitment returns the last data commitment.
 func (k Keeper) GetLastDataCommitment(ctx sdk.Context) (types.DataCommitment, error) {
```
rootulp, line 96: [nit][not-blocking] proposal to conform to the earliest/latest naming pattern because last is a bit confusing (is it most recent or least recent?)

```suggestion
// GetLastDataCommitment returns the last data commitment.
func (k Keeper) GetLatestDataCommitment(ctx sdk.Context) (types.DataCommitment, error) {
```

Also this conforms to `GetLatestValset`

rootulp (approved):

evan-forbes (approved):

SweeXordious:
@rootulp The remaining changes I will handle them in a subsequent PR. I don't want to wait another round of reviews for this one as it's taking too much time.

Thanks a lot everyone for the thorough reviews :+1: :+1:

x/qgb/keeper/keeper_data_commitment.go
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1821

x/qgb/abci.go
SweeXordious: https://github.com/celestiaorg/celestia-app/pull/1822

x/qgb/abci_test.go
SweeXordious: https://github.com/celestiaorg/celestia-app/pull/1823

PR #1803, opened 2023-05-21
feat!: upper bound for data commitment window

SweeXordious:
## Overview

Use the constant in core as an upper bound for the data commitment window

### DIFF
--- a/x/qgb/types/genesis.go
+++ b/x/qgb/types/genesis.go
@@ -3,6 +3,8 @@ package types
 import (
 	"fmt"
 
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+
 	"cosmossdk.io/errors"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 )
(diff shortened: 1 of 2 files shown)


x/qgb/types/genesis.go
```diff
@@ -61,6 +63,12 @@ func validateDataCommitmentWindow(i interface{}) error {
 			val,
 			MinimumDataCommitmentWindow,
 		))
+	} else if val > uint64(appconsts.DataCommitmentBlocksLimit) {
```
evan-forbes, line 66: not super related to this PR, but we don't need "else if" here, we can just have an if statement since we're returning

evan-forbes (approved):

evan-forbes (approved):

x/qgb/types/genesis.go
SweeXordious: [bebbbf9](https://github.com/celestiaorg/celestia-app/pull/1803/commits/bebbbf9b1343d52d2300b885c26465f633e2617d)

evan-forbes (approved):

PR #1822, opened 2023-05-23
chore: comprehensive expiry time constant

SweeXordious:
## Overview

Handles https://github.com/celestiaorg/celestia-app/pull/1799#discussion_r1202575777

## Checklist

### DIFF
--- a/x/qgb/abci.go
+++ b/x/qgb/abci.go
@@ -16,8 +16,12 @@ const (
 	// that would trigger the creation of a new valset request.
 	SignificantPowerDifferenceThreshold = 0.05
 
-	// AttestationExpiryTime the expiration time of an attestation after which it will be pruned.
-	AttestationExpiryTime = 3 * 7 * 24 * time.Hour
+	oneDay  = 24 * time.Hour
+	oneWeek = 7 * oneDay
+	// AttestationExpiryTime is the expiration time of an attestation.
+	// When this much time has passed after an attestation has been published, it will be
+	// pruned from state.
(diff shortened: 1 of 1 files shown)


rootulp (approved):

x/qgb/abci.go
```diff
@@ -16,8 +16,12 @@ const (
...
 	SignificantPowerDifferenceThreshold = 0.05
 
-	// AttestationExpiryTime the expiration time of an attestation after which it will be pruned.
-	AttestationExpiryTime = 3 * 7 * 24 * time.Hour
+	oneDay  = 24 * time.Hour
+	oneWeek = 7 * oneDay
+	// AttestationExpiryTime is the expiration time of an attestation.
+	// When this much time has passed after an attestation has been published, it will be
+	// pruned from state.
+	AttestationExpiryTime = 3 * oneWeek // 3 weeks
```
cmwaters, line 24: Since this is supposed to equate to the unbonding time and unbonding time is supposed to be a global constant (i.e. we've disallowed changing the unbonding time via governance), shall we also introduce a `DefaultUnbondingPeriod` in `appconsts` and ensure that these two values match

SweeXordious: Is there a constant somewhere tracking the unbonding period length? If so, yes, makes sense to use it instead of this.

For making this value upgradable, i dont think we will ever want to change it from the unbonding period. And if we do, then we can add the upgrade logic

SweeXordious: Cool. Thanks

PR #1843, opened 2023-05-26
chore: backport main for rc1

evan-forbes:
## Overview

cherry picks all changes from main for the v1.x branch, including #1842 

should be merged or rebased

## Checklist

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -70,6 +70,7 @@ type stakingModule struct {
 // DefaultGenesis returns custom x/staking module genesis state.
 func (stakingModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
 	params := stakingtypes.DefaultParams()
+	params.UnbondingTime = appconsts.DefaultUnbondingTime
 	params.BondDenom = BondDenom
 
 	return cdc.MustMarshalJSON(&stakingtypes.GenesisState{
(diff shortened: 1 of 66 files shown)


SweeXordious (approved):

SweeXordious (approved):

evan-forbes:
@MSevey when you get the chance can we have branch rules or something that require 2 reviews for release branches if we don't already? looks like I could merge this already with only 1 review

MSevey:
> @MSevey when you get the chance can we have branch rules or something that require 2 reviews for release branches if we don't already? looks like I could merge this already with only 1 review

it is enabled, the problem is that this branch is called `v1.x` which doesn't match any branch protection rule. 

It should be `v1.0.x`

evan-forbes:
> It should be v1.0.x

I see, the thinking there was that we would have a single release branch per major version, not minor, since afaiu we're not planning on maintaining old minor releases. are we planning on having and maintaining release branches for each minor release? such as 

`v1.1.x`, `v1.2.x`, etc

rootulp (approved):

cmwaters:
I don't think we need to maintain any minor release that is not the latest. We might still need to have branches v1.1.x and v1.2.x. Imagine we're working on a new minor release and there's a bug on the main network. We want to patch the version in current use. But anytime a user wants to upgrade. It should always be to the latest version within the major version not within the minor

Issue #1815, opened 2023-05-23
Investigate the valset creation normalization


SweeXordious:
We need to take a look on these lines, taken from the gravity modules, in the QGB state machine:

https://github.com/celestiaorg/celestia-app/blob/9f3539ebe6a65bd960baa43991ef245f3fc8e5f1/x/qgb/keeper/keeper_valset.go#L83-L103

And see if we really need to normalize the powers from 64 to 32 bits. Also, we need to check if this will be needed on the QGB smart contract.

cc @evan-forbes

SweeXordious:
Might also be worth looking at these: https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/types/validator.go

SweeXordious:
Around the normalization:
```
// Normalized voting power dramatically reduces how often we have to produce new
// validator set updates. For example if the total on chain voting power
// increases by 1% due to inflation, we shouldn't have to generate a new
// validator set, after all the validators retained their relative percentages
// during inflation and normalized Gravity bridge power shows no difference.
```

SweeXordious:
The implementation seems good. It is using the min-max normalization from the `[0, total validator power]` to `[0, 2^32]` to avoid creating valsets due to inflation for example, and only when a significant change happens to the valset.

PR #1846, opened 2023-05-26
docs: update the valset normalization docs

SweeXordious:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/1815

## Checklist


x/qgb/types/validator.go
```diff
@@ -97,22 +97,23 @@ func EVMAddrLessThan(e common.Address, o common.Address) bool {
...
 
-// PowerDiff returns the difference in power between two bridge validator sets
-// note this is Gravity bridge power *not* Cosmos voting power. Cosmos voting
-// power is based on the absolute number of tokens in the staking pool at any
-// given time Gravity bridge power is normalized using the equation.
+// PowerDiff returns the difference in power between two bridge validator sets.
+// Note this is QGB bridge power *not* Cosmos voting power.
+// Cosmos voting power is based on the absolute number of tokens in the staking
+// pool at any given time.
+// QGB bridge power is normalized using the equation.
```
evan-forbes, line 104: [not blocking]

I think we should remain consistent with the rest of the repo and either wrap or if we feel the need to space things out, then we should 
```go
// sentence
//
// Note: sentence 2

**Review by evan-forbes — APPROVED** [review]

File: x/qgb/types/validator.go (hunk 1 above)
Comment (SweeXordious) [reply x/qgb/types/validator.go:104]: I unwrapped the comment so that its easier to read. Should i wrap it again?

**Review by rootulp — APPROVED** [review]

File: x/qgb/keeper/keeper_valset.go [hunk 2]
```diff
@@ -125,15 +123,18 @@ func (k Keeper) GetCurrentValset(ctx sdk.Context) (types.Valset, error) {
 }
 
 // normalizeValidatorPower scales rawPower with respect to totalValidatorPower
-// to take a value between 0 and 2^32 Uses BigInt operations to avoid overflow
-// errors Example: rawPower = max (2^63 - 1), totalValidatorPower = 1 validator:
-// (2^63 - 1)
+// to take a value between 0 and 2^32.
+// Uses BigInt operations to avoid overflow errors.
+// Example: rawPower = max (2^63 - 1), totalValidatorPower = 1 validator: (2^63 - 1)
```
staheri14, line 128: [nit] Is a comma missing here?
```suggestion
// Example: rawPower = max (2^63 - 1), totalValidatorPower = 1, validator: (2^63 - 1)
```
The clarity of the description can be improved e.g., it is not immediate that there is only one validator (if I understood it correctly)

x/qgb/keeper/keeper_valset.go
```diff
@@ -125,15 +123,18 @@ func (k Keeper) GetCurrentValset(ctx sdk.Context) (types.Valset, error) {
...
 //
 //	result: (2^63 - 1) * 2^32 / (2^63 - 1) = 2^32 = 4294967296 [this is the multiplier value below, our max output]
 //
 // Example: rawPower = max (2^63 - 1), totalValidatorPower = 1000 validators with the same power: 1000*(2^63 - 1)
 //
 //	result: (2^63 - 1) * 2^32 / (1000(2^63 - 1)) = 2^32 / 1000 = 4294967
+//
+// This is using the min-max normalization from the interval [0, total validator power] to [0, 2^32].
+// Check the `PowerDiff` method under `types.validator.go` for more information.
 func normalizeValidatorPower(rawPower uint64, totalValidatorPower cosmosmath.Int) uint64 {
```
staheri14, line 138: [Question] Could you clarify the origin of the number `4294967296` which is used later in the code (in addition to what is already in the function doc)? Is the rationale behind its use for normalization explained in any specifications or ADR? I would appreciate pointers for further reading. 🙏

staheri14 (approved):
LGTM! included some questions.

x/qgb/types/validator.go
```diff
@@ -97,22 +97,23 @@ func EVMAddrLessThan(e common.Address, o common.Address) bool {
...
 // As an example if someone has 52% of the Cosmos voting power when a validator
-// set is created their Gravity bridge voting power is u32_max * .52
+// set is created their QGB bridge voting power is u32_max * .52
 //
 // Normalized voting power dramatically reduces how often we have to produce new
 // validator set updates. For example if the total on chain voting power
 // increases by 1% due to inflation, we shouldn't have to generate a new
 // validator set, after all the validators retained their relative percentages
-// during inflation and normalized Gravity bridge power shows no difference.
+// during inflation and normalized QGB bridge power shows no difference.
```
cmwaters, line 116:
```suggestion
// during inflation and normalized QGB power shows no difference.
```
Not sure if we should also be removing the `bridge` part as well since QGB = Quantum Gravity Bridge so QGB bridge stutters.

x/qgb/types/validator.go
evan-forbes: > I unwrapped the comment so that its easier to read. Should i wrap it again?

imo yeah, just so that we should probably remain consistent. In almost all, if not all, cases we wrap by line length and not sentence per standard go. we don't have a linter of it, so its not blocking, but we might want to to avoid future review feedback cycles

evan-forbes: I'm not super concerned about it however

SweeXordious: > In almost all, if not all, cases we wrap by line length and not sentence per standard go. we don't have a linter of it, so its not blocking, but we might want to to avoid future review feedback cycles

I understand, and that's  how I do it in the other docs. However, in this case, since it contains a lot of information, and is not really easy to understand, putting each sentence in a separate line made it easier to read and understand. And since we're not having any standards in the repo for comments, it made more sense to format it this way so anyone who reads it, doesn't have to separate it manually

x/qgb/keeper/keeper_valset.go
SweeXordious: Yes, could be improved. This comment was copied from the gravity bridge, that's why I left it the way it is and only tried to reformat it

x/qgb/keeper/keeper_valset.go
SweeXordious: This is taken from the gravity bridge implementation. 

The value comes from here: https://en.wikipedia.org/wiki/Feature_scaling

If we take this formula: 
image

with: `a=0`, `b=2^32`, `min(x) = 0`, `max(x) = total validator power`, we will find the multiplier `4294967296` being  `2^32`

x/qgb/types/validator.go
evan-forbes: imo its more difficult to read lol :smile:

I'm approving, but would prefer that wrap like normal in the future. Going back and forth on silly things like this is not worth our time :slightly_smiling_face: 

I don't know of any linters that wrap comments in a specific way, and would prefer to just agree on a max comment line length and let our IDEs do the rest.

evan-forbes (dismissed):

x/qgb/types/validator.go
SweeXordious: Got you. Lets change even the docs to be more understable when writing specs. I think a lot of things could be reformulated

x/qgb/keeper/keeper_valset.go
cmwaters: Could be a good idea to link the origins of that formula in the comments

cmwaters (dismissed):

x/qgb/keeper/keeper_valset.go
SweeXordious: [bf7af2d](https://github.com/celestiaorg/celestia-app/pull/1846/commits/bf7af2d18da840fbe3c3d4937279dd7a981b9001)

cmwaters (approved):

evan-forbes (approved):

x/qgb/keeper/keeper_valset.go
staheri14: Thank you for providing clarification on the formula. Just to confirm, when referring to the term "total validator power" in the equation `max(x) = total validator power,` does it represent the maximum power of the individual validators, or does it represent the combined power of all validators? According to the formula, it appears to refer to the former, but I would like to double-check for accuracy.

SweeXordious: > According to the formula, it appears to refer to the former,

Why is that? I think it refers to the total validator power of the whole validator set

staheri14: Nevermind, I figured the answer, it is the combined power of all the validators.

staheri14 (approved):

x/qgb/keeper/keeper_valset.go
staheri14: > According to the formula, it appears to refer to the former,

> Why is that? I think it refers to the total validator power of the whole validator set

Just saw your reply, yes, that's right.

staheri14: However, based on the example in the wiki link:
". For example, suppose that we have the students' weight data, and the students' weights span [160 pounds, 200 pounds]. "
It takes the max weight of individual students as the max for the normalization, not the aggregate of all of the weights. 
This was the reason I asked whether we consider the aggregate or the max of each individual validator.  @SweeXordious

Issue #1757, opened 2023-05-12
Add `genesis_time` to mint module's genesis state


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/1578#discussion_r1189580092

## Problem

Common fields (e.g. [`genesis_time`](https://github.com/celestiaorg/networks/blob/309f961dcdc7e7353e8143cc59c1797e4c0fc326/blockspacerace/genesis.json#L2)) in the `genesis.json` can't be accessed by modules so the x/mint module works around this by maintaining a field of state called `genesis_time` that is set when `block_height=1`. This is a bit hacky because the check for block_height=1 occurs on every block. 

## Proposal

One potential workaround is to add the `genesis_time` to the genesis state of the x/mint module. In this way, we'll have two fields in the genesis.json so something like:

```diff
{
  "genesis_time": "2023-03-14T16:00:00.000000000Z",
  "chain_id": "blockspacerace-0",
  "initial_height": "1",
  // truncated
    "mint": {
      "minter": {
        "inflation": "0.130000000000000000",
        "annual_provisions": "0.000000000000000000"
+       "genesis_time": "2023-03-14T16:00:00.000000000Z"
      },
```

PR #1851, opened 2023-05-29
feat!: remove genesis time from genesis state

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1757

## Description

Prior to this PR, the mint module contained a placeholder `genesis_time` in genesis state. It ignored the placeholder and used the block time for the block height `1` as the "genesis time". This PR removes the placeholder and that mechanism. 

```diff
      "minter": {
        "inflation_rate": "0.080000000000000000",
        "annual_provisions": "0.000000000000000000",
-       "genesis_time": "2023-06-22T18:46:11.873883Z",
        "previous_block_time": null,
        "bond_denom": "utia"
      }
```

After this PR, `genesis_time` in the mint module is set to the chain's genesis time which is the root level `genesis_time` specified in genesis.json. Genesis.json is created on chain initialization (i.e. `celestia-appd init chain-id`). 

This is breaking b/c it modifies a state transition for the mint module state and removes a field from the `Minter` state.

## Testing

Now there is only one `genesis_time` in genesis.json:

```shell
$ ./scripts/build-run-single-node.sh

$ cat /var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/celestia_app_XXXXXXXXXXXXX.wYBu5mwm/config/genesis.json | grep genesis_time -A 3 -B 3
{
  "genesis_time": "2023-06-23T16:13:24.719488Z",
  "chain_id": "private",
  "initial_height": "1",
  "consensus_params": {
```

Additionally it gets set correctly in mint module state

```shell
$ ./build/celestia-appd query mint genesis-time
2023-06-23 16:13:24.719488 +0000 UTC
```


rootulp:
Note we can use [GenesisOption](https://github.com/celestiaorg/celestia-app/blob/6a57fcd5fff2522fa321e39c483ce53262f714ea/test/util/testnode/genesis_options.go#L15-L18) to overwrite the genesis timestamp in the mint genesis state

evan-forbes (dismissed):
thinking on this more, although I really want to add this to avoid a weird if statement forever or override the genesis state, this technically isn't a bug or a security risk. so it might be too late to include in v1 not unlike #1985 :sob:

evan-forbes:
> this technically isn't a bug or a security risk

after sleeping on this, it is a bug since we are overriding the state in genesis, its just not a very important one. If we can get this into rc5, I think we should.

x/mint/keeper/genesis.go
```diff
@@ -5,13 +5,13 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// InitGenesis new mint genesis
+// InitGenesis initializes the x/mint store with data from the genesis state.
 func (keeper Keeper) InitGenesis(ctx sdk.Context, ak types.AccountKeeper, data *types.GenesisState) {
 	keeper.SetMinter(ctx, data.Minter)
 	ak.GetModuleAccount(ctx, types.ModuleName)
 }
```
cmwaters, line 12:
```suggestion
func (keeper Keeper) InitGenesis(ctx sdk.Context, ak types.AccountKeeper, data *types.GenesisState) {
        genesisTime := ctx.BlockTime()
        // override the genesis time with the actual genesis time supplied in `InitChain`
        data.Minter.GenesisTime = &genesisTime
	keeper.SetMinter(ctx, data.Minter)
	ak.GetModuleAccount(ctx, types.ModuleName)
}
```
I would really prefer that we keep the genesis time identical. Since the block time here is set in `InitChain` which is the [genesis time provided by Tendermint](https://github.com/celestiaorg/celestia-core/blob/a1b07a1e6c77595466da9c61b37c83b4769b47e7/consensus/replay.go#L312), I suggest we do this. Ideally I would not have two genesis times in the genesis file to avoid confusion and just have it set here

evan-forbes: we could do that, but my impression of this PR was to stop overriding the minter's genesis state

if we move it to a different spot, is that even consensus breaking?

evan-forbes: > if we move it to a different spot, is that even consensus breaking?

I guess so since this would be using block time of height 0 instead of 1

cmwaters: Can we modify mint's `GenesisState` so it doesn't include `GenesisTime`? 

> if we move it to a different spot, is that even consensus breaking?

Yes because time at height 1 is different to genesis time

test/util/test_app.go
```diff
@@ -279,6 +281,18 @@ func genesisStateWithValSet(
...
+func genesisStateWithGenesisTime(
+	app *app.App,
+	genesisState app.GenesisState,
+	time time.Time,
+) app.GenesisState {
+	minter := minttypes.DefaultMinter()
+	minter.GenesisTime = &time
+	mintGenesis := minttypes.NewGenesisState(minter)
+	genesisState[minttypes.ModuleName] = app.AppCodec().MustMarshalJSON(mintGenesis)
+	return genesisState
```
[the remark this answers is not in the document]
rootulp: Resolved by c02ea3cb71c3311f41f4e43e23eeb9ea5f527ea0

x/mint/keeper/genesis.go
rootulp: I'll try the approach @cmwaters proposed and (1) remove genesis time from the mint genesis state and (2) set it via `InitGenesis`.

I tried something similar to https://github.com/celestiaorg/celestia-app/pull/1578#discussion_r1196202725 before but couldn't get it to work.

evan-forbes: > Can we modify mint's GenesisState so it doesn't include GenesisTime?

we could but then I think we'd have to also implement extra protections

I think the other option is to do what this PR is doing, not overwrite, and then rely on either the user to copy and paste the value for an important genesis, or we find a way to pass the genesis time when setting the default genesis, which seems really doable imo

Less important as I don't think there are many instances of this, but if we ever wanted to regensis a testnet while also maintaining some minter state, then we couldn't do that if we keep overwriting

evan-forbes: cool, fortunatley whatever is chosen isn't very important. I think the most important thing is to merge #1987 and cut rc5 today

any improvements to the current minter are welcome, will defer @rootulp :slightly_smiling_face: , I'm full :shipit:

evan-forbes (approved):
LGTM

x/mint/keeper/genesis.go
```diff
@@ -5,13 +5,19 @@ import (
...
-// InitGenesis new mint genesis
+// InitGenesis initializes the x/mint store with data from the genesis state.
 func (keeper Keeper) InitGenesis(ctx sdk.Context, ak types.AccountKeeper, data *types.GenesisState) {
 	keeper.SetMinter(ctx, data.Minter)
+	// override the genesis time with the actual genesis time supplied in `InitChain`
+	blockTime := ctx.BlockTime()
+	gt := types.GenesisTime{
+		GenesisTime: &blockTime,
+	}
+	keeper.SetGenesisTime(ctx, gt)
```
SweeXordious, line 16: [Non blocking][To understand more]
This is still setting the genesis time in state, no?

x/mint/types/keys.go
```diff
@@ -1,8 +1,11 @@
 package types
 
-// MintKey is the key to use for the keeper store.
+// MintKey is the key to use for the Minter in the keeper store.
 var MintKey = []byte{0x00}
 
+// GenesisTimeKey is the key to use for GenesisTime in the keeper store.
+var GenesisTimeKey = []byte{0x01}
```
SweeXordious, line 7: [Non blocking]
I think something like this would be better so that if you're dumping the tree, you don't have to check what `0x01` means
```suggestion
var GenesisTimeKey = []byte{"MintGenesisTime"}
```
or similar

SweeXordious: Similar for the mint module

SweeXordious (approved):

x/mint/keeper/genesis.go
rootulp: Yep this is setting the genesis time in the state. Notably it doesn't use the mint module's genesis state `genesis_time` field because that was removed. It uses the root level genesis time from genesis.json

x/mint/types/keys.go
rootulp: > Similar for the mint module

Clarifying question, do you mean the `MintKey` on line 4?

SweeXordious: Yes

rootulp: Resolved via https://github.com/celestiaorg/celestia-app/pull/1851/commits/37d67309f4416186c4aedae41bf2262b0a59b453

evan-forbes (approved):

SweeXordious (approved):

cmwaters (approved):
Nice

As a note, it seems like the only thing Genesis state really needs is the bond denom. Everything else is set or calculated at genesis.

rootulp:
> As a note, it seems like the only thing Genesis state really needs is the bond denom. Everything else is set or calculated at genesis.

Yea good point. TBH I didn't realize I could separate the types so that the genesis state of the module could differ from the entire `Minter` struct but learned that it's possible with https://github.com/celestiaorg/celestia-app/pull/1851/commits/9e333b6d8679174b00df83d163e63fe46115d295

Issue #1555, opened 2023-03-28
specs: define which sdk modules we are using


evan-forbes:
we don't specify which sdk modules we're actually using, which is something we need to do. We can then point to the relevant specs in the sdk, and we should make sure that the few modifications we did are reflected in those specs. The only modification that we've done is, https://github.com/celestiaorg/cosmos-sdk/pull/112, where we added the (now named evm_address) to the validator struct.

PR #1856, opened 2023-05-30
docs: template for spec and initial files

### LINKED ISSUE #1818: Add new templates for the spec
During our internal audit, we will revamping the existing specs and still need to decide on the structure we want. In order to document that structure and remain consistent, we should create templates. 

For example, in the section that describes all of the block validity rules, we could have something along the lines of

```markdown
# Rule N: <name of rule>
<specifiy the exact block validity rule so that ... >

## Mechanism of Enforcement
<is there a fraud proof for this rule? if not, are there plans to include a fraud proof in the future? if so, what does that fraud proof look like?>

## Consequences of the rule being broken
```

or for modules we have 

```markdown
# x/<name

## Abstract
<hight level overview of what the module  does>

## State
<what state if any is being read, updated, written, deleted etc>

## Events
...
```

the end result can look very different from the above, just so long as we have a rough idea over what should be expected for different sections of the specs. This might also mean coming up a new way to organize the specs.

evan-forbes:
## Overview

This is just the initial "get the ball rolling" on the specs by adding the template @staheri14 created in #1818 and the initial links to each README in the state machine code as discussed in the internal audit/spec call. This is not intented to be the final structure :slightly_smiling_face: 

closes #1818, #1555 

## Checklist


specs/src/specs/block_validity_rules.md
```diff
@@ -0,0 +1 @@
+# Block Validity Rules TODO
```
rootulp, line 1:
```suggestion
# Block Validity Rules
```

rootulp:
I think we should add the new pages to these pages so that they show up in the PR preview: https://celestiaorg.github.io/celestia-app/pr-preview/pr-1856/
- https://github.com/celestiaorg/celestia-app/blob/main/specs/src/README.md?plain=1
- https://github.com/celestiaorg/celestia-app/blob/main/specs/src/SUMMARY.md?plain=1

My impression was that the short term approach would be to copy + paste the contents of the x/module/README.md to specs/src/mint/module.md and in x/module/README.md have a link to the corresponding page in the specs. If that's the case, spec owners can do that after this merges.

staheri14:
> My impression was that the short term approach would be to copy + paste the contents of the x/module/README.md to specs/src/mint/module.md and in x/module/README.md have a link to the corresponding page in the specs. If that's the case, spec owners can do that after this merges.

[Suggestion] If we proceed with this approach, then each spec owner would be able (in addition to copying the old content from the  x/module/README.md) to organize the content of the respective file based on the specs template.

staheri14 (approved):
LGTM! 🚀 
[Optional] We could have a README file specifically for the specs folder. This README can provide high-level information about the structure of the specs folder, including details on how the specs are formatted. It can include a link to the template that the specs should follow, as well as any other relevant information (which can be kept concise for now and expanded upon later through iteration).

evan-forbes:
> My impression was that the short term approach would be to copy + paste the contents of the x/module/README.md to specs/src/mint/module.md and in x/module/README.md have a link to the corresponding page in the specs.

this is what was originally suggested iirc, but I believe @MSevey suggested that we have the opposite mapping for now until we have some other setup that pulls the relevant readme's into that package. I'm open to suggestions about structure

MSevey:
> > My impression was that the short term approach would be to copy + paste the contents of the x/module/README.md to specs/src/mint/module.md and in x/module/README.md have a link to the corresponding page in the specs.
> 
> this is what was originally suggested iirc, but I believe @MSevey suggested that we have the opposite mapping for now until we have some other setup that pulls the relevant readme's into that package. I'm open to suggestions about structure

my initial naive thought was that we could just update the summary page to link to the actual pkg READMEs https://github.com/celestiaorg/celestia-app/blob/main/specs/src/SUMMARY.md?plain=1

then you the specs folder is really just a shell. 

a second naive approach was that instead of the spec README having a link to the pkg README, the spec file itself was a symbolic link to the README in the pkgs. 

Both are totally untested, so feel free to completely ignore me and do what makes the most sense to the team :-)

rootulp (dismissed):

staheri14 (dismissed):

liamsi:
I like what @MSevey suggested for the following reasons:
- less duplication / no copy paste necessary
- cosmos devs can still find docs where they would expect them (under x/modulename)
- my intuition is that devs will feel more responsibility over the specs/docs if they are in the same directory as the code (might be completely wrong about this too)

liamsi (approved):

cmwaters (changes requested):
> my initial naive thought was that we could just update the summary page to link to the actual pkg READMEs https://github.com/celestiaorg/celestia-app/blob/main/specs/src/SUMMARY.md?plain=1

I think we should just do this. Having directories with a file that just links to another file seems misleading. In the overview page where you lay out the contents of the spec, I would just do direct links to the readmes in the packages/modules we have. For the stuff that doesn't have a clear directory, like block validity rules, I would add those pages to the spec directory

rootulp (changes requested):
I think we should either:
Option A: keep https://github.com/celestiaorg/celestia-app/blob/main/specs/src/README.md up to date with https://github.com/celestiaorg/celestia-app/blob/main/specs/src/SUMMARY.md
Option B: delete https://github.com/celestiaorg/celestia-app/blob/main/specs/src/README.md

Note we need to keep SUMMARY.md b/c https://rust-lang.github.io/mdBook/guide/creating.html#summarymd

specs/src/README.md
```diff
@@ -8,3 +8,9 @@
   - [Public-Key Cryptography](./specs/public_key_cryptography.md)
```
rootulp, line 8: I think we should add Block Validity Rules above

```
  - [Block Validity Rules](./specs/block_validity_rules.md)
```

specs/src/specs/state_machine_modules.md
```diff
@@ -0,0 +1,30 @@
...
+
+## `celestia-app` Specific Modules
+
+- [blob](../../../x/blob/README.md)
+- [qgb](../../../x/qgb/README.md)
+- [mint](../../../x/mint/README.md)
+- [paramfilter](../../../x/paramfilter/README.md)
+- [upgrade](../../../x/upgrade/README.md)
+
+## Standard `cosmos-sdk` Modules
```
evan-forbes, line 13: note: added all the sdk modules we're using

evan-forbes:
@rootulp sorry, was just in the of refactoring when you reviewed. the readme should be up to date now

rootulp (approved):

evan-forbes:
@cmwaters @MSevey 
cool good idea, moved from having specific directories for state modules to just having links. I think the original thinking was that we would fill out more high level specs beyond the standard cosmos-sdk documentation, hence the specific directories for when we fill those out, but keeping those specs in their respective modules and linking in the readme as needed makes a lot of sense

rootulp (approved):

specs/src/specs/state_machine_modules.md
```diff
@@ -0,0 +1,30 @@
...
+
+- [auth](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/auth/spec/README.md)
+- [authz](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/authz/spec/README.md)
+- [bank](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/bank/spec/README.md)
+- [capability](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/capability/spec/README.md)
+- [crisis](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/crisis/spec/README.md)
+- [distribution](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/distribution/spec/README.md)
+- [evidence](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/evidence/spec/README.md)
+- [feegrant](https://github.com/celestiaorg/cosmos-sdk/blob/v1.14.0-sdk-v0.46.11/x/feegrant/spec/README.md)
+- [genutil](https://github.com/celestiaorg/cosmos-sdk/tree/v1.14.0-sdk-v0.46.11/x/genutil) (no spec)
```
cmwaters, line 23: Is this really a module?

cmwaters (approved):
I do echo Rootul's concern of having a duplicated README.md and a SUMMARY.md but other than that am happy with this initial direction

specs/src/specs/state_machine_modules.md
evan-forbes: I was tempted to not include it, but yeah it is and it can result in non-determinism for starting the genesis so I decided to include it.

https://github.com/celestiaorg/cosmos-sdk/blob/24903f4a983911f708fbf138a895ea4abc5dc752/x/genutil/module.go#L45-L53

evan-forbes:
> I do echo Rootul's concern of having a duplicated README.md and a SUMMARY.md but other than that am happy with this initial direction

I agree and we can definitely change this stucture in the future, it doesn't make any sense to have two. I think this is just an `mdbook` thing, which we could also change.

PR #1860, opened 2023-05-31
NOT READY FOR REVIEW

staheri14:
## Overview
Closes # 

## Checklist

### DIFF
--- a/pkg/square/square_test.go
+++ b/pkg/square/square_test.go
@@ -258,3 +258,26 @@ func TestSize(t *testing.T) {
 		assert.True(t, shares.IsPowerOfTwo(res))
 	}
 }
+
+// FuzzSquareDeconstruct test whether square.Deconstruct can correctly deconstruct a square.
+func FuzzSquareDeconstruct(f *testing.F) {
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	f.Fuzz(func(t *testing.T, normalTxCount int, pfbCount int, blobsPerPfb int, blobSize int) {
+		// skip negative values
+		if normalTxCount < 0 || pfbCount < 0 || blobsPerPfb < 0 || blobSize < 0 {
+			t.Skip()
+		}
+		// each PFB should have at least one non-zero size blob
+		if pfbCount > 0 && (blobSize == 0 || blobsPerPfb == 0) {
+			t.Skip()
+		}
+		txs := generateOrderedTxs(normalTxCount, pfbCount, blobsPerPfb, blobSize)
+		dataSquare, err := square.Construct(txs, appconsts.LatestVersion, appconsts.DefaultSquareSizeUpperBound)
+		if err != nil {
+			t.Skip()
+		}
+		recomputedTxs, err := square.Deconstruct(dataSquare, encCfg.TxConfig.TxDecoder())
+		require.NoError(t, err)
+		require.Equal(t, txs, recomputedTxs.ToSliceOfBytes())
+	})
+}
--- a/pkg/square/square_test.go
+++ b/pkg/square/square_test.go
@@ -259,9 +259,10 @@ func TestSize(t *testing.T) {
 	}
 }
 
-// FuzzSquareDeconstruct test whether square.Deconstruct can correctly deconstruct a square.
+// FuzzSquareDeconstruct tests whether square deconstruction function can correctly deconstruct a block back from a given square.
 func FuzzSquareDeconstruct(f *testing.F) {
 	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+
 	f.Fuzz(func(t *testing.T, normalTxCount int, pfbCount int, blobsPerPfb int, blobSize int) {
 		// skip negative values
 		if normalTxCount < 0 || pfbCount < 0 || blobsPerPfb < 0 || blobSize < 0 {
@@ -271,13 +272,16 @@ func FuzzSquareDeconstruct(f *testing.F) {
 		if pfbCount > 0 && (blobSize == 0 || blobsPerPfb == 0) {
 			t.Skip()
 		}
-		txs := generateOrderedTxs(normalTxCount, pfbCount, blobsPerPfb, blobSize)
-		dataSquare, err := square.Construct(txs, appconsts.LatestVersion, appconsts.DefaultSquareSizeUpperBound)
-		if err != nil {
-			t.Skip()
-		}
+
+		allTxs := generateOrderedTxs(normalTxCount, pfbCount, blobsPerPfb, blobSize)
+		// extract those transaction that fit into the block
+		_, blockTxs, err := square.Build(allTxs, appconsts.LatestVersion, appconsts.DefaultSquareSizeUpperBound)
+
+		// @TODO  may need to do a subset check for the blockTxs and the allTxs
+		dataSquare, err := square.Construct(blockTxs, appconsts.LatestVersion, appconsts.DefaultSquareSizeUpperBound)
+
 		recomputedTxs, err := square.Deconstruct(dataSquare, encCfg.TxConfig.TxDecoder())
 		require.NoError(t, err)
-		require.Equal(t, txs, recomputedTxs.ToSliceOfBytes())
+		require.Equal(t, blockTxs, recomputedTxs.ToSliceOfBytes())
 	})
 }


pkg/square/square_fuzz_test.go
```diff
@@ -70,3 +70,52 @@ func FuzzSquare(f *testing.F) {
...
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+
+	f.Add(0, 1)
+	f.Fuzz(func(t *testing.T, normalTxCount int, pfbCount int) {
+		// skip negative values
+		if normalTxCount < 0 || pfbCount < 0 {
+			t.Skip()
+		}
+		maxBlobSize := 1000 // @TODO there might be a global constant for this
+		allTxs := GenerateOrderedRandomTxs(encCfg.TxConfig, normalTxCount, pfbCount, maxBlobSize)
```
cmwaters, line 85: why not use `generateOrderedTxs`

pkg/square/square_fuzz_test.go
```diff
@@ -70,3 +70,52 @@ func FuzzSquare(f *testing.F) {
...
+		// construct the square
+		dataSquare, err := square.Construct(blockTxs, appconsts.LatestVersion, appconsts.DefaultSquareSizeUpperBound)
+		require.NoError(t, err)
+
+		require.Equal(t, builtSquare, dataSquare)
+
+		recomputedTxs, err := square.Deconstruct(dataSquare, encCfg.TxConfig.TxDecoder())
+		require.NoError(t, err)
+		require.Equal(t, len(blockTxs), len(recomputedTxs.ToSliceOfBytes()))
+		require.Equal(t, blockTxs, recomputedTxs.ToSliceOfBytes())
```
cmwaters, line 103: Couldn't we just add these 4 lines to `FuzzSquare`

cmwaters (commented):
There's a lot of existing test fixture code that we should try use instead of continually recreating the same functions.

staheri14:
> There's a lot of existing test fixture code that we should try use instead of continually recreating the same functions.

Agree with this, however, as far as I checked, the existing ones could not produce random blob count (for blob tx) and amount (for send tx). That is why I added the `GenerateOrderedRandomTxs`.

pkg/square/square_fuzz_test.go
staheri14: We could, As I pointed out earlier, the two tests currently generate transactions in different ways. However, I can modify the `FuzzSquare` to use the `GenerateOrderedRandomTxs` if you think it's a good idea. What are your thoughts on this? @cmwaters

pkg/square/square_fuzz_test.go
cmwaters: Well the main difference AFAICS between the two functions is that `generateOrderedTxs()` allows you to specify the blobsPerPFB as an argument while `GenerateOrderedRandomTxs()` generates a random number within the function. As I alluded to, the fuzzer should really be responsible for all randomness hence why I prefer `generateOrderedTxs()`

staheri14: > the fuzzer should really be responsible for all randomness 

I am onboarded and agree with the idea of delegating randomness generation to the fuzzer, yet, it is not the case even in the `generateOrderedTxs()` function. It introduces internal randomness when generating transactions, that is, with the same inputs it produces different outputs each time. Please see the following sample:
```go
func FuzzOrderedTxsRandmoness(f *testing.F) {
	var (
		normalTxCount uint = 12
		pfbCount      uint = 91
		blobsPerPfb   uint = 2
		blobSize      uint = 812
	)
	f.Add(normalTxCount, pfbCount, blobsPerPfb, blobSize)
	f.Fuzz(func(t *testing.T, normalTxCount, pfbCount, blobsPerPfb, blobSize uint) {
		set1 := generateOrderedTxs(int(normalTxCount), int(pfbCount), int(blobsPerPfb), int(blobSize))
		set2 := generateOrderedTxs(int(normalTxCount), int(pfbCount), int(blobsPerPfb), int(blobSize))
		require.Equal(t, set1, set2) // this fails, the transactions are not the same, regardless of the proportion of normal and pfb transactions i.e., normalTxCount and pfbCount
	})

}
```
There are many reasons why they are not the same including (but not limited to) the randomness produced internally in the [`RandBlobTxs`](https://github.com/celestiaorg/celestia-app/blob/53abe8824af0988d816bb241623b59a8d61387b3/pkg/square/builder_test.go#L63) which invokes [`RandMsgPayForBlobsWithSigner`](https://github.com/celestiaorg/celestia-app/blob/cb2d7ea4648d97ff32985cdd401a56ecda61de32/test/util/blobfactory/payforblob_factory.go#L238) which generates random blobs and namespaces.  

Considering the above observation (the internal randomness in the transaction generation), I incorporated additional randomness into the transaction generation process to provide further confidence in the absence of bugs in our code. This decision was based on the implicit assumption that our code is bug-free, but we wanted to validate this assumption. By generating more randomized data and testing our implementation against it, the aim is to obtain an additional guarantee of bug-free behavior. If our assumption holds true, the fuzz testing should not uncover any failing inputs. However, if we do encounter a failing input, we would inevitably need to rely on the logged error messages for debugging, considering the inherent randomness present in almost all of our test fixtures. 

Having said that, I am fine with merging the two tests. However, I want to highlight that the point you mentioned about there being "no outside randomness" would not be satisfied in either scenario, as far as I can tell. We may need actually do further refactoring to achieve this.  Please let me know your thoughts.

pkg/square/square_fuzz_test.go
staheri14: Thanks for the clarification 🙏 
> All randomness should be deterministic (there should be no outside randomness). If the fuzzer outputs some seed corpus that has an unexpected result, we should be able to run it over and over again and expect to see the same failure.

I covered this point in this [comment](https://github.com/celestiaorg/celestia-app/pull/1860#discussion_r1224638076).

staheri14: > What randomness is meaningful and not supposed to be covered elsewhere? For example changing the signer is meaningless given that square construction only cares about arrays of bytes and blobs

I agree with your point.  the `generateRandomSigner` function is not intended to augment the randomness of the final generated transactions. As I stated in my comment [here](https://github.com/celestiaorg/celestia-app/pull/1860#discussion_r1223413788), the only additional elements of randomness introduced in this PR are 1) the blob counts for each blob transaction and 2) the amount for the send transaction. However, since the inclusion of  `generateRandomSigner` appears to be causing confusion, I will proceed with the removal of it.

staheri14: Done in https://github.com/celestiaorg/celestia-app/pull/1860/commits/bf6b0bf5c312bc7826e4b5c45c1c12b1dcdba5ff

test/util/blobfactory/test_util.go
```diff
@@ -71,3 +72,30 @@ func CreateRawTx(txConfig client.TxConfig, msg sdk.Msg, signer *blobtypes.Keyrin
 
 	return rawTx
 }
+
+// GenerateRandomAmount generates a random amount for a Sand transaction.
```
rootulp, line 76:
```suggestion
// GenerateRandomAmount generates a random amount for a Send transaction.
```

rootulp (approved):

rootulp (dismissed):

test/util/blobfactory/payforblob_factory.go
```diff
@@ -23,6 +24,13 @@ import (
 
 var defaultSigner = testfactory.RandomAddress().String()
 
+const (
+	// TestMaxBlobSize is the maximum size of each blob in a blob transaction, for testing purposes
+	TestMaxBlobSize = 100
```
evan-forbes, line 29: [blocking]
I believe this is in bytes here

https://github.com/celestiaorg/celestia-app/blob/5787e52aed42663bf133a8649c0324b4eb45eeef/test/util/blobfactory/payforblob_factory.go#L305

and if that's the case, then should we have blobs of many different sizes?

evan-forbes (commented):
Callum raises good pionts, where ideally all randomness would be handled by go's fuzzing tools, then we could have an isolated test. As Sanaz has pointed out, since we already don't have that, I'm good with not for now though we might want to open an issue (it would be a good "good first issue"). I also think this is a meaningful addition to our share encoding testing and is worthwhile since we've caught weird edge cases in the past with similar tools having similar levels of randomness

left one blocking question, and would like to defer to @cmwaters for final approval

test/util/blobfactory/payforblob_factory.go
staheri14: Can you please elaborate on your question? 
> should we have blobs of many different sizes?

Currently, all the generated Blob transactions (according to the [RandMultiBlobTxs](https://github.com/celestiaorg/celestia-app/blob/1697b1a0e2c380a72e44678dbbf83865c89fee64/test/util/blobfactory/payforblob_factory.go#L526) function), have multiple blobs of different sizes each, are you suggesting we should not do that?

evan-forbes: yeah sure the blobs will be of different sizes, but they will all only use a single share since the max since is denominated in bytes afaict

staheri14: I see, good point, I now updated the max blob size in https://github.com/celestiaorg/celestia-app/pull/1860/commits/a3f07b75013b290f341184d2ffc1e37dae2050aa  so that a blob can roughly span two rows, at max.

staheri14: @evan-forbes Please let me know if the changes in https://github.com/celestiaorg/celestia-app/commit/a3f07b75013b290f341184d2ffc1e37dae2050aa addresses your concern.

evan-forbes (approved):

staheri14:
> Callum raises good pionts, where ideally all randomness would be handled by go's fuzzing tools, then we could have an isolated test. As Sanaz has pointed out, since we already don't have that, I'm good with not for now though we might want to open an issue (it would be a good "good first issue"). I also think this is a meaningful addition to our share encoding testing and is worthwhile since we've caught weird edge cases in the past with similar tools having similar levels of randomness
> 
> left one blocking question, and would like to defer to @cmwaters for final approval

Created a tracking issue, as you @evan-forbes  suggested https://github.com/celestiaorg/celestia-app/issues/1942

cmwaters (commented):
If we're going to add more randomness to the fuzzing than is part of the original corpus then can we add a int64 that we use as the seed for a `rand.Source` that is then passed to all the `GenerateRand...` functions such that we're capable of reproducing the exact same set of txs that we pass through the fuzzer.

I'm also not sure if we need to break this out into multiple tests if the aim is to simply run the fuzzer for `n` minutes every night.

staheri14:
Thank you, @cmwaters, for your insights on this matter. Based on our current discussion, I believe the conclusion for this PR can be summarized as follows:
1. Consolidating the square construct/deconstruct fuzzer (introduced in this PR) into the existing square fuzzer, as well as replacing the older transaction generation mechanism `GenerateOrderedTxs ` in the square fuzzer with a more randomized approach `GenerateOrderedRandomTxs`.
2. Ensuring that transactions are generated deterministically using the corpus generated by the fuzzer, thereby addressing this specific [issue](https://github.com/celestiaorg/celestia-app/issues/1942) (In line with this [comment](https://github.com/celestiaorg/celestia-app/pull/1860#pullrequestreview-1483090331)).

@cmwaters, please let me know if these two steps align with your expectations. If so, I will proceed to address the first step within this PR and create a separate PR for the second step.

rootulp (dismissed):

cmwaters:
Yup, exactly

test/util/blobfactory/payforblob_factory.go
```diff
@@ -23,6 +24,13 @@ import (
 
 var defaultSigner = testfactory.RandomAddress().String()
 
+var (
+	// TestMaxBlobSize is the maximum size of each blob in a blob transaction, for testing purposes
+	TestMaxBlobSize = appconsts.ShareSize * 4
```
staheri14, line 29: When setting sizes larger than this, specifically close to the square row size, the fuzzer exhibits non-deterministic halting behavior. I have also verified this issue by testing the original square fuzzer in the main branch, and encountered the same problem. I am currently investigating this issue and will submit another pull request as soon as I identify a fix.

staheri14: Tracking issue https://github.com/celestiaorg/celestia-app/issues/1956

evan-forbes: :eyes: :eyes: 

fwiw, we are testing with blobs larger than this in the integration and square size test, but this does seem like a very serious bug!

https://github.com/celestiaorg/celestia-app/blob/c1176b18ada45a2465cb48dc48eface32d301fe3/app/test/integration_test.go#L79-L107

staheri14: Thanks, @EvanForbes, for your input! It appears that the issue I encountered was specific to my local machine. I tried running it on another machine, and it worked perfectly fine. Therefore, I'm going to revert the max blob size back to the larger value, which is two rows of a square. https://github.com/celestiaorg/celestia-app/pull/1860/commits/087d8addcc1ad4edcebbcfbd8ff18a85270c6f99

rootulp (approved):

evan-forbes (approved):
LGTM

staheri14:
@cmwaters Please let me know whether the changes look good to you?

cmwaters (approved):
Thanks for that. This looks great!

Issue #1267, opened 2023-01-19
Document token filter middleware


evan-forbes:
I would also add here: documenting of IBC's token filter middleware

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/issues/1257#issuecomment-1397043029_

in #1219, we added a token filter to prevent tokens outside of the native one to be transfered via IBC, this change needs to be documented in:
- [ ] the specs
- [ ] developer focused documentation such as our doc site
- [ ] internally, such as readme for https://github.com/celestiaorg/celestia-app/tree/main/x/tokenfilter or a small ADR

cmwaters:
linking https://github.com/celestiaorg/docs/pull/518 as it is relevant

PR #1863, opened 2023-05-31
docs: specify the token filter middleware

cmwaters:
Closes: #1267

### DIFF
--- a/x/tokenfilter/README.md
+++ b/x/tokenfilter/README.md
@@ -1,0 +1,35 @@
+# IBC Token Filter
+
+## Abstract
+
+The IBC token filter prevents non-native tokens from being transferred through the IBC transfer module from a counterparty chain to the host chain. This is useful if a chain wishes to only permit
+their native token within their state machine.
+
+## Context
+
+When tokens are transferred through the IBC transfer module, the denomination of that token is modified at each hop. It's more accurate to think of the tokens transferring to another network to actually remain held in escrow at the origin chain and the destination chain minting a "wrapped" equivalent. Thus tokens moved to a non-native state machine are beholdent to the security of both chains. The denomination is denoted as the path it's taken from the origin chain. To provide an example, imagine 3 chains: A, B and C with native tokens "a", "b", "c". With IBC's transfer module, when a user transfers token "a" from A to B, the wrapped token is prefixed with the source port and source channel (i.e. `portidone/channel-0/a`), this is Ba. If the token is further transferred to C it becomes BCa (or `portidtwo/channel-1/portidone/channel-0/a`). This token is now beholdent to the security of B, C and A. Also note that this is different if "a" were to go directly to C. In other words: `Ca != BCa`.
+
+This context is important in recognising when a native token is returning to its origin state machine. Each IBC packet contains metadata including the source port and channel. Therefore if the denomination of the token is prefixed with the same source port and channel as detailed in the packet, we can conclude that the denomination originally came from the receiving chain.
+
+This logic is employed for path unwinding. Reversing the tokens through the path it came will eventually strip the denomination until it reaches the base denomination it started with.
+
+## Protocol
+
+The protocol targets inbound packets only. Outbound transfer messages are left unmodified. When a packet is sent to the IBC transfer module (denoted by the `transfer` port id), the token filter intercepts the message and attempts to unmarshal it as a `FungibleTokenPacketData`. If unmarshalling fails, the protocol should simply pass it down the stack for it to be handled elsewhere.
+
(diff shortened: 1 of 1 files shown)


x/tokenfilter/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+This context is important in recognising when a native token is returning to its origin state machine. Each IBC packet contains metadata including the source port and channel. Therefore if the denomination of the token is prefixed with the same source port and channel as detailed in the packet, we can conclude that the denomination originally came from the receiving chain.
```
rootulp, line 12: [question]

> is prefixed

Chain A port-id: `portidone`
Chain A channel: `channel-0`

In the example above, BCa is `portidtwo/channel-1/portidone/channel-0/a` so if it tries to return to Chain A, it is not prefixed with Chain A's port-id and channel because Chain A's port-id and channel come after Chain B. 

Does the port-id and channel need to be prefixed or suffixed?

x/tokenfilter/README.md
```diff
@@ -0,0 +1,35 @@
...
+their native token within their state machine.
+
+## Context
+
+When tokens are transferred through the IBC transfer module, the denomination of that token is modified at each hop. It's more accurate to think of the tokens transferring to another network to actually remain held in escrow at the origin chain and the destination chain minting a "wrapped" equivalent. Thus tokens moved to a non-native state machine are beholdent to the security of both chains. The denomination is denoted as the path it's taken from the origin chain. To provide an example, imagine 3 chains: A, B and C with native tokens "a", "b", "c". With IBC's transfer module, when a user transfers token "a" from A to B, the wrapped token is prefixed with the source port and source channel (i.e. `portidone/channel-0/a`), this is Ba. If the token is further transferred to C it becomes BCa (or `portidtwo/channel-1/portidone/channel-0/a`). This token is now beholdent to the security of B, C and A. Also note that this is different if "a" were to go directly to C. In other words: `Ca != BCa`.
```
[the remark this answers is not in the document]
cmwaters: I took it from here: https://github.com/cosmos/ibc-go/blob/ccfce10b62c8d2e7b88022bdd9a55c8185ac4108/modules/apps/transfer/types/trace.go#L24-L27

I'm not sure if thats the actual way its encoded. I'm happy to modify it

x/tokenfilter/README.md
rootulp: Sorry I think I'm taking this too literally but now I'm confused.

> Therefore if the denomination of the token is prefixed with the same source port and channel as detailed in the packet, we can conclude that the denomination originally came from the receiving chain.

Based on the example, I infer that each hop adds an additional prefix to the path. If that's the case, the prefix of the path doesn't indicate the originating chain, it indicates the most recent chain that the token was bridged to. 

What happens in this series of events:
1. Token a is bridged from chain A to B resulting in Ba on chain B
2. Token Ba is bridged from chain B to C resulting in CBa on chain C
3. Token CBa is bridged from chain C to A resulting in (ACBa or a)?

If we're trying to filter to just "when a native token is returning to its origin state machine", why does the criteria not enforce that there is only one port-id / channel and that port-id / channel is the originating chain? The prefix part of this spec makes me think the criteria only cares about the last bridged chain but I've heard IBC is path dependent so I don't think the criteria can only look at the last bridged chain.

evan-forbes (approved):
easy approve, dope!

x/tokenfilter/README.md
cmwaters: > the path doesn't indicate the originating chain, it indicates the most recent chain that the token was bridged to.

That it was bridged "from", yes that's true. You continuously prepend the metadata unless you notice that the the current prefix comes from your chain in which case you remove the most recent metadata (i.e. the prefix). This is where we get the winding and unwinding notion.

I have specified this further down but just for clarification. The token filter doesn't actually guarantee that only native tokens remain on the state machine, it guarantees that any token currently presiding in the state machine can always return. 

To illustrate this, if the cosmos hub were to perform an upgrade including this middleware. All existing IBC assets that had previously passed through the hub (i.e. had the hubs prefix) could unwind back through the cosmos hub to the original chain. This is necessary because if we only allowed the native asset back, then we could potentially sever assets from the original chain (most likely making them valueless).

Note with Celestia, because the genesis state only includes native tokens, we can conclude that only native tokens will ever exist in Celestia's state machine

> What happens in this series of events:

In this case, A would receive ACBa not a. To elaborate, if TIA were to go from Celestia to Osmosis, then to the Cosmos Hub, Celestia would reject a transaction trying to send the TIA from the Cosmos Hub to Celestia (you would have to go back the way you came)

rootulp: Thanks for explaining. I don't think any changes are necessary in the spec so resolving this thread.

PR #1867, opened 2023-06-01
chore: explicitly reject share commitments that are not == hash len

evan-forbes:
## Overview

Noticed this while I was writing the block validity rules. This was still failing as expected since the commitment is check, but now we're being more explicit.

## Checklist

### DIFF
--- a/pkg/appconsts/global_consts.go
+++ b/pkg/appconsts/global_consts.go
@@ -76,9 +76,17 @@ var (
 	// if another hash.Hash should be used as a base hasher in the NMT.
 	NewBaseHashFunc = consts.NewBaseHashFunc
 
+	// hashLength is the length of a hash in bytes.
+	hashLength = NewBaseHashFunc().Size()
+
 	// DefaultCodec is the default codec creator used for data erasure.
 	DefaultCodec = rsmt2d.NewLeoRSCodec
 
 	// SupportedShareVersions is a list of supported share versions.
 	SupportedShareVersions = []uint8{ShareVersionZero}
 )
+
+// HashLen returns the length of a hash in bytes.
+func HashLength() int {
+	return hashLength
+}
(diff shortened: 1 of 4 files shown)


pkg/appconsts/global_consts.go
```diff
@@ -76,9 +76,17 @@ var (
...
+	hashLength = NewBaseHashFunc().Size()
+
 	// DefaultCodec is the default codec creator used for data erasure.
 	DefaultCodec = rsmt2d.NewLeoRSCodec
 
 	// SupportedShareVersions is a list of supported share versions.
 	SupportedShareVersions = []uint8{ShareVersionZero}
 )
+
+// HashLen returns the length of a hash in bytes.
```
rootulp, line 89:
```suggestion
// HashLength returns the length of a hash in bytes.
```

x/blob/types/payforblob_test.go
```diff
@@ -147,6 +147,10 @@ func TestValidateBasic(t *testing.T) {
 	emptyShareCommitment := validMsgPayForBlobs(t)
 	emptyShareCommitment.ShareCommitments[0] = []byte{}
 
+	// MsgPayForBlobs that has an empty share commitment
```
rootulp, line 150:
```suggestion
	// MsgPayForBlobs that has an invalid share commitment size
```

pkg/appconsts/global_consts.go
evan-forbes: thaaaaannnkksss

rootulp (approved):

cmwaters (approved):
LGTM. I don't know what our stance is towards clearer error messages. I can imagine returning `ErrInvalidShareCommitment` in this scenario would be relatively confusing to users. It would be also helpful now we have multiple share commitments to return which one was invalid by wrapping the error

Issue #1870, opened 2023-06-02
AnnualProvisions may be updated more than once per year


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/348f795d8635324db4cca0367b599b5edc5c2fdb/x/mint/abci.go#L33-L43

## Problem

This code snippet uses `==` when it should use [`Dec.Equal`](https://github.com/celestiaorg/cosmos-sdk/blob/24903f4a983911f708fbf138a895ea4abc5dc752/types/decimal.go#L213). As a result, inflation rate and annual provisions may get updated more than once in the span of a year.

## Proposal

Use  [`Dec.Equal`](https://github.com/celestiaorg/cosmos-sdk/blob/24903f4a983911f708fbf138a895ea4abc5dc752/types/decimal.go#L213)

PR #1871, opened 2023-06-02
fix: update AnnualProvisions once per year

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1870

Note: this wasn't caught by the existing `TestAnnualProvisions` unit test because the test app used `stake` as its bond denomination but the mint module uses `utia` as the default bond denomination. So this PR also updates the test app to use `utia` as the default bond denomination. After this change, new tokens that get emitted due to inflation increase the total supply of `utia` which will impact the annual provisions calculation.


test/util/test_app.go
```diff
@@ -250,8 +250,10 @@ func genesisStateWithValSet(
 
 	}
 	// set validators and delegations
-	stakingGenesis := stakingtypes.NewGenesisState(stakingtypes.DefaultParams(), validators, delegations)
-	genesisState[stakingtypes.ModuleName] = app.AppCodec().MustMarshalJSON(stakingGenesis)
+	params := stakingtypes.DefaultParams()
+	params.BondDenom = app.BondDenom
```
rootulp, line 254: this line makes the test app's staking module use `utia` instead of `stake`

evan-forbes (approved):
really nice catch!

x/mint/abci.go
```diff
@@ -35,7 +35,9 @@ func maybeSetGenesisTime(ctx sdk.Context, k keeper.Keeper) {
 func maybeUpdateMinter(ctx sdk.Context, k keeper.Keeper) {
 	minter := k.GetMinter(ctx)
 	newInflationRate := minter.CalculateInflationRate(ctx)
-	if newInflationRate == minter.InflationRate {
+
+	isNonZeroAnnualProvisions := !minter.AnnualProvisions.IsZero()
+	if newInflationRate.Equal(minter.InflationRate) && isNonZeroAnnualProvisions {
 		// The minter's InflationRate AnnualProvisions already reflect the
 		// values for this year. Exit early because we don't need to update
 		// them.
```
evan-forbes, line 43: the good news is this appears to not break state or the result of the inflation, but previous to this change, this code was literally never getting hit, so we would always end up updating inflation. That might actually still change the apphash even if the state is not changing, depending on if this is causing an insertion to the iavl tree

evan-forbes:
@rootulp while we're fixing this, can we also only call `SetMinter` only once here? I don't think there is any benefit to calling this more than once, if anything only weirdness could occur

https://github.com/celestiaorg/celestia-app/blob/348f795d8635324db4cca0367b599b5edc5c2fdb/x/mint/abci.go#L44-L49

rootulp:
>  can we also only call `SetMinter` only once here?

Sure done in https://github.com/celestiaorg/celestia-app/pull/1871/commits/2d7dab5037f45c63e8095bcf9ec1c461eab9a794

evan-forbes (approved):

evan-forbes (approved):

SweeXordious (approved):

SweeXordious (approved):

PR #1878, opened 2023-06-05
chore(deps): bump core version to v1.21.2-tm-v0.34.27

SweeXordious:
## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -201,5 +201,5 @@ require (
 replace (
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.14.0-sdk-v0.46.11
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.21.1-tm-v0.34.27
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.21.2-tm-v0.34.27
 )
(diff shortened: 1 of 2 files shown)


rootulp:
[question] if https://github.com/celestiaorg/celestia-core/releases/tag/v1.21.2-tm-v0.34.27 only contains a feature, should it have been released as `v.1.22.0-tm-v0.34.27` instead?

rootulp (approved):

SweeXordious:
`v.1.22.0-tm-v0.34.27` according to semver, yes. But from what I noticed, the pattern we follow is something like:

- breaking change: we increment the minor version
- features/bug fixes:  we increment the bug version

While the major version stays the same

rootulp:
@cmwaters @evan-forbes what are your thoughts on https://github.com/celestiaorg/celestia-app/pull/1878#issuecomment-1576955459 ? If that's the case, I'd like to update https://github.com/celestiaorg/celestia-core#versioning b/c it's not obvious to me and I would've cut the wrong release in this instance.

evan-forbes:
I know our forks of the cosmos-sdk and tendermint don't follow semver, but we should try to as hard as possible

I think the confusion for me might have just been that https://github.com/celestiaorg/celestia-core/pull/1020 maybe should have just been a `refactor` or even `fix`, since afaiu we didn't break or add any apis we just fixed the implementation unecessarily needed the cached txs. Technically, we added an exported struct, but its only returned or accepted by unexpectored functions

rootulp:
> and we should try to as hard as possible to follow semver imo

Agreed so seems like no changes are needed to https://github.com/celestiaorg/celestia-core#versioning 

+1 to being careful with our PR conventional commit prefixes because:

> - Automatically determining a semantic version bump (based on the types of commits landed).
> - Communicating the nature of changes to teammates, the public, and other stakeholders.

[source](https://www.conventionalcommits.org/en/v1.0.0/#why-use-conventional-commits)

I could see how this particular scenario is a gray area on which prefix to use.

SweeXordious:
> but we should try to be correct with how we prefix our PRs to make this process easier

My bad, I used `feat` because I was unsure whether it's just a refactor or a feature. Because, it is a feature from the perspective that it changes a requirement for validators to run orchestrators. However, it doesn't make any change to the API or anything.

PR #1879, opened 2023-06-05
chore(deps): downgrade the version of docker

SweeXordious:
## Overview

We're setting this exact version so that we can update the orchestrator-relayer repository up to latest app main.

The reason for this is that orchestrator-relayer depends on testcontainers which only supports this version. And if we use a higher version here, the E2E tests break and we will be unable to bump the app version

## Checklist


evan-forbes (approved):
thanks, makes sense to me

SweeXordious:
Closing as no longer needed

PR #1893, opened 2023-06-08
feat: introduces NamespaceVersionMaxValue const

staheri14:
## Overview
Although we already have a constant `NamespaceVersionSize`, the comparison of the max value of the namespace version of the incoming Blobs is currently hardcoded to `math.uint8`. While this aligns with the current value of `NamespaceVersionSize` which is 1 byte, it would be safer to derive the maximum value from the original constant instead of manually hardcoding it. This  PR incorporates this modification.

## Checklist

- [ ] Visual proof for any user-facing features like CLI or documentation updates

### DIFF
--- a/pkg/appconsts/global_consts.go
+++ b/pkg/appconsts/global_consts.go
@@ -3,6 +3,7 @@ package appconsts
 import (
 	"github.com/celestiaorg/rsmt2d"
 	"github.com/tendermint/tendermint/pkg/consts"
+	"math"
 )
 
 // These constants were originally sourced from:
@@ -11,7 +12,8 @@ import (
 // They can not change throughout the lifetime of a network.
 const (
 	// NamespaveVersionSize is the size of a namespace version in bytes.
-	NamespaceVersionSize = 1
+	NamespaceVersionSize     = 1
+	NamespaceVersionMaxValue = math.MaxUint8 // this value should be set according to NamespaceVersionSize, i.e., 2^NamespaceVersionSize - 1
 
 	// NamespaceIDSize is the size of a namespace ID in bytes.
 	NamespaceIDSize = 28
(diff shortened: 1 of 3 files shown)


pkg/appconsts/global_consts.go
```diff
@@ -11,7 +13,8 @@ import (
 // They can not change throughout the lifetime of a network.
 const (
 	// NamespaveVersionSize is the size of a namespace version in bytes.
-	NamespaceVersionSize = 1
+	NamespaceVersionSize     = 1
+	NamespaceVersionMaxValue = math.MaxUint8 // this value should be set according to NamespaceVersionSize, i.e., 2^NamespaceVersionSize - 1
```
rootulp, line 17: I think there's a slight typo in `2^NamespaceVersionSize - 1` b/c it doesn't evaluate to `math.MaxUint8`. When `NamespaceVersionSize` is interpolated, this evaluates to `2 ^ 1 - 1`. Maybe this meant to say something like `... i.e., // math.Pow(2, 8 * NamespaceVersionSize) - 1` but proposal to do this instead:

```go
```

rootulp (commented):
PR LGTM after a small change

pkg/appconsts/global_consts.go
staheri14: You are right, when writing that formula, I was thinking of `NamespaceVersionSize` in bits, instead of bytes. The correction is as you suggested `2^(8* NamespaceVersionSize)-1)`. Going to fix it.

staheri14: Done in https://github.com/celestiaorg/celestia-app/pull/1893/commits/56a8fd21253afa4f986c5fb1d83ab58680a75b82.

rootulp (approved):

rootulp:
[nit] you may change the conventional commit prefix of this PR to `refactor: `

Issue #1894, opened 2023-06-08
Remove `useShareIndexes`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/cd21a8b5b8ed0f554ad6ac5f89098e41bbd80413/pkg/shares/share_splitting.go#L75

## Problem

We closed https://github.com/celestiaorg/celestia-app/issues/1110 but the flag `useShareIndexes` still exists

## Proposal

Deprecate and remove `useShareIndexes`

Issue #1321, opened 2023-01-31
Document all exported functions


evan-forbes:
some functions that we are exporting still do not have simple godocs, which needs to be fixed.

one such example in this https://github.com/celestiaorg/celestia-app/pull/1039#discussion_r1031803820_

also related to #963

PR #1895, opened 2023-06-08
feat!: remove `useShareIndexes`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1894

This is an API breaking change b/c it changes the function signature of `SplitBlobs`. celestia-node can adopt this change [here](https://github.com/celestiaorg/celestia-node/blob/0823f6b644eb6ba176ef5ae3c008c65018ed5f46/blob/helper.go#L77)

```diff
-	rawShares, err := shares.SplitBlobs(0, nil, b, false)
+	rawShares, err := shares.SplitBlobs(b)
```


pkg/shares/share_splitting.go
```diff
@@ -72,21 +72,10 @@ func SplitTxs(txs coretypes.Txs) (txShares []Share, pfbShares []Share, shareRang
 	return txShares, pfbShares, mergeMaps(txMap, pfbMap), nil
 }
 
-func SplitBlobs(cursor int, indexes []uint32, blobs []coretypes.Blob, useShareIndexes bool) ([]Share, error) {
-	if useShareIndexes && len(indexes) != len(blobs) {
-		return nil, ErrIncorrectNumberOfIndexes
-	}
+func SplitBlob(blob coretypes.Blob) ([]Share, error) {
```
cmwaters, line 75: Would appreciate if we could start to add go docs for our public functions. Maybe we can go do that in a separate PR

cmwaters (dismissed):
Nice clean up

pkg/shares/share_splitting.go
rootulp: +1 we have https://github.com/celestiaorg/celestia-app/issues/1321 to track existing functions but I'll add godoc to this function b/c added in this PR

pkg/shares/share_splitting.go
```diff
@@ -72,21 +72,22 @@ func SplitTxs(txs coretypes.Txs) (txShares []Share, pfbShares []Share, shareRang
...
+	writer := NewSparseShareSplitter()
+	if err := writer.Write(blob); err != nil {
+		return nil, err
 	}
+	return writer.Export(), nil
+}
+
+// SplitBlobs splits the provided blobs into shares.
+func SplitBlobs(blobs []coretypes.Blob) ([]Share, error) {
```
Wondertan, line 85: What about unifying them into a single `SplitBlobs` with variadic param? 
Looks like a perfect match.

pkg/shares/share_splitting.go
```diff
@@ -72,21 +72,22 @@ func SplitTxs(txs coretypes.Txs) (txShares []Share, pfbShares []Share, shareRang
...
+// SplitBlob splits the provided blob into shares.
+func SplitBlob(blob coretypes.Blob) ([]Share, error) {
+	writer := NewSparseShareSplitter()
+	if err := writer.Write(blob); err != nil {
+		return nil, err
 	}
+	return writer.Export(), nil
+}
+
+// SplitBlobs splits the provided blobs into shares.
```
cmwaters, line 84:
```suggestion
// SplitBlobs splits the provided blobs into shares belonging to the same sequence. All blobs should thus belong to the same namespace
```
I think it's important to specify that they should be of the same namespace. Does the `SparseShareSplitter` already panic if shares are from different namespaces?

rootulp: `SplitBlobs` can be called with blobs from different namespaces. SparseShareSplitter supports writing blobs from different namespaces.

Going to revert this comment.

pkg/shares/share_splitting.go
rootulp: Can do! Thanks for the suggestion.

pkg/shares/share_splitting.go
rootulp: Resolved by 245d2431d57eb7ff310ff3814d9fc41cfc9055fc

pkg/shares/share_splitting.go
rootulp: Resolved by [10f5f55](https://github.com/celestiaorg/celestia-app/pull/1895/commits/10f5f552855e7cb8888776fb6a34b3fbab25513d)

pkg/shares/share_splitting.go
```diff
@@ -72,21 +72,10 @@ func SplitTxs(txs coretypes.Txs) (txShares []Share, pfbShares []Share, shareRang
 	return txShares, pfbShares, mergeMaps(txMap, pfbMap), nil
 }
 
-func SplitBlobs(cursor int, indexes []uint32, blobs []coretypes.Blob, useShareIndexes bool) ([]Share, error) {
```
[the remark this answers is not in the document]
rootulp: Resolved by [10f5f55](https://github.com/celestiaorg/celestia-app/pull/1895/commits/10f5f552855e7cb8888776fb6a34b3fbab25513d)

cmwaters (approved):
🧹

rootulp:
https://github.com/celestiaorg/celestia-app/actions/runs/5260800176/jobs/9508250062?pr=1895#step:6:548 fails because https://github.com/celestiaorg/devops/issues/326

evan-forbes (approved):
nice, very clean

Wondertan:
Is this PR supposed to be in the following rc?

rootulp:
IMO it isn't critical to get into 1.0.0 so it can go in next 1.0.0-rc (if there is one) or in a subsequent release.

cmwaters:
I think this should just be part of v1.1

cmwaters (approved):

evan-forbes (approved):

PR #1902, opened 2023-06-10
fix: use the right key for `EarliestAvailableAttestationNonce` check method

SweeXordious:
## Overview

We were using the wrong key when checking if the `EarliestAvailableAttestationNonce` exists in store

## Checklist


cmwaters (approved):
Nice find 👍 . I assume we will need to backport this to v1.x right? Do we have tests for this btw?

SweeXordious:
Yes will need to be backported to `v1.x`. I will  open a PR once this one gets merged.

For the test, it needs to be done for other check methods as well before genesis block state is applied. I opened an issue to work on it separately https://github.com/celestiaorg/celestia-app/issues/1908

rootulp (approved):

evan-forbes:
is there anyway this could be consensus breaking?

SweeXordious:
yes, it is consensus breaking, but I believe everything is setup correctly, only the check was missing. So, it shouldn't halt  the network

PR #1903, opened 2023-06-11
chore!: bump sdk v0.46.13 and core v0.34.28

evan-forbes:
## Overview

bumps to our forks of the cosmos-sdk and comet to their latest versions. This is consensus breaking because an event was modified in the sdk, although that event may or may not have gotten hit on mocha-2 already.

aiming towards v1.x atm for max :shipit:, but we need this on main ofc

blocked by https://github.com/celestiaorg/cosmos-sdk/pull/326, https://github.com/celestiaorg/cosmos-sdk/pull/324, https://github.com/celestiaorg/celestia-core/pull/1021


app/prepare_proposal.go
```diff
@@ -14,11 +14,9 @@ import (
...
 // tendermint via the BlockData.
 func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePrepareProposal {
-	sdkCtx, err := app.NewProcessProposalQueryContext()
-	if err != nil {
-		panic(err)
-	}
-
+	// create a context using a branch of the state and loaded using the
+	// proposal height and chain-id
+	sdkCtx := app.NewProposalContext(core.Header{ChainID: app.GetChainID(), Height: app.LastBlockHeight() + 1})
```
evan-forbes, line 19: for posteriety, if we don't do this, then the signatures of some txs will get rejected because the chain-id and height is not consistently loaded into the context, resulting in `PrepareProposal` filtering out those transactions, and therefore those txs get caught in the mempool, which always produces confusing errors.

cmwaters:
> aiming towards v1.x atm for max :shipit:, but we need this on main ofc

Why not main first and then backport to v1.x?

evan-forbes:
ok, we can go through main closing in favor of https://github.com/celestiaorg/celestia-app/pull/1912

Issue #1872, opened 2023-06-03
specs: x/qgb


SweeXordious:
Write specs for `x/qgb` module

PR #1911, opened 2023-06-12
specs: add QGB specs

SweeXordious:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/1872

## Checklist


x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+    bytes32 dataRoot;
+    // Celestia block original square size.
+    uint256 squareSize;
 }
 ```
 
-I.e. the current block height is not 0, and we're at a data commitment window height.
+1. `height`: the height of the block.
+2. `dataRoot`: the data root, aka data hash, of the block.
+3. `squareSize`: the [square](https://celestiaorg.github.io/celestia-app/specs/data_structures.html#arranging-available-data-into-shares) size of the block.
```
cmwaters, line 131: Why is the square size important?

x/qgb/README.md
```diff
@@ -110,56 +380,32 @@ if err != nil {
...
-- `blob`: Takes a transaction hash, in hex format, and verifies that the blob paid for by the transaction has been committed to by the QGB contract. It only supports one blob for now.
-- `shares`: Takes a range of shares and a height, and verifies that these shares have been committed to by the QGB contract.
-- `tx`: Takes a transaction hash, in hex format, and verifies that it has been committed to by the QGB contract.
-
-## Useful links
+## Appendix
 
 The smart contract implementation is in [quantum-gravity-bridge](https://github.com/celestiaorg/quantum-gravity-bridge/).
 
 For QGB v2, the orchestrator and relayer implementations will be in a separate repository. Links will be added subsequently.
```
cmwaters, line 407: Do we have links for these?

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
 # `x/qgb`
 
-## Abstract
+## Concepts
 
 This module contains the [Quantum Gravity Bridge](https://blog.celestia.org/celestiums/) (QGB) state machine implementation.
 
-## State machine
+The QGB state machine is responsible for creating attestations which are signed by [orchestrators](https://github.com/celestiaorg/orchestrator-relayer/blob/main/docs/orchestrator.md), and submitted to EVM chains by [relayers](https://github.com/celestiaorg/orchestrator-relayer/blob/main/docs/relayer.md).
```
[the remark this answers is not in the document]
SweeXordious: > Does the orchestrator actually sign the data commitments or the validator?

The validator runs the orchestrator software

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+[Attestations](https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/types/attestation.go#L10-L18) are requests for signatures generated by the state machine during the `EndBlock` phase. They facilitate alignment between orchestrators and relayers on which data to sign, and when to relay it.
 
-### Conditions of generating a new attestation
+There are two types of attestations, [valsets](https://github.com/celestiaorg/celestia-app/blob/376a1d4c0f321f12ba78279d2bd34fc6cb5e6dc2/proto/celestia/qgb/v1/types.proto#L18-L33) and [data commitments](https://github.com/celestiaorg/celestia-app/blob/376a1d4c0f321f12ba78279d2bd34fc6cb5e6dc2/proto/celestia/qgb/v1/types.proto#L35-L55).
 
-#### New valset creation
+### Valsets
 
-A new valset is created in the following situations:
+A valset is an attestation type representing a validator set change. It Allows keeping track of all the previous validator set changes in state, and helps specify which orchestrator should sign which attestation.
```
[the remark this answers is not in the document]
SweeXordious: but this part is only for definitions. Do you mean we should add a link for that section in the md file?

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
 
-```golang
-if (latestValset == nil) || (lastUnbondingHeight == uint64(ctx.BlockHeight())) || significantPowerDiff {
-    ...
+In fact, when an orchestrator sees a newly generated valset, it queries the previous valset and checks whether it's part of its validator set. Then, the orchestrator signs the new valset and submits that signature to the [QGB P2P network](https://github.com/celestiaorg/orchestrator-relayer/pull/66). Otherwise, it ignores it and waits for new attestations.
```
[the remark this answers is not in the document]
SweeXordious: It could, but we don't do that at the moment. Every nonce is processed separately without any order.
We could sign nonce 10 then 2 then 11...

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+
+A new valset is generated by the valset [`handler`](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/abci.go#L84-L135) in the following cases:
+
+#### No valset in store
+
+When `EndBlock` is executed straight after genesis, the store doesn't have any valset created yet. So, it [checks](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/abci.go#L85-L93) wether there is any valset in store, and [generates](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/abci.go#L123-L134) a new one afterwards representing the initial validator set.
+
+#### Validator starts unbonding
+
+When a validator starts unbonding, the `LatestUnbondingBlockHeight` gets updated in a hook with the current block number. Then, inside `EndBlock`, we [check](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/abci.go#L123) whether the `LatestUnbondingBlockHeight` corresponds to the current block height. If so, then we generate a new valset that doesn't contain that unbonding validator.
```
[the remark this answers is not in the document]
SweeXordious: We could do that, but we don't want to expect from a validator to sign if he left the validator set. Also, when we will want to implement slashing, it wouldn't make sense to slash for liveness for a validator that started  unbonding

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+
+#### Significant power change
+
+The third scenario where a valset gets created is when there is a significant power change. As stated above, valsets contain an `evmAddress -> power` mapping for the validator sets they represent. When a [significant power change](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/abci.go#L99-L120) happens, a new valset gets created. The significant power threshold is defined by the [`SignificantPowerDifferenceThreshold`](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/abci.go#L15-L18) constant.
+
+A significant power change can happen if a validator's delegation got reduced or increased significantly, or the powers of multiple validators changed in a way that the whole validator set variation is higher than the threshold. This calculus is done inside the [`PowerDiff(...)`](https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/x/qgb/types/validator.go#L100-L140) method.
```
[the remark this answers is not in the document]
SweeXordious: No upper bound. And no rationale, inherited from the gravity bridge

x/qgb/README.md
SweeXordious: good catch :+1:

x/qgb/README.md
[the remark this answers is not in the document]
cmwaters: Even with this logic, we'd still expect situations where we're asking a validator to sign an attestation even although they are not part of the validator set. Imagine if they didn't unbond or anything but we're simply kicked out because another validator delegated enough to get itself in the set and kick out the lowest staked validator

x/qgb/README.md
SweeXordious: I didn't want to add it because there isn't any write-up for it so far.  I will open an issue to add it

SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1918

x/qgb/README.md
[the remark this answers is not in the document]
SweeXordious: Got it, but let's imagine the case where one validator which has 15% of the staking power leaves the validator set, then another validator joins with 15%. Then, if we don't have the unbonding hook, no change will happen to the valset because there was no power change

x/qgb/README.md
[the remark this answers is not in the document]
cmwaters: Ok, that's a bit confusing

x/qgb/README.md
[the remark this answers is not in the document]
cmwaters: Ah measurement of the power change is net? That doesn't sound too safe/correct

x/qgb/README.md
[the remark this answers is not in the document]
SweeXordious: We still can rename tho

x/qgb/README.md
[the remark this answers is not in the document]
cmwaters: I guess 5% errs on the side of caution but the upper bound might be something like 33% or 1/3. I haven't spent too much time thinking about the attack vectors here but it might be good to eventually write them up as a form of rationale

x/qgb/README.md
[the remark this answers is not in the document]
SweeXordious: It wouldn't need to be caught because its just a redelegation. 
We're even normalizing the powers to avoid creating valsets in the case of inflation or...

x/qgb/README.md
```diff
@@ -110,56 +380,32 @@ if err != nil {
...
-- `shares`: Takes a range of shares and a height, and verifies that these shares have been committed to by the QGB contract.
-- `tx`: Takes a transaction hash, in hex format, and verifies that it has been committed to by the QGB contract.
-
-## Useful links
+## Appendix
 
 The smart contract implementation is in [quantum-gravity-bridge](https://github.com/celestiaorg/quantum-gravity-bridge/).
 
-For QGB v2, the orchestrator and relayer implementations will be in a separate repository. Links will be added subsequently.
+For QGB v2, the orchestrator and relayer implementations is in the [orchestratory-relayer](https://github.com/celestiaorg/orchestrator-relayer) repo.
```
cmwaters, line 407:
```suggestion
For QGB v2, the orchestrator and relayer implementations are in the [orchestratory-relayer](https://github.com/celestiaorg/orchestrator-relayer) repo.
```

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+
+If the all the attestations in store are expired, which is an edge case that should never occur, the QGB state machine [doesn't prune](https://github.com/celestiaorg/celestia-app/blob/0629c757ef35a24187a8d7a4c706c7cdc894c8b6/x/qgb/abci.go#L161) the latest attestation.
+
+### Hooks
+
+#### Validator unbonding hook
+
+When a validator starts unbonding, a [hook](https://github.com/celestiaorg/celestia-app/blob/0629c757ef35a24187a8d7a4c706c7cdc894c8b6/x/qgb/keeper/hooks.go#L23-L34) is executed that [sets](https://github.com/celestiaorg/celestia-app/blob/0629c757ef35a24187a8d7a4c706c7cdc894c8b6/x/qgb/keeper/hooks.go#LL33C2-L33C2) the `LatestUnBondingBlockHeight` to the current block height. This allows creating a new valset that removes that validator from the valset members so that he doesn't need to sign attestations afterwards.
```
cmwaters, line 229: Is there a threshold for how much they should unbond before this gets triggered?

cmwaters (dismissed):
Documentation looks overall clear and helpful. I still have a few questions about the protocol itself but perhaps that can be addressed when we start the audit

x/qgb/README.md
SweeXordious: There is no threshold, the time a validator starts unbonding, this gets triggered

x/qgb/README.md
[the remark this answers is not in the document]
SweeXordious: I will open an issue to think on an upper bound for this value, thanks for pointing this out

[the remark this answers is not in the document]
SweeXordious: https://github.com/celestiaorg/celestia-app/issues/1920

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+                              
+Verifies that a transaction hash, a range of shares, or a blob referenced by its transaction hash were committed to by the QGB contract
+
+Usage:
+  celestia-appd verify [command]
+
+Available Commands:
+  blob        Verifies that a blob, referenced by its transaction hash, in hex format, has been committed to by the QGB contract.
+  shares      Verifies that a range of shares has been committed to by the QGB contract
+  tx          Verifies that a transaction hash, in hex format, has been committed to by the QGB contract
```
evan-forbes, line 285: [discusison around planning]
certainly not something we do in this PR, but as we've discussed in the past, we will eventually need to nail down the UX. All the tool that are needed are in these commands, but devs don't really have easy access to them. 

If they're looking at these specs, I would feel a bit confused if this were the only api, since these just logs a yes or no.

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+      --node string        <host>:<port> to Tendermint RPC interface for this chain (default "tcp://localhost:26657")
+  -o, --output string      Output format (text|json) (default "text")
+
+Global Flags:
+      --chain-id string      The network chain ID
+      --home string          directory for config and data (default "/home/midnight/.celestia-app")
+      --log-to-file string   Write logs directly to a file. If empty, logs are written to stderr
+      --log_format string    The logging format (json|plain) (default "plain")
+      --log_level string     The logging level (trace|debug|info|warn|error|fatal|panic) (default "info")
+      --trace                print out full stack trace on errors
```
evan-forbes, line 267: [not blocking suggestion]

imo we should be diligent in removing things that do not directly contribute to this module, we might even want to remove the other flags/aliases section.
```suggestion
```

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+After creating a new attestation, and adding it to the QGB store, an event is [emitted](https://github.com/celestiaorg/celestia-app/blob/0629c757ef35a24187a8d7a4c706c7cdc894c8b6/x/qgb/keeper/keeper_attestation.go#L16-L22) containing its nonce.
+
+## Client
+
+### Query attestation command
+
+The QGB query attestation command is part of the `celestia-appd` binary. It allows the user to query specific attestations by their corresponding nonce.
+
+```shell
+$ celestia-appd query qgb attestation --help                                                     INT ✘
```
evan-forbes, line 244: >  INT ✘ 

what's this?

evan-forbes, line 244: as also discussed with the comment on flags, I think it would be cleaner to just list the commands and a short description perhaps, and users can call `--help` if they want to view other flags. This also allows us to avoid manually updating this when those things change

x/qgb/README.md
```diff
@@ -110,56 +380,32 @@ if err != nil {
...
-## Useful links
+## Appendix
 
 The smart contract implementation is in [quantum-gravity-bridge](https://github.com/celestiaorg/quantum-gravity-bridge/).
 
-For QGB v2, the orchestrator and relayer implementations will be in a separate repository. Links will be added subsequently.
+For QGB v2, the orchestrator and relayer implementations are in the [orchestratory-relayer](https://github.com/celestiaorg/orchestrator-relayer) repo.
 
-QGB v1 implementation, including the orchestrator and relayer, is in the [qgb-integration](https://github.com/celestiaorg/celestia-app/tree/qgb-integration) branch.
+QGB v1 implementation, including the orchestrator and relayer, is in the [qgb-integration](https://github.com/celestiaorg/celestia-app/tree/qgb-integration) branch (deprecated).
```
evan-forbes, line 409: do we provide other information on versions elsewhere? afaik, I think we're the only ones who know what v1 or v2 are :sweat_smile: . If we don't distinguish between v1 and v2 in the code, we might want to only reference off hand here.

if we want to document the iterations we've done, then we might want to just do that in a short paragraph here, or in a separate document with more details. or maybe reference old ADRS or something not sure

evan-forbes (commented):
nice update! I think this will pair well with a higher level (but still very technical) document about trust assumptions, broader scope/connection contracts, etc whenever we get around to it.

no blocking comments, just some questions. I'll take a second review after chewing on it as well

x/qgb/README.md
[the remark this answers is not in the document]
cmwaters: Ok so just to confirm power change isn't just the total power now compared to the total power before? It actually looks to see what validators had what power?

x/qgb/README.md
cmwaters: So if a validator unbonds a single TIA token this will trigger? (Even though there would be no noticeable change to the validator set)

SweeXordious: In this case yes, because now he has lower power in the network, and it might make a difference when calculating the 2/3 relaying threshold

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
 
-A new data commitment is created in the following situation:
+The [QGB bridge power](https://github.com/celestiaorg/celestia-app/blob/6243f26fc419c32940d5dc4eb60b0e0aaf08eaa7/proto/celestia/qgb/v1/types.proto#L12-L13) is obtained by [normalizing](https://github.com/celestiaorg/celestia-app/blob/6243f26fc419c32940d5dc4eb60b0e0aaf08eaa7/x/qgb/keeper/keeper_valset.go#L125-L150) the validators' voting power using the [min-max normalization](https://en.wikipedia.org/wiki/Feature_scaling) formula. This formula takes into account the ratio of each validator's voting power to the total voting power in the block and scales it to a value between `0` and `2^32`.
 
-```golang
-if ctx.BlockHeight() != 0 && ctx.BlockHeight()%int64(k.GetDataCommitmentWindowParam(ctx)) == 0 {
-	...
+By normalizing the voting power, we can significantly reduce the frequency of generating new validator set updates. For example, if there is a small increase in the total on-chain voting power due to inflation, there is no need to create a new validator set. This is because the relative proportions of the validators remain the same, and the normalized QGB power doesn't show any significant difference.
```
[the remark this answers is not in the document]
SweeXordious: The percentage change reduces the frequency in case of redelegations for example.
The normalization reduces the frequency in case of, for example, inflation:
```
if there is a small increase in the total on-chain voting power due to inflation, there is no need to create a new validator set. This is because the relative proportions of the validators remain the same, and the normalized QGB power doesn't show any significant difference.
```

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+
+To ensure that the normalization process doesn't encounter overflow errors, the function normalizeValidatorPower uses [`BigInt`](https://github.com/celestiaorg/celestia-app/blob/6243f26fc419c32940d5dc4eb60b0e0aaf08eaa7/x/qgb/keeper/keeper_valset.go#LL142C1-L142C1) operations. It scales the raw power value with respect to the total validator power, making sure the result falls within the range of 0 to `2^32`.
+
+#### Power diff
+
+[`PowerDiff(...)`](https://github.com/celestiaorg/celestia-app/blob/6243f26fc419c32940d5dc4eb60b0e0aaf08eaa7/x/qgb/types/validator.go#L100-L141) is a function that calculates the difference in power between two sets of bridge validators. It's important to note that the power being compared is not the regular voting power in the Celestia consensus network, but a specific type called QGB bridge power (explained above).
+
+### Data commitments
+
+A data commitment is an attestation type representing a request to commit over a set of blocks. It provides an end exclusive range of blocks for orchestrators to sign over and propagate in the QGB P2P network.
```
SweeXordious, line 73:
```suggestion
A data commitment is an attestation type representing a request to commit over a set of blocks. It provides an end exclusive range of blocks for orchestrators to sign over and propagate in the QGB P2P network. The range is defined by the param [`DataCommitmentWindow`](https://github.com/celestiaorg/celestia-app/blob/fc83b04c3a5638ac8d415770e38a4046b84fa128/x/qgb/keeper/keeper_data_commitment.go#L44-L50), more on this below.
```

x/qgb/README.md
SweeXordious: good catch [c7e4a14](https://github.com/celestiaorg/celestia-app/pull/1911/commits/c7e4a144724941f118b321fce9053ab11bc17aa7)

x/qgb/README.md
```diff
@@ -1,50 +1,320 @@
...
+
+When an orchestrator sees a newly generated data commitment, it queries the previous valset and checks whether it's part of its validator set. Then, the orchestrator signs the new data commitment and submits that signature to the [QGB P2P network](https://github.com/celestiaorg/orchestrator-relayer/pull/66). Otherwise, it ignores it and waits for new attestations.
+
+A data commitment is comprised of the following fields:
```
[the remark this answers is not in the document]
SweeXordious: https://github.com/celestiaorg/celestia-app/pull/1911#discussion_r1231098193

x/qgb/README.md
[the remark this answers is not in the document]
cmwaters: So is the unbonding hook still necessary?

x/qgb/README.md
cmwaters: But why is a validator unbonding a single TIA token different to a delegator unbonding a single TIA token from the same validator. They both have the same voting power difference. I don't see we can't just use the significant voting power threshold as the sole trigger for a valset attestation.

Couldn't this also be some form of spam. i.e. every height I unbond 1 TIA, meaning that an attestation is made every height.

x/qgb/README.md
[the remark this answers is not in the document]
SweeXordious: yes, because if that sum is not significant, no change to the valset will happen, even if the validator unbonded. Meaning we will still keep waiting for them to sign or think we can slash them

x/qgb/README.md
```diff
@@ -1,50 +1,298 @@
...
+### Attestations types
 
-During their creation, the state machine might panic due to an unexpected behavior or event. These panics will be discussed below.
+[Attestations](https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/types/attestation.go#L10-L18) are requests for signatures generated by the state machine during the `EndBlock` phase. They facilitate alignment between orchestrators and relayers on which data to sign, and when to relay it.
 
-### Conditions of generating a new attestation
+There are two types of attestations, [valsets](https://github.com/celestiaorg/celestia-app/blob/376a1d4c0f321f12ba78279d2bd34fc6cb5e6dc2/proto/celestia/qgb/v1/types.proto#L18-L33) and [data commitments](https://github.com/celestiaorg/celestia-app/blob/376a1d4c0f321f12ba78279d2bd34fc6cb5e6dc2/proto/celestia/qgb/v1/types.proto#L35-L55).
 
-#### New valset creation
+### Valsets
```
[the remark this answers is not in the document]
SweeXordious: yes

x/qgb/README.md
SweeXordious: > I don't see we can't just use the significant voting power threshold as the sole trigger for a valset attestation.

We could but we won't be catching the case where a validator unbonds all his tokens, which are `4.9999999` of the power, and he would be able to double sign afterward without us slashing him.

> Couldn't this also be some form of spam. i.e. every height I unbond 1 TIA, meaning that an attestation is made every height.

They would also be paying the transaction fees. And we have pruning, so it would be around 500mb if they create a valset every block for 3 weeks for 15sec block time. 500mb of state that is not really manipulated inside the state machine that much is okey, I guess

cmwaters: > We could but we won't be catching the case where a validator unbonds all his tokens, which are 4.9999999 of the power, and he would be able to double sign afterward without us slashing him.

Don't you require 1/3 to double sign in order to genuinely equivocate (I'm just assuming light client verification security. I actually don't know the specifics)

Also it will take 21 days for the validator to truly unbond (or is the unbond hook for when their tokens are unlocked).

SweeXordious: > Don't you require 1/3 to double sign in order to genuinely equivocate

Yes, but even if they will be unable to halt the QGB, we still would want to slash any bad behavior

> Also it will take 21 days for the validator to truly unbond (or is the unbond hook for when their tokens are unlocked).

But, if after they start unbonding their tokens, we don't create a valset. Let's imagine a scenario where during that 21 days, no change happens to the valset. Then, day 22 we will still have the relayer getting the signature of that 
 malicious orchestrator, his tokens will be unlocked, and he will be able to submit any signature and will get relayed to the QGB. Of course, we're implementing the relayer to be smarter and not relay it, but also we wouldn't want this behaviour to go unpunished

x/qgb/README.md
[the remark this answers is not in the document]
evan-forbes: cool that part makes sense, and including the rational is a good thing.  The part I think we need to change is to also discuss the percentage change, which is the driver of the reduction. We only change if there is a 5% difference. We calculate that difference by first normalizing. This combined mechanism is responsible for safely reducing the frequency of valdiator set updates.

x/qgb/README.md
evan-forbes: sounds good

x/qgb/README.md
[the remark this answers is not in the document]
evan-forbes: cool

x/qgb/README.md
SweeXordious: [cca1dab](https://github.com/celestiaorg/celestia-app/pull/1911/commits/cca1dab597b78af7c4e6d38a624e906301e7a141)

x/qgb/README.md
cmwaters: > Yes, but even if they will be unable to halt the QGB, we still would want to slash any bad behavior

I'm usually a punishment minimalist. So long as there's not a 1/3 validator set change, there's nothing a malicious validator can do. I think 5% is a good threshold in this case.

cmwaters (approved):
Thanks for answering all my questions. 👍

x/qgb/README.md
```diff
@@ -110,56 +360,30 @@ if err != nil {
...
+### Hooks initialization panic
 
-Use "celestia-appd verify [command] --help" for more information about a command.
+When initializing the QGB hooks, if the QGB store key is not setup correctly, the state machine will [panic](https://github.com/celestiaorg/celestia-app/blob/0629c757ef35a24187a8d7a4c706c7cdc894c8b6/x/qgb/keeper/hooks.go#L14-L19):
 
+```golang
+// if startup is mis-ordered in app.go this hook will halt the chain when
+// called. Keep this check to make such a mistake obvious
+if k.storeKey == nil {
+panic("hooks initialized before GravityKeeper")
```
evan-forbes, line 379: adding this, but tbh probably not worth the extra round of reviews lol

used 3 " " instead of 4 since the rest are using 3
```suggestion
   panic("hooks initialized before GravityKeeper")
```

evan-forbes (approved):
dope I think this pairs really well with #1974!

x/qgb/README.md
SweeXordious: I will handle it in a subsequent PR, thanks :+1:

SweeXordious: https://github.com/celestiaorg/celestia-app/pull/1976

PR #1917, opened 2023-06-13
chore!: merge main to prepare for rc3

evan-forbes:
## Overview

merging main for v1.0.0-rc3


rootulp (approved):

SweeXordious (dismissed):

SweeXordious:
@evan-forbes can this PR also be included in this? https://github.com/celestiaorg/celestia-app/pull/1902

evan-forbes:
@SweeXordious yeah thanks for catchiing

that looks like it could be consensus breaking, is it?

evan-forbes:
it appears that the `TestSquareSizeIntegrationTest` is failing for the max hardcoded blocksize quite often in CI but I'm not getting it to fail locally. I'm investigating which change has increased it significantly. The recent update to the sdk might be the culprit, but I'm unsure.

SweeXordious:
> that looks like it could be consensus breaking, is it?

yes, it is consensus breaking, but I believe everything is setup correctly, only the check was missing. So, it shouldn't halt the network

SweeXordious (approved):

PR #1934, opened 2023-06-15
feat: introduce ibc redundant relay decorator

cmwaters:
Non state-breaking change that improves the efficiency of relayers by gating redundant transactions at the mempool layer.

This also extracts out the ante handler into its own file for readability


evan-forbes (approved):
neat

rootulp (approved):

Issue #1916, opened 2023-06-13
Mitigate spam PFBs


cmwaters:
### Summary

Add sufficient block validity rules that prevents users from get data availability for free. This can happen if a transaction makes it into a block but is invalid, thus the user doesn't pay for it. We don't necessarily need full validity checks in ProcessProposal so long as the signer has adequate fees to cover the gas for the transaction.

### Problem Definition

As a specific example, I could submit a 1MB PFB using an account that has insufficient funds for the tx. So long as there is enough gas to cover the `SigGasConsumeDecorator` and `ConsumeTxSizeGasDecorator` it will pass `CheckTx`, be proposed in a block and accepted by validators but run out of gas when it comes to being executed. 

### Proposal

Create a new `Decorator`, `CheckPFBGasDecorator` which is run in `PrepareProposal` and `ProcessPropsoal` and ensures that the PFB signer has sufficient gas to cover the costs of the PFB given the size of the blobs.

We should also move the `SigGasConsumeDecorator` and `ConsumeTxSizeGasDecorator` to Prepare/ProcessProposal to more accurately estimate the amount of gas that will be consumed in DeliverTx.

Note, here I'm talking about gas and not gas prices (where gas * gas price = fee). Currently, it's possible to set the gas price to 0 and not pay anything anyway. This is fine to start with but once we put together a fee system we will want to ensure that there are no loop holes.

Given the validity checks in Prepare/ProcessProposal, it may be possible to make `CheckTx` lighter than what it already is

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1936, opened 2023-06-15
feat: implement MinGasPFBDecorator

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1916

As this only affects CheckTx, this is a non-breaking change and can be in v1.1. However, ideally we apply this to `ProcessProposal` which is a state machine breaking change and will need to be done in v2

### DIFF
--- a/x/blob/ante/ante.go
+++ b/x/blob/ante/ante.go
@@ -1,0 +1,69 @@
+package ante
+
+import (
+	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/pkg/shares"
+	"github.com/celestiaorg/celestia-app/x/blob/types"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// MinGasPFBDecorator helps to prevent a PFB from being included in a block
+// but running out of gas in DeliverTx (effectively getting DA for free)
+// This decorator should be run after any decorator that consumes gas.
+type MinGasPFBDecorator struct {
+	k BlobKeeper
+}
+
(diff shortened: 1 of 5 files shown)


rootulp (dismissed):

x/blob/ante/ante.go
```diff
@@ -0,0 +1,65 @@
...
+func NewMinGasPFBDecorator(k BlobKeeper) MinGasPFBDecorator {
+	return MinGasPFBDecorator{k}
+}
+
+// AnteHandle implemnts the AnteHandler interface. It checks to see
+// if the transaction contains a MsgPayForBlobs and if so, checks that
+// the transaction has allocated enough gas.
+//
+// TODO: We need to run this antehandler in process proposal
+// hoever this will be a breaking change.
```
rootulp, line 29: [nit]

```suggestion
// however this will be a breaking change.
```

We should consider additional linters to catch typos like this. I would've expected https://github.com/celestiaorg/celestia-app/blob/c8f5c3376d6d8da214bd3af677c4d23970ea807b/.golangci.yml#L9 to catch 

Ref: https://golangci-lint.run/usage/linters/#misspell

rootulp (approved):

evan-forbes (approved):
LGTM! I tested locally just to make sure I wasn't missing anything, but if we really wanted to be comprehensive/paranoid, we might want to test how this works when paired with the other ante handlers that set the gas etc. Like, just add a quick unit test to our checktx tests or even the integration test.

cmwaters:
> LGTM! I tested locally just to make sure I wasn't missing anything, but if we really wanted to be comprehensive/paranoid, we might want to test how this works when paired with the other ante handlers that set the gas etc. Like, just add a quick unit test to our checktx tests or even the integration test.

Yeah I could look into adding an integration test if you want

evan-forbes (approved):

rootulp (approved):

Issue #1898, opened 2023-06-09
Guard against panics in `ProcessProposal`


evan-forbes:
Currently, we are not guarding against panics when the application calls process proposal. While this should not be possible, we could protect against potential exploitation of a bug by adding a deferred routine similar to:

```go
	defer func() {
		if err := recover(); err != nil {
			app.logger.Error(
				"panic recovered in PrepareProposal",
				"height", req.Height,
				"time", req.Time,
				"panic", err,
			)

			resp = abci.ResponsePrepareProposal{Txs: req.Txs}
		}
	}()

	resp := RelevantLogicEtc()
	return resp
```

PR #1948, opened 2023-06-16
feat: use panics and recovery appropriately

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1898

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -12,7 +12,8 @@ import (
 // preparing the proposal block data. The square size is determined by first
 // estimating it via the size of the passed block data. Then, this method
 // generates the data root for the proposal block and passes it back to
-// tendermint via the BlockData.
+// tendermint via the BlockData. Panics indicate a developer error and should
+// immediately halt the node for visibility and so they can be quickly resolved.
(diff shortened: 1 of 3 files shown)


rootulp (approved):

app/prepare_proposal.go
```diff
@@ -12,7 +12,8 @@ import (
 // preparing the proposal block data. The square size is determined by first
 // estimating it via the size of the passed block data. Then, this method
 // generates the data root for the proposal block and passes it back to
-// tendermint via the BlockData.
+// tendermint via the BlockData. Panics indicate a developer error and should
+// immediately halt the node for visibility and so they can be quickly resolved.
```
evan-forbes, line 16: are these docs up to date with the below?

>// In the case of a panic from an unexpected condition, it is better for the liveness of the
	// network that we catch it, log an error and vote nil than to crash the node.

evan-forbes (approved):

app/prepare_proposal.go
cmwaters: Yeah. `PrepareProposal` has panics, `ProcessProposal` catches them.

PR #1949, opened 2023-06-16
chore: `make proto-gen`

rootulp:
When I run `make proto-gen` on main, I observe changes.

https://github.com/celestiaorg/celestia-app/pull/1768 changed [this line](https://github.com/celestiaorg/celestia-app/commit/5b35a01f2955cafbd545c2b51e17195df1d39321#diff-94b9ac93697af06645344ee8040d0b891f934ad1a5e49c7c39c3cb062ba34a4fR71) without a change to params.proto

### DIFF
--- a/x/blob/types/params.pb.go
+++ b/x/blob/types/params.pb.go
@@ -68,7 +68,7 @@ func (m *Params) GetGasPerBlobByte() uint32 {
 	return 0
 }
 
-func (m *Params) GetGovSquareSizeUpperBound() uint64 {
+func (m *Params) GetGovMaxSquareSize() uint64 {
 	if m != nil {
 		return m.GovMaxSquareSize
 	}
--- a/x/qgb/types/query.pb.go
+++ b/x/qgb/types/query.pb.go
@@ -6,8 +6,8 @@ package types
 import (
 	context "context"
 	fmt "fmt"
-	_ "github.com/cosmos/cosmos-proto"
 	types "github.com/cosmos/cosmos-sdk/codec/types"
+	_ "github.com/cosmos/cosmos-proto"
 	_ "github.com/cosmos/cosmos-sdk/types/query"
 	_ "github.com/cosmos/gogoproto/gogoproto"
 	grpc1 "github.com/gogo/protobuf/grpc"


x/qgb/types/query.pb.go
```diff
@@ -6,8 +6,8 @@ package types
 import (
 	context "context"
 	fmt "fmt"
-	_ "github.com/cosmos/cosmos-proto"
 	types "github.com/cosmos/cosmos-sdk/codec/types"
+	_ "github.com/cosmos/cosmos-proto"
```
SweeXordious, line 10: This one is expected i guess. Could you please gofumpt and see if this change is reverted?

x/blob/types/params.pb.go
```diff
@@ -68,7 +68,7 @@ func (m *Params) GetGasPerBlobByte() uint32 {
 	return 0
 }
 
-func (m *Params) GetGovSquareSizeUpperBound() uint64 {
+func (m *Params) GetGovMaxSquareSize() uint64 {
```
SweeXordious, line 71: This one should have been caught by CI... weird

rootulp: It looks like https://github.com/celestiaorg/celestia-app/pull/1768 didn't modify a `.proto` file so the protobuf workflow wasn't run.

Ref: https://github.com/rootulp/celestia-app/blob/0464cdb652b165c25d4f963f60fe8848c2cfeb30/.github/workflows/protobuf.yml#L2

cmwaters (approved):

x/blob/types/params.pb.go
SweeXordious: Then I propose to run this workflow if a `.proto` changes and also if any `*.pb.*` so that if any change happens on any of these, we check if the proto files are consistent with the generated ones

SweeXordious (approved):

x/blob/types/params.pb.go
rootulp: Yea that works! Since proto-lint and proto-gen take much less time than tests, I proposed always running them in https://github.com/celestiaorg/celestia-app/pull/1950

Issue #743, opened 2022-09-20
specs: Precisely describe all block validity rules in `ProcessProposal`


evan-forbes:
A large portion of the responsibilities of the application is to define the block validity rules that are checked by validators during consensus (`ProcessProposal`), and enforced by creating fraud proofs for each rule. Things like namespaces must be ordered lexicographically, all messages must have a corresponding PFD transaction with a commitment that consists only of subtree roots, tail padding shares must consist of zeroed data, etc

We should define each of these rules in the specs.

rootulp:
In addition to the rules specified above: all namespaces must be valid (i.e. use a supported namespace version and ID)

PR #1966, opened 2023-06-20
docs: add block validity rules specs

evan-forbes:
## Overview

Figured we could at least use this as a place holder as a high level summary of celestia-app, and to point to all the other portions of the application.

part of and closes https://github.com/celestiaorg/celestia-app/issues/743
part of #650 

[rendered](https://github.com/celestiaorg/celestia-app/blob/b44960898f59ea3ed86430828606cdc72107a0be/specs/src/specs/block_validity_rules.md)

## Checklist


specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,99 @@
 # Block Validity Rules
```
evan-forbes, line 1: we might just want to call this a summary or something idk yet

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,98 @@
...
+
+Unlike most blockchains, Celestia derives most of its functionality from
+stateless commitments to data rather than stateful transitions. This means that
+the protocol relies heavily on block validity rules. Notably, resource
+constrained light clients must be able to detect when these validity rules have
+not been followed in order to avoid making an honest majority assumption.
+
+> **Note** Celestia relies on cometBFT (formerly tendermint) for consensus,
```
cmwaters, line 9:
```suggestion
> **Note** Celestia relies on CometBFT (formerly tendermint) for consensus,
```
Not capitalised?

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,98 @@
...
+construction of that square is also vital to correctly computing it.
+
+TODO
+
+#### Share Encoding
+
+Each chunk of block data is split into equally size shares for sampling
+purposes. The encoding was designed to allow for light clients to decode these
+shares to retrieve relevant data and to be future-proof yet backwards
+compatible. The share encoding is deeply integrated into square contraction, and
```
cmwaters, line 76:
```suggestion
compatible. The share encoding is deeply integrated into square construction, and
```

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,98 @@
...
+therefore critical to calculate the data root.
+
+See [shares spec](./shares.md)
+
+## `BlobTx` Validity Rules
+
+Each `BlobTx` consists of a transaction to pay for the blob, and the blob
+itself. Each `BlobTx` that is included in the block must be valid. Those rules
+are described in [`x/blob` module
+specs](../../../x/blob/README.md#validity-rules)
```
cmwaters, line 86: We should adjust this to explain that there can be multiple blobs in a `BlobTx`

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,98 @@
 # Block Validity Rules
+
+Unlike most blockchains, Celestia derives most of its functionality from
+stateless commitments to data rather than stateful transitions. This means that
+the protocol relies heavily on block validity rules. Notably, resource
+constrained light clients must be able to detect when these validity rules have
+not been followed in order to avoid making an honest majority assumption.
```
staheri14, line 7: [Question] Is the block validity rule the same as the logics implemented in the [ProcessProposal](https://github.com/celestiaorg/celestia-app/blob/8ae6a84b2c99e55625bbe99f70db1e5a985c9675/app/process_proposal.go#L20)? If we were to map it to a single function, would process proposal be the right one?

rootulp (dismissed):

specs/src/specs/block_validity_rules.md
cmwaters: Yup

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,98 @@
...
+
+> **Note** Celestia relies on cometBFT (formerly tendermint) for consensus,
+> meaning that it has single slot finality and is fork-free. Therefore, in order
+> to ensure that an invalid block is never committed to, each validator must
+> check that each block follows all validity rules before voting. If over two
+> thirds of the voting power colludes to break a validity rule, then fraud
+> proofs are created for light clients. After light clients verify fraud proofs,
```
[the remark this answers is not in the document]
evan-forbes: BEFPs do, and blob inclusion will sooner or later. Not sure when we'll get state fraud proofs

this does bring up a good point, where I feel like all fraud proofs are actually block validity fraud proofs, and perhaps I should try to change the first paragraph to emphasize that

specs/src/specs/block_validity_rules.md
evan-forbes: :+1: good idea 069bbd787e829ee0b710b55e921ced76ad502035

specs/src/specs/block_validity_rules.md
```diff
@@ -67,22 +67,23 @@ light clients. Since the data root is a commitment to the square, the
...
 
 #### Share Encoding
 
 Each chunk of block data is split into equally size shares for sampling
 purposes. The encoding was designed to allow for light clients to decode these
 shares to retrieve relevant data and to be future-proof yet backwards
-compatible. The share encoding is deeply integrated into square contraction, and
-therefore critical to calculate the data root.
+compatible. The share encoding is deeply integrated into square construction, and
+    therefore critical to calculate the data root.
```
rootulp, line 78: I think this was accidental

```suggestion
therefore critical to calculate the data root.
```

rootulp (approved):

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,99 @@
...
+
+> **Note** Celestia relies on CometBFT (formerly tendermint) for consensus,
+> meaning that it has single slot finality and is fork-free. Therefore, in order
+> to ensure that an invalid block is never committed to, each validator must
+> check that each block follows all validity rules before voting. If over two
+> thirds of the voting power colludes to break a validity rule, then fraud
+> proofs are created for light clients. After light clients verify fraud proofs,
+> they halt.
+
+Before any Celestia specific validation is performed, all cometBFT [block
```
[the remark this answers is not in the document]
evan-forbes: my bad, meant to grab the rest of those after the first one

corrected in https://github.com/celestiaorg/celestia-app/pull/1966/commits/c3e19fe2cafb75ce88b84014a4bec37bab99bc36

rootulp (dismissed):
🙌 fraud proof specs

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,77 @@
...
+> proofs are created for light clients. After light clients verify fraud proofs,
+> they halt.
+
+Before any Celestia specific validation is performed, all CometBFT [block
+validation
+rules](https://github.com/cometbft/cometbft/blob/v0.34.28/spec/core/data_structures.md#block)
+must be followed. The only deviation from these rules is how the data root
+([DataHash](https://github.com/cometbft/cometbft/blob/v0.34.28/spec/core/data_structures.md#header))
+is generated. Almost all of Celestia's functionality is derived from this
+change, including how it proves data availability to light clients.
```
staheri14, line 25: [Suggestion] I see there is a missing link between the block validity rules and the supplied subsections. My suggestion revolves around making this link a bit clear (feel free to rephrase it as you see fit, I got inspiration from the subsections you already included in the specs):
Below is my suggested addition to the text: 

In Celestia, the block validity rules encompass two main aspects: the rules governing the validity of individual transactions and the rules dictating how to construct the data hash.

- Transaction validity rules: Similar to any normal block, Celestia blocks are composed of a set of transactions. Therefore, the transaction validity rules defined by CometBFT form an integral part of the block validity rules in Celestia. (a reference to the appropriate documentation for the specific transaction validity rules.)

- Blob transaction: The Blob transaction is unique to Celestia, and its validity rules are covered in the corresponding [specifications](either link the specs or a subsection in this spec). ( I am assuming this is something we need to explain separately, that is why I have put it as a separate item)

- Data hash calculation steps: The following steps are specific to the calculation of the data hash:

  - Encoding/laying out transactions into a data square: Transactions within the block are serialized and organized into a square format. This process involves reorganizing the transactions, converting them into a series of bytes, splitting them into fixed-size shares, and arranging them in a square structure. All these steps adhere to specific rules, which are also considered part of the block validity rules. (Links to the respective specifications, OR subsections of the current doc, should be provided for further details.)

  - Constructing the data hash: The data hash is computed from the data square mentioned earlier. Detailed information on the construction of the data hash can be found in the [provided specifications OR a subsection in this doc](). The correct construction of the root hash is also included within the block validity rules.

staheri14 (commented):
LGTM! added some suggestions.

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,77 @@
...
+
+## Data Availability
```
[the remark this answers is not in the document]
evan-forbes: hmm good point. I guess this was just a section for block validity rules that were directly related to data availability

this point is usually glossed over in most blockchains, as it is assumed that the data is downloaded entirely in order to verify it

this section describes the difference between consensus nodes and light clients to emphasize that point

> The data for each block must be considered available before a given block can be
considered valid. For consensus nodes, this is done via an identical mechanism
to a normal CometBFT node, which involves downloading the entire block by each
node before considering that block valid.

> Light clients however do not download the entire block.

and then uses this preface for the reasoning behind the rest of the rules

do you think we should remove this header? While I'm less confident about the header, I do think the three paragraphs after the header are a useful preface

specs/src/specs/fraud_proofs.md
```diff
@@ -0,0 +1,25 @@
+# Fraud Proofs
+
+## Bad Encoding Fraud Proofs
```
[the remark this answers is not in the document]
evan-forbes: That is a good point that fraud proofs are not directly part of core/app atm, so including them in the spec might not make a lot of sense. I do think we should at least have a link to the relevant specs, since that will preserve our capability to send a newcomer a single link where they can find all the important portions.

The initial reason for including something here is that I feel that they have a large impact on the design and implementation. For example, the sole reason we have square layout/blob commitment rules are to accomodate blob inclusion proofs. If we eventually have at least a high level description of it, then we can see the end result, which I think makes the rest of the concepts a lot clearer.

rootulp (approved):

specs/src/specs/fraud_proofs.md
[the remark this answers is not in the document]
evan-forbes: That is a good point that fraud proofs are not directly part of core/app atm, so including them in the spec might not make a lot of sense. I do think we should at least have a link to the relevant specs, since that will preserve our capability to send a newcomer a single link where they can find all the important portions.

The initial reason for including something here is that I feel that they have a large impact on the design and implementation. For example, the sole reason we have square layout/blob commitment rules are to accomodate blob inclusion proofs. If we eventually have at least a high level description of it, then we can see the end result, which I think makes the rest of the concepts a lot clearer.

specs/src/specs/block_validity_rules.md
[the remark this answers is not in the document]
evan-forbes: I will attempt to address this when I address the above [comment](https://github.com/celestiaorg/celestia-app/pull/1966/files#r1242785436)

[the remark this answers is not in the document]
staheri14: I understand your point, and it does make sense. I would recommend consolidating all the relevant background into a dedicated section, such as "Background" or "Introduction".

specs/src/specs/fraud_proofs.md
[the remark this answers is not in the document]
staheri14: > I do think we should at least have a link to the relevant specs, since that will preserve our capability to send a newcomer a single link where they can find all the important portions.

Good idea and agree!

>The initial reason for including something here is that I feel that they have a large impact on the design and implementation. For example, the sole reason we have square layout/blob commitment rules are to accomodate blob inclusion proofs. If we eventually have at least a high level description of it, then we can see the end result, which I think makes the rest of the concepts a lot clearer. 

Agree that such information, at a high level,  will provide readers with a clear understanding of the motivations behind the design choices and the purpose of the validity rules. 
Following my previous comment, I recommend incorporating all the motivational information into a single section. Additionally, it would be beneficial to explicitly mention that the design of the Celestia block is influenced by these reasons, resulting in an extensive list of validity rules.

specs/src/specs/block_validity_rules.md
evan-forbes: > In Celestia, the block validity rules encompass two main aspects: the rules governing the validity of individual transactions and the rules dictating how to construct the data hash.

I like this a lot for the celestia specific rules and have updated accordingly. Note that I have simplified further by deleting a lot of the original descriptions and instead relying on the other portions of the spec that already cover it

evan-forbes:
updated this to be significantly smaller, and instead just rely on pointing to other portions of the spec for the time being

specs/src/specs/block_validity_rules.md
```diff
@@ -16,62 +19,33 @@ block validity rules can be foud in the [Fraud Proofs](./fraud_proofs.md) spec.
...
 
-#### Share Encoding
+All `BlobTx` transactions must be valid according to the [BlobTx validity rules](../../../x/blob/README.md#validity-rules)
 
-Each chunk of block data is split into equally size shares for sampling
-purposes. The encoding was designed to allow for light clients to decode these
-shares to retrieve relevant data and to be future-proof yet backwards
-compatible. The share encoding is deeply integrated into square construction, and
-therefore critical to calculate the data root.
+All remaining transaction types do not have to by valid if included in a block. For a complete list of modules see [state machine modules](./state_machine_modules.md).
```
rootulp, line 44:
```suggestion
All remaining transaction types do not have to be valid if included in a block. For a complete list of modules see [state machine modules](./state_machine_modules.md).
```

rootulp (approved):

specs/src/specs/block_validity_rules.md
```diff
@@ -1 +1,51 @@
...
+> with Dishonest Majorities"](https://arxiv.org/abs/1809.09044) and in the
+> [`celestia-node`](https://github.com/celestiaorg/celestia-node) repo.
+
+Celestia specifc validity rules can be categorized into two groups:
+
+### Transaction Validity Rules
+
+All `BlobTx` transactions must be valid according to the [BlobTx validity rules](../../../x/blob/README.md#validity-rules)
+
+All remaining transaction types do not have to by valid if included in a block. For a complete list of modules see [state machine modules](./state_machine_modules.md).
```
staheri14, line 44: I guess checking the validity of the transaction nonces goes under the CometBFT validity rules, right?

staheri14 (approved):
LGTM!

specs/src/specs/block_validity_rules.md
evan-forbes: good question! I don't think the ante handlers are really specified sufficiently anywhere, especially our usage of them, so I added #2044 

that might involve updating the above! the reasoning behind saying "all remaining transactions types do not have to be valid" were these lines

https://github.com/celestiaorg/celestia-app/blob/65046dd363fd5b50d9cfc8d83290c28de10a22a2/app/process_proposal.go#L55-L61

in the past, the rules were that validators could include invalid transactions in the block. It's very reasonable to see this as incorrect now since the most recent fix to use the entire antehandler instead of only incrementing the nonce!

https://github.com/celestiaorg/celestia-app/blob/65046dd363fd5b50d9cfc8d83290c28de10a22a2/app/process_proposal.go#L72-L79

we might want to say something along the lines of "random tx data is permitted, but decodable sdk.Txs must pass all antehandler checks in order for the block to be valid".

evan-forbes: linked this convo as part of that issue to ensure we address this specifically

specs/src/specs/fraud_proofs.md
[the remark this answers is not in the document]
cmwaters: I do think that wherever validity rules are described should be were fraud proofs which define the failure to comply with the aforementioned validity rules should reside. Therefore I think fraud proofs should be described in the specification here.

Issue #1913, opened 2023-06-12
Add visuals to the data square layout portion of the spec


evan-forbes:
This portion of the spec could benefit by adding more visuals. ADR013 has some good examples.

PR #1967, opened 2023-06-21
docs: add more context and figures to the data square layout spec

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/issues/1913


specs/src/specs/data_square_layout.md
```diff
@@ -14,7 +14,7 @@ Block data consists of:
...
 1. Blobs: binary large objects which do not modify the Celestia state, but which are intended for a Celestia application identified with a provided namespace.
 
-We want to arrange this data into a `k * k` matrix of fixed-sized [shares](../specs/shares.md), which will later be committed to in [Namespace Merkle Trees (NMTs)](https://github.com/celestiaorg/nmt/blob/v0.16.0/docs/spec/nmt.md) so that individual shares in this matrix can be proven to belong to a single data root.
+We want to arrange this data into a `k * k` matrix of fixed-sized [shares](../specs/shares.md), which will later be committed to in [Namespace Merkle Trees (NMTs)](https://github.com/celestiaorg/nmt/blob/v0.16.0/docs/spec/nmt.md) so that individual shares in this matrix can be proven to belong to a single data root. `k` must always be a power of 2 (e.g. 4, 8, 16, 32, etc.) as this is optimal for the erasure coding algorithm.
```
rootulp, line 17: [question] is `1` a valid square size? If yes, thoughts on 

```suggestion
We want to arrange this data into a `k * k` matrix of fixed-sized [shares](../specs/shares.md), which will later be committed to in [Namespace Merkle Trees (NMTs)](https://github.com/celestiaorg/nmt/blob/v0.16.0/docs/spec/nmt.md) so that individual shares in this matrix can be proven to belong to a single data root. `k` must always be a power of 2 (e.g. 1, 2, 4, 8, 16, 32, etc.) as this is optimal for the erasure coding algorithm.
```

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,40 @@ First, we impose some ground rules:
...
 
-Transactions can pay fees for a blob to be included in the same block as the transaction itself. It may seem natural to bundle the PayForBlob transaction that pays for a number of blobs with these blobs (which is the case in other blockchains with native execution, e.g. calldata in Ethereum transactions or OP_RETURN data in Bitcoin transactions), however this would mean that processes validating the state of the Celestia network would need to download all blob data. PayForBlob transactions must therefore only include a commitment to (i.e. some hash of) the blob they pay fees for. If implemented naively (e.g. with a simple hash of the blob, or a simple binary Merkle tree root of the blob), this can lead to a data availability problem, as there are no guarantees that the data behind these commitments is actually part of the block data.
+Given these rules, a square may look as follows:
+
+![](./figures/square_layout.svg)
+
+Padding is addressed in the [padding section](#padding). Namespace C contains two blobs of two shares each while Namespace D contains one blob of three shares.
+### Ordering
```
rootulp, line 33: [nit] I'm surprised markdown lint didn't complain and ask for a newline before this heading

```suggestion

### Ordering
```

specs/src/specs/figures/square_layout.svg
rootulp, line 1: [nit] I think the white cell between namespace C B2 and namespace D B1 should be light blue b/c it's a padding share with namespace C. Alternatively if it remains clear, it could be labeled "P".

[nit] I don't understand why there is a padding share in Namespace C between B1 and B2.

[optional] The white padding share in the top right of the square could be labeled "Reserved namespace padding" but doesn't strictly need to.

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,40 @@ First, we impose some ground rules:
...
+
+The order of blobs in a namespace is dictated by the priority of the PFBs that payed for the blob. A PFB with greater priority will have all blobs in that namespace strictly before a PFB with less priority.
```
rootulp, line 35: [optional] would it help to clarify that priority is based on fees?

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,40 @@ First, we impose some ground rules:
...
 
-Specifically, blobs must begin at a new share. We note a nice property from this rule: if the transaction sender knows 1) `k`, the size of the matrix, 2) the starting location of their blob in a row, and 3) the length of the blob (they know this since they are sending the blob), then they can actually compute a sequence of roots to _subtrees in the row NMTs_. More importantly, anyone can compute this, and can compute _the simple Merkle root of these subtree roots_.
+Specifically, a `MsgPayForBlobs` must include a `ShareCommitment` over the contents of each blob it is paying for. If the transaction sender knows 1) `k`, the size of the matrix, 2) the starting location of their blob in a row, and 3) the length of the blob (they know this since they are sending the blob), then they can actually compute a sequence of roots to _subtrees in the row NMTs_. Taking _the simple Merkle root of these subtree roots_ provides us with the `ShareCommitment` that gets included in `MsgPayForBlobs`. Using subtree roots instead of all the leafs makes blob inclusion proofs smaller.
```
rootulp, line 46: [nit] after square size independent blob commitments, the msg sender no longer needs to know `k` or the starting location.

^ disregard, I see this is addressed later.

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,40 @@ First, we impose some ground rules:
...
-With the above constraint, we can compute subtree roots deterministically. For example, a blob of 128 shares and `SubtreeRootThreshold` (SRT) = 64, must start on a share index that is a multiple of 2 because 128/64 = 2. In this case, there will be a maximum of 1 share of padding between blobs (more on padding below). The maximum subtree width in shares will also be 2, meaning that there will be 2 shares under each subtree root.
+![](./figures/subtree_width.svg)
 
-The last piece of the puzzle is determining _which_ row the blob is placed at (or, more specifically, the starting location). This is needed to keep the block producer accountable. To this end, the block producer simply augments each fee-paying transaction with the starting locations of the blobs the transaction pays for.
+The `SubtreeRootThreshold` is an arbitrary versioned protocol constant that aims to put a soft limit on the number of subtree roots included in a blob inclusion proof, as described in [ADR013](../../../docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md). A higher `SubtreeRootThreshold` means less padding and more tightly packed squares but also means greater proof sizes.
```
rootulp, line 59: [optional][nit] to disambiguate with other types of proofs (i.e. bad encoding fraud proofs)

```suggestion
The `SubtreeRootThreshold` is an arbitrary versioned protocol constant that aims to put a soft limit on the number of subtree roots included in a blob inclusion proof, as described in [ADR013](../../../docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md). A higher `SubtreeRootThreshold` means less padding and more tightly packed squares but also means greater blob inclusion proof sizes.
```

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,40 @@ First, we impose some ground rules:
...
 
-This, however, requires the block producer to interact with the transaction sender to provide them the starting location of their blob, so that the sender can sign over the commitment based on that starting location. This can be done selectively, but is not ideal as a default for e.g. end-user wallets.
+![](./figures/subtree_roots.svg)
 
-### Blob Share Commitment Rules
+Understanding 1) and 2) would usually require interaction with the block proposer. To make the possible starting locations of blobs sufficiently predictable and to make `ShareCommitment` independent of `k`, we impose an additional rule. The blob must start at an index that is either:
 
-To make the possible starting locations of blobs sufficiently predictable and constrained such that users can deterministically compute subtree roots, needed for the `ShareCommitment`s within a PFB, without the need to interact with the block proposer, we impose one additional rule:
+1. A multiple of the `SubtreeWidth`. 
+2. A multiple of `k`, the size of the square.
```
rootulp, line 53: [blocking] I don't think the version of non-interactive default rules that are implemented cares about the square size. If it did, then interaction would be needed between the block proposer and the msg sender.

I just updated ADR-013 w/ clarifying details but I think this meant to refer to the minimum square size that could contain the blob (a.k.a `MinSquareSize(blob)`)

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,40 @@ First, we impose some ground rules:
...
 
-### Caveats
+## Padding
 
-The blob placement rules described above conflict with the first rule that shares must be ordered by namespace ID, as shares between two blobs that are not placed adjacent to each other do not have a natural namespace they belong to. This is resolved by requiring that such shares have a value of zero and a namespace ID equal to the preceding blob's. Since their value is known, they can be omitted from NMT proofs of all shares of a given namespace ID.
+Given these rules whereby blobs in their share format can't be directly appended one after the other, we use padding shares to fill the gaps. These are shares with all `0` bytes. Padding always comes after the blobs in the namespace. The padding at the end of the square (Tail Padding) is unique in that it belongs to it's own namespace. All other padding shares the namespace of the blob before it.
```
rootulp, line 64: [blocking] in addition to tail padding, there is also reserved padding: https://celestiaorg.github.io/celestia-app/specs/shares.html#reserved-padding-share

[optional]

```suggestion
Given these rules whereby blobs in their share format can't be directly appended one after the other, we use padding shares to fill the gaps. These are shares with a particular format (see [padding](./shares.md#padding)). Padding always comes after all the blobs in the namespace. The padding at the end of the reserved namespace and at the end of the square are special in that they belong to unique namespaces. All other padding shares use the namespace of the blob before it in the data square.
```

specs/src/specs/data_square_layout.md
cmwaters: In [`NextShareIndex`](https://github.com/celestiaorg/celestia-app/blob/42433f34bce35741bb68e4ca29e3fc33e523211b/pkg/shares/blob_share_commitment_rules.go#LL51C13-L51C13) we pick the next spot as the minimum between the start of the next row and the current cursor + the `SubTreeWidth`

The `MinSquareSize(blob)` is something else but within `SubtreeWidth` function. Perhaps it does make sense to add it.

cmwaters: Looking at `NextShareIndex` I'm a little confused now about if what it's doing makes sense

specs/src/specs/data_square_layout.md
cmwaters: Yeah I can add that

specs/src/specs/data_square_layout.md
cmwaters: > there is also reserved padding

Is reserved padding doing anything helpful? i.e. why can't it just be the padding belonging to the last namespace in the set of reserved namespaces?

evan-forbes (dismissed):
:+1:

evan-forbes (approved):

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,41 @@ First, we impose some ground rules:
...
+The `SubtreeWidth` is calculated as the length of the blob in shares, divided by the [`SubtreeRootThreshold`](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/pkg/appconsts/v1/app_consts.go#L6) and rounded up to the nearest power of 2 ([implementation here](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/pkg/shares/non_interactive_defaults.go#L94-L116)). If the `SubtreeRootThreshold` is greater than the minimum square size that the blob can fit in (i.e. a blob of 15 shares has a minimum square size of 4) then we take that minimum value. This value is used as the width of the first mountain in the [Merkle Mountain Range](https://docs.grin.mw/wiki/chain-state/merkle-mountain-range/) that would represent the `ShareCommitment` (as shown below).
```
rootulp, line 56:
```suggestion
The `SubtreeWidth` is calculated as the length of the blob in shares, divided by the [`SubtreeRootThreshold`](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/pkg/appconsts/v1/app_consts.go#L6) and rounded up to the nearest power of 2 ([implementation here](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/pkg/shares/non_interactive_defaults.go#L94-L116)). If the `SubtreeRootThreshold` is greater than the minimum square size that the blob can fit in (i.e. a blob of 15 shares has a minimum square size of 4) then we take that minimum value. This value is used as the width of the first mountain in the [Merkle Mountain Range](https://docs.grin.mw/wiki/chain-state/merkle-mountain-range/) that would represent the `SubtreeWidth` (as shown below).
```

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,41 @@ First, we impose some ground rules:
...
+The `SubtreeRootThreshold` is an arbitrary versioned protocol constant that aims to put a soft limit on the number of subtree roots included in a blob inclusion proof, as described in [ADR013](../../../docs/architecture/adr-013-non-interactive-default-rules-for-zero-padding.md). A higher `SubtreeRootThreshold` means less padding and more tightly packed squares but also means greater blob inclusion proof sizes.
+With the above constraint, we can compute subtree roots deterministically. For example, a blob of 172 shares and `SubtreeRootThreshold` (SRT) = 64, must start on a share index that is a multiple of 2 because 172/64 = 3. 3 rounded up to the nearest power of 2 is 4. In this case, there will be a maximum of 3 shares of padding between blobs (more on padding below). The maximum subtree width in shares for the first mountain in the merkle range will be 4 (The actual mountain range would be 43 subtree roots of 4 shares each). The `ShareCommitment` is then the merkle tree over the peaks of the mountain range.
```
rootulp, line 61: [typo]

```suggestion
With the above constraint, we can compute subtree roots deterministically. For example, a blob of 172 shares and `SubtreeRootThreshold` (SRT) = 64, must start on a share index that is a multiple of 4 because 172/64 = 3. 3 rounded up to the nearest power of 2 is 4. In this case, there will be a maximum of 3 shares of padding between blobs (more on padding below). The maximum subtree width in shares for the first mountain in the Merkle range will be 4 (The actual mountain range would be 43 subtree roots of 4 shares each). The `ShareCommitment` is then the root of a Merkle tree where the leaves of the tree are the subtree roots.
```

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,41 @@ First, we impose some ground rules:
...
 
-This, however, requires the block producer to interact with the transaction sender to provide them the starting location of their blob, so that the sender can sign over the commitment based on that starting location. This can be done selectively, but is not ideal as a default for e.g. end-user wallets.
+![subtree roots](./figures/subtree_roots.svg)
 
-### Blob Share Commitment Rules
+Understanding 1) and 2) would usually require interaction with the block proposer. To make the possible starting locations of blobs sufficiently predictable and to make `ShareCommitment` independent of `k`, we impose an additional rule. The blob must start at an index that is either:
```
rootulp, line 51:
```suggestion
Understanding 1) and 2) would usually require interaction with the block proposer. To make the possible starting locations of blobs sufficiently predictable and to make `ShareCommitment` independent of `k`, we impose an additional rule. The blob must start at an index that is a multiple of the `SubtreeWidth`.
```

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,41 @@ First, we impose some ground rules:
...
 
-To make the possible starting locations of blobs sufficiently predictable and constrained such that users can deterministically compute subtree roots, needed for the `ShareCommitment`s within a PFB, without the need to interact with the block proposer, we impose one additional rule:
+1. A multiple of the `SubtreeWidth`.
+2. A multiple of `k`, the size of the square.
```
rootulp, line 54: [blocking] I don't think the blob must start at a multiple of `k` 

```suggestion
```

specs/src/specs/data_square_layout.md
```diff
@@ -25,28 +25,41 @@ First, we impose some ground rules:
...
+Specifically, a `MsgPayForBlobs` must include a `ShareCommitment` over the contents of each blob it is paying for. If the transaction sender knows 1) `k`, the size of the matrix, 2) the starting location of their blob in a row, and 3) the length of the blob (they know this since they are sending the blob), then they can actually compute a sequence of roots to _subtrees in the row NMTs_. Taking _the simple Merkle root of these subtree roots_ provides us with the `ShareCommitment` that gets included in `MsgPayForBlobs`. Using subtree roots instead of all the leafs makes blob inclusion proofs smaller.
 
-This, however, requires the block producer to interact with the transaction sender to provide them the starting location of their blob, so that the sender can sign over the commitment based on that starting location. This can be done selectively, but is not ideal as a default for e.g. end-user wallets.
+![subtree roots](./figures/subtree_roots.svg)
```
rootulp, line 49: [optional] I really appreciate the use of visuals but I think this diagram is a bit confusing b/c I don't think subtree 2 will ever actually be used.

Since the width of subtree root 1 is 2 shares, I think this diagram should have subtree root 2 over two shares and subtree root 3 over two shares.

rootulp (commented):
One blocking comment about the non-interactive default rules

specs/src/specs/data_square_layout.md
cmwaters: It did. Whenever padding was large enough that it would overflow into the next row, instead of starting there we would start at the beginning of the row.

I cleaned it up but in a way this rule still holds. For the implementation, instead of making it a multiple of `k`, we say that the blob needs to be able to fit in a square and since the `SubtreeWidth` is the `BlobMinSquareSize` when the `SubtreeRootThreshold` is high, any transaction that take up greater than 1/4 of the square they will always start at a multiple of `k` instead of having padding that might be `2k` or even `4k` (depending on what the `SubtreeRootTheshold` was).

I guess what's changed is that the multiple of `k` is encapsulated in the logic that calculates the `SubTreeWidth` (if you see what I mean)

So what I will do is take your suggestion and say that it's just a multiple of the `SubtreeWidth` but then add another section explaining how the subtree width is calculated.

specs/src/specs/data_square_layout.md
cmwaters: Yeah I see what you mean. That makes sense to me. Maybe I will try replicate the flow I have in the merkle mountain range diagram.

evan-forbes (approved):

specs/src/specs/data_square_layout.md
```diff
@@ -46,19 +46,16 @@ To that end, we impose some additional rules onto _blobs only_: blobs must be pl
...
+Understanding 1) and 2) would usually require interaction with the block proposer. To make the possible starting locations of blobs sufficiently predictable and to make `ShareCommitment` independent of `k`, we impose an additional rule. The blob must start at a multiple of the `SubtreeWidth`.
 
-1. A multiple of the `SubtreeWidth`.
-2. A multiple of `k`, the size of the square.
+The `SubtreeWidth` is calculated as the length of the blob in shares, divided by the [`SubtreeRootThreshold`](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/pkg/appconsts/v1/app_consts.go#L6) and rounded up to the nearest power of 2 ([implementation here](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/pkg/shares/non_interactive_defaults.go#L94-L116)). If the output is greater than the minimum square size that the blob can fit in (i.e. a blob of 15 shares has a minimum square size of 4) then we take that minimum value. This `SubtreeWidth` is used as the width of the first mountain in the [Merkle Mountain Range](https://docs.grin.mw/wiki/chain-state/merkle-mountain-range/) that would all together represent the `ShareCommitment` over the blob.
```
rootulp, line 53: 👍 I think this is way clearer. Thanks for improving this!

rootulp (approved):

PR #1979, opened 2023-06-22
chore: simplify next share index

cmwaters:
This PR pulls out a lot of redundant logic, simplifying the `NextShareIndex` function to make it more readable and understandable. I've also added more comments and updated the links.

Rationale: There is no subtreeWidth value that is not a multiple of the square size given that both must be powers of 2 and that the subtreewidth can't be greater than the minimum blob square size (and the blob has to fit in the square). This reinforces the notion that this algorithm is square size independent

### DIFF
--- a/pkg/shares/blob_share_commitment_rules.go
+++ b/pkg/shares/blob_share_commitment_rules.go
@@ -52,11 +52,6 @@ func BlobSharesUsedNonInteractiveDefaults(cursor, squareSize, subtreeRootThresho
 // See https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/data_square_layout.md
 // for more information.
 func NextShareIndex(cursor, blobShareLen, squareSize, subtreeRootThreshold int) int {
-	// if we're starting at the beginning of the row, then return as there are
-	// no cases where we don't start at 0.
-	if isStartOfRow(cursor, squareSize) {
-		return cursor
-	}
 
 	// Calculate the subtreewidth. This is the width of the first mountain in the
 	// merkle mountain range that makes up the blob share commitment (given the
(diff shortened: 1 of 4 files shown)


evan-forbes (approved):
good find and makes sense that we can refactor this. iirc it was originally written for the first version of non-interactive defaults

pkg/shares/blob_share_commitment_rules.go
```diff
@@ -45,35 +45,31 @@ func BlobSharesUsedNonInteractiveDefaults(cursor, squareSize, subtreeRootThresho
...
 //
-// https://github.com/celestiaorg/celestia-specs/blob/master/src/rationale/message_block_layout.md#non-interactive-default-rules
+// See https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/data_square_layout.md
+// for more information.
 func NextShareIndex(cursor, blobShareLen, squareSize, subtreeRootThreshold int) int {
 	// if we're starting at the beginning of the row, then return as there are
 	// no cases where we don't start at 0.
 	if isStartOfRow(cursor, squareSize) {
 		return cursor
 	}
```
rootulp, line 59: [optional] this optimization doesn't seem necessary. All tests pass with it's removal and the specs don't describe it.
```suggestion
```

rootulp (approved):
Nice clean up!

pkg/shares/blob_share_commitment_rules.go
cmwaters: Yeah it's a minor optimisation. I don't mind removing it

evan-forbes (approved):

rootulp (approved):

PR #1983, opened 2023-06-22
fix!: add namespace version check in pfbs

### LINKED ISSUE #1982: Make blobs that use the max namespace verison invalid
While it would require writing custom software, it is currenty possible that a blob could be submitted using the max namespace version. This is because we are only relying on the valdiate namespace function in the namespace pkg to check for namespace version validity instead of also checking for it in the blob module. 

We should add a check during PFB validation to prevent this from occuring. We should be safe to hardfork despite this is consensus breaking, since no one has exploited in on the testnets. We can double check by syncing from scratch.

evan-forbes:
## Overview

closes #1982 

## Checklist


rootulp (dismissed):

cmwaters (dismissed):
LGTM

cmwaters (approved):

rootulp (approved):

Issue #1971, opened 2023-06-21
MinGasPFBDecorator should also be run in ProcessProposal


cmwaters:
### Summary

https://github.com/celestiaorg/celestia-app/pull/1936 introduced a new decorator to ensure that PFBs provide a minimum amount of gas to cover for the bytes they are paying for. It would be consensus breaking to include this in `ProcessProposal` so the check initially only covers `CheckTx`. In v2, and once we upgrade SDK (so we can check the state that we call `runtx` in), we should introduce this decorator in `ProcessProposal`.

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

evan-forbes:
what are the benefits of running this in process proposal?

evan-forbes:
but can they still include zero fee transactions?

cmwaters:
> but can they still include zero fee transactions?

Yes so long as their is no fee mechanism (i.e. they can have a tx with the correct amount of gas but a gas-price of 0). But in the future, I assume we will put in place some fee mechanism that doesn't permit zero-fee transactions

cmwaters:
> I see so we just don't need to worry about this until we have a min fee?

kind of, although there's technically already a min fee. It's just on a per validator basis so I think we should still fix this as soon as possible

evan-forbes:
> It's just on a per validator basis so I think we should still fix this as soon as possible

I'm still unsure of what this adds, since a validator can still include arbitrarily priced txs. This won't contribute to the same issue as #1937 correct?

cmwaters:
Sure but I'm assuming that celestia eventually wants some sustainable economic model and when that happens we need to close off all the ways that users can cut corners. I'd rather adopt this now then have to wait for the next breaking release.

PR #1984, opened 2023-06-22
fix!: implement full antehandler checks

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1937

Issue #49, opened 2021-03-16
Fees: investigate and implement fee-burning


liamsi:
We want to implement eip-1559 style fee burning (but in the current tendermint deferred execution model).

This requires two parts:

1. Investigate how to achieve the above using the current execution model and within the existing ABCI methods (keeping in mind that abci++ is currently being spec'ed out). This needs to be summarized in an ADR and @adlerjohn volunteered to have a look into this.
2. Implement the fee-burning mechanism according to the above ADR in alignment with the specs.

Until then, we can probably just live with #48.

hxrts:
@fedekunze has done some work on this if you didn't catch his comment in discord https://github.com/celestiaorg/lazyledger-app/issues/49

liamsi:
Yeah, @fedekunze mentioned that to me but I didn't see the discord message nor the code so far. I'll check it out! Thanks for the pointer @hxrts 🙏🏼

Wondertan:
Yet an unpopular [concern](https://mtefagh.github.io/fee/) about EIP-1559 with in-depth analysis. It might be worth looking into.

evan-forbes:
closed with #380

evan-forbes:
just for posterity: as we've discussed in other issues/channles, we cannot have fee burning until we also have immediate execution.

evan-forbes:
> just for posterity: as we've discussed in other issues/channles, we cannot have fee burning until we also have immediate execution.

We don't actually need immediate execution, but we do need to ensure that each transaction that has fee burning (only PFBs in our case, can be executed properly when it gets included in a block. We are already doing a lot of this do our existing block validity rules, the only rules that we would have to add are:
- [ ] #979
- [ ] Signers of PFBs have enough funds to pay for the fee of their transaction
- [ ] Enforce a minimum fee

evan-forbes:
ref https://github.com/celestiaorg/celestia-app/issues/1971

PR #1985, opened 2023-06-22
feat!: run min gas pfb decorator in process proposal

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/1971

Yes, I'm aware that there's a lot of redundancy in the checks across the different stages in committing and executing a block. My plan is to start at a point of restrictiveness and then slowly remove redundancy rather than add more restrictiveness (which would be breaking)

### DIFF
--- a/x/blob/ante/ante.go
+++ b/x/blob/ante/ante.go
@@ -24,11 +24,8 @@ func NewMinGasPFBDecorator(k BlobKeeper) MinGasPFBDecorator {
 // AnteHandle implemnts the AnteHandler interface. It checks to see
 // if the transaction contains a MsgPayForBlobs and if so, checks that
 // the transaction has allocated enough gas.
-//
-// TODO: We need to run this antehandler in process proposal
-// however this will be a breaking change.
 func (d MinGasPFBDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
-	if !ctx.IsCheckTx() {
+	if ctx.IsReCheckTx() {
 		return next(ctx, tx, simulate)
 	}
 


x/blob/ante/ante.go
```diff
@@ -24,11 +24,8 @@ func NewMinGasPFBDecorator(k BlobKeeper) MinGasPFBDecorator {
 // AnteHandle implemnts the AnteHandler interface. It checks to see
 // if the transaction contains a MsgPayForBlobs and if so, checks that
 // the transaction has allocated enough gas.
-//
-// TODO: We need to run this antehandler in process proposal
-// however this will be a breaking change.
 func (d MinGasPFBDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
-	if !ctx.IsCheckTx() {
+	if ctx.IsReCheckTx() {
```
rootulp, line 28: [non blocking][question] how does modifying this line make the antehandler run in ProcessProposal?

cmwaters: ProcessProposal uses a context that has `isCheckTx` as false. Prior this would only run if the context was `CheckTx` now that this is removed it will run the handler in `ProcessProposal`. The reason for changing it to `IsReCheckTx` is that it doesn't make sense to re run this in `ReCheckTx` as nothing has changed

evan-forbes: > how does modifying this line make the antehandler run in ProcessProposal?

in #1984 we use the same antehandler for all things, which will then call the above handler as well if I'm not mistaken

evan-forbes:
I'm okay with merging this on the condition that we also get approval from leadership over adding a currently unplanned block validity rule.

we will need each pfb to be able to pay for gas should we enable fee burning or some sort of min fee, which was originally planned for #49, but has since been removed from mainnet.

after merging this, there will be an extra check that each PFB will have to have enough funds to pay for gas. While this is trivial for a validator to bypass by simply accepting 0 fee transactions (which is also a valid behavior), this is preparing us for whenever we do implement fee burning.

cc @musalbas @adlerjohn @liamsi

x/blob/ante/ante.go
rootulp: [no change needed][for my understanding] is there a downside to running this ante handler when `ctx.IsReCheckTx`? Is there a significant performance impact?

> The reason for changing it to IsReCheckTx is that it doesn't make sense to re run this in ReCheckTx as nothing has changed

To clarify, this ante handler is checking that the gas included the transaction exceeds the amount of gas expected to be consumed. Since it **does not** check that an account actually has the the amount of gas included in the transaction, this ante handler doesn't depend on state. Therefore the boolean result from this ante handler should be consistent before / after a new block is added even if that new block included a transaction that spent all of an account's funds.

evan-forbes: that's a good point @rootulp, but this ante handler works in tandem with the [`DeductFeeDecorator`](https://github.com/celestiaorg/cosmos-sdk/blob/dfc04d194ec23f99cd80e263a3cc75212e5cc1d0/x/auth/ante/fee.go#L104-L110) which runs before this one. So we actually first check if the account has enough funds, then we check that the fee is large enough for the PFB

rootulp: Does that mean the deduct fee decorator should be re-executed on ReCheckTx but the MinGasPFBDecorator in this PR shouldn't?

evan-forbes: perhpaps that's just prematurely optimal, but yeah

musalbas:
Is this a bug fix, or a new feature?

I feel it's important that we're disciplined with release candidates, and make sure that new release candidates only have bug fixes, and do not introduce new features that need to be additionally tested.

This looks like a trivial change - is there a reason that we cannot simply add this when we implement fee burning?

musalbas:
>My plan is to start at a point of restrictiveness and then slowly remove redundancy rather than add more restrictiveness (which would be breaking)

If there is an argument that this is actually a bug fix or security improvement due to the above - then I'm also fine with this being included, as that rationale makes sense.

cmwaters:
I would classify this as a bug fix but not an important one. Currently a validator could propose a block with a PFB that fills the entire block and instead of specifying the 20 million gas it would require to execute, only setting 100k. The tx would pass the validity checks in ProcessProposal but would run out of gas when it was executed in DeliverTx. However it doesn't matter if a PFB doesn't execute successfully, the data has already been published and the block proposer has got DA basically for free. The reason why this is not important is because regardless if the gas is correctly calculated or not, a user can always set a gas price of 0, and thus a fee of 0 (fee = gas * gas price) so long as the proposers min gas price that it will accept is 0 (default is non-zero). In other words you can get DA for free anyway. My understanding is that we will want to implement a proper and robust fee market relatively soon. Having this bug fixed will mean it's done now instead of later but given that a fee market implementation will likely be breaking in any case I think it's also fine if we punt on this change until v2.

evan-forbes (approved):
per sync discussion: :+1:

x/blob/ante/ante.go
cmwaters: > Is there a significant performance impact?

Probably not

PR #1993, opened 2023-06-27
[deprecated] test: genesis state tests for vesting accounts

### LINKED ISSUE #1806: EPIC: Pre-Genesis Tasks for Vesting Module
This epic serves as a centralized tracking mechanism for the essential tasks that must be completed before the genesis phase. The tasks primarily focus on thorough testing and validation of the vesting module and its associated functionalities.

Tasks:
- [x] Conduct comprehensive integration tests for the vesting module
  - [x] Develop test scenarios covering various types of vesting accounts

- [x] Design a suite of tests to cover different vesting schedules w.r.t. the allocation plans
  - [x] Delayed vesting accounts
  - [x] Periodic vesting accounts
  - [x] Continues vesting accounts

- [x] Implement testing of the genesis phase to test the following user actions
  - [x] Test delegations: the locked balance must be allowed to be delegated
  - [x] Test transfer: the locked balance is not allowed to be transferred elsewhere

mojtaba-esk:
closes #1806

## Overview
This PR is proposing a test suit to test the vesting accounts where introduced at genesis state. It tests the scenarios that we are intend to use for our mainnet.


app/test/vesting_test.go
```diff
@@ -0,0 +1,498 @@
...
+	assert.NoError(s.T(), err)
+
+	if mustSucceed {
+		assert.EqualValues(s.T(), 0, resQ.TxResult.Code, "the transfer TX must succeed")
+	} else {
+		assert.EqualValues(s.T(), 5, resQ.TxResult.Code, "the transfer TX must fail")
+	}
+}
+
+func (s *VestingModuleTestSuite) testDelegation(cctx testnode.Context, accName string) {
```
cmwaters, line 397: You could further test that vested accounts can claim rewards  for their delegations

app/test/vesting_test.go
```diff
@@ -0,0 +1,498 @@
...
+	del, err = GetAccountDelegations(cctx.GRPCClient, accAddress)
+	assert.NoError(s.T(), err)
+	assert.NotEmpty(s.T(), del, "delegations must not be empty")
+	assert.EqualValues(s.T(),
+		testfactory.BaseAccountDefaultBalance,
+		del[0].Balance.Amount.Int64(),
+		"delegation amount must match")
+}
+
+func AddAccountsToGenesisState(gs map[string]json.RawMessage, accounts ...authtypes.GenesisAccount) (map[string]json.RawMessage, error) {
```
cmwaters, line 431: I think I might have seen a similar function in `test/util` just fyi

app/test/vesting_test.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+}
+
+func (s *VestingModuleTestSuite) TestGenesisPeriodicVestingAccounts() {
+	// initial unlocked allocation to pay the gas fees
+	initBalanceForGasFee := sdk.NewCoin(app.BondDenom, sdk.NewInt(10))
+
+	startTime := tmtime.Now()
+	periods := vestingtypes.Periods{
+		vestingtypes.Period{Length: int64(10), Amount: sdk.Coins{sdk.NewInt64Coin(app.BondDenom, 2500)}},
```
cmwaters, line 205: Q: Is the length here in terms of blocks or time?

cmwaters (approved):
This looks good to me.

I'm surprised that there doesn't already exist tests in `cosmos/cosmos-sdk` that cover these cases.

app/test/vesting_test.go
mojtaba-esk: Good idea

app/test/vesting_test.go
mojtaba-esk: Thanks for the hint. I'll try to find that and use that one instead to avoid redundant code

app/test/vesting_test.go
mojtaba-esk: it is time in seconds

app/test/vesting_test.go
cmwaters: Up to you. We can also clean up all the test fixtures later

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+		_, err = cctx.WaitForTimestamp(startTime.Add(periods[i].Duration() + 10*time.Millisecond)) // Wait for the next period to be passed
+		assert.NoError(s.T(), err)
+	}
+
+	s.testTransferVestingAmount(cctx, accName, false)
+	s.testDelegation(cctx, accName)
+	s.testClaimDelegationReward(cctx, accName)
+}
+
+func (s *VestingModuleTestSuite) TestGenesisContinuesVestingAccounts() {
```
rootulp, line 296: I think it would be helpful to include a godoc comment above this test that explains the behavior it is verifying.

[question] Perhaps `Continues` meant to be `Continuous`?

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+		return gs
+	}
+
+	cctx, _, _ := s.NewNetworkWithGenesisOpt(kr, gsOpt)
+
+	assert.NoError(s.T(), cctx.WaitForNextBlock())
+
+	accAddress := getAddress(accName, cctx.Keyring).String()
+
+	// Test account details correctness
```
rootulp, line 334:
```suggestion
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+	assert.NoError(s.T(), err)
+
+	// Checking the queried account data
+	assert.Equal(s.T(), accAddress, qAcc.GetAddress().String(), "account address must match")
+	assert.EqualValues(s.T(), startTime.Unix(), qAcc.GetStartTime(), "start time must match")
+	assert.EqualValues(s.T(), endTime.Unix(), qAcc.GetEndTime(), "end time must match")
+	assert.EqualValues(s.T(),
+		testfactory.BaseAccountDefaultBalance,
+		qAcc.GetOriginalVesting().AmountOf(app.BondDenom).Int64(),
+		"original vesting must match")
```
rootulp, line 352:
```suggestion
		"original vesting amount must match")
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+	/*--------*/
```
rootulp, line 354:
```suggestion
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+	// Test available balance
```
rootulp, line 356: [nit] if available and spendable mean the same thing, proposal to use the same term.

[proposal] perhaps this comment can be replaced with a stand-alone test:

```diff
-// Test available balance
+s.testAvailableBalance(cctx, accName)
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+		assert.NoError(s.T(), err)
+
+		minExpectedSpendableBal := cva.GetVestedCoins(queryTime).Add(initBalanceForGasFee)
+
+		assert.LessOrEqual(s.T(),
+			balances.AmountOf(app.BondDenom).Int64(),
+			minExpectedSpendableBal.AmountOf(app.BondDenom).Int64(),
+		)
+
+		_, err = cctx.WaitForTimestamp(startTime.Add(5)) // Wait for a while
```
rootulp, line 370: [nit] IMO this line of code is more precise than the comment

```suggestion
		_, err = cctx.WaitForTimestamp(startTime.Add(5))
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+		assert.NoError(s.T(), err)
+	}
+	/*--------*/
```
rootulp, line 373:
```suggestion
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+	s.testTransferVestingAmount(cctx, accName, false)
+	s.testDelegation(cctx, accName)
+	s.testClaimDelegationReward(cctx, accName)
+}
+
+func (s *VestingModuleTestSuite) testTransferVestingAmount(cctx testnode.Context, accName string, mustSucceed bool) {
```
rootulp, line 380: [optional] this function would benefit from a godoc comment that explains that funds will be sent from `accName` to some unused account. Also it would help to clarify what `mustSucceed` is used for or rename it.

[optional] would it be beneficial for this function to accept an `amount` param instead of always trying to entire vesting balance?

[question] Are there scenarios where an account's vesting balance is partially vested? Do we want to verify scenarios where an account can send part of their vested balance?

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+	assert.NoError(s.T(), err)
+	assert.EqualValues(s.T(), 0, resQ.TxResult.Code, "the delegation TX must succeed")
+
+	// verify the delegations
+	del, err = GetAccountDelegations(cctx.GRPCClient, accAddress)
+	assert.NoError(s.T(), err)
+	assert.NotEmpty(s.T(), del, "delegations must not be empty")
+	assert.EqualValues(s.T(),
+		testfactory.BaseAccountDefaultBalance,
+		del[0].Balance.Amount.Int64(),
```
rootulp, line 431: [question] why does this only look at the first delegation response? Is only one response expected? If so, proposal to verify in the test via something like:

```go
	assert.Len(s.T(), del, 1)
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+	resQ, err := cctx.WaitForTx(resTx.TxHash, 10)
+	assert.NoError(s.T(), err)
+	assert.EqualValues(s.T(), 0, resQ.TxResult.Code, "the delegation TX must succeed")
+
+	// verify the delegations
+	del, err = GetAccountDelegations(cctx.GRPCClient, accAddress)
+	assert.NoError(s.T(), err)
+	assert.NotEmpty(s.T(), del, "delegations must not be empty")
+	assert.EqualValues(s.T(),
+		testfactory.BaseAccountDefaultBalance,
```
rootulp, line 430: [question] it's not immediately clear to me why the base account default balance should match the delegation amount.

Would it be helpful if this test file described it's own constant named `DelegationAmount` and used it here?

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+	assert.NotEmpty(s.T(), resR, "empty staking rewards")
+	assert.NotEmpty(s.T(), resR.Rewards)
+
+	rewardAmount := resR.Rewards[0].Reward.AmountOf(app.BondDenom).RoundInt().Int64()
+	assert.Greater(s.T(), rewardAmount, int64(0), "rewards must be more than zero")
+
+	balancesBefore, err := GetAccountSpendableBalance(cctx.GRPCClient, accAddress)
+	assert.NoError(s.T(), err)
+
+	// min is used as in the middle of the operation more tokens might be vested (unlocked)
```
rootulp, line 457: [optional] potentially clearer

```suggestion
	// minExpectedBalance is used because more tokens may be vested to the
	// account in the middle of this test
```

also can this variable be moved immediately above where it is used?

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+	return gs, nil
+}
+
+func GetValidators(grpcConn *grpc.ClientConn) (stakingtypes.Validators, error) {
+	scli := stakingtypes.NewQueryClient(grpcConn)
+	vres, err := scli.Validators(context.Background(), &stakingtypes.QueryValidatorsRequest{})
+
+	if vres == nil {
+		return stakingtypes.Validators{}, err
+	}
```
rootulp, line 523: [question] does this intentionally not check if `err != nil`? When I see 

```go
foo, err := someFunc()
```

I'm conditioned to expect

```go
foo, err := someFunc()
if err != nil {
		return stakingtypes.Validators{}, err
}

File: app/test/vesting_test.go [hunk 16]
```diff
@@ -0,0 +1,550 @@
...
+
+func GetAccountSpendableBalance(grpcConn *grpc.ClientConn, address string) (sdk.Coins, error) {
+	cli := banktypes.NewQueryClient(grpcConn)
+	res, err := cli.SpendableBalances(
+		context.Background(),
+		&banktypes.QuerySpendableBalancesRequest{
+			Address: address,
+		},
+	)
+	if err != nil || res == nil {
```
rootulp, line 546: [question] is the second part of this conditional necessary? If not, proposal to remove so that this follows the expected pattern

```suggestion
	if err != nil {
```

test/util/testnode/sign.go
```diff
@@ -20,7 +20,13 @@ func SignAndBroadcastTx(encCfg encoding.Config, c client.Context, account string
 			sdk.NewCoin(app.BondDenom, sdk.NewInt(1)),
 		)),
 	}
+	return SignAndBroadcastTxWithBuilderOption(opts, encCfg, c, account, msg...)
+}
 
+// SignAndBroadcastTxWithBuilderOption signs a transaction using the provided account and keyring
+// it accepts customized Tx builder options
+// inside the client.Context, then broadcasts it synchronously.
```
rootulp, line 28: [nit] rewrap + complete sentences

```suggestion
// SignAndBroadcastTxWithBuilderOption signs a transaction using the provided
// account and keyring. It accepts customized Tx builder options inside the
// client.Context, then broadcasts it synchronously.
```

rootulp (changes requested):

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+	cctx, _, _ := s.NewNetworkWithGenesisOpt(kr, gsOpt)
+
+	assert.NoError(s.T(), cctx.WaitForNextBlock())
+
+	for _, tt := range tests {
+		s.Run(tt.name, func() {
+			accAddress := getAddress(tt.accName, cctx.Keyring).String()
+
+			// Test account details correctness
```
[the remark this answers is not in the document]
mojtaba-esk: it tests if the given genesis data is read and processed correctly into the desired type of account.
They are implicitly tested in the lines below that, so I am ok with removing those.
@cmwaters and @evan-forbes do you see any benefit in keeping them?

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+			// Checking the queried account data
+			assert.Equal(s.T(), accAddress, qAcc.GetAddress().String(), "account address must match")
+			assert.EqualValues(s.T(), tt.endTime.Unix(), qAcc.GetEndTime(), "end time must match")
+
+			assert.EqualValues(s.T(),
+				testfactory.BaseAccountDefaultBalance,
+				qAcc.GetOriginalVesting().AmountOf(app.BondDenom).Int64(),
+				"original vesting must match")
+
+			/*--------*/
```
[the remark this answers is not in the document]
mojtaba-esk: Good idea, then I need to change the entire structure to create accounts and the genesis state in the `SetupSuite` method. Then all other tests can use it.
we might need to have separate test suites for each vesting account type or find a creative way to do it all in one suite

app/test/vesting_test.go
mojtaba-esk: Since some accounts like Periodic and Continues can have part of their balance locked (vesting), and it is hard to tell in advance how much, to test if locked balance can be delegated, the total initial vesting amount was set to delegate to cover it.
we need to explain it in comments

app/test/vesting_test.go
mojtaba-esk: I was not sure if the `Validators` method will return a non nil `err` if the `vres` is nil, so to avoid nil reference panic on `vres.Validators` if in any case `vres` is returned nil and err is also nil due to no validators found, I check it as well and return the `err` no matter what it is.
Do you think we should have an extra `if err != nil {...`  to have a better readability?

app/test/vesting_test.go
mojtaba-esk: it is to avoid panic on `res.GetBalances()` if `res` is nil. So if the address is not found, `res` will be nil.

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+			// Test available balance
+			balances, err := GetAccountSpendableBalance(cctx.GRPCClient, accAddress)
+			assert.NoError(s.T(), err)
+
+			expectedSpendableBal := int64(0)
+			if mustSucceed {
+				expectedSpendableBal = testfactory.BaseAccountDefaultBalance
```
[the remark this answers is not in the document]
mojtaba-esk: Since in delayed vesting account the locked is either all or none, here first zero amount is considered for spendable balance and then if the vesting amount is vested (unlocked), `testfactory.BaseAccountDefaultBalance` which is all vesting amount is assigned. 
yeah I agree with renaming some vars to add more clarity

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+	assert.EqualValues(s.T(),
+		testfactory.BaseAccountDefaultBalance,
+		qAcc.GetOriginalVesting().AmountOf(app.BondDenom).Int64(),
+		"original vesting must match")
+	assert.EqualValues(s.T(), periods, qAcc.VestingPeriods, "periods must match")
+
+	/*--------*/
+
+	// Test available balance
+	for i := 0; i < 2; i++ { // We just let one period to be unlocked to keep some locked amount to test other stuff
```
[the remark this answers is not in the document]
mojtaba-esk: Since I wanted to use one periodic vesting account to test multiple things, there is a confusion. Let me separate tests as you suggested to make thing clear

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+		currentLen := time.Since(startTime).Seconds()
+
+		balances, err := GetAccountSpendableBalance(cctx.GRPCClient, accAddress)
+		assert.NoError(s.T(), err)
+
+		expectedSpendableBal := initBalanceForGasFee.Amount.Int64()
+		passedLen := float64(0)
```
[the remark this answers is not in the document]
mojtaba-esk: yes renaming would help.
That `passedLen` iterates through periods of the account.

app/test/vesting_test.go
mojtaba-esk: Since the vested (unlocked) amount is transferable (Spendable), we do not test it. We only test if the vesting (locked) amount fails to transfer. So trying to transfer the entire allocation seems to be sufficient for the test purpose.

app/test/vesting_test.go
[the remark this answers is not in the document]
rootulp: Clarification: the suggestion is to remove the comment on line 126. I think lines 128-132 are fine but since they don't verify any details inside an account, I don't think the comment is necessary.

app/test/vesting_test.go
```diff
@@ -0,0 +1,542 @@
...
+		gs, err = AddAccountsToGenesisState(gs, vAccs...)
+		assert.NoError(s.T(), err)
+
+		gs, err = AddBalancesToGenesisState(gs, vBals...)
+		assert.NoError(s.T(), err)
+
+		return gs
+	}
+
+	cctx, _, _ := s.NewNetworkWithGenesisOpt(kr, gsOpt)
```
evan-forbes, line 118: [blocking]
minor, but this should be run during the setup of the suite instead of per test

app/test/vesting_test.go
```diff
@@ -0,0 +1,542 @@
...
+func (s *VestingModuleTestSuite) NewNetworkWithGenesisOpt(kr keyring.Keyring, genesisOpts ...testnode.GenesisOption) (cctx testnode.Context, rpcAddr, grpcAddr string) {
+	tmCfg := testnode.DefaultTendermintConfig()
+	tmCfg.Consensus.TargetHeightDuration = time.Millisecond * 1
+	appConf := testnode.DefaultAppConfig()
+
+	for _, name := range s.accounts {
+		testfactory.NewBaseAccount(kr, name)
+	}
+
+	cctx, rpcAddr, grpcAddr = testnode.NewNetwork(s.T(), testnode.DefaultParams(), tmCfg, appConf, []string{}, genesisOpts...)
```
evan-forbes, line 64: [blocking]
this has gotten updated recently, apologies for the refactor. this should probably look like this now
https://github.com/celestiaorg/celestia-app/blob/3f8c8a84ab2767dd5b0f5c22c7aab6653a788047/app/test/square_size_test.go#L46-L49

app/test/vesting_test.go
```diff
@@ -0,0 +1,542 @@
...
+
+	// set the accounts in the genesis state
+	authGenState.Accounts = append(authGenState.Accounts, pAccs...)
+	gs[authtypes.ModuleName] = encCfg.Codec.MustMarshalJSON(&authGenState)
+
+	return gs, nil
+}
+
+func AddBalancesToGenesisState(gs map[string]json.RawMessage, balances ...banktypes.Balance) (map[string]json.RawMessage, error) {
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
```
evan-forbes, line 496: [not blocking]
typically we don't creating an encoding config each time we call a function. tbh I don't think its really matters that much, so I'm definitely not blocking on it

app/test/vesting_test.go
```diff
@@ -0,0 +1,542 @@
...
+	assert.EqualValues(s.T(), 0, resQ.TxResult.Code, "the claim reward TX must succeed")
+
+	// Check if the reward amount in the account
+	balancesAfter, err := GetAccountSpendableBalance(cctx.GRPCClient, accAddress)
+	assert.NoError(s.T(), err)
+
+	assert.GreaterOrEqual(s.T(), balancesAfter.AmountOf(app.BondDenom).Int64(), minExpectedBalance, "Minimum balance after claiming reward")
+}
+
+func AddAccountsToGenesisState(gs map[string]json.RawMessage, accounts ...authtypes.GenesisAccount) (map[string]json.RawMessage, error) {
```
evan-forbes, line 475: [question]
is there ever a scenario where we want to add a balance or an account but not the other? if not, should we combine these to make sure that we don't accidently do that?

evan-forbes (commented):
great start!

app/test/vesting_test.go
```diff
@@ -0,0 +1,550 @@
...
+				testfactory.BaseAccountDefaultBalance,
+				qAcc.GetOriginalVesting().AmountOf(app.BondDenom).Int64(),
+				"original vesting must match")
+
+			/*--------*/
+
+			// Test the locking mechanism
+			// If the end time is already passed, the funds must be unlocked
+			// and we should be able to transfer some of it to another account
+			mustSucceed := tt.endTime.Before(time.Now())
```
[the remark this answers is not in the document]
mojtaba-esk: The entire structure is changed now. Thank you for your great suggestions.

app/test/vesting_test.go
mojtaba-esk: Done.

app/test/vesting_test.go
evan-forbes: does this portion of the refactor make sense @mojtaba-esk ? it won't look exactly like the above but similar. Essentially we can fit everything that need to change into that config

let me know if this doesn't make sense, happy to walk through it

mojtaba-esk: yes it does make the code simpler. I merged it into a copy of this branch and will open another PR shortly

rootulp:
closing in favor of https://github.com/celestiaorg/celestia-app/pull/2004

Issue #1994, opened 2023-06-27
Genesis-Utils: Create a tool to add accounts and generate the genesis file & test genesis file


mojtaba-esk:
We need a tool to assemble all accounts and details into a genesis file. 

The following account types are required:
- Regular accounts with liquid fund
- Delayed vesting accounts
- Periodic vesting accounts
- Continues vesting accounts

The requirements of the tool:
- Receive accounts' details in some forms of JSON or/and CSV
- Process the accounts and their balances
- Bake all of them into the genesis file
- Export the genesis file

UX design choice:
- a CLI tool
- with a web interface

evan-forbes:
what does the input look like?

mojtaba-esk:
Good question.
we get them in form of a bunch of CSV files, then I think we need to write some int. tools to convert them to JSON and I think JSON is a good format for input. We can follow the same structure of comos sdk for accounts in that JSON file.

evan-forbes:
we have a lot of test code that generate and edit genesis files and then spin up single validator networks over them, so starting their might be a good approach

https://github.com/celestiaorg/celestia-app/blob/b2fc98ccdea77fc00deaf56ddd4bd9e4258ca96d/test/util/testnode/full_node.go#L169-L205

rootulp:
Also worth noting that the celestia-appd binary has a `validate-genesis` command that may be useful here

Ref:
- celestia-app command gets added [here](https://github.com/celestiaorg/celestia-app/blob/b2fc98ccdea77fc00deaf56ddd4bd9e4258ca96d/cmd/celestia-appd/cmd/root.go#L168)
- celestiaorg/cosmos-sdk [here](https://github.com/celestiaorg/cosmos-sdk/blob/dfc04d194ec23f99cd80e263a3cc75212e5cc1d0/x/genutil/client/cli/validate_genesis.go#L17C1-L18)

mojtaba-esk:
**Checking for duplicate addresses**
After having a discussion [here](https://celestia-team.slack.com/archives/C04KS5R9VQU/p1691998367854779), it is decided to allow a single address be used in multiple allocation plans. e.g. user's can use one address for a regular (unlocked account) and one of the vesting accounts.
But to make sure we avoid accidental double import of a file, we can use hash file for the imported files.

mojtaba-esk:
Periodic vesting account is removed from the requirements as we do not have any to use. It can be simply added in future.

evan-forbes:
[...]

PR #2004, opened 2023-06-29
test: genesis state tests for vesting accounts

### LINKED ISSUE #1806: EPIC: Pre-Genesis Tasks for Vesting Module

mojtaba-esk:
closes #1806 

Overview

This PR is proposing a test suit to test the vesting accounts where introduced at genesis state. It tests the scenarios that we are intend to use for our mainnet.

**Note:** THIS PR is a copy of the PR #1993 as having the refactors of the main rebased into it to have easier review.

The suggested changes are applied to the tests.
Thank you @cmwaters and @evan-forbes and special thanks to @rootulp for his detailed points and for his great proposal to make the tests more fine grains and more special.

## Checklist


app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+
+	resQ, err := s.cctx.WaitForTx(resTx.TxHash, 10)
+	assert.NoError(s.T(), err)
+	assert.EqualValues(s.T(), 0, resQ.TxResult.Code, "the claim reward TX must succeed")
+
+	// Check if the reward amount in the account
+	balancesAfter, err := GetAccountSpendableBalance(s.cctx.GRPCClient, address)
+	assert.NoError(s.T(), err)
+
+	assert.GreaterOrEqual(s.T(), balancesAfter.AmountOf(app.BondDenom).Int64(), minExpectedBalance, "Minimum balance after claiming reward")
```
rootulp, line 545: [nit] I don't think the error message provides any additional information

```suggestion
	assert.GreaterOrEqual(s.T(), balancesAfter.AmountOf(app.BondDenom).Int64(), minExpectedBalance)
```

rootulp (changes requested):
Thanks for documenting some of the functions!

[optional] I think a helper function that gets a particular account (i.e. a vesting account that has some locked balance) may be helpful so that each test doesn't need a `for {` loop. 

A few blocking comments

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+	tmtime "github.com/tendermint/tendermint/types/time"
+	"google.golang.org/grpc"
+)
+
+const (
+	totalAccountsPerType = 300
+	initBalanceForGasFee = 10
+)
+
+type _accountDispenser struct {
```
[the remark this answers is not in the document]
rootulp: Thanks for explaining! In Go any field that starts with a lowercase letter is unexported so I don't think the leading `_` is necessary or desirable.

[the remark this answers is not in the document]
evan-forbes: agree with rootul here, unless we're generating go code or have some extra reason, I think we should stick to standard go practices since that's what the rest of the repo history does :slightly_smiling_face:

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+			expectedSpendableCoins.AmountOf(app.BondDenom).Int64(),
+			balances.AmountOf(app.BondDenom).Int64(),
+			"spendable balance must match")
+
+		// Stop testing once we hit an account with no spendable balance
+		if vestedCoins.IsZero() {
+			break
+		}
+		s.T().Log("waiting 3 seconds...")
+		time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
evan-forbes: agree with this comment, if we want to wait for a specific action, like a new block or something, then ideally we should aim to do that instead

also, is it possible for this test to only fail after reaching the timeout? if so, is it possible to constrain the failure case?

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+		err = vAcc.Unmarshal(resAccBytes)
+		assert.NoError(s.T(), err)
+
+		if vAcc.StartTime > tmtime.Now().Unix()+20 {
+			s.testDelegatingVestingAmount(name)
+			s.testClaimDelegationReward(name)
+			return
+		}
+	}
+}
```
evan-forbes, line 436: what are the differences between each of these tests? is there overlap between what is being tested? if so, is there a way to remove any overlap and reduce the cognitive load on readers?

are we testing everything in `TestGenesisContinuousVestingAccountsClaimDelegationRewards`?

evan-forbes (commented):
thanks for all of this @mojtaba-esk! this is overally really good!!

it's quite extensive, in that it appears to be thorough (which is really good!! :slightly_smiling_face:), but perhaps at the cost of cognitive load of readers. Admittedly, despite me being familar with how the vesting module works, it took a long time to understand what was occuring with these tests, which ideally (might not be possible!) should typically try to avoid since if they break we have to debug them.

there is also widespread usage of looping continuously until some condition is met, and I think there are ways to avoid that that might also reduce some of the issue described above

evan-forbes:
also, its difficult to tell what test failed or for why due to the stack dump, but it appears that one of the tests [here](https://github.com/celestiaorg/celestia-app/actions/runs/5414084334/jobs/9902199939?pr=2004) timed out?

```
panic: test timed out after 10m0s
running tests:
	TestVestingModule (1m27s)
	TestVestingModule/TestGenesisPeriodicVestingAccountsDelegationPartiallyVested (26s)
```

evan-forbes:
on a side note, I think we can run tests like this (and the standard sdk one) only in the nightly #2019

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+	accounts    sync.Map // map[_accountType]_accountDispenser
+	accountsMut sync.Mutex
+
+	kr   keyring.Keyring
+	cctx testnode.Context
+	ecfg encoding.Config
+}
+
+func TestVestingModule(t *testing.T) {
+	suite.Run(t, new(VestingModuleTestSuite))
```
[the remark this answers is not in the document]
mojtaba-esk: Agreed let's do it.
It takes some time because the vesting accounts are time sensitive and to assert some particular unlocking schemes, the test code waits for a number of seconds.

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+	genOpts = append(genOpts, s.initPeriodicVestingAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initContinuousVestingAccounts(totalAccountsPerType))
+
+	s.startNewNetworkWithGenesisOpt(genOpts...)
+}
+
+// NewNetworkWithGenesisOpt creates a new test network with the specified genesis options for the VestingModuleTestSuite.
+// It initializes a default Tendermint configuration (tmCfg) and default application configuration (appConf).
+// The target block time is set to 1 millisecond. It applies the given genesis options.
+// The function returns the created client context (cctx).
```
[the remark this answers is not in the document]
mojtaba-esk: thanks for pointing it out, the comment is outdated. will update it

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+		assert.EqualValues(s.T(),
+			expectedSpendableBal,
+			balances.AmountOf(app.BondDenom).Int64(),
+			"spendable balance must match")
+
+		// Continue testing until find an account with vesting (locked) balance
+		// because we want to test both vested & vesting accounts
+		if !alreadyVested {
+			break
+		}
```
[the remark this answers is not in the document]
mojtaba-esk: what do you mean by unvested? is it an account that its unlocked time is already passed or an account which is not vesting account?

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+		vestedCoins := vAcc.GetVestedCoins(tmtime.Now())
+		maxExpectedSpendableBalCoins := vestedCoins.Add(initCoinsForGasFee)
+		assert.LessOrEqual(s.T(),
+			balances.AmountOf(app.BondDenom).Int64(),
+			maxExpectedSpendableBalCoins.AmountOf(app.BondDenom).Int64())
+
+		// Stop testing once we hit an account with no spendable balance
+		if vestedCoins.IsZero() {
+			break
+		}
```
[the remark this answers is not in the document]
mojtaba-esk: to address that, the vesting accounts are generated using `startTime := tmtime.Now()` and it will increase by 5 seconds for each account. 
Since it takes a couple of seconds to start the network, there will always be some account with spendable balance.

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+// Parameters:
+// - count: The number of regular accounts to generate.
+//
+// Returns:
+// A GenesisOption function that takes the current genesis state as input, adds the generated accounts
+// and balances to it, and returns the modified genesis state.
+func (s *VestingModuleTestSuite) initContinuousVestingAccounts(count int) testnode.GenesisOption {
+	names := testfactory.GenerateAccounts(count)
+	s.accounts.Store(ContinuousVestingAccountType, _accountDispenser{names: names})
+	bAccounts, balances := generateBaseAccounts(s.kr, names)
```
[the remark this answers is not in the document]
mojtaba-esk: True. Let me see how we can do it efficiently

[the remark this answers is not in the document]
mojtaba-esk: Done

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+	bankGenState.Balances = append(bankGenState.Balances, balances...)
+	gs[banktypes.ModuleName] = encCfg.Codec.MustMarshalJSON(&bankGenState)
+
+	return gs, nil
+}
+
+// GetValidators retrieves the validators from the staking module using the provided gRPC client connection (grpcConn).
+// It takes a gRPC client connection (grpcConn) as input.
+// Then, it returns the validators from the response.
+func GetValidators(grpcConn *grpc.ClientConn) (stakingtypes.Validators, error) {
```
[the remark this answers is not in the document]
mojtaba-esk: Good idea, I moved them to testfactory/utils

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+	msgDelg := stakingtypes.NewMsgDelegate(
+		getAddress(name, s.cctx.Keyring),
+		validators[0].GetOperator(),
+		sdk.NewCoin(app.BondDenom, sdk.NewInt(testfactory.BaseAccountDefaultBalance)),
+	)
+	resTx, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, name, []sdk.Msg{msgDelg}...)
+	assert.NoError(s.T(), err)
+
+	resQ, err := s.cctx.WaitForTx(resTx.TxHash, 10)
+	assert.NoError(s.T(), err)
```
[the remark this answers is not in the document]
mojtaba-esk: I think it depends to the error. For example here, if there is an error usually means there was a problem in including the tx in the block or somehow the tx is not found. So I think the test should fail to investigate the root of the issue. what do you think about this situation?

app/test/vesting_test.go
mojtaba-esk: We can certainly move the overlapped code to a new function

mojtaba-esk:
> also, its difficult to tell what test failed or for why due to the stack dump, but it appears that one of the tests [here](https://github.com/celestiaorg/celestia-app/actions/runs/5414084334/jobs/9902199939?pr=2004) timed out?
> 
> ```
> panic: test timed out after 10m0s
> running tests:
> 	TestVestingModule (1m27s)
> 	TestVestingModule/TestGenesisPeriodicVestingAccountsDelegationPartiallyVested (26s)
> ```

10 minutes is too much. Since the tests are time sensitive, we have waits here and there in the code. We might be able to find a solution that does not require waits.

app/test/vesting_test.go
mojtaba-esk: Done

app/test/vesting_test.go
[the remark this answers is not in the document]
mojtaba-esk: True, done.

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+}
+
+// This test function tests delegation of a periodic vesting account that
+// has part of its allocation unlocked and part of it locked
+func (s *VestingModuleTestSuite) TestGenesisPeriodicVestingAccountsDelegationPartiallyVested() {
+	assert.NoError(s.T(), s.cctx.WaitForNextBlock())
+
+	// find and test a vesting account that has some vesting (locked) and
+	// some vested (unlocked) balance
+	for {
```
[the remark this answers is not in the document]
mojtaba-esk: Since in these tests, vesting accounts are created ahead of time, to test genesis, and they are time triggered, this is the only solution I could think of. If we can find a solution to avoid it, it will be great.

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+		assert.LessOrEqual(s.T(),
+			balances.AmountOf(app.BondDenom).Int64(),
+			maxExpectedSpendableBalCoins.AmountOf(app.BondDenom).Int64())
+
+		// Stop testing once we hit an account with no spendable balance
+		if vestedCoins.IsZero() {
+			break
+		}
+		s.T().Log("waiting 3 seconds...")
+		time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
mojtaba-esk: honestly I do not remember the exact reason. without it, I think it broke.

test/util/testfactory/utils.go
```diff
@@ -3,23 +3,31 @@ package testfactory
...
 
 const (
 	// nolint:lll
-	TestAccName  = "test-account"
-	TestAccMnemo = `ramp soldier connect gadget domain mutual staff unusual first midnight iron good deputy wage vehicle mutual spike unlock rocket delay hundred script tumble choose`
-	bondDenom    = "utia"
+	TestAccName               = "test-account"
+	TestAccMnemo              = `ramp soldier connect gadget domain mutual staff unusual first midnight iron good deputy wage vehicle mutual spike unlock rocket delay hundred script tumble choose`
+	bondDenom                 = "utia"
+	BaseAccountDefaultBalance = 10000
```
evan-forbes, line 33: why not use something large like in https://github.com/celestiaorg/celestia-app/blob/b5be4486cf1a62921ce4a329ed0305ee48d1316d/test/util/testfactory/utils.go#L89

is there a way to just use that function? or just use the standard mechanism in testnode? (where we create funded genesis accounts from the `testnode.Config.Accounts`?

https://github.com/celestiaorg/celestia-app/blob/b5be4486cf1a62921ce4a329ed0305ee48d1316d/test/util/testnode/full_node.go#L165

app/test/vesting_test.go
```diff
@@ -0,0 +1,696 @@
...
+func (s *VestingModuleTestSuite) initDelayedVestingAccounts(count int) testnode.GenesisOption {
+	initCoinsForGasFee := sdk.NewCoin(app.BondDenom, sdk.NewInt(initBalanceForGasFee))
+
+	names := testfactory.GenerateAccounts(count)
+	s.accounts.Store(DelayedVestingAccountType, accountDispenser{names: names})
+	vAccounts, balances := testfactory.GenerateDelayedVestingAccounts(s.kr, names, initCoinsForGasFee)
+	gAccounts := authtypes.GenesisAccounts{}
+	for i := range vAccounts {
+		gAccounts = append(gAccounts, vAccounts[i])
+	}
```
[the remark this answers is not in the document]
mojtaba-esk: it does not allow for this because the types are different. The code in the loop casts a vesting account type into a genesis account type and do them in batch with `...` did not work.

test/util/testfactory/utils.go
mojtaba-esk: We need to have a constant somewhere that can reference to it in our tests. something like `BaseAccountDefaultBalance` and it can be any value. We can change it to `99999999999999999`. We just need it to evaluate the balance for various tests.

mojtaba-esk: If we slightly change the standard mechanism function in testnode to adopt that constant, we can use it widely.

test/util/testfactory/utils.go
```diff
@@ -102,3 +110,282 @@ func GenerateAccounts(count int) []string {
...
+// Returns:
+// - []*vestingtypes.DelayedVestingAccount: The generated delayed vesting accounts.
+// - []banktypes.Balance: The balances of the accounts.
+func GenerateDelayedVestingAccounts(kr keyring.Keyring, names []string, initUnlockedCoins ...sdk.Coin) ([]*vestingtypes.DelayedVestingAccount, []banktypes.Balance) {
+	bAccounts, balances := GenerateBaseAccounts(kr, names, initUnlockedCoins...)
+	vAccounts := []*vestingtypes.DelayedVestingAccount{}
+
+	endTime := tmtime.Now().Add(-2 * time.Second)
+	for i := range bAccounts {
+		bal := FindBalanceByAddress(balances, bAccounts[i].GetAddress().String())
```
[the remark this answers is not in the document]
mojtaba-esk: I think we can omit it, after all they are test account and they are in sequence.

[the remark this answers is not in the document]
mojtaba-esk: done

go.mod
```diff
@@ -198,7 +198,8 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13
+	// github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13.0.20230725094623-1a1ad2fa875d
```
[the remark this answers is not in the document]
mojtaba-esk: yeah that was the query thing, now I reverted back to the code you proposed.

app/test/vesting_test.go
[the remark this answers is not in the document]
evan-forbes: ahh I see cool, do you mind if we add something that makes that more explicit then? 
```suggestion
	for i := range vAccounts {
		gAccounts = append(gAccounts, authtypes.GenesisAccount(vAccounts[i]))
	}
```

or perhaps since we're doing this conversion multiple times we can have a simple helper function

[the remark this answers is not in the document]
mojtaba-esk: Good idea. will do it

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+		getAddress(randomAcc, s.cctx.Keyring),
+		sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(vestingAmount))), // try to transfer the locked amount
+	)
+	resTx, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, name, []sdk.Msg{msgSend}...)
+	assert.NoError(s.T(), err)
+
+	resQ, err := s.cctx.WaitForTx(resTx.TxHash, 10)
+	assert.NoError(s.T(), err)
+
+	assert.EqualValues(s.T(), 5, resQ.TxResult.Code, "the transfer TX must fail")
```
rootulp, line 364: [optional]

```suggestion
	assert.EqualValues(s.T(), errors.ErrInsufficientFunds.ABCICode(), resQ.TxResult.Code, "the transfer TX must fail")
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,936 @@
...
+			return
+		}
+	}
+}
+
+func (s *VestingModuleTestSuite) TestGenesisDelayedVestingAccountsClaimDelegationRewards() {
+	assert.NoError(s.T(), s.cctx.WaitForNextBlock())
+
+	// find and test a vesting account with endTime which is
+	// 	at least 20 seconds away time from now
```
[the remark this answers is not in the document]
rootulp: [not blocking] do we want to extract a constant for time to wait per transaction (10 seconds). Can do in this PR or a follow-up issue.

app/test/vesting_test.go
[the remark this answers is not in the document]
rootulp: [not blocking] I think we should use `require` in places where the rest of the test should fail. This location seems like it should use `require`.

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+		getAddress(name, s.cctx.Keyring),
+		validators[0].GetOperator(),
+		sdk.NewCoin(app.BondDenom, sdk.NewInt(vestingAmount)),
+	)
+	resTx, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, name, []sdk.Msg{msgDelg}...)
+	assert.NoError(s.T(), err)
+
+	resQ, err := s.cctx.WaitForTx(resTx.TxHash, 10)
+	assert.NoError(s.T(), err)
+	assert.EqualValues(s.T(), 0, resQ.TxResult.Code, fmt.Sprintf("the delegation TX must succeed: \n%s", resQ.TxResult.String()))
```
rootulp, line 392: [nit] I think it's clearer to use the abci status codes

```suggestion
	assert.EqualValues(s.T(), abci.CodeTypeOK, resQ.TxResult.Code, fmt.Sprintf("the delegation TX must succeed: \n%s", resQ.TxResult.String()))
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+	msg := distributiontypes.NewMsgWithdrawDelegatorReward(
+		getAddress(name, s.cctx.Keyring),
+		validators[0].GetOperator(),
+	)
+	resTx, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, name, []sdk.Msg{msg}...)
+	assert.NoError(s.T(), err)
+
+	resQ, err := s.cctx.WaitForTx(resTx.TxHash, 10)
+	assert.NoError(s.T(), err)
+	assert.EqualValues(s.T(), 0, resQ.TxResult.Code, fmt.Sprintf("the claim reward TX must succeed: \n%s", resQ.TxResult.String()))
```
rootulp, line 446:
```suggestion
	assert.EqualValues(s.T(), abci.CodeTypeOK, resQ.TxResult.Code, fmt.Sprintf("the claim reward TX must succeed: \n%s", resQ.TxResult.String()))
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+}
+
+type accountType int
+
+const (
+	RegularAccountType accountType = iota + 1
+	DelayedVestingAccountType
+	PeriodicVestingAccountType
+	ContinuousVestingAccountType
+)
```
rootulp, line 45: [optional] since this test file is so large, I wonder if it would make sense to have one file per type of account under test

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+func (s *VestingModuleTestSuite) TestGenesisPeriodicVestingAccountsSpendableBalancePartiallyUnlocked() {
+	// Find a periodic vesting account that has some vested (unlocked) balance (its start time has already passed)
+	vAcc, name, err := s.getAnUnusedPeriodicVestingAccount(tmtime.Now().Unix() - 5)
+	assert.NoError(s.T(), err)
+	address := getAddress(name, s.cctx.Keyring).String()
+
+	// Since we want a partially unlocked balance we need to wait until
+	// the first period has passed if not already
+	for vAcc.GetVestedCoins(tmtime.Now()).IsZero() {
+		time.Sleep(time.Second)
```
rootulp, line 180: [optional] instead of forcing this test to wait for some unknown period of time, I wonder if it's possible to construct the testing accounts in such a way that this test always finds a partially vested periodic account and if it doesn't, it errors immediately.

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+	// find an unlocked delayed vesting account.
+	vAcc, name, err := s.getAnUnusedDelayedVestingAccount(0)
+	assert.NoError(s.T(), err)
+
+	address := getAddress(name, s.cctx.Keyring).String()
+	balances, err := testfactory.GetAccountSpendableBalanceByBlock(s.cctx.GRPCClient, address, block)
+	assert.NoError(s.T(), err)
+
+	expectedSpendableBal := vAcc.GetVestedCoins(block.Time).AmountOf(app.BondDenom).Int64() + initBalanceForGasFee
+	assert.EqualValues(s.T(),
```
rootulp, line 111: [question] should this additionally verify that expectedSpendableBal > initBalanceForGasFee to avoid the case mentioned above:

> [question] It seem possible for getAnUnusedDelayedVestingAccount to return a delayed vesting account whose start time and end time is in the future. If that's the case, this test doesn't verify that any tokens are unlocked 🤔

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+	// Find a periodic vesting account that that is currently in a vesting (locked) state
+	// i.e. its start time yet to be reached.
+	vAcc, name, err := s.getAnUnusedPeriodicVestingAccount(block.Time.Unix() + 10)
+	assert.NoError(s.T(), err)
+	address := getAddress(name, s.cctx.Keyring).String()
+
+	balances, err := testfactory.GetAccountSpendableBalanceByBlock(s.cctx.GRPCClient, address, block)
+	assert.NoError(s.T(), err)
+
+	expectedSpendableBal := vAcc.GetVestedCoins(block.Time).AmountOf(app.BondDenom).Int64() + initBalanceForGasFee
```
rootulp, line 211: [question] if the balance is locked, is expectedSpendableBal == initBalanceForGasFee ? Should this test add that as an assertion?

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+
+	// find a continuous vesting account with locked balance
+	vAcc, name, err := s.getAnUnusedContinuousVestingAccount(block.Time.Unix() + 10)
+	assert.NoError(s.T(), err)
+	address := getAddress(name, s.cctx.Keyring).String()
+
+	balances, err := testfactory.GetAccountSpendableBalanceByBlock(s.cctx.GRPCClient, address, block)
+	assert.NoError(s.T(), err)
+
+	expectedSpendableBalAmount := vAcc.GetVestedCoins(block.Time).AmountOf(app.BondDenom).Int64() + initBalanceForGasFee
```
rootulp, line 272: [question] should this additionally verify that expectedSpendableBalAmount == initBalanceForGasFee or put another way that 0 coins have unlocked?

app/test/vesting_test.go
rootulp: On second thought, there is so much similar code across tests that it may not make sense to split them up

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+	return func(gs map[string]json.RawMessage) map[string]json.RawMessage {
+		gs, err := AddGenesisAccountsWithBalancesToGenesisState(gs, gAccounts, balances)
+		assert.NoError(s.T(), err)
+		return gs
+	}
+}
+
+// initDelayedVestingAccounts initializes delayed vesting accounts for the VestingModuleTestSuite.
+// It generates the specified number of account names and stores them in the accounts map with
+// RegularAccountType as the key.
```
rootulp, line 480: [blocking]

```suggestion
// DelayedVestingAccountType as the key.
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+	case DelayedVestingAccountType:
+		return "delayed vesting"
+	case PeriodicVestingAccountType:
+		return "periodic vesting"
+	case ContinuousVestingAccountType:
+		return "continuous vesting"
+	default:
+		return "unknown"
+	}
+}
```
rootulp, line 652: [optional] should this be moved to the top of the file so that it is immediately below the location where accountType is defined?

go.mod
```diff
@@ -198,7 +198,9 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13
+	// SDK with blocktime in GRPC query: https://github.com/celestiaorg/cosmos-sdk/pull/337
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13.0.20230731181258-40cbd617f1ba
```
rootulp, line 202: [blocking] thanks for the fix in celestiaorg/cosmos-sdk! If possible, I think we should hold off on merging this PR until we get that PR merged and cut an official release of celestiaorg/cosmos-sdk

app/test/vesting_test.go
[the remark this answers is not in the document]
mojtaba-esk: we can have it. It is not actually wait for tx, it is just a safe time boundary to make sure when the tx is executed the targeted vesting account remains in the desired status e.g. locked, partially locked, etc.
but im ok with having some constant

app/test/vesting_test.go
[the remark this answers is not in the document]
mojtaba-esk: fixed in the new commit

app/test/vesting_test.go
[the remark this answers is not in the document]
mojtaba-esk: this does not exist anymore, closing it.

app/test/vesting_test.go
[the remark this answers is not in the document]
mojtaba-esk: fixed. Closing it

app/test/vesting_test.go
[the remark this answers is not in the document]
mojtaba-esk: you're right. Applying it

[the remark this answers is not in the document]
mojtaba-esk: done

test/util/testfactory/utils.go
```diff
@@ -102,3 +110,282 @@ func GenerateAccounts(count int) []string {
...
+// Returns:
+// - bAccounts: A slice of authtypes.BaseAccount representing the generated base accounts.
+// - balances: A slice of banktypes.Balance representing the balances of the generated accounts.
+func GenerateBaseAccounts(kr keyring.Keyring, names []string, initExtraCoins ...sdk.Coin) ([]authtypes.BaseAccount, []banktypes.Balance) {
+	bAccounts := []authtypes.BaseAccount{}
+	balances := []banktypes.Balance{}
+	for _, name := range names {
+		acc, coins := NewBaseAccount(kr, name)
+		bAccounts = append(bAccounts, *acc)
+		balances = append(balances, banktypes.Balance{
```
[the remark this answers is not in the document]
mojtaba-esk: from what I saw in the code. even if we can do that, the balance.amountOf stops when finds its first instance, but some other functions might make a summation of those. not investigated that part

test/util/testfactory/utils.go
```diff
@@ -102,3 +110,282 @@ func GenerateAccounts(count int) []string {
...
+	bAccounts, balances := GenerateBaseAccounts(kr, names, initUnlockedCoins...)
+	vAccounts := []*vestingtypes.ContinuousVestingAccount{}
+
+	startTime := tmtime.Now()
+
+	for i := range bAccounts {
+		bal := FindBalanceByAddress(balances, bAccounts[i].GetAddress().String())
+		coins := bal.Coins.Sub(initUnlockedCoins...)
+
+		endTime := startTime.Add(10 * time.Second)
```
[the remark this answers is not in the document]
mojtaba-esk: done

app/test/vesting_test.go
```diff
@@ -0,0 +1,696 @@
...
+	for i := range bAccounts {
+		gAccounts = append(gAccounts, &bAccounts[i])
+	}
+
+	return func(gs map[string]json.RawMessage) map[string]json.RawMessage {
+		gs, err := AddGenesisAccountsWithBalancesToGenesisState(gs, gAccounts, balances)
+		assert.NoError(s.T(), err)
+		return gs
+	}
+}
```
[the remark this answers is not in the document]
mojtaba-esk: as said we have no control on the amount of funds and we need to know the exact amount for some evaluation, so we either need to use an exported constant for the default funding or stick to the current approach

test/util/testfactory/utils.go
```diff
@@ -102,3 +110,282 @@ func GenerateAccounts(count int) []string {
...
+// It takes a gRPC client connection (grpcConn) and the account address (address) as inputs.
+// If no account found or an error occurs, it returns nil and the error.
+// Otherwise, it returns the value field of the account in the response, which represents the raw account information.
+//
+// Parameters:
+// - grpcConn: A gRPC client connection.
+// - address: The account address to retrieve the raw information for.
+//
+// Returns:
+// The raw account information as a byte slice, or nil and an error if the retrieval fails.
```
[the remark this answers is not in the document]
mojtaba-esk: made them more concise

test/util/testfactory/utils.go
```diff
@@ -102,3 +110,282 @@ func GenerateAccounts(count int) []string {
...
+
+	startTime := tmtime.Now()
+	for i := range bAccounts {
+		bal := FindBalanceByAddress(balances, bAccounts[i].GetAddress().String())
+		coins := bal.Coins.Sub(initUnlockedCoins...)
+		acc := vestingtypes.NewPeriodicVestingAccount(&bAccounts[i], coins, startTime.Unix(), periods)
+		vAccounts = append(vAccounts, acc)
+
+		// the startTime is increased for each account to be able to test various scenarios
+		startTime = startTime.Add(5 * time.Second)
```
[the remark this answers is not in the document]
mojtaba-esk: ok sure. 
We have a list of vesting accounts of type X (e.g. periodic). we need to have them baked into the genesis file, and the tests are time sensitive. that means at any point in time, we need to have some accounts that are already vested (unlocked) and some are vesting and for some types some with partially unlocked. So to have such thing, we make a list of accounts whose unlock time is increasing by for example 5 seconds. 
that means if for example we want to have an account with unlocked balance, we can find an account which has 
`startTime<CurrentTime` and if we want to have a locked account we can find an account with `startTime>CurrentTime+10` which will be unlocked in 10 seconds from now.
so we can test all possible scenarios this way.

test/util/testfactory/utils.go
```diff
@@ -102,3 +110,282 @@ func GenerateAccounts(count int) []string {
...
+	if err != nil || res == nil {
+		return nil, 0, err
+	}
+	return res.GetBalances(), res.QueryTime, nil
+}
+
+// GetRawAccountInfo retrieves the raw account information for the specified address using gRPC.
+// It takes a gRPC client connection (grpcConn) and the account address (address) as inputs.
+// If no account found or an error occurs, it returns nil and the error.
+// Otherwise, it returns the value field of the account in the response, which represents the raw account information.
```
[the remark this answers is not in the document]
mojtaba-esk: removed a lot of unnecessary comments as their function signature are descriptive enough.

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+
+func (s *VestingModuleTestSuite) TestGenesisDelayedVestingAccountsSpendableBalanceUnlocked() {
+	block, err := s.cctx.LatestBlock()
+	assert.NoError(s.T(), err)
+	// We need to wait for a block because sometimes querying based on the
+	// latest height throws an SDK error saying the height is in the future
+	assert.NoError(s.T(), s.cctx.WaitForNextBlock())
+
+	// find an unlocked delayed vesting account.
+	vAcc, name, err := s.getAnUnusedDelayedVestingAccount(0)
```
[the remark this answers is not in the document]
mojtaba-esk: Un used means that the tests have not used that account yet and it is untouched so its balance and schedule is as expected. Unlocked means their fund is unlocked i.e. their vesting time is over or partially over.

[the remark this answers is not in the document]
mojtaba-esk: yes it is possible, we need to add a wait here if the endtime is not yet passed

app/test/vesting_test.go
```diff
@@ -0,0 +1,674 @@
...
+	return func(gs map[string]json.RawMessage) map[string]json.RawMessage {
+		gs, err := AddGenesisAccountsWithBalancesToGenesisState(gs, gAccounts, balances)
+		assert.NoError(s.T(), err)
+		return gs
+	}
+}
+
+// unusedAccount returns an unused account name of the specified account type
+// for the VestingModuleTestSuite. If the account type is not found, it panics
+// with an error message.
```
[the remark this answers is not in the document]
mojtaba-esk: yeah it used to panic :D

app/test/vesting_test.go
mojtaba-esk: copy paste issue :D

mojtaba-esk: thanks for catching it

app/test/vesting_test.go
rootulp: Yes we can assert `vAcc.GetVestedCoins(block.Time).AmountOf(app.BondDenom).Int64()` is 0.

app/test/vesting_test.go
```diff
@@ -245,68 +219,48 @@ func (s *VestingModuleTestSuite) TestGenesisPeriodicVestingAccountsDelegationPar
...
 }
 
 func (s *VestingModuleTestSuite) TestGenesisContinuousVestingAccountsDelegation() {
-	assert.NoError(s.T(), s.cctx.WaitForNextBlock())
+	require.NoError(s.T(), s.cctx.WaitForNextBlock())
 
 	// find and test a vesting account that has some vesting (locked) balance
-	_, name, err := s.getAnUnusedContinuousVestingAccount(tmtime.Now().Unix() + 10)
+	_, name, err := s.getAnUnusedContinuousVestingAccount(tmtime.Now().Unix() + vestingDelayPerTx)
 	assert.NoError(s.T(), err)
```
rootulp, line 264:
```suggestion
	require.NoError(s.T(), err)
```

app/test/vesting_test.go
```diff
@@ -330,38 +284,57 @@ func (s *VestingModuleTestSuite) TestGenesisContinuousVestingAccountsDelegationP
...
+	assert.EqualValues(s.T(), sdkerrors.ErrInsufficientFunds.ABCICode(), txResultCode, "the transfer TX must fail")
+}
+
+// testTransferMustSucceed tests the transfer of a certain amount of funds from one account
+// to another. It asserts that the result code of the transaction is equal to 0, indicating
+// a success.
+func (s *VestingModuleTestSuite) testTransferMustSucceed(name string, amount int64) {
+	txResultCode, err := s.submitTransferTx(name, amount)
+	assert.NoError(s.T(), err)
+	assert.EqualValues(s.T(), 0, txResultCode, "the transfer TX must succeed")
```
rootulp, line 313:
```suggestion
	assert.EqualValues(s.T(), abci.CodeTypeOK, txResultCode, "the transfer TX must succeed")
```

app/test/vesting_test.go
```diff
@@ -545,8 +518,9 @@ func (s *VestingModuleTestSuite) initContinuousVestingAccounts(count int) testno
 }
 
 // unusedAccount returns an unused account name of the specified account type
-// for the VestingModuleTestSuite. If the account type is not found, it panics
-// with an error message.
+// unusedAccount returns an unused account name of the specified account type
```
rootulp, line 521: This line is duplicated

```suggestion
```

go.mod
```diff
@@ -198,9 +198,7 @@ require (
 )
 
 replace (
-	// SDK with blocktime in GRPC query: https://github.com/celestiaorg/cosmos-sdk/pull/337
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13.0.20230731181258-40cbd617f1ba
-	// github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.15.0-sdk-v0.46.13
```
rootulp: [question] is this blocked on a release with https://github.com/celestiaorg/cosmos-sdk/pull/337 ?

test/util/testfactory/utils.go
```diff
@@ -132,22 +132,6 @@ func NewBaseAccount(kr keyring.Keyring, name string) (*authtypes.BaseAccount, sd
...
-//
-// Parameters:
-// - kr: A keyring.Keyring instance used to create base accounts.
-// - names: A slice of account names.
-// - initExtraCoins: A slice of sdk.Coins representing the extra coins to be added to
-// the base accounts which can be used for example for gas fees.
-//
-// Returns:
-// - bAccounts: A slice of authtypes.BaseAccount representing the generated base accounts.
-// - balances: A slice of banktypes.Balance representing the balances of the generated accounts.
```
rootulp: [not blocking][question] why were all the comments in this file removed?

test/util/testfactory/utils.go
```diff
@@ -346,14 +289,15 @@ func GetAccountSpendableBalance(grpcConn *grpc.ClientConn, address string) (bala
...
-// Otherwise, it returns the spendable balances of the account as an sdk.Coins object.
+// GetAccountSpendableBalanceByBlock retrieves the spendable balance of an account for the specified address at the given block using gRPC.
 func GetAccountSpendableBalanceByBlock(grpcConn *grpc.ClientConn, address string, block *coretypes.Block) (balances sdk.Coins, err error) {
 	cli := banktypes.NewQueryClient(grpcConn)
 	ctx := metadata.AppendToOutgoingContext(context.Background(), grpctypes.GRPCBlockHeightHeader, fmt.Sprint(block.Height))
-	ctx = metadata.AppendToOutgoingContext(ctx, grpctypes.GRPCBlockTimeHeader, fmt.Sprint(block.Time.Unix()))
+
+	// Since the blockTime is not inferred from block height to the GRPC server, we use the following
+	// line; however, it is commented out because we do not want to modify the SDK code to support it
+	// ctx = metadata.AppendToOutgoingContext(ctx, grpctypes.GRPCBlockTimeHeader, fmt.Sprint(block.Time.Unix()))
```
rootulp, line 299: [blocking] I don't understand this. Why do we want to merge the commented code?

rootulp (commented):
One blocking question

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/celestiaorg/celestia-app
 go 1.20
 
 require (
-	github.com/celestiaorg/nmt v0.17.0
+	github.com/celestiaorg/nmt v0.18.0
```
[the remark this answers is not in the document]
mojtaba-esk: this must be a mistake, lemme fix it

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+	}
+}
+
+// initContinuousVestingAccounts function initializes continuous vesting accounts for testing purposes.
+// It takes the count of accounts as input and returns a GenesisOption. It generates account names,
+// base accounts, and balances. It defines start & end times to creates vesting accounts. The function
+// then updates the genesis state with the generated vesting accounts and balances. The start & endTime
+// of each account increases progressively to ensure some accounts have locked balances, catering to the
+// testing requirements.
+func (s *VestingModuleTestSuite) initContinuousVestingAccounts(count int) testnode.GenesisOption {
```
SweeXordious, line 503: [optional][refactor]
I guess this can be refactored to: 
```
func (s *VestingModuleTestSuite) initVestingAccounts(accountType accountType, count int) testnode.GenesisOption {
    ...
    switch accountType {
	case RegularAccountType:
                ....
	case DelayedVestingAccountType:
		....
	}
    ....
```

So that we don't have duplicate functions for every vesting account type

go.mod
mojtaba-esk: now not any more

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+	accounts.counter++
+	s.accounts.Store(accType, accounts)
+
+	return name, nil
+}
+
+// getAnUnusedContinuousVestingAccount returns an unused continuous vesting account and its name.
+//
+// It takes a minimum start-time as input and finds an unused account whose start time is greater than the input.
+func (s *VestingModuleTestSuite) getAnUnusedContinuousVestingAccount(minStartTime int64) (vAcc vestingtypes.ContinuousVestingAccount, name string, err error) {
```
SweeXordious, line 548: [optional][refactor]
Same here, we could refactor these 3 functions with one function and a cast after returning the value:
```
func (s *VestingModuleTestSuite) getNonRegularUnusedVestingAccount(accType accountType, minStartTime int64) (vAcc authtypes.AccountI, name string, err error) {
	for {
		switch accType {
		case RegularAccountType:
			return error
		case ContinuousVestingAccountType:
			...
		}
```

And then, casting like:
```
	acc, name, err := s.getNonRegularUnusedVestingAccount(ContinuousVestingAccountType, tmtime.Now().Unix() - 5)
	assert.NoError(s.T(), err)
	vAcc := acc.(*vestingtypes.ContinuousVestingAccount)
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+	for i := range vAccounts {
+		gAccounts = append(gAccounts, authtypes.GenesisAccount(vAccounts[i]))
+	}
+
+	return func(gs map[string]json.RawMessage) map[string]json.RawMessage {
+		gs, err := AddGenesisAccountsWithBalancesToGenesisState(gs, gAccounts, balances)
+		assert.NoError(s.T(), err)
+		return gs
+	}
+}
```
evan-forbes, line 519: one of the main things that we want from this test and our vesting module efforts is the ability to generically add vesting module accounts to a genesis. Meaning we need to be able to do that outside of this test. The main goal here is to have functionality that add vesting accounts to a genesis, and we can then test that they were installed correctly

test/util/testfactory/utils.go
mojtaba-esk: Just to clean up as the function signatures were descriptive enough

test/util/testfactory/utils.go
mojtaba-esk: do you think we should remove that function entirely as it is not used in the current test code?

test/util/testfactory/utils.go
```diff
@@ -102,3 +114,211 @@ func GenerateAccounts(count int) []string {
...
+	endTime := tmtime.Now().Add(-2 * time.Second)
+	for i := range bAccounts {
+		bal := balances[i]
+		// initUnlockedCoins are subbed to keep them unlocked
+		coins := bal.Coins.Sub(initUnlockedCoins...)
+		acc := vestingtypes.NewDelayedVestingAccount(&bAccounts[i], coins, endTime.Unix())
+		vAccounts = append(vAccounts, acc)
+
+		// the endTime is increased for each account to be able to test various scenarios
+		endTime = endTime.Add(5 * time.Second)
```
evan-forbes, line 163: [blocking]

by adding constant time here that make it almost exclusively useful to the vesting module test. I think in general we need to be focused on creating a generic tooling for creating tooling for installing vesting module accounts and then testing that tooling

evan-forbes (commented):
in general, we want to have the generic functionality to install a vesting account in the genesis, and then we simply want to test if that account was installed correctly. If we want to add test coverage to the vesting module, then I think we should actually upstream that.

as discussed synchronously, in theory we testing everything we need to test in six transactions. We first install the 3 different types of vesting accounts in the genesis using our generic tooling that we will use to create the actual genesis, then we test to ensure that the exist and are locked (3 failed txs), then we test that after whatever period of time/blocks the some predicted amount of funds are unlocked and are able to be used (3 passed txs)

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+// provided genesis state. It returns the updated genesis state and an error if any occurred.
+func AddGenesisAccountsWithBalancesToGenesisState(
+	gs map[string]json.RawMessage,
+	gAccounts []authtypes.GenesisAccount,
+	balances []banktypes.Balance,
+) (map[string]json.RawMessage, error) {
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+
+	var err error
+	gs, err = testfactory.AddAccountsToGenesisState(encCfg, gs, gAccounts...)
```
SweeXordious, line 622: [optional][nit]
```suggestion
	gs, err := testfactory.AddAccountsToGenesisState(encCfg, gs, gAccounts...)
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+	t := s.T()
+	t.Log("setting up vesting module test suite")
+
+	s.kr = testfactory.GenerateKeyring()
+
+	genOpts := []testnode.GenesisOption{}
+	genOpts = append(genOpts, s.initRegularAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initDelayedVestingAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initPeriodicVestingAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initContinuousVestingAccounts(totalAccountsPerType))
```
SweeXordious, line 95: [optional][style]
```suggestion
	genOpts := []testnode.GenesisOption{
		s.initRegularAccounts(totalAccountsPerType),
		s.initDelayedVestingAccounts(totalAccountsPerType),
		s.initPeriodicVestingAccounts(totalAccountsPerType),
		s.initContinuousVestingAccounts(totalAccountsPerType),
	}
```

test/util/testfactory/utils.go
```diff
@@ -102,3 +114,211 @@ func GenerateAccounts(count int) []string {
...
+		balances[i] = banktypes.Balance{
+			Address: acc.GetAddress().String(),
+			Coins:   coins.Add(initExtraCoins...),
+		}
+	}
+
+	return bAccounts, balances
+}
+
+func GenerateDelayedVestingAccounts(kr keyring.Keyring, names []string, initUnlockedCoins ...sdk.Coin) ([]*vestingtypes.DelayedVestingAccount, []banktypes.Balance) {
```
SweeXordious, line 150: I would prefer if we have a comment explaining how the delayed vesting accounts are created. Something like:

```suggestion
// GenerateDelayedVestingAccounts generates a set of delayed vesting accounts with the corresponding init unlocked coins.
// The end time of the vesting accounts starts from the past, i.e. -2sec, for the first name,
// and keeps increasing with every account at a rate of 5sec per account.
func GenerateDelayedVestingAccounts(kr keyring.Keyring, names []string, initUnlockedCoins ...sdk.Coin) ([]*vestingtypes.DelayedVestingAccount, []banktypes.Balance) {
```

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+	// find and test a vesting account with endTime which is already passed
+	vAcc, name, err := s.getAnUnusedDelayedVestingAccount(0)
+	require.NoError(s.T(), err)
+
+	// It is possible in some cases the given account is not unlocked
+	// so if that's the case, we wait a bit here
+	for vAcc.GetVestedCoins(tmtime.Now()).IsZero() {
+		time.Sleep(time.Second)
+	}
+	minExpectedSpendableBal := vAcc.GetVestedCoins(tmtime.Now()).AmountOf(app.BondDenom).Int64()
```
SweeXordious, line 138: [optional]
The `GetVestedCoins` returns nil if no coins are vested. Do you think it makes sense to split this into:
```
	minExpectedSpendableBal := vAcc.GetVestedCoins(tmtime.Now())
	require.NotNil(s.T(), minExpectedSpendableBal)
	require.NoError(s.T(), s.cctx.WaitForNextBlock())

	s.testTransferMustSucceed(name, minExpectedSpendableBal.AmountOf(app.BondDenom).Int64())
```
And anywhere this is used?

test/util/testfactory/utils.go
```diff
@@ -102,3 +114,211 @@ func GenerateAccounts(count int) []string {
 	}
 	return accs
 }
+
+func NewBaseAccount(kr keyring.Keyring, name string) (*authtypes.BaseAccount, sdk.Coins) {
```
SweeXordious, line 118: [optional]
```suggestion
// NewBaseAccount generates a new base account and returns its instance.
// If an empty string is passed as a name, a random one will be generated and used.
func NewBaseAccount(kr keyring.Keyring, name string) (*authtypes.BaseAccount, sdk.Coins) {
```

SweeXordious (commented):
Great stuff :tada: :tada: 

A lot of work has went into this, and it's amazing to see all of these scenarios being tested out :+1: :+1: 

I left a few comments in multiple rounds of reviews, but most of them are optional.

app/test/vesting_test.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// AddGenesisAccountsWithBalancesToGenesisState adds the given genesis accounts and balances to the
+// provided genesis state. It returns the updated genesis state and an error if any occurred.
+func AddGenesisAccountsWithBalancesToGenesisState(
+	gs map[string]json.RawMessage,
+	gAccounts []authtypes.GenesisAccount,
+	balances []banktypes.Balance,
+) (map[string]json.RawMessage, error) {
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
```
evan-forbes, line 494: is there a way that we can include this in the code adding vesting account to the genesis?

app/test/vesting_test.go
```diff
@@ -0,0 +1,507 @@
...
+		resAccBytes, err := testfactory.GetRawAccountInfo(s.cctx.GRPCClient, address)
+		if err != nil {
+			return vAcc, name, err
+		}
+
+		err = vAcc.Unmarshal(resAccBytes)
+		if err != nil || vAcc.EndTime > minEndTime {
+			return vAcc, name, err
+		}
+	}
```
evan-forbes, line 484: is there any way to avoid using for loops for simply lookups?

evan-forbes (commented):
this is headed in the right direction, thanks for taking the time to refactor @mojtaba-esk 

We really want everything to revolve around testing some code that adds vesting accounts to a genesis. Meaning we have a single function per each vesting account that given some address/pubkey, some balance, some vesting params, then that function adds it to the genesis. 

We should treat this code like normal tooling, so like what we discussed friday it should be off in its own package, we can unit test it, then we have tests whose entire purpose is to test that functionality.

just like what we discussed friday, we don't have to do all of those things in this PR. If we merge this then we are going to have to clean these up later, so an issue should be created for that before merging

app/test/vesting_test.go
mojtaba-esk: Since we have to set everything at genesis for these tests and it is time sensitive and we have no control over time, I could not come up with a loopless solution. Maybe if we have a paradigm shift and look at the problem from another angel, a solution could be found.

app/test/vesting_test.go
mojtaba-esk: Sure we can do that

pkg/genesis/vesting_accounts.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// set the accounts in the genesis state
+	authGenState.Accounts = append(authGenState.Accounts, pAccs...)
+	gs[authtypes.ModuleName] = encCfg.Codec.MustMarshalJSON(&authGenState)
+
+	return gs, nil
+}
+
+// AddBalancesToGenesisState adds the provided balances to the genesis state (gs) for the bank module.
+// It takes the raw genesis state (gs) and a variadic number of Balance objects (balances) as inputs.
```
evan-forbes, line 124: > and a variadic number of Balance objects (balances)

this doesn't appear to be the case, if not then we should update these docs

pkg/genesis/vesting_accounts_test.go
```diff
@@ -0,0 +1,498 @@
...
+	t := s.T()
+	t.Log("setting up vesting module test suite")
+
+	s.kr = testfactory.GenerateKeyring()
+
+	genOpts := []testnode.GenesisOption{}
+	genOpts = append(genOpts, s.initRegularAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initDelayedVestingAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initPeriodicVestingAccounts(totalAccountsPerType))
+	genOpts = append(genOpts, s.initContinuousVestingAccounts(totalAccountsPerType))
```
evan-forbes, line 93: as we've discussed previously, it looks like a lot of the logic here is actually things that we need in the generic code to add things to the genesis. If that's the case, then I don't think they should be specific to this test.

more specifically, these functions are both calling the function to create the vesting account, and then adding it to the genesis. We want a generic function that does this exact thing, then we test it.

does this make sense? if not I'm happy to sync on this to get more specific if needed :slightly_smiling_face:

evan-forbes (commented):
almost there! we're so close! this is definitely headed in the right direction

pkg/genesis/vesting_accounts.go
mojtaba-esk: yeah it seems like the comment is outdated

pkg/genesis/vesting_accounts_test.go
mojtaba-esk: well right now it is pretty much like this. It just does it from another perspective.

We have functions like this one which is generic and creates a generic genesis account with its balance with the given input for us.
```go
func NewGenesisDelayedVestingAccount(
	address string,
	vestingBalance,
	initUnlockedCoins sdk.Coins,
	endTime time.Time,
) (account authtypes.GenesisAccount, balance banktypes.Balance, err error) {
...
}
```

and then we just need to call the following function to add it to genesis:

```go
func AddGenesisAccountsWithBalancesToGenesisState(
	encCfg encoding.Config,
	gs map[string]json.RawMessage,
	gAccounts []authtypes.GenesisAccount,
	balances []banktypes.Balance,
) (map[string]json.RawMessage, error)
```

So does it do the same thing or I did not get your point properly?

evan-forbes: exactly, we're almost doing it, but we just need to go the extra step to do it

literally just combining these functions into a different function that accepts a genesis, list of accout to create, the returns a genesis back

app/test/vesting_test.go
mojtaba-esk: Thanks for pointing it out. Yes we can add it for the unlocked balances tests. will add it.

mojtaba-esk: Thanks for pointing it out. Yes we can add it for the unlocked balances tests. will add it.

test/util/testfactory/utils.go
mojtaba-esk: mark resolved as the code does not exist anymore

app/test/vesting_test.go
mojtaba-esk: done

app/test/vesting_test.go
mojtaba-esk: done

app/test/vesting_test.go
mojtaba-esk: done

app/test/vesting_test.go
mojtaba-esk: We need to pass `minEndTime` to the delayed vesting account function, so if we merge those, it might introduce confusion.

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+	totalAccountsPerType = 300
+	initBalanceForGasFee = 10
+	vestingAmount        = testfactory.BaseAccountDefaultBalance
+	vestingDelayPerTx    = 10 // this is a safe time (in seconds) to wait for a tx to be executed while the vesting period is not over yet
+)
+
+type accountDispenser struct {
+	names   []string
+	counter int
+}
```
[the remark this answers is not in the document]
mojtaba-esk: yes we do in a sync.map, it keeps the account type as the key for the map.

```go
type VestingModuleTestSuite struct {
	suite.Suite

	// accounts is a map from accountType to accountDispenser
	accounts    sync.Map
	accountsMut sync.Mutex

	kr   keyring.Keyring
	cctx testnode.Context
	ecfg encoding.Config
}
```

test/util/testfactory/utils.go
mojtaba-esk: marked as resolved as the feedback is applied

app/test/vesting_test.go
mojtaba-esk: the changes applied

app/test/vesting_test.go
```diff
@@ -0,0 +1,632 @@
...
+// The keyring of the context is set to the keyring (s.kr) of the VestingModuleTestSuite.
+func (s *VestingModuleTestSuite) startNewNetworkWithGenesisOpt(genesisOpts ...testnode.GenesisOption) {
+	s.ecfg = encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	cfg := testnode.DefaultConfig().WithGenesisOptions(testnode.ImmediateProposals(s.ecfg.Codec))
+	cfg.GenesisOptions = genesisOpts
+
+	cctx, _, _ := testnode.NewNetwork(s.T(), cfg)
+	s.cctx = cctx
+	s.cctx.Keyring = s.kr
+}
```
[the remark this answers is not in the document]
mojtaba-esk: indeed. Initially each test setup its own network, now we can consolidate all in one. Thanks for catching this point.

test/util/testfactory/utils.go
mojtaba-esk: removed

app/test/vesting_test.go
mojtaba-esk: thanks done

app/test/vesting_test.go
mojtaba-esk: now we have shrank the code a lot, so it is considered done ;)

app/test/vesting_test.go
mojtaba-esk: this is done in the new structure

app/test/vesting_test.go
mojtaba-esk: added

test/util/testfactory/utils.go
```diff
@@ -102,3 +109,89 @@ func GenerateAccounts(count int) []string {
...
+}
+
+func GetValidators(grpcConn *grpc.ClientConn) (stakingtypes.Validators, error) {
+	scli := stakingtypes.NewQueryClient(grpcConn)
+	vres, err := scli.Validators(context.Background(), &stakingtypes.QueryValidatorsRequest{})
+
+	if vres == nil {
+		return stakingtypes.Validators{}, err
+	}
+	return vres.Validators, err
```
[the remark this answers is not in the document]
mojtaba-esk: yes it is, the reason is because `vres` can be nil and if it is nil, then `vres.Validators` would panic due to nil reference err.
we can probably go for something like this:
```go
if err != nil || vres == nil{
		return stakingtypes.Validators{}, err
}
```

test/util/testfactory/utils.go
```diff
@@ -102,3 +109,89 @@ func GenerateAccounts(count int) []string {
...
+			Address: address,
+		},
+	)
+	if err != nil || res == nil {
+		return nil, err
+	}
+	return res.GetBalances(), nil
+}
+
+// GetAccountSpendableBalanceByBlock retrieves the spendable balance of an account for the specified address at the given block using gRPC.
```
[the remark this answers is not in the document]
mojtaba-esk: I removed the function

test/util/testfactory/utils.go
```diff
@@ -102,3 +109,89 @@ func GenerateAccounts(count int) []string {
...
+func GetAccountSpendableBalanceByBlock(grpcConn *grpc.ClientConn, address string, block *coretypes.Block) (balances sdk.Coins, err error) {
+	cli := banktypes.NewQueryClient(grpcConn)
+	ctx := metadata.AppendToOutgoingContext(context.Background(), grpctypes.GRPCBlockHeightHeader, fmt.Sprint(block.Height))
+
+	// Since the blockTime is not inferred from block height to the GRPC server, we use the following
+	// line; however, it is commented out because we do not want to modify the SDK code to support it
+	// ctx = metadata.AppendToOutgoingContext(ctx, grpctypes.GRPCBlockTimeHeader, fmt.Sprint(block.Time.Unix()))
```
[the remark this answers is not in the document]
mojtaba-esk: I removed the function

test/util/testfactory/utils.go
```diff
@@ -102,3 +109,89 @@ func GenerateAccounts(count int) []string {
...
+
+	res, err := cli.SpendableBalances(
+		ctx,
+		&banktypes.QuerySpendableBalancesRequest{
+			Address: address,
+		},
+	)
+	if err != nil || res == nil {
```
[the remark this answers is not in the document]
mojtaba-esk: I removed the function

test/util/testfactory/utils.go
```diff
@@ -102,3 +109,89 @@ func GenerateAccounts(count int) []string {
...
+	return res.GetBalances(), nil
+}
+
+func GetRawAccountInfo(grpcConn *grpc.ClientConn, address string) ([]byte, error) {
+	cli := authtypes.NewQueryClient(grpcConn)
+	res, err := cli.Account(context.Background(), &authtypes.QueryAccountRequest{
+		Address: address,
+	})
+
+	if err != nil || res == nil {
```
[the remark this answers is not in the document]
mojtaba-esk: if account does not exist and res is nil, we get panic in the next line:
```go
res.Account.Value
```

test/util/testfactory/utils.go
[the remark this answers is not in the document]
rootulp: > the reason is because vres can be nil

Right but the critical part is if `vres` can be nil at the same time as `err`. If `vres` is nil I would expect `err` to be non-nil. In other words, this part should be sufficient:

```
if err != nil
		return stakingtypes.Validators{}, err
}
// do something with vres because it should be non-nil
```

It would be really confusing if the `Validators()` invocation returned `err = nil` AND `vres = nil` because that gives us no signal on if the method succeeded or failed.

There exists an example in this codebase of invoking `Validators()` that performs error handling how I would expect [here](https://github.com/celestiaorg/celestia-app/blob/3810531e027e05f76e137820572c766b1bcda1c3/test/txsim/stake.go#L104-L108)

test/util/testfactory/utils.go
```diff
@@ -102,3 +109,89 @@ func GenerateAccounts(count int) []string {
...
+	cli := banktypes.NewQueryClient(grpcConn)
+	res, err := cli.SpendableBalances(
+		context.Background(),
+		&banktypes.QuerySpendableBalancesRequest{
+			Address: address,
+		},
+	)
+	if err != nil || res == nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
rootulp: If the account does not exist, I observe that result is non-nil so again the second part doesn't seem necessary or desirable.

```go
	doesNotExist := "celestia1jwzamm3ltkzce7ey5tn7uadt8uxg6k89c6ftnn"
	testfactory.GetAccountSpendableBalance(s.cctx.GRPCClient, doesNotExist)
```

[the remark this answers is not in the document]
mojtaba-esk: yes I realized it later and now fixed the code.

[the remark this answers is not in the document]
mojtaba-esk: Thanks for catching this point

evan-forbes (approved):
approving as the remaining feedback from me will be handled in #1994 instead as that is also a good place to add it.

:shipit: !

evan-forbes:
let's also add the necessary tests to the race detector block list

pkg/genesis/vesting_accounts.go
```diff
@@ -0,0 +1,155 @@
...
+		Address: address,
+		Coins:   balances,
+	}
+	bAccount := authtypes.NewBaseAccountWithAddress(sdkAddr)
+
+	return authtypes.GenesisAccount(bAccount), balance, nil
+}
+
+// NewGenesisDelayedVestingAccount creates a new DelayedVestingAccount with the specified parameters.
+// then returns the created account converted to genesis account type and the account balance
```
rootulp, line 33: [nits]
- capitalize first word in sentence
- period at end of sentence
- rewrap

```suggestion
// NewGenesisDelayedVestingAccount creates a new DelayedVestingAccount with the
// specified parameters. It returns the created account converted to genesis
// account type and the account balance.
```

I'm starting to think the comment rewrapping thing to 80 character width is a waste of dev productivity b/c we commonly have nits about it. Since it isn't enforced by a linter, I don't think it is worth doing.

rootulp: [no change needed] I'm not sure why the delayed vesting account constructor is the only function in this file that has a godoc. It seems comparable to `NewGenesisPeriodicVestingAccount` and `NewGenesisContinuousVestingAccount`.

rootulp (dismissed):

rootulp (dismissed):

evan-forbes (approved):

pkg/genesis/vesting_accounts_test.go
```diff
@@ -118,6 +118,12 @@ func (s *VestingModuleTestSuite) TestGenesisDelayedVestingAccountsTransferUnLock
 	vAcc, name, err := s.getAnUnusedDelayedVestingAccount(0)
 	require.NoError(s.T(), err)
 
+	// Since we want a partially unlocked balance we need to wait until
+	// the endTime is passed if not already
```
rootulp, line 122: this comment doesn't make sense to me. If this test needs a partially unlocked balance, I would expect it to wait until the start time has passed.

If it actually needed to wait until end time passed, then I would expect the helper above to be:

```go
	vAcc, name, err := s.getAnUnusedDelayedVestingAccount(tmtime.Now().Unix())
```

```suggestion
	// Since we want a partially unlocked balance we need to wait until
	// the startTime has passed if not already
```

rootulp (approved):

evan-forbes (approved):

PR #2012, opened 2023-06-30
fix!: use max square size worst case padding when constructing the square in all cases

### LINKED ISSUE #1952: Export function to go from block data to EDS
there's a [function](https://github.com/celestiaorg/celestia-node/blob/6654cdf4994dbd381efd0d6a29688c731177c855/core/eds.go#L19-L38) in celestia-node that we can move most of to the application

```go
// extendBlock extends the given block data, returning the resulting
// ExtendedDataSquare (EDS). If there are no transactions in the block,
// nil is returned in place of the eds.
func extendBlock(data types.Data) (*rsmt2d.ExtendedDataSquare, error) {
	if len(data.Txs) == 0 && data.SquareSize == uint64(1) {
		return nil, nil
	}

	sqr, err := square.Construct(data.Txs.ToSliceOfBytes(), appconsts.LatestVersion, share.MaxSquareSize)
	if err != nil {
		return nil, err
	}

	shares := make([][]byte, len(sqr))
	for i, s := range sqr {
		shares[i] = s.ToBytes()
	}

	return da.ExtendShares(shares)
}
```

note that I'm unsure if we should move the node specific logic on lines 23-25, as that's node specific logic, which feels like at initial glance that it should stay in node so they can change it if necessary without making changes here

### LINKED ISSUE #2013: Use the maximum square size when calculating the worst case share indexes in all scenarios
Square construction is supposed to be deterministic no matter the square size originally used to build the square during consensus. This property can currently fail since we are not using the same value to calculate the worse case share indexes.

https://github.com/celestiaorg/celestia-app/blob/c66ab98cba29e6cbbe8a03f107ae054e27b9a3ae/pkg/square/builder.go#L413-L419

For instance, during consensus if the gov max square size is smaller than maximum upper bound size (this is the case by default), then the `maxSquareCapacity` used in the above code will be `64*64` during consensus, but outside of consensus, for instance in celestia-node, this process is stateless and the max upper bound is used. In that case, we would use `128*128`, resulting in a different square construction, and thus different data root.

evan-forbes:
## Overview

While fixing #1952, I managed to get lucky and catch #2013 when running tests. imo, it combining the two isn't the worst thing since they help test #2013. Therefore, this PR is currently api breaking since we change `square.Construct`/`square.Build`, and consensus breaking because of the bug fix.

closes #1952
closes #2013 

## Checklist


app/test/integration_test.go
```diff
@@ -344,3 +346,24 @@ func (s *IntegrationTestSuite) TestShareInclusionProof() {
...
+	t := s.T()
+	emptyHeights := []int64{1, 2, 3}
+	for _, h := range emptyHeights {
+		blockRes, err := s.cctx.Client.Block(s.cctx.GoContext(), &h)
+		require.NoError(t, err)
+		require.True(t, app.EmptyBlock(blockRes.Block.Data, blockRes.Block.Header.Version.App))
+		s.ExtendBlobTest(t, blockRes.Block)
+	}
+
+}
```
evan-forbes, line 369: testing these via integration tests is likely a bit overkill, however since they are public functions this should give us a high degree of confidence that they will work as intended

rootulp (dismissed):

pkg/square/builder.go
```diff
@@ -410,10 +409,15 @@ func (e element) maxShareOffset() int {
...
+// encode.
+func worstCaseShareIndexes(blobs int, appVersion uint64) []uint32 {
+	maxSquareSize := appconsts.SquareSizeUpperBound(appVersion)
 	shareIndexes := make([]uint32, blobs)
 	for i := range shareIndexes {
-		shareIndexes[i] = uint32(maxSquareCapacity)
+		shareIndexes[i] = uint32(maxSquareSize * maxSquareSize)
 	}
 	return shareIndexes
 }
```
evan-forbes, line 423: Pls review!!

here's the bug, where in some edge cases (lots of txs) we'd end up using different values for the worst case, which would cause different square constructions when using different "maxSquareSize" values

app/extend_block.go
```diff
@@ -13,7 +12,7 @@ import (
 // version.
 func ExtendBlock(data coretypes.Data, appVersion uint64) (*rsmt2d.ExtendedDataSquare, error) {
 	// Construct the data square from the block's transactions
-	dataSquare, err := square.Construct(data.Txs.ToSliceOfBytes(), appVersion, appconsts.SquareSizeUpperBound(appVersion))
+	dataSquare, err := square.Construct(data.Txs.ToSliceOfBytes(), appVersion, 64)
```
rootulp, line 15: [blocking][question] why hard-code `64` instead of using the `appconsts.SquareSizeUpperBound(appVersion)`?

app/test/integration_test.go
```diff
@@ -349,11 +352,16 @@ func (s *IntegrationTestSuite) TestShareInclusionProof() {
...
+func ExtendBlobTest(t *testing.T, block *coretypes.Block) {
 	eds, err := app.ExtendBlock(block.Data, block.Header.Version.App)
 	require.NoError(t, err)
 	dah := da.NewDataAvailabilityHeader(eds)
-	require.Equal(t, dah.Hash(), block.DataHash.Bytes())
+	if !assert.Equal(t, dah.Hash(), block.DataHash.Bytes()) {
+		// save block to json file for further debugging if this occurs
+		b, err := json.MarshalIndent(block, "", "  ")
+		require.NoError(t, err)
+		require.NoError(t, ioutil.WriteFile(fmt.Sprintf("bad_block_%s.json", tmrand.Str(6)), b, 0644))
```
rootulp, line 363: [question] is this strictly for local debugging? Do we have access to this artifact if it occurs on CI?

pkg/square/builder.go
```diff
@@ -410,10 +409,15 @@ func (e element) maxShareOffset() int {
 	return e.numShares + e.maxPadding
 }
 
-func worstCaseShareIndexes(blobs, maxSquareCapacity int) []uint32 {
+// worstCaseShareIndexes returns the largest possible shared indexes for a set
```
rootulp, line 412:
```suggestion
// worstCaseShareIndexes returns the largest possible share indexes for a set
```

app/extend_block.go
evan-forbes: shit nice catch!!

app/test/integration_test.go
evan-forbes: yeah, this is only for local cases, but if it ever occurs then we automatically have a saved instance of the block so decided to keep this

its totally optional tho, so we can remove if we don't find it valuable

app/extend_block.go
evan-forbes: fixed in https://github.com/celestiaorg/celestia-app/pull/2012/commits/34957ae2b00a4471a170e0b6c9004233c9398d5e

pkg/square/square.go
```diff
@@ -48,8 +48,9 @@ func Build(txs [][]byte, appVersion uint64, maxSquareSize int) (Square, [][]byte
 //
 // Note that this function does not check the underlying validity of
 // the transactions.
-func Construct(txs [][]byte, appVersion uint64, maxSquareSize int) (Square, error) {
-	builder, err := NewBuilder(maxSquareSize, appconsts.SubtreeRootThreshold(appVersion), txs...)
+func Construct(txs [][]byte, appVersion uint64) (Square, error) {
+	maxSquareSize := appconsts.SquareSizeUpperBound(appVersion)
+	builder, err := NewBuilder(maxSquareSize, appVersion, txs...)
 	if err != nil {
```
evan-forbes, line 54: we no longer have a reason to pass the square size to the construct method, which should always use the value based on the appversion

app/extend_block.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	return da.ExtendShares(shares.ToBytes(dataSquare))
+}
+
+// EmptyBlock returns true if the given block data is considered empty by the
+// application at a given version.
+func EmptyBlock(data coretypes.Data, _ uint64) bool {
+	return len(data.Txs) == 0
+}
```
evan-forbes, line 27: These technically aren't related to the fix, but they are needed to test the fix and we need them to complete #1952 anyway

rootulp (dismissed):

app/extend_block.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return da.ExtendShares(shares.ToBytes(dataSquare))
+}
+
+// EmptyBlock returns true if the given block data is considered empty by the
+// application at a given version.
+func EmptyBlock(data coretypes.Data, _ uint64) bool {
```
cmwaters, line 25:
```suggestion
func IsEmptyBlock(data coretypes.Data, _ uint64) bool {
```

app/process_proposal.go
```diff
@@ -104,7 +104,7 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 	}
 
 	// Construct the data square from the block's transactions
-	dataSquare, err := square.Construct(req.BlockData.Txs, app.GetBaseApp().AppVersion(), app.GovSquareSizeUpperBound(sdkCtx))
+	dataSquare, err := square.Construct(req.BlockData.Txs, app.GetBaseApp().AppVersion())
```
cmwaters, line 107: [blocking] How do we enforce the governance based square size if we don't pass it here in Construct? From looking at this change, it seems a proposer could build an invalid block that is larger than the govmaxsquaresize and still get it committed.

cmwaters (dismissed):

app/extend_block.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/celestiaorg/celestia-app/pkg/da"
+	"github.com/celestiaorg/celestia-app/pkg/shares"
+	"github.com/celestiaorg/celestia-app/pkg/square"
+	"github.com/celestiaorg/rsmt2d"
+	coretypes "github.com/tendermint/tendermint/types"
+)
+
+// ExtendBlock extends the given block data into a data square for a given app
+// version.
+func ExtendBlock(data coretypes.Data, appVersion uint64) (*rsmt2d.ExtendedDataSquare, error) {
```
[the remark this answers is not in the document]
evan-forbes: this is just going off of what was requested by celestia-node in https://github.com/celestiaorg/celestia-app/issues/1952, so the same api of that function was used

pkg/square/square.go
evan-forbes: this was incorrect!!! #2015

app/extend_block.go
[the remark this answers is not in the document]
cmwaters: Do we want to add a check then that the square size calculated equals the square size provided in Data?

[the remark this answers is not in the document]
evan-forbes: actually after thinking about this more, I don't think we should add this check unless we are commiting to the value as it makes it easy to make the celestia-node to reject an otherwise valid block simply by modifying an uncommitted value

app/extend_block.go
```diff
@@ -0,0 +1,27 @@
...
+	// Construct the data square from the block's transactions
+	dataSquare, err := square.Construct(data.Txs.ToSliceOfBytes(), appVersion, 64)
```
rootulp, line 15: [blocking] I may be viewing an outdated commit but `64` still appears hard-coded

rootulp (commented):
One blocking comment about hard-coded 64, otherwise LGTM

app/extend_block.go
evan-forbes: gah sorry I reverted to fix different feedback, and it also reverted this change, fixed in https://github.com/celestiaorg/celestia-app/pull/2012/commits/5e7ea5e80f541ee358446435920b4ee1006e196f

rootulp (approved):

app/extend_block.go
```diff
@@ -0,0 +1,28 @@
...
+	// Construct the data square from the block's transactions
+	dataSquare, err := square.Construct(data.Txs.ToSliceOfBytes(), appVersion, appconsts.SquareSizeUpperBound(appVersion))
+	if err != nil {
+		return nil, err
+	}
+
+	return da.ExtendShares(shares.ToBytes(dataSquare))
+}
+
+// EmptyBlock returns true if the given block data is considered empty by the
```
staheri14, line 24:
```suggestion
// IsEmptyBlock returns true if the given block data is considered empty by the
```

app/extend_block.go
```diff
@@ -0,0 +1,28 @@
...
+// application at a given version.
```
staheri14, line 25: As the version is not used, maybe we can remove it from the func description as well. 

```suggestion
// EmptyBlock returns true if the given block data is considered empty.
```

pkg/square/builder.go
```diff
@@ -410,10 +409,15 @@ func (e element) maxShareOffset() int {
 	return e.numShares + e.maxPadding
 }
 
-func worstCaseShareIndexes(blobs, maxSquareCapacity int) []uint32 {
+// worstCaseShareIndexes returns the largest possible share indexes for a set
+// of blobs at a given appversion. Largest possible is "worst" in that protobuf
+// uses varints to encode integers, so larger integers can require more bytes to
+// encode.
+func worstCaseShareIndexes(blobs int, appVersion uint64) []uint32 {
```
staheri14, line 416: [suggestion] The function name isn't very clear about what it represents as the "worst". It might be better to use something like `shareIndexUpperBound` to make it more self-explanatory.

SweeXordious (approved):

app/extend_block.go
evan-forbes: gahh dang, will fix in a different PR for main

pkg/square/builder.go
```diff
@@ -410,10 +409,15 @@ func (e element) maxShareOffset() int {
...
+	maxSquareSize := appconsts.SquareSizeUpperBound(appVersion)
```
cmwaters, line 417: I don't think this would've been problematic because maxSquareCapacity was never going to be greater than the square size upper bound. I don't think it was possible to underestimate the square size

Issue #2029, opened 2023-07-04
Audit `x/mint`


SweeXordious:
## Context

https://github.com/celestiaorg/celestia-app/tree/main/x/mint

## Proposal

- Audit the specs and implementation
- Update the specs if anything is missing

PR #2030, opened 2023-07-04
docs: correct the mint inflation schedule

SweeXordious:
## Overview

Part of https://github.com/celestiaorg/celestia-app/issues/2029

Unless I'm missing something, one value needs to be updated in the specs:

image

## Checklist


rootulp:
Where is the screenshot from? I sourced the `1.647129056757192` value from ADR-019 https://github.com/celestiaorg/celestia-app/blob/965eafb4357376aec31f84f3628f7703c5587f9a/docs/architecture/adr-019-strict-inflation-schedule.md?plain=1#L45

SweeXordious:
@rootulp I did it manually in sheets: https://docs.google.com/spreadsheets/d/1XHc0o-2kK54vyn8lS_jd0-oeSQ96S2VX7JtMG4XhL8Q/edit?usp=sharing

I'll change in the ADR if this is correct

SweeXordious:
The thing I am investigating now is that we're using that exact value in the tests and it's not complaining... not sure what's the difference

rootulp:
I suspect Google sheets is dropping at that level of precision so I don't think the specs / ADR need to be updated but this raises a good point that all the values in the specs should have an associated unit test

PR #2008, opened 2023-06-29
feat: add multiple excluded directories to the test coverage script

evan-forbes:
## Overview

this pr enables us to add multiple directories to the exclusion list for measuring test coverage, and then adds the malicious package to that list

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #2051, opened 2023-07-10
test: coverage report to exclude test/util and to not run race detection

cmwaters:


rootulp (approved):
[non-blocking][question] why did we ignore `/malicious` from test coverage in https://github.com/celestiaorg/celestia-app/pull/2008 ?

I don't see `/malicious` ignored from the test commands in the Makefile: https://github.com/celestiaorg/celestia-app/blob/main/Makefile so it seems safe to run test on it + collect code coverage for it

cmwaters:
> I don't see `/malicious` ignored from the test commands in the Makefile: https://github.com/celestiaorg/celestia-app/blob/main/Makefile so it seems safe to run test on it + collect code coverage for it

Yeah I'm indifferent about this. I'm not too sure who's using the code coverage data at all. You could also argue that code coverage for parts that can't possibly be hit in production (i.e. test infrastructure) isn't important to publish.

rootulp (approved):

SweeXordious (approved):
I guess the `e2e` framework and `txsim` also could be excluded as they're part of the infrastructure. Or could be moved under `util` too.

scripts/test_cover.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 # Define the directories to exclude
-EXCLUDE_DIRS=("/malicious")
+EXCLUDE_DIRS=("/test/util")
```
staheri14, line 5: A suggestion about the PR title: This PR does more than "removing race detection", it also "excludes the "/test/util" directory from code coverage analysis".  You may consider updating the title.

staheri14 (commented):
It LGTM, however, I have a question: Am I correct in assuming that this script is exclusively for generating a test coverage report? In other words, do we still execute the tests located in the test/util directory, such as [this test](https://github.com/celestiaorg/celestia-app/blob/4725fd8925288fc03b42b1180526e07eaa2b927b/test/util/blobfactory/payforblob_factory_test.go#L15), or not?

SweeXordious:
> It LGTM, however, I have a question: Am I correct in assuming that this script is exclusively for generating a test coverage report? In other words, do we still execute the tests located in the test/util directory, such as [this test](https://github.com/celestiaorg/celestia-app/blob/4725fd8925288fc03b42b1180526e07eaa2b927b/test/util/blobfactory/payforblob_factory_test.go#L15), or not?

The tests are run in here: 
https://github.com/celestiaorg/celestia-app/blob/81ee99bf6a805cad7d3a20c544147c27c1b0aa82/.github/workflows/test.yml#L18-L19

What this PR touches is this, to generate the report:
https://github.com/celestiaorg/celestia-app/blob/81ee99bf6a805cad7d3a20c544147c27c1b0aa82/.github/workflows/test.yml#L21-L36

But as you said, it would be better if the title of the PR is adapted to also refer to the EXCLUDE directories change.

staheri14 (approved):
LGTM! (my suggestion for a change in the PR title still holds :) )

staheri14:
Thanks @SweeXordious for clarification on the tests 🙏

PR #2057, opened 2023-07-10
chore: docker common pipeline v0.2.1

jrmanes:
hello team!

I've created a new version in the common pipeline, this one should fix the issues with the forks and we'll have the version output (*when we'll be using a tag version*) that we're currently missing when using the docker containers. 

Thanks in advance!

Jose Ramon Mañes

closes: https://github.com/celestiaorg/devops/issues/369


rootulp:
I updated the PR title b/c from celestia-app's perspective updating this dependency seems like a chore. In other words, I don't think there's a user facing feature in this PR. But @jrmanes please rename if you think differently!

jrmanes:
> 

hello @rootulp :) 
thanks for mentioning this, https://github.com/celestiaorg/.github/pull/64 is pending because of this: https://github.com/celestiaorg/.github/pull/64#pullrequestreview-1522230288
I think we shouldn't have the forks issue using this version because we are not pushing always, anyway, we'll keep checking it

rootulp:
Merging b/c @jrmanes doesn't have permission to

Issue #2060, opened 2023-07-10
Upgrade to rsmt2d v0.10.0


rootulp:
## Context

https://github.com/celestiaorg/rsmt2d/releases/tag/v0.10.0

## Proposal

Upgrade rsmt2d from v0.9.0 to v0.10.0

PR #2062, opened 2023-07-10
chore!: bump to rsmt2d v0.10.0

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2060

Marked as breaking b/c after this PR `NewDataAvailabilityHeader` may return an error.

### DIFF
--- a/app/process_proposal.go
+++ b/app/process_proposal.go
@@ -122,7 +122,11 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 		return reject()
 	}
 
-	dah := da.NewDataAvailabilityHeader(eds)
+	dah, err := da.NewDataAvailabilityHeader(eds)
+	if err != nil {
+		logInvalidPropBlockError(app.Logger(), req.Header, "failure to create new data availability header", err)
+		return reject()
+	}
 	// by comparing the hashes we know the computed IndexWrappers (with the share indexes of the PFB's blobs)
 	// are identical and that square layout is consistent. This also means that the share commitment rules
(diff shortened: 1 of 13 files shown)


app/process_proposal.go
```diff
@@ -122,7 +122,11 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 		return reject()
 	}
 
-	dah := da.NewDataAvailabilityHeader(eds)
+	dah, err := da.NewDataAvailabilityHeader(eds)
```
staheri14, line 125: [Question] I'm curious to know if a malicious player can cause a validator to panic by using corrupted data. If that is indeed possible, would triggering a panic be considered a reasonable response?

staheri14 (approved):
I left a question about the panic behavior, it is non-blocking if we are sure that panic is a safe option.

app/process_proposal.go
rootulp: I don't understand the second part of the question, do you mind rephrasing?

I don't think it's possible for a malicious proposer to cause a validator to panic. If it were possible, then a malicious proposer could halt the chain by causing all honest validators to `panic`. I don't think `panic` is a reasonable response for an honest validator if they encounter a malicious block. In this scenario, an honest validator should reject the block and remain online.

staheri14: > If that is indeed possible, would triggering a panic be considered a reasonable response?

Thanks for the clarification. 🙏 
My question about panic does not apply to `process_proposal` as unlike `prepare_proposal` it does not panic (confused the two).

Issue #1961, opened 2023-06-19
Audit `x/upgrade`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/tree/main/x/upgrade

## Proposal
- Audit the specs and implementation
- Update the specs if anything is missing

rootulp:
Verified that it is possible to submit a legacy governance proposal via

```shell
$ celestia-appd tx gov submit-legacy-proposal --title="Test Proposal" --description="My awesome proposal" --type="Text" --deposit="10utia" --from validator --home $CELESTIA_APP_HOME --fees 21000utia
auth_info:
  fee:
    amount:
    - amount: "21000"
      denom: utia
    gas_limit: "210000"
    granter: ""
    payer: ""
  signer_infos: []
  tip: null
body:
  extension_options: []
  memo: ""
  messages:
  - '@type': /cosmos.gov.v1beta1.MsgSubmitProposal
    content:
      '@type': /cosmos.gov.v1beta1.TextProposal
      description: My awesome proposal
      title: Test Proposal
    initial_deposit:
    - amount: "10"
      denom: utia
    proposer: celestia1xq74xr0jpglx0vx7v77ar34w6fdv9nxhmgkmn5
  non_critical_extension_options: []
  timeout_height: "0"
signatures: []
confirm transaction before signing and broadcasting [y/N]: y
code: 0
codespace: ""
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: '[]'
timestamp: ""
tx: null
txhash: 8F5A709D6B743CD9A721E4C38FEE2265A8F71BF7657185F75B7AE069A1CC7EAD
```

and the legacy proposal failed to execute:

```
$ ./build/celestia-appd query tx 8F5A709D6B743CD9A721E4C38FEE2265A8F71BF7657185F75B7AE069A1CC7EAD
code: 9
codespace: gov
data: ""
events:
- attributes:
  - index: true
    key: c3BlbmRlcg==
    value: Y2VsZXN0aWExeHE3NHhyMGpwZ2x4MHZ4N3Y3N2FyMzR3NmZkdjlueGhtZ2ttbjU=
  - index: true
    key: YW1vdW50
    value: MjEwMDB1dGlh
  type: coin_spent
- attributes:
  - index: true
    key: cmVjZWl2ZXI=
[...]

PR #1562, opened 2023-03-29
docs: ADR 018 Network Upgrades

evan-forbes:
## Overview

ADR for social upgrades, still a work in progress in that we still need to finalize a few decisions before moving to the implementation detail phase, but this ADR is ready for the first round of reviews.

Related to [ADR 016](https://github.com/celestiaorg/celestia-app/pull/1491)

[rendered](https://github.com/celestiaorg/celestia-app/blob/evan/adr-018-social-upgrades/docs/architecture/adr-018-network-upgrades.md)

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #2072, opened 2023-07-10
chore: audit `x/upgrade` module

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1961


x/upgrade/test/integration_test.go
```diff
@@ -54,12 +53,8 @@ func (s *UpgradeTestSuite) SetupSuite() {
 		accounts[i] = tmrand.Str(9)
 	}
 
-	tmCfg := testnode.DefaultTendermintConfig()
-	tmCfg.Consensus.TimeoutCommit = 3 * time.Second
```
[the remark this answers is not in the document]
evan-forbes: we might need to add some more checks to ensure this is doing what its supposed to. The reasoning was the same as the square size test, where we changed the voting period to be 2s. We want to have a block time larger than that so that we can vote.

that's what the 
```go
WithGenesisOptions(testnode.ImmediateProposals(s.ecfg.Codec))
```
does at least.

 Since the upgrade module here removes the ability to upgrade, there should not be a way to do get a proposal through

x/upgrade/README.md
```diff
@@ -2,7 +2,17 @@
...
+This upgrade module is a fork of the cosmos-sdk's
+[x/upgrade](https://github.com/cosmos/cosmos-sdk/tree/main/x/upgrade) module
+that removes the entrypoints to the standard upgrade module by not registering a
+message server. It registers the standard upgrade module types to preserve the
+ability to marshal them. Note that the keeper of the standard upgrade module is
+still added to the application.
+
+A consequence of the removal of the entrypoints is that
+[cosmosvisor](https://github.com/cosmos/cosmos-sdk/tree/main/tools/cosmovisor)
+will not work for celestia-app.
```
evan-forbes, line 14: this is mostly true, but I feel like it might be useful to expand on this point somehow. We could still use it if we needed to. For example, we can still save a json with the relevant info or log things in the same way if we ever needed to have the same type of upgrade. Neutering the upgrade module as we've done here doesn't mean we will never be able to use cosmovisor, simply that we cannot vote on an upgrade with tokens.

Perhaps we should either merge https://github.com/celestiaorg/celestia-app/pull/1562 or I can maybe move portions of it here as well.

evan-forbes (commented):
> Why add the x/upgrade module to our state machine? What behavior do we want to use from the upstream x/upgrade module?

Does the abstract answer this question? I can add portions of https://github.com/celestiaorg/celestia-app/pull/1562 here as well if that would help. We don't want to be able to vote on upgrades, but we do still need some functionality so that IBC continue to compile/work. We also want to use the migration functionality, and we could even utilize the the halting/switching binaries mechanism as well should those things be needed.

x/upgrade/README.md
```diff
@@ -2,7 +2,17 @@
...
-The upgrade module removes the entrypoints to the standard upgrade module by not
-registering a message server. It registers the standard upgrade module types to
-preserve the ability to marshal them. Note that the keeper of the standard
-upgrade module is still added to the application.
+This upgrade module is a fork of the cosmos-sdk's
+[x/upgrade](https://github.com/cosmos/cosmos-sdk/tree/main/x/upgrade) module
+that removes the entrypoints to the standard upgrade module by not registering a
+message server. It registers the standard upgrade module types to preserve the
+ability to marshal them. Note that the keeper of the standard upgrade module is
+still added to the application.
```
cmwaters, line 10: It might be nice to add the motivation around this. 

I'm also a little unclear why we need to remove the message server. Only `x/gov` can invoke the methods right? Which we've removed the handler for

evan-forbes: this is a good point, we did remove the handler's so it should be sufficient to simply remove those handlers. I think the thinking was something along the lines of "why keep code that we can't access", and it was within reason to remove it entirely.

x/upgrade/test/integration_test.go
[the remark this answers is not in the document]
rootulp: Sorry I'm lost.

> We don't have to have a block time longer than that, however it is convenient.

Why does the test need a block time longer than voting period? FWIW voting period is 5 seconds per [these lines](https://github.com/rootulp/celestia-app/blob/16fbdfe2d17832a282b46b2297370c2eab3715fa/test/util/testnode/genesis_options.go#L38-L39).

> Since the upgrade module here removes the ability to upgrade, there should not be a way to do get a proposal through

Agreed. I assumed that's what the tests in this file are doing and if they pass without overriding `TimeoutCommit` then I don't see why it's necessary.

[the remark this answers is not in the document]
rootulp: I'm going to revert the changes here 🤷‍♂️ idk why TimeoutCommit is overridden but happy to include that in a code comment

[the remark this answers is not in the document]
rootulp: Reverted in 038a5d8af39746bb3a150cafc04e9060e1f0e5f5

x/upgrade/README.md
rootulp: +1 I think motivating context is in https://github.com/celestiaorg/celestia-app/pull/1491/files#diff-2441f8dbdfce977902fdfe6f74a28966c7b4202c387ee5d0999e394216dc013cR56-R59 so will try to copy some of that content in.

I kinda wish we merged ADR-016 considered we implemented option 2.

rootulp:
> but we do still need some functionality so that IBC continue to compile/work

It's not immediately clear what functionality is preserved in the forked x/upgrade module that is needed for IBC to work.

rootulp:
Does our fork intentionally not register the CLI command? I can't do

```
$ ./build/celestia-appd query upgrade
Error: unknown command "upgrade" for "query"
```

Ref: https://docs.cosmos.network/v0.46/modules/upgrade/04_client.html#cli

x/upgrade/README.md
rootulp: https://github.com/celestiaorg/celestia-app/pull/2072/commits/2d0503e33b9950937ddfe67f4f1b30f49faf3145

x/upgrade/README.md
rootulp: Removed the note on cosmovisor https://github.com/celestiaorg/celestia-app/pull/2072/commits/2d0503e33b9950937ddfe67f4f1b30f49faf3145

evan-forbes: thanks!

evan-forbes:
> It's not immediately clear what functionality is preserved in the forked x/upgrade module that is needed for IBC to work. Maybe we clarify that in a follow-up.

ref https://github.com/celestiaorg/celestia-app/pull/1629#discussion_r1170264709

cmwaters (approved):

evan-forbes (approved):

evan-forbes:
re: the cli commands

since we are unable to submit the software upgrade proposals, imo we should also remove the cli commands that create them since there is no reason to create them

PR #2079, opened 2023-07-11
feat: remove consensus config overrides in the cli

### LINKED ISSUE #2080: Remove the consensus configuration overrides in the config
per sync discussion, we're wanting to remove the overrides to the conensus configuration in the cli, mainly because its annoying to be unable to change it during tests and validators tend not to change the consensus configs.

evan-forbes:
## Overview

removes consensus config overrides in the cli

closes #2080

## Checklist


rootulp (approved):

pkg/appconsts/consensus_consts.go
```diff
@@ -4,5 +4,5 @@ import "time"
 
 const (
 	TimeoutPropose = time.Second * 10
-	TimeoutCommit  = time.Second * 15
+	TimeoutCommit  = time.Second * 11
```
evan-forbes, line 7: for BSR, we had 10 and 10, which typically resulted in 12-13 second empty blocks. This should allow for 13-14 second empty blocks. If there is consistent load on the network, then we should hit 13-15 seconds.

SweeXordious (approved):
[Question]
We're no longer hardcoding config values?

staheri14 (approved):

evan-forbes:
> We're no longer hardcoding config values?

yeah, atm, it seems like a premature change that gets in the way more than actually prevents wierd network activity. Instead, we should wait until we have better tools such as [proposer based timestamps](https://github.com/cometbft/cometbft/tree/main/spec/consensus/proposer-based-timestamp)

pkg/appconsts/consensus_consts.go
evan-forbes: the empty blocks just means that we won't need the extra 10 seconds to gossip proposals and prevote. With large bocks, tednermint's gossiping stack is rather inefficient atm, so we need the extra time to consistently gossip large blocks. When that occurs, we end up with up to 22 second block times. This was one of the goals with having a dynamic value for the 11 seconds set above. Instead of consistently waiting 11 seconds, we would change how long we wait depending on how long we have already waited.

Issue #2091, opened 2023-07-12
test is failing on main


rootulp:
## Problem

https://github.com/celestiaorg/celestia-app/actions/runs/5523740393/jobs/10075070503#step:4:31

```
test/util/malicious/out_of_order_prepare.go:59:9: assignment mismatch: 1 variable but da.NewDataAvailabilityHeader returns 2 values
```
How did https://github.com/celestiaorg/celestia-app/pull/2062 merge if it fails test [here](https://github.com/celestiaorg/celestia-app/actions/runs/5523740393/jobs/10075070503#step:4:187)?

My hypothesis is that it passes the test workflow in isolation but fails on `main` because https://github.com/celestiaorg/celestia-app/pull/2003 merged before it

## Proposal
1. Fix `main`
2. [Maybe] Enforce that PRs need to be up to date w/ main prior to merge

PR #2092, opened 2023-07-12
fix: malicious test and prepare proposal

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2091

## Testing

`make test` passes locally


cmwaters (approved):
Interesting that this snuck in

SweeXordious (approved):

PR #2095, opened 2023-07-12
feat: SquareSize helper for dah

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1011

cc: @Wondertan

### DIFF
--- a/pkg/da/data_availability_header.go
+++ b/pkg/da/data_availability_header.go
@@ -167,6 +167,14 @@ func (dah *DataAvailabilityHeader) IsZero() bool {
 	return len(dah.ColumnRoots) == 0 || len(dah.RowRoots) == 0
 }
 
+// SquareSize returns the number of rows in the original data square.
+func (dah *DataAvailabilityHeader) SquareSize() (int, error) {
+	if len(dah.RowRoots)%2 != 0 {
+		return 0, fmt.Errorf("len row roots is not divisible by 2: %d", len(dah.RowRoots))
+	}
+	return len(dah.RowRoots) / 2, nil
+}
+
 // MinDataAvailabilityHeader returns the minimum valid data availability header.
 // It is equal to the data availability header for a block with one tail padding
 // share.
--- a/pkg/da/data_availability_header.go
+++ b/pkg/da/data_availability_header.go
@@ -168,11 +168,8 @@ func (dah *DataAvailabilityHeader) IsZero() bool {
 }
 
 // SquareSize returns the number of rows in the original data square.
-func (dah *DataAvailabilityHeader) SquareSize() (int, error) {
-	if len(dah.RowRoots)%2 != 0 {
-		return 0, fmt.Errorf("len row roots is not divisible by 2: %d", len(dah.RowRoots))
-	}
-	return len(dah.RowRoots) / 2, nil
+func (dah *DataAvailabilityHeader) SquareSize() int {
+	return len(dah.RowRoots) / 2
 }
 
 // MinDataAvailabilityHeader returns the minimum valid data availability header.
--- a/pkg/da/data_availability_header_test.go
+++ b/pkg/da/data_availability_header_test.go
@@ -214,6 +214,47 @@ func Test_DAHValidateBasic(t *testing.T) {
 	}
 }
 
+func TestSquareSize(t *testing.T) {
+	type testCase struct {
+		name    string
+		dah     DataAvailabilityHeader
+		want    int
+		wantErr bool
+	}
+
+	testCases := []testCase{
+		{
+			name:    "min data availability header has an original square size of 1",
+			dah:     MinDataAvailabilityHeader(),
+			want:    1,
+			wantErr: false,
+		},
+		{
+			name:    "max data availability header has an original square size of 128",
+			dah:     maxDataAvailabilityHeader(),
+			want:    128,
+			wantErr: false,
+		},
+		{
+			name:    "returns an error if the number of row roots is not divisible by two",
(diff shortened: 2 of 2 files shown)


pkg/da/data_availability_header.go
```diff
@@ -167,6 +167,14 @@ func (dah *DataAvailabilityHeader) IsZero() bool {
 	return len(dah.ColumnRoots) == 0 || len(dah.RowRoots) == 0
 }
 
+// SquareSize returns the number of rows in the original data square.
+func (dah *DataAvailabilityHeader) SquareSize() (int, error) {
+	if len(dah.RowRoots)%2 != 0 {
+		return 0, fmt.Errorf("len row roots is not divisible by 2: %d", len(dah.RowRoots))
+	}
+	return len(dah.RowRoots) / 2, nil
```
Wondertan, line 175: Are we sure it's worth checking %2 and doing err != nil everywhere?

pkg/da/data_availability_header.go
```diff
@@ -167,6 +167,14 @@ func (dah *DataAvailabilityHeader) IsZero() bool {
 	return len(dah.ColumnRoots) == 0 || len(dah.RowRoots) == 0
 }
 
+// SquareSize returns the number of rows in the original data square.
+func (dah *DataAvailabilityHeader) SquareSize() (int, error) {
+	if len(dah.RowRoots)%2 != 0 {
```
staheri14, line 172: [optional] If we aim to do some sanity checks like this, then we may as well check whether the number of `RowRoots` and `ColRoots` are the same.

pkg/da/data_availability_header_test.go
```diff
@@ -239,3 +280,23 @@ func sortByteArrays(arr [][]byte) {
...
+
+// maxDataAvailabilityHeader returns a DataAvailabilityHeader the maximum square
+// size. This should only be used for testing.
+func maxDataAvailabilityHeader() (dah DataAvailabilityHeader) {
+	shares := generateShares(appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound)
+	eds, err := ExtendShares(shares)
+	if err != nil {
+		panic(err)
+	}
+	dah = NewDataAvailabilityHeader(eds)
```
staheri14, line 292: There seems to be an assignment mismatch.

pkg/da/data_availability_header_test.go
```diff
@@ -239,3 +280,23 @@ func sortByteArrays(arr [][]byte) {
...
+	return dah
+}
+
+// invalidDah returns a DataAvailabilityHeader with an odd number of row roots.
+// This should only be used for testing.
+func invalidDah() (dah DataAvailabilityHeader) {
+	dah = MinDataAvailabilityHeader()
+	dah.RowRoots = append(dah.RowRoots, bytes.Repeat([]byte{1}, 32))
```
staheri14, line 300: [Suggestion] Should we consider replacing direct access to individual fields of a `DataAvailabilityHeader`, such as `dah.RowRoots`, with a getter method to reduce unnecessary flexibility?

pkg/da/data_availability_header_test.go
```diff
@@ -214,6 +214,47 @@ func Test_DAHValidateBasic(t *testing.T) {
...
+		{
+			name:    "min data availability header has an original square size of 1",
+			dah:     MinDataAvailabilityHeader(),
+			want:    1,
+			wantErr: false,
+		},
+		{
+			name:    "max data availability header has an original square size of 128",
+			dah:     maxDataAvailabilityHeader(),
+			want:    128,
```
staheri14, line 235: [Suggestion] I think it would be better to derive the hardcoded value of `128` from the `appconsts.DefaultSquareSizeUpperBound` instead (as a function of it), as the constants may change in the future.

rootulp: Resolved via https://github.com/celestiaorg/celestia-app/pull/2095/commits/8ef4bd6158cf8338f92a44f87e57a2cfcd80254b

pkg/da/data_availability_header_test.go
rootulp: Deleted this function in https://github.com/celestiaorg/celestia-app/pull/2095/commits/8ef4bd6158cf8338f92a44f87e57a2cfcd80254b

pkg/da/data_availability_header.go
rootulp: Removed sanity checks b/c @Wondertan proposed removing the error return parameter

pkg/da/data_availability_header.go
rootulp: Defer to you b/c AFAIK there are no other usecases for this helper outside celestia-node.

Removed return parameter in https://github.com/celestiaorg/celestia-app/pull/2095/commits/8ef4bd6158cf8338f92a44f87e57a2cfcd80254b

cmwaters (approved):

pkg/da/data_availability_header_test.go
staheri14: My suggestion was something like the following:
```
want: appconsts.DefaultSquareSizeUpperBound
```
instead of:

```
want: 128
```

staheri14 (approved):
LGTM!

pkg/da/data_availability_header_test.go
rootulp: Sorry resolved via https://github.com/celestiaorg/celestia-app/pull/2095/commits/49dfa2e9271af22cb9ac5869ee378c9aa6d23670

Issue #2098, opened 2023-07-13
Txs endpoint returns error on range operators


nieejl:
### Summary of Bug

Endpoint to query transactions, fails when defining height ranges on transactions (tx).
Path: /cosmos/tx/v1beta1/txs

Tendermint has support for range queries using the >= and <= operators. Supported formats are described here: https://docs.tendermint.com/v0.34/rpc/#/Websocket/subscribe
Until recently (v. 0.34.20 , possibly later), this worked for Celestia nodes as well, but seems to be broken in v.0.34.28. This seems to have changed after the update to mocha-3.

Adding a filter for <= 11140, is expected to return all transactions where the height is less than 11140, but instead returns an error.
https://api-mocha.pops.one/cosmos/tx/v1beta1/txs?events=message.sender=%27celestiavaloper1amm0umxqxxw8j2q60x78xs32s089606sdmvx92%27&events=tx.height%3C=11140

Output:
```
{
  "code": 3,
  "message": "invalid event; event tx.height\u003c=11140 should be of the format: {eventType}.{eventAttribute}={value}",
  "details": [
  ]
}
```

I included a number of example queries below, to rule out some sources of error, and to showcase expected behavior:

Removing less than operator, returns the correct result:
https://api-mocha.pops.one/cosmos/tx/v1beta1/txs?events=message.sender=%27celestiavaloper1amm0umxqxxw8j2q60x78xs32s089606sdmvx92%27&events=tx.height=11137

Node similar to pops node which fails on same query, to rule out any configuration issues:
https://api-t.celestia.nodestake.top/cosmos/tx/v1beta1/txs?events=message.sender=%27celestiavaloper1amm0umxqxxw8j2q60x78xs32s089606sdmvx92%27&events=tx.height%3C=11140

Removing the range query, returns all transactions: https://api-mocha.pops.one/cosmos/tx/v1beta1/txs?events=message.sender=%27celestiavaloper1amm0umxqxxw8j2q60x78xs32s089606sdmvx92%27

Working example on a previous version (v 0.34.20): https://api.mocha.celestia.counterpoint.software/cosmos/tx/v1beta1/txs?events=message.sender=%27celestiavaloper1amm0umxqxxw8j2q60x78xs32s089606sdmvx92%27&events=tx.height%3C=57700

Working example on cosmos node (using v.0.34.29) at time of writing:
[...]

PR #2103, opened 2023-07-13
chore: always enable API in local dev scripts

rootulp:
While debugging https://github.com/celestiaorg/celestia-app/issues/2098 I frequently applied this patch to previous releases. Since these scripts exist for development purposes, I think it would save devs time by always enabling the API.


cmwaters (approved):
Is gRPC enabled by default? It would be good to ensure that this is also enabled when running test nodes

Issue #2071, opened 2023-07-10
Remove `fullNamespace`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/5bcf4ad51756e5f31dd7d9ea140844063b6ce391/pkg/square/builder.go#L361-L364

## Prerequisites
- [x] Merge https://github.com/celestiaorg/celestia-core/pull/1030 
- [x] Cut a new release of celestia-core: https://github.com/celestiaorg/celestia-core/releases/tag/v1.25.0-tm-v0.34.28

## Proposal
1. Bump to the new release of celestia-core
2. Replace usage of `fullNamespace` in celestia-app with celestia-core's exported `Namespace` method

MariusTanase:
I'm not aware of the codebase but I assume that the feat: add Namespace method to blob will work as intended(in the merge)

As for the Proposal 2, I'm assuming that this function will do the trick

func fullNamespace(blob core.Blob) []byte {
	return blob.Namespace()
}

Let me know if you have any bug.

PR #2104, opened 2023-07-14
refactor: use `blob.Namespace()` from celestia-core

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2071

### DIFF
--- a/pkg/square/builder.go
+++ b/pkg/square/builder.go
@@ -135,7 +135,7 @@ func (b *Builder) Export() (Square, error) {
 	// sort the blobs in order of namespace. We use slice stable here to respect the
 	// order of multiple blobs within a namespace as per the priority of the PFB
 	sort.SliceStable(b.Blobs, func(i, j int) bool {
-		return bytes.Compare(FullNamespace(b.Blobs[i].Blob), FullNamespace(b.Blobs[j].Blob)) < 0
+		return bytes.Compare(b.Blobs[i].Blob.Namespace(), b.Blobs[j].Blob.Namespace()) < 0
 	})
 
 	// write all the regular transactions into compact shares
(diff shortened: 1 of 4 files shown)


pkg/square/builder.go
```diff
@@ -366,11 +366,6 @@ func (b *Builder) IsEmpty() bool {
 	return b.TxCounter.Size() == 0 && b.PfbCounter.Size() == 0
 }
 
-// TODO: celestia-core should provide this method for `Blob`s
```
SweeXordious, line 369: [Question][Almost unrelated]
Why does core need to know about blobs and namespaces?

rootulp: hmm good question. Basic types like `Blob` and `BlobTx` are defined in celestia-core 

e.g. https://github.com/celestiaorg/celestia-core/blob/d5cd08c4e13a66acc087eda5d43109928e630846/types/block.go#L1039-L1055

So at least for now celestia-core is aware of blobs

SweeXordious: Thanks. Just checked, apparently, that's only used for tests

rootulp: We added `Namespace()` as a method to `Blob` in celestia-core as a convenience. Perhaps you're missing this context: https://github.com/celestiaorg/celestia-core/pull/1030

We can create a separate issue to explore moving types like `Blob` and `BlobTx` to celestia-app but TBH I don't expect it to be possible without a large refactor.

SweeXordious: Aaah just found where it's needed, thanks :+1:

SweeXordious (approved):

test/util/malicious/out_of_order_builder.go
```diff
@@ -72,14 +72,14 @@ func OutOfOrderExport(b *square.Builder) (square.Square, error) {
 	// sort the blobs in order of namespace. We use slice stable here to respect the
 	// order of multiple blobs within a namespace as per the priority of the PFB
 	sort.SliceStable(b.Blobs, func(i, j int) bool {
-		return bytes.Compare(square.FullNamespace(b.Blobs[i].Blob), square.FullNamespace(b.Blobs[j].Blob)) < 0
+		return bytes.Compare(b.Blobs[i].Blob.Namespace(), b.Blobs[j].Blob.Namespace()) < 0
```
staheri14, line 75: LGTM! what is the motivation behind this change? could not find it in the linked issue. Thanks.

staheri14 (approved):

PR #2116, opened 2023-07-18
fix: txsim uses the same gas estimation method

cmwaters:
This also simplifies some of the gas consumption logic between the ante handler and blob keeper

### DIFF
--- a/test/txsim/sequence.go
+++ b/test/txsim/sequence.go
@@ -42,9 +42,9 @@ type Operation struct {
 }
 
 const (
-	// set default gas limit to cover the costs of most transactions
-	// At 0.1 utia per gas, this equates to 20000utia per transaction
-	DefaultGasLimit = 200000
+	// Set the default gas limit to cover the costs of most transactions.
+	// At 0.1 utia per gas, this equates to 20_000utia per transaction.
+	DefaultGasLimit = 200_000
 )
 
 // ErrEndOfSequence is a special error which indicates that the sequence has been terminated
--- a/test/txsim/sequence.go
+++ b/test/txsim/sequence.go
@@ -43,8 +43,8 @@ type Operation struct {
 
 const (
 	// set default gas limit to cover the costs of most transactions
-	// At 0.001 utia per gas, this equates to 1000utia per transaction
-	DefaultGasLimit = 1000000
+	// At 0.1 utia per gas, this equates to 20000utia per transaction
+	DefaultGasLimit = 200000
 )
 
 // ErrEndOfSequence is a special error which indicates that the sequence has been terminated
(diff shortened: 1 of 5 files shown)


evan-forbes (dismissed):

test/txsim/sequence.go
```diff
@@ -43,8 +43,8 @@ type Operation struct {
 
 const (
 	// set default gas limit to cover the costs of most transactions
-	// At 0.001 utia per gas, this equates to 1000utia per transaction
-	DefaultGasLimit = 1000000
+	// At 0.1 utia per gas, this equates to 20000utia per transaction
+	DefaultGasLimit = 200000
```
rootulp, line 47: [naive question] I think the relationship may be: 200_000 default gas limit * 0.1 utia per gas = 20_000 utia per transaction.

So is this saying that most transactions will consume less than 20_000 utia?

x/blob/types/payforblob.go
```diff
@@ -21,6 +22,20 @@ import (
 const (
 	URLMsgPayForBlobs = "/celestia.blob.v1.MsgPayForBlobs"
 	ShareSize         = appconsts.ShareSize
+
+	// PFBGasFixedCost is a rough estimation of the gas that arises from several other places:
+	// - signature verification
+	// - tx size
+	// - read access to accounts
+	// From modelling PFB gas consumption we arrive at the following formula:
+	// 8 * 512 * shares + 65000. This has a correlation coefficient of 0.996
```
rootulp, line 31: [question] In this context does `shares` refer to the # of shares occupied by the blob? If yes, I interpret `8 * 512 * shares` as a variable cost so it's not clear to me why this impacts `PFBGasFixedCost`

rootulp (approved):
No blocking feedback. Are there already tests that verify that `DefaultEstimateGas` always overestimates the gas required for a transaction and never underestimates? If not, I think that should be a follow-up issue. Ideally the test in the follow-up fuzzes w/ various transaction types and blob sizes.

x/blob/types/payforblob.go
cmwaters: This is the linear model y = mx + c. The c component is extracted out into `PFBGasFixedCost`. Does that make sense?

x/blob/types/payforblob.go
```diff
@@ -21,6 +22,20 @@ import (
...
+	// To be conservative, we round up to 75000 because the first tx always
+	// takes up 10,000 more gas.
+	PFBGasFixedCost = 75000
+
+	// BytesPerBlobInfo is the amount of extra bytes in information a blob
+	// adds to the size of the underlying transaction
+	BytesPerBlobInfo = 70
```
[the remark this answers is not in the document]
cmwaters: Yeah surprisingly this wasn't constant. I was getting values between 65-68 bytes extra per blob. Thus I rounded it up to 70

evan-forbes (dismissed):

cmwaters:
> Are there already tests that verify that DefaultEstimateGas always overestimates the gas required for a transaction and never underestimates?

I used the tx sim to test this but could extract out a specific test for this

evan-forbes (approved):

Issue #2120, opened 2023-07-18
(untyped float constant 31556952000000000) to type int


rootulp:
### Summary of Bug

```
error:
# github.com/celestiaorg/celestia-app/x/mint/types
../../go/pkg/mod/github.com/celestiaorg/celestia-app@v1.0.0-rc9/x/mint/types/constants.go:15:27: cannot convert NanosecondsPerSecond * SecondsPerMinute * MinutesPerHour * HoursPerDay * DaysPerYear (untyped float constant 31556952000000000) to type int
make: *** [Makefile:22: build] Fehler 1
```

### Version

main

### Steps to Reproduce

I think: attempt to build celestia-app on a 32 bit machine

### For Admin Use

- [X] Not duplicate issue
- [X] Appropriate labels applied
- [X] Appropriate contributors tagged
- [X] Contributor assigned/self-assigned

jcstein:
tagging @waynewayner for more context and adding some here:

he is running into issue building on his arm32 machine. should it work on arm32? he tried to install it without modification to the makefile, but it built for amd64 instead of arm32. ideas?
error:

```
# github.com/celestiaorg/celestia-app/x/mint/types
../../go/pkg/mod/github.com/celestiaorg/celestia-app@v1.0.0-rc9/x/mint/types/constants.go:15:27: cannot convert NanosecondsPerSecond * SecondsPerMinute * MinutesPerHour * HoursPerDay * DaysPerYear (untyped float constant 31556952000000000) to type int
make: *** [Makefile:22: build] Fehler 1
```

his makefile:
```
## build: Build celestia-node binary.
build:
    @echo "--> Building Celestia"
    @GOOS=linux GOARCH=arm go build -o build/ ${LDFLAGS} ./cmd/celestia
.PHONY: build
```

rootulp:
Thanks for reporting @WayneWayner ! This appears to be a celestia-app bug which should be fixed by #2121

WayneWayner:
There is still a overflow issues on the latest commit: 

```
$ go get  github.com/celestiaorg/celestia-app@dec925336384241e55056937633018c64b2c0910
go: downloading github.com/celestiaorg/celestia-app v1.0.0-rc0.0.20230718175129-dec925336384
go: downloading google.golang.org/grpc v1.56.2
go: downgraded github.com/celestiaorg/celestia-app v1.0.0-rc9 => v1.0.0-rc0.0.20230718175129-dec925336384
[...]

PR #2121, opened 2023-07-18
fix: explicit int64 for NanosecondsPerYear

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2120

On a 32-bit system, an int can store values up to 2,147,483,647, which is much smaller than 31,556,952,000,000,000 (number of nanoseconds in a year).

After this merges I will cherry-pick it to the `v1.x` release branch


cmwaters (approved):
LGTM.

On a more meta note, if we are explicitly supporting 32 bit systems, then perhaps we need to extend our test suite to cover this

evan-forbes (approved):

Issue #2119, opened 2023-07-18
Add a min gas price gRPC endpoint for node and others to query for fee calculation


cmwaters:
### Summary

Add a gRPC endpoint to query the min gas price of the validator as set in the validator's app.toml file (and as also part of `BaseApp`) 

### Problem Definition

Validators currently don't make it transparent what the minimum gas price they are willing to accept for a transaction is. The only way a client can work out whether the tx has set a sufficient gas price is to try to submit a transaction and then parse the error message which will be of the form:

```
sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
```

### Proposal

Add a gRPC endpoint returning the value and a function for parsing the error message for node to handle. 

If we go down the road of adopting https://github.com/celestiaorg/celestia-app/issues/1621, then we could also continue to use this endpoint.

The other advantage with this change is nodes that set a fee less than the `DefaultMinGasPrice` can actually be used by the blob module rather than overpaying on gas-price. 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #2122, opened 2023-07-18
feat: add node service which exposes min gas price

cmwaters:
Closes: #2119


cmwaters:
Tested this on a local instance

image

rootulp (approved):
fancy :shipit:

evan-forbes (approved):

PR #2131, opened 2023-07-19
feat: add the ability to trace gas consumption

evan-forbes:
## Overview

bringing back the gas monitor, not sure if we should merge this, but I'm using to regenerate all the gas charts, which can be found in #2132 

## Checklist

### DIFF
--- a/tools/gasmonitor/gas_monitor.go
+++ b/tools/gasmonitor/gas_monitor.go
@@ -108,11 +108,11 @@ func (gm *GasConsumptionTrace) RefundGas(amount sdk.Gas, descriptor string) {
 
 // AnteHandle replaces the gas meter with the provided context with one that
 // monitors the amount of gas consumed it fulfills the ante.Decorator interface.
-func (h *Decorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+func (d *Decorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
 	if ctx.IsCheckTx() || ctx.IsReCheckTx() {
 		return next(ctx, tx, simulate)
 	}
 	trace, ctx := NewGasConsumptionTrace(ctx, tx)
-	h.Traces = append(h.Traces, trace)
+	d.Traces = append(d.Traces, trace)
 	return next(ctx, tx, simulate)
 }
--- a/tools/gasmonitor/gas_monitor.go
+++ b/tools/gasmonitor/gas_monitor.go
@@ -7,6 +7,7 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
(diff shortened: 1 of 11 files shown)


tools/gasmonitor/gas_monitor.go
```diff
@@ -0,0 +1,118 @@
...
+
+// ConsumeGas wraps the underlying RefundGas method to record the amount of gas
+// refunded.
+func (gm *GasConsumptionTrace) RefundGas(amount sdk.Gas, descriptor string) {
+	gm.Readings = append(gm.Readings, Reading{Description: descriptor, Refund: true, Amount: amount})
+	gm.GasMeter.RefundGas(amount, descriptor)
+}
+
+// AnteHandle replaces the gas meter with the provided context with one that
+// monitors the amount of gas consumed it fulfills the ante.Decorator interface.
```
rootulp, line 110:
```suggestion
// AnteHandle replaces the gas meter in the provided context with one that
// monitors the amount of gas consumed. It fulfills the ante.Decorator interface.
```

tools/gasmonitor/integration_test.go
```diff
@@ -0,0 +1,270 @@
...
+	for _, tc := range tests {
+		s.Run(tc.name, func() {
+			signer := blobtypes.NewKeyringSigner(s.cctx.Keyring, s.accounts[count], s.cctx.ChainID)
+			res, err := blob.SubmitPayForBlob(context.TODO(), signer, s.cctx.GRPCClient, tc.blobs, tc.opts...)
+			require.NoError(t, err)
+			require.NotNil(t, res)
+			require.Equal(t, abci.CodeTypeOK, res.Code, res.Codespace, res.Logs)
+		})
+		count++
+	}
```
rootulp, line 257: [optional refactor] Is it possible to remove the `count` variable entirely?

```suggestion
	for i, tc := range tests {
		s.Run(tc.name, func() {
			signer := blobtypes.NewKeyringSigner(s.cctx.Keyring, s.accounts[i], s.cctx.ChainID)
			res, err := blob.SubmitPayForBlob(context.TODO(), signer, s.cctx.GRPCClient, tc.blobs, tc.opts...)
			require.NoError(t, err)
			require.NotNil(t, res)
			require.Equal(t, abci.CodeTypeOK, res.Code, res.Codespace, res.Logs)
		})
	}
```

rootulp (dismissed):
Overall LGTM. Nothing in the PR seems particularly difficult to maintain and the Python script is a separate Gist so it seems safe to merge but I'm curious what other reviewers think.

tools/gasmonitor/integration_test.go
evan-forbes: yeah good catch forgot about that lil guy

tools/gasmonitor/integration_test.go
```diff
@@ -0,0 +1,270 @@
...
+			},
+		},
+		{
+			"small random typical",
+			blobfactory.ManyRandBlobs(t, tmrand.NewRand(), 1, 1),
+			[]blobtypes.TxBuilderOption{
+				blobtypes.SetFeeAmount(sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(9000)))),
+				blobtypes.SetGasLimit(90_000),
+			},
+		},
```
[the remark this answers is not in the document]
evan-forbes: yeah exactly, I think I just forgot to fix the names, will go do now

rootulp (approved):

tools/gasmonitor/integration_test.go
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/2131/commits/6ac9df335bbcaa63048562b77f70103215f82456

cmwaters:
> Do we see ourselves tracing the consumption in the future? If not, we should close again. Opening the discussion to the reviewers.

I guess do we see the calculations of gas being altered in the future (I'm not sure if an application can even set the `GasConfig` currently).

evan-forbes:
something like this might be useful if we ever run into another gas bug, or if we ever want to fine tune resource pricing. but until then... this is a lot of extra code!

PR #2135, opened 2023-07-19
fix: int overflow for 32 bit machines

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2129

https://github.com/celestiaorg/celestia-app/pull/2121 didn't actually fix the issue because the expression 

```go
NanosecondsPerYear = int64(NanosecondsPerSecond * SecondsPerYear)
```
was multiplying two `int`s together prior to the cast to `int64`. Since the product of those two `int`s overflows an `int32`, the value would overflow prior to the cast.

The fix proposed in this PR converts SecondsPerYear to an `int64`. Now the expression multiplies an `int` and `int64` so the product is an `int64` which won't overflow. 

## Testing

I added this to the Makefile

```go
## build-arm: Build the celestia-appd binary into the ./build directory. Target the linux OS and ARM (32-bit) architecture.
build-arm: mod
	@cd ./cmd/celestia-appd
	@mkdir -p build/
	@GOOS=linux GOARCH=arm go build $(BUILD_FLAGS) -o build/ ./cmd/celestia-appd
	@go mod tidy -compat=1.20
.PHONY: build-arm
```

Before

```shell
$ make build-arm
--> Updating go.mod
# github.com/celestiaorg/celestia-app/x/mint/types
x/mint/types/constants.go:15:29: NanosecondsPerSecond * SecondsPerYear (constant 31556952000000000 of type int) overflows int
make: *** [build-arm] Error 1
```

After

```shell
$ make build-arm
--> Updating go.mod
```

### DIFF
--- a/x/mint/types/constants.go
+++ b/x/mint/types/constants.go
@@ -11,8 +11,8 @@ const (
 	// value isn't 365 because 97 out of 400 years are leap years. See
 	// https://en.wikipedia.org/wiki/Year
 	DaysPerYear        = 365.2425
-	SecondsPerYear     = int(SecondsPerMinute * MinutesPerHour * HoursPerDay * DaysPerYear) // 31,556,952
-	NanosecondsPerYear = int64(NanosecondsPerSecond * SecondsPerYear)                       // 31,556,952,000,000,000
+	SecondsPerYear     = int64(SecondsPerMinute * MinutesPerHour * HoursPerDay * DaysPerYear) // 31,556,952
+	NanosecondsPerYear = int64(NanosecondsPerSecond * SecondsPerYear)                         // 31,556,952,000,000,000
 
 	InitialInflationRate = 0.08
 	DisinflationRate     = 0.1


cmwaters (approved):

evan-forbes (approved):

jcstein:
Nick is running into this error on an android, I'd imagine maybe the `make build-arm` isn't reflected in the makefile for node?

rootulp:
You're right the celestia-node [Makefile](https://github.com/celestiaorg/celestia-node/blob/main/Makefile) doesn't have a `make build-arm` command. Clarifying question: is Nick trying to run celestia-node or celestia-app on Android? 

If celestia-node, I think we should open an issue in celestia-node repo.
If celestia-app, I think we need a new issue in celestia-app and I can investigate

rootulp:
Circling back on this for visibility @jcstein:
- @nwhite1 was trying to run celestia-node on Android.
- The earliest celestia-app release with this fix was `v1.0.0-rc10` so releases >= `v1.0.0-rc10` shouldn't encounter this error. Releases < `v1.0.0-rc10` will encounter this error.
- I looked at a few releases of celestia-node to see what version of celestia-app they depend on. 
  - These celestia-node releases should not work: `v0.11.0-rc6`, `v0.11.0-rc7`, `v0.11.0-rc8`
  - These celestia-node releases should work: `v0.11.0-rc8-arabica-improvements`, `v0.11.0-rc9`

If ya'll still observe this error on celestia-node releases that should work, can you please open a new GH issue?

jcstein:
hey @rootulp thanks for following up here. i don’t think an issue is needed, and will see if one of the newer versions works!

jcstein:
assuming it doesn’t happen on a release where it is fixed

Issue #2143, opened 2023-07-21
ValidateBlobNamespace: cleaning up and future-proofing


ivan-gavran:
There seems to be some confusion about blob namespace validation:
- there are two functions called `ValidateBlobNamespace`: [one](https://github.com/celestiaorg/celestia-app/blob/341006e5a51e6646b2c1eeb91a3ec9781cb6adc7/pkg/namespace/namespace.go#L87-L87) (A) is a member function of the `Namespace` structure, and [the other](https://github.com/celestiaorg/celestia-app/blob/341006e5a51e6646b2c1eeb91a3ec9781cb6adc7/x/blob/types/payforblob.go#L135-L135) (B) takes `Namespace` as an argument. The two functions are almost identical, the only difference being that B checks if the namespace version is `NamespaceVersionZero`. 
- Interestingly, both functions are used in a similar context: A is called from within `ValidateBlobTx` (which is called both from `ProcessProposal` and `CheckTx`). B is called from within `ValidateBasic` of `MsgPayForBlobs`, which is also called from `ValidateBlobTx`.
- My guess is that calling both of these functions is unnecessary and that B should be the used one. 

Finally, a question of future-proving: would it make sense to replace the explicit check for version 0, 
```golang
if ns.Version != appns.NamespaceVersionZero {
		return ErrInvalidNamespaceVersion
	}
```
with a check for whether the `ns.Version` is among the supported user-namespace versions? (Along the lines of how it was done  for share versions, [here](https://github.com/celestiaorg/celestia-app/blob/341006e5a51e6646b2c1eeb91a3ec9781cb6adc7/x/blob/types/payforblob.go#L274-L276).)

There is another function which validates namespace versions, [ValidateVersion](https://github.com/celestiaorg/celestia-app/blob/341006e5a51e6646b2c1eeb91a3ec9781cb6adc7/pkg/namespace/namespace.go#L104-L104). In this validation, both zero-namespace and max-namespace are allowed (so, it is checking for any possible versions, and not just user-created blobs).  It would probably make sense to change the name of that function into something more concrete, e.g., `ValidateVersionSupported`.

relevant commit: [v1.0.0-rc6](https://github.com/celestiaorg/celestia-app/tree/v1.0.0-rc6)

PR #2145, opened 2023-07-21
refactor!: un-export `namespace.ValidateBlobNamespace`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2143

De-duplicate `ValidateBlobNamespace` and export it from `x/blob/types` package. The namespace package needs to define its own `isBlobNamespace` because it can't import `ValidateBlobNamespace` from `x/blob/types` as that would introduce a circular dependency.


cmwaters (approved):

x/blob/types/payforblob_test.go
```diff
@@ -174,29 +174,29 @@ func TestValidateBasic(t *testing.T) {
...
 			msg:     tailPaddingMsg,
-			wantErr: ErrTailPaddingNamespace,
+			wantErr: ErrInvalidBlobNamespace,
 		},
 		{
-			name:    "transaction namspace namespace id",
+			name:    "tx namespace",
 			msg:     txNamespaceMsg,
-			wantErr: ErrReservedNamespace,
+			wantErr: ErrInvalidBlobNamespace,
```
evan-forbes, line 189: I could be wrong, but I think switching all of these to the same value result in users seeing a single error code for many different errors. That's ok, but what do we gain for wrapping each of these? Just that we see their path faster?

admittedly, I've just added errors to our giant list of official errors in the past, but perhaps we're at the point where documenting all of these is needed. For example, what is the difference between `ErrInvalidBlobNamespace`, `ErrInvalidNamespace`, `ErrTxNamespace`. I basically have to look at the code each time.

rootulp: You're right, after this PR user's will see a single error code: `11138` for every case with `ErrInvalidBlobNamespace`.

> what do we gain for wrapping each of these? Just that we see their path faster?

Hmm not sure what you mean by path. I could see an argument for preserving the old error codes and maintaining some level of duplication of `ValidateBlobNamespace()`.

I think the pain point here is that ideally we have functions exposed by the `namespace` package that return error strings and not Cosmos SDK errors (because the namespace package should remain sufficiently general). On the other hand, x/blob/types/payforblob.go should use custom defined Cosmos SDK errors so that the error codes and strings are unique and informative instead of the more general "invalid blob namespace". 

I'll take another stab at this.

Wondertan:
@rootulp, could you open similar PR to celeria-node? I think we may have similar issue, as our code is mostly a port.

pkg/namespace/namespace.go
```diff
@@ -83,25 +83,8 @@ func (n Namespace) Bytes() []byte {
...
-		return fmt.Errorf("invalid blob namespace: %v cannot use tail padding namespace ID", n.Bytes())
-	}
-
-	return nil
-}
-
-// validateVersion returns an error if the version is not supported.
-func validateVersion(version uint8) error {
+// validateVersionSupported returns an error if the version is not supported.
+func validateVersionSupported(version uint8) error {
```
rootulp, line 87: [note to reviewers] this function was renamed per the comment in https://github.com/celestiaorg/celestia-app/issues/2143

I thought the original name was clear but I'm open to alternative naming suggestions.

rootulp:
> @rootulp, could you open similar PR to celeria-node? I think we may have similar issue, as our code is mostly a port.

As far as I can see, celestia-node only has one reference to `ValidateBlobNamespace` [here](https://github.com/celestiaorg/celestia-node/blob/6a6c68c42ff940467476c0a6876e7d986b0e84aa/blob/blob_test.go#L48) which is the `ValidateBlobNamespace` that remains exported in this PR.

Wondertan:
@rootulp, I mean the slice.Contains if should also be added to our validate function

x/blob/types/payforblob_test.go
```diff
@@ -244,6 +245,7 @@ func TestValidateBasic(t *testing.T) {
 				space, code, log := sdkerrors.ABCIInfo(err, false)
 				assert.Equal(t, tt.wantErr.Codespace(), space)
 				assert.Equal(t, tt.wantErr.ABCICode(), code)
+				fmt.Printf("code %v", code)
```
evan-forbes, line 248: I'm also fine with merging this PR and iterating if we have issues open, and this line ofc :smile:

x/blob/types/payforblob_test.go
rootulp: Took a different approach in https://github.com/celestiaorg/celestia-app/pull/2145/commits/14332d221fb0aed6f57ca08537d6395fc79cf521

x/blob/types/payforblob_test.go
rootulp: Resolved via https://github.com/celestiaorg/celestia-app/pull/2145/commits/6e4c660974f2933c7a8e2923638dfa1d71a862a4

evan-forbes (approved):

pkg/namespace/random_blob.go
```diff
@@ -36,3 +35,25 @@ func RandomBlobNamespaces(rand *tmrand.Rand, count int) (namespaces []Namespace)
 	}
 	return namespaces
 }
+
+// isBlobNamespace returns an true if this namespace is a valid user-specifiable
```
cmwaters, line 39:
```suggestion
// isBlobNamespace returns true if this namespace is a valid user-specifiable
```

cmwaters (approved):

PR #2151, opened 2023-07-21
doc: fix incorrect usage of `MaxBytes`

rootulp:
Fix a small issue in ADR-021. I don't think these lines meant to use `MaxBytes` b/c there is a separate field `MaxBlockSizeBytes` that acts as an upper bound for `MaxBytes`.


evan-forbes (approved):

cmwaters (approved):
Good catch

Issue #2108, opened 2023-07-14
Export function to determine max blob bytes in a block


rootulp:
## Context

celestia-node would like to add a validation step on blobs that checks that the blob size does not exceed the maximum possible bytes that a single blob can occupy in a block.

## Proposal

Export a function from celestia-app that informs celestia-node what is the maximum possible bytes in a block. The amount of data in a block is constrained by two parameters (see [ADR-021](https://github.com/celestiaorg/celestia-app/blob/e06116cb4cd2b3d535fc5de0df9e192bb6902692/docs/architecture/adr-021-restricted-block-size.md#decision)): `MaxBytes` and `GovMaxSquareSize`.

Rough steps:
1. Learn the current value for the params `GovMaxSquareSize` and `MaxBytes`. Note celestia-node has access to the celestia-app version but both these params are governance modifiable so they require a state access
2. Compute the max blob size that can fit into `GovMaxSquareSize` 
3. Compute the max blob size that can fit into `MaxBytes`
4. Take the minimum of steps 2 and 3

cc: @Wondertan @renaynay @musalbas

rootulp:
I think the proposed function will return an upper bound for blob size but it won't be an exact maximum size due to the number of bytes occupied by the PFB tx. In other words, smaller blobs than this upper bound may also be invalid. 

We should consider fuzz testing by attempting to submit blobs of various sizes and confirming that blob sizes > the returned value from the proposed function(s) are never included in a block.

Wondertan:
@rootulp, I am fine with such solution initially. Still, would be nice to think of the long term approach and if not complicated to aim to implement it first.

The stateful approach is annoying, but we can keep it inside stateful blob.Service. It already imports chain state methods and can query any param on start and reuse it throughout its lifespan.

rootulp:
> we can keep it inside stateful blob.Service

Specifically https://github.com/celestiaorg/celestia-node/blob/main/blob/service.go ? 

> can query any param on start

Does it already do this? Can you please link to an example? I'm not sure how celestia-node knows which major version of celestia-app it is connected to?

Wondertan:
My bad, I mean technically/in-theory can. It does not do this right now. The BlobSubmitter api can be extended with MaxBlobSize or something method that it could use.

evan-forbes:
[...]

PR #2152, opened 2023-07-22
feat: blob size upper bound

rootulp:
v0 implementation of https://github.com/celestiaorg/celestia-app/issues/2108
In order to resolve the TODOs, we need to make celestia-node aware of the governance params.

cc: @vgonkivs


rootulp:
I had a synchronous call w/ @evan-forbes and this PR may be obsolete if we had stateless validation to PFB transactions.

Issue #2157, opened 2023-07-24
review: Builder Init - consider not returning builder


tuxcanfly:
> Why are we returning the `Builder`? It isn't a new builder and the methods are just modifying the original?
> 
> Can't we just do this?

```suggestion
func (b *Builder) Init() (error) {
	if b.isCompactShare {
		return b.prepareCompactShare()
	} 
	return b.prepareSparseShare()
}
```

_Originally posted by @MSevey in https://github.com/rollkit/celestia-openrpc/pull/32#discussion_r1269600244_

PR #2165, opened 2023-07-25
refactor!: remove builder return param from init

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2157

A separate `Init()` function seems kinda like an anti pattern. If it must always be called after `NewBuilder`, why doesn't `NewBuilder` call it? @mojtaba-esk

### DIFF
--- a/pkg/shares/share_builder.go
+++ b/pkg/shares/share_builder.go
@@ -22,19 +22,23 @@ func NewEmptyBuilder() *Builder {
 	}
 }
 
-// NewBuilder returns a new share builder. Init() needs to be called right after
-// this method.
-func NewBuilder(ns appns.Namespace, shareVersion uint8, isFirstShare bool) *Builder {
-	return &Builder{
+// NewBuilder returns a new share builder.
+func NewBuilder(ns appns.Namespace, shareVersion uint8, isFirstShare bool) (*Builder, error) {
+	b := Builder{
 		namespace:      ns,
 		shareVersion:   shareVersion,
 		isFirstShare:   isFirstShare,
 		isCompactShare: isCompactShare(ns),
 	}
+	err := b.init()
+	if err != nil {
+		return nil, err
+	}
+	return &b, nil
 }
 
-// Init initializes the share builder. It should be called right after NewBuilder.
-func (b *Builder) Init() error {
+// init initializes the share builder by populating rawShareData.
+func (b *Builder) init() error {
 	if b.isCompactShare {
 		return b.prepareCompactShare()
 	}
--- a/pkg/shares/share_builder.go
+++ b/pkg/shares/share_builder.go
@@ -22,7 +22,8 @@ func NewEmptyBuilder() *Builder {
 	}
 }
 
-// Init() needs to be called right after this method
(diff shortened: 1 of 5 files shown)


pkg/shares/share_builder.go
```diff
@@ -32,18 +33,12 @@ func NewBuilder(ns appns.Namespace, shareVersion uint8, isFirstShare bool) *Buil
 	}
 }
 
-func (b *Builder) Init() (*Builder, error) {
+// Init initializes the share builder. It should be called right after NewBuilder.
+func (b *Builder) Init() error {
```
evan-forbes, line 37: apologies if I just forgot, but if this function should be called right after `NewBuilder` then why don't we just call it there?

evan-forbes (approved):

mojtaba-esk:
> A separate `Init()` function seems kinda like an anti pattern. If it must always be called after `NewBuilder`, why doesn't `NewBuilder` call it? @mojtaba-esk

Don't remember the rational behind it, I checked the code and probably the reason does not exist any more. So we can merge the two functions.

@evan-forbes ^

mojtaba-esk (changes requested):
LGTM, please go on with the proposal to merge the Init into the NewBuilder

pkg/shares/share_builder.go
rootulp: Resolved via https://github.com/celestiaorg/celestia-app/pull/2165/commits/cb86e485f193c363038fc598ded216c0f1863350

pkg/shares/share_builder.go
```diff
@@ -22,28 +22,27 @@ func NewEmptyBuilder() *Builder {
...
-	return &Builder{
+// NewBuilder returns a new share builder.
+func NewBuilder(ns appns.Namespace, shareVersion uint8, isFirstShare bool) (*Builder, error) {
+	b := Builder{
 		namespace:      ns,
 		shareVersion:   shareVersion,
 		isFirstShare:   isFirstShare,
 		isCompactShare: isCompactShare(ns),
 	}
+	err := b.init()
```
mojtaba-esk, line 33: looks good, could even be more concise by having the init in the if phrase like this:

```
if err := b.init(); err != nil {
    return nil, err
}
```

pkg/shares/share_builder_test.go
```diff
@@ -318,3 +311,10 @@ func TestShareBuilderImportRawData(t *testing.T) {
 		})
 	}
 }
+
+// mustNewBuilder returns a new builder with the given parameters. It fails the test if an error is encountered.
+func mustNewBuilder(t *testing.T, ns appns.Namespace, shareVersion uint8, isFirstShare bool) *Builder {
```
mojtaba-esk, line 316: smart idea

mojtaba-esk (dismissed):

pkg/shares/share_builder.go
rootulp: 109bff6aeee94f5a7bf7ab1b120343ea370ae945

mojtaba-esk (approved):

cmwaters (approved):
Yeah this makes sense to me

PR #2168, opened 2023-07-26
feat: migrate MsgSubmitProposal validity check to app as a ante handler

cmwaters:
This takes the PR https://github.com/celestiaorg/cosmos-sdk/pull/315/files which targeted the SDK and migrates it across to `celestia-app`.  This brings us one step closer to unforking us from the SDK

The only foreseeable change from this migration is that if a user is to locally run `ValidateBasic` it might show as being valid even although it gets rejected in `CheckTx` or `DeliverTx`.


app/ante/gov.go
```diff
@@ -0,0 +1,27 @@
...
+	govv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+)
+
+type GovProposalDecorator struct{}
+
+func NewGovProposalDecorator() GovProposalDecorator {
+	return GovProposalDecorator{}
+}
+
+// AnteHandle implements the AnteHandler interface. It ensures that MsgSubmitProposal
```
cmwaters, line 16:
```suggestion
// AnteHandle implements the AnteHandler interface. It ensures that MsgSubmitProposal has at least one message
```

evan-forbes (dismissed):

rootulp (dismissed):
[optional] would it be beneficial to port over a unit test that verifies:
1. a gov proposal with zero messages is invalid
2. a gov proposal with one message is valid

Ref: https://github.com/celestiaorg/cosmos-sdk/pull/315/files#diff-3d0ed5f02d7a5f2f3b4d7b3ab00c96748a4f862d7f70d9e9055572c415a70e2eR154

app/ante/gov_test.go
```diff
@@ -0,0 +1,75 @@
...
+			msg:    []types.Msg{msgEmptyProposal},
+			expErr: true,
+		},
+		{
+			name:   "good proposal; multiple messages in tx",
+			msg:    []types.Msg{msgProposal, msgSend},
+			expErr: false,
+		},
+		{
+			name:   "bad proposal; multiple messages in tx but proposal has no messages",
```
rootulp, line 55: 👍

rootulp (approved):

evan-forbes (approved):

Issue #1698, opened 2023-05-04
Move modified cosmos-sdk modules over from our fork to the app


evan-forbes:
We've resorted to modifying our fork of the cosmos-sdk in places, including the staking and the governance module, we should either copy those over, or find a way to incorporate the changes into the existing modules in a non-breaking way (similar to what we've done for the upgrade module)

[original comment](https://github.com/celestiaorg/cosmos-sdk/pull/315#pullrequestreview-1412647683)

PR #2169, opened 2023-07-26
feat!: handle registering evm addresses in QGB module

cmwaters:
Addresses: https://github.com/celestiaorg/celestia-app/issues/1698


proto/celestia/qgb/v1/tx.proto
```diff
@@ -0,0 +1,34 @@
...
+  // state machine
+  rpc RegisterEVMAddress(MsgRegisterEVMAddress)
+      returns (MsgRegisterEVMAddressResponse) {
+    option (google.api.http).get = "/qgb/v1/register_evm_address";
+  }
+}
+
+// MsgRegisterEVMAddress registers an evm address to a validator
+message MsgRegisterEVMAddress {
+  // The address of the validator
```
SweeXordious, line 25: might be good to specify whether it is an account address or a validator operator addres

test/util/common.go
```diff
@@ -355,7 +355,7 @@ func MakeTestMarshaler() codec.Codec {
 }
 
 // SetupFiveValChain does all the initialization for a 5 Validator chain using the keys here
-func SetupFiveValChain(t *testing.T) (TestInput, sdk.Context) {
+func SetupFiveValChain(t *testing.T, register bool) (TestInput, sdk.Context) {
```
SweeXordious, line 358: Why would we have an option to register? All validators should register an EVM address, no?

x/qgb/client/tx.go
```diff
@@ -0,0 +1,56 @@
...
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	cmd.AddCommand(CmdRegisterEVMAddress())
+
+	return cmd
+}
+
+func CmdRegisterEVMAddress() *cobra.Command {
```
SweeXordious, line 28: We can have this command to edit the EVM address for a validator (and also changing its name), but I thought that we were planning on including the message to register an EVM address as part of the create validator tx, then add a rule that a create validator should always have a register EVM address message.

x/qgb/client/tx.go
```diff
@@ -0,0 +1,56 @@
...
+send a new message overriding the previous one.
+`,
+		Args: cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			valAddress, evmAddress := args[0], args[1]
```
SweeXordious, line 44: Shouldn't we do more validation on the arguments at this level? so that a user doesn't end up sending a transaction and be a failed one because some address is not well formatted

x/qgb/keeper/keeper_valset.go
```diff
@@ -94,8 +94,15 @@ func (k Keeper) GetCurrentValset(ctx sdk.Context) (types.Valset, error) {
 
 		p := sdk.NewInt(k.StakingKeeper.GetLastValidatorPower(ctx, val))
 
-		// TODO make sure this  is always the case
-		bv := types.BridgeValidator{Power: p.Uint64(), EvmAddress: validator.EvmAddress}
+		evmAddress, exists := k.GetEVMAddress(ctx, val.String())
+		if !exists {
+			// validators may not have an EVM address associated with them.
+			// This is not enforced. We no longer include them from the validator
+			// set
```
SweeXordious, line 101: This changes the whole QGB security assumption, no? We rely on having all the validators signing, if not, then a Celestium becomes an ordinary DAC

x/qgb/client/tx.go
cmwaters: `NewMsgRegisterEVMAddress` below should sanitize the input (i.e. check that it is a ValAddress and a HexAddress)

cmwaters: Let me know if you think there is more validation that should be done i.e. maybe check size of the evm address

x/qgb/keeper/keeper_valset.go
cmwaters: Just for posterity (and correct me @SweeXordious if you are under a different impression): the decision here is that we always want to have the validators committed to some address. If no address is provided then we derive the EVM address from the cosmos sdk validator address. Ideally we have a single CLI command that registers both in the same tx.

x/qgb/types/msgs.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// GetSigner fulfills the sdk.Msg interface. The signer must be the validator address
+func (msg MsgRegisterEVMAddress) GetSigners() []sdk.AccAddress {
+	address, err := sdk.AccAddressFromBech32(msg.ValidatorAddress)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{address}
+}
```
evan-forbes, line 39: [blocking]
if we call this in the tests, it panics since `AccAddressFromBech32` can't handle a `celestiavalop` address. afaiu, this will stop anyone from being able to submit a tx with a MsgRegisterEVMAddress using their valop address. I think this is only working in the tests because we're not actually executing the tx with all of the antehandlers

I think this should work tho
```suggestion
// GetSigner fulfills the sdk.Msg interface. The signer must be the validator address
func (msg MsgRegisterEVMAddress) GetSigners() []sdk.AccAddress {
	valAddr, err := sdk.ValAddressFromBech32(msg.ValidatorAddress)
	if err != nil {
		panic(err)
	}
	return []sdk.AccAddress{sdk.AccAddress(valAddr)}
}
```

evan-forbes (commented):
only a few quick blocking comments/questions

[blocking]
do we also need to register the msg server here? https://github.com/celestiaorg/celestia-app/blob/4ec84e71d6bbf408d79bdec3b7ea97f8739f676c/x/qgb/module.go#L128-L132

given its importance, I think it would be good to have a test to ensure some of the guarantees around the new msg type, such as a rando account being unable to update the validator's evm address, and that the tx can get executed, wdyt?

x/qgb/types/msgs.go
evan-forbes: we also don't have to use this, as I'll defer to the author on how to test these things, but I wanted to make sure that the above section was working as expected, so I quickly wrote

```go
package qgb_test

import (
	"sync"
	"testing"

	"github.com/celestiaorg/celestia-app/test/util/testnode"
	qgbtypes "github.com/celestiaorg/celestia-app/x/qgb/types"
	"github.com/cosmos/cosmos-sdk/types/errors"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/suite"
	abci "github.com/tendermint/tendermint/abci/types"
)

func TestQGBIntegrationSuite(t *testing.T) {
	if testing.Short() {
		t.Skip("skipping qgb integration test in short mode.")
	}
	suite.Run(t, new(QGBIntegrationSuite))
}

type QGBIntegrationSuite struct {
	suite.Suite

	accounts []string
	cctx     testnode.Context
	ecfg     encoding.Config

	mut            sync.Mutex
	accountCounter int
}

func (s *QGBIntegrationSuite) SetupSuite() {
	t := s.T()
	t.Log("setting up integration test suite")

	accounts := make([]string, 10)
	for i := 0; i < len(accounts); i++ {
		accounts[i] = tmrand.Str(9)
	}

	cfg := testnode.DefaultConfig().WithAccounts(accounts)
	cctx, _, _ := testnode.NewNetwork(t, cfg)
	s.accounts = cfg.Accounts
	s.cctx = cctx
}

func (s *QGBIntegrationSuite) unusedAccount() string {
	s.mut.Lock()
	acc := s.accounts[s.accountCounter]
	s.accountCounter++
	s.mut.Unlock()
	return acc
}

func getAddress(account string, kr keyring.Keyring) sdk.AccAddress {
	rec, err := kr.Key(account)
	if err != nil {
		panic(err)
	}
	addr, err := rec.GetAddress()
	if err != nil {
		panic(err)
	}
	return addr
}

func (s *QGBIntegrationSuite) TestQGB() {
	t := s.T()
	type test struct {
		name                  string
		msgFunc               func() (msgs []sdk.Msg, signer string)
		hash                  string
		expectedCheckTxCode   uint32
		expectedDeliverTxCode uint32
	}
	tests := []test{
		{
			name: "edit a qgb validator address",
				account := "validator"
				valAcc, err := s.cctx.Keyring.Key(account)
				require.NoError(t, err)
				valAddr, err := valAcc.GetAddress()
				require.NoError(t, err)

				rvalAddr := sdk.ValAddress(valAddr)

				msg, err := qgbtypes.NewMsgRegisterEVMAddress(rvalAddr.String(), "0x95222290DD7278Aa3Ddd389Cc1E1d165CC4BAfe5")
				require.NoError(t, err)
				return []sdk.Msg{msg}, account
			},
			expectedCheckTxCode:   abci.CodeTypeOK,
			expectedDeliverTxCode: abci.CodeTypeOK,
		},
		{
			name: "unable to edit a qgb validator address",
				account := s.unusedAccount()
				valAcc, err := s.cctx.Keyring.Key("validator")
				require.NoError(t, err)
				valAddr, err := valAcc.GetAddress()
				require.NoError(t, err)

				rvalAddr := sdk.ValAddress(valAddr)

				msg, err := qgbtypes.NewMsgRegisterEVMAddress(rvalAddr.String(), "0x95222290DD7278Aa3Ddd389Cc1E1d165CC4BAfe5")
				require.NoError(t, err)
				return []sdk.Msg{msg}, account
			},
			expectedCheckTxCode:   abci.CodeTypeOK,
			expectedDeliverTxCode: errors.ErrInvalidPubKey.ABCICode(),
		},
	}

	// sign and submit the transactions
	for i, tt := range tests {
		msgs, signer := tt.msgFunc()
		res, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, signer, msgs...)
		require.NoError(t, err)
		require.NotNil(t, res)
		assert.Equal(t, abci.CodeTypeOK, res.Code, tt.name)
		tests[i].hash = res.TxHash
	}

	require.NoError(s.T(), s.cctx.WaitForNextBlock())

	for _, tt := range tests {
		res, err := testnode.QueryTx(s.cctx.Context, tt.hash, true)
		assert.NoError(t, err)
		require.NotNil(t, res)
		assert.Equal(t, tt.expectedDeliverTxCode, res.TxResult.Code, tt.name)
	}
}

```
feel free to copy paste or adjust if needed

x/qgb/client/tx.go
SweeXordious: that's the only thing I guess

x/qgb/types/errors.go
```diff
@@ -24,4 +24,5 @@ var (
 	ErrEarliestAvailableNonceStillNotInitialized = errors.Register(ModuleName, 33, "the earliest available nonce after pruning has still not been defined in store")
 	ErrRequestedNonceWasPruned                   = errors.Register(ModuleName, 34, "the requested nonce has been pruned")
 	ErrUnknownAttestationType                    = errors.Register(ModuleName, 35, "unknown attestation type")
+	ErrEVMAddressNotHex                          = errors.Register(ModuleName, 36, "the provided eth address is not a valid hex address")
```
SweeXordious, line 27:
```suggestion
	ErrEVMAddressNotHex                          = errors.Register(ModuleName, 36, "the provided evm address is not a valid hex address")
```

x/qgb/keeper/keeper_valset.go
```diff
@@ -193,3 +200,16 @@ func (k Keeper) GetLatestValsetBeforeNonce(ctx sdk.Context, nonce uint64) (*type
...
+
+func (k Keeper) SetEVMAddress(ctx sdk.Context, valAddress, evmAddress string) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetEVMKey(valAddress), []byte(evmAddress))
+}
+
+func (k Keeper) GetEVMAddress(ctx sdk.Context, valAddress string) (string, bool) {
+	store := ctx.KVStore(k.storeKey)
+	if !store.Has(types.GetEVMKey(valAddress)) {
+		return "", false
```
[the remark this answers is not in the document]
cmwaters: Not quite, they can still not have an evm address saved in state store. If it doesn't exist then we use the default which is just their validator address turned to an eth compatible HEX format

x/qgb/types/codec.go
```diff
@@ -3,12 +3,17 @@ package types
 import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 func RegisterCodec(_ *codec.LegacyAmino) {
 }
```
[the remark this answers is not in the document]
cmwaters: Let me see how it's done elsewhere and adapt it accordingly

x/qgb/types/msgs.go
cmwaters: Yeah adding an integration test here makes sense. Will try do this. As to whether we use a normal sdk address or a validator operator address for the mapping, I'm still not sure. Need to check how each gets used. It may be simpler to use the normal sdk address that's associated with the validator

proto/celestia/qgb/v1/tx.proto
```diff
@@ -0,0 +1,34 @@
...
+option go_package = "github.com/celestiaorg/celestia-app/x/qgb/types";
+
+// Msg is the message server for receiving qgb transactions
+service Msg {
+  // RegisterEVMAddress records an evm address for the validator which is used
+  // by the relayer to aggregate signatures. A validator can only register a
+  // single EVM address. The EVM address can be overridden by a later message.
+  // There are no validity checks of the EVM addresses existence on the Ethereum
+  // state machine
+  rpc RegisterEVMAddress(MsgRegisterEVMAddress)
```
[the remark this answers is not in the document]
cmwaters: Ideally we have a CLI that a new validator can do this all in a single transaction. I think registering is still a good wording for this. You can continually register new evm addresses and override the prior one

x/qgb/types/codec.go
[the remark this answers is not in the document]
cmwaters: Registered them in the same manner as the blob module

x/qgb/keeper/msg_server_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func TestRegisterEVMAddress(t *testing.T) {
+	input, sdkCtx := testutil.SetupFiveValChain(t)
+	k := input.QgbKeeper
+	vals := input.StakingKeeper.GetValidators(sdkCtx, 100)
+	require.GreaterOrEqual(t, len(vals), 1)
+	val := vals[0]
+	evmAddr, exists := k.GetEVMAddress(sdkCtx, val.OperatorAddress)
+	require.True(t, exists)
```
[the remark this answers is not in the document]
cmwaters: Yup good point

x/qgb/keeper/keeper_valset.go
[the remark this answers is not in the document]
SweeXordious: If so, then the implementation should change these parts: https://github.com/celestiaorg/celestia-app/blob/e2b154d8314872eba6f0d8d5ea18fdb58b201150/x/qgb/types/valset.go#L59-L87 and also parts of the `abci.go`, no?

evan-forbes (dismissed):
this makes sense to me!! there might be some additional testing we can do (as always), but I think we can do that in a follow up. Happy to help out with that too.

dope work @cmwaters, thanks for being so speedy and open to changing things.

as a side note I think this shipping process was really good. we ideated as a group well.

x/qgb/keeper/keeper_valset.go
```diff
@@ -95,11 +96,12 @@ func (k Keeper) GetCurrentValset(ctx sdk.Context) (types.Valset, error) {
...
-			// the protocol doesn't require that they register it, if they don't
-			// we take some unique string to derive their evm address from, namely
-			// the validators address
-			evmAddress = gethcommon.BytesToAddress(val).String()
+			// This should never happen and indicates a bug in the design of
+			// the system (most likely that a hook wasn't called). A validator
+			// should always have an associated EVM address. Fortunately we can
+			// safely recover from this by deriving the default again.
+			ctx.Logger().Error("validator does not have an evm address set")
+			evmAddress = types.DefaultEVMAddress(val).String()
```
SweeXordious, line 104: If we create a new EVM address here, this is not persisting it in store. Does it make sense to loop for correct default addresses in the hook?
Also, does it make sense to have a safer mechanism to derive the default EVM address? Something like hashing the operator address + something deterministic and specific to the current block?

x/qgb/keeper/hooks.go
```diff
@@ -38,7 +40,15 @@ func (h Hooks) BeforeDelegationCreated(_ sdk.Context, _ sdk.AccAddress, _ sdk.Va
...
 }
 
-func (h Hooks) AfterValidatorCreated(_ sdk.Context, _ sdk.ValAddress) error {
+func (h Hooks) AfterValidatorCreated(ctx sdk.Context, addr sdk.ValAddress) error {
+	defaultEvmAddr := types.DefaultEVMAddress(addr)
+	// This should practically never happen that we have a collision. It may be
+	// bad UX to reject the attempt to create a validator and require the user to
+	// generate a new set of keys but this ensures EVM address uniqueness
+	if !h.k.IsEVMAddressUnique(ctx, defaultEvmAddr.String()) {
+		return errors.Wrapf(types.ErrEVMAddressAlreadyExists, "create a validator with a different operator address to %s", addr.String())
```
SweeXordious, line 49: what would happen if a hook errors?

x/qgb/keeper/keeper_valset.go
cmwaters: > Also, does it make sense to have a safer mechanism to derive the default EVM address? Something like hashing the operator address + something deterministic and specific to the current block?

I assume the validator address is unique thus it's safe to derive the EVM address from? Is there a reason why you would like to add more variables? 

> If we create a new EVM address here, this is not persisting it in store. Does it make sense to loop for correct default addresses in the hook?

Hmm so in theory this should never happen but perhaps you're right that it's better that if there is a developer error that we still persist the new default EVM address

SweeXordious: > I assume the validator address is unique thus it's safe to derive the EVM address from? Is there a reason why you would like to add more variables?

the validator address is unique but as was discussed yesterday, some other validator can derive your validator EVM address and use it for themselves. Especially if the validator reuses keys

x/qgb/keeper/hooks.go
SweeXordious: cool, thanks :+1:

x/qgb/keeper/keeper_valset.go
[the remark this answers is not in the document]
SweeXordious: also pinging this conversation in case it was lost between the other ones

x/qgb/keeper/keeper_valset_test.go
```diff
@@ -170,3 +170,34 @@ func TestCheckingAttestationNonceInValsets(t *testing.T) {
 		})
 	}
 }
+
+func TestEVMAddresses(t *testing.T) {
```
SweeXordious, line 174: Does it make sense to also add a test where there is a duplicate EVM address and see the create validator fail?

cmwaters:
Some follow up issues:
- https://github.com/celestiaorg/celestia-app/issues/2294
- https://github.com/celestiaorg/celestia-app/issues/2293

x/qgb/keeper/keeper_valset_test.go
cmwaters: Yeah that shouldn't be too difficult to do. Let me give it a go

cmwaters: [add test for handling collision](https://github.com/celestiaorg/celestia-app/pull/2169/commits/fc5388b9dc1f3e7681d9eb5b876370b58b9002b4)

SweeXordious: :+1:

x/qgb/keeper/keeper_valset.go
```diff
@@ -194,3 +204,34 @@ func (k Keeper) GetLatestValsetBeforeNonce(ctx sdk.Context, nonce uint64) (*type
 		fmt.Sprintf("couldn't find valset before nonce %d", nonce),
 	)
 }
+
+func (k Keeper) SetEVMAddress(ctx sdk.Context, valAddress sdk.ValAddress, evmAddress gethcommon.Address) {
+	store := ctx.KVStore(k.storeKey)
+	// convert the address first from Hex (this may pad or prune the initial string)
```
SweeXordious, line 210: looks like a residual comment
```suggestion
```

SweeXordious (approved):
:tada: :tada: Awesome stuff :+1:  Thanks a lot for your patience with all the feedback

x/qgb/keeper/keeper_valset.go
```diff
@@ -94,9 +94,19 @@ func (k Keeper) GetCurrentValset(ctx sdk.Context) (types.Valset, error) {
...
+		evmAddress, exists := k.GetEVMAddress(ctx, val)
+		if !exists {
+			// This should never happen and indicates a bug in the design of
+			// the system (most likely that a hook wasn't called or a migration
+			// for existing validators wasn't conducted). A validator
+			// should always have an associated EVM address. Fortunately we can
+			// safely recover from this by deriving the default again.
+			ctx.Logger().Error("validator does not have an evm address set")
+			evmAddress = types.DefaultEVMAddress(val)
+			k.SetEVMAddress(ctx, val, evmAddress)
```
SweeXordious, line 106: [non-blocking][could be handled in a separate issue][personal opinion]
I guess at this level it's really important to panic if the EVM address doesn't exist because there is a serious issue with state.

The safety recovery system might be misleading in the case that a validator already changed his EVM address from the default one and then them having it being the default again. In that case, we're setting an invalid value in store.

I agree that if this happens the node will not be able to sync because of a different app hash, but it would be better to panic and just fail instead of putting an invalid value and an error log.

x/qgb/keeper/keeper_valset.go
SweeXordious: I guess it makes sense to open an issue to keep note that we can have a squatting issue and we need to investigate it when we have time

x/qgb/keeper/keeper_valset.go
cmwaters: > The safety recovery system might be misleading in the case that a validator already changed his EVM address from the default one and then them having it being the default again

If a validator has changed their EVM address then it must be in the store. EVM addresses can't be deleted. Once they're in state the validator always has an associated EVM address.

>  In that case, we're setting an invalid value in store.

How are we setting an invalid value?

> I guess at this level it's really important to panic

We can open this discussion (cc @evan-forbes, @staheri14 @rootulp). From my view, in this case we shouldn't actually panic. QGB is there to support an off-chain process. Any implementation bugs should be isolated and shouldn't result in taking down Celestia's primary service which is publishing blobs. The only problem with handling this error is the threat of an EVM address collision (someone set their EVM address to this default value). We can strengthen this by adding some deterministic randomness (like the blob time which is hard for a user to guess/manipulate). Or we can loop until we find a unique address (expensive but not computationally unbounded).

SweeXordious: Yes I agree, but this case is something that should never be hit. Hitting it means something is really wrong with the store and I think that if a node's store is changed in any way, it's a big red flag, and it can be a signal to also check if the other state is valid or not.

I consider it a general state error and not just QGB related.

SweeXordious (approved):

x/qgb/keeper/keeper_valset.go
SweeXordious: Makes sense :+1:, let's leave it for now then

x/qgb/keeper/keeper_valset.go
```diff
@@ -194,3 +204,33 @@ func (k Keeper) GetLatestValsetBeforeNonce(ctx sdk.Context, nonce uint64) (*type
...
+
+// IsEVMAddressUnique checks if the provided evm address is globally unique. This
+// includes the defaults we set validators when they initially create a validator
+// before registering
+func (k Keeper) IsEVMAddressUnique(ctx sdk.Context, evmAddress gethcommon.Address) bool {
+	store := ctx.KVStore(k.storeKey)
+	addrBytes := evmAddress.Bytes()
+	iterator := sdk.KVStorePrefixIterator(store, []byte(types.EVMAddress))
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
```
musalbas, line 230: Iterating over every EVM address won't be fraud proof friendly. Why does it need to be unique anyway?

evan-forbes (approved):

x/qgb/keeper/keeper_valset.go
cmwaters: My understanding is the system is not setup to handle duplicates. See https://github.com/celestiaorg/celestia-app/blob/e2b154d8314872eba6f0d8d5ea18fdb58b201150/x/qgb/types/validator.go#L181-L183

evan-forbes: we technically don't need it, but the currently implementation relies on maps heavily, so if each is not unique then we could end up overriding a previously added validator. Doing something different would require a dramatic re-write.

Adding this instance of iteration is bad, but we will also need to fix a lot of other instances as well before we can have uncompressed fraud proofs. Do you think its worth optimizing for that now?

cmwaters: Say there are two validators A and B with voting power 10 and 100. If they had the same EVM address, a signature from one of them would be impossible to distinguish if it was the one with 10 voting power or 100

musalbas: >Adding this instance of iteration is bad, but we will also need to fix a lot of other instances as well before we can have uncompressed fraud proofs. Do you think its worth optimizing for that now?

No, but just highlighting

Issue #2170, opened 2023-07-26
Rename GovSquareSizeUpperBound


rootulp:
LGTM.

Do we also want to rename it here https://github.com/celestiaorg/celestia-app/blob/3fc2fed84de759d3b82d80deff5df7f7ad7665b5/app/square_size.go#L10

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/pull/2150#pullrequestreview-1547059659_

rootulp:
We should audit all the variable and function names pertaining to square size and rename them to avoid confusing the two variables.

gaurav-11018:
@rootulp  i am working on this.
small clarification renaming should be done on only in this file :[celestia-app/app/square_size.go](https://github.com/celestiaorg/celestia-app/blob/3fc2fed84de759d3b82d80deff5df7f7ad7665b5/app/square_size.go#L10)

gaurav-11018:
after reading the docs i feel `GetGovernanceMaxSquareSize ` is more descriptive . what say @rootulp ?

rootulp:
IMO we should rename `GovSquareSizeUpperBound` => `GetMaxSquareSize`. It shouldn't use the word "gov" because this function doesn't strictly depend on the governance parameters. It shouldn't use the term `UpperBound` because the upper bound is one constant that influences the max square size.

rootulp:
@gaurav-11018 I'm thinking something like https://github.com/celestiaorg/celestia-app/pull/2173

rootulp:
Closing this as won't do. As confusing as it sounds, there are two parameters:
1. `GovMaxSquareSize`
2. `SquareSizeUpperBound`

And the current function name `GovSquareSizeUpperBound` (even though it is word soup) determines the max of those two parameters and returns it.

PR #2173, opened 2023-07-26
chore!: rename GovSquareSizeUpperBound

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2170

### DIFF
--- a/app/square_size.go
+++ b/app/square_size.go
@@ -5,9 +5,9 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// GovSquareSizeUpperBound returns the maximum square size that can be used for a block
-// using the governance parameter blob.GovMaxSquareSize.
-func (app *App) GovSquareSizeUpperBound(ctx sdk.Context) int {
+// GetMaxSquareSize returns the maximum square size that can be used for a
+// block.
+func (app *App) GetMaxSquareSize(ctx sdk.Context) int {
 	// TODO: fix hack that forces the max square size for the first height to
 	// 64. This is due to our fork of the sdk not initializing state before
 	// BeginBlock of the first block. This is remedied in versions of the sdk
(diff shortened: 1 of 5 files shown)


app/square_size.go
```diff
@@ -18,11 +18,15 @@ func (app *App) GovSquareSizeUpperBound(ctx sdk.Context) int {
...
+
+	return min(govMax, upperBound)
+}
 
-	return gmax
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
```
rootulp, line 31: [note for reviewers] can delete this in https://tip.golang.org/doc/go1.21

cmwaters (approved):

app/square_size.go
```diff
@@ -5,9 +5,9 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// GovSquareSizeUpperBound returns the maximum square size that can be used for a block
-// using the governance parameter blob.GovMaxSquareSize.
-func (app *App) GovSquareSizeUpperBound(ctx sdk.Context) int {
+// GetMaxSquareSize returns the maximum square size that can be used for a
+// block.
+func (app *App) GetMaxSquareSize(ctx sdk.Context) int {
```
evan-forbes, line 10: [discussion]
I forget what was the agreed upon scheme for naming. Per this PR, the term Max square size is now referring to current state of the largest square for a given height, which I believe differs from how this  was used in past (or at least how I've used this term in the past). 

That's not to say that this is a bad change! just a departure

evan-forbes: ref https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-021-restricted-block-size.md#decision

rootulp: Instead of trying to come up with a unique third name, I briefly explored what it would be like if `MaxSquareSize` actually acted as an upper bound for `GovMaxSquareSize` so that we could use `GovMaxSquareSize` throughout the app as the real max square size. 

See https://github.com/celestiaorg/celestia-app/issues/2185

It turns out we can't do that because the function signature of parameter validation functions 😑

Issue #2156, opened 2023-07-24
What happens to transactions that contain a blobs that are too large?


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/issues/2108

## Problem

Consensus nodes may not reject transactions that contain a blob that is too large to fit in a block. As a result, the transaction may sit in the mempool until it eventually expires (depends on mepool config).

## Proposal

Explore if the mempool will reject PFB transactions that contain a blob that will not fit in the maximum data square. If it doesn't, consider:

### Option A 

Add a stateless check to `ValidateBasic` on a transaction that verifies that the blob size is <= the maximum blob size.  Compute the maximum blob size based on the upper bound values for max square size and max block bytes.

### Option B
 
Decrease `max_tx_bytes` in the mempool config to the size of a tx that contains the largest blob that can fit in the largest data square.

### Option C

Add a stateful check via blob ante handler that verifies that the blob size is <= the maximum blob size. Compute the maximum blob size based on the governance parameters for max square size.

### Option D

Introduce a new versioned constant that acts as a max supported blob size.
- Pros: Since this is a versioned constant, a user can know prior to submitting their blob if it is valid. They don’t need to query the current chain state to determine if their constructed blob is valid.
- Cons: One additional parameter that needs to be re-evaluated when other square parameters are modified.

rootulp:
Since there isn't a versioned constant for the maximum supported blob size, it must be computed based on a number of other parameters (see https://github.com/celestiaorg/celestia-app/issues/2108). Since these parameters require access to the state, option A must be a stateful check.

rootulp:
I started working on validating that blobs are < max blob size in https://github.com/rootulp/celestia-app/tree/rp/improve-too-big-blob which is based on top of #2187

evan-forbes:
Option A makes a lot of sense, since this will get hit before even submitting a blob. Do we also have an option C (no mutually exclusive to A!), where we add an ante handler that checks the the size of the blob using the stateful parameters? This will get hit upon submitting a blobtx and when we are gossiping blobs

PR #2187, opened 2023-07-30
test: blob sizes that are too large

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/2156

### DIFF
--- a/app/test/check_tx_test.go
+++ b/app/test/check_tx_test.go
@@ -24,7 +24,7 @@ func TestCheckTx(t *testing.T) {
 	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
 	ns1 := appns.MustNewV0(bytes.Repeat([]byte{1}, appns.NamespaceVersionZeroIDSize))
 
-	accs := []string{"a", "b", "c", "d", "e", "f"}
+	accs := []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j"}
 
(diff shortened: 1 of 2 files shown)


app/test/check_tx_test.go
```diff
@@ -107,6 +107,54 @@ func TestCheckTx(t *testing.T) {
...
+			checkType: abci.CheckTxType_New,
+			getTx: func() []byte {
+				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 10_000_000, 1, false, testutil.ChainID, accs[8:9])[0]
+				return tx
+			},
+			// TODO: consider modifying CheckTx to return an error for this case
+			// so that consensus nodes do not propagate blobs that are too
+			// large.
+			expectedABCICode: abci.CodeTypeOK,
+		},
```
cmwaters, line 157: I think Tendermint catches this case i.e. an inbound transaction won't even reach checkTx because it's too large

cmwaters (approved):
It might be better to use a constant and test around that constant

rootulp:
> It might be better to use a constant and test around that constant

I thought about introducing a constant for the max blob size b/c currently one isn't enforced but I think that requires further discussion.

Since the largest supported blob size isn't a constant now, it has to be calculated based on a few other parameters (max square size, max block bytes) which requires access to state.

evan-forbes (approved):

PR #2190, opened 2023-07-31
fix: Use commit sha when there is no RC or tag available

jrmanes:
hello team!

This is a small PR, it contains a change to print the commit sha when we are not running an RC or a version, This might be useful when we want to run specific commits, branches, etc... *Otherwise the value will be empty.*

This comes from after testing the new metric `version`, I can see this when we are not running tags:
```
cometbft_consensus_step_duration_seconds_bucket{chain_id="<whatever>"",step="Commit",version="",le="0.2682695795279726"} 153
```

Please team, if you approve it, merge it after, I don't have permission to do it.

Thanks in advance! 🚀 

Jose Ramon Mañes


Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION := $(shell echo $(shell git describe --tags) | sed 's/^v//')
+VERSION := $(shell echo $(shell git describe --tags 2>/dev/null || git log -1 --format='%h') | sed 's/^v//')
```
jrmanes, line 1: I've added `2>/dev/null` bc, in case there's no tag, git prints something like:
```
fatal: No names found, cannot describe anything.
```
and we don't need this message

jrmanes: also mention:
`--format='%h'` -> short sha
`--format='%H'` -> long sha

evan-forbes (approved):

cmwaters (approved):
utACK 👍 

Don't forget to backport this to the v1.x branch

smuu (approved):
:+1:

PR #2202, opened 2023-08-01
feat: reject blobs that are too large

rootulp:
Implements Option A of https://github.com/celestiaorg/celestia-app/issues/2156

### DIFF
--- a/app/test/check_tx_test.go
+++ b/app/test/check_tx_test.go
@@ -4,8 +4,6 @@ import (
 	"bytes"
 	"testing"
 
-	tmrand "github.com/tendermint/tendermint/libs/rand"
-
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
@@ -15,6 +13,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
@@ -111,8 +110,7 @@ func TestCheckTx(t *testing.T) {
 			name:      "1,000 byte blob",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 1_000, 1, false, testutil.ChainID, accs[4:5])[0]
-				return tx
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[4], 1_000)
 			},
 			expectedABCICode: abci.CodeTypeOK,
 		},
@@ -120,8 +118,7 @@ func TestCheckTx(t *testing.T) {
 			name:      "10,000 byte blob",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 10_000, 1, false, testutil.ChainID, accs[5:6])[0]
-				return tx
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[5], 10_000)
 			},
 			expectedABCICode: abci.CodeTypeOK,
 		},
@@ -129,8 +126,7 @@ func TestCheckTx(t *testing.T) {
 			name:      "100,000 byte blob",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 100_000, 1, false, testutil.ChainID, accs[6:7])[0]
-				return tx
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[6], 100_000)
 			},
 			expectedABCICode: abci.CodeTypeOK,
 		},
@@ -138,23 +134,22 @@ func TestCheckTx(t *testing.T) {
 			name:      "1,000,000 byte blob",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 1_000_000, 1, false, testutil.ChainID, accs[7:8])[0]
-				return tx
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[7], 1_000_000)
 			},
 			expectedABCICode: abci.CodeTypeOK,
 		},
-		{
-			name:      "10,000,000 byte blob",
-			checkType: abci.CheckTxType_New,
-			getTx: func() []byte {
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 10_000_000, 1, false, testutil.ChainID, accs[8:9])[0]
-				return tx
-			},
-			// TODO: consider modifying CheckTx to return an error for this case
-			// so that consensus nodes do not propagate blobs that are too
-			// large.
-			expectedABCICode: abci.CodeTypeOK,
-		},
+		// {
+		// 	name:      "10,000,000 byte blob",
+		// 	checkType: abci.CheckTxType_New,
+		// 	getTx: func() []byte {
+		// 		tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 10_000_000, 1, false, testutil.ChainID, accs[8:9])[0]
+		// 		return tx
+		// 	},
+		// 	// TODO: consider modifying CheckTx to return an error for this case
+		// 	// so that consensus nodes do not propagate blobs that are too
+		// 	// large.
+		// 	expectedABCICode: abci.CodeTypeOK,
+		// },
 	}
 
 	for _, tt := range tests {
--- a/app/test/check_tx_test.go
+++ b/app/test/check_tx_test.go
@@ -138,18 +138,14 @@ func TestCheckTx(t *testing.T) {
 			},
 			expectedABCICode: abci.CodeTypeOK,
 		},
-		// {
-		// 	name:      "10,000,000 byte blob",
-		// 	checkType: abci.CheckTxType_New,
-		// 	getTx: func() []byte {
-		// 		tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 10_000_000, 1, false, testutil.ChainID, accs[8:9])[0]
-		// 		return tx
-		// 	},
-		// 	// TODO: consider modifying CheckTx to return an error for this case
-		// 	// so that consensus nodes do not propagate blobs that are too
-		// 	// large.
-		// 	expectedABCICode: abci.CodeTypeOK,
-		// },
+		{
+			name:      "10,000,000 byte blob",
+			checkType: abci.CheckTxType_New,
+			getTx: func() []byte {
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[8], 10_000_000)
+			},
+			expectedABCICode: blobtypes.ErrBlobSizeTooLarge.ABCICode(),
+		},
 	}
 
 	for _, tt := range tests {
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -14,7 +14,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/cosmos/cosmos-sdk/client"
-	"github.com/cosmos/cosmos-sdk/types/errors"
 
 	"github.com/stretchr/testify/suite"
 
@@ -388,6 +387,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob_blobSizes() {
 		blob *tmproto.Blob
 		// txResponseCode is the expected tx response ABCI code.
 		txResponseCode uint32
+		wantError      error
 	}
 	testCases := []testCase{
 		{
@@ -411,9 +411,9 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob_blobSizes() {
 			txResponseCode: abci.CodeTypeOK,
 		},
 		{
-			name:           "10,000,000 byte blob returns err tx too large",
-			blob:           mustNewBlob(t, 10_000_000),
-			txResponseCode: errors.ErrTxTooLarge.ABCICode(),
+			name:      "10,000,000 byte blob returns error blob size too large",
+			blob:      mustNewBlob(t, 10_000_000),
+			wantError: blobtypes.ErrBlobSizeTooLarge,
 		},
 	}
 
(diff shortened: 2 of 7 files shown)


app/test/check_tx_test.go
```diff
@@ -111,49 +110,41 @@ func TestCheckTx(t *testing.T) {
...
 		},
 		{
 			name:      "1,000,000 byte blob",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 1_000_000, 1, false, testutil.ChainID, accs[7:8])[0]
-				return tx
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[7], 1_000_000)
 			},
 			expectedABCICode: abci.CodeTypeOK,
```
evan-forbes, line 139: [comment] 

how are these tests passing without account numbers being added? shouldn't the signature be failing? also it appears we have two `RandBlobTxsWithAccounts`, one in `util` that uses the app to query the account number and sequence directly from the testapp, and then another in blobfactor where we don't...

test/util/blobfactory/payforblob_factory.go
```diff
@@ -260,6 +260,51 @@ func RandBlobTxs(enc sdk.TxEncoder, rand *tmrand.Rand, count, blobsPerTx, size i
...
+	address, err := signer.GetSignerInfo().GetAddress()
+	require.NoError(t, err)
+
+	coin := sdk.Coin{
+		Denom:  bondDenom,
+		Amount: sdk.NewInt(10),
+	}
+	options := []blobtypes.TxBuilderOption{
+		blobtypes.SetFeeAmount(sdk.NewCoins(coin)),
+		blobtypes.SetGasLimit(10_000_000),
```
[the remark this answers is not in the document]
rootulp: f6e3b8808c55a01983cf9927f9054ed53ee4346f

x/blob/types/payforblob.go
```diff
@@ -331,6 +332,51 @@ func ValidateBlobs(blobs ...*Blob) error {
...
+// parameter in application state (i.e. GovMaxSquareSize). Additionally, even if
+// the application state parameters are at their upper bounds, the number of
+// shares available for blob bytes may be less than estimated in this function
+// (i.e. if the PFB tx shares occupy more than one share). As a result, the
+// upper bound returned by this function may over-estimate the max valid blob
+// size but it should not under-estimate. Consequently, it may be used to
+// immediately reject blobs that are too large but blobs smaller than this upper
+// bound may still fail to be included in a block.
+func blobSizeUpperBound() int {
+	maxSquareSize := appconsts.SquareSizeUpperBound(appconsts.LatestVersion)
```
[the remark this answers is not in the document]
rootulp: We should pass it but at some point in the calling stack we have to use `appconsts.LatestVersion` because these validation functions are stateless and don't have access to a block's app version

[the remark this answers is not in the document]
rootulp: https://github.com/celestiaorg/celestia-app/pull/2202/commits/e6248f0961721ef8ec51b073865aef561be0ee65

app/test/check_tx_test.go
rootulp: I'm not sure what you mean by account numbers being added? This test uses `accs[7]`, other test cases use unique accounts.

rootulp: Refactored to use `RandBlobTxsWithAccounts` in all test cases except the last one b/c the last one would fail to create a blob tx for a 10,000,000 byte blob if it used `RandBlobTxsWithAccounts`.

 https://github.com/celestiaorg/celestia-app/pull/2202/commits/d8faa9760ca88b3ca9c928c9bc385bdf5ff04c91

app/test/check_tx_test.go
```diff
@@ -111,49 +110,41 @@ func TestCheckTx(t *testing.T) {
...
-				tx := blobfactory.RandBlobTxsWithAccounts(encCfg.TxConfig.TxEncoder(), tmrand.NewRand(), kr, nil, 10_000_000, 1, false, testutil.ChainID, accs[8:9])[0]
-				return tx
+				return blobfactory.BlobTxWithSize(t, encCfg.TxConfig.TxEncoder(), kr, testutil.ChainID, accs[8], 10_000_000)
 			},
-			// TODO: consider modifying CheckTx to return an error for this case
-			// so that consensus nodes do not propagate blobs that are too
-			// large.
-			expectedABCICode: abci.CodeTypeOK,
+			expectedABCICode: blobtypes.ErrBlobSizeTooLarge.ABCICode(),
 		},
```
rootulp, line 148: [optional] we can also remove this test case all together if maintainer's think the overhead of maintaining `blobfactory.BlobTxWithSize` is too high.

x/blob/types/payforblob.go
```diff
@@ -331,6 +332,51 @@ func ValidateBlobs(blobs ...*Blob) error {
...
 
+// validateBlobData returns an error if the size of data is zero or too large.
+func validateBlobData(appVersion uint64, data []byte) error {
+	if len(data) == 0 {
+		return ErrZeroBlobSize
+	}
+
+	maxBlobSize := blobSizeUpperBound(appVersion)
+	if len(data) > maxBlobSize {
+		return ErrBlobSizeTooLarge.Wrapf("max blob size is %d bytes", maxBlobSize)
```
rootulp, line 343: 🤔 the wrapped error should probably include the blob size for debugging purposes

x/blob/types/payforblob.go
```diff
@@ -331,6 +332,51 @@ func ValidateBlobs(blobs ...*Blob) error {
...
+	maxSquareSize := appconsts.SquareSizeUpperBound(appVersion)
+	maxShares := maxSquareSize * maxSquareSize
+
+	// Subtract one from maxShares because at least one share must be occupied
+	// by the PFB tx associated with this blob.
+	maxBlobShares := maxShares - 1
+	maxBlobBytes := maxBlobShares * appconsts.ContinuationSparseShareContentSize
+
+	return min(maxBlobBytes, coretypes.MaxBlockSizeBytes)
+}
```
rootulp, line 369: I think we should simplify this upper bound to compute the max number of bytes in a data square based on app version and reject based on that.

app/test/check_tx_test.go
evan-forbes: similar to the sequence or chain id, the account number of a given account is added to the tx sign bytes when we sign the tx but its not actually included in the transaction, therefore we have to look it up before we sign a tx.

since we're not looking up sequence or account number (passing nil for the conn arg), I would have expected the signature of the tx to fail...

the one in util can query directly using the testapp (doesn't use grpc). I'll take another look at this test and see if there's something wrong

x/blob/types/payforblob.go
```diff
@@ -49,7 +49,7 @@ const (
 var _ legacytx.LegacyMsg = &MsgPayForBlobs{}
 
 func NewMsgPayForBlobs(signer string, blobs ...*Blob) (*MsgPayForBlobs, error) {
-	err := ValidateBlobs(blobs...)
+	err := ValidateBlobs(appconsts.LatestVersion, blobs...)
```
rootulp, line 52: On second thought I think this may be dangerous for similar reasons mentioned in https://github.com/celestiaorg/celestia-app/pull/2203#discussion_r1283420399

This assumes the latest celestia-app version so the behavior varies based on when a node upgrades.

rootulp:
https://github.com/celestiaorg/celestia-app/actions/runs/5766459263/job/15634450248#step:4:246

ante handlers get run on `CheckTx`?! Were options A and C the same the whole time? 🤯

x/blob/ante/max_blob_size_ante.go
```diff
@@ -0,0 +1,93 @@
...
+func (d MaxBlobSizeDecorator) getMaxSquareSize(ctx sdk.Context) int {
+	// TODO: fix hack that forces the max square size for the first height to
+	// 64. This is due to our fork of the sdk not initializing state before
+	// BeginBlock of the first block. This is remedied in versions of the sdk
+	// and comet that have full support of PreparePropsoal, although
+	// celestia-app does not currently use those. see this PR for more details
+	// https://github.com/cosmos/cosmos-sdk/pull/14505
+	if ctx.BlockHeader().Height <= 1 {
+		return int(appconsts.DefaultGovMaxSquareSize)
+	}
```
cmwaters, line 65: This gets calculated in `d.k.GovMaxSquareSize` below so I don't think it's necessary

cmwaters (approved):
LGTM

x/blob/ante/max_blob_size_ante.go
```diff
@@ -0,0 +1,93 @@
...
+func NewMaxBlobSizeDecorator(k BlobKeeper) MaxBlobSizeDecorator {
+	return MaxBlobSizeDecorator{k}
+}
+
+// AnteHandle implements the AnteHandler interface. It returns an error if the
+// tx contains a MsgPayForBlobs where the total blob data size exceeds the upper
+// bound. The upper bound is calculated based on the number of bytes in a data
+// square with the maximum square size.
+func (d MaxBlobSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if ctx.IsReCheckTx() {
```
evan-forbes, line 26: [discussion]
imo we shouldn't run this during deliver tx, since 1) this would technically be adding block validity rule since it runs on PFBs 2) the only way a PFB can make it in the block is if the blob can fit. If a PFB gets included, but the blob doesn't, then we have to rely on block validity rules (via constructing the square) since by DeliverTx the PFB will have already been included and a blob inclusion fraud proof will be created.

this is mostly an efficiency thing, but I don't see any benefit atm, wdyt?
```suggestion
	if !ctx.IsCheckTx() {
```

evan-forbes (commented):
one comment to be discussed, but otherwise LGTM :+1:

x/blob/ante/max_blob_size_ante.go
evan-forbes: we could also be more specific here, like `if ctx.IsReCheckTx || if ctx.IsDeliverTx` if we don't like `if !ctx.CheckTx`

cmwaters (commented):
So I've been thinking about this and I think it's better if we have an AnteHandler that is overly restrictive than under restrictive. What I mean by that is that it's probably better if we reject a transaction that would've made it in a block than it is to accept a transaction that floats in the mempool but never makes its way to a block and the user is baffled as to why his/her transaction isn't being committed. Hence, I would use `ContinuationSparseShare` instead of `ShareSize`. I actually recall that we had a function that would more accurately calculate the amount of bytes in the `share` package. I think the more accurate we can be the better

x/blob/ante/max_blob_size_ante.go
rootulp: Hmm, I agree it doesn't seem necessary to run this ante handler on `DeliverTx()` b/c that is adding a block validity rule. However, I wonder if it makes sense to run this ante handler on `ReCheckTx()`. Consider the following scenario:

Assume that gov max square size is 128 at t=0.
t=0: User submits txA that is a large PFB. txA passes ante handler and gets gossiped to network. txA is valid but doesn't get included in a block yet.
t=1: New block lands that decreases gov max square size from 128 to 64
t=2: txA remains in mempool. Validators can't include it b/c it can only fit in a square size of 128.

I assume that txA will eventually be evicted from the mempool. The situation is a bit contrived and I think it probably makes more sense to only run this ante handler once on `IsCheckTx()`.


Note: our version of ctx doesn't have `ctx.IsDeliverTx()` 

Screenshot 2023-08-07 at 2 35 19 PM

so even though I prefer the more explicit:

```go
	if ctx.IsReCheckTx() || ctx.IsDeliverTx()
```

I'll go with

```go
	if !ctx.IsCheckTx() {
```

rootulp: Resolved by https://github.com/celestiaorg/celestia-app/pull/2202/commits/f8914cb53907c0023374be2cf050ec14b6b2d7e5

rootulp:
> What I mean by that is that it's probably better if we reject a transaction that would've made it in a block than it is to accept a transaction that floats in the mempool but never makes its way to a block and the user is baffled as to why his/her transaction isn't being committed.

Hmm given there isn't an explicit max total blob size constant, this ante handler has to estimate it. The current implementation over-estimates because if the ante handler rejects a transaction that could have fit in a block then we're adding an unspecified limitation on total blob size. In other words, if the ante handler under-estimates the max total blob size, the ante handler is effectively adding an explicit max total blob size constant on top of the constants already defined by ADR-021.

To your point, this ante handler can come closer to estimating the true total max blob size by using `ContinuationSparseShare`, but I think it should still over-estimate the max total blob size rather than under-estimate it.

app/test/integration_test.go
```diff
@@ -58,12 +58,14 @@ func (s *IntegrationTestSuite) SetupSuite() {
 		s.accounts[i] = tmrand.Str(20)
 	}
 
-	cfg := testnode.DefaultConfig().WithAccounts(s.accounts)
+	s.ecfg = encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	cfg := testnode.DefaultConfig().
+		WithAccounts(s.accounts).
+		WithGenesisOptions(testnode.SetBlobParams(s.ecfg.Codec, blobtypes.DefaultGenesis().Params))
```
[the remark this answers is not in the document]
rootulp: At one point while working on this, I thought they weren't being provided but tests pass without this change so will revert.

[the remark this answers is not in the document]
rootulp: Resolved by 467f57610e44c227bdeeea3cb2dbbf089dca8c44

app/test/check_tx_test.go
rootulp: Resolving this b/c no longer applicable.

cmwaters:
>  the ante handler is effectively adding an explicit max total blob size constant on top of the constants already defined by ADR-021.

I think it's fine that to add that extra constraint. For one, it's not actually a block validity rule, it serves to improve the UX around transaction submission, 2) they would be two different constants: one for a single PFB/tx the other, the total size of the square/block.

evan-forbes:
> To your point, this ante handler can come closer to estimating the true total max blob size by using ContinuationSparseShare, but I think it should still over-estimate the max total blob size rather than under-estimate it.

In this specific case, I think it makes sense to be as accurate as possible. The main reason being that we should never gossip a tx if it can never make it in a block. This is similar reasoning for having a minimum fee. Previously when we had 8MB blocks, we've arbitrarily set the value to 1 or 2MB per the thread https://github.com/celestiaorg/celestia-core/issues/867

rootulp:
Based on the above feedback, I'll add an explicit constant for max total blob size of a transaction and enforce it in this PR. Note this behavior change should be accompanied by a change to the specs.

rootulp:
>  I'll add an explicit constant for max total blob size of a transaction and enforce it in this PR.

This isn't feasible if we still want the max total blob to be based on the governance param for max square size because the param is modifiable and isn't a versioned constant.

rootulp:
TBH I'm not actually sure what the actionable feedback is. @cmwaters @evan-forbes 

> I think it makes sense to be as accurate as possible

I don't see a way to be more accurate than https://github.com/celestiaorg/celestia-app/pull/2202/files#diff-9baefd27a9e6df186608a7ef8c81d7aa14214357e72797c4c432510b3af9e1fdR50-R55

cmwaters:
I'm personally happy with what you've done here: https://github.com/celestiaorg/celestia-app/pull/2202/files#diff-9baefd27a9e6df186608a7ef8c81d7aa14214357e72797c4c432510b3af9e1fdR50-R55. I think it's as accurate as we can get

rootulp:
I have low confidence that this will backport to the v1.x branch cleanly but at least mergify can open a PR that I can fix.

cmwaters (approved):
Nice work 👍

evan-forbes (approved):
nice!

x/blob/ante/max_total_blob_size_ante.go
```diff
@@ -0,0 +1,88 @@
...
+	// celestia-app does not currently use those. see this PR for more details
+	// https://github.com/cosmos/cosmos-sdk/pull/14505
+	if ctx.BlockHeader().Height <= 1 {
+		return int(appconsts.DefaultGovMaxSquareSize)
+	}
+
+	upperBound := appconsts.SquareSizeUpperBound(ctx.ConsensusParams().Version.AppVersion)
+	govParam := d.k.GovMaxSquareSize(ctx)
+	return min(upperBound, int(govParam))
+}
```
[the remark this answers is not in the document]
rootulp: Makes sense. Created https://github.com/celestiaorg/celestia-app/issues/2267

Issue #2185, opened 2023-07-28
Enforce that `GovMaxSquareSize` <= `MaxSquareSize`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/2de3ed3f1c3bd636e917017c77374036d9c26b97/docs/architecture/adr-021-restricted-block-size.md?plain=1#L83

## Problem

I don't see where this is enforced:

>  Must be smaller than the `MaxSquareSize`.

which means the app needs to determine the limiting max square size via https://github.com/celestiaorg/celestia-app/blob/b3b8611c299a113a7283df2e986b5d7f2498c9e5/app/square_size.go#L8-L28

This means we have a third variable name for max square size which causes confusion (see https://github.com/celestiaorg/celestia-app/pull/2173). Additionally other modules (i.e. `blob`) doesn't have access to app so it can't readily import this function. That means it needs to redefine similar logic to derive the limiting factor for max square size.

## Proposal

Is it possible to enforce that the governance modifiable param is <= the upper bound constant?

1. In [`validateGovMaxSquareSize`](https://github.com/celestiaorg/celestia-app/blob/53e7b57a1a00c952ced0c590a4533b5d116a52f7/x/blob/types/params.go#L76-L77) add a check that ensures the gov max square size is <= the upper bound constant
2. Remove extra logic in [`GovSquareSizeUpperBound`](https://github.com/celestiaorg/celestia-app/blob/53e7b57a1a00c952ced0c590a4533b5d116a52f7/app/square_size.go#L8-L28)

rootulp:
Attempted to do this but it doesn't seem feasible b/c the validate param function needs to conform to a function signature from cosmos-sdk.  See the error [here](https://github.com/rootulp/celestia-app/commit/e375febd3efb881173fb0f1a50cfeabf8126c303#diff-32677395b5a229a23fb69148dd3bbe27316275ecb291928e09f162c38b2e0e67R44).

evan-forbes:
per a sync update, it is worth noting that we likely need to pass the app version to all param changes in the future so that we can determistically change restrictions over parameters.

rootulp:
Closing as won't do b/c we can't easily constraint the bounds of a gov param. That's why we added the kinda confusing logic to determine the `MaxEffectiveSquareSize`

PR #2203, opened 2023-08-01
feat: enforce that `GovMaxSquareSize` <= `MaxSquareSize`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2185

### DIFF
--- a/x/blob/types/params.go
+++ b/x/blob/types/params.go
@@ -41,7 +41,6 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 	return paramtypes.ParamSetPairs{
 		paramtypes.NewParamSetPair(KeyGasPerBlobByte, &p.GasPerBlobByte, validateGasPerBlobByte),
 		paramtypes.NewParamSetPair(KeyGovMaxSquareSize, &p.GovMaxSquareSize, validateGovMaxSquareSize),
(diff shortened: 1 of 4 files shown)


x/blob/types/params.go
```diff
@@ -90,6 +90,10 @@ func validateGovMaxSquareSize(v interface{}) error {
 			govMaxSquareSize,
 		)
 	}
+	upperBound := uint64(appconsts.SquareSizeUpperBound(appconsts.LatestVersion))
```
rootulp, line 93: @cmwaters is it concerning that this line always uses `appconsts.LatestVersion`? 

I think this means that when celestia-app cuts a new major version, the validation of this governance param _may_ change which seems like breaking behavior but that's fine because it will only happen in a breaking release.

What seems fishy is that this function now relies on a global const rather than plumbing app version for a particular block into here.

evan-forbes: [blocking]

I believe this is the reason why we are not doing this in the first place. We have to route the app version only when it changes if we want single binary syncs. Otherwise the node will change this value whenever they upgrade which will result in non-determinism

evan-forbes (changes requested):

x/blob/types/params.go
rootulp: Good point seems very sketchy. I still think it would be valuable to enforce that the gov max square size is always less than the upper bound constant but I don't think the approach in this PR is safe b/c the reason you mentioned.

rootulp: It is difficult to reject blobs that are too large considering (https://github.com/celestiaorg/celestia-app/pull/2202) because it isn't obvious how to fetch the max square size.

Issue #1964, opened 2023-06-19
test: skip data race tests without using `-short` flag


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/issues/1369 and https://github.com/celestiaorg/celestia-app/pull/1958#pullrequestreview-1486870364 

## Problem

The `-short` flag is used on this line: https://github.com/rootulp/celestia-app/blob/aa01da8c6d4ab46c5cdafd50ccc95749aa2a01c6/Makefile#L111

## Proposal

1. Remove the `-short` flag from the line above because the `-short` flag mechanism should be used to skip long running tests, not tests that have data races.
2. [short term] use a different mechanism to skip tests that have identified data races
3. [long term] fix https://github.com/celestiaorg/celestia-app/issues/1369 by root causing all data races

rootulp:
We may consider the `-skip` flag from `go help testflag`:

```
	-skip regexp
	    Run only those tests, examples, fuzz tests, and benchmarks that
	    do not match the regular expression. Like for -run and -bench,
	    for tests and benchmarks, the regular expression is split by unbracketed
	    slash (/) characters into a sequence of regular expressions, and each
	    part of a test's identifier must match the corresponding element in
	    the sequence, if any.
```

evan-forbes:
since we don't have direct access to build flags in runtime, in the past we've also considered using build flag switches (add an init to a module that will flip a bool to false that is protected by a build flag)

PR #2204, opened 2023-08-01
chore: skip data race tests with `-skip` flag

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/1964 via the **short term** approach. 

[Go 1.20](https://tip.golang.org/doc/go1.20#go-command) introduced the `-skip` flag and this PR skips all the tests that were previously skipped via `-short` when `make test-race` is run. It is possible that some of these tests don't contain data races in which case they can be removed from the argument to `-skip` in future PRs.


.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
-      - name: Run unit tests
+      - name: Run tests
```
rootulp, line 18: `make test` runs unit and integration tests

Makefile
```diff
@@ -153,10 +153,12 @@ test-e2e:
...
 .PHONY: test-e2e
 
-## test-race: Run unit tests in race mode.
+## test-race: Run tests in race mode.
 test-race:
+# TODO: Remove the -skip flag once the following tests no longer contain data races.
+# https://github.com/celestiaorg/celestia-app/issues/1369
 	@echo "--> Running tests in race mode"
-	@VERSION=$(VERSION) go test -mod=readonly -race -short ./...
+	@go test -mod=readonly ./... -v -race -skip "TestPrepareProposalConsistency|TestIntegrationTestSuite|TestQGBRPCQueries|TestSquareSizeIntegrationTest|TestStandardSDKIntegrationTestSuite|TestTxsimCommandFlags|TestTxsimCommandEnvVar|TestMintIntegrationTestSuite|TestQGBCLI|TestUpgrade"
```
evan-forbes, line 161: what do we think of doing something like this? 
https://github.com/celestiaorg/celestia-app/issues/1964#issuecomment-1597774436

we have a package that has an init func which modifies some global bool in a different package. then we use that bool in the tests similar to or as a replacement for t.Short()

cmwaters (approved):
Sounds good to me

Makefile
evan-forbes: in this case, there would only be a single init and build flag

if we wanted to run with the race detector, then we add those build flags, which causes a single init function to change a global var

the advantage here would be that we would be able to define things in an identical way to what we do now with t.Short() in the go test itself instead of having two different apis. 

The idea of just appending something to a list feels weird, but I won't block on it

evan-forbes (approved):

Makefile
rootulp: > there would only be a single init and build flag

Wouldn't we have to add a build flag to each test file?

> in an identical way to what we do now with t.Short() in the go test itself

Will your approach lead to something like:

```diff
func TestPrepareProposalConsistency(t *testing.T) {
	if testing.Short() {
		t.Skip("skipping TestPrepareProposalConsistency in short mode.")
	}
+	if testing.SkipRace() {
+		t.Skip("skipping TestPrepareProposalConsistency in race mode.")
+	}
```

If that's the case, it seems fine. TBH I'm not sure what you mean by single init and build flag so I'm going to merge this PR as is. Definitely open to reverting the `-skip` flag usage if there's a clearer way to do this.

PR #2207, opened 2023-08-02
chore: copy across cat pool spec from celestia-core

cmwaters:
## Overview

Given that it is only the specification of celestia-app that is published it makes more sense to include it there rather than in the directory of the cat pool itself

## Checklist

### DIFF
--- a/specs/src/SUMMARY.md
+++ b/specs/src/SUMMARY.md
@@ -7,6 +7,7 @@
   - [Namespace](./specs/namespace.md)
   - [Shares](./specs/shares.md)
   - [Consensus](./specs/consensus.md)
+    - [CAT Pool](./specs/cat_pool.md)
   - [Block Proposer](./specs/block_proposer.md)
   - [Block Validity Rules](./specs/block_validity_rules.md)
   - [Fraud Proofs](./specs/fraud_proofs.md)
--- a/specs/src/README.md
+++ b/specs/src/README.md
@@ -5,6 +5,7 @@
   - [Namespace](./specs/namespace.md)
   - [Shares](./specs/shares.md)
   - [Consensus](./specs/consensus.md)
+    - [CAT Pool](./specs/cat_pool.md)
   - [Block Proposer](./specs/block_proposer.md)
   - [Block Validity Rules](./specs/block_validity_rules.md)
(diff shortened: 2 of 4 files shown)


specs/src/SUMMARY.md
```diff
@@ -7,6 +7,7 @@
   - [Namespace](./specs/namespace.md)
   - [Shares](./specs/shares.md)
   - [Consensus](./specs/consensus.md)
+    - [CAT Pool](./specs/cat_pool.md)
```
rootulp, line 10: I think we may want to add this to the README here: https://github.com/celestiaorg/celestia-app/blob/492c3db844119da836295ca951d4548dec08133b/specs/src/README.md?plain=1#L7

It's a bit ridiculous to maintain 3 locations for the same table of contents so we may consider deleting that README alltogether.

rootulp (approved):
LGTM. In the past there was a push for specs to live in the same repo as the code they are relevant to. If we still want that, we can alternatively consider publishing celestia-core specs to their own website so that they don't need to be copied to celestia-app. I'm curious what @adlerjohn thinks.

evan-forbes (dismissed):

evan-forbes:
> In the past there was a push for specs to live in the same repo

alternatively we could simply link to the catpool spec in core. all that matters imo is that we have a single link to start with

cmwaters:
I tend to agree that they should live in their own repo but its a bit strange when it comes to forks. I would also prefer that from the outside Celestia is presented as a single cohesive spec

specs/src/SUMMARY.md
cmwaters: Yeah I added it to the `SUMMARY.md` which I think is the main one in terms of the rendered static site

staheri14 (approved):
Looks good! I'd recommend removing the original spec from the cat pool directory and adding a link to this one to avoid having multiple copies.

evan-forbes (approved):

cmwaters:
> it may as well make sense to have a separate website dedicated to the core related specs

Ideally I'd like to have a single site that contains all the specification of the celestia network (including `celestia-node` as well). For now, to make it more visible, I think it's better here and once we develop a better system we can remove it and just have a spec directory in core

PR #2208, opened 2023-08-03
fix: end exclusive parse namespace end share range

SweeXordious:
## Overview

Fixes the range to support end exclusive share ranges 

## Checklist

### DIFF
--- a/pkg/proof/proof_test.go
+++ b/pkg/proof/proof_test.go
@@ -225,12 +225,11 @@ func TestNewShareInclusionProof(t *testing.T) {
 // TestAllSharesInclusionProof creates proof for all the shares in a block.
 // Since we can't prove multiple namespaces at the moment, the block will contain all the shares with the same namespace.
 func TestAllSharesInclusionProof(t *testing.T) {
-	txs := testfactory.GenerateRandomTxs(64, 500)
+	txs := testfactory.GenerateRandomTxs(243, 500)
 
-	dataSquare, err := square.Construct(txs.ToSliceOfBytes(), appconsts.LatestVersion, appconsts.SquareSizeUpperBound(appconsts.LatestVersion))
-	if err != nil {
-		panic(err)
-	}
+	dataSquare, err := square.Construct(txs.ToSliceOfBytes(), appconsts.LatestVersion, 128)
+	require.NoError(t, err)
+	assert.Equal(t, 256, len(dataSquare))
 
 	// erasure the data square which we use to create the data root.
 	eds, err := da.ExtendShares(shares.ToBytes(dataSquare))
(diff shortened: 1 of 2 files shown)


pkg/proof/proof_test.go
```diff
@@ -221,3 +221,35 @@ func TestNewShareInclusionProof(t *testing.T) {
 		})
 	}
 }
+
+// TestAllSharesInclusionProof creates proof for all the shares in a block.
+// Since we can't prove multiple namespaces at the moment, the block will contain all the shares with the same namespace.
```
rootulp, line 226: [nit] rephrased + wrapped 

```suggestion
// TestAllSharesInclusionProof creates a proof for all shares in the data
// square. Since we can't prove multiple namespaces at the moment, all the
// shares use the same namespace.
```

pkg/proof/proof_test.go
```diff
@@ -221,3 +221,35 @@ func TestNewShareInclusionProof(t *testing.T) {
...
+func TestAllSharesInclusionProof(t *testing.T) {
+	txs := testfactory.GenerateRandomTxs(64, 500)
+
+	dataSquare, err := square.Construct(txs.ToSliceOfBytes(), appconsts.LatestVersion, appconsts.SquareSizeUpperBound(appconsts.LatestVersion))
+	if err != nil {
+		panic(err)
+	}
```
rootulp, line 233: [optional]

```suggestion
	require.NoError(t, err)
```

pkg/proof/proof_test.go
```diff
@@ -221,3 +221,35 @@ func TestNewShareInclusionProof(t *testing.T) {
...
+	eds, err := da.ExtendShares(shares.ToBytes(dataSquare))
+	require.NoError(t, err)
+
+	// create the new data root by creating the data availability header (merkle
+	// roots of each row and col of the erasure data).
+	dah, err := da.NewDataAvailabilityHeader(eds)
+	require.NoError(t, err)
+	dataRoot := dah.Hash()
+
+	actualNID, err := proof.ParseNamespace(dataSquare, 0, 64)
```
rootulp, line 245: [blocking][question] should this test assert that the len of the data square is 64 shares? When I add

```
	assert.Equal(t, 64, len(dataSquare))
```

I observe an error that makes me think the data square actually has 256 shares in which case this test doesn't actually verify the last share share. Also I confirmed that this test passes if I revert the code change back to

```go
	if endShare >= len(rawShares) {
```

rootulp, line 245: [nit] rename `actualNID` => `actualNamespace` or `actualNs` because ParseNamespace returns a namespace which contains a version and an ID

rootulp (commented):
One blocking question

pkg/proof/proof_test.go
SweeXordious: good catch [b747251](https://github.com/celestiaorg/celestia-app/pull/2208/commits/b747251305c05c3955d6e34930500612004be0ee)

Now the created square has 256 tx shares and it's proving them

pkg/proof/proof_test.go
SweeXordious: done

pkg/proof/proof_test.go
SweeXordious: done

SweeXordious: What I am not sure of is this:
```
square.Construct(txs.ToSliceOfBytes(), appconsts.LatestVersion, 128)
```
I give the `128` square size but the output is 256 shares... still investigating why is that

evan-forbes: 128 is the max square size, so what is likely happening is we're getting a 16x16 square as there are only enough txs for that

SweeXordious: aaaaah, makes sense, thanks :+1:

SweeXordious: resolving

rootulp (approved):

evan-forbes (approved):

Issue #2066, opened 2023-07-10
Shrink the interface exposed to the `mint` module


SweeXordious:
## Context

https://github.com/celestiaorg/celestia-app/blob/60f932802e30c31c2f9dfa810bce0e74902cf716/x/mint/types/expected_keepers.go#L17

and

https://github.com/celestiaorg/celestia-app/blob/60f932802e30c31c2f9dfa810bce0e74902cf716/x/mint/types/expected_keepers.go#L24

## Problem

These two methods are not used, it's better to remove them to reduce the API exposed to the `mint` module

## Proposal

Remove them

PR #2216, opened 2023-08-05
chore!: shrink interface of x/mint expected keepers

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2066


SweeXordious (approved):

evan-forbes (approved):
out of curiosity, are these methods used in the standard mint module? if so, what for?

cmwaters (approved):

cmwaters:
> out of curiosity, are these methods used in the standard mint module? if so, what for?

They're also not used by the mainline SDK. I think it's legacy from an older refactor

staheri14 (approved):
Assuming there is no future use for the removed methods, LGTM!

PR #2219, opened 2023-08-07
chore: bump the max tx bytes in the mempool config

cmwaters:


app/default_overrides.go
```diff
@@ -166,7 +166,13 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	cfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
 	cfg.Mempool.TTLNumBlocks = 10
-	cfg.Mempool.MaxTxBytes = appconsts.DefaultMaxBytes
+	// Given that there is a stateful transaction size check in CheckTx,
+	// We set a loose upper bound on what we expect the transaction to
+	// be based on the upper bound size of the entire block for the given
+	// version. This acts as a first line of DoS protection
+	upperBoundBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * appconsts.ContinuationSparseShareContentSize
+	cfg.Mempool.MaxTxBytes = upperBoundBytes
```
rootulp, line 174: I think this change makes sense b/c otherwise when governance bumps the gov max square size param from 64 to 128, a transaction that occupies (roughly) more than half the new square size will be rejected by all validators with the default max tx bytes. Validators can accept by modifying their local MaxTxBytes in `config.toml` but it seems prescient to bump the default now.

[optional] 
1. Do we want to change it here? https://github.com/celestiaorg/celestia-app/blob/a495d40f608dbae177b4cbd3ec2924e20b589847/test/util/testnode/config.go#L156-L159
2. Do we want to extract an `appconst` const for it?

[discussion]
> stateful transaction size check in CheckTx
The stateful check in CheckTx uses `appconsts.ShareSize` instead of `appconsts.ContinuationSparseShareContentSize` here:

```go
// squareBytes returns the number of bytes in a square for the given squareSize.
func squareBytes(squareSize int) int {
	numShares := squareSize * squareSize
	return numShares * appconsts.ShareSize
}
```

Mostly because the upper bound max blob size doesn't need to be exact. Similarly, `MaxTxBytes` doesn't need to be exact and I think it's slightly cleaner to rely on `appconsts.ShareSize` which is much less likely to change than `appconsts.ContinuationSparseShareContentSize` which could change with a share version bump. So we may consider changing `appconsts.DefaultMaxBytes` and `cfg.Mempool.MaxTxBytes` to calculate based on `appconsts.ShareSize` but definitely optional.

rootulp (approved):

evan-forbes (dismissed):

app/default_overrides.go
evan-forbes: I could see the benefit of switching to something simpler for readability. I'm also fine with using a more accurate value

cmwaters: > Do we want to change it here?

Yeah that makes sense

> Do we want to extract an appconst const for it?

I'm indifferent. If it were used in more places I'd probably be for it

evan-forbes (approved):

app/default_overrides.go
rootulp: I like appconsts for this but definitely not blocking on it.

rootulp (approved):

evan-forbes:
do we need to update this branch for the backport to work?

Issue #1899, opened 2023-06-09
Test for the ability to include transactions in the first block


evan-forbes:
Currently, we have never tested that transactions can actually be included in the first block. We also have issues with accessing the state in PrepareProposal, that's why we have these lines

https://github.com/celestiaorg/celestia-app/blob/9bf0cf1dd9ce31a3fecb51310c3913820b21a8c2/app/square_size.go#L11-L19

If we can't actually process txs in the first block, and there ends up being transactions in the mempool, then we could end up with a "frustrated" start to the chain where we won't be able to create a block until a block producer creates an empty one.

cmwaters:
Have we looked to see how much work it would be to copy over the cosmos-sdk's solution since I don't think they have this problem?

Specifically they have their own `prepareProposalState` and `processProposalState`

staheri14:
> Have we looked to see how much work it would be to copy over the cosmos-sdk's solution since I don't think they have this problem?

> Specifically they have their own prepareProposalState and processProposalState

We haven't identified the underlying cause of the problem yet. Could you please share your perspective on how copying over the prepareProposalState and processProposalState functions will address the issue?

staheri14:
Progress related to this issue is documented on the following Notion page: [Including Transactions in the First Block](https://www.notion.so/celestiaorg/Including-transactions-in-the-first-block-704ea09d81bf4b4f901b232553236889).

evan-forbes:
the linked notion is a great analysis. It appears that it will be very difficult to get a tx included in the first block. So difficult that a validator would have to build a separate binary. 

In an abundance of caution, imo we should still add the hacky yet simple check in `PrepareProposal` that returns zero tx and the empty data root. I doubt we could ever git in a scenario where it could get hit, but it would avoid the worst case scenario described above where proposers would progressively panic

staheri14:
> In an abundance of caution, imo we should still add the hacky yet simple check in PrepareProposal that returns zero tx and the empty data root. I doubt we could ever git in a scenario where it could get hit, but it would avoid the worst case scenario described above where proposers would progressively panic

Sounds good to me too, please see the tracking issue #2221

evan-forbes:
I think we can close with #2233

staheri14:
[...]

PR #2220, opened 2023-08-07
test: tests tx inclusion in the first block

staheri14:
## Overview
As part of #1899, we need to test the network/validator's behavior for transaction inclusion in the first block (more precisely, when the first block has not been committed yet). This PR introduces a new test that utilizes a local consensus config option, namely `timeout_commit`, to slow down block production. This delay allows transactions to arrive at the time of block height 1.

It is important to note that the current test demonstrates that transaction inclusion fails due to the absence of commit info for block height 0. There are two theories regarding this behavior:

1) Theory 1: This is expected behavior as the initial state or genesis state might not be available by the protocol definition. Consequently, the first block remains empty because there is no valid state available to validate transactions against. (Our fork of Cosmos SDK is not in parity with the mainline Cosmos SDK, however, with the initial inspection, it seems the initial state does not get committed in the mainline Cosmos SDK too, hence it is suspected that the same problem exists there as well. However, further tests should be developed to confirm this)

2) Theory 2: This is not an expected behavior. The genesis state should be committed when initializing the chain (in `abci.InitChain`). In this case, the root cause should be found and the fix should be implemented.  

In either case, subsequent PRs will be addressing this issue.

## Checklist


test/util/testnode/config.go
```diff
@@ -99,6 +99,12 @@ func (c *Config) WithGensisTime(t time.Time) *Config {
 	return c
 }
 
+// WithCommitTimeout sets the CommitTimeout and returns the Config.
+func (c *Config) WithCommitTimeout(d time.Duration) *Config {
```
evan-forbes, line 103: should we use TimeoutCommit instead of CommitTimeout here just for consistency?

staheri14: Of course, here is the fix https://github.com/celestiaorg/celestia-app/pull/2220/commits/4a29d4fc2385a517869757250116f890dc4c53d2.

staheri14:
> we don't have to handle it all in this PR, but after we add the hacky if statement protection in `PrepareProposal`, we could also add a different test that waits for block height 2 and then ensures that the first block is empty

Sure, I'll address this in https://github.com/celestiaorg/celestia-app/issues/2221.

app/test/block_production_test.go
```diff
@@ -0,0 +1,52 @@
...
+	t := s.T()
+
+	accounts := make([]string, 40)
+	for i := 0; i < 40; i++ {
+		accounts[i] = tmrand.Str(10)
+	}
+
+	cfg := testnode.DefaultConfig().
+		WithAccounts(accounts).
+		WithTimeoutCommit(10 * time.Second)
```
cmwaters, line 38: Just so I understand correctly, if timeout commit is set to 10 seconds, does the first height start 10 seconds after genesis or still at genesis. I'm trying to understand why the node doesn't propose a block instantly before the transaction gets posted.

cmwaters (dismissed):
Generally looks good, I just have a few questions

app/test/block_production_test.go
staheri14: I have made further clarifications in [this comment](https://github.com/celestiaorg/celestia-app/pull/2220/commits/124bb200c7f715479051e078f1d2a93c11f70278) to explicitly state that height 1 has not started yet.

cmwaters: Will we want for mainnet that the first height is actually started at `GenesisTime`?

staheri14: Can you please elaborate on your question?

cmwaters: Well currently if I set the genesis time to 5pm, the first height actually starts 10 seconds later. Thus the first block is produced perhaps ~12 seconds after genesis. I was wondering whether that is okay or if we'd want the first height to start immediately

staheri14: We don't expect validator nodes to use timeout commit config option in the mainnet (we don't advertise so), we use this option here for testing purposes only. Therefore, in mainnet, the network will start proposing the first block immediately after Genesis time I.e., after loading Genesis state. 

Trusted to your question, Validators can start the first block immediately in the mainnet (we are also going to address #2221 which ensures block 1 would be empty anyway).

app/test/block_production_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	cctx, _, _ := testnode.NewNetwork(t, cfg)
+	s.cctx = cctx
+	s.accounts = accounts
+}
+
+func (s *BlockProductionestSuite) Test_PostData() {
```
rootulp, line 45: [optional] can this test include a comment on what it it's attempting to verify or rename to something more descriptive? Two options b/c I'm not strongly opinionated

```go
func (s *BlockProductionestSuite) TestInabilityToIncludeTransactionsInBlockOne() {
```

```go
Test_PostData verifies that no transactions can be included in the block at height one.
func (s *BlockProductionestSuite) Test_PostData() {
```

staheri14: Good suggestion, thanks! please see https://github.com/celestiaorg/celestia-app/pull/2220/commits/17979e8aa4e38fa32100d2c77d06a5892fa42881

app/test/block_production_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/stretchr/testify/suite"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func TestBlockProductionestSuite(t *testing.T) {
+	suite.Run(t, new(BlockProductionestSuite))
+}
+
+type BlockProductionestSuite struct {
```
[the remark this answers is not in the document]
staheri14: Fixed in https://github.com/celestiaorg/celestia-app/pull/2220/commits/17979e8aa4e38fa32100d2c77d06a5892fa42881

app/test/block_production_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"time"
+
+	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
+	"github.com/celestiaorg/celestia-app/test/util/testnode"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/stretchr/testify/suite"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func TestBlockProductionestSuite(t *testing.T) {
```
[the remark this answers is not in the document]
staheri14: Fixed in https://github.com/celestiaorg/celestia-app/pull/2220/commits/17979e8aa4e38fa32100d2c77d06a5892fa42881

app/test/block_production_test.go
```diff
@@ -0,0 +1,52 @@
...
+type BlockProductionestSuite struct {
+	suite.Suite
+
+	accounts []string
+	cctx     testnode.Context
+}
+
+func (s *BlockProductionestSuite) SetupSuite() {
+	if testing.Short() {
+		s.T().Skip("skipping full node integration test in short mode.")
```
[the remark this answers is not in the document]
staheri14: Incorporated your suggestion in https://github.com/celestiaorg/celestia-app/pull/2220/commits/17979e8aa4e38fa32100d2c77d06a5892fa42881.

evan-forbes (approved):

app/test/block_production_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/celestiaorg/celestia-app/test/util/testnode"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/stretchr/testify/suite"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+func TestBlockProductionTestSuite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping block production test suite in short mode.")
+	}
```
SweeXordious, line 17: [Question]
As of https://github.com/celestiaorg/celestia-app/commit/9ea27d47e487debe3b44dcbf74c832a5940f47a8, I guess the test name also needs to be added to the Makefile so that it's not run in `race` mode. However, I am seeing the CI to be green. Does this mean that the test node no longer have races?

@rootulp is this expected?

cmwaters (approved):

PR #2224, opened 2023-08-08
chore: remove test cmd from cli

cmwaters:
The celestia-appd currently has the command: `celestia-appd tx blob TestRandBlob` which I believe was used for testing scripts but shouldn't be part of the production CLI as  is actually stated in a comment. This PR removes the command


cmwaters:
It seems like the devops team are still depending on this command for their scripts so I'm putting this back into draft mode

rootulp (approved):

evan-forbes (approved):

Issue #2227, opened 2023-08-08
cat_pool specs markdown issues


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/2207 just merged. It passes markdownlint on CI but fails for me locally. I have markdownlint 0.35.0 installed locally.

## Problem

```
$ make lint
--> Running golangci-lint
--> Running markdownlint
specs/src/specs/cat_pool.md:6 MD001/heading-increment/header-increment Heading levels should only increment by one level at a time [Expected: h2; Actual: h3]
specs/src/specs/cat_pool.md:90:1 MD007/ul-indent Unordered list indentation [Expected: 2; Actual: 4]
specs/src/specs/cat_pool.md:91:1 MD007/ul-indent Unordered list indentation [Expected: 2; Actual: 4]
specs/src/specs/cat_pool.md:102 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
make: *** [lint] Error 1
```

## Proposal

1. Fix the markdowlint errors above
2. https://github.com/celestiaorg/celestia-app/issues/2212

PR #2233, opened 2023-08-09
feat: implements block height check in PrepareProposal to avoid unexpected errors at height 1

### LINKED ISSUE #1899: Test for the ability to include transactions in the first block

### LINKED ISSUE #2221: Implementing block height check in PrepareProposal for improved behavior
The following suggestion is related to #1899, and is originally suggested by @evan-forbes in this [comment](https://github.com/celestiaorg/celestia-app/pull/2220#pullrequestreview-1565761861)  

> we don't have to handle it all in this PR, but after we add the hacky if statement protection in PrepareProposal, we could also add a different test that waits for block height 2 and then ensures that the first block is empty

staheri14:
## Overview
Closes #1899 
Closes #2221

## Checklist


specs/src/specs/cat_pool.md
```diff
@@ -3,25 +3,25 @@
 - 01.12.2022 | Initial specification (@cmwaters)
 - 09.12.2022 | Add Push/Pull mechanics (@cmwaters)
 
-### Outline
```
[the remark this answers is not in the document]
rootulp: Thanks for fixing! We could update the PR description to say "Closes https://github.com/celestiaorg/celestia-app/issues/2227" or extract the lint fix to a separate PR. Defer to you

app/prepare_proposal.go
```diff
@@ -32,7 +32,23 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 		ante.DefaultSigVerificationGasConsumer,
 		app.IBCKeeper,
 	)
-	txs := FilterTxs(sdkCtx, handler, app.txConfig, req.BlockData.Txs)
+
+	var txs [][]byte
+	if app.LastBlockHeight() == 0 {
```
evan-forbes, line 37: [optional]

I'll defer to the author if its needed or not, but it might be nice to briefly explain this or link to an issue that does for more context

evan-forbes (approved):
dope! thanks, can we also open an issue to remove this after mainnet?

specs/src/specs/cat_pool.md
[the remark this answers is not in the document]
staheri14: Thanks for mentioning the GH issue, I cut a separate PR for it, please see https://github.com/celestiaorg/celestia-app/pull/2242

app/prepare_proposal.go
```diff
@@ -32,7 +32,23 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
+		// the first block is expected to be empty; hence return an empty set of txs
+		txs = make([][]byte, 0)
+		if len(req.BlockData.Txs) != 0 {
+			// if the consensus layer sends non-empty set of transactions for
+			// block height 1, log it
+			app.Logger().Info(
+				"non-empty txs received from the consensus layer for block height 1",
+				"invalid",
+				len(req.BlockData.Txs),
```
SweeXordious, line 46:
```suggestion
				"non-empty txs received from the consensus layer for block height 1",
				"numberOfTransactions",
				len(req.BlockData.Txs),
```

evan-forbes (dismissed):

app/prepare_proposal.go
cmwaters: Yeah since this is somewhat of an unusual hack I think it would be good to point out what why this is needed

app/prepare_proposal.go
```diff
@@ -32,7 +32,23 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
-	txs := FilterTxs(sdkCtx, handler, app.txConfig, req.BlockData.Txs)
+
+	var txs [][]byte
+	if app.LastBlockHeight() == 0 {
+		// the first block is expected to be empty; hence return an empty set of txs
+		txs = make([][]byte, 0)
+		if len(req.BlockData.Txs) != 0 {
+			// if the consensus layer sends non-empty set of transactions for
+			// block height 1, log it
+			app.Logger().Info(
```
cmwaters, line 43: Why do we want to log this? i.e. what value does it bring?

cmwaters (dismissed):
LGTM just have one non-blocking question

app/prepare_proposal.go
staheri14: Thanks, fixed it in https://github.com/celestiaorg/celestia-app/pull/2233/commits/93692c5d4876e18d3c4f366e8d9126df608679c4.

app/prepare_proposal.go
staheri14: Sure, please see the added explanations https://github.com/celestiaorg/celestia-app/pull/2233/commits/93692c5d4876e18d3c4f366e8d9126df608679c4.

evan-forbes (approved):

staheri14:
> also open an issue to remove this after mainnet

Sure, done https://github.com/celestiaorg/celestia-app/issues/2271

app/test/block_production_test.go
```diff
@@ -55,3 +57,35 @@ func (s *BlockProductionTestSuite) Test_BlockOneTransactionNonInclusion() {
...
+	// wait until height 1 before posting transactions
+	// otherwise tx submission will fail
+	time.Sleep(1 * s.timeoutCommit)
+	// send some transactions, these should be included in the second block
+	_, err := s.cctx.PostData(s.accounts[0], flags.BroadcastBlock, appns.RandomBlobNamespace(), tmrand.Bytes(100000))
+	require.NoError(err)
+
+	// wait for 2*s.timeoutCommit+1*time.Second to ensure that the node is
+	// at height 2
+	_, err = s.cctx.WaitForHeightWithTimeout(2, 2*s.timeoutCommit+1*time.Second)
```
rootulp, line 73: [question] if this only cares about waiting for block height 2, is that specific time necessary? Very minor but It's difficult to read. Put another way, this seems to work:

```suggestion
	// Wait for block height 2.
	_, err = s.cctx.WaitForHeightWithTimeout(2, s.timeoutCommit)
	require.NoError(err)
```

app/test/block_production_test.go
```diff
@@ -55,3 +57,35 @@ func (s *BlockProductionTestSuite) Test_BlockOneTransactionNonInclusion() {
 	require.Error(err) // change this to require.NoError(err) to see the error
 	require.ErrorContains(err, "rpc error: code = Unknown desc = codespace sdk code 18: invalid request: failed to load state at height 0; no commit info found (latest height: 0)")
 }
+
+// Test_FirstBlockIsEmpty tests whether the first block is empty.
+func (s *BlockProductionTestSuite) Test_FirstBlockIsEmpty() {
+	require := s.Require()
+	// wait until height 1 before posting transactions
+	// otherwise tx submission will fail
+	time.Sleep(1 * s.timeoutCommit)
```
rootulp, line 66: [question][not blocking] it's not immediately obvious what the relationship is between timeout commit and block height 1. If this only cares about waiting for block height 1 then it could be

```suggestion
	// Wait until block height 1 before posting transactions otherwise tx
	// submission will fail.
	_, err := s.cctx.WaitForHeightWithTimeout(1, s.timeoutCommit)
	require.NoError(err)
```

rootulp (approved):

app/prepare_proposal.go
cmwaters: But we can conclude that it could be the case right? I'm just trying to work out what value it provides for node operators

staheri14: > But we can conclude that it could be the case right?

We expect that it would not be the case at all unless we observe this logged info. 

> I'm just trying to work out what value it provides for node operators

I understand your point. The current log doesn't tell node operators much unless they're interested in the problem of transactions in the mempool at height 0. However, having this log could help us check if this happens during local tests. I don't have a strong opinion on this though. If you think it's better, we can remove the log. Just let me know.

staheri14:
Oops, my bad. The auto-merge was on, so the PR got merged. I'll create a new PR to take care of the comments.

app/test/block_production_test.go
staheri14: Both methods work well: either sleeping for the timeoutCommit or waiting for height 1. However, for a clearer understanding, I've chosen to explicitly wait, and I've incorporated your suggestion in this [PR](https://github.com/celestiaorg/celestia-app/pull/2284). Thanks.

app/test/block_production_test.go
staheri14: Thanks, your comment is addressed in this [PR](https://github.com/celestiaorg/celestia-app/pull/2284).

app/prepare_proposal.go
staheri14: Created this PR to address your comment https://github.com/celestiaorg/celestia-app/pull/2285

PR #2238, opened 2023-08-09
chore: extract out signer into a separate package

cmwaters:
This adds a new `user` package which includes the modified `Signer` with a simple set of APIs for building, signing and broadcasting transactions.

One advantage of this `Signer` is that it locally tracks the nonce thus allowing for multiple blobs to be correctly submitted in a block (so long as FIFO order is maintained and not multiple signers are being user)

Addresses: https://github.com/celestiaorg/celestia-app/issues/1162


pkg/user/signer.go
```diff
@@ -0,0 +1,360 @@
...
+		chainID:               chainID,
+		accountNumber:         accountNumber,
+		lastSignedSequence:    sequence,
+		lastConfirmedSequence: sequence,
+		pollTime:              defaultPollTime,
+	}, nil
+}
+
+// SetupSingleSigner relies on the keyring with a single key. It extracts the address from the key and uses the
+// grpc connection to populate the chainID, account number and sequence number.
```
rootulp, line 81: [optional] I'm confused by this usage of "relies". Perhaps clearer:

```suggestion
// SetupSingleSigner sets up a signer based on the provided keyring. The keyring
// must contain exactly one key. It extracts the address from the key and uses
// the grpc connection to populate the chainID, account number, and sequence
// number.
```

pkg/user/tx_options.go
```diff
@@ -0,0 +1,99 @@
...
+
+func SetFeeGranter(feeGranter sdk.AccAddress) TxOption {
+	return func(builder sdkclient.TxBuilder) sdkclient.TxBuilder {
+		builder.SetFeeGranter(feeGranter)
+		return builder
+	}
+}
+
+// InheritTxConfig sets all of the accessible configurations from a given tx
+// into a a give client.TxBuilder
```
rootulp, line 62:
```suggestion
// into a given client.TxBuilder
```

pkg/user/signer.go
```diff
@@ -0,0 +1,360 @@
...
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmtypes "github.com/tendermint/tendermint/types"
+	"google.golang.org/grpc"
+)
+
+const defaultPollTime = 3 * time.Second
+
+// Signer is an abstraction for managing building, signing and broadcasting Celestia transactions
```
rootulp, line 29: [extremely optional] I re-read this a few times and "managing" doesn't seem necessary

```suggestion
// Signer is an abstraction for building, signing, and broadcasting Celestia transactions
```

rootulp (approved):
LGTM!

[optional] I think pkg/user could benefit from a README but def not blocking

pkg/user/signer_test.go
```diff
@@ -0,0 +1,56 @@
...
+	resp, err := signer.SubmitPayForBlob(ctx.GoContext(), blobs, fee, gas)
+	require.NoError(t, err)
+	require.EqualValues(t, 0, resp.Code)
+}
+
+func TestSubmitTx(t *testing.T) {
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	ctx, _, grpcAddr := testnode.NewNetwork(t, testnode.DefaultConfig())
+	conn, err := grpc.Dial(grpcAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
```
evan-forbes, line 43: if we want, the grpc connection is already added the client context that is returned by newnetwork

pkg/user/signer_test.go
```diff
@@ -0,0 +1,56 @@
...
+	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/libs/rand"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+func TestSubmitPayForBlob(t *testing.T) {
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	ctx, _, grpcAddr := testnode.NewNetwork(t, testnode.DefaultConfig())
```
evan-forbes, line 22: [optional]

if we use a suite, then the tests will run a bit faster, and it might save on some duplicate setup

evan-forbes (dismissed):
LGTM, I like the new names and api

only had a question and comment

evan-forbes:
if we want to close #1162 that's cool, but then we should open another issue to remove the keyring signer so we don't forget

cmwaters:
> if we want to close https://github.com/celestiaorg/celestia-app/issues/1162 that's cool, but then we should open another issue to remove the keyring signer so we don't forget

I think I will leave it open and we can close it once we remove the existing `KeyringSigner`

pkg/user/signer_test.go
cmwaters: Makes sense. Let me refactor this

evan-forbes (approved):

evan-forbes (approved):

pkg/user/signer.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/cosmos/cosmos-sdk/types/tx"
+	"github.com/cosmos/cosmos-sdk/types/tx/signing"
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmtypes "github.com/tendermint/tendermint/types"
+	"google.golang.org/grpc"
+)
+
+const defaultPollTime = 3 * time.Second
```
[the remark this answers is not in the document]
cmwaters: Both the test node and actual nodes. I do agree that this should be configurable but want to address it in a later PR

cmwaters:
I broke some test.... fix is on the way

rootulp (approved):

rootulp (approved):

rootulp (approved):

evan-forbes (approved):

evan-forbes:
do we need to add any tests to the race detector block list from the user package?

rootulp:
Yea potentially `TestSignerTestSuite` based on the failure [here](https://github.com/celestiaorg/celestia-app/actions/runs/5856698727/job/15877476279?pr=2238#step:4:2270).

cmwaters:
Yeah makes sense. Let me do that

evan-forbes (approved):

test/util/testnode/utils.go
```diff
@@ -111,7 +109,7 @@ func FundKeyringAccounts(accounts ...string) (keyring.Keyring, []banktypes.Balan
 		}
 
 		balances := sdk.NewCoins(
-			sdk.NewCoin(bondDenom, sdk.NewInt(99999999999999999)),
+			sdk.NewCoin(appconsts.BondDenom, sdk.NewInt(99999999999999999)),
```
rootulp, line 112: 👍

rootulp (approved):

test/util/testnode/utils_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestTestAccount(t *testing.T) {
+	kr := testnode.TestKeyring()
+	record, err := kr.Key(testnode.TestAccName)
+	require.NoError(t, err)
+	addr, err := record.GetAddress()
+	require.NoError(t, err)
+	fmt.Println(addr)
```
staheri14, line 17: [nit] I think this fmt is not required and can be removed.

test/util/testnode/genesis_options.go
```diff
@@ -41,3 +44,38 @@ func ImmediateProposals(codec codec.Codec) GenesisOption {
...
+	return func(state map[string]json.RawMessage) map[string]json.RawMessage {
+		// set the accounts in the genesis state
+		var authGenState authtypes.GenesisState
+		codec.MustUnmarshalJSON(state[authtypes.ModuleName], &authGenState)
+
+		genAccounts := make([]authtypes.GenesisAccount, len(addresses))
+		genBalances := make([]banktypes.Balance, len(addresses))
+		for idx, addr := range addresses {
+			genAccounts[idx] = authtypes.NewBaseAccount(addr, nil, uint64(idx+len(authGenState.Accounts)), 0)
+			fmt.Println("base account", genAccounts[idx])
```
staheri14, line 61: [nit] Is this Println intentional?

test/util/testnode/utils_test.go
```diff
@@ -0,0 +1,20 @@
...
+	require.Equal(t, testnode.TestAccAddr, addr.String())
+	require.Equal(t, testnode.TestAddress(), addr)
```
staheri14, line 19: It is a bit unclear why these equalities should hold, I think a function description could clarify the purpose of the test and the checks therein.

test/util/testnode/utils.go
```diff
@@ -0,0 +1,119 @@
+package testnode
```
staheri14, line 1: I suggest adding descriptions for each function in this file and shedding light on aspects like the outputs are randomized or deterministic (if the function name does not indicate so). This will help other developers understand the purpose and usage of the tools without having to delve deep into the code.

staheri14 (approved):
LGTM! added some comments, all non-blocking. 
I'd like to present a viewpoint regarding the size and content of PRs. It would be advantageous if we could divide PRs into smaller, self-contained units with singular purposes, as much as possible. This strategy would enhance our ability to monitor changes on the main branch and also simplify the process for reviewers to conduct reviews and distinguish between refactoring modifications and new implementations.

test/util/testnode/genesis_options.go
cmwaters: nope let me remove it

Wondertan:
I wonder if the changes in https://github.com/celestiaorg/celestia-node/pull/2511 should also leave in the Signer

test/util/testnode/utils.go
cmwaters: This file was moved across from the testfactory package. I didn't want to do this but we have a habit of writing tests using the same package name. As a lot of tests require the encoding config which imports the `app` package and essentially everything else we can often find ourselves caught in a dependency loop. I would prefer if people wrote tests using the `package types_test` instead of `package types` and would want to slowly move the repo to using that convention.

Alternatively, I thought about having the encoding.Config passed as a variable which might still be the better idea

cmwaters:
> I wonder if the changes in [celestiaorg/celestia-node#2511](https://github.com/celestiaorg/celestia-node/pull/2511) should also leave in the Signer

Kind of want to avoid overloading one class. This should just be responsible for signing, broadcasting, encoding and nonce management (even this is probably a little too much). I still want people to get the UX advantages of the `Signer` but still provide their own gas estimation (especially since we haven't got a mechanism for estimating gas for every transaction, just blobTxs) and gas pricing. But maybe this class still resides in app. I'm not sure

cmwaters:
> I'd like to present a viewpoint regarding the size and content of PRs. It would be advantageous if we could divide PRs into smaller, self-contained units with singular purposes, as much as possible. This strategy would enhance our ability to monitor changes on the main branch and also simplify the process for reviewers to conduct reviews and distinguish between refactoring modifications and new implementations.

I totally agree with this (although you might not think so from this PR 😅 ). Instead of migrating the signer package I planned on breaking it down into two PRs first to create the new package and then to remove the old one and migrate the tests across. However the tests were breaking even with just the new package and so I ended up modifying a lot more than planned. I'm wondering if I can still break it up and keep the refactoring diff to a minimum (although I still hold that our testing infra needs a lot of love)

cmwaters:
Per @staheri14's suggestion, I have modelled a new PR with just the new user package code here: https://github.com/celestiaorg/celestia-app/pull/2275. After merging, I will revamp this PR such that it targets cleaning up the test infra

cmwaters:
Going to close this in replacement of https://github.com/celestiaorg/celestia-app/pull/2314 and https://github.com/celestiaorg/celestia-app/pull/2275 which created the `user` package

Issue #2241, opened 2023-08-09
Inaccurate comment about `GovMaxSquareSize` being superseded


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/b70048b275a0e16e0509634c2175c99e2aad7bd1/x/blob/README.md?plain=1#L64-L70

## Problem

`GovMaxSquareSize` isn't superseded by `MaxSquareSize`. Maybe this sentence was trying to say that `GovMaxSquareSize` is upper bounded by `MaxSquareSize` but that also isn't the case (see https://github.com/celestiaorg/celestia-app/issues/2185).

## Proposal

Remove this sentence

> This value is superseded by the `MaxSquareSize`, which is hardcoded and cannot change without hardforking the chain.

sumanthsumu1418:
Hii can i work on this ..??

aliaaquib:
@rootulp i want to work on that issue!

rootulp:
Sorry @sumanthsumu1418 already has a PR open https://github.com/celestiaorg/celestia-app/pull/2245

PR #2245, opened 2023-08-11
 docs: improve comment about GovMaxSquareSize

sumanthsumu1418:
Closes https://github.com/celestiaorg/celestia-app/issues/2241

### DIFF
--- a/x/blob/README.md
+++ b/x/blob/README.md
@@ -64,8 +64,7 @@ value is set below that of normal transaction gas consumption, which is 10.
 #### `GovMaxSquareSize`
 
 `GovMaxSquareSize` is the maximum size of a data square that is considered valid
-by the validator set. This value is superseded by the `MaxSquareSize`, which is
-hardcoded and cannot change without hardforking the chain. See
+by the validator set. See
 [ADR021](../../docs/architecture/adr-021-restricted-block-size.md) for more
 details.
 


x/blob/README.md
```diff
@@ -64,8 +64,7 @@ value is set below that of normal transaction gas consumption, which is 10.
 #### `GovMaxSquareSize`
 
 `GovMaxSquareSize` is the maximum size of a data square that is considered valid
-by the validator set. This value is superseded by the `MaxSquareSize`, which is
-hardcoded and cannot change without hardforking the chain. See
+by the validator set. See
```
cmwaters, line 67: I think we should still mention the upper bound but instead say something like:
"This value is capped by the `SquareSizeUpperBound`, which is a versioned constant."

rootulp: I think that phrasing is misleading because `GovMaxSquareSize` isn't actually capped by `SquareSizeUpperBound`. See https://github.com/celestiaorg/celestia-app/issues/2185

cmwaters: True but I think we should somehow mention that the effective size of the square can't be exceeded by that value even if you change `GovMaxSquareSize` beyond it

rootulp: Makes sense. Evan had this idea: https://github.com/celestiaorg/celestia-app/issues/2267

rootulp:
Closing as stale.

PR #2251, opened 2023-08-11
fix(deps): bump godeltaprof version to support golang v1.21

SweeXordious:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/2249

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #2252, opened 2023-08-11
fix: TestLegacyGovUpgradeFailure flakiness

SweeXordious:
## Overview

This fixes the flakiness of the upgrade test sometimes perceived in CI and also locally.

The CI might fail if https://github.com/celestiaorg/celestia-app/pull/2251 is still not merged.

## Checklist

### DIFF
--- a/x/upgrade/test/integration_test.go
+++ b/x/upgrade/test/integration_test.go
@@ -112,10 +112,9 @@ func (s *UpgradeTestSuite) TestLegacyGovUpgradeFailure() {
 	res, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, acc, msg)
 	require.NoError(t, err)
 	require.Equal(t, abci.CodeTypeOK, res.Code) // we're only submitting the tx, so we expect everything to work
-	require.NoError(t, s.cctx.WaitForNextBlock())
 
 	// compare the result after the tx has been executed.
-	finalResult, err := testnode.QueryTx(s.cctx.Context, res.TxHash, false)
+	finalResult, err := s.cctx.WaitForTx(res.TxHash, 100)
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


x/upgrade/test/integration_test.go
```diff
@@ -112,10 +112,9 @@ func (s *UpgradeTestSuite) TestLegacyGovUpgradeFailure() {
 	res, err := testnode.SignAndBroadcastTx(s.ecfg, s.cctx.Context, acc, msg)
 	require.NoError(t, err)
 	require.Equal(t, abci.CodeTypeOK, res.Code) // we're only submitting the tx, so we expect everything to work
-	require.NoError(t, s.cctx.WaitForNextBlock())
 
 	// compare the result after the tx has been executed.
-	finalResult, err := testnode.QueryTx(s.cctx.Context, res.TxHash, false)
+	finalResult, err := s.cctx.WaitForTx(res.TxHash, 100)
```
evan-forbes, line 117: 100 seems a bit excessive, as if it actually took 99 blocks, something is very broken, but it would likely not be directly related to this test so definitely blocking on it

evan-forbes (approved):

x/upgrade/test/integration_test.go
SweeXordious: The 100 is an arbitrary number I used to be sure we waited enough. Since the timeout commit is in the order of ms, I think 100 is nothing, and it would allow us to know that something is really broken if the transaction is not included in any of the 100 blocks.
However, I can use a smaller number if you prefer it. If so, let me know which number you prefer

evan-forbes: we should expect it be accepted within a few blocks, so something as low as 5 should be sufficient. This change is not blocking

Issue #2253, opened 2023-08-11
Reserve the last 256 namespaces for the protocol


evan-forbes:
as originally proposed by @cmwaters, in a similar fashion to how we are reserving the first 256 and the last namespace for the protocol, we should also reserve the last 256 namespaces.

musalbas:
What's the rationale for this?

PR #2257, opened 2023-08-11
feat!: reserve the last 256 namespaces for protocol use

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2253

### DIFF
--- a/pkg/namespace/consts.go
+++ b/pkg/namespace/consts.go
@@ -51,10 +51,17 @@ var (
 	// (ordinary and PFBs) but before blobs.
 	ReservedPaddingNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 255})
 
-	// MaxPrimaryReservedNamespace represents the largest primary reserved
-	// namespace reserved for protocol use. Note that there may be other
-	// non-primary reserved namespaces beyond this upper limit.
+	// MaxPrimaryReservedNamespace is the highest primary reserved namespace.
+	// Namespaces lower than this are reserved for protocol use.
 	MaxPrimaryReservedNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 255})
+
+	// MinSecondaryReservedNamespace is the lowest secondary reserved namespace
+	// reserved for protocol use. Namespaces higher than this are reserved for
+	// protocol use.
+	MinSecondaryReservedNamespace = Namespace{
+		Version: math.MaxUint8,
+		ID:      append(bytes.Repeat([]byte{0xFF}, NamespaceIDSize-1), 0),
+	}
 
 	// TailPaddingNamespace is the namespace reserved for tail padding. All data
 	// with this namespace will be ignored.
--- a/pkg/namespace/consts.go
+++ b/pkg/namespace/consts.go
@@ -36,46 +36,50 @@ var (
 	// NamespaceVersionZeroPrefix is the prefix of a namespace ID for version 0.
 	NamespaceVersionZeroPrefix = bytes.Repeat([]byte{0}, NamespaceVersionZeroPrefixSize)
 
-	// TxNamespace is the namespace reserved for transaction data.
-	TxNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 1})
+	// TxNamespace is the namespace reserved for ordinary Cosmos SDK transactions.
+	TxNamespace = primaryReservedNamespace(0x01)
 
 	// IntermediateStateRootsNamespace is the namespace reserved for
 	// intermediate state root data.
-	IntermediateStateRootsNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 2})
+	IntermediateStateRootsNamespace = primaryReservedNamespace(0x02)
 
 	// PayForBlobNamespace is the namespace reserved for PayForBlobs transactions.
-	PayForBlobNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 4})
+	PayForBlobNamespace = primaryReservedNamespace(0x04)
 
-	// ReservedPaddingNamespace is the namespace used for padding after all
-	// reserved namespaces. In practice this padding is after transactions
-	// (ordinary and PFBs) but before blobs.
-	ReservedPaddingNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 255})
+	// PrimaryReservedPaddingNamespace is the namespace used for padding after all
+	// primary reserved namespaces.
+	PrimaryReservedPaddingNamespace = primaryReservedNamespace(0xFF)
 
 	// MaxPrimaryReservedNamespace is the highest primary reserved namespace.
 	// Namespaces lower than this are reserved for protocol use.
-	MaxPrimaryReservedNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 255})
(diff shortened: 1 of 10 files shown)


pkg/namespace/consts.go
```diff
@@ -36,39 +36,50 @@ var (
...
-	PayForBlobNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 4})
+	PayForBlobNamespace = primaryReservedNamespace(0x04)
 
-	// ReservedPaddingNamespace is the namespace used for padding after all
-	// reserved namespaces. In practice this padding is after transactions
-	// (ordinary and PFBs) but before blobs.
-	ReservedPaddingNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 255})
+	// PrimaryReservedPaddingNamespace is the namespace used for padding after all
+	// primary reserved namespaces.
+	PrimaryReservedPaddingNamespace = primaryReservedNamespace(0xFF)
```
rootulp, line 51: [optional] note to reviewers, I think it may be worth including one extra breaking change to this PR. Specifically:
 
```suggestion
	PrimaryReservedPaddingNamespace = primaryReservedNamespace(0xFE)
```

This change would differentiate MaxPrimaryReserved and PrimaryReservedPadding which seems redundant.

This proposal is partially motivated by the fact that tail padding is 0xFE which leaves one extra namespace at the very end for parity shares. I wonder if it may be helpful to have one extra namespace at the end of the primary range just in case.

specs/src/specs/namespace.md
```diff
@@ -57,30 +57,32 @@ The ID is encoded as a byte slice of length 28.
 
 ## Reserved Namespaces
```
rootulp, line 58: I rephrased this section a bit for clarity. Please LMK if you think I should revert anything @staheri14

pkg/namespace/consts.go
```diff
@@ -36,39 +36,50 @@ var (
...
 
-	// TxNamespace is the namespace reserved for transaction data.
-	TxNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 1})
+	// TxNamespace is the namespace reserved for ordinary Cosmos SDK transactions.
+	TxNamespace = primaryReservedNamespace(0x01)
 
 	// IntermediateStateRootsNamespace is the namespace reserved for
 	// intermediate state root data.
-	IntermediateStateRootsNamespace = MustNewV0([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 2})
+	IntermediateStateRootsNamespace = primaryReservedNamespace(0x02)
```
SweeXordious, line 44: [unrelated question]
Where is namespace `0x03` for? Is it reserved for something?

pkg/namespace/consts.go
```diff
@@ -36,39 +36,50 @@ var (
...
+func primaryReservedNamespace(lastByte byte) Namespace {
+	return Namespace{
+		Version: NamespaceVersionZero,
+		ID:      append(bytes.Repeat([]byte{0x00}, NamespaceIDSize-1), lastByte),
+	}
+}
+
+func secondaryReservedNamespace(lastByte byte) Namespace {
+	return Namespace{
+		Version: NamespaceVersionMax,
```
SweeXordious, line 82: [question to understand]
why is the version the max if it's meant to be used by us?

Also, probably unrelated. Before this PR, if someone wants to use the namespace `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFD`, the last namespace available, but with version 0, is that possible?

Wondertan (commented):
Does node needs to be updated with this?

pkg/namespace/consts.go
SweeXordious: [Question] Does this mean having one padding type for the primary reserved namespaces, and another padding type for the secondary ones? not the same padding namespace?

rootulp: Correct. Multiple namespaces are already used for different padding shares. See https://celestiaorg.github.io/celestia-app/specs/shares.html#reserved-padding-share

pkg/namespace/consts.go
rootulp: It was previously used for evidence but is no longer used.

pkg/namespace/consts.go
rootulp: > why is the version the max if it's meant to be used by us?

The protocol reserves namespaces at the end of namespace range (i.e. with version = max) so that these namespaces are ordered after all possible blob shares. See [here](https://github.com/celestiaorg/celestia-app/blob/86c06b5550bcee9be5065876c3ced770637f1de5/specs/src/specs/namespace.md?plain=1#L67-L70).

> Also, probably unrelated. Before this PR, if someone wants to use the namespace 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFD, the last namespace available, but with version 0, is that possible?

No because the namespace you provided doesn't satisfy the version 0 constraints. See [here](https://github.com/celestiaorg/celestia-app/blob/86c06b5550bcee9be5065876c3ced770637f1de5/specs/src/specs/namespace.md?plain=1#L32-L48).

rootulp:
> Does node needs to be updated with this?

Yes. From a cursory investigation:

1. Rename [`MaxReservedNamespace`](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L17C2-L17C2) to `MaxPrimaryReservedNamespace`
1. Rename [`ReservedPaddingNamespace`](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L20C8-L20C8) to `PrimaryReservedPaddingNamespace`
1. Add `MinSecondaryReservedNamespace` constant to [this list](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L15-L24)
1. [`ValidateForBlob`](https://github.com/celestiaorg/celestia-node/blob/87e9500120c27450e264ab4ffd59c87e92186194/share/namespace.go#L122-L131) should account for the secondary reserved namespace range.
1. Add unit tests

specs/src/specs/namespace.md
```diff
@@ -57,30 +57,32 @@ The ID is encoded as a byte slice of length 28.
...
-Applications MUST NOT use these reserved namespaces for their blob data.
+Celestia reserves some namespaces for protocol use.
+These namespaces are called "reserved namespaces".
+Reserved namespaces are used to arrange the contents of the [data square](./data_square_layout.md).
+Applications MUST NOT use reserved namespaces for their blob data.
+Reserved namespaces fall into two categories: _Primary_ and _Secondary_.
 
-Reserved namespaces fall into two categories, _Primary Reserved Namespaces_ and _Secondary Reserved Namespaces_.
+- Primary: Namespaces with values less than or equal to `0x00000000000000000000000000000000000000000000000000000000FF`.
+- Secondary: Namespaces with values greater than or equal to `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF00`.
```
cmwaters, line 67: Do we want to mention somewhere that secondary reserved namespaces always have a namespace version of 255

cmwaters (dismissed):
LGTM

specs/src/specs/namespace.md
cmwaters: Even the primary reserved namespaces are only relevant for a particular version

rootulp: Yea can do.

evan-forbes (approved):
nice! this is even simpler than the old implementation!

PR #2275, opened 2023-08-15
feat: create user package and signer class

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/pull/2238

Creates a user package based from the the `KeyringSigner` in the blob module. A future PR will refactor the tests and remove the `KeyringSigner`


pkg/user/signer.go
```diff
@@ -0,0 +1,375 @@
...
+		},
+	)
+	if err == nil {
+		return resp.TxResponse, err
+	}
+
+	// this is a bit brittle
+	if !strings.Contains(err.Error(), "not found") {
+		return &sdktypes.TxResponse{}, err
+	}
```
evan-forbes, line 223: I've never understood why the response returned by the client here is nil when an error is returned! I wish it wasn't the case so we could avoid this

evan-forbes (approved):
:+1: LGTM

pkg/user/signer.go
```diff
@@ -0,0 +1,375 @@
...
+	signature, _, err := s.keys.SignByAddress(s.address, bytesToSign)
+	if err != nil {
+		return nil, fmt.Errorf("error signing bytes: %w", err)
+	}
+
+	return signature, nil
+}
+
+// NewTxBuilder returns the default sdk Tx builder using the celestia-app encoding config
+func (s *Signer) txBuilder(opts ...TxOption) client.TxBuilder {
```
rootulp, line 341: [not blocking] I think the godoc didn't get updated after a function rename

```suggestion
// txBuilder returns the default sdk Tx builder using the celestia-app encoding config
func (s *Signer) txBuilder(opts ...TxOption) client.TxBuilder {
```

rootulp (dismissed):

pkg/user/signer.go
```diff
@@ -0,0 +1,375 @@
...
+	txClient := tx.NewServiceClient(s.grpc)
+
+	resp, err := txClient.GetTx(
+		ctx,
+		&tx.GetTxRequest{
+			Hash: txHash,
+		},
+	)
+	if err == nil {
+		return resp.TxResponse, err
```
staheri14, line 217: [nit] Just for improved readability: 
```suggestion
		return resp.TxResponse, nil
```

pkg/user/signer.go
```diff
@@ -0,0 +1,375 @@
...
+func (s *Signer) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxResponse, error) {
+	txClient := tx.NewServiceClient(s.grpc)
+
+	resp, err := txClient.GetTx(
+		ctx,
+		&tx.GetTxRequest{
+			Hash: txHash,
+		},
+	)
+	if err == nil {
```
staheri14, line 216: [Question] Is this equality check correct? or should it be an inequality check?

pkg/user/signer.go
```diff
@@ -0,0 +1,375 @@
...
+		case <-timer.C:
+			resp, err = txClient.GetTx(
+				ctx,
+				&tx.GetTxRequest{
+					Hash: txHash,
+				},
+			)
+
+			if err == nil {
+				return resp.TxResponse, err
```
staheri14, line 239: [nit] For improved readability
```suggestion
				return resp.TxResponse, nil
```

pkg/user/signer.go
```diff
@@ -0,0 +1,375 @@
...
+		},
+		Sequence: sequence,
+	}
+
+	err := builder.SetSignatures(draftsigV2)
+	if err != nil {
+		return fmt.Errorf("error setting draft signatures: %w", err)
+	}
+
+	// now we can use the data to produce the signature from each signer
```
staheri14, line 292:
```suggestion
	// now we can use the data to produce the signature from the signer
```

staheri14 (approved):
LGTM!

pkg/user/signer.go
cmwaters: One day if we change it we're going to break the entire stack lol

pkg/user/signer.go
cmwaters: Yes just because the error handling happens later on

rootulp (approved):

staheri14 (approved):

evan-forbes (approved):

Issue #2276, opened 2023-08-15
TestMaxTotalBlobSizeSuite fails test-race


rootulp:
## Problem

`TestMaxTotalBlobSizeSuite` fails test-race [here](https://github.com/celestiaorg/celestia-app/actions/runs/5865793655/job/15904306153?pr=2274#step:4:726)

## Proposal

1. Add the test to https://github.com/celestiaorg/celestia-app/blob/35c4b0e3ea82b19598827a450a94b1a678f1dd9b/Makefile#L161
2. All the tests that use testnode fail test-race. Tracked in https://github.com/celestiaorg/celestia-app/issues/1369

PR #2277, opened 2023-08-15
test: skip TestMaxTotalBlobSizeSuite during test-race

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2276

Note: doesn't need to be backported to v1.x because v1.x skips tests via `-short`
https://github.com/celestiaorg/celestia-app/blob/53b5dbf2438a085ded2952a0250829213eb77f55/Makefile#L114-L118


cmwaters (approved):
... and the list keeps growing 😢

SweeXordious (approved):

PR #2278, opened 2023-08-15
chore!: bump sdk fork to v1.17.0

cmwaters:
Bumps to the latest version of the SDK. Note that this deactivates the use of the EVM address. To fulfill that functionality we will need to reintroduce a mapping in https://github.com/celestiaorg/celestia-app/pull/2169

## Checklist

### DIFF
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -94,7 +94,8 @@ func TestValsetCreationWhenEditingEVMAddr(t *testing.T) {
 	qgb.EndBlocker(ctx, *pk)
 	ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 10)
 
-	assert.Equal(t, currentAttestationNonce+1, pk.GetLatestAttestationNonce(ctx))
+	// FIXME: this needs to change to 2 once we have a proper implementation of editing the EVM address
+	assert.Equal(t, currentAttestationNonce, pk.GetLatestAttestationNonce(ctx))
 }
 
 func TestSetValset(t *testing.T) {
--- a/x/qgb/abci_test.go
+++ b/x/qgb/abci_test.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/celestiaorg/celestia-app/x/qgb/types"
 
-	"github.com/cosmos/cosmos-sdk/x/staking/teststaking"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
 	testutil "github.com/celestiaorg/celestia-app/test/util"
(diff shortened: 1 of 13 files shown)


SweeXordious (commented):
[not blocking]
This is super cool, but as long as the mapping PR is not merged, the QGB will be broken, therefore main will be broken.
Does it make sense to postpone bumping until we have that PR reviewed and ready to be merge? so that we don't have a list of commits which all contain a broken QGB in between

cmwaters:
> [not blocking] This is super cool, but as long as the mapping PR is not merged, the QGB will be broken, therefore main will be broken. Does it make sense to postpone bumping until we have that PR reviewed and ready to be merge? so that we don't have a list of commits which all contain a broken QGB in between

I was hoping I could kind of do these as two separate commits. Maybe it makes sense to merge the other one first

x/qgb/abci_test.go
```diff
@@ -83,22 +82,20 @@ func TestValsetCreationWhenEditingEVMAddr(t *testing.T) {
...
 	)
-	_, err = msgServer.EditValidator(ctx, editMsg)
+	_, err := msgServer.EditValidator(ctx, editMsg)
 	require.NoError(t, err)
 	staking.EndBlocker(ctx, input.StakingKeeper)
 	qgb.EndBlocker(ctx, *pk)
 	ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 10)
 
-	assert.Equal(t, currentAttestationNonce+1, pk.GetLatestAttestationNonce(ctx))
+	// FIXME: this needs to change to 2 once we have a proper implementation of editing the EVM address
```
rootulp, line 97: [question] not sure what you mean by "2"

rootulp (approved):
I think this needs a make proto-gen

SweeXordious:
> I was hoping I could kind of do these as two separate commits

Yes, as separate commit, but better they are sequential. Like we merge this one and the next commit on main would be the mapping commit.
This would keep main not breaking the qgb as low as possible

x/qgb/abci_test.go
cmwaters: Ah it's currently 1 but we should be asserting that the latest nonce is 2

cmwaters: I also should have @-ed myself but forgot to

evan-forbes (approved):
this LGTM

fwiw I think its fine to merge to main and but then just not backport it yet to v1.x. This avoids using some temporaray feature branch, which is arguably just as bad as leaving main in a temporary unreleasable state (a property I'm not sure we need to preserve any more since we never want to cut releases directly from main)

rootulp (approved):

Issue #2262, opened 2023-08-14
Issue label automation fails


rootulp:
## Problem

The [issue label automation](https://github.com/celestiaorg/celestia-app/blob/main/.github/workflows/issue-label-automation.yml) fails [here](https://github.com/celestiaorg/celestia-app/actions/runs/5856952149/job/15877750700?pr=2260#step:2:10) for PRs opened by mergify bot

## Proposal

It may be possible to fix by adding `mergify[bot]` to [this line](https://github.com/celestiaorg/celestia-app/blob/cb8247ead807a52bb8915d450982544755efc1e6/.github/workflows/issue-label-automation.yml#L43)

rootulp:
Yea I introduced a regression

<img width="743" alt="Screenshot 2023-08-15 at 9 46 45 AM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/dd73c45e-1552-4cfb-85ec-e27f59811384">

[Source](https://docs.github.com/en/actions/learn-github-actions/expressions)

PR #2282, opened 2023-08-15
ci: fix issue label automation

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2262

Tested this on my fork and it works as expected. The problem with the last PR was that the environment variables are strings so they need to be compared against the string literals `'true'` and `'false'`.


SweeXordious (approved):
utACK

staheri14 (approved):

Issue #2070, opened 2023-07-10
Document the security assumptions for the time based inflation


SweeXordious:
The `x/mint` module uses a time based inflation schedule to determine the block provisions. At first glance, it seems that by delaying blocks, validators would be able to increase their allocation and receive more tokens. This is not the case due to the following:

- Light clients will halt if the headers are too far in the future, which would define an upper bound for how much deviation can be caused by validators voting in the future for light nodes to halt
- The annual allocations are not locked per year. Thus, the best a deviation can do is advance the inflation schedule. No extra tokens will be minted.
- Others

PR #2290, opened 2023-08-16
docs: add security section to x/mint README

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2070

### DIFF
--- a/x/mint/README.md
+++ b/x/mint/README.md
@@ -92,7 +92,7 @@ See [./test/mint_test.go](./test/mint_test.go) for an integration test suite for
 
 Source: <https://en.wikipedia.org/wiki/Year#Calendar_year>
 
-This module assumes `DaysPerYear = 365.2425`  so when modifying tests, developers must define durations based on this assumption because ordinary durations won't return the expected results. In other words:
+This module assumes `DaysPerYear = 365.2425` so when modifying tests, developers must define durations based on this assumption because ordinary durations won't return the expected results. In other words:
 
 ```go
 // oneYear is 31,556,952 seconds which will likely return expected results in tests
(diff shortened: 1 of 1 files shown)


x/mint/README.md
```diff
@@ -102,6 +102,18 @@ oneYear := time.Duration(minttypes.NanosecondsPerYear)
...
 
+### Security
+
+Q: Can validators manipulate the amount of tokens minted due to inflation?
+
+A:
+
+This x/mint module calculates block provisions based on block timestamps so it is important to understand how block timestamps work in CometBFT. In CometBFT, block timestamps are monotonically increasing. A block's timestamp is the median of the `Vote.Time` fields from vote messages, where each vote's timestamp is weighted based on the voting power of the validator that cast it.
+
+Based on [BFT time](https://docs.cometbft.com/v0.34/spec/consensus/bft-time), a block's timestamp is manipulatable by malicious validators if they control > 1/3 of the total voting power. Consequently, if malicious validators control > 1/3 of the total voting power, they could manipulate a block's timestamp to some arbitrary value (e.g. one year in the future) effectively advancing the inflation schedule by a year.
```
SweeXordious, line 113: > a block's timestamp is manipulatable by malicious validators if they control > 1/3 of the total voting power

[Question]
Why `>1/3`? Any validator can deviate the time for a certain portion except that it depends on their weight. If their weight is small, then the deviation would be small. If the weight is big, the deviation will be big.

evan-forbes (approved):
nice

x/mint/README.md
SweeXordious: Aah I see, thanks. But I still don't see how the 1/3 works:
> if malicious validators control > 1/3 of the total voting power, they could manipulate a block's timestamp to some arbitrary value (e.g. one year in the future)

I did a small test in here:
https://go.dev/play/p/bCku4cxDfu6

cmwaters: The proposer of the next height chooses the signatures to add to the commit but they don't theoretically need to add all the signatures they see. They only need to have 2/3+ signatures. So imagine there's 100 voting power in total and 34 is malicious. I as the proposer of the next height include a last commit that contains the 34 malicious voting power and the 33 honest voting power (to get to 67). 34 > 33 so the weighted median is whatever the 34 malicious nodes set which could be 10 years in the future for example

cmwaters (approved):
Helpful explanation

x/mint/README.md
SweeXordious: Aaah I see, this was the missing point for me:
>  the next height chooses the signatures to add to the commit

Thanks

Issue #2197, opened 2023-07-31
Loadtest mocha-3 with random blobs


evan-forbes:
We need to monitor mocha-3's performance with handling large amounts of blobs. This means that we should spam the mempool with blobs while monitoring various metrics.

- We should expect to see no to minor increases in block times since it can take multiple seconds to gossip large blocks, but we should not expect to see that it takes multiple rounds to reach consensus. It would be nice if we could know exactly how much longer it is taking to gossip the block.
- We should expect that bandwidth increases significantly, but it would be good to know exactly how much, and the exact distribution between the mempool and proposal gossiping.
- ~~We should expect that node's mempools will begin filtering out transactions with lower fees after hitting their local limit.~~
- ~~We should also look for memory leaks or other critical failures during this time.~~

edit: the two crossed issues will be tracked in a second set of issues dedicated to more formalized performance

it would be nice to combine this with https://github.com/celestiaorg/celestia-core/issues/1054

as a bonus: 
- We should increase the block size to 8MB and perform the same tests
- We should run the same tests with a large portion (if not all) of the validators running CAT

staheri14:
Created a [Notion page](https://www.notion.so/celestiaorg/Effect-of-spamming-on-network-performance-ca7d3100989b483cbdaff16670b2f216) exclusively aimed at compiling insights and discoveries pertaining to the performance evaluation of the network under a spamming attack. This encompasses the testing scenario, essential metrics, and performance indicators.

evan-forbes:
Here is a preliminary analysis

![time between cosecutive blocks and block size](https://github.com/celestiaorg/celestia-app/assets/42654277/23233227-1a4c-4176-8895-846a7b51fd74)

staheri14:
> We should expect that bandwidth increases significantly, but it would be good to know exactly how much, and the exact distribution between the mempool and proposal gossiping.

Link to the notion page where the second item of this issue is covered:

https://www.notion.so/celestiaorg/Analysis-of-Traffic-Distribution

evan-forbes:
I think we can close this as completed @staheri14 wdyt?

staheri14:
> I think we can close this as completed @staheri14 wdyt?

Certainly, we can proceed with closing it, as we have already obtained a variety of insightful test results concerning the network's health and have addressed all the questions in the issue.

PR #2292, opened 2023-08-17
feat: add the ability to specifiy the txsim master account

### LINKED ISSUE #2286: Add the ability to specify the master account in txsim
Currently, the only option is to use a master account with the largest funds. In general this has served as a good default, but this can be undesirable if we want to use a specific account with limited funds instead of a testnet account with more funds in the same wallet.

https://github.com/celestiaorg/celestia-app/blob/63a05a2245ad7a3520334b29084a0e8179730547/test/txsim/account.go#L76-L127

evan-forbes:
## Overview

This PR adds the ability to specify a specific account to use as master for txsim. This is useful for when we don't want to always use the account with the most funds in our wallet.

closes #2286
kinda sorta blocking #2197 

## Checklist

### DIFF
--- a/app/test/square_size_test.go
+++ b/app/test/square_size_test.go
@@ -136,6 +136,7 @@ func (s *SquareSizeIntegrationTest) fillBlocks(blobSize, blobsPerPFB, pfbsPerBlo
 		[]string{s.rpcAddr},
 		[]string{s.grpcAddr},
 		s.cctx.Keyring,
+		"",
 		rand.Int63(),
 		time.Second,
 		seqs...,
--- a/test/cmd/txsim/cli.go
+++ b/test/cmd/txsim/cli.go
@@ -22,22 +22,23 @@ import (
 
 // A set of environment variables that can be used instead of flags
 const (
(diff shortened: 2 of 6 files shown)


rootulp (approved):

SweeXordious (approved):

cmwaters (approved):
LGTM. We might be getting to the stage where we want to start using the options paradigm (or even having a config). `masterAccName`, `seed` and `pollTime` all have defaults and don't need to be set. Arguably `grpcEndpoints` and `rpcEndpoints` also have default values. This may make the constructor a bit more tidy and allow us to continue adding options in the future

evan-forbes:
> We might be getting to the stage where we want to start using the options paradigm (or even having a config).

yeah I was thinking the same thing. a config would be more readable to me than a bash script with 10 flags lol

cmwaters:
> > We might be getting to the stage where we want to start using the options paradigm (or even having a config).
> 
> yeah I was thinking the same thing. a config would be more readable to me than a bash script with 10 flags lol

I think the devops folks will want flags all the way

PR #2307, opened 2023-08-17
feat!: make submitPFB broadcast mode configurable

### LINKED ISSUE #1885: blob: SubmitPayForBlob should allow configurable broadcast mode
Or use the `sync` instead of `block`. 

As per @evan-forbes, `block` is going to be deprecated soonish, so we should stop using it. Additionally, it causes problems for the Dymension team

evan-forbes:
## Overview

closes #1885

## Checklist

### DIFF
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -28,6 +28,7 @@ import (
 	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
 	abci "github.com/tendermint/tendermint/abci/types"
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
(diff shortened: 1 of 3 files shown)


rootulp (approved):

Wondertan (commented):
[non-blocking]
Can this be an option with a sane default?

Wondertan (approved):

evan-forbes:
> Can this be an option with a sane default?

we could use options or something similar, but we are already accepting TxbuilderOptions so I didn't want to add two sets of options here. This function does have too many args tho, so its due for a refactor

Wondertan:
Ok, I was wondering if you can extend existing options.

app/test/integration_test.go
```diff
@@ -262,7 +263,14 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 			require.NoError(t, s.cctx.WaitForBlocks(3))
 
 			signer := blobtypes.NewKeyringSigner(s.cctx.Keyring, s.accounts[141], s.cctx.ChainID)
-			res, err := blob.SubmitPayForBlob(context.TODO(), signer, s.cctx.GRPCClient, []*blobtypes.Blob{tc.blob, tc.blob}, tc.opts...)
+			res, err := blob.SubmitPayForBlob(
+				context.TODO(),
+				signer,
+				s.cctx.GRPCClient,
+				sdktx.BroadcastMode_BROADCAST_MODE_BLOCK,
```
SweeXordious, line 270: Shouldn't this be broadcast mode sync? As per the issue?

evan-forbes: it was always block before, so I left it as is for now

cmwaters (approved):

SweeXordious (approved):

evan-forbes:
> Ok, I was wondering if you can extend existing options.

we could be it felt a bit too hacky and temporary. actually reducing the complexity and number of arguments I think requires a full refactor. Like, we'd have to wrap the txbuilder options in this more abstract option type, and apply the option two different times, once to see if there are tx builder options present when building the tx, then another time before we submit the tx. The alternative being to just add an extra argument, which felt like the least worst option for the time being

staheri14 (approved):
LGTM!

PR #2315, opened 2023-08-18
fix: InitUnlockedBalance unlocked from the vesting balance

### LINKED ISSUE #2295: InitUnlockedBalance will be unlocked from the vesting balance instead of being extra balance
### Summary

Currently, the behavior of the InitUnlockedBalance feature involves adding an extra unlocked balance to the locked accounts. However, there's a need to reconsider this approach and change it based on the [made decision](https://celestia-team.slack.com/archives/C04KS5R9VQU/p1692262592113819?thread_ts=1692205029.302239&cid=C04KS5R9VQU). 

### Problem Definition

To have a vesting account installed in the genesis file, the total allocation of that account is set to `OriginalVestingBalance` and the same amount is also set in the banking module.
I order to have a `total_allocation` locked and e.g. `1TIA` unlocked, the procedure is as follows:
`account.OriginalVestingBalance = total_allocation`
`account.BalanceInBankModule = total_allocation + 1TIA`
which means that account receives `1TIA` more, now the leadership decided instead of donating `1TIA` to each vesting account, we unlock `1TIA` from their locked allocation. Therefore, we need to subtract it from their total allocation. 
So:
`account.OriginalVestingBalance = total_allocation - 1TIA`
`account.BalanceInBankModule = total_allocation`

Same goal achieved without donating extra balance.

### Proposal

It is simple to do that.
Currently the functions that receive vesting account's details and generate a genesis version with its balances, have the following line of code to add extra unlocked amount:

```golang
balance = banktypes.Balance{
		Address: address,
		Coins:   initUnlockedCoins.Add(vestingBalance...),
	}
```

we need to change that code to the following:
```golang
balance = banktypes.Balance{
		Address: address,
		Coins:   vestingBalance,
	}
vestingBalance = vestingBalance.Sub(initUnlockedCoins...)
```

there are 3 functions that need to be modified:

1. NewGenesisDelayedVestingAccount
2. NewGenesisPeriodicVestingAccount
3. NewGenesisContinuousVestingAccount

### For Admin Use

- [ ] Not duplicate issue

mojtaba-esk:
closes #2295 
more context in the issue.
A short summary: To install a vesting account in the genesis file, allocate the desired amount to `OriginalVestingBalance` and the banking module. For instance, to lock a `total_allocation` and unlock `1TIA`, follow these steps:

    Set `account.OriginalVestingBalance = total_allocation`.
    Set `account.BalanceInBankModule = total_allocation + 1TIA`.

However, to avoid donating extra funds, a new approach is adopted: unlock `1TIA` from the vested amount. This is achieved by subtracting `1TIA` from the `OriginalVestingBalance`:

    Update `account.OriginalVestingBalance = total_allocation - 1TIA`.
    Keep `account.BalanceInBankModule = total_allocation`.

This achieves the same goal without additional donations.

## Checklist

### DIFF
--- a/pkg/genesis/vesting_accounts.go
+++ b/pkg/genesis/vesting_accounts.go
@@ -32,6 +32,8 @@ func NewGenesisRegularAccount(
 // NewGenesisDelayedVestingAccount creates a new DelayedVestingAccount with the
 // specified parameters. It returns the created account converted to genesis
 // account type and the account balance.
+// The final vesting balance is calculated by subtracting the initial unlocked coins
+// from the vesting balance.
 func NewGenesisDelayedVestingAccount(
 	address string,
 	vestingBalance,
@@ -55,6 +57,8 @@ func NewGenesisDelayedVestingAccount(
 	return authtypes.GenesisAccount(vAccount), balance, nil
 }
 
+// The final vesting balance is calculated by subtracting the initial unlocked coins
+// from the vesting balance.
 func NewGenesisPeriodicVestingAccount(
 	address string,
 	vestingBalance,
@@ -79,6 +83,8 @@ func NewGenesisPeriodicVestingAccount(
 	return authtypes.GenesisAccount(vAccount), balance, nil
 }
 
+// The final vesting balance is calculated by subtracting the initial unlocked coins
+// from the vesting balance.
 func NewGenesisContinuousVestingAccount(
 	address string,
 	vestingBalance,
--- a/pkg/genesis/vesting_accounts.go
+++ b/pkg/genesis/vesting_accounts.go
@@ -45,8 +45,9 @@ func NewGenesisDelayedVestingAccount(
 
 	balance = banktypes.Balance{
 		Address: address,
-		Coins:   initUnlockedCoins.Add(vestingBalance...),
+		Coins:   vestingBalance,
 	}
+	vestingBalance = vestingBalance.Sub(initUnlockedCoins...)
 
 	bAccount := authtypes.NewBaseAccountWithAddress(sdkAddr)
 	vAccount := vestingtypes.NewDelayedVestingAccount(bAccount, vestingBalance, endTime.Unix())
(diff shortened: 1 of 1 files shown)


pkg/genesis/vesting_accounts.go
```diff
@@ -45,8 +45,9 @@ func NewGenesisDelayedVestingAccount(
 
 	balance = banktypes.Balance{
 		Address: address,
-		Coins:   initUnlockedCoins.Add(vestingBalance...),
+		Coins:   vestingBalance,
 	}
+	vestingBalance = vestingBalance.Sub(initUnlockedCoins...)
```
staheri14, line 50: [Suggestion, optional] `Sub` may cause panic if it results in a negative value; it could be worth mentioning the possibility of panic in the `NewGenesisDelayedVestingAccount` description. Ditto other modified functions in this PR.

pkg/genesis/vesting_accounts.go
```diff
@@ -68,8 +69,9 @@ func NewGenesisPeriodicVestingAccount(
...
```
staheri14, line 74: [Suggestion] Since this change modifies how the function worked before, it's a good idea to note in the function description that the vesting account balance is now calculated by subtracting `initUnlockedCoins` from `vestingBalance`.

staheri14 (commented):
Looks good to me! I've included a few suggestions. Defer it to other reviewers for final approval.

cmwaters (approved):

evan-forbes (approved):
doesn't have to be done in this PR, but given its importance can we make sure that the total expected balance is tested for @mojtaba-esk? I believe we've discussed this before as invariants or having tx fail/pass depending on time and balance, but also for the perhaps we also need them for vesting module tests. Feel free to grab a slot on my calendar if we need to sync on this :slightly_smiling_face:

cmwaters (approved):

PR #2331, opened 2023-08-23
feat: add support for v2

cmwaters:
Adds v2 package. For now this is just the same as v1 but might change in the future. 

We will need at least the version number for a versioned state machine

### DIFF
--- a/pkg/appconsts/versioned_consts.go
+++ b/pkg/appconsts/versioned_consts.go
@@ -2,10 +2,11 @@ package appconsts
 
 import (
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
+	v2 "github.com/celestiaorg/celestia-app/pkg/appconsts/v2"
 )
 
 const (
-	LatestVersion = v1.Version
+	LatestVersion = v2.Version
 )
 
 // SubtreeRootThreshold works as a target upper bound for the number of subtree
--- a/pkg/appconsts/v2/app_consts.go
+++ b/pkg/appconsts/v2/app_consts.go
@@ -1,0 +1,7 @@
+package v2
+
+const (
+	Version              uint64 = 2
+	SquareSizeUpperBound int    = 128
+	SubtreeRootThreshold int    = 64
+)


pkg/appconsts/versioned_consts.go
```diff
@@ -2,10 +2,11 @@ package appconsts
 
 import (
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
+	v2 "github.com/celestiaorg/celestia-app/pkg/appconsts/v2"
 )
 
 const (
-	LatestVersion = v1.Version
+	LatestVersion = v2.Version
 )
```
rootulp, line 11: [not blocking][question] do we also want to add support for version 2 below like:

```go
func SubtreeRootThreshold(version uint64) int {
	switch version {
	case v1.Version:
		return v1.SubtreeRootThreshold
	case v2.Version:
		return v2.SubtreeRootThreshold
	default:
		panic(fmt.Sprintf("unsupported version: %v", version))
	}
}

func SquareSizeUpperBound(version uint64) int {
	switch version {
	case v1.Version:
		return v1.SquareSizeUpperBound
	case v2.Version:
		return v2.SquareSizeUpperBound
	default:
		panic(fmt.Sprintf("unsupported version: %v", version))
	}
}
```
```

File: pkg/appconsts/versioned_consts.go [hunk 2]
```diff
@@ -2,10 +2,11 @@ package appconsts
 
 import (
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
+	v2 "github.com/celestiaorg/celestia-app/pkg/appconsts/v2"
 )
 
 const (
-	LatestVersion = v1.Version
+	LatestVersion = v2.Version
```
rootulp, line 9: [blocking][question] if we merge this PR now, I'm concerned that the AppVersion that gets used everywhere will be `2`.

E.g. https://github.com/celestiaorg/celestia-app/blob/f36a031daee9e045b6df4f464872e4d3f8905998/app/default_overrides.go#L144

Is this desirable? Since this PR targets `main` and not the `v1.x` branch, I assume:
1. this PR won't ever be backported to `v1.x`
2. `main` is effectively going to be the development branch for the next major release (i.e. `v2.x`)

pkg/appconsts/versioned_consts.go
evan-forbes: noooooooooo noooo panics lol

pkg/appconsts/versioned_consts.go
evan-forbes: did you have another workflow in mind? or do you think we need to have the development go in a different branch?

my understanding of this change is that if we add a consensus breaking change we can add a quick switch statement to v2 so its easier to remember stuff

evan-forbes (approved):

pkg/appconsts/versioned_consts.go
cmwaters: Yes, this release flow is the most common one I've seen. `main` is the development branch (I've seen people also use `dev` to be more explicit) and users are only supposed to build against tagged releases or RCs. This PR of course won't ever be in v1.x and helps us to prepare v2.x which will be branched from `main`.

pkg/appconsts/versioned_consts.go
cmwaters: I had a discussion with @evan-forbes where we decided intentionally not to panic for unknown versions. 

I did think about adding the v2 values as the default but since they are currently the same, I thought to leave it as is although we should probably still make this change at some point even if to just be explicit about it

pkg/appconsts/versioned_consts.go
cmwaters: It looks like I didn't refresh my page when posting my comments so I didn't see what @evan-forbes had written

rootulp: I don't have another workflow in mind. I wanted to clarify the intended purpose of each branch b/c until this PR merges, `main` has been the primary development branch even for `v1.x` because we backport PRs from `main` to `v1.x`. 

After this PR merges, if a contributor has a hotfix they need to apply to `v1.x`, should they target `main` and backport to `v1.x` or target `v1.x` and forward port to `main`?

rootulp (approved):

pkg/appconsts/versioned_consts.go
cmwaters: @rootulp the problem I think you're trying to get at is if `main` diverges drastically from `v1.x` then it might be difficult to cherry-pick commits across and thus more work for the developer. This is true and becomes more problematic the longer our release cycles are so we should be wary of that.

I'm indifferent between forward porting and back porting. Forward porting means that you can actually cut the release out quicker (if timing is important) else the default flow I'm used to is to target `main` first and then backport to the relevant release branches

Issue #2321, opened 2023-08-21
Consider upgrading to Go 1.21


rootulp:
## Context

- [Go 1.21](https://tip.golang.org/doc/go1.21) was released 
- celestia-node is bumping to Go 1.21 in https://github.com/celestiaorg/celestia-node/pull/2589. Historically there was an effort to keep them on similar Go versions so that users don't need to download multiple Go versions when following docs.

## Proposal

Consider upgrading to Go 1.21. When we tackle this issue, we should bump the Go version in go.mod, README, and CI workflows (i.e. golangci-lint).

rootulp:
The Dockerfiles already use Go 1.21

https://github.com/celestiaorg/celestia-app/blob/c4886d3e16256278ea97fa6f22fb197e3de57bd1/docker/Dockerfile_txsim#L2

https://github.com/celestiaorg/celestia-app/blob/c4886d3e16256278ea97fa6f22fb197e3de57bd1/Dockerfile#L2

rootulp:
celestia-node bumped to Go 1.21 in https://github.com/celestiaorg/celestia-node/pull/2589. Since contributors may make PRs to both repos, it may be wise to bump celestia-app's Go version to 1.21 now so that contributors don't have to maintain multiple Go versions locally.

Issue #2332, opened 2023-08-23
Question on go mod `-compat` usage


rootulp:
## Context

The Makefile uses the `-compat` flag in 3 locations. Ex. https://github.com/celestiaorg/celestia-app/blob/c4886d3e16256278ea97fa6f22fb197e3de57bd1/Makefile#L66

Docs:
```
go help mod tidy

The -compat flag preserves any additional checksums needed for the
'go' command from the indicated major Go release to successfully load
the module graph, and causes tidy to error out if that version of the
'go' command would load any imported package from a different module
version. By default, tidy acts as if the -compat flag were set to the
version prior to the one indicated by the 'go' directive in the go.mod
file.
```

## Question

It's not immediately clear why the Makefile uses `-compat 1.20`. The version specified in go.mod is Go 1.20 so the preceeding version is Go 1.19. Why does the Makefile explicitly use `-compat 1.20` when the default would be Go 1.19? If the `-compat` flag isn't needed, proposal to remove it.

PR #2333, opened 2023-08-23
chore: upgrade to Go 1.21

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2321

Separately we can consider dropping the `-compat` flag usage. See https://github.com/celestiaorg/celestia-app/issues/2332 b/c I don't think it's necessary.


evan-forbes:
do we need to wait for celestia-node to be able to upgrade as well?

cmwaters (approved):

cmwaters:
`main` on celestia-node is on 1.21 (https://github.com/celestiaorg/celestia-node/blob/440e232999331adf1d1dc034e5ceb8b14e9db5c7/go.mod#L3). Their latest RC is on v1.20

rootulp:
Related: https://github.com/celestiaorg/celestia-app/issues/2321#issuecomment-1690519013

> in my experience, sometimes the go compiler will work fine if a dependency uses a higher version of go, but other times it won't

IDK how to explain that. In my experience, it should be safe for an app to use Go 1.20 and depend on a lib that uses Go 1.21.

Wondertan:
go mod tidy forced us to go1.21. I wanted to keep it 1.20 and just update deps supporting it, but that didn't work

evan-forbes (approved):

PR #2339, opened 2023-08-24
fix: patch default fee checker priority calculation

### LINKED ISSUE #2354: Rounding bug when calculating priority
> is there somewhere I can learn about the arabica issue and what this is supposed to fix?

The problem was that any gas price that was a decimal was getting rounded down to a priority of 0. So if you're submitting a tx and you double the fee you were still getting the same priority as the other transactions and hence weren't making it in a block. The only way to be prioritized above was to pay a fee 10x the amount in which case you would get a priority of 1. So basically a float-ints rounding problem

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/issues/2339#issuecomment-1691726050_

evan-forbes:
## Overview

This PR patches the default priority calculation and adds it to our antehandler. This approach copy paste forks [the code from the sdk](https://github.com/celestiaorg/cosmos-sdk/blob/64adc6e40ae679003d2f779387e06c4105c99121/x/auth/ante/validator_tx_fee.go#L1-L65) which is not exported. This avoids us from having to include a change in the sdk, add more unit tests, more easily change this in the future, and be more explicit over how priority is determined in the mempool.

This should fix the issue that we're seeing on arabica, however I'm still writing an integration test to triple check things.

closes #2354 

## Checklist


app/ante/fee_checker.go
```diff
@@ -0,0 +1,65 @@
...
+
+// getTxPriority returns a naive tx priority based on the amount of the smallest denomination of the gas price
+// provided in a transaction.
+// NOTE: This implementation should be used with a great consideration as it opens potential attack vectors
+// where txs with multiple coins could not be prioritize as expected.
+func getTxPriority(fee sdk.Coins, gas int64) int64 {
+	var priority int64
+	for _, c := range fee {
+		p := int64(math.MaxInt64)
+		gasPrice := c.Amount.Mul(sdk.NewInt(1000000)).QuoRaw(gas)
```
evan-forbes, line 55: > gasPrice := c.Amount.Mul(sdk.NewInt(1000000)).QuoRaw(gas)

note this is the patch relative to upstream

https://github.com/celestiaorg/cosmos-sdk/blob/64adc6e40ae679003d2f779387e06c4105c99121/x/auth/ante/validator_tx_fee.go#L55

app/ante/fee_checker_test.go
```diff
@@ -0,0 +1,49 @@
...
+			fee:         sdk.NewCoins(sdk.NewInt64Coin("utia", 2)),
+			gas:         1,
+			expectedPri: 2000000,
+		},
+		{
+			name:        "1_000_000 TIA fee normal gas tx",
+			fee:         sdk.NewCoins(sdk.NewInt64Coin("utia", 1_000_000_000_000)),
+			gas:         75000,
+			expectedPri: 13333333333333,
+		},
```
cmwaters, line 40: All of these have a gasPrice of 1 or more. Can we add a test with a gas price (i.e. fee/gas or 0.1). Maybe gas is 1 million and fee is 1000utia

cmwaters (commented):
If we wanted to clean this up we should separate `checkTxFeeWithValidatorMinGasPrices` to a separate function that actually calculates the priority. (it's not evident that comparing the min gas price of the local validator is where the priority is calculated). But maybe we can do this in a separate PR

app/ante/fee_checker.go
evan-forbes: yeah good idea https://github.com/celestiaorg/celestia-app/pull/2339/commits/57d5aabe09975f1eeb78ab5c0337e0254e56fde0

app/ante/fee_checker_test.go
evan-forbes: good point added

cmwaters (approved):

app/ante/fee_checker.go
```diff
@@ -0,0 +1,68 @@
...
+	sdkerror "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+const (
+	// priorityScalingFactor is a scaling factor to convert the gas price to a priority.
+	priorityScalingFactor = 1_000_000
+)
+
+// checkTxFeeWithValidatorMinGasPrices implements the default fee logic, where the minimum price per
+// unit of gas is fixed and set by each validator, can the tx priority is computed from the gas price.
```
SweeXordious, line 15:
```suggestion
// unit of gas is fixed and set by each validator, where the tx priority is computed from the gas price.
```
maybe?

SweeXordious (commented):
is there somewhere I can learn about the arabica issue and what this is supposed to fix?

cmwaters:
> is there somewhere I can learn about the arabica issue and what this is supposed to fix?

The problem was that any gas price that was a decimal was getting rounded down to a priority of 0. So if you're submitting a tx and you double the fee you were still getting the same priority as the other transactions and hence weren't making it in a block. The only way to be prioritized above was to pay a fee 10x the amount in which case you would get a priority of 1. So basically a float-ints rounding problem

app/ante/fee_checker.go
evan-forbes: can I fix this is a follow @SweeXordious so we can merge this today and begin backporting where needed?

rootulp (approved):
TBH I don't fully follow the logic for multiple fee coins but the test cases for a single coin LGTM

evan-forbes:
created https://github.com/celestiaorg/celestia-app/issues/2353 as a follow up

Issue #2364, opened 2023-08-25
test workflow doesn't run on v1.x


rootulp:
## Context

A test is failing on the v1.x branch: https://github.com/celestiaorg/celestia-app/issues/2363 which made me wonder how this passed CI. It looks like the test workflow isn't running on the v1.x branch

## Problem

The test workflow isn't run on the v1.x branch. See https://github.com/celestiaorg/celestia-app/actions/workflows/test.yml?query=branch%3Av1.x

<img width="1711" alt="Screenshot 2023-08-25 at 2 44 57 PM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/5e9619cb-ee04-4c7f-bccc-4a50c6f33bfd">

rootulp:
It looks like only the docker-security-build workflows run on v1.x

<img width="900" alt="Screenshot 2023-08-25 at 2 46 30 PM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/0cc74caa-becf-4f59-a4b7-5a6e662271b9">

We should also run the test and lint workflows on v1.x

rootulp:
I'm quite confused b/c ci-release is supposed to run on all PRs

https://github.com/celestiaorg/celestia-app/blob/c90e61d5a2d0c0bd0e123df4ab416f6f0d141b7f/.github/workflows/ci-release.yml#L14

but there are no records of it running on v1.x branch https://github.com/celestiaorg/celestia-app/actions/workflows/ci-release.yml?query=branch%3Av1.x

rootulp:
Oh maybe the branch dropdown is the source branch not the target branch.

rootulp:
Ahh I think I have an explanation. The view on https://github.com/celestiaorg/celestia-app/commits/v1.x only shows that 3 workflows are run on the commits after they have been applied to the v1.x branch

The CI checks are actually running on the PRs prior to being merged to the v1.x branch. IMO we should have them run on the PR and on the v1.x branch so that it's easy to quickly identify when a commit was introduced that started failing a particular workflow (especially relevant b/c not all workflows are required prior to merge see https://github.com/celestiaorg/celestia-app/issues/2305).

PR #2365, opened 2023-08-25
ci: run ci-release workflow on release branches

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2364

Note this needs to be backported to v1.x before it will take effect.

## Testing

Tested on my fork. Before this PR, only docker-security-publish workflows run on commits after they merge to the v1.x branch. After this PR, the ci-release workflows also run.

Before | After
--- | ---
Screenshot 2023-08-25 at 4 00 00 PM | Screenshot 2023-08-25 at 4 00 06 PM


.github/workflows/ci-release.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - v1.x
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
```
evan-forbes, line 11: could we generalize this branch to be any versioned branch like the tags below?

evan-forbes (commented):
:+1:

.github/workflows/ci-release.yml
evan-forbes: defo not blocking

cmwaters (approved):

.github/workflows/ci-release.yml
```diff
@@ -1,11 +1,12 @@
...
 
-# Trigger on push events to main (i.e. merges), pushing new semantic version
-# tags, all PRs, and manual triggers
+# Run this workflow on push events (i.e. PR merge) to main or release branches,
+# push events for new semantic version tags, all PRs, and manual triggers.
 on:
   push:
     branches:
       - main
+      - "v[0-9]+.x"
```
evan-forbes, line 9: maybe also
```suggestion
      - "v[0-9]+.x"
      - "v[0-9]+.[0-9]+.x"
```
stupid formatting question, does 0-9 just mean any number, or literally 0 through 9?

evan-forbes (approved):

.github/workflows/ci-release.yml
rootulp: - `[0-9]` means any digit
- `+` means 1 or more

hmm I think we only need `"v[0-9]+.[0-9]+.x"` if we intend on maintaining release branches for minor versions. In other words:

- `v1.x`
- `v1.0.x`
- `v1.1.x`
- `v1.2.x`

but I thought we were only planning on maintaining release branches per major version

Issue #2297, opened 2023-08-17
Blob Streaming


cmwaters:
### Summary

Redesign parts of the system to support blob streaming, a service where clients can continually append blobs to the same PFB avoiding the high fixed costs and the need to time transaction submission which would lead to systemic bandwidth spikes. 

### Problem Definition

Celestia aims to produce blocks every 15 seconds. Rollups (or shared sequencers) generally want their transactions to be published by Celestia as soon as possible. They will likely frame transactions into batches at a greater frequency than Celestia does (also as execution nodes can optimistically execute state).

What will likely end up happening is that sequencers will hold out aggregating as many transactions as possible until some trigger where they let go and fire it into the network. More specifically if I witness a block being published, I know that in about 10 seconds I need to submit my `BlobTx` and hope that in the roughly 1 second before the proposer enters the next height, that the network propagates the `BlobTx` to their mempool. What we will see is continual spikes in bandwidth usage as the network suddenly comes under intense load. TCP will throttle and some rollups will miss the bus and their transactions won't land in time, unnecessarily delaying them to the next height. For optimal performance we want bandwidth to be as flat and constant as possible. 

### Proposal

Rather than having the blob module work on a per PFB level, users can stream on a per blob level, the blobs get continually appended to the reference PFB and when the proposer pulls the `BlobTx` from the mempool aggregates as many of the blobs as possible. This is great as sequencers can focus on other framing parameters (i.e. shorter times; making sure the blob occupies the entirety of the shares; the amount of shares the blob occupies is a power of 2 (for smaller proofs)) and it means the network load remains relatively constant (well at least in proportion to the aggregated load of the rollups). 

The change isn't so easy as might be initially thought. We need to modify the CAT pool to tag content on a per blob basis rather than a per blobTx basis (and maybe using the `ShareCommitment` rather than the simple sha256. The other factor is handling the signatures. Every appended blob needs to be accompanied with a new signature which is over the new entire set of blobs (aka the new PFB). The tricky thing is that the cut off is now decided by the proposer not the user. The proposer may choose to include 3 blobs or 5 depending on the network conditions. This may lead to orphaned or "dangling" blobs that don't have an associated PFB. The client must thus read the published block (which it's doing already), update the nonce and send out the new PFB associated with the blob. The validator then reattaches it in their mempool and the client can continue streaming blobs to the new PFB.

cc @nashqueue @renaynay @Wondertan 

### For Admin Use

- [ ] Not duplicate issue
[...]

PR #2371, opened 2023-08-28
doc: ADR on reliable transaction handling

cmwaters:


docs/architecture/adr-022-transaction-handling.md
```diff
@@ -0,0 +1,50 @@
...
+
+- 2023/08/28: Initial draft
+
+## Status
+
+Proposed
+
+## Context
+
+Nonces are a crucial component in transactions for replay protection; preventing users from having their transaction executed multiple times. Each account has a monotonically increasing number associated with it: the nonce (also referred as the `sequence` in the Cosmos SDK). If the last executed nonce is `n`, then the next transaction signed by that account must have a nonce of `n + 1`. This, however, comes with a few downsides. If a user submits 3 transactions in a row with nonces 6, 7, 8 and while propogating throughout the network they are received by the next proposer as 8, 7, 6. Transactions with nonces 7 and 8 will be rejected and only 6 will make it into the mempool to be proposed in the next block. Furthermore, a local hash-based cache may prevent 7 or 8 from every making it into the mempool. Secondly, if the transaction with nonce 6 happened to have too low priority it would block all other transactions. A user would not be able to overwrite the transaction by submitting the same one yet with a higher fee.
```
rootulp, line 13: [typo] every => ever

```suggestion
Nonces are a crucial component in transactions for replay protection; preventing users from having their transaction executed multiple times. Each account has a monotonically increasing number associated with it: the nonce (also referred as the `sequence` in the Cosmos SDK). If the last executed nonce is `n`, then the next transaction signed by that account must have a nonce of `n + 1`. This, however, comes with a few downsides. If a user submits 3 transactions in a row with nonces 6, 7, 8 and while propogating throughout the network they are received by the next proposer as 8, 7, 6. Transactions with nonces 7 and 8 will be rejected and only 6 will make it into the mempool to be proposed in the next block. Furthermore, a local hash-based cache may prevent 7 or 8 from ever making it into the mempool. Secondly, if the transaction with nonce 6 happened to have too low priority it would block all other transactions. A user would not be able to overwrite the transaction by submitting the same one yet with a higher fee.
```

[question]
> Secondly, if the transaction with nonce 6 happened to have too low priority it would block all other transactions. A user would not be able to overwrite the transaction by submitting the same one yet with a higher fee.

- "too low priority" relative to what? 
- "all other transactions" sent by this user?
- if the user submits another transaction with a higher fee and nonce 6, why doesn't it replace the previous tx with a low fee and nonce 6?

docs/architecture/adr-022-transaction-handling.md
```diff
@@ -0,0 +1,50 @@
...
+
+`CheckTx` will thus also no longer use `NewIncrementSequenceDecorator` on its branch of state.
+
+The second part of the solution is to introduce an ordering mechanism in `PrepareProposal` which will take the priority ordered list of transactions and shuffle any transactions which have a greater nonce than the last executed one.
+
+Specifically, it will loop through all transactions. Any transaction that has a a nonce below the last executed for that account will be dropped. If the transaction has a nonce more than one greater than the last executed it will be put in a temporary map (`map[string]map[uint64]sdk.Tx`). If the tranasction with the correct nonce appears later in the list, the protocol will check the map, and append the saved transaction directly after it (we already know it has a greater priority).
```
rootulp, line 23: Is `map[string]map[uint64]sdk.Tx`: address => nonce => tx?

docs/architecture/adr-022-transaction-handling.md
```diff
@@ -0,0 +1,50 @@
...
+
+This change is not consensus/state machine breaking as it does not modify the behavior of `DeliverTx` or `ProcessProposal` (methods that must be deterministic).  It can be rolled out as soon as v1.1.0
+
+## Alternative Approaches
+
+An alternative approach would be to use nonce lanes. This works by having multiple sequence numbers that a user can choose to increment any one of. The problem with this is that it heavily breaks the current account system. The `AccountI` interface defines a `SetSequence` and `GetSequence` which use single `uint64` values (as opposed to a key value style). It would also require a state migration.
+
+## Consequences
+
+> This section describes the consequences, after applying the decision. All consequences should be summarized here, not just the "positive" ones.
```
rootulp, line 35:
```suggestion
```

rootulp (approved):

docs/architecture/adr-022-transaction-handling.md
cmwaters: > "too low priority" relative to what? 

relative to all other transactions that can make it in a block. 

> if the user submits another transaction with a higher fee and nonce 6, why doesn't it replace the previous tx with a low fee and nonce 6?

CheckTx will reject that transaction because it previous incremented the nonce from the previous transaction (from 5 to 6) so now it expects a nonce of 7

docs/architecture/adr-022-transaction-handling.md
cmwaters: yes do you want me to add that?

rootulp: [typo] transaction
[not blocking] I think it would help to clarify.so something like:

```suggestion
Specifically, it will loop through all transactions. Any transaction that has a a nonce below the last executed for that account will be dropped. If the transaction has a nonce more than one greater than the last executed it will be put in a temporary map (`map[string]map[uint64]sdk.Tx` where the map represents address => nonce => tx). If the transaction with the correct nonce appears later in the list, the protocol will check the map, and append the saved transaction directly after it (we already know it has a greater priority).
```

rootulp (approved):

evan-forbes (commented):
This is an elegant proposal! :clap: I like it better than nonce lanes, as nonce lanes never really solved our problem as I understand it, since presumably sequencers want txs to be included in order. Presumably, they don't want validators to be able exclude the first of their unconfirmed txs on purpose or accidentally. If they don't care, then nonce lanes or even skipping nonces could be viable.

AFAIU (could be wrong) this does have a issue not mentioned _yet_ which is spam. 

> 1) `CheckTx` and `ReCheckTx` will simply check that the provided sequence number is greater than the last executed one.

Malicious accounts could submit a transaction that could never be included in a block despite otherwise appearing valid by simply never submitting the correct nonce. The malicious actor could effectively censor the chain by submitting many high fee transactions that appear to be valid but could never be included in a block.

In order for this solution to work as intended, we must find a way for each mempool to only gossip transactions that are at least feasible to be included in a block, and require their peers to do the same.

My current favorite solution involves a few simple helpers that make the existing rules more viable, although there could be a different solution to this problem that would allow for the proposal to get past the spam issue.

### Potential Solution

- Restore the normal check for nonce during `CheckTx` since we never want to gossip a transaction in which we don’t have all of the sequences for.
- When pruning a transaction, we must check if we have a transaction from the same user but with a higher nonce. If so, prune that transaction first, even if it has a higher fee.
- Essentially require CAT
- When gossipping a `SeenTx` with CAT, also include the hashes of the lower sequence transactions. This way a node can request any missing transactions and call `CheckTx` in order, or ignore the transactions entirely if it thinks the lower nonce txs if need be (too low of fee for a lower nonce tx its already seen)
- When reaping from the mempool (especially when the mempool has more than 2MB worth of txs) we need to reap the txs with lower nonce first


We will also probably want to make account queries (not all state queries I don’t think), be derived from the mempool state, and/or have better nonce management when signing transactions.

Lastly, presumably we would keep txs from the same user indexed in the CAT’s tx store indexed together somehow to make this process/logic easier.

cmwaters:
> Malicious accounts could submit a transaction that could never be included in a block despite otherwise appearing valid by simply never submitting the correct nonce.

Yeah I did think about this although the TTL would eventually kick the transaction but I think you're right that you could essentially censor the chain if you made several accounts.

What came to mind is what if we switch the tag of the transaction to being the account number/sequence (rather than just the hash). You post the latest sequence and account number you have and then peers will request transactions as account number and sequence (also you could ranges)

evan-forbes:
I'll also ref #2297 here as if we have a mempool that handles these things, then applying such an optimization on top seems very doable

evan-forbes (approved):

PR #2389, opened 2023-08-30
chore(blob): clarifies that hex values should not be 0x prefixed in the PayForBlobs subcommand

staheri14:
## Overview

Prefixing hex values with the `0x` prefix is a common convention. Nevertheless, the `blob` module CLI does not have the capability to parse the `0x` prefix. Hence, including this clarification in the command description can greatly assist users.

## Checklist

### DIFF
--- a/x/blob/client/cli/payforblob.go
+++ b/x/blob/client/cli/payforblob.go
@@ -36,10 +36,9 @@ func CmdPayForBlob() *cobra.Command {
 		Long: "Pay for a data blob to be published to the Celestia blockchain. " +
 			"[hexNamespaceID] must be a 10 byte hex encoded namespace ID, " +
 			"(for the namespace version 0). " +
-			"The hex string should not include a 0x prefix." +
-			"[hexBlob] can be an arbitrary length hex encoded data blob." +
-			"The hex string should not include a 0x prefix." +
-			" " +
+			"The hex string should not include a 0x prefix. " +
+			"[hexBlob] can be an arbitrary length hex encoded data blob. " +
+			"The hex string should not include a 0x prefix. " +
 			"This command only supports a single blob per invocation. ",
 		Args: cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/x/blob/client/cli/payforblob.go
+++ b/x/blob/client/cli/payforblob.go
@@ -36,9 +36,8 @@ func CmdPayForBlob() *cobra.Command {
 		Long: "Pay for a data blob to be published to the Celestia blockchain. " +
 			"[hexNamespaceID] must be a 10 byte hex encoded namespace ID, " +
 			"(for the namespace version 0). " +
-			"The hex string should not include a 0x prefix. " +
 			"[hexBlob] can be an arbitrary length hex encoded data blob. " +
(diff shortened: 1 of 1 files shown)


x/blob/client/cli/payforblob.go
```diff
@@ -34,8 +34,11 @@ func CmdPayForBlob() *cobra.Command {
 		Use:   "PayForBlobs [hexNamespaceID] [hexBlob]",
 		Short: "Pay for a data blob to be published to the Celestia blockchain",
 		Long: "Pay for a data blob to be published to the Celestia blockchain. " +
-			"[hexNamespaceID] must be a 10 byte hex encoded namespace ID. " +
+			"[hexNamespaceID] must be a 10 byte hex encoded namespace ID, " +
+			"(for the namespace version 0). " +
+			"The hex string should not include a 0x prefix. " +
 			"[hexBlob] can be an arbitrary length hex encoded data blob. " +
+			"The hex string should not include a 0x prefix. " +
 			"This command only supports a single blob per invocation. ",
```
rootulp, line 42: 1. [not blocking] We can add support for parsing the `0x` prefix if one is provided. This could be a follow up issue.
2. [optional] Instead of repeating "The hex string should not include a 0x prefix." twice, this could state it once like:

```suggestion
			"[hexNamespaceID] must be a 10 byte hex encoded namespace ID for the namespace version 0. " +
			"[hexBlob] can be an arbitrary length hex encoded data blob. " +
			"Hex strings should not include a 0x prefix. " +
			"This command only supports a single blob per invocation. ",
```

rootulp (dismissed):

rootulp:
Added backport label b/c seems like something we should backport.

rootulp:
FWIW if we're going to include a scope in the commit message, I don't think we need a separating space so proposal to do: `chore(blob): clarifies that hex...`

Ref: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope

x/blob/client/cli/payforblob.go
staheri14: > [not blocking] We can add support for parsing the 0x prefix if one is provided. This could be a follow up issue.

Yes, that would be an option too. 

> [optional] Instead of repeating "The hex string should not include a 0x prefix." twice, this could state it once like:

Applied your suggestion https://github.com/celestiaorg/celestia-app/pull/2389/commits/6654226db2f024eb3a432e2c199f016c774a88a6

rootulp (approved):

staheri14:
> FWIW if we're going to include a scope in the commit message, I don't think we need a separating space so proposal to do: `chore(blob): clarifies that hex...`
> 
> Ref: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope

Thanks for sharing this. I revised the commit message accordingly.

x/blob/client/cli/payforblob.go
staheri14: > 1. [not blocking] We can add support for parsing the `0x` prefix if one is provided. This could be a follow up issue.

Please see that tracking issue https://github.com/celestiaorg/celestia-app/issues/2392

evan-forbes (approved):

PR #2405, opened 2023-08-31
chore(x/mint)!: shorten keys

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2497 
Closes https://github.com/celestiaorg/celestia-app/issues/2498

### DIFF
--- a/x/mint/types/keys.go
+++ b/x/mint/types/keys.go
@@ -1,10 +1,10 @@
 package types
 
 // KeyMinter is the key to use for the Minter in the mint store.
-var KeyMinter = []byte("Minter")
+var KeyMinter = []byte{0x0}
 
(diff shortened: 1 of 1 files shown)


cmwaters (approved):

rootulp:
Note this PR is breaking and we haven't come to a decision on whether we include it in v1.0.0 or close it and never do it.

cmwaters:
It's breaking but I don't see much risk in making that change if we already plan on making a breaking change. At the same time, I don't think there's too much to gain

evan-forbes (commented):
I don't think we should merge this without either deciding that we want to hardfork the testnets and include this in v1, or add a migration script

rootulp:
+1 we can re-open if we want this for v2.

PR #2406, opened 2023-08-31
docs(x/mint): explain `TargetInflationRate`

rootulp:
Addresses finding 004 of binary builders preliminary audit report. 

I don't think it's necessary to rename the variable to `MinimumTargetInflationRate` but can do if reviewers feel strongly that renaming is preferred over this comment.

### DIFF
--- a/x/mint/types/constants.go
+++ b/x/mint/types/constants.go
@@ -16,7 +16,10 @@ const (
 
 	InitialInflationRate = 0.08
 	DisinflationRate     = 0.1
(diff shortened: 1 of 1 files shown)


cmwaters (approved):

staheri14:
> I don't think it's necessary to rename the variable to MinimumTargetInflationRate but can do if reviewers feel strongly that renaming is preferred over this comment.

I would lean towards renaming it to `MinimumTargetInflationRate`.

PR #2408, opened 2023-08-31
refactor(x/mint): error if current before previous

rootulp:
Addresses finding 008 of Binary Builders preliminary audit.

### DIFF
--- a/x/mint/types/minter_test.go
+++ b/x/mint/types/minter_test.go
@@ -176,10 +176,11 @@ func BenchmarkCalculateBlockProvision(b *testing.B) {
 	r1 := rand.New(s1)
 	minter.AnnualProvisions = sdk.NewDec(r1.Int63n(1000000))
 	current := time.Unix(r1.Int63n(1000000), 0)
-	previous := current.Add(time.Second * 15)
+	previous := current.Add(-time.Second * 15)
 
 	for n := 0; n < b.N; n++ {
-		minter.CalculateBlockProvision(current, previous)
+		_, err := minter.CalculateBlockProvision(current, previous)
+		require.NoError(b, err)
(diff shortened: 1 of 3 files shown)


SweeXordious (approved):

x/mint/types/minter_test.go
```diff
@@ -109,10 +110,22 @@ func TestCalculateBlockProvision(t *testing.T) {
...
 	}
 	for _, tc := range testCases {
 		minter.AnnualProvisions = tc.annualProvisions
-		got := minter.CalculateBlockProvision(tc.current, tc.previous)
+		got, err := minter.CalculateBlockProvision(tc.current, tc.previous)
+		if tc.wantErr {
+			assert.Error(t, err)
+			return
+		}
+		assert.NoError(t, err)
```
staheri14, line 128: [optional]
```suggestion
	assert.Equal(t, err!=nil, tc.wantErr)
```

staheri14 (approved):
LGTM! suggested one optional change.

x/mint/types/minter_test.go
rootulp: I think the order of arguments in this suggestion is incorrect because the second argument should be expected and the third argument should be actual.

rootulp: The code in this PR is consistent with other tests in this repo. I don't feel strongly about which approach is preferred but don't want to wait for another two approvals so if you feel strongly about the suggestion, please create a GH issue and we can tackle it repo-wide.

staheri14: My point was to suggest replacing the if statement with a one-line equality check, and you can, of course, switch the order of arguments.

PR #2409, opened 2023-08-31
refactor(x/mint)!: change encoding of genesis time in state

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2499


rootulp:
Closing as won't do b/c breaking and we don't want to include anymore breaking changes in v1.x

PR #2424, opened 2023-09-05
chore!: merge main to v1.x

evan-forbes:
## Overview

Merged main to see the diff and make sure we we're not missing anything consensus breaking (technically we weren't, as its not consensus breaking, but just avoids a bad bug #2423)

I don't know if we want to bother with keeping up with v1.x, or attempting to push through this change, as we might just want to start cutting releases off main since all commits should in theory be backwards compatible and this PR is essentially impossible to review


evan-forbes:
closing as this PR has served its purpose to demonstrate the diff between v1.x and main

Issue #2404, opened 2023-08-31
Modify single-node.sh to use default home directory


rootulp:
## Context

We have two scripts for development purposes:

1. https://github.com/celestiaorg/celestia-app/blob/main/scripts/single-node.sh
2. https://github.com/celestiaorg/celestia-app/blob/main/scripts/build-run-single-node.sh

The first script uses the already installed celestia-appd binary and a randomly generated home directory.
The second script builds a new celestia-appd binary and uses a randomly generated home directory.

## Problem

If a developer modifies something in-between invocations of single-node.sh (i.e. modify a config parameter), the second invocation of single-node.sh will generate a new home directory so modifications in the previous home directory won't affect the next invocation. 

## Proposal

Modify the first script to use the default home directory. 
- When a developer wants to modify a config, they use the first script.
- When a developer wants a clean slate, they use the second script.

PR #1583, opened 2023-03-31
feat: add a tmp dir to the single node script

mojtaba-esk:
This PR is a simple change to the single node script to use a randomly generated tmp directory instead of the default directory.

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #2431, opened 2023-09-05
chore: default home directory for `single-node.sh`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2404

I frequently need to modify a config and test it out. When using the existing single-node.sh script, it's a bit tedious to find the randomly generated CELESTIA_APP_HOME directory and plumb that into subsequent ad-hoc commands.

This PR modifies the `single-node.sh` script to use the default CELESTIA_APP_HOME directory (usually `~/.celestia-appd`).  That way, if a developer needs to make a modification to a config, they can do that in the expected location (i.e. `~/.celestia-appd/config/config.toml`) and then re-run `celestia-appd start` and test out the modification.

Other improvements include:
- Adds a y/n prompt to confirm deleting any existing CELESTIA_APP_HOME directory
- Hoists variables to the top
- Silences noisy terminal output by redirecting to dev/null

In addition to reviewing the code diff, it may make sense for reviewers to try the script and see if it behaves as expected.

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -1,45 +1,94 @@
 #!/bin/sh
 
-set -o errexit -o nounset
+# Stop script execution if an error is encountered
+set -o errexit
+# Stop script execution if an undefined variable is used
+set -o nounset
 
-HOME_DIR=$(mktemp -d -t "celestia_app_XXXXXXXXXXXXX")
+CHAIN_ID="private"
+KEY_NAME="validator"
+KEYRING_BACKEND="test"
+COINS="1000000000000000utia"
+DELEGATION_AMOUNT="5000000000utia"
+CELESTIA_APP_HOME="${HOME}/.celestia-appd"
+CELESTIA_APP_VERSION=$(celestia-appd version 2>&1)
 
-echo "Home directory: ${HOME_DIR}"
+echo "celestia-app home: ${CELESTIA_APP_HOME}"
+echo "celestia-app version: ${CELESTIA_APP_VERSION}"
+echo ""
 
-CHAINID="private"
+# Ask the user for confirmation before deleting the existing celestia-app home
+# directory.
+read -p "Are you sure you want to delete: $CELESTIA_APP_HOME? [y/n] " response
 
-# Build genesis file incl account for passed address
-coins="1000000000000000utia"
-celestia-appd init $CHAINID --chain-id $CHAINID --home ${HOME_DIR}
-celestia-appd keys add validator --keyring-backend="test" --home ${HOME_DIR}
-celestia-appd add-genesis-account $(celestia-appd keys show validator -a --keyring-backend="test" --home ${HOME_DIR}) $coins --home ${HOME_DIR}
-celestia-appd gentx validator 5000000000utia \
-	--keyring-backend="test" \
-	--chain-id $CHAINID \
-	--home ${HOME_DIR}
+# Check the user's response
+if [[ $response != "y" ]]; then
+    # Exit if the user did not respond with "y"
+    echo "You must delete $CELESTIA_APP_HOME to continue."
+    exit 1
+fi
 
-celestia-appd collect-gentxs --home ${HOME_DIR}
+echo "Deleting $CELESTIA_APP_HOME..."
+rm -r "$CELESTIA_APP_HOME"
+
+echo "Initializing validator and node config files..."
+celestia-appd init ${CHAIN_ID} \
+  --chain-id ${CHAIN_ID} \
+  --home ${CELESTIA_APP_HOME} \
+  &> /dev/null # Hide output to reduce terminal noise
+
+echo "Adding a new key to the keyring..."
+celestia-appd keys add ${KEY_NAME} \
+  --keyring-backend=${KEYRING_BACKEND} \
+  --home ${CELESTIA_APP_HOME} \
+  &> /dev/null # Hide output to reduce terminal noise
+
+echo "Adding genesis account..."
+celestia-appd add-genesis-account \
+  $(celestia-appd keys show ${KEY_NAME} -a --keyring-backend=${KEYRING_BACKEND} --home ${CELESTIA_APP_HOME}) \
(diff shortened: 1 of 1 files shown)


SweeXordious (commented):
I think this was using the default directory before, then it was changed to use  a random temp one. I don't remember the rationale behind tho, but probably @evan-forbes has an idea

rootulp:
To provide more context: @mojtaba-esk made it use a randomly generated one in https://github.com/celestiaorg/celestia-app/pull/1583

Since we have two scripts that use randomly generated home directories, proposal for one of them to not use a randomly generated one so that a developer can modify configs across runs. More context in https://github.com/celestiaorg/celestia-app/issues/2404

scripts/single-node.sh
```diff
@@ -1,45 +1,94 @@
...
 
-celestia-appd collect-gentxs --home ${HOME_DIR}
+echo "Deleting $CELESTIA_APP_HOME..."
+rm -r "$CELESTIA_APP_HOME"
+
+echo "Initializing validator and node config files..."
+celestia-appd init ${CHAIN_ID} \
+  --chain-id ${CHAIN_ID} \
+  --home ${CELESTIA_APP_HOME} \
+  &> /dev/null # Hide output to reduce terminal noise
```
SweeXordious, line 38: [nit]
Does it make sense to only reduce `stdout` to `/dev/null`, and keep `stderr` so that if an error occurs, the user can see the issue?

Something like: 

```suggestion
  1> /dev/null # Hide output to reduce terminal noise
```

SweeXordious (approved):
utAck

scripts/single-node.sh
rootulp: You're right, it does make a lot of sense to only redirect `stdout`. But for some reason, `celestia-appd init` outputs normal logs to `stderr` so if we only redirect `stdout`, then we see these logs in the terminal...

```
{"app_message":{"auth":{"accounts":[],"params":{"max_memo_characters":"256","sig_verify_cost_ed25519":"590","sig_verify_cost_secp256k1":"1000","tx_sig_limit":"7","tx_size_cost_per_byte":"10"}},"authz":{"authorization":[]},"bank":{"balances":[],"denom_metadata":[{"base":"utia","denom_units":[{"aliases":["microtia"],"denom":"utia","exponent":0},{"aliases":[],"denom":"TIA","exponent":6}],"description":"The native staking token of the Celestia network.","display":"TIA","name":"TIA","symbol":"TIA","uri":"","uri_hash":""}],"params":{"default_send_enabled":true,"send_enabled":[]},"supply":[]},"blob":{"params":{"gas_per_blob_byte":8,"gov_max_square_size":"64"}},"capability":{"index":"1","owners":[]},"crisis":{"constant_fee":{"amount":"1000","denom":"utia"}},"distribution":{"delegator_starting_infos":[],"delegator_withdraw_infos":[],"fee_pool":{"community_pool":[]},"outstanding_rewards":[],"params":{"base_proposer_reward":"0.010000000000000000","bonus_proposer_reward":"0.040000000000000000","community_tax":"0.020000000000000000","withdraw_addr_enabled":true},"previous_proposer":"","validator_accumulated_commissions":[],"validator_current_rewards":[],"validator_historical_rewards":[],"validator_slash_events":[]},"evidence":{"evidence":[]},"feegrant":{"allowances":[]},"genutil":{"gen_txs":[]},"gov":{"deposit_params":{"max_deposit_period":"172800s","min_deposit":[{"amount":"10000000","denom":"utia"}]},"deposits":[],"proposals":[],"starting_proposal_id":"1","tally_params":{"quorum":"0.334000000000000000","threshold":"0.500000000000000000","veto_threshold":"0.334000000000000000"},"votes":[],"voting_params":{"voting_period":"172800s"}},"ibc":{"channel_genesis":{"ack_sequences":[],"acknowledgements":[],"channels":[],"commitments":[],"next_channel_sequence":"0","receipts":[],"recv_sequences":[],"send_sequences":[]},"client_genesis":{"clients":[],"clients_consensus":[],"clients_metadata":[],"create_localhost":false,"next_client_sequence":"0","params":{"allowed_clients":["06-solomachine","07-tendermint"]}},"connection_genesis":{"client_connection_paths":[],"connections":[],"next_connection_sequence":"0","params":{"max_expected_time_per_block":"30000000000"}}},"mint":{"bond_denom":"utia"},"params":null,"qgb":{"params":{"data_commitment_window":"400"}},"slashing":{"missed_blocks":[],"params":{"downtime_jail_duration":"600s","min_signed_per_window":"0.500000000000000000","signed_blocks_window":"100","slash_fraction_double_sign":"0.050000000000000000","slash_fraction_downtime":"0.010000000000000000"},"signing_infos":[]},"staking":{"delegations":[],"exported":false,"last_total_power":"0","last_validator_powers":[],"params":{"bond_denom":"utia","historical_entries":10000,"max_entries":7,"max_validators":100,"min_commission_rate":"0.000000000000000000","unbonding_time":"1814400s"},"redelegations":[],"unbonding_delegations":[],"validators":[]},"transfer":{"denom_traces":[],"params":{"receive_enabled":true,"send_enabled":true},"port_id":"transfer"},"vesting":{}},"chain_id":"private","gentxs_dir":"","moniker":"private","node_id":"07004beac9d706bf742146cd21469f3081ce17b3"}
```

SweeXordious: is this the same for all commands? maybe we should check how we're initializing the outputs. I can investigate later of needed

evan-forbes (approved):
I just use adhoc testnode scripts, so I don't use these, so approving not to block

utACK, I'll defer to anyone else that uses this tho

staheri14 (approved):
A great set of modifications! 👏

scripts/single-node.sh
rootulp: Yes it's the same for all commands. I don't see a setting in this script that is redirecting stderr to stdout globally so maybe the celestia-appd binary itself emits all output to stderr instead of stdout?

PR #2449, opened 2023-09-07
Makefile: simpily ledger build tags

MSevey:
In the process of working on the `goreleaser` PR I discovered that how we are building the `BUILD_TAGS` is overly verbose and unnecessary. 

Ultimately we are just trying to add `ledger` to `-tags`.  Hardcoding it and adding a note in the README for the 2 distributions that aren't supported seemed easier to maintain.


Makefile
```diff
@@ -5,52 +5,14 @@ DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bu
...
-# convert build_tags from a whitespace seperated list to a comma seperated list
-build_tags_comma_sep := $(subst $(whitespace),$(comma),$(build_tags))
-
 
 # process linker flags
 ldflags = -X github.com/cosmos/cosmos-sdk/version.Name=celestia-app \
 		  -X github.com/cosmos/cosmos-sdk/version.AppName=celestia-appd \
 		  -X github.com/cosmos/cosmos-sdk/version.Version=$(VERSION) \
 		  -X github.com/cosmos/cosmos-sdk/version.Commit=$(COMMIT) \
-		  -X "github.com/cosmos/cosmos-sdk/version.BuildTags=$(build_tags_comma_sep)"
```
MSevey, line 50: Removing this line had zero impact on ledger it seemed. If someone knows why this line is needed, I'm happy to add it back in.

Makefile
```diff
@@ -5,52 +5,14 @@ DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bu
...
-ldflags += $(LDFLAGS)
```
MSevey, line 51: `LDFLAGS` isn't defined anywhere so this line is a no-op.

Makefile
rootulp: It appears to have been added in https://github.com/celestiaorg/celestia-app/pull/360 but `build_tags_comma_sep` wasn't defined in that PR so it seems like it was a no-op when first introduced.

rootulp (approved):
Thanks for the clean up!

[nit] PR title could use a conventional commit prefix so that the squashed commit has a conventional commit prefix so that the auto generated release notes are easy to scan and identify if the next release should be a major, minor, or patch release.

evan-forbes (approved):

Issue #2418, opened 2023-09-04
Consider adding evidence parameters to the paramfilter


evan-forbes:
We've noticed that there are potentially 3 values that are currently modifiable via token voting that we might not want to be changed. These are the two evidence paramters, `evidence.MaxAgeNumBlocks` and `evidence.MaxAgeDuration`, along with the mint module's `mint.BondDenom` (the denomination sent to the distribution module's account)

evan-forbes:
cc @adlerjohn and @liamsi 

simlar to our latest last minute change, while technically consensus breaking if no one successfully changes these values in governance we should be clear to update by simply upgrading

rootulp:
I'm still investigating if governance can modify any of these 3 params but it doesn't look like it.

- I think`mint.BondDenom` is a genesis config and isn't a governance modifiable param. The x/mint module doesn't appear to use the params module (b/c no param subspace is provided [here](https://github.com/celestiaorg/celestia-app/blob/3f1b99bf940b198e03819b800f8fc7637e1c54cb/app/app.go#L312-L319)).
- For the evidence parameters, [this file](https://github.com/celestiaorg/cosmos-sdk/blob/c525ad72ab1449d31115815f5832a68bc79d4ffd/x/evidence/types/params.go) is mostly empty. Cosmos SDK [docs](https://docs.cosmos.network/v0.46/modules/evidence/05_params.html) claim: "The evidence module does not contain any parameters."

rootulp:
The evidence parameters are a subset of the consensus parameters. https://github.com/cosmos/cosmos-sdk/issues/6197 and https://github.com/cosmos/cosmos-sdk/pull/5952 leads me to think that consensus parameters are modifiable through `abci.EndBlocker` but I don't think they're modifiable through the governance module. The behavior of modifying consensus params isn't documented in the [gov docs](https://docs.cosmos.network/v0.46/modules/gov/) and I couldn't find references to consensus params or evidence params in the x/gov module so they still seem non governance modifiable.

rootulp:
Update: we're not including this in the v1.x release of celestia-app. We do intend on including it in a subsequent breaking change.

rootulp:
I'm not sure why the OP calls out those 3 params. IMO all the params that are currently governance modifiable should be made governance unmodifiable. Instead we should determine the values for those params off-chain via rough consensus and then update them (if necessary) via hard-forks.

rootulp:
[...]

PR #2453, opened 2023-09-08
feat!: add evidence params to blocklist

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2418 by adding evidence params to the blocklist.


cmwaters (dismissed):
LGTM

x/paramfilter/test/param_filter_test.go
```diff
@@ -1,63 +1,120 @@
...
+		p := testProposal(invalidChange)
+		err = handler(ctx, p)
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "parameter can not be modified")
 
-		ps := app.StakingKeeper.GetParams(ctx)
-		require.Equal(t, ps.MaxValidators, uint32(1))
+		// Ensure that the evidence params still haven't been modified
+		require.Equal(t, defaults, *testApp.GetConsensusParams(ctx).Evidence)
+	})
```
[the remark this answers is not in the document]
evan-forbes: this is weird, as it seems like the evidence params are getting updated in endblock :thinking: 

https://github.com/celestiaorg/cosmos-sdk/blob/c525ad72ab1449d31115815f5832a68bc79d4ffd/baseapp/baseapp.go#L457-L488

x/paramfilter/test/param_filter_test.go
```diff
@@ -1,63 +1,119 @@
...
+			require.Equal(t, stakingtypes.DefaultMaxEntries, ps.MaxEntries)
+
+			newMaxEntries := 8
+			validChange := proposal.NewParamChange(stakingtypes.ModuleName, string(stakingtypes.KeyMaxEntries), fmt.Sprint(newMaxEntries))
+			invalidChange := proposal.NewParamChange(p[0], p[1], "value")
+
+			p := testProposal(validChange, invalidChange)
+			err := handler(ctx, p)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), "parameter can not be modified")
```
staheri14, line 113: If it could be converted to an error type, it would enhance code readability and make potential future refactoring easier.

x/paramfilter/test/param_filter_test.go
```diff
@@ -1,63 +1,119 @@
...
+	t.Run("test that a proposal with a blocked param is rejected", func(t *testing.T) {
+		for _, p := range testApp.BlockedParams() {
+			p := testProposal(proposal.NewParamChange(p[0], p[1], "value"))
+			err := handler(ctx, p)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), "parameter can not be modified")
+		}
+	})
+
+	t.Run("test that a proposal with another blocked param (validator consensus params) is rejected", func(t *testing.T) {
```
staheri14, line 42: [Question] Can you please elaborate how this test differs from the first test? Thanks

staheri14 (commented):
LGTM! added some suggestion and question.
I'd defer it to other reviewers for final approval.

x/paramfilter/test/param_filter_test.go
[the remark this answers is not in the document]
rootulp: In theory it should be possible to modify evidence params if I could encode the updated evidence params in the appropriate format (I think via the legacy Amino JSON encoder) but I couldn't get it to work with the permutations I tried in this test. Happy to convert to draft if you think it's important to have confidence that we could modify evidence params prior to adding the evidence params key to the blocklist.

Can convert this back to draft and try to figure out encoding / decoding if blocking.

x/paramfilter/test/param_filter_test.go
rootulp: The first test uses the first blocked param from the list of blocked params.

The second test (this test) uses a different blocked param. Specifically the validator consensus params.

x/paramfilter/test/param_filter_test.go
rootulp: hmm it's defined here: https://github.com/rootulp/celestia-app/blob/90b761bb7c2918e0ca75fc488801cd9b914431c5/x/paramfilter/types.go#L15

idk if it can be "converted to an error type" b/c 

```
// Register returns an error instance that should be used as the base for
// creating error instances during runtime.
```

but I can replace the error assertions with this diff if you prefer:

```diff
-			require.Contains(t, err.Error(), "parameter can not be modified")
+			require.ErrorIs(t, err, paramfilter.ErrBlockedParameter)
```

rootulp: Attempted to address via https://github.com/celestiaorg/celestia-app/pull/2453/commits/4c452c5f4dc9de1225f91ea2618c6fca79297f44. Please unresolve if that doesn't address your concern.

x/paramfilter/test/param_filter_test.go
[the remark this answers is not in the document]
evan-forbes: nah, we have confidence in the paramfilter as is, so just adding it should be sufficient

evan-forbes (approved):
before we merge imo we should either:

1) decide to backport this and attempt to get validators on the testnets to update
2) write an if statement to make this compatible with v2

I'm good with either, so I'm approving! will defer to the author and other reviewers. If we go with option 1 we should expidite the next release candidate and mocha-4 upgraded asap

rootulp:
Agreed.

> write an if statement to make this compatible with v2

I'll attempt to do this now b/c I don't have confidence in my ability to coordinate all mocha-4 validators to upgrade to a new release and since mocha-4 is a live network, someone could submit a param change proposal with evidence params at any moment prior to all validators upgrading to a release with this PR. In that scenario, I expect a fork would occur.

evan-forbes:
> I expect a fork would occur.

we'd have to hardfork to fix it yeah

x/paramfilter/test/param_filter_test.go
staheri14: > The first test uses the first blocked param from the list of blocked params.

In the first test, it looks like it ranges through all of the blocked params, doesn't it? 
```go
                for _, p := range testApp.BlockedParams() {
			p := testProposal(proposal.NewParamChange(p[0], p[1], "value"))
			err := handler(ctx, p)
			require.Error(t, err)
			require.Contains(t, err.Error(), "parameter can not be modified")
		}
```

rootulp: oh true! I can remove the second test case b/c it should be covered by test case one.

rootulp: Resolved by f0a9951bd65a3e9ba6bbc265ab1bd3d255a5e212

x/paramfilter/test/param_filter_test.go
```diff
@@ -1,63 +1,99 @@
...
+	t.Run("test that a proposal with a blocked param is rejected", func(t *testing.T) {
+		for _, p := range testApp.BlockedParams() {
+			p := testProposal(proposal.NewParamChange(p[0], p[1], "value"))
+			err := handler(ctx, p)
+			require.Error(t, err)
+			require.ErrorIs(t, err, paramfilter.ErrBlockedParameter)
+		}
+	})
+
+	t.Run("test that another proposal with a blocked param (evidence consensus params) is rejected", func(t *testing.T) {
```
staheri14, line 42: > oh true! I can remove the second test case b/c it should be covered by test case one.

In light of this, wouldn't this third (now second) test be covered by the first one as well?

rootulp: True. I think we could replace the second test case with a test that verifies that evidence params is actually in the list of `testApp.BlockedParams()...` b/c that's something we don't cover with just the first test case.

I have to modify this PR to account for v1 / v2 branching so I expect that we'll want a standalone test case to verify that evidence params don't exist in v1 and do exist in v2... but will keep this unresolved until then.

rootulp:
Closing as won't do now but we do plan on doing https://github.com/celestiaorg/celestia-app/issues/2418 later.

Issue #2033, opened 2023-07-04
EPIC: core/app testground tooling


evan-forbes:
In the past, we've explored using testground for gathering insights by running network tests. Typically, that has involved manually running a test periodically when we are specifically checking on the performance of a specific network, while we also manually query for data to get the desired insights.

This seems to be what testground is good at, providing us with access to a _more_ realistic network to run quick benchmarks or experiements.

We want to continue that trend and empower our testground tooling even more. We want to thave the ability to 
- replicate gossiping issues from live networks
- find/experiment with gossiping opimizations
- test for performance regression

this will likely involve (but is not limited to):
- Separating out the core/app portion of the existing testground tests
- Refactoring the testground tests with an aim to make it easier to write proprietary tests
- Document the new tests
- Establish a testing pipeline
  - Include some mechanism to automatically record a standard set of metrics, ideally in a way that the same report can be generated on a live network as well (ie using pyroscrope/influx)

PR #2455, opened 2023-09-10
refactor!: genesis creation into its own package

evan-forbes:
## Overview

This PR refactors genesis creation into its own package. Besides a general simplification and isolation of logic, this PR also allows for the ease of adding multiple validators to the genesis. The goal being that we can eventually use this genesis creation code across all of our testing utilities that require it including testnode, e2e, the testapp integration tests, and specificlally testground.

This change was spun out of my recent efforts to refactor the testground tests and put them into celestia-app. #2033

## Checklist


test/util/testnode/config.go
```diff
@@ -93,31 +70,60 @@ func (c *Config) WithSupressLogs(sl bool) *Config {
...
-	return c
-}
-
 // WithTimeoutCommit sets the CommitTimeout and returns the Config.
 func (c *Config) WithTimeoutCommit(d time.Duration) *Config {
 	c.TmConfig.Consensus.TimeoutCommit = d
 	return c
 }
 
+// WithAccounts sets the genesis accounts and returns the Config.
```
SweeXordious, line 79:
```suggestion
// WithFundedAccounts sets the genesis accounts and returns the Config.
```

test/genesis/genesis_options.go
```diff
@@ -14,13 +15,13 @@ import (
 	v1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
 )
 
-// GenesisOption allows for arbitrary changes to be made on the genesis state
+// Modifier allows for arbitrary changes to be made on the genesis state
 // after initial accounts have been added. It accepts the genesis state as input
 // and is expected to return the modifed genesis as output.
```
SweeXordious, line 20:
```suggestion
// and is expected to return the modified genesis as output.
```

app/test/integration_test.go
```diff
@@ -249,7 +249,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 			"medium random with timeout height",
 			mustNewBlob(ns1, tmrand.Bytes(100000), appconsts.ShareVersionZero),
 			[]user.TxOption{
-				user.SetTimeoutHeight(1000),
+				user.SetTimeoutHeight(10000),
```
evan-forbes, line 252: The configuration is fixed in this PR as to actually use a default of 1 ms for a timeout commit. We have to either increase this since blocks are produced super fast, or add in a delay to slow block production. 

I'll defer to reviewers if they have an opinion. Adding a timeout commit also seems like a great option, since we have no real benefit of producing over 1000 blocks in a single test lol

rootulp (approved):

test/genesis/genesis.go
```diff
@@ -0,0 +1,267 @@
+package genesis
```
cmwaters, line 1: Wouldn't this be better placed under `util`

test/e2e/setup.go
```diff
@@ -23,12 +23,12 @@ import (
...
 )
 
-type GenesisAccount struct {
+type Account struct {
 	PubKey        cryptotypes.PubKey
 	InitialTokens int64
 }
 
-func MakeGenesis(nodes []*Node, accounts []*GenesisAccount) (types.GenesisDoc, error) {
+func MakeGenesis(nodes []*Node, accounts []*Account) (types.GenesisDoc, error) {
```
cmwaters, line 31: Do you think this function could use the `genesis` package instead of duplicating it?

cmwaters (commented):
Looks good

test/e2e/setup.go
evan-forbes: yeah definitely! was going to do that in a follow up allow with testground and testapp

test/genesis/genesis.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/2455/commits/c0ab03282a00aef2233422f74624bcb8e7187e93

rootulp (approved):

rootulp:
test is failing here: https://github.com/celestiaorg/celestia-app/actions/runs/6241326132/job/16943297954?pr=2455#step:4:12

evan-forbes:
thanks @rootulp, I might turn this into a draft. I still need to debug some things on the testground side and the result of that might change this a bit. Will update here by eod

evan-forbes:
updated this slightly to abstract out the function to create the genesis. This was needed for the testground repo as I had to create the gentxs outside of the genesis struct due to a really subtle marshaling issue with testground

rootulp (approved):
Still LGTM

cmwaters (approved):

PR #2456, opened 2023-09-11
feat: Testground tooling

### LINKED ISSUE #2033: EPIC: core/app testground tooling

evan-forbes:
## Overview

This is a refactor of the celestia-app tests in [test-infra](https://github.com/celestiaorg/test-infra). The refactor focuses on further removing boiler plate and using our existing testing utilities for devUX, but also allows for more elaborate test scenarios and data collection. 

Testground requires that each node is responsible for intializing, operating, performing tests, and collecting data on itself. To do this this refactor adds a new `Role` interface, where we define specific roles that `Plan`, `Execute`, and `Retro`. The first test case makes use of the interface to define a `Leader` and `Follower`, where we can have the `Leader` generate all the keys, configs, genesis, etc during `Plan` and then publishes that data to all `Followers`, which use their respective configuration to start their node. After execution of the test, each `Role` can run arbitrary `Retro` logic over the data produced or collected during the test run.

The first experiment is mean to add as similar environment to mainnet as possible. We will use this as a base control case for an upcoming PR that adds the unbounded blocksize test

Please see the [README](https://github.com/celestiaorg/celestia-app/blob/db0fdfd3197238c77e72f59222cfaa3ec6648396/test/testground/README.md) for more info and diagrams

closes #2033 blocking https://github.com/celestiaorg/celestia-core/issues/945

## Checklist


evan-forbes:
the test can run a 3 validator network atm, but there are still a few minor but blocking bugs left. I'll add a way to sync before the "planning" stage to handle those bugs, then we should be good write the larger tests :crossed_fingers:

test/genesis/files.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // InitFiles initializes the files for a new tendermint node with the provided
 // genesis. It will use the validatorIndex to save the validator's consensus
-// key.
+// key. TODO: abstract this function so that testground can use it.
```
Bidon15, line 16: king feature 👑

test/testground/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+#::: (OPTIONAL) RUNTIME CONTAINER
+#:::
+
+## The 'AS runtime' token is used to parse Docker stdout to extract the build image ID to cache.
+FROM ${RUNTIME_IMAGE} AS runtime
+RUN apk add --no-cache bash gcompat curl
+# PLAN_DIR is the location containing the plan source inside the build container.
+ENV PLAN_DIR /plan
+ENV GOLOG_LOG_FMT="json"
+ENV GOLOG_FILE /var/log/node.log
```
Bidon15, line 81: this is solely needed for ipfs/logger repo that node uses for logging. 
I think app can safely remove this if there are no plans to include celestia-node repo part into the build

test/testground/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+
+
+# HOME ENV is crucial for app/sdk -> remove at your OWN RISK!
+ENV HOME /
+
+COPY --from=builder /testground_dep_list /
+COPY --from=builder ${PLAN_DIR}/testplan.bin /testplan
+
+EXPOSE 9090 26657 26656 1317 26658 26660 26659 2121 4318 4317 30000
```
Bidon15, line 90: we can safely reduce 2121 4318 4317 bc of open-telemetry and multiaddr ports that node uses

test/testground/compositions/sanity/plan.toml
```diff
@@ -0,0 +1,49 @@
...
+latency = "0"
+bandwidth = "420Mib"
+validators = "3"
+topology = "connect_all"
+pex = "false"
+timeout_propose = "10s"
+timeout_commit = "11s"
+per_peer_bandwidth = "5Mib"
+blob_sequences = "1"
+blob_sizes = "100"
```
Bidon15, line 25: nice 
Screenshot 2023-04-26 at 16 57 51

Bidon15 (commented):
Some nits. Will continue once I get more free time. 
Sorry for the late review stages

evan-forbes:
There are some basic tests breaking due to the square size increase. I'll fix those soon, but in the meantime we can [finally] start the review process.

test/testground/compositions/utils.go
```diff
@@ -0,0 +1,95 @@
...
+		RoutingPolicy: network.AllowAll,
+	}
+
+	config.IPv4 = runenv.TestSubnet
+
+	// using the assigned `GlobalSequencer` id per each of instance
+	// to fill in the last 2 octets of the new IP address for the instance
+	ipC := byte((initCtx.GlobalSeq >> 8) + 1)
+	ipD := byte(initCtx.GlobalSeq)
+	config.IPv4.IP = append(config.IPv4.IP[0:2:2], ipC, ipD)
```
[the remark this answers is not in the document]
evan-forbes: @Bidon15 this was copied from the past implemetation, however I don't actually think it gets used and what code rabbit says seems like a legit concern. Can we remove this? Why doesn't testground do this for us?

pkg/appconsts/testground/app_consts.go
```diff
@@ -0,0 +1,7 @@
+package v1
```
rootulp, line 1: [blocking] I find it confusing that this package statement uses `v1` when this file isn't located it in the `v1` directory. Did this mean to say:

```suggestion
package testground
```

pkg/appconsts/versioned_consts.go
```diff
@@ -1,12 +1,12 @@
...
 
 import (
+	testground "github.com/celestiaorg/celestia-app/pkg/appconsts/testground"
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
-	v2 "github.com/celestiaorg/celestia-app/pkg/appconsts/v2"
 )
 
 const (
-	LatestVersion = v2.Version
+	LatestVersion = v1.Version
```
rootulp, line 9: [question] why was this downgraded from v2 to v1? Why is it v2 on main already?

pkg/appconsts/versioned_consts.go
```diff
@@ -1,12 +1,12 @@
 package appconsts
 
 import (
+	testground "github.com/celestiaorg/celestia-app/pkg/appconsts/testground"
```
rootulp, line 4: can the import alias be dropped?

```suggestion
	"github.com/celestiaorg/celestia-app/pkg/appconsts/testground"
```

pkg/appconsts/versioned_consts.go
```diff
@@ -24,8 +24,13 @@ func SubtreeRootThreshold(_ uint64) int {
...
 // governance. See `DefaultGovMaxSquareSize`.
-func SquareSizeUpperBound(_ uint64) int {
-	return v1.SquareSizeUpperBound
+func SquareSizeUpperBound(v uint64) int {
+	switch v {
+	case testground.Version:
+		return testground.SquareSizeUpperBound
+	default:
+		return v1.SquareSizeUpperBound
+	}
```
rootulp, line 33: [question] Why doesn't this have case statements for each version?

```go
	case testground.Version:
		return testground.SquareSizeUpperBound
	case v1.Version:
		return v1.SquareSizeUpperBound
	case v2.Version:
		return v2.SquareSizeUpperBound
```

even if v1 and v2 have the same constants, this seems more readable and future proof in case we update the v2 constant prior to cutting the v2 release

test/testground/go.mod
```diff
@@ -0,0 +1,229 @@
...
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	nhooyr.io/websocket v1.8.6 // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
+replace (
+	// use cosmos fork of keyring
+	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
```
rootulp, line 217: why is this replace needed? We removed a similar replace from celestia-app a while back.

Ref: https://github.com/celestiaorg/celestia-app/pull/548

test/testground/go.mod
```diff
@@ -0,0 +1,229 @@
...
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.18.3-sdk-v0.46.14
```
rootulp, line 218: should this always be kept in sync with the version in the root level go.mod?

test/testground/go.mod
```diff
@@ -0,0 +1,229 @@
...
+	// dgrijalva/jwt-go is deprecated and doesn't receive security updates.
+	// TODO: remove it: https://github.com/cosmos/cosmos-sdk/issues/13134
+	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.4.2
+	// Fix upstream GHSA-h395-qcrw-5vmq and GHSA-3vp4-m3rf-835h vulnerabilities.
+	// TODO Remove it: https://github.com/cosmos/cosmos-sdk/issues/10409
+	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.9.0
```
rootulp, line 224: should we create follow up issues to tackle resolving these TODOs?

test/testground/network/command_test.go
```diff
@@ -0,0 +1,105 @@
...
+	op.StopJob("fake_job")
+
+	if _, exists := op.jobs["fake_job"]; exists {
+		t.Errorf("Job should be removed after StopJob")
+	}
+}
+
+func TestOperator_Run(t *testing.T) {
+	// can't run this test if the runenv is actually used during the Run call
+	t.Skip("Skipping TestOperator_Run")
```
rootulp, line 33: do we want a follow up issue to enable this test?

test/testground/network/consensus_node.go
```diff
@@ -0,0 +1,326 @@
...
+// functions. It returns the last error that was not nil if any of the cleanup
+// functions returned an error.
+func (cn *ConsensusNode) Stop() error {
+	var err error
+	for _, stop := range cn.stopFuncs {
+		if sterr := stop(); sterr != nil {
+			err = sterr
+		}
+	}
+	return err
```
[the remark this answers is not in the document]
rootulp: cool idea!

go.work
```diff
@@ -0,0 +1,6 @@
+go 1.21.1
+
+use (
+	.
+	./test/testground
+)
```
rootulp, line 6: [question] are there any tips/tricks/gotchas contributors should know about using Go workspaces?

rootulp (commented):
One blocking comment about package names. 

I skimmed over the tesground code and it looks good to me! I'm a bit surprised at how good some of the CodeRabbit suggestions are.

[proposal] I think we'll want to update Dependabot so that it updates the Dockerfile at test/tesground/Dockerfile. Similar to [this](https://github.com/celestiaorg/celestia-app/blob/c680cf16e40d20da79b488a82860c23b3afc35c1/.github/dependabot.yml#L10-L17).

test/testground/network/consensus_node.go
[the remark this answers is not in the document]
evan-forbes: I originally left this here as this would only have an effect on a single run. Calling the stop functions is almost pointless. We can adopt the usage of multierror tho for good practice

test/testground/go.mod
evan-forbes: it might have been a remnant from a while ago not sure. iirc I just copied all the replaces and pasted in the go mod. it looks like we could remove tho

test/testground/go.mod
evan-forbes: it should always be kept in sync with whatever version of the app its importing. currently its importing the version of the app on this branch, which is that version

https://github.com/celestiaorg/celestia-app/blob/c680cf16e40d20da79b488a82860c23b3afc35c1/go.mod#L242

this PR is actually blocked by #2860, as we must have a mechanism for setting the app version

go.work
evan-forbes: ideally we'd use local references to everything, then they're way cooler. unfortunately when we have to reference a pushed version of code, like in a docker container, they kinda suck since there's always a chicken and egg problem.

for example we push v1.5.0 of the app, then we import v1.5.0 in the testground workspace, which would require cutting a new release if it was being used by anyone lol

test/testground/network/command_test.go
evan-forbes: we might just want to delete it since it won't ever really be ran without setting up a runenv just for it

test/testground/go.mod
evan-forbes: tbh idk what put those there, I didn't I don't think. Does go mod auto add these now?

pkg/appconsts/versioned_consts.go
evan-forbes: yeah we can good catch

pkg/appconsts/versioned_consts.go
evan-forbes: hmm no good catch that shouldn't have changed. maybe this was a messed up merge conflict or something

staheri14 (commented):
Excellent work! While I haven't fully reviewed the code changes yet, I'm currently setting up a network using this PR. I'll provide additional feedback and comments as I progress through this process.

test/testground/compositions/standard/plan.toml
```diff
@@ -0,0 +1,54 @@
+[metadata]
+name = "standard"
+author = "core-app"
+
+[global]
+plan = "core-app"
```
staheri14, line 6: Following this comment, may want to consider renaming it to 
```suggestion
plan = "celestia"
```

app/app.go
```diff
@@ -259,6 +262,15 @@ func New(
 	cdc := encodingConfig.Amino
 	interfaceRegistry := encodingConfig.InterfaceRegistry
 
+	// set the protocol version to a specific value if specified. This is used
+	// for testing purposes.
```
staheri14, line 266: [Question & Optional Suggestion] What is the impact of this if not used for testing purposes? Couldn't this alternatively be a function in the testground package?

test/util/testnode/full_node_test.go
```diff
@@ -132,3 +133,12 @@ func (s *IntegrationTestSuite) Test_FillBlock_InvalidSquareSizeError() {
 		})
 	}
 }
+
+// Test_defaultAppVersion tests that the default app version is set correctly in
+// testnode.
```
staheri14, line 138: [Question] Does `testnode` refer to a particular struct in the testnode package? if not, then it is a bit unclear what `the default app version is set correctly in testnode.` means

test/testground/network/sync.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func DownloadNodeConfigs(ctx context.Context, runenv *runtime.RunEnv, initCtx *run.InitContext) ([]RoleConfig, error) {
+	return DownloadSync(ctx, initCtx, MetaConfigTopic, RoleConfig{}, runenv.TestInstanceCount)
+}
+
+// GenesisWrapper is a simple struct wrapper that makes it easier testground to
+// properly serialize and distribute the genesis file.
+type GenesisWrapper struct {
+	Part    int    `json:"part"`
```
staheri14, line 135: Would be great to add a comment about what the `Part` field is for.

test/testground/network/sync.go
```diff
@@ -0,0 +1,218 @@
...
+	Genesis string `json:"genesis"`
+}
+
+// PublishGenesis publishes the genesis to the sync service. It splits the
+// genesis into 10 parts and publishes each part separately. This gets around the
+// 32Kb limit the underlying websocket has
```
staheri14, line 141:
```suggestion
// 32Kb limit the underlying websocket has.
```

test/testground/network/sync.go
```diff
@@ -0,0 +1,218 @@
...
+		_, err = initCtx.SyncClient.Publish(ctx, GenesisTopic, wrapper)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func DownloadGenesis(ctx context.Context, initCtx *run.InitContext) (json.RawMessage, error) {
```
staheri14, line 174: A func doc would be very helpful.

test/testground/network/sync.go
```diff
@@ -0,0 +1,218 @@
...
+		bz, err := hex.DecodeString(rawGen.Genesis)
+		if err != nil {
+			return nil, err
+		}
+		genesis = append(genesis, bz...)
+	}
+	return genesis, nil
+}
+
+func DownloadSync[T any](ctx context.Context, initCtx *run.InitContext, topic *sync.Topic, t T, count int) ([]T, error) {
```
staheri14, line 194: A func doc would be very helpful.

test/testground/network/sync.go
evan-forbes: 4fc5f8fbe93d43cd6e46644804c54d5364290c44

test/testground/network/sync.go
evan-forbes: 261214a5adbc10044fac9c6c3b4b07ba18c2aea0

test/util/testnode/full_node_test.go
evan-forbes: 3cf0697b18b10c7f36364767f9170336964c70b0

test/testground/network/sync.go
evan-forbes: 261214a5adbc10044fac9c6c3b4b07ba18c2aea0

app/app.go
evan-forbes: the app version must be set internally, so unfortnuately we can't do that since we don't have direct access to the baseapp, only the interface that wraps around it. However, we should be able to remove this now that we have the latest version of our fork of the sdk, since that version sets the appversion internally similar to this

to answer the question tho, it could be used to change the app version. The app version is included in and checked when verifying the header, so doing this without coordination with the rest of the network would result in the node halting.

pkg/appconsts/versioned_consts.go
evan-forbes: I actually disagree with this, as with each version, we'd have to go change each if statement across the entire codebase. besides being a bit annoying, there's no way to tell if we missed one or are being incosistent places. imo, just saying default reads as all versions use the same versions, which feels more concise then having to read each version and compare

evan-forbes:
@rootulp do you have any clue why the ledger test is breaking? its breaking locally as well, but wasn't breaking a few commits ago

pkg/appconsts/testground/app_consts.go
```diff
@@ -0,0 +1,7 @@
+package testground
+
+const (
+	Version              uint64 = 420420420
+	SquareSizeUpperBound int    = 512
+	SubtreeRootThreshold int    = 64
```
rootulp, line 6: [question] related to https://github.com/celestiaorg/celestia-app/pull/2456/files#r1417502424

why define this constant if it is never used?

rootulp (approved):
:shipit: I skipped reviewing the code in test/testground/* and instead focused on the celestia-app portions

test/util/testnode/rpc_client.go
```diff
@@ -24,26 +22,21 @@ var noOpCleanup = func() error { return nil }
...
 	cleanup := func() error {
-		cancel()
-		t.Log("stopping tmNode")
 		err := tmNode.Stop()
 		if err != nil {
 			return err
 		}
 		tmNode.Wait()
-		t.Log("tmNode has stopped")
-		return removeDir(t, path.Join([]string{cctx.HomeDir, "config"}...))
```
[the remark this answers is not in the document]
evan-forbes: yeah, so that we don't need to pass the `testing.T` to this function.

pkg/appconsts/versioned_consts.go
evan-forbes: > I don't understand why silently defaulting to an old version is preferable over a panic.

there's a thread of two discussing this in the version PRs, so apologies if you are already aware of this point, but essentially were relying on there not being any human error to avoid a network halting bug. There's also not really any benefit, since the appversion MUST be checked when verifying the header, both by Comet and the light client. 

If the appversion is not expected or the node/LC doesn't know how to handle that version, it must halt. This means that there are no scenarios where the panic would serve a positive purpose. The only time that it would get hit is when we accidentally forget to update one of the ever increasing if statements, in which case the entire network halts for no reason, even if they upgraded. 

imo we _definitely_ should not introduce a panic. I'm also of the opinion that being specifying only the _changes_ for a constant is actually more readable. If we specify each value for each version and we want to find when something changed, we have to look up each constant and manually compare, which feels weird to me. That's more just a readability opinion, and won't halt the network, so I don't hold that strong of a view on that.

I think we've casually discussed this synchronously, but since this has come up multiple times, I'm happy to schedule a sync or open a discussion specifically for it if need.

test/testground/README.md
```diff
@@ -0,0 +1,136 @@
...
+	Execute(ctx context.Context, runenv *runtime.RunEnv, initCtx *run.InitContext) error
+	// Retro is the last function called in a test by each node. It is
+	// responsible for collecting any data from the node and/or running any
+	// retrospective tests or benchmarks.
+	Retro(ctx context.Context, runenv *runtime.RunEnv, initCtx *run.InitContext) error
+}
+
+var _ Role = (*Leader)(nil)
+
+var _ Role = (*Follower)(nil)
```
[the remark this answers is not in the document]
evan-forbes: resolved in https://github.com/celestiaorg/celestia-app/pull/2456/commits/93cb5d45c879f2eb8f76ff034c4945f707f55708

pkg/appconsts/testground/app_consts.go
```diff
@@ -0,0 +1,7 @@
+package testground
+
+const (
+	Version              uint64 = 420420420
```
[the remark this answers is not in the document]
evan-forbes: resolved in https://github.com/celestiaorg/celestia-app/pull/2456/commits/93cb5d45c879f2eb8f76ff034c4945f707f55708

evan-forbes:
There's a weird bug where this test is failing consistently in ci bug not locally. It also doesn't make any sense, since it was just passing and the commit that broke it only added docs and removed an usused var

rootulp (approved):

staheri14 (approved):

test/util/testnode/config.go
```diff
@@ -119,7 +124,8 @@ func DefaultConfig() *Config {
 		WithAppConfig(DefaultAppConfig()).
 		WithAppOptions(DefaultAppOptions()).
 		WithAppCreator(cmd.NewAppServer).
-		WithSupressLogs(true)
+		WithSuppressLogs(true).
+		WithConsensusParams(DefaultConsensusParams())
```
evan-forbes, line 128: setting the default consensus params here fixes the failing test, which apparently we aren't doing doing on main! This doesn't fail on main, because we don't actually have a test that checks that the version is updated but now we do [here](https://github.com/celestiaorg/celestia-app/blob/50bc61ea72ddac7f7372ce7e4471be295670772b/test/util/testnode/full_node_test.go#L129-L136)

rootulp (approved):

PR #2467, opened 2023-09-11
fix(scripts): single node celestia app home fix

SweeXordious:
## Overview

The default Celestia app home directory contains a typo

## Checklist

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -10,7 +10,7 @@ KEY_NAME="validator"
 KEYRING_BACKEND="test"
 COINS="1000000000000000utia"
 DELEGATION_AMOUNT="5000000000utia"
-CELESTIA_APP_HOME="${HOME}/.celestia-appd"
+CELESTIA_APP_HOME="${HOME}/.celestia-app"
 CELESTIA_APP_VERSION=$(celestia-appd version 2>&1)
 
 echo "celestia-app home: ${CELESTIA_APP_HOME}"


scripts/single-node.sh
```diff
@@ -10,7 +10,7 @@ KEY_NAME="validator"
 KEYRING_BACKEND="test"
 COINS="1000000000000000utia"
 DELEGATION_AMOUNT="5000000000utia"
-CELESTIA_APP_HOME="${HOME}/.celestia-appd"
+CELESTIA_APP_HOME="${HOME}/.celestia-app"
```
rootulp, line 13: wow you're right `./celestia-app` looks correct b/c

Ref:
- https://github.com/celestiaorg/celestia-app/blob/159473dfbee3c2931cc34256b2471ff585cae5d7/app/app.go#L191
- https://github.com/celestiaorg/celestia-app/blob/159473dfbee3c2931cc34256b2471ff585cae5d7/app/app.go#L106

For some reason I thought `~./celestia-appd` was correct because locally I had a `~/.celestia-appd` directory. Since the rest of this script uses the environment variable `CELESTIA_APP_HOME`, it works before and after this change but I think we should merge it so that users who use this script don't get confused by two local directories.

rootulp (approved):

evan-forbes (approved):

scripts/single-node.sh
evan-forbes: django-unchained-drama

I've noticed this as well. I'm down to be consistent everywhere, however changing either would probably break a loooooooooot of stuff

Issue #2485, opened 2023-09-13
mint.BondDenom is not changeable via governance


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/41ead770b48d6b86dce3f252e1be844b4978e907/specs/src/specs/params.md?plain=1#L55

## Problem

BondDenom is a genesis param and isn't governance modifiable

## Proposal

Update the boolean in the last column to state that this param isn't governance modifiable

Issue #2481, opened 2023-09-13
Add x/mint constants to params.md


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/41ead770b48d6b86dce3f252e1be844b4978e907/specs/src/specs/params.md?plain=1#L56-L57

## Problem

We removed these params from the x/mint fork in this repo.

## Proposal

Remove them from params.md

evan-forbes:
I think these should be included in the specs as they are parameter that could be changed, they just require a hardfork. This is same for other hardcoded constants such as max square size

rootulp:
hmm okay since they can be changed with a hardfork, I understand your point. I think this line is slightly misleading 

https://github.com/celestiaorg/celestia-app/blob/0012451c4dc118767dd59bc8d341878b7a7cacdf/x/mint/README.md?plain=1#L81-L83

and we should clarify that although the parameters have been removed, hard-coded constants exist that can change outside token voting governance. Also I think we should use the same names in params.md as the constant names in 
https://github.com/celestiaorg/celestia-app/blob/0012451c4dc118767dd59bc8d341878b7a7cacdf/x/mint/types/constants.go#L17-L22

and we also need to add one more `TargetInflationRate`.

PR #2490, opened 2023-09-13
docs: update x/mint params.md

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2481
Closes https://github.com/celestiaorg/celestia-app/issues/2485

### DIFF
--- a/specs/src/specs/params.md
+++ b/specs/src/specs/params.md
@@ -52,7 +52,5 @@ are blocked by the `x/paramfilter` module.
 | staking.HistoricalEntries | 10000 | Number of historical entries to persist in store. | True |
 | staking.BondDenom | utia | Bondable coin denomination. | False |
 | staking.MinCommissionRate | 0.05 (5%) | Minimum commission rate used by all validators. | True |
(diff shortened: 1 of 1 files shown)


specs/src/specs/params.md
```diff
@@ -52,7 +52,5 @@ are blocked by the `x/paramfilter` module.
 | staking.HistoricalEntries | 10000 | Number of historical entries to persist in store. | True |
 | staking.BondDenom | utia | Bondable coin denomination. | False |
 | staking.MinCommissionRate | 0.05 (5%) | Minimum commission rate used by all validators. | True |
-| mint.BondDenom | utia | Denomination that is inflated and sent to the distribution module account. | True |
-| mint.InflationRateChange | 10.0% | The rate at which the annual provisions decrease each year. | False |
-| mint.InflationRate | 8.0% | Initial annual inflation rate used to calculate the annual provisions. | False |
+| mint.BondDenom | utia | Denomination that is inflated and sent to the distribution module account. | False |
```
evan-forbes, line 55: I don't think we should remove these for the same reasons we don't remove max square size

these can be changed, they just require a hardfork

Issue #2454, opened 2023-09-08
Transactions on mocha-4 got stuck in the mempool


evan-forbes:
When submitting some transaction using txsim, 4 managed to get stuck in the mempool, meaning that they were accepted by `CheckTx` but reject by `PrepareProposal`.

This should never happen, as malicious parties could fill the mempool with what appear to be valid high value transactions for free.

here are the transactions that got stuck

```json
[...]

PR #2515, opened 2023-09-15
fix: add time to the sdk.Context used in PrepareProposal

evan-forbes:
## Overview

I'm not sure if this fixes all of #2454, but it at least fixes one bug similar to it.

The fix in this PR allows for the antehandlers that are ran during prepare proposal to have access to a time close to the actual block time. It does this by simply adding `time.Now` to the header used for antehandler state access.

## Checklist

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -1,6 +1,8 @@
 package app
 
 import (
+	"time"
+
 	"github.com/celestiaorg/celestia-app/app/ante"
 	"github.com/celestiaorg/celestia-app/pkg/da"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
@@ -18,7 +20,11 @@ import (
 func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePrepareProposal {
 	// create a context using a branch of the state and loaded using the
 	// proposal height and chain-id
-	sdkCtx := app.NewProposalContext(core.Header{ChainID: req.ChainId, Height: app.LastBlockHeight() + 1})
+	sdkCtx := app.NewProposalContext(core.Header{
+		ChainID: req.ChainId,
+		Height:  app.LastBlockHeight() + 1,
+		Time:    time.Now(),
+	})
 	// filter out invalid transactions.
 	// TODO: we can remove all state independent checks from the ante handler here such as signature verification
 	// and only check the state dependent checks like fees and nonces as all these transactions have already
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -1,8 +1,6 @@
 package app
 
 import (
-	"time"
-
 	"github.com/celestiaorg/celestia-app/app/ante"
 	"github.com/celestiaorg/celestia-app/pkg/da"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
@@ -22,8 +20,8 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 	// proposal height and chain-id
 	sdkCtx := app.NewProposalContext(core.Header{
 		ChainID: req.ChainId,
-		Height:  app.LastBlockHeight() + 1,
-		Time:    time.Now(),
+		Height:  req.Height,
+		Time:    req.Time,
 	})
 	// filter out invalid transactions.
 	// TODO: we can remove all state independent checks from the ante handler here such as signature verification
--- a/app/test/prepare_proposal_context_test.go
+++ b/app/test/prepare_proposal_context_test.go
@@ -36,7 +36,7 @@ func TestTimeInPrepareProposalContext(t *testing.T) {
 	cfg := testnode.DefaultConfig().WithAccounts(accounts)
 	cctx, _, _ := testnode.NewNetwork(t, cfg)
 	ecfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-
+	vestAccName := "vesting"
 	type test struct {
 		name         string
 		msgFunc      func() (msgs []sdk.Msg, signer string)
@@ -46,7 +46,6 @@ func TestTimeInPrepareProposalContext(t *testing.T) {
 		{
 			name: "create continuous vesting account with a start time in the future",
 			msgFunc: func() (msgs []sdk.Msg, signer string) {
-				vestAccName := "vesting"
 				_, _, err := cctx.Keyring.NewMnemonic(vestAccName, keyring.English, "", "", hd.Secp256k1)
 				require.NoError(t, err)
 				sendAcc := accounts[0]
@@ -67,7 +66,6 @@ func TestTimeInPrepareProposalContext(t *testing.T) {
 		{
 			name: "send funds from the vesting account after it has been created",
 			msgFunc: func() (msgs []sdk.Msg, signer string) {
-				vestAccName := "vesting"
 				sendAcc := accounts[1]
 				sendingAccAddr := testfactory.GetAddress(cctx.Keyring, sendAcc)
 				vestAccAddr := testfactory.GetAddress(cctx.Keyring, vestAccName)
--- a/app/test/prepare_proposal_context_test.go
+++ b/app/test/prepare_proposal_context_test.go
@@ -27,7 +27,7 @@ import (
 // accounts never being able to spend funds.
 func TestTimeInPrepareProposalContext(t *testing.T) {
 	if testing.Short() {
-		t.Skip("skipping DebugVestingAccountsTime test in short mode.")
+		t.Skip("skipping TestTimeInPrepareProposalContext test in short mode.")
 	}
 	accounts := make([]string, 35)
 	for i := 0; i < len(accounts); i++ {
--- a/app/test/prepare_proposal_context_test.go
+++ b/app/test/prepare_proposal_context_test.go
@@ -1,0 +1,102 @@
+package app_test
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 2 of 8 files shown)


app/test/prepare_proposal_context_test.go
```diff
@@ -0,0 +1,102 @@
...
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+// TestTimeInPrepareProposalContext checks for an edge case where the block time
+// needs to be included in the sdk.Context that is being used in the
+// antehandlers. If a time is not included in the context, then the second
+// transaction in this test will always be filtered out, result in vesting
+// accounts never being able to spend funds.
+func TestTimeInPrepareProposalContext(t *testing.T) {
+	if testing.Short() {
```
evan-forbes, line 29: we could have included this in the standard cosmos-sdk test that we're already running, however it would be dependent upon the second testcase always being second

app/test/prepare_proposal_context_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Skip("skipping DebugVestingAccountsTime test in short mode.")
```
rootulp, line 30: [question] is this name supposed to match the test name?

```suggestion
		t.Skip("skipping TestTimeInPrepareProposalContext test in short mode.")
```

app/test/prepare_proposal_context_test.go
```diff
@@ -0,0 +1,102 @@
...
+	type test struct {
+		name         string
+		msgFunc      func() (msgs []sdk.Msg, signer string)
+		expectedCode uint32
+	}
+	tests := []test{
+		{
+			name: "create continuous vesting account with a start time in the future",
+			msgFunc: func() (msgs []sdk.Msg, signer string) {
+				vestAccName := "vesting"
```
rootulp, line 49: [optional] I think this could be hoisted up to line ~38 and removed from this line and the second test case b/c IIUC the `vestAccName` should be identical for both test cases.

rootulp (dismissed):

rootulp (dismissed):

app/test/prepare_proposal_context_test.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/2515/commits/5ad837147e243a008735eb1faaa5679240a7b2c4

app/prepare_proposal.go
```diff
@@ -18,7 +20,11 @@ import (
 func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePrepareProposal {
 	// create a context using a branch of the state and loaded using the
 	// proposal height and chain-id
-	sdkCtx := app.NewProposalContext(core.Header{ChainID: req.ChainId, Height: app.LastBlockHeight() + 1})
+	sdkCtx := app.NewProposalContext(core.Header{
+		ChainID: req.ChainId,
+		Height:  app.LastBlockHeight() + 1,
+		Time:    time.Now(),
```
cmwaters, line 26: There is a high possibility that time.Now() is greater than the block time proposed in the header (as it stems from the commit of the previous block). This could compromise the consistency property whereby a transaction seems valid in PrepareProposal but invalid in ProcessProposal. If the transaction isn't kicked from the mempool it could result in the chain halting (as proposers continue to propose a vesting transaction that they think is valid (it has a later time) but continues to get rejected (as validators are basing it off the last block time). Ideally we find a way to add the last block time to the context. If it's too difficult I would recommend setting Time to `time.Now().Sub(TargetBlockTime)` as an extra precaution but if there are multi rounds the same consistency problem could occur

cmwaters (changes requested):

app/prepare_proposal.go
evan-forbes: good point! I totally blanked this morning on us running the entire antehandler on all txs in process proposal

we definitely shouldn't do this

evan-forbes: I think the only solution here is to get the time in the request from comet similar to how its done on celestia-core/main or comet

liamsi, line 26: Yes, this must be the last block's time! Otherwise this is wrong. For example, when calculating the spendable amount, the only real reference the chain has is the last block, as the current block has not been finalized.

This looks about right:
https://github.com/cosmos/cosmos-sdk/blob/c7e0bd7b54d037a36ef7d31cc89ecabdf2124913/baseapp/abci.go#L409-L416

But please double-check that the `req` actually contains the correct time.

liamsi (changes requested):
See either: https://github.com/celestiaorg/celestia-app/pull/2515/files#r1327951681
or: https://github.com/celestiaorg/celestia-app/pull/2515/files#r1327582804

app/prepare_proposal.go
liamsi: Note: it does not seem to be the case that the correct time is set here:
https://github.com/celestiaorg/celestia-core/blob/d24c81afe0ecc4ccfbb7eb543a45a11794f04db4/state/execution.go#L118-L124

Only later in MakeBlock the time is set (correctly): 
- https://github.com/celestiaorg/celestia-core/blob/d24c81afe0ecc4ccfbb7eb543a45a11794f04db4/state/execution.go#L152
- https://github.com/celestiaorg/celestia-core/blob/3f3b7cc57f5cfc5e846ce781a9a407920e54fb72/state/state.go#L249

liamsi: Compare `main` of vanilla cometBFT: https://github.com/cometbft/cometbft/blob/1c6f86c840c64f83f4409ccca6b664234866cc38/state/execution.go#L128C9-L139

evan-forbes:
Thanks again @cmwaters and @liamsi for catching the time.Now bug, I don't know what I was thinking. Too many late nights deugging testground last week! :sweat_smile: . Per the first and second thread, the only solution to this is to use the same time that will be used in process proposal (the time in the header from the last commit). That's what's done in https://github.com/celestiaorg/celestia-core/pull/1081, which should be merged before this PR.

app/test/prepare_proposal_context_test.go
```diff
@@ -0,0 +1,100 @@
...
+				msg := banktypes.NewMsgSend(
+					vestAccAddr,
+					sendingAccAddr,
+					sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1))),
+				)
+				return []sdk.Msg{msg}, vestAccName
+			},
+			expectedCode: abci.CodeTypeOK,
+		},
+	}
```
evan-forbes, line 81: one additional test we can do is to have a third test case where we attempt to send too much and expect it to fail. I have a commit with this ready, however I'm worried that it will cause flakiness do to the process requiring a specific time delay window.

rootulp (approved):

liamsi (approved):
👏🏼

app/test/prepare_proposal_context_test.go
```diff
@@ -0,0 +1,100 @@
...
+	abci "github.com/tendermint/tendermint/abci/types"
+	tmrand "github.com/tendermint/tendermint/libs/rand"
+)
+
+// TestTimeInPrepareProposalContext checks for an edge case where the block time
+// needs to be included in the sdk.Context that is being used in the
+// antehandlers. If a time is not included in the context, then the second
+// transaction in this test will always be filtered out, result in vesting
+// accounts never being able to spend funds.
+func TestTimeInPrepareProposalContext(t *testing.T) {
```
cmwaters, line 28: I feel like we defer to using integration tests too often when unit tests would suffice

app/test/prepare_proposal_context_test.go
```diff
@@ -0,0 +1,100 @@
...
+				_, _, err := cctx.Keyring.NewMnemonic(vestAccName, keyring.English, "", "", hd.Secp256k1)
+				require.NoError(t, err)
+				sendAcc := accounts[0]
+				sendingAccAddr := testfactory.GetAddress(cctx.Keyring, sendAcc)
+				vestAccAddr := testfactory.GetAddress(cctx.Keyring, vestAccName)
+				msg := vestingtypes.NewMsgCreateVestingAccount(
+					sendingAccAddr,
+					vestAccAddr,
+					sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1000000))),
+					time.Now().Unix(),
```
cmwaters, line 58: This is not a start time in the future as is stated in the test name

app/test/prepare_proposal_context_test.go
cmwaters: I don't think it will be flaky if you set the end time well in the future and try send all the funds

cmwaters (approved):
Few minor comments but mostly looks good

app/test/prepare_proposal_context_test.go
evan-forbes: > well in the future and try send all the funds

yeah good point, sorry I was not clear enough. the thing that I was trying to test with the additional case here was not just that a tx would fail if there was not enough of a spendable balance, I think we have that somehwere already (#2004 iirc). That wouldn't be flakey. 

I was trying to create a test that checks that puts us in the exact scenario we would get in if we used `time.Now()`. Where a tx doesn't get filtered but is invalid.

app/test/prepare_proposal_context_test.go
evan-forbes: https://github.com/celestiaorg/celestia-app/issues/2533

app/test/prepare_proposal_context_test.go
evan-forbes: will grab this in a followup

Issue #2517, opened 2023-09-17
Add a metric for filtered txs


evan-forbes:
While there are currently known exceptions to this rule due to CheckTx not [ordering blob txs before normal ones](https://github.com/celestiaorg/celestia-app/blob/b59e876cd15ea9f17608d1c2ca865a86151b7e5f/pkg/square/square.go#L42) #1947, in general all transactions that make it through `CheckTx` should also be able to make it through `PrepareProposal`. Therefore we should keep track of when transactions [get filtered in `PrepareProposal`](https://github.com/celestiaorg/celestia-app/blob/b59e876cd15ea9f17608d1c2ca865a86151b7e5f/app/validate_txs.go#L36-L53). 

As suggested by by @cmwaters, we should monitor the filtered txs via metrics and and logged errors. We already have access to the logger where txs are filtered in the app, but the application currently does not have access to metrics.

Perhap not an exaustive list, but to fix this we could:
- return the metrics that we want to record in the `Prepare/ProcessProposalResponse`s 
- initialize the application with its own prometheus metrics (its own client and port)
- pass the existing comet prometheus client to the application
- query the application via a new abci query at specific points

PR #2518, opened 2023-09-18
chore: log filtered transactions and fix gas limits in tests

evan-forbes:
## Overview

This PR adds loggers back to the filter functions (but actually uses them this time), and logs when transaction get filtered during `PrepareProposal`. It also re-adds an obtuse gas limit to the transactions in the `TestPrepareProposalConsistency` test, where those errors were getting caught by the filter functions. 

part of #2517

## Checklist

### DIFF
--- a/app/test/fuzz_abci_test.go
+++ b/app/test/fuzz_abci_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/pkg/user"
 	testutil "github.com/celestiaorg/celestia-app/test/util"
 	"github.com/stretchr/testify/require"
 	abci "github.com/tendermint/tendermint/abci/types"
@@ -102,6 +103,7 @@ func TestPrepareProposalConsistency(t *testing.T) {
 						true,
 						testutil.ChainID,
 						accounts[:tt.count],
+						user.SetGasLimitAndFee(1_000_000_000, 0.1),
 					)
 					// create 100 send transactions
 					sendTxs := testutil.SendTxsWithAccounts(
@@ -113,6 +115,7 @@ func TestPrepareProposalConsistency(t *testing.T) {
 						accounts[0],
 						accounts[len(accounts)-sendTxCount:],
 						testutil.ChainID,
+						user.SetGasLimitAndFee(1_000_000, 0.1),
 					)
 					txs = append(txs, sendTxs...)
 					resp := testApp.PrepareProposal(abci.RequestPrepareProposal{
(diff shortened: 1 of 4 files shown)


app/validate_txs.go
```diff
@@ -55,7 +58,7 @@ func filterStdTxs(dec sdk.TxDecoder, ctx sdk.Context, handler sdk.AnteHandler, t
 // filterBlobTxs applies the provided antehandler to each transaction
 // and removes transactions that return an error. Panics are caught by the checkTxValidity
 // function used to apply the ante handler.
-func filterBlobTxs(dec sdk.TxDecoder, ctx sdk.Context, handler sdk.AnteHandler, txs []tmproto.BlobTx) ([]tmproto.BlobTx, sdk.Context) {
+func filterBlobTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handler sdk.AnteHandler, txs []tmproto.BlobTx) ([]tmproto.BlobTx, sdk.Context) {
```
evan-forbes, line 61: we are now hitting the unwritten 5 argument for a small function rule, so we might want to refactor this

app/test/fuzz_abci_test.go
```diff
@@ -102,6 +103,7 @@ func TestPrepareProposalConsistency(t *testing.T) {
 						true,
 						testutil.ChainID,
 						accounts[:tt.count],
+						user.SetGasLimitAndFee(1_000_000_000, 0.1),
```
evan-forbes, line 106: [for reviewers]
this fix could be separated out into its own PR. happy to do so if reviewers would appreciate that. originally went to go fix this issue and added the loggers simply to check if this change would fix the issue.

evan-forbes:
also see https://github.com/celestiaorg/celestia-app/issues/2519

rootulp (approved):

app/validate_txs.go
```diff
@@ -42,6 +44,7 @@ func filterStdTxs(dec sdk.TxDecoder, ctx sdk.Context, handler sdk.AnteHandler, t
 		// simply want to remove this tx, or we're catching a panic from one
 		// of the anteHanders which is logged.
 		if err != nil {
+			logger.Error("filtering already checked transaction", "tx", tmbytes.FromBytes(tx), "error", err)
```
cmwaters, line 47: Ideally would prefer not to dump the entire transaction. It wouldn't really help for debugability. I think it would be better either to decode it and print it out or just to print out the message type

evan-forbes: whoops good catch mean to hash that

evan-forbes: logging the hash and the msg types for the sdk txs

https://github.com/celestiaorg/celestia-app/pull/2518/commits/12eed6d464767cc1e773b97dc3c5b0639558733b

e486596441113fb01f318ba4d33a7edb85dceec1

rootulp (approved):

cmwaters (approved):

Issue #2555, opened 2023-09-20
markdown-linter doesn't run on PRs to main


rootulp:
## Context

We recently introduced a new CI workflow: https://github.com/celestiaorg/celestia-app/blob/9efd8b7fb9dabb551ebeb946baf829d3dd2f2323/.github/workflows/markdown-linter.yml

## Problem

markdown-linter isn't run on https://github.com/celestiaorg/celestia-app/pull/2554

### Option A

Add this to https://github.com/celestiaorg/celestia-app/blob/9efd8b7fb9dabb551ebeb946baf829d3dd2f2323/.github/workflows/markdown-linter.yml

```yaml
on:
  push:
    branches:
      - main
      - "v[0-9]+.x"
    tags:
      - "v[0-9]+.[0-9]+.[0-9]+"
      - "v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
      - "v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+"
  pull_request:
```

### Option B

Make https://github.com/celestiaorg/celestia-app/blob/main/.github/workflows/ci-release.yml invoke markdown-linter

rootulp:
Option A seems cleaner to me but I think Option B may enforce that markdown-lint runs if we adopt the approach of creating a tag and pushing it and/or cutting a release via this thingy

<img width="369" alt="Screenshot 2023-09-20 at 11 25 38 AM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/6fa174d8-5cbf-47d5-bb30-0c034afc3163">


instead of the process we currently use to create new releases which is:

<img width="184" alt="Screenshot 2023-09-20 at 11 25 10 AM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/be6b8525-c30c-408c-a84b-d9459cb00abb">

PR #2556, opened 2023-09-20
ci: run markdown-linter on ci-release

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2555

## Testing
1. https://github.com/rootulp/celestia-app/pull/36
2. https://github.com/rootulp/celestia-app/pull/37

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -31,13 +31,16 @@ jobs:
   lint:
     uses: ./.github/workflows/lint.yml
 
+  markdown-linter:
+    uses: ./.github/workflows/markdown-linter.yml
+
   test:
     uses: ./.github/workflows/test.yml
 
   # If this was a workflow dispatch event, we need to generate and push a tag
   # for goreleaser to grab
   version_bump:
-    needs: [lint, test]
+    needs: [lint, markdown-linter, test]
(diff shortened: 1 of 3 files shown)


.github/workflows/markdown-linter.yml
```diff
@@ -1,6 +1,6 @@
-name: "markdown-linter"
+name: markdown-linter
 on:
-  push:
+  workflow_call:
```
SweeXordious, line 3: Doesn't this mean that the linter will no be run on PRs and stuff? if we want to use this, we should also make the `lint` workflow call it, no?

SweeXordious (approved):

cmwaters (approved):

.github/workflows/ci-release.yml
```diff
@@ -31,13 +31,16 @@ jobs:
   lint:
     uses: ./.github/workflows/lint.yml
 
+  markdown-linter:
+    uses: ./.github/workflows/markdown-linter.yml
```
evan-forbes, line 35: do we want to use out fork of this so that we can ignore certain links? other wise I think we could end up in a situation where main and the release branch pass but this fails

evan-forbes (approved):

Issue #2557, opened 2023-09-20
Wrap panics in `PrepareProposal` with the msg type that caused it


evan-forbes:
as suggested by @cmwaters, we should wrap the panic in prepare proposal with the msg type that caused the issue. This will make debugging it in the future a lot easier.

PR #2559, opened 2023-09-21
feat: add decorator to wrap panic messages with the transaction that caused them

cmwaters:
Closes: #2557

This will allow us to understand when a panic happens, which transaction caused it.

As an example here is a mock panic:

```
mock panic
caused by transaction:
*types.MsgSend{from_address:"celestia15v4r83er48zfrm7u6uuqs3lytwxc63c34ltdgn" to_address:"celestia1t5m2f2m000twkmsgaa777rty36wd254292am6c" amount:<denom:"utia" amount:"10" > }
```

### DIFF
--- a/app/ante/panic_test.go
+++ b/app/ante/panic_test.go
@@ -1,0 +1,36 @@
+package ante_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/celestiaorg/celestia-app/app"
+	"github.com/celestiaorg/celestia-app/app/ante"
+	"github.com/celestiaorg/celestia-app/app/encoding"
+	"github.com/celestiaorg/celestia-app/test/util/testnode"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPanicHandlerDecorator(t *testing.T) {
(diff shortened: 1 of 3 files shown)


app/ante/panic_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/celestiaorg/celestia-app/app"
+	"github.com/celestiaorg/celestia-app/app/ante"
+	"github.com/celestiaorg/celestia-app/app/encoding"
+	"github.com/celestiaorg/celestia-app/test/util/testnode"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPanicHandlerDecorator(t *testing.T) {
```
rootulp, line 16: [optional] do we want to include a test case that verifies behavior of the `NewSetUpContextDecorator` is preserved? In other words,  a tx that runs out of gas still results in an `ErrOutOfGas`. I expect it will pass b/c that ante handler would catch an OutOfGasPanic before the panic handler decorator.

rootulp (approved):

app/ante/panic_test.go
cmwaters: Yeah the fact that the `PanicHandlerDecorator` comes before means all behaviour afterwards should theoretically be the same

evan-forbes (approved):

PR #2566, opened 2023-09-21
chore!: remove `MinShareCount` constant

### LINKED ISSUE #1183: Consider removing constants: `MinShareCount` and `MaxShareCount`
to generate the minimum data availability header, but we could probably just remove and calculate the value there if we want to remove one constant.

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/1177#discussion_r1062909867_

Chirag018:
Closes #1183

### DIFF
--- a/docs/architecture/adr-008-square-size-independent-message-commitments.md
+++ b/docs/architecture/adr-008-square-size-independent-message-commitments.md
@@ -81,7 +81,7 @@ Therefore the height of the tree over the subtree roots is in this implementatio
 ### Positive Rollmint changes
 
 1. A Rollup can include the commitment in the block header *before* posting to Celestia because it is size-independent and does not have to wait for Celestia to confirm the square size. In general, the roll-up needs access to this commitment in some form to verify a message inclusion proof guaranteeing data availability, which Rollmint currently does not have access to.
-2. In turn, this would serve as an alternative to [rollmint/adr-007](https://github.com/celestiaorg/optimint/blob/main/docs/lazy-adr/adr-007-header-commit-to-shares.md)
+2. In turn, this would serve as an alternative to [rollmint/adr-007](https://github.com/rollkit/rollkit/blob/f85b994e81bcacf33d187b2dcf2d40657f152408/specs/lazy-adr/adr-007-header-commit-to-shares.md)
 3. Here is one scheme on how a Rollup might use this new commitment in the block header. Let's assume a Rollup that looks like this:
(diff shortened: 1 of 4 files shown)


pkg/square/square.go
```diff
@@ -232,7 +232,10 @@ func (s Square) IsEmpty() bool {
 
 // EmptySquare returns a 1x1 square with a single tail padding share
 func EmptySquare() Square {
-	return shares.TailPaddingShares(appconsts.MinShareCount)
+	// MinshareCount is the minimum number of shares allowed in the original
+	// data square.
+	MinShareCount := appconsts.MinSquareSize
+	return shares.TailPaddingShares(MinShareCount * MinShareCount)
```
rootulp, line 238:
```suggestion
	// minShareCount is the minimum number of shares allowed in the original
	// data square.
	minShareCount := appconsts.MinSquareSize * appconsts.MinSquareSize
	return shares.TailPaddingShares(minShareCount)
```

rootulp (changes requested):

rootulp:
Thanks for the PR @Chirag018 ! I don't think we want to include any more breaking changes in v1.x. so we may not end up merging this

pkg/square/square.go
```diff
@@ -232,7 +232,10 @@ func (s Square) IsEmpty() bool {
 
 // EmptySquare returns a 1x1 square with a single tail padding share
 func EmptySquare() Square {
-	return shares.TailPaddingShares(appconsts.MinShareCount)
+	// minshareCount is the minimum number of shares allowed in the original
```
rootulp, line 235:
```suggestion
	// minShareCount is the minimum number of shares allowed in the original
```

pkg/square/square.go
```diff
@@ -232,7 +232,10 @@ func (s Square) IsEmpty() bool {
 
 // EmptySquare returns a 1x1 square with a single tail padding share
 func EmptySquare() Square {
-	return shares.TailPaddingShares(appconsts.MinShareCount)
+	// minShareCount is the minimum number of shares allowed in the original
+	// data square.
+	minShareCount := appconsts.MinSquareSize*appconsts.MinSquareSize
```
staheri14, line 237: [Optional Suggestion] If we anticipate that the minimum share count might be utilized beyond the scope of the `EmptySquare`, it could be advantageous to create a dedicated function, such as `MinShareCount()`.

staheri14 (approved):
LGTM!

evan-forbes (approved):

PR #2574, opened 2023-09-25
feat!: disable ibc upgrade proposal handler

cmwaters:
The proposal handler to upgrade IBC also invokes the standard `ScheduleUpgrade` within the upgrading module. See https://github.com/cosmos/ibc-go/blob/e2201aaf1b016356bbd40fcdc17988437adce5ae/modules/core/02-client/keeper/proposal.go#L82. This effectively means that one can create a governance proposal to upgrade the chain which is something we didn't want to support. As there is no need to handle updating the IBC client yet, this temporarily disables the proposal type.

### DIFF
--- a/x/upgrade/test/integration_test.go
+++ b/x/upgrade/test/integration_test.go
@@ -16,6 +16,8 @@ import (
 	v1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
 	v1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	ibctypes "github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
+	ibctmtypes "github.com/cosmos/ibc-go/v6/modules/light-clients/07-tendermint/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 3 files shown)


cmwaters:
I create a governance proposal with `*types.UpgradeProposal` as the `Content`. In there I include a `types.Plan` to upgrade at height 10,000. If it passes, [`ScheduleUpgrade`](https://github.com/cosmos/ibc-go/blob/e2201aaf1b016356bbd40fcdc17988437adce5ae/modules/core/02-client/keeper/proposal.go#L95) is called on our standard SDK upgrade keeper. This will create the plan and at height 10,000 will result in all nodes panicking and halting (until people switch to the new version)

cmwaters:
@evan-forbes do you want me to try write a test?

evan-forbes:
> do you want me to try write a test?

this working makes sense to me, but it would be good practice I think to write a test to ensure that this can actually happen and we're not missing something small or hidden before creating a consensus breaking change with some potetially important (albiet temporary as we could always fix with a hardfork) consequences with IBC

evan-forbes (dismissed):

x/upgrade/test/integration_test.go
```diff
@@ -149,6 +151,36 @@ func (s *UpgradeTestSuite) TestNewGovUpgradeFailure() {
...
+	require.NoError(t, err)
+
+	// submit the transaction and wait a block for it to be included
+	signer, err := testnode.NewSignerFromContext(s.cctx, acc)
+	require.NoError(t, err)
+	subCtx, cancel := context.WithTimeout(s.cctx.GoContext(), time.Minute)
+	defer cancel()
+	res, err := signer.SubmitTx(subCtx, []sdk.Msg{msg}, blobfactory.DefaultTxOpts()...)
+	require.Error(t, err)
+	t.Log(res)
```
renaynay, line 179: callum loves to leave logs in his tests

x/upgrade/test/integration_test.go
```diff
@@ -149,6 +151,36 @@ func (s *UpgradeTestSuite) TestNewGovUpgradeFailure() {
...
+	acc := s.unusedAccount()
+	accAddr := getAddress(acc, s.cctx.Keyring)
+	msg, err := v1beta1.NewMsgSubmitProposal(upgradeMsg, dep, accAddr)
+	require.NoError(t, err)
+
+	// submit the transaction and wait a block for it to be included
+	signer, err := testnode.NewSignerFromContext(s.cctx, acc)
+	require.NoError(t, err)
+	subCtx, cancel := context.WithTimeout(s.cctx.GoContext(), time.Minute)
+	defer cancel()
```
renaynay, line 176: alternatively t.Cleanup

renaynay (approved):

x/upgrade/test/integration_test.go
cmwaters: lol good catch

cmwaters: logs are the windows to the program

evan-forbes (approved):
thanks for adding the test! really great find! :+1:

SweeXordious (approved):

PR #2578, opened 2023-09-26
feat: add default time based TTL to mempool config

cmwaters:


staheri14:
How does this PR differ from https://github.com/celestiaorg/celestia-app/pull/2577?

SweeXordious (approved):
utAck 

[Question][not blocking]
Why would we keep the transactions only 5 blocks? Doesn't it make more sense to keep them for longer in case there was a sudden congestion then it was released? So the user won't have to submit the transaction another time

cmwaters:
> How does this PR differ from #2577? Thanks!

It's the same but one is targeting `main` the other targets `v1.x`

cmwaters:
> Why would we keep the transactions only 5 blocks? Doesn't it make more sense to keep them for longer in case there was a sudden congestion then it was released? So the user won't have to submit the transaction another time

I'm not too against having it back at 10. The reason is because it gives users a shorter feedback time where they know if their transaction made it or not and if they need to resubmit it with a higher fee. Right now we don't support submitting with a higher gas fee overriding the previous transaction so users need to wait for their transaction to be kicked out of the mempool before they can resubmit.

I'm curious to see if there is that much variance in load. The good thing with a short TTL is that you shed congestion quicker. either your transaction is paid enough to get in or it's not. 

I know this is a different mindset to ethereum where your transaction may stay in the mempool for hours before being committed on chain

Issue #2464, opened 2023-09-11
Proposal: Add the ability to Downgrade


evan-forbes:
Similar to having the ability to specifiy an increment to the app version in the binary or a config, we should also have the ability to decrement the app version. This might already be mostly possible, but we should add tests if that's the case.

evan-forbes:
to update the most recent thinking around this proposal, the app version should likely only incremented. The main reason being that downgrades would require any state migrations to be backwards compatible, which would require a rather insane amount of testing to have full confidence in. If a specific portion or entirety of an upgrade can be confirmed to be causing a specific regression, then the app version should be incremented to apply or revert those changes.

PR #2583, opened 2023-09-27
feat: allow for hardcoded upgrade schedules

cmwaters:
This PR implements the protocol work behind ADR018. It replaces the existing upgrade module with a custom one which is capable of modifying the app version in EndBlock so that the next block is proposed using the new version

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -226,11 +226,6 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts se
 		cache = store.NewCommitKVStoreCacheManager()
 	}
 
-	skipUpgradeHeights := make(map[int64]bool)
-	for _, h := range cast.ToIntSlice(appOpts.Get(server.FlagUnsafeSkipUpgrades)) {
-		skipUpgradeHeights[int64(h)] = true
-	}
-
 	pruningOpts, err := server.GetPruningOptionsFromFlags(appOpts)
 	if err != nil {
 		panic(err)
@@ -249,10 +244,10 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts se
 	}
 
 	return app.New(
-		logger, db, traceStore, true, skipUpgradeHeights,
-		cast.ToString(appOpts.Get(flags.FlagHome)),
+		logger, db, traceStore, true,
 		cast.ToUint(appOpts.Get(server.FlagInvCheckPeriod)),
 		encoding.MakeConfig(app.ModuleEncodingRegisters...), // Ideally, we would reuse the one created by NewRootCmd.
+		nil,
 		appOpts,
 		baseapp.SetPruning(pruningOpts),
 		baseapp.SetMinGasPrices(cast.ToString(appOpts.Get(server.FlagMinGasPrices))),
@@ -278,13 +273,13 @@ func createAppAndExport(
 	encCfg.Codec = codec.NewProtoCodec(encCfg.InterfaceRegistry)
 	var capp *app.App
 	if height != -1 {
-		capp = app.New(logger, db, traceStore, false, map[int64]bool{}, "", uint(1), encCfg, appOpts)
+		capp = app.New(logger, db, traceStore, false, uint(1), encCfg, nil, appOpts)
 
 		if err := capp.LoadHeight(height); err != nil {
 			return servertypes.ExportedApp{}, err
 		}
 	} else {
-		capp = app.New(logger, db, traceStore, true, map[int64]bool{}, "", uint(1), encCfg, appOpts)
+		capp = app.New(logger, db, traceStore, true, uint(1), encCfg, nil, appOpts)
 	}
 
 	return capp.ExportAppStateAndValidators(forZeroHeight, jailWhiteList)
--- a/x/upgrade/keeper.go
+++ b/x/upgrade/keeper.go
@@ -1,0 +1,89 @@
+package upgrade
+
+import (
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	ibctypes "github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
+)
+
+var _ ibctypes.UpgradeKeeper = (*Keeper)(nil)
+
+type Keeper struct {
+	// we use the same upgrade store key so existing IBC client state can
+	// safely be ported over without any migration
+	storeKey storetypes.StoreKey
+
+	// in memory copy of the upgrade schedule if any. This is local per node
+	// and configured from the config.
+	upgradeSchedule Schedule
+}
+
+// NewKeeper constructs an upgrade keeper
+func NewKeeper(storeKey storetypes.StoreKey, upgradeSchedule Schedule) Keeper {
+	return Keeper{
+		storeKey:        storeKey,
+		upgradeSchedule: upgradeSchedule,
+	}
+}
+
+// ScheduleUpgrade implements the ibc upgrade keeper interface. This is a noop as
+// no other process is allowed to schedule an upgrade but the upgrade keeper itself.
+// This is kept around to support the interface.
+func (k Keeper) ScheduleUpgrade(_ sdk.Context, _ types.Plan) error {
+	return nil
+}
+
+// GetUpgradePlan implements the ibc upgrade keeper interface. This is used in BeginBlock
+// to know when to write the upgraded consensus state. The IBC module needs to sign over
+// the next consensus state to ensure a smooth transition for counterparty chains. This
+// is implemented as a noop. Any IBC breaking change would be invoked by this upgrade module
+// in end blocker.
+func (k Keeper) GetUpgradePlan(_ sdk.Context) (plan types.Plan, havePlan bool) {
+	return types.Plan{}, false
+}
+
+// SetUpgradedClient sets the expected upgraded client for the next version of this chain at the last height the current chain will commit.
+func (k Keeper) SetUpgradedClient(ctx sdk.Context, planHeight int64, bz []byte) error {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.UpgradedClientKey(planHeight), bz)
+	return nil
+}
+
+// GetUpgradedClient gets the expected upgraded client for the next version of this chain
+func (k Keeper) GetUpgradedClient(ctx sdk.Context, height int64) ([]byte, bool) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.UpgradedClientKey(height))
+	if len(bz) == 0 {
+		return nil, false
+	}
+
+	return bz, true
+}
+
+// SetUpgradedConsensusState set the expected upgraded consensus state for the next version of this chain
+// using the last height committed on this chain.
+func (k Keeper) SetUpgradedConsensusState(ctx sdk.Context, planHeight int64, bz []byte) error {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.UpgradedConsStateKey(planHeight), bz)
+	return nil
+}
+
+// GetUpgradedConsensusState set the expected upgraded consensus state for the next version of this chain
+func (k Keeper) GetUpgradedConsensusState(ctx sdk.Context, lastHeight int64) ([]byte, bool) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.UpgradedConsStateKey(lastHeight))
+	if len(bz) == 0 {
+		return nil, false
+	}
+
+	return bz, true
+}
+
+// ClearIBCState clears any planned IBC state
+func (k Keeper) ClearIBCState(ctx sdk.Context, lastHeight int64) {
+	// delete IBC client and consensus state from store if this is IBC plan
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.UpgradedClientKey(lastHeight))
+	store.Delete(types.UpgradedConsStateKey(lastHeight))
+}
--- a/x/upgrade/upgrade.go
+++ b/x/upgrade/upgrade.go
@@ -1,0 +1,34 @@
+package upgrade
+
+type Schedule map[string]map[int64]uint64 // chain-id -> height -> app version
+
+// ShouldUpgradeNextHeight returns true if the network of the given chainID should
+// modify the app version in the following block. This can be used for both upgrading
+// and downgrading. This relies on social consensus to work. At least 2/3+ of the
+// validators must have the same app version and height in their schedule for the
+// upgrade to happen successfully.
+func (k Keeper) ShouldUpgradeNextHeight(chainID string, height int64) bool {
+	if k.upgradeSchedule == nil {
+		return false
+	}
+	if schedule, ok := k.upgradeSchedule[chainID]; ok {
+		if _, ok := schedule[height+1]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+// GetAppVersionForNextHeight returns the app version that the network of the given
+// chainID should modify in the following block. This is 0, if none is provided.
+func (k Keeper) GetAppVersionForNextHeight(chainID string, height int64) uint64 {
+	if k.upgradeSchedule == nil {
+		return 0
+	}
+	if schedule, ok := k.upgradeSchedule[chainID]; ok {
+		if appVersion, ok := schedule[height+1]; ok {
+			return appVersion
+		}
+	}
+	return 0
+}
--- a/app/app.go
+++ b/app/app.go
@@ -8,6 +8,7 @@ import (
 	"github.com/celestiaorg/celestia-app/x/mint"
 	mintkeeper "github.com/celestiaorg/celestia-app/x/mint/keeper"
 	minttypes "github.com/celestiaorg/celestia-app/x/mint/types"
+	"github.com/celestiaorg/celestia-app/x/upgrade"
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/client"
 	nodeservice "github.com/cosmos/cosmos-sdk/client/grpc/node"
@@ -66,8 +67,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/staking"
 	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
-	sdkupgradekeeper "github.com/cosmos/cosmos-sdk/x/upgrade/keeper"
-	sdkupgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 	"github.com/cosmos/ibc-go/v6/modules/apps/transfer"
 	ibctransferkeeper "github.com/cosmos/ibc-go/v6/modules/apps/transfer/keeper"
 	ibctransfertypes "github.com/cosmos/ibc-go/v6/modules/apps/transfer/types"
@@ -93,7 +92,6 @@ import (
 	blobmoduletypes "github.com/celestiaorg/celestia-app/x/blob/types"
 	"github.com/celestiaorg/celestia-app/x/paramfilter"
 	"github.com/celestiaorg/celestia-app/x/tokenfilter"
-	appupgrade "github.com/celestiaorg/celestia-app/x/upgrade"
 
 	qgbmodule "github.com/celestiaorg/celestia-app/x/qgb"
 	qgbmodulekeeper "github.com/celestiaorg/celestia-app/x/qgb/keeper"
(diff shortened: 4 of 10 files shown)


app/deliver_tx.go
```diff
@@ -0,0 +1,23 @@
...
+
+func (app *App) DeliverTx(req abci.RequestDeliverTx) abci.ResponseDeliverTx {
+	sdkTx, err := app.txConfig.TxDecoder()(req.Tx)
+	if err == nil {
+		if appVersion, ok := upgrade.IsUpgradeMsg(sdkTx.GetMsgs()); ok {
+			if !IsSupported(appVersion) {
+				panic(fmt.Sprintf("network has upgraded to version %d which is not supported by this node. Please upgrade and restart", appVersion))
+			}
+			app.SetProtocolVersion(appVersion)
+			// TODO: we may want to emit an event for this
```
evan-forbes, line 18: :+1: https://github.com/celestiaorg/celestia-app/issues/2611

app/process_proposal.go
```diff
@@ -66,19 +67,42 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
...
+				if !IsSupported(appVersion) {
+					logInvalidPropBlock(app.Logger(), req.Header, fmt.Sprintf("block proposes an unsupported app version %d", appVersion))
+					return reject()
+				}
+
+				// app version must always increase
+				if appVersion <= app.GetBaseApp().AppVersion() {
+					logInvalidPropBlock(app.Logger(), req.Header, fmt.Sprintf("block proposes an app version %d that is not greater than the current app version %d", appVersion, app.GetBaseApp().AppVersion()))
+					return reject()
+				}
```
evan-forbes, line 94: [comment/discussion, no changes needed]

imo I think preserving the ability to downgrade in scenarios that don't involve could still be super useful ref https://github.com/celestiaorg/celestia-app/issues/2464

if we wanted to be super careful, we could even require that all upgrades must also be revertable including ones that require migrations. Then we could revert in the advent of an unforseen bug. This would certainly add a testing burden, but potentially worth it.

cmd/celestia-appd/cmd/root.go
```diff
@@ -249,9 +243,6 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts se
 		baseapp.SetTrace(cast.ToBool(appOpts.Get(server.FlagTrace))),
 		baseapp.SetIndexEvents(cast.ToStringSlice(appOpts.Get(server.FlagIndexEvents))),
 		baseapp.SetSnapshot(snapshotStore, snapshottypes.NewSnapshotOptions(cast.ToUint64(appOpts.Get(server.FlagStateSyncSnapshotInterval)), cast.ToUint32(appOpts.Get(server.FlagStateSyncSnapshotKeepRecent)))),
-		func(b *baseapp.BaseApp) {
-			b.SetProtocolVersion(appconsts.LatestVersion)
-		},
```
evan-forbes, line 254: now that this is in init, will it still get called after rebooting? in the past we've had bugs where the chain-id was not set after rebooting because its supposed to be set during init

evan-forbes: [proposal for a future PR, no change needed]

perhaps we should just make this a stateful value? its a one off benefit as it would only apply to v1, however it would guarantee nodes running v1 will hit an apphash error instead of ignoring txs that they don't know about. We'll probably hit one of those anyway, but this just guarantees it asap. The other benefit being that we never have to think about loading a consensus critical value into the application.

app/process_proposal.go
```diff
@@ -66,19 +67,42 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
...
+			if appVersion, ok := upgrade.IsUpgradeMsg(msgs); ok {
+				if idx != 0 {
+					logInvalidPropBlock(app.Logger(), req.Header, fmt.Sprintf("upgrade message %d is not the first transaction", idx))
+					return reject()
+				}
+
+				if !IsSupported(appVersion) {
+					logInvalidPropBlock(app.Logger(), req.Header, fmt.Sprintf("block proposes an unsupported app version %d", appVersion))
+					return reject()
+				}
```
evan-forbes, line 88: is there a scenario where nodes running v1 that don't know how to parse this msg or have this logic vote for the proposal?

app/deliver_tx.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func (app *App) DeliverTx(req abci.RequestDeliverTx) abci.ResponseDeliverTx {
+	sdkTx, err := app.txConfig.TxDecoder()(req.Tx)
+	if err == nil {
+		if appVersion, ok := upgrade.IsUpgradeMsg(sdkTx.GetMsgs()); ok {
+			if !IsSupported(appVersion) {
+				panic(fmt.Sprintf("network has upgraded to version %d which is not supported by this node. Please upgrade and restart", appVersion))
+			}
+			app.SetProtocolVersion(appVersion)
```
evan-forbes, line 17: if this is meant to be done as the first transaction, should we just do this in BeginBlock?

evan-forbes (commented):
Nice, I like this approach and its kinda nice to have a version of signalling from the get go. Have a few questions/discussions for the first round, and I'll continue to review after chewing on it a bit more.

since we're incrementing the protocol version during DeliverTx instead of endblock, will it be weird that we're executing txs for a height using the incremented version, depsite the version for that header not yet being incremented? I could be missing something, and I know that deferred execution will sort of cover this up, but just seemed like a departure from the current system. Before we always used the app version in the header in every scenario, now we use the app version in the header in every scenario, except for the first height of an upgrade.

where do migrations occur? should these also occur during DeliverTx? If so, are we running into the same migration issue that we discussed earlier, where process proposal is running on a potentially different version of the state than when the txs are executed?

app/process_proposal.go
cmwaters: Yeah I thought about this. This was why I chose to wrap it as a tx rather than having it unparseable which would then be accepted. As a tx, it would fail the ante handler because there is no signature attached to it. Ideally I can write a test to check it out.

cmd/celestia-appd/cmd/root.go
cmwaters: > perhaps we should just make this a stateful value?

I agree. This is what later versions of the SDK have. Maybe we can add it in v2 as a migration

app/process_proposal.go
cmwaters: Yeah I thought about downgrades a bit and I think there are a few things we can add to keep the software robust. My concern at the moment is that someone could just propose a downgrade and everyone would accept. 

I think we would need a second signalling mechanism which signals our willingness to accept a downgrade proposal. 

Maybe we can make an issue for this and follow up later before the PR gets way too big

app/prepare_proposal.go
```diff
@@ -58,6 +59,27 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
+		// TODO: this would be improved if we only attempted the upgrade in the first round of the
+		// height to still allow transactions to pass through without being delayed from trying
+		// to coordinate the upgrade height
+		if newVersion, ok := app.UpgradeKeeper.ShouldProposeUpgrade(req.ChainId, req.Height); ok && newVersion > app.GetBaseApp().AppVersion() {
+			upgradeTx, err := upgrade.NewMsgVersionChange(app.txConfig, newVersion)
+			if err != nil {
+				panic(err)
+			}
+			// the upgrade transaction must be the first transaction in the block
+			txs = append([][]byte{upgradeTx}, txs...)
```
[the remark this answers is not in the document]
cmwaters: Yeah I want to add one

app/deliver_tx.go
```diff
@@ -0,0 +1,23 @@
...
+			// TODO: we may want to emit an event for this
+			return abci.ResponseDeliverTx{Code: abci.CodeTypeOK}
+		}
+	}
+	return app.BaseApp.DeliverTx(req)
+}
```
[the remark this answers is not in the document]
cmwaters: > since v1 nodes won't be running this, what will they do with this tx?

They will  just error and skip over it. In the following block they will panic (not sure if in cometbft or in app) because of the app version difference

app/process_proposal.go
evan-forbes: yeah I agree, downgrading (or incrementing the app version that effictively is lower verison) is far outside the scope of this PR.

app/deliver_tx.go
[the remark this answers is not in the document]
MSevey: [non-blocking] I think de-indenting and following an exit early strategy makes the code easier to follow and easier to test.  Additionally I think it reduces the chances of future changes causing bugs.
```suggestion
func (app *App) DeliverTx(req abci.RequestDeliverTx) abci.ResponseDeliverTx {
	sdkTx, err := app.txConfig.TxDecoder()(req.Tx)
	if err != nil {
	        return app.BaseApp.DeliverTx(req)
	}
	if appVersion, ok := upgrade.IsUpgradeMsg(sdkTx.GetMsgs()); !ok {
	        return app.BaseApp.DeliverTx(req)
	}
	if !IsSupported(appVersion) {
		panic(fmt.Sprintf("network has upgraded to version %d which is not supported by this node. Please upgrade and restart", appVersion))
	}
	app.SetProtocolVersion(appVersion)
	// TODO: we may want to emit an event for this
	return abci.ResponseDeliverTx{Code: abci.CodeTypeOK}
}
```

Now it's super clear that there are 4 test cases that need to be created.
1. `err != nil` from `app.txConfig.TxDecoder()(req.Tx)`
2. `!ok` from `upgrade.IsUpgradeMsg(sdkTx.GetMsgs())`
3. `!IsSupported(appVersion)`
4. happy case

x/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,92 @@
+package upgrade_test
```
MSevey, line 1: I'd recommend having unit tests for all the functions as well as this integration test.

app/deliver_tx.go
cmwaters: I think this would be more complex as we would need to track all the different proposed blocks and match them with the one that eventually gets committed.

cmwaters: This will be easier in finalize block because it can reside as part of the `preBlock` call back function

x/upgrade/upgrade_test.go
cmwaters: Yup for sure. Will add all the tests once we're happy with the design

app/version.go
```diff
@@ -0,0 +1,94 @@
...
+		"feegrant":     feegrantmodule.AppModule{}.ConsensusVersion(),
+		"evidence":     evidence.AppModule{}.ConsensusVersion(),
+		"crisis":       crisis.AppModule{}.ConsensusVersion(),
+		"genutil":      genutil.AppModule{}.ConsensusVersion(),
+		"capability":   capability.AppModule{}.ConsensusVersion(),
+		"blob":         blob.AppModule{}.ConsensusVersion(),
+		"qgb":          qgb.AppModule{}.ConsensusVersion(),
+		"ibc":          ibc.AppModule{}.ConsensusVersion(),
+		"transfer":     transfer.AppModule{}.ConsensusVersion(),
+	}
```
cmwaters, line 54: The way we map app versions to module versions will need to be modified later on as it doesn't quite fit to how the SDK are used to doing migrations. I will follow up with this in a later PR but for now this makes the tests pass and somewhat signals how migrations will work

evan-forbes: I can dig it

cmd/celestia-appd/cmd/root.go
```diff
@@ -234,10 +228,10 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts se
 	}
 
 	return app.New(
-		logger, db, traceStore, true, skipUpgradeHeights,
-		cast.ToString(appOpts.Get(flags.FlagHome)),
+		logger, db, traceStore, true,
 		cast.ToUint(appOpts.Get(server.FlagInvCheckPeriod)),
 		encoding.MakeConfig(app.ModuleEncodingRegisters...), // Ideally, we would reuse the one created by NewRootCmd.
+		nil,
```
evan-forbes, line 234: what's your current preferred way to pass this in the future?

x/upgrade/keeper.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return bz, true
+}
+
+// ClearIBCState clears any planned IBC state
+func (k Keeper) ClearIBCState(ctx sdk.Context, lastHeight int64) {
+	// delete IBC client and consensus state from store if this is IBC plan
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.UpgradedClientKey(lastHeight))
+	store.Delete(types.UpgradedConsStateKey(lastHeight))
```
evan-forbes, line 100: when would we call this?

x/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	testApp.BeginBlock(abci.RequestBeginBlock{Header: tmproto.Header{Height: 2}})
+	respDeliverTx := testApp.DeliverTx(abci.RequestDeliverTx{Tx: resp.BlockData.Txs[0]})
+	require.EqualValues(t, 0, respDeliverTx.Code, respDeliverTx.Log)
+	// app version should not have changed yet
+	require.EqualValues(t, 1, testApp.AppVersion())
+	respEndBlock := testApp.EndBlock(abci.RequestEndBlock{Height: 2})
+	// now the app version changes
+	require.EqualValues(t, 2, respEndBlock.ConsensusParamUpdates.Version.AppVersion)
```
evan-forbes, line 125: its not entirely evident where the consensus app version gets updated here, where in the sdk are we upgrading it? I remember running into this issue when we did the bsr upgrade

evan-forbes (approved):
this makes sense to me!! not only is this hardcoded upgrades, but this is "protected" or arguably even a form of signalling! had a few non-blocking questions

evan-forbes:
Certainly not suggesting we change this in this PR, with a more complex signalling mechanism we might want to consider only proceeding with the upgrade if more than some fraction greater than 2/3 signal their readiness.

x/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/celestiaorg/celestia-app/x/upgrade"
+	"github.com/stretchr/testify/require"
+	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
+)
+
+func TestUpgradeAppVersion(t *testing.T) {
+	testApp := setupTestApp(t, upgrade.NewSchedule(upgrade.NewPlan(3, 5, 2)))
```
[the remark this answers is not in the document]
cmwaters: Yeah I can also write an integration test

cmd/celestia-appd/cmd/root.go
cmwaters: I guess the main question I have from your question is whether we want to allow node operators to be able to modify this in their app.toml config. Either way we will have some hardcoded value that gets ported in the binary. I'm just not sure yet whether that value is a default that can be overrided or a built-in value that can only be changed by switching to a different binary

x/upgrade/keeper.go
cmwaters: IBC can call this when an upgrade message is executed. Since we don't allow IBC upgrade proposal messages, this path should never get hit.

cmwaters:
> With a more complex signalling mechanism, we might want to consider only proceeding with the upgrade if more than some fraction greater than 2/3 signal their readiness.

Yeah I also thought about this but theoretically since it only takes 2/3 to approve the block then only 2/3 are necessary to signal the upgrade. What we could do is inject some delay but I don't think it's necessary. 

Another idea is to have a more intelligent downgrade mechanism. If we're not reaching consensus across multiple rounds then a validator should be free to propose a lower version that might get the quorum required. While I've thought a little about downgrades, they haven't yet been properly addressed and should do in the future

app/prepare_proposal.go
```diff
@@ -58,6 +59,27 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
+
+			// because we are adding bytes, we need to check that we are not going over the limit
+			// if we are, we continually prune the last tx (the lowest paying blobTx).
+			size := sizeOf(txs)
+			for size > int(req.BlockDataSize) {
+				lastTx := txs[len(txs)-1]
+				txs = txs[:len(txs)-1]
+				size -= len(lastTx)
+			}
```
rootulp, line 81: [optional] it may be beneficial to include a test that verifies this logic executes as expected.

proto/celestia/upgrade/v1/types.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+package celestia.upgrade.v1;
+
+option go_package = "github.com/celestiaorg/celestia-app/x/upgrade";
+
+// MsgPayForBlobs pays for the inclusion of a blob in the block.
```
rootulp, line 6:
```suggestion
// MsgVersionChange is a message that signals an app version change 
```

proto/celestia/upgrade/v1/types.proto
```diff
@@ -0,0 +1,10 @@
...
+message MsgVersionChange {
+  // the new version to propose the change to
```
rootulp, line 8: [optional]

```suggestion
  // the app version this message proposes upgrading to
```

x/upgrade/keeper.go
```diff
@@ -0,0 +1,101 @@
...
+
+// SetUpgradedConsensusState set the expected upgraded consensus state for the next version of this chain
+// using the last height committed on this chain.
+func (k Keeper) SetUpgradedConsensusState(ctx sdk.Context, planHeight int64, bz []byte) error {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.UpgradedConsStateKey(planHeight), bz)
+	return nil
+}
+
+// GetUpgradedConsensusState set the expected upgraded consensus state for the next version of this chain
```
rootulp, line 84:
```suggestion
// GetUpgradedConsensusState get the expected upgraded consensus state for the next version of this chain
```

x/upgrade/upgrade.go
```diff
@@ -0,0 +1,40 @@
...
+
+type Plan struct {
+	Start   int64
+	End     int64
+	Version uint64
+}
+
+// ShouldUpgradeNextHeight returns true if the network of the given chainID should
+// modify the app version in the following block. This can be used for both upgrading
+// and downgrading. This relies on social consensus to work. At least 2/3+ of the
```
rootulp, line 13: [question] how does this work for downgrading? If the app version must increase each upgrade, I guess a new appVersion could be created that mimics the behavior of a previous app version so it upgrades appVersion but downgrades behavior. Is that "and downgrading" alludes to?

x/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,198 @@
...
+	// rejects the proposed block
+	respProcessProposal := testApp.ProcessProposal(abci.RequestProcessProposal{
+		Header: tmproto.Header{
+			Height: 3,
+		},
+		BlockData: &tmproto.Data{
+			Txs: [][]byte{upgradeTx},
+		},
+	})
+	require.True(t, respProcessProposal.IsRejected())
```
rootulp, line 141: [question] why is this `upgradeTx` proposing a version change that is not supported by the nodes own state machine? It looks like this `upgradeTx` is proposing an upgrade to version 3. 

Is it because the test app is setup with the schedule: `upgrade.NewSchedule(upgrade.NewPlan(3, 5, 2))` which means the app only supports version 1 and 2?

rootulp (dismissed):
LGTM

app/prepare_proposal.go
cmwaters: Yeah you're right. I don't think it will be too difficult to do

x/upgrade/upgrade.go
cmwaters: Downgrading isn't supported yet (I have some ideas I can look to implement in a later PR). 

Currently it would work as you predict that we need a new app version that mimics the behavior of a previous app version

x/upgrade/upgrade_test.go
cmwaters: This is ProcessProposal so it's mimicking receiving a proposal from another node (presumably with v3) capabilities. This is just to check that it will reject the proposal

evan-forbes (approved):

app/prepare_proposal.go
cmwaters: [add test for kicking out tx](https://github.com/celestiaorg/celestia-app/pull/2583/commits/68a7d44afb1933949ecd926e7226fd4a429db7df)

x/upgrade/upgrade_test.go
```diff
@@ -66,6 +79,25 @@ func TestUpgradeAppVersion(t *testing.T) {
...
+			BlockDataSize: 1e2,
+		})
+		require.Len(t, resp.BlockData.Txs, 1)
+		tx, err := testApp.GetTxConfig().TxDecoder()(resp.BlockData.Txs[0])
+		require.NoError(t, err)
+		require.Len(t, tx.GetMsgs(), 1)
+		msg, ok := tx.GetMsgs()[0].(*upgrade.MsgVersionChange)
+		require.True(t, ok)
+		require.EqualValues(t, 2, msg.Version)
+	}
```
rootulp, line 99: 👍

rootulp (approved):

rootulp (approved):

PR #2599, opened 2023-10-02
fix: register qgb msg server in codec

cmwaters:
This PR adds the QGB to the proto registry (noe this is different to the router)

### DIFF
--- a/x/qgb/types/codec.go
+++ b/x/qgb/types/codec.go
@@ -4,6 +4,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
 )
 
 const URLMsgRegisterEVMAddress = "/celestia.blob.v1.MsgRegisterEVMAddress"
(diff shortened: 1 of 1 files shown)


x/qgb/types/codec.go
```diff
@@ -23,4 +24,6 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 		&DataCommitment{},
 		&Valset{},
 	)
+
+	msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)
```
SweeXordious, line 28: [question to understand]
I checked the implementation of this, and it seems to be calling under the hood:

```
registry.RegisterImplementations((*sdk.Msg)(nil),&MsgRegisterEVMAddress{},)
registry.RegisterImplementations((*sdk.Msg)(nil),&MsgRegisterEVMAddressResponse{},)
```

Why do we need to re-register the `MsgRegisterEVMAddress`?
Why do we need to register the `MsgRegisterEVMAddressResponse` even if it's empty?

cmwaters: I just noticed this in passing when comparing with the blob module: 

https://github.com/celestiaorg/celestia-app/blob/f78b9459c8f3db1d85c3e579660e425683b14c70/x/blob/types/codec.go#L32

and sdk modules like bank:

https://github.com/celestiaorg/cosmos-sdk/blob/c525ad72ab1449d31115815f5832a68bc79d4ffd/x/bank/types/codec.go#L32

cmwaters: I think this just helps for service discovery actually

SweeXordious (approved):

PR #2609, opened 2023-10-03
fix(x/tokenfilter): return error OnRecvPacket

### LINKED ISSUE #2488: audit(tokenfilter): Return error OnRecvPacket
### Summary

As part of the TokenFilter module we attempt to unmarshal the incoming `FungibleTokenPacketData` [here](https://github.com/celestiaorg/celestia-app/blob/main/x/tokenfilter/ibc_middleware.go#L44). If we encounter an error we pass the data down the stack (to the transfer module). At present the transfer module only supports this packet type and thus will naturally error as expected but I think a cleaner approach would be to error early and return an error here instead.  

### Problem Definition

This is essentially a nitpick and it's likely the transfer module will not be supporting other packet types (to the best of my knowledge) in the near future but might be good to do to guarantee safety in the future. This is low priority. 

This issue arose as consequence of #1997 

### Proposal

```	
if err := transfertypes.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
        return channeltypes.NewErrorAcknowledgement(Err)
}
```

### For Admin Use

- [ ] Not duplicate issue

ninabarbakadze:
Resolves #2488

### DIFF
--- a/x/tokenfilter/ibc_middleware.go
+++ b/x/tokenfilter/ibc_middleware.go
@@ -45,9 +45,8 @@ func (m *tokenFilterMiddleware) OnRecvPacket(
 		// If this happens either a) a user has crafted an invalid packet, b) a
 		// software developer has connected the middleware to a stack that does
 		// not have a transfer module, or c) the transfer module has been modified
-		// to accept other Packets. The best thing we can do here is pass the packet
-		// on down the stack.
-		return m.IBCModule.OnRecvPacket(ctx, packet, relayer)
+		// to accept other Packets. Short-circuits and returns the error.
+		return channeltypes.NewErrorAcknowledgement(err)
 	}
 
 	// This checks the first channel and port in the denomination path. If it matches


rootulp (approved):

cmwaters:
So personally I don't think this is the correct action as I think it's outside the responsibility of the middleware to make such a validity check as I described in the comments, however I'm not married to this idea if we want to change it.

evan-forbes:
> I don't think this is the correct action as I think it's outside the responsibility of the middleware to make such a validity check as I described in the comments

This is also my hesitation in approving this. I'm not sure we have a suitable enough reason to change this behavior. The work is good @ninabarbakadze, and apologies for not discussing this in the issue before this work was taken on.

ninabarbakadze (commented):
no worries @evan-forbes thanks for the review guys @cmwaters @rootulp! I'm gonna go ahead and close this.

PR #2614, opened 2023-10-05
feat: add ability to check if a QGB attestation is in store

SweeXordious:
## Overview

I will need this to check whether attestations are in store or not, without having to query them directly.

## Checklist


evan-forbes (approved):
seems reasonable

rootulp (approved):
[optional] a unit test that verifies the endpoint works as expected

cmwaters:
Is this just for testing or is there another use case for this? Shouldn't the store maintain all attestations within a range

SweeXordious:
> Is this just for testing or is there another use case for this? Shouldn't the store maintain all attestations within a range

It's actually for the orchestrator to check if an attestation is there without querying (as it will be done subsequently). This would be handy to fix a certain orchestrator issue we've had with Arabica. I am still not merging this until I am sure this is what's needed and not something else.

evan-forbes:
> I am still not merging this until I am sure this is what's needed and not something else.

can we convert to a draft and document/discuss the reasoning behind this in an issue? (we should still be able to merge if nothing needs to be changed)

SweeXordious:
good idea, thanks

PR #2635, opened 2023-10-09
chore: remove unnecessary legacy handler from qgb module

SweeXordious:
## Checklist


x/qgb/module.go
```diff
@@ -114,7 +114,7 @@ func (am AppModule) Name() string {
 
 // Route returns the capability module's message routing key.
 func (am AppModule) Route() sdk.Route {
-	return sdk.NewRoute(types.RouterKey, NewHandler(am.keeper))
+	return sdk.Route{}
```
rootulp, line 117: [noob question] why is this Route method considered legacy? I've found it confusing when Cosmos SDK deprecates things b/c there isn't a consistent practice: some things are labeled "legacy" but not actually deprecated or removed.

SweeXordious: 100% confusing. Well, the rationale is if the msg service is defined, then it will route the transaction correctly. Only if it is unable to that it will call the Legacy handler.  In our case, we have the msg service, so we can delete this one instead of having to maintain almost the same logic in two places

cmwaters: Here's where we perform the if statement just FYI:

https://github.com/celestiaorg/cosmos-sdk/blob/c525ad72ab1449d31115815f5832a68bc79d4ffd/baseapp/baseapp.go#L807

cmwaters: Fwiw they do mention it is deprecated under the interface itself here:

https://github.com/celestiaorg/cosmos-sdk/blob/c525ad72ab1449d31115815f5832a68bc79d4ffd/types/module/module.go#L156

It's just that renaming the method would be breaking so they can only add a comment

cmwaters (approved):
👍

evan-forbes (approved):

Issue #2570, opened 2023-09-22
Make distinct go modules for the `/pkg` directory and celestia-app modules in `/x`


evan-forbes:
In order to help downstream teams avoid dependency hell, we should begin using go modules for some of our packages similar to what most rust projects do.

We should at least do this for the `/pkg` and `/x` directories, but could go further by including a go.mod in each subdirectory of `/x`.

ref https://github.com/celestiaorg/rsmt2d/issues/248 + a bunch of issues in node and rollkit

rootulp:
Note: this should be easier to do for `pkg/shares` because https://github.com/celestiaorg/celestia-app/pull/2659 merged.

rootulp:
This seems close-able because we extracted https://github.com/celestiaorg/go-square . Note: go-square doesn't include the `/x` modules stated in the OP but I think those are less useful for downstream apps compared to the `/pkg` modules. Please re-open or open a new issue if you feel differently @cmwaters @evan-forbes

Issue #1105, opened 2022-12-12
Proposal: Add a byte before each transaction that indicates its type


evan-forbes:
We currently wrap PFB containing transactions with the share index so that other validators and rollups can check for blob inclusion. This is done by wrapping the normal sdk.Tx with a stuct that has the relevant metadata. When decoding transactions, we have to have some mechanism to check that the transaction has the described metadata while we decode the transaction with protobuf. We currently use a [hacky mechnism](https://github.com/celestiaorg/celestia-core/blob/de02bdadc97b54ab7c9a0483aaa8295e6e1e034d/types/tx.go#L211-L230) that forces an additional check during the decoding process. 

Instead of using the hacky mechanism, we should add a single byte before each transaction, which determines the type. Then parsers of transactions can check this byte and know how to parse the rest of the transaction.

rootulp:
Raw notes from sync:
- a user creates and signs a normal transaction. Before publishing their transaction to the mempool, they prefix their transaction with a special byte that indicates the type of the transaction. Something like:

```golang
const (
	IndexWrappedTx = 0x01
	BlobTx         = 0x02
	NormalTx       = 0x03
)
```

- block proposers can tamper with the type byte of a transaction prior to inclusion in a block but the state machine will try to de-serialize a transaction with the incorrect type and this should result in a no-op.
- `CheckTx` should include a switch statement that is based on the first byte of the transaction
- raw pseudocode in this [Go playground](https://go.dev/play/p/ywwmo4Va-3e)

rootulp:
The motivation for this issue was to make it easier for clients to parse PFB transactions. See:

> Then parsers of transactions can check this byte and know how to parse the rest of the transaction.

After a synchronous discussion, we decided to pursue https://github.com/celestiaorg/celestia-app/issues/1173 so we're putting this issue on pause. We can keep this issue open in case we revisit this proposal.

PR #2659, opened 2023-10-11
chore!: move blob protos to app under it's own package

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/issues/2570

This PR creates a `pkg/blob` where the protos for `Blob` and `BlobTx` is created. The intention is that this becomes it's own go.mod to avoid needing to depend on `celestia-core` and our `cosmos-sdk` fork. Other common structures like namespace should do the same so we can achieve some cannonical representation of the structs between node and core/app

The reason for creating the types here instead of in `celestia-core` is so that only `celestia-app` becomes a multi go.mod repo as opposed to either. Additionally, `Blob` and `BlobTx` were never part of the upstream CometBFT.

I also made the decision to have a single representation of blobs rather than a proto and go native. The reason being is one of simplification and performance. With large blobs, copying the data across to another type can lead to a noticeable impact. Having a single struct however does have the dowside of potential coversion errors from uint32 to uint8 so data needs to be validated before it can be used (this is also the same as before).

This PR also ports over the proofs. I have put this all under a common/v1 proto package.


pkg/appconsts/global_consts.go
```diff
@@ -13,16 +14,16 @@ import (
 // They can not change throughout the lifetime of a network.
 const (
 	// NamespaceVersionSize is the size of a namespace version in bytes.
-	NamespaceVersionSize = 1
+	NamespaceVersionSize = ns.NamespaceVersionSize
 	// NamespaceVersionMaxValue is the maximum value a namespace version can be.
 	// This const must be updated if NamespaceVersionSize is changed.
 	NamespaceVersionMaxValue = math.MaxUint8
```
rootulp, line 20: [extremely optional] It seems like this constant could also be exported by `pkg/namespace` and imported here

```suggestion
	// NamespaceVersionMaxValue is the maximum value a namespace version can be.
	// This const must be updated if NamespaceVersionSize is changed.
	NamespaceVersionMaxValue = ns.NamespaceVersionMaxValue
```

rootulp (approved):
👏 this is so clean, great work!

> I also made the decision to have a single representation of blobs rather than a proto and go native

I like this!

[discussion][not blocking] I think it may help contributors if we include a note on why there exists an `x/blob` module and a `pkg/blob` for types. I don't think there's a good place to put this yet but I guess we could add a directory overview section in the README.md. In the absence of that, we could put this note in `x/blob/README.md` or `pkg/blob/README.md` (would need to be created).

cmwaters:
> [discussion][not blocking] I think it may help contributors if we include a note on why there exists an x/blob module and a pkg/blob for types. I don't think there's a good place to put this yet but I guess we could add a directory overview section in the README.md. In the absence of that, we could put this note in x/blob/README.md or pkg/blob/README.md (would need to be created).

Yeah this makes sense. Can follow up in another PR

pkg/blob/blob.go
```diff
@@ -0,0 +1,96 @@
...
+	fmt "fmt"
+	math "math"
+	"sort"
+
+	"github.com/celestiaorg/celestia-app/pkg/namespace"
+)
+
+// ProtoBlobTxTypeID is included in each encoded BlobTx to help prevent
+// decoding binaries that are not actually BlobTxs.
+const ProtoBlobTxTypeID = "BLOB"
```
evan-forbes, line 15: [unrelated to this PR, just a comment no change needed]

I know that we need something like this atm, but there _has_ to be better solution to this

evan-forbes (approved):
I, and probably everyone, really appreciate the refactor to only have a single struct :fire:

pkg/blob/blob.go
evan-forbes: could maybe be solved with a flavor of https://github.com/celestiaorg/celestia-app/issues/1105

PR #2670, opened 2023-10-12
fix: nightly e2e workflow

cmwaters:
The workflow wasn't correctly pulling the latest version. This fixes that. It also updates the default to "latest"


rootulp (approved):
[question] in https://github.com/celestiaorg/celestia-app/actions/runs/6688978782/job/18171816590#step:5:39 why is the `v` getting removed from `version`? Does this PR also fix that or is that a separate issue?

cmwaters:
> [question] in https://github.com/celestiaorg/celestia-app/actions/runs/6688978782/job/18171816590#step:5:39 why is the `v` getting removed from `version`? Does this PR also fix that or is that a separate issue?

Yes this should fix it

SweeXordious (approved):

Issue #2361, opened 2023-08-25
CI workflow for `make test-short`


rootulp:
## Problem

We don't have a CI workflow for `make test-short` so we may accidentally introduce a new test that takes 10+ minutes to run and forget to add the skip in short mode.

## Proposal

Add a CI workflow for `make test-short` that times out if `make test-short` takes longer than 60 seconds.

sayeedkhannabil:
Is it possible to give more clarification on this issue?

tungbq:
Hello, I've proposed the new PR #2367 for this task. Could you please review this?

PR #2673, opened 2023-10-12
ci: separate out PR tests and run fuzzing and race in nightlies

cmwaters:
This PR attempts to improve our CI. It breaks up our test suite into 4 parallel test jobs. It doesn't replicate the work by removing `test-short` and `test-coverage`, combining them into a single job.

This PR also moves our race tests to the nightlies and adds fuzzing to the nightlies

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,5 +1,6 @@
 name: nightly
 on:
+  workflow_dispatch:
   schedule:
     # runs every day at 6am UTC
     - cron: "0 6 * * *"
@@ -26,3 +27,26 @@ jobs:
 
       - name: Run e2e tests
         run: make test-e2e
+
+  race-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Run tests in race mode
+        run: make test-race
+
+  fuzz_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - uses: jidicula/go-fuzz-action@v1.1.0
+        with:
+          packages: './...'
+          fuzz-time: 2m
+          go-version: ${{ env.GO_VERSION }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,69 +7,69 @@ env:
 
 jobs:
   split-test-files:
-  runs-on: ubuntu-latest
-  steps:
-    - uses: actions/checkout@v4
-    - uses: actions/setup-go@v4
-      with:
-        go-version: ${{ env.GO_VERSION }}
-        check-latest: true
-    - name: Create a file with all Celestia pkgs
-      run: go list ./... > pkgs.txt
-    - name: Split pkgs into 4 files
-      run: split -d -n l/4 pkgs.txt pkgs.txt.part.
-    - uses: actions/upload-artifact@v3
-      with:
-        name: "${{ github.sha }}-00"
-        path: ./pkgs.txt.part.00
-    - uses: actions/upload-artifact@v3
-      with:
-        name: "${{ github.sha }}-01"
-        path: ./pkgs.txt.part.01
-    - uses: actions/upload-artifact@v3
-      with:
-        name: "${{ github.sha }}-02"
-        path: ./pkgs.txt.part.02
-    - uses: actions/upload-artifact@v3
-      with:
-        name: "${{ github.sha }}-03"
-        path: ./pkgs.txt.part.03
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+      - name: Create a file with all Celestia pkgs
+        run: go list ./... > pkgs.txt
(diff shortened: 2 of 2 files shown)


cmwaters:
In order for this to work we will need to disable the required tests as they no longer exist @evan-forbes (and require the new tests)

rootulp (commented):
> It breaks up our test suite into 4 parallel test jobs.

Why? In my experience, running 4 parallel jobs isn't faster than running 1 sequential job. See https://github.com/celestiaorg/celestia-app/issues/1439. IMO 1 sequential job is easier to monitor than 4 parallel jobs.

> It doesn't replicate the work by removing test-short and test-coverage, combining them into a single job.

It looks like `make test-short` got dropped from this PR and I think we should retain it so that CI can inform us if someone adds a long running test (i.e. 10+ mins) and forgets to add it to the list of tests skipped in short mode.

Ref: https://github.com/celestiaorg/celestia-app/issues/2361

> This PR also moves our race tests to the nightlies

Why? `make test-race` doesn't appear to be a bottleneck. It takes ~8m to run and test takes ~17m to run. Not blocking on it but it seems like we can run `make test-race` on every PR b/c `make test` takes longer and they run in parallel.

Ref: https://github.com/celestiaorg/celestia-app/actions/runs/6499346269

> and adds fuzzing to the nightlies

👍

cmwaters:
> Why? In my experience, running 4 parallel jobs isn't faster than running 1 sequential job. See https://github.com/celestiaorg/celestia-app/issues/1439. IMO 1 sequential job is easier to monitor than 4 parallel jobs.

I think the advantage is more that you only have to rerun the test split that failed instead of rerunning everything (I think you're right that it doesn't allocate more resources and most tests run in parallel anyway). There actually might be a way where we can cache the results so rerunning doesn't rerun the tests that already passed (I actually don't know how this works)

> It looks like make test-short got dropped from this PR and I think we should retain it so that CI can inform us if someone adds a long running test (i.e. 10+ mins) and forgets to add it to the list of tests skipped in short mode.
> 
> Ref: https://github.com/celestiaorg/celestia-app/issues/2361

I don't think there's a need to run `test-short` if you're going to run test. The problem with the integration tests that needed to be skipped wasn't that they took long but because of the race detector

> Why? make test-race doesn't appear to be a bottleneck. It takes ~8m to run and test takes ~17m to run. Not blocking on it but it seems like we can run make test-race on every PR b/c make test takes longer and they run in parallel.
> 
> Ref: https://github.com/celestiaorg/celestia-app/actions/runs/6499346269

Yes I guess the majority of the time of tests is taken is integration tests. Maybe I can just add this back. The reason for removing it is because race detection for a single threaded application is useless. There is only one place in the entire repo that we spin up a go routine and that is the txsim.


At a high level, I'm just trying to reduce the time our c.i. takes. It would be nice to get everything down to about 5mins. Open to other suggestions

rootulp:
Ahh thanks for pointing out the re-run thing for when test fails. I agree we have a problem with needing to re-run the test task frequently. IMO the problem is caused by frequent test flakes. Proposals that come to mind:
1. Fix test flakes
2. Modify tests to run faster (i.e. remove sleeps or WaitForNextHeights)
3. Auto retry tests if flake is observed. Maybe w/ [retry-step](https://github.com/marketplace/actions/retry-step)

Part of my hesitation is that running the tests in parallel on CI is more difficult to monitor and debug when a test fails. Additionally parallel tests in CI are less representative of how tests are run by contributors during development. So I'm concerned of edge cases where `make test` fails if run sequentially on a contributor's machine but passes on CI b/c they're run in parallel (or vice versa).

> I don't think there's a need to run test-short if you're going to run test. The problem with the integration tests that needed to be skipped wasn't that they took long but because of the race detector

Sorry I don't follow what you mean by integration tests that needed to be skipped... I'm repeating myself but I do think there's a reason to run `make test-short` even if CI runs `make test`. 

**Example**
CI runs `make test` and doesn't run `make test-short`. Developer A adds a test that takes 10+ mins and forgets to add [these lines](https://github.com/celestiaorg/celestia-app/blob/a3311e7e2d88d792f088e5d30cd6d6dc8a623b98/test/txsim/run_test.go#L28-L30) to the test. The PR passes `make test` in CI and merges. Later on, Developer B runs `make test-short` locally, they observe that a long running test was added that forgot those lines and needs to make a separate PR to add them. IMO running `make test-short` on CI with a low timeout (i.e. [1m](https://github.com/celestiaorg/celestia-app/blob/a3311e7e2d88d792f088e5d30cd6d6dc8a623b98/Makefile#L108)) catches this early so that the first PR wouldn't have merged.

> Yes I guess the majority of the time of tests is taken is integration tests. Maybe I can just add this back. The reason for removing it is because race detection for a single threaded application is useless. There is only one place in the entire repo that we spin up a go routine and that is the txsim.

Makes sense. Down to move `make test-race` to nightly given it hasn't really caught any issues for our application b/c as you pointed out, it's single threaded.

evan-forbes:
> In order for this to work we will need to disable the required tests as they no longer exist

cc @MSevey fwiw the only thing maintainers can do is to require merge/squash/rebase, otherwise its identical to write permissions

evan-forbes (approved):
I don't have a strong preference here, so I will defer to the author and second reviewer

we should definitely take the time to fix our tests and refactor them so that the tests that actually need to use comet can but in a robust way and the tests that don't need it don't use it

rootulp (approved):
LMK if you want me to disable required CI checks in order to merge this. I'm personally not a fan of split test files (for reasons mentioned above) but I won't block on it if other maintainers prefer it. 

Before we merge this, it may make sense to verify that all the new CI jobs pass on a fork

cmwaters:
Closing this in favour of #2787 which just focuses on fuzzing

PR #2681, opened 2023-10-16
chore: update deafult mempool size to match ttl and block size

evan-forbes:
## Overview

The default mempool size should match that of the default TTL and block size.

## Checklist

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -225,6 +225,7 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	// version. This acts as a first line of DoS protection
 	upperBoundBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * appconsts.ContinuationSparseShareContentSize
 	cfg.Mempool.MaxTxBytes = upperBoundBytes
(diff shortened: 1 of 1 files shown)


rootulp (dismissed):

app/default_overrides.go
```diff
@@ -225,6 +225,7 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	// version. This acts as a first line of DoS protection
 	upperBoundBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * appconsts.ContinuationSparseShareContentSize
 	cfg.Mempool.MaxTxBytes = upperBoundBytes
+	cfg.Mempool.MaxTxsBytes = appconsts.DefaultMaxBytes * cfg.Mempool.TTLNumBlocks
```
cmwaters, line 228: nit: I think this should probably be the upper bound not the governance upper bound. Reason is if we go from 2MB to 8MB then 10MB can only hold enough for a single block

evan-forbes: makes sense https://github.com/celestiaorg/celestia-app/pull/2681/commits/aff35667f73f03fd7141b79265d64923c2e3b49a

cmwaters (approved):

rootulp (approved):

PR #2610, opened 2023-10-03
refactor: extract variables for common utia and tia amounts

Chirag018:
Closes #2083

PR #2682, opened 2023-10-16
refactor: extract variables for common utia and tia amounts

### LINKED ISSUE #2083: Extract variables for common utia and TIA amounts
## Context

https://github.com/celestiaorg/celestia-app/blob/8b5db4ba20046d9fc7ab7a68b111b157c8fdf877/app/test/std_sdk_test.go#L101

## Problem

It's not immediately obvious how many coins this is

## Proposal

1. Extract variables for common amounts used in this test file
2. Consider refactoring all tests that use arbitrary coin amounts to use the variables extracted in step one

```go
var (
	oneUtia = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))
	// oneTia is 1e6 utia (a.k.a one million utia)
	oneTia = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1e6)))
)
```

Chirag018:
Closes #2083

Reference: https://github.com/celestiaorg/celestia-app/pull/2610 (due to conflict issues, while syncing up, the file got closed, so created a new one with updated changes)

### DIFF
--- a/test/util/genesis/accounts.go
+++ b/test/util/genesis/accounts.go
@@ -20,6 +20,12 @@ type Account struct {
 	InitialTokens int64
 }
 
+var (
+	Utia = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))
+	// Tia is 1e6 utia (a.k.a one million utia)
+	Tia = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1e6)))
+)
+
 func NewAccounts(initBal int64, names ...string) []Account {
 	accounts := make([]Account, len(names))
 	for i, name := range names {
@@ -115,7 +121,7 @@ func (v *Validator) GenTx(ecfg encoding.Config, kr keyring.Keyring, chainID stri
 		return nil, err
 	}
 
-	fee := sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))
+	fee := Utia
 	txBuilder := ecfg.TxConfig.NewTxBuilder()
 	err = txBuilder.SetMsgs(createValMsg)
 	if err != nil {
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/celestiaorg/celestia-app/test/util/blobfactory"
+	"github.com/celestiaorg/celestia-app/test/util/genesis"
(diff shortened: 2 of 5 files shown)


test/util/genesis/accounts.go
```diff
@@ -20,6 +20,12 @@ type Account struct {
 	InitialTokens int64
 }
 
+var (
+	Utia = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1)))
+	// Tia is 1e6 utia (a.k.a one million utia)
+	Tia = sdk.NewCoins(sdk.NewCoin(app.BondDenom, sdk.NewInt(1e6)))
+)
```
rootulp, line 27: [discussion] Since these variables depend on app.BondDenom, I wonder if it makes more sense to export them from app. Potential location after [this block](https://github.com/rootulp/celestia-app/blob/20d2ff2b6dc2a0008535858970cb0f061aeebe89/app/app.go#L101-L110).

Chirag018: Not sure at the moment, but I have opinions on it, that
1. defining Utia and tia variables in accounts.go makes sense as it is the center to use them, and it doesn't make the calling tougher.
2. if we have a look at `app.BondDenom` and as we are trying to export them, that's the only way we have at the moment(from my understanding)(just had surface understanding at the moment).

Chirag018: > [discussion] Since these variables depend on app.BondDenom, I wonder if it makes more sense to export them from app. Potential location after [this block](https://github.com/rootulp/celestia-app/blob/20d2ff2b6dc2a0008535858970cb0f061aeebe89/app/app.go#L101-L110).

Hey, @rootulp sorry for being too silent on this, I was a bit away from things, yup back working on this, I don't think my above comment makes any sense(my bad on it!), but I got the point you raised and found another good way to set the variables in `global_consts.go` file, which makes the overall variable useful and not conflicting anywhere.

And of the point of using the variables that are dependent on app.BondDenom, it also gets solved when we are storing the variables in `global_consts.go` file, by using directly BondDenom instead of app.BondDenom, does that make sense?

rootulp: hmm `global_consts.go` could be a viable location. Do other reviewers have thoughts on this?

rootulp:
Sorry @Chirag018 but it looks like this isn't a high priority issue so let's close this as won't do for now.

Issue #2716, opened 2023-10-18
Override `BaseProposerReward` and `BonusProposerReward` to 0%


rootulp:
## Context

We overrode the distribution module's `BaseProposerReward` and `BonusProposerReward` in https://github.com/celestiaorg/cosmos-sdk/pull/309 to 0%. However, that change was subsequently reverted in https://github.com/celestiaorg/cosmos-sdk/pull/339 b/c the override was made in cosmos-sdk and we pulled upstream changes from cosmos-sdk.

## Problem

We want to set both of these params to `0`. Note they are already `0` in the [specs](https://github.com/rootulp/celestia-app/blob/efcbdeaae3bf41a53e58792d84a68a3421b946e7/specs/src/specs/params.md?plain=1#L32-L33)

## Proposal

Make this change in celestia-app so we don't accidentally revert it again.

PR #2717, opened 2023-10-18
fix: override proposer rewards to 0%

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2716

## Testing

```bash
./scripts/single-node.sh
cat ~/.celestia-app/config/genesis.json
```

### Before

```genesis.json
    "distribution": {
      "params": {
        "community_tax": "0.020000000000000000",
        "base_proposer_reward": "0.010000000000000000",
        "bonus_proposer_reward": "0.040000000000000000",
        "withdraw_addr_enabled": true
      },
```

### After

```genesis.json
    "distribution": {
      "params": {
        "community_tax": "0.020000000000000000",
        "base_proposer_reward": "0.000000000000000000",
        "bonus_proposer_reward": "0.000000000000000000",
        "withdraw_addr_enabled": true
      },
```

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -141,7 +141,7 @@ var (
 		capability.AppModuleBasic{},
 		stakingModule{},
 		mintModule{},
-		distr.AppModuleBasic{},
+		distributionModule{},
 		newGovModule(),
 		params.AppModuleBasic{},
 		crisisModule{},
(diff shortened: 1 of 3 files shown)


rootulp:
Questions for reviewers: 
1. Do we backport to v1.x?
2. Is this conventional commit prefix `feat:` to `fix:`?

evan-forbes (approved):

evan-forbes:
we will have to backport in order to include in a release, which is probably the best idea just to be cautious and use tested code, but it is a good point that we don't technically need to. Whomever wants to create genesis could just use this commit.

using fix makes sense here

cmwaters (approved):

PR #2719, opened 2023-10-18
feat(blobstream): utilizes sdk.Dec type for power difference calculation 

### LINKED ISSUE #2699: Improper Use of Floats in PowerDiff
Context: Precision and performance concerns in financial computations.
Location: [validator.go](https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/types/validator.go#L115)
Issue: The method uses floats.
Remediation: Switch to using `sdk.Dec`

### LINKED ISSUE #2700: Use of Floats in SignificantPowerDiff
Context: Precision issues with floating-point numbers.
Location: [abci.go](https://github.com/celestiaorg/celestia-app/blob/main/x/qgb/abci.go#L18)
Issue: Floats are used for the `SignificantPowerDiff` constant.
Remediation: Transition to `sdk.Dec`.

staheri14:
Closes #2700 
Closes #2699


x/blobstream/types/validator.go
```diff
@@ -112,7 +113,8 @@ func EVMAddrLessThan(e common.Address, o common.Address) bool {
 // increases by 1% due to inflation, we shouldn't have to generate a new
 // validator set, after all the validators retained their relative percentages
 // during inflation and normalized Blobstream power shows no difference.
-func (ibv InternalBridgeValidators) PowerDiff(c InternalBridgeValidators) float64 {
+func (ibv InternalBridgeValidators) PowerDiff(
+	c InternalBridgeValidators) sdk.Dec {
 	powers := map[string]int64{}
```
staheri14, line 118: [note to the reviewers] Alternatively, we could perform all calculations using `sdk.Dec` from the outset. However, since there are no floating-point calculations involved until later in this function, I thought it might be more efficient to continue using integer operations until that point. Nevertheless, if it is preferable to use `sdk.Dec` arithmetic throughout, I can update the initial portion of the function accordingly.

x/blobstream/types/validator.go
```diff
@@ -112,11 +113,12 @@ func EVMAddrLessThan(e common.Address, o common.Address) bool {
...
 	// loop over ibv and initialize the map with their powers
 	for _, bv := range ibv {
-		powers[bv.EVMAddress.Hex()] = int64(bv.Power)
+		powers[bv.EVMAddress.Hex()] = int64(bv.Power) // TODO: overflow?
```
staheri14, line 121: The validator powers are of type `uint64`, and the current conversion could potentially lead to overflow. Upon reviewing the [`validateBasics`](https://github.com/celestiaorg/celestia-app/blob/6683a049b95729c2d055c8e08e3c81d9b7bb7914/x/blobstream/types/validator.go#L57) function, it appears that there is no range check performed on the power value. It would be prudent to confirm whether there might be any validators with powers falling within the range of [`math.MaxInt64`, `math.MaxUint64`]. Such power values could indeed cause an overflow issue in the code line mentioned above. (For example the power value of `math.MaxUint64` becomes `-1` when converted to `int64`)

staheri14:
It appears that this PR also resolves issue #2699. I couldn't locate any `float` calculations in validator.go except for those being handled by this PR. @SweeXordious, could you please confirm whether this is the case? then I'll update the description of PR accordingly.

x/blobstream/types/validator.go
SweeXordious: good catch :+1: But I think if the state machine got to a state where this overflows, then something is seriously messed up with the staking module. In fact, values like those will not be accepted by the staking module, then we normalize those values before this step. I think that we can use a validation at this level as a last line of defense, but reaching it is super unlikely if I understand corerctly

SweeXordious:
> It appears that this PR also resolves issue https://github.com/celestiaorg/celestia-app/issues/2699. I couldn't locate any float calculations in validator.go except for those being handled by this PR. @SweeXordious, could you please confirm whether this is the case? then I'll update the description of PR accordingly.

Apparently yes, this is also handling that issue, please do update. Thanks a lot for helping with this :pray:

x/blobstream/types/validator.go
SweeXordious: I think switching to `sdk.Dec` would be also future-proof in case we wanted to make any changes not to give to future devs the ability to choose. So, I would go with switching everything to `sdk.Dec`

SweeXordious (commented):
Awesome stuff :+1: 

I will approve after we switch all arithmetic operations to using `sdk.Dec`.

x/blobstream/types/validator.go
staheri14: > values like those will not be accepted by the staking module

So, does it mean that there will never be a case in which a validator's power reside in this range [math.MaxInt64, math.MaxUint64]? and it is guaranteed by the staking module?

SweeXordious: yes, the validator power is using `int64`, so no overflow would happen. However, I found something interesting:
https://github.com/celestiaorg/cosmos-sdk/blob/main/x/staking/types/validator.go#L369-L372

```
// PotentialConsensusPower returns the potential consensus-engine power.
func (v Validator) PotentialConsensusPower(r math.Int) int64 {
	return sdk.TokensToConsensusPower(v.Tokens, r)
}
```

The power is being reduced by a factor of:
```
// DefaultPowerReduction is the default amount of staking tokens required for 1 unit of consensus-engine power
var DefaultPowerReduction = NewIntFromUint64(1000000)
```

However, the tokens amount is represented by an int256... So, there might be overflow there when going from tokens to consensus-engine power unless there is the assumption that the staked tokens wouldn't be that much.

cc @evan-forbes any idea?

evan-forbes: sorry mind pointing out where's the int256 to save me a sec :sweat_smile: 

this is good to keep in the back of our heads, especially for networks such as arabica where the supply and voting power is getting rather insane. for normal networks tho the assumption of staying within the max and min values appears safe afaiu (please correct me if I'm wrong I didn't deep dive the code yet)

evan-forbes:
> I think switching to sdk.Dec would be also future-proof in case we wanted to make any changes not to give to future devs the ability to choose. So, I would go with switching everything to sdk.Dec

imo I think we should do this, but in a different issue/bump. We still need to test this on IRL networks to make sure that we're doing this in a non consensus breaking way or reevaluate and the simpler we can make that process the better

x/blobstream/types/validator.go
```diff
@@ -135,7 +137,11 @@ func (ibv InternalBridgeValidators) PowerDiff(c InternalBridgeValidators) float6
 		delta += math.Abs(float64(v))
 	}
 
-	return math.Abs(delta / float64(math.MaxUint32))
+	decDelta := sdk.NewDec(int64(delta))
```
evan-forbes, line 140: could we leave delta a int64 the entire time?

evan-forbes (dismissed):
one quick question to remove another instance of floats right above these, but still not blocking on it

evan-forbes:
before merging the backported version of this, we should sync from scratch and ideally perform other tests to ensure that this is not consensus breaking. If it is, then we can reevaluate our options from there

x/blobstream/types/validator.go
SweeXordious: In the [`validator`](https://github.com/celestiaorg/cosmos-sdk/blob/4c897c6e15c7c0b956ca9fa25d0625ff7a39553a/x/staking/types/staking.pb.go#L314-L339) struct, the delegated tokens are represented using [`Tokens`](https://github.com/celestiaorg/cosmos-sdk/blob/4c897c6e15c7c0b956ca9fa25d0625ff7a39553a/x/staking/types/staking.pb.go#L324C26-L324C26) which are [`Int`](https://github.com/celestiaorg/cosmos-sdk/blob/4c897c6e15c7c0b956ca9fa25d0625ff7a39553a/types/math.go#L14)s which are [`sdkmath.Int`](https://github.com/celestiaorg/cosmos-sdk/blob/4c897c6e15c7c0b956ca9fa25d0625ff7a39553a/math/int.go#L72-L77). 

Then, these are used when calculating the validator delegation and power in [here](https://github.com/celestiaorg/cosmos-sdk/blob/4c897c6e15c7c0b956ca9fa25d0625ff7a39553a/x/staking/keeper/val_state_change.go#L169) after getting the `consensus-engine power` which would [panic](https://github.com/celestiaorg/cosmos-sdk/blob/4c897c6e15c7c0b956ca9fa25d0625ff7a39553a/math/int.go#L156-L161) when converting from `sdkmath.Int` to `int64` if the amount is out of bounds.

x/blobstream/types/validator.go
staheri14: @evan-forbes Is the concern related to the `float64` conversion in the earlier lines? If so, then the short answer is yes, we can keep `delta` as `int64` the entire time.

The reason `delta` has been defined as `float64` is because `math.Abs()` accepts and outputs `float64` values. 

If the intention is to keep `delta` as an `int64`, there are two alternative approaches to achieve that:

1. Maintain `delta` as an `int64`, but continue using `math.Abs()` by converting individual powers to `float64`:
```go
var delta int64
for _, v := range powers {
    // NOTE: We are interested in the absolute value of the changes
    delta += int64(math.Abs(float64(v)))
}
```

2. Completely avoid the `float64` conversion of the powers by not using `math.Abs()`. However, please be aware that this approach may have undefined behavior in handling corner cases like when `v` is equal to ±Inf or NaN (which `math.Abs()` can handle: [Link](https://pkg.go.dev/math#Abs)):
```go
var delta int64
for _, v := range powers {
    // NOTE: We are interested in the absolute value of the changes
    if v < 0 {
        v = -v
    }
    delta += int64(v)
}
```
I lean towards option 2, only if the corner cases I mentioned are not conceivable. 
cc: @SweeXordious

staheri14:
> > I think switching to sdk.Dec would be also future-proof in case we wanted to make any changes not to give to future devs the ability to choose. So, I would go with switching everything to sdk.Dec

> imo I think we should do this, but in a different issue/bump. We still need to test this on IRL networks to make sure that we're doing this in a non consensus breaking way or reevaluate and the simpler we can make that process the better


Considering this comment of @evan-forbes, do we still want to refactor the initial section of the `PowerDiff` function to utilize `sdk.Dec` @SweeXordious in this PR? In my opinion, since the initial section of this function exclusively involves integer operations, converting it to utilize `sdk.Dec` arithmetic wouldn't enhance computational precision; instead, it would primarily improve code readability and clarity. 
PS: I have already started updating the first portion of the `PowerDiff` and it is in progress; there is one intricacy with the `Sub` method of sdk.Dec type, which converts zero values to `nil` which then causes the rest of computations to panic (and I am debugging and trying to find a fix). I can open a follow-up PR  to refactor the first portion of `PowerDiff` function, if there is an agreement on whether refactoring is needed or not. 

>  I think switching to sdk.Dec would be also future-proof in case we wanted to make any changes not to give to future devs the ability to choose.

I'm unsure how this change would effectively enforce any constraints on future developers when the function signature remains unchanged, and only its internals  are to be modified to utilize `sdk.Dec`. Could you please provide more insight into your perspective?

x/blobstream/types/validator.go
staheri14: Thanks for your comment! 
I'm unsure how this change would effectively enforce any constraints on future developers when the function signature remains unchanged, and only its internals are to be modified to utilize `sdk.Dec`. Could you please provide more insight into your perspective?

x/blobstream/types/validator.go
evan-forbes: yeah option 2 is what I think would be preferable

evan-forbes:
> In my opinion, since the initial section of this function exclusively involves integer operations, converting it to utilize sdk.Dec arithmetic wouldn't enhance computational precision; instead, it would primarily improve code readability and clarity.

my understanding was that this would potentially stop an overflow with large stake? correct me if I'm wrong @SweeXordious

reguardless I think we should merge this or some version of it sooner rather than later as ideally this would be included in the version that is used for mainnet

SweeXordious:
> my understanding was that this would potentially stop an overflow with large stake? correct me if I'm wrong @SweeXordious

Yes, that's the idea. But after investigating the cosmos-sdk implementation, this would fail in the staking module prior to reaching this level, since they're not handling the out of bounds case. source: https://github.com/celestiaorg/celestia-app/pull/2719#discussion_r1365261371

However, I agree that we should ship some version of this to be included in the upcoming release so that this change is also audited.

> Completely avoid the float64 conversion of the powers by not using math.Abs(). However, please be aware that this approach may have undefined behavior in handling corner cases like when v is equal to ±Inf or NaN (which math.Abs() can handle: [Link](https://pkg.go.dev/math#Abs)):

I also agree with going more with option 2. And we could even add explicit checks to see if `v` is out of bounds or `NaN` if that's possible.

Thanks a lot @staheri14 for helping with this stuff :pray:

staheri14:
I am working on option 2, will soon commit my changes.

x/blobstream/types/validator.go
```diff
@@ -129,13 +131,35 @@ func (ibv InternalBridgeValidators) PowerDiff(c InternalBridgeValidators) float6
...
+
+// Int64Abs returns the absolute value of an int64.
+// For math.MinInt64, it returns math.MaxInt64.
+// This is because the absolute value of math.inInt64 is
+// 1 greater than math.MaxInt64. So, we return math.MaxInt64 as the closest
+// approximation which has 1 bit of error.
+func Int64Abs(v int64) int64 {
+	if v < 0 {
+		if v == math.MinInt64 { // avoid overflow
+			return -(v + 1) // i.e., math.MaxInt64
```
staheri14, line 158: [Note to reviwers @evan-forbes @SweeXordious ] This is a tricky and problematic corner case, without this line, the negative of `-9223372036854775808` (math,MinInt64) equals itself but not `9223372036854775808`. The reason is explained in the function description.


Upon further thoughts, another issue came to my mind: if the sum of the absolute power diffs exceeds the maximum value representable by `int64` then there will be overflow issues (do we expect this happen?). This is while using `float64` it is less likely since the largest value representable by `float64` is much larger than the max value that an `int64` can accomodate. So, we may want to reconsider the original version in which delta was `float64` and we have been using math.Abs() instead. Please let me know your thoughts.

staheri14: Please note that float64 calculations also introduce their own errors, as illustrated below:

```go
v1 := math.MinInt64 // -9223372036854775808
v2 := float64(math.MinInt64) // -9223372036854776000
```

When examining the values of `v1` and `v2`, it becomes evident that the `float64` representation of `v1` does not precisely match its `int64` value.

evan-forbes: I don't think we have anything to worry about as long as we use normal values (most testnets have 1*10^15 total utia and 1*10^9 voting power), this is just being extra careful. 

I'm definitely okay with shipping and testing smaller changes, and was incorrect to assume that staying with int64 was a straightforward switch, will defer to the author on that

evan-forbes: I'm tempted to just use big.Int for everything

SweeXordious: Hum, yes `big.Int` or we just use `sdk.Dec` wrapper for everything, even the map of powers created in `powers := map[string]int64{}`. That would help us avoid all of these conversions but not sure about the consequences of it in functions calling the `PowerDiff`.

evan-forbes (approved):

x/blobstream/types/validator.go
staheri14: In the setting you mentioned i.e.,
> (most testnets have 1x10^15 total utia and 1x10^9 voting power)

it should be fine to use `int64` to hold the sum of power diffs, but if in Arabica it can get larger and the upper bound is not known yet, then I'd vote for keeping delta as `float64` just to avoid potential overflows.

@SweeXordious Once you mentioned powers are normalized? to what range they are normalized to? What is the max power value after normalization? Basically, what is the max power value that `PowerDiff` function can receive as its input? Is the sum of powers capped?

> I'm tempted to just use big.Int for everything

This or `sdk.Dec` which underneath is a `big.Int`.

SweeXordious: > @SweeXordious Once you mentioned powers are normalized, to what range they are normalized to? What is the max power value after normalization? Basically, what is the max power value that PowerDiff function can receive as its input? Is the sum of powers capped?

They're normalized to `2^32`:
https://github.com/celestiaorg/celestia-app/blob/fc348bf395e0c56787aa0fe587275adf2f04c40b/x/blobstream/keeper/keeper_valset.go#L136-L161

staheri14: Thank you for providing the context. While I believe that using `int64` with this normalization would have also been functional, I have since successfully refactored the `PowerDiff` function to utilize `sdk.Dec` from the outset (managed to resolve the bug I mentioned earlier). Regarding the code review, everything seems to be in order, except for the potential overflow issue when casting from `uint64` to `int64`; we have partially discussed it, however, it is out of scope of this PR and merits its own separate PR.

x/blobstream/types/validator.go
```diff
@@ -112,30 +114,35 @@ func EVMAddrLessThan(e common.Address, o common.Address) bool {
...
 // validator set, after all the validators retained their relative percentages
 // during inflation and normalized Blobstream power shows no difference.
-func (ibv InternalBridgeValidators) PowerDiff(c InternalBridgeValidators) float64 {
-	powers := map[string]int64{}
+func (ibv InternalBridgeValidators) PowerDiff(c InternalBridgeValidators) sdk.Dec {
+	powers := map[string]sdk.Dec{}
 	// loop over ibv and initialize the map with their powers
 	for _, bv := range ibv {
-		powers[bv.EVMAddress.Hex()] = int64(bv.Power)
+		powers[bv.EVMAddress.Hex()] = sdk.NewDec(int64(bv.Power))
```
SweeXordious, line 121: we can avoid this unsafe conversion using `sdk.NewDecFromBigInt(new(big.Int).SetUint64(bv.Power))`

staheri14: Thanks, done https://github.com/celestiaorg/celestia-app/pull/2719/commits/efae96232a7187b32179e37e298fddb4348b15b1.

.github/workflows/ci-release.yml
```diff
@@ -50,7 +50,7 @@ jobs:
   # If this was a workflow dispatch event, we need to generate and push a tag
   # for goreleaser to grab
   version_bump:
-    needs: [lint, markdown-linter, test]
+    # needs: [lint, markdown-linter, test]
```
SweeXordious, line 53:
```suggestion
    needs: [lint, markdown-linter, test]
```
probably?

.goreleaser.yaml
```diff
@@ -16,6 +16,8 @@ builds:
   - main: ./cmd/celestia-appd
     binary: celestia-appd
     env:
+      # Enable CGO for Ledger support
+      - CGO_ENABLED=1
```
SweeXordious, line 20:
```suggestion
```
I guess some other PR got confused into this

SweeXordious (commented):
awesome stuff :+1: :+1: 

just apply the comments and will approve

.github/workflows/ci-release.yml
staheri14: I did not make this change!

SweeXordious: This is funny :D :D

Probably wrong PR was targetted: https://github.com/celestiaorg/celestia-app/pull/2719/commits/21f244af1be156ba9891c21ff443759121b0f5ae

cc  @rootulp

x/blobstream/types/validator.go
staheri14: Actually this commit is the correct one https://github.com/celestiaorg/celestia-app/pull/2719/commits/bddfcfa3823718def693ee3293bb57183d0512d7

.github/workflows/ci-release.yml
staheri14: I merged a wrong branch into my branch, fixed it now!

x/blobstream/types/validator.go
```diff
@@ -112,30 +115,35 @@ func EVMAddrLessThan(e common.Address, o common.Address) bool {
...
 	// loop over ibv and initialize the map with their powers
 	for _, bv := range ibv {
-		powers[bv.EVMAddress.Hex()] = int64(bv.Power)
+		powers[bv.EVMAddress.Hex()] = sdk.NewDecFromBigInt(new(big.Int).SetUint64(bv.Power))
 	}
 
 	// subtract c powers from powers in the map, initializing
 	// uninitialized keys with negative numbers
 	for _, bv := range c {
+		bvPower := sdk.NewDec(int64(bv.Power))
```
SweeXordious, line 128:
```suggestion
		bvPower := sdk.NewDecFromBigInt(new(big.Int).SetUint64(bv.Power))
```

.goreleaser.yaml
staheri14: Indeed, I merged a wrong branch, though I fixed it!

x/blobstream/types/validator.go
staheri14: Done now https://github.com/celestiaorg/celestia-app/pull/2719/commits/d3c7ae0a54fd68413d3ad8072aa8dbd371df8f97

SweeXordious (approved):
:shipit:

evan-forbes (approved):

PR #2723, opened 2023-10-19
fix(malicious): use constructor NMT options

rootulp:
Reported by @vgonkivs who is working on a BEFP test and needs to pass `nmt.NodeVisitor()` in to the malicious package's tree.

### DIFF
--- a/test/util/malicious/tree.go
+++ b/test/util/malicious/tree.go
@@ -50,6 +50,7 @@ func (c constructor) NewTree(_ rsmt2d.Axis, axisIndex uint) rsmt2d.Tree {
 		nmt.NamespaceIDSize(appconsts.NamespaceSize),
 		nmt.IgnoreMaxNamespace(true),
 	}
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):
idk if celestia node is intending to use this package anymore, if not we should definitely delete it to stop maintaining it and can add it back if we ever need to commit over invalid blocks

vgonkivs:
I have added a test that checks out-of-order case, so, yeah, we need this package 🙂

rootulp:
@vgonkivs I assume you'll also need this in a v1.x release b/c AFAIK we don't have plans to cut a release from main anytime soon and celestia-node needs _some_ release to depend on.

We intend on only backporting things that are user-impacting but I think we need to make an exception for this PR b/c celestia-node is our consumer

evan-forbes:
celestia-node counts as users imo so yeah we should backport

staheri14 (approved):

Issue #2729, opened 2023-10-19
Ledger unsupported on pre-built binaries 


rootulp:
## Context

Reported by @adlerjohn 

## Problem

It appears the pre-built binaries for 1.1.0 on mac don't work with Ledger:
```
Default sign-mode 'direct' not supported by Ledger, using sign-mode 'amino-json'.
Error: ledger nano S: support for ledger devices is not available in this executable
```

## Proposal

Likely something to do with CGO. See [this comment](https://github.com/celestiaorg/celestia-app/blob/fc348bf395e0c56787aa0fe587275adf2f04c40b/.goreleaser.yaml#L4-L10).

Consider adding this to goreleaser.yml

```yaml
    env:
      - CGO_ENABLED=1
```

Ref: https://goreleaser.com/cookbooks/cgo-and-crosscompiling/?h=cgo

PR #2730, opened 2023-10-19
fix: Ledger support in pre-built binaries

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2729


rootulp:
Nvm this won't work because https://github.com/rootulp/celestia-app/actions/runs/6581110152/job/17880442212#step:6:42

PR #2744, opened 2023-10-20
ci: add support for go vulncheck

SweeXordious:
## Checklist

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -74,4 +74,6 @@ jobs:
 
   govulnech:
     name: Golang Vulncheck
-    uses: Templum/govulncheck-action@v1.0.0
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: Templum/govulncheck-action@v1.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -69,3 +71,26 @@ jobs:
         uses: github/codeql-action/analyze@v2
         with:
           category: "/language:${{matrix.language}}"
+
+  govulnech:
```
rootulp, line 75: [typo]

```suggestion
  govulncheck:
```

.github/workflows/static-analysis.yml
```diff
@@ -9,9 +9,11 @@
 # the `language` matrix defined below to confirm you have the correct set of
 # supported CodeQL languages.
 #
```
rootulp, line 11: [optional] if we repurpose this file, IMO we can remove the comment boilerplate from the top of the file.

Alternatively: we keep the boilerplate and move govulncheck to a new file.

SweeXordious: sure, good idea :+1: 

Originally, I opened this PR to test vulncheck and see what the results it will give and if it's worth it to add it next to codeQL

rootulp: I expect go vulncheck will become a best practice for repos so I think we should add it.

rootulp:
Can close b/c https://github.com/celestiaorg/celestia-app/pull/2969 merged

PR #2745, opened 2023-10-23
docs: ADR-022 Streamline system architecture

cmwaters:
[Rendered](https://github.com/celestiaorg/celestia-app/blob/cal/merge-adr/docs/architecture/adr-022-system-merge.md)


docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+  - Duplicate storage of squares
+  - Redundant computation of the extended data square from the block
+  - Extra round trip in retrieving the data and pushing it (latency costs).
+- A secondary implementation of a storage engine which largely overlaps with the KV based storage engine in the consensus node.
+- A secondary implementation for exchanging and verifying data: `go-header`
+- Two binaries that have different usage patterns: commands, config files, flags etc. This makes it more confusing for users.
+
+Having duplication requires both a broader knowledge base and more maintenance load for engineers. There is a larger surface area for bugs and gains from optimizations are not captured as efficiently as they could.
+
+Outside of the aforementioned tech debt, there is untapped potential in taking the benefits of LibP2P and porting it across to the Tendermint consensus algorithm. LibP2P has a more sophisticated and efficient gossiping protocols than Tendermint's. It also has a more robust peer discovery and peer management system.
```
rootulp, line 31: [grammar] 

```suggestion
Outside of the aforementioned tech debt, there is untapped potential in taking the benefits of LibP2P and porting it across to the Tendermint consensus algorithm. LibP2P has a more sophisticated and efficient gossiping protocol than Tendermint's. It also has a more robust peer discovery and peer management system.
```

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+- Phase 3 will likely coincide with bandwidth and latency improvements to both the consensus and mempool protocols such as: [Compact Blocks](https://github.com/celestiaorg/celestia-core/issues/883), Tx Client, Intermediate Blocks, Transaction Sharding (i.e. Narwhal), [BlobTx Pipe](https://github.com/celestiaorg/celestia-app/issues/2297), and Consensus Pipelining. Statesync will likely remain unchanged and be the easiest to support. Giving rolling upgrades, there will be one version that will support both Tendermint P2P and LibP2P for all components.
+
+### Risk Mitigation
+
+Replacing the networking layer is a significant undertaking and vulnerable to a number of risks. Each component will need to be benchmared to ensure no performance degradation and be sufficiently stress tested. Having both versions present, while adding complexity, has the advantage of enabling the option to fallback in case of unexpected failure.
```
rootulp, line 63:
```suggestion
Replacing the networking layer is a significant undertaking and vulnerable to a number of risks. Each component will need to be benchmarked to ensure no performance degradation and be sufficiently stress tested. Having both versions present, while adding complexity, has the advantage of enabling the option to fallback in case of unexpected failure.
```

rootulp (approved):

docs/architecture/adr-022-system-merge.md
Wondertan: As well as support for QUIC, which is the point of interest for Tendermint

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+- A secondary implementation for exchanging and verifying data: `go-header`
+- Two binaries that have different usage patterns: commands, config files, flags etc. This makes it more confusing for users.
+
+Having duplication requires both a broader knowledge base and more maintenance load for engineers. There is a larger surface area for bugs and gains from optimizations are not captured as efficiently as they could.
+
+Outside of the aforementioned tech debt, there is untapped potential in taking the benefits of LibP2P and porting it across to the Tendermint consensus algorithm. LibP2P has a more sophisticated and efficient gossiping protocols than Tendermint's. It also has a more robust peer discovery and peer management system.
+
+## Decision
+
+The decision is to converge to a LibP2P based network and to shape out reusable components that can be used by all node types: consensus, data and light. This is to be gradually done across four phases, each phase likely requiring it's own in-depth ADR:
```
Wondertan, line 35: Do you mean full?

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Phase 1: Removal of the bridge node
+
+The first step involves porting `Init` and `Start` methods from `celestia-app` and allowing consensus nodes to be initialized and run through the `celestia` binary. This will allow for a arguemnt to be passed in `Start` that enables the consensus nodes to directly push the extended data square (EDS) that is finalized in `FinalizeBlock` to the LibP2P based networks. Thus consensus nodes will temporarily be a hybrid of both networking systems and replace the need for a separate bridge node. Having a single binary will greatly simplify both the release process and end-to-end testing.
```
Wondertan, line 39:
```suggestion
The first step involves porting `Init` and `Start` methods from `celestia-app` and allowing consensus nodes to be initialized and run through the `celestia` binary. This will allow for a argument to be passed in `Start` that enables the consensus nodes to directly push the extended data square (EDS) that is finalized in `FinalizeBlock` to the LibP2P based networks. Thus consensus nodes will temporarily be a hybrid of both networking systems and replace the need for a separate bridge node. Having a single binary will greatly simplify both the release process and end-to-end testing.
```

Wondertan, line 39: I don't think it's worth doing this step without unifying storage.

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Phase 2: Migration of storage, syncing and fraud
+
+The next step migrates first the storage engine and then the syncing and fraud protocols (`go-header` and `go-fraud`) to `celestia-app`. This replaces the existing `BlockStore` and blocksync and evidence reactors. Streamlining the storage will require careful migration but will eliminate the duplication of storage and unify many of the data exchange protocols and the public APIs. Here, the consensus node can become more integrated with the `nodebuilder` package as it begins to be split up into a series of modules. This phase may also try to align configuration files and the general file directory.
```
Wondertan, line 43: From what we discussed, we wanted the syncing and storage will be backed by Node's API. Something like Put/GetEDS, which Consensus uses to write and read edses, respectively. Similarly, we would provide getters and putters for Headers over API. I don't yet see how consensus would work with go-header directly. Also the API to trim EDS to only partial namespace would be there,

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Phase 3: Rewiring consensus, txpool and statesync
+
+The third phase consists of replacing the interface between the consensus, mempool and statesync protocols to Tendermint's p2p layer with LibP2P. This can be done in tandem with phase 2 and may be coupled with other refactors and performance improvements to those components. This will manifest itself as four new modules to `celestia-node`: consensus, mempool, statesync, and execution (a standalone Celestia state machine. This is somewhat similar to `app.App`).
```
Wondertan, line 47: > This can be done in tandem with phase 2

We shouldn't do it in tandem with phase 2. Migrating storage and networking at the same time would be super risky and complicated. I see this as two separate step.

docs/architecture/adr-022-system-merge.md
cmwaters: I disagree but can you first explain why?

docs/architecture/adr-022-system-merge.md
cmwaters: The work can be done in tandem (as in they don't block each other). But this would be incorporated in later releases

docs/architecture/adr-022-system-merge.md
cmwaters: > I don't yet see how consensus would work with go-header directly

Why not? go-header, afaiu, will allow the consensus node to verify a series of headers. This is similar to the mechanism behind blocksync

docs/architecture/adr-022-system-merge.md
cmwaters: yes full

evan-forbes (approved):
great summary and start. all of the core protocol devs should read this!

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Decision
+
+The decision is to converge to a LibP2P based network and to shape out reusable components that can be used by all node types: consensus, data and light. This is to be gradually done across four phases, each phase likely requiring it's own in-depth ADR:
+
+### Phase 1: Removal of the bridge node
+
+The first step involves porting `Init` and `Start` methods from `celestia-app` and allowing consensus nodes to be initialized and run through the `celestia` binary. This will allow for a argument to be passed in `Start` that enables the consensus nodes to directly push the extended data square (EDS) that is finalized in `FinalizeBlock` to the LibP2P based networks. Thus consensus nodes will temporarily be a hybrid of both networking systems and replace the need for a separate bridge node. Having a single binary will greatly simplify both the release process and end-to-end testing.
```
Wondertan, line 39: This should mention that one of the goals is to compute EDS only once

Bidon15, line 39: does it mean that we just have a binary merge but the block storage + configs stay the same for both app node and ex-bridge node?

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Phase 2: Migration of storage, syncing and fraud
```
Bidon15, line 41: Will Phase 2 overhaul the general API design for storage that unlocks the potential to have a remoteDB? this helps reduce the storage footprint a lot for node operators

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Phase 3: Rewiring consensus, txpool and statesync
+
+The third phase consists of replacing the interface between the consensus, mempool and statesync protocols to Tendermint's p2p layer with LibP2P. This can be done in tandem with phase 2 and may be coupled with other refactors and performance improvements to those components. This will manifest itself as four new modules to `celestia-node`: consensus, mempool, statesync, and execution (a standalone Celestia state machine. This is somewhat similar to `app.App`).
+
+### Phase 4: Unification of Public APIs
+
+Lastly, legacy APIs like the RPC layer will be deprecated in favour of a unified public API. Every node should offer the same collection of endpoints and be capable of verifying all retrieved data.
+
+## Considerations
```
Bidon15, line 53: Curious to know, does every phase mean a dedicated ADR with more details?

Bidon15 (commented):
Awesome! Great doc 🥳

docs/architecture/adr-022-system-merge.md
Wondertan, line 41:
```suggestion
### Phase 2: Migration of storage, syncing networking and fraud
```

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+The next step migrates first the storage engine and then the syncing and fraud protocols (`go-header` and `go-fraud`) to `celestia-app`. This replaces the existing `BlockStore` and blocksync and evidence reactors. Streamlining the storage will require careful migration but will eliminate the duplication of storage and unify many of the data exchange protocols and the public APIs. Here, the consensus node can become more integrated with the `nodebuilder` package as it begins to be split up into a series of modules. This phase may also try to align configuration files and the general file directory.
+
+### Phase 3: Rewiring consensus, txpool and statesync
+
+The third phase consists of replacing the interface between the consensus, mempool and statesync protocols to Tendermint's p2p layer with LibP2P. This can be done in tandem with phase 2 and may be coupled with other refactors and performance improvements to those components. This will manifest itself as four new modules to `celestia-node`: consensus, mempool, statesync, and execution (a standalone Celestia state machine. This is somewhat similar to `app.App`).
+
+### Phase 4: Unification of Public APIs
```
Wondertan, line 49: The order between 3 and 4 may change depending on priorities

docs/architecture/adr-022-system-merge.md
cmwaters: Yes there is one binary but the file directory system and configs remain as they were

docs/architecture/adr-022-system-merge.md
cmwaters: It deduplicates the storage of blocks (ods + header). I don't think it modifies the API itself. What ideas do you have in mind?

docs/architecture/adr-022-system-merge.md
cmwaters: Yes, this is a high level ADR. Each of these phases will require a more dedicated description of the design and their tradeoffs

docs/architecture/adr-022-system-merge.md
cmwaters: Yeah we could do

docs/architecture/adr-022-system-merge.md
Wondertan: > Will Phase 2 overhaul the general API design for storage that unlocks the potential to have a remoteDB?

Yes. The node will be capable of remote db, and if the app migrates to it will also be able to support it

MSevey:
Great write up! 👏

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+Lastly, legacy APIs like the RPC layer will be deprecated in favour of a unified public API. Every node should offer the same collection of endpoints and be capable of verifying all retrieved data.
+
+## Considerations
+
+This is a project of considerable duration (roughly 18 months) and thus will coincide with other feature developments. Careful coordination can be extremely helpful in reducing the engineering resources required to complete the work and avoid unnecessary complexity.
```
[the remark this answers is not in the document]
cmwaters: Yes I agree. Phase 1 can be done in a minor release. Phase 2 arguable can be done in a minor release as well as we need to support both systems for one release and then in the following release deprecate the older system which would be breaking.

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Phase 2: Migration of storage, syncing and fraud
+
+The next step migrates first the storage engine and then the syncing and fraud protocols (`go-header` and `go-fraud`) to `celestia-app`. This replaces the existing `BlockStore` and blocksync and evidence reactors. Streamlining the storage will require careful migration but will eliminate the duplication of storage and unify many of the data exchange protocols and the public APIs. Here, the consensus node can become more integrated with the `nodebuilder` package as it begins to be split up into a series of modules. This phase may also try to align configuration files and the general file directory.
```
Wondertan, line 43: Let's mention somewhere the equivocation propagation as security improvement.

docs/architecture/adr-022-system-merge.md
staheri14, line 39: Are we talking about porting " porting `Init` and `Start` methods from `celestia-app`" to celestia-node, or entirely separate repo?

docs/architecture/adr-022-system-merge.md
staheri14, line 43: [Clarifying question] When we are discussing single binaries, I understand that the binary would allow a node to choose between different modes of operation, such as being a data node, light node, or consensus node, correct? Essentially, any combination of these would be possible, right? For example, a node operating as both consensus and data node?

If that's the case, then how do we distinguish between the different resource requirements of each node type? For example, as a light node, one might want to dedicate minimal bandwidth to some operations, while as a full consensus node, the bandwidth allocation could be different for the same operation. How does this new unified design address this aspect? Also, would this mean the same number of configuration flags (union of all the configs specific to each node type) would still be required to parameterize each mode of operation separately?

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+The Celestia network is divided into two subsystems owing largely to two different networking implementations. Some of the justification around this decision is as follows:
+
+- Tendermint consensus was battle tested and widely used yet tightly coupled to the networking layer. Using it as close to out of the box as possible meant a massive reduction in engineering resources to reach product launch.
+- Tendermint networking layer was largely unfamiliar to the team and had known frailties. LibP2P was more familiar, had more documentation and community support. It also provided functionality that was not available in Tendermint.
+
+This decision led to the creation of `celestia-node`, a new repository, responsible for data availability sampling and data retrieval. As a consequence of this decision:
+
+- A bridge node was introduced which had to be run side by side with the consensus node, relaying data from the consensus node to the nodes in the DA network (i.e. using LibP2P). This meant:
+  - Extra complexity for node operators
+  - Duplicate storage of squares
```
staheri14, line 22: The "duplicate storage of squares" mean duplicates in all 3 nodes i.e., consensus, bridge and data? or just node and consensus?

docs/architecture/adr-022-system-merge.md
```diff
@@ -0,0 +1,63 @@
...
+  - Redundant computation of the extended data square from the block
```
staheri14, line 23: Ditto this part: Is the extended square redundantly computed by DA and consensus nodes? or also the bridge?

docs/architecture/adr-022-system-merge.md
staheri14, line 39: Could you please elaborate on why you used 'temporarily'? Isn't it the case that the consensus node should always be subscribed to the P2P GossipSub layer dedicated to EDS propagation (assuming the GossipSub protocol is in use) in order to publish the EDSs?

docs/architecture/adr-022-system-merge.md
cmwaters: consensus and bridge. A validator is supposed to run both the consensus node and bridge node which means they are keeping two copies of the same data

docs/architecture/adr-022-system-merge.md
cmwaters: to `celestia-node` such that they can be integrated into the `celestia` binary

docs/architecture/adr-022-system-merge.md
cmwaters: We're trying to categorise node types by their user groups. After which you just run a single node type which out of the box should roughly correspond to your requirements.

cmwaters: Ideally we converge on flags and configuration files (at least stylistically) to avoid confusing node operators as much as possible

docs/architecture/adr-022-system-merge.md
cmwaters: temporarily in the sense of being hybrid. In the long term we will be libp2p based only

cmwaters:
Closing this PR as this has become stale and seems no longer relevant as an ADR

Issue #2749, opened 2023-10-23
v1.2.0 doesn't contain pre-built binaries


rootulp:
## Problem

The [v1.2.0](https://github.com/celestiaorg/celestia-app/releases/tag/v1.2.0) release doesn't contain pre-built binaries b/c markdown link check failed [here](https://github.com/celestiaorg/celestia-app/actions/runs/6592848859/job/17914267262) and that's a prerequisite for version bump (which is a prerequisite for goreleaser)

## Proposal

1. Remove `markdown-linter` as a prerequisite from version bump https://github.com/celestiaorg/celestia-app/blob/c0743df725cba331b42176977f2dc702277c8f18/.github/workflows/ci-release.yml#L52-L53
2. Fix the broken QGB links

Note: we can only re-run the CI workflow on v1.2.0 without modifications and I don't expect that to succeed so we'd have to do something hacky to get pre-built binaries attached to v1.2.0. We can fix this so we don't hit it on subsequent releases.

rootulp:
The 1.2.0 release still doesn't contain the pre-built binaries but we took steps to avoid this for future releases via #2750 and #2751.

PR #2750, opened 2023-10-23
docs: use QGB permalinks to appease markdown link check

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/2749 

Note: we've discussed avoiding backporting docs changes so we can close this PR if we're okay with markdown link check failing on the v1.x branch from now on.

## Testing

On the v1.x branch, locally `make markdown-link-check` passes


SweeXordious (commented):
Thanks for handling this. I think it would be better to use x/blobstream in the link and keep main branch instead of referring to old docs before the rename.

rootulp:
I think links to docs should use permalinks so that we don't break the links again due to a future rename or docs move. Also I think it's a bit confusing that the v1.x branch contains references to "qgb" and then links to "blobstream" docs. If we do that, we probably need to clarify the name change.

SweeXordious:
True, we're planning to write a small ADR around renaming so that it's more clear

rootulp:
IDK what to do for this PR but I don't think we should remain content with CI ❌ s on all future PRs to the v1.x branch. Three options I can see to get this moving forward (in order of preference from most preferred to least):

1. Remove the broken code links entirely
2. Replace the broken code links with permalinks to v1.x code
3. Replace the broken code links with permalinks to the same code on main. IMO confusing b/c the rename from QGB -> Blobstream which wasn't fully backported to v1.x

Ultimately I defer to you @SweeXordious b/c you're the primary owner on QGB docs

evan-forbes (approved):

SweeXordious (approved):

PR #2751, opened 2023-10-23
ci: remove markdown-linter as prerequisite to version bump

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/2749

The markdown-linter job contains markdown-link-check which is super flakey. If it fails on a release then the pre-built binaries won't be generated and attached to the release.


evan-forbes (approved):

staheri14 (approved):

rootulp:
I want to forward port this to main.

PR #2760, opened 2023-10-24
docs: fix spelling of compatibility

kevinheavey:
"Compatibility" is misspelt in a few places


rootulp (approved):
Thanks for the fix!

evan-forbes (approved):
I think we should change the spelling of that word if possible (:zany_face:)

Issue #2758, opened 2023-10-24
disable api, grpc and grpc-web by default in app.toml


mindstyle85:
### Summary

disable api, grpc and grpc-web by default in app.toml

### Problem Definition

n/a

### Proposal

disable api, grpc and grpc-web by default in app.toml

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

rootulp:
@mindstyle85 can you provide more details for why they should be disabled by default?

PR #2761, opened 2023-10-24
chore: disable API, GRPC, and GRPC web by default

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2758

## Testing
```
./scripts/single-node.sh
cat ~/.celestia-app/config/app.toml
```

Verified that 

```toml
[api]

# Enable defines if the API server should be enabled.
enable = false
```

```toml
[grpc]

# Enable defines if the gRPC server should be enabled.
enable = false
```

```toml
[grpc-web]

# GRPCWebEnable defines if the gRPC-web should be enabled.
# NOTE: gRPC must also be enabled, otherwise, this configuration is a no-op.
enable = false
```

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -253,7 +253,9 @@ func DefaultConsensusConfig() *tmcfg.Config {
 
 func DefaultAppConfig() *serverconfig.Config {
 	cfg := serverconfig.DefaultConfig()
-	cfg.API.Enable = true
+	cfg.API.Enable = false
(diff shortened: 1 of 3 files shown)


rootulp:
Great question. I think it was motivated by this [Slack thread](https://celestia-team.slack.com/archives/C05UB2NPJEA/p1698147415527029).

evan-forbes (approved):
thanks

SweeXordious (approved):

PR #2764, opened 2023-10-24
chore: discard abci responses by default

evan-forbes:
## Overview

sets the default to discard abci responses

closes #

## Checklist


staheri14 (approved):

SweeXordious (approved):
Question, how would the abci responses be used in crash recovery? 
https://github.com/tendermint/tendermint/issues/9054#issuecomment-1216247747

Cc @cmwaters

cmwaters:
> Question, how would the abci responses be used in crash recovery?

Super rare case but if the node crashes after calling `Commit` but before we persist state (in tendermint) then we're in a situation where the  application has executed height h and Tendermint has the block at height h but not the state at height h. We use the abci responses then to reconstruct Tendermint's state and persist it so everything is synced (at height h).

cmwaters (approved):

PR #2770, opened 2023-10-25
refactor!: move all blob share commitment code to the inclusion package

evan-forbes:
## Overview

We currently have code relating to creating and proving commitment across the inclusion, x/blob/types, and shares package. This api breaking refactor moves all of the commitment code to the inclusion package. 

The reasoning behind this is that it will eventually allow for users to import commitment creation and proving code while not having to import the blob module, tendermint, or the sdk. 

The reasoning behind keeping it in the inclusion instead of shares is that entire reason for making a commitment is to prove inclusion. So the name seems to fit better.

The reasoning behind note keeping it in the blob package was that it was difficult to not have an import cycle since this logic requires the share splitting logic for blobs.

## Checklist

### DIFF
--- a/pkg/da/data_availability_header.go
+++ b/pkg/da/data_availability_header.go
@@ -4,14 +4,15 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"math"
 
 	"github.com/celestiaorg/rsmt2d"
 	"github.com/tendermint/tendermint/crypto/merkle"
 	"github.com/tendermint/tendermint/types"
+	"golang.org/x/exp/constraints"
 
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/pkg/shares"
-	"github.com/celestiaorg/celestia-app/pkg/square"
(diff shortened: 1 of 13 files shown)


rootulp (dismissed):
LGTM

evan-forbes:
converting to a draft to fix the CI. which we should actually not expect to break given we just moved things around. will report back after finding the fix

cmwaters (approved):
This seems a logical change. I think further to this I would like to see `inclusion` and `proof` combined. Inclusion is simply a form of proof

evan-forbes:
> I think further to this I would like to see inclusion and proof combined. Inclusion is simply a form of proof

yeah I agree. When writing the blob share commitment proofs, it becomes very clear that this is an issue

rootulp (approved):

rootulp:
hmm I think conventional commit PR title check failed [here](https://github.com/celestiaorg/celestia-app/actions/runs/6657340776/job/18091848848#step:2:6) because the `!` was included prior to the scope and the spec states that it should follow the scope.

```
refactor(inclusion)!: I expect this to work
refactor!(inclusion): I expect this to fail
```

PR #2771, opened 2023-10-25
chore!: delete FitsIntoSquare

evan-forbes:
## Overview

we should do this regardless of #2770 (and I could swear we already did), but it was conflicting so I'm making a PR to that branch atm.

## Checklist


rootulp (dismissed):
Agreed. `FitsInSquare` seems unused outside these locations 

Screenshot 2023-10-25 at 4 30 13 PM

cmwaters:
It seems that `celestia-node` also don't import it. We should predominantly be conscious of them and rollkit before removing methods but this LGTM

evan-forbes:
yeah its a good point we should check with users before removing stuff.

I hope noone is using this! it was introduced forrever ago in the first (or one of the first) versions of the non-interactive defaults

rootulp:
A lot of this code has been extracted to https://github.com/celestiaorg/go-square so I think we should close this as won't do and track in https://github.com/celestiaorg/go-square/issues/4.

PR #2778, opened 2023-10-27
feat!: refactor blob, namespace and shares into single package

cmwaters:
## Overview

This is an incredibly broad change in terms of files changed but it chiefly consists of merging `pkg/blob`, `pkg/namespace` and `pkg/shares` into a `shares` package and updating all the import paths accordingly.

This is part of the move to create a separate go module for the shares package that can be used by downstream dependents like rollkit without needing to import tendermint or the cosmos-sdk. 

Logically, `Share`, `Namespace` and `Blob` are all part of the same encoding protocol that marshals/unmarshals blobs into fixed length shares.

## Checklist


rootulp:
- I think this needs a `make proto-gen` b/c https://github.com/celestiaorg/celestia-app/actions/runs/6667373763/job/18120712424?pr=2778#step:3:88
- The link in the root level README needs to be updated b/c https://github.com/celestiaorg/celestia-app/actions/runs/6667373902/job/18120714610?pr=2778#step:4:1041

rootulp:
Overall LGTM. Not actually approving b/c looks like merge conflicts. 

One remaining question: should this be marked as API breaking? I assume consumers who imported the previous packages will have to change their import path.

cmwaters:
Yeah this is definitely API breaking. I'll add the exclamation mark in the commit just to be clear

cmwaters:
So there's talk about just having this as a separate repo. I don't mind doing that but if that is the case we probably don't need to merge this PR.

pkg/appconsts/global_consts.go
```diff
@@ -14,16 +14,16 @@ import (
...
 	// NamespaceVersionSize is the size of a namespace version in bytes.
-	NamespaceVersionSize = ns.NamespaceVersionSize
+	NamespaceVersionSize = shares.NamespaceVersionSize
 	// NamespaceVersionMaxValue is the maximum value a namespace version can be.
 	// This const must be updated if NamespaceVersionSize is changed.
 	NamespaceVersionMaxValue = math.MaxUint8
 
 	// NamespaceIDSize is the size of a namespace ID in bytes.
-	NamespaceIDSize = ns.NamespaceIDSize
+	NamespaceIDSize = shares.NamespaceIDSize
```
staheri14, line 23: [suggestion] Now that the `share` package will have its own `go.mod`, it might be more logical to define these constants e.g., `NamespaceIDSize` at the app level (where the `share` package is used and where the core logic of app resides). We can then pass them as parameters to the relevant structs and methods within the `share` package.

rootulp:
Converting to draft b/c we may not merge this based on

> So there's talk about just having this as a separate repo. I don't mind doing that but if that is the case we probably don't need to merge this PR.

rootulp:
Assuming this is safe to close b/c we're moving most of this code to https://github.com/celestiaorg/go-square

PR #2780, opened 2023-10-27
docs: ADR-023 DA Fee Market

cmwaters:
This ADR tries to capture some of the problems and the general dynamic of Celestia's fee market between buyers (rollups) and sellers (validators).

It puts forward a preliminary design based on a global min gas price.

I have added a discussion section around extending the design with more dynamic pricing schemes

[Rendered](https://github.com/celestiaorg/celestia-app/blob/cal/adr-fee-market/docs/architecture/adr-023-data-fee-market.md)


docs/architecture/adr-023-data-fee-market.md
```diff
@@ -0,0 +1,60 @@
+# ADR 023: DA Fee Market
```
rootulp, line 1: [uber nit] can the ADR title and the filename match? Since DA != data, I got tripped up.

Either:
```suggestion
# ADR 023: Data Fee Market
```

or rename file to `adr-023-da-fee-market.md`

docs/architecture/adr-023-data-fee-market.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Status
+
+Proposed
+
+## Changelog
+
+- 2023/06/28: Initial draft
```
rootulp, line 9: [question] is this date correct? June 28th seems like a long time ago. Perhaps it meant to say October 27th? 

```suggestion
- 2023/10/27: Initial draft
```

docs/architecture/adr-023-data-fee-market.md
```diff
@@ -0,0 +1,60 @@
...
+- **Secondary Markets and MEV**: If Celestia is able to offer value to it's users but is not able to correctly capture that value, it risks being captured on secondary markets that auction block space off chain and use other currencies. Moving payment offchain reduces the utility of the TIA token that secures the network. Actors such as validators who wish to extract as much value as possible (MEV) may make selfish decisions that harm the integrity of the network as a whole.
+- **Price Elasticity and Competition**: Price elasticity refers to how much a buyers demand for a commodity changes as the price changes. Given a Rollups strong dependence on Celestia for liveness (their service stops if they are unable to submit data), the commodity is rather inelastic. The eventual increase in competition and fallback methods that allow a rollup to switch providers may reduce the elasticity. Inelastic goods generally gives greater power to the seller.
```
rootulp, line 25:
```suggestion
- **Price Elasticity and Competition**: Price elasticity refers to how much a buyer's demand for a commodity changes as the price changes. Given a Rollups strong dependence on Celestia for liveness (their service stops if they are unable to submit data), the commodity is rather inelastic. The eventual increase in competition and fallback methods that allow a rollup to switch providers may reduce the elasticity. Inelastic goods generally give greater power to the seller.
```

docs/architecture/adr-023-data-fee-market.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Decision
+
+Celestia's fee market should fall between the bounds of being cost-effective for both suppliers (validators) and buyers (rollups). It should endeavour to provide the greatest price certainty and use the resource of the network in the most efficient manner. To this extent the following decisions were made:
+
+- **Global Min Gas Price**: A new block validity rule has been introduced constraining the minimum gas price (in other words the fee divided by the gas limit) that a transaction can have. This is designed to mitigate the ability for payment to be settled off-chain. It also resolves the incongruence of nodes with different minimum limits. This value can be modified via on-chain governance. This therefore means there is a dependency on the community interested in the value of TIA to ensure that a minimum price does not lead to an uncompetitive price.
```
rootulp, line 31: [optional] the tense here makes it sound like a block validity rule has already been added but I don't think that's the case yet.

rootulp (approved):

docs/architecture/adr-023-data-fee-market.md
cmwaters: Lol I actually started writing this in June so it is somewhat correct but I'll update the date to now

docs/architecture/adr-023-data-fee-market.md
cmwaters: Technically ADRs should be written in the past tense because it is supposed to document a decision that has happened however since we use the same format as proposals people often write with future tense.

rootulp: Ahh that's confusing. Down to adopt either past or future tense as long as ADRs are consistent.

My fuzzy feeling is that ADRs should use future tense when in the Proposed state and then past tense if moved from Proposed to Accepted. If the ADR never gets accepted, it can remain in future tense. But that may be too much overhead.

cmwaters: Yeah I get that although it might be cumbersome to have to go through and change the tense (unless chatGPT can do that for us)

docs/architecture/adr-023-data-fee-market.md
```diff
@@ -0,0 +1,60 @@
...
+
+Proposed
+
+## Changelog
+
+- 2023/06/28: Initial draft
+
+## Introduction
+
+At it's core, the Celestia network can be described as a market for the commodity of published data. By outsourcing the publication of data to Celestia, a rollup can mitigate the possibility of equivocation that may cause different machines to execute on a different set and order of transactions resulting in different states. In the field of distributed systems, this is known as split-brain. Thus, the market encompasses buyers who are willing to pay various amounts on the publication of their data, and sellers who are willing to supply the machines to perform the computation, storage and networking as defined by the protocol. This document explores the various dynamics of this market and makes some preliminary decisions on how to best design the fee market.
```
evan-forbes, line 13: **definitely** not blocking, but I do think we could get away with distilling this

```suggestion
At it's core, the Celestia network can be described as a market for the commodity of published data. This document explores the various dynamics of this market and makes some preliminary decisions on how to best design the fee market.
```

docs/architecture/adr-023-data-fee-market.md
evan-forbes, line 31: could we elaborate on how using a min gas price would help with the items listed above that were described as important? or discuss how it would make the gas price more consistent? or is the discussion section supposed to be here? 

we could also link some of the research and retros of the min fee in other networks such as ETH.

docs/architecture/adr-023-data-fee-market.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Context
+
+There are several imporant considerations regarding Celestia's fee market model:
+
+- **A Single Decentralized Seller**: While there are multiple competing buyers, the group of "validators" acting as sellers must collectively perform the computation, storage and networking that is required to produce the service. They may incur different costs in running their machines and receive different slices of the overall revenue generated (see non-linear distribution) but nonetheless must perform mostly the equivalent service. More importantly, the quality of the service is proportional to the level of decentralization of the network. More evenly weighted nodes makes the network more resilient to attacks and censorship and improves the reliability to accessing the published data. In summary, the force to continually lower costs to buyers and to remain a competitive market may reduce the amount of sellers, yet counter to this, the security and thus the quality of the service is dependent on a healthy amount of sellers. The market needs to balance this.
```
evan-forbes, line 19: > More evenly weighted nodes makes the network more resilient to attacks and censorship and improves the reliability to accessing the published data.

we might want to distinguish between per block censorship resistance and more medium term censorship reistance. The reason being that having an even distribution of stake doesn't increase per block censorship resistance.

evan-forbes (commented):
Where the fees go might be outside the scope of this discussion, however it might be worth mentioning that explicitly just because the most famous instance of a min fee also burn the fee which has a lot of externalities

evan-forbes:
will give this another review soon as well after chewing and discussing it further

musalbas:
I think this should be a CIP (when CIPs are implemented), as this concerns high level protocol mechanics rather than celestia-app implementation details.

cmwaters:
Closing this in favour of: https://github.com/celestiaorg/CIPs/pull/24

Issue #2781, opened 2023-10-27
Address serialization in PayForBlobs event


lzrscg:
### Summary of Bug

In `celestia.blob.v1.EventPayForBlobs`, the signer address is represented incorrectly:
1. It does not match the signer address in the corresponding PayForBlobs message
2. It is a string containing quotes, which is not how addresses are usually represented in messages
3. It is an invalid address format

For example take [this transaction](https://celestiascan.com/mocha/transaction/e38d16712833ba89de7b37d269a92215528e62b6ceabd31c55090821e949f491). In the event, the signer is: `"celestia1vdjkcetnw35kzvtpwye8gdtvxfu8jcmh8y685emcvuenydekxp5rj6mxx36ksctn0y6rywfswahq0lqt35"`. This is has quotes and too long to be a valid address. By contrast, in the message, the address is represented as `celestia1aq2t5l2xycw94zgxg32760h9kf4uhasy4290wn`.

### Version

v1.1.0

### Steps to Reproduce

1. Visit https://celestiascan.com/mocha/transaction/e38d16712833ba89de7b37d269a92215528e62b6ceabd31c55090821e949f491 and click the source data icon in the top right (or just use this link: http://consensus-full-mocha-4.celestia-mocha.com:26657/tx?hash=0xe38d16712833ba89de7b37d269a92215528e62b6ceabd31c55090821e949f491)
2. Find the EventPayForBlobs in the events
```json
{
                    "type": "celestia.blob.v1.EventPayForBlobs",
                    "attributes": [
                        {
                            "key": "YmxvYl9zaXplcw==",
                            "value": "WzEyMzI5Ml0=",
                            "index": true
                        },
                        {
                            "key": "bmFtZXNwYWNlcw==",
                            "value": "WyJBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQWVOSTllb2gvN3dWSHM9Il0=",
                            "index": true
                        },
                        {
                            "key": "c2lnbmVy",
                            "value": "ImNlbGVzdGlhMXZkamtjZXRudzM1a3p2dHB3eWU4Z2R0dnhmdThqY21oOHk2ODVlbWN2dWVueWRla3hwNXJqNm14eDM2a3NjdG4weTZyeXdmc3dhaHEwbHF0MzUi",
                            "index": true
                        }
                    ]
                }
```
3. [Decode](https://www.base64decode.org) the last entry from base64
**Key:** `signer`
[...]

PR #2786, opened 2023-10-30
fix: address serialization in PayForBlobs event

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2781

Note this _shouldn't_ be consensus breaking b/c the ABCI response results hash shouldn't depend on events. [`deterministicResponseDeliverTx`](https://github.com/celestiaorg/celestia-core/blob/ffdb652634fb1b6b6426c0f1a2c5adec8af80253/types/results.go#L45-L55) filters events from the response. Thanks @cmwaters for identifying.

Draft PR because
- [x] Needs a unit test
- [ ] Confirm this isn't consensus breaking. Run a two node local devnet. One pre this change. One post this change. Submit a PFB.


rootulp:
Closing and re-opening this PR from a branch on celestia-app and not on my fork. That should make a Docker image that we can use for testing.

PR #2787, opened 2023-10-30
ci: add fuzz tests to nightly run

cmwaters:


rootulp (approved):
so clean

SweeXordious (approved):

cmwaters:
> so clean

It's a bit manual now because there's only a few tests but if it becomes problematic we can change it

evan-forbes (approved):

PR #2788, opened 2023-10-30
chore: submit a PFB in `single-node.sh`

rootulp:
While working on a separate issue, I found it helpful for this script to automatically submit a PFB after spinning up a devnet. Completely optional to merge.


scripts/single-node.sh
SweeXordious, line 1: I think would make sense to have a separate script that sends bunch of PFBs. But I have no strong opinion about this

rootulp: Yea works.

PR #2803, opened 2023-11-01
feat!: support coordinated v2 upgrade with flag

cmwaters:
## Overview

As discussed, given no pre-existing coordinating mechanism (i.e. signalling), v2 will have to work differently to future upgrades. Namely, nodes will run `start` with an `upgrade-height` flag. At the height before that in `EndBlock`, the protocol version will be updated and cometbft will receive the new v2 app version. There will be no state migrations. The block at the following height will be v2 and validators will begin validating and executing accordingly.

I plan on following up with this with a e2e test upgrading from v1.x to `main`.

## Checklist


app/app.go
```diff
@@ -246,16 +247,10 @@ func New(
 	loadLatest bool,
 	invCheckPeriod uint,
 	encodingConfig encoding.Config,
-	upgradeSchedule map[string]upgrade.Schedule,
+	upgradeHeight int64,
```
rootulp, line 250: [question] will this only ever be the upgrade height at which celestia-app should upgrade from v1 -> v2? 

Can we clarify via godoc: 
1. this value is only applicable to v1 -> v2 upgraded height
2. what happens if this value is `0`

cmd/celestia-appd/cmd/root.go
```diff
@@ -153,6 +156,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig encoding.Config) {
 
 func addModuleInitFlags(startCmd *cobra.Command) {
 	crisis.AddModuleInitFlags(startCmd)
+	startCmd.Flags().Int64(UpgradeHeightFlag, 0, "Upgrade height to switch to v2. Must be coordinated amongst all validators")
```
rootulp, line 159:
```suggestion
	startCmd.Flags().Int64(UpgradeHeightFlag, 0, "Upgrade height to upgrade from v1 to v2. Must be coordinated amongst all validators")
```

cmd/celestia-appd/cmd/root.go
```diff
@@ -44,6 +45,8 @@ const (
 
 	// FlagLogToFile specifies whether to log to file or not.
 	FlagLogToFile = "log-to-file"
+
+	UpgradeHeightFlag = "upgrade-height"
```
rootulp, line 49: [question][not blocking] I wonder if this flag should explicitly state `v2` so that users don't get confused by this flag for future upgrades. If it doesn't explicitly state `v2`, I think a godoc that explains it's only used for `v2` would be helpful.

rootulp (approved):
Overall LGTM and seems a bit easier to maintain than a map of upgradeSchedules. My only feedback is around clarifying (either via godoc or flag rename) that the `--upgrade-height` flag is only applicable to the v1 to v2 upgrade.

app/app.go
cmwaters: Yes, this will be removed in v3 and can only work when going from v1/v2. It will be ignored for other version.

cmwaters: I will add clarification in the godoc

cmd/celestia-appd/cmd/root.go
cmwaters: Yeah I think that's a good idea. So `v2-upgrade-height`?

evan-forbes (dismissed):
:+1:

evan-forbes:
this still LGTM, but out of curiosity, did we add a commit here and github is still letting us merge without getting more approves?

rootulp:
I think the approvals didn't get dismissed b/c this PR targets a feature branch `cal/minor...` that doesn't have this setting:

Screenshot 2023-11-07 at 5 41 46 PM

evan-forbes:
ahh I see thanks. assumed it was towards main

cmwaters:
I assume when it switches, it will dismiss the review

evan-forbes (approved):

PR #2823, opened 2023-11-08
test: mempool overrides

rootulp:
I wanted to see what value we use for `MaxTxsBytes`. Since it's computed and not hard-coded, I wrote a test to see it and verify we don't accidentally revert it.


evan-forbes (approved):

cmwaters (approved):
We should do these kind of compatibility vector tests in many other cases as well to prevent us from accidentally breaking stuff

Issue #2824, opened 2023-11-08
Add CLI support for initializing application stores from an existing snapshot


cmwaters:
### Summary

In SDK v0.47, a cli feature was introduced that allowed for node operators to initialise their `application.db` based off a trusted snapshot. This provides an alternative way (other than statesync) to bootstrap a node to a certain height. This can be super useful for starting up one node from another trusted node and is important in the case of wanting to debug the state transitions that happen at a certain height

### Proposal

Port over the features https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/client/snapshot/cmd.go

### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

rootulp:
It looks like the snapshots command was originally introduced in https://github.com/cosmos/cosmos-sdk/pull/16067. Our fork already includes https://github.com/celestiaorg/cosmos-sdk/tree/release/v0.46.x-celestia/client/snapshot so I think we just need to add the command to the root celestia-app command.

PR #2825, opened 2023-11-08
feat(cmd): snapshots command

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2824

## Testing

I verified the snapshots command was added

```shell
$ ./build/celestia-appd snapshots
Manage local snapshots

Usage:
  celestia-appd snapshots [command]

Available Commands:
  delete      Delete a local snapshot
  dump        Dump the snapshot as portable archive format
  export      Export app state to snapshot store
  list        List local snapshots
  load        Load a snapshot archive file (.tar.gz) into snapshot store
  restore     Restore app state from local snapshot

Flags:
  -h, --help   help for snapshots

```

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/cosmos/cosmos-sdk/client/keys"
 	"github.com/cosmos/cosmos-sdk/client/rpc"
+	"github.com/cosmos/cosmos-sdk/client/snapshot"
 	"github.com/cosmos/cosmos-sdk/server"
 	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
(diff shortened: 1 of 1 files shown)


cmd/celestia-appd/cmd/root.go
```diff
@@ -149,6 +150,9 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig encoding.Config) {
 		keys.Commands(app.DefaultNodeHome),
 		bscmd.VerifyCmd(),
 	)
+	rootCmd.AddCommand(
+		snapshot.Cmd(NewAppServer),
+	)
```
cmwaters, line 155: nit: I guess this could have just been appended to the `AddCommand` function above

cmwaters (approved):
ty! this turned out to be easier than I thought. Can we backport this to v1.x?

cmd/celestia-appd/cmd/root.go
rootulp: true. TBH the contents of this function don't make sense to me. IDK why we add some commands in a block at the top and then add more later

cmwaters: yeah it's all a bit messy but not too problematic

PR #2830, opened 2023-11-08
fix: handle context cancellations in txsim

cmwaters:
I have seen a few CI tests fail because although the context stops the process, it isn't correctly parsed causing the program that called the txsim to think that there was another error and not the context Canceled or DeadlineExceeded errors


staheri14 (approved):
LGTM!

PR #2832, opened 2023-11-09
feat!: add a signalling mechanism for coordinated upgrades

cmwaters:
## Overview

As per ADR-018, this PR extends the existing minimal upgrade module with a signalling mechanism.

Validators are expected to submit an on-chain message to signal that they wish to change version of the network. When a quorum has signalled the next version the upgrade module signals to the app that it is ready to switch to the next state machine. If the app version is not supported the node will panic. Note that this feature does not currently support downgrading. The only permissible app version change is the very next increment. To cancel the upgrade, the same validators need only to submit an on-chain message with the current version they are on.

There are some remaining design decisions that I would appreciate feedback on:

1. The current default signalling quorum is 5/6. That is halfway between 2/3 and 3/3. The reason being is that during an upgrade process, it's impossible to maintain 1/3 fault tolerance for both omission and commission faults. The best we can achieve is 1/6. Imagine it being set to 2/3. Say 2/3 - 1 had v3 and wanted to move to v3 and the other 1/3 only had v2 and did not want to upgrade. A malicious node with just 1 voting power could halt the network by signalling for v3 and then refusing to vote afterwards. This context is all to ask the question whether we want this parameter to be modified by governance or hard-coded to the version. I initially have set it as a parameter but I think it may be better to have it hard-coded to the version (since we should practically never change this parameter)
2. The way for tallying the votes is very inefficient. Every block, at the end, we loop through all votes, query the voting power of each node and tally accordingly. This is fine as a safe naive path but we may want to reduce the unnecessary load on the network. There are a few options:
   **a. Use diffs.** Whenever there is a voting power change to a validator we add a hook that fires with the old voting power and the new voting power and the upgrade module listens to this hook and updates the total accordingly. This would require modifying the staking module and may cause problems down the track with needing to maintain the staking fork. It's also more prone to errors.
   **b. Use epochs**. Tally all the signal votes once every 1000 blocks or once a day. There is no need to tally continuously and we spare computation by reducing the frequency that we need to update. Most major upgrades are relatively time agnostic and can be done at any point. Emergency major upgrades operate differently and would involve the network halting temporarily.

Some remaining work that will be done in follow up PRs:
- add integration tests for the signalling upgrade process
- add CLI commands and update readme
- add support for automation (use separate upgrade keys with authz)

## Checklist


rootulp:
Hmm proto-gen is still failing so I can push a commit that tries to resolve.

x/upgrade/tally_test.go
```diff
@@ -0,0 +1,249 @@
...
+		Version:          2,
+	})
+	require.NoError(t, err)
+
+	res, err := upgradeKeeper.VersionTally(goCtx, &types.QueryVersionTallyRequest{
+		Version: 2,
+	})
+	require.NoError(t, err)
+	require.EqualValues(t, 30, res.VotingPower)
+	require.EqualValues(t, 100, res.Threshold)
```
rootulp, line 96: [nit] this usage of `Threshold` makes me realize that the variable name `ThresholdPower` may make the code clearer

x/upgrade/tally_test.go
```diff
@@ -0,0 +1,249 @@
...
+	// remove one of the validators from the set
+	delete(mockStakingKeeper.validators, testutil.ValAddrs[2].String())
+	mockStakingKeeper.totalVotingPower -= 50
+
+	res, err = upgradeKeeper.VersionTally(goCtx, &types.QueryVersionTallyRequest{
+		Version: 2,
+	})
+	require.NoError(t, err)
+	require.EqualValues(t, 1, res.VotingPower)
+	require.EqualValues(t, 47, res.Threshold)
```
rootulp, line 172: [no change needed] this makes me realize that threshold got rounded up from `46.66666667`. I think it's unlikely but should the threshold always round up instead of rounding to the nearest int? 

Update: commented above on a more relevant line where Threshold is computed

x/upgrade/keeper.go
```diff
@@ -1,97 +1,202 @@
...
+}
+
+// GetVotingPowerThreshold returns the voting power threshold required to
+// upgrade to a new version. It converts the signal quorum parameter which
+// is a number between 0 and math.MaxUint32 representing a fraction and
+// then multiplies it by the total voting power
+func (k Keeper) GetVotingPowerThreshold(ctx sdk.Context) sdkmath.Int {
+	quorum := k.SignalQuorum(ctx)
+	totalVotingPower := k.stakingKeeper.GetLastTotalPower(ctx)
+	return quorum.MulInt(totalVotingPower).RoundInt()
```
rootulp, line 178: [blocking] I think it would be problematic if this rounds down and a quorum of less than 2 / 3 of validators is able to start the upgrade b/c the totalVotingPower is slightly less than 2/3... Should this always round up instead?

rootulp (commented):
Overall really great work! I left a few questions to understand more. I think my only blocking feedback pertains to the threshold power calculation. I think it could be problematic if it rounds down.

x/upgrade/keeper.go
```diff
@@ -1,97 +1,202 @@
...
 }
 
-// ScheduleUpgrade implements the ibc upgrade keeper interface. This is a noop as
-// no other process is allowed to schedule an upgrade but the upgrade keeper itself.
-// This is kept around to support the interface.
-func (k Keeper) ScheduleUpgrade(_ sdk.Context, _ types.Plan) error {
-	return nil
-}
+// SignalVersion is a method required by the MsgServer interface
+func (k Keeper) SignalVersion(ctx context.Context, req *types.MsgSignalVersion) (*types.MsgSignalVersionResponse, error) {
```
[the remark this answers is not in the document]
cmwaters: Well the address here has to match the signature that signs the message. In this method we also check that they are a validator

x/upgrade/keeper.go
```diff
@@ -1,97 +1,202 @@
...
+		if output[version] > threshold {
+			return true, version
+		}
+	}
+	return false, 0
+}
+
+// GetVotingPowerThreshold returns the voting power threshold required to
+// upgrade to a new version. It converts the signal quorum parameter which
+// is a number between 0 and math.MaxUint32 representing a fraction and
```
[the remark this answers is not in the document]
cmwaters: Sorry, it was originally a uint32 but I changed it to use the sdk's decimal struct and forgot to update the documentation. Thanks for pointing it out

x/upgrade/keeper.go
```diff
@@ -1,97 +1,202 @@
...
+		return nil, stakingtypes.ErrNoValidatorFound
+	}
 
-// SetUpgradedClient sets the expected upgraded client for the next version of this chain at the last height the current chain will commit.
-func (k Keeper) SetUpgradedClient(ctx sdk.Context, planHeight int64, bz []byte) error {
-	store := ctx.KVStore(k.storeKey)
-	store.Set(types.UpgradedClientKey(planHeight), bz)
-	return nil
+	// the signalled version must be either the current version (for cancelling an upgrade)
+	// or the very next version (for accepting an upgrade)
```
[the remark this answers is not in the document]
cmwaters: Well if the network is on v2 and I signal to upgrade to v3, I effectively cancel this by signalling v2 - indicating a preference to stay on v2

x/upgrade/module.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/x/upgrade/client/cli"
+)
+
+func init() {
+	types.RegisterLegacyAminoCodec(codec.NewLegacyAmino())
+}
+
+const (
+	consensusVersion uint64 = 2
```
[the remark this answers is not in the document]
cmwaters: This was just directly taken from the existing upgrade module (since it shares the same name). It should probably actually be 3 instead of 2 now that I think of it

x/upgrade/types/errors.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+import (
+	"cosmossdk.io/errors"
+)
+
+var ErrInvalidVersion = errors.Register(ModuleName, 1, "signalled version must be either the current version or one greater")
```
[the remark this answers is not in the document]
cmwaters: I don't think they need to be unique given that there is also a `ModuleName` included so it doesn't have to be global just locally unique to the module

x/upgrade/tally_test.go
cmwaters: Yup I can modify that

x/upgrade/tally_test.go
cmwaters: So I think it's okay to use `RoundInt()` instead of `Ceil()` (i.e round to the nearest integer instead of the next highest integer) because the comparison logic is strictly greater than this threshold rather than greater than or equal to. i.e. if we rounded down to 46, it would still take 47 in voting power to advance. 

Alternatively, I do `Ceil` and then invoke >= for the comparison logic

x/upgrade/keeper.go
cmwaters: As I said in the other comment, because the condition is strictly greater, this won't happen but you might get a case where it rounds up and with strictly greater ends up requiring more voting power than necessary so I'm going to change to using `Ceil()`

x/upgrade/types/codec.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+)
+
+// RegisterLegacyAminoCodec registers the upgrade types on the LegacyAmino codec.
+func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
```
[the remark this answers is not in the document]
cmwaters: I don't think so but I'm not entirely sure what legacy amino is for. IIRC, it's needed for ledger signing right? I'm not sure if our other modules support the legacy amino encoding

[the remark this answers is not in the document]
rootulp: I think Legacy Amino is needed for Ledger but in my experience the Cosmos SDK docs are poor at explaining when modules should implement this and if it's actually deprecated / slated for removal.

x/upgrade/types/errors.go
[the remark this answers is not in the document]
rootulp: Agreed. I think that issue was originally motivated by a scenario where only the error code was propagated and the module name was absent but idk where that occurs.

x/upgrade/keeper.go
rootulp: Resolving b/c this seems outdated now.

x/upgrade/keeper.go
```diff
@@ -169,13 +175,12 @@ func (k Keeper) TallyVotingPower(ctx sdk.Context, threshold int64) (bool, uint64
 }
 
 // GetVotingPowerThreshold returns the voting power threshold required to
-// upgrade to a new version. It converts the signal quorum parameter which
-// is a number between 0 and math.MaxUint32 representing a fraction and
-// then multiplies it by the total voting power
+// upgrade to a new version.
 func (k Keeper) GetVotingPowerThreshold(ctx sdk.Context) sdkmath.Int {
-	quorum := k.SignalQuorum(ctx)
+	// contract: totalVotingPower should not exceed MaxUit64
```
rootulp, line 180: [typo]

```suggestion
	// contract: totalVotingPower should not exceed MaxUint64
```

x/upgrade/keeper.go
```diff
@@ -169,13 +175,12 @@ func (k Keeper) TallyVotingPower(ctx sdk.Context, threshold int64) (bool, uint64
...
 	totalVotingPower := k.stakingKeeper.GetLastTotalPower(ctx)
-	return quorum.MulInt(totalVotingPower).RoundInt()
+	thresholdFraction := SignalThreshold(ctx.BlockHeader().Version.App)
+	return totalVotingPower.MulRaw(thresholdFraction.Numerator).QuoRaw(thresholdFraction.Denominator)
```
rootulp, line 183: We may want to investigate / enforce that this never overflows

rootulp (approved):
Great work! Excited for this.

Documenting from a call a future improvement: clearing out the list of validator votes for a version after a successful upgrade. That way `shouldUpgrade` doesn't return true for subsequent blocks.

evan-forbes (approved):
dope!

to document here notes from a sync discussion, we could also use a "crank" tx instead of checking if the upgrade is ready at some interval.

added #2875 and #2874 so I think we can merge and continue

x/upgrade/keeper.go
```diff
@@ -1,97 +1,207 @@
...
+	_ types.QueryServer = Keeper{}
+
+	defaultSignalTheshold = Fraction{Numerator: 5, Denominator: 6}
 )
 
-var _ ibctypes.UpgradeKeeper = (*Keeper)(nil)
+type Fraction struct {
+	Numerator   int64
+	Denominator int64
+}
```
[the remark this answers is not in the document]
cmwaters: Yeah okay, maybe I just make it private. There's no need to have it public as it's not queryable

x/upgrade/keeper.go
cmwaters: ref: https://github.com/celestiaorg/celestia-app/issues/2878

Issue #2672, opened 2023-10-12
`TestMaxTotalBlobSizeSuite/TestSubmitPayForBlob_blobSizes` flake


rootulp:
## Problem

I've seen this CI test failure a few times recently

```
--- FAIL: TestMaxTotalBlobSizeSuite (6.58s)
    --- FAIL: TestMaxTotalBlobSizeSuite/TestSubmitPayForBlob_blobSizes (0.34s)
        max_total_blob_size_test.go:107: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/max_total_blob_size_test.go:107
            	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.4/suite/suite.go:112
            	Error:      	Not equal: 
            	            	expected: 0x2b82
            	            	actual  : 0x20
            	Test:       	TestMaxTotalBlobSizeSuite/TestSubmitPayForBlob_blobSizes
        --- FAIL: TestMaxTotalBlobSizeSuite/TestSubmitPayForBlob_blobSizes/2_mebibyte_blob (0.19s)
            testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
    full_node.go:97: tearing down testnode
FAIL
FAIL	github.com/celestiaorg/celestia-app/app/test	771.257s
```

## Links
- https://github.com/celestiaorg/celestia-app/actions/runs/6487194321/job/17616955455?pr=2658#step:4:26
- https://github.com/celestiaorg/celestia-app/actions/runs/6487314634/job/17617347693#step:4:26
- https://github.com/celestiaorg/celestia-app/actions/runs/6579292956/job/17874840856?pr=2623#step:4:30

rootulp:
I think the test cases of this test suite are overlapping with each other. When I attempt to repro locally, I can't repro when I comment out the 1 byte and 1MiB test case. I can repro when I run all three test cases.

rootulp:
Note: all the recent test failures on main are from this flake.

PR #2834, opened 2023-11-10
test!: attempt to resolve flakes

rootulp:
Attempt to close https://github.com/celestiaorg/celestia-app/issues/2672. 

This PR refactors some of the tests. It also removes the `_Flaky` suffix from other tests that I can't locally reproduce failures for. I plan on keeping the other "flake" issues open for now.


app/test/max_total_blob_size_test.go
```diff
@@ -75,19 +71,9 @@ func (s *MaxTotalBlobSizeSuite) TestSubmitPayForBlob_blobSizes() {
...
-			name: "1 mebibyte blob",
-			blob: mustNewBlob(mebibyte),
-			want: abci.CodeTypeOK,
-		},
 		{
 			name: "2 mebibyte blob",
-			blob: mustNewBlob(2 * mebibyte),
+			blob: newBlobWithSize(2 * mebibyte),
 			want: blobtypes.ErrTotalBlobSizeTooLarge.ABCICode(),
 		},
```
rootulp, line 78: It seems like the multiple test cases of this test are affecting each other. Since the first two test cases aren't expected to encounter an error and that happy path is covered in other tests, proposal to drop them from this test.

app/test/integration_test.go
```diff
@@ -206,71 +185,6 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
 	}
 }
 
-func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
```
rootulp, line 209: this test seems redundant with TestSubmitPayForBlob [here](https://github.com/celestiaorg/celestia-app/blob/be3acf0a79849af0dafda5eecb0644cfc08c58b0/pkg/user/signer_test.go#L48-L58)

app/test/integration_test.go
```diff
@@ -395,35 +308,20 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob_blobSizes() {
...
-			txResponseCode: abci.CodeTypeOK,
-		},
-		{
-			name:           "100,000 byte blob",
-			blob:           mustNewBlob(100_000),
-			txResponseCode: abci.CodeTypeOK,
-		},
-		{
-			name:           "1,000,000 byte blob",
-			blob:           mustNewBlob(1_000_000),
```
rootulp, line 414: these test cases are all redundant b/c the expected txResponseCode is identical

app/test/integration_test.go
```diff
@@ -30,12 +30,16 @@ import (
...
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
+const (
+	Kibibyte = 1024
+	Mebibyte = 1024 * Kibibyte
+)
+
 func TestIntegrationTestSuite(t *testing.T) {
```
rootulp, line 43: passing
```
$ go test -count=2 . -run "TestIntegrationTestSuite"
ok  	github.com/celestiaorg/celestia-app/app/test	50.124s
```

app/test/max_total_blob_size_test.go
```diff
@@ -63,9 +59,9 @@ func (s *MaxTotalBlobSizeSuite) SetupSuite() {
 	require.NoError(t, cctx.WaitForNextBlock())
 }
 
-// TestSubmitPayForBlob_blobSizes verifies the tx response ABCI code when
-// SubmitPayForBlob is invoked with different blob sizes.
-func (s *MaxTotalBlobSizeSuite) TestSubmitPayForBlob_blobSizes() {
+// TestErrTotalBlobSizeTooLarge verifies that submitting a 2 MiB blob hits
+// ErrTotalBlobSizeTooLarge.
+func (s *MaxTotalBlobSizeSuite) TestErrTotalBlobSizeTooLarge() {
```
rootulp, line 64: seems less flaky

```
$ go test -count=5 . -run "TestMaxTotalBlobSizeSuite"
ok  	github.com/celestiaorg/celestia-app/app/test	31.621s
```

app/test/square_size_test.go
```diff
@@ -60,8 +60,7 @@ func (s *SquareSizeIntegrationTest) SetupSuite() {
 
 // TestSquareSizeUpperBound sets the app's params to specific sizes, then fills the
 // block with spam txs to measure that the desired max is getting hit
-// The "_Flaky" suffix indicates that the test may fail non-deterministically especially when executed in CI.
-func (s *SquareSizeIntegrationTest) TestSquareSizeUpperBound_Flaky() {
+func (s *SquareSizeIntegrationTest) TestSquareSizeUpperBound() {
```
rootulp, line 63: can't get it to fail locally but we should explore making this test faster

```
$ go test -count=1 . -run "TestSquareSizeIntegrationTest"
ok  	github.com/celestiaorg/celestia-app/app/test	106.361s
```

pkg/square/square_test.go
```diff
@@ -134,12 +134,10 @@ func generateBlobTxsWithNamespaces(t *testing.T, namespaces []ns.Namespace, blob
 	)
 }
 
-// The "_Flaky" suffix indicates that the test may fail non-deterministically especially when executed in CI.
-func TestSquareBlobShareRange_Flaky(t *testing.T) {
-	rand := tmrand.NewRand()
+func TestSquareBlobShareRange(t *testing.T) {
```
rootulp, line 137: can't repro locally

```
$ go test -count=1000 . -run "TestSquareBlobShareRange"
ok  	github.com/celestiaorg/celestia-app/pkg/square	12.168s
```

x/blob/client/testutil/integration_test.go
```diff
@@ -182,7 +177,9 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 	}
 }
 
-// The "_Flaky" suffix indicates that the test may fail non-deterministically especially when executed in CI.
-func TestIntegrationTestSuite_Flaky(t *testing.T) {
+func TestIntegrationTestSuite(t *testing.T) {
```
rootulp, line 180: Can't repro locally

```
$ go test ./... -count=3 -run "TestIntegrationTestSuite"
ok  	github.com/celestiaorg/celestia-app/x/blob/client/testutil	64.982s
```

app/test/integration_test.go
evan-forbes: originally this was to test for the canonical api, so removing makes sense. we also don't need to be testing user.TxOptions here

app/test/integration_test.go
evan-forbes: it looks like the purpose of this test is test different sizes, so if we remove these then I think we should rename it or remove it entirely if the error for too large of txs is being tested elsewhere

app/test/max_total_blob_size_test.go
```diff
@@ -105,16 +86,6 @@ func (s *MaxTotalBlobSizeSuite) TestSubmitPayForBlob_blobSizes() {
...
 			require.Equal(t, tc.want, res.Code, res.Logs)
-
-			sq, err := square.Construct([][]byte{blobTx}, appconsts.LatestVersion, squareSize)
-			if tc.want == abci.CodeTypeOK {
-				// verify that if the tx was accepted, the blob can fit in a square
-				assert.NoError(t, err)
-				assert.False(t, sq.IsEmpty())
-			} else {
-				// verify that if the tx was rejected, the blob can not fit in a square
-				assert.Error(t, err)
```
evan-forbes, line 116: [blocking]
do we still want to keep this assertion?

test/util/testnode/node_interaction_api.go
```diff
@@ -277,7 +277,7 @@ func (c *Context) PostData(account, broadcastMode string, ns appns.Namespace, bl
 // create a square of the desired size. broadcast mode indicates if the tx
 // should be submitted async, sync, or block. (see flags.BroadcastModeSync). If
 // broadcast mode is the string zero value, then it will be set to block.
-func (c *Context) FillBlock(squareSize int, accounts []string, broadcastMode string) (*sdk.TxResponse, error) {
+func (c *Context) FillBlock(squareSize int, account string, broadcastMode string) (*sdk.TxResponse, error) {
```
evan-forbes, line 280: imo this should count as api breaking as well since celestia-node imports this function

x/blob/client/testutil/integration_test.go
```diff
@@ -40,17 +39,12 @@ func NewIntegrationTestSuite(cfg cosmosnet.Config) *IntegrationTestSuite {
...
-		s.T().Skip("skipping integration test in short mode.")
-	}
 	s.T().Log("setting up integration test suite")
 
-	net := network.New(s.T(), s.cfg, username)
-
-	s.network = net
-	s.kr = net.Validators[0].ClientCtx.Keyring
+	network := network.New(s.T(), s.cfg, username)
+	s.network = network
```
evan-forbes, line 46: just my 2cents, not blocking, but I don't think we should ever override the name of a package, which was the reason this was called net and not network

x/blob/client/testutil/integration_test.go
evan-forbes: we could get this to go way faster if we refactored it to not use the cosmos-sdk's network code https://github.com/celestiaorg/celestia-app/issues/829

x/blob/client/testutil/integration_test.go
```diff
@@ -89,58 +81,14 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 			expectedCode: 0,
 			respType:     &sdk.TxResponse{},
 		},
-		{
-			name: "unsupported share version",
```
evan-forbes, line 93: :+1: :+1:

evan-forbes (dismissed):

app/test/integration_test.go
rootulp: Good point. We're not testing for `ErrTxTooLarge` elsewhere but it seems highly related to a different test: `TestErrTotalBlobSizeTooLarge` which checks for a ErrTotalBlobSizeTooLarge error.

Since this test still tests two different blob sizes (1 mebibyte, 10,000,000 byte), if we need to keep it then I guess the test could be renamed to `TestSubmitPayForBlob_txResponseCode`

rootulp: TBH I don't think the second test case for ErrTxTooLarge is providing much value so I'm inclined to delete it and refactor the test.

rootulp: After refactoring, I realized there's already a test for `SubmitPayForBlob` in signer_test.go so deleting this entirely.

app/test/max_total_blob_size_test.go
rootulp: Good catch. I elected to remove b/c it seems like this test is covering behavior of something unrelated (square construction).

I brought back the test in: 3e0c90216e81eeb52a3fa91e7044d97d0fe98da6

test/util/testnode/node_interaction_api.go
rootulp: great catch

x/blob/client/testutil/integration_test.go
rootulp: ohhh great point, that makes a lot of sense. IMO we should probably do that for instances where variables are named `blob` too but will keep in mind for the future and revert this change.

rootulp: 4a9d8e50081a3f64ed2544b4b7a1f0df23f2a3f2

app/test/integration_test.go
```diff
@@ -80,47 +72,41 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
 			tmrand.NewRand(),
 			s.cctx.Keyring,
 			c.GRPCClient,
-			950000,
+			600*kibibyte,
 			1,
 			false,
 			s.accounts[:20],
 		)
 	}
```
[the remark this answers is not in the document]
rootulp: CodeRabbit seems wrong here.

app/test/integration_test.go
```diff
@@ -131,27 +117,21 @@
...
 			hashes := make([]string, len(txs))
 
 			for i, tx := range txs {
+				// The default CometBFT mempool MaxTxBytes is 1 MiB so the generators in
+				// this test must create transactions that are smaller than that.
+				require.LessOrEqual(t, len(tx), 1*mebibyte)
+
 				res, err := s.cctx.Context.BroadcastTxSync(tx)
 				require.NoError(t, err)
 				assert.Equal(t, abci.CodeTypeOK, res.Code, res.RawLog)
```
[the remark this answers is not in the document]
rootulp: hmm I think it's okay to use require instead of assert here so resolving.

app/test/integration_test.go
```diff
@@ -195,7 +172,7 @@
 				require.EqualValues(t, v2.Version, blockRes.Block.Header.Version.App)
 
 				sizes = append(sizes, size)
-				ExtendBlobTest(t, blockRes.Block)
+				ExtendBlockTest(t, blockRes.Block)
 			}
 			// ensure that at least one of the blocks used the max square size
 			assert.Contains(t, sizes, uint64(appconsts.DefaultGovMaxSquareSize))
```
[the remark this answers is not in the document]
rootulp: this suggestion doesn't make sense to me.

evan-forbes (approved):

app/test/integration_test.go
```diff
@@ -167,10 +147,7 @@
...
 				resp, err := testnode.QueryTx(s.cctx.Context, hash, true)
 				require.NoError(t, err)
-				assert.NotNil(t, resp)
-				if resp == nil {
-					continue
-				}
+				require.NotNil(t, resp)
 				require.Equal(t, abci.CodeTypeOK, resp.TxResult.Code, resp.TxResult.Log)
 				heights[resp.Height]++
 				// ensure that some gas was used
```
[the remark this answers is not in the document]
rootulp: there is an assertion literally on the next line... silly CodeRabbit.

test/util/testnode/node_interaction_api.go
```diff
@@ -291,7 +291,7 @@ func (c *Context) FillBlock(squareSize int, accounts []string, broadcastMode str
 
 	// we use a formula to guarantee that the tx is the exact size needed to force a specific square size.
 	blobSize := shares.AvailableBytesFromSparseShares(shareCount)
-	return c.PostData(accounts[0], broadcastMode, appns.RandomBlobNamespace(), tmrand.Bytes(blobSize))
+	return c.PostData(account, broadcastMode, appns.RandomBlobNamespace(), tmrand.Bytes(blobSize))
```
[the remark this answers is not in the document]
rootulp: There doesn't need to be error handling within this function b/c it returns an error if one is encountered.

staheri14 (approved):
LGTM!

app/test/integration_test.go
```diff
@@ -80,47 +72,41 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
-	// tendermint's default tx size limit is 1Mb, so we get close to that
-	equallySized1MbTxGen := func(c client.Context) []coretypes.Tx {
+	singleBlobTxGen := func(c client.Context) []coretypes.Tx {
 		return blobfactory.RandBlobTxsWithAccounts(
 			s.ecfg,
 			tmrand.NewRand(),
 			s.cctx.Keyring,
 			c.GRPCClient,
-			950000,
+			600*kibibyte,
```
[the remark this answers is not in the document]
rootulp: This change wasn't strictly needed. I thought it was misleading that the variable name stated `1Mb` when the generator created transactions that weren't 1Mb. Opted to change the variable name and the size of txs created.

app/test/integration_test.go
```diff
@@ -80,47 +72,41 @@ func (s *IntegrationTestSuite) SetupSuite() {
 func (s *IntegrationTestSuite) TestMaxBlockSize() {
 	t := s.T()
 
-	// tendermint's default tx size limit is 1Mb, so we get close to that
```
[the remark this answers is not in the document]
rootulp: The comment is no longer accurate b/c now this tx generator creates transactions that are ~600 KiB

app/test/integration_test.go
```diff
@@ -80,47 +72,41 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
 			c.GRPCClient,
-			200000, // 200 KB
+			200*kibibyte,
 			3,
 			false,
 			s.accounts[20:40],
 		)
 	}
 
-	// Generate 80 randomly sized txs (max size == 50 KB). Generate these
```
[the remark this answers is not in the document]
rootulp: Yes. 

> Generate these transactions using some of the same accounts as the previous generator to ensure that the sequence number is being utilized correctly in blob txs

was incorrect because this tx generator used accounts 40 - 120 which does not overlap with the previous tx generator (20 - 40). So I deleted the misleading comment.

app/test/unit_test.go
```diff
@@ -0,0 +1,6 @@
+package app_test
+
+const (
+	kibibyte = 1024
+	mebibyte = 1024 * kibibyte
```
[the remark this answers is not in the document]
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/2873

test/util/testnode/full_node_test.go
```diff
@@ -31,40 +38,31 @@ type IntegrationTestSuite struct {
...
 	err := s.cctx.WaitForNextBlock()
 	require.NoError(err)
-	// check that we're actually able to set the consensus params
+
 	var params *coretypes.ResultConsensusParams
 	// this query can be flaky with fast block times, so we repeat it multiple
 	// times in attempt to decrease flakiness
-	for i := 0; i < 40; i++ {
-		params, err = s.cctx.Client.ConsensusParams(context.TODO(), nil)
+	for i := 0; i < 100; i++ {
```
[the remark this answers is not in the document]
rootulp: to decrease flakiness

test/util/testnode/full_node_test.go
```diff
@@ -127,7 +123,7 @@ func (s *IntegrationTestSuite) Test_FillBlock_InvalidSquareSizeError() {
 
 	for _, tc := range tests {
 		s.Run(tc.name, func() {
-			_, actualErr := s.cctx.FillBlock(tc.squareSize, s.accounts, flags.BroadcastAsync)
+			_, actualErr := s.cctx.FillBlock(tc.squareSize, s.accounts[2], flags.BroadcastAsync)
```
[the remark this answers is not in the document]
rootulp: `s.accounts` is still a slice with multiple accounts. This line just uses one of those accounts.

x/blob/client/testutil/integration_test.go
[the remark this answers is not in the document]
rootulp: they didn't contribute to flakiness but they were removed b/c they don't strictly need to be included in this integration test. Instead these cases should be covered by unit tests

test/util/testnode/full_node_test.go
```diff
@@ -19,7 +19,14 @@ import (
 	coretypes "github.com/tendermint/tendermint/rpc/core/types"
 )
 
+const (
+	kibibyte = 1024
```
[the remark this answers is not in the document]
rootulp: I really want to commit this suggestion but I'm going to skip it b/c accepting it will require re-approvals which takes too much time

PR #2843, opened 2023-11-14
chore: lower the default mempool min gas price

evan-forbes:
## Overview

The mempool minfee seems too high as the goal of it is just to reduce spam if the blocks are empty. The proper solution will come with a consensus critical dynamic minfee. For now the default should be lowered by an order of magnitude.

## Checklist

### DIFF
--- a/pkg/appconsts/initial_consts.go
+++ b/pkg/appconsts/initial_consts.go
@@ -20,7 +20,7 @@ const (
 	// DefaultMinGasPrice is the default min gas price that gets set in the app.toml file.
 	// The min gas price acts as a filter. Transactions below that limit will not pass
 	// a nodes `CheckTx` and thus not be proposed by that node.
-	DefaultMinGasPrice = 0.1
+	DefaultMinGasPrice = 0.01
 
 	// DefaultUnbondingTime is the default time a validator must wait
 	// to unbond in a proof of stake system. Any validator within this


rootulp (dismissed):
should this be backported to v1.x?

staheri14 (dismissed):

cmwaters (dismissed):

cmwaters (approved):

cmwaters:
#makedatafreeagain

app/default_overrides_test.go
```diff
@@ -62,7 +62,7 @@ func TestDefaultAppConfig(t *testing.T) {
 
 	assert.Equal(t, uint64(1500), cfg.StateSync.SnapshotInterval)
 	assert.Equal(t, uint32(2), cfg.StateSync.SnapshotKeepRecent)
-	assert.Equal(t, "0.1utia", cfg.MinGasPrices)
+	assert.Equal(t, "0.002utia", cfg.MinGasPrices)
```
rootulp, line 65: [question] how was `.002utia` arrived at? IIUC `0.1utia` was arrived at based on rough estimates. How did the estimates change to go from 0.1 -> 0.01 -> 0.002?

rootulp (approved):
[not blocking] should we include this default in `params.md`?

app/default_overrides_test.go
rootulp: If the original value was off by two orders of magnitude, I would expect 0.1 -> 0.01 -> 0.001

rootulp: Since `.002utia` got merged here, I think we may want to update chain-registry to 

```
        "low_gas_price": 0.002,
        "average_gas_price": 0.004,
        "high_gas_price": 0.008
```

but we likely only want to do this after a decent chunk of validators have updated to a release that includes this. On second thought, I think their locally configured default won't get overridden with this new default so our release notes should probably explicitly call out the manual action that validators need to take.

jcstein: https://github.com/cosmos/chain-registry/pull/3239
https://github.com/chainapsis/keplr-chain-registry/pull/290

updating these now, but leaving in draft

Issue #744, opened 2022-09-20
Remove temporary import hack for protobuf any


rootulp:
## Context
- https://github.com/celestiaorg/celestia-app/issues/507#issue-1284883031
- https://github.com/celestiaorg/celestia-app/pull/554

## Problem
We want to remove the temporary import hack in https://github.com/celestiaorg/celestia-app/pull/742/files#diff-63113d53e02a4e334ffa1a1ba0c308329c7acd431153571a2c02ce9268af5816R28-R32

evan-forbes:
is this still a thing @SweeXordious ?

SweeXordious:
it's still here: https://github.com/celestiaorg/celestia-app/blob/b8168ac2a29e8dda5ac895b540f3ff328933b490/scripts/protocgen.sh#L26-L29
not sure if it's worth fixing tho

rootulp:
I think this issue is worth fixing. Ideally we remove the lines that perform the hack b/c it looks like they modify the generated `query.pb.go` file and the top of that file says:

https://github.com/celestiaorg/celestia-app/blob/b75c7e3b80f91efae1164e8889bed9dd2c96414f/x/mint/types/query.pb.go#L1

PR #2846, opened 2023-11-15
chore: remove import hack for protobuf any

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/744

This import hack doesn't seem necessary b/c it doesn't actually modify `x/blobstream/types/query.pb.go`

## Testing

```
$ rm x/blobstream/types/query.pb.go
$ make proto-gen
```

and then observe that there is no `git diff` so no changes to the already checked in `x/blobstream/types/query.pb.go` and the generated one.

Also proto-gen passes in CI: https://github.com/celestiaorg/celestia-app/actions/runs/6881660922/job/18718509598?pr=2846

### DIFF
--- a/scripts/protocgen.sh
+++ b/scripts/protocgen.sh
@@ -23,11 +23,6 @@ done
 
 cd ..
 
-# temporary import hack to use cosmos-sdk implementation of Any type.
-# check https://github.com/celestiaorg/celestia-app/issues/507 for more information.
-sed -i 's/types "github.com\/celestiaorg\/celestia-app\/codec\/types"/types "github.com\/cosmos\/cosmos-sdk\/codec\/types"/g' \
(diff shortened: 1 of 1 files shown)


staheri14 (approved):
I ran the testing process (after removing the replacement command in the bash script); I can also confirm that the import path of the `types` package points to the cosmos types package i.e.,`github.com/cosmos/cosmos-sdk/codec/types`. This renders the import hack unnecessary and safe to remove.

evan-forbes (commented):
it seems like we can remove this as the protobuf is no longer changing, but I'm curious as to what changed that fixed this. Was it the update to gogo proto?

will defer @SweeXordious for final approval to be abundantly cautious and for additional context

SweeXordious (approved):
Weird, tested it and it's not needed anymore 🎉

Issue #2434, opened 2023-09-05
feat(blob): allow for more than one blob to be sumbmitted via the CLI


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/2411/files#r1316337240

The blob CLI only allows one blob to be submitted.

## Proposal

Add the ability to submit multiple blobs via the CLI

bao1029p:
hey @rootulp can i work on this

rootulp:
Sure! I think multiple blobs via the CLI may be an interesting UX problem so we may want to design the expected user experience before we start working on code.

bao1029p:
yes absolutely, right now the simplest way i came up with are just allow a bunch of hex blob as arguments but this could be improve by accept file path as input, this file can contain multiple blob data in json format or hex string

bao1029p:
oh sorry for the late reply didn't see your reaction, i will try the file path implementation

PR #2856, opened 2023-11-19
feat: Use file input in payForBlobs CLI to allow to submit multiple blobs

bao1029p:
Close [#2434](https://github.com/celestiaorg/celestia-app/issues/2434)

## Changes made
- Add blobJson proto type define how user can submit blob in file
- Add file path as argument input for `payForBlob` command
- Add `parseSubmitBlobs` to parse content from the file to array of defined `blobJSON`. Each blobJSON contain a namespaceID and blob in hex encoded format ( can modify what the user can put in the file later on, this need further conversation on this )
- Modifed the `broadcastPFB` func to accept multiple blobs instead

## Testing
test the command with single-node script running and got this result ( thanks @sontrinh16 for providing the test result )
image

using the test_blob.json file contain:
`{
    "Blobs": [
        {
            "namespaceId": "0x00010203040506070809",
            "blob": "0x48656c6c6f2c20576f726c6421"
        },
        {
            "namespaceId": "0x00010203040506070809",
            "blob": "0x48656c6c6f2c20576f726c6421"
        }
    ]
}`
## Checklist


bao1029p:
@coderabbitai pause

sontrinh16:
@bao1029p hey gud ser can i permision to push some code to this PR

bao1029p:
> @bao1029p hey gud ser can i permision to push some code to this PR

yes plss feel free to contribute to this PR

bao1029p:
thanks @sontrinh16 for the help, i think we are close command work now, will do some testing

proto/celestia/core/v1/blob/blob.proto
```diff
@@ -21,4 +21,12 @@ message BlobTx {
...
\ No newline at end of file
+}
+
+
+// BlobJson define a simple file in JSON format for user to pass in as argument 
+// in payForBlob CLI. 
+message BlobJson {
+  string namespace_id = 1;
+  string blob = 2;
+}
```
evan-forbes, line 32: adding this is understandable, but since the type is only internal then we should only define it as a go struct

x/blob/client/cli/payforblob.go
```diff
@@ -32,55 +32,79 @@ const (
...
+				rawblob, err := hex.DecodeString(hexStr)
+				if err != nil {
+					fmt.Printf("failure to decode hex blob value %s: %s", hexStr, err.Error())
+					continue
+				}
+
+				shareVersion, _ := cmd.Flags().GetUint8(FlagShareVersion)
+				blob, err := types.NewBlob(namespace, rawblob, shareVersion)
+				if err != nil {
+					fmt.Printf("failure to create blob with hex blob value %s: %s", hexStr, err.Error())
```
evan-forbes, line 101: [blocking]
as is would this submit a transaction without the blob that failed? if that's the case then I think we need to return with error. that could be immediatly or using something like multierror

x/blob/client/cli/payforblob.go
```diff
@@ -32,55 +32,79 @@ const (
...
+		namespaceID is the user-specifiable portion of a version 0 namespace. It must be a hex encoded string of 10 bytes.\n
+		blob must be a hex encoded string of any length.\n
+		
+		`,
 		Aliases: []string{"PayForBlob"},
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) < 2 {
-				return fmt.Errorf("PayForBlobs requires two arguments: namespaceID and blob")
+			if len(args) < 1 {
+				return fmt.Errorf("PayForBlobs requires one arguments: path to blob.json")
```
evan-forbes, line 67: [blocking]
I think there's a way to avoid a breaking change, and we should pursue that. for example we could use a flag to indicate that its pointing to json file (or accepted piped input etc). we could even check if the input has the last five characters equal to ".json" and then proceed to try and parse a json

evan-forbes (commented):
dope left a few comments on things I think we should handle but otherwise this is a really great start. thanks for doing this @bao1029p !!

x/blob/client/cli/payforblob.go
sontrinh16: yes agreed will update this

rootulp:
Thanks for this contribution! Looks like its heading in the right direction. Two small additions to Evan's review:

1. golangci-lint fails [here](https://github.com/celestiaorg/celestia-app/actions/runs/7046369381/job/19182518697?pr=2856#step:5:37536) which you may be able to reproduce locally if you run `make lint`. We'll want to fix those issues prior to merging.
2. proto-gen fails [here](https://github.com/celestiaorg/celestia-app/actions/runs/7046369275/job/19182516529?pr=2856). We usually fix via `make proto-gen` but in this case, we can replace the protobuf definition with a Go struct as Evan suggested.

sontrinh16:
hey all @bao1029p agreed i would jump in and finish this PR, happy to contribute

proto/celestia/core/v1/blob/blob.proto
sontrinh16: removed as requested

x/blob/client/cli/payforblob.go
sontrinh16: added the check, thanks for the suggestion

sontrinh16:
@rootulp @evan-forbes hey guys can you approve the CI run again, i think i resolve all the problem. Thanks

x/blob/client/cli/payforblob.go
```diff
@@ -32,55 +33,83 @@ const (
...
+
+		namespaceID is the user-specifiable portion of a version 0 namespace. It must be a hex encoded string of 10 bytes.\n
+		blob must be a hex encoded string of any length.\n
+		
+		`,
 		Aliases: []string{"PayForBlob"},
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) < 2 {
-				return fmt.Errorf("PayForBlobs requires two arguments: namespaceID and blob")
+			if len(args) < 1 {
```
rootulp, line 67: [blocking] note this is still a breaking change to the CLI because the number of arguments has decreased. Is it possible to introduce this feature in a non-breaking way (i.e. via an optional flag)?

x/blob/client/testutil/integration_test.go
```diff
@@ -81,6 +132,47 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
...
+				fmt.Sprintf("--from=%s", username),
+				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
+				fmt.Sprintf("--%s=%s", flags.FlagFees, sdk.NewCoins(sdk.NewCoin(s.cfg.BondDenom, sdk.NewInt(2))).String()),
+				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
+				fmt.Sprintf("--%s=1", paycli.FlagNamespaceVersion),
+			},
+			expectErr:    true,
+			expectedCode: 0,
+			respType:     &sdk.TxResponse{},
+		},
```
rootulp, line 175: these test cases were intentionally removed from main so they should be removed here. It looks like they were accidentally added back via a merge conflict.

x/blob/client/testutil/integration_test.go
```diff
@@ -70,8 +122,7 @@ func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
 		{
 			name: "valid transaction",
 			args: []string{
-				hexNamespace,
-				hexBlob,
+				validPropFile.Name(),
```
rootulp, line 125: Note: if we add this feature in a non-breaking way, then we can leave this test case untouched so that we have test coverage of the existing single PFB submission via CLI. We can additionally add a new test case that uses multiple PFBs via a file.

x/blob/client/cli/payforblob.go
sontrinh16: hmmm a more simpler way is to add another cmd for this feature like how cosmos sdk did with their gov cmd, what do you think? With this aproach we dont have to modify the test too, only need to add 1 new testcase for the new cmd

sontrinh16: hmmm one option we can do is too move this new feature to a new cmd, that way old testcase wont be affect since we can just add a new one in. Other solution might be, we add an optional flag for the file path and disable the args number check if the flag is being use instead check for the file extension only ( because since we only need the flag value to submit blobs so the 2 args is useless by this point ). What do you think ?

cmwaters: I would have a preference to keep this to a single command if possible but am also open to having two if we think it's cleaner. 

One option is that we just introduce a switch statement. If there is a single argument we treat it as a file, if it is two arguments we treat it as the namespace and blob

sontrinh16: @rootulp @cmwaters hey guys, i modified the code so it would handle both use cases ( single blob and multiple blobs )

x/blob/client/testutil/integration_test.go
rootulp: Awesome, thanks!

x/blob/client/cli/payforblob.go
```diff
@@ -28,59 +29,98 @@ const (
 	// FlagNamespaceVersion allows the user to override the namespace version when
 	// submitting a PayForBlob.
 	FlagNamespaceVersion = "namespace-version"
+
+	// FlagNamespaceVersion allows the user to override the namespace version when
+	// submitting a PayForBlob.
```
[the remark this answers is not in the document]
sontrinh16: good catch, will remove this

x/blob/client/cli/payforblob.go
```diff
@@ -28,59 +29,98 @@ const (
...
-		Short: "Pay for a data blob to be published to Celestia.",
-		Long: "Pay for a data blob to be published to Celestia.\n" +
-			"namespaceID is the user-specifiable portion of a version 0 namespace. It must be a hex encoded string of 10 bytes.\n" +
-			"blob must be a hex encoded string of any length.\n" +
-			// TODO: allow for more than one blob to be sumbmitted via the CLI
-			"This command currently only supports a single blob per invocation.\n",
+		Short: "Pay for a data blobs to be published to Celestia.",
+		Long: "Pay for a data blobs to be published to Celestia.\n" +
+			"User can use namespaceID and blob or path to a json file as argument" +
+			`Where blob.json contains:
```
staheri14, line 51: [Suggestion] To enhance the user experience, it would be beneficial to consider the implementation of flexible file naming conventions, rather than restricting to a specific file name like `blob.json`. This can be achieved by allowing users to include their chosen file name in the filepath, which then can be parsed accordingly.

sontrinh16: oh yeah, i think i need to specify that blob.json is an example file name

x/blob/client/cli/payforblob.go
```diff
@@ -28,59 +29,98 @@ const (
...
-		Use: "PayForBlobs namespaceID blob",
+		Use: "PayForBlobs [namespaceID, blob] [path/to/blob.json]",
 		// This example command can be run in a new terminal after running single-node.sh
-		Example: "celestia-appd tx blob PayForBlobs 0x00010203040506070809 0x48656c6c6f2c20576f726c6421 \\\n" +
+		Example: "celestia-appd tx blob PayForBlobs path/to/blob.json \\\n" +
 			"\t--chain-id private \\\n" +
 			"\t--from validator \\\n" +
 			"\t--keyring-backend test \\\n" +
 			"\t--fees 21000utia \\\n" +
 			"\t--yes",
```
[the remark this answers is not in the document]
sontrinh16: added as suggested

x/blob/client/cli/payforblob.go
```diff
@@ -28,59 +29,98 @@ const (
...
-		Short: "Pay for a data blob to be published to Celestia.",
-		Long: "Pay for a data blob to be published to Celestia.\n" +
-			"namespaceID is the user-specifiable portion of a version 0 namespace. It must be a hex encoded string of 10 bytes.\n" +
-			"blob must be a hex encoded string of any length.\n" +
-			// TODO: allow for more than one blob to be sumbmitted via the CLI
-			"This command currently only supports a single blob per invocation.\n",
+		Short: "Pay for a data blobs to be published to Celestia.",
+		Long: "Pay for a data blobs to be published to Celestia.\n" +
```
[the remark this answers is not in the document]
sontrinh16: updated

x/blob/client/cli/payforblob.go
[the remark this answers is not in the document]
sontrinh16: updated as requested

x/blob/client/cli/payforblob.go
```diff
@@ -28,69 +29,149 @@ const (
...
+				if filepath.Ext(path) != ".json" {
+					return fmt.Errorf("invalid file extension, require json got %s", filepath.Ext(path))
+				}
+
+				return nil
+			}
+
 			if len(args) < 2 {
-				return fmt.Errorf("PayForBlobs requires two arguments: namespaceID and blob")
+				return fmt.Errorf("PayForBlobs requires two argument: namespaceID and blob")
```
evan-forbes, line 95: if we're going to use fmt.Errorf, then we should include something to format in the string other wise I think the linter is going to block us

evan-forbes (commented):
only one comment from the linter otherwise lgtm. thanks @bao1029p

x/blob/client/cli/payforblob.go
sontrinh16: updated as request

evan-forbes (dismissed):

x/blob/client/cli/payforblob.go
```diff
@@ -28,69 +30,149 @@ const (
...
-			"This command currently only supports a single blob per invocation.\n",
+			"\t--yes \\\n" +
+			"\n" +
+			"celestia-appd tx blob PayForBlobs --input-file path/to/blobs.json \\\n" +
+			"\t--chain-id private \\\n" +
+			"\t--from validator \\\n" +
+			"\t--keyring-backend test \\\n" +
+			"\t--fees 21000utia \\\n" +
+			"\t--yes \\\n" +
+			"\\\n",
```
rootulp, line 56: [nit] when the example is printed, it contains trailing `\` after the last `--yes` and it shouldn't. Also there looks like one extra `\` after the second command.

```suggestion
			"\t--yes \n\n" +
			"celestia-appd tx blob PayForBlobs --input-file path/to/blobs.json \\\n" +
			"\t--chain-id private \\\n" +
			"\t--from validator \\\n" +
			"\t--keyring-backend test \\\n" +
			"\t--fees 21000utia \\\n" +
			"\t--yes \n",
```

rootulp (commented):
Fantastic work! Thanks for persisting through multiple rounds of review. I have one nit and then this LGTM

rootulp (approved):

evan-forbes (approved):

PR #2864, opened 2023-11-21
fix: set the protocol version during integration tests

evan-forbes:
## Overview

Ever since we implemented upgrades, we moved the baseapp.Option to set the protocol version outside of the app creator. This is a desired property to make v2 compatible for v1, however that also means that we stopped setting the app version in testnode. This PR remedies that by adding an app option to set manually set the protocol version in testnode. This also allows for the testground tooling to set its own app version, which is necessary for extremely large blonks.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to set a specific protocol version for testing purposes.

- **Tests**
  - Added a new test to verify that the default application version matches the latest protocol version.

- **Refactor**
  - Updated default application options to include the latest protocol version.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -108,6 +108,9 @@ const (
 	BondDenomAlias = "microtia"
 	// DisplayDenom defines the name, symbol, and display value of the Celestia token.
 	DisplayDenom = "TIA"
+	// SetProtocolVersionOptionKey is the key used to set the protocol version
+	// to a specific value during tests.
+	SetProtocolVersionOptionKey = "set-protocol-version"
 )
 
 // These constants are derived from the above variables.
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -259,6 +262,15 @@ func New(
 	cdc := encodingConfig.Amino
 	interfaceRegistry := encodingConfig.InterfaceRegistry
 
+	// set the protocol version to a specific value if specified. This is used
+	// for testing purposes.
+	if pvo := appOpts.Get(SetProtocolVersionOptionKey); pvo != nil {
+		baseAppOptions = append(baseAppOptions, func(ba *baseapp.BaseApp) {
+			appVersion := pvo.(uint64)
+			ba.SetProtocolVersion(appVersion)
```
evan-forbes, line 270: this is done in v1.x in https://github.com/celestiaorg/celestia-app/blob/2dbfabf1849e166974c1287c35b43e5e07727643/cmd/celestia-appd/cmd/root.go#L249-L251

but not on main

cmwaters:
Can we not have the protocol version set in the genesis file. 

I removed the option to set it in the constructor because it should be persisted to disk and retrieved from there rather than be initialized everytime. The application doesn't know what the app version is. It works it out when it initiates (through the `Info` call)

I would like the cannonical way is that the app version is initially set through the genesis which is read through in `InitGenesis`. It is read subsequently upon startup through `Info` (prior to that the node doesn't know the app version of the network). It is only updated via the upgrade mechanism in which the new app version gets persisted to disk.

evan-forbes:
I'm assuming this was an accidently contradiction, if not mind clarifying?

> Can we **not** have the protocol version set in the genesis file.

> I would like the cannonical way is that the app version is initially set through the genesis which is read through in InitGenesis.

if I'm understanding correctly, I completely agree with the second sentiment. That would also fix the issue in the tests, which this PR is strictly aimed at. I didn't recognize that this would be fixed in upgrading the sdk, so will close.

Issue #1944, opened 2023-06-16
Support gas refunds


cmwaters:
Currently when a user estimates the amount of gas required for their transaction, if they underestimate, their transaction is not executed but they still pay the gas, if they overestimate they pay more than what was necessary. Striking this balance is difficult but doesn't need to be. We can simply refund any remaining fee back to the user so they can safely overestimate and only pay for what they use.

For the implementation this would mean defining a post antehandler which is run after the transaction

rootulp:
Rough steps:

1. Test the existing behavior: https://github.com/celestiaorg/celestia-app/pull/2885
2. Draft a prototype of fee refunds based on gas consumption
3. Write a CIP
4. Polish prototype and get it ready to merge

rootulp:
During a discussion w/ @evan-forbes it became clear that "gas refunds" may not be the best way to describe this line of work to users. It's not clear why user's would receive a "gas refund". Also current name collides with the Cosmos SDK method [`RefundGas`](https://github.com/cosmos/cosmos-sdk/blob/6bea474d3ac99c61b14b50b047ee356b7f38b6ef/store/gas.go#L71-L73) which actually isn't needed for this.

Since all of the tx fee is deducted from the fee payer via the ante handler, we're exploring adding a posthandler that refunds the fee payer a portion of the tx fee that is based on the gas consumption of the tx. As a result, tx fees experienced by users are gas consumption-based rather than always the entire fee.

As a result, we could describe the entire fee system after this change as: gas consumption based tx fees. The portion we're adding to the posthandler could be described as gas consumption based tx fee refunds.

rootulp:
>  if they underestimate, their transaction is not executed but they still pay the gas

Hmm I can't repro this behavior. See this [pastebin](https://pastebin.com/tEnCWJZ3)

rootulp:
We put this line of work on pause b/c the mitigation strategies [here](https://github.com/celestiaorg/CIPs/blob/3243f1916110737c4f707038c4ec38f0bf24981e/cips/cip-11.md?plain=1#L63-L66) were deemed too complicated.

PR #2885, opened 2023-11-30
test: TestGasConsumption

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/1944

Note: we don't strictly need to merge this. I opted to write a test for the existing behavior so that we have a straight-forward way to verify the behavior change after we implement gas refunds.

### DIFF
--- a/pkg/user/signer_test.go
+++ b/pkg/user/signer_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/celestiaorg/celestia-app/test/util/testnode"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
-	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	abci "github.com/tendermint/tendermint/abci/types"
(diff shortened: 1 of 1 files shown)


pkg/user/signer_test.go
```diff
@@ -74,3 +75,43 @@ func (s *SignerTestSuite) ConfirmTxTimeout() {
...
 }
+
+// TestGasConsumption verifies that the amount deducted from a user's balance is
+// based on the fee provided in the tx instead of the gas used by the tx. This
+// behavior leads to poor UX because tx submitters must over-estimate the amount
+// of gas that their tx will consume and they are not refunded for the excess.
+func (s *SignerTestSuite) TestGasConsumption() {
+	t := s.T()
+
+	utiaToSend := int64(1)
```
staheri14, line 86: [Optional] This test already has a fantastic set of comments 👍. Adding one more line to explain that this transaction specifies a fee amount of ` 1 tia`, which is higher than the actual gas consumption for this transaction, would elevate the explanation to the next level.

staheri14 (approved):
👍

pkg/user/signer_test.go
rootulp: Good point, can do!

staheri14 (approved):

cmwaters (approved):

PR #2887, opened 2023-12-01
feat!: refund gas remaining

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/1944

Remaining tasks
- [x] Define an upper bound for the amount of gas to refund (initially 50% of the original tx's gas)
- [x] Investigate the max gas consumed to perform the refund operation. See https://github.com/celestiaorg/celestia-app/pull/2887#issuecomment-1870542935
- [x] Define a new constant for the max gas consumed to perform the refund operation. Modify the refund operation to only refund gas remaining - max gas consumed to perform the refund operation.
- [x] Investigate how to handle the `simulate` flag.

Test cases
- [x] If the tx fees are paid by a feegranter, then the feegranter should receive the refund.
- [x] If the tx fees are paid by the signer, then the signer should receive the refund.
- [x] If the gas remaining after tx processing isn't greater than the amount of gas needed to process a refund, no gas should be refunded.
- [x] If a tx specifies a gas price != 1, the amount refunded is based on the gas price specified and doesn't assume a gas price of 1.
- [ ] If a tx specifies more than one fee coin
- [ ] Improve test coverage on max portion of gas consumed to refund


rootulp:
😄 looks like its working! https://github.com/celestiaorg/celestia-app/actions/runs/7065552202/job/19235786103#step:4:42

cmwaters:
Good start!

Did you work out how much gas this operation would cost on average?

Some design decisions to consider:
- Do we want to still charge a fixed gas cost which is the average we assume such an operation to consume? (and then only if there is still unspent gas do we perform the refund)
- Do we want an upper bound to the amount of gas that can be refunded? Perhaps as a portion of the original gas allocated. This stops people from over consuming. It's not too much of a problem now but say in the future we set an upper bound on gas consumed, someone could allocate the entire block and get it fully refunded.

app/posthandler/fee_refund_decorator.go
```diff
@@ -0,0 +1,127 @@
...
+	refundRecipientAccount := frd.accountKeeper.GetAccount(ctx, refundRecipient)
+	if refundRecipientAccount == nil {
+		return errors.Wrapf(sdkerrors.ErrUnknownAddress, "refund recipient address: %s does not exist", refundRecipientAccount)
+	}
+
+	if err := frd.processRefund(frd.bankKeeper, ctx, refundRecipientAccount, coinsToRefund); err != nil {
+		return err
+	}
+
+	events := sdk.Events{newFeeRefundEvent(coinsToRefund, refundRecipient)}
```
evan-forbes, line 72: I'm not 100% sure adding an extra event to each tx is worth it, especially since all the information that a user/wallet needs is already contained in gas used

evan-forbes (commented):
nice! just as an optimization, I was trying to think of way to avoid transfering the funds twice, once to deduct gas and then again to refund. I think we might be able to do this if we change the existing antehandler that deducts funds to only check if enough funds are present. Then this post handler might be able to panic if there are not enough funds after the tx has been executed.

There might be issues with that approach when it comes to the guarantees that we want to provide in ProcessProposal, but would have to examine further.

I'm not sure if it would be worth it, but in the future we might consider only checking if there are enough funds to remove

evan-forbes:
yeah definitely this is a future thing

> in the future we might consider only checking if there are enough funds to remove

app/posthandler/fee_refund_decorator.go
rootulp: 26fc642efa31b9e176a30c6652ff0dbcb63779b5

rootulp:
I'm investigating the gas cost to perform the refund. 

```
test case: part of the fee should be refunded
gasConsumedDuringRefund 61934
gasConsumedDuringRefund 61931
test case: at most half of the fee should be refunded
gasConsumedDuringRefund 62030
gasConsumedDuringRefund 62027
test case: refund should be sent to fee payer if specified
gasConsumedDuringRefund 73154
gasConsumedDuringRefund 73151
```

I plan on investigating why the log is emitted twice per test case. The max appears to be ~73K gas which seems like a lot considering an ordinary [send transaction](https://celenium.io/tx/d86031d574b398d1cd4f77f45ea8c13b2c3a95dbcca3167305551e95397d2bff) consumes ~91K gas.

rootulp:
Disregard the previous comment. I was inspecting the wrong gas meter. 

```
test case: part of the fee should be refunded
gasConsumedDuringRefund 14405
test case: at most half of the fee should be refunded
gasConsumedDuringRefund 14438
test case: refund should be sent to fee payer if specified
gasConsumedDuringRefund 14438
```

I think we can round up to 15,000 gas for the constant.

rootulp:
I'm investigating how to handle the `simulate` flag. The only other posthandler implementation in cosmos-sdk is [tips](https://github.com/celestiaorg/cosmos-sdk/blob/bc89c06feb984c6bc6b9ebf453f9db24b6794b74/x/auth/posthandler/tips.go#L27) which ignores `simulate`. Created https://github.com/cosmos/cosmos-sdk/issues/18907 for docs on the `simulate` flag.

rootulp:
[for post-break] @cmwaters @evan-forbes this PR is re-reviewable

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// AnteHandle implements the cosmos-sdk AnteHandler interface. Note: the
+// AnteHandler interface is also used for post-handlers.
+func (d RefundGasRemainingDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if err := d.maybeRefund(ctx, tx, simulate); err != nil {
+		return ctx, err
+	}
+	return next(ctx, tx, simulate)
+}
```
evan-forbes, line 52: blocking

can we get an if statement for the app version here along with a quick unit test to make sure its not making any changes for appversion v1?

evan-forbes (commented):
super dope, well done! really excited to get this out! people are going to love this

had a few blocking comments around versioning and some potential overflows, but other than that I think this is good to go!

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,144 @@
...
+
+// maybeRefund conditionally refunds a portion of the tx fee to the fee payer.
+func (d RefundGasRemainingDecorator) maybeRefund(ctx sdk.Context, tx sdk.Tx, simulate bool) error {
+	// If this is a simulation, then no refund needs to be issued.
+	if simulate {
+		return nil
+	}
```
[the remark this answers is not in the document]
rootulp: Yea, as far as I can tell `simulate=true` means no state changes should be made so gas isn't actually deducted from the fee payer and therefore no refund needs to be issued.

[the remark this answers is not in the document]
rootulp: I can elaborate in the comment if that would help.

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,144 @@
...
+	gasMeter := ctx.GasMeter()
+	ctx = ctx.WithGasMeter(sdk.NewInfiniteGasMeter())
+	// Restore the original gas meter after this posthandler is done.
+	defer ctx.WithGasMeter(gasMeter)
+
+	// If the gas meter doesn't have enough gas remaining to cover the
+	// refund gas cost, then no refund needs to be issued.
+	if gasMeter.GasRemaining() < RefundGasCost {
+		return nil
+	}
```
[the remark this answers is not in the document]
rootulp: I think we want the tx to succeed and skip any gas refund processing if this conditional is true.
If this conditional caused the tx to fail, then tx submitters would need to over-estimate gas limit by a bit more than they currently do just to get a gas refund.

app/posthandler/refund_gas_remaining.go
rootulp: Great call! d69402889f896b2dc2d548860c5dce2f232d8b56

app/posthandler/refund_gas_remaining.go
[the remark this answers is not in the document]
rootulp: 30b6e5011fdb2487a76b4c31d994f13bd1521031

cmwaters:
>  If a tx doesn't specify a fee payer, unspent gas should be refunded to the tx submitter.

If the tx doesn't have a fee payer then I don't think there should be any refund

rootulp:
>> If a tx doesn't specify a fee payer, unspent gas should be refunded to the tx submitter.

> If the tx doesn't have a fee payer then I don't think there should be any refund

hmm I may not have phrased the test case correctly but I meant to say that if the tx doesn't use the feegrant module then the gas remaining should be refunded to the signer and not the fee granter.

app/posthandler/refund_gas_remaining_test.go
```diff
@@ -0,0 +1,153 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			gotCtx, err := decorator.AnteHandle(tc.ctx, tc.tx, tc.simulate, tc.next)
+			if tc.wantErr {
+				assert.Error(t, err)
+				return
+			}
+			assert.Equal(t, tc.wantCtx, gotCtx)
+		})
+	}
+}
```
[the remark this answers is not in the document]
rootulp: it's silly but ante handlers just pass contexts from one to the next so the only way to verify that a refund was issues is to make assertions on the wantCtx.

app/posthandler/suite_test.go
```diff
@@ -0,0 +1,182 @@
...
+	require.NoError(err)
+	require.Equal(abci.CodeTypeOK, resp.Code)
+}
+
+func (s *RefundGasRemainingSuite) TestDecorator() {
+	t := s.T()
+
+	type testCase struct {
+		name                string
+		gasLimit            uint64
```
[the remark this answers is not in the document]
rootulp: Good call, added two test cases in 617a5e2b87b98e43cd762b0312fe2c4ebf338006 and 6af56c4e1ca6c9f8b77bb27ac5cc179fb704ab65

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,144 @@
...
+		return errors.Wrapf(err, "error refunding %s from fee collector module account to %s", refund, recipient)
+	}
+
+	return nil
+}
+
+// getRefund returns the coins that should be refunded to the recipient.
+func getRefund(gasMeter sdk.GasMeter, feeTx sdk.FeeTx) sdk.Coins {
+	gasPrice := getGasPrice(feeTx)
+	toRefund := gasPrice.Amount.MulInt64(int64(gasMeter.GasRemaining())).TruncateInt()
```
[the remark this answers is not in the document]
rootulp: Good point, prob safer to do: 925e81b05eb91f6bb6f8edf7e192023d7d37145e

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return feeTx.FeePayer()
+}
+
+// getGasPrice returns the gas price of the feeTx.
+// gasLimit * gasPrice = fee. So gasPrice = fee / gasLimit.
+func getGasPrice(feeTx sdk.FeeTx) sdk.DecCoin {
+	feeCoins := feeTx.GetFee()
+	gas := feeTx.GetGas()
+	gasPrice := sdk.NewDecFromInt(feeCoins.AmountOf(appconsts.BondDenom)).Quo(sdk.NewDec(int64(gas)))
```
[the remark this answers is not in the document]
rootulp: Resolved by removing `int64()` usage

cmwaters:
You mean it should be refunded to the fee granter as the fee granter paid not the signer

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,155 @@
...
+	return next(ctx, tx, simulate)
+}
+
+// maybeRefund conditionally refunds a portion of the tx fee to the fee payer.
+func (d RefundGasRemainingDecorator) maybeRefund(ctx sdk.Context, tx sdk.Tx, simulate bool) error {
+	// If this tx was included in a v1 block then no refund needs to be issued.
+	// Gas refunds are only applicable for blocks >= v2.Version
+	if ctx.BlockHeader().Version.App == v1.Version {
+		return nil
+	}
```
cmwaters, line 61: 👍

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return nil
+}
+
+// getRefund returns the coins that should be refunded to the recipient.
+func getRefund(gasMeter sdk.GasMeter, feeTx sdk.FeeTx) sdk.Coins {
+	gasPrice := getGasPrice(feeTx)
+	toRefund := gasPrice.Amount.MulInt(sdk.NewIntFromUint64(gasMeter.GasRemaining())).TruncateInt()
+	maxToRefund := MaxPortionOfFeeToRefund.MulInt(feeTx.GetFee().AmountOf(appconsts.BondDenom)).TruncateInt()
```
cmwaters, line 127: [block]

So I think this should be proportional to the gas used * gas price not the fee. If someone submitted a send tx with a gas limit of 50 million (or whatever is needed to full a block) they would prevent others from using the block at half the gas price they provided. If we use `gas_used * gas_price` then they will get charged basically the full price as they will only get refunded for the 100,000 + 50% of the gas used

cmwaters (changes requested):

app/posthandler/refund_gas_remaining.go
rootulp: Sorry I don't quite follow your example. Where does 100,000 come from? 

> So I think this should be proportional to the gas used * gas price not the fee.

this refers to `maxToRefund`? If yes are you proposing something like:

```go
maxToRefund := gas_used * gas_price * SomeConstant
```

cmwaters: 100,00 in this example is the actual gas used for the message

cmwaters: Yes I'm proposing something like this

rootulp: Got it. Here are some names + numbers we can use to make it clear what we're referring to

```go
// constants
MaxPortionOfFee := .5
MaxPortionOfFeeUtilizedToRefund := .5

// user specified
gasLimit := 50,000,000
gasPrice := 1
fees := 50,000,000
// execution of tx
gasUsed := 100,000

// existing method
maxToRefund := MaxPortionOfFee * fees // 25,000,000
// proposed method
maxToRefund := gasUsed * gasPrice * MaxPortionOfFeeUtilizedToRefund // 50,000
```

rootulp: Good point. This is actually the proposed design in https://github.com/cosmos/cosmos-sdk/issues/2150#issue-354029283 and setting `MaxPortionOfFeeUtilizedToRefund` to 1.5 or 2

cmwaters: Yes exactly!

rootulp: Resolved in 161664b11a22b089c2ba823691b18dc15b1456e0 but need to improve tests.

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return nil
+}
+
+// getRefund returns the coins that should be refunded to the recipient.
+func getRefund(gasMeter sdk.GasMeter, feeTx sdk.FeeTx) sdk.Coins {
+	gasPrice := getGasPrice(feeTx)
+	toRefund := gasPrice.Amount.MulInt(sdk.NewIntFromUint64(gasMeter.GasRemaining())).TruncateInt()
+	maxToRefund := MaxRefundGasConsumed.Mul(gasPrice.Amount.MulInt(sdk.NewIntFromUint64(gasMeter.GasConsumed()))).TruncateInt()
+	amountToRefund := minimum(toRefund, maxToRefund)
```
cmwaters, line 128: 👍

cmwaters (approved):
Great work!

The test cases are very thorough, couldn't think of anything you may have missed. 

Given this diverges from the standard SDK, we should include a small specification somewhere in the spec directory (can be a follow up PR)

app/posthandler/refund_gas_remaining.go
```diff
@@ -0,0 +1,155 @@
...
+// MaxRefundGasConsumed is the maximum amount of gasConsumed that can be refunded.
+var MaxRefundGasConsumed = sdk.NewDecWithPrec(5, 1) // 50%
+
+// RefundGasCost is the amount of gas consumed during the execution of this
+// posthandler. If a tx reaches this posthandler with gas remaining in excess of
+// this amount, then a refund will be issued for the gas remaining -
+// RefundGasCost.
+//
+// NOTE: this value was determined empirically.
+const RefundGasCost = 15_000
```
cmwaters, line 26: I assume if there are any changes to the gas cost that the tests will catch it

cmwaters (approved):
Still LGTM

Let's bring this up one last time in the call tomorrow and if everyone's still happy with the design we can merge this

rootulp:
During a meeting, we decided to stop pursuing this line of work because the mitigation ideas for the [DoS attack](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-11.md#dos-attack) are complicated and we'd rather prioritize other work.

Issue #2916, opened 2023-12-08
Increase test coverage for governance modifiable and governance unmodifiable params


rootulp:
## Context

Our specs have [params.md](https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/params.md) which has a column for whether a parameter is governance modifiable. 

## Problem

AFAICT there aren't tests that verify whether the parameters in the table are governance modifiable. https://github.com/celestiaorg/celestia-app/blob/ce1c6cd93278031bb56ed9910e5c706aa2c3c6cc/x/paramfilter/test/param_filter_test.go#L17 exists but it only verifies the 4 blocked params aren't modifiable: https://github.com/celestiaorg/celestia-app/blob/ce1c6cd93278031bb56ed9910e5c706aa2c3c6cc/app/app.go#L733-L746

## Proposal

Increase the test coverage of the blocked / unblocked params. Ideally each param in the table is covered by a test.

fahimahmedx:
Taking this on!

PR #2919, opened 2023-12-09
feat!: prevent modifications on governance unmodifiable params

fahimahmedx:
## Overview
1 of 2 PRs that will address #2916. This PR adds test cases for the governance unmodifiable params.
The three params below are not added as they are already constants in `mint/types/constants.go`

> mint.DisinflationRate
> mint.InitialInflationRate
> mint.TargetInflationRate

## Checklist


fahimahmedx:
In `paramfilter/test/param_filter_test.go`, is this test still valid considering `staking.MaxValidators` can not be changed by governance?
image

rootulp:
> In paramfilter/test/param_filter_test.go, is this test still valid considering staking.MaxValidators can not be changed by governance?

Good point. params.md claims that MaxValidators is not governance-modifiable
https://github.com/celestiaorg/celestia-app/blame/74b3404679e8b258b58fad5b791794f522baa460/specs/src/specs/params.md#L64

So in my opinion that:
1. The existing test should be re-written to include a valid change (i.e. modify a param that is governance modifiable) and an invalid change (modify a param that is not governance modifiable or not even a real param) and verify that no changes are observed to the first param.
2. Add a separate test that verifies that `MaxValidators` is not governance modifiable

app/app.go
```diff
@@ -734,14 +734,22 @@ func (app *App) setPostHanders() {
...
 		// consensus.validator.PubKeyTypes
 		{baseapp.Paramspace, string(baseapp.ParamStoreKeyValidatorParams)},
+		// consensus.Version.AppVersion
+		{baseapp.Paramspace, string(baseapp.ParamStoreKeyVersionParams)},
+		// staking.BondDenom
+		{stakingtypes.ModuleName, string(stakingtypes.KeyBondDenom)},
+		// staking.MaxValidators
+		{stakingtypes.ModuleName, string(stakingtypes.KeyMaxValidators)},
+		// staking.UnbondingTime
+		{stakingtypes.ModuleName, string(stakingtypes.KeyUnbondingTime)},
```
rootulp, line 752: [blocking] this adds more params to the list of blocked params which is potentially something we need to do so that there isn't a divergence between what is stated in params.md and what is actually modifiable. However, it's not clear if these need to be added without appropriate test coverage.

IMO we should tackle this in a few steps:
1. Write a table-driven test with test cases for each param listed in params.md. The test should attempt to modify the param and see if it was actually modified. Each test case should have a boolean field that indicates if modification was expected to happen (i.e. that param is governance modifiable) or not happen (i.e. that param is not governance modifiable).
2. Once we have the test above, we can compare the test results to params.md. If any divergences exist, create a GH issue with the divergence so that we can get alignment on whether the param should be governance modifiable.
3. [Optional] If there are any params that are governance modifiable and shouldn't be, we may need to add them to this BlockedParams list.

x/paramfilter/test/param_filter_test.go
```diff
@@ -34,30 +33,6 @@ func TestParamFilter(t *testing.T) {
...
-	err := handler(ctx, p)
-	require.NoError(t, err)
-
-	ps := app.StakingKeeper.GetParams(ctx)
-	require.Equal(t, ps.MaxValidators, uint32(1))
-
-	// ensure that we're throwing out entire proposals if any of the changes are blocked
-	for _, p := range app.BlockedParams() {
-		// try to set the max validators to 2, unlike above this should fail
-		validChange := proposal.NewParamChange(stakingtypes.ModuleName, string(stakingtypes.KeyMaxValidators), "2")
```
rootulp, line 50: [blocking] good catch! IMO we should update this test to use a governance modifiable param instead of deleting it entirely.

rootulp (commented):
Thanks for tackling this! I left a few comments on how I think we can proceed

rootulp:
> This PR adds test cases for the governance unmodifiable params.

@fahimahmedx is it possible you didn't push a commit? I don't see the new test cases.

fahimahmedx:
@rootulp Hi Rootul, thank you for your fast review!

Regarding the adding test cases, I believe that was a typo on my part as the current test case should test the newly added parameters.

As another question regarding the `consensus.block.TimeIotaMs` param, I noticed the `consensustypes` from `cosmos-sdk` wasn't added and so I don't have access to `consensustypes`. I also noticed that in [/x/consensus/types](https://github.com/cosmos/cosmos-sdk/tree/main/x/consensus/types), there is no `params.go` file to contains a field along the lines of `TimeIotaMs`. Any suggestions on how I might get this parameter?

> go: github.com/celestiaorg/celestia-app/app imports
>         github.com/cosmos/cosmos-sdk/x/consensus/types: module github.com/cosmos/cosmos-sdk@latest found (v0.50.1, replaced by github.com/celestiaorg/cosmos-sdk@v1.20.0-sdk-v0.46.16), but does not contain package github.com/cosmos/cosmos-sdk/x/consensus/types

rootulp:
>  is it not the case that the already existing test case should test the newly added parameters (for the unmodifiable params specifically)? I will add additional test cases for the modifiable params in a seperate PR.

Rephrasing a bit to something I understand: does the existing test `TestParamFilter` already cover unmodifiable params? IMO it doesn't because it only tests the `BlockedParams()` which as you've noticed doesn't contain all of the claimed unmodifiable params in params.md

So I think we should add test cases for those params _before_ modifying the `BlockedParams` to see if they're actually necessary to add to `BlockedParams`.

cmwaters (commented):
Can we rephrase the name of this PR from test to feat! as it's a breaking change which adds functionality

fahimahmedx:
Hi @rootulp, I noticed that that there isn't a `mint.BondDenom` in celestia's `cosmos-sdk`. Rather, there is a `mint.MintDenom` as seen [here](https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.46.x-celestia/x/mint/types/params.go#L33).

Is the docs supposed to say `mint.MintDenom` instead?

Additionally, as parameters like `mint.DisinflationRate` are constants as seen [here](https://github.com/celestiaorg/celestia-app/blob/d13d002cee0fea22940b1eb2a0e5a67afb05a695/x/mint/types/constants.go#L20), is that enough reasoning to prove this governance unmodifiable parameter is indeed unmodifiable?

rootulp:
Sorry for my delay @fahimahmedx. 

> I noticed that that there isn't a mint.BondDenom in celestia's cosmos-sdk. Rather, there is a mint.MintDenom as seen [here](https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.46.x-celestia/x/mint/types/params.go#L33). Is the docs supposed to say mint.MintDenom instead?

celestia-app doesn't use the mint module defined in celestiaorg/cosmos-sdk so [MintDenom](https://github.com/celestiaorg/cosmos-sdk/blob/bc89c06feb984c6bc6b9ebf453f9db24b6794b74/x/mint/types/params.go#L33) doesn't need to be covered by these tests. celestia-app uses the x/mint module defined in this repo so `mint.BondDenom` is likely referring to [this]( https://github.com/celestiaorg/celestia-app/blob/f57f5c104ea03b644bb6e50e262f212036eba99e/proto/celestia/mint/v1/mint.proto#L33).

> is that enough reasoning to prove this governance unmodifiable parameter is indeed unmodifiable?

Yep. Since mint.InitialInflationRate, mint.DisinflationRate, and mint.TargetInflationRate are constants, I think they don't strictly need to be covered by tests here.

fahimahmedx:
Hi @rootulp, hope your holidays are going well!

While trying to create a test for `mint.BondDenom`, I get a panic of `panic: parameter Minter not registered` when trying to submit a `ParamChange` proposal for subspace `minttypes.ModuleName` and key `string(minttypes.KeyMinter)`.

When investigating this, I noticed that in [app.go line 318](https://github.com/celestiaorg/celestia-app/blob/a0a14a3c2bc07d3635528b40488d71ab6c5b1ccf/app/app.go#L318), we don't actually pass in a `paramtypes.Subspace` variable into the `NewKeeper` function which the `NewKeeper` function would use to set the `KeyTable`.

As parameters are registered in a `paramtypes.Subspace` in Cosmos SDK, which means that if a parameter is not registered in the subspace, it cannot be modified it, is this enough justification to show that `mint.BondDenom` is not modifiable by governance?

rootulp:
> is this enough justification to show that mint.BondDenom is not modifiable by governance?

Yes I think it is. Thanks for your thorough research!  We may want to include a note in the mint module README.md that the mint module doesn't have any governance modifiable parameters for the reasons you've outlined. We can include the note in a separate PR, it doesn't have to be in this one.

fahimahmedx:
Hi @rootulp, I've implemented the changes and the PR is now ready for review!
Addressing [point 2 on this comment](https://github.com/celestiaorg/celestia-app/pull/2919#discussion_r1421568115), the following parameters that were supposed to be unmodifiable were able to be modified through a proposal request.

- `bank.SendEnabled`
- `conensus.validator.PubKeyTypes`
- `consensus.Version.AppVersion`
- `staking.BondDenom`
- `staking.MaxValidators`
- `staking.UnbondingTime`

Do I proceed with opening an issue for each parameter (given this divergence) to get alignment on if they should be modifiable or not?

I have added these parameters to `App.BlockedParams()` to prevent them from being modified in the meanwhile.

rootulp:
Thanks for all your hard work here! 

> Do I proceed with opening an issue for each parameter (given this divergence) to get alignment on if they should be modifiable or not?

Yea can you please open one GH issue and reference the list of params that _are_ governance modifiable even though the specs claim they shouldn't be modifiable. Then we can get input from the team (cc: @evan-forbes ) on how to proceed. 

IMO we should have tests covering all the params before we add any params to the blocked list.

specs/src/specs/params.md
```diff
@@ -64,3 +64,7 @@ are blocked by the `x/paramfilter` module.
...
 | staking.MinCommissionRate                     | 0.05 (5%)                                   | Minimum commission rate used by all validators.                                                                                                                                                 | True                      |
 | staking.UnbondingTime                         | 1814400 (21 days)                           | Duration of time for unbonding in seconds.                                                                                                                                                      | False                     |
+
+### Tests
+See [/x/paramfilter/test/governance_params_test.go](/x/paramfilter/test/governance_params_test.go) for tests that attempt to modify a parameter.
+Note that `TimeIotaMs` is not in `conensus.block`, so a governance proposal can not be submitted to modify it. Additionally, the `x/mint` module does not have any governance modifiable params as justified in [/x/mint/README.md#params](/x/mint/README.md#params)
```
rootulp, line 70: [question] if `TimeIotaMs` is not in `consensus.block`, should it be removed from the list of params above? Do you know if this field exists elsewhere?

x/paramfilter/test/governance_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+	suite.ctx = suite.app.BaseApp.NewContext(false, tmproto.Header{})
+	suite.govHandler = params.NewParamChangeProposalHandler(suite.app.ParamsKeeper)
+	suite.pph = paramfilter.NewParamBlockList([2]string{})
+
+	minter := minttypes.DefaultMinter()
+	suite.app.MintKeeper.SetMinter(suite.ctx, minter)
+}
+
+func TestHandlerTestSuite(t *testing.T) {
+	suite.Run(t, new(HandlerTestSuite))
```
rootulp, line 44: I expect this test to fail when running `make test-short`. If that's the case, we may want to skip it in short mode via something like:

```go
	if testing.Short() {
		t.Skip("skipping handler test suite in short mode.")
	}
	suite.Run(t, new(HandlerTestSuite))
```

x/paramfilter/test/governance_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/cosmos/cosmos-sdk/baseapp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
+	params "github.com/cosmos/cosmos-sdk/x/params"
+	"github.com/cosmos/cosmos-sdk/x/params/types/proposal"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+type HandlerTestSuite struct {
```
rootulp, line 24: [question] what does `Handler` refer to?

x/paramfilter/test/governance_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+func TestHandlerTestSuite(t *testing.T) {
+	suite.Run(t, new(HandlerTestSuite))
+}
+
+func (suite *HandlerTestSuite) TestUnmodifiableParameters() {
+	testCases := []struct {
+		name     string
+		proposal *proposal.ParameterChangeProposal
+		onHandle func()
```
rootulp, line 51: [question] what is `onHandle` meant to refer to? It looks like verification logic for each test case

x/paramfilter/test/governance_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+		{
+			"bank.SendEnabled",
+			testProposal(proposal.ParamChange{
+				Subspace: banktypes.ModuleName,
+				Key:      string(banktypes.KeySendEnabled),
+				Value:    `[{"denom": "test", "enabled": false}]`,
+			}),
+			func() {
+				sendEnabledParams := suite.app.BankKeeper.GetParams(suite.ctx).SendEnabled
+				suite.Require().Equal([]*banktypes.SendEnabled{banktypes.NewSendEnabled("test", false)}, sendEnabledParams)
```
rootulp, line 64: [optional] consider extracting variables to make the test more readable:

```suggestion
				got := suite.app.BankKeeper.GetParams(suite.ctx).SendEnabled
				want := []*banktypes.SendEnabled{banktypes.NewSendEnabled("test", false)}
				suite.Require().Equal(want, got)
```

x/paramfilter/test/governance_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+				suite.Require().Equal(
+					stakingParams.UnbondingTime,
+					time.Duration(1))
+			},
+			false,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
rootulp, line 150: [question] is this idiom needed if this test isn't executed asynchronously? I've observed that this line is needed for concurrent code but isn't strictly necessary in non concurrent code.

```suggestion
```

app/app.go
```diff
@@ -736,12 +736,16 @@ func (*App) BlockedParams() [][2]string {
...
```
rootulp, line 748: [blocking] this is a breaking change so if it is included in this PR then the PR title needs a `!` to conform to conventional commits.

IMO we shouldn't add this to the blocked list yet. We should create a GH issue with the list of params that are governance modifiable even though the specs claim they shouldn't be. We should get team alignment on what to do for those params. Then we may update this list or update the specs accordingly. cc: @evan-forbes

specs/src/specs/params.md
fahimahmedx: For context, `ConsensusParams` in `celestia-core/abci/types/types.pb.go` as seen [here](https://github.com/celestiaorg/celestia-core/blob/e67de4517780e8db93c74d8eec4e9c66edffbc99/abci/types/types.pb.go#L2809) uses `BlockParam` from `celestia-core/abci/types/types.pb.go` instead of `types1.BlockParams` from `celestia-core/proto/tendermint/types/params.pb.go`.

This version of `BlockParam` does not contain `TimeIotaMs`, however I am not sure why `types1.BlockParams` isn't used instead, or if `ConsensusParams` from `celestia-core/proto/tendermint/types/types.pb.go` is another valid alternative that could be used instead in `baseapp.go`. 

With this context in mind and knowing the field exists somewhere else (but seems to not be used), I think it would be good to clarify if the intended version of `BlockParams` is being used in the first place before it is removed from `params.md`. Do you have any thoughts on this or know of anyone else who might?

fahimahmedx: For further context on why I ask this, it is because I use `BaseApp.GetConsensusParams()` when testing if the governance modification was successful. [This commit here](https://github.com/celestiaorg/celestia-app/pull/2919/commits/39fede889d8cf6504427c1df639792fae1994764) shows this code.

x/paramfilter/test/governance_params_test.go
fahimahmedx: Initially it referred to handling test cases, but I have now improved the naming.

x/paramfilter/test/governance_params_test.go
fahimahmedx: Yes that is correct. I have updated the naming to be `subTest`.

x/paramfilter/test/governance_params_test.go
fahimahmedx: Thank you for spotting that! I have removed it now.

app/app.go
evan-forbes: yeah any consens breaking change needs a CIP

@fahimahmedx, this is consensus breaking since nodes running this won't change a param, where nodes that are not will change the param

evan-forbes: the CIP doesn't have to be super in depth, since the change is pretty small, but we do still need to explain the benefits of adding or removing any params

rootulp:
Converting this to draft until we have a CIP with alignment.

Issue #2894, opened 2023-12-04
Tool to monitor network upgrade status and auto submit `MsgTryUpgrade`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/2832
https://github.com/celestiaorg/celestia-app/pull/2880#issuecomment-1835786783

## Problem

There now exists a new msg type that initiates an upgrade (`MsgTryUpgrade`).

## Proposal

Create a stand-alone binary (likely that lives in `/tools`) that monitors that status of upgrades on the network. When the tool recognizes that the upgrade threshold has been met, it should auto-submit a transaction that includes the message to initiate an upgrade. The tool likely needs to be configured with a GRPC endpoint to a full-consensus node. It likely also needs a keyring that is funded with a small amount of tia to submit the transaction.

PR #2922, opened 2023-12-11
feat: upgrademonitor

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2894

Demo on [Loom](https://www.loom.com/share/2f7c08e950d14b1eab5857e059948342?sid=1dee106f-da93-4d13-a136-ba91f35d92d1)


rootulp:
On second thought, this tool can accept a pre-signed transaction that is generated but not published. When the upgrademonitor detects that the network is upgrade-able, it can publish the pre-signed transaction.

.gitignore
```diff
@@ -16,4 +16,6 @@ square/testdata
 **/*.html
 .release-env
 **/*.DS_Store
-vendor
\ No newline at end of file
+vendor
+tools/upgrademonitor/upgrademonitor
+tools/upgrademonitor/*.json
```
rootulp, line 21: added so that we don't accidentally include the upgrademonitor binary or unsigned / signed transaction JSONs in version control

tools/upgrademonitor/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+					return nil // stop the upgrademonitor
+				}
+			}
+		}
+	},
+}
+
+func Execute() {
+	// Bind the version variable to the --version flag
+	rootCmd.Flags().Uint64Var(&version, "version", defaultVersion, "version to monitor")
```
cmwaters, line 52: To make this a little more convenient for the user, you could use the grpc endpoint to query the current version of the chain and assume by default that the version to be upgraded to is the next one.

Currently only the next version number is supported i.e. it has to be monotonically increasing

tools/upgrademonitor/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/celestiaorg/celestia-app/tools/upgrademonitor
```
cmwaters, line 1: What was the rationale for this having it's own go.mod?

tools/upgrademonitor/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+	// Bind the grpcEndpoint variable to the --grpc-endpoint flag
+	rootCmd.Flags().StringVar(&grpcEndpoint, "grpc-endpoint", defaultGrpcEndpoint, "GRPC endpoint of a consensus node")
+	// Bind the pollFrequency variable to the --poll-frequency flag
+	rootCmd.Flags().Int64Var(&pollFrequency, "poll-frequency", defaultPollFrequency, "poll frequency in seconds")
+	// Bind the pathToTransaction variable to the value provided for the --auto-publish flag
+	rootCmd.Flags().StringVar(&pathToTransaction, "auto-publish", defaultPathToTransaction, "auto publish a signed transaction when the network is upgradeable")
```
cmwaters, line 58: This flag kind of has to be set else it doesn't really make sense. I'd suggest instead that we have the path built in directly in the CLI  ie.

```
Use: "upgrademonitor path-to-signed-tx.json"
```
and require one argument

We should also parse this into the tx bytes first so that if their is an error it fails on start up and not when trying to submit.

tools/upgrademonitor/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+				}
+				fmt.Printf("version: %v, voting: %v, threshold: %v, total: %v\n", version, resp.GetVotingPower(), resp.GetThresholdPower(), resp.GetTotalVotingPower())
+
+				if internal.IsUpgradeable(resp) {
+					fmt.Printf("the network is upgradeable so publishing %v\n", pathToTransaction)
+					resp, err := internal.Publish(conn, pathToTransaction)
+					if err != nil {
+						return err
+					}
+					fmt.Printf("published transaction: %v\n", resp.TxHash)
```
cmwaters, line 42: I think we also need to check that resp.Code is zero else print out the log (as the transaction failed)

cmwaters (commented):
Great work. Left a few comments to go over

tools/upgrademonitor/cmd/root.go
rootulp: Good point, can do!

tools/upgrademonitor/go.mod
rootulp: hmm upgrademonitor is distinct from celestia-app in that the binary will be used independently of celestia-app. A distinct go.mod allows upgrademonitor to manage its own dependencies.

Your question made me realize that we could even extract upgrademonitor entirely to its own repo.

tools/upgrademonitor/cmd/root.go
rootulp: 0ec39692e14af1b13da3bfae8adaece89ed254ad

tools/upgrademonitor/cmd/root.go
rootulp: 7271be8f0f1531b88defb0683d694ac2722a5d43

tools/upgrademonitor/cmd/root.go
rootulp: 50fcc9a9e230e04817f60588fa5f6738fcd6f0d5

tools/upgrademonitor/go.mod
rootulp: If we want it in a stand alone repo: https://github.com/celestiaorg/upgrade-monitor/pull/1

rootulp:
Closing in favor of putting this in a stand-alone repo.

PR #2928, opened 2023-12-12
test: simple predicate verification test for snark accounts

### LINKED ISSUE #2923: Integration and testing of SNARK library for simple predicate verification
This issue marks the initial step towards implementing SNARK accounts, addressing one of the tasks in https://github.com/celestiaorg/celestia-app/issues/2905.

The task involves integrating a SNARK library (TB specified) into our system. Its application will be demonestrated through a test involving a simple predicate (e.g., proof of correct multiplication). The results of this test (plus some further development on top of it in the future) will  serve as a mock SNARK account predicate and will be used for assessing the SNARK transaction flow and ensuring correct verification.

staheri14:
Closes #2923


staheri14:
cc @karotoka for review

pkg/snarkaccount/snarkaccount_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+// SquareRoot represents a simple circuit to compute the square root.
+// It defines three inputs::
+// - Input X: The base of the square operation.
+// - Input Z: An additional input with no effect on the constraints.
+// - Input Y: The result of squaring X (i.e., X * X).
+type SquareRoot struct {
+	X frontend.Variable `gnark:"x"`       // Base of the square operation
+	Y frontend.Variable `gnark:",public"` // Result of the square operation
+	Z frontend.Variable `gnark:",public"` // Additional input, but not used in constraints
```
rootulp, line 21: silly question why is the leading comma necessary in ",public"?

rootulp (approved):
Overall LGTM. Thanks for the clearly defined test!

pkg/snarkaccount/snarkaccount_test.go
staheri14: Good question. I am not sure either, this is the format used and instructed to be used according to the library's documentation.

evan-forbes (commented):
dope! I think we can add this to whatever the snarkaccount module is just as an example predicate (can still be a test ofc, just something to indicate that this is not testing something specific to the protocol)

staheri14:
> dope! I think we can add this to whatever the snarkaccount module instead of the pkg, but other than that this example predicate LGTM

@evan-forbes I initially considered your approach for this PR, but I thought code clarity could be improved by introducing a dedicated package in the 'pkg' folder specifically for snark-related logic e.g., verification logic. This package would include a generic interface and house multiple implementations of this interface. The 'x/snarkaccount' [assuming the name would be `snarkaccount`] module would then utilize this package. What are your thoughts on this strategy? Also, if the pkg folder is intended for a different use, I'd appreciate any insights you might have.

evan-forbes (dismissed):
We can separate out some of the logic, I don't see that being a big deal. The initial thinking was assuming that this code would only be imported by the snarkaccount package, which might be a bad assumption! 

It's definitely not necessary and could be done in a different way, but it would also be nice to have unique package names. The reasoning being that we then won't have to fix the imports each time to specify which snarkaccount package we're importing from.

staheri14:
> it would also be nice to have unique package names. The reasoning being that we then won't have to fix the imports each time to specify which snarkaccount package we're importing from.

Certainly, I update the package name to `snakrutils`. I am open to alternative names as well.

evan-forbes (approved):

evan-forbes (approved):

staheri14:
@evan-forbes Shall I merge this PR?

evan-forbes:
> Shall I merge this PR?

that's a good question... last sync we didn't have time to discuss the implementation at all. since the implementation is very oriented at being a prototype, we don't have a ton of confidence in the implementation details yet. Therefore, unfortunately, we might want to keep any changes isolated from main. We should definitely keep all implementation work together on a branch somewhere. 

feature branches are no fun, but I'm not sure we have a better mechanism. provided we're diligent about keeping commits isolated in scope, that should make things easier for future us should we decide to begin maintaining it.

staheri14:
@evan-forbes A feature branch sounds good to me, I'll proceed with making one named `snark-account` (or if there is a particular naming convention to follow, please let me know and I'll name it accordingly).

staheri14:
Will open it in the future if need be.

PR #2930, opened 2023-12-13
feat!: remove namespace and share version from `MsgPayForBlobs`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2932

## Notes for reviewers
- a few tests had to be updated because the removal of these fields means that the MsgPayForBlobs encoded in the PFB namespace are slightly smaller. Tests that hard-coded share indices had to be adjusted slightly.
- proto-check-breaking is expected to fail


rootulp:
Ready for review but I don't plan on merging until @evan-forbes is back and has time to review.

proto/celestia/blob/v1/tx.proto
```diff
@@ -18,19 +18,14 @@ message MsgPayForBlobs {
...
   // share_commitments is a list of share commitments (one per blob).
   repeated bytes share_commitments = 4;
-  // share_versions are the versions of the share format that the blobs
-  // associated with this message should use when included in a block. The
-  // share_versions specified must match the share_versions used to generate the
-  // share_commitment in this message.
-  repeated uint32 share_versions = 8;
+  // 8 was previously used for share_versions;
+  reserved 8;
 }
```
evan-forbes, line 29: blocking

since this is breaking, do we need to create a v2 of the proto msg?

I think https://github.com/celestiaorg/celestia-app/issues/2932#issuecomment-1856474966 makes sense, and we might have to add an intermediate type to convert between v1 and v2 so that we don't have to have an extra copy of the logic to execute the msg.

evan-forbes (commented):
yikes, big-oof on us leaving these in. :see_no_evil:

rootulp:
Yea good point. Going to close this PR and leave the issue open b/c I think it's prob safe to introduce a new type like `MsgPayForBlobsV2` without these fields.

PR #2934, opened 2023-12-13
feat(txsim): add CLI Flag for chainID

### LINKED ISSUE #2471: Add CLI Flag for chainID in txsim for improved transaction control
# Problem
txsim lacks a CLI flag for specifying the chain ID. Instead, it dynamically determines the chain ID by querying the target endpoint and assigns this ID to the generated transactions. This approach assumes that the transactions are intended for the same chain ID as the receiver's. However, if the endpoint changes its chain ID, submitting transactions may encounter issues, such as the master account having insufficient balance in the new chain or other failures due to a mismatch in chain IDs. This can create confusion for users trying to identify the root cause of the failure. Furthermore, there's a possibility that the transaction will succeed, but it might inadvertently be submitted to the wrong chain, diverging from the intended goal.

# Solution
Providing user with a chain id flag to determine their intended chain.

phamminh0811:
## Overview

Close #2471 
This pr adding CLI Flag for chainID in txsim and make txsim runs with chainID option.

## Checklist


hoangdv2429:
we need to add prefix `feat:` for this @phamminh0811

rootulp (approved):
Code changes LGTM. Thanks for the contribution! Defer to @staheri14 and @cmwaters for another review + merge

staheri14 (commented):
Thanks for the contribution! I have left one comment.
Also, would like to defer it to other reviewers  for the final approval.

hoangdv2429:
I believe [here](https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#files) is the describe of go.work.sum and we also have an on going issue related to changes of go.work.sum in [here](https://github.com/golang/go/issues/51941). I think we can not track it for now. wdyt @staheri14 @phamminh0811

rootulp:
Based on https://github.com/celestiaorg/celestia-app/issues/2471#issuecomment-1862250164 it looks like we may not need this PR.

Issue #2939, opened 2023-12-15
Document how to get tx hash for a base64 encoded blob tx


rootulp:
## Context

- https://github.com/celestiaorg/celestia-app/issues/2806#issuecomment-1798776619
- https://github.com/celestiaorg/celestia-app/issues/2868

## Problem

A few issues have been opened because the CometBFT API endpoint `/cosmos/base/tendermint/v1beta1/blocks/{block_number}` (and potentially other endpoints) returns a list of base64 encoded transactions. Users perform the following steps:
1.  Iterate over the transactions
2. Base64 decode the transaction
3. Hash the tx bytes to determine the tx hash <- problem step
4. Query for a tx by tx hash

Step 3 is problematic because this works for all ordinary transactions but fails for blob transactions which are encoded as protobuf encoded `BlobTx`. In order to convert from the tx bytes to the tx hash needed for step 4, a user likely needs to [`UnmarshalBlobTx`](https://github.com/celestiaorg/celestia-app/blob/7e9fb84642dde8d51853e73efb817bf8067ffc7f/pkg/blob/blob.go#L58) and then hash the tx field inside the `BlobTx`.

## Proposal

1. Verify the proposed workaround above via unit test or Go playground
2. Add this workaround to docs or specs

Issue #2806, opened 2023-11-02
tx not found


hot-westeros:
### Summary of Bug

Hi, After testing, we found that there are many txs not found, such as the hash ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326. Our wallet program decoded and detected this transaction, but could not obtain the detailed information.

curl -X 'GET' 'https://api.celestia.pops.one/cosmos/tx/v1beta1/txs/ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326'
{
  "code": 5,
  "message": "tx not found: ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326",
  "details": [
  ]
}

What causes this transaction? Is it a special transaction? Can your team help to check it? Thanks!

### Version

v1.3.0

### Steps to Reproduce

1. Get the data of block height 408
curl https://api.celestia.pops.one/cosmos/base/tendermint/v1beta1/blocks/408
2. Use program decoding to obtain the transaction list
ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326 in block
3. Report an error after obtaining transaction details
curl -X 'GET' 'https://api.celestia.pops.one/cosmos/tx/v1beta1/txs/ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326'
{
   "code": 5,
   "message": "tx not found: ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326",
   "details": [
   ]
}

### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

rootulp:
Can you please elaborate on

> Use program decoding to obtain the transaction list

https://www.mintscan.io/celestia/block/408 contains 100 txs but I can't find the tx hash you pasted. For example https://www.mintscan.io/celestia/tx/ba68a0fbb70b062f1a3452dda918c75cddbdb2bdc8f7eb34ee91d42e8e505326?height=408 does not display a result.

hot-westeros:
Use program decoding to obtain the transaction list --- You can find a Cosmos SDK to parse out the transactions contained in block 408.

[...]

PR #2940, opened 2023-12-15
docs: add FAQ to blob specs

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2939

The blob specs are the x/blob/README.md file. While updating it, I noticed a few things that were outdated and decided to remove outdated fields (when incomplete) or update them. I spent more time on the Go Playground than explaining in words how to convert from Comet BFT API response to tx hash.


x/blob/README.md
```diff
@@ -71,38 +62,31 @@ details.
...
-- namespace: namespace is a byte slice of length 29 where the first byte is the
-  namespaceVersion and the subsequent 28 bytes are the namespaceId.
-- blob_sizes: sizes of each blob in bytes.
-- share_commitments is a list of share commitments (one per blob).
-- share_versions are the versions of the share format that the blobs associated
-  with this message should use when included in a block. The share_versions
-  specified must match the share_versions used to generate the share_commitment
-  in this message. See
-  [ADR007](../../docs/architecture/adr-007-universal-share-prefix.md) for more
-  details on how this effects the share encoding and when it is updated.
```
rootulp, line 101: I inlined this and added the missing fields to the source proto definition

x/blob/README.md
```diff
@@ -3,31 +3,22 @@
...
-    1. `Data         []byte`: the data to be published.
-    1. `ShareVersion uint32`: the version of the share format used to encode
-       this blob into a share.
-1. A single [`sdk.Tx`](https://github.com/celestiaorg/cosmos-sdk/blob/v1.15.0-sdk-v0.46.13/docs/architecture/adr-020-protobuf-transaction-encoding.md) which encapsulates a `MsgPayForBlobs` message that is composed of:
-    1. `Signer string`: the transaction signer
-    1. `NamespaceIds []byte`: the namespaces they wish to publish each blob to.
-       The namespaces here must match the namespaces in the `Blob`s.
-    1. `ShareCommitments [][]byte`: a share commitment that is the root of a Merkle
-       tree where the leaves are share commitments to each blob associated with
-       this `BlobTx`.
```
rootulp, line 20: Removed these fields because they're incomplete

x/blob/README.md
```diff
@@ -3,31 +3,22 @@
...
 
 The `x/blob` module enables users to pay for arbitrary data to be published to
-the Celestia blockchain. Users create a single `BlobTx` that is composed of:
-
-1. Multiple `Blob`s (Binary Large OBjects): the data they wish to publish. A
-   single `Blob` is composed of:
-    1. `NamespaceId  []byte`: the namespace this blob should be published to.
-    1. `Data         []byte`: the data to be published.
-    1. `ShareVersion uint32`: the version of the share format used to encode
-       this blob into a share.
```
rootulp, line 13: Removed these fields b/c they're incomplete

x/blob/README.md
```diff
@@ -210,3 +194,16 @@ The steps in the
...
+Q: Why do the responses from Comet BFT API endpoints fail to decode to valid transaction hashes?
+
+The response of CometBFT API endpoints (e.g. `/cosmos/base/tendermint/v1beta1/blocks/{block_number}`) will contain a field called `txs` with base64 encoded transactions. On most Cosmos based chains, one can base64 decode the transactions in the response and unmarshal the result into a `sdk.Tx` type. However, on Celestia, the `BlobTx` type is a wrapper around the Cosmos SDK transaction. For this reason, the transactions in the response can't directly be decoded and used. In order to use the response, one must unpack the transaction inside any `BlobTx`s. In other words:
+
+1. Base64 decode the transaction
+1. Check to see if the transaction is a `BlobTx` by unmarshalling it into a `BlobTx` type.
+   1. If it is a `BlobTx`, then unmarshal the `BlobTx`'s `Tx` field into a `sdk.Tx` type.
+   1. If it is not a `BlobTx`, then unmarshal the transaction into a `sdk.Tx` type.
+
+See this [Go Playground](https://go.dev/play/p/rdB1uVlQRml) for a code exmaple.
```
rootulp, line 209: I just realized that it's probably not best practice to link to a Go playground from our specs. I could replace the link with a new test in this repo if reviewers prefer that.

rootulp: https://github.com/celestiaorg/celestia-app/pull/2940/commits/78d249dfc2f7ade142e9e92cebd2487f61cbab47

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error decoding transaction: %v", err)
+	}
+	return tx, nil
+}
+
+func sum(bz []byte) []byte {
+	h := sha256.Sum256(bz)
+	return h[:]
+}
```
cmwaters, line 111: nit: `hash` is probably a better name for this function

x/blob/test/testdata/block_response.json
```diff
@@ -0,0 +1,1852 @@
+{
```
cmwaters, line 1: Why did you pick a block with so many transactions?

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+	Block Block `json:"block"`
+}
+
+type Block struct {
+	Data BlockData `json:"data"`
+}
+
+type BlockData struct {
+	Txs []string `json:"txs"`
+}
```
cmwaters, line 31: Why not use the types provided in cometbft?

x/blob/README.md
```diff
@@ -210,3 +194,16 @@ The steps in the
 function can be reverse engineered to submit blobs programmatically.
 
 <!-- markdownlint-enable MD010 -->
+
+## FAQ
+
+Q: Why do the responses from Comet BFT API endpoints fail to decode to valid transaction hashes?
```
cmwaters, line 200: Can you mention that this is just for PFB transactions

x/blob/test/testdata/block_response.json
rootulp: Mostly because this was the one referenced in https://github.com/celestiaorg/celestia-app/issues/2806

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+		if i == 273 {
+			// https://celenium.io/tx/C55BDD3DF3348A9F8D9206528051804754F009A1B9D0F69CCC2F9D4334215D21
+			wantHash := "C55BDD3DF3348A9F8D9206528051804754F009A1B9D0F69CCC2F9D4334215D21"
+			gotHash := strings.ToUpper(hex.EncodeToString(sum(txBytes)))
+			assert.Equal(t, wantHash, gotHash)
+
+			wantSigner := "celestia18y3ydyn7uslhuxu4lcm2x83gkdhrrcyaqvg6gk"
+			gotSigner := tx.GetMsgs()[0].GetSigners()[0].String()
+			assert.Equal(t, gotSigner, wantSigner)
+		}
```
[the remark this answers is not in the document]
rootulp: Can do. I didn't because it didn't seem strictly necessary to assert on all the fields.

x/blob/test/decode_blob_tx_test.go
rootulp: Good point, can do.

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+type Block struct {
+	Data BlockData `json:"data"`
+}
+
+type BlockData struct {
+	Txs []string `json:"txs"`
+}
+
+// TestDecodeBlobTx demonstrates how one can take the response from the Comet
+// BFT API endpoint /cosmos/base/tendermint/v1beta1/blocks/{block_number} and
```
staheri14, line 22: [Question] Why isn't it named "Celestia-core API" instead of "CometBFT API"? This choice might be leading to confusion, as readers could associate it with Comet-BFT APIs (while they are very similar, yet they have their own differences like this one) .

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+			t.Errorf("error decoding base64 tx: %v", err)
+		}
+
+		txBytes := getTxBytes(decoded)
+		tx, err := decodeTx(txBytes)
+		if err != nil {
+			t.Errorf("error decoding tx: %v", err)
+		}
+
+		if i == 273 {
```
staheri14, line 38: [Question] Why 273? mind explaining where does it come from?

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+			assert.Equal(t, wantHash, gotHash)
+
+			wantSigner := "celestia18y3ydyn7uslhuxu4lcm2x83gkdhrrcyaqvg6gk"
+			gotSigner := tx.GetMsgs()[0].GetSigners()[0].String()
+			assert.Equal(t, gotSigner, wantSigner)
+		}
+	}
+}
+
+func getBlockResponse(t *testing.T) BlockResponse {
```
staheri14: A func description is helpful as the name of the function sounds like it retrieves live block data.

x/blob/test/decode_blob_tx_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+func getTxBytes(txBytes []byte) []byte {
+	bTx, isBlob := blob.UnmarshalBlobTx(txBytes)
+	if isBlob {
+		return bTx.Tx
+	}
+	return txBytes
+}
+
+func decodeTx(txBytes []byte) (types.Tx, error) {
```
staheri14, line 86: [Question] Why not putting these utility functions in a main package instead of a test package?

staheri14 (commented):
LGTM!

x/blob/README.md
```diff
@@ -3,31 +3,22 @@
...
-transaction i.e., `sdk.Tx` from the blob. Both components get included in the
-data square in different namespaces: the `sdk.Tx` of the original `BlobTx`
-together with some metadata about the separated blobs get included in the
-PayForBlobNamespace (one of the [reserved
-namespaces](../../specs/src/specs/namespace.md#reserved-namespaces)) and the
-associated blob gets included in the namespace the user specified in the
-original `BlobTx`. Further reading: [Data Square
-Layout](../../specs/src/specs/data_square_layout.md)
+the `sdk.Tx` from the blob(s). Both components get included in the
+data square in different namespaces:
```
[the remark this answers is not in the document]
rootulp: A link exists on line 21. I can inline though

x/blob/README.md
```diff
@@ -71,38 +62,31 @@ details.
...
   repeated uint32 blob_sizes = 3;
+  // share_commitments is a list of share commitments (one per blob).
   repeated bytes share_commitments = 4;
+  // share_versions are the versions of the share format that the blobs
+  // associated with this message should use when included in a block. The
+  // share_versions specified must match the share_versions used to generate the
+  // share_commitment in this message.
   repeated uint32 share_versions = 8;
 }
 ```
```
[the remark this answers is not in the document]
rootulp: This was how it was previously but my concern is then the docs may get stale with respect to the fields. If the docs are inlined with the code, I think they're less likely to get stale b/c in a PR that just changes one of these fields, the surrounding lines are visible in the PR diff.

x/blob/README.md
```diff
@@ -71,38 +62,31 @@ details.
...
 
-- [`MsgPayForBlobs`](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/proto/celestia/blob/v1/tx.proto#L16-L31)
-  pays for a set of blobs to be included in the block. Blob transactions that contain
-  this `sdk.Msg` are also referred to as "PFBs".
+`MsgPayForBlobs` pays for a set of blobs to be included in the block. Blob transactions that contain this `sdk.Msg` are also referred to as "PFBs".
 
 ```proto
+// MsgPayForBlobs pays for the inclusion of a blob in the block.
 message MsgPayForBlobs {
+  // signer is the bech32 encoded signer address
```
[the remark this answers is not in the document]
rootulp: 3411146071f36d2fa019e2964073972ea0979d63

x/blob/test/decode_blob_tx_test.go
rootulp: dd6e0d829e8b72b1c4776ef52a4d0d69dd023516

x/blob/test/decode_blob_tx_test.go
rootulp: c6dfa14c058a02aec9c6987de938c6bb58dc3780

x/blob/test/decode_blob_tx_test.go
rootulp: hmm because they're not used outside this test file

x/blob/test/decode_blob_tx_test.go
rootulp: 4942be3e963d0936043f31d5161ae6aceda8a94b

x/blob/test/decode_blob_tx_test.go
staheri14: Wouldn't they be useful?

staheri14 (dismissed):

x/blob/test/decode_blob_tx_test.go
rootulp: I guess they may be useful. After moving them to a non-test file, are you proposing also exporting them? If we don't export them then they won't be useful to external consumers. If we do export them, then we need to consider whether we should add to the exported public API.

I'm inclined to merge this as-is and create a follow-up issue if it's something we want to pursue.

rootulp:
Tagging @cmwaters for re-review.

x/blob/test/decode_blob_tx_test.go
staheri14: Yes, I meant also exporting them as they are likely to be used by the users if there is no such function available in the app.

staheri14: A follow up PR sounds good to me.

x/blob/README.md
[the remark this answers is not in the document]
staheri14: I understand your perspective. However, generally, when someone intends to update a file or specification to reflect a code change, it's anticipated that they will review the entire file (not only a portion of it) to ensure that all references, relevant sections (in this case the descriptions of protobuf message fields) are properly updated.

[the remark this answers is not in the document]
rootulp: IMO the benefit of inline links does not outweigh the maintenance burden of maintaining 
1. code comments above the protobuf definition in the source file and 
2. a separate format for the same docs in the x/blob/README.md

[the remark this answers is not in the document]
staheri14: I'm open to your perspective on this matter. It was just a suggestion from my side.

staheri14 (approved):

Issue #1158, opened 2023-01-02
Panic on submit-legacy-proposal


mojtaba-esk:
### Summary of Bug

Getting panic on trying to submit a legacy proposal.
here is the raw log on querying the tx:
```sh
recovered: runtime error: invalid memory address or nil pointer dereference\nstack:\ngoroutine
  138 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:24
  +0x65\ngithub.com/cosmos/cosmos-sdk/baseapp.newDefaultRecoveryMiddleware.func1({0x1c4f980,
  0x37447e0})\n\t/root/go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.4.0-sdk-v0.46.0/baseapp/recovery.go:71
  +0x27\ngithub.com/cosmos/cosmos-sdk/baseapp.newRecoveryMiddleware.func1({0x1c4f980?,
  0x37447e0?})\n\t/root/go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.4.0-sdk-v0.46.0/baseapp/recovery.go:39
  +0x30\ngithub.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x1c4f980, 0x37447e0},
  0xc00e4f1680?)\n\t/root/go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.4.0-sdk-v0.46.0/baseapp/recovery.go:28
  +0x37\ngithub.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x1c4f980, 0x37447e0},
  0x28e79c0?)\n\t/root/go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.4.0-sdk-v0.46.0/baseapp/recovery.go:33
  +0x5e\ngithub.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).runTx.func1()\n\t/root/go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.4.0-sdk-v0.46.0/baseapp/baseapp.go:605
[...]

PR #2944, opened 2023-12-17
chore: bump core to use more precise mempool metrics

evan-forbes:
## Overview

bumps core to have more precise mempool metrics. I'll be using this branch on testground, but leaving as draft for now


evan-forbes:
we don't need to use the backport label here as I manually backported in #1158 due to the guaranteed conflicts

PR #2955, opened 2023-12-20
refactor: removes redundant for loop

fahimahmedx:
## Overview
There are two for loops that iterate over the same values. This small PR combines them into one.

## Checklist

### DIFF
--- a/x/paramfilter/gov_handler.go
+++ b/x/paramfilter/gov_handler.go
@@ -56,9 +56,7 @@ func (pbl ParamBlockList) handleParameterChangeProposal(
 		if pbl.IsBlocked(c.Subspace, c.Key) {
 			return ErrBlockedParameter
 		}
-	}
 
-	for _, c := range p.Changes {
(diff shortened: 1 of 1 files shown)


cmwaters (commented):
I think we need the two for loops here because the rule is that if one parameter change is blocked then all other changes are void. 

That being said it might be fine because if we perform an `Update` and later realise that one of the parameters is blocked, it should simply fail the transaction and not apply it to state.

Generally there are going to only be a few parameters that get changed in a single proposal so I don't think the optimisation provides much gains

rootulp:
> That being said it might be fine because if we perform an Update and later realise that one of the parameters is blocked, it should simply fail the transaction and not apply it to state.

I think [this test](https://github.com/celestiaorg/celestia-app/blob/f20b0a96d639ec628bdf8de80c9107c15a9e6897/x/paramfilter/test/param_filter_test.go#L49-L55) covers that scenario but I agree, the optimization doesn't seem strictly necessary and IMO the code is a bit more readable with two for loops.

PR #2964, opened 2023-12-27
chore: update go work

cmwaters:
It seems my the go.work.sum is out of date and go keeps trying to modify it in my other PR. This is done by running `go mod tidy`


rootulp (approved):
I've seen this locally too. We can potentially explore a dependabot style auto go mod tidy workflow

PR #2970, opened 2023-12-28
refactor: Imporve signer confirm tx

htiennv:
This PR using a `Ticker` for periodically call `GetTx` by given hash instead of a `Timer`. This makes the code clean and avoids calling reset `Timer` multiple times.

It's similar to `WaitMined` in [geth](https://github.com/ethereum/go-ethereum/blob/09e0208029ff96a9cda0c69dbaebfd3f31a39771/accounts/abi/bind/util.go#L32).

Other changes:
* Add lock when get `pollTime`

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
@@ -16,6 +15,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/types/tx"
 	"github.com/cosmos/cosmos-sdk/types/tx/signing"
 	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
@@ -153,7 +153,7 @@ func (s *Signer) SubmitPayForBlob(ctx context.Context, blobs []*blob.Blob, opts 
 		return nil, err
 	}
 	if resp.Code != 0 {
-		return resp, fmt.Errorf("tx failed with code %d: %s", resp.Code, resp.RawLog)
+		return nil, fmt.Errorf("tx failed with code %d: %s", resp.Code, resp.RawLog)
 	}
 
 	return s.ConfirmTx(ctx, resp.TxHash)
@@ -206,36 +206,32 @@ func (s *Signer) BroadcastTx(ctx context.Context, txBytes []byte) (*sdktypes.TxR
 	return resp.TxResponse, nil
 }
 
-// ConfirmTx periodically pings the provided node for the commitment of a transaction by its
-// hash. It will continually loop until the context is cancelled, the tx is found or an error
-// is encountered.
+// ConfirmTx waits for tx to be confirmed on the blockchain.
+// It stops waiting when the context is canceled.
 func (s *Signer) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxResponse, error) {
 	txClient := tx.NewServiceClient(s.grpc)
-	timer := time.NewTimer(0)
-	defer timer.Stop()
+
+	poolTicker := time.NewTicker(s.pollTime)
+	defer poolTicker.Stop()
+
 	for {
+		resp, err := txClient.GetTx(ctx, &tx.GetTxRequest{Hash: txHash})
+		if err == nil {
+			if resp.TxResponse.Code != 0 {
+				return nil, fmt.Errorf("tx failed with code %d: %s", resp.TxResponse.Code, resp.TxResponse.RawLog)
+			}
+			return resp.TxResponse, nil
+		}
+
+		if !errors.Is(err, sdkerrors.ErrNotFound) {
+			return nil, err
+		}
+
+		// Wait for the next round.
 		select {
 		case <-ctx.Done():
-			return &sdktypes.TxResponse{}, ctx.Err()
-		case <-timer.C:
-			resp, err := txClient.GetTx(
-				ctx,
-				&tx.GetTxRequest{
-					Hash: txHash,
-				},
-			)
-			if err == nil {
-				if resp.TxResponse.Code != 0 {
-					return resp.TxResponse, fmt.Errorf("tx failed with code %d: %s", resp.TxResponse.Code, resp.TxResponse.RawLog)
-				}
-				return resp.TxResponse, nil
-			}
-
-			if !strings.Contains(err.Error(), "not found") {
-				return &sdktypes.TxResponse{}, err
-			}
-
-			timer.Reset(s.pollTime)
+			return nil, ctx.Err()
+		case <-poolTicker.C:
 		}
 	}
 }
@@ -257,8 +253,6 @@ func (s *Signer) Address() sdktypes.AccAddress {
 
 // SetPollTime sets how often the signer should poll for the confirmation of the transaction
 func (s *Signer) SetPollTime(pollTime time.Duration) {
-	s.mtx.Lock()
-	defer s.mtx.Unlock()
 	s.pollTime = pollTime
 }
 
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -15,7 +16,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/types/tx"
 	"github.com/cosmos/cosmos-sdk/types/tx/signing"
 	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
@@ -153,7 +153,6 @@ func (s *Signer) SubmitPayForBlob(ctx context.Context, blobs []*blob.Blob, opts 
 		return nil, err
 	}
 	if resp.Code != 0 {
-		// TODO(htienv): Should return nil here? 
 		return resp, fmt.Errorf("tx failed with code %d: %s", resp.Code, resp.RawLog)
 	}
 
@@ -223,19 +222,18 @@ func (s *Signer) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxResp
 		resp, err := txClient.GetTx(ctx, &tx.GetTxRequest{Hash: txHash})
 		if err == nil {
 			if resp.TxResponse.Code != 0 {
-				return nil, fmt.Errorf("tx failed with code %d: %s", resp.TxResponse.Code, resp.TxResponse.RawLog)
+				return resp.TxResponse, fmt.Errorf("tx failed with code %d: %s", resp.TxResponse.Code, resp.TxResponse.RawLog)
 			}
 			return resp.TxResponse, nil
 		}
-
-		if !errors.Is(err, sdkerrors.ErrNotFound) {
-			return nil, err
+		if !strings.Contains(err.Error(), "not found") {
+			return &sdktypes.TxResponse{}, err
 		}
 
 		// Wait for the next round.
 		select {
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			return &sdktypes.TxResponse{}, ctx.Err()
 		case <-poolTicker.C:
 		}
 	}
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -215,8 +215,8 @@ func (s *Signer) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxResp
 	pollTime := s.pollTime
 	s.mtx.RUnlock()
 
-	poolTicker := time.NewTicker(pollTime)
-	defer poolTicker.Stop()
+	pollTicker := time.NewTicker(pollTime)
+	defer pollTicker.Stop()
 
 	for {
 		resp, err := txClient.GetTx(ctx, &tx.GetTxRequest{Hash: txHash})
@@ -234,7 +234,7 @@ func (s *Signer) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxResp
 		select {
 		case <-ctx.Done():
 			return &sdktypes.TxResponse{}, ctx.Err()
-		case <-poolTicker.C:
+		case <-pollTicker.C:
 		}
 	}
 }
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -206,8 +206,9 @@ func (s *Signer) BroadcastTx(ctx context.Context, txBytes []byte) (*sdktypes.TxR
 	return resp.TxResponse, nil
 }
 
-// ConfirmTx waits for tx to be confirmed on the blockchain.
-// It stops waiting when the context is canceled.
+// ConfirmTx periodically pings the provided node for the commitment of a transaction by its
+// hash. It will continually loop until the context is cancelled, the tx is found or an error
+// is encountered.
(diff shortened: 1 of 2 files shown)


pkg/user/signer.go
```diff
@@ -257,8 +253,6 @@ func (s *Signer) Address() sdktypes.AccAddress {
 
 // SetPollTime sets how often the signer should poll for the confirmation of the transaction
 func (s *Signer) SetPollTime(pollTime time.Duration) {
-	s.mtx.Lock()
-	defer s.mtx.Unlock()
 	s.pollTime = pollTime
 }
```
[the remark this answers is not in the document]
htiennv: Done

evan-forbes (dismissed):
lgtm

pkg/user/signer.go
```diff
@@ -153,7 +153,7 @@ func (s *Signer) SubmitPayForBlob(ctx context.Context, blobs []*blob.Blob, opts
 		return nil, err
 	}
 	if resp.Code != 0 {
-		return resp, fmt.Errorf("tx failed with code %d: %s", resp.Code, resp.RawLog)
+		return nil, fmt.Errorf("tx failed with code %d: %s", resp.Code, resp.RawLog)
```
evan-forbes, line 156: blocking

wait this change seems unrelated. why do we want to return nil and not the response here?

htiennv: Thanks @evan-forbes for your review!
I see here we are already returning an error with a code, so there is no need to return a response. Returning nil would be better.

htiennv: And I don't understand why the old code returns response here.

evan-forbes: your point is definitley valid, where ideally we would return only the tx response or nil and an error. Since the sdk includes so much information in the tx response, I think we need to keep returning it as well. We can't get rid of the error entirely either since there are instances where the sdk will return a nil tx response. Perhaps we can not perform the check on the code to keep things logically cleaner, however that check is performed in almost all cases.

regardless, that change is outside of the scope of the refactor imo, and changing the return value seems very close to a breaking change if it isn't one. Existing scripts that I've written would at least result in a panic should this get hit.

evan-forbes: we can see this actually break the `TestLegacyUpgrade` test

evan-forbes: I'd approve without this change, and we should open an issue to track the returning nil thing

htiennv: Ok, thanks! I will ignore this change.

htiennv: Done @evan-forbes

pkg/user/signer.go
```diff
@@ -153,6 +153,7 @@ func (s *Signer) SubmitPayForBlob(ctx context.Context, blobs []*blob.Blob, opts
 		return nil, err
 	}
 	if resp.Code != 0 {
+		// TODO(htienv): Should return nil here?
```
rootulp, line 156: good question. It seems odd to return a response and an error from this function.

htiennv: Yes, I think need reconsider for ensure it doesn't affect other things

rootulp: I'm not strongly opinionated on this but do you mind removing your handle from the TODO comment?

```suggestion
		// TODO: should this return nil? 
```

or removing the line all together and creating a GH issue?

rootulp (commented):
[question] IDK how this was missed in a previous PR but should https://github.com/htiennv/celestia-app/blob/563a154b4f76e1489bda929b64024677117723ff/pkg/user/signer_test.go#L72 contain a leading `Test` so that it actually runs in the test suite?

evan-forbes (commented):
it looks like we're still seeing quite a few tests fail, so we need to figure out why before we can merge anything. per the other review we should also clean up todo comments

pkg/user/signer.go
htiennv: Yes, thank you!

htiennv:
Oh sorry, I will fix this PR for handle test cases!

htiennv:
I have update this PR, please review it @evan-forbes @rootulp. Thanks!

pkg/user/signer.go
```diff
@@ -206,36 +206,35 @@ func (s *Signer) BroadcastTx(ctx context.Context, txBytes []byte) (*sdktypes.TxR
...
-				}
-				return resp.TxResponse, nil
-			}
-
-			if !strings.Contains(err.Error(), "not found") {
-				return &sdktypes.TxResponse{}, err
-			}
-
-			timer.Reset(s.pollTime)
+		case <-poolTicker.C:
```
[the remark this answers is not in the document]
htiennv: Thanks

rootulp:
Thanks for the ping @htiennv . I'm observing that `TestConfirmTxTimeout` fails so I pushed a commit that makes that test actually run (by adding the `Test` prefix).

htiennv:
Thanks @rootulp, I understand why this test fails, because currently we're checking 	
```go
require.Equal(s.T(), context.DeadlineExceeded, err)
```

But in this change I call `GetTx` before call `ctx.Done()`

so it reach error looks like: 
`rpc error: code = DeadlineExceeded desc = context deadline exceeded`

So `require.Equal(s.T(), context.DeadlineExceeded, err)` will fails

I think we can check as follows:
```go
require.Equal(s.T(), strings.Contains(err.Error(), context.DeadlineExceeded.Error()), true)
```
How do you feel about that?

rootulp:
Hmm, slight nit but also appears fixed with:

```go
	assert.Error(t, err)
	assert.Contains(t, err.Error(), context.DeadlineExceeded.Error())
```

pkg/user/signer.go
```diff
@@ -206,36 +206,35 @@ func (s *Signer) BroadcastTx(ctx context.Context, txBytes []byte) (*sdktypes.TxR
 	return resp.TxResponse, nil
 }
 
-// ConfirmTx periodically pings the provided node for the commitment of a transaction by its
-// hash. It will continually loop until the context is cancelled, the tx is found or an error
-// is encountered.
+// ConfirmTx waits for tx to be confirmed on the blockchain.
+// It stops waiting when the context is canceled.
```
rootulp, line 210: I think the previous godoc contains information that is still relevant. How about reverting the godoc change?

```go

```
```suggestion
// ConfirmTx periodically pings the provided node for the commitment of a transaction by its
// hash. It will continually loop until the context is cancelled, the tx is found or an error
// is encountered.
```

pkg/user/signer.go
```diff
@@ -206,36 +206,35 @@ func (s *Signer) BroadcastTx(ctx context.Context, txBytes []byte) (*sdktypes.TxR
...
 func (s *Signer) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxResponse, error) {
 	txClient := tx.NewServiceClient(s.grpc)
-	timer := time.NewTimer(0)
-	defer timer.Stop()
+
+	s.mtx.RLock()
+	pollTime := s.pollTime
+	s.mtx.RUnlock()
```
rootulp, line 216: 👍 seems help to add these read lock/unlocks given 

```go
// SetPollTime sets how often the signer should poll for the confirmation of the transaction
func (s *Signer) SetPollTime(pollTime time.Duration) {
	s.mtx.Lock()
	defer s.mtx.Unlock()
	s.pollTime = pollTime
}
```

pkg/user/signer.go
```diff
@@ -206,36 +206,35 @@ func (s *Signer) BroadcastTx(ctx context.Context, txBytes []byte) (*sdktypes.TxR
...
+
+	pollTicker := time.NewTicker(pollTime)
```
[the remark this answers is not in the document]
htiennv: Oh no, i have poll `GetTx` immediately. 
In fact, the difference is that I didn't check `ctx.Done()` immediately like the previous version.

pkg/user/signer.go
htiennv: yes, thank you

pkg/user/signer.go
htiennv: I see it was missed in the previous version.

pkg/user/signer.go
htiennv: Done

pkg/user/signer_test.go
```diff
@@ -69,12 +69,15 @@ func (s *SignerTestSuite) TestSubmitTx() {
...
+	t := s.T()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	_, err := s.signer.ConfirmTx(ctx, string("E32BD15CAF57AF15D17B0D63CF4E63A9835DD1CEBB059C335C79586BC3013728"))
-	require.Error(s.T(), err)
-	require.Equal(s.T(), err, context.DeadlineExceeded)
+	_, err := s.signer.ConfirmTx(ctx, "E32BD15CAF57AF15D17B0D63CF4E63A9835DD1CEBB059C335C79586BC3013728")
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), context.DeadlineExceeded.Error())
 }
```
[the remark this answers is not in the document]
htiennv: I think it's good. I will add them

evan-forbes (approved):

pkg/user/signer.go
rootulp: thanks for adding! Resolving this.

htiennv:
@rootulp I added some tests for ConfirmTx as you suggest. But I think case 3 (What happens if the tx if found after the second poll) is not necessary and can be ignored.

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
 	require.EqualValues(t, 0, resp.Code)
 }
 
-func (s *SignerTestSuite) ConfirmTxTimeout() {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+// TestConfirmTx verifies that the ConfirmTx method returns the expected results.
+func (s *SignerTestSuite) TestConfirmTx() {
+	t := s.T()
+
+	// 1. ConfirmTx returns an error when the context time out
```
rootulp, line 78:
```suggestion
	// 1. ConfirmTx returns an error when the context times out
```

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
...
+	// 2. ConfirmTx returns an error when tx is not found
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
-	_, err := s.signer.ConfirmTx(ctx, string("E32BD15CAF57AF15D17B0D63CF4E63A9835DD1CEBB059C335C79586BC3013728"))
-	require.Error(s.T(), err)
-	require.Equal(s.T(), err, context.DeadlineExceeded)
+	_, err = s.signer.ConfirmTx(ctx, "not found tx")
+	assert.Error(t, err)
+
+	// 3. ConfirmTx has no error if tx is found immediately
```
rootulp, line 91:
```suggestion
	// 3. ConfirmTx returns no error if tx is found immediately
```

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
...
+	fee := user.SetFee(1e6)
+	gas := user.SetGasLimit(1e6)
+	msg := bank.NewMsgSend(s.signer.Address(), testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 10)))
+	resp, err := s.submitTxWithoutConfirm([]sdk.Msg{msg}, fee, gas)
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
+	resp, err = s.signer.ConfirmTx(s.ctx.GoContext(), resp.TxHash)
+	assert.NoError(t, err)
+	assert.Equal(t, resp.Code, uint32(0))
```
rootulp, line 100: [nit] the order of arguments to assert.Equal is important for debugging. It should be expected then actual so that if this test fails, the logs help debug.

```suggestion
	assert.Equal(t, abci.CodeTypeOK, resp.Code)
```

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
...
+	// 4. ConfirmTx returns error if the tx is found with a non-zero error code
+	balance := s.queryCurrentBalance(t)
+	// Create a msg send with out of balance, ensure this tx is failed
+	msg = bank.NewMsgSend(s.signer.Address(), testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 1+balance)))
+	resp, err = s.submitTxWithoutConfirm([]sdk.Msg{msg}, fee, gas)
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
+	resp, err = s.signer.ConfirmTx(s.ctx.GoContext(), resp.TxHash)
+	assert.Error(t, err)
+	assert.NotEqual(t, resp.Code, uint32(0))
```
rootulp, line 111:
```suggestion
	assert.Equal(t, abci.CodeTypeOK, resp.Code)
```

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
...
+	resp, err := s.submitTxWithoutConfirm([]sdk.Msg{msg}, fee, gas)
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
+	resp, err = s.signer.ConfirmTx(s.ctx.GoContext(), resp.TxHash)
+	assert.NoError(t, err)
+	assert.Equal(t, resp.Code, uint32(0))
+
+	// 4. ConfirmTx returns error if the tx is found with a non-zero error code
+	balance := s.queryCurrentBalance(t)
+	// Create a msg send with out of balance, ensure this tx is failed
```
rootulp, line 104:
```suggestion
	// Create a msg send with out of balance, ensure this tx fails
```

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
 	require.EqualValues(t, 0, resp.Code)
 }
 
-func (s *SignerTestSuite) ConfirmTxTimeout() {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+// TestConfirmTx verifies that the ConfirmTx method returns the expected results.
```
rootulp, line 74: In my opinion, "the expected results" is vague. If this comment doesn't provide value, let's remove it entirely.

```suggestion
```

pkg/user/signer_test.go
```diff
@@ -69,12 +71,44 @@ func (s *SignerTestSuite) TestSubmitTx() {
...
+func (s *SignerTestSuite) TestConfirmTx() {
```
rootulp, line 75: [blocking] Can the test cases in this test be refactored to use table driven tests?

Ref: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

rootulp (commented):
Thanks for adding the test cases! I have a few comments on them

pkg/user/signer_test.go
htiennv: Thanks

pkg/user/signer_test.go
htiennv: Thanks for suggest! But I think this seems difficult to do, because each case requires different steps and inputs. For example, some cases need to call submit tx with different msg, some cases do not need to.

What do you think about dividing it into different function: TestConfirmTx_Timeout, TestConfirmTx_NotFoundTx,....?

pkg/user/signer_test.go
htiennv: `assert.NotEqual(t, abci.CodeTypeOK, resp.Code)`

pkg/user/signer_test.go
rootulp: Hmm if the setup is distinct for each test case, we could use [this pattern](https://github.com/celestiaorg/celestia-app/blob/ebd733d1db2e75d849a637f68587534447ceccd0/pkg/square/square_test.go#L38-L53) with multiple `t.Run("test case for...")`

htiennv: oh thanks, I didn't know about this...

htiennv: @rootulp I just push a new commit. Please review it, thank you!

rootulp: Awesome, looks great!

rootulp (approved):
Great work! Thanks for the PR

htiennv:
Thanks!

evan-forbes (approved):

staheri14 (approved):
Nice!

PR #2971, opened 2023-12-31
refactor: remove ValidateBasic in cli

lukema95:
## Overview
The `ValidateBasic` function has been called in the `GenerateOrBroadcastTxWithFactory` function, so we no longer need to add the `ValidateBasic` function when writing Cli. Related PRs can be viewed at: https://github.com/cosmos/cosmos-sdk/pull/9236#discussion_r623803504

## Checklist

### DIFF
--- a/x/blobstream/client/tx.go
+++ b/x/blobstream/client/tx.go
@@ -46,9 +46,6 @@ send a new message overriding the previous one.
 				ValidatorAddress: valAddress,
 				EvmAddress:       evmAddress,
 			}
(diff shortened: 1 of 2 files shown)


evan-forbes (commented):
I think this is okay, but we should dig further to be abundantly cautious that we aren't missing something and failing later than needed

good find tho @lukema95

cmwaters (approved):
This makes sense as being redundant.

rootulp (approved):

PR #2975, opened 2024-01-02
chore: remove unnecessary conversions

estensen:
Use linter to detect unnecessary type conversions
https://github.com/mdempsky/unconvert


cmwaters (approved):
This change looks good to me. Thanks for contributing

rootulp (approved):

PR #2982, opened 2024-01-04
chore: unslice slice expressions

estensen:
Found with [gocritic](https://github.com/go-critic/go-critic) ran with `golangci-lint run -E gocritic`

### DIFF
--- a/pkg/square/square.go
+++ b/pkg/square/square.go
@@ -262,7 +262,7 @@ func WriteSquare(
 	}
 
 	square := make([]shares.Share, totalShares)
-	copy(square[:], txShares)
+	copy(square, txShares)
 	copy(square[pfbStartIndex:], pfbShares)
 	if blobWriter.Count() > 0 {
 		copy(square[paddingStartIndex:], padding)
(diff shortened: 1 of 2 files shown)


estensen:
@rootulp I'm more than happy to add `gocritic` and fix all the linters you listed. I just thought it was easier to review multiple tiny PRs. 

Also, it's up to you if you agree with all the issues the linter found.

rootulp:
Yea I generally agree, smaller PRs are easier to review. However, in this case there's nothing stopping a subsequent PR from reverting the modifications you make in this PR. IMO if we're going to enable the gocritic linter (which I'm not strongly opinionated on for/against) then I'd rather we enable it in the config so that it runs in CI and subsequent PRs can't introduce any regressions. Does that make sense?

Before we proceed with this line of work, it may be worthwhile to create a GH issue with a short description on gocritic so that we can get alignment from the team on whether it's something we should enable. Our team is generally in favor of linters that reduce the # of nits on PRs and enforce more codebase consistency so gocritic seems like something we would want.

estensen:
Got it! Created https://github.com/celestiaorg/celestia-app/issues/2987

rootulp:
Converted this to draft b/c if possible, I think it's preferable to fix all the go-critic violations in the same PR that enables the linter in golangci.yml.

rootulp:
Closing because stale. We have another PR that does something similar: https://github.com/celestiaorg/celestia-app/pull/3000 and enables the linter in golangci.yml so thanks for the idea and creating the issue!

Issue #1621, opened 2023-04-12
Consider a global minimum gas price


cmwaters:
### Summary

Add a new consensus validity rule which sets a lower bound to the gas price of any transaction (that can also be assumed as the fee/gas). 

### Problem Definition

Currently we set an arbitrary 0.002 utia as the default minimum gas price in each nodes `app.toml`. This is a local parameter which prevents a transaction from being accepted in a node's mempool if it has a gas price below that value. Block producers can alter this value. Including setting it to 0. Block rewards for minting a new block + all transaction fees are proportionally distributed to all validators not to the block producer. Therefore there is an incentive for block producers to sell their blockspace through secondary markets. This undercuts the rest of the validators who provide similar compute, bandwidth and storage. Thus each validators strategy would swing to using secondary markets to maximize payoff. Having the transfer of value being conducted in off-chain auctions isn't necessarily that bad, however it likely means an increasingly centralised block production system; the potential exclusion of delegators who might not be privy to the auctions; and less general demand for the TIA token as the auction may be conducted other denominations. 

### Proposal

If we want the blockspace market to remain within Celestia, we should explore economic mechanisms such as setting a global min gas price (possibly controlled through governance) which better distributes value to those that provide it. 

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

evan-forbes:
noting that this would require the validation of all state transitions during `Prepare/ProcessProposal`, since we cannot have an invalid PFB be included in the block, and this is adding a block validity rule where there can no longer be PFBs with a fee of zero

 ref #49 since that would also require some sort of adjustable enforced min fee

liamsi:
Quoting @musalbas here: 
> Due to deferred execution, making it a consensus rule may not be trivial, and then we might as well implement fee burning.


That is a good point. Gas only gets accounted for in the next block. It is probably sufficient to enforce this on the mempool level only.

cmwaters:
You don't necessarily need to know how much gas a transaction will consume to implement this. In process proposal you simply check that the fee and gas limit equate to a minimum agreed upon gas price. The tx may be invalid and not execute but we allow for that already. This would simply assert: given the tx is valid, the fee payer will pay for at least this price per unit of gas

rootulp:
[...]

PR #2985, opened 2024-01-05
feat!: use harcoded global min gas price check in ProcessProposal

ninabarbakadze:
## Overview

Fixes - #1621 

Introducing a global minimum gas price, a consensus change enforcing a lower bound for all transactions.

_sidenote_ - updated the fee markets section in the docs but since I haven't yet been able to thoroughly read the specs section please do point out if I missed something.

## Checklist


app/ante/min_fee_test.go
```diff
@@ -0,0 +1,86 @@
...
+			expErr:     false,
+		},
+		{
+			name:       "good tx; fee equal to required minimum",
+			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, feeAmount)),
+			appVersion: uint64(2),
+			expErr:     false,
+		},
+		{
+			name:       "good tx; with no fee",
```
cmwaters, line 61: nit: this should probably mention the v1 app version

app/test/max_total_blob_size_test.go
```diff
@@ -78,7 +78,7 @@ func (s *MaxTotalBlobSizeSuite) TestErrTotalBlobSizeTooLarge() {
 
 	for _, tc := range testCases {
 		s.Run(tc.name, func() {
-			blobTx, err := signer.CreatePayForBlob([]*blob.Blob{tc.blob}, user.SetGasLimit(1e9))
+			blobTx, err := signer.CreatePayForBlob([]*blob.Blob{tc.blob}, user.SetGasLimit(1e9), user.SetFee(2000000))
```
cmwaters, line 81: nit: Since the minimum gas price is hardcoded, you could probably create a function that takes the gas limit and returns the minimum fee instead of having to calculate it yourself

specs/src/specs/resource_pricing.md
```diff
@@ -168,12 +168,14 @@ entire fee is then deducted from the transaction no matter what.
...
-their `app.toml`. Transactions that do not exceed that given gas price will not
-be able to enter that node's mempool, and thus they will also not be
-gossiped by that node.
+By default, Celestia's consensus nodes prioritize transactions in their mempools
+based on fees. Previously, there was no minimum fee enforced, allowing each 
+consensus node to independently set its own minimum gas fee in `app.toml`, even 
+allowing for a fee of 0, creating the possibility of secondary markets. Now, 
+Celestia introduces a global minimum gas price, a consensus constant, unaffected
+by individual node configurations. Although nodes retain the freedom to increase
+gas prices locally, transactions that do not meet the global minimum threshold
```
cmwaters, line 177: more importantly a transaction below the minimum will lead to the entire block being rejected as invalid

cmwaters (commented):
Nice work! Almost done

specs/src/specs/resource_pricing.md
ninabarbakadze: updated

app/ante/min_fee_test.go
ninabarbakadze: added the app version

evan-forbes (commented):
had one blocking comment, but otherwise LGTM!! great work and debugging :+1: 

its an optimization and therefore optional, and can be done in a different PR, but if we wanted to fail even faster we _could_ account for the new min fee in the [user package](https://github.com/celestiaorg/celestia-app/blob/03aa13a6cee41536afb32d7b7dc9d0331be42985/pkg/user/signer.go#L124-L141).

app/test/max_total_blob_size_test.go
ninabarbakadze: do you suggest I put that helper in a separate package or do you have a specific spot in mind?

app/ante/fee_checker.go
```diff
@@ -11,41 +15,28 @@ const (
...
-			requiredFees := make(sdk.Coins, len(minGasPrices))
-
-			// Determine the required fees by multiplying each required minimum gas
-			// price by the gas limit, where fee = ceil(minGasPrice * gasLimit).
-			glDec := sdk.NewDec(int64(gas))
-			for i, gp := range minGasPrices {
-				fee := gp.Amount.Mul(glDec)
-				requiredFees[i] = sdk.NewCoin(gp.Denom, fee.Ceil().RoundInt())
-			}
+	minFee := sdk.NewInt(int64(math.Ceil(appconsts.GlobalMinGasPrice * float64(gas))))
```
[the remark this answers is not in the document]
ninabarbakadze: I'll try do some testing around this to sanity check

specs/src/specs/resource_pricing.md
```diff
@@ -168,12 +168,14 @@ entire fee is then deducted from the transaction no matter what.
...
 ## Fee market
 
-By default, Celestia consensus nodes will use mempools that prioritize fees,
-however mempool usage cannot be enforced at the protocol level. There is
-currently no enforced minimum fee, this value is set by each consensus node in
-their `app.toml`. Transactions that do not exceed that given gas price will not
-be able to enter that node's mempool, and thus they will also not be
-gossiped by that node.
+By default, Celestia's consensus nodes prioritize transactions in their mempools
+based on fees. Previously, there was no minimum fee enforced, allowing each
```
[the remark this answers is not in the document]
ninabarbakadze: done!

app/ante/min_fee_test.go
```diff
@@ -0,0 +1,86 @@
...
+			name:       "good tx; fee equal to required minimum",
+			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, feeAmount)),
+			appVersion: uint64(2),
+			expErr:     false,
+		},
+		{
+			name:       "good tx; with no fee (v1)",
+			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, 0)),
+			appVersion: uint64(1),
+			expErr:     false,
```
rootulp, line 64: [question] do we need to preserve the previous behavior where if appVersion v1 is used, the validator checks if the gas price in the tx exceeds the locally configured min gas price defined in `app.toml`?

I don't quite understand what behavior needs to be preserved from v1 to support single binary syncs but we should get some clarity around that and ideally document it. cc: @evan-forbes @cmwaters

specs/src/specs/resource_pricing.md
```diff
@@ -168,12 +168,14 @@ entire fee is then deducted from the transaction no matter what.
...
 ## Fee market
 
-By default, Celestia consensus nodes will use mempools that prioritize fees,
-however mempool usage cannot be enforced at the protocol level. There is
-currently no enforced minimum fee, this value is set by each consensus node in
-their `app.toml`. Transactions that do not exceed that given gas price will not
-be able to enter that node's mempool, and thus they will also not be
-gossiped by that node.
+By default, Celestia's consensus nodes prioritize transactions in their mempools
+based on fees. In v1, there was no minimum fee enforced, allowing each
```
rootulp, line 172: [not blocking] gas price seems more accurate. Tx A can have more fees than Tx B but whichever tx has a higher gas price will be prioritized higher.

```suggestion
By default, Celestia's consensus nodes prioritize transactions in their mempools
based on gas price. In v1, there was no minimum gas price enforced, allowing each 
```

specs/src/specs/resource_pricing.md
```diff
@@ -168,12 +168,14 @@ entire fee is then deducted from the transaction no matter what.
...
 
-By default, Celestia consensus nodes will use mempools that prioritize fees,
-however mempool usage cannot be enforced at the protocol level. There is
-currently no enforced minimum fee, this value is set by each consensus node in
-their `app.toml`. Transactions that do not exceed that given gas price will not
-be able to enter that node's mempool, and thus they will also not be
-gossiped by that node.
+By default, Celestia's consensus nodes prioritize transactions in their mempools
+based on fees. In v1, there was no minimum fee enforced, allowing each 
+consensus node to independently set its own minimum gas fee in `app.toml`, even
```
rootulp, line 173:
```suggestion
consensus node to independently set its own minimum gas price in `app.toml`, even 
```

specs/src/specs/resource_pricing.md
```diff
@@ -168,12 +168,14 @@ entire fee is then deducted from the transaction no matter what.
...
-By default, Celestia consensus nodes will use mempools that prioritize fees,
-however mempool usage cannot be enforced at the protocol level. There is
-currently no enforced minimum fee, this value is set by each consensus node in
-their `app.toml`. Transactions that do not exceed that given gas price will not
-be able to enter that node's mempool, and thus they will also not be
-gossiped by that node.
+By default, Celestia's consensus nodes prioritize transactions in their mempools
+based on fees. In v1, there was no minimum fee enforced, allowing each 
+consensus node to independently set its own minimum gas fee in `app.toml`, even 
+allowing for a fee of 0, creating the possibility of secondary markets. In v2,
```
rootulp, line 174:
```suggestion
allowing for a gas price of 0, creating the possibility of secondary markets. In v2, 
```

specs/src/specs/resource_pricing.md
```diff
@@ -168,12 +168,14 @@ entire fee is then deducted from the transaction no matter what.
...
-be able to enter that node's mempool, and thus they will also not be
-gossiped by that node.
+By default, Celestia's consensus nodes prioritize transactions in their mempools
+based on fees. In v1, there was no minimum fee enforced, allowing each 
+consensus node to independently set its own minimum gas fee in `app.toml`, even 
+allowing for a fee of 0, creating the possibility of secondary markets. In v2, 
+Celestia introduces a global minimum gas price, a consensus constant, unaffected
+by individual node configurations. Although nodes retain the freedom to increase
+gas prices locally, transactions that do not meet the global minimum threshold
+will lead to the entire block being rejected as invalid.
```
rootulp, line 178: [nit] 

```suggestion
by individual node configurations. Although nodes retain the freedom to increase
gas prices locally, all transactions in a block must be greater than or equal to the 
global minimum threshold. If a block is proposed that contains a tx with a gas price
below the global min gas price, the block will be rejected rejected as invalid.
```

rootulp (commented):
Overall LGTM! Great work. I have a few nits about wording in the specs that are optional and one question about whether we need to preserve the v1 behavior to support single binary syncs.

For context: https://github.com/celestiaorg/celestia-app/issues/1568

app/ante/min_fee_test.go
rootulp: Actually on second thought I don't think we need to preserve the old behavior because when a validator is processing a block, they don't have access to the min gas price in the app.toml of the block proposer.

ninabarbakadze: resolving this since we want to preserve all pre-upgrade consensus behaviour

app/ante/fee_checker.go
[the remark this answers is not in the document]
cmwaters: Yeah that's a good point, I think we should be adding overflow checks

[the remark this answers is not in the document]
cmwaters: Are there some linters / static analysis tools out there that would do it for us?

app/ante/min_fee_test.go
```diff
@@ -0,0 +1,86 @@
+package ante_test
```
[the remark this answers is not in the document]
ninabarbakadze: `fee_checker_test.go` already exists. I had to split up the test suit because of circular dependency issues since `getTxPriority` is not exported  and has to be part of the same package as `fee_checker.go`, nevertheless I think you're right this is confusing. I'll rename `fee_checker_test.go` `to tx_priority_test.go` and leave this file as is.

app/ante/fee_checker.go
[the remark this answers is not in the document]
ninabarbakadze: if we use max uint64 as a gas limit and global min fee becomes 1.1 there will be an overflow when we convert it to int64 since uint64 does not need to reserve any bits for negative numbers max uint64 is inherintly larger. @cmwaters @evan-forbes

I think this could be resolved by incorporating bigInt. we could also throw an error if a gas limit is larger than the max int64.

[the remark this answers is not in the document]
cmwaters: I'm not sure if big.Int would work unless you scale up the `MinimumGasPrice` to an int. Maybe copy the same `priorityScalingFactor` which is `1_000_000`.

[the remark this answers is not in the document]
cmwaters: We will need to check that gas <= math.MaxInt64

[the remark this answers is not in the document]
ninabarbakadze: but even then if gas is `math.MaxInt64-1` and globalMinGasPrice is > 1 we could still overflow.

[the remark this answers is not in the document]
ninabarbakadze: we could check that `minFee<math.MaxInt64-1`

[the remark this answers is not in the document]
cmwaters: I had something like this in mind
```
if gas > math.MaxInt64 {
    return err
}
minFee := sdk.NewInt(int64(globalMinGasPrice * priorityScalingFactor).MulRaw(int64(gas)).QuoRaw(priorityScalingFactor)
```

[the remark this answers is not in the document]
cmwaters: Check if the division rounds down. If so, we will need to change GTE to GT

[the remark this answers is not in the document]
evan-forbes:
```go
if gas > math.MaxInt64 {
    return err
}
```
I don't think it matters too too much that this a new block validity rule, however we should document that there is a new effective max on gas (and maybe a new if statement)

the most important thing though is that we are [risking non-determinism with a max int64 and a float64](https://stackoverflow.com/questions/74164159/how-to-keep-precision-for-big-numbers-in-golang-when-converting-from-float-to-bi#:~:text=All%20positive%20integers%20up%20to,is%20happening%20in%20your%20case.) 

I'm a `big.Fan` of just using `big.Int` in one way or another for everything, and we probably need to do that here if we want a bullet proof solution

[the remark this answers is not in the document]
cmwaters: > I'm a `big.Fan` of just using `big.Int`

looool I hope this was intentional

ninabarbakadze:
some of the nits will be addressed in a different pr when I continue working on the second part of this feature. https://github.com/celestiaorg/celestia-app/issues/3016

app/ante/fee_checker.go
```diff
@@ -11,41 +16,38 @@ const (
...
-	if ctx.IsCheckTx() {
-		minGasPrices := ctx.MinGasPrices()
-		if !minGasPrices.IsZero() {
-			requiredFees := make(sdk.Coins, len(minGasPrices))
+	// global minimum fee only applies to app versions greater than one
+	if ctx.BlockHeader().Version.App > v1.Version {
+		// check for overflow, gas is now capped at MaxInt64
+		if gas > math.MaxInt64 {
+			return nil, 0, errors.Wrap(sdkerror.ErrInvalidGasLimit, "overflow: gas exceeds MaxInt64")
+		}
```
cmwaters, line 35: It looks like we actually don't need this check.

We can go:

`gasInt  := sdk.NewIntFromUint64(gas)`

and then go `globalMinGasPrice.MulInt(gasInt).TruncateInt`

app/ante/fee_checker.go
```diff
@@ -11,41 +16,38 @@ const (
...
+		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.GlobalMinGasPrice))
+		if err != nil {
+			return nil, 0, errors.Wrap(err, "invalid GlobalMinGasPrice")
+		}
 
-			if !feeCoins.IsAnyGTE(requiredFees) {
-				return nil, 0, errors.Wrapf(sdkerror.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
-			}
+		minFee := globalMinGasPrice.MulInt64(int64(gas)).TruncateInt()
+		if !fee.GTE(minFee) {
```
cmwaters, line 44: Because we are truncating this will need to be `!fee.GT(minFee)`

ninabarbakadze: I think it's unintuitive to replace gte with gt since we should allow fee to be equal to. minFee. I think the only time we'd need gt is if minFee truncated to 0. but I added a rule for that edge case.

app/ante/fee_checker.go
```diff
@@ -11,41 +15,40 @@ const (
...
-	// if this is a CheckTx. This is only for local mempool purposes, and thus
-	// is only ran on check tx.
-	if ctx.IsCheckTx() {
-		minGasPrices := ctx.MinGasPrices()
-		if !minGasPrices.IsZero() {
-			requiredFees := make(sdk.Coins, len(minGasPrices))
+	// global minimum fee only applies to app versions greater than one
+	if ctx.BlockHeader().Version.App > v1.Version {
+		// convert the global minimum gas price to a big.Int
+		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.GlobalMinGasPrice))
```
cmwaters, line 32: note: not sure if the sdk have planned to deprecate sdk.Dec in the future (ref: https://github.com/cosmos/cosmos-sdk/issues/11783)

app/ante/fee_checker.go
```diff
@@ -11,41 +15,40 @@ const (
...
+		if err != nil {
+			return nil, 0, errors.Wrap(err, "invalid GlobalMinGasPrice")
+		}
+
+		gasInt := sdk.NewIntFromUint64(gas)
+		minFee := globalMinGasPrice.MulInt(gasInt).TruncateInt()
```
cmwaters, line 38: I think we should be using `RountInt` instead of `TruncateInt`.

Imagine the min gas price is 0.16 and the gas is 10, thus the fee is 1.6. We would want the user to have to pay 2 or more. If we truncate then the user only needs to pay 1.

app/ante/fee_checker.go
```diff
@@ -11,41 +15,40 @@ const (
...
-			// price by the gas limit, where fee = ceil(minGasPrice * gasLimit).
-			glDec := sdk.NewDec(int64(gas))
-			for i, gp := range minGasPrices {
-				fee := gp.Amount.Mul(glDec)
-				requiredFees[i] = sdk.NewCoin(gp.Denom, fee.Ceil().RoundInt())
-			}
+		// if min fee truncates to zero, set it to one
+		if minFee.LT(sdk.NewInt(appconsts.MinFee)) {
+			minFee = sdk.NewInt(appconsts.MinFee)
+		}
```
cmwaters, line 43: I think it should be okay for a user to pay zero fees if the min gas price is 0

app/ante/fee_checker.go
```diff
@@ -11,41 +15,40 @@ const (
...
 
-			if !feeCoins.IsAnyGTE(requiredFees) {
-				return nil, 0, errors.Wrapf(sdkerror.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
-			}
+		if !fee.GTE(minFee) {
```
rootulp, line 45: [no change needed] we may want to compare / contrast this implementation with https://github.com/skip-mev/feemarket/blob/e6f642fbeac30f78ec7061e2f7b23a6371bf0e85/x/feemarket/ante/fee.go#L76-L103

app/ante/min_fee_test.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		{
+			name:       "bad tx; gas limit and fee are 0",
+			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, 0)),
+			gasLimit:   0,
+			appVersion: uint64(2),
+			expErr:     true,
+		},
+		{
+			name:       "good tx; rounds down to 0 gets reset to 1",
```
rootulp, line 84: [nit] it's not immediately obvious what gets rounded down to 0 and reset to 1.

Perhaps a better test name would be "expect no error when minFee = 0.8"

Related, should there be another test case like:

```go
		{
			name:       "expect error when tx fee is 0",
			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, 0)),
			gasLimit:   400,
			appVersion: uint64(2),
			expErr:     false,
		},
```

ninabarbakadze: we're allowing tx fee of 0

app/ante/fee_checker.go
ninabarbakadze: removed

app/ante/fee_checker.go
cmwaters: Apparently not so don't worry about this comment

cmwaters (approved):
thumbs up from me 👍

app/ante/ante.go
```diff
@@ -39,7 +39,7 @@ func NewAnteHandler(
 		ante.NewConsumeGasForTxSizeDecorator(accountKeeper),
 		// Ensure the feepayer (fee granter or first signer) has enough funds to pay for the tx.
 		// Side effect: deducts fees from the fee payer. Sets the tx priority in context.
-		ante.NewDeductFeeDecorator(accountKeeper, bankKeeper, feegrantKeeper, checkTxFeeWithValidatorMinGasPrices),
+		ante.NewDeductFeeDecorator(accountKeeper, bankKeeper, feegrantKeeper, CheckTxFeeWithGlobalMinGasPrices),
```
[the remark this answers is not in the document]
cmwaters: I was thinking about having a generic version wrapper that wraps the decorator and only executes it for certain versions. This might be more cleaner than having version checks sporadically added to every decorator

app/ante/fee_checker.go
cmwaters: I think the main thing we should be wary of is this implementation assumes only a single token whereas skip's implementation allows for multiple native tokens. At the moment we only have one token and I don't see that changing in the foreseeable future but maybe we want to avoid the footgun in the future

pkg/appconsts/initial_consts.go
```diff
@@ -22,6 +22,13 @@ const (
 	// a nodes `CheckTx` and thus not be proposed by that node.
 	DefaultMinGasPrice = 0.002
 
+	// GlobalMinGasPrice is used in the processProposal to ensure
+	// that all transactions have a gas price greater than or equal to this value.
+	GlobalMinGasPrice = DefaultMinGasPrice
```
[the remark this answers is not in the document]
cmwaters: This should be viewed as a temporary home for now as we will want to either make it a governance parameter or have some other mechanism which picks this value. I don't think v2 will simply have this hardcoded value.

If we want to move it, we can shift it to the v2 directory. I'm easy either way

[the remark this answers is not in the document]
rootulp: Sounds good, I like `v2/app_consts.go`. 

I think I'd rather it _not_ be a governance modifiable param but I think we can discuss that as a follow up. Mostly b/c I don't think we should have any governance params long term and instead we should modify params via hard-forks after reaching rough consensus off-chain. So eventually we would move from governance params to versioned params.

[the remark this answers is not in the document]
rootulp: Since this PR already has two approvals, I don't think we need to block on just this constant move. We can move the constant in a follow up PR if that's easier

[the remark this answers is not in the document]
ninabarbakadze: I'll make this change in a follow-up pr @rootulp @cmwaters

Issue #2901, opened 2023-12-05
Consider kebab-casing `PayForBlobs` in the CLI


rootulp:
## Context

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/issues/2856#issuecomment-1840594264_
            

## Problem

All commands in the celestia-app CLI use kebab-casing. I.e. `celestia-appd tx ibc-transfer`.

## Proposal

Consider the breaking change of renaming `PayForBlobs` / `PayForBlob` to `pay-for-blob`

rootulp:
We could do this in a non-breaking way if we retain the PascalCase names as aliases. In other words, these would all be valid and for the same command:

```go
celestia-appd tx blob pay-for-blobs
celestia-appd tx blob pay-for-blob
celestia-appd tx blob PayForBlobs
celestia-appd tx blob PayForBlob
```

PR #2989, opened 2024-01-08
feat: kebab-case pay-for-blob command

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2901

## Testing

The previous commands still work b/c they are aliases:

```
$ ./build/celestia-appd tx blob pay-for-blob
Error: pay-for-blob requires two arguments: namespaceID and blob
$ ./build/celestia-appd tx blob pay-for-blobs
Error: pay-for-blob requires two arguments: namespaceID and blob
$ ./build/celestia-appd tx blob PayForBlob
Error: pay-for-blob requires two arguments: namespaceID and blob
$ ./build/celestia-appd tx blob PayForBlobs
Error: pay-for-blob requires two arguments: namespaceID and blob
```

### DIFF
--- a/x/blob/client/cli/payforblob.go
+++ b/x/blob/client/cli/payforblob.go
@@ -38,15 +38,15 @@ const (
 
 func CmdPayForBlob() *cobra.Command {
 	cmd := &cobra.Command{
-		Use: "PayForBlobs [namespaceID blob]",
+		Use: "pay-for-blob [namespaceID blob]",
 		// This example command can be run in a new terminal after running single-node.sh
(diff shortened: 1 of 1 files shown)


cmwaters (approved):
Perfect 👍

evan-forbes:
do we need the feat! for this?

ninabarbakadze:
> do we need the feat! for this?

I don't think so since @rootulp set up aliases it's no longer a breaking change

evan-forbes:
nice!

Issue #2990, opened 2024-01-08
`PayForBlob --help` indentation is off


rootulp:
## Problem

The help message for PayForBlob contains some unnecessary indentation and unescaped newlines.

```
$ celestia-appd tx blob PayForBlob --help
Pay for a data blob(s) to be published to Celestia.
User can use namespaceID and blob as argument for single blob submission
or use --input-file flag with the path to a json file for multiple blobs submission,
where the json file contains:

		{
			"Blobs": [
				{
					"namespaceID": "0x00010203040506070809",
					"blob": "0x48656c6c6f2c20576f726c6421"
				},
				{
					"namespaceID": "0x00010203040506070809",
					"blob": "0x48656c6c6f2c20576f726c6421"
				}
			]
		}

		namespaceID is the user-specifiable portion of a version 0 namespace. It must be a hex encoded string of 10 bytes.\n
		blob must be a hex encoded string of any length.\n
```

## Proposal

1. Remove the `\n`s in the help message
2. Fix the indentation
3. Reword anything that seems off

PR #2991, opened 2024-01-08
fix: CLI help for PayForBlob

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2990

## Testing

```shell
$ ./build/celestia-appd tx blob PayForBlob --help
Pay for data blob(s) to be published to Celestia.
To publish a single blob, specify the namespaceID and blob via CLI arguments.
To publish multiple blobs, use the --input-file flag with the path to a JSON file.
The JSON should look like:

{
	"Blobs": [
		{
			"namespaceID": "0x00010203040506070809",
			"blob": "0x48656c6c6f2c20576f726c6421"
		},
		{
			"namespaceID": "0x00010203040506070809",
			"blob": "0x48656c6c6f2c20576f726c6421"
		}
	]
}

The namespaceID is the user-specifiable portion of a version 0 namespace.
The namespaceID must be a hex encoded string of 10 bytes.
The blob must be a hex encoded string of non-zero length.

Usage:
  celestia-appd tx blob PayForBlobs [namespaceID blob] [flags]
```

### DIFF
--- a/x/blob/client/cli/payforblob.go
+++ b/x/blob/client/cli/payforblob.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -31,15 +30,18 @@ const (
 	// submitting a PayForBlob.
 	FlagNamespaceVersion = "namespace-version"
 
-	// FlagFileInput allows the user to provide file path to the json file
-	// for submitting multiple blobs.
+	// FlagFileInput allows the user to provide the path to a JSON file for
+	// submitting multiple blobs.
 	FlagFileInput = "input-file"
+
+	// FileInputExtension is the only file extension supported for
+	// FlagFileInput.
+	FileInputExtension = ".json"
 )
 
 func CmdPayForBlob() *cobra.Command {
 	cmd := &cobra.Command{
 		Use: "PayForBlobs [namespaceID blob]",
-		// This example command can be run in a new terminal after running single-node.sh
 		Example: "celestia-appd tx blob PayForBlobs 0x00010203040506070809 0x48656c6c6f2c20576f726c6421 \\\n" +
 			"\t--chain-id private \\\n" +
 			"\t--from validator \\\n" +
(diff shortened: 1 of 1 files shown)


cmwaters (approved):

evan-forbes (approved):

x/blob/client/cli/payforblob.go
```diff
@@ -52,27 +54,28 @@ func CmdPayForBlob() *cobra.Command {
...
+			"blob": "0x48656c6c6f2c20576f726c6421"
 		}
+	]
+}
 
-		namespaceID is the user-specifiable portion of a version 0 namespace. It must be a hex encoded string of 10 bytes.\n
-		blob must be a hex encoded string of any length.\n
+The namespaceID is the user-specifiable portion of a version 0 namespace.
+The namespaceID must be a hex encoded string of 10 bytes.
+The blob must be a hex encoded string of any length.
```
staheri14, line 78: [Question] Is zero-length blob an acceptable input? if not, then it is worth mentioning that blobs should have non-zero length.

x/blob/client/cli/payforblob.go
```diff
@@ -81,17 +84,16 @@ func CmdPayForBlob() *cobra.Command {
 				return err
 			}
 
-			// If there is a file path input we'll check for the file extension
 			if path != "" {
-				if filepath.Ext(path) != ".json" {
-					return fmt.Errorf("invalid file extension, require json got %s", filepath.Ext(path))
+				if filepath.Ext(path) != FileInputExtension {
```
staheri14, line 88: [Question] Is there a check in the code for whether the file exists in the specified path (hence showing proper error if not)?

staheri14 (dismissed):
LGTM!

x/blob/client/cli/payforblob.go
rootulp: Agreed, good call!

rootulp: 2c37b373a4152e2d67faeb8d0725dbdc8e1a4444

staheri14 (approved):

x/blob/client/cli/payforblob.go
rootulp: Looks like yes.
```
$ ./build/celestia-appd tx blob pay-for-blob --input-file foo.json --from validator
Error: open foo.json: no such file or directory
```

PR #2992, opened 2024-01-08
feat: expose sequence number correctly to users of the signer

cmwaters:
It's probably misleading that `GetSequence` should not only return the sequence number but increment it. The incrementing sequence should be internal and the user should just be able to use `Sequence` if they wish to query the sequence number of the signer

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -267,8 +267,24 @@ func (s *Signer) PubKey() cryptotypes.PubKey {
 	return s.pk
 }
 
+// Sequence returns the last signed sequence number of the signer
+func (s *Signer) Sequence() uint64 {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+	return s.lastSignedSequence
+}
+
 // GetSequence gets the latest signed sequence and increments the local sequence number
+// Deprecated: Use Sequence if you want to get the latest signed sequence number
 func (s *Signer) GetSequence() uint64 {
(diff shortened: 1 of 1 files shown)


evan-forbes (dismissed):
nice

rootulp (approved):

pkg/user/signer.go
```diff
@@ -267,14 +267,30 @@ func (s *Signer) PubKey() cryptotypes.PubKey {
...
+// Deprecated: Use Sequence if you want to get the latest signed sequence number
 func (s *Signer) GetSequence() uint64 {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 	defer func() { s.lastSignedSequence++ }()
 	return s.lastSignedSequence
 }
 
+// incrementSequence gets the latest signed sequence and increments the local sequence number
+func (s *Signer) incrementSequence() uint64 {
```
staheri14, line 287: [Suggestion/Optional] I suggest renaming the function to `GetAndIncrementLastSequence`. This name provides a clearer understanding of the function's behavior.

staheri14 (approved):

pkg/user/signer.go
cmwaters: done although I kept the method private

evan-forbes (approved):

rootulp (approved):

staheri14 (approved):

PR #3001, opened 2024-01-10
test: remove redundant test suite

ninabarbakadze:
## Overview

This file just starts running the test suit and shuts it down therefore we can remove it.

## Checklist


rootulp (commented):
Strange but I think the CLITestSuite is actually used by `x/blobstream/client/query_test.go`

ninabarbakadze:
> Strange but I think the CLITestSuite is actually used by `x/blobstream/client/query_test.go`

you're right! me and

evan-forbes:
yeah this was my bad, I just saw the file that wasn't doing anything and assumed it wasn't needed

PR #3002, opened 2024-01-10
chore: update core to revert breaking config change

evan-forbes:
## Overview

updates core to revert the breaking config change. we'll get the config change when we update comet to something other than that based on v0.34. backporting manually since this is guaranteed to have conflicts thanks to the change in go mod


staheri14 (approved):

rootulp (approved):

rootulp:
Do we want to backport to v1.x?

PR #3017, opened 2024-01-16
feat: add gas estimation to the signer

cmwaters:
The eventual goal is to have the signer be able to automatically estimate gas


pkg/user/signer.go
```diff
@@ -410,3 +404,17 @@ func QueryAccount(ctx context.Context, conn *grpc.ClientConn, encCfg encoding.Co
...
+func (s *Signer) getSignatureV2(sequence uint64, signature []byte) signing.SignatureV2 {
+	sigV2 := signing.SignatureV2{
+		Data: &signing.SingleSignatureData{
+			SignMode:  signing.SignMode_SIGN_MODE_DIRECT,
+			Signature: signature,
+		},
+		Sequence: sequence,
+	}
+	if sequence == 0 {
+		sigV2.PubKey = s.pk
```
cmwaters, line 417: It seems like the PubKey needs to be set for every message

evan-forbes (dismissed):

cmwaters:
Blocked on https://github.com/celestiaorg/cosmos-sdk/pull/371

ninabarbakadze (approved):

pkg/user/signer.go
staheri14: [Question] "needs" or "does not need"?

staheri14: Also, wondering whether this change aligns with the PR description or is an independent additional fix?

staheri14 (approved):

pkg/user/signer.go
cmwaters: > [Question] "needs" or "does not need"?

In the past it always needed the pubkey because of some bug in the SDK but I've since fixed it now so it no longer needs the pubKey

> Also, wondering whether this change aligns with the PR description or is an independent additional fix?

Yeah I could extract it out

staheri14: > Yeah I could extract it out

Good idea 👍

evan-forbes (dismissed):

rootulp (approved):

Wondertan:
Does estimation add a roundtrip?

staheri14 (approved):

evan-forbes (approved):

cmwaters:
> Does estimation add a roundtrip?

Yes it would

evan-forbes (approved):

pkg/user/signer_test.go
```diff
@@ -114,6 +114,18 @@ func (s *SignerTestSuite) TestConfirmTx() {
 	})
 }
 
+func (s *SignerTestSuite) TestGasEstimation() {
+	msg := bank.NewMsgSend(s.signer.Address(), testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 10)))
+	seq := s.signer.Sequence()
+	txBytes, err := s.signer.CreateTx([]sdk.Msg{msg})
+	require.NoError(s.T(), err)
+	// reset the sequence
+	s.signer.ForceSetSequence(seq)
```
rootulp, line 123: [question] why does the sequence need to be reset? Based on the `EstimateGas` function body, it seems like it should be possible to estimate the gas for a tx regardless of if the sequence is 0 or not.

```suggestion
```

[tangent] if it is important to test that estimate gas works when sequence == 0 and sequence != 0, perhaps two tests cases would make sense here.

rootulp (dismissed):

pkg/user/signer_test.go
cmwaters: Hmm maybe I could rewrite the `EstimateGas` function signature so instead of txBytes (which need to be generated using `CreateTx` and which automatically increment the sequence) to only take an array of `sdk.Msgs` and option types and create an unsigned transaction that way. What do you think?

cmwaters: > why does the sequence need to be reset?

Because `CreateTx` automatically increments the sequence

rootulp: Hmm that refactor idea seems okay to me but I don't follow why the refactor is needed. If I remove the line to `ForceSetSequence(seq)`, the test still passes.

cmwaters: > If I remove the line to ForceSetSequence(seq), the test still passes.

Really? It shouldn't. You should get a sequence mismatch for the following test in the suite

rootulp: oh I only ran that individual test not the remainder of the test suite. It seems like a code smell if other tests break based on this tests side-effects. Definitely not blocking on it but I'm open to your refactor idea if that means tests won't impact each other in this suite.

cmwaters: The reason for refactoring is more a matter of better dev UX. it's a bit of a foot gun if creating the tx you need to estimate the gas increments the sequence number

pkg/user/signer_test.go
```diff
@@ -114,6 +114,14 @@ func (s *SignerTestSuite) TestConfirmTx() {
 	})
 }
 
+func (s *SignerTestSuite) TestGasEstimation() {
+	msg := bank.NewMsgSend(s.signer.Address(), testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 10)))
+	// reset the sequence
```
rootulp, line 119: No longer needed

```suggestion
```

rootulp (dismissed):
LGTM

evan-forbes (approved):

rootulp (approved):

Issue #3018, opened 2024-01-17
govulncheck is failing


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/actions/runs/7547699875/job/20548198793?pr=3017#step:4:12

## Problem

govulncheck is failing in CI and locally:

```
$ make govulncheck
go: downloading golang.org/x/vuln v1.0.2
go: downloading golang.org/x/tools v0.17.0
govulncheck: loading modules: /opt/homebrew/bin/go list -m -json -mod=mod all: exit status 1
go: -mod may only be set to readonly when in workspace mode, but it is set to "mod"
	Remove the -mod flag to use the default readonly value,
	or set GOWORK=off to disable workspace mode.

exit status 1
make: *** [govulncheck] Error 1
```

## Proposal

1. Fix govulncheck so it can run locally
2. Ensure it passes in CI

rootulp:
I think this was caused by the recent 1.0.2 release: https://go.googlesource.com/vuln/+/refs/tags/v1.0.2 because this works with no issues if I downgrade to v1.0.1

Will do that and then investigate the issue more after. We likely want to pin to a particular version locally and in CI so that PRs don't randomly start failing.

PR #3019, opened 2024-01-17
chore(deps): pin to govulncheck v1.0.1

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3018

Will fix CI too because CI just calls `make govulncheck`. I kinda like that pattern b/c then we don't have to keep two versions in sync (one in Makefile and one in CI workflow file)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,5 +197,5 @@ prebuilt-binary:
 
 ## govulncheck: Check for vulnerabilities in dependencies.
 govulncheck:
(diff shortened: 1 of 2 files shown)


cmwaters (approved):
LGTM 

Is there anything else that calls latest that we want to pin. We should also probably have some way of manually updating these so we can use the latest release

ninabarbakadze (approved):

evan-forbes (approved):

PR #3027, opened 2024-01-18
chore: decrease voting period in single-node.sh

rootulp:
While working on https://github.com/celestiaorg/celestia-app/compare/main...rootulp:celestia-app:rp/gov-proposal-for-app-version I found it helpful to make these two modifications to the single node script. These are strictly optional so we don't _need_ to merge this.


cmwaters (approved):
I don't mind merging these changes

evan-forbes (approved):

Issue #2191, opened 2023-07-31
Stop returning the square size in PrepareProposal


evan-forbes:
In https://github.com/celestiaorg/celestia-core/pull/1051 we are removing the square size from block data and therefore no longer need the application to return it during PrepareProposal

rootulp:
https://github.com/celestiaorg/celestia-core/pull/1051 merged but it's on main and not on the release branch. Since it constitutes a breaking change, I guess we intentionally didn't backport it.

PR #3031, opened 2024-01-21
feat!: remove square size from prepare proposal response

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2191


rootulp:
Nvm can't do this without a breaking celestia-core release.

Issue #3016, opened 2024-01-16
Use nodes local min gas price for prepare proposal


cmwaters:
### Summary

The global min gas price is used as the minimum price for a transaction to enter a nodes mempool. The node still has a locally configured min price. This should be used when preparing the proposal such that all transaction included must be greater than that price

### Proposal

In `PrepareProposal`, while looping through all transactions, if one has a gas price lower than the nodes own threshold skip over it and do not include it in the list of transactions to be proposed

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

evan-forbes:
> The global min gas price is used as the minimum price for a transaction to enter a nodes mempool. The node still has a locally configured min price. This should be used when preparing the proposal such that all transaction included must be greater than that price

do transactions make it in the mempool that are less than the locally configured min fee? if not, then would the filter in prepare proposal every filter anything out?

cmwaters:
> do transactions make it in the mempool that are less than the locally configured min fee?

Yes so long as they are greater than the global min gas price. 

The locally configured min fee just means that this proposer won't include them but they may be included by another proposer. If the transaction has too low of a fee, it will eventually be kicked.

rootulp:
Sorry for the silly question but why do we want to support this?

ninabarbakadze:
update: we stopped working on this until we figure out next steps including whether or not we even want this feature.

rootulp:
I think we can close this as won't do b/c we decided to not modify the local min gas price. Copying from Slack:

> Decision to not modify the purpose of the local config. It can still be used to increase the min fee to be accepted in the mempool.

Please re-open if I misinterpreted @ninabarbakadze @cmwaters

PR #3034, opened 2024-01-22
feat: use nodes min gas price in prepare proposal

ninabarbakadze:
## Overview

Add a check in prepare proposal to make sure that all transaction fees meet node's configured local threshold.

Fixes - #3016


app/ante/fee_checker.go
```diff
@@ -23,27 +23,34 @@
...
+	if err != nil {
+		return errors.Wrap(err, "invalid minGasPrice: %s")
 	}
 
-	priority := getTxPriority(feeTx.GetFee(), int64(gas))
-	return feeTx.GetFee(), priority, nil
+	minFee := minGasPriceDec.MulInt(sdk.NewIntFromUint64(gas)).RoundInt()
+	if !fee.GTE(minFee) {
+		return errors.Wrapf(sdkerror.ErrInsufficientFee, "%s; got: %s required: %s", errMsg, fee, minFee)
+	}
```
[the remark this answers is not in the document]
rootulp: +1 this suggestion seems way more readable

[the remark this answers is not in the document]
rootulp: Perhaps also

```suggestion
	minFee := minGasPriceDec.MulInt(sdk.NewIntFromUint64(gas)).RoundInt()
	if !fee.GTE(minFee) {
		return errors.Wrapf(sdkerror.ErrInsufficientFee, "%s; got: %s required at least: %s", errMsg, fee, minFee)
	}
```

app/ante/min_fee_test.go
```diff
@@ -100,11 +100,19 @@ func TestCheckTxFeeWithGlobalMinGasPrices(t *testing.T) {
...
 			if tc.expErr {
-				require.Error(t, err)
+				require.Error(t, globalMinGasPriceErr)
+				require.Error(t, minGasPriceErr)
+				require.Equal(t, globalMinGasPriceErr.Error(), minGasPriceErr.Error())
 			} else {
-				require.NoError(t, err)
+				require.NoError(t, globalMinGasPriceErr)
+				require.NoError(t, minGasPriceErr)
+				require.Equal(t, globalMinGasPriceErr, minGasPriceErr)
```
rootulp, line 115: [optional] do we need this assertion if both errors are `nil`? I would never expect this line to catch something that wasn't caught by the previous two lines 🤔 

```suggestion
```

app/test/prepare_proposal_test.go
```diff
@@ -112,10 +113,39 @@ func TestPrepareProposalFiltering(t *testing.T) {
 		testutil.ChainID,
 	)).ToSliceOfBytes()
 
+	// create 3 MsgSend transactions with that are equal to the minimum fee
```
rootulp, line 116:
```suggestion
	// create 3 MsgSend transactions that are equal to the minimum fee
```

app/validate_txs.go
```diff
@@ -44,6 +46,18 @@ func filterStdTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handler
 			logger.Error("decoding already checked transaction", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()), "error", err)
 			continue
 		}
+
+		feeTx, ok := sdkTx.(sdk.FeeTx)
+		if !ok {
+			logger.Error("error")
+		}
```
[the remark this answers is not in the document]
rootulp: [blocking] I think the log message should be more informative here. The CodeRabbit suggestion seems to make sense.

app/ante/fee_checker.go
```diff
@@ -23,27 +23,34 @@ func CheckTxFeeWithGlobalMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, in
...
 	}
 
-	fee := feeTx.GetFee().AmountOf(appconsts.BondDenom)
-	gas := feeTx.GetGas()
-
 	// global minimum fee only applies to app versions greater than one
 	if ctx.BlockHeader().Version.App > v1.Version {
-		// convert the global minimum gas price to a big.Int
-		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.GlobalMinGasPrice))
+		err := CheckTxFeeWithMinGasPrices(feeTx, appconsts.GlobalMinGasPrice, "insufficient global minimum fee")
```
evan-forbes, line 28: [blocking question]

is this called during processproposal? if so we need to not to avoid any non-determinism. I didn't see a test for this but could have missed it, but that's something that we probably need. ProcessProposal should have the same result for a block with txs that are below the locally configured value

evan-forbes (commented):
unfortunately I don't think we can merge this until we reach social consensus on whether or not we want all nodes to accept and gossip txs below their configured fee.

also had a question on about if this is being called in processproposal, and we probably need a test to make sure that its not. 

otherwise LGTM

ninabarbakadze:
> unfortunately I don't think we can merge this until we reach social consensus on whether or not we want all nodes to accept and gossip txs below their configured fee.
> 
> also had a question on about if this is being called in processproposal, and we probably need a test to make sure that its not.
> 
> otherwise LGTM

Thanks for the review @evan-forbes @rootulp I will hold off working on this until we're all on the same page about wanting to introduce this feature.

app/ante/fee_checker.go
ninabarbakadze: `CheckTxFeeWithGlobalMinGasPrices` is called in processproposal through an antehandler, checking the global fee

rootulp:
@ninabarbakadze can this be closed b/c we're not doing it?

Ref: https://github.com/celestiaorg/celestia-app/pull/3094#discussion_r1530483923

ninabarbakadze:
> @ninabarbakadze can this be closed b/c we're not doing it?
> 
> Ref: [#3094 (comment)](https://github.com/celestiaorg/celestia-app/pull/3094#discussion_r1530483923)

closing it!

Issue #3035, opened 2024-01-23
move v2 related constants to v2 directory


ninabarbakadze:
## Summery

              [blocking][question] for @cmwaters. The breakdown of files in `pkg/appconsts` has me wondering where this constant should be defined. Candidates:

1. `consensus_consts.go` because this is a new "consensus constant"
2. `v2/app_consts.go` because this is a new param introduced in app version v2

If we do keep this in `initial_consts.go` then I think we need to update the comment at the top of this file because this constant can't be modified via on-chain governance or the node's local config.

```
// The following defaults correspond to initial parameters of the network that can be changed, not via app versions
// but other means such as on-chain governance, or the nodes local config
```

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/2985#discussion_r1456403580_

PR #3040, opened 2024-01-24
chore: move v2 consts to v2 directory

ninabarbakadze:
## Overview

moving GlobalMinGasPrice constant under v2

Fixes - #3035

### DIFF
--- a/app/ante/fee_checker.go
+++ b/app/ante/fee_checker.go
@@ -6,6 +6,7 @@ import (
 	errors "cosmossdk.io/errors"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
+	v2 "github.com/celestiaorg/celestia-app/pkg/appconsts/v2"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerror "github.com/cosmos/cosmos-sdk/types/errors"
 )
@@ -29,7 +30,7 @@ func CheckTxFeeWithGlobalMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, in
 	// global minimum fee only applies to app versions greater than one
 	if ctx.BlockHeader().Version.App > v1.Version {
 		// convert the global minimum gas price to a big.Int
-		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.GlobalMinGasPrice))
+		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", v2.GlobalMinGasPrice))
 		if err != nil {
 			return nil, 0, errors.Wrap(err, "invalid GlobalMinGasPrice")
 		}
--- a/app/ante/fee_checker.go
+++ b/app/ante/fee_checker.go
@@ -29,19 +29,19 @@ func CheckTxFeeWithGlobalMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, in
 
 	// global minimum fee only applies to app versions greater than one
 	if appVersion > v1.Version {
-		gmgp, err := appconsts.GlobalMinGasPrice(appVersion)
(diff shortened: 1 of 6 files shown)


app/ante/fee_checker.go
```diff
@@ -29,7 +30,7 @@ func CheckTxFeeWithGlobalMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, in
 	// global minimum fee only applies to app versions greater than one
 	if ctx.BlockHeader().Version.App > v1.Version {
 		// convert the global minimum gas price to a big.Int
-		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.GlobalMinGasPrice))
+		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", v2.GlobalMinGasPrice))
```
rootulp, line 33: [not blocking][question] if the constant is now app version specific, should we define a getter like this in `versioned_consts.go`:

```go
func GlobalMinGasPrice(version uint64) (float64, error) {
	switch version {
	case v2.Version:
		return v2.GlobalMinGasPrice, nil
	default:
		return 0, fmt.Errorf("global min gas price not defined for version %d", version)
	}
}
```

and then use here

```go
		gmgp, err := appconsts.GlobalMinGasPrice(ctx.BlockHeader().Version.App)
		if err != nil {
			return nil, 0, errors.Wrap(err, "invalid GlobalMinGasPrice")
		}

		// convert the global minimum gas price to a big.Int
		globalMinGasPrice, err := sdk.NewDecFromStr(fmt.Sprintf("%f", gmgp))
		if err != nil {
			return nil, 0, errors.Wrap(err, "invalid GlobalMinGasPrice")
		}
```

rootulp (approved):

app/ante/fee_checker.go
rootulp: Feel free to rename variables, add godocs, etc. Mostly for illustrative purposes ^

staheri14 (commented):
Looks good to me! I've added a suggestion to consider treating the `GlobalMinGasPrice` in a manner similar to other versioned constants.

app/ante/fee_checker.go
ninabarbakadze: thanks for the review and suggestions! I completely missed the `versioned_consts.go` file 🤦🏻‍♀️ updated now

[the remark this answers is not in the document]
ninabarbakadze: thanks for the review @staheri14! updated!

rootulp: No problemo. I wonder if the appconsts in v1 and v2 should actually be un-exported so that consumers are forced to use the getters rather than hard-coding a particular versioned constant.

I know Callum found a few instances where code was relying on a default or a particular versioned constant rather than querying for a versioned constant based on app version. Can discuss more outside this PR if maintainers think we should pursue this idea.

pkg/appconsts/versioned_consts.go
```diff
@@ -22,6 +24,17 @@ func SubtreeRootThreshold(_ uint64) int {
...
+// GlobalMinGasPrice is used in the processProposal to ensure
+// that all transactions have a gas price greater than or equal to this value.
+func GlobalMinGasPrice(version uint64) (float64, error) {
+	switch version {
+	case v2.Version:
+		return v2.GlobalMinGasPrice, nil
+	default:
+		return 0, fmt.Errorf("global min gas price not defined for version %d", version)
+	}
+}
```
rootulp, line 36: [optional] add tests in `versioned_consts_test.go`. The unit test may look crazy simple for now but I expect it may be more helpful down the line when we introduce new app versions.

rootulp (dismissed):

staheri14 (approved):

rootulp (approved):

app/ante/fee_checker.go
ninabarbakadze: yes, I think making them unexported makes sense. we should probably go through the app and replace all hardcoded values with getters+constants based on app version as you suggested. cc @cmwaters @evan-forbes

PR #3041, opened 2024-01-24
test: add unit test for global min gas price getter

ninabarbakadze:
## Overview

Added a unit test for GlobalMinGasPrice constant getter. https://github.com/celestiaorg/celestia-app/pull/3040#discussion_r1465446053

Refactored existing unit tests to make it more readable.

### DIFF
--- a/pkg/appconsts/versioned_consts_test.go
+++ b/pkg/appconsts/versioned_consts_test.go
@@ -65,7 +65,6 @@ func TestGlobalMinGasPrice(t *testing.T) {
 	}
 }
 
-
 func TestSquareSizeUpperBound(t *testing.T) {
 	type testCase struct {
 		version uint64
--- a/pkg/appconsts/versioned_consts_test.go
+++ b/pkg/appconsts/versioned_consts_test.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


rootulp (approved):

evan-forbes (approved):

pkg/appconsts/versioned_consts_test.go
```diff
@@ -13,39 +13,89 @@ import (
...
+
+	for _, tc := range testCases {
+		name := fmt.Sprintf("version %v", tc.version)
+		t.Run(name, func(t *testing.T) {
+			got, err := appconsts.GlobalMinGasPrice(tc.version)
+			if tc.expErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, tc.expected, got)
```
staheri14, line 69: I think this line is meant to be under the `if` block.

ninabarbakadze: I thought it was sufficient enough to just check that the error was thrown. if you think that it's better if I take that line out of the `if`  block I'm more than happy to!

staheri14: I previously thought that `tc.expected` referred to the expected error. No modification is necessary, the current version is correct.

staheri14 (approved):
LGTM!

Issue #3043, opened 2024-01-25
Markdown link check failures


rootulp:
## Context

markdown-link check has been failing on recent PRs.

## Problem
- https://github.com/celestiaorg/celestia-app/actions/runs/7649482027/job/20843954506?pr=3007
- https://github.com/celestiaorg/celestia-app/actions/runs/7654901109/job/20859806747

The problematic links identified so far are:
- https://ycharts.com/indicators/ethereum_cumulative_unique_addresses
- https://learnmeabitcoin.com/technical/little-endian
- https://www.rfc-editor.org/rfc/rfc6979.txt

## Proposal

1. Wrap those links with a markdown link disable
2. Or remove the links entirely

PR #3044, opened 2024-01-25
chore: ignore links that fail markdown-link-check

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3043 by ignoring links that actually work but sometimes fail in CI when markdown-link-check is run.

## Testing

`make markdown-link-check` passes locally


ninabarbakadze (approved):

evan-forbes (approved):

evan-forbes:
@rootulp do we need to backport these?

rootulp:
The most recent commits on v1.x don't fail for markdown-link-check: https://github.com/celestiaorg/celestia-app/commits/v1.x/ but I expect they would if there were more recent commits. So it's optional for us to backport. I'm indifferent on it, you?

PR #3046, opened 2024-01-25
refactor: extract init and denom from app.go

rootulp:
A small refactor to extract some of the contents of `app.go`. Motivated by a separate PR that tried to modify app.go which is currently at 700+ line file with very little consistency.

### DIFF
--- a/app/denom.go
+++ b/app/denom.go
@@ -1,0 +1,12 @@
+package app
+
+import "github.com/celestiaorg/celestia-app/pkg/appconsts"
+
+const (
+	// BondDenom defines the native staking token denomination.
+	BondDenom = appconsts.BondDenom
+	// BondDenomAlias defines an alias for BondDenom.
+	BondDenomAlias = "microtia"
+	// DisplayDenom defines the name, symbol, and display value of the Celestia token.
+	DisplayDenom = "TIA"
+)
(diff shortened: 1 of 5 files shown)


app/denom.go
```diff
@@ -0,0 +1,12 @@
+package app
```
cmwaters, line 1: Personally, I'd say this file is too small to constitute a new file but it's a slight preference

cmwaters (approved):

ninabarbakadze (approved):
lgtm

evan-forbes (approved):

app/denom.go
rootulp: hmm I'm not sure how to proceed. I'm trying to extract stuff from app.go because it's massive. IMO the denomination constants don't need to be located in the same file as the rest of the app set up logic.

If this doesn't constitute a new file, where should I put it? I don't think reverting and putting it back in app.go makes sense.

rootulp: I guess I could rename this file consts.go and put other constants defined in the `package app` in here? But calling it `consts.go` now felt premature

rootulp: Going to merge for now b/c we can easily change this in a follow up PR.

PR #3055, opened 2024-01-26
refactor: extract posthandler package

rootulp:
Another small refactor to make app.go more readable.


evan-forbes (approved):
I'd also be fine with adding the post handler to the ante handler file if we want to avoid too many small files

I don't have a strong opinion tho

ninabarbakadze (approved):

PR #3065, opened 2024-01-30
chore(deps): upgrade to rsmt2d v0.12.0 release candidate

rootulp:
Upgrade to https://github.com/celestiaorg/rsmt2d/releases/tag/v0.12.0-rc0 

This PR is mostly for testing purposes to verify CI passes and that we can cut an official release.


rootulp:
This fails https://github.com/celestiaorg/celestia-app/actions/runs/7715652203/job/21030690879?pr=3065#step:4:268

rootulp:
I think the square size and NMT options need to be encoded into the treeName to avoid conflicts with other trees. Trees also need to be registered once (and only once to avoid an error) prior to using them. This change feels extremely error-prone so proposal to revert the upstream breaking changes in rsmt2d and think through the `RegisterTree` design a bit more. Ideally the `treeName` isn't leaked down to consuming apps for them to manage.

Issue #3072, opened 2024-01-31
Potential DoS vector: blobs that are too big to be included in a square


rootulp:
## Problem

Originally reported via @tuxcanfly. See https://gist.github.com/tuxcanfly/207ab3dc1fac4bb9c38bedad2a053f18

## Proposal

Is it possible for `CheckTx` to reject this type of transaction prior admittance into the mempool? 
- A stateful check could be based on the `GovSquareSize` of 64 which would reject transactions larger than ~2MiB
- A stateless check could be based on the `SquareSizeUpperBound` of 128 which would reject transactions larger than ~8MiB

rootulp:
Note we already have an ante handler that rejects txs where the total blob size is too large. It's possible that ante handler doesn't catch edge cases near the total blob size because it derives an upper bound based on a number of parameters. See 
- https://github.com/celestiaorg/celestia-app/issues/2156
- https://github.com/celestiaorg/celestia-app/pull/2202
- https://github.com/celestiaorg/celestia-app/issues/2214

cmwaters:
I think we calculate the space taken by the blobs incorrectly https://github.com/celestiaorg/celestia-app/blob/2d261a2a0f7f627044342929b534a16b762c47e0/x/blob/ante/max_total_blob_size_ante.go#L74-L79

it should be in regards of the amount of shares * 512 bytes i.e.

```go
for _, size := range sizes {
	sum += uint64(shares.SparseSharesNeeded(size)) * appconsts.ShareSize
}

```

Also I would consider of adding an extra share or two as buffer for padding

rootulp:
`getTotal` is only used [here](https://github.com/celestiaorg/celestia-app/blob/2d261a2a0f7f627044342929b534a16b762c47e0/x/blob/ante/max_total_blob_size_ante.go#L33) where we want the number of total bytes in the blob sizes.

> Also I would consider of adding an extra share or two as buffer for padding

Agreed. That ante handler doesn't account for padding shares so this is related to https://github.com/celestiaorg/celestia-app/issues/3081

cmwaters:
> getTotal is only used [here](https://github.com/celestiaorg/celestia-app/blob/2d261a2a0f7f627044342929b534a16b762c47e0/x/blob/ante/max_total_blob_size_ante.go#L33) where we want the number of total bytes in the blob sizes.

yes but we want the total bytes based on the amount of shares that those blobs will occupy; not the raw bytes of the blobs

[...]

PR #3082, opened 2024-02-06
fix!: reject PFBs based on share count

rootulp:
@cmwaters noticed that the ante handler rejects txs but doesn't account for the shares used by each blob size. See https://github.com/celestiaorg/celestia-app/issues/3072#issuecomment-1930466074

This means that a PFB with many small blobs would not be rejected by the ante handler even though it can't possibly fit. This PR modifies the ante handler to reject based on the # of shares occupied by each `blobSize` instead of the # of total bytes. Also adds a test case that would fail with the previous implementation but passes with this implementation.

--- 

Note to future selves: we should add to the v2 release notes:

- breaking!: the blob ante handler became more restrictive in that it rejects transactions that would have previously entered the mempool but failed to be included in a block. Transactions that previously failed with `ErrTotalBlobSizeTooLarge` should now fail with `ErrBlobTooLarge`.


x/blob/ante/max_total_blob_size_ante.go
```diff
@@ -20,36 +20,33 @@ func NewMaxBlobSizeDecorator(k BlobKeeper) MaxTotalBlobSizeDecorator {
...
-// total blob size less than this max still fail to be included in a block due
-// to overhead from the PFB tx and/or padding shares.
-func (d MaxTotalBlobSizeDecorator) maxTotalBlobSize(ctx sdk.Context) int {
+// getMaxBlobShares returns the max the number of shares available for blob data.
+func (d MaxTotalBlobSizeDecorator) getMaxBlobShares(ctx sdk.Context) int {
 	squareSize := d.getMaxSquareSize(ctx)
 	totalShares := squareSize * squareSize
 	// The PFB tx share must occupy at least one share so the # of blob shares
 	// is at least one less than totalShares.
 	blobShares := totalShares - 1
```
cmwaters, line 48: Suggestion: do we want to bump this from 1 to 2 to have one intranamespace padding as buffer

I just think it's better to be conservative about this

cmwaters (dismissed):
LGTM

x/blob/ante/max_total_blob_size_ante.go
rootulp: Yea I can see the argument to account for padding as buffer. But how much padding do we account for?
- The square could contain one or more [namespace padding](https://celestiaorg.github.io/celestia-app/specs/shares.html#namespace-padding-share) 
- The square could contain one or more [primary reserved padding](https://celestiaorg.github.io/celestia-app/specs/shares.html#primary-reserved-padding-share)

My concern is that if this ante handler accounted for worst case padding then it may reject blobs from inclusion into the mempool even though that tx could be included in a square.

One idea for a more sophisticated ante handler is to construct a data square with just this one tx and verify that the # of shares used is < maxSquareSize * maxSquareSize

rootulp:
This ante handler could be refactor to use `GovSquareSizeUpperBound` in `square_size.go`

x/blob/ante/max_total_blob_size_ante.go
```diff
@@ -20,36 +20,33 @@ func NewMaxBlobSizeDecorator(k BlobKeeper) MaxTotalBlobSizeDecorator {
 }
 
 // AnteHandle implements the Cosmos SDK AnteHandler function signature. It
-// returns an error if tx contains a MsgPayForBlobs where the total blob size is
-// greater than the max total blob size.
+// returns an error if tx contains a MsgPayForBlobs where the blob sizes would
+// occupy more shares than the largest possible data square.
```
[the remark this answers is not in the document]
rootulp: Fixed in [fef0a87](https://github.com/celestiaorg/celestia-app/pull/3082/commits/fef0a87e52a85a057a8f409209d38abe8d3e7c1e)

x/blob/ante/blob_share_decorator.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return next(ctx, tx, simulate)
+}
+
+// getMaxBlobShares returns the max the number of shares available for blob data.
+func (d BlobShareDecorator) getMaxBlobShares(ctx sdk.Context) int {
+	squareSize := d.getMaxSquareSize(ctx)
+	totalShares := squareSize * squareSize
+	// The PFB tx share must occupy at least one share so the number of blob shares
+	// is at least one less than totalShares.
```
staheri14, line 48: [nit]
```suggestion
	// is at most one less than totalShares.
```

x/blob/ante/blob_share_decorator.go
```diff
@@ -0,0 +1,78 @@
...
+	// BeginBlock of the first block. This is remedied in versions of the sdk
+	// and comet that have full support of PreparePropsoal, although
+	// celestia-app does not currently use those. see this PR for more details
+	// https://github.com/cosmos/cosmos-sdk/pull/14505
+	if ctx.BlockHeader().Height <= 1 {
+		return int(appconsts.DefaultGovMaxSquareSize)
+	}
+
+	upperBound := appconsts.SquareSizeUpperBound(ctx.BlockHeader().Version.App)
+	govParam := d.k.GovMaxSquareSize(ctx)
```
staheri14, line 67: [Question] Is it possible for the `govParam` value to exceed the maximum limit permitted by the app version?

x/blob/ante/blob_share_decorator_test.go
```diff
@@ -67,7 +66,20 @@ func TestMaxTotalBlobSizeAnteHandler(t *testing.T) {
 			// single blob that is 2 MiB returns an error.
 			wantErr: true,
 		},
-		// tests based on shares
+		{
+			name: "PFB with many single byte blobs should fit",
+			pfb: &blob.MsgPayForBlobs{
+				BlobSizes: repeat(4095, 1), // (64 * 64) - 1 = 4095 so this should fit
```
staheri14, line 72: [Question] What is the size of each share in the square? it seems to be assumed to be 1 byte?

x/blob/types/errors.go
```diff
@@ -33,5 +33,7 @@ var (
 	ErrNoShareCommitments             = errors.Register(ModuleName, 11135, "no share commitments provided")
 	ErrInvalidNamespace               = errors.Register(ModuleName, 11136, "invalid namespace")
 	ErrInvalidNamespaceVersion        = errors.Register(ModuleName, 11137, "invalid namespace version")
-	ErrTotalBlobSizeTooLarge          = errors.Register(ModuleName, 11138, "total blob size too large")
+	// ErrTotalBlobSize is deprecated, use ErrBlobTooLarge instead.
+	ErrTotalBlobSizeTooLarge = errors.Register(ModuleName, 11138, "total blob size too large")
+	ErrBlobTooLarge          = errors.Register(ModuleName, 11139, "blob(s) too large")
```
staheri14, line 38: [nit-suggestion]  
```suggestion
	ErrBlobsTooLarge          = errors.Register(ModuleName, 11139, "blob(s) too large")
```

x/blob/ante/blob_share_decorator_test.go
```diff
@@ -124,11 +136,12 @@ func TestMaxTotalBlobSizeAnteHandler(t *testing.T) {
 			require.NoError(t, txBuilder.SetMsgs(tc.pfb))
 			tx := txBuilder.GetTx()
 
-			mbsd := ante.NewMaxBlobSizeDecorator(mockBlobKeeper{})
+			mbsd := ante.NewBlobShareDecorator(mockBlobKeeper{})
 			_, err := mbsd.AnteHandle(ctx, tx, false, mockNext)
 
 			if tc.wantErr {
 				assert.Error(t, err)
+				assert.ErrorIs(t, err, blob.ErrBlobTooLarge)
```
staheri14, line 144: [suggestion-optional] For better readability, the error type could have been part of the `testCase` struct.

staheri14 (dismissed):
LGTM! nice documentation 👍

x/blob/ante/blob_share_decorator.go
rootulp: Yes 😞 see https://github.com/celestiaorg/celestia-app/issues/2185

x/blob/ante/blob_share_decorator_test.go
rootulp: this is trying to convey that if there are 4095 blobs each of one byte then the entire PFB should fit. Recall that even if a blob only contains one byte, it must occupy an entire blob share because multiple blobs cannot occupy one share.

So this could theoretically be `repeat(4095, 100)` or 4095 blobs each of size 100 bytes because a single 100 byte blob occupies a single share.

x/blob/types/errors.go
rootulp: Updated in 0774127b5058d1610413aedc16ec21df87ca44ca

x/blob/ante/blob_share_decorator_test.go
rootulp: Clarified in 9c448ae940b500fae327095b5f3cd049adca1b02

x/blob/ante/blob_share_decorator_test.go
rootulp: Good point, resolved in 5bdc057d4a24c248c72e24767ed8dfca202eed60

staheri14 (approved):
Thanks for addressing the comments, still LGTM!

x/blob/ante/max_total_blob_size_ante.go
cmwaters: > My concern is that if this ante handler accounted for worst case padding then it may reject blobs from inclusion into the mempool even though that tx could be included in a square.

That's probably okay if we're talking about a blob that is still 64 * 64 shares i.e. it's a pretty big blob. It's probably better that the user gets the feedback immediately via `CheckTx` then having their transaction mysteriously unable to get on-chain

cmwaters (approved):
👍

Issue #3092, opened 2024-02-08
Make GlobalMinGasPrice constant gov modifiable


ninabarbakadze:
**Description**

Currently GlobalMinGasPrice is a hard-coded constant but should be gov modifiable.

rootulp:
Note, if we make it gov modifiable we should also add it to the specs page: [Mainnet Parameters](https://celestiaorg.github.io/celestia-app/specs/params.html) and [cip-13](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-13.md)

PR #3094, opened 2024-02-08
feat: make global min gas price gov modifiable

ninabarbakadze:
## Overview

Makes GlobalMinGasPrice hard-coded constant gov modifiable

Fixes - #3092


x/blob/types/estimate_gas_test.go
```diff
@@ -39,7 +39,7 @@ func TestPFBGasEstimation(t *testing.T) {
 	for idx, tc := range testCases {
 		t.Run(fmt.Sprintf("case %d", idx), func(t *testing.T) {
 			accnts := testfactory.GenerateAccounts(1)
-			testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultConsensusParams(), accnts...)
+			testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultInitialConsensusParams(), accnts...)
```
ninabarbakadze, line 42: note for the reviewer: after i updated DefaultConsensusParams function this test started estimating a tiny bit higher gas, hence why i left it with initial consensus params. 

From the initial look, I'm not sure why updating the app version would result into higher estimates.

app/ante/fee_checker.go
```diff
@@ -7,7 +7,7 @@ import (
 	"cosmossdk.io/math"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
-	"github.com/celestiaorg/celestia-app/x/minfee"
+	minfee "github.com/celestiaorg/celestia-app/x/minfee"
```
cmwaters, line 10: the alias isn't necessary

app/ante/fee_checker.go
```diff
@@ -47,19 +47,19 @@ func CheckTxFeeWithGlobalMinGasPrices(ctx sdk.Context, tx sdk.Tx, paramKeeper pa
 	// global minimum fee only applies to app versions greater than one
 	if ctx.BlockHeader().Version.App > v1.Version {
 		// fmt.Println("GOT HEREEEEEEE")
```
cmwaters: Can you remove print statement

app/ante/ante.go
```diff
@@ -75,3 +77,10 @@ func NewAnteHandler(
 }
 
 var DefaultSigVerificationGasConsumer = ante.DefaultSigVerificationGasConsumer
+
+// The purpose of this wrapper is to ensure that CheckTxFeeWithGlobalMinGasPrices complies with the ante.TxFeeChecker type.
+func CheckTxFeeWithGlobalMinGasPricesWrapper(paramKeeper paramkeeper.Keeper) ante.TxFeeChecker {
```
cmwaters, line 82: We actually use this wrapper as a means of passing the paramKeeper into the ante.TxFeeChecker function because it's needed to get the global min gas price

app/ante/fee_checker.go
```diff
@@ -4,51 +4,85 @@ import (
...
+	// is only ran on check tx.
+	if ctx.IsCheckTx() {
+		defaultMinGasPriceDec, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.DefaultMinGasPrice))
+		if err != nil {
+			return nil, 0, errors.Wrapf(err, "invalid defaultMinGasPrice: %f", defaultMinGasPriceDec)
+		}
 
-		// convert the global minimum gas price to a big.Int
-		globalMinGasPriceInt, err := sdk.NewDecFromStr(fmt.Sprintf("%f", globalMinGasPrice))
+		err = verifyMinFee(fee, gas, defaultMinGasPriceDec, "insufficient validator minimum fee")
```
cmwaters, line 42:
```suggestion
		err = verifyMinFee(fee, gas, defaultMinGasPriceDec, "insufficient minimum gas price for this validator")
```

cmwaters: might make it a little more understandable

app/ante/fee_checker.go
```diff
@@ -4,51 +4,85 @@ import (
...
+		if !subspace.Has(ctx, minfee.KeyGlobalMinGasPrice) {
+			return nil, 0, errors.Wrap(sdkerror.ErrKeyNotFound, "GlobalMinGasPrice")
+		}
+
+		var globalMinGasPrice sdk.Dec
+		// Gets the global minimum gas price from the param store
+		// panics if not configured properly
+		subspace.Get(ctx, minfee.KeyGlobalMinGasPrice, &globalMinGasPrice)
+
+		err := verifyMinFee(fee, gas, globalMinGasPrice, "insufficient global minimum fee")
```
cmwaters, line 65:
```suggestion
		err := verifyMinFee(fee, gas, globalMinGasPrice, "insufficient gas price for the network")
```

app/default_overrides.go
```diff
@@ -198,6 +198,17 @@ func getLegacyProposalHandlers() (result []govclient.ProposalHandler) {
...
 func DefaultConsensusParams() *tmproto.ConsensusParams {
+	return &tmproto.ConsensusParams{
+		Block:     DefaultBlockParams(),
+		Evidence:  DefaultEvidenceParams(),
+		Validator: coretypes.DefaultValidatorParams(),
+		Version: tmproto.VersionParams{
+			AppVersion: appconsts.LatestVersion,
+		},
+	}
+}
```
cmwaters, line 209: Is this used anywhere. Note: we have a duplicate in the `testnode` package

cmwaters (commented):
Nice PR. Logic seems sounds and sufficiently tested. Just have a few nits before we can merge

app/ante/fee_checker.go
```diff
@@ -4,51 +4,85 @@ import (
...
 
 	priority := getTxPriority(feeTx.GetFee(), int64(gas))
 	return feeTx.GetFee(), priority, nil
 }
 
+// verifyMinFee validates that the provided transaction fee is sufficient given the provided minimum gas price.
+func verifyMinFee(fee math.Int, gas uint64, minGasPrice sdk.Dec, errMsg string) error {
+	// Determine the required fee by multiplying required minimum gas
+	// price by the gas limit, where fee = minGasPrice * gas.
+	minFee := minGasPrice.MulInt(sdk.NewIntFromUint64(gas)).RoundInt()
```
rootulp, line 79: [question] should this take the ceiling instead of round? If it rounds down then a transaction could get included that pays a fee slightly less than the one derived used global min gas price.

app/ante/min_fee_test.go
```diff
@@ -29,67 +39,88 @@ func TestCheckTxFeeWithGlobalMinGasPrices(t *testing.T) {
...
+			expErr:     false,
+		},
+		{
+			name:       "good tx; fee above node's required minimum",
+			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, feeAmount+1)),
+			gasLimit:   uint64(float64(feeAmount) / appconsts.DefaultGlobalMinGasPrice),
+			appVersion: uint64(1),
+			isCheckTx:  true,
 			expErr:     false,
 		},
```
rootulp, line 115: [question] would it be helpful to include a test case where a node operators locally configured min gas price differs from the constant at appconsts.DefaultGlobalMinGasPrice?

rootulp (commented):
Do you want to update the specs in this PR or a follow-up PR? We have to update CIP-13 in a follow-up PR because separate repo. Do you mind creating an issue or PR in the cips repo to track?

Ref: https://github.com/celestiaorg/celestia-app/issues/3092#issuecomment-1934518838

[question] did we use the Cosmos Hub [globalfee](https://github.com/cosmos/gaia/tree/main/x/globalfee) module as inspiration for this?

app/ante/ante.go
ninabarbakadze: yeah I wanted to express that function signatures were otherwise incompatible because we added extra parameter - paramKeeper

x/minfee/upgrade_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}})
+
+	// app version should not have changed yet
+	require.EqualValues(t, 1, testApp.AppVersion())
+
+	// global min gas price should not have been set yet
+	gotBefore, err := testApp.ParamsKeeper.Params(ctx, &proposal.QueryParamsRequest{
+		Subspace: minfee.ModuleName,
+		Key:      string(minfee.KeyGlobalMinGasPrice),
+	})
```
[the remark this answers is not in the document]
ninabarbakadze: I think it's because the subspace exists? I tried querying it with multiple random keys and I got the same results back
Screenshot 2024-03-18 at 20 04 43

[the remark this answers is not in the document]
ninabarbakadze: when subspace doesn't exist it returns nil and throws error

x/paramfilter/test/gov_params_test.go
```diff
@@ -486,6 +487,23 @@ func (suite *GovParamsTestSuite) TestModifiableParams() {
 				assert.Equal(want, got)
 			},
 		},
+		{
+			"minfee.GlobalMinGasPrice",
+			testProposal(proposal.ParamChange{
+				Subspace: minfeetypes.ModuleName,
+				Key:      string(minfeetypes.KeyGlobalMinGasPrice),
+				Value:    `"0.002000000000000000"`,
```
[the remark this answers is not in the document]
ninabarbakadze: yeah good call! I'll set it to a new value.

app/ante/fee_checker.go
```diff
@@ -4,51 +4,85 @@ import (
...
 const (
 	// priorityScalingFactor is a scaling factor to convert the gas price to a priority.
 	priorityScalingFactor = 1_000_000
 )
 
-// CheckTxFeeWithGlobalMinGasPrices implements the default fee logic, where the minimum price per
-// unit of gas is fixed and set globally, and the tx priority is computed from the gas price.
-func CheckTxFeeWithGlobalMinGasPrices(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
+// CheckTxFeeWithMinGasPrices implements default fee validation logic for transactions.
+// It ensures that the provided transaction fee meets a minimum threshold for the validator
```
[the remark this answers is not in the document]
ninabarbakadze: the gov modifiable param is more of a general minimum that we expect where we still want validators to be able to increase their local fee

proto/celestia/minfee/v1/genesis.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package celestia.minfee.v1;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/minfee";
+
+// GenesisState defines the minfee module's genesis state.
+message GenesisState {
+    double global_min_gas_price = 1;
```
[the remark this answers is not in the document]
ninabarbakadze: since the original value was a float type I wanted to save it in original form, but we could change it if you think it's necessary

app/ante/fee_checker.go
```diff
@@ -4,51 +4,85 @@ import (
...
-	appVersion := ctx.BlockHeader().Version.App
 
-	// global minimum fee only applies to app versions greater than one
-	if appVersion > v1.Version {
-		globalMinGasPrice := appconsts.GlobalMinGasPrice(appVersion)
+	// Ensure that the provided fee meets a minimum threshold for the validator.
+	// This is only for local mempool purposes, and thus
+	// is only ran on check tx.
+	if ctx.IsCheckTx() {
+		defaultMinGasPriceDec, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.DefaultMinGasPrice))
```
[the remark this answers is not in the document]
ninabarbakadze: you're right! it should be `ctx.MinGasPrices()`

x/minfee/module.go
```diff
@@ -0,0 +1,145 @@
...
+func (am AppModule) RegisterInvariants(_ sdk.InvariantRegistry) {}
+
+// Route returns the message routing key for the minfee module.
+func (am AppModule) Route() sdk.Route {
+	return sdk.Route{}
+}
+
+// QuerierRoute returns the minfee module's querier route name.
+func (am AppModule) QuerierRoute() string {
+	return ""
```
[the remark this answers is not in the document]
ninabarbakadze: do we need a querier route?

ninabarbakadze:
> Do you want to update the specs in this PR or a follow-up PR? We have to update CIP-13 in a follow-up PR because separate repo. Do you mind creating an issue or PR in the cips repo to track?
> 
> Ref: [#3092 (comment)](https://github.com/celestiaorg/celestia-app/issues/3092#issuecomment-1934518838)
> 
> [question] did we use the Cosmos Hub [globalfee](https://github.com/cosmos/gaia/tree/main/x/globalfee) module as inspiration for this?

1. I updated the `params.md` to include this new param. we still need to update state_machine_modules.md with link to our new modules + minfee module once it's merged. I will create an issue for it.
2. I tried to follow the pattern in our existing modules.

x/minfee/module.go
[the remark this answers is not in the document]
cmwaters: No, we're just mocking the rest of the interface for the module. We only really need the module for initialising the genesis

app/ante/fee_checker.go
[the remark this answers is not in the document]
cmwaters: > For some reason, I thought we were going to ignore the value set by validators in config and instead use the gov modifiable param.

We definitely went back and forward a couple times on this but the decision was to leave the existing behaviour as is i.e. the local validator minimum for check tx alongside the global minimum check for prepare/process proposal. 

Iirc the justification, we weren't yet sure of the importance of mempool synchrony

cmwaters (approved):
Great work!

app/ante/fee_checker.go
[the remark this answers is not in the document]
rootulp: Ahh now I remember that. I think that means we can close https://github.com/celestiaorg/celestia-app/pull/3034 as won't do.

[optional] we could update the comment here to say that the config minimum only applies for `CheckTx` and the global minimum applies for `Prepare/ProcessProposal`.

Update: it's described in the body of this function so definitely optional / not blocking.

app/ante/min_fee_test.go
```diff
@@ -27,69 +37,96 @@ func TestCheckTxFeeWithGlobalMinGasPrices(t *testing.T) {
...
+		{
+			name:       "bad tx; fee below node's required minimum",
+			fee:        sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, feeAmount-1)),
+			gasLimit:   uint64(float64(feeAmount) / validatorMinGasPrice),
+			appVersion: uint64(1),
+			isCheckTx:  true,
+			expErr:     true,
+		},
```
rootulp, line 129: 👍

specs/src/specs/params.md
```diff
@@ -64,5 +64,6 @@ are blocked by the `x/paramfilter` module.
...
 | staking.MinCommissionRate                     | 0.05 (5%)                                   | Minimum commission rate used by all validators.                                                                                                                                                 | True                      |
 | staking.UnbondingTime                         | 1814400 (21 days)                           | Duration of time for unbonding in seconds.                                                                                                                                                      | False                     |
+| minfee.GlobalMinGasPrice                      | 0.002                                        | All transactions must have a gas price greater than or equal to this value                                                                                                                             | True                      |
```
rootulp, line 67: [nit] this param table is alphabetically sorted. Can you please move `minfee` up?

rootulp: [tangent] we don't have to figure this out in this PR but I wonder if it would make sense to include the app version that a constant is applicable for in this table. For example, I think some of these params could be modified to be different in different app versions and at that point it's not immediately obvious if this table should reflect the original or updated value. Perhaps both with a new column for app version?

rootulp (approved):
+1 great work!

x/blob/types/estimate_gas_test.go
ninabarbakadze: thoughts? @cmwaters @rootulp @evan-forbes

x/minfee/module.go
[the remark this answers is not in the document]
ninabarbakadze: yeah thought so but I added it anyway. we could remove it or we could leave it.

specs/src/specs/params.md
ninabarbakadze: yes, i can move it in a follow up pr. thanks @rootulp 

and to you second point i think that's a good call. I'll export your comment as an issue.

x/minfee/module.go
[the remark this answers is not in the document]
rootulp: Sorry Cosmos SDK docs are awful. 

https://docs.cosmos.network/v0.46/building-modules/module-manager.html#appmodule says:

```
	// Deprecated: use RegisterServices
	QuerierRoute() string
```

> QuerierRoute() (deprecated): Returns the name of the module's query route, for [queries](https://docs.cosmos.network/v0.46/building-modules/messages-and-queries.html#queries) to be routes to the module by [BaseApp](https://docs.cosmos.network/v0.46/core/baseapp.html#query-routing).

and this separate page: https://docs.cosmos.network/v0.46/building-modules/messages-and-queries.html#queries which makes me think `QuerierRoute` is only relevant if the file has defined a `query.proto`. Since the minfee module doesn't have any queries, we likely don't need it but I don't see the harm in returning the module name.

app/ante/ante.go
```diff
@@ -75,3 +77,11 @@ func NewAnteHandler(
 }
 
 var DefaultSigVerificationGasConsumer = ante.DefaultSigVerificationGasConsumer
+
+// The purpose of this wrapper is to enable the passing of an additional paramKeeper parameter
+// whilst still satisfying the ante.TxFeeChecker type.
+func CheckTxFeeWithGlobalMinGasPricesWrapper(paramKeeper paramkeeper.Keeper) ante.TxFeeChecker {
```
evan-forbes, line 83: [optional]

the name `CheckTx` could be confusing

app/ante/fee_checker.go
```diff
@@ -1,54 +1,84 @@
...
-	// global minimum fee only applies to app versions greater than one
-	if appVersion > v1.Version {
-		globalMinGasPrice := appconsts.GlobalMinGasPrice(appVersion)
+	// Ensure that the provided fee meets a minimum threshold for the validator.
+	// This is only for local mempool purposes, and thus
+	// is only ran on check tx.
+	if ctx.IsCheckTx() {
+		minGasPrice := ctx.MinGasPrices().AmountOf(appconsts.BondDenom)
+		if !minGasPrice.IsZero() {
+			err := verifyMinFee(fee, gas, minGasPrice, "insufficient minimum gas price for this validator")
```
evan-forbes, line 37: [optoinal]

since all nodes will run this, validator might be confusing to users

evan-forbes (approved):

evan-forbes:
thanks for going through a rigorous testing process

ninabarbakadze:
> thanks for testing rigorously!
> 
> only had optional naming related things left!

thanks evan! I'll update in a follow up pr.

PR #3104, opened 2024-02-13
chore: adds a note about namespace collisions in the namespace spec

### LINKED ISSUE #2125: namespace (audit): clarifying the implications of namespace overlap
# Problem
The following issue aims to capture the potential consequences of two applications choosing the same namespace. It seeks to address the following questions:

1. What can happen if two applications choose the same namespace?
2. Could it cause any security issues, such as spamming each other's data space?
3. Would it make application transaction retrieval less efficient, as the data of one application is mixed with another? (my initial thought is that this mixing of data should not inherently affect efficiency. Each blob transaction typically contains the share indices that allow for the retrieval of application-specific data. However, it is possible that Rollkit may employ a different approach for fetching application data, not solely relying on the share indices of the submitted transaction.)

If any of the above problems are found to be problematic, it would be advisable to suggest that applications choose their namespaces randomly or use any other approach that minimizes the likelihood of collision.

# Acceptance Criteria
Based on the answers to the questions above, it is recommended to update the Consideration and Assumption section of the namespace specification to include any necessary recommendations.

staheri14:
Closes #2125

### DIFF
--- a/specs/src/specs/namespace.md
+++ b/specs/src/specs/namespace.md
@@ -87,6 +87,10 @@ For additional information on the significance and application of the reserved n
 
 Applications MUST refrain from using the [reserved namespaces](#reserved-namespaces) for their blob data.
 
+Celestia does not ensure the prevention of non-reserved namespace collisions. 
+Consequently, two distinct applications might use the same namespace.
+It is the responsibility of these applications to be cautious and manage the implications and consequences arising from such namespace collisions.
+
 ## Implementation
 
 See [go-square/namespace](https://github.com/celestiaorg/go-square/tree/main/namespace).


rootulp (approved):

specs/src/specs/namespace.md
```diff
@@ -87,6 +87,10 @@ For additional information on the significance and application of the reserved n
 
 Applications MUST refrain from using the [reserved namespaces](#reserved-namespaces) for their blob data.
 
+Celestia does not ensure the prevention of non-reserved namespace collisions. 
+Consequently, two distinct applications might use the same namespace.
+It is the responsibility of these applications to be cautious and manage the implications and consequences arising from such namespace collisions.
```
evan-forbes, line 92: we might also want to link to the "Woods Attack" if this is not already linked somewhere

https://forum.celestia.org/t/woods-attack-on-celestia/59

evan-forbes (approved):

ninabarbakadze (approved):

cmwaters (approved):

specs/src/specs/namespace.md
staheri14: Sure, will do in a separate PR [to not to dismiss the current approvals].

Issue #3105, opened 2024-02-13
Add support for migrations during the upgrade process


cmwaters:
### Summary

The upgrade protocol currently isn't wiring up the `Configurator` which is where modules can register migrations that should be executed in the `EndBlocker` when going from one version to another.

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

Issue #3052, opened 2024-01-26
Consider using one order for BeginBlockers, EndBlockers, and InitGenesis


rootulp:
## Context

app.go has 3 separate orders for BeginBlockers, EndBlockers, and InitGenesis

https://github.com/celestiaorg/celestia-app/blob/30aa1b0dddf5871f51b99e60b925dca6caeacc95/app/app.go#L399-L475

## Proposal

If the 3 orders don't actually need to be different, proposal to refactor into something like how [Osmosis does it](https://github.com/osmosis-labs/osmosis/blob/7c81b90825ab2efe92444ac167191b8d041e0c21/app/app.go#L311-L317).

PR #3112, opened 2024-02-15
feat: support migrations during version upgrades

cmwaters:
Closes: #3105


app/app.go
```diff
@@ -365,32 +366,35 @@ func New(
...
+		module.NewVersionedModule(distr.NewAppModule(appCodec, app.DistrKeeper, app.AccountKeeper, app.BankKeeper, app.StakingKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(staking.NewAppModule(appCodec, app.StakingKeeper, app.AccountKeeper, app.BankKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(evidence.NewAppModule(app.EvidenceKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(authzmodule.NewAppModule(appCodec, app.AuthzKeeper, app.AccountKeeper, app.BankKeeper, app.interfaceRegistry), v1.Version, v2.Version),
+		module.NewVersionedModule(ibc.NewAppModule(app.IBCKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(params.NewAppModule(app.ParamsKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(transfer.NewAppModule(app.TransferKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(blob.NewAppModule(appCodec, app.BlobKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(blobstream.NewAppModule(appCodec, app.BlobstreamKeeper), v1.Version, v2.Version),
+		module.NewVersionedModule(upgrade.NewAppModule(app.UpgradeKeeper), v2.Version, v2.Version),
```
rootulp, line 393: [question] why is v2.Version used as both the second and third param? Is it because the upgrade module only supports v2 and isn't enabled in v1?

See other comment but Go docs on what the `fromVersion` and `toVersion` describe would help.

app/app.go
```diff
@@ -519,28 +523,37 @@ func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.R
...
+				panic(err)
+			}
 		}
 	}
 	return res
 }
 
+func (app *App) Upgrade(ctx sdk.Context, version uint64) error {
+	app.SetAppVersion(ctx, version)
+	return app.mm.RunMigrations(ctx, app.configurator, version, version)
```
rootulp, line 544: [question] why is `version` used for both `fromVersion` and `toVersion`? If we're upgrading, from a version to a different version, I would expect these two versions to be distinct.

Can a clarifying comment be added?

app/module/configurator.go
```diff
@@ -0,0 +1,93 @@
...
+
+// QueryServer implements the Configurator.QueryServer method
+func (c configurator) QueryServer() grpc.Server {
+	return c.queryServer
+}
+
+// RegisterMigration implements the Configurator.RegisterMigration method
+func (c configurator) RegisterMigration(moduleName string, fromVersion uint64, handler module.MigrationHandler) error {
+	if fromVersion == 0 {
+		return sdkerrors.ErrInvalidVersion.Wrap("module migration versions should start at 1")
```
rootulp, line 48: does this mean no migrations can occur from v0 -> v1? Does other code already assume version starts at v1?

app/module/configurator.go
```diff
@@ -0,0 +1,93 @@
...
+	c.migrations[moduleName][fromVersion] = handler
+
+	return nil
+}
+
+// runModuleMigrations runs all in-place store migrations for one given module from a
+// version to another version.
+func (c configurator) runModuleMigrations(ctx sdk.Context, moduleName string, fromVersion, toVersion uint64) error {
+	// No-op if toVersion is the initial version or if the version is unchanged.
+	if toVersion <= 1 || fromVersion == toVersion {
```
rootulp, line 68: given the logic `fromVersion == toVersion`, I expect the other line of code to never run any migrations

```go
	return app.mm.RunMigrations(ctx, app.configurator, version, version)
```

app/module/module.go
```diff
@@ -0,0 +1,349 @@
...
+	moduleMap := make(map[uint64]map[string]sdkmodule.AppModule)
+	allModules := make([]sdkmodule.AppModule, len(modules))
+	modulesStr := make([]string, 0, len(modules))
+	firstVersion, lastVersion := uint64(0), uint64(0)
+	for idx, module := range modules {
+		if module.fromVersion == 0 {
+			return nil, sdkerrors.ErrInvalidVersion.Wrapf("v0 is not a valid version for module %s", module.module.Name())
+		}
+		if module.fromVersion > module.toVersion {
+			return nil, sdkerrors.ErrLogic.Wrapf("toVersion can not be less than fromVersion for module %s", module.module.Name())
```
rootulp, line 54: [uber nit] can the conditional logic and the error message match so that it's easier to mentally verify 

```suggestion
			return nil, sdkerrors.ErrLogic.Wrapf("fromVersion can not be greater than toVersion for module %s", module.module.Name())
```

app/module/module.go
```diff
@@ -0,0 +1,349 @@
...
+	m.OrderInitGenesis = moduleNames
+}
+
+// SetOrderExportGenesis sets the order of export genesis calls
+func (m *Manager) SetOrderExportGenesis(moduleNames ...string) {
+	m.assertNoForgottenModules("SetOrderExportGenesis", moduleNames)
+	m.OrderExportGenesis = moduleNames
+}
+
+// SetOrderBeginBlockers sets the order of set begin-blocker calls
```
rootulp, line 96:
```suggestion
// SetOrderBeginBlockers sets the order of begin-blocker calls
```

app/module/module.go
```diff
@@ -0,0 +1,349 @@
...
+func (m *Manager) SetOrderBeginBlockers(moduleNames ...string) {
+	m.assertNoForgottenModules("SetOrderBeginBlockers", moduleNames)
+	m.OrderBeginBlockers = moduleNames
+}
+
+// SetOrderEndBlockers sets the order of set end-blocker calls
```
rootulp, line 102:
```suggestion
// SetOrderEndBlockers sets the order of end-blocker calls
```

app/module/module.go
```diff
@@ -0,0 +1,349 @@
...
+	if !ok {
+		return sdkerrors.ErrInvalidType.Wrapf("expected %T, got %T", configurator{}, cfg)
+	}
+	modules := m.OrderMigrations
+	if modules == nil {
+		modules = DefaultMigrationsOrder(m.ModuleNames(toVersion))
+	}
+	currentVersionModules, exists := m.versionedModules[fromVersion]
+	if !exists {
+		return sdkerrors.ErrInvalidVersion.Wrapf("version %d not supported", fromVersion)
```
rootulp, line 231: [optional] to disambiguate from the error below

```suggestion
		return sdkerrors.ErrInvalidVersion.Wrapf("fromVersion %d not supported", fromVersion)
```

app/module/module.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	nextVersionModules, exists := m.versionedModules[toVersion]
+	if !exists {
+		return sdkerrors.ErrInvalidVersion.Wrapf("version %d not supported", toVersion)
```
rootulp, line 235: [optional] to disambiguate from the error above

```suggestion
		return sdkerrors.ErrInvalidVersion.Wrapf("toVersion %d not supported", toVersion)
```

app/module/module.go
```diff
@@ -0,0 +1,349 @@
...
+
+	return &Manager{
+		versionedModules:   moduleMap,
+		allModules:         allModules,
+		firstVersion:       firstVersion,
+		lastVersion:        lastVersion,
+		OrderInitGenesis:   modulesStr,
+		OrderExportGenesis: modulesStr,
+		OrderBeginBlockers: modulesStr,
+		OrderEndBlockers:   modulesStr,
```
rootulp, line 80: [tangent] I don't think we need to do this at the same time as this PR but I wonder if all these could use the same order: https://github.com/celestiaorg/celestia-app/issues/3052.

rootulp (commented):
Kinda makes sense to me. My takeaway is this introduces a new versioned module type that is aware of the version range it supports. That seems reasonable. The logic around running migrations also seems sound so no big blocking feedback

cmwaters:
>  My takeaway is this introduces a new versioned module type that is aware of the version range it supports

Yeah it's basically that and it multiplexes based on the version it expects to be on

app/module/module.go
cmwaters: Yeah I was merely copying across from the SDK

app/module/configurator.go
cmwaters: Correct. Yes we started on v1 and there are assumptions that v0 doesn't exist and implies that the version has forgotten to be set

app/module/configurator.go
cmwaters: Yeah sorry, I just patched this up quickly without correctly implementing it

app/module/configurator.go
rootulp: thanks! 

[tangent] Is there already an assertion somewhere that fails on chain start up if a chain is initialized with v0? Asking because I haven't seen an assertion like that but I expect it exists _somewhere_.

cmwaters: Yes I believe I've added it. I'll double check

app/app.go
cmwaters: Changed it

app/module/module.go
```diff
@@ -0,0 +1,353 @@
...
+		} else if !currentModuleExists && nextModuleExists {
+			ctx.Logger().Info(fmt.Sprintf("adding a new module: %s", moduleName))
+			moduleValUpdates := nextModule.InitGenesis(ctx, c.cdc, nextModule.DefaultGenesis(c.cdc))
+			// The module manager assumes only one module will update the
+			// validator set, and it can't be a new module.
+			if len(moduleValUpdates) > 0 {
+				return sdkerrors.ErrLogic.Wrap("validator InitGenesis update is already set by another module")
+			}
+		}
+		// TODO: handle the case where a module is no longer supported (i.e. removed from the state machine)
```
evan-forbes, line 245: > // TODO: handle the case where a module is no longer supported (i.e. removed from the state machine)

Moving this todo to an issue

evan-forbes (dismissed):
this LGTM

the mocks are great essential unit tests, and this might be planned and I just missed it, but it seems v2 blocking to have some test that will break if things outside of the Manager change. Similar to #3137, if there are other things that are necessary for the complete integration, can we make a point to document them before merging and then discuss them in the next v2 sync? cc @rootulp

app/module/module.go
```diff
@@ -0,0 +1,353 @@
...
+type VersionedModule struct {
+	module                 sdkmodule.AppModule
+	fromVersion, toVersion uint64
+}
+
+func NewVersionedModule(module sdkmodule.AppModule, fromVersion, toVersion uint64) VersionedModule {
+	return VersionedModule{
+		module:      module,
+		fromVersion: fromVersion,
+		toVersion:   toVersion,
```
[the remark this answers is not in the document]
cmwaters: It's meant as a range. The same module could be used over several app versions

evan-forbes (approved):

app/app.go
```diff
@@ -519,35 +524,52 @@ func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.R
...
+
 // InitChainer application update at chain initialization
 func (app *App) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
 	var genesisState GenesisState
 	if err := tmjson.Unmarshal(req.AppStateBytes, &genesisState); err != nil {
 		panic(err)
 	}
-	return app.mm.InitGenesis(ctx, app.appCodec, genesisState)
+	// genesis must always contain the consensus params. The validator set howerver is derived from the
+	// initial genesis state
```
rootulp, line 555: I don't understand this comment, can it be rephrased? The consensus params and the validator set seem orthogonal. 

Also the validator set doesn't seem applicable to the following three lines so if it's not relevant in this context, proposal to delete that portion of the comment.

rootulp (approved):

app/module/module.go
```diff
@@ -0,0 +1,353 @@
...
+	modulesStr := make([]string, 0, len(modules))
+	firstVersion, lastVersion := uint64(0), uint64(0)
+	for idx, module := range modules {
+		if module.fromVersion == 0 {
+			return nil, sdkerrors.ErrInvalidVersion.Wrapf("v0 is not a valid version for module %s", module.module.Name())
+		}
+		if module.fromVersion > module.toVersion {
+			return nil, sdkerrors.ErrLogic.Wrapf("fromVersion can not be greater than toVersion for module %s", module.module.Name())
+		}
+		for version := module.fromVersion; version <= module.toVersion; version++ {
```
staheri14, line 56: Mind explaining in the code what `fromVersion` and `toVersion` indicates for each module?

app/module/module.go
```diff
@@ -0,0 +1,353 @@
...
+		toVersion:   toVersion,
+	}
+}
+
+// NewManager creates a new Manager object
+func NewManager(modules ...VersionedModule) (*Manager, error) {
+	moduleMap := make(map[uint64]map[string]sdkmodule.AppModule)
+	allModules := make([]sdkmodule.AppModule, len(modules))
+	modulesStr := make([]string, 0, len(modules))
+	firstVersion, lastVersion := uint64(0), uint64(0)
```
staheri14, line 48: A description for what firstVersion and lastVersion hold would be helpful.

app/module/module.go
```diff
@@ -0,0 +1,353 @@
...
+			return nil, sdkerrors.ErrInvalidVersion.Wrapf("v0 is not a valid version for module %s", module.module.Name())
+		}
+		if module.fromVersion > module.toVersion {
+			return nil, sdkerrors.ErrLogic.Wrapf("fromVersion can not be greater than toVersion for module %s", module.module.Name())
+		}
+		for version := module.fromVersion; version <= module.toVersion; version++ {
+			if moduleMap[version] == nil {
+				moduleMap[version] = make(map[string]sdkmodule.AppModule)
+			}
+			moduleMap[version][module.module.Name()] = module.module
```
staheri14, line 60: [question] Why the `fromVersion` is taken into account but not the `toVersion` (I guess it is related to my earlier question about the usage of fromVersion and toVersion).

app/module/module.go
```diff
@@ -0,0 +1,353 @@
...
+			if err != nil {
+				return err
+			}
+		} else if !currentModuleExists && nextModuleExists {
+			ctx.Logger().Info(fmt.Sprintf("adding a new module: %s", moduleName))
+			moduleValUpdates := nextModule.InitGenesis(ctx, c.cdc, nextModule.DefaultGenesis(c.cdc))
+			// The module manager assumes only one module will update the
+			// validator set, and it can't be a new module.
+			if len(moduleValUpdates) > 0 {
+				return sdkerrors.ErrLogic.Wrap("validator InitGenesis update is already set by another module")
```
staheri14, line 242: [Question] In the `EndBlock` , under a similar circumstances, where there are multiple validator updates, the method panics, but here it returns an error. Just wondering whether the two should behave similarly or not.

staheri14 (commented):
LGTM, added some suggestions and questions, will defer to other reviewers for the final approval.

app/app.go
cmwaters: I could perhaps move this comment to the function signature. I think it's important since normally there are multiple ways of setting the consensus params and validator set that we clarify which one we use

app/module/module.go
cmwaters: This section was taken from the original module manager in the cosmos sdk

app/module/module.go
cmwaters: I don't quite understand this question. All version `fromVersion` to `toVersion` are added to the `moduleMap`

rootulp (dismissed):

evan-forbes (dismissed):

app/module/module.go
staheri14: What I meant to say is that the `version` variable contains the value obtained from `module.fromModule`. In the map, however, we only store the module corresponding to `fromVersion` and not `toVersion`. This leads to my confusion: shouldn't the same process apply to all versions ranging from `fromVersion` to `toVersion`? This mostly relates to my earlier question regarding the purpose of `toVersion`.

evan-forbes (approved):

rootulp (approved):

PR #3122, opened 2024-02-21
refactor: first block prepare proposal test

cmwaters:
There has been some fakiness in the  block production tests. We don't need a full integration test.


ninabarbakadze (approved):
lgtm

rootulp (approved):
LGTM. The current PR title works. FYI we can also use the `refactor:` commit prefix 

Ref: https://www.conventionalcommits.org/en/v1.0.0/

PR #3141, opened 2024-02-29
pkg/proof: reject negative indices in QueryTxInclusionProof

### LINKED ISSUE #3140: pkg/proof: QueryTxInclusionProof should validate the range of index to avoid negative values being cast to uint64 and overflowing
If we look at this code freshly parsed from path[0] https://github.com/celestiaorg/celestia-app/blob/2754df43e6393555c88bf557b1172208aded9f32/pkg/proof/querier.go#L33

the next place that it is used is https://github.com/celestiaorg/celestia-app/blob/2754df43e6393555c88bf557b1172208aded9f32/pkg/proof/querier.go#L50

That code is directly used by the ABCI client and thus susceptible to marauders who could pass in negative values that then overflow when cast to uint64(-1) and could take down a client with that large indices being added to the ShareInclusionProof

## Fix
Please perform range checks before using it
```diff
+if index < 0 {
+    return nil, fmt.Errorf("path[0] element produced a negative value
+}
```

/cc @elias-orijtech @liamsi @rootulp

odeke-em:
This change ensures that any negative indices are outright rejected to avoid any future overflows due to the wrapping around of uint64(v) and control from and outside caller.

Fixes #3140.


rootulp (approved):
Thanks for the PR!

cc: @evan-forbes should we backport to v1.x and cut a patch release?

cmwaters (approved):
Thanks for catching and fixing this. Minor nit: can we either move the test to proof_test.go or find a more apt name than regression_test.go 

@rootulp yes this should be backported

evan-forbes (approved):

rootulp (approved):

rootulp:
Lot of CI checks failing

rootulp (approved):
LGTM if CI passes

rootulp (approved):

cmwaters (approved):
Thanks @odeke-em

pkg/proof/proof_test.go
```diff
@@ -256,3 +259,21 @@ func TestAllSharesInclusionProof(t *testing.T) {
 	require.NoError(t, err)
 	assert.NoError(t, proof.Validate(dataRoot))
 }
+
+// Ensure that we reject negative index values and avoid overflows.
+// https://github.com/celestiaorg/celestia-app/issues/3140
+func TestQueryTxInclusionProofRejectsNegativeValues(t *testing.T) {
```
rootulp, line 265: On re-review, I failed to catch that this test function could've been added to a new file `querier_test.go` considering it's testing a function defined in `querier.go`.

odeke-em: I initially had it in regression_test.go then go asked to move it here

odeke-em: image

rootulp: Yea IMO that comment is valid. It comes from Callum not Go. `regression_test.go` doesn't make sense to me because there is no source code file called `regression.go`. The convention followed in this code base is that test files suffix the source file with `_test.go`. For example:

- Implementation in: `querier.go`
- Tests in: `querier_test.go`

Issue #2117, opened 2023-07-18
docker: Prettify comment sections in Dockerfile and Dockerfile.ephemeral.


Bidon15:
It's a copy-cat from Dockerfile and Dockerfile.ephemeral. 
I guess a follow-up PR will fix other Dockerfiles. 

Editing this one, tho

_Originally posted by @Bidon15 in https://github.com/celestiaorg/celestia-app/pull/2052#discussion_r1266621067_

```[tasklist]
### Tasks
- [ ] Edit comments like `# stage 1 .....` and `# stage 2` in DockerFile and DockerFile ephemeral to the style like in tx simulator
```

rootulp:
There is no longer Dockerfile.ephemeral but there are two Dockerfiles that could use improved comments:
1. Dockerfile
2. Dockerfile_txsim

The Dockerfile for testground seem really helpful and could be used for inspiration.

rootulp:
Honestly going to ignore the txsim Dockerfile b/c it's not as important as the other one.

PR #3145, opened 2024-03-04
docs: Dockerfile comments

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2117

I can think of a few follow-ups after reviewing this Dockerfile:

1. Can we use a different username than `celestia`? Perhaps `celestia-app` to avoid any user name collision with celestia-node.
2. Does `$BUILDPLATFORM` actually get populated?
3. Can we get rid of the hadolint ignore by pinning versions for all the dependencies?
4. Bump the Alpine version for the BUILDER image
5. `make hadolint` only runs on Dockerfile and not the other Dockerfiles in this repo

cc: @Bidon15


evan-forbes (approved):

evan-forbes (approved):

ninabarbakadze (approved):
lgtm

PR #3149, opened 2024-03-04
ci: use Go version from go.mod

rootulp:
This should make it slightly easier to bump Go versions in the future because there will be fewer places to update. 
 
Thanks for the tip @ramin! See https://github.com/celestiaorg/celestia-core/pull/1242/files#r1511342490


evan-forbes (approved):

ramin (approved):
genius

Issue #3086, opened 2024-02-07
TestMaxBlockSize flake


rootulp:
## Problem

```
--- FAIL: TestIntegrationTestSuite (21.86s)
    --- FAIL: TestIntegrationTestSuite/TestMaxBlockSize (13.12s)
        integration_test.go:[27](https://github.com/celestiaorg/celestia-app/actions/runs/7817375365/job/21325106901?pr=3085#step:4:28)1: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/integration_test.go:271
            	            				/home/runner/work/celestia-app/celestia-app/app/test/integration_test.go:174
            	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.8.4/suite/suite.go:112
            	Error:      	Not equal: 
            	            	expected: []byte{0xe8, 0xf8, 0x5b, 0xba, 0xe5, 0x85, 0xc6, 0x19, 0xf3, 0xbd, 0x98, 0x4, 0x5e, 0xe5, 0x4d, 0x63, 0xc1, 0xd, 0x77, 0x8d, 0x22, 0x8a, 0x33, 0xea, 0xc6, 0x60, 0x[28](https://github.com/celestiaorg/celestia-app/actions/runs/7817375365/job/21325106901?pr=3085#step:4:29), 0x78, 0x82, 0x28, 0xbf, 0xbd}
            	            	actual  : []byte{0xd5, 0x5c, 0xaf, 0xbb, 0x44, 0x65, 0x4e, 0xc1, 0xb5, 0x7e, 0xf9, 0x43, 0x83, 0x97, 0x9e, 0x7c, 0x[31](https://github.com/celestiaorg/celestia-app/actions/runs/7817375365/job/21325106901?pr=3085#step:4:32), 0xc, 0xd, 0xcc, 0x92, 0x75, 0x7c, 0xde, 0xfd, 0xdd, 0x9b, 0xa5, 0x53, 0x6b, 0xc4, 0x5}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,4 +1,4 @@
            	            	 ([]uint8) (len=[32](https://github.com/celestiaorg/celestia-app/actions/runs/7817375365/job/21325106901?pr=3085#step:4:33)) {
[...]

PR #3154, opened 2024-03-05
fix!: TestMaxBlockSize flake

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3086


rootulp:
Closing b/c this PR assumed we were going to merge https://github.com/celestiaorg/go-square/pull/48 into go-square but we found a better approach with https://github.com/celestiaorg/go-square/pull/49 so celestia-app should just need to bump the go-square dep after we cut a release there.

Issue #3134, opened 2024-02-27
Make the encoding require the app version


evan-forbes:
The encoding could differ between app versions. This could result in non-determinism. The safest thing to do imo would be to require the app version when creating an encoding config, and we create a new encoding config each height. This way we ensure that there encoding is done deterministically across heights and nodes. There might be other mechanisms for forcing this.

cmwaters:
Another option is to have the encoding config simply support all versions

evan-forbes:
> Another option is to have the encoding config simply support all versions

does the user need to specify which version of the encoding is needed at a given point?

cmwaters:
I guess what we are trying to prevent is someone submitting a transaction that is not yet or no longer supported right?

Or is there any other use cases you had in mind?

evan-forbes:
the main thing is trying to avoid non-determinism when upgrading since being able to decode or not decode a tx could result in that if we didn't get everything perfect. I was worried about differences in the antehandlers during process and even finalize block. Like, if someone submits a new tx type, will that get passed to ante handlers on nodes that have been upgraded where nodes that have not it would not get passed since it couldn't be decoded.

by requiring the app version, we ensure that this doesn't occur. We could also ensure that all old tx types are treated as unsupported, instead of quasi-supported

cmwaters:
For the antehandler, I was planning on trying to decode it and if it was successful then checking that the msg type was supported by the current version

evan-forbes:
I could see that working as well, can we document that somewhere if its not already? this is marked as `needs-discussion` with the intention to disucss this during the next v2 sync so we can discuss further there. 

while requiring the app version is annoying, imo I think its annoying in a good way similar to a strict type system. Especially if it avoids a bug that requires manually digging through protobuf bytes

evan-forbes:
we don't need this because of #3162

PR #3162, opened 2024-03-06
feat!; prototype for gatekeeping messages based on their version

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/issues/3134

This PR solves the problem of ensuring the messages belonging to modules not part of the current app version aren't executed.

It does this in two ways:
- Introducing an antehandler decorator to be predominantly used in CheckTx to immediately reject transactions giving users a useful error message (instead of just failing to execute)
- Introduces a `CircuitBreaker` implementation to the `MsgServiceRouter` which prevents the execution of messages not belonging to the current app version. We need this because another module may call a message that is not current supported (think a governance proposal)

I had several complications with the wiring of this given the structure of the SDK and tried a few different variants - this one I think being the better.

It uses the configurator which is reponsible for registering services to read all the methods a modules grpc Server supports and extracting out the message names and mapping them to one or more versions that they are supported for.

Note: this is a prototype and is not yet complete but is open to feedback. In order to finish this I will need to write tests and more comprehensive documentation.

There is currently an unresolved problem in that prepare proposal's context does not have access to the app version


app/app.go
```diff
@@ -367,31 +373,88 @@ func New(
...
+		},
+		{
+			Module:      blobstream.NewAppModule(appCodec, app.BlobstreamKeeper),
+			FromVersion: v1, ToVersion: v2,
+		},
+		{
+			Module:      upgrade.NewAppModule(app.UpgradeKeeper),
+			FromVersion: v2, ToVersion: v2,
+		},
+	})
```
cmwaters, line 457: I made this change predominantly for readability reasons but happy to revert if people think this does not help

app/module/configurator.go
```diff
@@ -91,3 +125,35 @@ func (c configurator) runModuleMigrations(ctx sdk.Context, moduleName string, fr
...
+		}, noopInterceptor)
+	}
+	s.addMessages(msgs)
+	// call the underlying msg server to actually register the grpc server
+	s.msgServer.RegisterService(sd, v)
+}
+
+func noopInterceptor(_ context.Context, _ interface{}, _ *grpc.UnaryServerInfo, _ grpc.UnaryHandler) (interface{}, error) {
+	return nil, nil
+}
```
cmwaters, line 159: This is where the fun happens (I basically insert a wrapper to read off the grpc service descriptor before passing it on to the underlying service

app/app.go
rootulp: Looks fine to me

app/module/configurator.go
rootulp: 🤯

app/module/configurator.go
```diff
@@ -91,3 +125,35 @@ func (c configurator) runModuleMigrations(ctx sdk.Context, moduleName string, fr
...
+// This allows us to create a map of which messages are accepted across which
+// versions
+type serverWrapper struct {
+	addMessages func(msgs []string)
+	msgServer   pbgrpc.Server
+}
+
+func (s *serverWrapper) RegisterService(sd *grpc.ServiceDesc, v interface{}) {
+	msgs := make([]string, len(sd.Methods))
+	for idx, method := range sd.Methods {
```
rootulp, line 141: are messages always 1:1 with methods?

rootulp (commented):
Seems clever! How does this handle the case where the new version of a module removes a message type? I infer the message will no longer appear on the module service description methods and so it won't get registered in the map for a particular version. Is that right?

Wait on re-read it looks like a module defines a range of versions it supports so addMessages can't selectively remove a message for a module at a particular version.

Maybe I'm asking about a scenario that isn't possible. But if it is possible (and unlikely) I think we need a plan for it.

app/module/configurator.go
```diff
@@ -1,49 +1,83 @@
...
+	}
+}
+
+func (c VersionedConfigurator) addMessages(msgs []string) {
+	for version := c.fromVersion; version <= c.toVersion; version++ {
+		if _, exists := c.acceptedMessages[version]; !exists {
+			c.acceptedMessages[version] = map[string]struct{}{}
+		}
+		for _, msg := range msgs {
+			c.acceptedMessages[version][msg] = struct{}{}
```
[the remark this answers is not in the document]
cmwaters: I actually don't know. It's how I've seen other engineers do it so I just copied it here but I never really benchmarked

app/module/configurator.go
cmwaters: Yup, each method only ever has one message which is the request argument

cmwaters:
> I infer the message will no longer appear on the module service description methods and so it won't get registered in the map for a particular version. Is that right?

Yeah that's correct. Say we remove staking.MsgDelegate. It will be registered for v1 and not part of v2

app/module/configurator.go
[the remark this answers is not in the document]
rootulp: Turns out `struct{}` is more efficient: https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b

Ppl argue back and forth on [Reddit](https://www.reddit.com/r/golang/comments/gnge5m/stop_using_mapstringstruct_for_existence_checks/) about it. My takeaways is `struct{}` is a bit worse syntax wise but better because the map can only be used for existence checks. If a `bool` is used, at some point later a developer could interpret `bool`s in the map for something else besides just existence.

app/ante/msg_gatekeeper.go
```diff
@@ -0,0 +1,55 @@
...
+	return next(ctx, tx, simulate)
+}
+
+func (mgk MsgVersioningGateKeeper) IsAllowed(ctx context.Context, msgName string) (bool, error) {
+	appVersion := sdk.UnwrapSDKContext(ctx).BlockHeader().Version.App
+	acceptedMsgs, exists := mgk.acceptedMsgs[appVersion]
+	if !exists {
+		return false, sdkerrors.ErrNotSupported.Wrapf("app version %d is not supported", appVersion)
+	}
+	_, exists = acceptedMsgs[msgName]
```
evan-forbes, line 50: What is the worst that can happen if a wannabe attacker crafts a proto msg with a different name to get past this?

or perhaps a better question is, could an attacker sneak a tx through that would fail differently between two different versions? :thinking:

app/ante/msg_gatekeeper.go
```diff
@@ -0,0 +1,55 @@
...
+			return ctx, sdkerrors.ErrNotSupported.Wrapf("transaction type %s is not supported in version %d", msgTypeURL, ctx.BlockHeader().Version.App)
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func (mgk MsgVersioningGateKeeper) IsAllowed(ctx context.Context, msgName string) (bool, error) {
+	appVersion := sdk.UnwrapSDKContext(ctx).BlockHeader().Version.App
+	acceptedMsgs, exists := mgk.acceptedMsgs[appVersion]
```
evan-forbes, line 46: is this making the assumption that app version has been checked to be supported before hand? I think this is the case, and don't think its an issue, but just want to confirm

app/app.go
```diff
@@ -367,31 +373,88 @@ func New(
...
+			Module:      feegrantmodule.NewAppModule(appCodec, app.AccountKeeper, app.BankKeeper, app.FeeGrantKeeper, app.interfaceRegistry),
+			FromVersion: v1, ToVersion: v2,
+		},
+		{
+			Module:      crisis.NewAppModule(&app.CrisisKeeper, skipGenesisInvariants),
+			FromVersion: v1, ToVersion: v2,
+		},
+		{
+			Module:      gov.NewAppModule(appCodec, app.GovKeeper, app.AccountKeeper, app.BankKeeper),
+			FromVersion: v1, ToVersion: v2,
```
evan-forbes, line 407: I like the formatting change a lot

unrelated to this PR, does switching the `ToVersion` here to 0 allow for any future version to work? if so, can we do that fix the testground issues (app version == to some large number to have large block sizes)?

evan-forbes (commented):
overall makes a lot of sense! still chewing on a few questions that I listed below

app/ante/msg_gatekeeper.go
cmwaters: If say an attacker uses bank.v1.MsgSend as the msg url, then the proto registry will try unmarshal the `Value` bytes as that type. If it fails to unmarshal it because the fields are different then it would simply reject the message. If the fields are also the same then the message is basically treated as a MsgSend and the normal rules would apply

app/ante/msg_gatekeeper.go
cmwaters: Yes that assumption is made but should be caught in the block before if we are going to upgrade to a version that the node doesn't currently support

app/app.go
cmwaters: > does switching the ToVersion here to 0 allow for any future version to work?

No everything currently must be explicit (and continuous). I think it's better if we have a different mechanism for testing the large block sizes - we could disconnect them from the appversion by simply having a mapping that is introduced in the app constructor

cmwaters:
Blocked on https://github.com/celestiaorg/cosmos-sdk/pull/376

rootulp (dismissed):

app/app.go
```diff
@@ -521,26 +593,37 @@ func (app *App) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.R
...
 
-func (app *App) Upgrade(ctx sdk.Context, version uint64) error {
-	app.SetAppVersion(ctx, version)
-	return app.mm.RunMigrations(ctx, app.configurator, app.AppVersion(ctx), version)
+func (app *App) Upgrade(ctx sdk.Context, fromVersion, toVersion uint64) error {
+	if err := app.mm.RunMigrations(ctx, app.configurator, fromVersion, toVersion); err != nil {
+		return err
+	}
+	if toVersion == v2 {
+		// we need to set the app version in the param store for the first time
```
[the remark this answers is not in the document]
cmwaters: The crux was that since we pass the sdk context everywhere, we should use that to get the app version. The app version should technically be present in the context because it includes the header that has the app version. The problem however was that in most cases when we create the context we only populate parts of the header (the height for example). So I needed to make several changes in `PrepareProposal`, `ProcessProposal`, `InitGenesis` etc. that we actually populated the app version

app/module/module.go
```diff
@@ -18,75 +18,85 @@ import (
...
 	// versions the state machine supports
 	firstVersion, lastVersion := uint64(0), uint64(0)
 	for idx, module := range modules {
-		if module.fromVersion == 0 {
-			return nil, sdkerrors.ErrInvalidVersion.Wrapf("v0 is not a valid version for module %s", module.module.Name())
+		name := module.Module.Name()
+		moduleVersion := module.Module.ConsensusVersion()
+		if module.FromVersion == 0 {
+			return nil, sdkerrors.ErrInvalidVersion.Wrapf("v0 is not a valid version for module %s", module.Module.Name())
 		}
```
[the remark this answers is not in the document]
cmwaters: I have the check below that asserts that FromVersion cannot be greater than ToVersion which would catch this scenario

PR #3166, opened 2024-03-07
feat(scripts): mainnet.sh

rootulp:
Totally optional PR to add a `mainnet.sh` script that can be used to start a consensus full node on mainnet. I'm using this to try to sync from scratch with a celestia-appd binary built from `main`.

Example script output

```
root@rootul-mainnet-consensus:~/celestia-app# ./scripts/mainnet.sh
celestia-app home: /root/.celestia-app
celestia-app version: 1.0.0-rc0-564-g82703272

Are you sure you want to delete: /root/.celestia-app? [y/n] y
Deleting /root/.celestia-app...
Initializing config files...
Settings seeds in config.toml...
Downloading genesis file...
Starting celestia-appd in the background and piping logs to mainnet.log
You can check the node's status via: celestia-appd status
```


jcstein:
Notes from @rootulp on how to view logs:

View log file:
```
cat mainnet.log
```

View actively running process which is piping into a log file:
```
tail -f mainnet.log
```

To stop the process:
```
ps aux | grep celestia-app
kill <PID>

# if that doesnt' work
kill -9 <PID>
``

evan-forbes (approved):
I don't see the harm in merging since the api's are pretty stable and shouldn't break too often. ofc if after a period of time we stop using this we should prune it (we should probably do this for other scripts as well) 

would also be useful I think to dump state (and optionally the block) upon hitting an error. Then we could debug each error independently quickly

ninabarbakadze (approved):

Issue #3165, opened 2024-03-06
specs: all non-PFB transactions must be ordered before PFB transactions


rootulp:
## Context

- https://celestiaorg.github.io/celestia-app/specs/block_validity_rules.html
- https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#layout-rationale

## Problem

The current specs don't describe a block validity rule that is verified by [this test](https://github.com/celestiaorg/celestia-app/blob/44ca7316f8ea604fe45dbefe1809e09cb4c06f4f/app/test/process_proposal_test.go#L278-L285). My best effort at describing the rule:

In block data, all non-PFB transactions must be ordered before PFB transactions.

## Proposal

Add this block validity rule to the specs. In my opinion, we should add a third section to https://celestiaorg.github.io/celestia-app/specs/block_validity_rules.html#validity-rules for **Block validity rules** where we add this rule.

PR #3168, opened 2024-03-07
specs: block rules

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3165

While I was working on this, I noticed that the transaction validity rules section was incorrect because it failed to differentiate the rules for app version v1 and v2. Can extract that to a stand-alone PR if reviewers prefer that.


staheri14 (approved):
Nice 👍

evan-forbes (approved):

PR #3177, opened 2024-03-14
fix: throw an error if incorrect ledger is used

rootulp:
Upgrade to https://github.com/celestiaorg/cosmos-sdk/releases/tag/v1.20.2-sdk-v0.46.16 on the v1.x branch to pick up the Ledger fix in https://github.com/celestiaorg/cosmos-sdk/pull/374


rootulp:
I think we accidentally made a cosmos-sdk patch release with a breaking change @cmwaters

rootulp:
I'm taking inspiration from https://github.com/celestiaorg/celestia-app/pull/3162 on all the other places app version needs to be plumbed because it looks like this PR broke a few tests

cmd/celestia-appd/cmd/root.go
```diff
@@ -246,9 +245,6 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts se
 		baseapp.SetTrace(cast.ToBool(appOpts.Get(server.FlagTrace))),
 		baseapp.SetIndexEvents(cast.ToStringSlice(appOpts.Get(server.FlagIndexEvents))),
 		baseapp.SetSnapshot(snapshotStore, snapshottypes.NewSnapshotOptions(cast.ToUint64(appOpts.Get(server.FlagStateSyncSnapshotInterval)), cast.ToUint32(appOpts.Get(server.FlagStateSyncSnapshotKeepRecent)))),
-		func(b *baseapp.BaseApp) {
-			b.SetAppVersion(sdk.Context{}, appconsts.LatestVersion)
-		},
```
rootulp, line 251: @cmwaters I'd appreciate your review here. These lines are missing from the main branch so opted to delete them here which resolved some test failures. Was that correct? I don't understand the flow of setting app version before / after the cosmos SDK bump

cmwaters: So in v1, the app version is hardcoded through this exact mechanism (i.e. it's manually set in the constructor). In v2 and onwards, it is read from the genesis file. It could be possible to make the same modifications to celestia-core for v1 that would allow v1.x to also read and set the app version from genesis. However, it's probably easier if you simply keep these lines around which will hardcode the version to 1

rootulp: It's not actually possible to keep these lines here because v1.x panics on this line. On v1.x, `app.ParamStore` is not defined before hitting this line.

PR #3180, opened 2024-03-15
docs: ADR022 - multi versioned state machine

cmwaters:


docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
+
+- 2024/03/15: Initial draft (@cmwaters)
+
+## Status
+
+Implemented
+
+## Context
+
+The Celestia application required a modification from the existing Cosmos SDK to support multiple versions of the state machine simultaneously. This capability is crucial for single binary syncs and ensuring a smooth transitions during upgrades, allowing nodes to upgrade independently and then switch to the next state machine without any downtime to the network. This is important for a network which is depended on by a large number of rollups and users.
```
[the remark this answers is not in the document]
rootulp: CodeRabbit left useful suggestions 👍

docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Decision
+
+We decided to implement a multi-versioned state machine by making significant modifications to the module manager, the configurator, and introducing a new ante handler. These changes enable the application to maintain and execute different versions of the state machine based on the app version associated with each transaction and prevents the execution of transactions that are not part of the current state machine.
```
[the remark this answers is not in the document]
rootulp: > based on the app version associated with each transaction

is the app version associated with a transaction or with a block? I think the latter.

docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
... application (as 1). An application with multiple versions needs to be able to persist them to disk and load them in the case the node crashes or is shutdown. The `Info` abci call must pass the app version to Comet for Comet to generate headers with it. The decision was made to always require the app version to be set in the consensus params of Genesis (prior `ConsensusParams` wasn't mandatory). Upon first use, `Info` will return 0 (as no app version has been persisted) and then take the app version passed in `InitChain` to initialize the app version. It is persisted both to disk in the `ParamStore` and in memory in the `BaseApp` struct. Upon later startups, `Info` will load the app version from disk, set it in memory and also return it to Comet. The `AppVersion` is only later adjusted when `Upgrade` is called on the `Application`. Upgrading is no longer a module but is more a native part of the application. Modules register migrations and `Upgrade`, which is executed in `EndBlock` will trigger those migrations and then set the new version. Upgrading needs to happen in `EndBlock`, such that `PrepareProposal` for the following height already is run with the logic of the new version.
```
rootulp, line 25: `abci` => `ABCI`

```suggestion
The app version was previously never persisted to disk, but was simply hardcoded into the application (as 1). An application with multiple versions needs to be able to persist them to disk and load them in the case the node crashes or is shutdown. The `Info` ABCI call must pass the app version to Comet for Comet to generate headers with it. The decision was made to always require the app version to be set in the consensus params of Genesis (prior `ConsensusParams` wasn't mandatory). Upon first use, `Info` will return 0 (as no app version has been persisted) and then take the app version passed in `InitChain` to initialize the app version. It is persisted both to disk in the `ParamStore` and in memory in the `BaseApp` struct. Upon later startups, `Info` will load the app version from disk, set it in memory and also return it to Comet. The `AppVersion` is only later adjusted when `Upgrade` is called on the `Application`. Upgrading is no longer a module but is more a native part of the application. Modules register migrations and `Upgrade`, which is executed in `EndBlock` will trigger those migrations and then set the new version. Upgrading needs to happen in `EndBlock`, such that `PrepareProposal` for the following height already is run with the logic of the new version.
```

docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
+
+### SDK Context
+
+The `sdk.Context` type is passed to almost every component in the application and is thus ideal for storing and accessing the current app version. It already contains the app version in the block header. However this is not always populated. Thus `InitChain` and `PrepareProposal` were modified to include the app version as either passed from the genesis file or loaded from the application itself.
+
+### Module Manager Changes
+
+The module manager was modified to contain a mapping of modules to their version.
```
rootulp, line 33: Should this be something like:

```suggestion
The module manager was modified to contain a mapping from app version -> module name -> module.
```

docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
+
+```go
+type Manager struct {
+    versionedModules map[uint64]map[string]sdkmodule.AppModule
+    ...
+}
+```
+
+Each module has a `Name` and a `ConsensusVersion` that marks it unique. No two modules can have the same `Name` within a single app version. When constructing the application, we specify the contiguous range (inclusive) of versions that the module is part of:
```
rootulp, line 42:
```suggestion
Each module has a `Name` and a `ConsensusVersion` that marks it unique. No two modules can have the same `Name` within a single app version. When constructing the application, we specify the contiguous range (inclusive) of app versions that the module is part of:
```

docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
+        FromVersion: v1, ToVersion: v2,
+    },
+    {
+        Module:      upgrade.NewAppModule(app.UpgradeKeeper),
+        FromVersion: v2, ToVersion: v2,
+    },
+})
+```
+
+`InitGenesis`, `BeginBlock` and `EndBlock` will only be called for the modules that belong to the current version of the application.
```
rootulp, line 57: Is `InitGenesis` invoked for subsequent app versions (i.e. after app version 1)?

docs/architecture/adr-022-multi-versioned-state-machine.md
```diff
@@ -0,0 +1,98 @@
...
+The configurator is an object that is used upon initialisation to register the msg server, query server and migrations. In order to know what `sdk.Msg`s corresponded with which app version, a wrapper was added to the configurator that would scrape all the `sdk.Msg`s from the msg server as they were being registered and add them to a map:
+
+```go
+// acceptedMsgs is a map from appVersion -> msgTypeURL -> struct{}.
+acceptedMessages map[uint64]map[string]struct{}
+```
+
+### Transaction Filtering through the `MsgVersioningGateKeeper`
+
+The messages within a transactions are filtered based on the app version. The map constructed from the configurator is passed to a new AnteHandler which will reject transactions of a different app version in `CheckTx` so that clients get an immediate response. The `MsgVersioningGateKeeper` is also passed to the `MsgServiceRouter` to reject transactions upon execution. The reason for needing both is that another module may try to execute a transaction for a module that is not part of the app version.
```
rootulp, line 72:
```suggestion
The messages within a transaction are filtered based on the app version. The map constructed from the configurator is passed to a new AnteHandler which will reject transactions of a different app version in `CheckTx` so that clients get an immediate response. The `MsgVersioningGateKeeper` is also passed to the `MsgServiceRouter` to reject transactions upon execution. The reason for needing both is that another module may try to execute a transaction for a module that is not part of the app version.
```

rootulp (dismissed):
Overall LGTM, thanks for writing this up! Only non-blocking questions to clarify.

docs/architecture/adr-022-multi-versioned-state-machine.md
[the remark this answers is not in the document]
cmwaters: There is a mapping between transaction type and what versions it can be executed during. If a block is a version and includes a transaction not belonging to that version it is considered invalid

docs/architecture/adr-022-multi-versioned-state-machine.md
cmwaters: InitGenesis at the module manager level is called once for all modules of that version. However when we upgrade, if we have a new module (for example minfee) we call InitGenesis on specifically the module with the DefaultGenesis of the module

docs/architecture/adr-022-multi-versioned-state-machine.md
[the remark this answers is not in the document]
rootulp: Oh I see what this was saying. On re-read looks fine, can resolve.

rootulp (approved):

ninabarbakadze (approved):

PR #3186, opened 2024-03-18
feat: add workdir to dockerfile

smuu:
## Overview

This should be a non-breaking change.

This only changes the work directory when the container is started. This means that if you want to execute a command in a running container, it will be automatically in that directory instead of '/'. This improves maintaining and debugging our environments.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,8 @@ COPY --from=builder /celestia-app/build/celestia-appd /bin/celestia-appd
 COPY --chown=${USER_NAME}:${USER_NAME} docker/entrypoint.sh /opt/entrypoint.sh
 # Set the user to celestia.
 USER ${USER_NAME}
+# Set the working directory to the home directory.
+WORKDIR ${CELESTIA_HOME}
 # Expose ports:
 # 1317 is the default API server port.
 # 9090 is the default GRPC server port.
--- a/docker/Dockerfile_txsim
+++ b/docker/Dockerfile_txsim
@@ -53,6 +53,9 @@ COPY --chown=${USER_NAME}:${USER_NAME} docker/txsim.sh /opt/entrypoint.sh
 
 USER ${USER_NAME}
 
+# Set the working directory to the home directory.
+WORKDIR ${CELESTIA_HOME}
+
 # grpc, rpc, api ports
 EXPOSE 26657 1317 9090
 


Dockerfile
```diff
@@ -57,6 +57,8 @@ COPY --from=builder /celestia-app/build/celestia-appd /bin/celestia-appd
 COPY --chown=${USER_NAME}:${USER_NAME} docker/entrypoint.sh /opt/entrypoint.sh
 # Set the user to celestia.
 USER ${USER_NAME}
+# Set the working directory to the home directory.
+WORKDIR ${CELESTIA_HOME}
```
rootulp, line 61: > This should be a non-breaking change.

Won't this change be breaking to scripts that assume the working directory is `/`? In other words, scripts that were written against the previous Docker container _may_ break.

> If the WORKDIR instruction is not specified in a Dockerfile, the default WORKDIR is the root directory (/).

Ref: https://www.baeldung.com/ops/docker-default-workdir

rootulp (dismissed):
I think this may break scripts that assume the working directory is `/`.

[question] do we want to apply something similar to the two other Dockerfiles in this repo? 
1. celestia-app/docker/Dockerfile_txsim
2. celestia-app/test/testground/Dockerfile

staheri14 (approved):
👍

ninabarbakadze (approved):

evan-forbes (approved):

Dockerfile
smuu: Thanks for bringing this up. I checked out scripts, and we always use absolute paths.

I'm not sure if externals use the container with custom scripts as well.

smuu: Also, @cmwaters uses absolute paths in the e2e tests using knuu

smuu:
> I think this may break scripts that assume the working directory is `/`.
> 
> [question] do we want to apply something similar to the two other Dockerfiles in this repo?
> 
>     1. celestia-app/docker/Dockerfile_txsim
> 
>     2. celestia-app/test/testground/Dockerfile

I added it for the txsim Dockerfile.

I would not add it to the testground Dockerfile, as I don't want to break things as we sunset testground.

rootulp (approved):

evan-forbes (approved):

smuu:
> I think this may break scripts that assume the working directory is `/`.
> 
> [question] do we want to apply something similar to the two other Dockerfiles in this repo?
> 
>     1. celestia-app/docker/Dockerfile_txsim
> 
>     2. celestia-app/test/testground/Dockerfile

I added it for the txsim Dockerfile.

I would not add it to the testground Dockerfile, as I don't want to break things as we sunset testground.

PR #3201, opened 2024-03-20
fix: fixes failing TestE2ESimple by setting the latest app version

### LINKED ISSUE #3010: `make test-e2e` fails
## Context

The nightly workflow is failing and has been for a while https://github.com/celestiaorg/celestia-app/actions/runs/7498578218/job/20413921302

## Problem

```
=== RUN   TestE2ESimple
    simple_test.go:44: Running simple e2e test version 3[8](https://github.com/celestiaorg/celestia-app/actions/runs/7498578218/job/20413921302#step:5:9)02ea[9](https://github.com/celestiaorg/celestia-app/actions/runs/7498578218/job/20413921302#step:5:10)
    simple_test.go:72: 
        	Error Trace:	/home/runner/work/celestia-app/celestia-app/test/e2e/simple_test.go:72
        	Error:      	Not equal: 
        	            	expected: 0x1
        	            	actual  : 0x0
        	Test:       	TestE2ESimple
--- FAIL: TestE2ESimple (302.66s)
=== RUN   TestMinorVersionCompatibility
    upgrade_test.go:31: skipping e2e test: E2E_VERSIONS not set
--- SKIP: TestMinorVersionCompatibility (0.00s)
=== RUN   TestMajorUpgradeToV2
    upgrade_test.go:[19](https://github.com/celestiaorg/celestia-app/actions/runs/7498578218/job/20413921302#step:5:20)2: 
        	Error Trace:	/home/runner/work/celestia-app/celestia-app/test/e2e/upgrade_test.go:192
        	Error:      	Not equal: 
        	            	expected: 0x2
        	            	actual  : 0x1
        	Test:       	TestMajorUpgradeToV2
        	Messages:   	version mismatch after upgrade
```

staheri14:
Addresses the issue with the TestE2ESimple failure reported in #3010


test/e2e/simple_test.go
```diff
@@ -64,12 +63,16 @@ func TestE2ESimple(t *testing.T) {
...
+		testnet.Node(1).AddressRPC())
+	require.NoError(t, err)
+
 	blockchain, err := testnode.ReadBlockchain(context.Background(), testnet.Node(0).AddressRPC())
 	require.NoError(t, err)
 
 	totalTxs := 0
 	for _, block := range blockchain {
-		require.Equal(t, v1.Version, block.Version.App)
+		require.Equal(t, expectedAppVersion, block.Version.App)
```
rootulp, line 75: [question] do we run this e2e test on the v1.x branch? If not, it seems like we could've done this:

```diff
-		require.Equal(t, v1.Version, block.Version.App)
+		require.Equal(t, v2.Version, block.Version.App)
```

but your approach seems more resilient to future app versions so seems preferable to me.

test/util/testnode/read.go
```diff
@@ -39,6 +39,30 @@ func ReadBlockchain(ctx context.Context, rpcAddress string) ([]*types.Block, err
...
+	client, err := http.New(rpcAddress, "/abci_info")
+	if err != nil {
+		return 0, err
+	}
+	resp, err := client.ABCIInfo(ctx)
+	if err != nil {
+		return 0, err
+	}
+	return resp.Response.AppVersion, nil
+}
```
rootulp, line 52: [optional] this doesn't appear used so we could delete it.

```suggestion
```

rootulp (approved):

staheri14:
>  However since we set the app version when constructing it, we don't really have to query it. We could simply pass the app version we want in the constructor when we're creating the genesis here

I agree, we could set it manually without querying. 👍 
Nevertheless, I have a concern: while this works for the nightly tests, as we know the expected app version, this isn't universally applicable to all versions of the celestia-app image utilized in the `TestE2ESimple` test. This could lead to a situation where the image is outdated and doesn't support v2, even though the genesis file specifies app version 2. So, my primary concern is: shouldn't the expected app version in the `TestE2ESimple` be aligned with the version of the image provided to the test?

> To take a step back, I don't think the TestE2ESimple provides much value beyond a starting point for using knuu and to help with my familiarity. I'm leaning towards removing it. 

I support retaining it; it serves as an effective baseline test for assessing the health and progress of the app's latest version and the network. Moreover, as you highlighted, it provides an easy-to-understand test (and a starting point) to write tests using knuu in celestia-app.

cmwaters:
> I support retaining it; it serves as an effective baseline test for assessing the health and progress of the app's latest version and the network

Yes but this is sufficiently covered by the other test. There's nothing new that this test tests

cmwaters:
> This could lead to a situation where the image is outdated and doesn't support v2, even though the genesis file specifies app version 2.

If this were to happen then the node would merely fail upon `InitGenesis` when it realises it doesn't support the app version passed to it via the genesis. 

> So, my primary concern is: shouldn't the expected app version in the TestE2ESimple be aligned with the version of the image provided to the test?

No, remember an image can support several versions. There is no one app version that aligns with the image.

staheri14:
> If this were to happen then the node would merely fail upon InitGenesis when it realises it doesn't support the app version passed to it via the genesis.

This is precisely what I'm saying: when trying out older images, the test will encounter failures right from the start, making it unsuitable for older versions of the celestia app images.

> No, remember an image can support several versions. 

This is correct only for the latest version of the Celestia app that is compatible with older versions. However, older versions of the Celestia app will not support future versions.

> Yes but this is sufficiently covered by the other test. There's nothing new that this test tests

I agree that it is covered by other tests, but the other test scenarios target more aspects than this one, for example, upgradability. This one focuses on simple network progress for a specific version, which makes it easier to reason about due to its smaller scope and serves its own purpose. I don't see any harm in keeping it in the repository; moreover, I see benefits.

cmwaters:
> This is precisely what I'm saying: when trying out older images, the test will encounter failures right from the start, making it unsuitable for older versions of the celestia app images.

I don't see what difference there is between setting the app version in the test and querying it soon after starting. In both cases, if the app version is incorrectly set with respect to the image the test will fail, we will realise why and we will adjust the app version or the image. At least with the former you are being more explicit and deliberate about the app version you are expecting

staheri14:
> I don't see what difference there is between setting the app version in the test and querying it soon after starting. In both cases, if the app version is incorrectly set with respect to the image the test will fail, we will realise why and we will adjust the app version or the image. At least with the former you are being more explicit and deliberate about the app version you are expecting


Perhaps my previous messages weren't clear enough, but I want to clarify that my intention isn't to suggest querying the app version from the node. Instead, I'm posing a different question entirely: how can we determine and set the app version based on the image?

cmwaters:
> I'm posing a different question entirely: how can we determine and set the app version based on the image?

I guess if we were to run `celestia-appd init test` it would produce a default genesis with the latest app version that it supports. But I don't think having an automated way to do this is necessary. The test writer should specify the app version and the image and if the test writers does this in an incompatible way then the test fails.

test/e2e/simple_test.go
```diff
@@ -69,7 +69,9 @@ func TestE2ESimple(t *testing.T) {
 
 	totalTxs := 0
 	for _, block := range blockchain {
-		require.Equal(t, v1.Version, block.Version.App)
+		require.Equal(t, appconsts.LatestVersion,
+			block.Version.App) // the app version in the genesis is set to the
+		// latest version, so we should see this in the blocks
```
rootulp, line 74: the indentation looks off

```suggestion
		// The app version in the genesis is set to the
		// latest version, so we should see this in the blocks.
		require.Equal(t, appconsts.LatestVersion, block.Version.App) 
```

rootulp (approved):

test/e2e/simple_test.go
staheri14: Fixed now [4f67e83](https://github.com/celestiaorg/celestia-app/pull/3201/commits/4f67e83005d592c11871d782cf938bebf081453b)

test/e2e/simple_test.go
```diff
@@ -69,7 +69,10 @@ func TestE2ESimple(t *testing.T) {
 
 	totalTxs := 0
 	for _, block := range blockchain {
-		require.Equal(t, v1.Version, block.Version.App)
+		// the app version in the genesis is set to the latest version,
+		// so we should see this in the blocks
+		require.Equal(t, appconsts.LatestVersion,
+			block.Version.App)
```
rootulp, line 75: indentation still looks off to me. This assertion is easier to read on one line

```suggestion
		require.Equal(t, appconsts.LatestVersion, block.Version.App)
```

rootulp (approved):

rootulp:
This PR seems close-able because https://github.com/celestiaorg/celestia-app/pull/3211 merged.

Ref: https://github.com/celestiaorg/celestia-app/pull/3211#discussion_r1541206098

PR #3214, opened 2024-03-26
test: add two node experiment

cmwaters:
Still a wip


cmwaters:
This is no longer needed (as it's been done another way)

PR #3216, opened 2024-03-27
chore: replace go.work with go.work.example

cmwaters:
As agreed upon, go.work should be used for local development.

This removes it, replacing it with a go.work.example that people can use. It adds the two files `go.work` and `go.work.sum` to gitignore


ninabarbakadze (dismissed):

evan-forbes (approved):

evan-forbes (approved):

rootulp (approved):
Thanks!

Issue #3212, opened 2024-03-26
`TestPriorityByGasPrice` flake


rootulp:
## Problem

```
--- FAIL: TestPriorityTestSuite (11.08s)
    --- FAIL: TestPriorityTestSuite/TestPriorityByGasPrice (0.00s)
        priority_test.go:89: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/priority_test.go:89
            	Error:      	Not equal: 
            	            	expected: 0x0
            	            	actual  : 0xd
            	Test:       	TestPriorityTestSuite/TestPriorityByGasPrice
            	Messages:   	insufficient gas price for the network; insufficient gas price for the network; got: 32 required at least: 160.000000000000000000: insufficient fee
```

## Occurrences
1. https://github.com/celestiaorg/celestia-app/actions/runs/8398308541/job/23002976415?pr=3208#step:4:26

PR #3218, opened 2024-03-27
feat: remove versioned min gas price

cmwaters:
This PR removes the versioned global min gas price (since we use governance to change the price)

This PR also fixes `TestPriorityTestSuite` which could sometimes have a min fee which is less than the default

Closes https://github.com/celestiaorg/celestia-app/issues/3212


rootulp (approved):
Awesome thanks! I'll link in PR description that this closes https://github.com/celestiaorg/celestia-app/issues/3212

evan-forbes (approved):

Issue #2878, opened 2023-11-30
Investigate / enforce that totalVotingPower in the upgrade module never overflows


cmwaters:
_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/2832#discussion_r1409624164_

evan-forbes:
to clarify a bit per sync, this involves writing a test that checks for an overflow in the upgrade module

PR #3222, opened 2024-03-28
refactor(x/upgrade): voting power threshold 

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2878

### DIFF
--- a/x/upgrade/keeper.go
+++ b/x/upgrade/keeper.go
@@ -90,7 +90,10 @@ func (k Keeper) SignalVersion(ctx context.Context, req *types.MsgSignalVersion) 
 // which the application can use as signal to upgrade to that version.
 func (k *Keeper) TryUpgrade(ctx context.Context, _ *types.MsgTryUpgrade) (*types.MsgTryUpgradeResponse, error) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
-	threshold := k.GetVotingPowerThreshold(sdkCtx)
+	threshold, err := k.GetVotingPowerThreshold(sdkCtx)
+	if err != nil {
+		return nil, err
+	}
 	hasQuorum, version := k.TallyVotingPower(sdkCtx, threshold.Int64())
 	if hasQuorum {
 		k.quorumVersion = version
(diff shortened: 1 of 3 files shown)


x/upgrade/keeper.go
```diff
@@ -169,15 +175,25 @@ func (k Keeper) TallyVotingPower(ctx sdk.Context, threshold int64) (bool, uint64
...
+	}
+	mod, err := product.SafeMod(sdkmath.NewInt(thresholdFraction.Denominator))
+	if err != nil {
+		return sdkmath.ZeroInt(), err
+	}
+	if mod.IsZero() {
+		return res, nil
 	}
-	return product.QuoRaw(thresholdFraction.Denominator).AddRaw(1)
+	return res.AddRaw(1), nil
```
rootulp, line 196: This code seems complicated so I'm open to refactoring it if we're okay with dropping the `Fraction` type

rootulp: Found a way to refactor using Cosmos SDK decimal. See 3fdc1d7838338ce74f047c3347619b656b70cf59. Note this results in a small amount of precision loss when dealing with extremely large total voting powers but IMO the precision loss is tolerable in favor of cleaner code. Thoughts @cmwaters ?

x/upgrade/keeper.go
```diff
@@ -16,21 +16,17 @@ var (
 	_ types.MsgServer   = &Keeper{}
 	_ types.QueryServer = Keeper{}
 
-	defaultSignalTheshold = Fraction{Numerator: 5, Denominator: 6}
+	// defaultSignalThreshold is 5/6 or approximately 83.33%
+	defaultSignalThreshold = sdk.NewDec(5).Quo(sdk.NewDec(6))
```
cmwaters, line 20: Dec has 6 decimals of precision right?

cmwaters (approved):
Given that we use 5/6ths as the hard coded threshold, I think it is tolerable that there may be some loss in the precision. I don't think it's critical to security that the threshold is exactly 5/6 so to make this cleaner I would agree with this

x/upgrade/keeper.go
rootulp: Looks like it has more than 6:

```
defaultSignalThreshold (string): 0.833333333333333333
defaultSignalThreshold (float64): 0.8333333333333334
```

Issue #3223, opened 2024-03-29
Improve docs for `x/upgrade`


rootulp:
## Context

Josef at Informal Systems asked questions about the docs for `x/upgrade`.

## Proposal

Revise and improve the docs for those modules.

- [x] Update [this section](https://github.com/celestiaorg/celestia-app/blob/c7bef58d058899de23d2cc9d47403c3898e21f53/x/upgrade/README.md?plain=1#L10-L13) of `x/upgrade` README  because off-chain social consensus is only needed to agree on the upgrade height for v1 -> v2. Upgrades from v2 onwards use an on-chain signaling mechanism + crank message.
- [x] Remove [this line](https://github.com/celestiaorg/celestia-app/blob/c7bef58d058899de23d2cc9d47403c3898e21f53/x/upgrade/README.md?plain=1#L21) because it links to a closed PR. Can it be replaced with a link to a merged PR?
- [x] Add a link to https://github.com/celestiaorg/CIPs/blob/main/cips/cip-10.md in the x/upgrade README

PR #3224, opened 2024-03-29
docs: improve `x/upgrade` docs

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3223

x/upgrade README format inspired by https://docs.cosmos.network/v0.50/build/spec/SPEC_MODULE

## FLUPs
1. Can we move some of the unit tests from tally_test.go to keeper_test.go?
2. Why does ResetTally only delete versions less than the current version? Why not delete all versions in the map?

### DIFF
--- a/x/upgrade/README.md
+++ b/x/upgrade/README.md
@@ -21,22 +21,6 @@ The map from validator address to version is updated when a validator signals fo
 
 See [types/msgs.go](./types/msgs.go) for the message types.
 
-## Begin Block
-
-N/A
-
-## End Block
-
-N/A
-
-## Hooks
-
-N/A
-
-## Events
-
-N/A
-
 ## Client
 
 ### CLI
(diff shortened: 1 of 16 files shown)


x/upgrade/README.md
```diff
@@ -1,23 +1,77 @@
...
+
+N/A
+
+## Future Improvements
+
+N/A
+
+## Tests
+
+N/A
```
cmwaters, line 70: Is it necessary to keep the sections that are N/A?

rootulp: Not strictly necessary in my opinion. Can definitely remove

rootulp: https://github.com/celestiaorg/celestia-app/pull/3224/commits/cbea0ebb070b29c9c04914fc039ac961886fd4e9

cmwaters (approved):

ninabarbakadze (approved):

cmwaters:
> Can we move some of the unit tests from tally_test.go to keeper_test.go?

Yeah we can if you want

> Why does ResetTally only delete versions less than the current version? Why not delete all versions in the map?

I guess the one case is if you've already voted for v3 before upgrading to v2. Then you'd have to revote because it would be deleted. But this is not really an expected use case so I have nothing against deleting all the versions in the map

Issue #2271, opened 2023-08-14
Revert First Block Transaction Check


staheri14:
# Problem
Following issue #1899, we decided to add some checks to the `prepare_proposal`  process to ensure that the first block does not contain any transactions (#2221).
However, the incorporation of this check is just to avoid unexpected errors, hence after mainnet this check becomes unnecessary and this issue is to remove this check (basically revert changes introduced in #2233) after mainnet.

rootulp:
Mainnet launched so should we revisit reverting https://github.com/celestiaorg/celestia-app/pull/2233? I thought of this because https://github.com/celestiaorg/celestia-app/pull/3122

cc: @cmwaters

PR #3225, opened 2024-04-01
revert: last block height == 0 check

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2271

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -47,25 +47,7 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 		app.ParamsKeeper,
 		app.MsgGateKeeper,
 	)
-
-	var txs [][]byte
-	// This if statement verifies whether the preparation of the proposal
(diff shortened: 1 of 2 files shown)


staheri14 (approved):
I am fine with removing it! Though, wouldn't the same check would be required for v2?

evan-forbes (approved):

cmwaters (approved):

cmwaters:
> Though, wouldn't the same check would be required for v2?

We haven't done anything that may handle a first block with transactions but given your analysis that it's highly unlikely to populate the mempool prior to the first block and the fact that there is unlikely to be a chain we start in the future that is important as mainnet (i.e. most of them will be devnets or testnets), I am okay with removing this check

rootulp:
> I am fine with removing it! Though, wouldn't the same check would be required for v2?

Why would it be required for v2? v2 will run on the existing networks (i.e. mainnet, mocha, arabica) and new testnets. The existing networks already have an empty block height at 1 and it's unlikely for new testnets to have transactions in the first block.

Issue #3231, opened 2024-04-01
Add more CODEOWNERS to v1.x branch


rootulp:
## Problem

https://github.com/celestiaorg/celestia-app/pull/3219 has two approvals but can't be merged because the v1.x branch doesn't have @staheri14 or @ninabarbakadze listed as CODEOWNERS

## Proposal

Add @staheri14 and @ninabarbakadze as CODEOWNERS on v1.x branch

rootulp:
Separately, it looks like the v1.x branch file has a CODEOWNERS file with errors

<img width="717" alt="Screenshot 2024-04-01 at 1 59 08 PM" src="https://github.com/celestiaorg/celestia-app/assets/3699047/bb3fe8bc-6e3f-4508-98d3-11720aad41b1">

rootulp:
Closed by #3232

PR #3232, opened 2024-04-01
chore: add Sanaz and Nina as codeowners

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3231

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,11 +8,11 @@
 # directory-level definitions.
 
 # global owners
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):

cmwaters (approved):
So I think the more long term solution is to add @celestiaorg/celestia-core as code owners to all the repos (I've done this already with go-square). That way when we add someone to the github team, they are automatically added as global owners to all our code bases

PR #3239, opened 2024-04-02
fix: missing validator name in SimpleE2ETest

### LINKED ISSUE #3240: Validators fail to be added since their accounts are already part of the testnet genesis keyring
The `SimpleE2ETest` fails with this err  `key not found`, we can resolve it by this [PR](https://github.com/celestiaorg/celestia-app/pull/3239), however, there is still another issue which does not let the test be run successfully.  Below is the full explanation of the root cause:

When creating genesis node, their accounts are by default added to the testnet genesis keyring. see [here](https://github.com/celestiaorg/celestia-app/blob/9440dea585422ddf6d89074037dd906cf78930a7/test/e2e/testnet.go#L48) where the `t.genesis.Keyring()` is passed. 

Then at the next step, they are added as validator to the `testnet.genesis` via `t.genesis.AddValidator(...)` call, see [here](https://github.com/celestiaorg/celestia-app/blob/9440dea585422ddf6d89074037dd906cf78930a7/test/e2e/testnet.go#L52)
During this process, the same account is to be added to the `t.genesis` which makes it fail at this [line](https://github.com/celestiaorg/celestia-app/blob/897b448cb8d2b655743bf1e786f6823775468964/test/util/genesis/genesis.go#L129), there are two reasons for failure
- One in [here](https://github.com/celestiaorg/celestia-app/blob/897b448cb8d2b655743bf1e786f6823775468964/test/util/genesis/genesis.go#L101C16-L101C29 ) when the account name is empty 
- And when we set the name of the account to resolve the above (see this [PR](https://github.com/celestiaorg/celestia-app/pull/3239)), it again fails because the name already exists, in [here](https://github.com/celestiaorg/celestia-app/blob/897b448cb8d2b655743bf1e786f6823775468964/test/util/genesis/genesis.go#L107) 

This suggests the code needs a bit of refactoring to resolve the issue.

staheri14:
Closes #3240

### DIFF
--- a/test/e2e/node.go
+++ b/test/e2e/node.go
@@ -261,6 +261,7 @@ func (n *Node) Start() error {
 func (n *Node) GenesisValidator() genesis.Validator {
 	return genesis.Validator{
 		Account: genesis.Account{
+			Name:          n.Name,
 			PubKey:        n.AccountPubKey,
 			InitialTokens: n.SelfDelegation,
 		},
--- a/test/util/genesis/genesis.go
+++ b/test/util/genesis/genesis.go
@@ -126,9 +126,9 @@ func (g *Genesis) AddValidator(val Validator) error {
 	}
 
 	// Add the validator's genesis account
-	//if err := g.AddAccount(val.Account); err != nil {
+	// if err := g.AddAccount(val.Account); err != nil {
 	//	return err
-	//}
+	// }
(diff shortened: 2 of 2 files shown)


staheri14:
Closing as the full fix has been implemented in https://github.com/celestiaorg/celestia-app/pull/3243

PR #3243, opened 2024-04-02
fix: genesis account creation in e2e tests

cmwaters:
Closes: #3240

This is an attempt at fixing the problem @staheri14 pointed out with the creation of the validator accounts in the genesis file.

### DIFF
--- a/test/e2e/node.go
+++ b/test/e2e/node.go
@@ -7,9 +7,6 @@ import (
 
 	"github.com/celestiaorg/celestia-app/test/util/genesis"
 	"github.com/celestiaorg/knuu/pkg/knuu"
-	"github.com/cosmos/cosmos-sdk/crypto/hd"
-	"github.com/cosmos/cosmos-sdk/crypto/keyring"
-	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
 	"github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/crypto"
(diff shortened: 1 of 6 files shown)


test/e2e/node.go
```diff
@@ -260,8 +246,8 @@ func (n *Node) Start() error {
 
 func (n *Node) GenesisValidator() genesis.Validator {
 	return genesis.Validator{
-		Account: genesis.Account{
-			PubKey:        n.AccountPubKey,
+		KeyringAccount: genesis.KeyringAccount{
```
staheri14, line 249: [question] Where do we store the public key for this account?

test/util/genesis/accounts.go
```diff
@@ -20,40 +19,28 @@ const (
 	DefaultInitialBalance = 1e15 // 1 billion TIA
 )
 
-// Account represents a user account on the Celestia network.
+// KeyringAccount represents a user account on the Celestia network.
```
staheri14, line 22: Perhaps this part of the comment does not apply anymore:
```
or an address needs to be provided
```

staheri14: [suggestion] The name account in `KeyringAccount` is a bit confusing, as it is more of some info about the account not the actual account. I'd suggest using an alternative name  like `KeyringAccountInfo`.

staheri14 (approved):
Thanks, Callum. After reviewing the new logic, I believe it addresses the issue with the `SimpleE2ETest` and duplicate accounts. However, I haven't tested it yet. Once it's merged, I'll run the e2e test and provide you with an update.

test/util/genesis/accounts.go
cmwaters: So I tried to distinguish `KeyringAccount` as an `Account` that is paired with a Keyring belonging to the `Genesis` struct. If you don't want to use the keyring provided by the `Genesis` struct then you use an `Account` and therefore must include the public key

evan-forbes (approved):
LGTM

test/util/genesis/accounts.go
staheri14: Thanks, your explanation has made things clearer. I think it would be helpful to also add these explanations as a comment to both structs.

PR #3248, opened 2024-04-03
feat: add ntrace for network visualising

cmwaters:
As part of the performance workstream, this is a draft version of a tool for plotting the sending and receiving of data by nodes

example


cmwaters:
Closing this as I think we're going to go with python scripts (cc: @evan-forbes)

Issue #3270, opened 2024-04-05
specs: add minfee module to home page


rootulp:
## Problem

Minfee is listed on https://celestiaorg.github.io/celestia-app/specs/state_machine_modules.html but it isn't listed on https://celestiaorg.github.io/celestia-app/

## Proposal

1. Add minfee to https://celestiaorg.github.io/celestia-app/
2. Figure out a way to not maintain two separate lists of state machine modules

Issue #3151, opened 2024-03-04
Ignore mdbook `theme/`


rootulp:
## Context
https://github.com/celestiaorg/celestia-node/pull/3205#discussion_r1511840410

> highlight.js is the JavaScript that is used to highlight code snippets, you should not need to modify this.

Ref: https://rust-lang.github.io/mdBook/format/theme/index.html

## Proposal
Investigate if the highlight.js in this repo differs from the default. If not, consider gitignoring it.

Ref: https://github.com/celestiaorg/celestia-app/blob/main/specs/theme/highlight.js

PR #3275, opened 2024-04-08
docs(specs): remove root level state machine modules

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3270 by removing the redundant list of state machine modules on the root level README.md.
Closes https://github.com/celestiaorg/celestia-app/issues/3151 by git ignoring `theme`.


specs/book.toml
```diff
@@ -2,7 +2,6 @@
 authors = ["Celestia Labs"]
 language = "en"
 multilingual = false
-src = "src"
```
rootulp, line 5: This config wasn't doing anything because "src" is the default

> src: By default, the source directory is found in the directory named src directly under the root folder. But this is configurable with the src key in the configuration file.

Ref: https://rust-lang.github.io/mdBook/format/configuration/general.html#general-metadata

ninabarbakadze (approved):
much better, thanks

Issue #3276, opened 2024-04-08
x/signal: follow-ups from docs


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/3224 merged and there are some untracked follow ups. It seems like there's agreement on doing them.

## Proposal

1. Can we move some of the unit tests from tally_test.go to keeper_test.go?
2. Why does ResetTally only delete versions less than the current version? Why not delete all versions in the map?

rootulp:
> Can we move some of the unit tests from tally_test.go to keeper_test.go?

Was handled in https://github.com/celestiaorg/celestia-app/pull/3290

> Why does ResetTally only delete versions less than the current version? Why not delete all versions in the map?

Still needs to be handled.

PR #3277, opened 2024-04-08
refactor(x/signal): reset all versions via `ResetTally`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3276


rootulp:
Keeping in draft for now. I plan on rebasing after https://github.com/celestiaorg/celestia-app/pull/3290 merges.

app/ante/fee_checker.go
```diff
@@ -16,6 +17,14 @@ const (
 	priorityScalingFactor = 1_000_000
 )
 
+// The purpose of this wrapper is to enable the passing of an additional paramKeeper parameter in
+// ante.NewDeductFeeDecorator whilst still satisfying the ante.TxFeeChecker type.
+func ValidateTxFeeWrapper(paramKeeper params.Keeper) ante.TxFeeChecker {
+	return func(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
+		return ValidateTxFee(ctx, tx, paramKeeper)
+	}
+}
```
cmwaters: 👍

cmwaters (dismissed):
LGTM

The only edge case against doing this is imagine we release v2 and shortly after v3 (before the network has upgraded to v2). Some nodes upgrade to v2 and signal v2. Some upgrade to v3 and signal v3. v2 eventually receives quorum and the upgrade happens. Now the nodes who signalled v3 have had their "votes" wiped and so they must revote if they want to move to v3

rootulp:
I thought about that edge case and I don't think it's possible because [these lines](https://github.com/celestiaorg/celestia-app/blob/1d73e334a48732ba2ef07caf064a7060175381f0/x/signal/keeper.go#L67-L72). In other words, when the app version is 1, the nodes that try to signal for 3 will hit an `ErrInvalidVersion` because they are signaling for a version too far in advance.

evan-forbes (approved):
LGTM

cmwaters:
> I thought about that edge case and I don't think it's possible because [these lines](https://github.com/celestiaorg/celestia-app/blob/1d73e334a48732ba2ef07caf064a7060175381f0/x/signal/keeper.go#L67-L72). In other words, when the app version is 1, the nodes that try to signal for 3 will hit an ErrInvalidVersion because they are signaling for a version too far in advance.

Makes sense. I always wasn't sure whether we should allow skip versions or not

rootulp:
I'm not sure why golangci-lint is failing on prepare environment. Error looks unrelated to this PR.

rootulp:
I think we can try skipping the pkg cache

evan-forbes (dismissed):

cmwaters (approved):

ninabarbakadze (approved):
sorry for the delay. lgtm

PR #3285, opened 2024-04-09
chore(deps): upgrade to ibc-go v6.3.0

rootulp:
Upgrade to ibc-go [v6.3.0](https://github.com/cosmos/ibc-go/releases/tag/v6.3.0) on the v1.x branch.


ninabarbakadze (commented):
does this mean i should be adding the [block header version fix](https://github.com/cosmos/ibc-go/pull/6099) against v6.3.x?

rootulp:
Good question. IMO yes that way we're not stuck on 6.2.x.

evan-forbes (approved):

ninabarbakadze (approved):

PR #3286, opened 2024-04-09
fix: uses the specified volume amount for the genesis nodes

staheri14:
This PR implements a correction to utilize the disk space amount specified in the resources for genesis nodes, instead of the default `1Gi`.


rach-id (approved):
LGTM, 

Not blocking: would be nice to add some validation to the volume in some subsequent PR

rootulp (approved):

Issue #3284, opened 2024-04-09
Potential bug preventing app from syncing from scratch 


ninabarbakadze:
## Description

while investigating #3167 and going through historical commits using git bisect it output the first commit that supposedly broke syncing a52ed2612d9070af7133dd1c8b3a519d1b9eefaa introduced by this pr https://github.com/celestiaorg/celestia-app/pull/2583

cmwaters:
I'm going to start working on this. My hunch is that the app hash mismatch is caused by the fact that `main` doesn't have an `upgrade` module in it's v1 yet `v1.x` has. This is because the `upgrade` module was extended to include the signalling component for coordination. My plan is to actually separate these out into two modules: an `x/upgrade` which just includes the ibc stuff and should look the same as v1, and `x/signal` for coordinating upgrades

PR #3290, opened 2024-04-10
fix!: split out upgrade into a separate signal module

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/3284

We've been experienceing an app hash mismatch when running `main` on mainnet. I believe this is caused because in v1 we had an upgrade module (albeit unreachable) which was still part of state (kept around for compatibility with IBC). In `main` we coped the upgrade modules state and extended it with the signalling mechanism however this module only applied from v2 to v2 meaning there wasn't the same state in v1. 

This attempts to fix this. It separates out the signal protocol into it's own module (which correctly should only be v2) and reimport the upgrade keeper from the sdk (as was in v1.x)


cmwaters:
I will try run this on `mainnet` to see if the app hash persists

cmwaters:
In v1, we registered all the upgrade types in the encoding register:

https://github.com/celestiaorg/celestia-app/blob/91b6827a80ef8a4d0048ca81f9eda4c88f37a2ef/app/app.go#L163

These are types like `upgrade.Plan` which don't and shouldn't be used. I am wondering if we should still register them

cmwaters:
This hasn't completely fixed the app hash mismatch problem but I think this change is still important.

What nina and I have noticed is that there is a mismatch in the store keys between v1.x and main (which is the source for how the app hash is created). So what it seems we will need to do is:
1) add back the v1.x keys (i.e. blobmodule keys)
2) remove the new keys and instead add the keys to the store during the upgrade by calling `LoadVersionAndUpgrade` in the `MultiCommitStore`

x/signal/README.md
```diff
@@ -27,18 +27,18 @@ See [types/msgs.go](./types/msgs.go) for the message types.
 
 ```shell
 celestia-appd query upgrade tally
```
rootulp, line 29: I assume the query command is also moving from the upgrade module to the signal module

```suggestion
celestia-appd query signal tally
```

x/signal/keeper.go
```diff
@@ -20,12 +20,12 @@ var (
 	defaultSignalThreshold = sdk.NewDec(5).Quo(sdk.NewDec(6))
 )
 
-// SignalThreshold is the fraction of voting power that is required
+// Threshold is the fraction of voting power that is required
```
rootulp, line 23: 👍

x/signal/keeper_test.go
```diff
@@ -20,6 +24,52 @@ import (
 	tmdb "github.com/tendermint/tm-db"
 )
 
+func TestGetVotingPowerThreshold(t *testing.T) {
```
rootulp, line 27: Thanks for moving tests to `keeper_test.go`! I just find it easier to find the test if a function is defined in foo.go then the unit test is in foo_test.go.

I'll revert the change from https://github.com/celestiaorg/celestia-app/pull/3277 and/or rebase after this merges

x/signal/types/msgs.go
```diff
@@ -5,23 +5,17 @@ import (
...
 	QuerierRoute = ModuleName
+	RouterKey    = ModuleName
 
-	// RouterKey defines the module's message routing key
-	RouterKey = ModuleName
-
-	URLMsgSignalVersion = "/celestia.upgrade.v1.Msg/SignalVersion"
-	URLMsgTryUpgrade    = "/celestia.upgrade.v1.Msg/TryUpgrade"
+	URLMsgSignalVersion = "/celestia.signal.v1.Msg/SignalVersion"
+	URLMsgTryUpgrade    = "/celestia.signal_test.v1.Msg/TryUpgrade"
```
rootulp, line 18: [blocking]

```suggestion
	URLMsgTryUpgrade    = "/celestia.signal.v1.Msg/TryUpgrade"
```

rootulp (commented):
I think it's okay if proto-check-breaking fails [here](https://github.com/celestiaorg/celestia-app/actions/runs/8628726730/job/23651430463?pr=3290#step:3:19) because we haven't cut a release from main yet.

One blocking comment but the rest LGTM. Thanks for diving deep on this investigation!

x/signal/types/msgs.go
cmwaters: good catch. Don't know how I changed that.

rootulp (approved):
I just remembered we may need to modify the specs to include a link to the new signal module but that doesn't strictly need to happen in this PR

cmwaters:
> I just remembered we may need to modify the specs to include a link to the new signal module but that doesn't strictly need to happen in this PR

good point

evan-forbes (approved):
I'm _signaling_ my approval of this PR

Issue #2267, opened 2023-08-14
Un-export `GovMaxSquareSize`


rootulp:
definitely not blocking on it, but since this code is identical to https://github.com/celestiaorg/celestia-app/blob/92be1fcc7135bd1150d2064da2d7ac300642adba/app/square_size.go#L8-L28 we could probably find a way to abstract this logic. for instance, it might be best to unexport the logic to retrieve the [raw govmaxsquaresize](https://github.com/celestiaorg/celestia-app/blob/92be1fcc7135bd1150d2064da2d7ac300642adba/x/blob/keeper/params.go#L27-L31) and instead create a new exported method that runs this logic.

While we are no longer able to enforce the underlying value for gov square being smaller than the max square size since we have versioned constants, we could maybe also close https://github.com/celestiaorg/celestia-app/issues/2185. Instead of forcing the underlying value is less, we'd enforce it at read time.

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/2202#discussion_r1293841406_

rootulp:
I would prefer https://github.com/celestiaorg/celestia-app/issues/2185 over this issue but #2185 is unlikely to make it into v2 so I agree we can consider this as a "nice to have" for v2.

rootulp:
> we could probably find a way to abstract this logic. for instance, it might be best to unexport the logic to retrieve the [raw govmaxsquaresize](https://github.com/celestiaorg/celestia-app/blob/92be1fcc7135bd1150d2064da2d7ac300642adba/x/blob/keeper/params.go#L27-L31) and instead create a new exported method that runs this logic.

On re-read, this involves leaking the app version to the blob keeper which seems less than ideal.

PR #3294, opened 2024-04-11
refactor: deduplicate max effective square size

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/2267


cmwaters (approved):
This change makes me realise that it's kinds of odd that the blob keeper has the `SquareSize` parameter in the first place (which is "app wide")

rootulp:
+100

evan-forbes (approved):
makes sense

PR #3295, opened 2024-04-11
docs: improve prepare proposal comments

rootulp:
Fixes a few inconsistencies between the code and comments.

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -15,15 +15,13 @@ import (
 )
 
 // PrepareProposal fulfills the celestia-core version of the ABCI interface by
-// preparing the proposal block data. The square size is determined by first
-// estimating it via the size of the passed block data. Then, this method
-// generates the data root for the proposal block and passes it back to
-// tendermint via the BlockData. Panics indicate a developer error and should
-// immediately halt the node for visibility and so they can be quickly resolved.
+// preparing the proposal block data. This method generates the data root for
+// the proposal block and passes it back to tendermint via the BlockData. Panics
+// indicate a developer error and should immediately halt the node for
+// visibility and so they can be quickly resolved.
 func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePrepareProposal {
 	defer telemetry.MeasureSince(time.Now(), "prepare_proposal")
-	// create a context using a branch of the state and loaded using the
-	// proposal height and chain-id
+	// Create a context using a branch of the state.
 	sdkCtx := app.NewProposalContext(core.Header{
 		ChainID: req.ChainId,
 		Height:  req.Height,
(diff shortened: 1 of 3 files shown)


app/prepare_proposal.go
```diff
@@ -15,15 +15,13 @@ import (
...
 // PrepareProposal fulfills the celestia-core version of the ABCI interface by
-// preparing the proposal block data. The square size is determined by first
-// estimating it via the size of the passed block data. Then, this method
-// generates the data root for the proposal block and passes it back to
-// tendermint via the BlockData. Panics indicate a developer error and should
-// immediately halt the node for visibility and so they can be quickly resolved.
+// preparing the proposal block data. This method generates the data root for
+// the proposal block and passes it back to tendermint via the BlockData. Panics
+// indicate a developer error and should immediately halt the node for
+// visibility and so they can be quickly resolved.
```
rootulp, line 21: Removed a sentence about square size estimating b/c that doesn't seem like an explicit step in this function

app/prepare_proposal.go
```diff
@@ -32,10 +30,6 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 			App: app.BaseApp.AppVersion(),
 		},
 	})
-	// filter out invalid transactions.
-	// TODO: we can remove all state independent checks from the ante handler here such as signature verification
-	// and only check the state dependent checks like fees and nonces as all these transactions have already
-	// passed CheckTx.
```
rootulp, line 38: I think this TODO is technically still possible but it seems like an optimization that we don't need to do. If we want to do it, proposal to create a GH issue to track and I can add the comment back.

app/prepare_proposal.go
```diff
@@ -32,10 +30,6 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 			App: app.BaseApp.AppVersion(),
 		},
 	})
-	// filter out invalid transactions.
```
rootulp, line 35: Moved this closer to the line that filters transactions.

app/prepare_proposal.go
```diff
@@ -72,8 +68,6 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
 		panic(err)
 	}
 
-	// create the new data root by creating the data availability header (merkle
-	// roots of each row and col of the erasure data).
```
rootulp, line 76: The following lines don't actually create the data root. That only happens at the very end of the function body with `dah.Hash()`.

app/prepare_proposal.go
cmwaters: Yeah can we create a GH issue to track this

cmwaters (approved):

app/prepare_proposal.go
rootulp: https://github.com/celestiaorg/celestia-app/issues/3300

ninabarbakadze (approved):

PR #3303, opened 2024-04-11
chore(scripts): persist ABCI responses for single-node.sh

rootulp:
I had to do this for an IBC test and an external engineer who is using this script also has to do this b/c they are querying block results. The relevant config:

```
#######################################################
###         Storage Configuration Options           ###
#######################################################
[storage]

# Set to true to discard ABCI responses from the state store, which can save a
# considerable amount of disk space. Set to false to ensure ABCI responses are
# persisted. ABCI responses are required for /block_results RPC queries, and to
# reindex events in the command-line tool.
discard_abci_responses = true
```

Note: this may increase disk space usage but this script is used for dev purposes and every run deletes the previous run's `data/` directory so I'm not concerned about it.


evan-forbes (approved):

cmwaters (approved):
makes sense

PR #3304, opened 2024-04-11
refactor(app/module): minor readability improvements and more tests

rootulp:
I was reviewing the `app/module` directory because I want to write a README for it but I noticed a few small refactors that I started working on. The refactors include: 
- adding more unit tests
- simplify how `firstVersion` and `lastVersion` is set in the manager
- extracting types to a `types.go` file
- extracting utils to a `utils.go` file and adding tests
- improving code comments
- renaming `VersionedConfigurator` => `Configurator`
- ordering public methods above private methods
- renaming files to match the type they define (i.e. `manager.go`, `configurator.go`)


evan-forbes (approved):

ninabarbakadze (commented):
let's resolve lint and i'll take a look.

p.s sorry i merged main. it's a habit

app/module/utils.go
```diff
@@ -0,0 +1,45 @@
+package module
+
+import (
+	"slices"
+	"sort"
+
+	sdkmodule "github.com/cosmos/cosmos-sdk/types/module"
+)
+
+const authName = "auth"
```
cmwaters, line 10: nit: I wonder if you could just get that name from the auth module. Not that it should ever change

cmwaters (dismissed):
Interesting pattern to separate out into types and utils files. I'm not sure if I'd agree that this improves readability but I don't want to bike shed on this

app/module/utils.go
rootulp: Great call b363d9f3dfdeb1d390a052c9374986bbf517ad3e

evan-forbes (approved):

ninabarbakadze (approved):

PR #3323, opened 2024-04-15
chore: remove checkboxes from issue templates

rootulp:
We basically never use the checkboxes on issues so they're just visual noise. Proposal to delete them from the issue templates.


ninabarbakadze (commented):
do we not want these for external contributors?

evan-forbes (approved):
yas delete

rootulp:
> do we not want these for external contributors?

We (the maintainers) aren't actually filling out the checkboxes so no. External contributors aren't encouraged to fill them out.

ninabarbakadze (approved):

PR #3329, opened 2024-04-16
chore: resolve loopclosure warning

rootulp:
Fixes this warning

Screenshot 2024-04-16 at 11 45 35 AM

Ref: https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/loopclosure

Note: since we're already on Go 1.22, the loop already creates a new variable per loop iteration so we don't technically need to merge this but idk how to inform the warning generator that we're already on Go 1.22 and the warning doesn't apply to us.

### DIFF
--- a/app/test/priority_test.go
+++ b/app/test/priority_test.go
@@ -80,6 +80,7 @@ func (s *PriorityTestSuite) TestPriorityByGasPrice() {
 	wg := &sync.WaitGroup{}
 	for _, signer := range s.signers {
 		wg.Add(1)
+		signer := signer // new variable per iteration
 		go func() {
 			defer wg.Done()
 			// ensure that it is greater than the min gas price


staheri14 (approved):

app/test/priority_test.go
```diff
@@ -80,6 +80,7 @@ func (s *PriorityTestSuite) TestPriorityByGasPrice() {
 	wg := &sync.WaitGroup{}
 	for _, signer := range s.signers {
 		wg.Add(1)
+		signer := signer // new variable per iteration
```
evan-forbes, line 83: while this gets rid of the linting error, does this actually get rid of the problem?

would it be best to pass this to the goroutine?

rootulp: The problem doesn't actually exist for us because we're on Go 1.22. Go 1.22 has new lifetime rules:

>  In Go 1.22, loop variables are created anew for each iteration, so such references are no longer at risk of using a variable after it has been updated by the loop.

Ref: https://tip.golang.org/doc/go1.22

My hunch is that `go vet` in my editor (potentially others) is using an older version of Go.

evan-forbes: Woah, dope

Issue #3336, opened 2024-04-16
Consensus failure on reboot on `main`


evan-forbes:
using current `main`, if we reboot, we currently panic

perhaps we're not peristing some state or value for the context? In theory, I don't think the recent change to `MaxEffectiveSquareSize` should have changed anything, but perhaps we missed something? This doesn't occur on 97f09de75098e4bb5176b845b97628f73058745d , the commit directly before that change.

To recreate, we can run the single node script, shut down after producing a block or two, then reboot.

[...]

PR #3338, opened 2024-04-17
revert: deduplicate max effective square size

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3336 by reverting #3294
Opens https://github.com/celestiaorg/celestia-app/issues/2267

## Testing

I observe a consensus failure if I reboot celestia-app on `main`. I don't observe a consensus failure if I reboot celestia-app on this branch after the revert.

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -4,7 +4,6 @@ import (
 	"time"
 
 	"github.com/celestiaorg/celestia-app/v2/app/ante"
-	"github.com/celestiaorg/celestia-app/v2/app/squaresize"
 	"github.com/celestiaorg/celestia-app/v2/pkg/appconsts"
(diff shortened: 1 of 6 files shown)


evan-forbes:
do we know what the exact cause is? is the refactor fundementally flawed, or is it just some simple state thingy?

I know its just a nice to have, so perhaps we shelve it for a while, but naively without looking into it, it seems like we could have the refactor w/o the the state issues?

cmwaters:
Yeah I would kind of like to have an idea of why this failed instead of simply blindly reverting it

rootulp:
I was planning on investigating it after getting `main` back to a functional state

evan-forbes (approved):
> I was planning on investigating it after getting main back to a functional state

makes sense, sgtm!

staheri14 (approved):

cmwaters (approved):

Issue #3167, opened 2024-03-07
Can't sync from genesis using celestia-appd main


rootulp:
## Context

I'm trying to do a manual sync from genesis test using a celestia-appd binary build from main. 

## Problem

I can't get past block height 1. I'm observing this log:

```
2:57AM ERR Error in validation err="wrong Block.Header.AppHash.  Expected 869EBEF734A657323408029077A7B0F835DFBB771582BBCE4084D29D5661DFF8, got F43C8B84340E6803AC86AD934B03E7145E155F7E8E2E7E19412D6ED8A44CBF14" module=blockchain
2:57AM ERR Stopping peer for error err="blockchainReactor validation error: wrong Block.Header.AppHash.  Expected 869EBEF734A657323408029077A7B0F835DFBB771582BBCE4084D29D5661DFF8, got F43C8B84340E6803AC86AD934B03E7145E155F7E8E2E7E19412D6ED8A44CBF14" module=p2p peer={"Data":{},"Logger":{}}
```

More info if helpful

```
root@rootul-mainnet-consensus:~# celestia-appd status
[...]

PR #3353, opened 2024-04-19
feat: versioned ibc middleware

ninabarbakadze:
## Overview

Addresses: https://github.com/celestiaorg/celestia-app/issues/3167

Adds a versioning wrapper around the IBC middleware so that they are not run when they are not supposed to be

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -329,17 +329,22 @@ func New(
 	)
 	// transfer stack contains (from top to bottom):
 	// - Token Filter
+	// - Packet Forwarding Middleware
 	// - Transfer
 	var transferStack ibcporttypes.IBCModule
 	transferStack = transfer.NewIBCModule(app.TransferKeeper)
-	transferStack = tokenfilter.NewIBCMiddleware(transferStack)
-	transferStack = packetforward.NewIBCMiddleware(
+	packetForwardMiddleware := packetforward.NewIBCMiddleware(
 		transferStack,
 		app.PacketForwardKeeper,
 		0, // retries on timeout
 		packetforwardkeeper.DefaultForwardTransferPacketTimeoutTimestamp, // forward timeout
 		packetforwardkeeper.DefaultRefundTransferPacketTimeoutTimestamp,  // refund timeout
 	)
+	// packetForwardMiddleware is used only for version 2
+	transferStack = module.NewWithVersions(packetForwardMiddleware, transferStack, v2, v2)
+	// token filter wraps packer forward middleware and is thus the first module in the transfer stack
+	tokenFilterMiddelware := tokenfilter.NewIBCMiddleware(transferStack)
+	transferStack = module.NewWithVersions(tokenFilterMiddelware, transferStack, v1, v2)
 
 	// Create evidence Keeper for to register the IBC light client misbehaviour evidence route
 	evidenceKeeper := evidencekeeper.NewKeeper(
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):
dang

app/module/ibc.go
```diff
@@ -0,0 +1,149 @@
...
+	channelCap *capabilitytypes.Capability,
+	counterparty channeltypes.Counterparty,
+	version string,
+) (string, error) {
+	currentAppVersion := ctx.BlockHeader().Version.App
+	if currentAppVersion >= v.fromVersion && currentAppVersion <= v.toVersion {
+		return v.wrappedModule.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, channelCap, counterparty, version)
+	}
+	return v.nextModule.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, channelCap, counterparty, version)
+}
```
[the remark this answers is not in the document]
rootulp: not blocking but seems like a decent suggestion

app/app.go
```diff
@@ -329,17 +329,22 @@ func New(
...
 		transferStack,
 		app.PacketForwardKeeper,
 		0, // retries on timeout
 		packetforwardkeeper.DefaultForwardTransferPacketTimeoutTimestamp, // forward timeout
 		packetforwardkeeper.DefaultRefundTransferPacketTimeoutTimestamp,  // refund timeout
 	)
+	// packetForwardMiddleware is used only for version 2
+	transferStack = module.NewVersionedIBCModule(packetForwardMiddleware, transferStack, v2, v2)
+	// token filter wraps packer forward middleware and is thus the first module in the transfer stack
+	tokenFilterMiddelware := tokenfilter.NewIBCMiddleware(transferStack)
```
rootulp, line 346: [question] is it possible to version middleware or only the surrounding IBC modules? The line below this:

```
	transferStack = module.NewVersionedIBCModule(tokenFilterMiddelware, transferStack, v1, v2)
```

is kinda difficult to follow b/c IIUC it composed the previous transfer stack that contained PFM (v2 to v2) into a new transfer stack that is (v1 to v2). Just an idea but I wonder if it would it be easier to follow if we versioned middleware instead of the surrounding IBC module. IDK what that would look like yet so feel free to ignore.

app/module/ibc.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+	channeltypes "github.com/cosmos/ibc-go/v6/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/v6/modules/core/05-port/types"
+	exported "github.com/cosmos/ibc-go/v6/modules/core/exported"
+)
+
+func NewVersionedIBCModule(
```
rootulp, line 11: [question] do we want a unit test that verifies this new versioned behavior?

rootulp (approved):

app/app.go
cmwaters: This is how middleware works though: it's essentially modules that keep wrapping one another (similar to the antehandler design)

rootulp: Makes sense. Resolving.

cmwaters (approved):

rootulp (approved):

PR #3357, opened 2024-04-19
fix!: resolve CodeQL warnings

rootulp:
I think this should resolve all the CodeQL issues identified [here](https://github.com/celestiaorg/celestia-app/security/code-scanning). Most of the issues were from x/blobstream or pkg/proof so would appreciate @rach-id 's review.


pkg/proof/querier.go
```diff
@@ -123,7 +129,7 @@ func QueryShareInclusionProof(_ sdk.Context, path []string, req abci.RequestQuer
 
 // ParseNamespace validates the share range, checks if it only contains one namespace and returns
 // that namespace ID.
-func ParseNamespace(rawShares []shares.Share, startShare, endShare int) (appns.Namespace, error) {
+func ParseNamespace(rawShares []shares.Share, startShare int64, endShare int64) (appns.Namespace, error) {
```
rootulp, line 132: this is technically a breaking change

x/blobstream/client/verify.go
```diff
@@ -194,7 +198,7 @@ func sharesCmd() *cobra.Command {
 	return addVerifyFlags(command)
 }
 
-func VerifyShares(ctx context.Context, logger tmlog.Logger, config VerifyConfig, height uint64, startShare uint64, endShare uint64) (isCommittedTo bool, err error) {
+func VerifyShares(ctx context.Context, logger tmlog.Logger, config VerifyConfig, height int64, startShare uint64, endShare uint64) (isCommittedTo bool, err error) {
```
rootulp, line 201: this is also breaking

rach-id (commented):
As a first look, I think the blobstream related ones can be ignored since all will be deprecated, no need to keep fixing stuff there (there is a button to set the reported errors as won't fix). For the remaining ones, will review shortly

pkg/proof/proof.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	math "math"
```
rach-id, line 7: nit
```suggestion
	"math"
```

pkg/proof/proof_test.go
```diff
@@ -123,8 +123,8 @@ func TestNewShareInclusionProof(t *testing.T) {
 
 	type test struct {
 		name          string
-		startingShare int
-		endingShare   int
+		startingShare int64
```
rach-id, line 126: why are we changing these to `int64` then converting them back to `int` in line 223?

pkg/proof/querier.go
```diff
@@ -3,6 +3,7 @@ package proof
 import (
 	"bytes"
 	"fmt"
+	math "math"
```
rach-id, line 6: nit
```suggestion
	"math"
```

pkg/proof/querier.go
rach-id: Why are we switching here to `int64`? I don't think the shares numbers will go to reach `math.MaxInt`

pkg/proof/querier.go
```diff
@@ -156,3 +162,10 @@ func ParseNamespace(rawShares []shares.Share, startShare, endShare int) (appns.N
 	}
 	return startShareNs, nil
 }
+
+func safeConvert(x int64) (int, error) {
+	if math.MinInt <= x && x <= math.MaxInt {
+		return int(x), nil
+	}
+	return 0, fmt.Errorf("invalid int: %d", x)
```
rach-id, line 170: [optional]
guess it's better to stick to the same error message in the other function. Alternatively, we could create an error type for it.

```suggestion
	return 0, fmt.Errorf("value %d is too large to convert to int", x)
```

pkg/proof/querier.go
```diff
@@ -156,3 +162,10 @@ func ParseNamespace(rawShares []shares.Share, startShare, endShare int) (appns.N
 	}
 	return startShareNs, nil
 }
+
+func safeConvert(x int64) (int, error) {
```
rach-id, line 166: [optional]
Guess it would make sense for this function to specify exactly which conversion it will do in the title so that if we need other conversions, we don't end up renaming this one.

```suggestion
func safeConvertInt64ToInt(x int64) (int, error) {
```

pkg/proof/querier_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_safeConvert(t *testing.T) {
+	testCases := []struct {
+		input int64
+		want  int
+	}{
+		{input: math.MinInt64, want: math.MinInt64},
```
rach-id, line 15: The safe convert implementation is correct, but this test is platform dependent, if I understand correctly, and should return an error in the general case. This is passing because `int == int64` for 64 bit systems but I don't think we can assume that everywhere, no?

If we're fine with that assumption :shipit: as only devs will run the tests and we can assume they don't have 90s computers.

test/testground/network/consensus_node.go
```diff
@@ -321,3 +326,10 @@ func parsePeerID(input string) (string, net.IP, int, error) {
 
 	return address, ip, portInt, nil
 }
+
+func safeConvert(x int) (uint16, error) {
```
rach-id, line 330: [optional]
Same, would be nice if this is renamed `safeConvertIntToUint16`

test/testground/network/consensus_node.go
```diff
@@ -321,3 +326,10 @@ func parsePeerID(input string) (string, net.IP, int, error) {
...
+	if x > 0 && x <= math.MaxUint16 {
```
rach-id, line 331:
```suggestion
	if x >= 0 && x <= math.MaxUint16 {
```

`uint16(0)` is a valid value

test/testground/network/consensus_node.go
```diff
@@ -321,3 +326,10 @@ func parsePeerID(input string) (string, net.IP, int, error) {
...
+		return uint16(x), nil
+	}
+	return 0, fmt.Errorf("invalid uint16: %d", x)
```
rach-id, line 334: [optional]
Similar error message as above

rach-id (commented):
utAck

Left a few comment to understand more

pkg/proof/querier.go
rootulp: Agreed! Thanks

pkg/proof/querier.go
rootulp: Resolved via fc530d880eb6b484dfcdc126f3a7696d96750e1f

test/testground/network/consensus_node.go
rootulp: a9327a7dc995ecd5e76b03ec02f371aa852fc7ec

test/testground/network/consensus_node.go
rootulp: good catch! 65e4601bccfb52143f43061d2c39c9494c93a147

test/testground/network/consensus_node.go
rootulp: a9327a7dc995ecd5e76b03ec02f371aa852fc7ec

pkg/proof/querier_test.go
rootulp: yea totally agreed, I can delete the test entirely if that would help.

rootulp: 8821560e7576dbb677852f2b97194a63b4b8b1f0

pkg/proof/proof_test.go
rootulp: because `ParseNamespace` now accepts an `int64`

rootulp:
>  I think the blobstream related ones can be ignored since all will be deprecated

Agreed they can be ignored or fixed. Since this PR already fixes them, I'm not going to revert it.

pkg/proof/querier.go
rootulp: This removes the need to do additional range checking inside the body of this function.

rach-id: I am not sure if I see where this range check is happening, can you point the line please?

x/blobstream/client/verify.go
rach-id: this change is for consistency, right?

rach-id:
> Agreed they can be ignored or fixed. Since this PR already fixes them, I'm not going to revert it.

100%, I meant if there are any left, it's fine not to waste time on them

x/blobstream/client/verify.go
rootulp: Mostly motivated by consistency. If this was reverted to `uint64` then we'd have to do bounds checks / converting inside this function body.

- `trpc.Block` accepts a `*int64 height`
- `VerifyDataRootInclusion` accepts a `int64 height`.

rach-id (approved):

evan-forbes (approved):

ninabarbakadze (approved):

PR #3359, opened 2024-04-20
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:
If you don't need to format the string, it is recommended to use error.New()


rach-id (commented):
Is there a reason why the proto were regenerated?

ChengenH:
> Is there a reason why the proto were regenerated?

Fixed, please check again

rach-id (approved):

evan-forbes (approved):

PR #3368, opened 2024-04-23
test: packet forward middleware

ninabarbakadze:


test/interchain/packet_forward_test.go
```diff
@@ -0,0 +1,192 @@
...
+	Receiver       string        `json:"receiver"`
+	Port           string        `json:"port"`
+	Channel        string        `json:"channel"`
+	Timeout        time.Duration `json:"timeout"`
+	Retries        *uint8        `json:"retries,omitempty"`
+	Next           *string       `json:"next,omitempty"`
+	RefundSequence *uint64       `json:"refund_sequence,omitempty"`
+}
+
+// TestPacketForwardMiddleware verifies that Packet Forward Middleware (PFM) works as expected by sending a transfer from Celestia to Cosmos Hub, back to Celestia which then forwards it to Cosmos Hub 2. Celestia -> CosmosHub -> Celestia -> CosmosHub2 .
```
rootulp, line 35: [question] can this be simplified to CosmosHub -> Celestia -> CosmosHub2?

Basically it reduces one extra hop. And AFAICT we only care about verifying that Celestia can forward the metadata along

rootulp (commented):
Have you tried keeping the containers alive after the test completes? That way it's easier to see container logs to see which step isn't completing

Ref: https://github.com/celestiaorg/celestia-app/tree/main/test/interchain#troubleshooting

test/interchain/packet_forward_test.go
ninabarbakadze: I was doing that in order to get some tia on cosmos and also test forwarding all in one tx

rootulp:
Closing this b/c we decided in a sync to pursue a unit test instead of interchain tests.

PR #3310, opened 2024-04-12
feat!: disable x/blobstream in v2

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3305

## Testing
<details>
<summary>Manually verified on a local devnet that a user can register an EVM address if the chain is on app version 1.</summary>

```shell
$   celestia-appd tx qgb register \
    "$(celestia-appd keys show validator --home ${CELESTIA_APP_HOME} --bech val -a)" \
    0x966e6f22781EF6a6A82BBB4DB3df8E225DfD9485 \
    --from ${KEY_NAME} \
    --home ${CELESTIA_APP_HOME} \
    --fees 30000utia \
    --broadcast-mode block \
    --yes
code: 0
codespace: ""
data: 12300A2E2F63656C65737469612E7167622E76312E4D7367526567697374657245564D41646472657373526573706F6E7365
events:
- attributes:
  - index: true
    key: c3BlbmRlcg==
    value: Y2VsZXN0aWExc3B6YWU4MzNnM2E4dDQwdXN3NTAyZXR4dHo2c2czaG5meDg4OHc=
  - index: true
    key: YW1vdW50
    value: MzAwMDB1dGlh
  type: coin_spent
- attributes:
  - index: true
    key: cmVjZWl2ZXI=
    value: Y2VsZXN0aWExN3hwZnZha20yYW1nOTYyeWxzNmY4NHoza2VsbDhjNWxwbmpzM3M=
  - index: true
    key: YW1vdW50
    value: MzAwMDB1dGlh
  type: coin_received
- attributes:
  - index: true
    key: cmVjaXBpZW50
    value: Y2VsZXN0aWExN3hwZnZha20yYW1nOTYyeWxzNmY4NHoza2VsbDhjNWxwbmpzM3M=
  - index: true
    key: c2VuZGVy
    value: Y2VsZXN0aWExc3B6YWU4MzNnM2E4dDQwdXN3NTAyZXR4dHo2c2czaG5meDg4OHc=
  - index: true
    key: YW1vdW50
    value: MzAwMDB1dGlh
  type: transfer
- attributes:
  - index: true
    key: c2VuZGVy
    value: Y2VsZXN0aWExc3B6YWU4MzNnM2E4dDQwdXN3NTAyZXR4dHo2c2czaG5meDg4OHc=
  type: message
- attributes:
  - index: true
    key: ZmVl
    value: MzAwMDB1dGlh
  - index: true
    key: ZmVlX3BheWVy
[...]

PR #3372, opened 2024-04-23
refactor(app): app.go and modules.go

rootulp:
A bunch of small refactors I made while working on https://github.com/celestiaorg/celestia-app/pull/3310


cmwaters (approved):
LGTM

It is feasible that we use the hard coded upgrade height for future upgrades (say in the example that we decide the signalling mechanism is too complicated or isn't right for us in someway).

I had thought about whether we still wanted to keep this coordination mechanism around for future upgrades but I guess this is a conversation for a later date right now `upgradeHeightV2` makes sense

rootulp:
SGTM

ninabarbakadze (approved):

PR #3375, opened 2024-04-24
feat: check if the celestia binary is installed in single-node.sh

rach-id:
## Overview

Given that we're using `set -o errexit` in the `single-node.sh` script, and we're setting the variable `CELESTIA_APP_VERSION=$(celestia-appd version 2>&1)`, if `celestia-appd` is not in the path, the execution will return nothing and just exist, and users can be confused why the script is not doing anything. This PR adds a check whether the celestia-appd binary is in the PATH and is executable, and returns an error otherwise to help users.

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -5,6 +5,12 @@ set -o errexit
 # Stop script execution if an undefined variable is used
 set -o nounset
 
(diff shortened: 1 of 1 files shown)


rootulp (approved):
Thanks!!!

staheri14 (approved):
👍

PR #3376, opened 2024-04-24
chore: POSIX compatible single-node

rach-id:
## Overview

Make the `single-node.sh` script POSIX compatible so it runs on macOS/Linux etc without issue


rootulp (approved):
LGTM! Thanks

staheri14 (approved):
utAck

Issue #3380, opened 2024-04-24
`ERROR E2ESimple: expected context.DeadlineExceeded, got rpc error: code = DeadlineExceeded desc = context deadline exceeded`


rootulp:
## Context

I'm trying to run the end-to-end tests

## Problem

```shell
$ make test-e2e
--> Running end to end tests
go run ./test/e2e
test-e2e2024/04/24 15:03:09 No particular test specified. Running all tests.
test-e2e2024/04/24 15:03:09 go run test/e2e/*.go <test_name> to run a specific test
test-e2e2024/04/24 15:03:09 Valid tests are: E2ESimple

test-e2e2024/04/24 15:03:09 === RUN E2ESimple
test-e2e2024/04/24 15:03:09 Running simple e2e test version 32accba
INFO[2024-04-24T15:03:09-04:00]knuu/knuu.go:168 LOG_LEVEL: info
INFO[2024-04-24T15:03:09-04:00]knuu/knuu.go:66 Initializing knuu with scope: E2ESimple_20240424_150309
test-e2e2024/04/24 15:03:23 Creating testnet validators
test-e2e2024/04/24 15:03:23 Creating account
{"level":"info","name":"alice","pk":"PubKeySecp256k1{023765DF9278C8766053D3ED009D175A77D47906787134F9A289EB692BC15EE264}","time":"2024-04-24T15:03:23-04:00","message":"txsim account created and added to genesis"}
test-e2e2024/04/24 15:03:23 Setting up testnets
{"level":"info","name":"val0","directory":"/var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/val0","time":"2024-04-24T15:03:29-04:00","message":"Creating validator's config and data directories"}
{"level":"info","name":"val1","directory":"/var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/val1","time":"2024-04-24T15:03:47-04:00","message":"Creating validator's config and data directories"}
{"level":"info","name":"val2","directory":"/var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/val2","time":"2024-04-24T15:04:00-04:00","message":"Creating validator's config and data directories"}
{"level":"info","name":"val3","directory":"/var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/val3","time":"2024-04-24T15:05:29-04:00","message":"Creating validator's config and data directories"}
[...]

PR #3382, opened 2024-04-24
test(e2e): isDeadlineExceeded

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3380

## Testing

I make it past the place where I hit the error in the issue. Now I'm hitting a different error which makes me think this PR resolved at least one issue.

```
test-e2e2024/04/24 15:42:27 Starting testnets
test-e2e2024/04/24 15:45:31 Running txsim
test-e2e2024/04/24 15:46:01 Reading blockchain
test-e2e2024/04/24 15:46:51 --- ERROR E2ESimple: expected at least 10 transactions, got 1
exit status 1
make: *** [test-e2e] Error 1
```

### DIFF
--- a/test/e2e/simple.go
+++ b/test/e2e/simple.go
@@ -2,9 +2,9 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/celestiaorg/celestia-app/v2/app"
@@ -50,10 +50,8 @@ func E2ESimple(logger *log.Logger) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 	opts := txsim.DefaultOptions().WithSeed(seed).SuppressLogs()
-	err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...)
-
-	if !errors.Is(err, context.DeadlineExceeded) {
-		return fmt.Errorf("expected context.DeadlineExceeded, got %w", err)
+	if err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...); !isDeadlineExceeded(err) {
+		return fmt.Errorf("expected error to contain context deadline exceeded but got %w", err)
 	}
 
 	logger.Println("Reading blockchain")
@@ -71,4 +69,11 @@ func E2ESimple(logger *log.Logger) error {
 		return fmt.Errorf("expected at least 10 transactions, got %d", totalTxs)
 	}
 	return nil
+}
+
+// isDeadlineExceeded returns true if the error contains "context deadline
+// exceeded". It should return true for a context.DeadlineExceeded error and an
+// RPC error that wraps a DeadlineExceeded.
+func isDeadlineExceeded(err error) bool {
+	return strings.Contains(err.Error(), "context deadline exceeded")
 }


test/e2e/simple.go
```diff
@@ -50,10 +50,8 @@ func E2ESimple(logger *log.Logger) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 	opts := txsim.DefaultOptions().WithSeed(seed).SuppressLogs()
-	err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...)
-
-	if !errors.Is(err, context.DeadlineExceeded) {
-		return fmt.Errorf("expected context.DeadlineExceeded, got %w", err)
+	if err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...); !isDeadlineExceeded(err) {
+		return fmt.Errorf("expected error to contain context deadline exceeded but got %w", err)
```
rach-id, line 54: context deadline exceeded is only one type of errors that txsim can return, and there are others. Why are we expecting it to be the only one?

rach-id (commented):
> I make it past the place where I hit the error in the issue. Now I'm hitting a different error which makes me think this PR resolved at least one issue.

I personally prefer to leave this PR open until all the issues are resolved, but I defer to your judgement

test/e2e/simple.go
rach-id: if that's the happy path, why would we need to execute the rest of the tests? 

also, why wouldn't it be something like:

```

if err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...); !isDeadlineExceeded(err) {
	return fmt.Errorf("expected error to contain context deadline exceeded but got %w", err)
}

=>

err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...)
assert.Falsef(t, isDeadlineExceeded(err), "expected error to contain context deadline exceeded but got %w", err)
```
or similar?

test/e2e/simple.go
```diff
@@ -50,10 +50,8 @@ func E2ESimple(logger *log.Logger) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 	opts := txsim.DefaultOptions().WithSeed(seed).SuppressLogs()
-	err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...)
-
-	if !errors.Is(err, context.DeadlineExceeded) {
```
cmwaters, line 55: Does this not work because the error gets wrapped? If that is the case I would prefer that the error gets unwrapped / or otherwise handled in txsim itself

cmwaters (changes requested):

test/e2e/simple.go
rootulp: I think a context deadline exceeded error gets wrapped in an RPC error but TBH then I would expect the original implementation `errors.Is` to have worked.

I can hold off on merging this b/c I don't have a convincing way to prove this helped or hurt b/c the test is flaky for other reasons outside this issue.

test/e2e/simple.go
rootulp: oh I kinda see what you mean. I think you meant:

```go
err = txsim.Run(ctx, testnet.GRPCEndpoints()[0], kr, encCfg, opts, sequences...)
assert.True(t, isDeadlineExceeded(err), "expected error to contain context deadline exceeded but got %w", err)
```

but yea that would make it clearer that the test expects a context deadline exceeded error.

test/e2e/simple.go
cmwaters: Yeah I think we should then unwrap it in txsim and just return the raw deadline exceeded error

rootulp:
Not planning on working on this

Issue #3258, opened 2024-04-04
Tx consumes more gas in case of set GranterFee


vgonkivs:
I'm currently working on enabling GrantFee functionality in celestia-node for SubmitPFB transactions and I'm observing that the [gas estimation](https://github.com/celestiaorg/celestia-app/blob/ac9428872d803264f9c6ea595cbb2d6cc2d8c519/x/blob/types/payforblob.go#L171) is not working as I'm getting "out of gas" each time I set a tx granter.

cmwaters:
Yeah, the model for estimating works on the assumption that there are no added options to the transaction i.e. it's just a vanialla PFB

Bidon15:
> Yeah, the model for estimating works on the assumption that there are no added options to the transaction i.e. it's just a vanilla PFB

So does it mean this is node issue or app?

cmwaters:
Its an app issue. The best short term solution is to use the `EstimateGas` on signer or `SimulateTx` via grpc which should provide an accurate gas estimate

cmwaters:
Ref: https://github.com/celestiaorg/celestia-app/issues/3342

PR #3385, opened 2024-04-25
feat: introduce gas multiplier to increase gas limit

vgonkivs:
## Overview
Resolves https://github.com/celestiaorg/celestia-app/issues/3258

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -61,7 +61,7 @@ type Signer struct {
 	localSequence uint64
 	// the chains last known sequence number
 	networkSequence uint64
-	// gasMultiplier is used to increase gas limit as it is sometimes underestimate
+	// gasMultiplier is used to increase gas limit as it is sometimes underestimated
 	gasMultiplier float64
 	// lookup map of all pending and yet to be confirmed outbound transactions
 	outboundSequences map[uint64]struct{}
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -8,9 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/celestiaorg/celestia-app/v2/app/encoding"
-	apperrors "github.com/celestiaorg/celestia-app/v2/app/errors"
-	blobtypes "github.com/celestiaorg/celestia-app/v2/x/blob/types"
 	"github.com/celestiaorg/go-square/blob"
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/grpc/tmservice"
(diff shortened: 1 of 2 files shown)


cmwaters (dismissed):
LGTM 👍

pkg/user/signer_test.go
```diff
@@ -43,7 +44,7 @@ func (s *SignerTestSuite) SetupSuite() {
 	s.Require().NoError(err)
 	addr, err := rec.GetAddress()
 	s.Require().NoError(err)
-	s.signer, err = user.SetupSigner(s.ctx.GoContext(), s.ctx.Keyring, s.ctx.GRPCClient, addr, s.encCfg)
+	s.signer, err = user.SetupSigner(s.ctx.GoContext(), s.ctx.Keyring, s.ctx.GRPCClient, addr, s.encCfg, user.WithGasMultiplier(1.2))
```
rootulp, line 47: [question] if this test needs a gas multiplier of 1.2 should we consider bumping the default to 1.2?

rootulp (dismissed):
LGTM! Great work

pkg/user/signer_test.go
vgonkivs: It was added just to overwrite the default value.

rootulp: ahh got it. Okay nvm

pkg/user/signer.go
```diff
@@ -48,6 +61,8 @@ type Signer struct {
 	localSequence uint64
 	// the chains last known sequence number
 	networkSequence uint64
+	// gasMultiplier is used to increase gas limit as it is sometimes underestimate
```
rach-id, line 64:
```suggestion
	// gasMultiplier is used to increase gas limit as it is sometimes underestimated
```

pkg/user/signer.go
```diff
@@ -23,9 +20,25 @@ import (
...
+
+	"github.com/celestiaorg/celestia-app/v2/app/encoding"
+	apperrors "github.com/celestiaorg/celestia-app/v2/app/errors"
+	blobtypes "github.com/celestiaorg/celestia-app/v2/x/blob/types"
 )
 
-const DefaultPollTime = 3 * time.Second
+const (
+	DefaultPollTime              = 3 * time.Second
+	DefaultGasMultiplier float64 = 1.1
```
rach-id, line 31: [not blocking]
Wouldn't this be better if it's configurable and exposed to the user someway? At some point, 1.1 might not be enough too, if I understand correctly

vgonkivs: I'm not sure that the user should be aware of any multipliers that we have. The current multiplier covers the case when our Estimation is not perfect(in case of additional options). I also think we'll never face the issue when we have to increase the `DefaultMultiplier`.  Signer has an option that allows to configure the multiplier if it's required and imo it's enough.

cmwaters (approved):

pkg/user/signer.go
cmwaters: Yeah this value is configurable as a `signer.Option`

evan-forbes (approved):
:+1:

rootulp (approved):

PR #3394, opened 2024-04-26
refactor(app): app.go

rootulp:
A few small refactors to `app.go`.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -81,6 +81,7 @@ import (
 	ibcporttypes "github.com/cosmos/ibc-go/v6/modules/core/05-port/types"
 	ibchost "github.com/cosmos/ibc-go/v6/modules/core/24-host"
 	ibckeeper "github.com/cosmos/ibc-go/v6/modules/core/keeper"
+	ibctesting "github.com/cosmos/ibc-go/v6/testing"
 	ibctestingtypes "github.com/cosmos/ibc-go/v6/testing/types"
 	"github.com/spf13/cast"
 	abci "github.com/tendermint/tendermint/abci/types"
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -109,6 +110,7 @@ const (
 )
 
 var _ servertypes.Application = (*App)(nil)
+var _ ibctesting.TestingApp = (*App)(nil)
```
rootulp, line 113: A few comments in this file claim that a method implements the TestingApp interface so I think we should enforce that via a compile-time check.

app/app.go
```diff
@@ -665,11 +668,8 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, _ config.APIConfig) {
 	authtx.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
 	// Register new tendermint queries routes from grpc-gateway.
 	tmservice.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
-
 	// Register node gRPC service for grpc-gateway.
 	nodeservice.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
-
-	// Register the
```
rootulp, line 672: This comment was noise before

app/app.go
```diff
@@ -724,16 +724,3 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
...
-func extractRegisters(m sdkmodule.BasicManager) []encoding.ModuleRegister {
-	// TODO: might be able to use some standard generics in go 1.18
-	s := make([]encoding.ModuleRegister, len(m))
-	i := 0
-	for _, v := range m {
-		s[i] = v
-		i++
-	}
-	return s
-}
```
rootulp, line 739: This implementation is unnecessarily complex. Refactored and moved to modules.go because it is only used once in that file.

cmwaters (approved):

staheri14 (approved):

PR #3408, opened 2024-04-29
feat: adds functionality to read the blockchain headers into a csv file

staheri14:
While we can obtain traced data and use it to calculate block time, size, and throughput, this is merely an alternative approach. We can achieve the same results at the end of the test using Python scripts and without analyzing the traced data.


staheri14:
[To the reviewers] Since the PR is not based on the `main` branch (is based off the `experimentation` branch), reviews are not strictly necessary. However, we plan to merge the `experimentation` branch as soon as we finalize the branch management for the celestia-core and release a new version of that repository. Therefore, it would be beneficial to review changes in the `experimentation` branch incrementally (for each PR), rather than postponing reviews until the time of the merge.

evan-forbes (approved):

Issue #2942, opened 2023-12-15
TestPriorityByGasPrice flake


rootulp:
## Description 

We're currently experiencing a flake see: 

- https://github.com/celestiaorg/celestia-app/actions/runs/7225039499/job/19687642759?pr=2941#step:4:31
- https://github.com/celestiaorg/celestia-app/actions/runs/8559843485/job/23457500233?pr=3266

```
--- FAIL: TestPriorityTestSuite (7.08s)
    --- FAIL: TestPriorityTestSuite/TestPriorityByGasPrice (1.02s)
        priority_test.go:118: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/priority_test.go:118
            	Error:      	Should be true
            	Test:       	TestPriorityTestSuite/TestPriorityByGasPrice
    network.go:48: tearing down testnode
FAIL
```

## Acceptance Criteria

- [ ] Investigate and resolve the flake

rootulp:
Slightly different flake for the same test:

```
--- FAIL: TestPriorityTestSuite (6.08s)
    --- FAIL: TestPriorityTestSuite/TestPriorityByGasPrice (0.00s)
        priority_test.go:89: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/priority_test.go:89
            	Error:      	Not equal: 
            	            	expected: 0x0
            	            	actual  : 0xd
            	Test:       	TestPriorityTestSuite/TestPriorityByGasPrice
    network.go:48: tearing down testnode
```

Ref: https://github.com/celestiaorg/celestia-app/actions/runs/7789508543/job/21241291035?pr=3079

evan-forbes:
```
--- FAIL: TestPriorityTestSuite (12.07s)
    --- FAIL: TestPriorityTestSuite/TestPriorityByGasPrice (1.01s)
        priority_test.go:124: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/priority_test.go:124
            	Error:      	Should be true
            	Test:       	TestPriorityTestSuite/TestPriorityByGasPrice
    network.go:46: tearing down testnode
```

same issue slightly different flake

rootulp:
[...]

PR #3413, opened 2024-04-30
fix(ci): update from published to released trigger for deployment epic

MSevey:
## Overview

I realized that the `published` event is triggered by pre-releases too.
The `released` event is only triggered when we set the release to the latest release, which is what we want to track for deployments.


MSevey:
@rootulp is this a known issue: https://github.com/celestiaorg/celestia-app/actions/runs/8895827778/job/24427042390?pr=3413#step:4:35

rootulp:
Yes but thanks for flagging: https://github.com/celestiaorg/celestia-app/issues/2942

staheri14 (approved):

Issue #3187, opened 2024-03-18
TestMinorVersionCompatibility broken


cmwaters:
### Summary

This test currently panics in InitGenesis when it is run

### Problem Definition

f

### Proposal

f

### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

rootulp:
Hmm these tests pass locally for me

rootulp:
This issue needs setup information on Knuu, see [Notion](https://www.notion.so/celestiaorg/Knuu-94e5ce0fa4764803b4333deb23893fba?p=0848a70da653413c848ac1ba6f5e3957&pm=s).

Blocked on https://github.com/celestiaorg/celestia-app/issues/3255

ninabarbakadze:
the port-forwarding error in this test is caused by the pod getting deleted after the upgrade. The pod spins up again but we're not forwarding port to the said pod. more context from my conversation with devops: https://celestia-team.slack.com/archives/C02PEA73QM6/p1713355805981239

PR #3414, opened 2024-04-30
fix: MinorVersionCompatibility test

smuu:
## Overview

This fixes the port forwarding issues nina faces

cc: @ninabarbakadze 

Closes https://github.com/celestiaorg/celestia-app/issues/3187

### DIFF
--- a/test/e2e/main.go
+++ b/test/e2e/main.go
@@ -24,8 +24,7 @@ func main() {
 	logger := log.New(os.Stdout, "test-e2e", log.LstdFlags)
 
 	tests := []Test{
-		// FIXME the following test is currently failing.
-		// {"MinorVersionCompatibility", MinorVersionCompatibility},
+		{"MinorVersionCompatibility", MinorVersionCompatibility},
 		{"MajorUpgradeToV2", MajorUpgradeToV2},
 		{"E2ESimple", E2ESimple},
 	}
--- a/test/e2e/check_upgrades.go
+++ b/test/e2e/check_upgrades.go
@@ -20,10 +20,6 @@ import (
 	"github.com/tendermint/tendermint/rpc/client/http"
 )
 
-const (
-	txsimVersion = "a92de72"
-)
-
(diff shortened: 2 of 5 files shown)


test/e2e/main.go
```diff
@@ -24,8 +24,7 @@ func main() {
 	logger := log.New(os.Stdout, "test-e2e", log.LstdFlags)
 
 	tests := []Test{
-		// FIXME the following test is currently failing.
-		// {"MinorVersionCompatibility", MinorVersionCompatibility},
```
rootulp, line 28: you're a beast @smuu thanks so much!

rootulp (approved):
Thanks! I think you can update the PR description to state this PR closes https://github.com/celestiaorg/celestia-app/issues/3187

Also, I think golangci-lint is complaining so we prob need to fix that before merging but we really appreciate this fix

ninabarbakadze (dismissed):
thank you so much! I can resolve the lint issues for you if you like

smuu:
Adding some context/explanation:

Using port-forwarding in Kubernetes is error-prone, and I try to avoid it whenever possible.

In this case, the problem is:
- We start an Instance
- We create a port-forward
- We replace the running Pod with another Pod with another version without closing the port-forward
- We try to use the still existing port-forward where the target changed and therefore fails

The workaround I propose does not use port-forwarding. Instead, it executed a shell command on another Instance (Executer) to fetch the height from the RPC.
I understand this is not optimal, as using the RPC client inside app makes more sense than shell scripts. If more resources are available, we will see if we can solve it in knuu.

I hope this workaround is sufficient enough for the time beeing.

smuu:
> thank you so much! I can resolve the lint issues for you if you like

That would be great!

ninabarbakadze (commented):
everything should be green now @smuu @rootulp

rootulp (approved):

ninabarbakadze (approved):

PR #3423, opened 2024-05-02
feat: enable querying state through grpc

cmwaters:
We seemed to have forgotten to add the query function which means `ABCIQuery` wasn't working even although the grpc endpoint was exposed


renaynay (approved):
👑

liamsi (approved):
nil

rootulp (approved):

evan-forbes (approved):

Issue #3259, opened 2024-04-04
Support multiple keys within signer


cmwaters:
### Summary

Kinda doesn't need one

### Problem Definition

Currently, the `signer` can only be set with a single address that it uses to sign all the transactions. Some users may want to submit `sdk.Msg`s with different accounts. This is the case with Raas providers who may want a dedicated account per rollup.

### Proposal

Each transaction has a concept of signers. What is needed is that the signer needs to be read from the transaction and then checked whether the keyring has it. If so we sign with that account. We still won't support transactions with multiple signers but this is a rare requirement. For PFBs the signer will need to be provided. It may be possible to do this as an optional argument (especially if we know that the keyring only has a single signer).

PR #3433, opened 2024-05-06
feat: add multi-account support

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/3259

This PR adds support for using multiple kerying keys with a single Signer. It splits signer into a purely offline signer for signing messages and the txclient responsible for tx submission and handling


pkg/user/tx_client.go
```diff
@@ -0,0 +1,412 @@
...
+	accounts := make([]*Account, len(records))
+	for idx, record := range records {
+		addr, err := record.GetAddress()
+		if err != nil {
+			return nil, err
+		}
+		accNum, seqNum, err := QueryAccount(ctx, conn, encCfg.InterfaceRegistry, addr)
+		if err != nil {
+			return nil, fmt.Errorf("querying account: %w", err)
+		}
```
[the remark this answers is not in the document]
cmwaters: They should be able to add keys after setting up the tx client. If it's the first time the key is being used, the txclient will automatically query for the sequence number etc... before submitting the transactions.

app/test/check_tx_test.go
```diff
@@ -49,6 +48,7 @@ func TestCheckTx(t *testing.T) {
 			name:      "normal transaction, CheckTxType_New",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
+				// TODO might have to change back to 1
```
rootulp, line 51: change what back to 1?

app/test/check_tx_test.go
```diff
@@ -160,8 +160,9 @@ func TestCheckTx(t *testing.T) {
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
 				signer := createSigner(t, kr, accs[8], encCfg.TxConfig, 9)
-				_, blobs := blobfactory.RandMsgPayForBlobsWithSigner(tmrand.NewRand(), signer.Address().String(), 1_000_000, 1)
-				tx, err := signer.CreatePayForBlob(blobs, opts...)
+				_, blobs := blobfactory.RandMsgPayForBlobsWithSigner(tmrand.NewRand(), signer.Account(accs[8]).Address().String(), 1_000_000, 1)
+				// TODO: go back to the sequences here and handle it
```
rootulp, line 164: is this TODO still relevant?

pkg/proof/proof_test.go
```diff
@@ -194,15 +194,15 @@ func TestNewShareInclusionProof(t *testing.T) {
 		},
 		{
 			name:          "blob shares for first namespace",
-			startingShare: 55,
-			endingShare:   57,
+			startingShare: 56,
+			endingShare:   58,
```
rootulp, line 198: [no change needed] do you have any idea why these blob shares got move over by one share? Perhaps something in the tx namespace occupied one extra share?

pkg/user/signer.go
```diff
@@ -282,300 +126,133 @@ func (s *Signer) DecodeTx(txBytes []byte) (authsigning.Tx, error) {
...
-	if _, exists := s.outboundSequences[sequence]; exists {
-		return s.retryBroadcastingTx(ctx, txBytes, sequence+1)
-	}
-
-	if sequence < s.networkSequence {
-		s.localSequence = s.networkSequence
-		return s.retryBroadcastingTx(ctx, txBytes, s.localSequence)
-	}
+// Account returns an account of the signer from the key name. Also returns a bool if the
+// account exists.
```
rootulp, line 135:
```suggestion
// Account returns an account of the signer from the key name.
```

x/blob/types/estimate_gas_test.go
```diff
@@ -40,13 +40,13 @@ func TestPFBGasEstimation(t *testing.T) {
 		t.Run(fmt.Sprintf("case %d", idx), func(t *testing.T) {
 			accnts := testfactory.GenerateAccounts(1)
 			testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultInitialConsensusParams(), accnts...)
-			addr := testfactory.GetAddress(kr, accnts[0])
-			signer, err := user.NewSigner(kr, nil, addr, encCfg.TxConfig, testutil.ChainID, 1, 0, appconsts.LatestVersion)
+			// addr := testfactory.GetAddress(kr, accnts[0])
```
rootulp, line 43:
```suggestion
```

rootulp (dismissed):

evan-forbes (approved):
nice! its been cool to see this code evolve. can't wait to play with it

pkg/user/signer.go
```diff
@@ -1,260 +1,104 @@
...
-	networkSequence uint64
-	// gasMultiplier is used to increase gas limit as it is sometimes underestimated
-	gasMultiplier float64
-	// lookup map of all pending and yet to be confirmed outbound transactions
-	outboundSequences map[uint64]struct{}
-	// a reverse map for confirming which sequence numbers have been committed
-	reverseTxHashSequenceMap map[string]uint64
+	// set of accounts that the signer can manage. Should match the keys on the keyring
+	accounts            map[string]*Account
+	addressToAccountMap map[string]string
```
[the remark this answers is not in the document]
cmwaters: Turned this into an issue because I don't believe we have one currently

pkg/user/tx_client.go
```diff
@@ -0,0 +1,461 @@
...
+		resp, err := txClient.GetTx(ctx, &sdktx.GetTxRequest{Hash: txHash})
+		if err == nil {
+			if resp.TxResponse.Code != 0 {
+				return resp.TxResponse, fmt.Errorf("tx was included but failed with code %d: %s", resp.TxResponse.Code, resp.TxResponse.RawLog)
+			}
+			return resp.TxResponse, nil
+		}
+		// FIXME: this is a relatively brittle of working out whether to retry or not. The tx might be not found for other
+		// reasons. It may have been removed from the mempool at a later point. We should build an endpoint that gives the
+		// signer more information on the status of their transaction and then update the logic here
```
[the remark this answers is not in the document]
cmwaters: This will be resolved in the future with `TxStatus`

pkg/proof/proof_test.go
cmwaters: Yeah I was actually curious that this needed to change. I looked through the code and couldn't see why the PFBs would take up more space. One thing I think we should add if we don't have already is regression tests to the square / share construction to make sure that breaking changes are caught

app/test/check_tx_test.go
cmwaters: I don't think so, I think nina might have added it later

rootulp (approved):

evan-forbes (approved):

PR #3446, opened 2024-05-08
fix: tokenfilter test

### LINKED ISSUE #3238: Fix tokenfilter tests
### Summary

Right now our token filter tests have been commented out:

https://github.com/celestiaorg/celestia-app/blob/897b448cb8d2b655743bf1e786f6823775468964/test/tokenfilter/tokenfilter_test.go#L135-L141

The reason being that the app version isn't passed in the header (so the tests simply panic).

We should fix this and re-enable the tests. Currently they rely on the `testing` package provided by the ibc-go repo. It's possible to suggest the fix there. The problem is that we would want the changes backported to the v6.x branch which is no longer supported (only v7 and v8 are supported). I would prefer not to have to fork another repo so the options would be to:

1. Rewrite the tests using `interchaintest` (the docker image based one that strangelove maintain)
2. Upgrade to v8 in v2 and submit the patch that supports different app versions

ninabarbakadze:
## Overview

Fixes #3238

### DIFF
--- a/test/tokenfilter/setup.go
+++ b/test/tokenfilter/setup.go
@@ -2,6 +2,7 @@ package tokenfilter
 
 import (
 	"encoding/json"
+	// "fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/tokenfilter/setup.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/celestiaorg/celestia-app/v2/app/encoding"
 	"github.com/celestiaorg/celestia-app/v2/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v2/test/util/testnode"
+
```
rootulp, line 13: I think this newline is unintentional

```suggestion
```

test/tokenfilter/tokenfilter_test.go
```diff
@@ -136,6 +136,5 @@ func TestTokenFilterTestSuite(t *testing.T) {
 	// FIXME: this is because the ibctesting framework we use
 	// doesn't set the version in the header which we require
 	// to know which state machine to execute against
```
rootulp, line 138: [blocking] this comment appears no longer relevant

rootulp (approved):
Thanks!

evan-forbes (approved):

PR #3459, opened 2024-05-10
feat: export BroadcastPayForBlob to support more async blob submission.

evan-forbes:
## Overview

This PR exports an existing function to block on submitting a PFB to the mempool, but not block on getting the tx included in a block.


pkg/user/signer.go
```diff
@@ -189,15 +189,19 @@ func (s *Signer) SubmitTx(ctx context.Context, msgs []sdktypes.Msg, opts ...TxOp
...
 	}
 
 	return s.ConfirmTx(ctx, resp.TxHash)
 }
 
-func (s *Signer) broadcastPayForBlob(ctx context.Context, blobs []*blob.Blob, opts ...TxOption) (*sdktypes.TxResponse, error) {
+// BroadcastPayForBlob forms a transaction from the provided blobs, signs it,
+// and broadcasts it to the chain. TxOptions may be provided to set the fee and
+// gas limit. This function does not block on the tx actually being included in
+// a block.
```
evan-forbes, line 203:
```suggestion
// gas limit. This function does not block on the tx actually being included in
// a block, only on being included in the mempool.
```

rach-id (approved):

rootulp (approved):

Issue #3378, opened 2024-04-24
`NodeInfo.protocol_version.app: 1` after upgrade to app version 2


rootulp:
## Context

I'm trying to manually test https://github.com/celestiaorg/celestia-app/pull/3310 by submitting blobstream txs in app version v1 and v2.

## Problem

I made a slight modification to the `./single-node.sh` script:

```diff
+# Override the genesis to use app version 1 and then upgrade to app version 2 later.
+sed -i'.bak' 's#"app_version": "2"#"app_version": "1"#g' "${CELESTIA_APP_HOME}"/config/genesis.json

# Start celestia-app
echo "Starting celestia-app..."
celestia-appd start \
  --home ${CELESTIA_APP_HOME} \
  --api.enable \
  --grpc.enable \
  --grpc-web.enable \
+  --v2-upgrade-height 3
```

When I run the script, I observe that the node appears to upgrade at the end of block height 2 because:

```
1:05PM INF executed block height=2 module=state num_invalid_txs=0 num_valid_txs=0
1:05PM INF adding a new module: interchainaccounts
1:05PM INF created new capability module=ibc name=ports/icahost
1:05PM INF port binded module=x/ibc/port port=icahost
1:05PM INF claimed capability capability=2 module=icahost name=ports/icahost
1:05PM INF adding a new module: minfee
1:05PM INF adding a new module: packetfowardmiddleware
1:05PM INF adding a new module: signal
```

However, even after block height 3, when I query for status I observe `NodeInfo.protocol_version.app: 1`

```
$ celestia-appd status

{
  "NodeInfo": {
    "protocol_version": {
      "p2p": "8",
      "block": "11",
      "app": "1"
    },
    "id": "a7a360053ddc46c7ce9c8534ca9cab8b051db880",
    "listen_addr": "tcp://0.0.0.0:26656",
    "network": "private",
    "version": "1.0.0-rc0-354-g32accbaa",
    "channels": "40202122233038606100",
    "moniker": "private",
    "other": {
      "tx_index": "on",
      "rpc_address": "tcp://0.0.0.0:26657"
    }
  },
  "SyncInfo": {
    "latest_block_hash": "04F2786B5CC8DF780771E521F7DF0CBEA245524154201104522E91A3E5EB11E5",
    "latest_app_hash": "4B4014F397D62A797E0C48C12D7A2823F3190E3491659060BB59F9CD4287BE35",
    "latest_block_height": "19",
[...]

PR #3469, opened 2024-05-12
chore(deps): upgrade to celestia-core v1.36.1-tm-v0.34.29

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3378

## Testing

I updated single-node.sh to use an upgrade height so that I could verify `celestia-appd status` reports:

Block height | App version
--- | ---
1 | 1
2 | 1
3 | 2
4 | 2

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -85,10 +85,14 @@ sed -i'.bak' 's#discard_abci_responses = true#discard_abci_responses = false#g' 
 # Override the VotingPeriod from 1 week to 1 minute
 sed -i'.bak' 's#"604800s"#"60s"#g' "${CELESTIA_APP_HOME}"/config/genesis.json
 
+# Override the genesis to use app version 1 and then upgrade to app version 2 later.
+sed -i'.bak' 's#"app_version": "2"#"app_version": "1"#g' "${CELESTIA_APP_HOME}"/config/genesis.json
+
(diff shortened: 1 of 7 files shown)


scripts/single-node.sh
```diff
@@ -85,10 +85,14 @@ sed -i'.bak' 's#discard_abci_responses = true#discard_abci_responses = false#g'
...
+
 # Start celestia-app
 echo "Starting celestia-app..."
 celestia-appd start \
   --home "${CELESTIA_APP_HOME}" \
   --api.enable \
   --grpc.enable \
-  --grpc-web.enable
+  --grpc-web.enable \
+  --v2-upgrade-height 3
```
rach-id, line 98: Why do we want to have these changes here? is there a reason to upgrade in the `single-node.sh`?

rach-id (commented):
LGTM, left one question to understand

scripts/single-node.sh
rootulp: Mostly for debugging purposes. IMO this is how arabica, mocha, and mainnet will behave so we should be testing it locally.

It seems non representative of reality to use a genesis file that has app version 2 in it without an upgrade height.

rach-id (approved):

evan-forbes (approved):

Issue #3462, opened 2024-05-10
`panic: loading latest version: failed to load latest version: version of store qgb mismatch root store's version; expected 1 got 0`


rootulp:
## Problem

I can't start a node, stop it, and start it again on current `main`

```
$ celestia-appd start
1:02PM INF starting node with ABCI Tendermint in-process
1:02PM INF service start impl=multiAppConn module=proxy msg={}
1:02PM INF service start connection=query impl=localClient module=abci-client msg={}
1:02PM INF service start connection=snapshot impl=localClient module=abci-client msg={}
1:02PM INF service start connection=mempool impl=localClient module=abci-client msg={}
1:02PM INF service start connection=consensus impl=localClient module=abci-client msg={}
1:02PM INF service start impl=EventBus module=events msg={}
1:02PM INF service start impl=PubSub module=pubsub msg={}
1:02PM INF service start impl=IndexerService module=txindex msg={}
panic: loading latest version: failed to load latest version: version of store qgb mismatch root store's version; expected 1 got 0

goroutine 8 [running]:
github.com/celestiaorg/celestia-app/v2/app.(*App).Info(0x140012ba008, {{0x1041410ea?, 0x14001033748?}, 0x1027e7ab4?, 0x0?})
	/Users/rootulp/git/rootulp/celestiaorg/celestia-app/app/app.go:528 +0x370
github.com/tendermint/tendermint/abci/client.(*localClient).InfoSync(0x14000113020, {{0x1041410ea?, 0x14001033878?}, 0x1027dc1e8?, 0x10413f3f6?})
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.35.0-tm-v0.34.29/abci/client/local_client.go:250 +0x10c
github.com/tendermint/tendermint/proxy.(*appConnQuery).InfoSync(0x14001728560?, {{0x1041410ea?, 0x40?}, 0x12dfdce68?, 0x70?})
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.35.0-tm-v0.34.29/proxy/app_conn.go:170 +0x2c
github.com/tendermint/tendermint/consensus.(*Handshaker).HandshakeWithContext(0x14001033c90, {0x105375888, 0x1067bb940}, {0x10538a4f0, 0x140012ad860})
[...]

PR #3477, opened 2024-05-13
fix: restart node for chain that starts on v2

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3462

## Testing

```shell
./scripts/single-node.sh
# wait a block then stop the node with CTRL + C
celestia-appd start
```

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -87,6 +87,7 @@ import (
 	tmjson "github.com/tendermint/tendermint/libs/json"
 	"github.com/tendermint/tendermint/libs/log"
 	tmos "github.com/tendermint/tendermint/libs/os"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	dbm "github.com/tendermint/tm-db"
 )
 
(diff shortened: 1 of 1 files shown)


ninabarbakadze (approved):

app/app.go
```diff
@@ -547,13 +548,21 @@ func (app *App) InitChain(req abci.RequestInitChain) (res abci.ResponseInitChain
...
+		app.mountKeysAndInit(appVersion)
 	}
 
-	return app.BaseApp.InitChain(req)
+	res = app.BaseApp.InitChain(req)
+
+	ctx := app.NewContext(false, tmproto.Header{})
+	if appVersion != v1 {
+		app.SetInitialAppVersionInConsensusParams(ctx, appVersion)
+		app.SetAppVersion(ctx, appVersion)
```
evan-forbes, line 563: unrelated comment:

setting up context for state access is a good idea in general, but when we have to initialize with an empty header it feels so broken since we are now relying on everything downstream of that context to not need or get data from that header. I don't like it lol.

evan-forbes (approved):

app/app.go
rootulp: +1 don't like it. I couldn't see an alternative to save consensus params that doesn't use a context so not sure how to avoid it.

cmwaters: Yeah I agree, it's not great.

PR #3496, opened 2024-05-17
chore: extends single-node.sh script to support local tracing and traced data retrieval

staheri14:
This PR extends the single-node.sh script by providing an option for the user to enable local tracing and retrieve traced data.

**How to test**
1. Run `single-node.sh` under the `scripts` folder
2. Respond `y` to `Do you want to set up local tracing with the ability to pull traced data? [y/n]`
3. Once the node starts, navigate to ./celestia-app/data/traces folder and you will find the locally traced files.

Also, the trace package from celestia-core has a method to pull the traced data which can be found  [here](https://github.com/celestiaorg/celestia-core/blob/a281e871c70f4c8bd4f6ab7c7e39e28c715d65b2/pkg/trace/fileserver.go#L102).

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -101,7 +101,7 @@ if [[ $response == "y" ]]; then
   sed -i.bak -e "s/^trace_pull_address *=.*/trace_pull_address = \"$trace_pull_address\"/" ${CELESTIA_APP_HOME}/config/config.toml
   trace_push_batch_size=1000
   sed -i.bak -e "s/^trace_push_batch_size *=.*/trace_push_batch_size = \"$trace_push_batch_size\"/" ${CELESTIA_APP_HOME}/config/config.toml
-  echo "Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:${trace_pull_address}"
+  echo "Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1${trace_pull_address}"
 fi
 
 # Start celestia-app
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -91,6 +91,19 @@ sed -i'.bak' 's#"604800s"#"60s"#g' "${CELESTIA_APP_HOME}"/config/genesis.json
 # Override the genesis to use app version 1 and then upgrade to app version 2 later.
 sed -i'.bak' 's#"app_version": "2"#"app_version": "1"#g' "${CELESTIA_APP_HOME}"/config/genesis.json
 
+
+echo "Do you want to set up local tracing with the ability to pull traced data? [y/n]"
+read -r response
+if [[ $response == "y" ]]; then
+  trace_type="local"
+  sed -i.bak -e "s/^trace_type *=.*/trace_type = \"$trace_type\"/" ${CELESTIA_APP_HOME}/config/config.toml
+  trace_pull_address=":26661"
+  sed -i.bak -e "s/^trace_pull_address *=.*/trace_pull_address = \"$trace_pull_address\"/" ${CELESTIA_APP_HOME}/config/config.toml
+  trace_push_batch_size=1000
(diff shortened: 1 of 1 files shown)


scripts/single-node.sh
```diff
@@ -91,6 +91,19 @@ sed -i'.bak' 's#"604800s"#"60s"#g' "${CELESTIA_APP_HOME}"/config/genesis.json
...
+echo "Do you want to set up local tracing with the ability to pull traced data? [y/n]"
+read -r response
+if [[ $response == "y" ]]; then
+  trace_type="local"
+  sed -i.bak -e "s/^trace_type *=.*/trace_type = \"$trace_type\"/" ${CELESTIA_APP_HOME}/config/config.toml
+  trace_pull_address=":26661"
+  sed -i.bak -e "s/^trace_pull_address *=.*/trace_pull_address = \"$trace_pull_address\"/" ${CELESTIA_APP_HOME}/config/config.toml
+  trace_push_batch_size=1000
+  sed -i.bak -e "s/^trace_push_batch_size *=.*/trace_push_batch_size = \"$trace_push_batch_size\"/" ${CELESTIA_APP_HOME}/config/config.toml
+  echo "Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:${trace_pull_address}"
```
rootulp, line 104: this prints with two colons so I think we want to remove one:

```
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1::26661
```

```suggestion
  echo "Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1${trace_pull_address}"
```

rootulp (dismissed):
silly question but how do I pull the traced data? Do I need to have Grafana + Prometheus running locally and configure it to fetch from `http://127.0.0.1:26661`?

[not blocking] but it would be cool if we had a script (perhaps in this one or a separate one) to launch a local Grafana + Prometheus that is already configured to fetch traced data from that endpoint. Alternatively, if we can use Grafana Cloud and not run Grafana + Prometheus locally, it would be nice if there was a step or screenshot on how to configure Grafana Cloud to read traced data from the single node running via this script.

staheri14:
> silly question but how do I pull the traced data? Do I need to have Grafana + Prometheus running locally and configure it to fetch from `http://127.0.0.1:26661`?
> 
> [not blocking] but it would be cool if we had a script (perhaps in this one or a separate one) to launch a local Grafana + Prometheus that is already configured to fetch traced data from that endpoint. Alternatively, if we can use Grafana Cloud and not run Grafana + Prometheus locally, it would be nice if there was a step or screenshot on how to configure Grafana Cloud to read traced data from the single node running via this script.

Sure, I'll create an issue for this one, but, in the meanwhile, I have updated the PR description and explained how to locate the local traced data.

scripts/single-node.sh
staheri14: Thanks, fixed in https://github.com/celestiaorg/celestia-app/pull/3496/commits/3982f8838e51c908ad0df85dc3a362bb454158c3

staheri14:
@rootulp The trace package from the celestia-core has a method to pull the traced data https://github.com/celestiaorg/celestia-core/blob/a281e871c70f4c8bd4f6ab7c7e39e28c715d65b2/pkg/trace/fileserver.go#L102

rootulp:
Cool thanks, asked out of curiosity :)

rootulp (approved):
The log looks fixed to me: `Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661`.

Tangent: I would also be okay with always enabling tracing and getting rid of the conditional prompt for `y/n`

rach-id (approved):

staheri14:
> The log looks fixed to me: `Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661`.
> 
> Tangent: I would also be okay with always enabling tracing and getting rid of the conditional prompt for `y/n`

The amount of traced data can become quite large, so I think it is prudent to let the user decide (note that local tracing by default is disabled).

Issue #3342, opened 2024-04-18
pkg/user: Signer automatically estimates gas


renaynay:
https://github.com/celestiaorg/celestia-node/pull/3298#discussion_r1568997897

cmwaters:
For simple PFBs we should use the modelled formula for estimating gas. For any other transaction we should use the `EstimateGas` function.

We should have a gas multiplier field which we multiply by when using `EstimateGas` (as it occasionally underestimates). This should be adjustable with a default of 1.1

PR #3510, opened 2024-05-29
feat: add auto gas and fee estimations

vgonkivs:
Closes https://github.com/celestiaorg/celestia-app/issues/3342

## Overview
Some points to consider:
* The current version of EstimateGas utilizes `Simulate` that brings additional roundtrips to the estimation process. As we discussed internally with @cmwaters we should try to get rid of roundtrips in favour of static gas estimations(for different types of messages), as it has been done for blob estimation.
* ideally I'd add two different Fee handlers to the txClient that estimates gas/fee for PFB transactions and all other types of txs. But it will make sense after reaching the proposal described in 1.


pkg/user/tx_client.go
```diff
@@ -166,83 +170,124 @@ func SetupTxClient(
...
 	}
 
-	txBytes, err := s.signer.EncodeTx(tx)
+	estimated := false
+	for _, coin := range txBuilder.GetTx().GetFee() {
+		if coin.Denom == appconsts.BondDenom {
+			estimated = true
+			break
+		}
+	}
```
cmwaters, line 256: So if I understand this correctly, if the user hasn't set the fee then the `GetFee()` would return an empty array?

pkg/user/tx_client.go
```diff
@@ -166,83 +170,124 @@ func SetupTxClient(
...
 	}
 
-	tx, account, _, err := s.signer.SignTx(msgs, opts...)
+	txBuilder, err := client.signer.txBuilder(msgs, opts...)
 	if err != nil {
 		return nil, err
 	}
 
-	txBytes, err := s.signer.EncodeTx(tx)
+	estimated := false
```
cmwaters, line 250:
```suggestion
	isFeeEstimated := false
```
For readability it may be better to call this feeEstimated (since there is also gas estimating)

cmwaters: Actually this should be the other way around. Something like `hasUserSetFee` right?

pkg/user/tx_client.go
```diff
@@ -166,83 +170,124 @@ func SetupTxClient(
...
+		gasLimit, err = client.estimateGas(ctx, txBuilder)
+		if err != nil {
+			return nil, err
+		}
+		txBuilder.SetGasLimit(gasLimit)
+	}
+
+	if !estimated {
+		fee := int64(math.Ceil(appconsts.DefaultMinGasPrice * float64(gasLimit)))
+		txBuilder.SetFeeAmount(sdktypes.NewCoins(sdktypes.NewCoin(appconsts.BondDenom, sdktypes.NewInt(fee))))
```
cmwaters, line 273: I think there's another option like `SetFee` where you don't need to specify the `BondDenom`

pkg/user/tx_client.go
```diff
@@ -361,81 +408,83 @@ func (s *TxClient) ConfirmTx(ctx context.Context, txHash string) (*sdktypes.TxRe
 	}
 }
 
-func (s *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, opts ...TxOption) (uint64, error) {
-	s.mtx.Lock()
-	defer s.mtx.Unlock()
+func (client *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, opts ...TxOption) (uint64, error) {
```
cmwaters, line 411: nit: maybe we should document that the gas estimate we get is always multiplied by the multiplier in case that's not clear

pkg/user/tx_client_test.go
```diff
@@ -34,103 +34,139 @@ type TxClientTestSuite struct {
...
 }
 
-func (s *TxClientTestSuite) SetupSuite() {
-	s.encCfg = encoding.MakeConfig(app.ModuleEncodingRegisters...)
-	s.ctx, _, _ = testnode.NewNetwork(s.T(), testnode.DefaultConfig().WithFundedAccounts("a"))
-	_, err := s.ctx.WaitForHeight(1)
-	s.Require().NoError(err)
-	s.txClient, err = user.SetupTxClient(s.ctx.GoContext(), s.ctx.Keyring, s.ctx.GRPCClient, s.encCfg, user.WithGasMultiplier(1.2))
-	s.Require().NoError(err)
+func (testClient *TxClientTestSuite) SetupSuite() {
```
cmwaters, line 37: nit: It's more a `testSuite` than a `testClient`

pkg/user/tx_client.go
```diff
@@ -166,83 +170,124 @@ func SetupTxClient(
...
-	txBytes, _, err := s.signer.CreatePayForBlobs(account, blobs, opts...)
+	blobSizes := make([]uint32, len(blobs))
+	for i, blob := range blobs {
+		blobSizes[i] = uint32(len(blob.Data))
+	}
+
+	gasLimit := uint64(float64(types.DefaultEstimateGas(blobSizes)) * client.gasMultiplier)
+	fee := uint64(math.Ceil(appconsts.DefaultMinGasPrice * float64(gasLimit)))
+	// prepend calculated params, so it can be overwritten in case the user has specified it.
+	opts = append([]TxOption{SetGasLimit(gasLimit), SetFee(fee)}, opts...)
```
cmwaters, line 212: This currently breaks down if someone wants us to estimate the gas but wants to provide a gas price themselves (if there's congestion for example). We should create an issue and look to address that in the future

pkg/user/tx_client.go
```diff
@@ -166,83 +170,124 @@ func SetupTxClient(
...
+	estimated := false
+	for _, coin := range txBuilder.GetTx().GetFee() {
+		if coin.Denom == appconsts.BondDenom {
+			estimated = true
+			break
+		}
+	}
+
+	gasLimit := txBuilder.GetTx().GetGas()
+	if float32(gasLimit) < appconsts.DefaultMinGasPrice {
```
cmwaters, line 259: Can you explain you're reasoning here? So if someone submits a transaction with a gas limit that is less than the default min gas price? Shouldn't it be if the gasLimit is 0?

cmwaters (commented):
Good work! Added a few comments

pkg/user/tx_client.go
vgonkivs: Correct. But if the user wants to set a 0 fee, then he has to explicitly set it(was confirmed by @ninabarbakadze)

pkg/user/tx_client.go
vgonkivs: `hasUserSetFee` sgtm

pkg/user/tx_client.go
vgonkivs: The option you've mentioned is a TxOption, that invokes `builder.SetFeeAmount `with utia as denomination under the hood.

pkg/user/tx_client.go
vgonkivs: It covers the case when the user provides a gasLimit that is less than `DefaultMinGasPrice`, so the tx will be failed. I think it could be useful if at some point we decide to update MinGasPrice.

pkg/user/tx_client.go
vgonkivs: This case works fine, as the gas will be estimated but the calculated gas price will be overwritten. The issue might be if the user wants us to estimate gas price, providing a gas limit. But in this case, I guess we will get an insufficient fee error that is already handled, so the tx will be retried with higher fees. Additionally, I can check the gas limit before estimating.

pkg/user/tx_client.go
cmwaters: Makes sense. It could be simpler to just check if length of fee is greater than zero (since only tia is supported)

pkg/user/tx_client.go
cmwaters: Ah I get it.

pkg/user/tx_client.go
vgonkivs: My motivation was to check the denomination because:
1) ALL values that are greater or equal to zero are valid;
2) the user may mistakenly use an invalid denomination. So, we'll catch it and re-estimate;

pkg/user/tx_client.go
cmwaters: Yeah it doesn't quite work as would be intuitive. Ideally, I as a user could set the gas price and request that the client estimates the gas limit and sets the fee by multiplying the gas price I provided with the gas limit that was estimated. This is a bit tricky to do so I think we should follow it up later.

vgonkivs:
Generally, I'd make a FeeHandler `middleware` that can intercept all requests that are coming to the Client and perform all needed computations.

cmwaters (approved):
LGTM

cmwaters:
> Generally, I'd make a FeeHandler `middleware` that can intercept all requests that are coming to the Client and perform all needed computations.

Can you elaborate on this?

vgonkivs:
Can create an issue, describing all ideas so we can have a clear understanding of what should be done and do we need it.

cmwaters:
Yes let's create an issue and discuss it further there

evan-forbes (approved):
LGTM

ninabarbakadze (approved):
lgtm. thanks!

Issue #3509, opened 2024-05-28
Extend txsim to send all state machine messages


ninabarbakadze:
## Description

Every time we add a new module we should add it to txsim

## Acceptance Criteria

Issue #2535, opened 2023-09-18
Refactor TestApp


evan-forbes:
iirc, we essentially just took our TestApp from the simapp in the sdk. The api for using this test application could be greatly improved to provide test writers with fine grained and each control over the ABCI. This should include easily spinning up a custom application, creating new blocks, execution transactions, querying state, a mock mempool, and even manual state manipulation. The overall goal should be to create a tool taht allows us to replace many of our integration tests, which we have used mainly because we don't have a good test app api

blocking #2533

evan-forbes:
either with testapp or even with testnode, we might want to consider using libraries such as https://github.com/informalsystems/CometMock

PR #3513, opened 2024-05-29
test: consistent appHash between commits

### LINKED ISSUE #2414: Non-determinism unit test
## Description

Write a non-determinism unit test for the state machine. The test will involve creating a set of transactions, executing them, and comparing the resulting apphash with an expected value.

## Acceptance Criteria

- [ ] Submit and execute txs using txsim
- [ ] Should cover v1 and v2
- [ ] Assert expected state and data roots

ninabarbakadze:
## Overview

Fixes #2414 

### This test is split up in a few parts:

- Execute transactions deterministically on v1 and save the AppHash. https://github.com/celestiaorg/celestia-app/pull/3522 ([includes genesis folder backport from utils](https://github.com/celestiaorg/celestia-app/pull/3520))
- Set up the same testApp environment on main, execute transactions and assert that the AppHash matches the one from v1

### Verifying the AppHash: 

If you want to make sure that the expected AppHash is correct, you can run the [non determinism test on v1.x](https://github.com/celestiaorg/celestia-app/pull/3522) that generates it.

### Follow-ups:
#3509 
https://github.com/celestiaorg/celestia-app/issues/3540


rach-id (dismissed):
LGTM, left some optional nits

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// TestNonDeterminismBetweenMainAndV1 executes a set of different transactions,
+// produces an app hash and compares it with the app hash produced by v1.x
+func TestNonDeterminismBetweenMainAndV1(t *testing.T) {
+	const (
+		numBlobTxs, numNormalTxs = 5, 5
+	)
+
+	expectedAppHash := []byte{100, 237, 125, 126, 116, 10, 189, 82, 156, 116, 176, 136, 169, 92, 185, 12, 72, 134, 254, 175, 234, 13, 159, 90, 139, 192, 190, 248, 67, 9, 32, 217}
```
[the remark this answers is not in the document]
ninabarbakadze: good point i'll link the executable from v1.x in the comments

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	// Commit the state
+	testApp.Commit()
+
+	// Get the app hash
+	appHash := testApp.LastCommitID().Hash
+
+	// Assert that the app hash is equal to the app hash produced by v1.x
+	require.Equal(t, expectedAppHash, appHash)
```
rootulp, line 92:
```suggestion
	assert.Equal(t, expectedAppHash, appHash)
```

rootulp (commented):
[future issue / PR] This is a good starting point. I think we could improve this test by executing transaction types that include more unique messages because AFAIU this only tests MsgSend, MsgPayForBlob.

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+		"fee vapor thing fish fan memory negative raven cram win quantum ozone job mirror shoot sting quiz black apart funny sort cancel friend curtain",
+		"skin beef review pilot tooth act any alarm there only kick uniform ticket material cereal radar ethics list unlock method coral smooth street frequent",
+		"ecology scout core guard load oil school effort near alcohol fancy save cereal owner enforce impact sand husband trophy solve amount fish festival sell",
+		"used describe angle twin amateur pyramid bitter pool fluid wing erode rival wife federal curious drink battle put elbow mandate another token reveal tone",
+		"reason fork target chimney lift typical fine divorce mixture web robot kiwi traffic stove miss crane welcome camp bless fuel october riot pluck ordinary",
+		"undo logic mobile modify master force donor rose crumble forget plate job canal waste turn damp sure point deposit hazard quantum car annual churn",
+		"charge subway treat loop donate place loan want grief leg message siren joy road exclude match empty enforce vote meadow enlist vintage wool involve",
+	}
+	kb := keyring.NewInMemory(cdc)
```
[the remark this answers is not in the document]
ninabarbakadze: I was following the convention from our code https://github.com/celestiaorg/celestia-app/blob/fd4b75bb6145da14bdb62476ff065ade9530d0ea/test/util/testnode/utils.go#L45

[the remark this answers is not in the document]
ninabarbakadze: this function is pretty much identical

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+		addresses = append(addresses, rec.Name)
+	}
+
+	// Apply genesis state to the app
+	_, _, err = testutil.ApplyGenesisState(testApp, pubKeys, 1_000_000_000, app.DefaultInitialConsensusParams())
+	require.NoError(t, err)
+
+	accinfos := queryAccountInfo(testApp, addresses, kr)
+
+	// Create a set of 10 deterministic sdk transactions
```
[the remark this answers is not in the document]
ninabarbakadze: not if you have any randomness like a random namespace every time a test is run. I just wanted to imply that in the comment for anyone debugging this test in the future.

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+	expectedAppHash := []byte{100, 237, 125, 126, 116, 10, 189, 82, 156, 116, 176, 136, 169, 92, 185, 12, 72, 134, 254, 175, 234, 13, 159, 90, 139, 192, 190, 248, 67, 9, 32, 217}
+
+	// Initialize testApp
+	testApp := testutil.NewTestApp()
+
+	enc := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	// Create deterministic keys
+	kr, pubKeys := DeterministicKeyRing(enc.Codec)
+
+	recs, err := kr.List()
```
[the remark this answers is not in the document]
ninabarbakadze: I try to use the same conventions as our other testing infra https://github.com/celestiaorg/celestia-app/blob/fd4b75bb6145da14bdb62476ff065ade9530d0ea/test/util/testfactory/common.go#L69

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+		enc.TxConfig,
+		kr,
+		1000,
+		addresses[0],
+		addresses[:numNormalTxs],
+		testutil.ChainID,
+	)
+
+	// Create a set of 5 deterministic blob transactions
+	blobTxs := blobfactory.ManyMultiBlobTx(t, enc.TxConfig, kr, testutil.ChainID, addresses[numBlobTxs+1:], accinfos[numBlobTxs+1:], testfactory.Repeat([]*blob.Blob{
```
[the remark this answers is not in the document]
ninabarbakadze: how do you mean different transactions? I'm just using different addresses for blob txs. I just want to start the slice at index 6

app/test/non_determinism_test.go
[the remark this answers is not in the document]
rootulp: kk can resolve. I have no idea what `kb` means though

app/test/non_determinism_test.go
rootulp: Thanks, updating the comment so that they both say assert or require would be nice. Since there are no more assertions after this line an assert will behave identically to a require.

app/test/non_determinism_test.go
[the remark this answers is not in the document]
rootulp: Fair enough, can preserve that for now. In a separate PR if we really care, we can rename recs => records

cmwaters:
> I think we could improve this test by executing transaction types that include more unique messages because AFAIU this only tests MsgSend, MsgPayForBlob.

I agree but I think first merge this with the basic structure and then we can further elaborate

app/test/non_determinism_test.go
```diff
@@ -0,0 +1,132 @@
...
+	hd "github.com/cosmos/cosmos-sdk/crypto/hd"
+	keyring "github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/stretchr/testify/require"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+// TestNonDeterminismBetweenMainAndV1 executes a set of different transactions,
+// produces an app hash and compares it with the app hash produced by v1.x
+func TestNonDeterminismBetweenMainAndV1(t *testing.T) {
```
cmwaters, line 25: This shouldn't necessary be between v1 and main. It should be comparing the app hash generated from this commit compared to the app hash generated from the previous commit

cmwaters (commented):
I would have constructed this test a little differently. I would use a single `user.Signer` that has the keyring with all accounts. Then I would have something like.

```go
txs := []struct {
    sdkMsgs []sdk.Msg
    author string
    blobs []blob.Blob
    txOptions []user.TxOption
}{
   // add a mixture of blob txs and non blob txs with various signers
}
// you may want to break this out over multiple blocks
txBytes := make([]byte, len(txs))
for idx, tx := range txs {
    if len(tx.sdkMsgs) > 0 {
        txBytes[idx] = signer.CreateTx(tx.sdkMsgs, tx.txOptions)
    } else {
       txBytes[idx] = signer.CreatePayForBlob(tx.author, tx.blobs, tx.txOptions)
   }
}
// run prepare proposal and process proposal (make sure data hashes are deterministic)

// run `BeginBlock`, `DeliverTx`,  and `EndBlock` with the txBytes

// run `Commit` and get the app hash. Compare with the static one

```

app/test/non_determinism_test.go
evan-forbes: agree that I think we can rename this function since the commitment should be the same across all branches if I'm understanding this comment correctly.

evan-forbes (dismissed):

app/test/non_determinism_test.go
rootulp: [optional] slight tangent: the test could be renamed something like `TestConsistentAppHash` because it is asserting some expected app hash matches some calculated app hash. I think the goal is to identify cases of non-determinism but just because this test passes, it doesn't mean we prevented all potential cases of non-determinism.

ninabarbakadze: i agree with all of your point, i only wanted to underline that the hardcoded apphash was generated on v1.x because technically we could do it between any branches but practically the logic for generating apphash lives on v1.x.

app/test/non_determinism_test.go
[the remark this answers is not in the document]
ninabarbakadze: I'll link the file once this pr is merged as a follow-up https://github.com/celestiaorg/celestia-app/pull/3522

app/test/consistent_apphash_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Get the app hash
+	appHash := testApp.LastCommitID().Hash
+
+	// Require that the app hash is equal to the app hash produced on a different commit
+	require.Equal(t, expectedAppHash, appHash)
+}
+
+// DeterministicNamespace returns a deterministic namespace
+func Namespace() appns.Namespace {
```
rootulp, line 137: 1. comment name should match function name
2. If this isn't used outside this test file, proposal to un-export it.

```suggestion
// namespace returns a deterministic namespace
func namespace() appns.Namespace {
```

app/test/consistent_apphash_test.go
```diff
@@ -0,0 +1,173 @@
...
+	return appns.Namespace{
+		Version: 0,
+		ID:      []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 37, 67, 154, 200, 228, 130, 74, 147, 162, 11},
+	}
+}
+
+// DeterministicKeyRing returns a deterministic keyring and a list of deterministic public keys
+func DeterministicKeyRing(cdc codec.Codec) (keyring.Keyring, []types.PubKey) {
```
rootulp, line 145: [optional] un-export

```suggestion
// deterministicKeyRing returns a deterministic keyring and a list of deterministic public keys
func deterministicKeyRing(cdc codec.Codec) (keyring.Keyring, []types.PubKey) {
```

test/util/test_app.go
```diff
@@ -73,23 +76,99 @@ func SetupTestAppWithGenesisValSet(cparams *tmproto.ConsensusParams, genAccounts
...
+		NextValidatorsHash: genDoc.ValidatorHash(),
+		Version: tmversion.Consensus{
+			App: cparams.Version.AppVersion,
+		},
+	}})
 
+	return gen.Keyring(), gen.Accounts(), nil
+}
+
+// NewTestAppWithGenesisSet initializes a new app with a validator set and genesis accounts.
```
rootulp, line 169: [nit] since this function doesn't take a validator set as a parameter, I think the current godoc is misleading.

```suggestion
// NewTestAppWithGenesisSet initializes a new app with genesis accounts and returns the testApp, validator set and keyring.
```

test/util/testnode/utils.go
```diff
@@ -65,10 +65,9 @@ func RandomAddress() sdk.Address {
 	return addresses[0]
 }
 
-func FundKeyringAccounts(accounts ...string) (keyring.Keyring, []banktypes.Balance, []authtypes.GenesisAccount) {
-	kr, addresses := NewKeyring(accounts...)
-	genAccounts := make([]authtypes.GenesisAccount, len(accounts))
-	genBalances := make([]banktypes.Balance, len(accounts))
+func FundKeyringAccounts(addresses []sdk.AccAddress) ([]banktypes.Balance, []authtypes.GenesisAccount) {
```
rootulp, line 68: [informational] note this modification is technically Go API breaking. We technically don't need to preserve Go API compatability because this is targeting `main` so FYI but no change needed

rootulp (approved):

test/util/test_app.go
ninabarbakadze: i was not only referring to the arguments it takes but also the overall functionality of the method and what it initialises. happy to change it if you think it's still misleading.

app/test/consistent_apphash_test.go
ninabarbakadze: i can't just have namespace because it conflicts with the package name but i renamed it to `fixedNamespace`

evan-forbes (approved):

test/util/test_app.go
```diff
@@ -73,23 +76,99 @@ func SetupTestAppWithGenesisValSet(cparams *tmproto.ConsensusParams, genAccounts
...
 		log.NewNopLogger(), db, nil,
 		cast.ToUint(emptyOpts.Get(server.FlagInvCheckPeriod)),
 		encCfg,
 		0,
 		emptyOpts,
 	)
+}
+
+// ApplyGenesisState sets genesis on initialized testApp with the provided arguments.
+func ApplyGenesisState(testApp *app.App, pubKeys []cryptotypes.PubKey, balance int64, cparams *tmproto.ConsensusParams) (keyring.Keyring, []genesis.Account, error) {
```
cmwaters, line 98: why not?
```suggestion
func ApplyGenesisState(testApp *app.App, gen *genesis.Genesis) error {
```
The way you currently have it is not applying genesis but creating a genesis.

test/util/test_app.go
```diff
@@ -126,6 +205,48 @@ func NewTestAppWithGenesisSet(cparams *tmproto.ConsensusParams, genAccounts ...s
 	return testApp, valSet, kr
 }
 
+// AddDeterministicValidatorToGenesis adds a single deterministic validator to the genesis.
+func AddDeterministicValidatorToGenesis(g *genesis.Genesis) error {
+	// hardcoded keys for deterministic account creation
+	mnemo := "body world north giggle crop reduce height copper damp next verify orphan lens loan adjust inform utility theory now ranch motion opinion crowd fun"
+	consensusKey := ed25519.GenPrivKeyFromSecret([]byte("12345678901234567890123456389012"))
+	networkKey := ed25519.GenPrivKeyFromSecret([]byte("12345678901234567890123456786012"))
```
cmwaters, line 213: If you look at `test/util/common.go`, they already have a bunch of keys set up that you can use

test/util/testnode/utils.go
cmwaters: I would prefer if we add this as a completely new function called `FundAddresses` since it's quite a different function

test/util/test_app.go
ninabarbakadze: preferably i'd have a function that lets you customise genesis without having to do the setup yourself. i'm happy to change the function signature to correctly reflect the behaviour. we also have [a ticket refactoring test_app ](https://github.com/celestiaorg/celestia-app/issues/2535) so we'll spend more time properly designing this process.

test/util/test_app.go
ninabarbakadze: I don't think those keys are deterministic

test/util/testnode/utils.go
ninabarbakadze: I'll revert this change completely. I thought that decoupling those functions made sense but we don't really use them elsewhere.

test/util/test_app.go
ninabarbakadze: technically this method is applying genesis on already initialised testapp

test/util/test_app.go
rootulp: I think it's misleading so I prefer the comment I proposed but not blocking on it.

rootulp (approved):

evan-forbes (approved):

PR #3520, opened 2024-05-31
chore: add test/util/genesis on v1.x

ninabarbakadze:
## Overview

adding genesis to utils as part of #2414


rach-id (commented):
As a first pass, why is this needed to be in v1.x?

ninabarbakadze:
> As a first pass, why is this needed to be in v1.x?

non determinism unit test. testApp needs to be constructed the same way so i wanted to reuse this logic.

test/util/testnode/config.go
```diff
@@ -7,42 +7,37 @@ import (
 	"github.com/celestiaorg/celestia-app/cmd/celestia-appd/cmd"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
 	v1 "github.com/celestiaorg/celestia-app/pkg/appconsts/v1"
+	"github.com/celestiaorg/celestia-app/test/util/genesis"
 	"github.com/cosmos/cosmos-sdk/server"
 	srvconfig "github.com/cosmos/cosmos-sdk/server/config"
 	srvtypes "github.com/cosmos/cosmos-sdk/server/types"
 	tmconfig "github.com/tendermint/tendermint/config"
-	tmrand "github.com/tendermint/tendermint/libs/rand"
+	// tmrand "github.com/tendermint/tendermint/libs/rand"
```
rootulp, line 15:
```suggestion
```

app/test/integration_test.go
```diff
@@ -206,78 +201,6 @@ func (s *IntegrationTestSuite) TestMaxBlockSize() {
 	}
 }
 
-func (s *IntegrationTestSuite) TestSubmitPayForBlob() {
```
rootulp, line 209: [question] was this test moved or just deleted? It doesn't look like the original PR deleted them: https://github.com/celestiaorg/celestia-app/pull/2455/files#diff-d66f948da147c1b292dcf8131bd525c02c6e23daa6c64712a85e53476271cc65L60

app/test/integration_test.go
```diff
@@ -385,66 +308,6 @@ func (s *IntegrationTestSuite) TestEmptyBlock() {
 	}
 }
 
-// TestSubmitPayForBlob_blobSizes verifies the tx response ABCI code when
-// SubmitPayForBlob is invoked with different blob sizes.
-func (s *IntegrationTestSuite) TestSubmitPayForBlob_blobSizes() {
```
rootulp, line 390: [question] was this test moved or just deleted? It doesn't look like the original PR deleted them: https://github.com/celestiaorg/celestia-app/pull/2455/files#diff-d66f948da147c1b292dcf8131bd525c02c6e23daa6c64712a85e53476271cc65L60

rootulp (approved):
LGTM. One question about whether tests were moved or deleted.

app/test/integration_test.go
ninabarbakadze: it was deleted like in this pr https://github.com/celestiaorg/celestia-app/pull/2834/files#diff-d66f948da147c1b292dcf8131bd525c02c6e23daa6c64712a85e53476271cc65

app/test/integration_test.go
ninabarbakadze: ditto

app/test/integration_test.go
rootulp: Thanks. Resolving

rootulp (approved):

cmwaters (changes requested):
There's a lot of go API breaking changes in this PR. If you want to add the genesis package, it needs to be a purely additive change without removing several files and functions.

rach-id (approved):
utAck

rach-id:
> There's a lot of go API breaking changes in this PR

Should we worry about breaking changes to the testing infrastructure in release branches?

evan-forbes (approved):

cmwaters:
> Should we worry about breaking changes to the testing infrastructure in release branches?

rootulp:
I revise my review b/c the testing infra is exported so this is technically API breaking like Callum mentioned. 

@ninabarbakadze is it possible to add new exported functions and deprecate existing functions to avoid all API breaks?

ninabarbakadze:
> I revise my review b/c the testing infra is exported so this is technically API breaking like Callum mentioned.
> 
> @ninabarbakadze is it possible to add new exported functions and deprecate existing functions to avoid all API breaks?

Yes, I'm going to manually add the genesis folder without cherry picking evan's commit and see if it'll make the process of maintaining the API easier.

rootulp (approved):

cmwaters (approved):

Issue #3516, opened 2024-05-29
Add exception to get go-vuln passing 


evan-forbes:
## Description

There is rightfully a vuln in ibc v6.2, so the ci will fail. This vuln doesn't affect celestia-app since cosmwasm is not installed, therefore we can add an exception to avoid the red x in CI

## Acceptance Criteria
- [ ] go-vuln CI to passes after adding an exception to ibc v6.2.2

rootulp:
Doesn't seem possible to ignore a particular vulerability per 
- https://github.com/golang/go/issues/59507
- https://github.com/golang/go/issues/61211

PR #3529, opened 2024-06-04
ci: remove govulncheck

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3516
Opens https://github.com/celestiaorg/celestia-app/issues/3530

This PR removes govulncheck from CI because govulncheck does not support ignoring a particular vulnerability. Since we're on ibc-go v6.2.x which has a vulnerability, CI will report a red X on all future PRs because govulncheck fails. We can re-enable govulncheck when it adds support for ignoring a particular vulnerability.


rach-id (approved):

Makefile
```diff
@@ -213,8 +213,3 @@ prebuilt-binary:
 		ghcr.io/goreleaser/goreleaser-cross:${GOLANG_CROSS_VERSION} \
 		release --clean
 .PHONY: prebuilt-binary
-
-## govulncheck: Check for vulnerabilities in dependencies.
-govulncheck:
-	@go run golang.org/x/vuln/cmd/govulncheck@v1.0.1 ./...
-.PHONY: govulncheck
```
staheri14, line 220: [Optional] I find govulncheck useful and valuable in general, so would like to suggest keeping it if possible. Would something like below work? 
Directing the output of govulncheck to a txt file and then excluding the IBC portion of the error, and displaying the remaining part as the result.

rootulp: Something hacky like that could work but we'd also need to modify the exit code of the script so that CI only fails if the output contains anything after the filtering. 

IMO it isn't worth investing time to implement and maintain the hacky solution so I'd rather drop it entirely for now.

evan-forbes (approved):

Makefile
staheri14: I see there is a tracking [issue](https://github.com/celestiaorg/celestia-app/issues/3530) for adding govulncheck back when it supports ignoring errors, so, all good!

PR #3555, opened 2024-06-11
fix!: msg gatekeeper for authz messages

rootulp:
Resolves this [security advisory](https://github.com/celestiaorg/celestia-app/security/advisories/GHSA-j4w8-q5hh-6xf9). Thanks @Reecepbcups for the report and fix!


app/ante/msg_gatekeeper.go
```diff
@@ -32,15 +33,35 @@ func (mgk MsgVersioningGateKeeper) AnteHandle(ctx sdk.Context, tx sdk.Tx, simula
...
+
+func (mgk MsgVersioningGateKeeper) hasInvalidMsg(ctx sdk.Context, acceptedMsgs map[string]struct{}, msgs []sdk.Msg) error {
+	for _, msg := range msgs {
+		// Recursively check for invalid messages in nested authz messages.
+		if execMsg, ok := msg.(*authz.MsgExec); ok {
+			nestedMsgs, err := execMsg.GetMessages()
+			if err != nil {
+				return err
+			}
+			if err = mgk.hasInvalidMsg(ctx, acceptedMsgs, nestedMsgs); err != nil {
```
evan-forbes, line 52: do we want to put a limit on recursion? I'm not sure if this would be super valuable, but seems like good practice. I'm unaware of a usecase for authz that requires a bunch of nested msgs, but could be missing something.

if this wasn't consensus breaking before, it would be now

evan-forbes (approved):
is this consensus breaking?

app/ante/msg_gatekeeper.go
cmwaters: What you could do is remove authz.MsgExec from the accepted messages list here. This would enforce that it's not possible to MsgExec a MsgExec which I don't think is useful and could have unknown repercussions if we allow it

cmwaters (approved):
Fwiw I wouldn't count this as a bug because I intended this behaviour. In my original design the ante handler didn't exist and thus messages could be accepted but would just be rejected when it came to executing them. I later decided to add the antehandler as fail fast mechanism thinking it's better that the user knows right away that their message is invalid

app/ante/msg_gatekeeper.go
rootulp: An alternative option to the one Callum proposed is:

```diff
func (mgk MsgVersioningGateKeeper) hasInvalidMsg(ctx sdk.Context, acceptedMsgs map[string]struct{}, msgs []sdk.Msg) error {
	for _, msg := range msgs {
		// Recursively check for invalid messages in nested authz messages.
		if execMsg, ok := msg.(*authz.MsgExec); ok {
			nestedMsgs, err := execMsg.GetMessages()
			if err != nil {
				return err
			}
+			if err = mgk.hasNestedMsgExec(ctx, nestedMsgs); err != nil {
+				return err
+			}
			if err = mgk.hasInvalidMsg(ctx, acceptedMsgs, nestedMsgs); err != nil {
				return err
			}
		}

		msgTypeURL := sdk.MsgTypeURL(msg)
		_, exists := acceptedMsgs[msgTypeURL]
		if !exists {
			return sdkerrors.ErrNotSupported.Wrapf("message type %s is not supported in version %d", msgTypeURL, ctx.BlockHeader().Version.App)
		}
	}

	return nil
}

+func (mgk MsgVersioningGateKeeper) hasNestedMsgExec(_ sdk.Context, nestedMsgs []sdk.Msg) error {
+	for _, nestedMsg := range nestedMsgs {
+		if _, ok := nestedMsg.(*authz.MsgExec); ok {
+			return sdkerrors.ErrNotSupported.Wrapf("nested MsgExec is not allowed in MsgExec")
+		}
+	}

	return nil
}
```

but my concern with adding this is that even though a MsgExec inside a MsgExec seems like a strange use case, what if it's actually desirable for certain scenarios. 

I don't see a major risk from not adding this constraint because if a user crafts a deeply nested MsgExec wrapper (i.e. `MsgExec(MsgExec(MsgExect(...)))` then the transaction will be rejected based on tx size.

evan-forbes: sgmt, that's likely better than a simple limit.

rootulp: I spun out an issue for this this discussion: https://github.com/celestiaorg/celestia-app/issues/3586

Issue #3337, opened 2024-04-17
Add a test to restart the app


evan-forbes:
## Description 

In the past, we've ran into at least two bugs that resulted in a panic after restarting (one being #3336).

## Acceptance criteria

- [ ] Unit test that starts stops and restarts a node. If a unit test format is not possible e2e test will also do

cmwaters:
Good shout. it's hard to do a full reboot as a unit test. I think this would be better suited as a knuu test.

In cometbft's test they include perturbations which involve killing and restarting the docker instance

evan-forbes:
we might be able to with testnode if we manually call these and shutdown, then restart https://github.com/celestiaorg/celestia-app/blob/9db2f8b8bffe3dd1dd4eaf7d9b1d38afb71a193a/test/util/testnode/network.go#L36-L43

that might not work tho, in which case I agree, knuu is the way to go

cmwaters:
FYI, I think the upgrade tests already kill the node and restart it

PR #3505, opened 2024-05-27
feat: use proxy instead of port forwards

smuu:
## Overview

This pull request removes the need for port-forwarding. Instead, it facilitates the reverse proxy that is available in the scope.

I will comment in code to explain the changes.

@staheri14 Feel free to merge this as well to your branch

PR #3556, opened 2024-06-12
test: make nodes also restart after upgrading in knuu

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/3337

### DIFF
--- a/test/e2e/major_upgrade_v2.go
+++ b/test/e2e/major_upgrade_v2.go
@@ -68,6 +68,13 @@ func MajorUpgradeToV2(logger *log.Logger) error {
 
 	heightBefore := upgradeHeight - 1
 	for i := 0; i < numNodes; i++ {
+		// check for early errors from txsim
+		select {
+		case <-errCh:
+			return fmt.Errorf("txsim err: %w", ctx.Err())
+		default:
+		}
+
 		client, err := testNet.Node(i).Client()
 		testnet.NoError("failed to get client", err)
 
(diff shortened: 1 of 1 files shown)


cmwaters:
Currently dealing with this error:

```
E0612 12:34:54.754686   51154 portforward.go:409] an error occurred forwarding 49552 -> 9090: error forwarding port 9090 to pod c1fa51134fec5c9363b80fa36a57202e3c208bbe37128429d0eea31ad1662cb7, uid : network namespace for sandbox "c1fa51134fec5c9363b80fa36a57202e3c208bbe37128429d0eea31ad1662cb7" is closed
test-e2e2024/06/12 12:36:11 --- ERROR MajorUpgradeToV2: expected context.Canceled error, got: sequence 0: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:49552: connect: connection refused"
```

evan-forbes:
would there be any tradeoffs in running this in testnode? Perhaps not as thorough, although with my current understanding in theory it could catch many of the same issues. We'd also be able to block PRs that don't pass that test instead of waiting until after they are merged.

cmwaters:
> would there be any tradeoffs in running this in testnode?

1. cmd/celestia-appd (i.e. our exported binary) and testnode have different ways of constructing the node and running them
2. There may be incongruences between crashing a node (and the process that was running it) and programatically stopping and starting a node)
3. testnode is only equipped for running single node networks (so far) which wouldn't emulate what would happen if the network still progressed and the node had to catch up.

Outside of these three, I would think it's relatively similar

cmwaters:
I may be mistaken, but I think this is blocked on https://github.com/celestiaorg/celestia-app/pull/3505 and potential other changes that can avoid the port forwarding errors that arise

cmwaters:
This is passing for me locally now

ninabarbakadze (approved):

test/e2e/major_upgrade_v2.go
```diff
@@ -79,7 +80,24 @@ func MajorUpgradeToV2(logger *log.Logger) error {
...
+		client, err := node.Client()
+		testnet.NoError("failed to get client", err)
+
+		height, err := getHeight(ctx, client, time.Minute)
+		if err != nil {
+			return fmt.Errorf("failed to get height: %w", err)
+		}
+
+		if err := node.Upgrade(latestVersion); err != nil {
+			return fmt.Errorf("failed to restart node: %w", err)
```
rootulp, line 94: [optional]

```suggestion
			return fmt.Errorf("failed to upgrade node: %w", err)
```

rootulp (approved):

staheri14 (approved):

test/e2e/major_upgrade_v2.go
cmwaters: I know the function says upgrade, but it's actually upgrading to the same version, effectively a restart. It could be cleaner if we just have a `Restart` method in testnet that does the same thing

rootulp: +1 restart would be clearer in that case

Issue #3550, opened 2024-06-11
Remove constraint that enforces signals for current or next app version


rootulp:
## Context

Informal Systems v2 audit finding 1.

## Problem

Assume the network is currently on v2.0.0 (app version 2) and we release a buggy v3.0.0 (app version 3). We want the network to skip that version so we release a v3.0.1 (app version 4). However, there is a constraint that prevents validators from signaling for app version 4.

## Proposal

- Remove [this constraint](https://github.com/celestiaorg/celestia-app/blob/f9a7e04ef537ba5ec5af1e9a29055628778cc3f1/x/signal/keeper.go#L67-L72) that enforces signals must be for the current or next app version.
- Inspect the rest of the x/signal module to ensure there isn't anything that depends on this constraint.

PR #3558, opened 2024-06-12
feat!: modify constraint for signal versions

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3550

### DIFF
--- a/x/signal/keeper.go
+++ b/x/signal/keeper.go
@@ -63,6 +63,12 @@ func (k Keeper) SignalVersion(ctx context.Context, req *types.MsgSignalVersion) 
 		return nil, err
 	}
 
+	// The signalled version can not be less than the current version.
+	currentVersion := sdkCtx.BlockHeader().Version.App
+	if req.Version < currentVersion {
+		return nil, types.ErrInvalidVersion.Wrapf("signalled version %d, signalled version %d", req.Version, currentVersion)
+	}
+
 	_, found := k.stakingKeeper.GetValidator(sdkCtx, valAddr)
 	if !found {
 		return nil, stakingtypes.ErrNoValidatorFound
--- a/x/signal/keeper.go
+++ b/x/signal/keeper.go
@@ -63,13 +63,6 @@ func (k Keeper) SignalVersion(ctx context.Context, req *types.MsgSignalVersion) 
 		return nil, err
 	}
 
-	// The signalled version must be either the current version or the next
-	// version.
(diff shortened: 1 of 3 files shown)


x/signal/keeper.go
```diff
@@ -63,13 +63,6 @@ func (k Keeper) SignalVersion(ctx context.Context, req *types.MsgSignalVersion)
 		return nil, err
 	}
 
-	// The signalled version must be either the current version or the next
-	// version.
-	currentVersion := sdkCtx.BlockHeader().Version.App
-	if req.Version != currentVersion && req.Version != currentVersion+1 {
-		return nil, types.ErrInvalidVersion
-	}
```
cmwaters, line 71: I think we should still assert that:

```go
req.Version >= currentVersion
```

so as to not support downgrades

rootulp: Yea good point so that we match this constraint: https://github.com/rootulp/celestia-app/blob/c270c424020b146168a10361c190eae4b073dab0/app/app.go#L468-L472

rootulp: Fixed

cmwaters: The reason for wanting to allow to vote for the current version is: 
- Network is on v2
- We release v3
- Nodes start signalling for v3
- Bug is realised or for whatever reason we decide the release is not ready
- _Nodes signal back to the current version to cancel the upgrade_

I get your point about the delay but I think that can be overcome i.e. we write it as a noop that also does not trigger the delay

rootulp: ahh makes sense. An alternative would be:

Introduce a new transaction type that allows a validator to clear their previous signal.
Pro: clearer than signaling for current version
Con: larger change than allowing validators to signal for current version

We can allow voting for the current version and do a no-op for the delay if quorum version == current version

x/signal/keeper_test.go
```diff
@@ -122,11 +121,13 @@ func TestTallyingLogic(t *testing.T) {
 		Version:          0,
 	})
 	require.Error(t, err)
+	require.ErrorIs(t, err, types.ErrInvalidVersion)
```
staheri14, line 124: [nit] Would be helpful to add a one liner for each testcase explaining why an error is required or not required.

staheri14 (approved):
LGTM!

x/signal/keeper.go
cmwaters: Yeah I would be fine with that, like a `MsgCancelSignal` which under the hood signals for the current version

Issue #3552, opened 2024-06-11
Add a delay between `MsgTryUpgrade` and upgrade height


rootulp:
## Context

Informal Systems v2 audit finding 3.

## Problem

A user can submit a `MsgTryUpgrade` as soon as 5/6 voting power has signaled for a version. The remaining 1/6 of voting power won't have any time to upgrade prior to the upgrade height.

## Proposal

Add a delay between the block that includes a successful `MsgTryUpgrade` and the upgrade height.

1. Determine a reasonable delay time period. Ex: 2 days or 1 week.
2. Add the delay. Note: during this time all signals and future cranks should be no-ops.

PR #3560, opened 2024-06-12
feat!: add a delay between quorum and upgrade height

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3552

Ready for review but I won't merge until we reach social consensus on what the upgrade height delay should be.

## Testing

Manually tested the new CLI commands by staring single node script and then:

```shell
$ celestia-appd query signal upgrade
no upgrade is pending.

$ celestia-appd tx signal signal 3 --from $VALIDATOR --fees $DEFAULT_FEES --broadcast-mode $BROADCAST_MODE --yes

$ celestia-appd tx signal try-upgrade  --from $VALIDATOR --fees $DEFAULT_FEES --broadcast-mode $BROADCAST_MODE --yes

$ celestia-appd query signal upgrade
upgrade is pending to app version 3 at height 151345.
```


cmwaters (changes requested):
Given the upgrade process now spans across multiple heights we need to be aware of node crashes, thus the signal keeper will need to persist the upgrade height and the quorum version to state.

I think it would also be helpful if we expose a grpc query for that upgrade height so users are able to know when exactly that upgrade height is.

x/signal/keeper.go
```diff
@@ -171,23 +201,39 @@ func (k Keeper) GetVotingPowerThreshold(ctx sdk.Context) sdkmath.Int {
...
 	for ; iterator.Valid(); iterator.Next() {
+		if bytes.Equal(iterator.Key(), types.UpgradeKey) {
+			// skip over the upgrade key
+			continue
+		}
 		store.Delete(iterator.Key())
 	}
-	k.quorumVersion = 0
+	// delete the upgrade value
+	store.Delete(types.UpgradeKey)
```
cmwaters, line 236: Is it necessary to delete the upgrade last instead of first?

x/signal/keeper.go
```diff
@@ -102,6 +126,9 @@ func (k Keeper) VersionTally(ctx context.Context, req *types.QueryVersionTallyRe
 	iterator := store.Iterator(nil, nil)
 	defer iterator.Close()
 	for ; iterator.Valid(); iterator.Next() {
+		if bytes.Equal(iterator.Key(), types.UpgradeKey) {
+			continue
+		}
```
cmwaters, line 131: You don't need to include the if statement if you want to start the iterator after the upgrade key (since the upgrade key is presumably the first key in the kv store).

proto/celestia/signal/v1/query.proto
```diff
@@ -13,6 +14,12 @@ service Query {
       returns (QueryVersionTallyResponse) {
     option (google.api.http).get = "/upgrade/v1/tally/{version}";
   }
+
+  // GetUpgrade enables a client to query for upgrade information if an upgrade is pending.
```
cmwaters, line 18: nit: maybe add a little note that the response will be empty if there is no current upgrade pending

cmwaters (commented):
Thanks for adding the query endpoint and tests. Great work! 

I just have a few more minor comments and then we should be good to merge

x/signal/keeper.go
```diff
@@ -18,6 +20,11 @@ var (
 
 	// defaultSignalThreshold is 5/6 or approximately 83.33%
 	defaultSignalThreshold = sdk.NewDec(5).Quo(sdk.NewDec(6))
+
+	// defaultUpgradeHeightDelay is the number of blocks after a quorum has been
+	// reached that the chain should upgrade to the new version. Assuming a
+	// block interval of 12 seconds, this is 48 hours.
+	defaultUpgradeHeightDelay = int64(3 * 7 * 24 * 60 * 60 / 12) // 3 weeks * 7 days * 24 hours * 60 minutes * 60 seconds / 12 seconds per block = 151,200 blocks.
```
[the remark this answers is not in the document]
rootulp: yea can to make it const and public. 

We can discuss the duration of 3 weeks in today's v2 call. Yaz suggested 3 weeks so I went with that.

proto/celestia/signal/v1/query.proto
rootulp: 01ec9f3d0e8089483f7e5d887b121622e5088719

x/signal/keeper.go
rootulp: nope but I thought this was clearer. I refactored to something else that may be clearer: c7a9f7d43646cb168750d4709ea93646be226b3c

x/signal/keeper.go
rootulp: how do I start the iterator at the first key _after_ the upgrade key? I expect this will start the iterator _at_ the upgrade key:

```
	iterator := store.Iterator(types.UpgradeKey, nil)
```

and I don't have a constant for the first value _after_ the upgrade key. If we start an iterator like this, we'll need to skip the first value which seems just as janky as the explicit conditional:

```go
		if bytes.Equal(iterator.Key(), types.UpgradeKey) {
			continue
		}
```

x/signal/keeper.go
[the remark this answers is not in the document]
cmwaters: Ok, I'll just state my reason here for posterity. Having a three week delay means all networks, include our testnets will take at minimum three weeks. This could mean that it may take two to three months from the initial final release to being on mainnet. Including auditing and other internal testing, this is frankly way too slow of a turnaround time. Especially when we want to try increase the cadence of upgrades (it's almost been 9 months since mainnet launch).

x/signal/keeper.go
cmwaters: What's wrong with adding a constant like:

```go
var FirstSignalKey = []byte{0x000} 
```

and start the iterator there

x/signal/keeper.go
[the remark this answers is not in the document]
rootulp: We could make the delay period different durations based on the network. 
- Arabica: 1 week
- Mocha: 2 weeks
- Mainnet: 3 weeks

x/signal/keeper.go
rootulp: hmm nothing seems wrong with that but I wasn't sure which would be logically after `[]byte{0x00}`: A: `[]byte{0x000}` or B: `[]byte{0x01}`.

 I wonder if we should reserve more keys at the beginning of the range in case we want the signal keeper to persist other things in the state store. Although if we do that, we would need a state migration at which point we can reserve more keys so may not need them now.

rootulp: 3e791ddd25518cfa34cc5068f0f06026e478babd

x/signal/keeper.go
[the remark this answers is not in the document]
rootulp: Discussed synchronously and landed on 1 week.

x/signal/integration_test.go
```diff
@@ -54,7 +55,28 @@ func TestUpgradeIntegration(t *testing.T) {
...
+	require.ErrorIs(t, err, types.ErrUpgradePending)
+
+	// Verify that if a validator tries to change their signal version, it
+	// returns an error because an upgrade is pending.
+	_, err = app.SignalKeeper.SignalVersion(ctx, &types.MsgSignalVersion{
+		ValidatorAddress: valAddr.String(),
+		Version:          3,
+	})
+	require.Error(t, err)
+	require.ErrorIs(t, err, types.ErrUpgradePending)
```
cmwaters, line 71: 👍

cmwaters (dismissed):
Perfect, thanks

x/signal/types/keys.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+var (
+	// UpgradeKey is the key in the signal store used to persist a upgrade if one is
+	// pending.
+	UpgradeKey = []byte{0x00}
+
+	// FirstSignalKey is used as a divier to separate the UpgradeKey from all
```
staheri14, line 8:
```suggestion
	// FirstSignalKey is used as a divider to separate the UpgradeKey from all
```

x/signal/keeper_test.go
```diff
@@ -291,32 +313,118 @@ func TestThresholdVotingPower(t *testing.T) {
...
+		_, err = upgradeKeeper.SignalVersion(ctx, &types.MsgSignalVersion{ValidatorAddress: testutil.ValAddrs[3].String(), Version: 2})
+		require.NoError(t, err)
+
+		// This TryUpgrade should succeed.
+		_, err = upgradeKeeper.TryUpgrade(goCtx, &types.MsgTryUpgrade{})
+		require.NoError(t, err)
+
+		got, err := upgradeKeeper.GetUpgrade(ctx, &types.QueryGetUpgradeRequest{})
+		require.NoError(t, err)
+		assert.Equal(t, uint64(2), got.Upgrade.AppVersion)
```
staheri14, line 425: [nit]
If there is a constant for the app version e.g., [this](https://github.com/celestiaorg/celestia-app/blob/4c732d312032d96466b06e5f5b7db98268f2b1b1/pkg/appconsts/v2/app_consts.go#L4), I'd suggest using the const instead of literals.

x/signal/keeper.go
```diff
@@ -78,16 +89,28 @@ func (k Keeper) SignalVersion(ctx context.Context, req *types.MsgSignalVersion)
...
 	if hasQuorum {
-		k.quorumVersion = version
+		if version <= sdkCtx.BlockHeader().Version.App {
+			return &types.MsgTryUpgradeResponse{}, types.ErrInvalidUpgradeVersion.Wrapf("can not upgrade to version %v because it is less than or equal to current version %v", version, sdkCtx.BlockHeader().Version.App)
+		}
+		upgrade := types.Upgrade{
+			AppVersion:    version,
+			UpgradeHeight: sdkCtx.BlockHeader().Height + DefaultUpgradeHeightDelay,
+		}
+		k.setUpgrade(sdkCtx, upgrade)
```
staheri14, line 113: [Question] Would it be possible that two validators invoke `TryUpgrade` at two different heights hence resulting in them calculating two different `UpgradeHeight`s? Would it be essential to ensure the entire network end up calculating the same `UpgradeHeight`?

staheri14 (commented):
LGTM! I'd defer to other reviewers for the final approval.

x/signal/keeper.go
rootulp: Not that I would expect because if validator A and B are on the same chain. Both validators will execute the try upgrade included in the earlier block.

x/signal/keeper_test.go
rootulp: 8e9c37d86a8c5efd3522f4298d3802c124e3c4a2

staheri14 (approved):
LGTM!

cmwaters (approved):

Issue #3539, opened 2024-06-07
Add grpc querier for minfee


cmwaters:
### Summary

Right now, if I want to find the minfee of the chain I have to use the `Params` `QueryClient`. I have to know the module name and key that I am looking for and then I have to unwrap the string which has double quotations. It may be more intuitive to define a grpc query client in `minfee` itself.

PR #3565, opened 2024-06-14
feat: add grpc query server for minfee

cmwaters:
Closes: #3539

Creates a grpc query server for easier querying of the min gas price.

During this PR I had a change of heart and wanted to call `GlobalMinGasPrice` to `NetworkMinGasPrice` as I felt this was somewhat more accurate. Let me know how you guys feel?


proto/celestia/minfee/v1/query.proto
```diff
@@ -0,0 +1,26 @@
...
+
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/minfee";
+
+// Query defines the gRPC querier service.
+service Query {
+  // NetworkMinGasPrice queries the network wide minimum fee
```
rootulp, line 12: I interpret fee as gasPrice * gas so IMO gas price is more correct here

```suggestion
  // NetworkMinGasPrice queries the network wide minimum gas price.
```

if we wanted to be really verbose it could say

```suggestion
  // NetworkMinGasPrice queries the network wide minimum gas price per unit of gas.
```

x/minfee/grpc_query.go
```diff
@@ -0,0 +1,34 @@
...
+type QueryServerImpl struct {
+	paramsKeeper keeper.Keeper
+}
+
+// NewQueryServerImpl creates a new QueryServerImpl.
+func NewQueryServerImpl(paramsKeeper keeper.Keeper) *QueryServerImpl {
+	return &QueryServerImpl{paramsKeeper: paramsKeeper}
+}
+
+// NetworkMinGasPrice returns the global minimum gas price.
```
rootulp, line 24:
```suggestion
// NetworkMinGasPrice returns the network minimum gas price.
```

x/minfee/grpc_query.go
```diff
@@ -0,0 +1,34 @@
...
+func (q *QueryServerImpl) NetworkMinGasPrice(ctx context.Context, _ *QueryNetworkMinGasPrice) (*QueryNetworkMinGasPriceResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	var params Params
+	subspace, found := q.paramsKeeper.GetSubspace(ModuleName)
+	if !found {
+		return nil, status.Errorf(codes.NotFound, "subspace not found for minfee. Minfee is only active v2 and onwards")
```
rootulp, line 30:
```suggestion
		return nil, status.Errorf(codes.NotFound, "subspace not found for minfee. Minfee is only active in app version 2 and onwards")
```

x/minfee/grpc_query.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	subspace.GetParamSet(sdkCtx, &params)
+	return &QueryNetworkMinGasPriceResponse{NetworkMinGasPrice: params.GlobalMinGasPrice}, nil
```
rootulp, line 33: nit: I personally slightly prefer NetworkMinGasPrice but I think it's more important if all the terms related to this param were consistent. 

In other words, `params.GlobalMinGasPrice` should also be renamed to `params.NetworkMinGasPrice`. If we're not doing that, then this PR should revert back to the previous name GlobalMinGasPrice so that the codebase is consistent and uses one term for it.

x/minfee/grpc_query_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	sdkCtx := testApp.NewContext(false, tmproto.Header{Height: 1})
+	ctx := sdk.WrapSDKContext(sdkCtx)
+
+	// Perform a query for the network minimum gas price
+	resp, err := queryServer.NetworkMinGasPrice(ctx, &minfee.QueryNetworkMinGasPrice{})
+	require.NoError(t, err)
+
+	// Check the response
+	require.Equal(t, resp.NetworkMinGasPrice.MustFloat64(), v2.GlobalMinGasPrice)
```
rootulp, line 27: nit I think the order of these is swapped because the expected value is `v2.GlobalMinGasPrice` and the actual value is `resp...`

```suggestion
	require.Equal(t, v2.GlobalMinGasPrice, resp.NetworkMinGasPrice.MustFloat64())
```

it is minor but will help if we need to debug this test in the future.

rootulp (dismissed):
LGTM. I think we'll need a follow-up PR to rename all instances of GlobalMinGasPrice to the new name NetworkMinGasPrice

x/minfee/grpc_query.go
cmwaters: Yes I agree. I was planning to do that in a follow up PR

rootulp (dismissed):

x/minfee/grpc_query.go
ninabarbakadze: I also like NetworkMinGasPrice better

ninabarbakadze (approved):
lgtm

rootulp (approved):

PR #3571, opened 2024-06-17
fix: minfee key table not registered on app restarts

### LINKED ISSUE #3553: GlobalMinGasPrice not getting persisted
### Summary of Bug

Apparently, there is a problem persisting the app version and the GlobalMinGasPrice. I will continue digging.

```
celestia-appd tx blob pay-for-blob 0x00010203040506070809 0x000102030405060708094314314324324234234234234234234324234234234232 --fees 21000utia --chain-id 
private --keyring-backend test -b block --yes --from validator
code: 111222
codespace: undefined
data: ""
events: []
gas_used: "8786"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: "recovered: parameter GlobalMinGasPrice not registered\ncaused by transaction:\n*types.MsgPayForBlobs{signer:\"celestia147x5c0hxx4vt7j8njx4s000vwut8vlxm7mz5ua\"
  namespaces:\"\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\000\\001\\002\\003\\004\\005\\006\\007\\010\\t\"
  blob_sizes:33 share_commitments:\"w3\\223!\\357\\222\\346$\\206\\022dp\\244ARt\\373\\\\\\356@\\266\\210\\\"\\206\\363\\212\\2430\\r\\224\\320\\366\"
  share_versions:0 }\n\nstack:\ngoroutine 468 [running]:\nruntime/debug.Stack()\n\t/Users/midnight/go/go1.22.3/src/runtime/debug/stack.go:24
  +0x64\ngithub.com/cosmos/cosmos-sdk/baseapp.newDefaultRecoveryMiddleware.func1({0x107789c80,
  0x14001665e70})\n\t/Users/midnight/go2/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.22.0-sdk-v0.46.16/baseapp/recovery.go:71
  +0x2c\ngithub.com/cosmos/cosmos-sdk/baseapp.newRecoveryMiddleware.func1({0x107789c80,
  0x14001665e70})\n\t/Users/midnight/go2/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.22.0-sdk-v0.46.16/baseapp/recovery.go:39
  +0x4c\ngithub.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x107789c80, 0x14001665e70},
  0x1400023c780)\n\t/Users/midnight/go2/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.22.0-sdk-v0.46.16/baseapp/recovery.go:28
  +0x50\ngithub.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x107789c80, 0x14001665e70},
  0x1400140d8a8)\n\t/Users/midnight/go2/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.22.0-sdk-v0.46.16/baseapp/recovery.go:33
  +0xb8\ngithub.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).runTx
```

### Version

2

### Steps to Reproduce

1. checkout v2.0.0-rc1
2. bash scripts/single-node.sh
3. let it run past block 3
4. stop celestia-appd
5. run it again using:
```
celestia-appd start \
  --api.enable \
  --grpc.enable \
  --grpc-web.enable \
  --v2-upgrade-height 3
```
6. execute:
```
celestia-appd tx blob pay-for-blob 0x00010203040506070809 0x000102030405060708094314314324324234234234234234234324234234234232 --fees 21000utia --chain-id 
private --keyring-backend test -b block --yes --from validator
```

Second one, if you do the above steps, and at step 5, remove the --v2-upgrade-height 3 flag, and execute the PFB command:

```
celestia-appd tx blob pay-for-blob 0x00010203040506070809 0x000102030405060708094314314324324234234234234234234324234234234232 --fees 21000utia --chain-id 
private --keyring-backend test -b block --yes --from validator
code: 37
codespace: sdk
data: ""
events: []
gas_used: "8786"
gas_wanted: "-1"
height: "0"
info: ""
logs: []
raw_log: 'app version 0 is not supported: feature not supported'
timestamp: ""
tx: null
txhash: C586BBD76F4AAFBFAD7064472A677AB8FC5D0E0429AFBBC29A05DB91550E2C35
```

ninabarbakadze:
## Overview

Fixes #3553

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/celestiaorg/celestia-app/v2/app"
 	"github.com/celestiaorg/celestia-app/v2/app/encoding"
+	"github.com/celestiaorg/celestia-app/v2/x/minfee"
 	"github.com/stretchr/testify/assert"
 	"github.com/tendermint/tendermint/libs/log"
 	tmdb "github.com/tendermint/tm-db"
@@ -38,6 +39,11 @@ func TestNew(t *testing.T) {
 	})
 	t.Run("should not have sealed the baseapp", func(t *testing.T) {
 		assert.False(t, got.IsSealed())
+	})
+	t.Run("should have set the minfee key table", func(t *testing.T) {
+		subspace := got.GetSubspace(minfee.ModuleName)
+		hasKeyTable := subspace.HasKeyTable()
+		assert.True(t, hasKeyTable)
 	})
 }
 
--- a/x/minfee/module_test.go
+++ b/x/minfee/module_test.go
@@ -32,7 +32,7 @@ func TestNewModuleInitializesKeyTable(t *testing.T) {
 	paramsKeeper := paramkeeper.NewKeeper(codec.NewProtoCodec(registry), codec.NewLegacyAmino(), storeKey, tStoreKey)
 	subspace := paramsKeeper.Subspace(minfee.ModuleName)
 
-	// Initialize the minfee module which register set the key table
+	// Initialize the minfee module which registers the key table
 	minfee.NewAppModule(paramsKeeper)
 
 	// Require key table to be initialized
--- a/x/minfee/module_test.go
+++ b/x/minfee/module_test.go
@@ -1,0 +1,41 @@
+package minfee_test
+
+import (
+	"testing"
+
+	"github.com/celestiaorg/celestia-app/v2/x/minfee"
+	"github.com/cosmos/cosmos-sdk/codec"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/cosmos/cosmos-sdk/store"
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	paramkeeper "github.com/cosmos/cosmos-sdk/x/params/keeper"
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	"github.com/stretchr/testify/require"
+	tmdb "github.com/tendermint/tm-db"
+)
+
+func TestNewModuleInitializesKeyTable(t *testing.T) {
+	storeKey := sdk.NewKVStoreKey(paramtypes.StoreKey)
+	tStoreKey := storetypes.NewTransientStoreKey(paramtypes.TStoreKey)
+
+	// Create the state store
+	db := tmdb.NewMemDB()
+	stateStore := store.NewCommitMultiStore(db)
+	stateStore.MountStoreWithDB(storeKey, storetypes.StoreTypeIAVL, db)
+	stateStore.MountStoreWithDB(tStoreKey, storetypes.StoreTypeTransient, nil)
+	require.NoError(t, stateStore.LoadLatestVersion())
+
+	registry := codectypes.NewInterfaceRegistry()
+
+	// Create a params keeper
+	paramsKeeper := paramkeeper.NewKeeper(codec.NewProtoCodec(registry), codec.NewLegacyAmino(), storeKey, tStoreKey)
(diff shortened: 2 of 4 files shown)


cmwaters (commented):
It's common for modules to have the `RegisterMinFeeParamTable` or equivalent in the constructor of the keeper. Since `minfee` doesn't have a keeper per say (there is nothing in `minfee` that is committed to state), perhaps it's cleaner to include it in the constructor of the module itself. That way we know whenever the module is used the param should be registered rather than have it haphazardly chucked somewhere in `app.go`

https://github.com/celestiaorg/celestia-app/blob/34f2443ba639e68a489307a75dcab84d3030c40b/x/minfee/module.go#L79-L84

app/app_test.go
```diff
@@ -39,6 +40,11 @@ func TestNew(t *testing.T) {
 	t.Run("should not have sealed the baseapp", func(t *testing.T) {
 		assert.False(t, got.IsSealed())
 	})
+	t.Run("should have set the minfee key table", func(t *testing.T) {
+		subspace := got.GetSubspace(minfee.ModuleName)
+		hasKeyTable := subspace.HasKeyTable()
+		assert.True(t, hasKeyTable)
+	})
```
rootulp, line 47: 👍

rootulp (commented):
+1 to Callum's comment: I think it would be clearer if the module registered it's own param table. Then we can add a unit test to to the x/minfee module to ensure it always registered it's own param table (in addition) to the unit test you added to app_test.go

ninabarbakadze:
> ## Overview
> Fixes #3553



> +1 to Callum's comment: I think it would be clearer if the module registered it's own param table. Then we can add a unit test to to the x/minfee module to ensure it always registered it's own param table (in addition) to the unit test you added to app_test.go

Addressed your recommendations 👍🏼

x/minfee/module_test.go
```diff
@@ -0,0 +1,41 @@
...
+	stateStore.MountStoreWithDB(tStoreKey, storetypes.StoreTypeTransient, nil)
+	require.NoError(t, stateStore.LoadLatestVersion())
+
+	registry := codectypes.NewInterfaceRegistry()
+
+	// Create a params keeper
+	paramsKeeper := paramkeeper.NewKeeper(codec.NewProtoCodec(registry), codec.NewLegacyAmino(), storeKey, tStoreKey)
+	subspace := paramsKeeper.Subspace(minfee.ModuleName)
+
+	// Initialize the minfee module which register set the key table
```
rootulp:
```suggestion
	// Initialize the minfee module which registers the key table
```

rootulp (approved):
Thanks for fixing!

x/minfee/module_test.go
ninabarbakadze: it's very strange to me how sometimes i push a commit fixing/refactoring a piece of code that you still see later on. for example i fixed that comment here https://github.com/celestiaorg/celestia-app/pull/3571/commits/563355e919493e4923b8419b1eae4320f4aad064 and somehow it still showed you an old version 🤔

rootulp: oh hmm I have no idea. I've seen that happen if I start reviewing a PR while the author pushes a commit and I leave a review before refreshing to view what they pushed after I started reviewing.

But in this case it looks like you pushed that commit 3 hours ago and I reviewed ~24 mins ago so I should've seen the latest changes 🤷

ninabarbakadze:
> Thanks nina!
> 
> We should probably look to turn the construction of the state store (needed to initialise the keeper) into some test helper as I see it used often

np! yeah, good idea. exporting into an issue.

PR #3572, opened 2024-06-17
refactor: test/util/common

rootulp:
Completely optional refactor that I made while reviewing this code.

### DIFF
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -10,7 +10,7 @@ import (
 	cosmosmath "cosmossdk.io/math"
 	"github.com/celestiaorg/celestia-app/v2/app"
 	"github.com/celestiaorg/celestia-app/v2/x/blobstream/keeper"
-	bstypes "github.com/celestiaorg/celestia-app/v2/x/blobstream/types"
+	blobstreamtypes "github.com/celestiaorg/celestia-app/v2/x/blobstream/types"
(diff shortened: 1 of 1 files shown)


test/util/common.go
```diff
@@ -290,22 +285,21 @@ func CreateTestEnvWithoutBlobstreamKeysInit(t *testing.T) TestInput {
 		marshaler,
 		keySlashing,
 		&stakingKeeper,
-		getSubspace(paramsKeeper, slashingtypes.ModuleName).WithKeyTable(slashingtypes.ParamKeyTable()),
```
rootulp, line 293: This was the original motivation for reviewing this file. We don't need the `..WithKeyTable(slashingtypes.ParamKeyTable())` here because the slashing keeper does that in the keeper constructor:

```go
// NewKeeper creates a slashing keeper
func NewKeeper(cdc codec.BinaryCodec, key storetypes.StoreKey, sk types.StakingKeeper, paramspace types.ParamSubspace) Keeper {
	// set KeyTable if it has not already been set
	if !paramspace.HasKeyTable() {
		paramspace = paramspace.WithKeyTable(types.ParamKeyTable())
	}
```

cmwaters (approved):
I like that it's slightly more explicit

ninabarbakadze (approved):

Issue #3573, opened 2024-06-17
Unit test prepare proposal


rootulp:
## Context

Informal Systems audit v2 informational finding 5: Proposed block could exceed size limit

## Problem

https://github.com/celestiaorg/celestia-app/blob/main/app/prepare_proposal.go

## Proposal

Add a unit test to ensure that prepare proposal never returns a `ResponsePrepareProposal.BlockData.Txs` where the total size of Txs exceeds the value of `RequestPrepareProposal.BlockDataSize` because if it does then celestia-core will panic [here](https://github.com/rootulp/celestia-core/blob/7e97bb0f2b8747e4ad144589295c078268ce1a89/state/execution.go#L146-L153)

cmwaters:
> Informal Systems audit v2 informational finding 5: Proposed block could exceed size limit

Can you explain how this could happen. I don't believe we ever add data only remove transactions

rootulp:
Correct, the application only ever filters transactions out.

rootulp:
Discussed with @cmwaters offline and we decided this informational audit finding should just be acknowledged and the unit test doesn't add much value.

PR #3584, opened 2024-06-18
test: prepare proposal block data size

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3573

Note: we could add an explicit conditional at the end of prepare proposal like: 

```go
	if getSize(txs) > req.BlockDataSize {
		panic("the size of the proposal block data exceeds the block data size")
	}
```

but I'm not sure that actually gets us much. Instead of panic'ing in celestia-core (like it does currently), it'll panic in celestia-app. Since this is a change to prepare proposal which has already been audited, I'm hesitant to add the conditional unless reviewers feel strongly that it is helpful.

### DIFF
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -25,17 +25,17 @@ import (
 
 func TestPrepareProposalPutsPFBsAtEnd(t *testing.T) {
 	numBlobTxs, numNormalTxs := 3, 3
-	accnts := testfactory.GenerateAccounts(numBlobTxs + numNormalTxs)
-	testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultConsensusParams(), accnts...)
+	accounts := testfactory.GenerateAccounts(numBlobTxs + numNormalTxs)
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultConsensusParams(), accounts...)
 	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
-	infos := queryAccountInfo(testApp, accnts, kr)
+	infos := queryAccountInfo(testApp, accounts, kr)
 
 	blobTxs := blobfactory.ManyMultiBlobTx(
 		t,
 		encCfg.TxConfig,
 		kr,
 		testutil.ChainID,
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):

cmwaters:
Is this some form of defensive programming in the instance that we modify the algorithm such that there is a possibility that bytes are added and that therefore this test would catch it?

app/test/prepare_proposal_test.go
```diff
@@ -205,6 +205,61 @@ func TestPrepareProposalFiltering(t *testing.T) {
...
+			Txs: txs,
+		},
+		BlockDataSize: getSize(txs),
+		ChainId:       testutil.ChainID,
+		Height:        testApp.LastBlockHeight() + 1,
+		Time:          time.Now(),
+	}
+
+	got := testApp.PrepareProposal(request)
+	require.LessOrEqual(t, getSize(got.BlockData.Txs), request.BlockDataSize)
```
rach-id, line 260: [not blocking]
Does it make sense to have a test where the size is bigger than the BlockDataSize? and check if prepare proposal fails?

rach-id (approved):

app/test/prepare_proposal_test.go
rootulp: Unfortunately it won't fail unless we add the explicit check proposed in the PR description.

rootulp:
> Is this some form of defensive programming in the instance that we modify the algorithm such that there is a possibility that bytes are added and that therefore this test would catch it?

Unfortunately I don't think this test would actually catch it because it doesn't fuzz inputs into prepare proposal. If we wanted to be defensive then we would have to add the explicit check in prepare proposal (see PR description).

This PR is motivated by added a test case to for the audit finding. It communicates the "intention" of prepare proposal but doesn't actually enforce any new constraints.

cmwaters:
Yeah, I'm a little confused by it's existence. Are you saying it's there just to communicate to people that we don't expect `PrepareProposal` to somehow increase the bytes returned?

rootulp:
Yea exactly

rootulp:
Discussed with @cmwaters offline and we decided this informational audit finding should just be acknowledged and the unit test doesn't add much value.

Issue #3586, opened 2024-06-18
Limit the recursion for msg gatekeeper authz messages?


rootulp:
An alternative option to the one Callum proposed is:

```diff
func (mgk MsgVersioningGateKeeper) hasInvalidMsg(ctx sdk.Context, acceptedMsgs map[string]struct{}, msgs []sdk.Msg) error {
	for _, msg := range msgs {
		// Recursively check for invalid messages in nested authz messages.
		if execMsg, ok := msg.(*authz.MsgExec); ok {
			nestedMsgs, err := execMsg.GetMessages()
			if err != nil {
				return err
			}
+			if err = mgk.hasNestedMsgExec(ctx, nestedMsgs); err != nil {
+				return err
+			}
			if err = mgk.hasInvalidMsg(ctx, acceptedMsgs, nestedMsgs); err != nil {
				return err
			}
		}

		msgTypeURL := sdk.MsgTypeURL(msg)
		_, exists := acceptedMsgs[msgTypeURL]
		if !exists {
			return sdkerrors.ErrNotSupported.Wrapf("message type %s is not supported in version %d", msgTypeURL, ctx.BlockHeader().Version.App)
		}
	}

	return nil
}

+func (mgk MsgVersioningGateKeeper) hasNestedMsgExec(_ sdk.Context, nestedMsgs []sdk.Msg) error {
+	for _, nestedMsg := range nestedMsgs {
+		if _, ok := nestedMsg.(*authz.MsgExec); ok {
+			return sdkerrors.ErrNotSupported.Wrapf("nested MsgExec is not allowed in MsgExec")
+		}
+	}

	return nil
}
```

but my concern with adding this is that even though a MsgExec inside a MsgExec seems like a strange use case, what if it's actually desirable for certain scenarios. 

I don't see a major risk from not adding this constraint because if a user crafts a deeply nested MsgExec wrapper (i.e. `MsgExec(MsgExec(MsgExec(...)))` then the transaction will be rejected based on tx size.

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/3555#discussion_r1636764149_

rootulp:
The alternative approach would be what @cmwaters originally suggested which is to remove `authz.MsgExec` from the list of acceptedMsgs.

cmwaters:
> but my concern with adding this is that even though a MsgExec inside a MsgExec seems like a strange use case, what if it's actually desirable for certain scenarios.

[...]

PR #3590, opened 2024-06-18
feat!: reject MsgExec inside MsgExec

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3586

cc: @Reecepbcups

### DIFF
--- a/app/ante/msg_gatekeeper.go
+++ b/app/ante/msg_gatekeeper.go
@@ -50,7 +50,7 @@ func (mgk MsgVersioningGateKeeper) hasInvalidMsg(ctx sdk.Context, acceptedMsgs m
 				return err
 			}
 			// App version 2 introduces a new constraint that prevents a MsgExec inside a MsgExec.
-			if ctx.BlockHeader().Version.App > 2 {
+			if ctx.BlockHeader().Version.App >= 2 {
 				if err = mgk.hasNestedMsgExec(ctx, nestedMsgs); err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):

app/ante/msg_gatekeeper.go
```diff
@@ -49,6 +49,12 @@ func (mgk MsgVersioningGateKeeper) hasInvalidMsg(ctx sdk.Context, acceptedMsgs m
...
 			}
+			// App version 2 introduces a new constraint that prevents a MsgExec inside a MsgExec.
+			if ctx.BlockHeader().Version.App >= 2 {
+				if err = mgk.hasNestedMsgExec(ctx, nestedMsgs); err != nil {
+					return err
+				}
+			}
 			if err = mgk.hasInvalidMsg(ctx, acceptedMsgs, nestedMsgs); err != nil {
 				return err
 			}
```
Reecepbcups, line 60: Is `mgk.hasInvalidMsg` still needed since the new nested check would catch it before? (unless ofc this was shipped into prod already w/ v1)

cmwaters: `hasInvalidMsg` is still necessary to disallow `TryMsgUpgrade` within the `MsgExec`

cmwaters (commented):
Hmm this is now getting messy. If we want to prevent recursion then I think this should be it's own antehandler / or separated out from this antehandler in some way.

rootulp:
Closing this as won't do for now. Needs more investigation so created https://github.com/celestiaorg/celestia-app/issues/3609

PR #3611, opened 2024-06-21
chore(e2e test): extends the wait time for genesis nodes to produce the first block in e2e tests #3610

### LINKED ISSUE #3610: Extend the wait time for genesis nodes to produce the first block in e2e tests
## Problem

In the e2e tests, we wait for a total of 1 second for the genesis nodes to progress to height 1, however, this amount is insufficient, given that block times are typically around 10 seconds and more. This causes the e2e benchmark tests fail to be initialized nondeterministically, specially for the big block tests, requiring to rerun the test multiple times. 

## Proposal
Proposal to change this wait time to 10 seconds.

staheri14:
Closes #3610

### DIFF
--- a/test/e2e/testnet/testnet.go
+++ b/test/e2e/testnet/testnet.go
@@ -367,7 +367,7 @@ func (t *Testnet) Start() error {
 				return fmt.Errorf("failed to start node %s", node.Name)
 			}
 			fmt.Printf("node %s is not synced yet, waiting...\n", node.Name)
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(1000 * time.Millisecond)
 		}
 	}
 	return nil


test/e2e/testnet/testnet.go
```diff
@@ -367,7 +367,7 @@ func (t *Testnet) Start() error {
 				return fmt.Errorf("failed to start node %s", node.Name)
 			}
 			fmt.Printf("node %s is not synced yet, waiting...\n", node.Name)
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(1000 * time.Millisecond)
```
rootulp, line 370: [nit] seems more readable
```suggestion
			time.Sleep(10 * time.Second)
```

rootulp (dismissed):

test/e2e/testnet/testnet.go
staheri14: I have updated the unit from milliseconds  to seconds.
Since this loop iterates 10 times and checks the node's status in each iteration, to achieve a total wait time of 10 seconds, the wait time for each iteration should be adjusted to 1 second.

ninabarbakadze (approved):

rootulp (approved):

evan-forbes (approved):

Issue #3371, opened 2024-04-23
Dedicated feature branch for e2e benchmark tests with main branch sync


staheri14:
In the Knuu e2e benchmark tests, we sometimes encounter the need to use versions of celestia-core that include unreleased breaking changes or features not fully integrated into celestia-app. This situation becomes a barrier to developing and merging benchmark tests into the celestia-app main branch. To address this issue, a decision has been made to establish a feature branch specifically for benchmark tests. Future PRs will be directed to this branch, facilitating seamless integrations without affecting the main branch of celestia-app.

Additionally, it would be useful to have an automated way to backport features from the main branch to this new feature branch, ensuring it remains up-to-date with the latest developments in the main.

evan-forbes:
the goal is to merge this feature branch after v2

evan-forbes:
we can unblock with https://github.com/celestiaorg/celestia-app/issues/3370

staheri14:
> we can unblock with #3370

And https://github.com/celestiaorg/celestia-core/issues/1363

Issue #3557, opened 2024-06-12
Merging the knuu network tests feature branch


evan-forbes:
We have a branch that was created to help unblock the network tests in knuu, and now we need to merge it. Below is a list of issues that are classified as either needed or nice to have. All issues in the needed section are required to be closed to also close this issue. Please feel free to edit as you see fit @staheri14 @smuu 

### Needed
- [x] #3589 
  - [x]  https://github.com/celestiaorg/celestia-app/pull/3591
- [x] https://github.com/celestiaorg/celestia-app/issues/3437
  - [x] https://github.com/celestiaorg/celestia-app/pull/3541
- [x] https://github.com/celestiaorg/celestia-app/pull/3505
- [x] https://github.com/celestiaorg/celestia-app/pull/3515
- [x] https://github.com/celestiaorg/celestia-app/issues/3588
  - [x]  #3595
- [x] https://github.com/celestiaorg/celestia-app/pull/3612
  - [x] https://github.com/celestiaorg/celestia-app/pull/3415 (its advanced version was merged by #3612)

PR #3541, opened 2024-06-07
fix: increases txsim grpc client max receive message size

staheri14:
Closes #3437

PR #3612, opened 2024-06-21
test: adding big block tests to the main branch

### LINKED ISSUE #3638: Improve E2E Benchmark Test Correctness Verification to Avoid False Positives
## Problem
The current method of confirming the correctness of benchmark tests is prone to false positives. Presently, a transaction count exceeding 10 is used as an indicator that the network has properly progressed and that enough transactions have been submitted. However, this approach can lead to incorrect conclusions. Depending on the number of sequences per txclient, merely initializing the subaccounts for those sequences can result in more than 10 transactions. In such cases, while the tx count condition is met, yet, the actual submitted PFBs may not have reached the network due to reasons such as large transaction size or other network issues [it has been witnessed in the recent experiments].

## Acceptance Criteria
- [ ] The transaction count check must be replaced with a method that more accurately reflects network progress and expected behavior. A potential solution is to verify whether the maximum block size has been reached for at least one block during the experiment.

staheri14:
Part of #3557
Closes https://github.com/celestiaorg/celestia-app/issues/3614
Closes #3638 
Closes https://github.com/celestiaorg/celestia-app/issues/3371


test/e2e/benchmark/throughput.go
```diff
@@ -1,21 +1,58 @@
...
+	},
+	TxClientsResource: testnet.Resources{
+		MemoryRequest: "1Gi",
+		MemoryLimit:   "3Gi",
+		CPU:           "2",
+		Volume:        "1Gi",
+	},
+	SelfDelegation:     10000000,
+	CelestiaAppVersion: "pr-3541", // "pr-3261",
+	TxClientVersion:    "pr-3541", //"pr-3261",
```
staheri14, line 36: We can point it to the main as soon as #3541 gets merged.

staheri14:
The markdown linter issue should be resolved by #3639

staheri14:
I have tested all the benchmark tests in this branch. However, I made a few recent changes (a few hours ago) that I am still re-running tests for them to ensure they haven't caused any regression. I will mark the PR as a draft, but any early review is welcome!

staheri14:
I re-ran a subset of benchmark tests, and they looked good, hence opening this PR for review.

staheri14:
The CI failures will be resolved by #3653

test/e2e/testnet/node.go
```diff
@@ -49,14 +49,14 @@ type Node struct {
 	traceProxyHost string
 }
 
-// PullRoundStateTraces retrieves the round state traces from a node.
+// PullBlockSummaryTraces retrieves the block summary traces from a node.
 // It will save them to the provided path.
-func (n *Node) PullRoundStateTraces(path string) ([]trace.Event[schema.RoundState], error,
+func (n *Node) PullBlockSummaryTraces(path string) ([]trace.Event[schema.BlockSummary], error,
```
evan-forbes, line 54: do we care that this is technically breaking? (test! instead of test)

test/e2e/benchmark/manifest.go
```diff
@@ -95,3 +96,21 @@ func (m *Manifest) GetConsensusParams() *tmproto.ConsensusParams {
...
 	return cparams
 }
+
+// summary generates a summary of the Manifest struct to be used as chain id.
+func (m *Manifest) summary() string {
+	latency := 0
+	if m.EnableLatency {
+		latency = 1
+	}
+	maxBlockMB := m.MaxBlockBytes / mb
```
evan-forbes, line 106: are `mb` and `MB` duplicates? https://github.com/celestiaorg/celestia-app/blob/77042f80d51b32afac48201a4bcf956976a14f8f/test/e2e/testnet/defaults.go#L12

I didn't look, but I'm assuming its some import cycle thing, if so is it possible to remove that in a different PR?

evan-forbes (commented):
could we do the refactors coderabbit is suggesting to simplify the tests? if so, are there downsides to that?

evan-forbes (dismissed):

test/e2e/benchmark/manifest.go
staheri14: Not actually, there is no import cycle, it is now using the testnet's constants https://github.com/celestiaorg/celestia-app/pull/3612/commits/d93c2aad8914c58324936b4e2f8698ecd618e906

test/e2e/testnet/node.go
staheri14: Good point, I retracted the old `PullRoundStateTraces` func to make this PR non-breaking.

evan-forbes (approved):

test/e2e/benchmark/throughput.go
```diff
@@ -1,26 +1,65 @@
...
+}
+
 func TwoNodeSimple(logger *log.Logger) error {
 	latestVersion, err := testnet.GetLatestVersion()
 	testnet.NoError("failed to get latest version", err)
 
-	logger.Println("=== RUN TwoNodeSimple", "version:", latestVersion)
+	testName := "TwoNodeSimple"
+	logger.Printf("Running %s\n", testName)
+	logger.Println("version", latestVersion)
```
rootulp, line 62: is this missing a formatting directive?

```suggestion
	logger.Printf("version %v\n", latestVersion)
```

test/e2e/benchmark/benchmark.go
```diff
@@ -118,3 +121,56 @@ func (b *BenchmarkTest) Run() error {
...
+			return fmt.Errorf("expected app version %d, got %d", appconsts.LatestVersion, block.Version.App)
+		}
+		size := int64(block.Size())
+		if size >= expectedBlockSizeBytes {
+			targetSizeReached = true
+			break
+		}
+		if size > maxBlockSize {
+			maxBlockSize = size
+		}
```
rootulp, line 169: nit: I don't think these lines will be executed if the previous conditional evaluates to `true` b/c the `break` so consider re-ordering this above the previous conditional so that `maxBlockSize` accounts for the block that actually exceeds >= expectedBlockSizeBytes

rootulp (approved):

staheri14:
The markdown linter complaint is unrelated to this PR and will be resolved independently in https://github.com/celestiaorg/celestia-app/issues/3668

test/e2e/benchmark/benchmark.go
staheri14: The current version is correct as well. I see what causes confusion here: the value of `maxBlockSize` is only used when the target size is not reached. That is, `if size >= expectedBlockSizeBytes` is never hit, and if that condition is never met, then `maxBlockSize` correctly reflects the most recent maximum size observed. If we do hit the `expectedBlockSizeBytes` condition and `if size >= expectedBlockSizeBytes` evaluates to true, then the value of `maxBlockSize` does not matter. Nevertheless, to resolve future confusion, I will apply your suggestion in a follow up PR, thanks!

test/e2e/benchmark/throughput.go
staheri14: They both indeed work identically:
The `Println` function prints the supplied inputs with comma separation, whereas in `Printf`, I achieved the same result using formatting. Here is a sample output from the tests:
```
test-e2e-benchmark2024/07/08 11:34:05 Running TwoNodeSimple
test-e2e-benchmark2024/07/08 11:34:05 version 8caa580
```
Their implementation is just inconsistent, which I can fix in a follow-up PR.

test/e2e/benchmark/benchmark.go
staheri14: Please see https://github.com/celestiaorg/celestia-app/pull/3670

test/e2e/benchmark/throughput.go
staheri14: Please see https://github.com/celestiaorg/celestia-app/pull/3670

Issue #3587, opened 2024-06-18
`make test-interchain` fails on main


rootulp:
## Problem

```
$ make test-interchain
--> Running interchain tests
main module (github.com/celestiaorg/celestia-app/v2) does not contain package github.com/celestiaorg/celestia-app/v2/test/interchain
make: *** [test-interchain] Error 1
```

I think this may have been caused by adding the `v2` suffix to package names but I'm not 100% sure

## Acceptance criteria

- [ ] Fix `make test-interchain` so that it can run on `main`

PR #3623, opened 2024-06-25
fix: make test-interchain

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3587

## Testing

```
$ make test-interchain
--> Running interchain tests
=== RUN   TestInterChainAccounts
    setup.go:203: Pruned 3 volumes, reclaiming approximately 5.0 MB
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ test-fuzz:
 ## test-interchain: Run interchain tests in verbose mode. Requires Docker.
 test-interchain:
 	@echo "--> Running interchain tests"
-	@go test ./test/interchain -v
+	@cd ./test/interchain && go test . -v
 .PHONY: test-interchain
 
 ## txsim-install: Install the tx simulator.


ninabarbakadze (approved):
thank you 🙏🏻

Makefile
```diff
@@ -168,7 +168,7 @@ test-fuzz:
 ## test-interchain: Run interchain tests in verbose mode. Requires Docker.
 test-interchain:
 	@echo "--> Running interchain tests"
-	@go test ./test/interchain -v
+	@cd ./test/interchain && go test . -v
```
ninabarbakadze, line 171: do we need the "." ?

cmwaters (approved):
I had the same problem. Thanks for fixing. Is this part of CI already?

rootulp:
Not part of CI yet but I'll work on adding it separately.

staheri14 (approved):
utAck

Makefile
rootulp: - `go test .` runs only tests in the current directory, 
- `go test` runs tests in the current package directory, but if run from a module root, it will run tests in all packages in the module.

In this scenario, there's only one test in that directory and both commands run it.

Issue #3627, opened 2024-06-26
Add a log statement for v1 -> v2 upgrade


rootulp:
## Context

At the `--v2-upgrade-height` a consensus node operators sees the following logs:

```log
6:24PM INF finalizing commit of block hash={} height=4 module=consensus num_txs=1 root=2F3D9603863BA5781CACA6BF9B1DDFA20D87997C6D49BCA38277C3073034BE57
6:24PM INF minted coins from module account amount=27946719utia from=mint module=x/bank
6:24PM INF executed block height=4 module=state num_invalid_txs=0 num_valid_txs=1
6:24PM INF adding a new module: interchainaccounts
6:24PM INF created new capability module=ibc name=ports/icahost
6:24PM INF port binded module=x/ibc/port port=icahost
6:24PM INF claimed capability capability=2 module=icahost name=ports/icahost
6:24PM INF adding a new module: minfee
6:24PM INF adding a new module: packetfowardmiddleware
6:24PM INF adding a new module: signal
6:24PM INF commit synced commit=436F6D6D697449447B5B3232372032303620313534203139362031203738203336203133392035203234312036392032353320323134203232322031333320323234203139203133302039312031323320313338203735203239203631203531203420313539203138352031383120393620313231203131385D3A347D
6:24PM INF committed state app_hash=E3CE9AC4014E248B05F145FDD6DE85E013825B7B8A4B1D3D33049FB9B5607976 height=4 module=state num_txs=1
6:24PM INF indexed block exents height=4 module=txindex
6:25PM INF Timed out dur=10978.895468 height=5 module=consensus round=0 step=1
6:25PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"B7C6AC32AF6ADEF0DBBCF1A25CBA58538FBC5E0A7AEE7251F705990FD7D697C0","parts":{"hash":"C52415B6F0D38C97CFCC41283FE91E593865D6BF7999BA8CC37B4B2BBC19E209","total":1}},"height":5,"pol_round":-1,"round":0,"signature":"8zlCAqYeMookLUCl2PPflV4+VD9od9Sj569kc4x2a2wkGPE2gnN+6EkUjlYZc26+CM6aqscYtpseVIdhdUk7DA==","timestamp":"2024-06-26T18:25:01.097207476Z"} proposer=D8807A87724448A7C93AE2CCC44CEBC4B858D2B0
[...]

PR #3628, opened 2024-06-26
chore: log v2 upgrade

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3627

## Testing

Confirmed the log shows up during an upgrade

```
4:29PM INF upgrading from app version 1 to 2
4:29PM INF executed block height=2 module=state num_invalid_txs=0 num_valid_txs=0
4:29PM INF adding a new module: interchainaccounts
4:29PM INF created new capability module=ibc name=ports/icahost
4:29PM INF port binded module=x/ibc/port port=icahost
4:29PM INF claimed capability capability=2 module=icahost name=ports/icahost
4:29PM INF adding a new module: minfee
4:29PM INF adding a new module: packetfowardmiddleware
4:29PM INF adding a new module: signal
```


evan-forbes (approved):

cmwaters (approved):
Do we want a log at the end when the upgrade has been completed or people just wait for the next block to be produced

rootulp:
> Do we want a log at the end when the upgrade has been completed or people just wait for the next block to be produced

Hmm good point. We can add a log in `BeginBlock` so that it shows up when the next block is produced.

PR #3637, opened 2024-06-27
feat(tools): blockheight

rootulp:
Extremely optional tool I used to determine the block height that a Robusta chain should use for `--v2-upgrade-height`. We may want to use a variation of this for testnets when we roll-out v2 there.

## Testing

```
$ go run main.go
currentHeight: 5627
currentTime: 2024-06-28 10:42:03 -0400 EDT
targetTime: 2024-07-04 12:30:00 -0400 EDT
diffInSeconds: 524877
diffInBlockHeight: 44256
targetHeight: 49883
```

### DIFF
--- a/tools/blockheight/README.md
+++ b/tools/blockheight/README.md
@@ -1,0 +1,15 @@
+# blockheight
+
+blockheight is a small tool to help predict the block height of a chain at a given time.
+
+## Usage
+
+Update values in `main.go` and then run it:
+
+```go
+$ go run main.go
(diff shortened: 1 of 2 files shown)


cmwaters (commented):
I thought this tool would use RPC to read the current height and time of a chain and then the argument you would provide a future time or duration and it would spit out the calculated height

rootulp:
yea that could also work. Since this PR didn't get approvals, I'm guessing ppl don't want it so will close.

PR #3642, opened 2024-06-28
fix!: use POST for pfb submission

cmwaters:
Instead of `GET` so the user can actually add the blob in the body


evan-forbes (approved):

ninabarbakadze (approved):
thanks cal

rootulp:
Silly question, where does this `POST` get used? How can I test it?

Issue #3472, opened 2024-05-12
Populate `consensus_params.version` for exported app


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/3450 fixed a few issues to make it possible to invoke `celestia-appd export`

## Problem

The exported [output](https://gist.github.com/rootulp/dfea2b5b40f7366b03706fc39321ceee) doesn't populate `consensus_params.version`

Note: this isn't a big deal b/c we could populate this field manually.

## Proposal

Populate the `consensus_params.version` field for an exported app.

loglapa:
I'd like to take this. I reviewed the closed #6007 and will avoid its type/build issue. The plan is to populate the exported consensus params' app version using the current proto type and canonical app version, preserve a non-zero version if one is already present, and add a focused export regression test before opening the PR.

PR #3650, opened 2024-07-02
fix export app-version

najeal:
Related to #3472 

## Overview

ParamStoreKeyVersionParams key was never set in paramStore.
The changes forces storage of app-version and fix test for exporting version 2.


rootulp (dismissed):
This looks good to me, thanks! I plan on manually testing this PR by stop + starting a node before and after the upgrade height b/c persisting the app version has a lot of edge cases.

rootulp (commented):
Not sure how to explain this b/c the unit test seems to pass but when I manually tried this:

```shell
./scrips/single-node.sh

// wait until after block height 3
// stop node
celestia-appd export
```

the resulting JSON does not populate the app version in consensus params:

```json
		"version": {

		}
```

najeal:
@rootulp yes I need to open a PR an cosmos-sdk side to close the issue.
This PR partially resolves the issue as we are now providing an expected v1 or v2 version. Previously only v1 was working regarding the tests.
But cosmos-sdk still does not populate this value.

najeal:
In addition to my previous anwer. This is where Version is not populated:
https://github.com/celestiaorg/cosmos-sdk/blob/83c372bc533d09c332c9dc0ee422ee8d087443b1/server/export.go#L83-L97

rootulp:
Thanks! I think we'll have to make a PR to https://github.com/celestiaorg/cosmos-sdk first, get it merged, cut a release, bump to that version in celestia-app and then proceed with this PR so will mark as draft for now.

app/export.go
```diff
@@ -19,7 +19,13 @@ func (app *App) ExportAppStateAndValidators(forZeroHeight bool, jailAllowedAddrs
 		return servertypes.ExportedApp{}, err
 	}
 
-	app.InitializeAppVersion(ctx)
+	versionToInit := v1
+	if app.LastBlockHeight() >= app.upgradeHeightV2 {
+		versionToInit = v2
+	}
```
cmwaters, line 25: This won't work if the user hasn't set the appropriate flag. I would prefer trying to read the consensus params to know what the version is

rootulp:
@najeal celestia-app bumped to the version of cosmos-sdk with your fix on all 3 supported branches (`main`, `v1.x`, and `v2.x`) so your PR should be unblocked.

najeal:
@rootulp normally, the PR just needs to uncomment the test right now, so my first commit can be deleted.
One problem is, when uncommenting, the test fails whereas testing from using the script `single-node.sh` works.
It looks like the problem comes from the test state itself, not the feature.
I will investigate a little bit more

najeal:
There is some weird stuff. Checking version from param store after the upgrade: `version := testApp.GetAppVersionFromParamStore(ctx)` fails. It falls in the `ParamStoreKeyVersionParams` key not stored logic.
But looking with the debugger, this key is stored during the upgrade with v2 value.

Either I don't initialise the context the way I should or the store is reset during the commit of the upgrade (see: cosmos-sdk/store/rootmulti.store.go  `loadVersion(ver int64, upgrades *types.StoreUpgrades)` function). What do you think?
I suspect new param store to be created without old values moved in.

PR #3653, opened 2024-07-02
chore: fixes the link to namespace implementation 

staheri14:
This link was previously fixed, but apparently, the file for namespace implementation has been moved again (the directory name has changed). Hence, here is the new fix.


specs/src/specs/namespace.md
```diff
@@ -94,7 +94,7 @@ Among the potential consequences is the _Woods Attack_, as elaborated in this fo
 
 ## Implementation
 
-See [go-square/namespace](https://github.com/celestiaorg/go-square/blob/main/shares/namespace.go).
+See [go-square/namespace](https://github.com/celestiaorg/go-square/blob/main/share/namespace.go).
```
staheri14, line 97: I intentionally didn't use a permalink to ensure that we always track its most recent location and receive errors in CIs if it changes.

rootulp: [no change needed] it's a bit confusing currently b/c we don't have versioned specs so it's not immediately clear if these specs should describe the celestia-app v1.x behavior or v2.x behavior. A permalink to the previous implementation(s) is correct for v1.x and v2.x. A link to the new location (this PR) isn't actually correct yet because celestia-app doesn't use `main` of go-square. There is no official release of go-square with those breaking changes so if we want to be 100% accurate, we probably shouldn't point to them from these specs yet. Not super opinionated about it and your change future-proofs the specs a bit so seems fine to me.

rootulp (dismissed):

specs/src/specs/namespace.md
staheri14: Good point. Considering this new information, I have slightly revised the text and provided two links: one to the version of the code that matches the current specifications, and one pointing to the most recent version of the code.

specs/src/specs/namespace.md
```diff
@@ -94,7 +94,8 @@ Among the potential consequences is the _Woods Attack_, as elaborated in this fo
 
 ## Implementation
 
-See [go-square/namespace](https://github.com/celestiaorg/go-square/blob/main/shares/namespace.go).
+See the [namespace implementation in go-square](https://github.com/celestiaorg/go-square/blob/be3c2801e902a0f90f694c062b9c4e6a7e01154e/namespace/namespace.go).
```
staheri14, line 97: The permalink provided here is based on [this file](https://github.com/celestiaorg/go-square/blob/v1.1.0/namespace/namespace.go) from v1.1.0 of go-square.

evan-forbes (approved):

rootulp (approved):

Issue #3624, opened 2024-06-26
AppHash mismatch test between main and v2 


ninabarbakadze:
## Description

Currently we have an AppHash mismatch test on main and v1.x executing state machine messages on v1. We should have the same test for v2.

## Acceptance Criteria

- [ ] We have a consistent apphash test between v2 and main executing v2 state machine messages
- [ ] Refactor the existing test to table driven test

rootulp:
For clarification: the [consistent app hash test](https://github.com/celestiaorg/celestia-app/blob/9475c4bcb9f9cb3f27c9be054201303526c11f2f/app/test/consistent_apphash_test.go#L38-L42) on `main` executes v1 state machine messages and compares against an app hash generated from executing the same messages on the `v1.x` branch.

Which git branch do you propose adding the test in this issue to (presumably `main`)? If `main` where you plan to source the app hash used to compare against (presumably `v2.0.0-rc2`)?

rootulp:
@ninabarbakadze does this issue include a new app hash test for v3? If not, do we need a separate issue for that?
AFAIK v3 doesn't introduce any new message types so IMO it doesn't need one but if that's the case, idk why this issue blocks v3.

rootulp:
Answering my own questions from above:

> Which git branch do you propose adding the test in this issue to (presumably main)? 

main

> If main where you plan to source the app hash used to compare against (presumably v2.0.0-rc2)?

Looks like v2.x based on [this line](https://github.com/celestiaorg/celestia-app/pull/3879/files#diff-de1da0384e69ff8f40e98d1cd0fa0520d40b0ef1afff1b74e2be009820ce30d7R91) but unclear if it was v2.1.2 or an earlier release.

PR #3665, opened 2024-07-05
test(v1.x): expand apphash test with all state machine msgs

ninabarbakadze:
## Overview

Consistent Apphash test on v1


app/consistent_apphash_test.go
```diff
@@ -3,134 +3,303 @@ package app_test
...
 
 	"github.com/celestiaorg/celestia-app/app"
 	"github.com/celestiaorg/celestia-app/app/encoding"
 	"github.com/celestiaorg/celestia-app/pkg/appconsts"
-	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
 	"github.com/celestiaorg/celestia-app/pkg/user"
 	testutil "github.com/celestiaorg/celestia-app/test/util"
 	"github.com/celestiaorg/celestia-app/test/util/blobfactory"
+	"github.com/celestiaorg/celestia-app/test/util/testfactory"
+
```
rootulp, line 15: does your editor automatically add these newlines between imports? They seem unecessary to me. If all the imports are in one block, then golangci-lint will order them correctly.

```suggestion
```

app/consistent_apphash_test.go
```diff
@@ -3,134 +3,303 @@ package app_test
...
+	appns "github.com/celestiaorg/celestia-app/pkg/namespace"
 	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/crypto/hd"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"github.com/cosmos/cosmos-sdk/crypto/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
```
rootulp, line 22:
```suggestion
```

app/consistent_apphash_test.go
```diff
@@ -3,134 +3,303 @@ package app_test
...
+	blobtypes "github.com/celestiaorg/celestia-app/x/blob/types"
+	"github.com/cosmos/cosmos-sdk/x/authz"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	crisisTypes "github.com/cosmos/cosmos-sdk/x/crisis/types"
+	distribution "github.com/cosmos/cosmos-sdk/x/distribution/types"
+	"github.com/cosmos/cosmos-sdk/x/feegrant"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	slashingtypes "github.com/cosmos/cosmos-sdk/x/slashing/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+
```
rootulp, line 32:
```suggestion
```

app/consistent_apphash_test.go
```diff
@@ -3,134 +3,303 @@ package app_test
...
 type BlobTx struct {
 	author    string
 	blobs     []*tmproto.Blob
 	txOptions []user.TxOption
 }
 
-// TestConsistentAppHash executes a set of txs, generates an app hash,
-// and compares it against a previously generated hash from the same set of transactions.
+// TestConsistentAppHash executes all state machine messages, generates an app hash and a data root,
+// and compares it against a previously generated hashes from the same set of transactions.
```
rootulp, line 48:
```suggestion
// and compares it against previously generated hashes from the same set of transactions.
```

evan-forbes (approved):
doesn't have to be changed here, but I still think that pursuing a table driven approach is best here, even if that means we have to add some functionality in testapp

the main reasons were similar to what we've discussed before, just that adding things to table driven tests is intuitive and we can isolate msgs. atm this test can break and we don't immediately know which module changed its state machine in a breaking way

we also don't have the ability to change the app version, which presumably is something else that we would need if we want to make it easy on ourselves in the future when we expand these tests

app/consistent_apphash_test.go
```diff
@@ -3,134 +3,303 @@ package app_test
...
 
-	// Create Blob Tx
-	rawBlobTx, _, err := signer.CreatePayForBlobs(blobTx.author, blobTx.blobs, blobTx.txOptions...)
+	// NewMsgWithdrawDelegatorReward - withdraws delegator rewards
+	msgWithdrawDelegatorReward := distribution.NewMsgWithdrawDelegatorReward(accountAddresses[0], genValidators[0].GetOperator())
+	secondBlockSdkMsgs = append(secondBlockSdkMsgs, msgWithdrawDelegatorReward)
+
+	// ------------ Third Block ------------
+
+	// Txs within the third block are signed by the validator's signer
```
rootulp, line 250: > are signed by the validator's signer

is this unique to the third block? If not, why describe it here and not for blocks one and two?

app/consistent_apphash_test.go
```diff
@@ -172,16 +341,119 @@ func deterministicKeyRing(cdc codec.Codec) (keyring.Keyring, []types.PubKey) {
...
+
+	// Deliver SDK Txs
+	for i, tx := range encodedSdkTxs {
+		resp := testApp.DeliverTx(abci.RequestDeliverTx{Tx: tx})
+		if resp.Code != abci.CodeTypeOK {
+			return nil, nil, fmt.Errorf("DeliverTx failed for the message at index %d: %s", i, resp.Log)
+		}
+	}
+
+	// Deliver Blob Txs
```
rootulp, line 433:
```suggestion
	// Deliver Blob Tx
```

rootulp (approved):

app/consistent_apphash_test.go
ninabarbakadze: it is unique to the third block

ninabarbakadze:
> doesn't have to be changed here, but I still think that pursuing a table driven approach is best here, even if that means we have to add some functionality in testapp
> 
> the main reasons were similar to what we've discussed before, just that adding things to table driven tests is intuitive and we can isolate msgs. atm this test can break and we don't immediately know which module changed its state machine in a breaking way
> 
> we also don't have the ability to change the app version, which presumably is something else that we would need if we want to make it easy on ourselves in the future when we expand these tests

I'd say that I can revisit this when i have to write consistent apphash test for v2. We can then figure out how we want different versions to be tested + the structure. Then I'd also refactor this to the table driven approach. Updated the AC on the v2 testing issue: https://github.com/celestiaorg/celestia-app/issues/3624

Although the way i see this test breaking mostly is by application producing a different hash to what's expected because we accidentally made an apphash breaking change to the state machine.

app/consistent_apphash_test.go
```diff
@@ -197,6 +466,44 @@ func queryAccountInfo(capp *app.App, accs []string, kr keyring.Keyring) []blobfa
...
+	rec, err := kr.Key(account)
+	if err != nil {
+		panic(err)
+	}
+	addr, err := rec.GetAddress()
+	if err != nil {
+		panic(err)
+	}
+	return addr
+}
```
cmwaters, line 479: I thought we already had a function like this in test util package

cmwaters (approved):

evan-forbes (approved):

app/consistent_apphash_test.go
ninabarbakadze: on main we do

PR #3672, opened 2024-07-08
test: improves e2e tests performance by reading the block metadata instead of the entire block data

### LINKED ISSUE #3667: Improving benchtest performance by reading the block metadata instead of the entire block data
# Problem

At the end of benchmark tests, the entire blockchain is read to conduct some sanity check on the entire experiment e.g., if certain tx count or block size is met. However, the same sanity check can be done without pulling all the blocks data. 

# Proposal
- Only fetch the metadata related to each block to perform the sanity checks.

staheri14:
Closes #3667

I also used this PR to make a change to E2ESimple to address a flaky behavior. The sequence count of the txclient has been increased to 10 to ensure there are sufficient transactions at the end of the test.


test/e2e/simple.go
```diff
@@ -31,7 +31,8 @@ func E2ESimple(logger *log.Logger) error {
 	logger.Println("Creating txsim")
 	endpoints, err := testNet.RemoteGRPCEndpoints()
 	testnet.NoError("failed to get remote gRPC endpoints", err)
-	err = testNet.CreateTxClient("txsim", testnet.TxsimVersion, 1, "100-2000", 100, testnet.DefaultResources, endpoints[0])
+	err = testNet.CreateTxClient("txsim", testnet.TxsimVersion, 10,
```
staheri14, line 34: [To reviewers] The sequence count has been increased to 10 to address a flaky behavior in the E2ESimple test. This change ensures there are sufficient transactions at the end of the experiment.

rootulp (approved):
LGTM. Conventional commit prefix can prob use `test:` instead of `feat:` b/c only test files are modified.

test/util/testnode/read.go
```diff
@@ -39,6 +40,25 @@ func ReadBlockchain(ctx context.Context, rpcAddress string) ([]*types.Block, err
...
+	client, err := http.New(rpcAddress, "/websocket")
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Status(ctx)
+	if err != nil {
+		return nil, err
+	}
+	lastHeight := resp.SyncInfo.LatestBlockHeight
+	res, err := client.BlockchainInfo(ctx, 0, lastHeight)
```
cmwaters, line 55: This will on return up to 20 blocks so it will mean we miss data if the chain is longer than 20

staheri14: Mind elaborating why `20`? it reads from the first block up to the `lastHeight`, so the entire history of the chain

staheri14: Never mind, I see your point, the implementation limits the max height to 20, thanks, will work on a different method then.

staheri14: I refactored the code, and now we can read the entire chain's headers.

staheri14:
I also ran `E2ESimple` and can confirm it works fine with this new change.

rootulp (approved):

smuu (approved):
Nice once

evan-forbes (dismissed):

test/util/testnode/read.go
```diff
@@ -39,6 +39,63 @@ func ReadBlockchain(ctx context.Context, rpcAddress string) ([]*types.Block, err
...
+			break
+		}
+
+		// set the new maxHeight to fetch the next batch of headers
+		maxHeight = lastFetchedHeight - 1
+
+	}
+
+	// reverse the order of headers to be ascending (lowest first).
+	reverseSlice(blockHeaders)
```
rach-id, line 87: [nit|personal opinion|feel free to ignore]
might be refactored to:

```suggestion
	blockchainInfoLimit := int64(20) // taken from: celestia-core/rpc/core/blocks.go: L32
	for startHeight := int64(1); startHeight < maxHeight; startHeight += blockchainInfoLimit {
		endHeight := startHeight + blockchainInfoLimit - 1
		res, err := client.BlockchainInfo(ctx, startHeight, endHeight)
		if err != nil {
			return nil, err
		}

		metas := res.BlockMetas
		reverseSlice(metas)
		blockHeaders = append(blockHeaders, metas...)
	}
	
	if int64(len(blockHeaders)) != maxHeight {
		// this in case we want all the headers, and we don't want to miss any. 
		// because BlockchainInfo might not return all the headers if the node is not synced or pruned
		// without the API returning an error.
		return nil, fmt.Errorf("couldn't fetch all the necessary headers")
	}
```

test/util/testnode/read.go
```diff
@@ -39,6 +39,63 @@ func ReadBlockchain(ctx context.Context, rpcAddress string) ([]*types.Block, err
...
+
+	return blockHeaders, nil
+}
+
+// reverseSlice reverses the order of elements in a slice in place.
+func reverseSlice[T any](s []T) {
+	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
+		s[i], s[j] = s[j], s[i]
+	}
+}
```
rach-id, line 97: I guess some tests could be nice to have for this. Some chatGPT table tests I just generated:
```
func TestReverseSlice(t *testing.T) {
	tests := []struct {
		input    interface{}
		expected interface{}
	}{
		{[]int{1, 2, 3, 4, 5}, []int{5, 4, 3, 2, 1}},
		{[]string{"a", "b", "c", "d"}, []string{"d", "c", "b", "a"}},
		{[]int{1, 2}, []int{2, 1}},
		{[]int{1}, []int{1}},
		{[]string{}, []string{}},
	}

	for _, tt := range tests {
		switch v := tt.input.(type) {
		case []int:
			reverseSlice(v)
			if !reflect.DeepEqual(v, tt.expected) {
				t.Errorf("reverseSlice(%v) = %v, want %v", tt.input, v, tt.expected)
			}
		case []string:
			reverseSlice(v)
			if !reflect.DeepEqual(v, tt.expected) {
				t.Errorf("reverseSlice(%v) = %v, want %v", tt.input, v, tt.expected)
			}
		}
	}
}
```

staheri14: Thanks, If you don't mind, I'll add it in a follow up PR.

test/util/testnode/read.go
staheri14: Thank you for your suggestion. I initially intended to do it this way, but I realized the 20 block limit might not always apply. Therefore, I adopted an approach that isn't dependent on the hardcoded 20 block limit (to make it future proof).

test/util/testnode/read.go
staheri14: Initially, I intended to address this in a follow-up PR to maintain the current approvals. However, I mistakenly dismissed them. Therefore, I will add the tests in the current PR.

staheri14: Please see https://github.com/celestiaorg/celestia-app/pull/3672/commits/56dc5507a83fd56e013baeaca6dfe87a7a560060

rootulp (dismissed):

test/util/testnode/read.go
rach-id: I see 👍

cmwaters (approved):

evan-forbes (approved):

test/util/testnode/read.go
```diff
@@ -39,6 +39,63 @@ func ReadBlockchain(ctx context.Context, rpcAddress string) ([]*types.Block, err
 	return ReadBlockHeights(ctx, rpcAddress, 1, status.SyncInfo.LatestBlockHeight)
 }
 
+// ReadBlockchainHeaders retrieves the blockchain headers from height 1 up to
+// latest available height from the node at rpcAddress and returns it.
+// The headers are returned in ascending order (lowest first).
+func ReadBlockchainHeaders(ctx context.Context, rpcAddress string) ([]*types.BlockMeta, error) {
```
rootulp, line 45: [nit] this function name is a bit misleading b/c `Header` is a distinct struct in celestia-core. This function returns a list of `BlockMeta` so IMO it could be:

```suggestion
func ReadBlockMetas(ctx context.Context, rpcAddress string) ([]*types.BlockMeta, error) {
```

test/util/testnode/read_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReverseSlice(t *testing.T) {
```
rootulp, line 11: [optional] `TestReverseSlice` makes me think that this tests the exported function `ReverseSlice`. But it actually tests `reverseSlice` so this could be

```suggestion
func Test_reverseSlice(t *testing.T) {
```

test/e2e/simple.go
```diff
@@ -43,15 +44,17 @@ func E2ESimple(logger *log.Logger) error {
 	logger.Println("Waiting for 30 seconds to produce blocks")
 	time.Sleep(30 * time.Second)
 
-	blockchain, err := testnode.ReadBlockchain(context.Background(), testNet.Node(0).AddressRPC())
-	testnet.NoError("failed to read blockchain", err)
+	logger.Println("Reading blockchain headers")
+	blockchain, err := testnode.ReadBlockchainHeaders(context.Background(), testNet.Node(0).AddressRPC())
```
rootulp, line 48: [nit] related to other comment about renaming function to ReadBlockMetas

```suggestion
	blockMetas, err := testnode.ReadBlockchainHeaders(context.Background(), testNet.Node(0).AddressRPC())
```

test/e2e/simple.go
```diff
@@ -43,15 +44,17 @@ func E2ESimple(logger *log.Logger) error {
...
+	testnet.NoError("failed to read blockchain headers", err)
 
 	totalTxs := 0
-	for _, block := range blockchain {
-		if appconsts.LatestVersion != block.Version.App {
-			return fmt.Errorf("expected app version %d, got %d in block %d", appconsts.LatestVersion, block.Version.App, block.Height)
+	for _, blockMeta := range blockchain {
```
rootulp, line 52: [nit] related to other comment about renaming function to ReadBlockMetas, this would make names consistent across the function name and callsite.

```suggestion
	for _, blockMeta := range blockMetas {
```

rootulp (approved):

PR #3680, opened 2024-07-13
test(testnode): configure custom min gas price

rootulp:
Context: https://github.com/celestiaorg/celestia-node/pull/3453#issuecomment-2220710282

Fix testnode so that it is possible to query the local min gas price again. I plan on backporting this to v2.x to fix celestia-node integration tests which query local min gas price.


rootulp:
The sdk context on v1.x is populated with a min gas price but on main it isn't. I think this regression was caused by a refactor to testnode b/c the v1.x testnode sets min gas price [here](https://github.com/rootulp/celestia-app/blob/8ca917507e3c97ef83896823369ea4c45e417c72/test/util/network/network.go#L84) but on main, it uses the default app creator which doesn't set it.

test/util/testnode/config.go
```diff
@@ -166,3 +172,19 @@ func DefaultTendermintConfig() *tmconfig.Config {
...
+		encodingConfig := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+		return app.New(
+			log.NewNopLogger(),
+			tmdb.NewMemDB(),
+			nil,            // trace store
+			0,              // invCheckPerid
+			encodingConfig, //encoding config
+			0,              // upgrade height v2
+			simapp.EmptyAppOptions{},
+			baseapp.SetMinGasPrices(fmt.Sprintf("%v%v", appconsts.DefaultMinGasPrice, app.BondDenom)),
```
rootulp, line 187: this line is the fix

cmwaters (approved):

test/util/testnode/config.go
ninabarbakadze: when did this break? out of curiosity

rootulp: It works on v1.x but it is broken on `main` / v2.x. I think it broken during a refactor to testnode, potentially https://github.com/celestiaorg/celestia-app/pull/1991/files#diff-1408d16ad448ff1be464a1eb5af128736aca3302ba3a5d7fbebd7de86473a478R107

evan-forbes (approved):

PR #3685, opened 2024-07-15
refactor(scripts): improve user prompts for single-node.sh

### LINKED ISSUE #3652: Improve confirmation prompts for single-node.sh
## Context

https://github.com/celestiaorg/celestia-app/blob/main/scripts/single-node.sh

## Problem

There are two confirmation prompts for `y/n`

## Proposal

1. Remove the second prompt and always set up the local tracing stuff
2. Replace the first prompt with something like: "Do you want to start a new local testnet?". If the user responds with `y` then delete `~/.celestia-app` and run `celestia-appd init`, etc. If the user responds with `n` then the script should just run `celestia-appd start ...` so that the node continues running the previous local testnet.

najeal:
Closes #3652 

## Overview

It modifies `single-node.sh` script.
This rewrites the first prompt and deletes the second one.
I wanted write `Do you want to start a new local testnet? [y/n]` only when `$HOME/.celestia-app` does not exist. But it looks like as soon as `celestia-app` is invoked, it creates the directory. Hence I cannot use this information.


rootulp:
> I wanted write Do you want to start a new local testnet? [y/n] only when $HOME/.celestia-app does not exist. 

Good idea!

> But it looks like as soon as celestia-app is invoked, it creates the directory.

The directory should only be created when `celestia-appd init` is invoked`. So I think you can do something like:

```code 
if $HOME/.celestia-app exists {
    // promt user to start a new local testnet (delete existing $HOME/.celestia-app) or use existing local testnet
} else {
    // skip prompt and create a new local testnet
}
```

rootulp:
Pushed a commit with the changes I had in mind. Tested:

```bash
$ ./scripts/single-node.sh
celestia-app home: /Users/rootulp/.celestia-app
celestia-app version: 2.0.0-rc2

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661
Starting celestia-app...
```

```bash
$ ./scripts/single-node.sh
celestia-app home: /Users/rootulp/.celestia-app
celestia-app version: 2.0.0-rc2

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
n
Starting celestia-app...
```

rootulp (approved):

najeal:
Makes sense ! I did not think about the genesis file ! 👍

rach-id (approved):

PR #3697, opened 2024-07-17
test(testnode): query for events

rootulp:
Motivated by a failing integration test in celestia-node when bumping to celestia-app v2.x


rootulp:
hmm I'm stumped. The first commit on this PR failed locally but passed in CI so I'm quite confused.

PR #3706, opened 2024-07-18
fix(test/e2e): name cannot be empty error

rootulp:
Recent e2e tests failed with:

```
2024/07/18 17:22:08 Failed to setup testnet: converting accounts into sdk types: invalid account 4: name cannot be empty
```

because https://github.com/celestiaorg/celestia-app/pull/3690 merged. Since e2e tests aren't run as part of CI on each PR, I didn't learn about the failure until @ninabarbakadze pinged about it.

## Testing

```
make test-e2e
```

gets past that error locally.

### DIFF
--- a/test/e2e/testnet/testnet.go
+++ b/test/e2e/testnet/testnet.go
@@ -218,6 +218,7 @@ func (t *Testnet) CreateAccount(name string, tokens int64, txsimKeyringDir strin
 	err = t.genesis.AddAccount(genesis.Account{
 		PubKey:  pk,
 		Balance: tokens,
+		Name:    name,
 	})
 	if err != nil {
 		return nil, err
--- a/test/util/genesis/genesis.go
+++ b/test/util/genesis/genesis.go
@@ -122,6 +122,9 @@ func (g *Genesis) WithKeyringAccounts(accs ...KeyringAccount) *Genesis {
 
 // AddAccount adds an existing account to the genesis.
 func (g *Genesis) AddAccount(account Account) error {
+	if err := account.ValidateBasic(); err != nil {
+		return err
+	}
 	for _, acc := range g.accounts {
 		if bytes.Equal(acc.PubKey.Bytes(), account.PubKey.Bytes()) {
 			return fmt.Errorf("account with pubkey %s already exists", account.PubKey.String())


ninabarbakadze (approved):
thanks rootul!!

test/util/genesis/genesis.go
```diff
@@ -122,6 +122,9 @@ func (g *Genesis) WithKeyringAccounts(accs ...KeyringAccount) *Genesis {
 
 // AddAccount adds an existing account to the genesis.
 func (g *Genesis) AddAccount(account Account) error {
+	if err := account.ValidateBasic(); err != nil {
```
evan-forbes, line 125: are there any scenarios where this could consensus breaking? if so, are those scenarios relevant and do we need to document this somewhere?

evan-forbes (approved):
approving as this makes sense even if it is consensus breaking

the only time it ever could get hit is with a new chain, we just _**might**_ not be able to backport this to v2

evan-forbes:
tentatively removed the backport command out of an abundance of caution, but we can add it back if we don't need to be worried about this

test/util/genesis/genesis.go
cmwaters: Why would this be consensus breaking? This is just about creating the genesis - I don't quite see how it affects actual running of a chain

cmwaters (approved):
👍

test/e2e/testnet/testnet.go
```diff
@@ -218,6 +218,7 @@ func (t *Testnet) CreateAccount(name string, tokens int64, txsimKeyringDir strin
 	err = t.genesis.AddAccount(genesis.Account{
 		PubKey:  pk,
 		Balance: tokens,
+		Name:    name,
 	})
```
rootulp, line 222: [note to self] this may be why it is preferable to use a struct constructor like `NewGenesisAccount` instead of creating a struct via `genesis.Account{}` in many files. That way the PR that added a new field to the struct can add it to the `NewGenesisAccount` constructor and doesn't have to CMD + F for all instances of the struct in the entire codebase to make sure one isn't missed.

Would be nice if Go compiler complained that the field wasn't initialized.

rootulp:
I think this can be backported to v2.x b/c AFAICT it only impacts `test/` files so shouldn't be consensus breaking.

PR #3713, opened 2024-07-19
feat!: compact blocks prototype

cmwaters:


evan-forbes:
:saluting_face: closing due to neccessary changes in the design, but we're re-using portions of the implementation from this in https://github.com/celestiaorg/celestia-core/issues/1532

PR #3718, opened 2024-07-22
test: fixes e2e test failures due to increased timeout propose and timeout commit

staheri14:
The e2e tests started failing due to a recent change to the node configurations, which unintentionally increased the timeout propose and timeout commit of all validators in all e2e tests to `10` and `11` seconds ([link](https://github.com/celestiaorg/celestia-app/blob/9fa9fbf2fc347752dfe67aa3eca6dc157d3f3291/pkg/appconsts/consensus_consts.go#L6-L7)). This caused the `E2ESimple` test to fail due to its short test duration `30` seconds and insufficient network progress due to the high block time (due to these, there were not enough txs at the end of the test). The `MajorUpgradeToV2` also seemed to fail for similar reasons (related to timeouts).

The current PR addresses the problem by retracting the initial timeout propose and timeout commit i.e., `3` and `1` seconds, respectively. This will resolve the issue and make all the e2e tests pass.


staheri14:
Will open it for reviews once this [PR](https://github.com/celestiaorg/celestia-app/pull/3672) gets merged.

staheri14:
I can confirm the all the e2e tests (i.e., `MajorUpgradeToV2`, `MinorVersionCompatibility`, and `E2ESimple`) pass in this branch. Among the benchmark tests, I ran `TwoNodeBigBlock64MB` for sample, it also worked fine.

evan-forbes (approved):

rootulp:
> The e2e tests started failing due to a recent change to the node configurations, which unintentionally increased the timeout propose and timeout commit of all validators in all e2e tests to 10 and 11 seconds ([link](https://github.com/celestiaorg/celestia-app/blob/9fa9fbf2fc347752dfe67aa3eca6dc157d3f3291/pkg/appconsts/consensus_consts.go#L6-L7)).

I'm confused because `TimeoutPropose` and `TimeoutCommit` hasn't been modified for a year. See [git blame](https://github.com/celestiaorg/celestia-app/blame/main/pkg/appconsts/consensus_consts.go). That's the case on `main`, `v1.x`, and `v2.x`. What am I missing?

test/e2e/testnet/setup.go
```diff
@@ -15,8 +15,8 @@ import (
 func MakeConfig(node *Node, opts ...Option) (*config.Config, error) {
 	cfg := app.DefaultConsensusConfig()
 	cfg.TxIndex.Indexer = "kv"
-	cfg.Mempool.MaxTxsBytes = 1 * GiB
-	cfg.Mempool.MaxTxBytes = 8 * MiB
+	cfg.Consensus.TimeoutPropose = config.DefaultConsensusConfig().TimeoutPropose
+	cfg.Consensus.TimeoutCommit = config.DefaultConsensusConfig().TimeoutCommit
```
rootulp, line 19: [no change needed][for other reviewers] these lines override

```
TimeoutPropose: 10s => 3s 
TimeoutCommit 11s => 1s
```

it's kinda confusing that these lines aren't the defaults we use:
- https://github.com/celestiaorg/celestia-core/blob/6333e56ff954fae056f628ad0fe459409a0b3fcd/config/config.go#L989
- https://github.com/celestiaorg/celestia-core/blob/6333e56ff954fae056f628ad0fe459409a0b3fcd/config/config.go#L995

rootulp (approved):

test/e2e/testnet/setup.go
staheri14: In the e2e tests, we initially used the default consensus configurations with timeout commit and timeout propose set to 1 and 3 seconds, respectively. However, with the introduction of the benchmark test (this [PR](https://github.com/celestiaorg/celestia-app/pull/3612/files#diff-3d744f73561b631eaa5fc3b543a34858ab95ba21b53c768fdcffcb6b58f954bdR16)), we aligned the configurations with those used by celestia-app, which set higher values for timeout commit and propose. These higher values caused the normal e2e tests to time out. In this PR, I have reset the default timeouts to 1 and 3 seconds, and increased them only for the benchmark tests.

staheri14:
> I'm confused because `TimeoutPropose` and `TimeoutCommit` hasn't been modified for a year (see [git blame](https://github.com/celestiaorg/celestia-app/blame/main/pkg/appconsts/consensus_consts.go)). That's the case on `main`, `v1.x`, and `v2.x`. What am I missing?

I see where the confusion comes from, the issue resolved in this PR is not about the values of timeout_commit and timeout_propose in the app default configs (or the cometBFT default configs), rather which configs we use in the test.

test/e2e/testnet/setup.go
staheri14: PS: decreasing timeout commit and propose also make the e2e tests faster.

rootulp: ahh got it, thanks for clarifying!

PR #3720, opened 2024-07-23
fix: lint suggestions

ninabarbakadze:
## Overview

There were non-blocking lint suggestions shown on every run. This pr addresses those suggestions so we don't have to keep seeing the red lines.

### DIFF
--- a/test/e2e/testnet/node.go
+++ b/test/e2e/testnet/node.go
@@ -303,17 +303,12 @@ func (n *Node) Start() error {
 	if err := n.StartAsync(); err != nil {
 		return err
 	}
-	if err := n.WaitUntilStartedAndForwardPorts(); err != nil {
-		return err
-	}
-	return nil
+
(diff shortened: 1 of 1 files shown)


cmwaters (approved):

ninabarbakadze:
> showing

run `make lint` on main

rach-id (approved):
@ninabarbakadze 
```
~/Celestia/celestia-core on main *2 > make lint                                                                                                                                                     took 19s
--> Running linter
~/Celestia/celestia-core on main *2 >     
```

It's not showing anything locally. I guess that I have a different version. 

Generally, if we want to enforce specific linting rule, we should add a workflow that runs the linter and complains if there are any changes to the repo. Otherwise, we will need to make this change manually every time, and not expect contributors to do it themselves.

Not blocking on this tho, thanks for handling it :pray:

PR #3735, opened 2024-07-25
refactor!: version gas scheduler variables

### LINKED ISSUE #3404: Version gas schedule variables
All on-chain governance modifiable parameters associated with the gas scheduler (i.e. GasCostPayForBlobs) should become versioned parameters i.e. they can only change with a hard fork upgrade.

The reason is to help clients who are submitting transactions make it easier to estimate the gas (as it won't fluctuate from block to block and you won't need additional queries for these values - they can be hardcoded into your estimator)

Context: https://github.com/celestiaorg/celestia-app/issues/3011#issuecomment-2071835562_
            
Acceptance Criteria

- [x] Write a CIP 
- [ ] Implement the change

### LINKED ISSUE #3748: Come back to versioned gas scheduler variables
## Description

Come back to #3404 https://github.com/celestiaorg/celestia-app/pull/3735 after versioned app module upgrades and see if we want to keep the current design. 

## Acceptance Criteria

- [ ] reevaluate the current status of refactor
- [ ] write docs for the final version

### LINKED ISSUE #3762: Add a params v3 page to the specs
Overall LGTM. We'll also need to add a params v3 page to the specs and mark the params: `blob.GasPerBlobByte` and `auth.TxSizeCostPerByte` as non governance modifiable on that page. 

Ref: https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/parameters_v2.md

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/3735#pullrequestreview-2221789398_

ninabarbakadze:
## Overview

- Fixes #3404 
- Fixes #3748 
- Fixes #3762


app/ante/tx_size.go
```diff
@@ -0,0 +1,146 @@
...
+
+// consumeGasForTxSize consumes gas based on the size of the transaction.
+// It uses different parameters depending on the app version.
+func consumeGasForTxSize(ctx sdk.Context, cost uint64, params auth.Params) {
+	// For app v2 and below we should get txSizeCostPerByte from auth module
+	if ctx.BlockHeader().Version.App <= v2.Version {
+		ctx.GasMeter().ConsumeGas(params.TxSizeCostPerByte*cost, "txSize")
+	} else {
+		// From v3 onwards, we should get txSizeCostPerByte from appconsts
+		ctx.GasMeter().ConsumeGas(appconsts.TxSizeCostPerByte(v3.Version)*cost, "txSize")
```
cmwaters, line 144: 👍

app/ante/tx_size_test.go
```diff
@@ -0,0 +1,189 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/tx/signing"
+	authante "github.com/cosmos/cosmos-sdk/x/auth/ante"
+	xauthsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/stretchr/testify/require"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+func setup() (*app.App, sdk.Context, client.Context, sdk.AnteHandler, error) {
```
cmwaters, line 28: It looks like `sdk.AnteHandler` isn't used by the test. Do you mind removing it

app/ante/tx_size_test.go
```diff
@@ -0,0 +1,189 @@
...
+	app, _, _ := testutil.NewTestAppWithGenesisSet(app.DefaultConsensusParams())
```
cmwaters, line 29: Q: If we're using `DefaultConsensusParams` that will mean we're using app version 3 right? Has this been updated yet?

x/blob/keeper/gas_test.go
```diff
@@ -23,33 +24,33 @@ func TestPayForBlobGas(t *testing.T) {
 		{
 			name:            "1 byte blob", // occupies 1 share
 			msg:             types.MsgPayForBlobs{BlobSizes: []uint32{1}},
-			wantGasConsumed: uint64(1*appconsts.ShareSize*types.DefaultGasPerBlobByte + paramLookUpCost), // 1 share * 512 bytes per share * 8 gas per byte + 1060 gas for fetching param = 5156 gas
+			wantGasConsumed: uint64(1*appconsts.ShareSize*appconsts.GasPerBlobByte(v3.Version) + paramLookUpCost), // 1 share * 512 bytes per share * 8 gas per byte + 1060 gas for fetching param = 5156 gas
```
cmwaters, line 27: These tests should be updates so that they always use the latest version instead of being hardcoded to v3

x/blob/keeper/keeper.go
```diff
@@ -43,7 +45,14 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) PayForBlobs(goCtx context.Context, msg *types.MsgPayForBlobs) (*types.MsgPayForBlobsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	gasToConsume := types.GasToConsume(msg.BlobSizes, k.GasPerBlobByte(ctx))
+	// GasPerBlobByte is a versioned variable from version 3 onwards.
+	var gasToConsume uint64
+	if ctx.BlockHeader().Version.App <= v2.Version {
+		gasToConsume = types.GasToConsume(msg.BlobSizes, k.GasPerBlobByte(ctx))
+	} else {
+		gasToConsume = types.GasToConsume(msg.BlobSizes, appconsts.GasPerBlobByte(appconsts.LatestVersion))
```
cmwaters, line 53: This should not be hardcoded but be `ctx.BlockHeader().Version.App`

cmwaters (changes requested):

pkg/appconsts/initial_consts.go
```diff
@@ -13,10 +13,6 @@ const (
 	// maximum number of bytes allowed in a valid block.
 	DefaultMaxBytes = DefaultGovMaxSquareSize * DefaultGovMaxSquareSize * ContinuationSparseShareContentSize
 
-	// DefaultGasPerBlobByte is the default gas cost deducted per byte of blob
-	// included in a PayForBlobs txn
-	DefaultGasPerBlobByte = 8
-
```
rootulp, line 19: nit: this is breaking so we should prob add ! to the PR title.

[no change needed] If we want to prep for the v3 release, we could also write a changelog entry on how consumers of this package need to modify their code if they used `DefaultGasPerBlobByte`. We could write these entires in a new markdown file in a new directory like `changelog/v3.md`

pkg/appconsts/v3/app_consts.go
```diff
@@ -0,0 +1,7 @@
+package v3
+
+const (
+	Version           uint64 = 3
+	TxSizeCostPerByte uint64 = 10
+	GasPerBlobByte    uint32 = 8
+)
```
rootulp, line 7: should this also include all the other constants from v2?

```go
	SquareSizeUpperBound int    = 128
	// NetworkMinGasPrice is used by x/minfee to prevent transactions from being
	// included in a block if they specify a gas price lower than this.
	NetworkMinGasPrice float64 = 0.000001 // utia
```

pkg/appconsts/versioned_consts.go
```diff
@@ -26,7 +27,17 @@ func SquareSizeUpperBound(_ uint64) int {
...
+
+func GasPerBlobByte(_ uint64) uint32 {
+	return v3.GasPerBlobByte
+}
+
 var (
 	DefaultSubtreeRootThreshold = SubtreeRootThreshold(LatestVersion)
 	DefaultSquareSizeUpperBound = SquareSizeUpperBound(LatestVersion)
+	DefaultTxSizeCostPerByte    = TxSizeCostPerByte(LatestVersion)
+	DefaultGasPerBlobByte       = GasPerBlobByte(LatestVersion)
```
rootulp, line 42: Blocking: these constants aren't defined for v2 and `LatestVersion = v2.Version` so these are broken.

app/ante/tx_size_test.go
ninabarbakadze: No, it hasn't been updated yet but I left it as DefaultConsensusParams because we're planning to do so.

pkg/appconsts/v3/app_consts.go
ninabarbakadze: I'm not really sure. wdyt? @cmwaters

pkg/appconsts/versioned_consts.go
ninabarbakadze: I only did that because we're about to upgrade the app to v3 and I believe no matter what you pass into the function as a version it will return v3.GasPerBlobByte or v3.TxSizeCostPerByte

x/blob/ante/ante.go
```diff
@@ -33,8 +36,12 @@ func (d MinGasPFBDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool
 		// NOTE: here we assume only one PFB per transaction
 		if pfb, ok := m.(*types.MsgPayForBlobs); ok {
 			if gasPerByte == 0 {
-				// lazily fetch the gas per byte param
-				gasPerByte = d.k.GasPerBlobByte(ctx)
+				if ctx.BlockHeader().Version.App <= v2.Version {
+					// lazily fetch the gas per byte param
+					gasPerByte = d.k.GasPerBlobByte(ctx)
+				} else {
+					gasPerByte = appconsts.GasPerBlobByte(v3.Version)
```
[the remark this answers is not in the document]
ninabarbakadze: I think it'll still work in v4 because no matter what version you pass into the function it will return v3.GasPerBlobByte  which Is slightly unintuitive. I propose we remove the option to pass arguments unless we actually use them.

pkg/appconsts/v3/app_consts.go
cmwaters: Yes, that makes sense. 

I'm unsure why `NetworkMinGasPrice` needs to be a versioned constant. This seems just an initial constant that we set min fee to at genesis (or whenever the module begins to be used)

x/blob/ante/ante.go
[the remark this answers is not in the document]
cmwaters: Yes but that's incorrect behaviour. It should be

```suggestion
					gasPerByte = appconsts.GasPerBlobByte(ctx.BlockHeader().Version.App)
```

pkg/appconsts/v3/app_consts.go
ninabarbakadze: yeah it's used as a default

app/ante/tx_size.go
cmwaters, line 144:
```suggestion
		ctx.GasMeter().ConsumeGas(appconsts.TxSizeCostPerByte(ctx.BlockHeader().Version.App)*cost, "txSize")
```
Similar to the blob/keeper.go, this should just be the app version not hardcoded to v3

app/ante/tx_size.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	}
+
+	return false
+}
+
+// consumeGasForTxSize consumes gas based on the size of the transaction.
+// It uses different parameters depending on the app version.
+func consumeGasForTxSize(ctx sdk.Context, cost uint64, params auth.Params) {
```
cmwaters, line 138:
```suggestion
func consumeGasForTxSize(ctx sdk.Context, txBytes uint64, params auth.Params) {
```
I think `txBytes` would be more of an accurate description of the variable

pkg/appconsts/initial_consts.go
```diff
@@ -26,4 +22,8 @@ const (
 	// to unbond in a proof of stake system. Any validator within this
 	// time can be subject to slashing under conditions of misbehavior.
 	DefaultUnbondingTime = 3 * 7 * 24 * time.Hour
+
+	// NetworkMinGasPrice is used by x/minfee to prevent transactions from being
+	// included in a block if they specify a gas price lower than this.
+	DefaultNetworkMinGasPrice = 0.000001 // utia
```
cmwaters, line 28: 👍

app/ante/tx_size.go
```diff
@@ -0,0 +1,146 @@
...
+	simSecp256k1Pubkey = &secp256k1.PubKey{Key: key}
+	simSecp256k1Sig    [64]byte
+)
+
+func init() {
+	// Decodes a valid hex string into a sepc256k1Pubkey for use in transaction simulation
+	bz, _ := hex.DecodeString("035AD6810A47F073553FF30D2FCC7E0D3B1C0B74B61A1AAA2582344037151E143A")
+	copy(key, bz)
+	simSecp256k1Pubkey.Key = key
+}
```
rootulp, line 35: I see this was copied from https://github.com/rootulp/cosmos-sdk/blob/ed62a67e3c9309e865ff337b0a4e48ae0dae38fc/x/auth/ante/sigverify.go#L32-L37 so no change needed but IMO we could avoid the `init()` and inline this below before it is used

app/ante/tx_size_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	cgtsd := ante.NewConsumeGasForTxSizeDecorator(app.AccountKeeper)
+	antehandler := sdk.ChainAnteDecorators(cgtsd)
+
+	testCases := []struct {
+		name  string
+		sigV2 signing.SignatureV2
+	}{
+		{"SingleSignatureData", signing.SignatureV2{PubKey: priv1.PubKey()}},
+		{"MultiSignatureData", signing.SignatureV2{PubKey: priv1.PubKey(), Data: multisig.NewMultisig(2)}},
```
rootulp, line 65: I think it would be beneficial to have test cases for the different app versions to test this logic:

```go
	// For app v2 and below we should get txSizeCostPerByte from auth module
	if ctx.BlockHeader().Version.App <= v2.Version {
		ctx.GasMeter().ConsumeGas(params.TxSizeCostPerByte*cost, "txSize")
	} else {
		// From v3 onwards, we should get txSizeCostPerByte from appconsts
		ctx.GasMeter().ConsumeGas(appconsts.TxSizeCostPerByte(v3.Version)*cost, "txSize")
```

which seems like the major difference from upstream

app/ante/tx_size_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+			// require that antehandler passes and does not underestimate decorator cost
+			require.Nil(t, err, "ConsumeTxSizeGasDecorator returned error: %v", err)
+			require.True(t, consumedSimGas >= expectedGas, "Simulate mode underestimates gas on AnteDecorator. Simulated cost: %d, expected cost: %d", consumedSimGas, expectedGas)
+		})
+	}
+}
+
+// CreateTestTx creates a test tx given multiple inputs.
+func CreateTestTx(txBuilder client.TxBuilder, clientCtx client.Context, privs []cryptotypes.PrivKey, accNums []uint64, accSeqs []uint64, chainID string) (xauthsigning.Tx, error) {
```
rootulp, line 130: can this be made private?

pkg/appconsts/initial_consts.go
```diff
@@ -26,4 +22,8 @@ const (
 	// to unbond in a proof of stake system. Any validator within this
 	// time can be subject to slashing under conditions of misbehavior.
 	DefaultUnbondingTime = 3 * 7 * 24 * time.Hour
+
+	// NetworkMinGasPrice is used by x/minfee to prevent transactions from being
```
rootulp, line 26:
```suggestion
	// DefaultNetworkMinGasPrice is used by x/minfee to prevent transactions from being
```

pkg/appconsts/initial_consts.go
rootulp, line 28: Should this include a node that it only applies to app version >= 2?

x/blob/README.md
```diff
@@ -46,7 +46,7 @@ message Params {
 
 #### `GasPerBlobByte`
 
-`GasPerBlobByte` is the amount of gas that is consumed per byte of blob data
+In v1 and v2`GasPerBlobByte` is the amount of gas that is consumed per byte of blob data
```
rootulp, line 49: This is also true for v3 so maybe this should clarify that before it was a governance modifiable param and then in v3 onwards it is a versioned param.

x/blob/keeper/keeper.go
```diff
@@ -43,7 +45,14 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) PayForBlobs(goCtx context.Context, msg *types.MsgPayForBlobs) (*types.MsgPayForBlobsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	gasToConsume := types.GasToConsume(msg.BlobSizes, k.GasPerBlobByte(ctx))
+	// GasPerBlobByte is a versioned variable from version 3 onwards.
```
rootulp, line 48:
```suggestion
	// GasPerBlobByte is a versioned param from version 3 onwards.
```

rootulp (commented):
Overall LGTM. We'll also need to add a params v3 page to the specs and mark the params: `blob.GasPerBlobByte` and `auth.TxSizeCostPerByte` as non governance modifiable on that page. 

Ref: https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/parameters_v2.md

app/ante/tx_size_test.go
ninabarbakadze: I'd say after we update the application to v3 so i can just use latest version

app/ante/tx_size.go
```diff
@@ -0,0 +1,145 @@
...
+	sigTx, ok := tx.(authsigning.SigVerifiableTx)
+	if !ok {
+		return ctx, errors.Wrap(sdkerrors.ErrTxDecode, "invalid tx type")
+	}
+	params := cgts.ak.GetParams(ctx)
+
+	consumeGasForTxSize(ctx, sdk.Gas(len(ctx.TxBytes())), params)
+
+	// simulate gas cost for signatures in simulate mode
+	if simulate {
```
evan-forbes, line 65: could we delete this section and some of the docs above since we don't use simulate?

app/ante/tx_size.go
```diff
@@ -0,0 +1,145 @@
...
+// slightly over estimated due to the fact that any given signing account may need
+// to be retrieved from state.
+//
+// CONTRACT: If simulate=true, then signatures must either be completely filled
+// in or empty.
+// CONTRACT: To use this decorator, signatures of transaction must be represented
+// as legacytx.StdSignature otherwise simulate mode will incorrectly estimate gas cost.
+type ConsumeTxSizeGasDecorator struct {
+	ak ante.AccountKeeper
+}
```
evan-forbes, line 47: might want to add some docs that this is a copy paste fork from the sdk, and list the modifications

evan-forbes (commented):
only minor comments from me, deferring to other's for approval after they resolve their comments/blocks

app/ante/tx_size.go
cmwaters: Why do you think we don't use simulate?

app/ante/tx_size.go
ninabarbakadze: done!

app/ante/tx_size_test.go
rootulp: 1. Why not update the latest version constant to v3 in this PR? 
2. Why should the latest version constant block unit tests in this PR? Without unit tests, it's not easy to determine if this feature works as expected.

pkg/appconsts/v2/app_consts.go
```diff
@@ -4,7 +4,4 @@ const (
 	Version              uint64 = 2
 	SquareSizeUpperBound int    = 128
 	SubtreeRootThreshold int    = 64
-	// NetworkMinGasPrice is used by x/minfee to prevent transactions from being
-	// included in a block if they specify a gas price lower than this.
-	NetworkMinGasPrice float64 = 0.000001 // utia
```
rootulp, line 9: [nit] another breaking change we could add to changelog / v3 release notes.

x/blob/keeper/keeper_test.go
```diff
@@ -27,34 +29,44 @@ import (
...
-
-	// emit an event by submitting a PayForBlob
-	msg := createMsgPayForBlob(t, signer, namespace, blobData)
-	_, err := k.PayForBlobs(ctx, msg)
-	require.NoError(t, err)
+	// Testing transition from v2 -> v3
+	// v2 uses the GasPerBlobByte key from the params store
+	// v3 uses the GasPerBlobByte from the appconsts
+	// TODO: Replace this with appversion latest
+	versions := []uint64{v2.Version, v3.Version}
```
rootulp, line 36: [question] why only test appversion latest instead of all 3 app versions?

```suggestion
	// TODO: Replace this with appversion latest
	versions := []uint64{v1.Version, v2.Version, v3.Version}
```

rootulp (approved):
Not going to block on it but unit tests that the ante handler behaves differently depending on the app version would be really nice.

app/ante/tx_size.go
evan-forbes: ahh I see I just dug through and found that by simulate, we mean calculate the gas value. up until now I assumed this was for the x/simulate package, which I don't think we use but perhaps I have misunderstanding there as well!

x/blob/keeper/keeper_test.go
ninabarbakadze: we don't really need v1.Version here because we care that the transitionary versions work. For example for versions v2 and v1 it's the same behaviour and they'd fall under the same if statement same goes for versions v3+. I don't think it'd be sustainable to have all versions hardcoded in tests.

app/ante/tx_size_test.go
ninabarbakadze: 1. I didn't want to update LatestVersion constant because we haven't properly upgraded the app to v3 and iirc @cmwaters was planning on doing it.
3. It's not really blocking, to my understanding we hadn't yet bumped the app to the latest version hence why I had a follow up ticket for looking back on the implementation + I manually verified that both cases worked. I updated the test to test v2.Version and v3.Version so we don't have to have reservations around testing. I'll go back to this and change all v3.Version references to LatestVersion

pkg/appconsts/v2/app_consts.go
ninabarbakadze: where are the release notes?

app/ante/tx_size_test.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/commit/2ddaa9f97df64e703e04ce12378b607963ef6265

x/blob/keeper/keeper_test.go
rootulp: On re-read, I find this test confusing b/c it doesn't actually make any assertions for version 2 and version 3. I think this test would be useful if it looked something like:

```go
	type testCase struct {
		version uint64
		want    int64
	}
	testCases := []testCase{
		{version: v1.Version, want: 10},
		{version: v2.Version, want: 8},
	}
```

but since it doesn't do that, we could just make it non-version aware:

```go
// TestPayForBlobs verifies the attributes on the emitted event.
func TestPayForBlobs(t *testing.T) {
	k, _, ctx := CreateKeeper(t, v3.Version)
	signer := "celestia15drmhzw5kwgenvemy30rqqqgq52axf5wwrruf7"
	namespace := share.MustNewV0Namespace(bytes.Repeat([]byte{1}, share.NamespaceVersionZeroIDSize))
	namespaces := [][]byte{namespace.Bytes()}
	blobData := []byte("blob")
	blobSizes := []uint32{uint32(len(blobData))}

	// verify no events exist yet
	events := ctx.EventManager().Events().ToABCIEvents()
	assert.Len(t, events, 0)

	// emit an event by submitting a PayForBlob
	msg := createMsgPayForBlob(t, signer, namespace, blobData)
	_, err := k.PayForBlobs(ctx, msg)
	require.NoError(t, err)

	// verify that an event was emitted
	events = ctx.EventManager().Events().ToABCIEvents()
	assert.Len(t, events, 1)
	protoEvent, err := sdk.ParseTypedEvent(events[0])
	require.NoError(t, err)
	event, err := convertToEventPayForBlobs(protoEvent)
	require.NoError(t, err)

	// verify the attributes of the event
	assert.Equal(t, signer, event.Signer)
	assert.Equal(t, namespaces, event.Namespaces)
	assert.Equal(t, blobSizes, event.BlobSizes)
}
```

which I realize now is probably what you meant by the TODO.

pkg/appconsts/v2/app_consts.go
rootulp: we don't have a place to store it yet so no change needed. See [Slack thread](https://celestia-team.slack.com/archives/C02PFFCGMNW/p1723141674451709) for more context

x/blob/keeper/keeper_test.go
ninabarbakadze: yeah exactly. I tried to avoid hardcoding the version. In hindsight I would've only started working on this after the app was properly upgraded to v3. Do you think it'd be better if I update the application to v3 and then I can also address the issues in tests.

app/ante/tx_size_test.go
ninabarbakadze: Added both versions now @rootulp

ninabarbakadze (commented):
moving this to draft until app upgrades to v3 and/or multi app work is finished

app/ante/tx_size.go
rootulp: Just to close out this thread: `simulate` is used. For example when a user submits a tx with the flag:

```
      --dry-run                   ignore the --gas flag and perform a simulation of a transaction, but don't broadcast it (when enabled, the local Keybase is not accessible)
```

related

```
      --gas-adjustment float      adjustment factor to be multiplied against the estimate returned by the tx simulation; if the gas limit is set manually this flag is ignored  (default 1)
```

app/ante/tx_size_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	testCases := []struct {
+		version uint64
+		name    string
+		sigV2   signing.SignatureV2
+	}{
+		{v2.Version, "SingleSignatureData v2", signing.SignatureV2{PubKey: priv1.PubKey()}},
+		{v2.Version, "MultiSignatureData v2", signing.SignatureV2{PubKey: priv1.PubKey(), Data: multisig.NewMultisig(2)}},
+		{appconsts.LatestVersion, fmt.Sprintf("SingleSignatureData v%d", appconsts.LatestVersion), signing.SignatureV2{PubKey: priv1.PubKey()}},
+		{appconsts.LatestVersion, fmt.Sprintf("MultiSignatureData v%d", appconsts.LatestVersion), signing.SignatureV2{PubKey: priv1.PubKey(), Data: multisig.NewMultisig(2)}},
```
rootulp, line 70: These test cases don't contain an expected amount of gas to be consumed. IMO this test would be more likely to catch bugs if it verified that the amount of gas consumed by v2 differed from v3. Since the params for v2 and v3 are the same, one way to do that would be to change the value of the param for v2 and/or v3 and verify that the gas consumed was based on that.

app/test/upgrade_test.go
```diff
@@ -145,8 +145,8 @@ func TestAppUpgradeV3(t *testing.T) {
 
 // TestAppUpgradeV2 verifies that the all module's params are overridden during an
 // upgrade from v1 -> v2 and the app version changes correctly.
-func TestAppUpgradeV2(t *testing.T) {
-	NetworkMinGasPriceDec, err := sdk.NewDecFromStr(fmt.Sprintf("%f", v2.NetworkMinGasPrice))
+func TestAppUpgrades(t *testing.T) {
```
rootulp, line 148:
```suggestion
func TestAppUpgradeV2(t *testing.T) {
```

pkg/appconsts/initial_consts.go
rootulp: https://github.com/celestiaorg/celestia-app/blob/main/docs/release-notes/pending-release.md now exists so we should add a line that states this constant was removed and what library consumers should use instead.

pkg/appconsts/v2/app_consts.go
rootulp: We now have a place for this: https://github.com/celestiaorg/celestia-app/blob/main/docs/release-notes/pending-release.md

pkg/appconsts/versioned_consts.go
rootulp: This convo can be resolved bc LatestVersion = 3.

x/blob/keeper/keeper_test.go
```diff
@@ -101,6 +101,7 @@ func CreateKeeper(t *testing.T) (*keeper.Keeper, store.CommitMultiStore, sdk.Con
 		cdc,
 		paramsSubspace,
 	)
+	// change the params to a non default value
```
rootulp, line 104: this comment doesn't make sense to me b/c the next line sets the params to the default values.

```suggestion
```

x/blob/types/payforblob.go
```diff
@@ -165,7 +164,7 @@ func EstimateGas(blobSizes []uint32, gasPerByte uint32, txSizeCost uint64) uint6
 // DefaultEstimateGas runs EstimateGas with the system defaults. The network may change these values
 // through governance, thus this function should predominantly be used in testing.
```
rootulp, line 165: The network can't change the params via governance any more so this comment looks incorrect.

```suggestion
// DefaultEstimateGas runs EstimateGas with the system defaults. 
```

rootulp (approved):

ninabarbakadze:
DefaultGasPerBlobByte is still accessible through `appconsts.DefaultGasPerBlobByte` so I don't think library consumers would need to change anything

app/ante/tx_size_test.go
ninabarbakadze: done

rootulp:
> DefaultGasPerBlobByte is still accessible through appconsts.DefaultGasPerBlobByte so I don't think library consumers would need to change anything

Oh good point, nvm disregard my comment about adding it to release notes. Does that mean the only breaking change in this PR is the removal of `NetworkMinGasPrice`? Should we add that to pending-release notes?

rootulp (approved):
Since this PR is marked breaking, I think we should at the very least add notes to pending-release.md about what was API breaking.

If we wanted to go above and beyond, we could add a line describing what this refactor does to pending-release.md so that we can use it in the release notes.

ninabarbakadze:
> Since this PR is marked breaking, I think we should at the very least add notes to pending-release.md about what was API breaking.
> 
> If we wanted to go above and beyond, we could add a line describing what this refactor does to pending-release.md so that we can use it in the release notes.

I want to merge this but will do in a follow up

x/blob/keeper/keeper.go
```diff
@@ -43,7 +45,14 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	gasToConsume := types.GasToConsume(msg.BlobSizes, k.GasPerBlobByte(ctx))
+	// GasPerBlobByte is a versioned param from version 3 onwards.
+	var gasToConsume uint64
+	if ctx.BlockHeader().Version.App <= v2.Version {
+		gasToConsume = types.GasToConsume(msg.BlobSizes, k.GasPerBlobByte(ctx))
+	} else {
+		gasToConsume = types.GasToConsume(msg.BlobSizes, appconsts.GasPerBlobByte(ctx.BlockHeader().Version.App))
+	}
```
cmwaters, line 54: 👍

cmwaters (approved):
Great work!

PR #3738, opened 2024-07-25
test: adds network test with latency

staheri14:


test/e2e/testnet/defaults.go
```diff
@@ -4,7 +4,7 @@ var DefaultResources = Resources{
 	MemoryRequest: "400Mi",
 	MemoryLimit:   "400Mi",
 	CPU:           "300m",
-	Volume:        "1Gi",
+	Volume:        "0Gi",
```
smuu, line 7: The minimum size for a volume is `1Gi`.
If you don't want a volume, you should not call `AddVolume` for that Instance.

smuu (changes requested):

PR #3747, opened 2024-07-31
test: makes network latency effective at earlier heights in benchmark tests

### LINKED ISSUE #3746: Applying network latency in the earlier heights in benchmark tests
# Problem

In the benchmark tests, latency is added to the nodes, but it only takes effect very late in the experiment, which is not desirable.

# Acceptance Criteria
- [ ] Identify and resolve the issue causing the delayed latency effect.
- [ ] Implement a solution where latency is applied immediately after the nodes are started, without waiting for them to sync, as suggested by DevOps. That is,  the code needs to be modified to apply latency and jitter immediately after the instance starts (without waiting for the first block), i.e.,  start all nodes and wait for txsim to initiate: 
https://github.com/celestiaorg/celestia-app/blob/68d1e42a409f14b56b296df24ec2c0fa0c72a829/test/e2e/benchmark/benchmark.go#L104
then, apply latency and jitter: 
https://github.com/celestiaorg/celestia-app/blob/68d1e42a409f14b56b296df24ec2c0fa0c72a829/test/e2e/benchmark/benchmark.go#L112

staheri14:
Closes #3746

### DIFF
--- a/test/e2e/benchmark/benchmark.go
+++ b/test/e2e/benchmark/benchmark.go
@@ -130,7 +130,6 @@ func (b *BenchmarkTest) Run() error {
 	err = b.StartTxClients()
 	if err != nil {
 		return fmt.Errorf("failed to start tx clients: %v", err)
-
 	}
 
 	// wait some time for the tx clients to submit transactions
--- a/test/e2e/benchmark/benchmark.go
+++ b/test/e2e/benchmark/benchmark.go
@@ -100,8 +100,10 @@ func (b *BenchmarkTest) SetupNodes() error {
 
 // Run runs the benchmark test for the specified duration in the manifest.
 func (b *BenchmarkTest) Run() error {
-	log.Println("Starting testnet")
-	err := b.Start()
+	log.Println("Starting benchmark testnet")
+
+	log.Println("Starting nodes")
+	err := b.StartNodes()
 	if err != nil {
 		return fmt.Errorf("failed to start testnet: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


staheri14:
I verified that the latency gets added at earlier stages of the experiment by inspecting the block times and the average delay between sending and receiving individual block parts.

**Experiment without latency:**

block_time_vs_size_1v2-t2-b60-bw5mb-tc11-tp80-l0-v1-8mb_8

**Experiment with latency:**

block_time_vs_size_1v2-t2-b60-bw5mb-tc11-tp80-l1-v1-8mb_8

As we can see, the block time increased in the experiment with latency compared to the one without latency, starting from the beginning of the experiment. The average block time also confirms this observation.

Furthermore, I inspected the average block part delay for height 10 in both experiments, and the results indicate that the latency was applied as expected:
- **Average block part delay without latency:** 577.99 milliseconds
- **Average block part delay with latency:** 1258.19 milliseconds

evan-forbes (approved):

cmwaters (approved):
So if I understand this correctly, previously because we were waiting until all genesis nodes had produced their first block, this meant latency would only be added later on. Now we would do it as soon as the machine says it's up and running.

nit: I would prefer we integrate latency into the start function so test writers don't have to worry about that but that can be addressed later

staheri14:
>So if I understand this correctly, previously because we were waiting until all genesis nodes had produced their first block, this meant latency would only be added later on. Now we would do it as soon as the machine says it's up and running. 

Your understanding is correct. 

> nit: I would prefer we integrate latency into the start function so test writers don't have to worry about that but that can be addressed later

This is handled in the [BenchmarkTest.Run()](https://github.com/celestiaorg/celestia-app/blob/c4196c517c963a892e43720f3c516e8c86786396/test/e2e/benchmark/benchmark.go#L102), so test developers don't need to take any additional actions. As soon as latency is specified in the manifest, it will take effect in the test. However, this is not part of the testnet functionalities (note that [BenchmarkTest](https://github.com/celestiaorg/celestia-app/blob/c4196c517c963a892e43720f3c516e8c86786396/test/e2e/benchmark/benchmark.go#L16) wraps around `testnet.Testnet`), as (I thought) adding latency is not necessarily required for other non-benchmark e2e tests. If you think we can benefit from net latency for those tests, let me know and I'll integrate it into Testnet.Start instead.

Issue #3753, opened 2024-08-02
gRPC TxStatus endpoint


cmwaters:
We now have a tx status endpoit in celestia-core which returns the real time status of a transaction. The TxClient, for whom this new endpoint is chiefly designed around, only relied on grpc. We need to bridge the gap by creating a new gRPC service that proxies to the underlying RPC server

Issue #3421, opened 2024-05-02
Expose all celestia-node required endpoints through gRPC


liamsi:
### Summary

Node has to use core's/comet's RPC additionally to gRPC. That is very confusing. 
The current app API is not sufficient and lead to bad design decisions in celestia-node which led to proposals that will manifest these even further https://github.com/celestiaorg/celestia-node/issues/2931

### Problem Definition

Node currently uses both RPC (comet) and gRPC (app). Apparently, ~~the only~~ one reason for that is only comet's RPC exposes proofs for state queries currently, or rather ABCI queries (see #3422). 
Another reason, is that bridge nodes currently get their block data via RPC. That is also not ideal as the block data should be binary/protobuf encoded instead of a large JSON file. 

If these two changed, celestia-node could solely rely on app's gRPC. 

One of the places ~~The only place~~ where node uses rpc is here: 
- https://github.com/celestiaorg/celestia-node/blob/2d2adb3a2504f42bb56e08ef2b8bacba52c54c86/state/core_access.go#L348
- reason being that node wants to verify proofs here: https://github.com/celestiaorg/celestia-node/blob/2d2adb3a2504f42bb56e08ef2b8bacba52c54c86/state/core_access.go#L369-L377

### Proposal

- [x] Expose ABCI queries from comet RPC through gRPC proxy or returns proofs in other ways. #3423
- [ ] Expose a way to subscribe to block data via gRPC:
   - essentially subscribe to the same event as node does (signedBlock) but with local client and then stream result to new endpoint

liamsi:
false alarm: https://github.com/celestiaorg/cosmos-sdk/blob/02244ae2a73e053234ca2de6e68a97974edb952a/client/grpc/tmservice/service.go#L221

liamsi:
Actually, not quite: https://github.com/celestiaorg/celestia-app/pull/3423

Ferret-san:
is it possible to expose these two methods from core, and eventually move the logic to node if possible?

https://github.com/celestiaorg/celestia-core/blob/793ece9bbd732aec3e09018e37dc31f4bfe122d9/rpc/core/routes.go#L33-L34

liamsi:
> is it possible to expose these two methods from core, and eventually move the logic to node if possible?

Yes, should be very, very easy!

> and eventually move the logic to node if possible?

Also. Do you mind opening an issue in node for that?

walldiss:
[...]

PR #3754, opened 2024-08-02
feat: gRPC server for TxStatus endpoint

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/3753

This in some ways ties into https://github.com/celestiaorg/celestia-app/issues/3421


rootulp (approved):
Should we backport this to v1.x and/or v2.x?

app/grpc/tx/server.go
```diff
@@ -0,0 +1,81 @@
...
+
+	txID, err := hex.DecodeString(req.TxId)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid tx id: %s", err)
+	}
+
+	resTx, err := node.TxStatus(ctx, txID)
+	if err != nil {
+		return nil, err
+	}
```
evan-forbes, line 73: after we move node initialization to the app and have access to the dbs, it would soooo nice if we would just move the logic here and not wrap the existing rpc

evan-forbes (approved):

cmwaters:
> Should we backport this to v1.x and/or v2.x?

Yes I think we should

app/grpc/tx/server.go
cmwaters: Yeah, although I don't think proxying through the local RPC has much overhead

PR #3761, opened 2024-08-06
feat: blockheight tool

rootulp:
I used this tool to determine that the v2 upgrade height that we should configure on Arabica.


evan-forbes (approved):

evan-forbes (approved):

cmwaters (approved):
Could have the target time as a cli argument and have the user input the rpc endpoint so we can do it for all chains, but happy to merge this as a basic tool

PR #3765, opened 2024-08-07
feat!: support authored blobs

cmwaters:
This PR uses the go-square to support authored blobs (using the v1 share version).

In this PR is the new block validation logic which ensures that all "authored" blobs have a signer that matches the signer of the PFB

TxSim now will randomly use blobs of share version 0 and 1


test/util/testnode/config.go
```diff
@@ -139,7 +139,7 @@ func DefaultConfig() *Config {
 func DefaultConsensusParams() *tmproto.ConsensusParams {
 	cparams := types.DefaultConsensusParams()
 	cparams.Block.TimeIotaMs = 1
-	cparams.Block.MaxBytes = appconsts.DefaultMaxBytes
+	cparams.Block.MaxBytes = int64(appconsts.DefaultUpperBoundMaxBytes)
```
rootulp, line 142: [question] why change the default? AFAICT the value used on mainnet is `1974272`. Ref: 

Screenshot 2024-08-08 at 12 59 54 PM

```
DefaultMaxBytes= 1974272
DefaultUpperBoundMaxBytes=7897088
```

and the testnode should by default use those consensus params. With the exception of the max square size integration test, when do we need to create blocks bigger?

test/util/testnode/config.go
```diff
@@ -153,7 +153,7 @@ func DefaultTendermintConfig() *tmconfig.Config {
 	// Override the mempool's MaxTxBytes to allow the testnode to accept a
 	// transaction that fills the entire square. Any blob transaction larger
 	// than the square size will still fail no matter what.
-	maxTxBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * appconsts.ContinuationSparseShareContentSize
+	maxTxBytes := appconsts.DefaultUpperBoundMaxBytes
```
rootulp, line 156: 👍  thanks for the refactor to use a defined constant for this value.

[no change needed][separate issue]: why does the default here not match the default used on mainnet? IMO only one test (max square size test) should override this value and the rest of our tests should use the `appconsts.MaxTxBytes`

x/blob/types/blob_tx_test.go
```diff
@@ -22,16 +22,16 @@ import (
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 )
 
-func TestNewBlob(t *testing.T) {
+func TestNewV0Blob(t *testing.T) {
```
rootulp, line 25: [question] do we want a new test in this file for `TestNewV1Blob`?

rootulp (commented):
My only blocking comment is: should this feature be gated behind a new app version? If yes, then the tests in this PR that currently pass with app version 2 should fail and should only pass if app version is 3.

evan-forbes (dismissed):
:rocket:

evan-forbes:
started to address the versioning thing in https://github.com/celestiaorg/celestia-app/pull/3769

pkg/appconsts/global_consts.go
```diff
@@ -13,55 +11,9 @@ import (
...
-	// ShareInfoBytes is the number of bytes reserved for information. The info
-	// byte contains the share version and a sequence start indicator.
-	ShareInfoBytes = 1
-
-	// SequenceLenBytes is the number of bytes reserved for the sequence length
-	// that is present in the first share of a sequence.
-	SequenceLenBytes = 4
-
-	// ShareVersionZero is the first share version format.
-	ShareVersionZero = uint8(0)
```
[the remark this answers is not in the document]
cmwaters: I agree. Let me make this an issue so we can remember better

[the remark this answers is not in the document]
cmwaters: In general I think it might be a good idea to start an UPGRADING.md page

test/util/testnode/config.go
cmwaters: I think I mistook this for the mempool MaxTxsBytes. This should stay the same

app/test/process_proposal_test.go
```diff
@@ -259,6 +262,43 @@ func TestProcessProposal(t *testing.T) {
...
+			mutator: func(d *tmproto.Data) {
+				addr := signer.Account(accounts[0]).Address()
+				blob, err := share.NewV1Blob(ns1, data, addr)
+				require.NoError(t, err)
+				rawTx, _, err := signer.CreatePayForBlobs(accounts[0], []*share.Blob{blob}, user.SetGasLimit(100000), user.SetFee(100000))
+				require.NoError(t, err)
+				d.Txs[0] = rawTx
+				d.Hash = calculateNewDataHash(t, d.Txs)
+			},
+			appVersion:     appconsts.LatestVersion,
```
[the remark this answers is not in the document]
cmwaters: I was waiting on this until we have more information regarding your v3 prototype. At the moment it's not gated but it should be

pkg/appconsts/global_consts.go
```diff
@@ -13,55 +11,9 @@ import (
...
-	ContinuationCompactShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes - CompactShareReservedBytes
-
-	// FirstSparseShareContentSize is the number of bytes usable for data in the
-	// first sparse share of a sequence.
-	FirstSparseShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes - SequenceLenBytes
-
-	// ContinuationSparseShareContentSize is the number of bytes usable for data
-	// in a continuation sparse share of a sequence.
-	ContinuationSparseShareContentSize = ShareSize - NamespaceSize - ShareInfoBytes
+	DefaultShareVersion = share.ShareVersionZero
```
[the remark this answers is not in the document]
cmwaters: I actually don't know what is meant by the default here. I have left it as is because v1 is a new feature and I a) don't know if many people will use it and b) should be opt in anyway.

Maybe if this is just for tests we can change it but it would require a large refactor since the new version requires new arguments

test/util/testnode/config.go
cmwaters: Yeah I agree that there should be a single config that we use in all our tests (and only change them when they need to be changed). I think this should be addresses as a follow up

evan-forbes (approved):
I'm good merging without the gate

https://github.com/celestiaorg/celestia-app/issues/3771
https://github.com/celestiaorg/celestia-app/issues/3772

test/util/testnode/config.go
rootulp: +1 as a follow-up

rootulp (approved):
LGTM. 

[follow-up] maybe we should have a follow-up issue to ensure the authored blobs feature is only supported for app version 3. Even if we adopt the multiplexed app prototype, that shouldn't prevent us from adding defensive app version checks for new features (like authored blobs) in low level tests like `x/blob/types/blob_tx_test.go`.

PR #3767, opened 2024-08-08
refactor: signer to use txstatus

ninabarbakadze:

### DIFF
--- a/pkg/user/tx_client.go
+++ b/pkg/user/tx_client.go
@@ -26,6 +26,7 @@ import (
 	"github.com/celestiaorg/celestia-app/v3/app"
 	"github.com/celestiaorg/celestia-app/v3/app/encoding"
 	apperrors "github.com/celestiaorg/celestia-app/v3/app/errors"
+	"github.com/celestiaorg/celestia-app/v3/app/grpc/tx"
 	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v3/x/blob/types"
 	"github.com/celestiaorg/celestia-app/v3/x/minfee"
@@ -199,13 +200,17 @@ func SetupTxClient(
 
 // SubmitPayForBlob forms a transaction from the provided blobs, signs it, and submits it to the chain.
 // TxOptions may be provided to set the fee and gas limit.
-func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*blob.Blob, opts ...TxOption) (*sdktypes.TxResponse, error) {
+func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*blob.Blob, opts ...TxOption) (*sdktypes.TxResponse, *tx.TxStatusResponse, error) {
 	resp, err := client.BroadcastPayForBlob(ctx, blobs, opts...)
 	if err != nil {
-		return resp, err
+		return resp, nil, err
 	}
 
-	return client.ConfirmTx(ctx, resp.TxHash)
+	statusResponse, err := client.ConfirmTx(ctx, resp.TxHash)
+	if err != nil {
+		return resp, nil, err
+	}
+	return resp, statusResponse, nil
 }
 
 func (client *TxClient) SubmitPayForBlobWithAccount(ctx context.Context, account string, blobs []*blob.Blob, opts ...TxOption) (*sdktypes.TxResponse, error) {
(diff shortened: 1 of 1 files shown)


pkg/user/tx_client_test.go
```diff
@@ -45,6 +47,9 @@ func (suite *TxClientTestSuite) SetupSuite() {
 	suite.Require().NoError(err)
 	suite.txClient, err = user.SetupTxClient(suite.ctx.GoContext(), suite.ctx.Keyring, suite.ctx.GRPCClient, suite.encCfg, user.WithGasMultiplier(1.2))
 	suite.Require().NoError(err)
+	// FIXME: Temporary way of querying the raw log.
+	// TxStatus will natively support this in the future.
+	suite.serviceClient = sdktx.NewServiceClient(suite.ctx.GRPCClient)
```
cmwaters, line 52: Well we will still need this to get the gas wanted and gas used

pkg/user/tx_client.go
```diff
@@ -253,10 +268,12 @@ func (client *TxClient) BroadcastPayForBlobWithAccount(ctx context.Context, acco
...
-func (client *TxClient) SubmitTx(ctx context.Context, msgs []sdktypes.Msg, opts ...TxOption) (*sdktypes.TxResponse, error) {
+func (client *TxClient) SubmitTx(ctx context.Context, msgs []sdktypes.Msg, opts ...TxOption) (*TxResponse, error) {
 	resp, err := client.BroadcastTx(ctx, msgs, opts...)
-	if err != nil {
-		return resp, err
+	if err != nil && resp != nil {
+		return &TxResponse{Code: resp.Code, TxHash: resp.TxHash}, fmt.Errorf("failed to broadcast tx: %v", err)
+	} else if err != nil {
+		return &TxResponse{}, fmt.Errorf("failed to broadcast tx: %v", err)
 	}
```
cmwaters, line 277: nit: would be a little cleaner to use a helper function:
```
if err != nil {
    return parseTxResp(resp), fmt.Errorf("failed to broadcast tx: %w", err)
}
```
which turns the broadcastTxResponse to a TxResponse or nil if it's nil

cmwaters (approved):
LGTM - just a few minor nits

pkg/user/tx_client.go
ninabarbakadze: yeah i considered it but I thought that maybe this was more straight-forward and easy to understand

pkg/user/tx_client_test.go
ninabarbakadze: true i think i automatically copied this. will remove!

pkg/user/tx_client.go
```diff
@@ -414,32 +431,44 @@ func (client *TxClient) retryBroadcastingTx(ctx context.Context, txBytes []byte)
...
+					continue
+				}
+			case "COMMITTED":
+				txResponse := &TxResponse{
+					Height: resp.Height,
+					TxHash: txHash,
+					Code:   resp.ExecutionCode,
+				}
+				if resp.ExecutionCode != 0 {
+					return txResponse, fmt.Errorf("tx was included but failed with code %d: %s", resp.ExecutionCode, resp.Status)
```
evan-forbes, line 464: [not blocking]

using the word committed instead of included here I think is a smidge clearer as there's a less of a chance to confuse with inclusion into the mempool
```suggestion
					return txResponse, fmt.Errorf("tx was committed but failed with code %d: %s", resp.ExecutionCode, resp.Status)
```

pkg/user/tx_client.go
```diff
@@ -39,6 +40,14 @@ const (
 
 type Option func(client *TxClient)
 
+// TxResponse is a response from the chain after
+// a transaction has been submitted.
+type TxResponse struct {
+	Height int64
+	TxHash string
+	Code   uint32
+}
```
evan-forbes, line 49: what's the benefit of having our own type here? if this is different from the sdk TxResponse, is there a way to indicate it in the name?

pkg/user/tx_client.go
```diff
@@ -414,32 +431,44 @@ func (client *TxClient) retryBroadcastingTx(ctx context.Context, txBytes []byte)
...
+		if err != nil {
+			return &TxResponse{}, err
 		}
 
-		// Wait for the next round.
-		select {
-		case <-ctx.Done():
-			return &sdktypes.TxResponse{}, ctx.Err()
-		case <-pollTicker.C:
+		if err == nil && resp != nil {
```
evan-forbes, line 446: where are we notifying the user that their tx is in the mempool?

evan-forbes (approved):

pkg/user/tx_client.go
ninabarbakadze: I'll address this in a follow-up

pkg/user/tx_client.go
ninabarbakadze: I don't think we need to notify a user in this instance cause we're just confirming the inclusion in a block

PR #3769, opened 2024-08-08
chore: add version 3

evan-forbes:
## Overview

add support for app version 3

### DIFF
--- a/app/module/versioned_ibc_module_test.go
+++ b/app/module/versioned_ibc_module_test.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/celestiaorg/celestia-app/v3/app/module"
 	mocks "github.com/celestiaorg/celestia-app/v3/app/module/mocks"
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
+	v2 "github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v2"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/ibc-go/v6/modules/core/04-channel/types"
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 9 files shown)


app/module/versioned_ibc_module_test.go
```diff
@@ -188,6 +190,17 @@ func TestVersionedIBCModule(t *testing.T) {
...
+			name:    "OnAcknowledgementPacket with supported version",
+			version: 3,
+			setupMocks: func(ctx sdk.Context) {
+				mockWrappedModule.EXPECT().OnAcknowledgementPacket(ctx, types.Packet{}, []byte{}, sdk.AccAddress{}).Return(nil)
+			},
+			method: func(ctx sdk.Context) (interface{}, error) {
+				return nil, versionedModule.OnAcknowledgementPacket(ctx, types.Packet{}, []byte{}, sdk.AccAddress{})
+			},
+			expectedValue: nil,
+		},
```
evan-forbes, line 203: it would be nice to refactor these tests to make it easier to add a version without copy pasting gobs of code

app/modules.go
```diff
@@ -96,95 +97,95 @@ func (app *App) setupModuleManager(skipGenesisInvariants bool) error {
 	app.manager, err = module.NewManager([]module.VersionedModule{
 		{
 			Module:      genutil.NewAppModule(app.AccountKeeper, app.StakingKeeper, app.BaseApp.DeliverTx, app.txConfig),
-			FromVersion: v1, ToVersion: v2,
+			FromVersion: v1, ToVersion: appconsts.LatestVersion,
```
evan-forbes, line 100: for things that are not being removed, I think we should just use the latest version. when they get removed or replaced, then we add a ToVersion.

open to not donig this

pkg/appconsts/v2/app_consts.go
```diff
@@ -1,9 +1,7 @@
 package v2
 
 const (
-	Version              uint64 = 2
-	SquareSizeUpperBound int    = 128
-	SubtreeRootThreshold int    = 64
```
evan-forbes, line 6: 1) these weren't being used 2) atm, I don't think we benefit from having copies of constants in each version. Perhaps we can version these differently? Just use the v1 version until they change, then specify the new version

evan-forbes:
I'm assuming that this PR is still missing things as there are version checks eeevvvveeerrrryyyywwwwhhhheeeerrrrreeee so leaving as a draft

evan-forbes:
closing in favor of #3870

PR #3774, opened 2024-08-09
feat: require bbr by default

### LINKED ISSUE #3745: By default, stop the node unless using bbr
Since using bbr as a congestion control algo increases throughput significantly, by default the node should not start without it. Node operators still need the ability to bypass this default.

## AC

- [ ] Add a mechanism that will not start the node unless using bbr as a congestion control algo in any environment (docker, vm, hardware etc)
- [ ] Add a flag that bypasses the above mechanism

evan-forbes:
## Overview

This PR adds a requirement to use bbr by default. This requirement can be bypassed by using a flag. Warnings are returned when installing, and there is a makefile command that will configure the system to use bbr.

### ~~~ Note for Reviewers ~~~

Most of this PR is copying over the start command from the sdk. Alternatively, we could simply modify the sdk. That would not require copy past forking, but the trend has been that every change that we make to the sdk eventually gets moved here, so I figured we should just skip to doing that. This also gives us finer grained control in the future. I'm definitely not tied to this approach, and if we'd rather change our fork, then I'm happy go do that instead.

closes #3745


cmd/celestia-appd/cmd/start.go
```diff
@@ -0,0 +1,627 @@
...
+			}
+
+			_, err := server.GetPruningOptionsFromFlags(serverCtx.Viper)
+			return err
+		},
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			err := checkBBR(cmd)
+			if err != nil {
+				return err
+			}
```
evan-forbes, line 115: besides adding imports and not exporting a few functions that previously were, this is the only change to the code :upside_down_face:

scripts/single-node.sh
```diff
@@ -102,7 +102,8 @@ startCelestiaApp() {
     --api.enable \
     --grpc.enable \
     --grpc-web.enable \
-    --v2-upgrade-height 3
+    --v2-upgrade-height 3 \
+    --force-no-bbr // no need to require bbr usage on a local node
```
rootulp, line 106:
```suggestion
    --force-no-bbr // no need to require BBR usage on a local node
```

README.md
```diff
@@ -77,6 +77,22 @@ See <https://docs.celestia.org/nodes/celestia-app> for more information.
 
 ## Usage
 
+First, make sure that the BBR congestion control algorithm is enabled in the
```
rootulp, line 80: [nit] a link would be helpful. Also expand the abbreviation b/c this is the first usage a reader will see

```suggestion
First, make sure that the [BBR](https://www.ietf.org/archive/id/draft-cardwell-iccrg-bbr-congestion-control-01.html) ("Bottleneck Bandwidth and Round-trip propagation time") congestion control algorithm is enabled in the
```

README.md
```diff
@@ -77,6 +77,22 @@ See <https://docs.celestia.org/nodes/celestia-app> for more information.
...
+system's kernel. The result should contain `bbr`:
+
+```sh
+sysctl net.ipv4.tcp_congestion_control
+```
+
+If not, enable it in Linux by calling the `make use-bbr` or by running:
```
rootulp, line 87:
```suggestion
If not, enable it on Linux by calling the `make use-bbr` or by running:
```

Makefile
```diff
@@ -212,3 +212,27 @@ prebuilt-binary:
...
+check-bbr:
+	@echo "Checking if BBR is enabled..."
+	@if [ "$(sysctl net.ipv4.tcp_congestion_control | awk '{print $3}')" != "bbr" ]; then \
+	    echo "WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README for further instruction"; \
+	else \
+	    echo "BBR is enabled."; \
+	fi
+.PHONY: check-bbr
+
+# Target to configure the system to use BBR
```
rootulp, line 225: So that this shows up in `make help` commands

```suggestion
## enable-bbr: Enable BBR congestion control algorithm. Only works on Linux.
```

Makefile
```diff
@@ -212,3 +212,27 @@ prebuilt-binary:
 		ghcr.io/goreleaser/goreleaser-cross:${GOLANG_CROSS_VERSION} \
 		release --clean
 .PHONY: prebuilt-binary
+
+check-bbr:
```
rootulp, line 216:
```suggestion

## check-bbr: Check if your system uses BBR congestion control algorithm. Only works on Linux.
check-bbr:
```

Makefile
```diff
@@ -212,3 +212,27 @@ prebuilt-binary:
...
+	@echo "Checking if BBR is enabled..."
+	@if [ "$(sysctl net.ipv4.tcp_congestion_control | awk '{print $3}')" != "bbr" ]; then \
+	    echo "WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README for further instruction"; \
```
rootulp, line 219:
```suggestion
	    echo "WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README."; \
```

cmd/celestia-appd/cmd/start.go
```diff
@@ -0,0 +1,632 @@
...
+	rootCmd.AddCommand(
+		startCmd,
+		tendermintCmd,
+		server.ExportCmd(appExport, defaultNodeHome),
+		version.NewVersionCommand(),
+		server.NewRollbackCmd(appCreator, defaultNodeHome),
+	)
+}
+
+// checkBBR checks is bbr is configured to be used as a congestion control algo.
```
rootulp, line 569:
```suggestion
// checkBBR checks if BBR is configured to be used as a congestion control algorithm.
```

cmd/celestia-appd/cmd/start.go
```diff
@@ -0,0 +1,632 @@
...
+func checkBBR(command *cobra.Command) error {
+	const (
+		errorMsg = `
+// The BBR congestion control algorithm does not appear to be enabled in this 
+// system's kernel. This is important for the p2p stack to be performant.
+//
+// to enable bbr call:
```
rootulp, line 576:
```suggestion
// To enable BBR call:
```

rootulp (approved):
I think this will need a corresponding PR to https://github.com/celestiaorg/docs to instruct users to enable BBR or provide the `--force-no-bbr` flag

cmwaters:
I think this should just be in V3. We could do a patch release for v2 but it will require extra communication to tell node operators of the change and hopefully v3 will shortly be out

cmwaters (approved):
utAck

FYI: I think a lot of the start/stop cli logic will get refactored when we use the new upgrade tooling that Rootul is working on

rootulp (approved):

scripts/single-node.sh
```diff
@@ -102,7 +102,8 @@ startCelestiaApp() {
     --api.enable \
     --grpc.enable \
     --grpc-web.enable \
-    --v2-upgrade-height 3
+    --v2-upgrade-height 3 \
+    --force-no-bbr // no need to require BBR usage on a local node
```
qezz, line 106: I believe it's not a valid bash syntax

so, `#` for comments

cmd/celestia-appd/cmd/start.go
```diff
@@ -0,0 +1,632 @@
...
+// 
+// and can be verified to be running using
+//
+sysctl net.ipv4.tcp_congestion_control
+
+// This might not work for all systems, you might have to search online to 
+// figure out how to enable bbr for your system.
+//
+// While this node will get worse performance using something other than bbr,
+// If you need to bypass this block use the "--force-no-bbr true" flag.
```
qezz, line 591: depending on the parser, `--x true` can give different results, i.e. it will probably be parsed as `--x` and `true` as a positional argument

if you want to keep explicit value for it, then it should be `--force-no-bbr=true`, if it's supported by the args parser

PR #3786, opened 2024-08-16
refactor: remove unnecessary `config.Seal()`

rootulp:
- Add unit tests to cosmos SDK config set up
- Remove an unnecessary `config.Seal` invocation that has been around for a [long time](https://github.com/celestiaorg/celestia-app/commit/5e4a1dc56e42e7341b6da9526430b12dcc7a6691#diff-9e0425d31b7cf4072a746feb0c6bc1e0045b639e01dd0b5ac2f08e8c2952c886R120) but isn't necessary because the `init()` command always sets and seals the config.


rootulp:
I may expose a method on the config to check if it's sealed or not and then conditionally seal it.

cmd/celestia-appd/cmd/root.go
```diff
@@ -100,9 +99,6 @@ func NewRootCmd() *cobra.Command {
 
 // initRootCommand performs a bunch of side-effects on the root command.
 func initRootCommand(rootCommand *cobra.Command, encodingConfig encoding.Config) {
-	config := sdk.GetConfig()
```
staheri14, line 103: Was this line meant to be deleted? Asking in case it was not intentional

app/sdk_config_test.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func Test_setCosmosSDKConfig(t *testing.T) {
+	config := sdk.GetConfig()
+	assert.Equal(t, Bech32PrefixAccAddr, config.GetBech32AccountAddrPrefix())
+	assert.Equal(t, Bech32PrefixAccPub, config.GetBech32AccountPubPrefix())
+	assert.Equal(t, Bech32PrefixValAddr, config.GetBech32ValidatorAddrPrefix())
+	assert.Equal(t, Bech32PrefixValPub, config.GetBech32ValidatorPubPrefix())
+	assert.Equal(t, Bech32PrefixConsAddr, config.GetBech32ConsensusAddrPrefix())
+	assert.Equal(t, Bech32PrefixConsPub, config.GetBech32ConsensusPubPrefix())
```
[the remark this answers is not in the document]
staheri14: @rootulp I also agree with this comment, the `setCosmosSDKConfig()` looks like missing, (or if not, adding some clarifying comments would be helpful).

staheri14 (commented):
LGTM! Included some comments.

cmd/celestia-appd/cmd/root.go
rootulp: Yep it meant to be deleted. It doesn't need to be invoked here because it's already invoked via `init()`.

app/sdk_config_test.go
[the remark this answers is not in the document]
rootulp: It's invoked via `init()`. In general I think `init()` should be avoided because it makes it harder to unit test.

evan-forbes (approved):

cmd/celestia-appd/cmd/root.go
staheri14: >it's already invoked via init().

By `init()`, do you mean the following function from`app/sdk_config.go`?
```go
func init() {
	setCosmosSDKConfig()
}
```
If yes, then it seems it has no usage in the codebase (I looked up for its usage in your branch, and nothing showed up). Am I missing something?

rootulp: `init` is automatically invoked. https://go.dev/doc/effective_go#init

staheri14 (approved):
LGTM!

PR #3729, opened 2024-07-24
feat: prototype versioned app module upgrades

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3625

Still WIP

PR #3787, opened 2024-08-16
feat: distinct go modules for all state machine modules

rootulp:
Unblocks https://github.com/celestiaorg/celestia-app/pull/3729

## Description

One of the issues encountered when working on the [multi-app prototype](https://github.com/celestiaorg/celestia-app/pull/3729) is that protobuf definitions get registered multiple times (once per state machine imported into the multiplexer). This results in errors like:

```
panic: concrete type *types.MsgPayForBlobs has already been registered under typeURL /, cannot register *types.MsgRegisterEVMAddress under same typeURL. This usually means that there are conflicting modules registering different concrete types for a same interface implementation
```

To resolve, this PR introduces distinct Go modules per state machine module. This enables the v2 and v3 state machine to import the exact same module (i.e. `github.com/celestiaorg/celestia-app/x/blob`) which will result in just one proto type being registered per actual type.


rootulp:
Discussed with Callum offline and closing this for now. As a prerequisite to this work we need to make all state machine modules not import `app`.

PR #3788, opened 2024-08-18
feat: add error log to txstatus

ninabarbakadze:
## Overview

- [x] Bump celestia-core verison including refactored TxStatus work
- [x] Extend the rpc response to return the error log
- [x] Update TxClient with extended TxStatus struct


ninabarbakadze (commented):
I think breaking proto should not be a problem here

cmwaters (dismissed):

pkg/user/tx_client.go
```diff
@@ -445,26 +441,26 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+			case core.TxStatusCommitted:
 				txResponse := &TxResponse{
 					Height: resp.Height,
 					TxHash: txHash,
 					Code:   resp.ExecutionCode,
+					Error:  resp.Error,
 				}
 				if resp.ExecutionCode != 0 {
-					return txResponse, fmt.Errorf("tx was included but failed with code %d: %s", resp.ExecutionCode, resp.Status)
+					return txResponse, fmt.Errorf("tx was committed but failed with code %d: %s", resp.ExecutionCode, resp.Status)
```
cmwaters, line 460: This could change to include the error
```suggestion
					return txResponse, fmt.Errorf("tx was committed but failed with code %d: %s", resp.ExecutionCode, resp.Error)
```

pkg/user/tx_client_test.go
```diff
@@ -47,8 +48,6 @@ func (suite *TxClientTestSuite) SetupSuite() {
 	suite.Require().NoError(err)
 	suite.txClient, err = user.SetupTxClient(suite.ctx.GoContext(), suite.ctx.Keyring, suite.ctx.GRPCClient, suite.encCfg, user.WithGasMultiplier(1.2))
 	suite.Require().NoError(err)
-	// FIXME: Temporary way of querying the raw log.
-	// TxStatus will natively support this in the future.
 	suite.serviceClient = sdktx.NewServiceClient(suite.ctx.GRPCClient)
```
cmwaters, line 51: Do we still need the service client here?

ninabarbakadze: yes because we are polling GasWanted GasUsed

pkg/user/tx_client.go
```diff
@@ -46,6 +47,7 @@ type TxResponse struct {
 	Height int64
 	TxHash string
 	Code   uint32
+	Error  string
```
cmwaters, line 50: I would advocate for removing error here because it should be returned in the `error` type

ninabarbakadze: done!

pkg/user/tx_client.go
```diff
@@ -445,26 +440,25 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
 					Code:   resp.ExecutionCode,
 				}
 				if resp.ExecutionCode != 0 {
-					return txResponse, fmt.Errorf("tx was included but failed with code %d: %s", resp.ExecutionCode, resp.Status)
+					return txResponse, fmt.Errorf("tx was committed but failed with code %d: %s", resp.ExecutionCode, resp.Error)
 				}
 				return txResponse, nil
-			case "EVICTED":
+			case core.TxStatusEvicted:
 				return &TxResponse{}, fmt.Errorf("tx: %s was evicted from the mempool", txHash)
```
cmwaters, line 462: sorry looks like I missed this in the first run.
```suggestion
				return &TxResponse{TxHash: txHash}, fmt.Errorf("tx: %s was evicted from the mempool", txHash)
```

proto/celestia/core/v1/tx/tx.proto
```diff
@@ -33,6 +33,8 @@ message TxStatusResponse {
     // and returns whether it was successful or errored. A non zero
     // execution code indicated an error.
     uint32 execution_code = 3;
+    // error log for failed transactions.
+    string error = 4;
     // status is the status of the transaction.
-    string status = 4;
+    string status = 5;
```
cmwaters, line 39: So whenever you modify the field number, that's a breaking change. To do this in a non-breaking change you should add `string error = 5;` not `4`

ninabarbakadze: we have never released this so we don't have to maintain any backwards compatibility

evan-forbes (approved):

cmwaters (approved):

Issue #3805, opened 2024-08-22
goreleaser-cross doesn't support `version_template`


rootulp:
## Problem

Goreleaser failed to publish binaries on v2.1.1 because https://github.com/celestiaorg/celestia-app/actions/runs/10497069730/job/29079086420#step:6:66

## Proposal

Fix it

PR #3806, opened 2024-08-22
ci: use goreleaser-cross for goreleaser-check command

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3805

[goreleaser v2.2.0](https://github.com/goreleaser/goreleaser/releases/tag/v2.2.0) deprecated the `name_template` field in favor of `version_template`. Our goreleaser-check command used a GH action that was using goreleaser v2.2.0 so we fixed the deprecation in https://github.com/celestiaorg/celestia-app/pull/3797 but the make prebuilt-binary command uses goreleaser-cross which was using a previous version of goreleaser that fails for the new field.

This PR updates the goreleaser-check command to use the goreleaser-cross Docker image so that the check command and prebuilt-binary command uses the same version.

Note: we'll need to update `name_template` to `version_template` when we upgrade to Go 1.23.0 because that will use goreleaser-cross 1.23.0 which bumped to goreleaser 2.22.0. See https://github.com/goreleaser/goreleaser-cross/pull/77


cmwaters (approved):
utAck

ninabarbakadze (approved):
another utAck from me

Issue #3804, opened 2024-08-21
`multistore restore: cannot import into non-IAVL store \"acc\": internal logic error`


rootulp:
## Context

I'm trying to start an Arabica consensus node and state sync it to the tip of the chain so that I can make progress on a separate issue.

## Problem

When I run [arabica.sh](https://gist.github.com/rootulp/467e7fe3617cf5937185a8f63ac767d0) I encounter an error with state sync:

```
$ ./scripts/arabica.sh
celestia-app home: /Users/rootulp/.celestia-app
celestia-app version: 2.0.0

Are you sure you want to delete: /Users/rootulp/.celestia-app? [y/n] y
Deleting /Users/rootulp/.celestia-app...
Initializing config files...
Settings seeds in config.toml...
Block height: 1769034
Trust hash: 110BF83440E24CDD5DA5F4D4872B85555EBBE52A590A8730F301C0BD046038F3
Enabling state sync in config.toml...
Downloading genesis file...
Starting celestia-appd...
4:25PM INF starting node with ABCI Tendermint in-process
4:25PM INF service start impl=multiAppConn module=proxy msg={}
4:25PM INF service start connection=query impl=localClient module=abci-client msg={}
4:25PM INF service start connection=snapshot impl=localClient module=abci-client msg={}
4:25PM INF service start connection=mempool impl=localClient module=abci-client msg={}
4:25PM INF service start connection=consensus impl=localClient module=abci-client msg={}
4:25PM INF service start impl=EventBus module=events msg={}
4:25PM INF service start impl=PubSub module=pubsub msg={}
4:25PM INF service start impl=IndexerService module=txindex msg={}
4:25PM INF Version info abci=0.17.0 block=11 cmtbft_version=0.34.29 commit_hash= p2p=8
4:25PM INF This node is not a validator addr=20EB400293E85D22227E1A403D46B39B52AA754C module=consensus pubKey=W/cXE0Gbhz5HR4k++QUOHao7XZYOZhnc7ekBiunNbGI=
4:25PM INF P2P Node ID ID=b7816c37f3a5a2e4f16727a7719a2cfb52f6242a file=/Users/rootulp/.celestia-app/config/node_key.json module=p2p
4:25PM INF Adding persistent peers addrs=[] module=p2p
4:25PM INF Adding unconditional peer ids ids=[] module=p2p
[...]

Issue #3807, opened 2024-08-22
Add a test for state sync


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/issues/3804

## Problem

We discovered a bug with state sync on a testnet. Ideally this should've been caught via a unit or integration test.

## Proposal

Add a test that tries to state sync in a manner that will catch regressions like https://github.com/celestiaorg/celestia-app/issues/3804

PR #3808, opened 2024-08-22
fix: state sync on v2.x

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3804 with a hacky approach that infers the app version based on the snapshot height and mounts KV stores accordingly.

Kinda related to https://github.com/celestiaorg/celestia-app/issues/3804 but we didn't hit this because we don't have tests that try to state sync so I created https://github.com/celestiaorg/celestia-app/issues/3807

## Testing

Using both the `arabica.sh` and `mocha.sh` scripts in this PR, I can start state syncing. I let the Arabica node fully state sync to the tip of the chain and it worked.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -798,13 +798,16 @@ func (app *App) OfferSnapshot(req abci.RequestOfferSnapshot) abci.ResponseOfferS
 	if app.upgradeHeightV2 == 0 {
 		app.Logger().Debug("v2 upgrade height not set, assuming app version 2")
 		app.mountKeysAndInit(v2)
-	} else if req.Snapshot.Height >= uint64(app.upgradeHeightV2) {
-		app.Logger().Debug("snapshot height is greater than or equal to upgrade height, assuming app version 2")
-		app.mountKeysAndInit(v2)
-	} else {
-		app.Logger().Debug("snapshot height is less than upgrade height, assuming app version 1")
-		app.mountKeysAndInit(v1)
+		return app.BaseApp.OfferSnapshot(req)
 	}
 
+	if req.Snapshot.Height >= uint64(app.upgradeHeightV2) {
+		app.Logger().Debug("snapshot height is greater than or equal to upgrade height, assuming app version 2")
+		app.mountKeysAndInit(v2)
+		return app.BaseApp.OfferSnapshot(req)
+	}
+
+	app.Logger().Debug("snapshot height is less than upgrade height, assuming app version 1")
+	app.mountKeysAndInit(v1)
 	return app.BaseApp.OfferSnapshot(req)
 }
--- a/app/app.go
+++ b/app/app.go
@@ -167,6 +167,9 @@ type App struct {
 	// upgradeHeightV2 is used as a coordination mechanism for the height-based
 	// upgrade from v1 to v2.
 	upgradeHeightV2 int64
+	// HACK: remove this once we have a better way to handle versioning while
+	// state syncing
+	stateSyncVersion uint64
 	// MsgGateKeeper is used to define which messages are accepted for a given
 	// app version.
 	MsgGateKeeper *ante.MsgVersioningGateKeeper
@@ -519,11 +522,17 @@ func (app *App) Info(req abci.RequestInfo) abci.ResponseInfo {
 		if err != nil {
 			panic(err)
 		}
-		appVersion := app.GetAppVersionFromParamStore(ctx)
-		if appVersion > 0 {
-			app.SetAppVersion(ctx, appVersion)
+		// Info get's called after state sync is complete. Here
+		// we check for this case and set the app version accordingly.
+		if app.stateSyncVersion > 0 {
+			app.SetAppVersion(ctx, app.stateSyncVersion)
 		} else {
-			app.SetAppVersion(ctx, v1)
+			appVersion := app.GetAppVersionFromParamStore(ctx)
+			if appVersion > 0 {
+				app.SetAppVersion(ctx, appVersion)
+			} else {
+				app.SetAppVersion(ctx, v1)
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


app/app.go
```diff
@@ -784,3 +785,26 @@ func (app *App) InitializeAppVersion(ctx sdk.Context) {
...
+	if app.upgradeHeightV2 == 0 {
+		app.Logger().Debug("v2 upgrade height not set, assuming app version 2")
+		app.mountKeysAndInit(v2)
+	} else if req.Snapshot.Height >= uint64(app.upgradeHeightV2) {
+		app.Logger().Debug("snapshot height is greater than or equal to upgrade height, assuming app version 2")
+		app.mountKeysAndInit(v2)
+	} else {
+		app.Logger().Debug("snapshot height is less than upgrade height, assuming app version 1")
+		app.mountKeysAndInit(v1)
+	}
```
rootulp, line 807: @cmwaters is it necessary to also invoke something like:

```
			app.SetAppVersion(ctx, v2)
```

here? When I tried that, I got a panic because

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x10585c2e0]

goroutine 135 [running]:
github.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).NewContext(_, _, {{0x0, 0x0}, {0x0, 0x0}, 0x0, {0x0, 0x0, 0x0}, ...})
	/Users/rootulp/git/rootulp/celestiaorg/cosmos-sdk/baseapp/test_helpers.go:46 +0x70
github.com/celestiaorg/celestia-app/v2/app.(*App).OfferSnapshot(0x140007dd508, {0x14001266b40?, {0x140010be540?, 0x104c7cbac?, 0x1?}})
	/Users/rootulp/git/rootulp/celestiaorg/celestia-app/app/app.go:794 +0x6c
github.com/tendermint/tendermint/abci/client.(*localClient).OfferSnapshotSync(0x14001c003c0, {0x14001266b40?, {0x140010be540?, 0x107597fc0?, 0x1?}})
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/abci/client/local_client.go:330 +0x110
github.com/tendermint/tendermint/proxy.(*appConnSnapshot).OfferSnapshotSync(0x1400035c0a0?, {0x14001266b40?, {0x140010be540?, 0x140013bfec0?, 0x20?}})
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/proxy/app_conn.go:199 +0x30
github.com/tendermint/tendermint/statesync.(*syncer).offerSnapshot(0x14001710240, 0x140005b9da0)
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/statesync/syncer.go:325 +0x220
github.com/tendermint/tendermint/statesync.(*syncer).Sync(_, _, _)
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/statesync/syncer.go:269 +0x178
github.com/tendermint/tendermint/statesync.(*syncer).SyncAny(_, _, _)
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/statesync/syncer.go:185 +0x3f0
github.com/tendermint/tendermint/statesync.(*Reactor).Sync(_, {_, _}, _)
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/statesync/reactor.go:303 +0x2f8
github.com/tendermint/tendermint/node.startStateSync.func1()
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/node/node.go:725 +0x90
created by github.com/tendermint/tendermint/node.startStateSync in goroutine 14
	/Users/rootulp/go/pkg/mod/github.com/celestiaorg/celestia-core@v1.40.0-tm-v0.34.29/node/node.go:724 +0x31c
```

app/app_test.go
```diff
@@ -47,6 +53,45 @@ func TestNew(t *testing.T) {
 	})
 }
 
+func TestOfferSnapshot(t *testing.T) {
```
rootulp, line 56: Note: this unit test doesn't actually hit the `error` observed before the fix. To do that we need to start a channel and stream chunks of the snapshot so I may opt to do that in an integration test instead.

evan-forbes (approved):
might want to mention somewhere in the scripts or in their names that they do statesync, but not blocking

thanks for testing state sync!! great catch. ideally we test this with knuu

app/app.go
cmwaters: Looks like the panic is from how you initialise the context.

cmwaters: But yes, we should be setting the app version here

cmwaters:
I am going to test this out manually. If it passes we can merge this for now and then work on adding it to the e2e tests and a more long term fix

app/app.go
```diff
@@ -784,3 +794,31 @@ func (app *App) InitializeAppVersion(ctx sdk.Context) {
...
+
+	if req.Snapshot.Height >= uint64(app.upgradeHeightV2) {
+		app.Logger().Debug("snapshot height is greater than or equal to upgrade height, assuming app version 2")
+		app.mountKeysAndInit(v2)
+		app.stateSyncVersion = v2
+		return app.BaseApp.OfferSnapshot(req)
+	}
+
+	app.Logger().Debug("snapshot height is less than upgrade height, assuming app version 1")
+	app.mountKeysAndInit(v1)
```
rootulp, line 822: [question] why not?

```suggestion
	app.stateSyncVersion = v1
	app.mountKeysAndInit(v1)
```

app/app.go
cmwaters: Turns out we don't because the state that we sync has the app version and when we call `Info` we update it from the store

rootulp: Callum tested offline and since `Info` gets invoked after state sync and app version is set there, it doesn't need to be set here.

cmwaters (approved):
So long as everything is green this should be good to merge

ninabarbakadze (approved):

Issue #3824, opened 2024-08-27
ICA AllowMessages on Arabica is incorrect


rootulp:
## Problem

On Arabica, the ICA host `AllowMessages` param is incorrect. It is currently a wildcard when it should be an allowlist of messages:

```
root@rootulp-arabica:~# celestia-appd query params subspace icahost AllowMessages
key: AllowMessages
subspace: icahost
value: '["*"]'
```

Wildcard is the default however it should've been overwritten with this list: https://github.com/celestiaorg/celestia-app/blob/2f74d9b84353b714295d3179b6df6c8315ffb813/app/ica_host.go#L3-L18

rootulp:
On Arabica and Mocha we can use governance to modify it but we should fix the override so that it doesn't need a governance vote on mainnet.

rootulp:
Please see [meeting notes](https://docs.google.com/document/d/1XraEd_nDnxDs-M4Qp8ogUwENzHbrPcjeF7g2BrJltQI) and [hackMD doc](https://hackmd.io/vjutJsiHTrWliMVmvRfyaQ) for full details. 

Summary:
- Short term: we'll fix on Arabica, Mocha, and Mainnet via a governance proposal
- Long term: in celestia-app v3.x we will hard-code the param to an explicit allowlist and make the param non governance modifiable
- Note: this means there will be a ~1 week time period on mainnet where the ICA allowlist will be a wildcard instead of an explicit allowlist. We think this isn't a serious security concern b/c other Cosmos chains (Cosmos Hub, Neutron) have this param.

rootulp:
We're tackling this with three solutions:

1. https://github.com/celestiaorg/celestia-app/issues/3827
2. https://github.com/celestiaorg/celestia-app/issues/3836
3. https://github.com/celestiaorg/celestia-app/issues/3837

PR #3826, opened 2024-08-27
fix: ICA module params on v2.x

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3824 by fixing the issue for subsequent releases + mainnet

## Testing

On Arabica, I can state sync from a recent snapshot the tip. To truly test this I need to state sync from a snapshot prior to the v2 upgrade height and then execute all the blocks after that but my node can't discover snapshots from that long ago. Created https://github.com/celestiaorg/celestia-app/issues/3828 which would help


evan-forbes (approved):
lgtm

rootulp:
Converted to draft b/c we're evaluating our options. Will have a discussion about potential fixes tomorrow.

Issue #3768, opened 2024-08-08
Improve Tx submission: handling evictions


cmwaters:
### Summary

The tools we provide users to assist in tx submission has come a long way and we're starting to see a more production grade system that can be relied upon by rollup sequencers or RaaS providers.

As to date, we now have:
- Local tracking of nonce to allow multiple transactions in a single block
- Multiple accounts so a single light node can submit on behalf of multiple rollups
- A TxStatus endpoint that is more lightweight and gives us more information on the lifecycle of a transaction [In progress].

### Problem Definition

There are still a few missing areas where we can improve robustness. One important one is handling evictions which cause nonce mismatches for an entire batch of transactions. We also can't rely on the consensus node we submit to for the new nonces because they could lie and cause us to break replay protection.

### Proposal

What I'm proposing is a little more elaborate. Some of the ideas I took from maelstrom which focused a lot on reliable tx submission. Essentially, the txClient needs to have a small pool (like a mempool) that tracks the transactions over their lifecycle and only discards them once they have been confirmed on chain. As transactions can potentially live forever, all transactions must have a TimeoutHeight which defaults to 5 blocks above the current height. That way as we track the blockchain we can always reach a conclusive result as to whether a transaction has failed or been successful. I propose a default of 5 that can be configured by the user.

Now, when it comes to [handling nonce mismatches](https://github.com/celestiaorg/celestia-app/blob/b684acd65178df2361c0bce6f7cab29fcc142b25/pkg/user/tx_client.go#L334), if the client knows all the unconfirmed transactions they have submitted, they can replay the transactions that they have signed to the receiving node up to the current transaction. If a transaction is evicted or times out, the other transactions can all have their nonces update accordingly (they may also need to wait past their expiration date to avoid double execution). Having the user set the expiration time also allows the user to resubmit sooner with a higher fee (although I think pricing holistically should be solved elsewhere).

cmwaters:
There are two cases where evictions are causing the `TxClient` to submit transactions that are incorrect according to the consensus nodes local view:

1. Consensus node’s view of nonce is **greater** than `TxClient`: https://github.com/celestiaorg/celestia-app/issues/4784
2. Consensus node’s view of nonce is **less** than `TxClient`: https://github.com/celestiaorg/celestia-node/issues/3986 (used to be more common until we implemented a tracking mechanism)

For 1. this can happen if:
- the user is submitting transactions using the same keys on a different `TxClient` so it's not aware that other transactions are being committed
[...]

PR #3830, opened 2024-08-28
refactor: index submitted txs in tx client and improve nonce management

### LINKED ISSUE #3899: Index submitted transactions for nonce tracking in tx client
## Description

Currently, if a transaction has a nonce mismatch, we query the transaction nonce and resubmit transactions with the updated nonce. We should stop resubmitting transactions on users' behalf and instead track all submitted transactions and their nonces locally. If a transaction gets evicted during ConfirmTx(), we can readjust the nonce to what it was before the transaction was submitted on-chain, allowing users to resubmit transactions if they choose to do so.

ninabarbakadze:
## Overview

Fixes #3899


ninabarbakadze (commented):
square size test started failing, investigating that.

pkg/user/tx_client.go
```diff
@@ -137,6 +148,9 @@ type TxClient struct {
 	defaultGasPrice float64
 	defaultAccount  string
 	defaultAddress  sdktypes.AccAddress
+	txPool          map[string]poolTxInfo
```
ninabarbakadze, line 151: [Question] do we want to have a way of periodically clearing the cache?

pkg/user/tx_client_test.go
```diff
@@ -208,11 +245,62 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
...
+					break
+				}
+			}
+		}
+
+		seqAfterEviction := sender.Sequence()
+		require.Equal(t, seqBeforeEviction, seqAfterEviction)
+
+		// Increase ttl again
+		suite.testnodeConfig.TmConfig.Mempool.TTLDuration = 120 * time.Second
```
evan-forbes, line 294: do we need to restart the node for it to take effect?

pkg/user/tx_client_test.go
```diff
@@ -208,11 +245,62 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
...
+		msg := bank.NewMsgSend(sender.Address(), testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(app.BondDenom, 10)))
+		var seqBeforeEviction uint64
+		for {
+			// There is a chance one of the txs will get into a block so we have to keep updating the sequence
+			seqBeforeEviction = sender.Sequence()
+			resp, err := suite.txClient.BroadcastTx(suite.ctx.GoContext(), []sdk.Msg{msg}, fee, gas)
+			require.NoError(t, err)
+			_, err = suite.txClient.ConfirmTx(suite.ctx.GoContext(), resp.TxHash)
+			if err != nil {
+				if err.Error() == "tx was evicted from the mempool" {
```
evan-forbes, line 284: if there's a bug, could this just loop indefinitely?

pkg/user/tx_client_test.go
```diff
@@ -208,11 +245,62 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
...
-		_, err = suite.txClient.ConfirmTx(suite.ctx.GoContext(), resp.TxHash)
+		require.Equal(t, resp.Code, abci.CodeTypeOK)
+
+		nonce, signer, exists := suite.txClient.GetTxFromLocalMempool(resp.TxHash)
+		require.True(t, exists)
+		require.Equal(t, suite.txClient.DefaultAccountName(), signer)
+		seq := suite.txClient.Signer().Account(suite.txClient.DefaultAccountName()).Sequence()
+		// Successfully broadcast transaction increases the nonce
+		// txInfo is indexed before the nonce is increased
+		require.Equal(t, seq, nonce+1)
```
evan-forbes, line 256: [not blocking]

looks like we repeat a few things in these test runs, is it worth using a table driven test or pulling this out?

evan-forbes (dismissed):

evan-forbes:
fwiw, Evan likes the tx client having a local mempool. Quite powerful.

pkg/user/tx_client.go
ninabarbakadze: @evan-forbes @cmwaters I'd appreciate your input here

pkg/user/tx_client_test.go
ninabarbakadze: no because it's a pointer in the config and mempool just reads the config so if you update smth in the config it will automatically reference this new value

pkg/user/tx_client_test.go
ninabarbakadze: if for some reason this loops indefinitely(which should not be the case) the test will hang until it times out so we'll know that the loop is broken

pkg/user/tx_client_test.go
ninabarbakadze: I can pull this out into a helper yea

pkg/user/tx_client_test.go
evan-forbes: not blocking, but what would be a way to avoid the test taking 10 minutes to fail?

ninabarbakadze: I can limit the number of loops to 5 for example. It will fail before then.

pkg/user/tx_client_test.go
evan-forbes: would this then be the source a data race in our tests?

pkg/user/tx_client.go
evan-forbes: ideally, it seems like we should never be a scenario where we need this to be cleared. Either a tx is included and we continually try to get it included, or we handle each case when that does not occur and remove the tx at that point

ninabarbakadze: the only time tx doesn't get removed from this local cache is if context gets cancelled while polling for tx status which means that the said tx was never confirmed. If the transaction doesn't get confirmed it can linger in that cache until it ooms

pkg/user/tx_client_test.go
ninabarbakadze: I don't think so because in this test we're not sending so many txs plus other tests using tesnode would initialise the network with the default config overriding this ttl. I'd be happy to reset this back to the default though.

evan-forbes: it seems like we're writing to some state that is continually being read by the mempool, and there are no guarantees that we are not getting a data race, which is less than ideal. It's only in a tests and not prod, but I think the only reason we're not seeing it atm is because we aren't running this test with the race detector

the only way around this is to not change the config after it is passed to testnode, or we have to modify the usage in comet to first check a mutex before accessing the state

pkg/user/tx_client.go
```diff
@@ -468,40 +416,67 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+// the signer's sequence.
+func (client *TxClient) handleEvictions(txHash string) error {
+	client.mtx.Lock()
+	defer client.mtx.Unlock()
+	// Get transaction from the local pool
+	txInfo, exists := client.localMempool[txHash]
+	if !exists {
+		return fmt.Errorf("tx not found in tx client local pool: %s", txHash)
+	}
+	// The sequence should not be incremented if the transaction was evicted
```
cmwaters, line 465:
```suggestion
	// The sequence should not be rolled back to the sequence of the transaction that was evicted to be
	// ready for resubmission. All transactions with a later nonce will be kicked by the nodes tx pool.
```
nit: some more context here would be helpful

pkg/user/tx_client_test.go
cmwaters: Yeah I think if we want this to be done safely, it should be a separate test and we should only set it at the start

cmwaters (commented):
Logically looks good. I just have a small set of nits for making things easier to understand

pkg/user/tx_client_test.go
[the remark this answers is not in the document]
ninabarbakadze: we don't have evictions error type

pkg/user/tx_client_test.go
ninabarbakadze: yeah if i decrease the ttl at initialisation do you suggest that I reinitialise the network once I want to submit another tx which requires me to increase the ttl again?

pkg/user/tx_client.go
evan-forbes: by periodically? do we mean a ttl? imo, the most important thing here is to be able to consistently reason about what is occurring in the local pool.

pkg/user/tx_client_test.go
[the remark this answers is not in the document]
cmwaters: so just an `Rejection` error type? It probably makes sense to also have an eviction error

pkg/user/tx_client_test.go
```diff
@@ -281,3 +313,35 @@ func (suite *TxClientTestSuite) queryCurrentBalance(t *testing.T) int64 {
...
+		WithTendermintConfig(defaultTmConfig).
+		WithFundedAccounts("a", "b", "c").
+		WithAppCreator(testnode.CustomAppCreator("0utia"))
+	ctx, _, _ := testnode.NewNetwork(t, testnodeConfig)
+	_, err := ctx.WaitForHeight(1)
+	require.NoError(t, err)
+	txClient, err := user.SetupTxClient(ctx.GoContext(), ctx.Keyring, ctx.GRPCClient, encCfg, user.WithGasMultiplier(1.2))
+	require.NoError(t, err)
+	return encCfg, txClient, ctx
+}
```
cmwaters, line 347: nit: this doesn't make too much sense as it's own function because it's so specific to one test.

pkg/user/tx_client_test.go
```diff
@@ -281,3 +313,35 @@ func (suite *TxClientTestSuite) queryCurrentBalance(t *testing.T) int64 {
...
+// asserts that a tx was indexed in the tx tracker and that the sequence does not increase
+func assertTxInTxTracker(t *testing.T, txClient *user.TxClient, txHash string, expectedSigner string) {
+	nonce, signer, exists := txClient.GetTxFromTxTracker(txHash)
+	require.True(t, exists)
+	require.Equal(t, expectedSigner, signer)
+	seq := txClient.Signer().Account(expectedSigner).Sequence()
+	// Successfully broadcast transaction increases the nonce
+	// txInfo is indexed before the nonce is increased
+	require.Equal(t, seq, nonce+1)
+}
```
cmwaters, line 331: What do you mean by asserting that it doesn't increase yet expecting the sequence to be 1 ahead of the sequence of the transaction (i.e. it was incremented)?

pkg/user/tx_client.go
```diff
@@ -468,40 +433,68 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+func (client *TxClient) handleEvictions(txHash string) error {
+	client.mtx.Lock()
+	defer client.mtx.Unlock()
+	// Get transaction from the local pool
+	txInfo, exists := client.txTracker[txHash]
+	if !exists {
+		return fmt.Errorf("tx not found in tx client local pool: %s", txHash)
+	}
+	// The sequence should not be rolled back to the sequence of the transaction that was evicted to be
+	// ready for resubmission. All transactions with a later nonce will be kicked by the nodes tx pool.
```
cmwaters, line 483:
```suggestion
	// The sequence should be rolled back to the sequence of the transaction that was evicted to be
	// ready for resubmission. All transactions with a later nonce will be kicked by the nodes tx pool.
```
I think you meant should instead of should not

pkg/user/tx_client.go
```diff
@@ -468,40 +433,68 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+// handleEvictions handles the scenario where a transaction is evicted from the mempool.
+// It removes the evicted transaction from the local tx pool without incrementing
+// the signer's sequence.
+func (client *TxClient) handleEvictions(txHash string) error {
+	client.mtx.Lock()
+	defer client.mtx.Unlock()
+	// Get transaction from the local pool
+	txInfo, exists := client.txTracker[txHash]
+	if !exists {
+		return fmt.Errorf("tx not found in tx client local pool: %s", txHash)
```
cmwaters, line 480: I think this could be a better error. How does this situation happen and what should we be telling the user?

pkg/user/tx_client.go
```diff
@@ -27,20 +26,28 @@ import (
...
 )
 
 type Option func(client *TxClient)
 
+// txInfo is a struct that holds the sequence and the signer of a transaction
+// in the local tx pool.
+type txInfo struct {
+	sequence  uint64
+	signer    string
+	timeStamp time.Time
```
rootulp, line 48: [nit] timestamp is one word. https://en.wikipedia.org/wiki/Timestamp

```suggestion
	timestamp time.Time
```

pkg/user/tx_client.go
```diff
@@ -368,23 +385,20 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
 		broadcastTxErr := &BroadcastTxError{
 			TxHash:   resp.TxResponse.TxHash,
 			Code:     resp.TxResponse.Code,
 			ErrorLog: resp.TxResponse.RawLog,
 		}
-		return resp.TxResponse, broadcastTxErr
+		return nil, broadcastTxErr
+	}
+
+	// save the sequence and signer of the transaction in the local pool
```
rootulp, line 396: IIUC txTracker isn't a mempool, it's a cache of recently seen transactions

```suggestion
	// save the sequence and signer of the transaction in the local txTracker
```

pkg/user/tx_client.go
```diff
@@ -395,62 +409,13 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
 
-// retryBroadcastingTx creates a new transaction by copying over an existing transaction but creates a new signature with the
-// new sequence number. It then calls `broadcastTx` and attempts to submit the transaction
-func (client *TxClient) retryBroadcastingTx(ctx context.Context, txBytes []byte) (*sdktypes.TxResponse, error) {
-	blobTx, isBlobTx, err := blobtx.UnmarshalBlobTx(txBytes)
-	if isBlobTx {
-		// only check the error if the bytes are supposed to be of type blob tx
-		if err != nil {
-			return nil, err
+// pruneTxTracker removes transactions from the local tx pool that are older than 10 minutes
```
rootulp, line 412: nit: elsewhere in this PR it's called a txTracker. Can we use that term consistently instead of tx pool sometimes and txTracker other times.

```suggestion
// pruneTxTracker removes transactions from the local tx tracker that are older than 10 minutes
```

pkg/user/tx_client.go
```diff
@@ -468,40 +433,68 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+			return nil, client.handleEvictions(txHash)
+		default:
+			client.deleteFromTxTracker(txHash)
+			return nil, fmt.Errorf("transaction with hash %s not found; it was likely rejected", txHash)
 		}
 	}
 }
 
+// handleEvictions handles the scenario where a transaction is evicted from the mempool.
+// It removes the evicted transaction from the local tx pool without incrementing
```
rootulp, line 472:
```suggestion
// It removes the evicted transaction from the local tx tracker without incrementing
```

pkg/user/tx_client.go
```diff
@@ -468,40 +433,68 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+// the signer's sequence.
+func (client *TxClient) handleEvictions(txHash string) error {
+	client.mtx.Lock()
+	defer client.mtx.Unlock()
+	// Get transaction from the local pool
```
rootulp, line 477:
```suggestion
	// Get transaction from the local tx tracker
```

pkg/user/tx_client.go
```diff
@@ -468,40 +433,68 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+	txInfo, exists := client.txTracker[txHash]
+	if !exists {
+		return fmt.Errorf("tx: %s not found in tx client local pool; likely failed during broadcast", txHash)
```
rootulp, line 480:
```suggestion
		return fmt.Errorf("tx: %s not found in tx client txTracker; likely failed during broadcast", txHash)
```

pkg/user/tx_client.go
```diff
@@ -468,40 +433,68 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+	// The sequence should be rolled back to the sequence of the transaction that was evicted to be
+	// ready for resubmission. All transactions with a later nonce will be kicked by the nodes tx pool.
+	if err := client.signer.SetSequence(txInfo.signer, txInfo.sequence); err != nil {
+		return fmt.Errorf("setting sequence: %w", err)
+	}
+	delete(client.txTracker, txHash)
+	return fmt.Errorf("tx was evicted from the mempool")
+}
+
+// deleteFromTxTracker safely deletes a transaction from the local tx pool.
```
rootulp, line 491:
```suggestion
// deleteFromTxTracker safely deletes a transaction from the local tx tracker.
```

pkg/user/tx_client.go
```diff
@@ -604,6 +598,14 @@ func (client *TxClient) getAccountNameFromMsgs(msgs []sdktypes.Msg) (string, err
 	return record.Name, nil
 }
 
+// GetTxFromTxTracker gets transaction info from the tx client's local tx pool by its hash
```
rootulp, line 601:
```suggestion
// GetTxFromTxTracker gets transaction info from the tx client's local tx tracker by its hash
```

rootulp (approved):
nit: consistently use the term txTracker instead of txTracker in some places and local pool in other places.

pkg/user/tx_client_test.go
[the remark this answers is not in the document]
cmwaters: There wouldn't be any fields, but now it's easier to type cast it when handling errors

pkg/user/pruning_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestPruningInTxTracker(t *testing.T) {
+	txClient := &TxClient{
+		txTracker: make(map[string]txInfo),
+	}
+	numTransactions := 10
+
+	// Add 10 transactions to the tracker that are 10 minutes old
```
rootulp, line 17: this comment seems incorrect. It looks like the implementation below adds 5 transactions that are 10 minutes old and 5 transactions that are 5 minutes old.

```suggestion
```

rootulp (approved):
> Fixes https://github.com/celestiaorg/celestia-app/issues/3768

The PR description will close that issue when this PR merges which doesn't seem like what you want. Since https://github.com/celestiaorg/celestia-app/issues/3768 is large in scope, it is not immediately obvious which parts of that issue this PR addresses. One idea is to convert that issue into a tracking issue with several smaller issues. Then this PR can close one of the smaller issues and leave the remaining ones open.

cmwaters (approved):

ninabarbakadze:
> > Fixes #3768
> 
> The PR description will close that issue when this PR merges which doesn't seem like what you want. Since #3768 is large in scope, it is not immediately obvious which parts of that issue this PR addresses. One idea is to convert that issue into a tracking issue with several smaller issues. Then this PR can close one of the smaller issues and leave the remaining ones open.

Created a new issue that this pr directly resolves https://github.com/celestiaorg/celestia-app/issues/3899 i decided to keep the old one for context. Me and Callum can groom it/close it/delete it when we pm best txs stuff.

staheri14 (dismissed):
LGTM!

pkg/user/tx_client_test.go
```diff
@@ -174,48 +168,90 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
...
+		confirmTxResp, err := suite.txClient.ConfirmTx(suite.ctx.GoContext(), resp.TxHash)
 		require.Error(t, err)
+		require.Nil(t, confirmTxResp)
 		code := err.(*user.ExecutionError).Code
 		require.NotEqual(t, abci.CodeTypeOK, code)
+		require.True(t, wasRemovedFromTxTracker(resp.TxHash, suite.txClient))
 	})
 }
 
+func TestEvictions(t *testing.T) {
```
evan-forbes, line 228: I think we just need to add this test to the exceptions list to get past the race detector issue with testnode/sdk

ninabarbakadze: resolved now. can I please get another approval? 🙏🏻

rootulp (approved):

evan-forbes (approved):

Issue #3836, opened 2024-08-30
Implement block validity rule to reject ICA transactions with a message not on hard-coded allowlist


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/issues/3824

## Problem

Arabica and Mocha have an incorrect ICA host param for AllowMessages because it is the default [*]. It should be an explicit allowlist.

## Proposal

1. Implement block validity rule to reject ICA transactions with a message not on hard-coded allowlist
2. Release this change as a v2.x.x release

rootulp:
I created an ICA on Cosmos Hub testnet that controls an account on Arabica. When I submit an execute tx on Cosmos, it manifests as `MsgRecvPacket` on the host chain so that's the message type we need to inspect in prepare / process proposal.

See https://github.com/celestiaorg/celestia-app/issues/3792#issuecomment-2322304935

evan-forbes:
closing as unplanned per the most recent sync

instead, the approach being taken is to simply submit the gov proposals and wait

PR #3839, opened 2024-08-31
feat!: reject ICA transactions with a message not on hard-coded allowlist

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3836

This PR rejects blocks that include ICA transactions with a message not on the hard-coded allowlist `icaAllowedMessages()`. This is a temporary solution that we may want until governance votes to modify the allowlist from wildcard to the explicit allowlist. 

This targets the v2.x branch and if merged, we would release it as a minor release (e.g. v2.2.x). The original motivation behind this PR is that we could introduce this change as a "soft-fork" and remove it quickly after governance votes to fix the param.

I would rather not merge this. I think there is more risk from merging + releasing this than there is from having the param be a wildcard for a week. This PR is complex and there isn't a lot of time to let this bake on testnets. Note: this PR invalidates the requirement:

> Requirement 5 [ProcessProposal, determinism-2]: For any two correct processes p and q, and any arbitrary block u, if p’s (resp. q’s) CometBFT calls RequestProcessProposal on u at height h, then p’s Application accepts u if and only if q’s Application accepts u. Note that this requirement follows from Requirement 4 and the Agreement property of consensus.

From [CometBFT docs](https://docs.cometbft.com/v0.38/spec/abci/abci++_app_requirements). In other words for a block with an ICA tx that contains a message not on `icaAllowedMessages()`:
- celestia-app nodes < v2.2.x will accept it
- celestia-app ndoes > v2.2.x will reject it

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -8,7 +8,11 @@ import (
 	"github.com/celestiaorg/celestia-app/v2/pkg/da"
 	"github.com/celestiaorg/go-square/shares"
 	"github.com/celestiaorg/go-square/square"
+	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/telemetry"
+	icahosttypes "github.com/cosmos/ibc-go/v6/modules/apps/27-interchain-accounts/host/types"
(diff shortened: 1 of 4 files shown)


rootulp:
Closing b/c this is too complex. We're instead going to modify the param via governance.

Issue #3850, opened 2024-09-09
Request reviews from all core/app engineers


rootulp:
## Context

Breakfast discussion w/ @cmwaters @ninabarbakadze 

## Proposal

@cmwaters @ninabarbakadze don't like how the auto PR request reviewer action only requests reviews from two people on core/app. 

I personally don't have a strong opinion for or against this.

## Proposal

They would prefer if the PR request reviewer action requested reviews from all core/app engineers.

ninabarbakadze:
thanks! I think @evan-forbes should have the perms for this

evan-forbes:
we just need to change this line

https://github.com/celestiaorg/celestia-app/blob/main/.github%2Fauto_request_review.yml#L18

PR #3856, opened 2024-09-10
ci: request reviews from all celestia-core members

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3850

I'm not sure if this explicitly assigns all celestia-core members as reviewers or just the entire group.


cmwaters (approved):
Based on the comments, it sounds like this should work as expected

staheri14 (approved):

PR #3870, opened 2024-09-12
feat: support v3 app

cmwaters:
Adds v3 as an app version.


app/test/std_sdk_test.go
```diff
@@ -307,7 +307,7 @@ func (s *StandardSDKIntegrationTestSuite) TestStandardSDK() {
 			name: "signal a version change",
 			msgFunc: func() (msgs []sdk.Msg, signer string) {
 				valAccount := s.getValidatorAccount()
-				msg := signal.NewMsgSignalVersion(valAccount, 2)
+				msg := signal.NewMsgSignalVersion(valAccount, 4)
```
ninabarbakadze, line 310:
```suggestion
				msg := signal.NewMsgSignalVersion(valAccount, 3)
```

app/test/upgrade_test.go
```diff
@@ -27,9 +34,118 @@ import (
 	dbm "github.com/tendermint/tm-db"
 )
 
+func TestAppUpgradeV3(t *testing.T) {
```
ninabarbakadze, line 37: more comments would be useful for the reader

app/app.go
```diff
@@ -341,10 +343,10 @@ func New(
 		packetforwardkeeper.DefaultRefundTransferPacketTimeoutTimestamp,  // refund timeout
 	)
 	// PacketForwardMiddleware is used only for version 2.
-	transferStack = module.NewVersionedIBCModule(packetForwardMiddleware, transferStack, v2, v2)
+	transferStack = module.NewVersionedIBCModule(packetForwardMiddleware, transferStack, v2, v3)
```
ninabarbakadze, line 346: should we have some kind of test to avoid forgetting to bump versioned ibc modules?

app/app.go
```diff
@@ -341,10 +343,10 @@ func New(
 		packetforwardkeeper.DefaultRefundTransferPacketTimeoutTimestamp,  // refund timeout
 	)
 	// PacketForwardMiddleware is used only for version 2.
```
rootulp, line 345:
```suggestion
	// PacketForwardMiddleware is used only for version >= 2.
```

app/modules.go
```diff
@@ -341,6 +341,26 @@ func versionedStoreKeys() map[uint64][]string {
 			stakingtypes.StoreKey,
 			upgradetypes.StoreKey,
 		},
+		v3: {
```
rootulp, line 344: [no change needed] verified this is the exact same list as v2 https://www.diffchecker.com/pHyajXyz/

app/prepare_proposal.go
```diff
@@ -47,18 +50,39 @@ func (app *App) PrepareProposal(req abci.RequestPrepareProposal) abci.ResponsePr
...
 	// The txs returned are the ones used in the square and block.
-	dataSquare, txs, err := square.Build(txs,
-		app.MaxEffectiveSquareSize(sdkCtx),
-		appconsts.SubtreeRootThreshold(app.GetBaseApp().AppVersion()),
+	var (
+		dataSquareBytes [][]byte
+		err             error
+		size            uint64
 	)
+	switch app.AppVersion() {
```
rootulp, line 58: should we add a unit test to prepare proposal to verify this logic?

app/process_proposal.go
```diff
@@ -122,24 +125,40 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
...
+	var (
+		dataSquareBytes [][]byte
+		err             error
 	)
-	if err != nil {
-		logInvalidPropBlockError(app.Logger(), req.Header, "failure to compute data square from transactions:", err)
+
+	switch app.AppVersion() {
+	case v3:
+		var dataSquare squarev2.Square
```
rootulp, line 135: should we add a unit test to process proposal to verify this logic?

app/test/std_sdk_test.go
rootulp: [blocking] 4 seems incorrect b/c that version isn't supported by the state machine? Why not use 3 here like Nina suggests.

app/test/upgrade_test.go
```diff
@@ -27,9 +34,118 @@ import (
...
+
+	deliverTxResp = testApp.DeliverTx(abci.RequestDeliverTx{
+		Tx: blobTx.Tx,
+	})
+	require.Equal(t, abci.CodeTypeOK, deliverTxResp.Code, deliverTxResp.Log)
+
+	_ = testApp.EndBlock(abci.RequestEndBlock{})
+}
+
 // TestAppUpgrades verifies that the all module's params are overridden during an
```
rootulp, line 146:
```suggestion
// TestAppUpgradeV2 verifies that the all module's params are overridden during an
```

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cosmossdk.io/errors v1.0.1
 	cosmossdk.io/math v1.3.0
 	github.com/celestiaorg/blobstream-contracts/v3 v3.1.0
+	github.com/celestiaorg/go-square v1.0.0
 	github.com/celestiaorg/go-square/v2 v2.0.0-rc2
```
rootulp, line 10: [separate PR] should we create an official release of go-square so that it can be used here?

rootulp (commented):
Overall LGTM. One blocking comment

app/test/std_sdk_test.go
cmwaters: Because 3 is the current version. It would be weird to signal for the current version. What is more normal is to signal for the next version

cmwaters: I could do something like `LatestVersion + 1` so we don't have to continually increment it

go.mod
cmwaters: Yes, I noticed this as well

cmwaters: Everything seems to be passing so it's probably safe to cut the final release

app/process_proposal.go
cmwaters: What logic would this test be verifying? Do we want to build a v3 square with an authored blob and assert that we can parse out the signer or something like that?

app/test/std_sdk_test.go
rootulp: Oh I see, thanks! No longer blocking.

4 or `LatestVersion + 1` both work for me.

rootulp (approved):

rootulp (approved):
LGTM

ninabarbakadze (dismissed):

ninabarbakadze (approved):

rootulp (approved):

PR #3879, opened 2024-09-16
test: consistent app hash on v2

### LINKED ISSUE #3624: AppHash mismatch test between main and v2

ninabarbakadze:
## Overview

Fixes #3624


app/test/consistent_apphash_test.go
```diff
@@ -35,66 +39,119 @@ import (
...
+			encodedBlobTxs:     createEncodedBlobTx,
+			expectedDataRoot:   []byte{100, 59, 112, 241, 238, 49, 50, 64, 105, 90, 209, 211, 49, 254, 211, 83, 133, 88, 5, 89, 221, 116, 141, 72, 33, 110, 16, 78, 5, 48, 118, 72},
+			// Expected app hash produced by v1.x - https://github.com/celestiaorg/celestia-app/blob/v1.x/app/consistent_apphash_test.go
+			expectedAppHash: []byte{84, 216, 210, 48, 113, 204, 234, 21, 150, 236, 97, 87, 242, 184, 45, 248, 116, 127, 49, 88, 134, 197, 202, 125, 44, 210, 67, 144, 107, 51, 145, 65},
+		},
+		{
+			name:    "execute sdk messages and blob tx on v2",
+			version: v2.Version,
+			encodedSdkMessages: func(t *testing.T, accountAddresses []sdk.AccAddress, genValidators []stakingtypes.Validator, testApp *app.App, signer *user.Signer, valSigner *user.Signer) ([][]byte, [][]byte, [][]byte) {
+				firstBlockEncodedTxs, secondBlockEncodedTxs, thirdBlockEncodedTxs := encodedSdkMessagesV1(t, accountAddresses, genValidators, testApp, signer, valSigner)
```
rootulp, line 80: v2 removed x/blobstream so I think the blobstream transactions need to be removed the test case on v2. Alternatively if they aren't removed, maybe this test case should verify that x/blobstream messages are no-ops on v2.

rootulp (approved):

app/test/consistent_apphash_test.go
ninabarbakadze: I think those messages are supposed to be rejected on v2. I'll investigate.

cmwaters: It looks like the blobstream messages were never part of the v1 test suite anyway so I think doing it like this is fine

cmwaters (commented):
Looks good. Probably a good idea to add ICA and IBC messages as well to expand the surface area of the tests

app/test/consistent_apphash_test.go
```diff
@@ -388,7 +440,7 @@ func executeTxs(testApp *app.App, encodedBlobTx []byte, encodedSdkTxs [][]byte,
 	dataHash := resPrepareProposal.BlockData.Hash
 
 	header := tmproto.Header{
-		Version:        version.Consensus{App: 1},
+		Version:        version.Consensus{App: appVersion},
```
cmwaters, line 443:
```suggestion
		Version:        version.Consensus{App: testApp.AppVersion()},
```
You could just do this which avoids needing the extra argument

cmwaters (commented):
Don't know why this was not included in my original review 🤷

ninabarbakadze:
> Looks good. Probably a good idea to add ICA and IBC messages as well to expand the surface area of the tests

I'm extending the test to include all IBC messages for all verisons

ninabarbakadze:
> > Looks good. Probably a good idea to add ICA and IBC messages as well to expand the surface area of the tests
> 
> I'm extending the test to include all IBC messages for all verisons

After looking into adding IBC messages I realised that it would require me to add more test chains, set up relayers, etc. I think it's best if we have a separate test for executing IBC messages and comparing AppHash between versions. Thoughts? @cmwaters @rootulp

cmwaters:
Don't we already have tests where we send IBC messages (i.e. ICA and Transfer) can't we just add the app hash test vectors within those tests?

cmwaters (approved):
LGTM

ninabarbakadze:
> Don't we already have tests where we send IBC messages (i.e. ICA and Transfer) can't we just add the app hash test vectors within those tests?

yeah we could. I'll open an issue for it.

app/test/consistent_apphash_test.go
rootulp: If the v1 test doesn't have blobstream messages, is it missing any other messages? Do we have a GH issue to make this test exhaustive of all message types supported by each app version?

rootulp (approved):
I'm okay with expanding this test or adding a new test to expand message types that are supported (including IBC message types).

app/test/consistent_apphash_test.go
ninabarbakadze: it covers most of it now. the only message we don't execute is `MsgSubmitEvidence` and IBC messages that require a different setup. Initially, we wanted to cover as much as possible and we're now extending to cover IBC messages too.

Issue #3859, opened 2024-09-11
Override the `timeout_commit` and `timeout_propose` when switching to v3


evan-forbes:
We should override the timeouts upon the switch to app version v3

rootulp:
Notes from discussion w/ @evan-forbes :

- This issue diverges from ADR-115 because we want to override two params not just one (as proposed in ADR-115).
- This is blocked on experiments for determining what the values are.
    - @evan-forbes hypothesizes that TimeoutCommit = ~4.7 seconds and TimeoutProposal = ~3.5 seconds.
    - @evan-forbes teach @rootulp how to run experiment to determine the correct timeouts. Blocked on @evan-forbes syncing with @sysrex because he's currently using his personal accounts for Congest.

staheri14:
**Question:**
At height X-1 (before upgrade height X), should we suggest timeouts for the current app version (v2) or the upcoming version (v3)?

### **Option 1: Use Timeouts for Next Version (v3)**  
- **Pros:**
  - Fast block production starts immediately at upgrade height (X).
- **Cons:**
  - Inconsistency: Suggested timeouts at height X-1 don’t match the app version in the current block header (v2 vs. v3).

### **Option 2: Use Timeouts for Current Version (v2)**  
- **Pros:**
  - Consistency: Timeouts always align with the app version in the current block header.
- **Cons:**
  - Delayed effect: Timeouts based on v3 will only apply at height X+1 (one height after the upgrade).

evan-forbes:
good question, my understanding of this is that we should only change the timeouts when the app version is changed, so we should do option 2

evan-forbes:
to add more reasoning to the above, we don't know if the upgrade is successful until we produce a block that has that version, therefore that is the soonest we possibly could change.

staheri14:
Thanks, @evan-forbes! That makes sense—I'll go ahead with the second option.

evan-forbes:
either or should be okay if I'm understanding correctly, especially for our existing testnets

we could also just use a default timeouts for the first block, as it shouldn't be super important for us atm

PR #3882, opened 2024-09-17
feat!: versioned timeouts

staheri14:
Closes https://github.com/celestiaorg/celestia-app/issues/3859

manually tested in https://github.com/celestiaorg/celestia-app/pull/3882#issuecomment-2415280773

also tested in knuu


pkg/appconsts/global_consts.go
```diff
@@ -30,7 +30,7 @@ const (
 	// DefaultUpgradeHeightDelay is the number of blocks after a quorum has been
 	// reached that the chain should upgrade to the new version. Assuming a block
 	// interval of 12 seconds, this is 7 days.
-	DefaultUpgradeHeightDelay = int64(7 * 24 * 60 * 60 / 12) // 7 days * 24 hours * 60 minutes * 60 seconds / 12 seconds per block = 50,400 blocks.
+	DefaultUpgradeHeightDelay = 3 // int64( 7 * 24 * 60 * 60 / 12) // 7 days * 24 hours * 60 minutes * 60 seconds / 12 seconds per block = 50,400 blocks.
```
jcstein, line 33:
```suggestion
	DefaultUpgradeHeightDelay = 3 // int64( 7 * 24 * 60 * 60 / 6) // 7 days * 24 hours * 60 minutes * 60 seconds / 6 seconds per block = 100,800 blocks.
```

app/default_overrides.go
```diff
@@ -274,8 +274,8 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	cfg.Mempool.MaxTxsBytes = int64(upperBoundBytes) * cfg.Mempool.TTLNumBlocks
```
jcstein, line 274: Should line 146 have this?
```golang
	// time. The expected block time is 6 seconds, therefore this value is 30
```

jcstein:
Should `GoalBlockTime = time.Second * 6` on `pkg/appconsts/consensus_consts.go`?

Linking discussion from my PR which I plan to close: https://github.com/celestiaorg/celestia-app/pull/3953/files#r1794435786

scripts/single-node.sh
```diff
@@ -4,7 +4,7 @@
 
 # Stop script execution if an error is encountered
 set -o errexit
-# Stop script execution if an undefined variable is used
+# Stop scyript execution if an undefined variable is used
```
jcstein, line 7:
```suggestion
# Stop script execution if an undefined variable is used
```

app/test/timeouts_test.go
```diff
@@ -0,0 +1,92 @@
...
+type TimeoutTestSuite struct {
+	suite.Suite
+
+	ecfg     encoding.Config
+	accounts []string
+	cctx     testnode.Context
+}
+
+const (
+	timeoutPropose = 7 * time.Second
```
jcstein, line 36: Should these be the same as below? in pkg/appconsts/v3/app_consts.go

Screenshot 2024-10-10 at 15 39 46

staheri14: No, they shouldn't, the reason is that these are the initial values configured for a node, intentionally set to differ from the default versioned timeouts to verify that nodes derive timeouts from the app version rather than the config settings.

jcstein: nice, thank you for explaining!

pkg/appconsts/global_consts.go
staheri14: [To reviewers] This change isn't intended for merging in this PR. It's a temporary workaround to address [this issue](https://github.com/celestiaorg/celestia-app/issues/3947#issuecomment-2401175659). I reduced the delay to ensure the e2e tests finish in a reasonable time—otherwise, we would have to wait for 50,400 blocks before the upgrade tests complete.

go.mod
```diff
@@ -260,5 +260,5 @@ replace (
 	github.com/cosmos/ledger-cosmos-go => github.com/cosmos/ledger-cosmos-go v0.12.4
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.42.0-tm-v0.34.35
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.42.0-tm-v0.34.35.0.20241011205050-363058a623c9
```
staheri14, line 263: This needs to point to a version of core after this [PR](https://github.com/celestiaorg/celestia-core/pull/1494) is merged.

test/e2e/timeouts.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/v3/test/e2e/testnet"
+	"github.com/celestiaorg/celestia-app/v3/test/util/testnode"
+)
+
+// VersionedTimeouts runs a simple testnet with 4 validators. It submits both MsgPayForBlobs
+// and MsgSends over 30 seconds and then asserts that at least 10 transactions were
+// committed.
+func VersionedTimeouts(logger *log.Logger) error {
```
staheri14, line 17: [To reviewers] This test is still in progress.

pkg/appconsts/v3/app_consts.go
```diff
@@ -1,10 +1,14 @@
...
+
 const (
 	Version              uint64 = 3
 	SquareSizeUpperBound int    = 128
 	SubtreeRootThreshold int    = 64
 	TxSizeCostPerByte    uint64 = 10
 	GasPerBlobByte       uint32 = 8
 	MaxTxSize            int    = 2097152 // 2 MiB in bytes
+	TimeoutPropose              = time.Second * 3
+	TimeoutCommit               = time.Second * 1
```
staheri14, line 13: [To the reviewers] We may need further research to properly set these values.

evan-forbes: By default, we should use the timeouts from the 8MB 6s chain, so 4.2s timeout commit and 3.5s timeout propose. This gets us 5-7s blocks for a 0-8MB block

If we want closer to 6s second blocks we can adjust for testnets and max block size

test/e2e/benchmark/throughput.go
```diff
@@ -139,6 +140,17 @@ func TwoNodeBigBlock8MB(logger *log.Logger) error {
...
+func TwoNodeBigBlock8MBV3(logger *log.Logger) error {
+	manifest := bigBlockManifest
+	manifest.MaxBlockBytes = 8 * testnet.MB
+	// HACKHACK: use a version of celestia-app built from a commit on this PR.
+	// This can be removed after the PR is merged to main.
+	manifest.CelestiaAppVersion = "pr-3882"
+	manifest.TxClientVersion = "pr-3882"
+	manifest.GenesisAppVersion = v3.Version
+	return runBenchmarkTest(logger, "TwoNodeBigBlock8MBV3", manifest)
+}
```
evan-forbes, line 152: @staheri14 is this just due this PR being created before the build tags? we can remove this, correct?

test/e2e/major_upgrade_v3.go
```diff
@@ -97,12 +99,51 @@ func MajorUpgradeToV3(logger *log.Logger) error {
...
+	startDur := blockTimes[0]
+	endDur := blockTimes[len(blockTimes)-1]
+
+	if startDur < time.Second*10 {
+		testnet.NoError("", fmt.Errorf("blocks for v2 are too short %v", len(blockTimes)))
+	}
+
+	if endDur > time.Second*7 {
+		testnet.NoError("", fmt.Errorf("blocks for v3 are too long %v", len(blockTimes)))
+	}
```
evan-forbes, line 146: We need to change this to be more robust against flakes by accounting for rounds with more than 1 round.

test/e2e/timeouts.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/v3/test/e2e/testnet"
+	"github.com/celestiaorg/celestia-app/v3/test/util/testnode"
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+// VersionedTimeouts runs a simple testnet with 4 validators. It submits both MsgPayForBlobs
+// and MsgSends over 30 seconds and then asserts that at least 10 transactions were
+// committed.
+func VersionedTimeouts(logger *log.Logger) error {
```
evan-forbes, line 18: @staheri14 do we need this test with the modifications to the upgrade test? this seems like we don't need to check that the times were changed twice

test/util/testnode/app_wrapper.go
```diff
@@ -0,0 +1,19 @@
...
+
+func WrapEndBlocker(app *app.App, timeoutCommit time.Duration) func(ctx sdk.Context, req abci.RequestEndBlock) abci.ResponseEndBlock {
+	endBlocker := func(ctx sdk.Context, req abci.RequestEndBlock) abci.ResponseEndBlock {
+		resp := app.EndBlocker(ctx, req)
+		resp.Timeouts.TimeoutCommit = timeoutCommit
+		return resp
+	}
+
+	return endBlocker
+}
```
evan-forbes, line 19: :innocent: 

hehe testnode still goes fast

test/e2e/benchmark/throughput.go
cmwaters, line 152: Are we adding these benchmarks to see if there's a performance regression? Don't the existing benchmarks use the latest version?

Just trying to understand the point of this.

test/e2e/timeouts.go
cmwaters: I agree that extending the major_upgrade_v3 should suffice

test/txsim/run_test.go
```diff
@@ -170,6 +170,8 @@ func TestTxSimUpgrade(t *testing.T) {
 		WithFundedAccounts("txsim-master")
 	cctx, _, grpcAddr := testnode.NewNetwork(t, cfg)
 
+	time.Sleep(time.Second * 2)
```
cmwaters, line 173: Why was this sleep added. Could we instead just use the cctx to wait for the first block to be produced?

test/util/testnode/app_wrapper.go
cmwaters: lol sneaky

test/util/testnode/read.go
```diff
@@ -96,6 +99,67 @@ func reverseSlice[T any](s []T) {
...
+
+	var result []HeightTime
+	for i := 0; i < len(blockMetas)-1; i++ {
+		// the time difference between the current header and the next one
+		timeDiff := blockMetas[i+1].Header.Time.Sub(blockMetas[i].Header.Time)
+		// shift the time difference to the next height
+		result = append(result, HeightTime{Height: blockMetas[i+1].Header.Height, Time: timeDiff})
+	}
+	return result
+}
```
cmwaters, line 121: This stuff is probably duplicated with the blocktime tool that I wrote but we can come back and clean that up later

cmwaters (changes requested):
Need to revert back to the 7 days. Most of it looks good though

test/e2e/benchmark/throughput.go
evan-forbes: agree, I like the idea of testing this and checking for basic metrics, and I'm guessing the purpose was to sanity check the timeout change with full blocks manually and then remove later so I'm removing these in favor of #3965

test/util/testnode/read.go
evan-forbes: moved to a gist on this comment for now https://github.com/celestiaorg/celestia-app/issues/3965#issuecomment-2413746001

rootulp (commented):
1. Isn't this technically breaking so we should add `!` in the PR title.
2. [Can be follow up PR] should we add a line item to https://github.com/celestiaorg/celestia-app/blob/main/docs/release-notes/pending-release.md
3. How was this tested? When I run `./scripts/single-node.sh` blocks are produced on average every 11 seconds. Might be related to https://github.com/celestiaorg/celestia-app/issues/3975

test/e2e/benchmark/throughput.go
staheri14: Are you referring to the version "pr-3882" used for validator? it points to the celestia-app image built based on the current PR, and the reason we had this test was to ensure we can produce 8MB blocks with the new versioned timeouts, as in e2e benchmark tests, we measure the size of blocks when the test is concluded.

test/util/testnode/config.go
```diff
@@ -184,13 +184,15 @@ func DefaultAppCreator() srvtypes.AppCreator {
 			simapp.EmptyAppOptions{},
 			baseapp.SetMinGasPrices(fmt.Sprintf("%v%v", appconsts.DefaultMinGasPrice, app.BondDenom)),
 		)
+		app.SetEndBlocker(wrapEndBlocker(app, time.Millisecond*0))
```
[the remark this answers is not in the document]
evan-forbes: all testnode tests previously had blocks with timeout commits equal to zero (or something very small), so changing this here is assuming the opposite, that testnode tests are keeping this by default, but if they want to test the end blocker, then they use the normal one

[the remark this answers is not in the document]
rootulp: Ahh got it, thanks for explaining. Resolving.

test/e2e/major_upgrade_v3.go
```diff
@@ -97,12 +99,50 @@ func MajorUpgradeToV3(logger *log.Logger) error {
...
+
+	if len(blockTimes) < 7 {
+		testnet.NoError("", fmt.Errorf("not enough block times collected: %v", len(blockTimes)))
+	}
+
+	startDur := blockTimes[0]
+	endDur := blockTimes[len(blockTimes)-1]
+
+	if startDur < time.Second*10 {
```
staheri14, line 139: afaik, the `upgradeHeightV3` is not the exact height at which upgrade to v3 happens, rather A height at which we saw version v3 in the block header (the transition could have happened earlier as well), this means  block at height `upgradeHeightV3 - 4` is not necessarily a v2 block. If we want to know the exact height at which transition has happened, we need to iterate over all the headers, and find that height.

test/e2e/timeouts.go
staheri14: In the `VersionedTimeouts`, in addition to verifying the overall block time, I was working  on adding a scenario in which we could start one of the nodes later (the reason we have  `indices`  parameter in the methods like `StartNodes` is to control which nodes to start first) and this way we trigger fast sync reactor, and verify if it works properly (however such a test would have been possible with the rpc endpoint which is now deleted).

test/e2e/major_upgrade_v3.go
staheri14: I notice that the ticker is set to 10s, which is half the block time of a v2 block. Therefore, it is likely that upgradeHeightV3 - 4 has a v2 app version in its header. However, to ensure clarity and make the test more readable, I suggest explicitly checking the app version in the loop where block times are read.

evan-forbes:
the knuu cluster is on the fritz, and @rach-id 's local devnet tool comes to the rescue again! here are the timestamps from the logs

```
15D71B38D250F6D2CB0ADD7E711A8658FD63DA993EE4939E77","total":1}},"height":49,"pol_round":-1,"round":0,"signature":"AVp2K3YL3zxlwbROldfOCbLalitHc5DNU9Q7kZzfFxMZMoy30IsCq3tXaXpCs42GakLe2yWhGtsmFQEGpbzhCg==","timestamp":"2024-10-15T22:21:01.230269104Z"} proposer=12C509ED04EA103C84944017B9E7ADE494D486A3
10:21PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"5A9EDAE7C7F68A36D4D4604693F997C00DFC399D007AC8DDEA48C6E0BC649756","parts":{"hash":"DD427A879508A4D6D3EB497071F437AC0938EB0E5DFB78D1569A5D2CFFA149A4","total":1}},"height":50,"pol_round":-1,"round":0,"signature":"3BzjkhFr/sVGlzWc1yfQyFH1fC3McGKBFdJV00pYG+cyviBjLvCoS6CAqCA7SmVJKlcSshzL+7hPCd98I24FBA==","timestamp":"2024-10-15T22:21:12.783355131Z"} proposer=223467FF1BD29B323D8CF4AFCFECD5ED1C0D4383
10:21PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"ED5731ADB3F97BD54515C3C0B5CBE0E1F95AE79312DDEFC3DA98AC10B32581A1","parts":{"hash":"CDD6126E7D55BA813BAA811637355CC57037E74B278AA453FAAF681EC998E2AC","total":1}},"height":51,"pol_round":-1,"round":0,"signature":"tIyE8v9HPbuA/ZRVJMobBiC1WwCU62aSkKlgmbT2NfhJ26AZxu6RXx8XctuhwU0ompYVOdRoQWGHh2KuMrAJBw==","timestamp":"2024-10-15T22:21:24.202601875Z"} proposer=BC381AF58CB970ACAF795A003461EE02057C5F92
10:21PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"5A95DDAEE62A7DAAA2997E604D513A6B7063C332E8138C2908B6D705DF0B8E11","parts":{"hash":"1580CCD32169396E8DFBB2AD55C3D50A111ADBE8F84E586990A40D723A64A9D0","total":1}},"height":52,"pol_round":-1,"round":0,"signature":"VC1F7OHMIaLxAcwsgjLRUgUNYdNWFPQBjHwDdnJFQvW0V9Njd3YI3XN1KoriwXA42aqbt3xmOCKLAqNqBkyLDQ==","timestamp":"2024-10-15T22:21:35.613200507Z"} proposer=E0B63563D528BDA8AA3D26A703FC8EAAD1556016
10:21PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"E6325454000B16A6CB1963B76F0AF12ED03FE89AC47A8B677C140AD432416862","parts":{"hash":"A72AF13D9118850C4B5904CDABAD78022FAE2B8554E503915493C8F51C84B9C5","total":1}},"height":53,"pol_round":-1,"round":0,"signature":"+sXsU+NF1iNEvaJElb8ax9nXPHh951aVeLNr4hzvCt2TUZKg60HZtvLEBXoFAiKnlM/n3KO+bu/xrmc75A9lAQ==","timestamp":"2024-10-15T22:21:47.032329431Z"} proposer=12C509ED04EA103C84944017B9E7ADE494D486A3
10:21PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"960877E0053D87126B3D3E2991066F1D9D9BA1110CFC811D2C5E9A0864A71F91","parts":{"hash":"7F7D31404FF5AF5C3F52BDE7E769CDE6B00C498EDA7A02FDA0C8DAB9D9AB307B","total":1}},"height":54,"pol_round":-1,"round":0,"signature":"x10mNLY9SZMYTTEpAait+9PZrxW08/k/t2begQiPX9jrHkRlNnRjBVDwYeDQEx655iMTT6xDB3IgDhKhnEadBQ==","timestamp":"2024-10-15T22:21:51.747512718Z"} proposer=223467FF1BD29B323D8CF4AFCFECD5ED1C0D4383
10:21PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"6BF50B6FA84A6CEB80F48E6EBE2344EDAE394B14802D1A5F3F9F3DD13D364540","parts":{"hash":"AF1A821BF5CEA6DC96BCECF65678E07444196050F8034D7336DB59C36179E9FE","total":1}},"height":55,"pol_round":-1,"round":0,"signature":"+v33QJhJ4rJYn/JKzmarV3PZU1MHpKKDWUTO6K5YMNC6Ne5XJJbaMX89lQK8tK39fffP7JnYocSVn28tDSWhCg==","timestamp":"2024-10-15T22:21:56.356056546Z"} proposer=BC381AF58CB970ACAF795A003461EE02057C5F92
10:22PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"AB963F8637A4183BAB94A1AA5016FFC6DE7550D7118928C91C34FE919F412578","parts":{"hash":"964C27E0F75BD14570B1EE2433712B6048667E92A5BECDCB5F407A156814E398","total":1}},"height":56,"pol_round":-1,"round":0,"signature":"T850ryJPbgrtGeDmrHkFtUVopAb+OBSEFiWZO5uueXnbzi7R/U287447BBZH/9wEp+5+UGHPvWKt6t8vROIpBQ==","timestamp":"2024-10-15T22:22:00.964394091Z"} proposer=E0B63563D528BDA8AA3D26A703FC8EAAD1556016
10:22PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"2760E8C5A57C9D9C5BC87ED2D4BCB9F81C2065B3560849FBE6DD7F3EF7C7A570","parts":{"hash":"4C0EF005E86D3274B75353C0B37099B9DCE6E0013BE412ADCDEABC7E4A4E23C5","total":1}},"height":57,"pol_round":-1,"round":0,"signature":"opOXXP6M2PMhZ1pNXniOaJUKinCMJ/dnmhoDIHJpoPxrURvH4TToEWYg3bYtpyqdaGTmPXeu1kDvEwVUqjjTCw==","timestamp":"2024-10-15T22:22:05.484655108Z"} proposer=12C509ED04EA103C84944017B9E7ADE494D486A3
10:22PM INF received proposal module=consensus proposal={"Type":32,"block_id":{"hash":"27861014C6F14575F1C75DCE134BBE7A1B4D3B56885207A78992EA51C2454026","parts":{"hash":"82C556A57098F168F0E71E3A7D9361B654BE2182F0D33610D3592B2D2E2E666C","total":1}},"height":58,"pol_round":-1,"round":0,"signature":"o3Bcw/bfS79fTT3AiWgTBBDuBdhTOAdB42Iut2UKgXwHvIprn7EDiy75l1JqWf7WwxMBCf76VBmYGRHic8VhBA==","timestamp":"2024-10-15T22:22:10.202010713Z"} proposer=223467FF1BD29B323D8CF4AFCFECD5ED1C0D4383
```

the upgrade works as expected and the timeouts are overwritten. therefore, we can create a release in core and merge.

go.mod
```diff
@@ -260,5 +260,5 @@ replace (
 	github.com/cosmos/ledger-cosmos-go => github.com/cosmos/ledger-cosmos-go v0.12.4
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.42.0-tm-v0.34.35
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.42.0-tm-v0.34.35.0.20241014174725-23e3fb130177
```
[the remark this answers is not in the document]
evan-forbes: apologies, I kept waiting on the knuu cluster which was working earlier, but konked out on the last run

https://github.com/celestiaorg/celestia-app/pull/3882#issuecomment-2415280773

creating a release now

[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/3882/commits/fc6a897a4b707b96f829f9c617960711e35eb8a7

test/e2e/major_upgrade_v3.go
evan-forbes: :+1: 

updated in

https://github.com/celestiaorg/celestia-app/pull/3882/commits/8c019af12a85885553a57016969713aa70a47c5e

test/txsim/run_test.go
evan-forbes: forget which commit this was fixed, but it was fixed. sometimes it still flakes

pkg/appconsts/versioned_consts.go
```diff
@@ -52,3 +55,25 @@ var (
...
+func GetTimeoutPropose(v uint64) time.Duration {
+	switch v {
+	case v1.Version:
+		return TimeoutPropose
+	case v2.Version:
+		return TimeoutPropose
+	default:
+		return v3.TimeoutPropose
+	}
+}
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/3882/commits/955a6584636b88fe6fcae9ad0e3e331796d2489f

pkg/appconsts/v1/app_consts.go
```diff
@@ -1,7 +1,11 @@
 package v1
 
+import "time"
+
 const (
 	Version              uint64 = 1
 	SquareSizeUpperBound int    = 128
 	SubtreeRootThreshold int    = 64
+	TimeoutPropose              = time.Millisecond * 10000
+	TimeoutCommit               = time.Millisecond * 11000
```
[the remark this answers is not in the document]
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/3882/commits/955a6584636b88fe6fcae9ad0e3e331796d2489f

app/app.go
```diff
@@ -535,11 +537,15 @@ func (app *App) Info(req abci.RequestInfo) abci.ResponseInfo {
 	if resp.AppVersion > 0 && !app.IsSealed() {
 		app.mountKeysAndInit(resp.AppVersion)
 	}
+
+	resp.Timeouts.TimeoutPropose = appconsts.GetTimeoutPropose(resp.AppVersion)
+	resp.Timeouts.TimeoutCommit = appconsts.GetTimeoutCommit(resp.AppVersion)
```
cmwaters, line 542: Just FYI: If this is called the first time a node starts up. The version may be 0

test/e2e/benchmark/manifest.go
```diff
@@ -80,6 +80,8 @@ type Manifest struct {
 	GovMaxSquareSize int64
 
 	DisableBBR bool
+
+	GenesisAppVersion uint64
```
cmwaters, line 84: What's this for?

test/e2e/major_upgrade_v3.go
```diff
@@ -97,12 +99,54 @@ func MajorUpgradeToV3(logger *log.Logger) error {
 				testnet.NoError("failed to get header", err)
 				if resp.Header.Version.App == v3.Version {
 					upgradeComplete = true
+					if upgradedHeight == 0 {
+						upgradedHeight = resp.Header.Height
+					}
```
cmwaters, line 104: If you wanted to be extra sure you could make the assertion that no other node has a different upgrade height (i.e. all nodes did actually upgrade at the same height)

test/e2e/major_upgrade_v3.go
```diff
@@ -97,12 +99,54 @@ func MajorUpgradeToV3(logger *log.Logger) error {
...
+	startDur := blockTimes[0]
+	endDur := blockTimes[len(blockTimes)-1]
+
+	if startDur < time.Second*10 {
+		testnet.NoError("", fmt.Errorf("blocks for v2 are too short %v", len(blockTimes)))
+	}
+
+	if endDur > time.Second*7 {
+		testnet.NoError("", fmt.Errorf("blocks for v3 are too long %v", len(blockTimes)))
+	}
```
cmwaters, line 149: So if I understand correctly, this just picks the duration of the block 5 before the signal height and 1 after the upgrade height and asserts the first is greater than 10 seconds and the last is less than 7 seconds.

I think it should check all the heights up to the actual upgrade height (i.e. during the delay period) to be sure that it still remains at the expected times

x/blobstream/integration_test.go
```diff
@@ -46,6 +46,8 @@ func (s *BlobstreamIntegrationSuite) SetupSuite() {
 	cctx, _, _ := testnode.NewNetwork(t, cfg)
 	s.ecfg = encoding.MakeConfig(app.ModuleEncodingRegisters...)
 	s.cctx = cctx
+
+	require.NoError(t, s.cctx.WaitForBlocks(10))
```
cmwaters, line 50: Why was this added?

test/util/testnode/node_interaction_api.go
```diff
@@ -50,6 +52,10 @@ func NewContext(goContext context.Context, keyring keyring.Keyring, tmConfig *tm
 	return Context{goContext: goContext, Context: clientContext, apiAddress: apiAddress}
 }
 
+func (c *Context) GetTMNode() *node.Node {
+	return c.tmNode
+}
+
```
cmwaters, line 58: Why was this added? It doesn't look used

test/e2e/testnet/testnet.go
```diff
@@ -437,20 +482,20 @@ func (t *Testnet) StartNodes(ctx context.Context) error {
...
 	log.Info().Msg("waiting for genesis nodes to sync")
-	err = t.WaitToSync(ctx)
+	err = t.WaitToSync(ctx, indices...)
 	if err != nil {
 		return err
 	}
 
-	return t.StartTxClients(ctx)
+	return t.StartTxClients(ctx, indices...)
 }
```
cmwaters, line 499: Why was this added?

cmwaters (commented):
It looks like a bunch of stuff was added to the e2e testnet files (something to do with indices) which does not appear to be used. Generally, it seems like there's a bunch of stuff in this PR which seems legacy and doesn't need to be in the PR

test/e2e/benchmark/manifest.go
evan-forbes: looks like just to set the app version in the genesis

cc @staheri14

x/blobstream/integration_test.go
evan-forbes: it kept flaking

test/util/testnode/node_interaction_api.go
evan-forbes: not sure, removed in 6c70dac7ce849046c8e7dbbb32f68614ef523cb9

test/e2e/testnet/testnet.go
evan-forbes: not sure, looks like a sanity check https://github.com/celestiaorg/celestia-app/blob/6c70dac7ce849046c8e7dbbb32f68614ef523cb9/test/e2e/testnet/testnet.go#L428-L435

cc @staheri14 

went ahead and removed in 61a747baeff12a779ac54d5f8d8fb50ebc57e398

test/e2e/major_upgrade_v3.go
evan-forbes: refactored in 07c1fc98836d040ed48a9ae84e76969def2ab656

cmwaters (approved):
LGTM

Can we just create a follow up issue to fix the v3 upgrade test such that we don't need a fixed commit tag

evan-forbes:
> Can we just create a follow up issue to fix the v3 upgrade test such that we don't need a fixed commit tag

ahh good catch https://github.com/celestiaorg/celestia-app/issues/3981

test/e2e/major_upgrade_v3.go
```diff
@@ -97,12 +101,93 @@ func MajorUpgradeToV3(logger *log.Logger) error {
...
+
+		if b.dur < appconsts.GetTimeoutCommit(b.block.Version.App) {
+			return fmt.Errorf(
+				"block was too fast for corresponding version: version %v duration %v upgrade height %v height %v",
+				b.block.Version.App,
+				b.dur,
+				preciseUpgradeHeight,
+				b.block.Height,
+			)
+		}
```
rootulp, line 184: [question] it looks like this will throw an error if b.dur doesn't exactly match the timeout commit which seems extremely fragile. I would expect this test to pass if blocks times are ~6 seconds which is > the timeout commit of 3.5 seconds. Maybe I'm missing something.

rootulp (approved):
[question] now that `TimeoutPropose` and `TimeoutCommit` are versioned constants, should they be removed from consensus constants [here](https://github.com/rootulp/celestia-app/blob/b1f43f5ed19d2fcf2cfb199ebca8988c6a272476/pkg/appconsts/consensus_consts.go)?

cmwaters:
> [question] now that TimeoutPropose and TimeoutCommit are versioned constants, should they be removed from consensus constants [here](https://github.com/rootulp/celestia-app/blob/b1f43f5ed19d2fcf2cfb199ebca8988c6a272476/pkg/appconsts/consensus_consts.go)?

Yeah they probably can be

test/e2e/major_upgrade_v3.go
evan-forbes: ahh whoops yeah good point, will fix in a flup

knuu isn't working still so this test was only manually performed and this wasn't hit

it needs to be 
```go
if b.dur > appconsts.GetTimeoutCommit([the other app version]) {
}
```

evan-forbes: https://github.com/celestiaorg/celestia-app/pull/3882/commits/9669f8c996b445637132231bff70ed1d53c0f6ab

pkg/appconsts/v3/app_consts.go
```diff
@@ -1,10 +1,18 @@
...
 	SquareSizeUpperBound int    = 128
 	SubtreeRootThreshold int    = 64
 	TxSizeCostPerByte    uint64 = 10
 	GasPerBlobByte       uint32 = 8
 	MaxTxSize            int    = 2097152 // 2 MiB in bytes
+	TimeoutPropose              = time.Millisecond * 3500
+	TimeoutCommit               = time.Millisecond * 4200
+	// UpgradeHeightDelay is the number of blocks after a quorum has been
+	// reached that the chain should upgrade to the new version. Assuming a block
+	// interval of 12 seconds, this is 7 days.
```
rootulp, line 16:
```suggestion
	// interval of 6 seconds, this is 7 days.
```

pkg/appconsts/versioned_consts.go
```diff
@@ -52,3 +55,45 @@ var (
...
+		return parsedValue
+	}
+	switch v {
+	case v1.Version:
+		return v1.UpgradeHeightDelay
+	case v2.Version:
+		return v2.UpgradeHeightDelay
+	default:
+		return v3.UpgradeHeightDelay
+
```
rootulp, line 97:
```suggestion
```

pkg/appconsts/consensus_consts.go
```diff
@@ -3,11 +3,9 @@ package appconsts
...
 
 const (
-	TimeoutPropose = time.Second * 10
-	TimeoutCommit  = time.Second * 11
 	// GoalBlockTime is the target time interval between blocks. Since the block
 	// interval isn't enforced at consensus, the real block interval isn't
 	// guaranteed to exactly match GoalBlockTime. GoalBlockTime is currently targeted
 	// through static timeouts (i.e. TimeoutPropose, TimeoutCommit).
-	GoalBlockTime = time.Second * 15
+	GoalBlockTime = time.Second * 6
```
rootulp: Modifying this constant will impact some of the defaults in default_overrides. If we do this, there are comments that need to be updated. For example: 

https://github.com/rootulp/celestia-app/blob/17b3a727e698c298658f329362ab77d5fb045684/app/default_overrides.go#L144-L147

also we prob need to update specs with the new defaults even though the old defaults are the ones that are running on mainnet.

So we should prob reconsider modifying this constant. IMO the easiest thing to do would be to delete this constant and modify the defaults that use it to hard-coded values that match the current implementation (so based on 15 seconds).

evan-forbes: we can technically update this, as it is only ever updated in the genesis state, but it correctly causes the app hash tests to fail so its staying 15s for now

evan-forbes: tbc, this change has been reverted

rootulp (approved):
LGTM thanks for pushing this through. 116 commits 🤯 

1. I left a comment about modifying GoalBlockTime that we can address in a FLUP
2. In the future, would be nice if PR description contained a testing section so reviewers know how this change was manually tested

pkg/appconsts/consensus_consts.go
rootulp: yea +1 on not modifying this rn

evan-forbes:
> In the future, would be nice if PR description contained a testing section so reviewers know how this change was manually tested

@rootulp good idea, added a link to the comment and we can also test in knuu now. it appears to pass if a I manually log it and check, but there's an rpc that keeps failing that seems unrelated

x/signal/keeper.go
```diff
@@ -109,7 +104,7 @@ func (k *Keeper) TryUpgrade(ctx context.Context, _ *types.MsgTryUpgrade) (*types
 		}
 		upgrade := types.Upgrade{
 			AppVersion:    version,
-			UpgradeHeight: sdkCtx.BlockHeader().Height + k.upgradeHeightDelayBlocks,
+			UpgradeHeight: sdkCtx.BlockHeader().Height + appconsts.UpgradeHeightDelay(version),
```
cmwaters, line 107: Now we can't override it :(

test/util/testnode/node_interaction_api.go
```diff
@@ -33,6 +34,7 @@ type Context struct {
 	goContext context.Context
 	client.Context
 	apiAddress string
+	tmNode     *node.Node
```
cmwaters, line 37: this can be removed (but you can do it as a FLUP)

test/util/testnode/network.go
```diff
@@ -33,6 +33,7 @@ func NewNetwork(t testing.TB, config *Config) (cctx Context, rpcAddr, grpcAddr s
 	})
 
 	cctx = NewContext(ctx, config.Genesis.Keyring(), config.TmConfig, config.Genesis.ChainID, config.AppConfig.API.Address)
+	cctx.tmNode = tmNode
```
cmwaters, line 36: same here

cmwaters (approved):
Ship it!

(But the override no longer works so we need to fix that in a later PR)

x/signal/keeper.go
evan-forbes: I think it does in that function afaiu, but I did not test

https://github.com/celestiaorg/celestia-app/blob/fbbd9e613628225edae690d91a2525ba8f50c15c/pkg/appconsts/versioned_consts.go#L83-L89

cmwaters: Ah, yeah I think you might be right (I had skipped over that part)

We could probably move the string parsing to an `init` function

PR #3892, opened 2024-09-20
fix: force no bbr when starting local devnet

rach-id:
## Overview

adds the `--force-no-bbr` flag when starting the local devnet so that it works.


rootulp (approved):

staheri14 (approved):
utAck!

evan-forbes (approved):

ninabarbakadze (approved):

cmwaters (approved):

PR #3900, opened 2024-09-24
chore: add a knuu getter

mojtaba-esk:
While attempting to use testnet pkg in the node repo. There is a need to use `knuu` object that is initialized by the `testnet.New()` to create new instances and interact with `knuu`. Since it is not exported, this PR proposed to have a getter func for it, so users can interact with knuu easily.

### DIFF
--- a/test/e2e/testnet/testnet.go
+++ b/test/e2e/testnet/testnet.go
@@ -57,6 +57,10 @@ func New(ctx context.Context, name string, seed int64, grafana *GrafanaInfo, cha
 	}, nil
 }
 
+func (t *Testnet) Knuu() *knuu.Knuu {
+	return t.knuu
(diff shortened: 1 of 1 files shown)


ninabarbakadze (approved):

rootulp (approved):

cmwaters (approved):
It's probably cleaner to pass the `*knuu.Knuu` struct to the testnet constructor but this is also fine

mojtaba-esk:
> It's probably cleaner to pass the `*knuu.Knuu` struct to the testnet constructor but this is also fine

Thanks for the feedback. yes that could make it better. Most parts of the e2e could be refactored as for example I see a number of funcs have too many arguments which we could use options instead.

PR #3909, opened 2024-09-25
feat: cap tx size AnteHandler

ninabarbakadze:


app/ante/max_tx_size.go
```diff
@@ -0,0 +1,28 @@
...
+func NewMaxTxSizeDecorator() MaxTxSizeDecorator {
+	return MaxTxSizeDecorator{}
+}
+
+// AnteHandle implements the AnteHandler interface. It ensures that tx size is under application's configured threshold.
+func (d MaxTxSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	// This is a tx validity check, therefore it only applies to CheckTx.
+	// Tx size rule applies to app versions v3 and onwards.
+	if ctx.IsCheckTx() && ctx.BlockHeader().Version.App >= v3.Version {
+		if len(ctx.TxBytes()) >= appconsts.TxMaxBytes(ctx.BlockHeader().Version.App) {
```
rootulp, line 23: [nit] can we rename the appconst so that it uses the same name as the constant

```suggestion
		if len(ctx.TxBytes()) >= appconsts.MaxTxBytes(ctx.BlockHeader().Version.App) {
```

app/ante/max_tx_size.go
```diff
@@ -0,0 +1,28 @@
...
+			return ctx, sdkerrors.ErrTxTooLarge
```
rootulp, line 24: [question]

`sdkerrors.ErrTxTooLarge` appears already defined. Out of curiosity, in what scenarios is it thrown outside of this PR?

pkg/appconsts/v3/app_consts.go
```diff
@@ -6,4 +6,5 @@ const (
 	SubtreeRootThreshold int    = 64
 	TxSizeCostPerByte    uint64 = 10
 	GasPerBlobByte       uint32 = 8
+	MaxTxBytes           int    = 2097152 // 2MG (2 * 1024 * 1024)
```
rootulp, line 9:
```suggestion
	MaxTxBytes           int    = 2097152 // 2 MiB in bytes
```

pkg/appconsts/versioned_consts.go
```diff
@@ -42,6 +42,10 @@ func GasPerBlobByte(_ uint64) uint32 {
 	return v3.GasPerBlobByte
 }
 
+func TxMaxBytes(_ uint64) int {
+	return v3.MaxTxBytes
+}
```
rootulp, line 47: [nit] 
```suggestion
func MaxTxBytes(_ uint64) int {
	return v3.MaxTxBytes
}
```

specs/src/ante_handler_v3.md
```diff
@@ -3,6 +3,7 @@
 The AnteHandler chains together several decorators to ensure the following criteria are met for app version 3:
 
 - The tx does not contain any messages that are unsupported by the current app version. See `MsgVersioningGateKeeper`.
+- The tx size is not larger than the application's configured versioned constant MaxTxSize.
```
ninabarbakadze, line 6: I'll add the link to MaxTxSize in appconsts once this pr is merged

rootulp (approved):
Implementation LGTM, great work! I think we need a FLUP issue to write a CIP for this.

nit: PR title should include `!` because I think this is consensus breaking.

rootulp:
Another nit: PR title could be more readable to end-users b/c it will show up in release notes. For example: `feat!: limit the max tx size to 2 MiB`.

ninabarbakadze:
> Another nit: PR title could be more readable to end-users b/c it will show up in release notes. For example: "feat!: limit the max tx size to 2 MiB".
> 
> Can also add this to release notes.

Thanks @rootulp! Updated :)

app/ante/max_tx_size.go
```diff
@@ -0,0 +1,27 @@
...
+type MaxTxSizeDecorator struct{}
+
+func NewMaxTxSizeDecorator() MaxTxSizeDecorator {
+	return MaxTxSizeDecorator{}
+}
+
+// AnteHandle implements the AnteHandler interface. It ensures that tx size is under application's configured threshold.
+func (d MaxTxSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	// Tx size rule applies to app versions v3 and onwards.
+	if ctx.BlockHeader().Version.App >= v3.Version {
```
evan-forbes, line 21: [very optional]

imo, if we can exit with a negative, then we should. this follows golang's norms

so 

```
if ctx.BlockHeader().Version.App < v3.Version {
    next() // or whatever
   }
   ```

File: app/ante/max_tx_size.go [hunk 7]
```diff
@@ -0,0 +1,27 @@
...
+		if len(ctx.TxBytes()) >= appconsts.MaxTxBytes(ctx.BlockHeader().Version.App) {
+			return ctx, sdkerrors.ErrTxTooLarge
```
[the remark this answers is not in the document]
evan-forbes: this is good to note, whichever is used, we simply need to document it

using strictly greater than makes sense to me, as we can just use that constant everywhere instead of subtracting 1 everywhere as well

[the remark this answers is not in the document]
cmwaters: I second this that it would be better if it were strictly greater than 2MB

cmwaters (commented):
A few nits otherwise looks good

app/ante/max_tx_size.go
```diff
@@ -0,0 +1,29 @@
...
+
+// AnteHandle implements the AnteHandler interface. It ensures that tx size is under application's configured threshold.
+func (d MaxTxSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	// Tx size rule applies to app versions v3 and onwards.
+	if ctx.BlockHeader().Version.App < v3.Version {
+		return next(ctx, tx, simulate)
+	}
+
+	if len(ctx.TxBytes()) > appconsts.MaxTxBytes(ctx.BlockHeader().Version.App) {
+		return ctx, sdkerrors.ErrTxTooLarge
```
rootulp, line 26: [nit] The current error lacks context for a user to understand how large their transaction is and how large of a tx the system can support.

Proposal to include two more pieces of information in this error: the size of the current tx and the max tx bytes.

rootulp: Also for posterity it looks like `mempool.ErrTxTooLarge` is thrown by Tendermint if the tx exceeds the mempool's max tx bytes parameter.

In BroadcastTx the Cosmos SDK translates the `mempool.ErrTxTooLarge` into a `sdkerrors.ErrTxTooLarge` which is the same error used here.

[proposal] I wonder if we should use a different error to differentiate that this error is _not_ because the tx exceeds the mempool's configured max tx bytes but instead is the application's max tx bytes. Since the mempool can be configured locally, user's may or may not observe a `mempool.ErrTxTooLarge` (and therefore `sdkerrors.ErrTxTooLarge` depending on which node they submit transactions to.

rootulp (approved):
LGTM with one nit and one proposal

app/ante/max_tx_size.go
ninabarbakadze: I can address these in a follow-up since I need to link the new versioned constant in ante handler docs anyway

evan-forbes (approved):
nice, thanks for applying feedback :relaxed: 


:point_right: :point_left: in the future, do you mind responding on the threads after the feedback is applied (ideally with a commit :innocent: ), so then its even fewer brain cycles to hit the green button. thanks!! :pray:

app/ante/max_tx_size.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// AnteHandle implements the AnteHandler interface. It ensures that tx size is under application's configured threshold.
+func (d MaxTxSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	// Tx size rule applies to app versions v3 and onwards.
+	if ctx.BlockHeader().Version.App < v3.Version {
+		return next(ctx, tx, simulate)
+	}
+
+	if len(ctx.TxBytes()) > appconsts.MaxTxBytes(ctx.BlockHeader().Version.App) {
```
evan-forbes, line 25: can we also delete the MaxTotalBlobSizeDecorator?

are there any scenarios where this context is not set properly? like in prepare proposal?

cmwaters: I think we should still keep it because one limits by number of shares and the other number of bytes. I think you pointed out that you could have a PFB that is full of 1 byte blobs which would exceed the share limit but no the byte limit

rootulp: `MaxTotalBlobSizeDecorator` only applies to app version 1

```
		// Ensure that the tx's total blob size is <= the max blob size.
		// Only applies to app version == 1.
		blobante.NewMaxTotalBlobSizeDecorator(blobKeeper),
```

and since the ante handler added in this PR only applies to app version >= 3, I think we need to retain `MaxTotalBlobSizeDecorator`.

The ante handler that limits based on number of shares is `NewBlobShareDecorator` and that one is applied to app version >= 2.

```
		// Ensure that the blob shares occupied by the tx <= the max shares
		// available to blob data in a data square. Only applies to app version
		// >= 2.
		blobante.NewBlobShareDecorator(blobKeeper),
```

rootulp (changes requested):
lint is failing: https://github.com/celestiaorg/celestia-app/actions/runs/11278126505/job/31365885695?pr=3909#step:5:30

test-short is failing: https://github.com/celestiaorg/celestia-app/actions/runs/11278126505/job/31365889567?pr=3909#step:4:219

ninabarbakadze:
> lint is failing: https://github.com/celestiaorg/celestia-app/actions/runs/11278126505/job/31365885695?pr=3909#step:5:30
> 
> test-short is failing: https://github.com/celestiaorg/celestia-app/actions/runs/11278126505/job/31365889567?pr=3909#step:4:219

oops my local linter seems to be not working. one sec!

app/ante/max_tx_size.go
```diff
@@ -0,0 +1,31 @@
...
+// AnteHandle implements the AnteHandler interface. It ensures that tx size is under application's configured threshold.
+func (d MaxTxSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	// Tx size rule applies to app versions v3 and onwards.
+	if ctx.BlockHeader().Version.App < v3.Version {
+		return next(ctx, tx, simulate)
+	}
+
+	maxTxBytes := appconsts.MaxTxBytes(ctx.BlockHeader().Version.App)
+	if len(ctx.TxBytes()) > maxTxBytes {
+		return ctx, fmt.Errorf("tx size is larger than the application's configured threshold: %d bytes", maxTxBytes)
```
rootulp, line 28: [nit] provide the current tx bytes in the error message so that a user knows how many bytes they need to trim from their tx.

app/process_proposal.go
```diff
@@ -90,6 +90,9 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 				return reject()
 			}
 
+			// Set the tx size on the context before calling the AnteHandler
+			sdkCtx = sdkCtx.WithTxBytes(tx)
```
rootulp, line 94: [question] does this need to be gated behind a version gate? @evan-forbes mentioned that it is consensus breaking

```suggestion
		// Set the tx bytes in the context for app version v3 and greater
		if sdkCtx.BlockHeader().Version.App >= 3 {
			sdkCtx = sdkCtx.WithTxBytes(tx)
		}
```

app/process_proposal.go
```diff
@@ -115,6 +118,9 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			return reject()
 		}
 
+		// Set the tx size on the context before calling the AnteHandler
+		sdkCtx = sdkCtx.WithTxBytes(tx)
```
rootulp, line 122: [same question] does this need to be gated behind a version check?

Can this be hoisted up so that it only needs to happen once in this file? Perhaps after this line:

```go
		sdkTx, err := app.txConfig.TxDecoder()(tx)
		// Set the tx bytes in the context for app version v3 and greater
		if sdkCtx.BlockHeader().Version.App >= 3 {
			sdkCtx = sdkCtx.WithTxBytes(tx)
		}
		// rest of file
```

app/test/prepare_proposal_test.go
```diff
@@ -137,6 +140,28 @@ func TestPrepareProposalFiltering(t *testing.T) {
 	require.NoError(t, err)
 	noAccountTx := []byte(testutil.SendTxWithManualSequence(t, encConf.TxConfig, kr, nilAccount, accounts[0], 1000, "", 0, 6))
 
+	// memo is 2MG resulting in the transaction being over limit
```
rootulp, line 143:
```suggestion
	// memo is 2 MiB resulting in the transaction being over limit
```

app/test/process_proposal_test.go
```diff
@@ -49,6 +50,26 @@ func TestProcessProposal(t *testing.T) {
...
+	// create 2 single blobTxs that include a large memo making the transaction
+	// larger than the configured max tx bytes
+	largeBlobTxs := blobfactory.ManyMultiBlobTx(
+		t, enc, kr, testutil.ChainID, accounts[3:], infos[3:],
+		blobfactory.NestedBlobs(
+			t,
+			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 4),
+			[][]int{{100}, {1000}, {420}, {300}},
+		),
+		user.SetMemo(largeMemo))
```
rootulp, line 67: [question] how does this function create 2 single blobTxs? I don't see `2` used as an argument to `ManyMultiBlobTx`

app/validate_txs.go
```diff
@@ -50,6 +50,10 @@ func filterStdTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handler
 			logger.Error("decoding already checked transaction", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()), "error", err)
 			continue
 		}
+
+		// Set the tx size on the context before calling the AnteHandler
+		ctx = ctx.WithTxBytes(tx)
```
rootulp, line 55: [question] does this need to be version gated? In other words is it consensus breaking?

app/validate_txs.go
```diff
@@ -83,6 +87,10 @@ func filterBlobTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handle
 			logger.Error("decoding already checked blob transaction", "tx", tmbytes.HexBytes(coretypes.Tx(tx.Tx).Hash()), "error", err)
 			continue
 		}
+
+		// Set the tx size on the context before calling the AnteHandler
+		ctx = ctx.WithTxBytes(tx.Tx)
```
rootulp, line 92: [question] does this need to be version gated? In other words is it consensus breaking?

rootulp (changes requested):
Pretty confident the process_proposal change needs to be gated behind an app version check. There are a few other instances that should be verified if they are consensus breaking and if they are then we need version gates.

app/process_proposal.go
ninabarbakadze: me and @evan-forbes briefly discussed this yesterday but I don't think we reached a firm conclusion. `ConsumeTxSizeGasDecorator` also uses TxBytes set on the context so I guess it is consensus breaking? I could also reset it back to empty after `MaxTxSizeDecorator` is executed.

rootulp: It seems like a bug that it wasn't set before `ConsumeTxSizeGasDecorator`. 

> do we want it to be set during the ante handler execution in v3?

IMO yes

evan-forbes: I added this gate in https://github.com/celestiaorg/celestia-app/pull/3960 (which we can close after this)

I think there's a chance it could be consensus breaking because of the ante handle that subtracts gas for the size of the tx. we don't currently set it in process proposal afaik, which would mean that's its possible to reject or accept a block if an account does or doesn't have enough gas. very unlikely, but I think it could happen.

evan-forbes: so we need a gate

app/process_proposal.go
evan-forbes: referencing this comment https://github.com/celestiaorg/celestia-app/pull/3909#discussion_r1796032074

we need a gate for process proposal, we don't need a gate prepare proposal

app/ante/max_tx_size_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestMaxTxSizeDecorator(t *testing.T) {
+	decorator := ante.NewMaxTxSizeDecorator()
+	anteHandler := sdk.ChainAnteDecorators(decorator)
+
+	testCases := []struct {
+		name        string
+		txSize      int
+		expectError bool
+		appVersion  uint64
+	}{
```
evan-forbes, line 24: optional, but useful 

we can add a bool to determine if this is checkTx here, then add that bool to the test sdk.Context as will with `sdkCtx.WithIsCheckTx(true)`

evan-forbes (commented):
nice, I think we only need the gates

also left an optional change for including `ischecktx` in our tests

app/process_proposal.go
ninabarbakadze: makes sense! thanks!

app/validate_txs.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/3909#discussion_r1796032074

app/validate_txs.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/3909#discussion_r1796032074

app/test/process_proposal_test.go
ninabarbakadze: I'll create as many blobs as the accounts you pass into the function

app/ante/max_tx_size.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/3909/commits/54b2b4d1ae3874579b1617aee8da258be4bd7f4c

app/process_proposal.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/3909/commits/54b2b4d1ae3874579b1617aee8da258be4bd7f4c

app/ante/max_tx_size_test.go
ninabarbakadze: both addressed https://github.com/celestiaorg/celestia-app/pull/3909/commits/54b2b4d1ae3874579b1617aee8da258be4bd7f4c

app/process_proposal.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/3909/commits/a873099ce3d6f205a1bcf5c6768c3a9926852a7f

app/process_proposal.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/3909/commits/a873099ce3d6f205a1bcf5c6768c3a9926852a7f

evan-forbes (approved):
LGTM

nice work!

app/test/process_proposal_test.go
```diff
@@ -49,6 +50,26 @@ func TestProcessProposal(t *testing.T) {
...
+
+	// create 1 large sendTx that includes a large memo making the
+	// transaction over the configured max tx bytes limit
+	largeSendTx := testutil.SendTxsWithAccounts(
+		t, testApp, enc, kr, 1000, accounts[0], accounts[1:2], testutil.ChainID, user.SetMemo(largeMemo),
 	)
```
cmwaters, line 73: I think the memo size is capped (256 bytes IIRC) so this will fail regardless

app/validate_txs.go
cmwaters: nit: this would have been easier to follow if it was done once in `PrepareProposal` where the context actually is created

app/process_proposal.go
```diff
@@ -68,6 +68,11 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 		}
 
 		sdkTx, err := app.txConfig.TxDecoder()(tx)
+		// Set the tx bytes in the context for app version v3 and greater
+		if sdkCtx.BlockHeader().Version.App >= 3 {
+			sdkCtx = sdkCtx.WithTxBytes(tx)
+		}
```
cmwaters, line 74: same nit here: this would be easier to follow if it was placed right after sdkCtx is initialised on line 54

cmwaters (commented):
A few nits which would be nice to address but logically looks good

app/process_proposal.go
ninabarbakadze: it needs to be done while we're iterating over transactions in later lines no?

app/test/process_proposal_test.go
ninabarbakadze: do you know where the memo check is? because it was failing with transaction too large error in the logs

app/process_proposal.go
ninabarbakadze: so you suggest i iterate over block txs another time after context is created and set `WithTxBytes` there?

cmwaters: ignore me. I misunderstood this

app/test/process_proposal_test.go
cmwaters: It's called `ValidateMemoDecorator`

ninabarbakadze: `NewMaxTxSizeDecorator` is called before so it'll be rejected for transaction being too large

specs/src/ante_handler_v3.md
```diff
@@ -3,6 +3,7 @@
 The AnteHandler chains together several decorators to ensure the following criteria are met for app version 3:
 
 - The tx does not contain any messages that are unsupported by the current app version. See `MsgVersioningGateKeeper`.
+- The tx size is not larger than the application's configured versioned constant MaxTxBytes.
```
rootulp, line 6: [not blocking] we should have considered naming it something else to avoid conflicting with the mempool parameter `MaxTxBytes`. For example `MaxTxSize` would've worked

rootulp (approved):
LGTM great work.

Do we need a follow up issue to track writing a CIP for this change? I noted this elsewhere but perhaps the CIP can explain why we're using a tx validity rule instead of a mempool parameter.

specs/src/ante_handler_v3.md
ninabarbakadze: I can change it in a follow-up since i need to update the ante handler anyway :)

app/validate_txs.go
```diff
@@ -51,6 +51,10 @@ func filterStdTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handler
 			logger.Error("decoding already checked transaction", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()), "error", err)
 			continue
 		}
+
+		// Set the tx size on the context before calling the AnteHandler
+		ctx = ctx.WithTxBytes(tx)
```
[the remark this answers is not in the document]
ninabarbakadze: I think this is due to the merge. I'll update in a follow-up

PR #3911, opened 2024-09-26
chore: update e2e tests to knuu 0.16.1

mojtaba-esk:
This PR proposes updating the e2e tests code to use the latest version of `knuu` ([v0.16.1](https://github.com/celestiaorg/knuu/releases/tag/v0.16.1))

The main changes of this release are:
1. Use unique static names for instances: now `knuu` does not add a random string to the names of the instances to make them predictable; therefore, it paves the path to the long running tests.
2. CTRL+C to run timeout handler with no sleep: optimizes the termination request by the user
3. Ability to fetch logs from instances
4. Adjust stopped status actions for the update image: From now on, user can `Stop` the instance, then update the image, arguments, start command and or env vars, then commit and start the instance. This is handy for upgrade test.


go.mod
```diff
@@ -255,4 +248,6 @@ replace (
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.41.0-tm-v0.34.29
+
+	nhooyr.io/websocket => nhooyr.io/websocket v1.8.17
```
mojtaba-esk, line 252: I had to bump the version for the websocket as there was a complain on not finding the `v1.8.6`

cmwaters: Is it not possible to do without the replace statement (I.e. can we bump it directly?)

test/e2e/testnet/testnet.go
```diff
@@ -61,11 +66,11 @@ func (t *Testnet) Knuu() *knuu.Knuu {
 	return t.knuu
 }
 
-func (t *Testnet) NewPreloader() (*preloader.Preloader, error) {
+func (t *Testnet) NewPreloader(name string) (*preloader.Preloader, error) {
```
cmwaters, line 69: Nit: could we just use the name of the testnet here and not need to add the argument?

cmwaters (approved):

ninabarbakadze (approved):
lgtm

go.mod
mojtaba-esk: It is indirect and used in some dependencies, so I am not sure if that's possible. Do you know if we can do that?

rach-id: maybe try changing line 236 to use the new version?

test/e2e/simple.go
```diff
@@ -15,7 +15,9 @@ import (
 // and MsgSends over 30 seconds and then asserts that at least 10 transactions were
 // committed.
 func E2ESimple(logger *log.Logger) error {
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
```
rach-id, line 18: why do you need a cancellable context here? is it getting cancelled somewhere other than the end of the function call?

go.mod
mojtaba-esk: BTW, this `nhooyr.io/websocket` is not used in `knuu` nor the `e2e test`, but not sure why it did not find it on my machine.
```sh
go mod why nhooyr.io/websocket
# nhooyr.io/websocket
github.com/celestiaorg/celestia-app/v3/cmd/celestia-appd/cmd
github.com/cosmos/cosmos-sdk/server/grpc
github.com/improbable-eng/grpc-web/go/grpcweb
nhooyr.io/websocket
```

there might be something wrong with my `goproxy`, could you please remove it and then test it locally to see if everything is alright; so we can remove that replace entirely?
thanks

rach-id (approved):
LGTM

test/e2e/simple.go
mojtaba-esk: good point. will fix it.

test/e2e/testnet/testnet.go
mojtaba-esk: applied

rootulp (dismissed):

test/e2e/major_upgrade_v2.go
```diff
@@ -16,13 +16,14 @@ import (
 )
 
 func MajorUpgradeToV2(logger *log.Logger) error {
-	numNodes := 4
-	upgradeHeight := int64(10)
-	ctx, cancel := context.WithCancel(context.Background())
```
evan-forbes, line 21: for my own understanding, why don't we need the cancel any longer?

test/e2e/minor_version_compatibility.go
```diff
@@ -27,14 +32,10 @@ func MinorVersionCompatibility(logger *log.Logger) error {
...
 	}
-	numNodes := 4
 	r := rand.New(rand.NewSource(seed))
 	logger.Println("Running minor version compatibility test", "versions", versions)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
-	testNet, err := testnet.New(ctx, "runMinorVersionCompatibility", seed, nil, "test")
+	testNet, err := testnet.New(ctx, "MinorVersionCompatibility", seed, nil, "test")
```
evan-forbes, line 38: Why change the name here?

evan-forbes (approved):

test/e2e/major_upgrade_v2.go
mojtaba-esk: The removal was suggested in the previous review process. Since `ctx.Background()` is tied to the main goroutine, it will remain valid until the entire test suite completes. We could put  `ctx.WithCancel()` back to tie the lifecycle explicitly to this particular test function, allowing for more precise resource management.

Also, the line `defer testNet.Cleanup(ctx)` ensures that all resources related to this specific test, including the Knuu instances, are cleaned up properly. So, using either approach is ok, but using `ctx.WithCancel()` provides more control over the test's lifecycle.

test/e2e/minor_version_compatibility.go
mojtaba-esk: It seems like the term `run` does not seem to add any value to the test name. If it is preferred to keep it, we can put it back.

rootulp (approved):

PR #3914, opened 2024-09-26
docs: remaining docs from versioned gas scheduler variables

### LINKED ISSUE #3762: Add a params v3 page to the specs

ninabarbakadze:
## Overview

- Fixes #3762
- Adds ante handler v3 with updated links

### DIFF
--- a/specs/src/ante_handler_v3.md
+++ b/specs/src/ante_handler_v3.md
@@ -1,0 +1,25 @@
+# AnteHandler v3
+
+The AnteHandler chains together several decorators to ensure the following criteria are met for app version 3:
+
+- The tx does not contain any messages that are unsupported by the current app version. See `MsgVersioningGateKeeper`.
+- The tx does not contain any [extension options](https://github.com/cosmos/cosmos-sdk/blob/22c28366466e64ebf0df1ce5bec8b1130523552c/proto/cosmos/tx/v1beta1/tx.proto#L119-L122).
+- The tx passes `ValidateBasic()`.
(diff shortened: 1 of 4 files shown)


specs/src/ante_handler_v3.md
```diff
@@ -0,0 +1,25 @@
+# AnteHandler v3
```
rootulp, line 1: this file won't show up in the markdown book because it isn't listed in the summary.

For subsequent PRs: can you please try running `mdbook serve` in the root directory of the specs? Then you can visually verify that the specs match what you expect them to look like.

rootulp (changes requested):

specs/src/ante_handler_v3.md
ninabarbakadze: Apologies. I haven't spent a lot of time looking at specs. I'll go through it properly.

cmwaters:
Shouldn't we hold out on this until we know if there will be a tx size capping antehandler in v3?

specs/src/ante_handler_v3.md
rootulp: No worries! The https://github.com/celestiaorg/celestia-app/blob/main/specs/README.md should have setup steps

rootulp (approved):
LGTM, thanks for adding this! IMO we can merge this as-is and update the specs if we modify any other things in v3

rach-id (approved):

PR #3833, opened 2024-08-28
ci(goreleaser): fix release candidate suffix

rootulp:
Prevents https://github.com/celestiaorg/celestia-app/issues/3819 from happening again. Also adds a document so other maintainers can generate + upload prebuilt binaries if they don't work again in the future.

The current `-rc*` [prerelease_suffix](https://goreleaser.com/customization/git/#git) was not working because goreleaser treated v2.1.2-rc1 as more recent than v2.1.2 in the [failing workflow](https://github.com/celestiaorg/celestia-app/actions/runs/10523365886/job/29157983992#step:6:68). This resolves by removing the `*`.

## Testing

1. Worked on my fork: https://github.com/rootulp/celestia-app/releases/tag/v2.1.2
2. Worked on this repo: https://github.com/celestiaorg/celestia-app/releases/tag/v2.1.2

### Before

```
    • git state                                      commit=48173df3dc78f9348eedb3796f29ef9e9e5dc584 branch=HEAD current_tag=v2.1.2-rc1 previous_tag=v2.1.2-rc0 dirty=true
```
### After

```
    • git state                                      commit=48173df3dc78f9348eedb3796f29ef9e9e5dc584 branch=HEAD current_tag=v2.1.2 previous_tag=v2.1.2-rc0 dirty=true
```

It is curious that it thinks the previous tag is v2.1.2-rc0 and not v2.1.2-rc1 but I couldn't fix with various [tag_sort](https://goreleaser.com/customization/git/?h=tag_sort#git) options. I think this only impacts Goreleaser changelog generation which we don't use b/c we use Github's auto generate release notes feature.

PR #3929, opened 2024-10-01
fix(ci): classify `-arabica` and `-mocha` releases as prereleases

### LINKED ISSUE #3926: Prebuilt binaries don't get attached for `-arabica` or `-mocha` tag
## Context

We recently shifted to creating `-arabica`, `-mocha` tags for each release.

https://github.com/celestiaorg/celestia-app/blob/363a8c5a81bddcd9a4c88a7ee8c9f2516e1c36a6/.github/workflows/goreleaser.yml#L42-L43

https://github.com/celestiaorg/celestia-app/blob/363a8c5a81bddcd9a4c88a7ee8c9f2516e1c36a6/.goreleaser.yaml#L120-L121

## Problem

GoReleaser didn't attach prebuilt binaries to:
- https://github.com/celestiaorg/celestia-app/releases/tag/v2.2.0-arabica
- https://github.com/celestiaorg/celestia-app/releases/tag/v2.2.0-mocha

GoReleaser thinks the latest tag is v2.2.0-rc0 instead of v2.2.0-arabica or v2.2.0-mocha. See https://github.com/celestiaorg/celestia-app/actions/runs/11020450957/job/30605337119#step:6:99

## Proposal

Fix GoReleaser config so that it can attach prebuilt binaries to these releases. A few tips:
1. Ask questions in GoReleaser Discord or submit a GH issue with details
2. Investigate the prerelease_suffix

pippokr:
Update suffix to match -mocha and -arabica. Fixes #3926


.goreleaser.yaml
```diff
@@ -118,4 +118,4 @@ changelog:
 release:
   prerelease: auto
 git:
-  prerelease_suffix: "-rc"
+  prerelease_suffix: "-*"
```
rootulp, line 121: Are you sure wildcard like `*` works here? It didn't work before. See https://github.com/celestiaorg/celestia-app/pull/3833

I think we may want to use this config:

```
# .goreleaser.yaml
git:
  # What should be used to sort tags when gathering the current and previous
  # tags if there are more than one tag in the same commit.
  #
  # Default: '-version:refname'.
  tag_sort: -version:creatordate

  # What should be used to specify prerelease suffix while sorting tags when gathering
  # the current and previous tags if there are more than one tag in the same commit.
  prerelease_suffix: "-"
```

See https://goreleaser.com/customization/git/?h=prerelease_suffix

rootulp (changes requested):

.goreleaser.yaml
pippokr: Yes, I did some test and everything is working fine.

See https://github.com/pippokr/celestia-app/releases/tag/v2.2.7-arabica

rootulp: Thanks for testing! Do you mind creating a few more tags to verify this works in all edge cases. For example: can you please create:
1. v2.3.0-rc0
2. v2.3.0-arabica
3. v2.3.0-mocha
4. v2.3.0

The GoReleaser CI for each invocation should use fetch the correct tag and attach binaries to the correct release. Also all releases should be considered prerelease except for the last one

pippokr: No bother at all :)

I've created some additional tags as you requested and as you can see, the binaries have been created and attached to the correct release.
https://github.com/pippokr/celestia-app/releases

rootulp: Confirmed it seems to work:

v2.3.0-rc0: https://github.com/pippokr/celestia-app/actions/runs/11158938550/job/31016375536#step:6:93
v2.3.0-arabica: https://github.com/pippokr/celestia-app/actions/runs/11161716747/job/31024948289#step:6:93
v2.3.0-mocha: https://github.com/pippokr/celestia-app/actions/runs/11161986186/job/31025762310#step:6:92
v2.3.0: https://github.com/pippokr/celestia-app/actions/runs/11162196323/job/31026407515#step:6:93

rootulp (commented):
Fairly confident this works based on @pippokr testing but I asked in Discord just to be sure: https://discord.com/channels/890434333251362866/1291397273414139915

🤞 we can merge after GoReleaser confirms this config is valid.

.goreleaser.yaml
pippokr: Happy days :)

rootulp:
From Discord

> I'm pretty sure you can leave prerelease_suffix to the default -.
> See how [goreleaser's config](https://github.com/goreleaser/goreleaser/blob/main/.goreleaser.yaml#L22-L24) does not set that value and can use vX.X.X-nightly release tags.

.goreleaser.yaml
rootulp, line 121: Based on Discord we can use just hyphen

```suggestion
  prerelease_suffix: "-"
```

rootulp (approved):

evan-forbes (approved):

PR #3952, opened 2024-10-08
chore: pass knuu as an argument to the testnet.New

mojtaba-esk:
As pointed out before in [one of the comments](https://github.com/celestiaorg/celestia-app/pull/3900#pullrequestreview-2325619673), if we pass knuu as an argument to the testnet constructor, user will have more flexibility in playing with knuu options and can optimize the resource consumption based on the test requirements.

This PR attempts to achieve that goal.


rootulp (commented):
CI is failing b/c NewBenchmarkTest also needs to be updated to account for the TestnetOptions modification

mojtaba-esk:
> CI is failing b/c NewBenchmarkTest also needs to be updated to account for the TestnetOptions modification

Thanks for pointing it out. now it is fixed.

test/e2e/testnet/testnet.go
```diff
@@ -30,42 +30,25 @@ type Testnet struct {
...
+		keygen:  newKeyGenerator(opts.Seed),
+		grafana: opts.Grafana,
+		knuu:    opts.Knuu,
 	}, nil
 }
 
-func (t *Testnet) Knuu() *knuu.Knuu {
-	return t.knuu
-}
-
```
rootulp, line 68: [question] is this getter no longer helpful?

rootulp (dismissed):
LGTM

test/e2e/testnet/testnet.go
mojtaba-esk: Not really as the user passes the knuu object, so they can use it on their end the way they need.

rootulp (commented):
golangci-lint is complaining

mojtaba-esk:
> golangci-lint is complaining

forgot to save a file, pushed again

test/e2e/benchmark/benchmark.go
```diff
@@ -20,14 +21,32 @@ type BenchmarkTest struct {
...
 
 func NewBenchmarkTest(name string, manifest *Manifest) (*BenchmarkTest, error) {
-	// create a new testnet
-	testNet, err := testnet.New(context.Background(), name, seed,
-		testnet.GetGrafanaInfoFromEnvVar(), manifest.ChainID,
-		manifest.GetGenesisModifiers()...)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	identifier := fmt.Sprintf("%s_%s", name, time.Now().Format("20060102_150405"))
```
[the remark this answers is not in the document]
mojtaba-esk: done

rootulp (approved):

test/e2e/testnet/testnet.go
```diff
@@ -30,39 +30,25 @@ type Testnet struct {
...
-	kn.HandleStopSignal(ctx)
+type Options struct {
+	Seed             int64
+	Grafana          *GrafanaInfo
+	ChainID          string
+	GenesisModifiers []genesis.Modifier
+	Knuu             *knuu.Knuu
+}
 
+func New(opts Options) (*Testnet, error) {
```
cmwaters, line 41: nit: I prefer the use of `Options` for arguments to a constructor which are in fact optional. All fields here can have working defaults except for the Knuu instance which is mandatory thus I'd prefer 

```suggestion
func New(knuu *knuu.Knuu, opts Options) (*Testnet, error) {
```

so now it's more explicit or better still if you go with

```go
type Option func (*Testnet) 

func WithSeed(seed int64) Option {
    return func (t *Testnet) {
        t.seed = seed
    }
}

func New(knuu *knuu.Knuu, options Option...) *Testnet
```

mojtaba-esk: yes the term "option" is kinda counter intuitive here. The main reason for using it to reduce the number of arguments to the constructor. 
Function option could be another way of using it; it is subjective, but I think option alone is cleaner than than the function option way.
we could extract `knuu` out of that option thing. a question might arise here: do we have default values for the rest of options if we extract knuu out? if so, I will do that.

rootulp: Here's a guess at defaults for the options:

```
	Seed             int64 => 42
	Grafana          *GrafanaInfo => &GrafanaInfo{}
	ChainID          string => test-chain-id
	GenesisModifiers []genesis.Modifier => []genesis.Modifier{}
```

mojtaba-esk: ok thanks Rootul, then let's do it

mojtaba-esk: done

test/e2e/testnet/testnet.go
```diff
@@ -30,39 +35,26 @@ type Testnet struct {
...
-		Msg("Knuu initialized")
-
-	kn.HandleStopSignal(ctx)
+type Options struct {
+	Seed             int64
+	Grafana          *GrafanaInfo
+	ChainID          string
+	GenesisModifiers []genesis.Modifier
+	Knuu             *knuu.Knuu
+}
```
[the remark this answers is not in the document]
mojtaba-esk: thanks, will do it asap

test/e2e/benchmark/benchmark.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v3/test/e2e/testnet"
 	"github.com/celestiaorg/celestia-app/v3/test/util/testnode"
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+const (
+	timeFormat = "20060102_150405"
 )
```
rootulp, line 20: [uber nit] a const block isn't necessary if there's only one constant

```suggestion
const timeFormat = "20060102_150405"
```

test/e2e/benchmark/benchmark.go
```diff
@@ -20,14 +25,30 @@ type BenchmarkTest struct {
...
-		testnet.GetGrafanaInfoFromEnvVar(), manifest.ChainID,
-		manifest.GetGenesisModifiers()...)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	identifier := fmt.Sprintf("%s_%s", name, time.Now().Format(timeFormat))
+	kn, err := knuu.New(ctx, knuu.Options{
+		Scope:        identifier,
+		ProxyEnabled: true,
+	})
```
rootulp, line 35: [nit] could be named `scope` because that's how it is used later

```suggestion
	scope := fmt.Sprintf("%s_%s", name, time.Now().Format(timeFormat))
	kn, err := knuu.New(ctx, knuu.Options{
		Scope:        scope,
		ProxyEnabled: true,
	})
```

test/e2e/benchmark/benchmark.go
```diff
@@ -20,14 +25,30 @@ type BenchmarkTest struct {
...
 	if err != nil {
 		return nil, err
 	}
 
+	// context.Background() is used to allow the stopSignal to be functioning even after this function returns
```
rootulp, line 40:
```suggestion
	// context.Background() is used to allow the stopSignal to be functional even after this function returns
```

test/e2e/benchmark/benchmark.go
```diff
@@ -20,14 +25,30 @@ type BenchmarkTest struct {
...
+	kn.HandleStopSignal(context.Background())
+
+	log.Printf("Knuu initialized with scope %s", kn.Scope)
```
rootulp, line 43: [nit] I would move this log statement immediately beneath the Knuu initialization command (and error handling):

```go
	kn, err := knuu.New(ctx, knuu.Options{
		Scope:        identifier,
		ProxyEnabled: true,
	})
	if err != nil {
		return nil, err
	}
	log.Printf("Knuu initialized with scope %s", kn.Scope)
```

test/e2e/major_upgrade_v2.go
```diff
@@ -12,28 +12,40 @@ import (
...
+	"github.com/celestiaorg/knuu/pkg/knuu"
 	"github.com/tendermint/tendermint/rpc/client/http"
 )
 
 func MajorUpgradeToV2(logger *log.Logger) error {
 	var (
+		testName      = "MajorUpgradeToV2"
 		numNodes      = 4
 		upgradeHeight = int64(10)
 	)
```
rootulp, line 24: [nit] this codebase doesn't make use of `var` blocks inside of function bodies so I would've expected to see

```suggestion
	testName := "MajorUpgradeToV2"
	numNodes := 4
	upgradeHeight := int64(10)
```

which is conveniently 2 fewer lines 👍

test/e2e/major_upgrade_v2.go
```diff
@@ -12,28 +12,40 @@ import (
...
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	identifier := fmt.Sprintf("%s_%s", testName, time.Now().Format(timeFormat))
+	kn, err := knuu.New(ctx, knuu.Options{
+		Scope:        identifier,
+		ProxyEnabled: true,
+	})
```
rootulp, line 33: [nit] same nit regarding scope

```suggestion
	scope := fmt.Sprintf("%s_%s", testName, time.Now().Format(timeFormat))
	kn, err := knuu.New(ctx, knuu.Options{
		Scope:        scope,
		ProxyEnabled: true,
	})
```

test/e2e/major_upgrade_v2.go
```diff
@@ -12,28 +12,40 @@ import (
...
+	testnet.NoError("failed to initialize Knuu", err)
+
+	kn.HandleStopSignal(ctx)
+	logger.Printf("Knuu initialized with scope %s", kn.Scope)
```
rootulp, line 37: [nit] same nit about moving this immediately after knuu initialization.

test/e2e/major_upgrade_v3.go
```diff
@@ -10,16 +10,31 @@ import (
...
 )
 
 func MajorUpgradeToV3(logger *log.Logger) error {
-	numNodes := 4
-	upgradeHeightV3 := int64(20)
+	var (
+		testName        = "MajorUpgradeToV3"
+		numNodes        = 4
+		upgradeHeightV3 = int64(10)
+	)
```
rootulp, line 21: [nit]

```suggestion
	testName := "MajorUpgradeToV3"
	numNodes := 4
	upgradeHeightV3 := int64(10)
```

test/e2e/major_upgrade_v3.go
```diff
@@ -10,16 +10,31 @@ import (
...
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	identifier := fmt.Sprintf("%s_%s", testName, time.Now().Format(timeFormat))
+	kn, err := knuu.New(ctx, knuu.Options{
+		Scope:        identifier,
+		ProxyEnabled: true,
+	})
```
rootulp, line 30: [nit]

```suggestion
	scope := fmt.Sprintf("%s_%s", testName, time.Now().Format(timeFormat))
	kn, err := knuu.New(ctx, knuu.Options{
		Scope:        scope,
		ProxyEnabled: true,
	})
```

test/e2e/major_upgrade_v3.go
```diff
@@ -10,16 +10,31 @@ import (
...
+	testnet.NoError("failed to initialize Knuu", err)
+
+	kn.HandleStopSignal(ctx)
+	logger.Printf("Knuu initialized with scope %s", kn.Scope)
```
rootulp, line 34: [nit] same nit about moving this up

test/e2e/benchmark/benchmark.go
rootulp: [question] the other tests use `ctx` instead of `context.Background()` here. Since ctx is:

```
	ctx, cancel := context.WithCancel(context.Background())
```

that seems possible to do here too.

rootulp (commented):
I had a bunch of nits but decided to address them myself by pushing a commit.

rootulp:
Oops nvm I can't push to your fork until you add me as a collaborator.

rootulp (dismissed):

rootulp:
Approving to unblock. I can address nits in a follow-up if this merges before I get permission to push my commit that addresses nits.

test/e2e/benchmark/benchmark.go
mojtaba-esk: In other tests the ctx remains alive until the end of the test, so it is ok to use it. In the benchmark, this function does some initialization and ends so the ctx dies and to handle Stop signal we need a ctx that is alive until the end of the test/program. that's why `context.Background()` is used.

test/e2e/major_upgrade_v2.go
mojtaba-esk: :) it is kinda subjective. I think the first one looks cleaner, but don't have a strong opinion on it.

test/e2e/major_upgrade_v2.go
mojtaba-esk: it is after knuu initialization.  `knuu.New()` does the initialization and if it does not fail (error check), then we handle the stop signal.

or you mean removing that empty line?

rootulp: I mean moving the log statement.

```diff
	kn, err := knuu.New(ctx, knuu.Options{
		Scope:        scope,
		ProxyEnabled: true,
	})
	testnet.NoError("failed to initialize Knuu", err)
+	logger.Printf("Knuu initialized with scope %s", kn.Scope)

	kn.HandleStopSignal(ctx)
-	logger.Printf("Knuu initialized with scope %s", kn.Scope)
```

test/e2e/simple.go
```diff
@@ -9,24 +9,35 @@ import (
...
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	testNet, err := testnet.New(ctx, "E2ESimple", seed, nil, "test")
+	identifier := fmt.Sprintf("%s_%s", testName, time.Now().Format(timeFormat))
+	kn, err := knuu.New(ctx, knuu.Options{
+		Scope:        identifier,
+		ProxyEnabled: true,
+	})
```
rootulp, line 28:
```suggestion
	scope := fmt.Sprintf("%s_%s", testName, time.Now().Format(timeFormat))
	kn, err := knuu.New(ctx, knuu.Options{
		Scope:        scope,
		ProxyEnabled: true,
	})
```

rootulp (approved):

evan-forbes:
also, as a broad question, why were we able to merge this PR when only a single core app engineer approved?

mojtaba-esk:
> also, as a broad question, why were we able to merge this PR when only a single core app engineer approved?

yes there was another issue in the ci as well. There were some unresolved conversations too which is fixed by Rootul.

PR #3953, opened 2024-10-08
feat: decrease block time to 6s

jcstein:

### DIFF
--- a/pkg/appconsts/consensus_consts.go
+++ b/pkg/appconsts/consensus_consts.go
@@ -9,5 +9,5 @@ const (
 	// interval isn't enforced at consensus, the real block interval isn't
 	// guaranteed to exactly match GoalBlockTime. GoalBlockTime is currently targeted
 	// through static timeouts (i.e. TimeoutPropose, TimeoutCommit).
-	GoalBlockTime = time.Second * 15
+	GoalBlockTime = time.Second * 6
 )
--- a/pkg/appconsts/global_consts.go
+++ b/pkg/appconsts/global_consts.go
@@ -29,8 +29,8 @@ const (
 
 	// DefaultUpgradeHeightDelay is the number of blocks after a quorum has been
 	// reached that the chain should upgrade to the new version. Assuming a block
-	// interval of 12 seconds, this is 7 days.
-	DefaultUpgradeHeightDelay = int64(7 * 24 * 60 * 60 / 12) // 7 days * 24 hours * 60 minutes * 60 seconds / 12 seconds per block = 50,400 blocks.
+	// interval of 6 seconds, this is 7 days.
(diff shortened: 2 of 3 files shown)


pkg/appconsts/global_consts.go
```diff
@@ -29,8 +29,8 @@ const (
 
 	// DefaultUpgradeHeightDelay is the number of blocks after a quorum has been
 	// reached that the chain should upgrade to the new version. Assuming a block
-	// interval of 12 seconds, this is 7 days.
-	DefaultUpgradeHeightDelay = int64(7 * 24 * 60 * 60 / 12) // 7 days * 24 hours * 60 minutes * 60 seconds / 12 seconds per block = 50,400 blocks.
+	// interval of 6 seconds, this is 7 days.
+	DefaultUpgradeHeightDelay = int64(7 * 24 * 60 * 60 / 6) // 7 days * 24 hours * 60 minutes * 60 seconds / 6 seconds per block = 100,800 blocks.
 )
```
cmwaters, line 34: Hmm I think this is doable because users of the earlier version will still crash in any case when the state machine transitions to v3 (even if nodes in the network have different values of the `DefaultUpgradeHeightDelay`

jcstein: Isn't doable? Or is?

Thanks for the review!

pkg/appconsts/consensus_consts.go
```diff
@@ -9,5 +9,5 @@ const (
 	// interval isn't enforced at consensus, the real block interval isn't
 	// guaranteed to exactly match GoalBlockTime. GoalBlockTime is currently targeted
 	// through static timeouts (i.e. TimeoutPropose, TimeoutCommit).
-	GoalBlockTime = time.Second * 15
+	GoalBlockTime = time.Second * 6
```
jcstein, line 12: I'm curious if this will cause block time to be less than 6s? As before it was 15, but block times average around 12s on all networks.

rootulp: Decreasing the value for this constant won't actually lower block times. It looks like this constant is used in 3 places:

1. To determine Mempool TTLDuration
2. To determine Evidence MaxAgeNumBlocks
3. To determine x/ibc MaxExpectedTimePerBlock

jcstein: thank you for following up here. I think then it may be safe to close this PR?

rootulp: yea we re-evaluate what those values should be. It doesn't have to strictly happen in #3882 but let's create a follow-up to tackle that. Thanks!

jcstein: sweet, thanks for the help and context.

PR #3954, opened 2024-10-09
chore: optimize checkTx

evan-forbes:
## Overview

this PR makes a simple optimization for checkTx

### DIFF
--- a/app/process_proposal.go
+++ b/app/process_proposal.go
@@ -67,6 +67,8 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			tx = blobTx.Tx
 		}
 
+		ctx := sdkCtx.WithTxBytes(tx)
+
 		sdkTx, err := app.txConfig.TxDecoder()(tx)
 		if err != nil {
 			if req.Header.Version.App == v1 {
@@ -93,7 +95,7 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			// we need to increment the sequence for every transaction so that
 			// the signature check below is accurate. this error only gets hit
 			// if the account in question doesn't exist.
-			sdkCtx, err = handler(sdkCtx, sdkTx, false)
+			ctx, err = handler(ctx, sdkTx, false)
 			if err != nil {
 				logInvalidPropBlockError(app.Logger(), req.Header, "failure to increment sequence", err)
 				return reject()
--- a/app/process_proposal.go
+++ b/app/process_proposal.go
@@ -67,7 +67,7 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			tx = blobTx.Tx
 		}
 
-		ctx := sdkCtx.WithTxBytes(tx)
+		sdkCtx = sdkCtx.WithTxBytes(tx)
 
 		sdkTx, err := app.txConfig.TxDecoder()(tx)
 		if err != nil {
@@ -95,7 +95,7 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			// we need to increment the sequence for every transaction so that
 			// the signature check below is accurate. this error only gets hit
 			// if the account in question doesn't exist.
-			ctx, err = handler(ctx, sdkTx, false)
+			sdkCtx, err = handler(sdkCtx, sdkTx, false)
 			if err != nil {
 				logInvalidPropBlockError(app.Logger(), req.Header, "failure to increment sequence", err)
 				return reject()
--- a/app/process_proposal.go
+++ b/app/process_proposal.go
@@ -67,7 +67,8 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			tx = blobTx.Tx
 		}
 
-		sdkCtx = sdkCtx.WithTxBytes(tx)
+		// todo: uncomment once we're sure this isn't consensus breaking
+		// sdkCtx = sdkCtx.WithTxBytes(tx)
 
 		sdkTx, err := app.txConfig.TxDecoder()(tx)
 		if err != nil {
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -137,6 +137,22 @@ func TestPrepareProposalFiltering(t *testing.T) {
 	require.NoError(t, err)
 	noAccountTx := []byte(testutil.SendTxWithManualSequence(t, encConf.TxConfig, kr, nilAccount, accounts[0], 1000, "", 0, 6))
 
+	// create a tx that can't be included in a 64 x 64 when accounting for the
+	// pfb along with the shares
+	tooManyShareBtx := blobfactory.ManyMultiBlobTx(
+		t,
+		encConf.TxConfig,
+		kr,
+		testutil.ChainID,
+		accounts[3:4],
+		infos[3:4],
+		blobfactory.NestedBlobs(
+			t,
+			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 4000),
+			[][]int{repeat(4000, 1)},
+		),
+	)[0]
+
 	type test struct {
 		name      string
 		txs       func() [][]byte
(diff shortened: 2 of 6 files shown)


app/test/process_proposal_test.go
```diff
@@ -299,6 +314,19 @@ func TestProcessProposal(t *testing.T) {
...
+			input: &tmproto.Data{
+				Txs: [][]byte{},
+			},
+			mutator: func(d *tmproto.Data) {
+				// this tx will get filtered out by prepare proposal before this
+				// so we add it here
+				d.Txs = append(d.Txs, tooManyShareBtx)
+			},
+			appVersion:     v3.Version,
+			expectedResult: abci.ResponseProcessProposal_REJECT,
```
evan-forbes, line 328: this reject isn't new, but it should still be covered by tests

app/validate_txs.go
```diff
@@ -45,6 +45,7 @@ func FilterTxs(logger log.Logger, ctx sdk.Context, handler sdk.AnteHandler, txCo
 func filterStdTxs(logger log.Logger, dec sdk.TxDecoder, ctx sdk.Context, handler sdk.AnteHandler, txs [][]byte) ([][]byte, sdk.Context) {
 	n := 0
 	for _, tx := range txs {
+		ctx = ctx.WithTxBytes(tx)
```
evan-forbes, line 48: this needs to be added for future things anyway, although it has no impact atm

x/blob/ante/blob_share_decorator_test.go
```diff
@@ -24,141 +30,138 @@ const (
...
+				tc.appVersion,
+				user.NewAccount(testfactory.TestAccName, 1, 0),
+			)
+			require.NoError(t, err)
+
+			blobTx := blobfactory.RandBlobTxs(signer, rand, 1, tc.blobsPerPFB, tc.blobSize)
+
+			btx, isBlob, err := blobtx.UnmarshalBlobTx([]byte(blobTx[0]))
+			require.NoError(t, err)
+			require.True(t, isBlob)
```
evan-forbes, line 154: made this a much more meaningful test by using actual signed txs instead of mocking a pfb size

x/blob/ante/blob_share_decorator.go
```diff
@@ -32,11 +34,11 @@ func (d BlobShareDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool
 	if ctx.BlockHeader().Version.App == v1.Version {
 		return next(ctx, tx, simulate)
 	}
-
+	fmt.Println("calling the ante handler lol")
```
rach-id, line 37:
```suggestion
```

rach-id (approved):

app/test/prepare_proposal_test.go
```diff
@@ -137,6 +137,22 @@ func TestPrepareProposalFiltering(t *testing.T) {
 	require.NoError(t, err)
 	noAccountTx := []byte(testutil.SendTxWithManualSequence(t, encConf.TxConfig, kr, nilAccount, accounts[0], 1000, "", 0, 6))
 
+	// create a tx that can't be included in a 64 x 64 when accounting for the
+	// pfb along with the shares
```
rootulp, line 141: [optional]

```suggestion
	// create a tx that can't be included in a 64 x 64 when accounting for the
	// PFB shares along with the blob shares.
```

app/test/prepare_proposal_test.go
```diff
@@ -137,6 +137,22 @@ func TestPrepareProposalFiltering(t *testing.T) {
...
+		encConf.TxConfig,
+		kr,
+		testutil.ChainID,
+		accounts[3:4],
+		infos[3:4],
+		blobfactory.NestedBlobs(
+			t,
+			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 4000),
+			[][]int{repeat(4000, 1)},
+		),
```
rootulp, line 153: [question] out of curiosity how many PFB shares is this? It seems like 4000 blobs shares + x PFB shares. 

The PFB shares need to be >= 96 in order for the total number of shares to be >= 4096.

x/blob/ante/blob_share_decorator.go
```diff
@@ -75,7 +76,8 @@ func (d BlobShareDecorator) getMaxSquareSize(ctx sdk.Context) int {
 
 // getSharesNeeded returns the total number of shares needed to represent all of
 // the blobs described by blobSizes.
-func getSharesNeeded(blobSizes []uint32) (sum int) {
+func getSharesNeeded(base uint32, blobSizes []uint32) (sum int) {
+	sum = share.CompactSharesNeeded(base)
```
rootulp, line 80: [optional] rename base to something more meaningful

```suggestion
func getSharesNeeded(pfbTxSize uint32, blobSizes []uint32) (sum int) {
	sum = share.CompactSharesNeeded(pfbTxSize)
```

x/blob/ante/blob_share_decorator_test.go
```diff
@@ -24,141 +30,138 @@ const (
...
-			name: "PFB with 1 blob that occupies total square - 1",
-			pfb: &blob.MsgPayForBlobs{
-				BlobSizes: []uint32{uint32(share.AvailableBytesFromSparseShares((squareSize * squareSize) - 1))},
-			},
-			appVersion: v2.Version,
+			name:        "PFB with 1 blob that occupies total square - 1",
+			blobsPerPFB: 1,
+			blobSize:    share.AvailableBytesFromSparseShares(squareSize*squareSize - 1),
+			appVersion:  v2.Version,
+			wantErr:     blob.ErrBlobsTooLarge,
```
rootulp, line 109: [question] why does this error now? If the blob occupies the total square - 1 then there should be room for 1 share to contain the PFB tx.

app/test/prepare_proposal_test.go
```diff
@@ -137,6 +137,22 @@ func TestPrepareProposalFiltering(t *testing.T) {
 	require.NoError(t, err)
 	noAccountTx := []byte(testutil.SendTxWithManualSequence(t, encConf.TxConfig, kr, nilAccount, accounts[0], 1000, "", 0, 6))
 
+	// create a tx that can't be included in a 64 x 64 when accounting for the
+	// pfb along with the shares
+	tooManyShareBtx := blobfactory.ManyMultiBlobTx(
```
rootulp, line 142: [optional] rename for clarity

```suggestion
	tooManyShareBlobTx := blobfactory.ManyMultiBlobTx(
```

rootulp (dismissed):
[follow up] we should improve the docs on blob share decorator in a follow up.

app/test/prepare_proposal_test.go
evan-forbes: good point, the back of the napkin reasoning here is each blob has a size, namespace, and blob share commitment, so the overhead is atleast ~72 bytes + encoding overhead. So by this math we're well over the <50KB found in 96 shares

x/blob/ante/blob_share_decorator_test.go
evan-forbes: hmm that's a good question, let me check

evan-forbes: good catch!! 45251c1757086723335e2625fdc48af9cd4bb4ad

rootulp (approved):
nice fix, ty!

app/process_proposal.go
```diff
@@ -67,6 +67,9 @@ func (app *App) ProcessProposal(req abci.RequestProcessProposal) (resp abci.Resp
 			tx = blobTx.Tx
 		}
 
+		// todo: uncomment once we're sure this isn't consensus breaking
+		// sdkCtx = sdkCtx.WithTxBytes(tx)
```
evan-forbes, line 71: sorry for one more review @rootulp 

I commented this out as I think there are scenarios where this could be consensus breaking and this is not explicitly tested. It should likely be protected by an if statement for v3

evan-forbes: cc @ninabarbakadze

rootulp (approved):

app/process_proposal.go
rootulp: hmm this PR targets main so we can technically include consensus breaking things b/c main is the basis for v3.0.0.

IIUC we just can't backport this to v2.x

evan-forbes:
for posterity, this is the PR that is being backported to v2, therefore, its a non-breaking version. ontop of this PR for v3, a breaking change can be made by uncommenting out the ctx change in process proposal

app/process_proposal.go
ninabarbakadze: do we want to remove this for now? and after this is backported to v2 we'll have it for main with version gates

ninabarbakadze (approved):
goos stuff

PR #3960, opened 2024-10-10
feat: version gate potential consensus breaking changes for v3 in pro…

evan-forbes:
## Overview

ready for review, however it is blocked by #3954 

this PR gates the potentially consensus breaking portions of #3954. It should not be backported to v2 as it requires references to v3

### DIFF
--- a/x/blob/ante/blob_share_decorator.go
+++ b/x/blob/ante/blob_share_decorator.go
@@ -3,6 +3,7 @@ package ante
 import (
 	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
 	v1 "github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v1"
+	v2 "github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v2"
 	blobtypes "github.com/celestiaorg/celestia-app/v3/x/blob/types"
 	"github.com/celestiaorg/go-square/v2/share"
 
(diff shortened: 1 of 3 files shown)


x/blob/ante/blob_share_decorator.go
```diff
@@ -25,12 +26,17 @@ func NewBlobShareDecorator(k BlobKeeper) BlobShareDecorator {
...
+	case v2.Version:
+		if !ctx.IsCheckTx() {
+			return next(ctx, tx, simulate)
+		}
+	case v1.Version:
 		return next(ctx, tx, simulate)
+	default:
+		if ctx.IsReCheckTx() {
+			return next(ctx, tx, simulate)
+		}
```
rootulp, line 39: [question] this appears to be missing a `case v3.Version`

[nit] it's cognitively difficult to understand if this case block is correct. Is it really necessary to skip the ante handler for recheck tx? That seems like a minor performance enhancement and/or premature optimization. 

I would expect the switch statement to look like this:

In other words, not depend on ReCheckTx or CheckTx
```go
	switch ctx.BlockHeader().Version.App {
	case v1.Version:
		// this ante handler was not used in v1
		return next(ctx, tx, simulate)
	case v2.Version:
		// run ante handler
	case v3.Version:
		// run ante handler
	default:
		return ctx, fmt.Errorf("unsupported app version %v", ctx.BlockHeader().Version.App)
	}
```

x/blob/ante/blob_share_decorator_test.go
```diff
@@ -169,3 +170,53 @@ func TestBlobShareDecorator(t *testing.T) {
...
 }
+
+func TestBSDAppVersion(t *testing.T) {
+	type test struct {
+		appVersion uint64
+		isCheckTx  bool
+		expectErr  bool
+	}
+	tests := []test{
+		{appVersion: v1.Version, isCheckTx: false, expectErr: false},
```
rootulp, line 181: [question] can we make the antehandler behave the same regardless of `isCheckTx`? That way we can eliminate a whole class of these test cases.

If not, should we add a test case for `v1.Version` and `isCheckTx: true`?

evan-forbes:
after talking with others, I think we can actually close this PR

we're not trying to impose any new rules, so there's no need to include a change in process proposal. If the change is checked in the mempool, then there's not need to check it again in prepare proposal. The most important aspect is already protected by the square construction itself

Issue #3959, opened 2024-10-10
Modify other constants based on decreased block time


rootulp:
## Context

_Originally posted by @jcstein in https://github.com/celestiaorg/celestia-app/pull/3953#discussion_r1794537751_

## Problem

The constant `GoalBlockTime` is used:
 
> 1. To determine Mempool TTLDuration
> 2. To determine Evidence MaxAgeNumBlocks
> 3. To determine x/ibc MaxExpectedTimePerBlock

#3882 proposes decreasing block times from 12 to 6 seconds.

## Proposal

Re-evaluate if those constants need to be modified if we decrease block time from 12 to 6 seconds.

rootulp:
## Mempool TTL duration

```go
// TTLDuration, if non-zero, defines the maximum amount of time a transaction
// can exist for in the mempool.
//
// Note, if TTLNumBlocks is also defined, a transaction will be removed if it
// has existed in the mempool at least TTLNumBlocks number of blocks or if it's
// insertion time into the mempool is beyond TTLDuration.
TTLDuration time.Duration `mapstructure:"ttl-duration"`
```

Currently

```toml
ttl-duration = "1m15s"
```

which is 12.5 blocks. Since we also have

```
ttl-num-blocks = 5
```

we don't need to modify this value.

## Evidence MaxAgeNumBlocks

```
consensus.evidence.MaxAgeDuration = 1814400000000000 (21 days)
consensus.evidence.MaxAgeNumBlocks = 120960
```

120960 blocks * 12 seconds per block = 1451520 seconds = 16.8 days
120960 blocks * 6 seconds per block = 725760 seconds = 8.4 days

we don't need to modify this parameter because evidence is only considered expired after 21 days and 120960 blocks (see [here](https://github.com/celestiaorg/celestia-core/blob/a263c2ca632398271ddb57fc0966c83fa4b2299a/evidence/pool.go#L272-L273)). Since 21 days remains intact, evidence won't be pruned prematurely.

## x/ibc MaxExpectedTimePerBlock

TBD

Issue #3686, opened 2024-07-16
Block validity rule to cap the size of the transaction and the number of txs in the block


cmwaters:
### Summary

Similar to how Celestia has constraints on block and square size to ensure that the network does not fall over under excess demand, we should also consider adding constraints to the max tx size and max number of transactions. 

Currently we do have an antehandler that sets the max tx size to the size of the current max block (note that nodes can locally set a smaller max tx size for their mempool). However, as the block jumps to 8MB, having to gossip an 8MB transaction without any chunking could be detrimental to the network. As a precaution we should set a max size that is no longer equal to the block size. This also makes it easier for submitters to know how large they should batch their transactions.

We also want to prevent situations that could lead to excessive computation which could also slow down the node. We could introduce an upper bound on the amount of txs or we could use the `MaxGas` field in the `ConsensusParams`.

### Problem Definition

g

### Proposal

g

cmwaters:
Does this require a CIP (It is technically a consensus breaking rule)?

What would we want to set this size to? 1MB? 2MB?

evan-forbes:
yeah, this would require at least 1 CIP, but as touched on above, the reasoning for the two is quite distinct so we might want to do 2 separate things.

The cap on tx size is difficult, as users have indicated that larger blobs can help things like "super blobs". The protocol also might not need such a limit until blocks are increased to meaningfully larger than > 4MB. Configurations and expectations could also just need to be changed for larger txs. Larger txs take longer to gossip, therefore users should not expect them to be included within the typical 2 block period. For the same reason, proposers using any flavor of compact blocks change their calculations for inclusion of large txs. If that's the case, then maybe we don't need a limit. 

The cap on the number of txs seems much less difficult to find an effective number and to see the urgency. At least for the medium term. This might be needed sooner rather than later as well, considering 5000 txs can already fit in a 2MB block.

cmwaters:
> The cap on the number of txs seems much less difficult to find an effective number and to see the urgency. At least for the medium term. This might be needed sooner rather than later as well, considering 5000 txs can already fit in a 2MB block.

But if the problem here is with execution costs, then don't we want to go about this by setting a max gas?

cmwaters:
> The protocol also might not need such a limit until blocks are increased to meaningfully larger than > 4MB

That could be the case, but that's on the horizon. Ideally we add the cap before we increase such that it has less impact on current users

evan-forbes:
> setting a max gas?

[...]

PR #3961, opened 2024-10-10
feat: increase default mempool `TTLNumBlocks` to 12

rootulp:
Motivation in the **Mempool TTL duration** section of https://github.com/celestiaorg/celestia-app/issues/3959#issuecomment-2405769487

## Description

Modify the default consensus node config for mempools so that:

- TTLNumBlocks from 5 to 12 because 12 six second blocks = 72 seconds which is comparable to the existing 75 second TTLDuration.
- ~~MaxTxBytes from approx 7.5 MiB to 8 MiB because it's easier to understand and communicate~~
- ~~MaxTxsBytes from approx 37.6 MiB to 40 MiB because it's easier to understand and communicate~~

## Testing

`make install && ./scripts/single-node.sh` then cat the config

```toml
version = "v1"
ttl-duration = "1m15s"
ttl-num-blocks = 12
```

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -8,7 +8,6 @@ import (
 	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v3/x/mint"
 	minttypes "github.com/celestiaorg/celestia-app/v3/x/mint/types"
-	"github.com/celestiaorg/go-square/v2/share"
 	"github.com/cosmos/cosmos-sdk/codec"
 	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -36,6 +35,8 @@ import (
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	coretypes "github.com/tendermint/tendermint/types"
 )
+
+const mebibyte = 1024 * 1024
 
 // bankModule defines a custom wrapper around the x/bank module's AppModuleBasic
 // implementation to provide custom default genesis state.
@@ -259,20 +260,14 @@ func DefaultEvidenceParams() tmproto.EvidenceParams {
 func DefaultConsensusConfig() *tmcfg.Config {
 	cfg := tmcfg.DefaultConfig()
 	// Set broadcast timeout to be 50 seconds in order to avoid timeouts for long block times
-	// TODO: make TimeoutBroadcastTx configurable per https://github.com/celestiaorg/celestia-app/issues/1034
 	cfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
 
-	cfg.Mempool.TTLNumBlocks = 5
-	cfg.Mempool.TTLDuration = time.Duration(cfg.Mempool.TTLNumBlocks) * appconsts.GoalBlockTime
-	// Given that there is a stateful transaction size check in CheckTx,
-	// We set a loose upper bound on what we expect the transaction to
-	// be based on the upper bound size of the entire block for the given
-	// version. This acts as a first line of DoS protection
-	upperBoundBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * share.ContinuationSparseShareContentSize
-	cfg.Mempool.MaxTxBytes = upperBoundBytes
-	cfg.Mempool.MaxTxsBytes = int64(upperBoundBytes) * cfg.Mempool.TTLNumBlocks
-	cfg.Mempool.Version = "v1" // prioritized mempool
+	cfg.Mempool.TTLNumBlocks = 12
+	cfg.Mempool.TTLDuration = time.Duration(75 * time.Second)
+	cfg.Mempool.MaxTxBytes = 8 * mebibyte   // 8 MiB
+	cfg.Mempool.MaxTxsBytes = 40 * mebibyte // 40 MiB
+	cfg.Mempool.Version = "v1"              // prioritized mempool
 
 	cfg.Consensus.TimeoutPropose = appconsts.TimeoutPropose
 	cfg.Consensus.TimeoutCommit = appconsts.TimeoutCommit
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -264,7 +264,7 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
 
 	cfg.Mempool.TTLNumBlocks = 12
-	cfg.Mempool.TTLDuration = time.Duration(75 * time.Second)
+	cfg.Mempool.TTLDuration = 75 * time.Second
 	cfg.Mempool.MaxTxBytes = 8 * mebibyte   // 8 MiB
 	cfg.Mempool.MaxTxsBytes = 40 * mebibyte // 40 MiB
 	cfg.Mempool.Version = "v1"              // prioritized mempool
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -36,8 +36,6 @@ import (
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
-const mebibyte = 1024 * 1024
-
 // bankModule defines a custom wrapper around the x/bank module's AppModuleBasic
 // implementation to provide custom default genesis state.
 type bankModule struct {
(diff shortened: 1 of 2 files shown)


app/default_overrides.go
```diff
@@ -259,20 +260,14 @@ func DefaultEvidenceParams() tmproto.EvidenceParams {
...
 	cfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
 
-	cfg.Mempool.TTLNumBlocks = 5
-	cfg.Mempool.TTLDuration = time.Duration(cfg.Mempool.TTLNumBlocks) * appconsts.GoalBlockTime
-	// Given that there is a stateful transaction size check in CheckTx,
-	// We set a loose upper bound on what we expect the transaction to
-	// be based on the upper bound size of the entire block for the given
-	// version. This acts as a first line of DoS protection
-	upperBoundBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * share.ContinuationSparseShareContentSize
```
rootulp, line 272: part of this PR is motivated by a desire to remove this.

app/default_overrides_test.go
```diff
@@ -80,11 +80,11 @@ func TestDefaultConsensusConfig(t *testing.T) {
...
 
-			// overrides
-			MaxTxBytes:   7_897_088,
-			MaxTxsBytes:  39_485_440,
+			// Overrides
+			TTLNumBlocks: 12,
 			TTLDuration:  75 * time.Second,
-			TTLNumBlocks: 5,
+			MaxTxBytes:   8 * mebibyte,
+			MaxTxsBytes:  40 * mebibyte,
```
evan-forbes, line 87: Per sync discussions elsewhere I think we can increase this to 1 GB

app/default_overrides_test.go
```diff
@@ -80,11 +80,11 @@ func TestDefaultConsensusConfig(t *testing.T) {
...
 			WalPath:               tmcfg.DefaultMempoolConfig().WalPath,
 
-			// overrides
-			MaxTxBytes:   7_897_088,
-			MaxTxsBytes:  39_485_440,
+			// Overrides
+			TTLNumBlocks: 12,
 			TTLDuration:  75 * time.Second,
-			TTLNumBlocks: 5,
+			MaxTxBytes:   8 * mebibyte,
```
evan-forbes, line 86: I don't see the harm in keeping this at 8, but it might be less confusing to keep it at 2MiB since that will be the hard coded limit

evan-forbes (dismissed):

app/default_overrides_test.go
ninabarbakadze: yes! let's change this and I'll approve @rootulp

cmwaters: The main problem with dropping this to 2MB is when we decide to increase the tx size limit, we will also need to increase it here (and tell validators to modify their configs else large transactions won't get in). I would opt to leave it 8MB.

Once again we still have the issue of not knowing whether a config has been modified by a user or not to know whether we can override it if we decide to change it

rootulp: > I don't see the harm in keeping this at 8, but it might be less confusing to keep it at 2MiB since that will be the hard coded limit

The current default is ~7.5 MiB so you're proposing a large decrease. Do you mind re-reading the PR description? 

 cc: @evan-forbes @ninabarbakadze

rootulp: To clarify: I think it's confusing to keep the existing ~7.5 MiB. 

If the proposal is to decrease `MaxTxBytes` to 2 MiB, I thought that was actually a solution to https://github.com/celestiaorg/celestia-app/issues/3686 and I asked about it [here](https://github.com/celestiaorg/celestia-app/issues/3686#issuecomment-2393738597). AFAICT there wasn't alignment on setting `MaxTxBytes` to 2 MiB.

app/default_overrides_test.go
rootulp: Created https://github.com/celestiaorg/celestia-app/issues/3966

app/default_overrides_test.go
ninabarbakadze: I do see @cmwaters point but I still think it's counter intuitive to set the default to ~7.5 MiB when we inevitably reject txs that are > 2MiB in CheckTx

app/default_overrides.go
```diff
@@ -259,19 +260,13 @@ func DefaultEvidenceParams() tmproto.EvidenceParams {
 func DefaultConsensusConfig() *tmcfg.Config {
 	cfg := tmcfg.DefaultConfig()
 	// Set broadcast timeout to be 50 seconds in order to avoid timeouts for long block times
-	// TODO: make TimeoutBroadcastTx configurable per https://github.com/celestiaorg/celestia-app/issues/1034
```
rootulp, line 262: The linked GH issue is closed

app/default_overrides.go
```diff
@@ -259,19 +260,13 @@ func DefaultEvidenceParams() tmproto.EvidenceParams {
...
-	cfg.Mempool.TTLNumBlocks = 5
-	cfg.Mempool.TTLDuration = time.Duration(cfg.Mempool.TTLNumBlocks) * appconsts.GoalBlockTime
-	// Given that there is a stateful transaction size check in CheckTx,
-	// We set a loose upper bound on what we expect the transaction to
-	// be based on the upper bound size of the entire block for the given
-	// version. This acts as a first line of DoS protection
-	upperBoundBytes := appconsts.DefaultSquareSizeUpperBound * appconsts.DefaultSquareSizeUpperBound * share.ContinuationSparseShareContentSize
-	cfg.Mempool.MaxTxBytes = upperBoundBytes
-	cfg.Mempool.MaxTxsBytes = int64(upperBoundBytes) * cfg.Mempool.TTLNumBlocks
+	cfg.Mempool.TTLNumBlocks = 12
```
rootulp, line 266: This is the only implementation change

app/default_overrides.go
```diff
@@ -259,19 +260,13 @@ func DefaultEvidenceParams() tmproto.EvidenceParams {
...
+	cfg.Mempool.TTLDuration = 75 * time.Second
+	cfg.Mempool.MaxTxBytes = 7_897_088
+	cfg.Mempool.MaxTxsBytes = 39_485_440
```
rootulp, line 269: These 3 lines just set the values in a more clear way. The actual values are identical before + after this PR.

app/default_overrides_test.go
rootulp: Resolved via https://github.com/celestiaorg/celestia-app/pull/3961/commits/e7f9a6c000f1234369c7f6366ded2d7271587e40

ninabarbakadze (approved):

cmwaters (commented):
I'm pretty indifferent about this. Keep in mind this PR is defaults so it won't change any existing nodes. Just new nodes.

I would slightly prefer to increase to 8MB which can path the way for increases in the cap when they come (which we will want them to). 

I think it would be more useful to also update the descriptions in the TOML file since for example, MaxTxBytes isn't the only length cap, in fact the main one is in the antehandler (so we might even consider replacing this with a hard coded number).

Something I would advocate for is that the created toml file actually comments out all config values to start with (like .zshrc files do) and the user uncomments them if they want to change them, that way we can actually update the defaults and it will apply to all users.

rootulp:
> I would slightly prefer to increase to 8MB which can path the way for increases in the cap when they come (which we will want them to).

This PR doesn't modify `MaxTxBytes` and IMO that should be discussed separately. Agreed that we may want to change it based on the recently added tx size antehandler. 

The motivation for this PR is to update constants based on the decrease of block time from 12 to 6 seconds (https://github.com/celestiaorg/celestia-app/issues/3959#issue-2578922762). Specifically, prior to this change, the mempool would evict transactions that are either 1m 15 seconds old or 6 blocks (i.e. 6 * 11.77 seconds = 70.62 seconds) old. After the block time reduction, transactions will be evicted from the mempool after 6 * 6 seconds = 36 seconds. If we want to retain existing mempool behavior, we should bump the `TTLNumBlocks`.

app/default_overrides.go
cmwaters: Can we have an issue to bump this value past 8MB

cmwaters (approved):
Ok looks good.

Based on what you're saying why not just have a TTL based on time only (rather than num blocks)

rootulp:
> Based on what you're saying why not just have a TTL based on time only (rather than num blocks)

We could. I supposed that's a breaking change to end-users but we can consider removing the TTL num blocks feature.

app/default_overrides.go
evan-forbes: isn't this limited to 2MB anyway?

PR #3962, opened 2024-10-10
docs(specs): update app version for params v3

rootulp:
I missed this when reviewing https://github.com/celestiaorg/celestia-app/pull/3735


ninabarbakadze (approved):

evan-forbes (approved):

cmwaters (approved):
It's a bit weird because I don't feel like app version has a default

rootulp:
Oh good point, maybe we should rename that column to Default / value or split the parameters into ones that are modifiable with a column heading of default and non-modifiable with a column heading of value.

Issue #3979, opened 2024-10-16
Version `DefaultUpgradeHeightDelay`


evan-forbes:
In order to keep roughly a week delay, we should version the `DefaultUpgradeHeightDelay`. To close this issue, version `DefaultUpgradeHeightDelay`. Ideally, such a change would get into v3 so that when upgrading to v4, the delay would be the same.

rootulp:
This appears to have been done via https://github.com/celestiaorg/celestia-app/pull/3882

Please re-open if I'm mistaken @evan-forbes @staheri14

PR #3980, opened 2024-10-16
feat!: version `UpgradeHeightDelay`

### LINKED ISSUE #3978: Stop requiring `TestAppUpgradeV3` produce enough blocks to hit the DefaultUpgradeHeightDelay
https://github.com/celestiaorg/celestia-app/blob/899ad7c56f62560a5666a604606f550ae1804cb3/app/test/upgrade_test.go#L106-L124

we're currently requiring a test to produce >50,400 block to pass, in order to prevent flakes in the race detector and simply to reduce the amount of time required for this test to pass, we should reduce the number of blocks that this has to take. This could be via build flags, or configuring the testapp specifically to not use the same constant.

evan-forbes:
## Overview

closes #3978 #3979

this is just a nice to have for v3. If we're worried about this causing issues at all, we should just postpone this imo.


cmwaters:
Another option is to set a time instead of a number of blocks, which becomes the delay. This makes it easier to know when exactly the switch to vX will happen

evan-forbes:
> Another option is to set a time instead of a number of blocks, which becomes the delay. This makes it easier to know when exactly the switch to vX will happen

I like that the most!

if we postpone to v4, then I definitely think we should do that and bother with the migration etc

pkg/appconsts/v3/app_consts.go
```diff
@@ -11,4 +11,8 @@ const (
 	MaxTxSize            int    = 2097152 // 2 MiB in bytes
 	TimeoutPropose              = time.Millisecond * 3500
 	TimeoutCommit               = time.Millisecond * 4200
+	// UpgradeHeightDelay is the number of blocks after a quorum has been
+	// reached that the chain should upgrade to the new version. Assuming a block
+	// interval of 12 seconds, this is 7 days.
```
rootulp, line 16:
```suggestion
	// reached that the chain should upgrade to the new version. Assuming a block
	// interval of 6 seconds, this is 7 days.
```

rootulp (approved):
LGTM

evan-forbes:
merging now but I'm assigning myself the flups

thanks everyone!!

Issue #3973, opened 2024-10-15
Version the parameters used to limit tx types


evan-forbes:
[question][potentially blocking] I have a strong suspicion that we will need to modify these constraints in a future release. Should we make them versioned constants now to make it easier to modify them in future releases?

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/3942#discussion_r1799630393_

PR #3982, opened 2024-10-16
refactor: version constants for msg limits

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3973

### DIFF
--- a/pkg/appconsts/versioned_consts_test.go
+++ b/pkg/appconsts/versioned_consts_test.go
@@ -72,6 +72,30 @@ func TestVersionedConsts(t *testing.T) {
 			expectedConstant: v3.MaxTxSize,
 			got:              appconsts.MaxTxSize(v3.Version),
 		},
+		{
+			name:             "NonPFBTransactionCap v2",
+			version:          v2.Version,
+			expectedConstant: v2.NonPFBTransactionCap,
+			got:              appconsts.NonPFBTransactionCap(v2.Version),
+		},
+		{
+			name:             "NonPFBTransactionCap v3",
+			version:          v3.Version,
+			expectedConstant: v3.NonPFBTransactionCap,
+			got:              appconsts.NonPFBTransactionCap(v3.Version),
+		},
+		{
+			name:             "PFBTransactionCap v2",
+			version:          v2.Version,
+			expectedConstant: v2.PFBTransactionCap,
+			got:              appconsts.PFBTransactionCap(v2.Version),
+		},
+		{
+			name:             "PFBTransactionCap v3",
+			version:          v3.Version,
+			expectedConstant: v3.PFBTransactionCap,
+			got:              appconsts.PFBTransactionCap(v3.Version),
+		},
 	}
 
 	for _, tc := range testCases {
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -297,7 +297,7 @@ func TestPrepareProposalCappingNumberOfMessages(t *testing.T) {
 		signers = append(signers, signer)
 	}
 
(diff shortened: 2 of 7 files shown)


rach-id (approved):
Thanks for handling this 🙏

pkg/appconsts/versioned_consts_test.go
```diff
@@ -72,6 +72,30 @@ func TestVersionedConsts(t *testing.T) {
 			expectedConstant: v3.MaxTxSize,
 			got:              appconsts.MaxTxSize(v3.Version),
 		},
+		{
```
ninabarbakadze, line 75: doesn't have to be addressed in this pr but we probably need to come up with a more efficient way of testing this before it gets out of hand with all the versions

ninabarbakadze (approved):

cmwaters:
These values are just in prepare proposal. They are not part of consensus and thus don't need to be versioned. We can change them at will from minor release to minor release

rootulp:
> These values are just in prepare proposal. They are not part of consensus and thus don't need to be versioned. We can change them at will from minor release to minor release

is a good point. I think we should close this PR and https://github.com/celestiaorg/celestia-app/issues/3973 as won't do.

evan-forbes:
> These values are just in prepare proposal. They are not part of consensus and thus don't need to be versioned. We can change them at will from minor release to minor release

this is what I thought as well, but I think someone pointed out that this would just save us the scenario where we can't increase them until some consensus breaking change. In that case, we don't have to have the consensus breaking change first, then cut a release that has the increased values.

therefore I'm good with this now

evan-forbes:
we can also version them later if that is the case, so I'm also good with keeping this closed until then

rootulp:
> but I think someone pointed out that this would just save us the scenario where we can't increase them until some consensus breaking change.

Why can't we change them until some consensus breaking change? It's confusing because the constants are located in a file named `global_consts.go` but they don't actually need to remain constant for the lifetime of the network like the other constants in that file:

https://github.com/rootulp/celestia-app/blob/17b3a727e698c298658f329362ab77d5fb045684/pkg/appconsts/global_consts.go#L14

They can change whenever. In retrospect, it might have been clearer to define these constants where they're used in `validate_txs.go`.

rach-id:
We could argue that versioning can be beneficial in scenarios like one version is running 2mb blocks, with these limitations softly enforced by validators not changing the code, and the next version has some outstanding optimisations that won't be activated until a breaking upgrade is done that allows for higher limitations and 8mb blocks for example. In this case we would want to enforce a soft control on the blocks being produced until the upgrade height, and versioning would be beneficial.

But I guess that it's fine to leave it for future us to worry about it and implement it if needed. We can always version anything in the code 😄 

> In retrospect, it might have been clearer to define these constants where they're used in validate_txs.go

I think the confusion arises from these constants being recommended but not mandatory, being only softly enforced. The reason why I think they belong in app consts is to avoid having limitations scattered around the repo that devs will need to source dive to find them (similar to what sometimes happens in LibP2P where some consts are set in weird places, and you have no idea why the code is behaving in a certain way until you deep dive). Having the consts in the same place helps with this. 

We might think of adding a new file that contains non consensus breaking consts and put them all in there, to make the distinction between consensus breaking vs non-breaking. But I guess we don't need to worry much about this as this code will change in the near future with blocks increasing in size, and we might not even need to enforce such limitations.

TL;DR: I see no harm in merging this PR or leaving it for our future selves to worry about 😄

rootulp:
> We might think of adding a new file that contains non consensus breaking consts and put them all in there, to make the distinction between consensus breaking vs non-breaking.

I agree, this could help

PR #3984, opened 2024-10-16
refactor: rename limits from tx to msg

rootulp:
Motivation: https://github.com/celestiaorg/CIPs/pull/220#discussion_r1803281819

IMO we should either include this in v3.0.0-rc0 or close it b/c not worth it to break these constant names after that.


rach-id (approved):

rootulp:
Thoughts @evan-forbes ? It's technically breaking but it's a really small constant rename and there are no downstream users of the old constants (yet).

PR #3990, opened 2024-10-18
docs: add vacuum! spec

evan-forbes:
## Overview

creating a PR for the vacuum spec.

[rendered](https://github.com/celestiaorg/celestia-app/blob/e666c7d38940ef32c475d8347eee301fa91fe327/specs/src/vacuum.md)

### DIFF
--- a/specs/src/vacuum.md
+++ b/specs/src/vacuum.md
@@ -1,0 +1,473 @@
+# Vacuum! Spec
+
+## Intro
+
+Vacuum! is a high throughput, extremely efficient, and very robust blob propagation protocol. It uses highly optimized lazy gossiping and clever prioritization to distribute unique blobs over unique connections. This enables close to the most optimal theoretical performance for any given topoplogy or network load.
+
+Validator Availability Ceritificates (VACs) allow for multi-height pipelining of block data before an arbitrary future block is created by providing a mechanism for validators to signal to proposers which block data they have. This is similar, but simpler, to DAG based protocols.
+
+Unlike other high efficiency block propagation protocols such as Turbine, the topology of the network and the route of the block data is not known ahead of time. Instead, to acheive high throughput, the topology of the network and path of each blob is discovered on the fly as data is propagated via the pipelined lazy gossip of `VAC`s and `WantBlob`s. Beyond the simplicity benefits, using a JIT approach to routing blobs results in a more optimal path than a AOT approach. This is derived from incorporating variables that are impossible to know a head of time, such as the realtime congestion and latency between each peer.
+
+### Assumptions
+
+Vacuum! must be built ontop of a protocol such as QUIC or TCP, meaning it assumes there are guarantees in the order in which messages are read.
+
+## Constants
+
+SAME_VAC_SEND_LIMIT: the number of times a peer can send the same VAC to a peer before being removed. Default = 1
+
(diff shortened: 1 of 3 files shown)


specs/src/vacuum.md
```diff
@@ -0,0 +1,473 @@
...
+
+#### Outbound Logic
+
+- nodes MUST only send a `WantBlob` to a peer after receiving a `VAC` from that same peer.
+- nodes MUST send a `WantBlob` after receiving the first `VAC` for a given `RootVAC`
+- nodes MAY request a `BlobTx` by sending a `WantBlob` to a peer after receiving a `VAC` for that `BlobTx`
+
+> nodes MUST send a `WantBlob` after receiving the first `VAC` for a given `RootVAC`
+
+To expand on the reasoning behind this, the first `VAC` from each validator is prioritized throughput the network to handle one of the "worst case" scenarios where all of the validator's mempools naturally differ. When prioritizing these, the network is maintains its ability to get a high throughput and prioritize the highest value txs. This is also why validators send the highest priority `VAC` to all of their peers (more on this in the Outbound Logic for VACs).
```
rootulp, line 121:
```suggestion
To expand on the reasoning behind this, the first `VAC` from each validator is prioritized throughout the network to handle one of the "worst case" scenarios where all of the validator's mempools naturally differ. When prioritizing these, the network maintains its ability to get a high throughput and prioritize the highest value txs. This is also why validators send the highest priority `VAC` to all of their peers (more on this in the Outbound Logic for VACs).
```

rootulp (commented):
Only partially reviewed. I think a diagram or high level description of the protocol would help prior to diving into inbound / outbound logic.

Historically the specs were filled with lots of content that we thought we would implement but didn't end up implementing. As a result, the specs still contain information that doesn't match the current implementation. 

TBH I don't think the vacuum spec should be added to the specs yet because vaccum isn't implemented or used. IMO it makes more sense to add this as an ADR in celestia-app or potentially even celestia-core. After it gets implemented and incorporated in the protocol, we can add the ADR to the specs.

evan-forbes:
> TBH I don't think the vacuum spec should be added to the specs yet because vaccum isn't implemented or used. 

fair! this is for the upcoming blog post. folks suggested it was better to have this up on github instead of hackmd

evan-forbes:
converted to a draft for now, and we can leave it here until we are confident the implementation and spec match

cmwaters (commented):
Congrats on writing the spec! This gives us a more solid basis to chip away at and improve. 

We will probably need to do the same with Compact Blocks. I'd like to weave them in a way where it's clear how Vacuum! extends the protocol. I was thinking about explaining the model of how there are three things that can improve Compact Blocks efficacy:
- Tx Dissemination Speed - how quickly you can send a transaction to all nodes in a network
- Preparation - how can you increase the likelihood that 2/3+ validators have all the transactions in the block you propose
- Recovery - how quickly can nodes retrieve the missing transactions.

Does this seem like an okay model?

I see Vacuum! as mostly providing benefits to the preparation aspect (but perhaps also to tx dissemination with chunking). I think chunking could be written as a separate spec as it's not fundamental to vacuum!

evan-forbes:
closing as this is a bit stale and have too much on my plate to finish

will likely reopen or start from scratch soon

PR #4003, opened 2024-10-23
feat(x/minfee): extract go module

rootulp:
## Motivation

As part of the refactor to a multiplexed app, we want a mechanism to avoid importing state machine modules (e.g. `x/minfee`) once per state machine version. We want this because each time a state machine module is imported by the multiplexer, it registers proto types in the proto registry and interfaces in the Cosmos SDK interface registry. When we import the same state machine module more than once, we hit duplicate proto registry errors (same for Cosmos SDK interface registry).

This PR is the first of many that spin out a distinct Go module per state machine module. Expect subsequent PRs for `x/blob`, `x/signal`, etc.

Note: if we didn't do this for the multiplexed app, it is likely that we would do this as part of an update to new Cosmos SDK versions. For example Cosmos SDK [v0.52.x](https://github.com/cosmos/cosmos-sdk?tab=readme-ov-file#module-dependencies) spun out Go modules for all the state machine modules. 

Note: the diff of `x/minfee` from v2.x to main is minimal (see `git diff v2.x..main -- x/minfee/`) so I choose to do this on main. We'll need to backport this to v2.x as well.

## Testing

I did this on my fork and verified that we can:
1. Publish multiple Go modules from the same repo (i.e. celestia-app, and x/minfee)
2. From celestia-app import x/minfee. See [here](https://github.com/rootulp/celestia-app/blob/2831fa7f1a111774de2339cfc5ee2860da24e1fc/go.mod#L27)

Ref:
- https://github.com/rootulp/celestia-app/releases/tag/v3.0.0
- https://github.com/rootulp/celestia-app/releases/tag/x%2Fminfee%2Fv1.1.0


x/minfee/params.go
```diff
@@ -13,18 +12,9 @@ const ModuleName = "minfee"
...
-	DefaultNetworkMinGasPrice sdk.Dec
+	KeyNetworkMinGasPrice = []byte("NetworkMinGasPrice")
 )
 
-func init() {
-	DefaultNetworkMinGasPriceDec, err := sdk.NewDecFromStr(fmt.Sprintf("%f", appconsts.DefaultNetworkMinGasPrice))
-	if err != nil {
-		panic(err)
-	}
-	DefaultNetworkMinGasPrice = DefaultNetworkMinGasPriceDec
```
rootulp, line 25: this doesn't need to happen inside an `init()` so I refactored it

evan-forbes (approved):
:+1:

x/minfee/go.mod
```diff
@@ -0,0 +1,136 @@
+module github.com/celestiaorg/celestia-app/x/minfee
```
rach-id, line 1: since we're already in v3 version in the app, shouldn't this also be v3?

go.mod
```diff
@@ -254,6 +255,9 @@ require (
 )
 
 replace (
+	// Replace this import until there's an official release of x/minfee.
+	// After this PR merges, we can create a release of x/minfee v1.0.0
```
rach-id, line 259: same, minfee is already at v3 with app, no?

rach-id (commented):
LGTM. Just a few questions to understand more.

Are we planning on having independent releases of the modules without releasing the app? if so, should we also define workflows to handle module specific pipeline?

x/minfee/go.mod
rootulp: It could be but since there hasn't been a release of the Go module `github.com/celestiaorg/celestia-app/x/minfee` yet, I intended on creating v1.0.0

go.mod
rootulp: We don't have any releases of x/minfee yet so I intended on starting with v1.0.0 instead of v3.0.0

x/minfee/go.mod
rach-id: Cool. If it won't create confusion for different modules having different versions used in a different celestia-app version, im fine with it :+1:

rootulp: TBH I do think it will be confusing. Because now the state machine modules will have versions different from the application.

But IMO it would make more sense for the `x/minfee` version to bump according to SemVer rather than arbitrarily bumping it alongside the app version. For example, we may cut celestia-app v4 soon with no changes to `x/minfee` and then the question is whether we re-release `x/minfee` v4.0.0

rach-id: Sounds good, thanks for clarifying

rach-id (approved):

evan-forbes:
clear to merge @rootulp

rootulp:
During a sync w/ Binary Builders we learned that extracting Go modules for state machine modules isn't a prerequisite for upgrading to SDK v0.52.x so we may not need this.

We may still want it if we pursue our own implementation of a multiplexer so will leave as draft for now.

rootulp:
Closing for now. Can re-open if we need this.

PR #4004, opened 2024-10-23
fix: make check-bbr

evan-forbes:
## Overview

very minor fix that fixes the `check-bbr` command


cmwaters (dismissed):
utAck

rach-id (approved):

ninabarbakadze (approved):
utAck

cmwaters (approved):
re-utAck

PR #4011, opened 2024-10-29
fix: enable tcp_bbr to load automatically after reboot

bombermine3:
## Overview

Fix: Enable automatic loading of tcp_bbr module after reboot


evan-forbes (approved):
dope, thanks

rach-id (approved):

ninabarbakadze (approved):

PR #4025, opened 2024-11-07
fix: replace bash syntax with sh in configure-v3

bombermine3:
## Overview

Since sh is the default shell in the Makefile, running configure-v3 resulted in the error: `/bin/sh: 2: [[: not found`. This fix modifies the syntax to be compatible with sh, allowing the script to run successfully without errors.


evan-forbes (approved):
thanks for fixing these @bombermine3 :heart:

rach-id (approved):

PR #4026, opened 2024-11-07
fix: replace any `ttl-num-blocks` value

kinrokinro:
## Overview
On one of our node I found the default `ttl-num-blocks` value as `10` instead of `5`, so the default approach will not work in this case and it should be fixed with this commit.


kinrokinro:
I also found `ttl-num-blocks = 0` and `ttl-num-blocks = 15` on our RPC nodes (these have been around since the `mocha-2` genesis I guess so that's probably the reason)

evan-forbes:
thanks @kinrokinro !!

rach-id (approved):

ninabarbakadze (approved):
utAck

evan-forbes (approved):

Issue #3975, opened 2024-10-15
`celestia-appd status` reports app version 3 prematurely


rootulp:
## Context

On `main`, I added a log line to print the current version in EndBlocker and ran

```
make install
./scripts/single-node.sh
```

## Problem

The app thinks it is on app version 3 from the beginning even though I think it should start on app version 1, upgrade to app version 2 at block height 3.

Also there are log lines for an upgrade from app version 1 -> 2 at block height 3 which doesn't make sense if the app version is already 3.

Note the genesis doesn't have an app version specified.

```
$ ./scripts/single-node.sh
celestia-app home: /Users/rootulp/.celestia-app
celestia-app version: 3.0.0-alpha-5-ga7476c6c7

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Initializing validator and node config files...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661
Starting celestia-app...
10:55AM INF starting node with ABCI Tendermint in-process
10:55AM INF service start impl=multiAppConn module=proxy msg={}
10:55AM INF service start connection=query impl=localClient module=abci-client msg={}
10:55AM INF service start connection=snapshot impl=localClient module=abci-client msg={}
10:55AM INF service start connection=mempool impl=localClient module=abci-client msg={}
10:55AM INF service start connection=consensus impl=localClient module=abci-client msg={}
10:55AM INF service start impl=EventBus module=events msg={}
10:55AM INF service start impl=PubSub module=pubsub msg={}
10:55AM INF service start impl=IndexerService module=txindex msg={}
10:55AM INF ABCI Replay Blocks appHeight=0 module=consensus stateHeight=0 storeHeight=0
10:55AM INF mounting KV stores for app version 3
10:55AM INF initializing blockchain state from genesis.json
10:55AM INF asserting crisis invariants inv=1/11 module=x/crisis name=bank/nonnegative-outstanding
10:55AM INF asserting crisis invariants inv=2/11 module=x/crisis name=bank/total-supply
10:55AM INF asserting crisis invariants inv=3/11 module=x/crisis name=gov/module-account
10:55AM INF asserting crisis invariants inv=4/11 module=x/crisis name=distribution/nonnegative-outstanding
[...]

PR #4041, opened 2024-11-20
chore: improve single-node.sh

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3975

Fixes a few issues I discovered with single-node.sh

1. `single-node.sh` previously created a genesis file with `appVersion: 3` so even though it attempted to upgrade from v1 -> v2, that step doesn't make any sense because the node was already on v3. This PR fixes the issue by removing the `--v2-upgrade-height` flag from the script
2. Add a few more logs
3. Add a new script: `single-node-upgrades.sh` that can be invoked to start a single node on v1 and upgrade it: v1 -> v2 -> v3. After running the command, the user will see

```
An upgrade is pending to app version 3 at height 50406.
```

## FLUPs

1. Modify single-node.sh to override the upgrade height delay and instead upgrade from v2 -> v3 instantly: https://github.com/celestiaorg/celestia-app/issues/4053


app/app.go
```diff
@@ -459,6 +459,7 @@ func (app *App) EndBlocker(ctx sdk.Context, req abci.RequestEndBlock) abci.Respo
 	res := app.manager.EndBlock(ctx, req)
 	currentVersion := app.AppVersion()
 	// For v1 only we upgrade using an agreed upon height known ahead of time
+	app.BaseApp.Logger().Debug(fmt.Sprintf("currentVersion %v", currentVersion))
```
rach-id, line 462: [nit]
why not something like:
```suggestion
	app.BaseApp.Logger().Debug("current version",  "version", currentVersion)
```

app/app.go
```diff
@@ -476,6 +477,7 @@ func (app *App) EndBlocker(ctx sdk.Context, req abci.RequestEndBlock) abci.Respo
 	} else if shouldUpgrade, newVersion := app.SignalKeeper.ShouldUpgrade(ctx); shouldUpgrade {
 		// Version changes must be increasing. Downgrades are not permitted
 		if newVersion > currentVersion {
+			app.BaseApp.Logger().Info(fmt.Sprintf("upgrading from app version %v to %v", currentVersion, newVersion))
```
rach-id, line 480: [nit]
same
```suggestion
			app.BaseApp.Logger().Info("upgrading app version", "current version", currentVersion, "new version", newVersion))
```

scripts/upgrade-to-v3.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh
+
+# Prerequisites: ensure ./single-node.sh is running in another terminal.
+# Wait until block height is 3 for the node to upgrade from v1 -> v2.
+# This script will upgrade the node from v2 -> v3.
```
rach-id, line 5: [nit]

```suggestion
# This script will upgrade the node from v2 -> v3.
# Prerequisites: ensure ./single-node.sh is running in another terminal.
# Wait until block height is 3 for the node to upgrade from v1 -> v2.
```

rach-id (dismissed):

app/app.go
```diff
@@ -476,6 +477,7 @@
 	} else if shouldUpgrade, newVersion := app.SignalKeeper.ShouldUpgrade(ctx); shouldUpgrade {
 		// Version changes must be increasing. Downgrades are not permitted
 		if newVersion > currentVersion {
+			app.BaseApp.Logger().Info("upgrading app version", "current version", currentVersion, "new version", newVersion))
```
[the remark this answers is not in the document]
rach-id: This should be applied before merging. I made a mistake leaving an extra ) with the suggestion

evan-forbes (dismissed):

rach-id (dismissed):

app/app.go
```diff
@@ -459,6 +459,7 @@ func (app *App) EndBlocker(ctx sdk.Context, req abci.RequestEndBlock) abci.Respo
 	res := app.manager.EndBlock(ctx, req)
 	currentVersion := app.AppVersion()
 	// For v1 only we upgrade using an agreed upon height known ahead of time
+	app.BaseApp.Logger().Debug("current version", "version", currentVersion)
```
cmwaters, line 462: Don't you think this is a little too spammy (I know it's debug logs) to have the version printed for each block?

cmwaters (commented):
I would actually think it would be more convenient that the `single-node.sh` scripts starts from the latest version. We're no longer that concerned with v1 and soon to be v2

app/app.go
rootulp: Yea fair enough, it's mostly for my debugging but prob not useful for node operators. I can remove this line and add it back when I debug version issues.

rootulp:
> I would actually think it would be more convenient that the single-node.sh scripts starts from the latest version. We're no longer that concerned with v1 and soon to be v2

That is fine with me. I'll move the current `single-node.sh` stuff to a new script that actually starts the chain at v1, upgrades to v2, then to v3 so that we can easily test for app version edge cases.

Even though we only care about the v3 state machine for current modifications, I think it's still important to test that we can do v1 -> v2 -> v3 correctly b/c that's what mainnet nodes do.

cmwaters (approved):
Happy to merge this. The `single-node-upgrades.sh` works as expected. As a side note: the chain won't actually upgrade until height 50406, so this seems good just for testing that the upgrade works

rootulp:
> Happy to merge this. The single-node-upgrades.sh works as expected. As a side note: the chain won't actually upgrade until height 50406, so this seems good just for testing that the upgrade works

https://github.com/celestiaorg/celestia-app/pull/4065 should make it upgrade from v2 -> v3 faster (like 3 blocks)

evan-forbes (approved):

PR #4063, opened 2024-11-27
chore: check if a block is empty by taking its reference

rach-id:
This will be helpful in Celestia-node to avoid copying the block when checking if it's empty

### DIFF
--- a/app/extend_block.go
+++ b/app/extend_block.go
@@ -25,8 +25,17 @@ func ExtendBlock(data coretypes.Data, appVersion uint64) (*rsmt2d.ExtendedDataSq
 	return da.ExtendShares(share.ToBytes(dataSquare))
 }
 
-// EmptyBlock returns true if the given block data is considered empty by the
+// IsEmptyBlock returns true if the given block data is considered empty by the
 // application at a given version.
+//
+// Deprecated: Use IsEmptyBlockRef for better performance with large data structures.
 func IsEmptyBlock(data coretypes.Data, _ uint64) bool {
+	return len(data.Txs) == 0
+}
+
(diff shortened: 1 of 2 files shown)


rootulp (approved):

ninabarbakadze (approved):

rootulp (approved):

app/extend_block.go
```diff
@@ -25,8 +25,17 @@ func ExtendBlock(data coretypes.Data, appVersion uint64) (*rsmt2d.ExtendedDataSq
...
+//
+// Deprecated: Use IsEmptyBlockRef for better performance with large data structures.
 func IsEmptyBlock(data coretypes.Data, _ uint64) bool {
 	return len(data.Txs) == 0
 }
+
+// IsEmptyBlockRef returns true if the application considers the given block data
+// empty at a given version.
+// This method passes the block data by reference for improved performance.
+func IsEmptyBlockRef(data *coretypes.Data, _ uint64) bool {
```
cmwaters, line 39: what's the second argument for? Do we still need to keep it around? Also if we're not planning on backporting this can we just break the go API?

rach-id: > what's the second argument for?

in case at some point we want to have specific method for each version. For now, it's not used and not sure why it was added in first place

> Also if we're not planning on backporting this can we just break the go API?

I'm backporting this to v3 to be released and used in node

rootulp: 1. It's for app version
2. We are backporting it

Wondertan, line 39: nit: there are no references in Golang, only pointers

see `unsafe.Pointer` or `reflect.PointerTo` for idiomatic way to name/document pointers

Issue #4053, opened 2024-11-25
Override upgrade height delay for local devnets


rootulp:
## Context

After the signal module receives a successful MsgTryUpgrade, there is an upgrade height delay of N blocks before the state machine upgrades from v2 -> v3

## Problem

The upgrade height delay is too long for testing and local development purposes.

## Proposal

For local devnets (i.e. one started via `./single-node.sh`) consider overriding the upgrade height delay to something like 1 block.

rootulp:
Related: https://github.com/celestiaorg/celestia-app/issues/3979

rootulp:
Looks like we can do this in two ways:
1. Via `OverrideUpgradeHeightDelayStr`
2. Via special casing the upgrade height for blockchains with the chain-id "test" in [here](https://github.com/celestiaorg/celestia-app/blob/593fc8603fdeb49d6d40fa41bf3b40f074e18487/pkg/appconsts/versioned_consts.go#L90-L103)

rootulp:
It's a bit confusing that we have multiple ways to override this field. In a call right after I got back from vacation, I recall discussing this with @evan-forbes @cmwaters and pretty sure we agreed to use the chain-id to override configs so I plan on removing the ability to override the upgrade height via `OverrideUpgradeHeightDelayStr`

PR #4065, opened 2024-11-27
refactor: override the upgrade height for chain-id `test` to 3

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4053

## Description
1. Override the upgrade height for chain-id `test` to 3 so that there are still a few blocks of delay in tests
1. Remove `OverrideUpgradeHeightDelayStr` because it was confusing to have multiple ways to override the upgrade height (1) chain-id and (2) via ENV variables

### DIFF
--- a/pkg/appconsts/versioned_consts.go
+++ b/pkg/appconsts/versioned_consts.go
@@ -87,6 +87,9 @@ func UpgradeHeightDelay(chainID string, v uint64) int64 {
 		}
 		return parsedValue
 	}
+	if chainID == "test" {
+		return 3
+	}
 	switch v {
 	case v1.Version:
 		return v1.UpgradeHeightDelay
@@ -100,6 +103,5 @@ func UpgradeHeightDelay(chainID string, v uint64) int64 {
 		return v2.UpgradeHeightDelay
 	default:
 		return v3.UpgradeHeightDelay
-
 	}
 }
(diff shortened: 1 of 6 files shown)


rootulp:
`make test-e2e` doesn't pass locally before or after this change so idk if this change broke it any more.

pkg/appconsts/versioned_consts.go
```diff
@@ -80,12 +80,8 @@ func GetTimeoutCommit(v uint64) time.Duration {
 
 // UpgradeHeightDelay returns the delay in blocks after a quorum has been reached that the chain should upgrade to the new version.
 func UpgradeHeightDelay(chainID string, v uint64) int64 {
-	if OverrideUpgradeHeightDelayStr != "" {
-		parsedValue, err := strconv.ParseInt(OverrideUpgradeHeightDelayStr, 10, 64)
-		if err != nil {
-			panic("Invalid OverrideUpgradeHeightDelayStr value")
-		}
-		return parsedValue
+	if chainID == "test" {
```
cmwaters, line 83: nit can we use a constant name (that can also be used in tests)

cmwaters (approved):
👍  - thanks for removing the override. Yes there's no need for it. I think there may still be reference to the override in the Dockerfile

PR #4067, opened 2024-11-28
fix: nonce_mismatch log parsing

InoMurko:
The actual error log that would bubble up is actually (using this in celestia-node)
```
broadcast tx error: account sequence mismatch, expected 357545, got 357544: incorrect account sequence
```
So seeking a prefix for `account sequence mismatch` would not work.

### DIFF
--- a/app/errors/nonce_mismatch.go
+++ b/app/errors/nonce_mismatch.go
@@ -32,7 +32,7 @@ func ParseNonceMismatch(err error) (uint64, error) {
 // ParseExpectedSequence extracts the expected sequence number from the
 // ErrWrongSequence error.
 func ParseExpectedSequence(str string) (uint64, error) {
-	if !strings.HasPrefix(str, "account sequence mismatch") {
+	if !strings.Contains(str, "incorrect account sequence") {
 		return 0, fmt.Errorf("unexpected wrong sequence error: %s", str)
 	}
 


rootulp:
hmm the existing error message parsing seems fragile but why would it not work? If the error is: `account sequence mismatch, expected 357545, got 357544: incorrect account sequence` then that error message does have the prefix `account sequence mismatch`.

Before we change the behavior here, I think we need unit test coverage for the expected error messages.

InoMurko:
> hmm the existing error message parsing seems fragile but why would it not work? If the error is: `account sequence mismatch, expected 357545, got 357544: incorrect account sequence` then that error message does have the prefix `account sequence mismatch`.
> 
> 
> 
> Before we change the behavior here, I think we need unit test coverage for the expected error messages. 
> 
> 

The error starts with: 'broadcast tx error: '

rootulp:
> The error starts with: 'broadcast tx error: '

Is that just when the error gets bubbled up to celestia-node? I think that at the point where this error parsing is happening, the error message is just: `account sequence mismatch, expected 357545, got 357544: incorrect account sequence`. We need unit tests before we can be confident this change has the desired effect.

InoMurko:
@rootulp let me know if a unit test is needed and if you have any particular flow you'd want to test, given above info!

rootulp:
> Given that this piece of code (ParseExpectedSequence, ParseNonceMismatch) is not used

This PR changes the behavior of `ParseExpectedSequence` so how does it fix the issue if `ParseExpectedSequence` is not used? 

Do you have concrete steps on how I can repro the issue you're observing? It would be helpful for me to repro if you created a [bug report issue](https://github.com/celestiaorg/celestia-app/issues/new?assignees=&labels=bug&projects=&template=bug-report.yml&title=%3Ctitle%3E)

jcstein:
> We run a fork of celestia-node and we run it in our infra (a light node for mocha4 and mainnet).

We can only support on official releases from @celestiaorg here, FWIW.

jcstein:
I think it would make sense to contribute those changes to celestia-node so that others can use it and so it can be upstreamed to prevent you from running a fork. have you made a PR?

InoMurko:
> I think it would make sense to contribute those changes to celestia-node so that others can use it and so it can be upstreamed to prevent you from running a fork. have you made a PR?

Not yet, ran them over @renaynay (Rene Lubov) and was waiting for feedback.

rootulp:
Closing this because stale. Please open an issue if there is still a problem.

PR #4068, opened 2024-11-28
refactor: remove sysctl dependency from start command

mircea-c:
## Overview

This change was prompted by an attempt to run the application in an unprivileged docker container based on a slim image. The absence of the `sysctl` command causes the application to fail the BBR check, even though that module is loaded and available.

The goals of this PR are:
- allow the application to run the BBR check in a restricted environment such as a docker container
- make the application distribution agnostic by removing the `systemd` dependency
- remove an instance of process forking with `exec`


rach-id (approved):

evan-forbes (approved):
nice! thanks for this @mircea-c

Issue #3823, opened 2024-08-26
`celestia-appd status` sends output to stderr


rootulp:
## Context

I was wondering why I could never pipe the output of celestia-appd status into jq like this:

```shell
celestia-appd status | jq .
```

## Problem

It turns out, `celestia-appd status` sends output to stderr instead of stdout. To repro:

```
root@rootulp-arabica:~# celestia-appd status >stdout.txt 2>stderr.txt

root@rootulp-arabica:~# cat stdout.txt

root@rootulp-arabica:~# cat stderr.txt
{"NodeInfo":{"protocol_version":{"p2p":"8","block":"11","app":"2"},"id":"8066c315325974182830634cbf9086de6f5c5e42","listen_addr":"tcp://0.0.0.0:26656","network":"arabica-11","version":"2.1.1-1-g3c860c38","channels":"40202122233038606100","moniker":"node-name","other":{"tx_index":"off","rpc_address":"tcp://127.0.0.1:26657"}},"SyncInfo":{"latest_block_hash":"5FE738C5C861C7A90737CAA6283DC86BD22CFEFB6F20D8599013867208DF061F","latest_app_hash":"97D2F0E29D5193AA70772266E8C51CB97E53B04FF474A9B3413761F41B5A8464","latest_block_height":"1807125","latest_block_time":"2024-08-26T20:00:39.040946298Z","earliest_block_hash":"EF5E3D7EC80B7E9F6AD9EE1F2870D00859238796910133CFC3858CDA90A1EE4F","earliest_app_hash":"569DF84025DFBA9E8DAFE9B7ED774E02A280CE828055BD105066DFE06BC1F7FF","earliest_block_height":"1770001","earliest_block_time":"2024-08-21T23:27:44.509291497Z","catching_up":false},"ValidatorInfo":{"Address":"EFE52D240D92F81A033DB40EBD42AD9C8C6D900D","PubKey":{"type":"tendermint/PubKeyEd25519","value":"0m7QrYxl0XLzmaQx51IcmgF1yS69nTmhZwqkzGL2V6c="},"VotingPower":"0"}}
```

## Proposal

Send `celestia-appd status` to stdout instead of stderr

rootulp:
As a temporary workaround I do this:

```
$ celestia-appd status 2>&1 | jq .
```

rootulp:
Blocked on a release of celestiaorg/cosmos-sdk with https://github.com/celestiaorg/cosmos-sdk/pull/420

PR #4071, opened 2024-12-02
fix(cmd): set defaults for command output

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3823
Motivation https://github.com/celestiaorg/cosmos-sdk/pull/420#issuecomment-2508725213

## Testing

```bash
make build
./build/celestia-appd status | jq .
``` 
works

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -56,6 +56,9 @@ func NewRootCmd() *cobra.Command {
 	rootCommand := &cobra.Command{
 		Use: "celestia-appd",
 		PersistentPreRunE: func(command *cobra.Command, _ []string) error {
+			command.SetOut(command.OutOrStdout())
+			command.SetErr(command.ErrOrStderr())
+
 			clientContext, err := client.ReadPersistentCommandFlags(initClientContext, command.Flags())
 			if err != nil {
 				return err


rach-id (approved):

cmd/celestia-appd/cmd/root.go
```diff
@@ -56,6 +56,9 @@ func NewRootCmd() *cobra.Command {
 	rootCommand := &cobra.Command{
 		Use: "celestia-appd",
 		PersistentPreRunE: func(command *cobra.Command, _ []string) error {
+			command.SetOut(command.OutOrStdout())
+			command.SetErr(command.ErrOrStderr())
```
evan-forbes, line 60: does this change all logs to go to the output instead of the error? if so, is a breaking change for things that collect logs?

evan-forbes (commented):
I'm good with this change, but if this is across all outputs we need to document that in the release notes

cmd/celestia-appd/cmd/root.go
rach-id: if I understand correctly, this is just for the prints. the logs are handled separatly:
```
	var logWriter io.Writer
	if strings.ToLower(serverCtx.Viper.GetString(flags.FlagLogFormat)) == tmcfg.LogFormatPlain {
		logWriter = zerolog.ConsoleWriter{Out: os.Stderr}
	} else {
		logWriter = os.Stderr
	}

	logLvlStr := serverCtx.Viper.GetString(flags.FlagLogLevel)
	logLvl, err := zerolog.ParseLevel(logLvlStr)
	if err != nil {
		return fmt.Errorf("failed to parse log level (%s): %w", logLvlStr, err)
	}

	serverCtx.Logger = ZeroLogWrapper{zerolog.New(logWriter).Level(logLvl).With().Timestamp().Logger()}
```

Rootul can correct me

rootulp: > does this change all logs to go to the output instead of the error?

Output that targets stderr will still go to stderr. This fixes an issue where commands that used Cobra's 

```go
// Println is a convenience method to Println to the defined output, fallback to Stderr if not set.
func (c *Command) Println(i ...interface{}) {
	c.Print(fmt.Sprintln(i...))
}
```

would previously go to stderr because stdout wasn't configured for the command. The `status` command is an example that used this Println method

evan-forbes: nice, makes sense

it might still be a bit confusing to anyone reading the code, as the logs are set to go to stderr in the sdk

evan-forbes (approved):

cmd/celestia-appd/cmd/root.go
rootulp: > as the logs are set to go to stderr in the sdk

huh that seems unexpected. I'll try running a node pre / post this change to double check that I didn't break something

rootulp: wow you're right. the SDK does send logs to STDERR. Before (and after) this change, `error.txt` contains logs from the node:

```bash
./scripts/single-node.sh > output.txt 2> error.txt
# output.txt contains output from the script 
# error.txt contains logs from the node which is unexpected
```
```

PR #4084, opened 2024-12-05
fix: cap tx size

ninabarbakadze:

### DIFF
--- a/app/check_tx.go
+++ b/app/check_tx.go
@@ -15,6 +15,15 @@ import (
 // transactions that contain blobs.
 func (app *App) CheckTx(req abci.RequestCheckTx) abci.ResponseCheckTx {
 	tx := req.Tx
+
+	// all txs should be under the max tx size limit
+	maxTxSize := appconsts.MaxTxSize(app.AppVersion())
+	currentTxSize := len(tx)
+	if currentTxSize > appconsts.MaxTxSize(app.AppVersion()) {
+		err := fmt.Errorf("tx size %d bytes is larger than the application's configured threshold of %d bytes", currentTxSize, maxTxSize)
+		return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false)
+	}
+
 	// check if the transaction contains blobs
 	btx, isBlob, err := blobtx.UnmarshalBlobTx(tx)
 	if isBlob && err != nil {
--- a/app/test/check_tx_test.go
+++ b/app/test/check_tx_test.go
@@ -217,6 +217,31 @@ func TestCheckTx(t *testing.T) {
 			},
 			expectedABCICode: abci.CodeTypeOK,
 		},
+		{
+			name:      "v1 blob over 2MiB",
+			checkType: abci.CheckTxType_New,
+			getTx: func() []byte {
+				signer := createSigner(t, kr, accs[10], encCfg.TxConfig, 11)
+				blob, err := share.NewV1Blob(share.RandomBlobNamespace(), bytes.Repeat([]byte{1}, 2_000_000), signer.Account(accs[10]).Address())
+				require.NoError(t, err)
+				blobTx, _, err := signer.CreatePayForBlobs(accs[10], []*share.Blob{blob}, opts...)
+				require.NoError(t, err)
+				return blobTx
+			},
+			expectedABCICode: blobtypes.ErrBlobsTooLarge.ABCICode(),
+		},
+		{
+			name:      "v0 blob over 2MiB",
+			checkType: abci.CheckTxType_New,
+			getTx: func() []byte {
+				signer := createSigner(t, kr, accs[10], encCfg.TxConfig, 11)
+				blob, err := share.NewV0Blob(share.RandomBlobNamespace(), bytes.Repeat([]byte{1}, 2_000_000))
+				require.NoError(t, err)
+				blobTx, _, err := signer.CreatePayForBlobs(accs[10], []*share.Blob{blob}, opts...)
+				require.NoError(t, err)
+				return blobTx
+			},
+		},
 	}
 
 	for _, tt := range tests {
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -2,6 +2,7 @@ package app_test
 
 import (
 	"crypto/rand"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
@@ -51,7 +52,6 @@ func TestPrepareProposalPutsPFBsAtEnd(t *testing.T) {
 		accnts[:numBlobTxs],
 		infos[:numBlobTxs],
 		testfactory.Repeat([]*share.Blob{protoBlob}, numBlobTxs),
-		blobfactory.DefaultTxOpts()...,
 	)
 
 	normalTxs := testutil.SendTxsWithAccounts(
@@ -97,75 +97,74 @@ func TestPrepareProposalFiltering(t *testing.T) {
 
 	// create 3 single blob blobTxs that are signed with valid account numbers
 	// and sequences
-	blobTxs := blobfactory.ManyMultiBlobTx(
-		t,
-		encConf.TxConfig,
-		kr,
-		testutil.ChainID,
-		accounts[:3],
-		infos[:3],
-		blobfactory.NestedBlobs(
-			t,
-			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 3),
-			[][]int{{100}, {1000}, {420}},
-		),
-		blobfactory.DefaultTxOpts()...,
-	)
+	// blobTxs := blobfactory.ManyMultiBlobTx(
+	// 	t,
+	// 	encConf.TxConfig,
+	// 	kr,
+	// 	testutil.ChainID,
+	// 	accounts[:3],
+	// 	infos[:3],
+	// 	blobfactory.NestedBlobs(
+	// 		t,
+	// 		testfactory.RandomBlobNamespaces(tmrand.NewRand(), 3),
+	// 		[][]int{{100}, {1000}, {420}},
+	// 	),
+	// )
 
 	// create 3 MsgSend transactions that are signed with valid account numbers
 	// and sequences
-	sendTxs := coretypes.Txs(testutil.SendTxsWithAccounts(
-		t,
-		testApp,
-		encConf.TxConfig,
-		kr,
-		1000,
-		accounts[0],
-		accounts[len(accounts)-3:],
-		testutil.ChainID,
-	)).ToSliceOfBytes()
+	// sendTxs := coretypes.Txs(testutil.SendTxsWithAccounts(
+	// 	t,
+	// 	testApp,
+	// 	encConf.TxConfig,
+	// 	kr,
+	// 	1000,
+	// 	accounts[0],
+	// 	accounts[len(accounts)-3:],
+	// 	testutil.ChainID,
+	// )).ToSliceOfBytes()
 
-	validTxs := func() [][]byte {
-		txs := make([][]byte, 0, len(sendTxs)+len(blobTxs))
-		txs = append(txs, blobTxs...)
-		txs = append(txs, sendTxs...)
-		return txs
-	}
+	// validTxs := func() [][]byte {
+	// 	txs := make([][]byte, 0, len(sendTxs)+len(blobTxs))
+	// 	txs = append(txs, blobTxs...)
+	// 	txs = append(txs, sendTxs...)
+	// 	return txs
+	// }
 
 	// create 3 MsgSend transactions that are using the same sequence as the
 	// first three blob transactions above
-	duplicateSeqSendTxs := coretypes.Txs(testutil.SendTxsWithAccounts(
-		t,
-		testApp,
-		encConf.TxConfig,
-		kr,
-		1000,
-		accounts[0],
-		accounts[:3],
-		testutil.ChainID,
-	)).ToSliceOfBytes()
+	// duplicateSeqSendTxs := coretypes.Txs(testutil.SendTxsWithAccounts(
+	// 	t,
+	// 	testApp,
+	// 	encConf.TxConfig,
+	// 	kr,
+	// 	1000,
+	// 	accounts[0],
+	// 	accounts[:3],
+	// 	testutil.ChainID,
+	// )).ToSliceOfBytes()
 
 	// create a transaction with an account that doesn't exist. This will cause the increment nonce
 	nilAccount := "carmon san diego"
 	_, _, err := kr.NewMnemonic(nilAccount, keyring.English, "", "", hd.Secp256k1)
 	require.NoError(t, err)
-	noAccountTx := []byte(testutil.SendTxWithManualSequence(t, encConf.TxConfig, kr, nilAccount, accounts[0], 1000, "", 0, 6))
+	// noAccountTx := []byte(testutil.SendTxWithManualSequence(t, encConf.TxConfig, kr, nilAccount, accounts[0], 1000, "", 0, 6))
 
 	// create a tx that can't be included in a 64 x 64 when accounting for the
 	// pfb along with the shares
-	tooManyShareBtx := blobfactory.ManyMultiBlobTx(
-		t,
-		encConf.TxConfig,
-		kr,
-		testutil.ChainID,
-		accounts[3:4],
-		infos[3:4],
-		blobfactory.NestedBlobs(
-			t,
-			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 4000),
-			[][]int{repeat(4000, 1)},
-		),
-	)[0]
+	// tooManyShareBtx := blobfactory.ManyMultiBlobTx(
+	// 	t,
+	// 	encConf.TxConfig,
+	// 	kr,
+	// 	testutil.ChainID,
+	// 	accounts[3:4],
+	// 	infos[3:4],
+	// 	blobfactory.NestedBlobs(
+	// 		t,
+	// 		testfactory.RandomBlobNamespaces(tmrand.NewRand(), 4000),
+	// 		[][]int{repeat(4000, 1)},
+	// 	),
+	// )[0]
 
 	// memo is 2 MiB resulting in the transaction being over limit
 	largeString := strings.Repeat("a", 2*1024*1024)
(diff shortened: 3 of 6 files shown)


app/check_tx.go
```diff
@@ -15,6 +15,15 @@ import (
 // transactions that contain blobs.
 func (app *App) CheckTx(req abci.RequestCheckTx) abci.ResponseCheckTx {
 	tx := req.Tx
+
+	// all txs should be under the max tx size limit
```
ninabarbakadze, line 19: we can extract this logic into a helper to make it less repetitive once we decide what to do with the AnteHandler because otherwise it becomes confusing.

app/test/process_proposal_test.go
```diff
@@ -349,26 +328,6 @@ func TestProcessProposal(t *testing.T) {
...
 		},
-		{
-			name:  "blob txs larger than configured max tx size",
-			input: validData(),
-			mutator: func(d *tmproto.Data) {
-				d.Txs = append(d.Txs, largeBlobTxs...)
-				d.Hash = calculateNewDataHash(t, d.Txs)
-			},
-			appVersion:     appconsts.LatestVersion,
-			expectedResult: abci.ResponseProcessProposal_REJECT,
```
evan-forbes, line 360: [blocking]

wait, isn't the problem that this doesn't occur now, and chaning it is consensus breaking? Is this just getting rejected for the reason we think it is?

app/validate_txs.go
```diff
@@ -34,7 +36,20 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []*tx.BlobTx) {
...
+	//nolint:prealloc
+	var txsBelowLimit [][]byte
+	for idx, tx := range txs {
+		if len(tx) > maxTxSize {
+			err := fmt.Sprintf("tx size %d bytes at index %d exceeds the application's configured threshold of %d bytes", len(tx), idx, maxTxSize)
+			logger.Error(err)
+			continue
+		}
+		txsBelowLimit = append(txsBelowLimit, tx)
+	}
```
evan-forbes, line 50: why not check sizes of PFBs since that's what we care about? This would also allow us to keep logic in antehandlers instead of making a special case and having to remember that we have this logic hidden in this function.


side note: in theory, a consensus breaking version would be handled in the `ValidateBasic` method since this is a stateless check on PFBs only. We can't do that now without also passing the app version to all validate basic methods, but we need to that eventually anyway  ref #4088

app/test/process_proposal_test.go
```diff
@@ -51,26 +50,6 @@ func TestProcessProposal(t *testing.T) {
 			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 4),
 			[][]int{{100}, {1000}, {420}, {300}},
 		),
-		blobfactory.DefaultTxOpts()...,
-	)
-
-	largeMemo := strings.Repeat("a", appconsts.MaxTxSize(appconsts.LatestVersion))
```
evan-forbes, line 57: aren't memo's capped at 256? if so, is this a realistic test? Won't all tx with this big of a memo be rejected as is?

evan-forbes (commented):
I think we're okay here but we need to double check a test to make sure. We can't be adding any new rejection cases to processproposal

I'm okay with adding this block in a non-breaking way like this now on the condition that we fix this is a more maintainble way in the near future. meaning, if this is a tx validity rule, then we should use the tools to add tx validity rules (antehandlers). this could involve either #4088 and/or #1166

app/test/prepare_proposal_test.go
```diff
@@ -184,9 +183,23 @@ func TestPrepareProposalFiltering(t *testing.T) {
 		blobfactory.NestedBlobs(
 			t,
 			testfactory.RandomBlobNamespaces(tmrand.NewRand(), 3),
-			[][]int{{100}, {1000}, {420}},
```
rootulp, line 187: [no change needed] out of curiosity why were these 3 blob txs pruned before? They are all smaller than the max tx size.

app/check_tx.go
rootulp, line 19: why does this check need to be moved from the antehandler to check tx?

If we do this PR, do we still need the ante handler: max tx size?

app/test/process_proposal_test.go
evan-forbes: I found https://github.com/celestiaorg/celestia-app/pull/3909#discussion_r1797069274

which I find incredibly opaque and like an enormous footgun

this test should still be changed imo

evan-forbes: the thing we're trying to prevent is blobs that are too big, therefore that's what needs to be tested. using large memo's as a proxy should be avoided

app/test/big_blob_test.go
```diff
@@ -67,8 +67,8 @@ func (s *BigBlobSuite) TestErrBlobsTooLarge() {
 	}
 	testCases := []testCase{
 		{
-			name: "2 mebibyte blob",
-			blob: newBlobWithSize(2 * mebibyte),
+			name: "~ 1.9 mebibyte blob",
+			blob: newBlobWithSize(2_000_000),
```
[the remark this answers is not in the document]
ninabarbakadze: this test is testing for this case `broadcast tx error: the number of shares occupied by blobs in this MsgPayForBlobs 4151 exceeds the max number of shares available for blob data 4096: blob(s) too large` if i set it to 2 MiB it'd no longer hit that error and will directly be rejected for being too large.

we can add a test case for exactly 2 MiB blob. I'll post a commit with the test.

app/test/prepare_proposal_test.go
ninabarbakadze: because they were running out of gas

app/check_tx.go
ninabarbakadze: we can not change the AnteHandler in a patch because it'll be technically consensus breaking because it's called in process proposal. I have an issue open to clean up this logic + AnteHandlers in the next major release https://github.com/celestiaorg/celestia-app/issues/4087

app/test/process_proposal_test.go
ninabarbakadze: I'm removing this test here in process proposal because it won't be removing large blobs and we should add proper testing once we fix the ante handler. wdyt?

app/validate_txs.go
ninabarbakadze: @cmwaters suggested that we should reject encoded transactions directly if the size is over 2MiB. This makes sense to me since we care about transactions not being over limit not only the blobs right? because when this tx is getting gossiped it'd be encoded right? 

I'm not sure we can change the AnteHandler in non breaking way. I see it as a temporary fix in place until we can clean this up and break things in v4.

cmwaters: We can have the check in the antehandler, we would just need to have specific logic for the case of a PFB where we count the the blob sizes as well as the the tx bytes in the context.

> This would also allow us to keep logic in antehandlers instead of making a special case and having to remember that we have this logic hidden in this function.

Antehandlers aren't the only source of validation, we do a lot of blob specific validity checks outside the antehandlers

Having it prior to being unmarshalled is the simplest and most likely safest way to implement a max tx size limit. There is no custom logic based on specific messages

app/check_tx.go
```diff
@@ -15,6 +15,15 @@ import (
 // transactions that contain blobs.
 func (app *App) CheckTx(req abci.RequestCheckTx) abci.ResponseCheckTx {
 	tx := req.Tx
+
+	// all txs must be less than or equal to the max tx size limit
+	maxTxSize := appconsts.MaxTxSize(app.AppVersion())
+	currentTxSize := len(tx)
+	if currentTxSize > maxTxSize {
+		err := fmt.Errorf("tx size %d bytes is larger than the application's configured threshold of %d bytes", currentTxSize, maxTxSize)
```
cmwaters, line 23: Could we make this a standard sdk error with an error code and could we also include the app version in the error message

cmwaters (dismissed):

app/test/big_blob_test.go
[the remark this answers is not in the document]
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/4084/commits/826f56b8bcfa13dfcc2ec31678e7a4b87ee61eec

app/test/check_tx_test.go
```diff
@@ -173,11 +176,11 @@ func TestCheckTx(t *testing.T) {
 			expectedABCICode: abci.CodeTypeOK,
 		},
 		{
-			name:      "10,000,000 byte blob",
+			name:      "2,000,000 byte blob",
 			checkType: abci.CheckTxType_New,
 			getTx: func() []byte {
 				signer := createSigner(t, kr, accs[9], encCfg.TxConfig, 10)
-				_, blobs := blobfactory.RandMsgPayForBlobsWithSigner(tmrand.NewRand(), signer.Account(accs[9]).Address().String(), 10_000_000, 1)
+				_, blobs := blobfactory.RandMsgPayForBlobsWithSigner(tmrand.NewRand(), signer.Account(accs[9]).Address().String(), 2_000_000, 1)
```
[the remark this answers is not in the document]
evan-forbes: we should be good not to version in this case yeah

app/test/process_proposal_test.go
evan-forbes: we definitely don't have to add tests to processproposal since we shouldn't be changing it. if anything we might want a test that confirms no breaking change was made.

I just found this test very confusing, as it was on process proposal and we're adding a memo that should be invalid by itself.

app/validate_txs.go
evan-forbes: I'm good with putting the logic in umarshaling or an antehandler :+1:  I still don't like the idea of not being able to version validate basic logic. 

I think its already a block validity that everything must be decodable. do we pass the app version to the encoding and decoding logic?

app/test/process_proposal_test.go
evan-forbes: cool x-post https://github.com/celestiaorg/celestia-app/pull/4084#discussion_r1876539901

> changing in process proposal is consensus breaking

yeah this is just why I was confused we were testing things there

app/validate_txs.go
evan-forbes, line 50: if we adding this in checkTx, we don't need to add it here too, correct?

evan-forbes (dismissed):
I'm good with merging

app/errors/errors.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"cosmossdk.io/errors"
+)
+
+const AppErrorsCodespace = "app"
+
+// general application errors
+var (
+	ErrTxExceedsMaxSize = errors.Register(AppErrorsCodespace, 11142, "exceeds max tx size limit")
```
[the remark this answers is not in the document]
cmwaters: It happens at runtime in the initialization phase (which is prior to when `main()` is called)

app/validate_txs.go
cmwaters: Yes I think so. In theory we should be able to remove this and rely just on `CheckTx`. There might be an edge cases when we change version

app/test/check_tx_test.go
```diff
@@ -44,6 +45,7 @@ func TestCheckTx(t *testing.T) {
 		checkType        abci.CheckTxType
 		getTx            func() []byte
 		expectedABCICode uint32
+		expectedLog      string
```
cmwaters, line 48: Does the expected log really help if we now have error codes we can assert

cmwaters (dismissed):

app/test/process_proposal_test.go
ninabarbakadze: good idea. I'll add a test with large blob and make sure it doesn't get rejected.

app/test/check_tx_test.go
ninabarbakadze: i guess it's redundant in check tx

app/validate_txs.go
```diff
@@ -34,7 +36,20 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []*tx.BlobTx) {
 //
 // Side-effect: arranges all normal transactions before all blob transactions.
 func FilterTxs(logger log.Logger, ctx sdk.Context, handler sdk.AnteHandler, txConfig client.TxConfig, txs [][]byte) [][]byte {
-	normalTxs, blobTxs := separateTxs(txConfig, txs)
+	// all transactions should be below the max tx size
```
rootulp, line 39:
```suggestion
	// all transactions should be less than or equal to the max tx size
```

app/validate_txs.go
```diff
@@ -34,7 +36,20 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []*tx.BlobTx) {
...
+	maxTxSize := appconsts.MaxTxSize(ctx.BlockHeader().Version.App)
+	//nolint:prealloc
+	var txsBelowLimit [][]byte
+	for idx, tx := range txs {
+		if len(tx) > maxTxSize {
+			err := fmt.Sprintf("tx size %d bytes at index %d exceeds the application's configured threshold of %d bytes", len(tx), idx, maxTxSize)
```
rootulp, line 45: [uber nit] in some places we call the max tx size a "limit". In other places we call it a "threshold". It would be nice if we always referred to this with the same vocabulary. 

In my opinion it would be sufficient to call it `MaxTxSize` and not use "limit" or "threshold" because the prefix "max" already conveys that.

rootulp (approved):
> I'm a little lost on the rationale for this change. Is there a Github issue or document with more context on the bug that this PR is fixing?

I re-read the [Slack thread](https://celestia-team.slack.com/archives/C067E40LZDG/p1732943002247859) b/c there isn't a GitHub issue for the bug. IMO the Slack thread should be linked in PR description to help reviewers get more context. Better yet a GitHub issue would've really helped. Here's the bug as I understand it:

## Context

https://github.com/celestiaorg/CIPs/blob/main/cips/cip-28.md 

## Problem

The 2 MiB tx limit proposed in CIP-28 doesn't work as expected. Currently it only applies to non-blob txs. Blob txs can be larger than 2 MiB.

## Investigation

Our `CheckTx` wraps the Cosmos SDK base app `CheckTx`. Both blob txs and non-blob txs have the base app `CheckTx` invoked on them. BaseApp `CheckTx` invokes BaseApp `runTx` which invokes `app.anteHandler`. The `app.anteHandler` includes the `MaxTxSizeDecorator`. Our `PrepareProposal` constructs an ante handler that includes the `MaxTxSizeDecorator`. Our `ProcessProposal` also constructs an ante handler that includes the `MaxTxSizeDecorator`. 

The problem with the current implementation of `MaxTxSizeDecorator` is that it gets the inner tx and not the blobs attached because of [these lines](https://github.com/celestiaorg/celestia-app/blob/818026ac446570605d3db582f2d261fb392e1973/app/check_tx.go#L54-L55) in `CheckTx` and [this line](https://github.com/celestiaorg/celestia-app/blob/818026ac446570605d3db582f2d261fb392e1973/app/validate_txs.go#L95) in `PrepareProposal`. The fix in this PR is to take the implementation of the `MaxTxSizeDecorator` and move it into `CheckTx` and `PrepareProposal` before the blobs are split from the blob tx. This PR doesn't do the same in `ProcessProposal` because that would be consensus breaking. 

## Questions

1. The `MaxTxSizeDecorator` unit tests don't include a BlobTx so we didn't catch this at development time. Should we add a unit test for BlobTx if we are keeping MaxTxSizeDecorator around? If not we should deprecate MaxTxSizeDecorator because it doesn't work as expected. Can be a follow-up.
2. Do we need to copy + paste the logic into `PrepareProposal`, specifically `FilterTxs`? Given this now won't be a consensus rule, `CheckTx` seems sufficient because validators that are running unmodified code will reject BlobTxs larger than 2 MiB from entering their mempool and therefore won't be included in a block. IMO we can remove it from `FilterTxs` before merging but not blocking on it.
3. Can we rename the PR title to something a bit more informative? Perhaps: "fix: reject BlobTxs larger than 2 MiB"

ninabarbakadze:
> #4087

1. MaxTxSizeDecorator will be reowrked in

> > I'm a little lost on the rationale for this change. Is there a Github issue or document with more context on the bug that this PR is fixing?
> 
> I re-read the [Slack thread](https://celestia-team.slack.com/archives/C067E40LZDG/p1732943002247859) b/c there isn't a GitHub issue for the bug. IMO the Slack thread should be linked in PR description to help reviewers get more context. Better yet a GitHub issue would've really helped. Here's the bug as I understand it:
> 
> ## Context
> https://github.com/celestiaorg/CIPs/blob/main/cips/cip-28.md
> 
> ## Problem
> The 2 MiB tx limit proposed in CIP-28 doesn't work as expected. Currently it only applies to non-blob txs. Blob txs can be larger than 2 MiB.
> 
> ## Investigation
> Our `CheckTx` wraps the Cosmos SDK base app `CheckTx`. Both blob txs and non-blob txs have the base app `CheckTx` invoked on them. BaseApp `CheckTx` invokes BaseApp `runTx` which invokes `app.anteHandler`. The `app.anteHandler` includes the `MaxTxSizeDecorator`. Our `PrepareProposal` constructs an ante handler that includes the `MaxTxSizeDecorator`. Our `ProcessProposal` also constructs an ante handler that includes the `MaxTxSizeDecorator`.
> 
> The problem with the current implementation of `MaxTxSizeDecorator` is that it gets the inner tx and not the blobs attached because of [these lines](https://github.com/celestiaorg/celestia-app/blob/818026ac446570605d3db582f2d261fb392e1973/app/check_tx.go#L54-L55) in `CheckTx` and [this line](https://github.com/celestiaorg/celestia-app/blob/818026ac446570605d3db582f2d261fb392e1973/app/validate_txs.go#L95) in `PrepareProposal`. The fix in this PR is to take the implementation of the `MaxTxSizeDecorator` and move it into `CheckTx` and `PrepareProposal` before the blobs are split from the blob tx. This PR doesn't do the same in `ProcessProposal` because that would be consensus breaking.
> 
> ## Questions
> 1. The `MaxTxSizeDecorator` unit tests don't include a BlobTx so we didn't catch this at development time. Should we add a unit test for BlobTx if we are keeping MaxTxSizeDecorator around? If not we should deprecate MaxTxSizeDecorator because it doesn't work as expected. Can be a follow-up.
> 2. Do we need to copy + paste the logic into `PrepareProposal`, specifically `FilterTxs`? Given this now won't be a consensus rule, `CheckTx` seems sufficient because validators that are running unmodified code will reject BlobTxs larger than 2 MiB from entering their mempool and therefore won't be included in a block. IMO we can remove it from `FilterTxs` before merging but not blocking on it.
> 3. Can we rename the PR title to something a bit more informative? Perhaps: "fix: reject BlobTxs larger than 2 MiB"

1. `MaxTxSizeDecorator` will be reworked in v4 because removing or changing it is consensus breaking. We're tracking the issue https://github.com/celestiaorg/celestia-app/issues/4087
2. I removed the FilterTxs change but technically you could still modify check tx
3. yes

evan-forbes (approved):

app/validate_txs.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/4084/commits/5cddecc314196fe8e5e1e444c687d16ee502ce56

app/validate_txs.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/4084/commits/5cddecc314196fe8e5e1e444c687d16ee502ce56

rootulp (approved):
Thanks for pushing this through!

Issue #3995, opened 2024-10-21
Consider including v2 upgrade heights in binary for common networks


rootulp:
## Context

We used a `--v2-upgrade-height` flag on networks

## Problem

If a consensus node is block syncing from genesis, it needs to provide the `--v2-upgrade-height`

## Proposal

Consider hard-coding the `--v2-upgrade-height`s for arabica, mocha, mainnet beta so that node operators don't have to specify it.

cmwaters:
I am for this

rootulp:
Note: this will be cumbersome to test b/c we'll need to block sync on Arabica, Mocha, or Mainnet through the v2 upgrade height.

To make this issue easier, we should build a utility that makes it easier + faster to block sync.

PR #4091, opened 2024-12-06
feat: set defaults for `--v2-upgrade-height`

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/3995

If a user specifies the `--v2-upgrade-height` flag then it takes precedence over any defaults. If the user doesn't specify a `--v2-upgrade-height` flag and the chainID is `arabica-11`, `mocha-4`, or `celestia` then the v2 upgrade height is set to the block height that was coordinated on those chains. 

See: https://docs.celestia.org/how-to-guides/network-upgrade-process#lemongrass-network-upgrade

## Testing

1. `./scripts/single-node.sh` doesn't set a v2 upgrade height
2. `./arabica.sh` if we remove the `--v2-upgrade-height` flag in the script, gets defaulted correctly
3. `./arabica.sh` if we override the `--v2-upgrade-height` flag in the script, that gets used


cmd/celestia-appd/cmd/start.go
```diff
@@ -117,6 +118,20 @@ is performed. Note, when enabled, gRPC will also be automatically enabled.
 				return err
 			}
 
+			switch clientCtx.ChainID {
```
rach-id, line 121: if a user specifies no chain ID, what's the default value?

rootulp: This will get logged but no default value will be supplied:

```
				serverCtx.Logger.Info(fmt.Sprintf("No default value exists for the v2 upgrade height when the chainID is %v", clientCtx.ChainID))
```

evan-forbes (approved):

rach-id (approved):

cmwaters (approved):
Thanks for making this change 👍

PR #4113, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


evan-forbes (approved):

rootulp (commented):
IMO this is a style choice. I would prefer if a linter was configured so that the codebase always used `fmt.Errorf` so that this codebase is consistent.

IMO if it's a style choice and it isn't enforced via a linter then I'm inclined to not merge this b/c nothing prevents a future PR from using `fmt.Errorf` without arguments.

PR #4142, opened 2024-12-19
docs: fix URL

goodfirm:
## Overview

The link has expired, replace it with the latest address

### DIFF
--- a/specs/src/proto/types.proto
+++ b/specs/src/proto/types.proto
@@ -4,7 +4,7 @@ import "google/protobuf/timestamp.proto";
 
 // Define data structures.
 
-// https://github.com/celestiaorg/celestia-specs/blob/master/specs/data_structures.md#header
+// https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/data_structures.md#header
 message Header {
   ConsensusVersion version = 1;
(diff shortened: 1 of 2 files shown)


ninabarbakadze (commented):
lint failing

goodfirm:
> lint failing

The reason for the failure is that `doc` is not an available type, but `docs` is.

Modified, please review again.

rootulp (approved):

rootulp:
Note: you may need to run `make proto-gen` and push the changed files to this PR b/c I'm surprised that the docs change didn't manifest in the generated `.pb.go` files yet

goodfirm:
> `make proto-gen`

image

Thank you, I executed this command, but no new files were generated

rootulp:
Oh sorry, on re-read I noticed the only changes were in specs/**/*.proto files so I don't think they will result in any changed generated files. Thanks for the fix!

goodfirm:
> Oh sorry, on re-read I noticed the only changes were in specs/**/*.proto files so I don't think they will result in any changed generated files. Thanks for the fix!

My pleasure.
Thanks for your review!

rach-id (approved):

PR #4157, opened 2024-12-24
refactor: using maps.Clone

pengbanban:
## Overview

using maps.Clone


rootulp:
Doesn't seem necessary.

PR #4159, opened 2024-12-25
chore: fix quotation mark

taozui472:


rootulp (commented):
What was wrong with the previous quotation marks?

taozui472:
The unexpected marks might confuse readers and be distracted by them. Also some editors might not render correctly depending on the software or platform used, appearing as gibberish or causing formatting problems.

rootulp (approved):

ninabarbakadze (approved):

PR #4201, opened 2025-01-09
feat: allow resubmission after unknown status

jcstein:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/rakyll/statik v0.1.7
-	github.com/rs/zerolog v1.33.0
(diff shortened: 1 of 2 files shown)


pkg/user/tx_client.go
```diff
@@ -673,3 +676,23 @@ func QueryNetworkMinGasPrice(ctx context.Context, grpcConn *grpc.ClientConn) (fl
 	}
 	return networkMinPrice, nil
 }
+
+// handleUnknownStatus handles transactions that return an unknown status.
+// It reverts the sequence number and removes the transaction from tracking.
+func (client *TxClient) handleUnknownStatus(txHash string) error {
```
cmwaters, line 682: This is really similar to the handleEvictions function. Maybe we can combine the two and just return different errors.

jcstein:
welp i accidentally closed this as I refreshed my fork

Issue #4204, opened 2025-01-10
`grpc: received message larger than max (14682092 vs. 10485760)" when fetching blocks.`


rootulp:
## Context

Currently the default `app.toml` has:

```toml
# MaxRecvMsgSize defines the max message size in bytes the server can receive.
# The default value is 10MB.
max-recv-msg-size = "10485760"
```

## Problem

> Community APIs (Mocha) seem to be all returning "message": "grpc: received message larger than max (14682092 vs. 10485760)" when fetching blocks.
Example: https://mocha.api.cumulo.me/cosmos/base/tendermint/v1beta1/blocks/4050500
It seems this is after the latest proposal to increase the block size to 8 MiB. This likely requires adjusting the max-recv-msg-size for affected endpoints.
https://github.com/cosmos/gaia/issues/3151

Reported by @qubelabsio

## Proposal
1. Determine a new default value for `max-recv-msg-size`
2. Update the default
3. Ask Mocha node operators to manually bump locally

rootulp:
Copying from Slack

> The default is 10 MiB which I expect is sufficient for 8 MiB blocks.
https://mocha-4.celenium.io/block/4050500?tab=transactions is a 7 MB block but the error message claims it's trying to return a 14 MiB response :thinking_face:
It looks like the response for that endpoint includes txs twice (see [example](https://mocha.api.cumulo.me/cosmos/base/tendermint/v1beta1/blocks/4050499)). So we need to be bump the default to at least 16 MiB, prob 20 MiB in case the other fields are large. Surprised we didn't hit this in network tests.

> I confirmed that our endpoint is unmodified from upstream. For example Cosmos mainnet REST API also has txs twice (see [example](https://cosmos-rest.publicnode.com/cosmos/base/tendermint/v1beta1/blocks/21002943)). Once in sdk_block and once in block.

PR #4205, opened 2025-01-10
feat: increase max-recv-msg-size default to 20 MiB

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4204

Note this won't modify already existing `app.toml` files so node operators need to modify their local `app.toml`. Example command:

```shell
sed -i.bak 's/^max-recv-msg-size = "10485760"/max-recv-msg-size = "20971520"/' ~/.celestia-app/config/app.toml
```

## Testing

```shell
# Before
$ cat ~/.celestia-app/config/app.toml | grep max-recv
max-recv-msg-size = "10485760"

# After
$ ./scripts/single-node.sh
$ cat ~/.celestia-app/config/app.toml | grep max-recv
max-recv-msg-size = "20971520"
```

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -294,5 +294,8 @@ func DefaultAppConfig() *serverconfig.Config {
 	cfg.StateSync.SnapshotInterval = 1500
 	cfg.StateSync.SnapshotKeepRecent = 2
 	cfg.MinGasPrices = fmt.Sprintf("%v%s", appconsts.DefaultMinGasPrice, BondDenom)
+
+	const mebibyte = 1048576
+	cfg.GRPC.MaxRecvMsgSize = 20 * mebibyte
 	return cfg
 }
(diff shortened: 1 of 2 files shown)


app/default_overrides.go
```diff
@@ -294,5 +294,8 @@ func DefaultAppConfig() *serverconfig.Config {
 	cfg.StateSync.SnapshotInterval = 1500
 	cfg.StateSync.SnapshotKeepRecent = 2
 	cfg.MinGasPrices = fmt.Sprintf("%v%s", appconsts.DefaultMinGasPrice, BondDenom)
+
+	const mebibyte = 1048576
```
rach-id, line 298: a super tiny nit for a follow up issue would be to extract the `mebibyte` constant and export it somewhere. We have a lot of redefinitions of it, so it would be good to use the same one to avoid any mistakes

evan-forbes (approved):

PR #4214, opened 2025-01-13
docs: configure-v3

rootulp:
Before the `configure-v3` command didn't show up in Makefile help


rach-id (approved):
utAck

rootulp:
Yes. IMO we should have docs for all the commands

evan-forbes (approved):

Issue #4212, opened 2025-01-13
Spam Mocha and analyze block times


rootulp:
## Context

Mocha increased to 8 MiB blocks.

## Problem

We want to be confident that block times can remain ~6 seconds with full 8 MiB blocks.

## Proposal

1. Spam Mocha via txsim for a couple hours. Use a range of blob sizes. Verify that blocks are near full.
2. After a couple hours, analyze the block times. There exists a Go binary for this.

rootulp:
It's kinda tough to get consistently full blocks b/c the txs land in different blocks. For example logs look like

```
{"level":"info","height":4156070,"address":"celestia1tj5t0chjtv44fs3du4f3mjlsexz7s9v5sckx23","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
{"level":"info","height":4156070,"address":"celestia1gt45dyfjjxtd58l3kjesc83uttk8ewyqljtu65","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
{"level":"info","height":4156072,"address":"celestia1fzkmlt3mee0gv6svpwg90lsmrwdncmmdrw68ac","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
{"level":"info","height":4156073,"address":"celestia10ny9aqs63fg8ysr6vvpfh3mtev9kypejagz0g2","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
{"level":"info","height":4156075,"address":"celestia17s9ampc4u7whp6fu2emjv8y437kjc7yk8254s0","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
{"level":"info","height":4156077,"address":"celestia1n5yl702v2t4273qp4un83cjcxkuqyzelj38sfj","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
{"level":"info","height":4156079,"address":"celestia15cgjxztmxtugmppfa6et6kwuj7z532ttcysl7e","blobs count":"1","total byte size of blobs":1720203,"time":"2025-01-13T15:56:18-05:00","message":"tx committed"}
[...]

PR #4218, opened 2025-01-14
feat(tools): bytes in block tool

rootulp:
Extremely optional tool for https://github.com/celestiaorg/celestia-app/issues/4212

## Testing
```
$ go run main.go
Fetched a total of 1400 blocks (from ~4156770 up to ~4158132).
Average bytes_in_block: 2634636.38 bytes (~2.51 MiB)
Average block_time:     5967.89 ms (~5.97 seconds)
```

another invocation

```
$ go run main.go
Fetched a total of 110 blocks (from ~4170798 up to ~4170898).
Average bytes_in_block: 5617659.89 bytes (~5.36 MiB)
Average block_time:     6761.83 ms (~6.76 seconds)
```

### DIFF
--- a/tools/bytes-in-block/main.go
+++ b/tools/bytes-in-block/main.go
@@ -22,7 +22,6 @@ type Block struct {
 }
 
 func main() {
-	// We want to fetch blocks from height=4156770 up to height=4158208
 	start := 4170798
 	end := 4170898
 	limit := 100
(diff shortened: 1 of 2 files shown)


tools/bytes-in-block/main.go
```diff
@@ -0,0 +1,103 @@
...
+	limit := 100
+
+	var totalBlocks int64
+	var sumBytes int64
+	var sumBlockTime int64
+
+	// We'll loop offset in increments of `limit`.
+	// For each iteration, we do a GET request with the given offset + limit=100
+	// until we've covered the range [start, end).
+	for offset := start; offset <= end; offset += limit {
```
rach-id, line 36: [nit][optional]
rename offset and limit to more understandable names like: `blockRangeStart`, `numberOfBlocks` or similar

rach-id (approved):
I think it's good to keep all kind of tooling we build somewhere 👍 

I would suggest adding a small readme or a godoc describing what this does and how to run it to make it easier to use

rach-id (approved):

ninabarbakadze (approved):

PR #4227, opened 2025-01-15
chore(deps): bump to core v1.45.0-tm-v0.34.35 and cosmos-sdk v1.27.0-sdk-v0.46.16

rach-id:
## Overview

Bump to latest core and cosmos-sdk to support the new block gRPC endpoints


rootulp (approved):

evan-forbes (approved):
I think we just want to be careful not to create a release until we are confident that we are not delaying consensus with rechecks too much

rach-id:
Thanks a lot for the reminder. Will be working on the benchmark next before the release

PR #4238, opened 2025-01-17
docs: gas used and gas price estimation ADR

rach-id:
## Overview

Adds an ADR for CIP-18

### DIFF
--- a/docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
+++ b/docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
@@ -1,0 +1,104 @@
+# ADR 023: Gas used and gas price estimation
+
+## Changelog
+
+- 2025/01/17: Initial draft (@rach-id)
+
+## Status
+
+Proposed
+
+## Context
+
+As per [CIP-18: Standardised Gas and Pricing Estimation Interface](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md), we will provide a base implementation for the gas and fee estimation in celestia-app so it can be used when building transactions.
+
+## Decision
+
+We will proceed with the implementation in celestia-app.
+
+The fee estimation mechanism is TBD.
+
+## Detailed Design
+
+The API will be provided through gRPC following the interface provided in [CIP-18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md).
+
+### Gas used estimation
+
+The gas used estimation will be calculated using the baseapp simulation mode:
+
+```go
+	gasInfo, _, err := app.BaseApp.Simulate(tx)
+	if err != nil {
+		return nil, err
+	}
+```
+
+### Gas price estimation
+
+There are multiple ways to have a gas price estimation for a priority level. For a first implementation, we will use the following:
+
+- High Priority: The gas price is the median of the top 10% of transactions' gas prices from the last 5 blocks.
+- Medium Priority: The gas price is the median of all gas prices from the last 5 blocks.
+- Low Priority: The gas price is the median of the lowest 10% of gas prices from the last 5 blocks.
+- None Priority: This is equivalent to the Medium priority, using the median of all gas prices from the last 5 blocks.
+
+The calculation of the top 10% and bottom 10% will be done using the [standard deviation and z-scores](https://en.wikipedia.org/wiki/Standard_normal_table#Cumulative_(less_than_Z)).
+A z-score or a standard score represents the number of standard deviations a data point is from the mean in a standard normal distribution.
+
+Provided the z-scores table, we can see that:
+
+- High priority: corresponds to `mean + 1.28 standard_deviation` value.
+- Medium priority: is the `mean` value.
+- Low priority: corresponds to `mean - 1.28 standard_deviation`
+- None priority: is the `mean` value.
+
+Note: if the last 5 blocks are all empty, we will return the node's min gas price.
+
+The following is a basic implementation of the standard deviation that we can use:
+
+```go
+// mean calculates the mean value of the provided gas prices.
+func mean(gasPrices []float64) float64 {
+	sum := 0.0
+	for _, gasPrice := range gasPrices {
+		sum += gasPrice
+	}
+	return sum / float64(len(gasPrices))
+}
+
+// standardDeviation calculates the standard deviation of the provided gas prices.
+func standardDeviation(gasPrices []float64) float64 {
+	meanGasPrice := mean(gasPrices)
+	var variance float64
+	for _, gasPrice := range gasPrices {
+		variance += math.Pow(gasPrice-meanGasPrice, 2)
+	}
+	variance /= float64(len(gasPrices))
+	return math.Sqrt(variance)
+}
+```
+
+## Alternative Approaches
+
+There are multiple ways to estimate the gas price. However, a first default implementation can be as basic as described above. 
+
+Better estimations can be done subsequently.
+
+## Consequences
+
+### Positive
+
+- Provide a gas used and gas price estimation for clients to use
+
+### Negative
+
+None.
+
+### Neutral
+
+None.
+
+## References
+
+- [CIP-18: Standardised Gas and Pricing Estimation Interface](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md).
+- [standard deviation and z-scores](https://en.wikipedia.org/wiki/Standard_normal_table#Cumulative_(less_than_Z)).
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
```diff
@@ -0,0 +1,110 @@
...
+
+There are multiple ways to have a gas price estimation for a priority level. For a first implementation, we will use the following:
+
+- High Priority: The gas price is the median of the top 10% of transactions' gas prices from the last 5 blocks.
+- Medium Priority: The gas price is the median of all gas prices from the last 5 blocks.
+- Low Priority: The gas price is the median of the lowest 10% of gas prices from the last 5 blocks.
+- None Priority: This is equivalent to the Medium priority, using the median of all gas prices from the last 5 blocks.
+
+The calculation of the top 10% and bottom 10% will be done using the [standard deviation and z-scores](https://en.wikipedia.org/wiki/Standard_normal_table#Cumulative_(less_than_Z)).
+A z-score or a standard score represents the number of standard deviations a data point is from the mean in a standard normal distribution.
```
rootulp, line 46: [blocking] These two paragraphs describe two different approaches. For the high priority case:
1. Use median of the top 10%
2. Calculate mean + 1.28 * standard_deviation

Those two approaches result in two different results. Can you clarify which one this plans on using?

IMO the second one doesn't make sense if the data isn't normally distributed and we have no guarantees the distribution of gas prices in a block is normally distributed. The first approach is easier to explain and seems "good enough".

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
```diff
@@ -0,0 +1,110 @@
...
+## Detailed Design
+
+The API will be provided through gRPC following the interface provided in [CIP-18](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md).
+
+### Gas used estimation
+
+The gas used estimation will be calculated using the baseapp simulation mode:
+
+```go
+	gasInfo, _, err := app.BaseApp.Simulate(tx)
```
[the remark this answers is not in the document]
rach-id: Thanks. That's a good point to clarify.

From my perspective, the endpoint can be neutral and independent on nodes configuration. It should only reflect what the state machine returns.

However, since this endpoint will be queried from clients, it's up to them to set the multiplier to their desired values.

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+### Gas price estimation
+
+There are multiple ways to have a gas price estimation for a priority level. For a first implementation, we will use the following:
+
+- High Priority: The gas price is the median of the top 10% of transactions' gas prices from the last 5 blocks.
+- Medium Priority: The gas price is the median of all gas prices from the last 5 blocks.
+- Low Priority: The gas price is the median of the lowest 10% of gas prices from the last 5 blocks.
+- None Priority: This is equivalent to the Medium priority, using the median of all gas prices from the last 5 blocks.
```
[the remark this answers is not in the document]
rach-id: None is the default. In case no value is provided, none will be returned so that the priority is optional

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
rach-id: the first paragraph explains that we will take the top 10% and bottom 10% in a high level.

the second paragraph explains how we will do that. We will calculate standard deviation, then use a value from the z-scores table that coincides with the percentile we want to use:
image

which is the value that coincides with top/bottom 10%. 

Alternatively, we will need to implement the percentile approach ourselves, which I prefer not to since it's a bit complicated math and it's better to import it. And to import it, I couldn't find a good library that we can use it from that doesn't have license limitations.

Mean and standard deviation implementations are straightforward. So I guess it's better to use them.

rach-id: Also for a default API, we don't need to be precise of the values. I guess that's for the community to build spot on estimators

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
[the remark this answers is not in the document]
rach-id: added a note here to clarify: 84b48a66c762ed6fea2dad7957508d515d181a6d

evan-forbes (approved):

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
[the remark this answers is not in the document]
rootulp: SGTM

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
rootulp: the first paragraph says:

> - High Priority: The gas price is the median of the top 10% of transactions' gas prices from the last 5 blocks.

which diverges from the paragraph below which says:

> - High priority: corresponds to `mean + 1.28 * standard_deviation` value.

The median of the top 10% != mean + 1.28 * standard_deviation so I still think we should just describe one of those approaches and remove the other one.

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Context
+
+As per [CIP-18: Standardised Gas and Pricing Estimation Interface](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md), celestia-app will provide a base implementation for the gas and fee estimation so it can be used when building transactions.
+
+## Decision
+
+We will proceed with the implementation in celestia-app.
+
+The fee estimation mechanism is TBD.
```
rootulp, line 19: Is this still TBD? This ADR describes two different fee estimation mechanisms so seems like we can remove the TBD.

```suggestion
The fee estimation mechanism is described below.
```

rootulp (dismissed):
Approving to unblock. 

I'm still a little confused on which approach this ADR proposes:
- median of the top 10% 
- mean + 1.28 * standard deviation of gas prices

so seems clearer to remove one of those and describe the one that we'll use

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
rach-id: nice catch, the word median shouldn't be there

```suggestion
- High Priority: The gas price is the price at the start of the top 10% of transactions’ gas prices from the last 5 blocks.
- Medium Priority: The gas price is the median of all gas prices from the last 5 blocks.
- Low Priority: The gas price is the value at the end of the lowest 10% of gas prices from the last 5 blocks.
- None Priority: This is equivalent to the Medium priority, using the median of all gas prices from the last 5 blocks.

The calculation of the top 10% and bottom 10% will be done using the [standard deviation and z-scores](https://en.wikipedia.org/wiki/Standard_normal_table#Cumulative_(less_than_Z)).
A z-score or a standard score represents the number of standard deviations a data point is from the mean in a standard normal distribution.
```

rootulp (approved):

evan-forbes (approved):

docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
```diff
@@ -0,0 +1,113 @@
...
+Clients using this endpoint are responsible for accounting for any necessary multipliers.
+
+### Gas price estimation
+
+There are multiple ways to have a gas price estimation for a priority level. For a first implementation, we will use the following:
+
+- High Priority: The gas price is the price at the start of the top 10% of transactions’ gas prices from the last 5 blocks.
+- Medium Priority: The gas price is the median of all gas prices from the last 5 blocks.
+- Low Priority: The gas price is the value at the end of the lowest 10% of gas prices from the last 5 blocks.
+- None Priority: This is equivalent to the Medium priority, using the median of all gas prices from the last 5 blocks.
```
cmwaters, line 46: Just something to be wary of: this doesn't consider the fullness of the blocks. For example if there is only one transaction in the last 5 blocks, The new pricing will be in direct relation of what the price of that transaction was. Yet if the blocks continue to be empty, the estimation should be the minimum

rach-id: yes, in this case, the returned gas price will be that of that transaction. There are many edge cases we can consider in this and multiple ways to enhance this estimation. 

The most important IMO is to return a gas price estimate > than the minimum, only in the case of the last 5 blocks are all > 50% full or some other number.

But I guess we just want to provide an API for now, even if the numbers are skewed, to lay ground for the community to build more accurate estimation algorithms.

If not, we can rework the estimation mechanism.

PR #4257, opened 2025-01-22
feat: implement the gas estimation API

rach-id:
## Overview

Implements the gas estimation endpoint following: https://github.com/celestiaorg/CIPs/blob/main/cips/cip-18.md

Is there an issue for this or I create one?


app/grpc/gas_estimation/gas_estimator_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			meanVal := Mean(tt.gasPrices)
+			got := StandardDeviation(meanVal, tt.gasPrices)
+			// We'll do a tolerance check for floating-point comparisons.
+			if math.Abs(got-tt.want) > 1e-9 {
+				t.Errorf("stdDev(%v) = %v, want %v", tt.gasPrices, got, tt.want)
+			}
```
rach-id, line 85: this is chatgpt's doing and I think it's cool. we can remove if it looks complicated :D

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -0,0 +1,246 @@
...
+	blobtx "github.com/celestiaorg/go-square/v2/tx"
+	"github.com/cosmos/cosmos-sdk/client"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	gogogrpc "github.com/gogo/protobuf/grpc"
+	coretypes "github.com/tendermint/tendermint/rpc/core/types"
+)
+
+// EstimationZScore is the z-score corresponding to 10% and 90% of the gas prices distribution.
+// More information can be found in: https://en.wikipedia.org/wiki/Standard_normal_table#Cumulative_(less_than_Z)
+const EstimationZScore = 1.28
```
rootulp, line 21: Before we use Z score, IMO we should verify that the gas prices in historical blocks form a normal distribution.

I tried to convey in the ADR that median of top 10% is likely easier and good enough so we don't need to use z scores.

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// extractGasPriceFromTransactions takes a list of transaction results
+// and returns their corresponding gas prices.
+func extractGasPriceFromTransactions(txs []*coretypes.ResultTx) ([]float64, error) {
+	gasPrices := make([]float64, 0)
+	for _, tx := range txs {
+		var feeWithDenom string
+		for _, event := range tx.TxResult.Events {
+			// resp.TxResult.Events[4].Attributes[0].Key
```
rootulp, line 198:
```suggestion
```

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// estimateGasPriceForTransactions takes a list of transactions and priority
+// and returns a gas price estimation.
+// The priority sets the estimation as follows:
+// - High Priority: The gas price is the price at the start of the top 10% of transactions’ gas prices from the last five blocks.
+// - Medium Priority: The gas price is the median of all gas prices from the last five blocks.
+// - Low Priority: The gas price is the value at the end of the lowest 10% of gas prices from the last five blocks.
+// - Unspecified Priority (default): This is equivalent to the Medium priority, using the median of all gas prices from the last five blocks.
+// More information can be found in ADR-023.
```
rootulp, line 172: nit this comment says median but the implementation say mean.

```suggestion
// estimateGasPriceForTransactions takes a list of transactions and priority
// and returns a gas price estimation.
// The priority sets the estimation as follows:
// - High Priority: The gas price is the price at the start of the top 10% of transactions’ gas prices from the last five blocks.
// - Medium Priority: The gas price is the mean of all gas prices from the last five blocks.
// - Low Priority: The gas price is the value at the end of the lowest 10% of gas prices from the last five blocks.
// - Unspecified Priority (default): This is equivalent to the Medium priority, using the mean of all gas prices from the last five blocks.
// More information can be found in ADR-023.
```

buf.lock
```diff
@@ -1,15 +1,18 @@
 # Generated by buf. DO NOT EDIT.
-version: v1
+version: v1beta1
```
rootulp, line 2: what motivated this change?

proto/celestia/core/v1/gas_estimation/gas_estimator.proto
```diff
@@ -0,0 +1,71 @@
...
+  // The gas used is estimated using the state machine simulation.
+  rpc EstimateGasPriceAndUsage(EstimateGasPriceAndUsageRequest) returns (EstimateGasPriceAndUsageResponse) {}
+}
+
+// TxPriority is the priority level of the requested gas price.
+// The following priority levels are defined:
+// - High Priority: The gas price is the price at the start of the top 10% of transactions’ gas prices from the last 5 blocks.
+// - Medium Priority: The gas price is the median of all gas prices from the last 5 blocks.
+// - Low Priority: The gas price is the value at the end of the lowest 10% of gas prices from the last 5 blocks.
+// - Unspecified Priority (default): This is equivalent to the Medium priority, using the median of all gas prices from the last 5 blocks.
```
rootulp, line 35:
```suggestion
// - Medium Priority: The gas price is the mean of all gas prices from the last 5 blocks.
// - Low Priority: The gas price is the value at the end of the lowest 10% of gas prices from the last 5 blocks.
// - Unspecified Priority (default): This is equivalent to the Medium priority, using the mean of all gas prices from the last 5 blocks.
```

rootulp (dismissed):
LGTM. 

One suggestion is to use medians like was originally stated in the ADR (median of top 10%, median, median of bottom 10%).

If we use averages and z-scores, then we should confirm that the gas prices in historical blocks on Mainnet follow a normal distribution.

app/grpc/gasestimation/gas_estimator.go
rach-id: > Before we use Z score, IMO we should verify that the gas prices in historical blocks form a normal distribution.

Having a normal distribution is optional in our case IMO since we don't need spot on values. There are three cases:

- a flat distribution: where the values will generally be close the to the mean which is fine
- a U distribution (I mean just the shape of the distribution forming an U or a W or similar): where we have two cases: either the low prices will be higher than 10%. Or the high prices will be lower than 90%. but I guess that's fine.

> I tried to convey in the ADR that median of top 10% is likely easier and good enough so we don't need to use z scores.

got you 👍  I avoided implementing the percentile because it means implementing a bit of complicated math method and maintaining it and I wanted to avoid that. If we can find a library we can use, that has an open source license, I'll definitely change this implementation.

app/test/gas_estimation_test.go
```diff
@@ -0,0 +1,191 @@
...
+	// test setup: create a test chain, submit a few PFBs to it, keep track of their gas
+	// price, then test the gas estimator API.
+	accountNames := testfactory.GenerateAccounts(150) // using 150 to have 2 pages of txs
+	cfg := testnode.DefaultConfig().WithFundedAccounts(accountNames...).
+		WithTimeoutCommit(10 * time.Second) // to have all the transactions in just a few blocks
+	cctx, _, _ := testnode.NewNetwork(t, cfg)
+	require.NoError(t, cctx.WaitForNextBlock())
+
+	// create the gas estimation client
+	gasEstimationAPI := gasestimation.NewGasEstimatorClient(cctx.GRPCClient)
```
evan-forbes, line 41: these could be different follow upPRs, or addressed here up to you:
- add simple helper function/method in the signer that creates this client, queries for the gas price, and returns it 
- add docs somewhere for this client or that method/function

evan-forbes (approved):

app/test/gas_estimation_test.go
rach-id: sounds good 👍 I'll open an issue and do it once this is merged.

Where do you suggest this to be documented?

rach-id: https://github.com/celestiaorg/celestia-app/issues/4258

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -0,0 +1,245 @@
...
+			return nil, fmt.Errorf("couldn't find fee denom for transaction %s: %s", tx.Hash, feeWithDenom)
+		}
+		fee, err := strconv.ParseFloat(feeWithoutDenom, 64)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't parse fee for transaction %s: %w", tx.Hash, err)
+		}
+		gasPrices = append(gasPrices, fee/float64(tx.TxResult.GasWanted))
+	}
+	return gasPrices, nil
+}
```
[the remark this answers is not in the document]
rach-id: 5946b5108ae34d7e712bc326c1a118df448feb1d

app/test/gas_estimation_test.go
evan-forbes: probably wherever the signer docs are

evan-forbes (approved):

rootulp (approved):
Great work shipping this so quickly!

My question about do historical blocks have gas prices that follow a normal distribution shouldn't block merging this

PR #4265, opened 2025-01-25
fix: Fix incorrect namespace byte length Update adr-014-versioned-nam…

mdqst:
## Overview

I noticed a mismatch in the byte length for `MinReservedNamespace` and `MaxReservedNamespace` in the **Implementation Details** section. According to Option A, the Namespace ID should be 8 bytes when Namespace Version = 0, but the current implementation uses 9 bytes.  

I’ve corrected the values to align with the specification:  
- `MinReservedNamespace` and `MaxReservedNamespace` now use 8 bytes.  
- Adjusted `MinBlobNamespace` and `MaxBlobNamespace` accordingly.

### DIFF
--- a/docs/architecture/adr-014-versioned-namespaces.md
+++ b/docs/architecture/adr-014-versioned-namespaces.md
@@ -145,10 +145,10 @@ Regardless of the option chosen, end-users must explicitly specify the namespace
 
 Since the NMT needs to be aware of the version byte, Option A is equivalent to increasing the namespace ID size to `9` bytes and then constraining the namespaces available for use to namespaces that have a leading `0` byte. In other words:
 
-- MinReservedNamespace: []byte{0, 0, 0, 0, 0, 0, 0, 0, 1}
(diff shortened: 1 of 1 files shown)


rach-id (changes requested):
If I understand correctly, this change is not needed since we keep the first byte to be 0 and the rest 8 bytes are the namespace specific bytes. 

@rootulp can confirm

rootulp:
> According to Option A, the Namespace ID should be 8 bytes when Namespace Version = 0, but the current implementation uses 9 bytes.

The namespace includes the version + ID. So the implementation should use 9 bytes (1 for the version, 8 for the ID).

rach-id:
@mdqst for future contributions, please refer to our good-first-issue issues:

- https://github.com/celestiaorg/celestia-app/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22
- https://github.com/celestiaorg/celestia-core/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

PR #4266, opened 2025-01-25
docs: Fix incorrect proof size calculation in Proposed Non-Interactiv…

0xbryer:
## Overview

An error in the proof size formula where `k/2` and `k/4` were not combined, leading to a misleading representation. The correct simplification should be `3k/4` since `k/2 + k/4 = 3k/4`. This updates the formula to reflect this correction.  

Before:  
```  
Proof size = (k/2 + log(k) + k/4) * NMT-Node size + 2*log(k) * MT-Node size  
```  

After:  
```  
Proof size = (3k/4 + log(k)) * NMT-Node size + 2*log(k) * MT-Node size  
```  

This change ensures the formula is accurate and avoids potential confusion.


rach-id (changes requested):
I don't think any change is needed here, since having the terms separate helps to match each term with its corresponding explanation:

```
Proof size = subtree roots (rows) + subtree roots (last row) + blue nodes (parity shares) + 2 \* blue nodes (`DataRoot`)
```

@nashqueue can confirm.

PR #4267, opened 2025-01-26
chore: Fix offset calculation to correctly skip blocks from startHeight

fragchain:
## Overview

The `offset` parameter wasn't properly accounting for the `startHeight` when skipping blocks.
This change replaces `offset` with `offset-startHeight` to ensure the correct number of blocks are skipped starting from `startHeight`.
This should resolve any inconsistencies in block processing.

### DIFF
--- a/tools/bytes-in-block/main.go
+++ b/tools/bytes-in-block/main.go
@@ -38,7 +38,7 @@ func main() {
 		// you don't exceed the last block. But the endpoint might just return fewer
 		// blocks if you overshoot. In practice, `offset` is just how many items to skip.
 		// We'll keep it simple:
-		url := fmt.Sprintf("https://api-mocha.celenium.io/v1/block?limit=%d&offset=%d&sort=asc&stats=true", numberOfBlocksPerRequest, offset)
+		url := fmt.Sprintf("https://api-mocha.celenium.io/v1/block?limit=%d&offset=%d&sort=asc&stats=true", numberOfBlocksPerRequest, offset-startHeight)
 
 		resp, err := http.Get(url)
 		if err != nil {


tools/bytes-in-block/main.go
```diff
@@ -38,7 +38,7 @@ func main() {
 		// you don't exceed the last block. But the endpoint might just return fewer
 		// blocks if you overshoot. In practice, `offset` is just how many items to skip.
 		// We'll keep it simple:
-		url := fmt.Sprintf("https://api-mocha.celenium.io/v1/block?limit=%d&offset=%d&sort=asc&stats=true", numberOfBlocksPerRequest, offset)
+		url := fmt.Sprintf("https://api-mocha.celenium.io/v1/block?limit=%d&offset=%d&sort=asc&stats=true", numberOfBlocksPerRequest, offset-startHeight)
```
rach-id, line 41: I don't think this is how it's supposed to work. Apparently in this API, this request queries for the blocks between `[offset, offset + numberOfBlocksPerRequest]`. In this case `offset` starts at `startHeight` and keeps increasing by `numberOfBlocksPerRequest` on each iteration until it reaches the end. 

So no change to this call is required.

rach-id: @rootulp can confirm

rach-id (changes requested):

tools/bytes-in-block/main.go
fragchain: [rach-id](https://github.com/rach-id) thank you for your feedback! I see your point, but I’d like to clarify why I believe this change is necessary:  

The current implementation assumes that `offset` always starts at `startHeight` and increments by `numberOfBlocksPerRequest`. However, if `offset` isn't explicitly reset to account for `startHeight`, it can lead to incorrect behavior in scenarios where `startHeight` is modified dynamically or when the API is used in a context where `offset` doesn’t align perfectly with `startHeight`.  

My change ensures that the calculation explicitly accounts for `startHeight`, guaranteeing that the query starts from the intended block range every time. Without this adjustment, there’s a risk of missing or duplicating blocks in edge cases, especially when the API is reused with varying `startHeight` values.  

I’d love to hear your thoughts on this and can provide additional examples if needed.

rootulp: > scenarios where startHeight is modified dynamically

I don't see when or how this can happen. The script works as-is. We want to include blocks starting at start height and querying 100 blocks each time.

Issue #4258, opened 2025-01-22
gas estimation helper function and documentation


rach-id:
these could be different follow upPRs, or addressed here up to you:
- add simple helper function/method in the signer that creates this client, queries for the gas price, and returns it 
- add docs somewhere for this client or that method/function

_Originally posted by @evan-forbes in https://github.com/celestiaorg/celestia-app/pull/4257#discussion_r1925481851_

PR #4270, opened 2025-01-27
feat: add the gas estimation API to the signer

rach-id:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/4258

If there are other places where it would be beneficial to document the API, please let me know to make the corresponding changes.

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -1,15 +1,17 @@
 package user
 
 import (
+	"context"
 	"errors"
 	"fmt"
-
+	"github.com/celestiaorg/celestia-app/v3/app/grpc/gasestimation"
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):

rootulp (approved):
LGTM

rach-id:
this means that the txclient will always need to have an app gRPC endpoint. is this something we want?

vgonkivs:
but it already keeps it https://github.com/celestiaorg/celestia-app/blob/d43ec72a043d55c0f964b356f6f5bae0dc662688/pkg/user/tx_client.go#L138. Or you mean gas estimator endpoint, provided by the user? The node can do it for you and pass it if it is configured.

rach-id:
then let's do it :shipit:

PR #4274, opened 2025-01-27
chore: chainbuilder cosmetics + use PebbleDB for block and state stores

rach-id:
## Overview

adds some cosmetics to the chain builder and switches to use PebbleDB for the block and state stores


evan-forbes (dismissed):

evan-forbes:
closing this PR specifically, but we have a link to it in an open issue to not forget what we did to fix chainbuilder should we need to in the future

PR #4284, opened 2025-02-03
chore: update mempool defaults

evan-forbes:
updates the default mempool values

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -265,8 +265,8 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	cfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
 
-	cfg.Mempool.TTLNumBlocks = 12
-	cfg.Mempool.TTLDuration = 75 * time.Second
+	cfg.Mempool.TTLNumBlocks = 24
+	cfg.Mempool.TTLDuration = 150 * time.Second
 	cfg.Mempool.MaxTxBytes = 2 * mebibyte
 	cfg.Mempool.MaxTxsBytes = 80 * mebibyte
 	cfg.Mempool.Version = "v2" // prioritized mempool
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -36,6 +36,10 @@ import (
 	coretypes "github.com/tendermint/tendermint/types"
 )
 
+const (
+	mebibyte = 1048576
+)
+
 // bankModule defines a custom wrapper around the x/bank module's AppModuleBasic
 // implementation to provide custom default genesis state.
 type bankModule struct {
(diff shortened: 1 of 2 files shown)


app/default_overrides.go
```diff
@@ -263,9 +267,9 @@ func DefaultConsensusConfig() *tmcfg.Config {
 
 	cfg.Mempool.TTLNumBlocks = 12
 	cfg.Mempool.TTLDuration = 75 * time.Second
-	cfg.Mempool.MaxTxBytes = 7_897_088
-	cfg.Mempool.MaxTxsBytes = 39_485_440
-	cfg.Mempool.Version = "v1" // prioritized mempool
+	cfg.Mempool.MaxTxBytes = 2 * mebibyte
+	cfg.Mempool.MaxTxsBytes = 80 * mebibyte
+	cfg.Mempool.Version = "v2" // prioritized mempool
```
evan-forbes, line 272: It shouldn't be a breaking change to set the default to 2MiB since that should already be checked by the antihandler, but there's no reason to keep this at 8

the throughput was quadrupled, and the block times were halved, so the mempool's size was increased

rootulp, line 272:
```suggestion
	cfg.Mempool.Version = "v2" // Content Addressable Transaction (CAT) mempool
```

rootulp: Confirmed this is the CAT mempool. Ref: https://github.com/celestiaorg/celestia-core/blob/ca0c9d7c93a4f66fc2257ff345cb690359060233/consensus/common_test.go#L422-L424

https://github.com/celestiaorg/celestia-core/blob/ca0c9d7c93a4f66fc2257ff345cb690359060233/consensus/common_test.go#L32

rootulp (dismissed):
[optional] in a FLUP we can add a make command that makes it easy for node operators to modify their config in-place.

app/default_overrides_test.go
```diff
@@ -86,9 +85,9 @@ func TestDefaultConsensusConfig(t *testing.T) {
 			// Overrides
 			MaxTxBytes:   7_897_088,
 			MaxTxsBytes:  39_485_440,
-			TTLDuration:  75 * time.Second,
-			TTLNumBlocks: 12,
-			Version:      "v1",
+			TTLDuration:  150 * time.Second,
+			TTLNumBlocks: 24,
```
rootulp, line 89: I expect this test to fail b/c these values diverge from what is specified in `default_overrides.go`

rootulp (approved):

app/default_overrides.go
```diff
@@ -261,11 +265,11 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	cfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
 
-	cfg.Mempool.TTLNumBlocks = 12
-	cfg.Mempool.TTLDuration = 75 * time.Second
-	cfg.Mempool.MaxTxBytes = 7_897_088
-	cfg.Mempool.MaxTxsBytes = 39_485_440
-	cfg.Mempool.Version = "v1" // prioritized mempool
+	cfg.Mempool.TTLNumBlocks = 24
+	cfg.Mempool.TTLDuration = 150 * time.Second
```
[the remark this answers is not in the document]
evan-forbes: ahhh good catch

sorry I was just doubling everything will revert these

rootulp (approved):

rach-id (approved):

ninabarbakadze (approved):

PR #4285, opened 2025-02-03
WIP: spec for PBBT

evan-forbes:

### DIFF
--- a/specs/src/pbbt.md
+++ b/specs/src/pbbt.md
@@ -1,0 +1,356 @@
+ # Vacuum! Part I: High Throughput Recovery
+
+## Summary
+
+High throughput recovery (the second half of block propagation) could be achieved using a "Pull Based Broadcast Tree" (PBBT). PBBT works by enabling the pipelining of `Have` and `Want` messages in otherwise standard pull based gossip, and by distributing the burden of propagation across the broadcaster's peers. Unlike push based broadcast trees, the route to distribute data is generated in a JIT fashion.
+
+Determined attackers can sybil attack both broadcast trees and pull based gossip in time restricted applications such as recovery. To get around these attacks, PBBT relies on preparation (the first part of block propagation), erasure encoding data exchanged during recovery, and an overlay network.
+
+Lastly, we suggest a simpler design that makes better use of the validator overlay network for future consideration.
+
+## Intro
+
+The Celestia protocol will likely separate block propagation into two phases. "Preparation", for distributing data *before* the block is created, and "recovery" for distributing data *after* the block has been created. This document focuses on recovery.
+
+![edited final map](https://hackmd.io/_uploads/ryD-gvZuJl.png)
+
+In order to use the data propagated during the preparation phase, the recovery phase has to use some form of pull based gossip.
+
+Recovery has the following constraints:
+- **100% of the Block data MUST be delivered to >2/3 of the voting power before the `ProposalTimeout` is reached**
+- **MUST be able to utilize the data already distributed during preparation**
+
+
+Given these constraints, we can isolate the different bottlenecks to throughput:
+- **efficiency:** the overhead % of the block the validators must download (0% == downloading the block exactly once)
+- **speed:** the rate at which the block data is delivered to a % of validators
+- **broadcaster's bandwidth:** The % of the block the proposer must upload
+
+Enter the Pull Based Broadcast Tree (PBBT).
+
+PBBT addresses:
+- **efficiency:** by utilizing pull based logic
+- **speed:** by pipelining `Have` and `Want` messages, and by finding the optimal path for distributing data in JIT fashion
+- **broadcaster's bandwidth:** by distributing the burden of propagation to the broadcaster's peers via only sending portions of a block data to each
+
+This document focuses on "recovery", however, both share many mechanisms. Therefore, we first describe the relevant shared components in an abstract fashion.
+
+## Pull Based Block Propagation
+
+PBBT relies heavily on slightly modified pull based gossiping mechanisms. There are four generalized message types.
+
+- `Commitment` is a commitment from the broadcaster over the `Haves` and `Data` that is being distributed.
+- `Have` messages are propagated by nodes to search for the route `Data` messages will take while being distributed (the broadcast route)
+- `Want` messages are used by nodes to indicate missing data and confirm the broadcast route
+- `Data` the actual blobs of data
+
+The only difference between typical pull based block propagation and what is used here, is that the ***`Have` messages can be verified to originate from the broadcaster***. Beyond simply avoiding gossiping `Have` and `Data` messages that do not originate from the broadcaster, this extra verification step unblocks the pipelining of `Have` and `Want` messages.
+
+```mermaid
+sequenceDiagram
+    participant NodeA as Node A
+    participant NodeB as Node B
+    
+    NodeA->>NodeB: COMMITMENT
+    Note over NodeB: Verify COMMITMENT
+    NodeA->>NodeB: HAVE
+    Note over NodeB: Verify HAVE
+    NodeB->>NodeA: WANT
+    NodeA->>NodeB: DATA
+```
+
+**Figure 0: General pull based gossip steps.**
+
+### Distribution Rules
+
+The rules below goal is to facilitate safe pull based propagation.
+
+- **Nodes MUST only propagate validated messages**
+    - `Commitment` messages MUST originate from the Proposer
+    - `Have` messages MUST match those committed to in the `Commitment` message
+    - `Data` messages MUST match their corresponding `Have` and `Commitment`
+- **Nodes MUST Send the `Commitment` message before sending `Have` messages**
+- **Nodes MUST only send `Data` messages if that data has been requested**
+- **Nodes MUST avoid sending the same `Have` or `Want` message to the same peer more than once**
+- **Nodes SHOULD request `Data` they don't have**
+- **Upon first receiving `Commitment` and then `Have` messages, nodes SHOULD broadcast those messages to all peers.**
+
+### Disconnection Rules
+
+- **Nodes MUST disconnect from peers that send invalid messages**
+- **Nodes MUST disconnect from peers that send `Data` messages that have not been requested**
+- **Nodes MUST disconnect from peers that send more than one of the same `Have` or `Want` message**
+
+## Pipelining `Have` and `Want` Messages
+
+Pull based mechanisms are very efficient, but sacrifice speed. Here we describe how to pipeline the `Have` and `Want` messages so that push based mechanisms can get close to the speed of push.
+
+Instead of waiting until data is downloaded to gossip `Have` messages, in PBT, they are gossiped after verifying them.
+
+- **Nodes MUST gossip a `Have` message only if EITHER the `Want` message for that data has been successfully queued to for sending OR the `Data` for the `Have` has already been downloaded**
+
+Keep the other pull based rules in context as well, where nodes cannot send the same message twice or unauthenticated data.
+
+```mermaid
+sequenceDiagram
+    participant NodeA as Node A
+    participant NodeB as Node B
+    participant NodeC as Node C
+    
+    Note over NodeA, NodeC: time 0
+    NodeA->>NodeB: HAVE
+    
+    Note over NodeA, NodeC: time 1
+    NodeB->>NodeA: WANT
+    NodeB->>NodeC: HAVE
+    
+    Note over NodeA, NodeC: time2
+    NodeC->>NodeB: WANT
+    NodeA->>NodeB: DATA
+    
+    Note over NodeA, NodeC: time3
+    NodeB->>NodeC: DATA
+```
+
+**Figure 1: Sequence graph of pipelining. Note the simultaneous transfer of `Want` and `Have`**. Times are added to indicate which operation occurs at essentially the same time.
+
+## Broadcast Tree
+
+In order to increase the "recovery" throughput substantially, the network must consider both speed, efficiency, and how much data the broadcaster must upload.
+
+The first bottleneck for throughput with a single broadcaster is the broadcaster's upload bandwidth.
+
+Where:
+- **T**: Max Theoretical Throughput (e.g., in bytes per second)
+- **B**: Upload bandwidth of the broadcaster (e.g., in bytes per second)
+- **n**: Number of times the block has to be uploaded
+
+```
+T = B / n
+```
+
+Therefore, for recovery, to increase throughput the network should attempt to limit the number of times each proposer is required to upload the block data. 
+
+This can be done simply by the proposer uploading only a portion to each of its peers. Instead of the proposer using its bandwidth to upload the same parts to different peers, the peers can gossip each portion to each other.
+
+
+```mermaid
+sequenceDiagram
+    participant NodeA as Node A
+    participant NodeB as Node B
+    participant NodeC as Node C
+
+    Note over NodeA, NodeC: time 0
+    NodeA->>NodeB: HAVE_0
+    NodeA->>NodeC: HAVE_1
+
+    Note over NodeA, NodeC: time 1
+    NodeB->>NodeA: WANT_0
+    NodeB->>NodeC: HAVE_0
+    NodeC->>NodeA: WANT_1
+    NodeC->>NodeB: HAVE_1
+
+    Note over NodeA, NodeC: time 2
+    
+    NodeA->>NodeB: DATA_0
+    NodeA->>NodeC: DATA_1
+
+    Note over NodeA, NodeC: time 3
+    NodeB->>NodeC: DATA_0
+    NodeC->>NodeB: DATA_1
+```
+
+**Figure 3: Sequence graph of the broadcast tree. Note how portions of the block `Data` are transferred between nodes.** Times are added to indicate which operation occur at essentially the same time.
+
+### Optimal Routing
+
+One of the benefits of generating the route of data on fly instead of pre-planning the route is that the plan can incorporate real time data. 
+
+This inherently includes:
+- network latency
+- as timely as possible info on the node's status 
+- already distributed data 
+
+However, it can also include congestion if the `Have` messages and entered into the send queue in a FIFO fashion as `Data`. This ensures that a haves take longer to be delivered to a congested connection, making it more likely peers will recieve that same `Have` message first from a different peer. 
+
+- **Nodes SHOULD send `Have` and `Data` messages from a FIFO queue**
+- **Nodes SHOULD request `Data` from the first peer that sends them a `Have` message**
+
+Lastly, there's still a chance that a peer could already have all `Data` included in the `Commitment`, recreate the `Have` messages and send them to many nodes in the network extremely quickly. Without additional rules, this would cause that node to become a bottleneck in distribution, therefore we add a new rule to ensure this doesn't occur.
+
+- **Nodes MUST have fewer than `PerPeerConcurrentRequestCap` concurrent requests to a single peer**
+
+If this value is set to 2 for example, then there will be a max of two concurrent requests open to each peer. After the first request is fullfilled, then the second is sent.
+
+
+
+## Security
+
+Currently, >1/3 voting power could halt the network, therefore any new broadcasting mechanism must be held to same standard. Meaning that if 2/3 of the voting power is honest, there should not be attacks that can break the block propagation mechanism either.
+
+The only assumptions that we can make are:
+
+**- 1) Nodes are not eclipse attacked**
+**- 2) 2/3 of the validators are participating as described in this spec**
+
+The first assumption means that there exists *a* path of honest nodes connecting the broadcaster and every other honest node. The second is self explanatory.
+
+Recall the constraints for recovery:
+
+>- **100% of the Block data MUST be delivered to >2/3 of the validators**
+>- **The time to distribute that data MUST be <= the `ProposalTimeout`**
+
+While pull based gossip and broadcast trees are incredibly powerful tools for increasing throughput, it's easy to derive attacks that fail these constraints even with the assumptions. First we discuss those attacks, and then we will discuss strategies for mitigation.
+
+### Denial of Service
+
+The purpose of the strict diconnection rules are to prevent all scenarios where an attacker can flood the network with valid messages. If invalid messages are sent, then the attacker is kicked from their address book.
+
+Currently, there are no known gaps. Please comment if you find one.
+
+### Sybil Attacks
+
+#### Pull Based Gossip
+
+Upon requesting data in a standard pull based protocol, faulty peers can simply not send requested data. Typically, honest nodes add a timeout, after which they will request the data from a different peer. Given our assumption that nodes are not being eclipse attacked, nodes will eventually find an honest peer with the data.
+
+The problem arises when applications have less time to distribute data than they can avoid or remove an attacker. Recovery is such as case. If an attacker wants to disrupt a network, it only needs to slow the distribution of data sufficiently beyond the `ProposalTimeout`. While the network will eventually distribute that data, the throughput will be dramatically decreased until connections to malicious peers are removed.
+
+> Side note: its easy to confuse a congested peer as a malicious one. This makes peer scoring incredibly difficult for pull based systems. If timeouts for scoring or addional requests are too agressive, feedback loops emerge. Congested nodes can hit timeouts, resulting in requesting more duplicate data, which causes more congestion. 
+
+**Conclusions:**
+- Time sensitive applications require additional mechanisms in order to use pull based gossip.
+
+#### Broadcast Trees
+
+Using the most naive design of broadcast trees, where the broadcaster distributes equal amounts of block data to each of its peers, a single faulty peer will break the first constraint of recovery. Even if parity data is added, determined attackers can simply spin up a large majority of nodes in the network and not propagate data.
+
+**Conclusions:**
+1) The usage of broadcast trees requires mechanisms that allow for faulty nodes to exist.
+2) Validators must be able to differentiate between themselves and non-validators. Without this, attackers can also increase the number of nodes in the network and poison the tree.
+
+### Defending Against Attacks
+
+The above attacks all rely on an attacker's ability to add many malicious nodes to the network. Our strategy towards defending against attacks will likely involve mechanisms that prevent this from occurring in the first place, alongside ones that mitigate damage when it does.
+
+#### Erasure Encoding Recovery Data
+
+Distributing erasure encoded data alongside the original data introduces interesting tradeoffs to the attacks above.
+
+```go
+var (
+    // originalData is the block data without erasure encoding
+    originalData = []byte{...}
+    // totalData include parity data
+    totalData = []byte{...}
+)
+
+// e is the erasure encoding factor that specifies the ratio of parity data
+e := len(totalData) / len(originalData) 
+```
+
+While the maximum throughput is now divided by `e`, the network no longer has to distribute 100% of the block to 2/3 of the voting power. Now the constraint is:
+
+- **(100% / e) of the Block data MUST be delivered to >2/3 of the validators**
+
+Alone, distributing parity data during recovery only makes attacks more difficult. Both the pull based and broadcast tree attacks can still occur, they just require more malicious nodes.
+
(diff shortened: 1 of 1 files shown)


specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+
+Currently, there are no known gaps. Please comment if you find one.
+
+### Sybil Attacks
+
+#### Pull Based Gossip
+
+Upon requesting data in a standard pull based protocol, faulty peers can simply not send requested data. Typically, honest nodes add a timeout, after which they will request the data from a different peer. Given our assumption that nodes are not being eclipse attacked, nodes will eventually find an honest peer with the data.
+
+The problem arises when applications have less time to distribute data than they can avoid or remove an attacker. Recovery is such as case. If an attacker wants to disrupt a network, it only needs to slow the distribution of data sufficiently beyond the `ProposalTimeout`. While the network will eventually distribute that data, the throughput will be dramatically decreased until connections to malicious peers are removed.
```
rach-id, line 218: this could halt the network if the timeout is higher than the proposal timeout, right?

specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+  repeated TxMetaData txs       = 4;
+  bytes               signature = 5;
+}
+```
+
+The compact block is signed over by the proposer, and verified by converting to signbytes, and verifying using the proposer's public key to verify the included signature. This siganture is separate from the proposal signature as it is purely related to block propagation, and not meant to be part of the proposal. This allows for block propagation to be backwards compatible with older implementations.
+
+The `BlobMetaData` contains the hash of the PFB for the blob transaction that it commits to, alongside the `size`. `size` is the length of the blobTx in bytes.
+
+The `bp_hash` is generated by taking the merkle root over of each of the blob transactions in `BlobMetaData` and `Have` messasges.
```
rach-id, line 318:
```suggestion
The `bp_hash` is generated by taking the merkle root over each of the blob transactions in `BlobMetaData` and `Have` messages.
```

rach-id (commented):
🎉 🎉 

Awesome work 🚀 

Left some comments to understand more and some cosmetics

specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+In order to use the data propagated during the preparation phase, the recovery phase has to use some form of pull based gossip.
+
+Recovery has the following constraints:
+
+- **100% of the Block data MUST be delivered to >2/3 of the voting power before the `ProposalTimeout` is reached**
+- **MUST be able to utilize the data already distributed during preparation**
+
+Given these constraints, we can isolate the different bottlenecks to throughput:
+
+- **efficiency:** the overhead % of the block the validators must download (0% == downloading the block exactly once)
```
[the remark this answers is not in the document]
evan-forbes: determining upload efficiency is impossible to determine without either setting hard caps, or hardcoding the topology including fixed latency and bandwidth. If we make assumptions around even bandwidth and network latency throughput a random network, if all nodes only download the block once, and the proposer only uploads the block once, then nodes should upload the block ~once.

specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+- **speed:** the rate at which the block data is delivered to a % of validators
+- **broadcaster's bandwidth:** The % of the block the proposer must upload
```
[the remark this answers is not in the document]
evan-forbes: yeah it is common to refer to this as broadcaster overhead in terms of a percent, however note the comment above discussion measuring efficiency for uploads vs downloads for this protocol.
 
 I also think that this is still worth distinguishing between the two. The reason to use pull based gossip is to get efficiency as described here. The reason to use a broadcast tree is to be efficient with a specific participant's bandwidth.

specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+
+Recovery has the following constraints:
+
+- **100% of the Block data MUST be delivered to >2/3 of the voting power before the `ProposalTimeout` is reached**
+- **MUST be able to utilize the data already distributed during preparation**
+
+Given these constraints, we can isolate the different bottlenecks to throughput:
+
+- **efficiency:** the overhead % of the block the validators must download (0% == downloading the block exactly once)
+- **speed:** the rate at which the block data is delivered to a % of validators
```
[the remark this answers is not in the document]
evan-forbes: yeah

specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+- **Nodes MUST only send `Data` messages if that data has been requested**
+- **Nodes MUST avoid sending the same `Have` or `Want` message to the same peer more than once**
+- **Nodes SHOULD request `Data` they don't have**
+- **Upon first receiving `Commitment` and then `Have` messages, nodes SHOULD broadcast those messages to all peers.**
+
+### Disconnection Rules
+
+- **Nodes MUST disconnect from peers that send invalid messages**
+- **Nodes MUST disconnect from peers that send `Data` messages that have not been requested**
+- **Nodes MUST disconnect from peers that send more than one of the same `Have` or `Want` message**
```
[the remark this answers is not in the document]
evan-forbes: definitive imo. I'd rather it break than peer scoring mask problems slowly. 

bugs in the validation logic are almost always vulns, therefore imo its better to break than for peer scoring to temporarily mask the actual problem.

specs/src/recovery.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Pipelining `Have` and `Want` Messages
+
+Pull based mechanisms are very efficient, but sacrifice speed. Here we describe how to pipeline the `Have` and `Want` messages so that push based mechanisms can get close to the speed of push.
+
+Instead of waiting until data is downloaded to gossip `Have` messages, in PBT, they are gossiped after verifying them.
+
+- **Nodes MUST gossip a `Have` message only if EITHER the `Want` message for that data has been successfully queued to for sending OR the `Data` for the `Have` has already been downloaded**
```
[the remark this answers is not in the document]
evan-forbes: go doesn't actually know when something was sent as we don't have full access to the tcp buffers. the only way to do this would be to wait for an ack that the message was received. All we need here is to know that we will send message.

> a malicious node can send multiple HAVE messages for data that's not propagated and all the nodes will be waiting for it

see the security section. this can happen for all pull based logic and similar attacks can be performed on broadcast trees. forwarding a message before its received doesn't actually add any new security considerations beyond normal pull based logic. if timeouts can be used, all data eventually gets transmitted. It is more difficult to attributable of a fault for a timeout, however in theory there exists some peer scoring that eventually prunes malicious peers. Even without pipelining the haves and wants, pull based logic is prone to hit timeouts during congestion. Fortunately, using parity data and overlay network remove the need for peer scoring.

specs/src/recovery.md
evan-forbes: perhaps I can make this clearer, but overlay networks and parity data remove the need for timeouts or peer scoring during recovery.

during preparation however, we will still likely need timeouts.

specs/src/recovery.md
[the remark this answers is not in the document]
rach-id: I see. So we're optimizing for the proposer to only upload the block once, but for the other nodes, they can upload as much as they can?

a separate idea, if the proposer has perfect internet, is well-connected, and has very low latency, why not let it upload more than once? that would speed up propagation.

Maybe we could do something like upload as much as you can, or even use streamed erasure encoding

specs/src/recovery.md
rach-id: >  but overlay networks and parity data remove the need for timeouts or peer scoring during recovery

how does this work?

overlay network allow connecting directly to validators, meaning the data will eventually reach them, and they can vote on it? even with a malicious minority

and parity data allow reconstruction in case a malicious peer withholds the data it received from the proposer?

evan-forbes: sorry, the proposal timeout effectively is the single global timeout. normally a timeout implies that we retry via asking another peer for data (this is what we would do during preparation or using CAT for example). Here we are not retrying for any specific part, because that part is only relevant to that height and round. Instead we retry via a new round, which may or may not be over the exact same block data.

the typical reasoning for timeouts is to handle malicious or slow peers. The overlay network combats malicious nodes by enabling validators to focus on propagating to each other instead of nodes that could be attacking.

parity data stops single validators from attacking

therefore combining the two gets us to the point where 1/3 of the validators need to be faulty to halt the network, which is the status quo

specs/src/recovery.md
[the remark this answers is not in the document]
evan-forbes: > I see. So we're optimizing for the proposer to only upload the block once, but for the other nodes, they can upload as much as they can?

we're letting the optimal routing mechanisms handle the rest yeah exactly. Since the broadcaster is different each height and round, the optimal route of data is different each time. Sometimes one node could end up utilizing more bandwidth than others simply due to the topology, other times they could not.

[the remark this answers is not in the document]
rach-id: makes sense. maybe an optimization would be not to limit uploading the block only once but upload as much as possible. 

In this case, if the node has low connectivity to the network, it can upload even 1/4 (or whichever the erasure encoding coefficient chosen), and be good since the rest of the network can reconstruct the block.

However, if it's very well-connected, it can upload as much as possible.

The idea is not to put any upload limitations to the proposer

specs/src/recovery.md
rach-id: makes sense, thanks 🙏

specs/src/recovery.md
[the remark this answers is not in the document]
evan-forbes: the proposer isn't actually capped at uploading the block once. I agree that this is a really good point to bring up though as its not specified. There could be scenarios where it would be more optimal to download more data directly from the proposer if both the proposer and the downloading peer have available bandwidth.

I think @nashqueue brought up something similar. There likely exists a rule we can add that pairs well with the existing optimal routing rules

>  - Nodes SHOULD send Have and Data messages from a FIFO queue
> - Nodes SHOULD request Data from the first peer that sends them a Have message
> - Nodes MUST have <= `PerPeerConcurrentRequestCap` concurrent requests to a single peer

I think @nashqueue suggested that the proposer could include another `Have` that is appended to the FIFO queue. This `Have` could simply be the entire rest of the block.

This also improves the fault tolerance of broadcast trees and pull based logic significantly, as if a peer is faulty or slow, then the data would naturally route through the next highest node

evan-forbes:
I don't think this is 110% perfect, but imo we can merge just to clear out the PRs. If we need changes can update after

rach-id (approved):

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+encoding data exchanged during recovery, and an overlay network.
+
+Lastly, we suggest a simpler design that makes better use of the validator
+overlay network for future consideration.
+
+## Intro
+
+The Celestia protocol will likely separate block propagation into two phases.
+"Preparation", for distributing data *before* the block is created, and
+"recovery" for distributing data *after* the block has been created. This
```
cmwaters, line 24:
```suggestion
"Recovery" for distributing data *after* the block has been created. This
```
To keep in line with the capitalisation used for Preparation. It would also be nice to capture block creation as an event such that block propagation has three phases: Preparation, Creation, Recovery.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+    NodeC->>NodeB: DATA_1
+```
+
+**Figure 3: Sequence graph of the broadcast tree. Note how portions of the block
+`Data` are transferred between nodes.** Times are added to indicate which
+operation occur at essentially the same time.
+
+### Optimal Routing
+
+One of the benefits of generating the route of data on fly instead of
```
cmwaters, line 215:
```suggestion
One of the benefits of generating the route of data on the fly instead of
```

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+This inherently includes:
+
+- network latency
+- as timely as possible info on the node's status
+- already distributed data
+
+However, it can also include congestion if the `Have` messages and entered into
+the send queue in a FIFO fashion as `Data`. This ensures that a haves take
+longer to be delivered to a congested connection, making it more likely peers
+will recieve that same `Have` message first from a different peer.
```
cmwaters, line 227:
```suggestion
Additionally, if the `Have` and `Data` messages are part of the same FIFO channel, it ensures, when a node is congested that haves take longer to be delivered to a congested connection. This is good as it makes it more likely that peers will receive that same `Have` message first from a different uncongested peer. 
```
Trying to improve readability here

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+  message**
+
+Lastly, there's still a chance that a peer could already have all `Data`
+included in the `Commitment`, recreate the `Have` messages and send them to many
+nodes in the network extremely quickly. Without additional rules, this would
+cause that node to become a bottleneck in distribution, therefore we add a new
+rule to ensure this doesn't occur.
+
+- **Nodes MUST have fewer than `PerPeerConcurrentRequestCap` concurrent requests
+  to a single peer**
```
cmwaters, line 240: 👍  - good thinking, but wouldn't we also want to throttle/stagger it on the server side i.e. only send out x amount of `Have` messages at a time. My concern is that we'd still get congestion to that node if all it's peers (are sending `PerPeerConcurrentRequestCap` requests.

cmwaters: By restricting the amount of haves you're effectively limiting how many branches can be created from your side of the tree.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+attacks.
+
+Currently, >1/3 voting power could halt the network, therefore any new
+broadcasting mechanism must be held to same standard. Meaning that if 2/3 of the
+voting power is honest, there should not be attacks that can break the block
+propagation mechanism either.
+
+The only assumptions that we can make are:
+
+**- 1) Nodes are not eclipse attacked** **- 2) 2/3 of the validators are
```
cmwaters, line 262:
```suggestion
**- 1) Nodes are not eclipse attacked** **- 2) 2/3 of the validators are
```

Can we prevent 1 by ensuring (with the overlay network) that each node is connected to more than 1/3 of voting power to other nodes i.e. this will mean the node is connected to at least one honest node

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+#### Pull Based Gossip
+
+Upon requesting data in a standard pull based protocol, faulty peers can simply
+not send requested data. Typically, honest nodes add a timeout, after which they
+will request the data from a different peer. Given our assumption that nodes are
+not being eclipse attacked, nodes will eventually find an honest peer with the
+data.
+
+The problem arises when applications have less time to distribute data than they
+can avoid or remove an attacker. Recovery is such as case. If an attacker wants
```
cmwaters, line 297:
```suggestion
The problem arises when applications have less time to distribute data than is needed to
bypass or remove an attacker. Recovery is such as case. If an attacker wants
```
Is this what you mean?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+to disrupt a network, it only needs to slow the distribution of data
+sufficiently beyond the `ProposalTimeout`. While the network will eventually
+distribute that data, the throughput will be dramatically decreased until
+connections to malicious peers are removed.
+
+> Side note: its easy to confuse a congested peer as a malicious one. This makes
+> peer scoring incredibly difficult for pull based systems. If timeouts for
+> scoring or addional requests are too agressive, feedback loops emerge.
+> Congested nodes can hit timeouts, resulting in requesting more duplicate data,
+> which causes more congestion.
```
cmwaters, line 307: Ideally, a peer (and the wider network) can manage it's congestion in a way that we can limit the amount of times that congested nodes behave like malicious nodes (there are also behaviours which are only done by malicious nodes i.e. over requesting data).

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+determined attackers can simply spin up a large majority of nodes in the network
+and not propagate data.
+
+**Conclusions:**
+
+1) The usage of broadcast trees requires mechanisms that allow for faulty nodes
+   to exist.
+2) Validators must be able to differentiate between themselves and
+   non-validators. Without this, attackers can also increase the number of nodes
+   in the network and poison the tree.
```
cmwaters, line 328: 👍  - valid points

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+### Defending Against Attacks
+
+In order to prevent attackers from sending `Have` messages to all peers and simply not responding, we use the `PerPeerConcurrentRequestCap` discussed above.
```
cmwaters, line 332: Oh does this work for both sending have messages and want messages?

cmwaters: `PerPeerConcurrentRequestCap` <- this value would be a consensus parameter (i.e. global and part of the state machine)

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+```go
+var (
+    // originalData is the block data without erasure encoding
+    originalData = []byte{...}
+    // totalData include parity data
+    totalData = []byte{...}
+)
+
+// e is the erasure encoding factor that specifies the ratio of parity data
+e := len(totalData) / len(originalData)
```
cmwaters, line 365: Is there some recommended value of e?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+}
+
+// CompactBlock commits to the transaction included in a proposal. It enalbes
+// clients to reuse already downloaded blobs instead of gossiping them all again
+// during the block propagation.
+message CompactBlock {
+  bytes                     bp_hash      = 1;
+  repeated TxMetaData       blobs        = 2;
+  bytes                     signature    = 3;
+  tendermint.types.Proposal proposal     = 4;
```
cmwaters, line 446: Why does the compact block also have the proposal? In the past the proposal and the block have been sent separately

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+  uint32                    last_length  = 5;
```
cmwaters, line 447: What is this used for?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+  bytes  hash  = 1;
+  uint32 start = 2;
+  uint32 end   = 3;
+}
+
+// CompactBlock commits to the transaction included in a proposal. It enalbes
+// clients to reuse already downloaded blobs instead of gossiping them all again
+// during the block propagation.
+message CompactBlock {
+  bytes                     bp_hash      = 1;
```
cmwaters, line 443: is this the block part hash?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+The `bp_hash` is generated by taking the merkle root over of each of the blob
+transactions in `BlobMetaData` and `Have` messasges.
+
+Verification:
+
+- The signature MUST be valid and from the expected proposer for that height and
+  round.
+
+### Have
```
cmwaters, line 470: So a single have could refer to multiple parts?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+- If parity data is also gossiped, then the acceptable number of failures is
+  raised relative to the erasure encoding factor.
+- If an overlay network is added, then validators are able to avoid sybil
+  attacks from non-validators
+- If some portion of the proposed block is already distributed to a majority of
+  validators, then it has already been sufficiently distributed to bypass sybil.
+
+## Messages
+
+### Commitment
```
cmwaters, line 427: Is the CompactBlock the Commitment? Could we just use one name?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+PBBT addresses:
+
+- **efficiency:** by utilizing pull based logic
+- **speed:** by pipelining `Have` and `Want` messages, and by finding the
+  optimal path for distributing data in JIT fashion
+- **broadcaster's bandwidth:** by distributing the burden of propagation to the
+  broadcaster's peers via only sending portions of a block data to each
+
+This document focuses on "recovery", however, both share many mechanisms.
```
liamsi, line 55: Question: "both" refers to recovery and ...? I assume it's Preparation as it is the other phase discussed above but it is a bit unclear.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+- **Nodes MUST disconnect from peers that send more than one of the same `Have`
+  or `Want` message**
+
+## Pipelining `Have` and `Want` Messages
+
+Pull based mechanisms are very efficient, but sacrifice speed. Here we describe
+how to pipeline the `Have` and `Want` messages so that push based mechanisms can
+get close to the speed of push.
+
+Instead of waiting until data is downloaded to gossip `Have` messages, in PBT,
```
liamsi, line 123:
```suggestion
Instead of waiting until data is downloaded to gossip `Have` messages, in PBBT,
```
?
Otherwise the abbreviation PBT is used for the first time here and needs to be introduced somewhere.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+pre-planning the route is that the plan can incorporate real time data. 
+
+This inherently includes:
+
+- network latency
+- as timely as possible info on the node's status
+- already distributed data
+
+However, it can also include congestion if the `Have` messages and entered into
+the send queue in a FIFO fashion as `Data`. This ensures that a haves take
```
liamsi, line 225:
```suggestion
the send queue in a FIFO fashion as `Data`. This ensures that a `Have`s take
```
?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+One of the benefits of generating the route of data on fly instead of
+pre-planning the route is that the plan can incorporate real time data. 
+
+This inherently includes:
+
+- network latency
+- as timely as possible info on the node's status
+- already distributed data
+
+However, it can also include congestion if the `Have` messages and entered into
```
liamsi, line 224: Ah, probably better to rephrase as Callum suggested below instead: https://github.com/celestiaorg/celestia-app/pull/4285/files#r2152219493

```suggestion
However, it can also include congestion if the `Have` messages are entered into
```
?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+While pull based gossip and broadcast trees are incredibly powerful tools for
+increasing throughput, it's easy to derive attacks that fail these constraints
+even with the assumptions. First we discuss those attacks, and then we will
+discuss strategies for mitigation.
+
+### Denial of Service
+
+The purpose of the strict diconnection rules are to prevent all scenarios where
+an attacker can flood the network with valid messages. If invalid messages are
```
liamsi, line 281: Shouldn't this be "flood the network with valid and invalid messages"?

valid messages cases:
- Nodes MUST disconnect from peers that send Data messages that have not been requested
- Nodes MUST disconnect from peers that send more than one of the same Have or Want message

invalid messages case:

- Nodes MUST disconnect from peers that send invalid messages

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+### Defending Against Attacks
+
+In order to prevent attackers from sending `Have` messages to all peers and simply not responding, we use the `PerPeerConcurrentRequestCap` discussed above.
+
+```go
+// the largest number of validators to compose 1/3 of the voting power. If we
+// assume even voting power percentage, then that is simply 1/3 of the number of
+// validators.
+faultyValidators := max(1, validatorSet.Size() / 3)
```
liamsi, line 338: In practice this will still be stake weighted, right?

Also, would call this var sth like maxAllowedFaultyVals or sth like that.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+// the faultyPartCount is determined by amount of parity data added
+faultyPartCount = math.Abs(parts.Total() - parityParts.Total())
+
+concurrentPartLimit = faultyPartCount / faultyValCount
```
liamsi, line 343: Is faultyValidators == faultyValCount ?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+- **(100% / e) of the Block data MUST be delivered to >2/3 of the validators**
+
+Alone, distributing parity data during recovery only makes attacks more
+difficult. Both the pull based and broadcast tree attacks can still occur, they
+just require more malicious nodes.
+
+While additional mechanisms are still needed, adding parity data to the recovery
+process is a powerful tool to get to our goal:
+
+>Its important that new broadcasting mechanisms do not introduce a lower barrier
```
liamsi, line 380:
```suggestion
>It is important that new broadcasting mechanisms do not introduce a lower barrier
```

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+  network**
+
+We do not specify an overlay network further here. The benefit of an overlay
+network is simply that validators can avoiding sybil nodes to the best of their
+knowledge.
+
+If we can make the attacks require >= 1/3 of the voting power, arguably we have
+achieved our original goal.
+
+>Its important that new broadcasting mechanisms do not introduce a lower barrier
```
liamsi, line 397:
```suggestion
>It is important that new broadcasting mechanisms do not introduce a lower barrier
```

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+The more voting power that has signaled that they have a transaction, then the
+more difficult it is to stop the distribution of that transaction.
+
+### Synthesis
+
+Alone, none of the mechanisms listed above are sufficient for preventing sybil
+attacks. However, when we combine them we can meet our original goal:
+
+>Its important that new broadcasting mechanisms do not introduce a lower barrier
```
liamsi, line 415:
```suggestion
>It is important that new broadcasting mechanisms do not introduce a lower barrier
```

specs/src/recovery.md
liamsi: Assuming this is the same erasure coding we use currently? In that case `e` would be 4 (as `len(totalData)==4*len(originalData)`

liamsi (commented):
This is a beautiful and thought through design! 🚀 💜 I left some comments while reading through it but they shouldn't be blocking anything really.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+
+Verification:
+
+- The part MUST exist for the compact block of that height and round.
+
+### Data
+
+```proto
+// RecoveryPart is a wire message for sending a portion of the block to a peer.
+message RecoveryPart {
```
liamsi, line 514: Is there / will there be a MaxPartSize of some sorts? Or parts somehow guaranteed to have the same size (as with good old tendermint block parts)?

specs/src/recovery.md
liamsi: This is likely to contain height, round, and pol_round but also not sure.

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+message PartMetaData {
+  uint32 index = 1;
+  bytes  hash  = 2;
+}
+
+// HaveParts is a message sent by a peer to inform that it has or will receive a
+// part of a block.
+message HaveParts {
+  int64                 height = 1;
+  int32                 round  = 2;
```
liamsi, line 483: polRound isn't necessary because we are past that stage?

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+  repeated PartMetaData parts  = 3;
```
liamsi, line 484: Is it important to cap this s.t. a HaveParts with too many indices would immediately be rejected? 🤔

specs/src/recovery.md
```diff
@@ -0,0 +1,525 @@
...
+}
+```
+
+Verification:
+
+- The merkle proof MUST be verified using the `bp_hash` in the `CompactBlock`
+  for that height and round.
+
+### Want
```
liamsi, line 493: Would it make sense for peers to be able to "cancel" previously sent wants? e.g. if they received the data through a faster route / node somehow?

specs/src/recovery.md
cmwaters: Ok but does that mean we only send the compact block (not a separate proposal). And if it's just for those fields, could those fields be added directly instead of the entire proposal struct?

evan-forbes:
merging to clear this out, but looks like there are still some flups so will have to go back through those

PR #4299, opened 2025-02-05
test(cip-29): reduce inflation

rootulp:
cc: @tac0turtle


rootulp:
Discussed w/ @tac0turtle offline and this PR currently assumes that the constants for initial inflation rate and disinflation rate decreased by 33% but the CIP is actually suggesting retaining the constants as is and then decreasing them by 33% when calculating inflation. It means we can't use this formula any more:

```
InitialInflationRate * ((1 - DisinflationRate) ^ YearsSinceGenesis)
``` 

or rather the formula needs a conditional where it selectively switches based on the app version.

rootulp:
@tac0turtle when I use the reference implementation in CIP-29, the inflation numbers diverge from what is presented in the table. I'm observing

```
		testCases := []testCase{
			{0, 0.0536}, // note this value won't be used in production because celestia-app was on app version 3 in year 0.
			{1, 0.04824},
			{2, 0.040507128},
			{3, 0.0364564152},
		}
```

when the table says https://github.com/celestiaorg/CIPs/pull/249/files#diff-5bc579390f911ecfc45fde43055ce471aea57812b146e1669a507682deb3ef15R57-R59

rootulp:
Closing for now. @tac0turtle feel free to include this in the PR that implements CIP-29.

PR #4305, opened 2025-02-06
feat(pkg/user): Implement gas estimation in TxClient. Fix issue #4282

### LINKED ISSUE #4282: use the gas estimation when building the transaction
Hey, are there any plans to rework `EstimateGas` in TxClient using the new functionality? I had an impression that the signer's responsibility is to build and sign the transactions. IMO, the most effective way of interacting with this API is through the TxClient, since it already has a relevant method.

_Originally posted by @vgonkivs in https://github.com/celestiaorg/celestia-app/issues/4270#issuecomment-2627070041_

VolodymyrBg:
Closes #4282

This PR implements gas estimation functionality in the TxClient when building transactions. The changes include:

1. Added a new `estimateAndSetGas` method to handle gas estimation and setting in the transaction builder
2. Updated the `BroadcastTx` method to use gas estimation by default when no explicit gas limit is provided
3. Added comprehensive tests for the new gas estimation functionality

Key changes:
- Gas estimation is now performed automatically when no explicit gas limit is provided
- Existing behavior is preserved when gas limit is explicitly set through options
- Added unit tests to verify both automatic gas estimation and explicit gas setting scenarios

The implementation ensures that:
- Gas estimation is performed efficiently using the existing `estimateGas` method
- Transaction fees are calculated correctly based on the estimated gas
- The process is thread-safe with proper mutex locking
- Existing functionality for custom gas limits is maintained through TxOptions

Testing:
- Added new test cases in `tx_client_test.go`
- Verified both automatic gas estimation and explicit gas setting scenarios
- Ensured backward compatibility with existing transaction building process

### DIFF
--- a/pkg/user/tx_client.go
+++ b/pkg/user/tx_client.go
@@ -310,56 +310,47 @@ func (client *TxClient) SubmitTx(ctx context.Context, msgs []sdktypes.Msg, opts 
 	return client.ConfirmTx(ctx, resp.TxHash)
 }
 
+// estimateAndSetGas estimates the gas for a transaction and sets it in the builder
+func (client *TxClient) estimateAndSetGas(ctx context.Context, txBuilder client.TxBuilder, opts *TxOptions) error {
+	// If gas is already set in options, skip estimation
+	if opts != nil && opts.Gas != 0 {
+		txBuilder.SetGasLimit(opts.Gas)
+		return nil
+	}
+
+	// Estimate gas
+	gasLimit, err := client.estimateGas(ctx, txBuilder)
+	if err != nil {
+		return fmt.Errorf("estimating gas: %w", err)
+	}
+
+	// Set the estimated gas limit
+	txBuilder.SetGasLimit(gasLimit)
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


ninabarbakadze (commented):
Thanks! Let's fix CI and I'll take a look

vgonkivs:
Thanks for your contribution, @VolodymyrBg. Unfortunately, this is not exactly what I meant in my comment. Let me clarify.

The current estimation process is inefficient because it relies on SimulateTx, which includes a round trip. To improve efficiency, the app team has implemented the [GasEstimator service](https://github.com/celestiaorg/celestia-app/pull/4270). This service connects to a third-party estimator that provides GasPrice and Gas.

By default, every consensus node runs this service. However, if the user wants to query another node, they should be able to provide a gRPC connection. Only if the GasEstimator fails should we fall back to SimulateTx as the default solution.

Additionally, the user should have the option to specify a maximum gas price (by default, minGasPrice * 100). If the calculated gas price exceeds this limit, the transaction should not be executed.

Question: What is TxOptions? I couldn't find any references to it in your code.

PR #4315, opened 2025-02-10
docs: ADR for High Throughput Recovery

evan-forbes:
## Overview

still a draft for now as we'll add more details as the implementation comes along


evan-forbes:
closing this PR for now, if we want an ADR we can go through some of the implementation of the prop reactor and document the reasoning for it

PR #4332, opened 2025-02-14
docs: fix typo

tomasandroil:
- **"dependant"** → **"dependent"**  
- **"medata"** → **"metadata"**  

#### File modified:  
- `local_devnet/celestia-app/app.toml`


rootulp (approved):

ninabarbakadze (approved):

rach-id (commented):
this is just a generated file. Maybe we should fix this in this cosmos-sdk so all generated files are valid

PR #4333, opened 2025-02-14
fix prepare_proposal_test.go

ruslan0012:
There is a typo in the character’s name. The correct spelling is “carmen san diego” (a well-known franchise), not “carmon san diego”.


rootulp:
This typo doesn't matter.

PR #4340, opened 2025-02-18
fix(txsim): improved grpc server discovery, added default keypath value

### LINKED ISSUE #4223: txsim nice to haves
## Context

I've been using txsim to spam Mocha

## Problems

1. txsim expects a `--key-path` as a CLI argument. We could make it optional and by default look in `~/.celestia-app`
2. txsim populates N accounts with 1000 TIA which is wasteful. Using the `--feegrant` flag eliminates this problem but that's not mentioned in the example. Let's make the example command include the `--feegrant` flag
3. If you provide an incorrect GRPC endpoint, you get this confusing error:

```
{"level":"error","error":"error getting balance for celestia1rmqkyypuly8u5cp7ggrt5cj788887w2tadcvyu: rpc error: code = Unavailable desc = name resolver error: produced zero addresses","account":"tx-sim-9","time":"2025-01-14T14:13:00-05:00","message":"error getting initial account balance"}
```

Instead it should return an error that the GRPC endpoint isn't accessible

4. It's difficult to actually fill blocks on a testnet. It would be nice if txsim had a command to fill blocks so that users don't have to fiddle with the flags `--blob`, `--blob-amounts`, `--blob-sizes`

zatarx:
## Overview
Fixes #4223

1. Added default key path lookup.
2. Added `--feegrant` flag to the example of txsim.
3. `buildGRPCConn` now filters out most of the invalid URLs, but if for example google.com:443 is provided, then it passes thru the function. IMO, healthcheck grpc endpoint must be introduced to solve this item (mb thru a separate issue/adr).
4. @rootulp Could you elaborate on having an extra command for blobs? You mentioned filling up the blocks, but I didn't quite get that. Should it dynamically produce a PFB number, blob amount per PFB and blob sizes based on some network parameters and then spam the chain?
5. grpc urls with 433 port(tls support) after introducing `buildGrpcConn` aren't failing anymore.


rootulp:
> @rootulp Could you elaborate on having an extra command for blobs? You mentioned filling up the blocks, but I didn't quite get that. Should it dynamically produce a PFB number, blob amount per PFB and blob sizes based on some network parameters and then spam the chain?

Yes. I want a way to more easily fill blocks to their max capacity. Currently Celestia blocks can contain 8 MiB of blob data but each individual blob is limited to ~2 MiB. This means that a user needs to submit 4 x ~2 MiB (or slightly smaller) blobs in order to fill a block. Even when I tried doing that via txsim, I wasn't able to consistently fill blocks so there may be other issues needed to accomplish my goal. You can skip this and/or extract it to a new issue so that it doesn't block the work in this PR.

rootulp (dismissed):
Overall LGTM but I haven't tested it manually.

rootulp (approved):
Tested the default keypath lookup and `--feegrant` in the example command. Both work.

zatarx:
I noticed that test account keyring backend is `memory`, addressing it right now

rootulp (dismissed):

zatarx:
Looking at the ci right now

zatarx:
@rootulp could you trigger the ci please? If there's anything else needs to be done within this PR, lmk

rootulp:
Triggered

rootulp (approved):

ninabarbakadze (approved):
thanks!

PR #4344, opened 2025-02-18
revert: default mempool to prioritized (v1)

rootulp:
We got reports from @mindstyle85 that the v2 mempool caused issues for users so we're reverting the default back to the prioritized mempool v1.


evan-forbes (approved):

rach-id (approved):
LGTM. 

Question:

this will only work if we're initializing a new node. Should we provide a way for nodes to automatically switch to v1 mempool?

rootulp:
> this will only work if we're initializing a new node. Should we provide a way for nodes to automatically switch to v1 mempool?

Good point. We never introduced a way for user's to switch to v2 mempool so IMO we don't need one to go back to v1. When we roll-out the switch to v2, we can optionally include `make` commands to modify the config in-place.

PR #4347, opened 2025-02-18
docs: fix typo

Bugarim:
indicated - indicates
represent - represents

thanks devs

### DIFF
--- a/app/grpc/tx/tx.pb.go
+++ b/app/grpc/tx/tx.pb.go
@@ -80,7 +80,7 @@ type TxStatusResponse struct {
 	Index  uint32 `protobuf:"varint,2,opt,name=index,proto3" json:"index,omitempty"`
 	// execution_code is returned when the transaction has been committed
 	// and returns whether it was successful or errored. A non zero
-	// execution code indicates an error.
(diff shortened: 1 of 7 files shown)


app/grpc/tx/tx.pb.go
```diff
@@ -80,7 +80,7 @@ type TxStatusResponse struct {
 	Index  uint32 `protobuf:"varint,2,opt,name=index,proto3" json:"index,omitempty"`
 	// execution_code is returned when the transaction has been committed
 	// and returns whether it was successful or errored. A non zero
-	// execution code indicated an error.
+	// execution code indicates an error.
```
rach-id, line 83: this should be fixed in the proto file

rach-id (changes requested):

rach-id:
@Bugarim fix the PR title to follow the conventional commits spec https://www.conventionalcommits.org/en/v1.0.0/ .

rach-id:
fix the commit history

rootulp:
@rach-id I think we need to update the the `typos` branch with the latest change from `main` so that they don't appear in this diff. Will try

PR #4352, opened 2025-02-21
chore: fixed bash array syntax issue

letshard:
## Overview

I noticed an issue in the loop where the array elements were not properly quoted, which could lead to problems if any of the package paths contained spaces. The correct syntax is to use `"${PKGS[@]}"` instead of `${PKGS[@]}`.


rach-id (approved):

letshard:
[rach-id](https://github.com/rach-id), please merge changes.

rootulp:
None of the packages contain spaces. This seems unnecessary.

PR #4356, opened 2025-02-25
fix: added missing semicolons

hexcow:
## Overview

Added missing semicolons at the end of elements in .proto files to fix a compilation error.
This aligns with the Protocol Buffers syntax requirements.


rach-id (changes requested):
Please verify your changes before opening a PR!!!!!!

PR #4365, opened 2025-03-01
docs: fix typo

sky-coderay:
`and TxClientVersion` - deleted double space


rach-id:
doesn't add any value

PR #4366, opened 2025-03-03
chore: fix issue with incorrect use of single quotes in sed command

hexcow:
## Overview

I’ve updated the sed command where single quotes were incorrectly used around strings containing variables. This could cause unexpected behavior, as variables were not being correctly interpolated.

I’ve replaced the single quotes with double quotes to ensure the variables are properly handled during substitution.


rach-id (commented):
Using a single quote makes it easier to read the regex and doesn't require escaping it. So this is not needed. Closing

Issue #4370, opened 2025-03-03
UnpackAny limit exceeded for historical txs


rootulp:
## Context

> Hey! We think there might be a bug in Cosmos SDK. A few months ago, a constant MaxUnpackAnySubCalls‎ = 100 was added, which - if we understand correctly - sets a limit on the number of calls inside MsgExec. However, in this transaction ([link](https://celenium.io/tx/999b89260b384256f32722f7a1c54b319889cc6dd34c73b99ec1f47e560c5b66?tab=events)), there seem to be around 175 calls. As a result, when trying to parse the transaction, we get a call limit exceeded error. For now, we’ve rolled back to version v1.25.0-sdk-v0.46.16, where this change wasn’t present. But if there are breaking changes in the future, we won’t be able to re-index using the newer versions. Is this expected behavior or a bug? And what would be the best way to handle this?

From @vvuwei

## Problem

[v1.25.1-sdk-v0.46.16](https://github.com/celestiaorg/cosmos-sdk/releases/tag/v1.25.1-sdk-v0.46.16) included a fix that addressed a security vulnerability but celestia-app releases with that change won't be able to process block [7013](https://www.mintscan.io/celestia/block/7013) because [this tx](https://www.mintscan.io/celestia/tx/999b89260b384256f32722f7a1c54b319889cc6dd34c73b99ec1f47e560c5b66?height=7013) has more than 100 UnpackAny calls. This effectively breaks single-binary syncs.
 
## Options

1. Identify txs included on-chain with a UnpackAny >= 100. If it's not that many, consider modifying the constant in celestia-app v4 to handle the max UnpackAnys already included on-chain.
2. Consider wrapping https://github.com/celestiaorg/cosmos-sdk/pull/422 in a version gate so that it only applies to app version >= 3
3.  Consider wrapping https://github.com/celestiaorg/cosmos-sdk/pull/422 in a block height gate so that it only applies to block heights >= N

tac0turtle:
height gating seems sufficient. You shouldnt need to worry about the vulnerability in the historical context

rootulp:
Height based gating has the downside that we must add height constants for each supported chain-id (arabica, mocha, mainnet beta). If we pursue option 2 then we can avoid any chain-id specific logic.

[...]

PR #4373, opened 2025-03-03
feat: version gate MsgExec limit

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4370

### DIFF
--- a/cmd/celestia-appd/cmd/app_server.go
+++ b/cmd/celestia-appd/cmd/app_server.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/celestiaorg/celestia-app/v3/app"
 	"github.com/celestiaorg/celestia-app/v3/app/encoding"
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
+	v3 "github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v3"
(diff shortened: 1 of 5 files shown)


cmd/celestia-appd/cmd/app_server.go
```diff
@@ -43,12 +46,27 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOptions
...
+	var encodingConfig encoding.Config
+	if appVersion < v3.Version {
+		// For versions 1 and 2, use the standard encoding config without recursion limits
+		logger.Info("Using standard encoding config without recursion limits", "app_version", appVersion)
+		encodingConfig = encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	} else {
+		// For version 3+, use the versioned encoding config with recursion limits
+		logger.Info("Using versioned encoding config with recursion limits", "app_version", appVersion)
+		encodingConfig = encoding.MakeVersionedConfig(appVersion, app.ModuleEncodingRegisters...)
+	}
```
cmwaters, line 62: would this work as expected? If I only start up once, then I would be using the encodingConfig for version 1 even if I eventually sync up to a height with app version 3 or greater

rootulp: I had the same concern. This needs testing

PR #4392, opened 2025-03-08
fix: correct typo in image filename (effizicient -> efficient)

leopardracer:
fix: correct typo in image filename

effizicient -> efficient


rootulp:
Yes the filename looks like a typo but the weird submodule seems undesirable. Also if the filename was referenced in docs as the previous name then we also need to update that.

PR #4394, opened 2025-03-09
chore: Fix precision loss in time conversion

0xkazak:
## Overview

Noticed a potential precision loss when converting `diffInSeconds` to `time.Duration`. The previous approach used `fmt.Sprintf("%.0fs", diffInSeconds)`, which could introduce rounding issues. Replaced it with a direct multiplication by `time.Second` to ensure accuracy.

### DIFF
--- a/tools/blocketa/main.go
+++ b/tools/blocketa/main.go
@@ -72,7 +72,7 @@ func Run() error {
 	}
 	diffInBlockHeight := targetBlockHeight - currentHeight
 	diffInSeconds := blockTime * float64(diffInBlockHeight)
-	diffInTime, err := time.ParseDuration(fmt.Sprintf("%.0fs", diffInSeconds))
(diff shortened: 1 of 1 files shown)


tools/blocketa/main.go
```diff
@@ -72,7 +72,7 @@ func Run() error {
 	}
 	diffInBlockHeight := targetBlockHeight - currentHeight
 	diffInSeconds := blockTime * float64(diffInBlockHeight)
-	diffInTime, err := time.ParseDuration(fmt.Sprintf("%.0fs", diffInSeconds))
+	diffInTime := time.Duration(diffInSeconds * float64(time.Second))
 	if err != nil {
 		return err
 	}
```
rootulp, line 78: If there's no error then there is no need for error handling below.

rootulp (changes requested):
IMO the loss of precision is not a concern in this utility tool

PR #4396, opened 2025-03-10
fix: correct typo in image filename (effizicient -> efficient)

leopardracer:
#4392 Reopened


rach-id (approved):

rootulp (changes requested):
What is this? It shouldn't be included

Screenshot 2025-03-10 at 9 48 37 AM

rootulp (changes requested):
Still has a submodule commit

PR #4399, opened 2025-03-10
refactor: remove duplicate `SetMinRetainBlocks` invocation

alingse:
## Overview

I am verifying some lint idea in my project see https://github.com/alingse/sundrylint/pull/10
it report the function call with repeat args from a sub-function call.

we think the sub-function call is heavy, so the repeat args probably a copy-and-paste mistake.

I check the top 4000 github repos, and found this repo has this case.

### DIFF
--- a/cmd/celestia-appd/cmd/app_server.go
+++ b/cmd/celestia-appd/cmd/app_server.go
@@ -57,7 +57,6 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOptions
 		baseapp.SetMinRetainBlocks(cast.ToUint64(appOptions.Get(server.FlagMinRetainBlocks))),
 		baseapp.SetHaltHeight(cast.ToUint64(appOptions.Get(server.FlagHaltHeight))),
 		baseapp.SetHaltTime(cast.ToUint64(appOptions.Get(server.FlagHaltTime))),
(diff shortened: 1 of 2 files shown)


cmd/celestia-appd/cmd/app_server.go
```diff
@@ -57,7 +57,6 @@ func NewAppServer(logger log.Logger, db dbm.DB, traceStore io.Writer, appOptions
 		baseapp.SetMinRetainBlocks(cast.ToUint64(appOptions.Get(server.FlagMinRetainBlocks))),
 		baseapp.SetHaltHeight(cast.ToUint64(appOptions.Get(server.FlagHaltHeight))),
 		baseapp.SetHaltTime(cast.ToUint64(appOptions.Get(server.FlagHaltTime))),
-		baseapp.SetMinRetainBlocks(cast.ToUint64(appOptions.Get(server.FlagMinRetainBlocks))),
```
rootulp, line 60: wow I think the linter is correct b/c this line is repeated 3 lines above

rootulp (approved):

cmd/celestia-appd/cmd/app_server.go
alingse: yes, it's check the func call with repeat args

rach-id (approved):

PR #4411, opened 2025-03-13
feat: update the gas estimation mechanism to use the mempool

rach-id:
## Overview

updates the gas estimation mechanism to rely on the mempool. 

also, uses percentiles for estimation instead of z-scores

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -2,10 +2,11 @@ package app
 
 import (
 	"fmt"
-	"github.com/celestiaorg/go-square/v2/share"
 	"io"
 	"slices"
 	"time"
+
+	"github.com/celestiaorg/go-square/v2/share"
 
 	"github.com/celestiaorg/celestia-app/v3/app/grpc/gasestimation"
 
--- a/app/app.go
+++ b/app/app.go
@@ -766,7 +766,7 @@ func (app *App) getGovMaxSquareBytes() (uint64, error) {
 		return 0, err
 	}
 	maxSquareSize := app.BlobKeeper.GovMaxSquareSize(ctx)
-	return uint64(maxSquareSize * maxSquareSize * share.ContinuationSparseShareContentSize), nil
+	return maxSquareSize * maxSquareSize * share.ContinuationSparseShareContentSize, nil
 }
 
 // RegisterTendermintService implements the Application.RegisterTendermintService method.
--- a/app/app.go
+++ b/app/app.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"fmt"
+	"github.com/celestiaorg/go-square/v2/share"
 	"io"
 	"slices"
 	"time"
@@ -755,7 +756,16 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, _ config.APIConfig) {
 func (app *App) RegisterTxService(clientCtx client.Context) {
 	authtx.RegisterTxService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.BaseApp.Simulate, app.interfaceRegistry)
 	celestiatx.RegisterTxService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
-	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.txConfig.TxDecoder(), app.BaseApp.Simulate)
+	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.txConfig.TxDecoder(), app.getGovMaxSquareBytes, app.BaseApp.Simulate)
+}
+
+func (app *App) getGovMaxSquareBytes() (uint64, error) {
+	ctx, err := app.CreateQueryContext(app.LastBlockHeight(), false)
+	if err != nil {
+		return 0, err
+	}
+	maxSquareSize := app.BlobKeeper.GasPerBlobByte(ctx)
+	return uint64(maxSquareSize * maxSquareSize * share.ContinuationSparseShareContentSize), nil
 }
 
 // RegisterTendermintService implements the Application.RegisterTendermintService method.
--- a/app/app.go
+++ b/app/app.go
@@ -765,7 +765,7 @@ func (app *App) getGovMaxSquareBytes() (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
-	maxSquareSize := app.BlobKeeper.GasPerBlobByte(ctx)
+	maxSquareSize := app.BlobKeeper.GovMaxSquareSize(ctx)
 	return uint64(maxSquareSize * maxSquareSize * share.ContinuationSparseShareContentSize), nil
 }
 
--- a/app/app.go
+++ b/app/app.go
@@ -755,7 +755,7 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, _ config.APIConfig) {
 func (app *App) RegisterTxService(clientCtx client.Context) {
 	authtx.RegisterTxService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.BaseApp.Simulate, app.interfaceRegistry)
 	celestiatx.RegisterTxService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
-	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.BaseApp.Simulate)
+	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.txConfig.TxDecoder(), app.BaseApp.Simulate)
 }
 
 // RegisterTendermintService implements the Application.RegisterTendermintService method.
--- a/app/app.go
+++ b/app/app.go
@@ -766,7 +766,7 @@ func (app *App) getGovMaxSquareBytes() (uint64, error) {
 		return 0, err
 	}
 	maxSquareSize := app.BlobKeeper.GovMaxSquareSize(ctx)
-	return maxSquareSize * maxSquareSize * share.ContinuationSparseShareContentSize, nil
+	return maxSquareSize * maxSquareSize * share.ShareSize, nil
 }
 
 // RegisterTendermintService implements the Application.RegisterTendermintService method.
--- a/app/benchmarks/gas_estimator_test.go
+++ b/app/benchmarks/gas_estimator_test.go
@@ -34,7 +34,7 @@ func BenchmarkGasPriceEstimation(b *testing.B) {
 	require.NoError(b, err)
 
 	// 8 mb block
-	blockSizeInBytes := 128 * 128 * share.ContinuationSparseShareContentSize
+	blockSizeInBytes := 8 * mebibyte
 
 	benchmarks := []struct {
 		name                 string
--- a/app/benchmarks/gas_estimator_test.go
+++ b/app/benchmarks/gas_estimator_test.go
@@ -4,6 +4,8 @@ package benchmarks_test
 
 import (
 	"context"
+	"testing"
+
 	"github.com/celestiaorg/celestia-app/v3/app"
 	"github.com/celestiaorg/celestia-app/v3/app/encoding"
 	"github.com/celestiaorg/celestia-app/v3/app/grpc/gasestimation"
@@ -21,7 +23,6 @@ import (
 	"github.com/tendermint/tendermint/rpc/client"
 	rpctypes "github.com/tendermint/tendermint/rpc/core/types"
 	"github.com/tendermint/tendermint/types"
-	"testing"
 )
 
 func BenchmarkGasPriceEstimation(b *testing.B) {
--- a/app/benchmarks/gas_estimator_test.go
+++ b/app/benchmarks/gas_estimator_test.go
@@ -1,0 +1,124 @@
+package benchmarks_test
+
+import (
+	"context"
+	"github.com/celestiaorg/celestia-app/v3/app"
+	"github.com/celestiaorg/celestia-app/v3/app/encoding"
+	"github.com/celestiaorg/celestia-app/v3/app/grpc/gasestimation"
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/v3/pkg/user"
+	testutil "github.com/celestiaorg/celestia-app/v3/test/util"
+	"github.com/celestiaorg/celestia-app/v3/test/util/blobfactory"
+	"github.com/celestiaorg/celestia-app/v3/test/util/testfactory"
+	blobtypes "github.com/celestiaorg/celestia-app/v3/x/blob/types"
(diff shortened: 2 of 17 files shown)


app/benchmarks/gas_estimator_test.go
```diff
@@ -0,0 +1,126 @@
...
+	txs []types.Tx
+}
+
+func (b benchMempool) UnconfirmedTxs(ctx context.Context, limit *int) (*rpctypes.ResultUnconfirmedTxs, error) {
+	return &rpctypes.ResultUnconfirmedTxs{Txs: b.txs}, nil
+}
+
+func (b benchMempool) NumUnconfirmedTxs(ctx context.Context) (*rpctypes.ResultUnconfirmedTxs, error) {
+	//TODO implement me
+	panic("implement me")
```
rootulp, line 107: can this return empty values instead of `panic`ing?

app/benchmarks/gas_estimator_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (b benchMempool) CheckTx(ctx context.Context, tx types.Tx) (*rpctypes.ResultCheckTx, error) {
+	//TODO implement me
+	panic("implement me")
```
rootulp, line 112: can this return empty values instead of panicing?

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -105,144 +88,120 @@ func (s *gasEstimatorServer) EstimateGasPriceAndUsage(ctx context.Context, reque
 	}, nil
 }
 
+// gasPriceEstimationThreshold the threshold of mempool transactions to
+// estimate the gas price.
```
rootulp, line 92: [nit] this comment doesn't explain what the threshold is used for. Can it be revised to say something about how min gas price is used if the block is less than this full?

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -105,144 +88,120 @@ func (s *gasEstimatorServer) EstimateGasPriceAndUsage(ctx context.Context, reque
...
-		ctx,
-		lastFiveBlocksTransactionsQuery(latestHeight),
-		false,
-		&page,
-		&numberOfTransactionsPerPage,
-		"asc",
-	)
+	if txsResp.TotalBytes < int64(appconsts.DefaultMaxBytes*gasPriceEstimationThreshold) {
+		return appconsts.DefaultMinGasPrice, nil
+	}
```
rootulp, line 108: [question] this doesn't account for padding in a data square. The txs can take up < 70% of the bytes of DefaultMaxBytes and still take up > 100% of the data square.

[blocking] appconsts.DefaultMaxBytes shouldn't be used here because it's still ~1.9 MiB. On all networks, governance has modified it to be 8 MiB so this won't work as expected.

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -105,144 +88,120 @@ func (s *gasEstimatorServer) EstimateGasPriceAndUsage(ctx context.Context, reque
...
 	case TxPriority_TX_PRIORITY_UNSPECIFIED:
-		return meanGasPrice, nil
+		return Median(gasPrices)
 	case TxPriority_TX_PRIORITY_LOW:
-		stDev := StandardDeviation(meanGasPrice, gasPrices)
-		return meanGasPrice - EstimationZScore*stDev, nil
+		bottom10PercentIndex := len(gasPrices) * 10 / 100
+		if bottom10PercentIndex == 0 {
+			// the case of a slice containing less than 10 elements
+			bottom10PercentIndex = 1
```
rootulp, line 132: [question] what if there are 0 elements then an index of 1 will be out of bounds

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -105,144 +88,120 @@ func (s *gasEstimatorServer) EstimateGasPriceAndUsage(ctx context.Context, reque
...
-	for _, tx := range txs {
-		var feeWithDenom string
-		for _, event := range tx.TxResult.Events {
-			if event.GetType() == "tx" {
-				for _, attr := range event.Attributes {
-					if string(attr.Key) == "fee" {
-						feeWithDenom = string(attr.Value)
-					}
-				}
+// The total size of the returned transactions won't exceed the max byte parameter.
```
rootulp, line 146:
```suggestion
// The total size of the returned transactions won't exceed the maxBytes parameter.
```

app/test/gas_estimation_test.go
```diff
@@ -73,49 +140,45 @@ func TestEstimateGasPrice(t *testing.T) {
...
-	require.NoError(t, err)
 
 	close(gasPricesChan)
 	gasPrices := make([]float64, 0, len(accountNames))
 	for price := range gasPricesChan {
 		gasPrices = append(gasPrices, price)
 	}
+	sort.Float64s(gasPrices)
+
+	medianGasPrices, err := gasestimation.Median(gasPrices)
```
rootulp, line 151:
```suggestion
	medianGasPrice, err := gasestimation.Median(gasPrices)
```

rootulp (commented):
[blocking] `appconsts.DefaultMaxBytes` won't work as expected because it's 1.9 MiB. If we want to use the max bytes in a square then we should prob use the value of the governance param but that requires querying state which would be unideal every time a user tries to estimate gas.

I think the logic around .7 * appconsts.DefaultMaxBytes needs to be rethought.

The rest of the changes LGTM.

app/benchmarks/gas_estimator_test.go
rach-id: f084da0bd97346b6278b7cb784cd1bf444065f06

app/benchmarks/gas_estimator_test.go
rach-id: f084da0bd97346b6278b7cb784cd1bf444065f06

app/grpc/gasestimation/gas_estimator.go
rach-id: 44c2934518157793282f70b8b1c3719fed64c2f4

app/grpc/gasestimation/gas_estimator.go
rach-id: this expects 70% of the block to be filled, which means the list of gas prices can't be empty. But I guess it makes more sense to add an explicit check:

c9582bce2b719989fa475406ba3ff822f65e4e07

app/grpc/gasestimation/gas_estimator_test.go
```diff
@@ -37,59 +39,27 @@ func TestMean(t *testing.T) {
...
-			got := Mean(tt.gasPrices)
+			got, err := Median(tt.gasPrices)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
+			}
+			assert.NoError(t, err)
 			if got != tt.want {
 				t.Errorf("mean(%v) = %v, want %v", tt.gasPrices, got, tt.want)
 			}
```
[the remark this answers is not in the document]
rach-id: f4dcc44a5433e6c0bf188c52b1115d4f86fccaa6

app/grpc/gasestimation/gas_estimator_test.go
```diff
@@ -37,59 +39,27 @@ func TestMean(t *testing.T) {
...
-		})
-	}
-}
-
 func TestEstimateGasPriceForTransactions(t *testing.T) {
-	gasPrices := []float64{10, 20, 30, 40, 50}
-	meanGasPrices := Mean(gasPrices)
-	stDev := StandardDeviation(meanGasPrices, gasPrices)
+	gasPrices := []float64{10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110}
+	medianGasPrices, err := Median(gasPrices)
```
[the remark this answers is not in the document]
rach-id: c01b85a42050aff2f93e3ee7349dd3ab1b20432a

app/test/gas_estimation_test.go
rach-id: 673f765f15bdc5dce5b83bce3c9be9aad60729b9

app/grpc/gasestimation/gas_estimator.go
rach-id: > [question] this doesn't account for padding in a data square. The txs can take up < 70% of the bytes of DefaultMaxBytes and still take up > 100% of the data square.

Yes, that's true. The issue if we want to account for the padding is needing to build a square and that's too expensive to do on every estimation call. I guess this value is enough since it covers at least 70% of the block or more.

> [blocking] appconsts.DefaultMaxBytes shouldn't be used here because it's still ~1.9 MiB. On all networks, governance has modified it to be 8 MiB so this won't work as expected.

good catch 👍 

6460fd1ad74985817d945cb762743f4c47c98330

rootulp (changes requested):
Has this been tested manually? I don't expect it to work as expected because it's querying `GasPerBlobByte` instead of params related to square size.

app/app.go
```diff
@@ -755,7 +757,16 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, _ config.APIConfig) {
...
-	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.BaseApp.Simulate)
+	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.txConfig.TxDecoder(), app.getGovMaxSquareBytes, app.BaseApp.Simulate)
+}
+
+func (app *App) getGovMaxSquareBytes() (uint64, error) {
+	ctx, err := app.CreateQueryContext(app.LastBlockHeight(), false)
+	if err != nil {
+		return 0, err
+	}
+	maxSquareSize := app.BlobKeeper.GasPerBlobByte(ctx)
```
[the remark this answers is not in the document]
rach-id: really good catch 👍  I was in a hurry I called the wrong function 7460b78bf0dd10710fe1d2807d742dda8f23a130

app/app.go
```diff
@@ -755,7 +757,16 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, _ config.APIConfig) {
 func (app *App) RegisterTxService(clientCtx client.Context) {
 	authtx.RegisterTxService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.BaseApp.Simulate, app.interfaceRegistry)
 	celestiatx.RegisterTxService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
-	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.BaseApp.Simulate)
+	gasestimation.RegisterGasEstimationService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.txConfig.TxDecoder(), app.getGovMaxSquareBytes, app.BaseApp.Simulate)
```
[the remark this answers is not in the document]
rach-id: > this only queries the gov max square bytes once at start-up

if I understand correctly, this is not querying the gov max square at startup, it queries it whenever that function is called. So, it will query the state with every request

app/benchmarks/gas_estimator_test.go
```diff
@@ -0,0 +1,128 @@
...
+func BenchmarkGasPriceEstimation(b *testing.B) {
+	encfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	accounts := testfactory.GenerateAccounts(1)
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultConsensusParams(), accounts...)
+	infos := queryAccountInfo(testApp, accounts, kr)
+	signer, err := user.NewSigner(kr, encfg.TxConfig, testutil.ChainID, appconsts.LatestVersion, user.NewAccount(accounts[0], infos[0].AccountNum, infos[0].Sequence))
+	require.NoError(b, err)
+
+	// 8 mb block
+	blockSizeInBytes := 128 * 128 * share.ContinuationCompactShareContentSize
```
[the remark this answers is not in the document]
rach-id: changed to this: 853e6179dcb038dccedc866a216fd660ae9daf53

since we're only benchmarking PFBs, it's fine to use sparse shares size

app/app.go
```diff
@@ -755,7 +757,16 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, _ config.APIConfig) {
...
+}
+
+func (app *App) getGovMaxSquareBytes() (uint64, error) {
+	ctx, err := app.CreateQueryContext(app.LastBlockHeight(), false)
+	if err != nil {
+		return 0, err
+	}
+	maxSquareSize := app.BlobKeeper.GovMaxSquareSize(ctx)
+	return maxSquareSize * maxSquareSize * share.ContinuationSparseShareContentSize, nil
```
rootulp, line 769: [nit] this still uses a constant (`share.ContinuationSparseShareContentSize`) that is an implementation detail. Given the estimate and logic around this code is not exact, IMO it would be clearer to use `share.ShareSize`.

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -105,144 +94,130 @@ func (s *gasEstimatorServer) EstimateGasPriceAndUsage(ctx context.Context, reque
...
-// If no transaction is found in the last five blocks, return the network
-// min gas price.
-// It's up to the light client to set the gas price in this case
-// to the minimum gas price set by that node.
+// on the gas prices of the transactions in the mempool.
+// If the mempool transactions can't fill more than 70% of the block, the min gas
+// price is returned.
 func (s *gasEstimatorServer) estimateGasPrice(ctx context.Context, priority TxPriority) (float64, error) {
-	status, err := s.clientCtx.Client.Status(ctx)
+	// using -1 to return all the transactions.
```
rootulp, line 109:
```suggestion
	// Use -1 to query all the unconfirmed transactions.
```

rootulp (approved):

evan-forbes (approved):

app/benchmarks/gas_estimator_test.go
```diff
@@ -0,0 +1,129 @@
...
+func BenchmarkGasPriceEstimation(b *testing.B) {
+	encfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	accounts := testfactory.GenerateAccounts(1)
+	testApp, kr := testutil.SetupTestAppWithGenesisValSet(app.DefaultConsensusParams(), accounts...)
+	infos := queryAccountInfo(testApp, accounts, kr)
+	signer, err := user.NewSigner(kr, encfg.TxConfig, testutil.ChainID, appconsts.LatestVersion, user.NewAccount(accounts[0], infos[0].AccountNum, infos[0].Sequence))
+	require.NoError(b, err)
+
+	// 8 mb block
+	blockSizeInBytes := 128 * 128 * share.ContinuationSparseShareContentSize
```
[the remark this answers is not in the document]
rach-id: 4aa799533647c6d6e547c17492c32971550545e7

app/app.go
rach-id: aef21be44da31797e2fcc878ecc3e804290a9429

rootulp (approved):

evan-forbes (approved):

Issue #4413, opened 2025-03-14
Decrease `UpgradeHeightDelay` on networks


rootulp:
## Context

https://celestia-team.slack.com/archives/C087852KEVC/p1741966964952959

## Problem

Currently the `UpgradeHeightDelay` is 7 days on Arabica, Mocha, and Mainnet Beta. This means it will take more than 3 weeks to deploy a new celestia-app major version across all networks. 

The UpgradeHeightDelay is useful to give some grace period to the 1/6 of voting power that haven't upgraded prior to the activation height. 

```mermaid
graph LR
  A[Create v4.0.0-arabica. Ask validators to upgrade binary & signal for v4] -- Wait until 5/6 voting power has signalled for v4--> B[Submit MsgTryUpgrade]
  B -- Wait UpgradeHeightDelay --> C[v4 activates on Arabica]
```

## Proposal

1. Arabica: decrease `UpgradeHeightDelay` from 7 days to 1 day
2. Mocha: decrease `UpgradeHeightDelay` from 7 days to 2 days
3. ~~Mainnet: decrease `UpgradeHeightDelay` from 7 days to 3 days~~

Note: this is consensus breaking and we're going to release it in a minor release. Nodes must upgrade to a release with this change _prior_ to the upgrade process starting on that network (specifically before the `MsgTryUpgrade` is submitted).

rootulp:
Revised this issue to decrease the upgrade height on all networks. cc: @cmwaters @evan-forbes @adlerjohn

jcstein:
I'm in favor of these changes

adlerjohn:
Also fine with these changes

rootulp:
In core/app sprint planning we agreed to decrease the upgrade height delay on Arabica and Mocha but not on Mainnet. I updated the issue description.

PR #4414, opened 2025-03-14
feat!: reduce upgrade height delay for Arabica and Mocha

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4413

Note: This targets main but it won't ever be used on main because https://github.com/celestiaorg/celestia-app/pull/4400 simplifies the versioned constants. Main (and v4+) won't have versioned constants because they just need to support constants for the latest release. This change will be backported to v3.x where it will be used.

### DIFF
--- a/pkg/appconsts/versioned_consts.go
+++ b/pkg/appconsts/versioned_consts.go
@@ -11,6 +11,9 @@ import (
 
 const (
 	LatestVersion = v3.Version
+	// numBlocksPerDay is the number of blocks in a day assuming a block
+	// interval of 6 seconds. Equivalent to 14,400 blocks.
+	numBlocksPerDay = int64(24 * 60 * 60 / 6)
 )
 
 // SubtreeRootThreshold works as a target upper bound for the number of subtree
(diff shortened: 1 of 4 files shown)


evan-forbes (dismissed):
Almost certainly fine, but out of good habit we might just to document that all software < this release are now incompatible with Arabica since this will change the state machine upon a successful crank message

rach-id (approved):

rootulp:
Putting in draft because I want to amend this PR to decrease the upgrade height delay on all networks.

evan-forbes (approved):

pkg/appconsts/versioned_consts.go
```diff
@@ -94,7 +100,16 @@ func UpgradeHeightDelay(chainID string, v uint64) int64 {
...
+		switch chainID {
+		case ArabicaChainID:
+			return numBlocksPerDay
+		case MochaChainID:
+			return numBlocksPerDay * 2
+		case MainnetChainID:
+			return v3.UpgradeHeightDelay
+		}
 	}
+	// TODO: this should panic because this should never be invoked for v4+
```
rach-id, line 113: [not blocking]
is there an issue for this TODO?

rach-id (approved):

pkg/appconsts/versioned_consts.go
rootulp: No issue but it may not be worth including this TODO because it will live in the v3.x line which we won't really be updating b/c we're about to merge all the v4 stuff to `main` soon

PR #4416, opened 2025-03-17
refactor: remove redundant import aliases

argentpapa:
Enable a linter that catches redundant import aliases.


rach-id (dismissed):

x/blob/types/payforblob.go
```diff
@@ -2,7 +2,7 @@ package types
 
 import (
 	"bytes"
-	fmt "fmt"
```
rootulp, line 5: Can you please enable a linter to catch all occurrences where a package is imported with an alias that is the same name?

argentpapa: since revive is already enabled, I config https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#redundant-import-alias

rootulp:
Thanks! This is great!

golangci-lint is complaining: https://github.com/celestiaorg/celestia-app/actions/runs/13902535794/job/38897608828?pr=4416 so there may be a few more imports that need fixing

argentpapa:
for case:
`  Error: app/module/versioned_ibc_module_test.go:7:2: redundant-import-alias: Import alias "mocks" is redundant (revive)
  	mocks "github.com/celestiaorg/celestia-app/v3/app/module/mocks"`
  	
[bug: redundant-import-alias rule check for import path instead of package name:](https://github.com/mgechev/revive/issues/936)
maybe we need add  something like this: `//revive:disable:redundant-import-alias`

app/module/versioned_ibc_module_test.go
argentpapa, line 1: ref: https://github.com/mgechev/revive/issues/936
"github.com/celestiaorg/celestia-app/v3/app/module/mocks"- > package mock_types

app/test/circuit_breaker_test.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/celestiaorg/celestia-app/v3/app"
 	"github.com/celestiaorg/celestia-app/v3/app/encoding"
-	v1 "github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v1"
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v1"
```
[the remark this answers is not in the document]
argentpapa: ref: https://pkg.go.dev/github.com/celestiaorg/celestia-app/v3/pkg/appconsts/v1 
package name is v1, no need change to `appconsts `

rootulp:
I think this is going to cause merge conflicts with https://github.com/celestiaorg/celestia-app/pull/4400 . This PR is a nice to have and that one is required so I'm inclined to postpone merging this until after that one lands. Hope that's okay with you @argentpapa

rootulp:
We upgraded from golangci-lint v1 to v2 so this PR will have to change the config used in .golangci.yml

rootulp:
IMO worth blocking on merging https://github.com/celestiaorg/celestia-app/pull/4400 first.

rootulp:
No longer blocked but sorry this has a bunch of merge conflicts. `main` has the v4 PR merged so if you want to resume this PR, we'd appreciate it @argentpapa . Thanks in advance!

rootulp (approved):

rach-id (approved):

PR #4417, opened 2025-03-17
fix: fixed arithmetic syntax in bash script

0xstratus:
## Overview

I replaced `$[...]` with the correct `$((...))` syntax for arithmetic operations in the script.
This ensures proper functionality when selecting a random container.


rach-id:
not providing any added value

PR #4418, opened 2025-03-17
Fix вocumentation: correct public key and share terminology 

Marcofann:
Fixed incorrect Secp256k1 key size notation:  
  - Updated "257-bits" to "33 bytes (264 bits)" in `public_key_cryptography.md`. 

Fixed incorrect grammar in "Blob data are opaque bytes":  
  - Changed to "Blob data is a sequence of opaque bytes" in `shares.md`. 
 
Corrected numbering in ordered lists:  
  - Fixed Markdown list numbering in `shares.md` to properly display **1, 2** instead of two **1s**.  

Clarified "big-endian" description in share format:  
  - Replaced "big-endian unsigned integer" with "MSB-first unsigned integer" for better technical accuracy.

### DIFF
--- a/specs/src/public_key_cryptography.md
+++ b/specs/src/public_key_cryptography.md
@@ -14,7 +14,7 @@ A highly-optimized library is available in C (<https://github.com/bitcoin-core/s
 
 ### Public-keys
 
-Secp256k1 public keys can be compressed to 257-bits (or 33 bytes) per the format described [here](https://github.com/cosmos/cosmos-sdk/blob/v0.46.15/docs/basics/accounts.md#public-keys).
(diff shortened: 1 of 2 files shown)


specs/src/public_key_cryptography.md
```diff
@@ -14,7 +14,7 @@ A highly-optimized library is available in C (<https://github.com/bitcoin-core/s
 
 ### Public-keys
 
-Secp256k1 public keys can be compressed to 257-bits (or 33 bytes) per the format described [here](https://github.com/cosmos/cosmos-sdk/blob/v0.46.15/docs/basics/accounts.md#public-keys).
+Secp256k1 public keys can be compressed to 33 bytes (264 bits) per the format described [here](https://github.com/cosmos/cosmos-sdk/blob/v0.46.15/docs/basics/accounts.md#public-keys).
```
rootulp, line 17: This is wrong.

rootulp (changes requested):

rootulp:
Does not add value. Some of the changes are wrong.

PR #4430, opened 2025-03-19
feat!: set upgrade on upgrade keeper in endblocker

julienrbrt:
Similar to https://github.com/01builders/celestia-app/pull/68 but for v3.

After this PR, signaling a version on v3 will as well set the upgrade on the upgrade keeper.
It then expects a binary that contains the upgrade handler with the name `v`+`AppVersion`, at the next block.
This will be useful for running upgrade with the multiplexer.
Migration handler and Store migrator logic is skipping v4, as the migration is done on the new binary.

This should be backported to v3.

### DIFF
--- a/pkg/appconsts/versioned_consts_test.go
+++ b/pkg/appconsts/versioned_consts_test.go
@@ -137,7 +137,6 @@ func TestUpgradeHeightDelay(t *testing.T) {
 			expectedUpgradeHeightDelay: 3,
 		},
 		{
-
 			name:                       "v3 upgrade delay on arabica is 1 day",
 			chainID:                    appconsts.ArabicaChainID,
 			version:                    v3.Version,


rootulp (commented):
The module manager has a [RunMigrations](https://github.com/rootulp/celestia-app/blob/1f9ccbe3d33c8090e9f47b56e7a955127bcbab40/app/module/manager.go#L220-L270) that was intended to be used instead of x/upgrade but I don't recall why we needed a custom `RunMigrations` instead of letting x/upgrade do it. @cmwaters do you recall?

Note: if we do use x/upgrade instead of the module manager `RunMigrations` then we can remove that in a FLUP.

Left a few nits regarding a rename (version => upgrade) now that a function signature has changed.

cmwaters:
> The module manager has a [RunMigrations](https://github.com/rootulp/celestia-app/blob/1f9ccbe3d33c8090e9f47b56e7a955127bcbab40/app/module/manager.go#L220-L270) that was intended to be used instead of x/upgrade but I don't recall why we needed a custom RunMigrations instead of letting x/upgrade do it. @cmwaters do you recall?

x/upgrade was only kept around for IBC which depends on it (and probably shouldn't). IIRC migrations were never in upgrade. It was always a part of the module manager and modules could register their handlers to the manager and then baseapp would call the module manager upon app version change

app/app.go
```diff
@@ -484,11 +484,19 @@ func (app *App) EndBlocker(ctx sdk.Context, req abci.RequestEndBlock) abci.Respo
...
-		if newVersion > currentVersion {
-			app.BaseApp.Logger().Info("upgrading app version", "current version", currentVersion, "new version", newVersion)
-			app.SetAppVersion(ctx, newVersion)
+		if upgrade.AppVersion > currentVersion {
+			app.BaseApp.Logger().Info("upgrading app version", "current version", currentVersion, "new version", upgrade.AppVersion)
+
+			if err := app.UpgradeKeeper.ScheduleUpgrade(ctx, upgradetypes.Plan{
+				Name:   fmt.Sprintf("v%d", upgrade.AppVersion),
+				Height: upgrade.UpgradeHeight,
+			}); err != nil {
```
rootulp, line 495: since this is being backported to v3.x, should this be gated to only apply to v4? We don't want to schedule an upgrade on the upgrade keeper for v2 -> v3 which should still be supported by the v3.x binary.

app/module/manager.go
```diff
@@ -234,6 +235,14 @@ func (m Manager) RunMigrations(ctx sdk.Context, cfg sdkmodule.Configurator, from
 	}
 	nextVersionModules, exists := m.versionedModules[toVersion]
 	if !exists {
+		// For celestia-appd 4, we do not want to migrate on commit (https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.46.x-celestia/baseapp/abci.go#L342-L366).
+		// Celestia App uses Cosmos SDK v0.50, and upgrade handlers
+		// The migration will be done at PreBlocker on the new binary.
+		// If the binary isn't switched, v3 will anyway halt before the next block is produced (the module manager will panic).
```
rootulp, line 241:
```suggestion
		// Celestia App v4 uses Cosmos SDK v0.50 and upgrade handlers.
		// The migration will be occur in PreBlocker on the new binary.
		// If the binary isn't switched, v3 will halt before the next block is produced because the module manager will panic.
```

rootulp (commented):
[question] should the `upgradeKeeper.ScheduleUpgrade` be version gated to only apply to the v3 -> v4 upgrade? The v3.x binary still needs to support the v2.x -> v3.x upgrade which did not include an upgrade plan.

app/app.go
julienrbrt: Good point 👍

rootulp:
Converting to draft b/c commits still being added to this. Please mark ready for review if/when you want reviews.

julienrbrt:
> Converting to draft b/c commits still being added to this. Please mark ready for review if/when you want reviews.

Yes! Should be fixed now, I missed you have something that were resetting the keys.

app/app.go
```diff
@@ -272,7 +273,8 @@ func New(
 	// The upgrade keeper is initialised solely for the ibc keeper which depends on it to know what the next validator hash is for after the
 	// upgrade. This keeper is not used for the actual upgrades but merely for compatibility reasons. Ideally IBC has their own upgrade module
 	// for performing IBC based upgrades. Note, as we use rolling upgrades, IBC technically never needs this functionality.
```
rootulp, line 275: This comment is no longer applicable because we use the upgrade keeper for v4

```suggestion
```

app/app.go
```diff
@@ -154,7 +155,7 @@ type App struct {
 	DistrKeeper         distrkeeper.Keeper
 	GovKeeper           govkeeper.Keeper
 	CrisisKeeper        crisiskeeper.Keeper
-	UpgradeKeeper       upgradekeeper.Keeper // This is included purely for the IBC Keeper. It is not used for upgrading
+	UpgradeKeeper       upgradekeeper.Keeper // Upgrades are set in endblock when signaled
```
rootulp, line 158: [nit] Upgrades are set in endblocker only for v3 -> v4. All previous upgrades didn't use the upgrade keeper. Proposal to remove the comment 

```suggestion
	UpgradeKeeper       upgradekeeper.Keeper
```

app/app.go
```diff
@@ -484,11 +486,27 @@ func (app *App) EndBlocker(ctx sdk.Context, req abci.RequestEndBlock) abci.Respo
...
-			app.SetAppVersion(ctx, newVersion)
+	} else if shouldUpgrade, upgrade := app.SignalKeeper.ShouldUpgrade(ctx); shouldUpgrade {
+		// Version changes must be increasing. Downgrades are not permitted.
+		if upgrade.AppVersion > currentVersion {
+			app.BaseApp.Logger().Info("upgrading app version", "current version", currentVersion, "new version", upgrade.AppVersion)
+
+			if currentVersion == v3 { // v3 -> v4 needs to schedule an upgrade
+				plan := upgradetypes.Plan{
+					Name:   fmt.Sprintf("v%d", upgrade.AppVersion),
+					Height: upgrade.UpgradeHeight + 1, // next block executing the upgrade.
```
rootulp, line 497: [question] why is the + 1 needed here? Will the block at UpgradeHeight have app version 3 or 4 in it?

```suggestion
					Height: upgrade.UpgradeHeight + 1, // the block after the upgrade height executes the upgrade.
```

rootulp (approved):
LGTM with two nits and one question

app/app.go
julienrbrt: Because of how your fork was working. The migrations were done in `Commit` within the same block, and the app version updated. For the v3 -> v4, and all successive migrations, the app version will be updated at `UpgradeHeight`, and the same block will be running the new binary containing the migrations.

rootulp (approved):
Still LGTM

evan-forbes (dismissed):

rootulp (dismissed):

evan-forbes (dismissed):

evan-forbes (approved):

pkg/appconsts/v3/app_consts.go
```diff
@@ -4,6 +4,7 @@ import "time"
 
 const (
 	Version              uint64 = 3
+	NextVersion          uint64 = 4
```
cmwaters, line 7: Why does this need to exist? Can't you just add a v4 package with the v4 constants

pkg/appconsts/chain_ids.go
```diff
@@ -6,6 +6,8 @@ const (
 	MainnetChainID = "celestia"
 	// TestChainID is the chain ID used for testing.
 	TestChainID = "test"
+	// LocalDevnetChainID is another chain ID used for testing.
+	LocalDevnetChainID = "local_devnet"
```
cmwaters, line 10: Why do we need to add support for another testnet chain?

rootulp: This chain id is used by the `local_devnet` directory. The `local_devnet` directory could be refactored to use `test` so IMO we can do that in a follow up and then remove it from here.

Ref: https://github.com/celestiaorg/celestia-app/blob/8f9b3dcab883046eaca060d6aebdee5337bc4ebd/local_devnet/scripts/start_node_and_create_validator.sh#L47

pkg/appconsts/v3/app_consts.go
julienrbrt: This is only useful for the v3 to v4 upgrade. v3 doesn't need to know anything else about v4.
Felt cleaner than creating a whole new package or do v3.Version+1

rootulp: Maybe this PR didn't create a v4 package because this was backported to v3.x and could be confusing to introduce a v4 package on v3.x 🤷‍♂️ 

IMO nothing prevents us from doing that though

PR #4433, opened 2025-03-21
fix: nil signer panic in `findAccount` and correct error message in `AddAccount`

Pronoss:
1. Nil Signer Safety Check:
   - Enhanced `findAccount` method with an additional check for `nil` signers to prevent potential `panic`.
   - New condition: `if len(signers) == 0 || signers[0] == nil` ensures that signer access is safe.

2. Correct Error Message Context:
   - In `AddAccount`, the error message incorrectly referenced `acc.pubKey`. 
   - Replaced it with `acc.name` for proper context when reporting failure to get the address.

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -161,8 +161,8 @@ func (s *Signer) Accounts() []*Account {
 
 func (s *Signer) findAccount(txbuilder client.TxBuilder) (*Account, error) {
 	signers := txbuilder.GetTx().GetSigners()
-	if len(signers) == 0 {
-		return nil, fmt.Errorf("message has no signer")
+	if len(signers) == 0 || signers[0] == nil {
+		return nil, fmt.Errorf("message has no valid signer")
 	}
 	accountName, exists := s.addressToAccountMap[signers[0].String()]
 	if !exists {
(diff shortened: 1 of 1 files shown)


rootulp:
Can you please explain why this is needed? It would be helpful to see a Github issue that demonstrates how you encountered the `panic`.

Pronoss:
@rootulp
This check prevents a potential panic when `GetSigners()` returns a slice with `nil` as the first element — which can happen if the message is incomplete or the signer address is missing. Without the check, calling `signers[0].String()` would cause a panic.

The updated condition if `len(signers) == 0 || signers[0] == nil` ensures safe access and improves robustness.

I didn’t encounter the panic during runtime but spotted this as a potential issue while reviewing the code and thinking through possible edge cases in transaction building. I can create a test case if needed, but wanted to raise this as a preventive fix to avoid runtime panics.

rootulp:
> which can happen if the message is incomplete or the signer address is missing

Is this possible?

Unit tests increase confidence so PRs should usually include them. In this case, this fix seems unnecessary so inclined to close. Please see https://github.com/celestiaorg/celestia-app/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22 if you'd like to contribute.

PR #4435, opened 2025-03-22
chore(all): replace interface{}

argentpapa:
## Overview

replacing interface{} by the 'any' type added in go1.18;


rootulp:
Why?

rootulp:
Why?

argentpapa:
> Why?

just walkthrough codebase:
image


ref: https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize#pkg-functions

PR #4439, opened 2025-03-23
chore: fix array indexing error in bash script

0xstratus:
## Overview

Fixed an issue with array indexing. The old syntax:

```sh
rand_container=${docker_containers["$[RANDOM % ${#docker_containers[@]}]"]};
```

was incorrect. Updated it to the correct version:

```sh
rand_container=${docker_containers[$((RANDOM % ${#docker_containers[@]}))]};
```

This ensures proper syntax for both arithmetic and array indexing in bash.


rach-id:
not necessary

PR #4440, opened 2025-03-23
chore: fix error message for startShare validation

0xwitty:
## Overview

I’ve updated the error message to clarify that startShare should not be negative. The previous message was a bit misleading, as startShare can be zero, which isn’t a negative number.

Now the message reads: "start share %d should not be negative," making it clearer.


rach-id:
doesn't add any value

PR #4442, opened 2025-03-24
fix(proto): improve comments and consistency in types.proto and wire.proto

tomasandroil:
### Key changes:
- Removed incorrect `: uint8_t` from `AccountStatus` enum definition.
- Fixed `Decimal` message field numbering (`denominator` moved from 1 to 2).
- Corrected typo in `Delegation` comment (`delegating` → `delegated`).
- Replaced "32-bytes" with "32-byte" in comments for consistency.
- Improved clarity in `MessageCommitmentAndSignature` by updating the signature field comment.
- Updated `graffiti` field comment to clarify its purpose as optional arbitrary data.

These changes improve readability, accuracy, and maintainability of the proto files.

### DIFF
--- a/specs/src/proto/types.proto
+++ b/specs/src/proto/types.proto
@@ -42,7 +42,7 @@ message AvailableDataHeader {
 
 // https://github.com/celestiaorg/celestia-specs/blob/master/specs/networking.md#account
 // AccountStatus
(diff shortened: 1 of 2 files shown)


rootulp:
The protos in the specs aren't up to date. We should remove them instead of modifying them.

PR #4447, opened 2025-03-24
feat: implement gas price and gas limit customization for txsim

### LINKED ISSUE #4445: feature request: add gas-limit and gas-price flags to txsim
### Summary

Add `--gas-limit` and `--gas-price` flags to the `txsim` tool to allow users to customize transaction gas parameters, addressing insufficient fee errors when running against networks with higher minimum gas requirements.

### Problem Definition

Users running txsim against certain networks (e.g., validator-1.celestia-arabica-11.com) encounter "insufficient fee" errors due to the default gas parameters being too low. For example:

`error funding accounts: broadcast tx error: insufficient minimum gas price for this node; got: 7200 required at least: 360000.000000000000000000: insufficient fee`

The current implementation uses fixed gas parameters without providing a way for users to adjust them based on network requirements. This prevents txsim from working properly on networks with higher minimum fee requirements.

### Proposal

Add two new command-line flags to `txsim`:
1. `--gas-limit uint`: Allows users to set a custom gas limit for transactions (0 = auto-estimate)
1. `--gas-price float`: Allows users to set a custom gas price for transactions (0 = use default)
These parameters will be passed through the Options struct to the AccountManager, which will then use them when calculating transaction fees and applying gas limits.
With these flags, users can adjust their gas parameters to meet network requirements. For example:

`txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 360000 --gas-price 1.0`

Implementation details:
- Add two new flags to the CLI interface
- Extend the Options struct to store the custom values
- Add methods to pass these values to the AccountManager
- Update the transaction submission logic to use these custom values when specified
- This change allows `txsim` to be more flexible and work across different network configurations without code changes.

jcstein:
## Overview

Resolves #4445 

## Testing results

`txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 360000 --gas-price 1.0`

The fees were outrageous here - 16+ TIA. - [example block](https://arabica.celenium.io/block/4973806?tab=transactions)

`txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 720000 --gas-price 0.5`

This fills 2MB with 8.25 TIA - [example block](https://arabica.celenium.io/block/4973865?tab=transactions)

then using
`txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 3600000 --gas-price 0.1`

works and 2MB costs 1.65 TIA, much more reasonable - [example block](https://arabica.celenium.io/block/4973906?tab=transactions)

### Failed testing

this also failed but idk why
```
txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 18000000 --gas-price 0.02

Error: error funding accounts: broadcast tx error: insufficient minimum gas price for this node; got: 72000 required at least: 360000.000000000000000000: insufficient fee
```

going lower
```
txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 12000000 --gas-price 0.03

Error: error funding accounts: broadcast tx error: insufficient minimum gas price for this node; got: 108000 required at least: 360000.000000000000000000: insufficient fee
```

and up a bit:
```
txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 15000000 --gas-price 0.03

Error: error funding accounts: broadcast tx error: insufficient minimum gas price for this node; got: 108000 required at least: 360000.000000000000000000: insufficient fee
```


rootulp:
nit: golangci-lint is complaining: https://github.com/celestiaorg/celestia-app/actions/runs/14042058383/job/39314321206#step:5:38

test/txsim/account.go
```diff
@@ -225,11 +227,29 @@ func (am *AccountManager) Submit(ctx context.Context, op Operation) error {
...
+			opts = append(opts, user.SetGasLimit(am.gasLimit))
+			
+			// Calculate fee based on gas price
+			var gasPrice float64
+			if am.gasPrice > 0 {
+				gasPrice = am.gasPrice
+			} else {
+				gasPrice = appconsts.DefaultMinGasPrice
+			}
+			opts = append(opts, user.SetFee(uint64(math.Ceil(float64(am.gasLimit)*gasPrice))))
```
rootulp, line 241: [question] this seems a bit difficult to reason about. Is it possible to extract the `fee` calculation to _after_ the custom gas limit and gas price have been set. That way this code can remain less indented:

```go
if am.gasLimit > 0 {
    // set custom gas limit
}

if am.gasPrice > 0 {
  // set custom gas price
}

fee := calculateFee(gasLimit, gasPrice)
opts = append(opts, user.SetFee(fee))
```

That way there's only one line where we set the fee instead of two.

rootulp (approved):
One idea for an optional refactor to make it easier to reason about the fee calculation

test/txsim/account.go
```diff
@@ -225,11 +227,29 @@ func (am *AccountManager) Submit(ctx context.Context, op Operation) error {
...
 	if op.GasLimit == 0 {
-		opts = append(opts, user.SetGasLimit(DefaultGasLimit), user.SetFee(defaultFee))
+		if am.gasLimit > 0 {
+			// Use the custom gas limit set on the AccountManager
+			opts = append(opts, user.SetGasLimit(am.gasLimit))
+			
+			// Calculate fee based on gas price
+			var gasPrice float64
+			if am.gasPrice > 0 {
+				gasPrice = am.gasPrice
```
rach-id, line 237: [nit] 
maybe add a log specifying the custom gas price chosen, especially if it's `< appconsts.DefaultMinGasPrice`. To avoid having to assist people setting values lower than `appconsts.DefaultMinGasPrice` and complaining why it doesn't work

rach-id (approved):
utack

jcstein:
tested with `txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 3600000 --gas-price 0.1`, [example block](https://arabica.celenium.io/block/4977553?tab=transactions)

test/txsim/account.go
jcstein: thank you for the feedback! https://github.com/celestiaorg/celestia-app/pull/4447/commits/36d2c38486f3494a6426ea38fb5a481f7a454153

test/txsim/account.go
jcstein: thank you! https://github.com/celestiaorg/celestia-app/pull/4447/commits/36d2c38486f3494a6426ea38fb5a481f7a454153

test/txsim/account.go
```diff
@@ -224,15 +226,43 @@ func (am *AccountManager) Submit(ctx context.Context, op Operation) error {
...
+
+	// Determine gas limit
+	switch {
+	case op.GasLimit > 0:
+		gasLimit = op.GasLimit
+	case am.gasLimit > 0:
+		gasLimit = am.gasLimit
+	default:
+		gasLimit = DefaultGasLimit
+		fee = defaultFee
```
rootulp, line 240: [question] I'm a little confused by why this line is here. 
```suggestion
```

can this code set fee _after_ determining the gas limit and gas price?

rootulp (commented):
I'm still a little confused on why this calculates fee in the middle of setting gas limit and gas price. IMO the code could be clearer if it:
1. Determine gas limit (either user specified or default)
2. Determine gas price (either user specified or default)
3. Calculate fee based on gas limit and gas price from step 1 and 2

I haven't tried any of this code and it's mostly for testing purposes so if this works then I'm okay with merging it

test/txsim/account.go
jcstein: but gave it a shot https://github.com/celestiaorg/celestia-app/pull/4447/commits/a71af690d82df86406738487af31a3b538ea0882

rootulp (approved):
LGTM

rootulp:
FYI golangci-lint is failing but it wasn't caused by this PR: https://github.com/celestiaorg/celestia-app/issues/4454 will fix separately.

jcstein:
thank you for the feedback! I think https://github.com/celestiaorg/celestia-app/pull/4447/commits/6c7bfea012af7a718e76cf94b51dc1b56606be15 should resolve. before merging this though I'd like to test it once

jcstein:
worked as expected: `txsim --key-path ~/.celestia-app --grpc-endpoint validator-1.celestia-arabica-11.com:9090 --feegrant --blob 4 --blob-amounts 1 --blob-sizes 475000 --gas-limit 3600000 --gas-price 0.1`

example block can't be found rn as celenium is behind on arabica. here's a gist: https://gist.github.com/jcstein/020ed3b16e4c368bf10775dbeab75d38

rootulp (approved):

evan-forbes (approved):
utack

jcstein:
> The gas limit is a bit of strange add because txsim commonly is used for simulating multiple types of transactions at once all requiring different amounts of gas. @jcstein was there a specific use case for wanting to set the gas limit? I can understand the gas price.

I will trust core/app team and your judgement here @cmwaters! In https://github.com/celestiaorg/celestia-app/issues/4445 I hit issues with `error funding accounts: broadcast tx error: insufficient minimum gas price for this node; got: 7200 required at least: 360000.000000000000000000: insufficient fee`, but did not try to work this out using just ability to set gas limit. want to rework it?

rach-id:
Proposition: 

Why not use the gas estimation endpoint in txsim? would be a good stress test for it every time someone runs it, and it should return accurate results.

jcstein:
I'm not sure what's best here. I'd defer to core/app team to take this over if possible, I just was trying to use txsim and it didn't work without changing the minimum gas price to be enough compared to what the endpoint is asking for

evan-forbes (dismissed):

rootulp (approved):

jcstein:
I'm still a bit concerned this implementation is wrong, even though it "works"

cmwaters:
> Why not use the gas estimation endpoint in txsim? would be a good stress test for it every time someone runs it, and it should return accurate results.

Eventually we should. But I think allowing to manually set it is also good. We should open an issue for it

cmwaters:
> I hit issues with error funding accounts: broadcast tx error: insufficient minimum gas price for this node; got: 7200 required at least: 360000.000000000000000000: insufficient fee,

Yeah this is from the gas price being insufficient

cmwaters (dismissed):

cmwaters (approved):

Issue #4426, opened 2025-03-18
Pin SHAs for GH actions


rootulp:
## Context

https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

## Problem

Git tags are mutable so an attacker can compromise a GH action, modify the code that the tag points to, and get it to run in our CI.

## Proposal

1. Modify a GH action in this repo to use SHAs instead of tags. Include a comment with the version so that dependabot can update it.
2. After verifying that dependabot can update the SHA and version, apply the same change to all workflows and GH actions in this repo.
3. [Optional] Repeat for all upstream workflows (i.e. https://github.com/celestiaorg/.github)

RyanCarlisle:
Hi, can I work on this?

RyanCarlisle:
Hey @rootulp, here's a PR with the workflows running in a repo in my account.
PR link : https://github.com/RyanCarlisle/celestia-app/pull/1

Please check this and let me know if the changes can be merged or if any other changes are required.

rootulp:
It looks good. Are you able to confirm that dependabot can still bump those versions and SHAs?

RyanCarlisle:
@rootulp I am sure that it would be able to update the SHAs, however, I am unable to find out whether it will update the commented versions. The blog you gave here had this repo and I checked one of the dependabot PRs where it updates the versions (PR link: https://github.com/nodejs/security-wg/pull/1246/files). The dependabot file link is: https://github.com/nodejs/security-wg/blob/main/.github/dependabot.yml. However, when I search for this, it says that we have to update the commented versions manually.

Is there any way in which I can be sure of this?

rootulp:
The example in that PR seems fishy b/c https://github.com/nodejs/security-wg/pull/1246/files/bf5f4ba87b9ab222109b07472d5e3c5bdc111be3#diff-fbf0b5c5b8fe0c87675d921f38b5abd81847e474398a5250b8b21ed5158b8aa2

> However, when I search for this, it says that we have to update the commented versions manually.

If that's the case, then I'm inclined to not do this b/c updating the versions in comments seems tedious and not worth it.

> Could you suggest a way in which I can be sure of this?

Yes. Create a repo in your personal account, add GH actions with a SHA and version comment, enable dependabot, see if dependabot opens a PR to bump the SHA and version comment.

RyanCarlisle:
Just confirmed it works: https://github.com/RyanCarlisle/celestia-app/pull/7
[...]

PR #4448, opened 2025-03-24
ci: add SHAs and commented versions in workflow .yml files

RyanCarlisle:
This PR fixes: https://github.com/celestiaorg/celestia-app/issues/4426

## Overview

Modify a GH action in this repo to use SHAs instead of tags. Include a comment with the version so that dependabot can update it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,5 @@
 version: 2
-updates: # trying to run dependabot
+updates:
   - package-ecosystem: docker
     directory: "/"
     schedule:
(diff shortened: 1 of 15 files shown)


rootulp:
`!` in the conventional commit prefix is for breaking changes. This PR doesn't contain a breaking change.

Ref: https://www.conventionalcommits.org/en/v1.0.0/#summary

rootulp (approved):
LGTM if CI passes

rach-id (approved):

Issue #4444, opened 2025-03-24
Adjust the estimated gas price during congestion when it's equal to the minimum


rach-id:
If the mempool is full of transactions that are all using the minimum gas price, the current estimation mechanism will return the minimum gas price regardless of the priority. We can fix this by applying an adjustment to the estimation given the priority to increase the inclusion probability.

PR #4449, opened 2025-03-24
feat: adjust the estimated gas price for high and medium priority

rach-id:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/4444

I compared the benchmark between this adjusted mechanism and the previous one, and no performance changes were perceived. Benchmarks were done in a MacBook Pro m3 max on 8mb blocks of transactions

### DIFF
--- a/app/grpc/gasestimation/gas_estimator.go
+++ b/app/grpc/gasestimation/gas_estimator.go
@@ -130,7 +130,7 @@ func (s *gasEstimatorServer) estimateGasPrice(ctx context.Context, priority TxPr
 const (
 	// highPriorityGasAdjustmentRate is the percentage increase applied to the
 	// estimated gas price when the block is more than 70% full, i.e., gasPriceEstimationThreshold,
-	// and the estimated gas price equals the minimum gas price. This ensures that high-priority
+	// but the gas prices are tightly clustered. This ensures that high-priority
 	// transactions still have a competitive fee to improve inclusion probability.
 	highPriorityGasAdjustmentRate = 1.3
 	// mediumPriorityGasAdjustmentRate similar to highPriorityGasAdjustmentRate but for
--- a/app/grpc/gasestimation/gas_estimator.go
+++ b/app/grpc/gasestimation/gas_estimator.go
@@ -157,7 +157,7 @@ func estimateGasPriceForTransactions(gasPrices []float64, priority TxPriority) (
 	}
 	stDev := StandardDeviation(Mean(gasPrices), gasPrices)
 	switch priority {
-	case TxPriority_TX_PRIORITY_UNSPECIFIED:
+	case TxPriority_TX_PRIORITY_MEDIUM, TxPriority_TX_PRIORITY_UNSPECIFIED:
 		estimation, err := Median(gasPrices)
 		if err != nil {
 			return 0, err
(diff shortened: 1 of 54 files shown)


evan-forbes (dismissed):
nice

do we have an issue to add a specified maximum somewhere to make sure this doesn't just increase forever?

do we also have an issue to only trigger these if the mempool is ~mostly full? or is that already implemeted? I see adding this as blocking shipping this code (shipping != merging)

rach-id:
> do we have an issue to add a specified maximum somewhere to make sure this doesn't just increase forever?

I think we can leave this endpoint just returning the estimation needed. For the max gas price, Celestia-node provides that as a flag.

> do we also have an issue to only trigger these if the mempool is ~mostly full? or is that already implemeted? 

Yes, this is only triggered when the mempool has > 70% full block: https://github.com/celestiaorg/celestia-app/blob/25e222beabc97c3ecee0bd710df9d88530a1f6da/app/grpc/gasestimation/gas_estimator.go#L119-L121

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -126,21 +127,45 @@ func (s *gasEstimatorServer) estimateGasPrice(ctx context.Context, priority TxPr
...
+)
+
 // estimateGasPriceForTransactions takes a list of transactions and priority
 // and returns a gas price estimation.
 // The priority sets the estimation as follows:
 // - High Priority: The gas price is the median price of the top 10% of transactions’ gas prices in the mempool.
 // - Medium Priority: The gas price is the median price of the all gas prices in the mempool.
 // - Low Priority: The gas price is the median price of the bottom 10% of gas prices in the mempool.
 // - Unspecified Priority (default): This is equivalent to the Medium priority, using the median price of all gas prices in the mempool.
+// If the list of gas prices has a standard deviation < gasPriceAdjustmentThreshold, meaning the gas price values are tightly clustered,
```
rootulp, line 151: [question] what is the intuition around only applying the gasAdjustmentRate if the standard deviation is under some threshold?

If the gas prices aren't tightly clustered, then don't we still want to adjust up?

app/grpc/gasestimation/gas_estimator.go
```diff
@@ -149,9 +174,23 @@ func estimateGasPriceForTransactions(gasPrices []float64, priority TxPriority) (
...
 	case TxPriority_TX_PRIORITY_MEDIUM:
-		return Median(gasPrices)
+		estimation, err := Median(gasPrices)
+		if err != nil {
+			return 0, err
+		}
+		if stDev < gasPriceAdjustmentThreshold {
+			return estimation * mediumPriorityGasAdjustmentRate, nil
+		}
+		return estimation, nil
```
rootulp, line 184: [nit] this can combine two case statements so that the implementation doesn't have to be duplicated

```go
	case TxPriority_TX_PRIORITY_MEDIUM, TxPriority_TX_PRIORITY_UNSPECIFIED:
```

rootulp (dismissed):
LGTM. One question about intuition for why we need `gasPriceAdjustmentThreshold`.

Naively it seems sufficient to apply a multiplier for medium and high priority transactions regardless of the standard deviation.

app/grpc/gasestimation/gas_estimator.go
rach-id: The adjustment is added to avoid the situation where all the transactions keep being of the minimum gas price regardless of the priority.

 We could technically add the adjustment in all cases, but it would be increasing the gas price unnecessarily. The original mechanism is fine and gives accurate results when gas prices are different.

app/grpc/gasestimation/gas_estimator.go
rach-id: good idea 👍  
9237373

rootulp (dismissed):
Still LGTM

evan-forbes (approved):

rootulp:
golangci-lint failing https://github.com/celestiaorg/celestia-app/actions/runs/14068631182/job/39397223591?pr=4449#step:5:39

evan-forbes (approved):

rootulp (approved):

PR #4459, opened 2025-03-26
feat(pkg/user): Implement gas price + usage estimation endpoints on `TxClient`, and option for using separate estimator service

rach-id:
## Overview

This PR implements two endpoints on `TxClient`: 
`EstimateGasPrice` and `EstimateGasPriceAndUsage` on the `TxClient`, deprecating the respective endpoints on `Signer` (as signer's responsibilities should remain purely with signing).

Additionally, there is now an option `WithEstimatorService` which takes a grpc connection to a user-specified estimator endpoint and uses it for all gas price + usage estimation queries. If this option is **not** used, then default connection passed to TxClient on setup will be used.

This PR also removes the concept of the gasMultiplier from the client-side and puts it server-side. When using Simulate, the node can occasionally return a value that would be less than the gas required to submit the transaction at a slightly different time. To lean on the conservative side, this PR adds a 1.1 multiplier to whatever gas is returned by Simulate


rootulp (approved):

pkg/user/tx_client.go
```diff
@@ -512,6 +516,20 @@ func (client *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, op
 	return client.estimateGas(ctx, txBuilder)
 }
 
+// EstimateGasPrice calls the gas estimation endpoint to return the estimated gas price based on priority.
+func (client *TxClient) EstimateGasPrice(ctx context.Context, priority gasestimation.TxPriority) (float64, error) {
```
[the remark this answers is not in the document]
renaynay: @rach-id yeah basically I want 

`client.EstimateGasPriceAndLimit` 

and

`client.EstimateGasPrice`

the reason being - we don't need to estimate gas for blobs (which is majority of txs thru node) so let's keep a clean endpoint for just estimating price based on tx priority.

The other thing is, we need to remove those methods of the signer as they don't have anything to do with signing. Can this be in scope for the PR?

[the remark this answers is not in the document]
rach-id: [f091aa9](https://github.com/celestiaorg/celestia-app/pull/4459/commits/f091aa9842e3e34dd040cb4c8ce3f1a552e3e875)

pkg/user/tx_client.go
```diff
@@ -516,6 +517,47 @@ func (client *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, op
...
 
+// EstimateGasPriceAndUsage simulates the transaction, calculating the amount of gas that was consumed during execution. The final
+// result will be multiplied by gasMultiplier(that is set in TxClient).
+// Also returns the estimated gas price given the provided priority.
+func (client *TxClient) EstimateGasPriceAndUsage(
+	ctx context.Context,
+	msgs []sdktypes.Msg,
+	priority gasestimation.TxPriority,
+	opts ...TxOption,
+) (float64, uint64, error) {
```
rootulp, line 528: [nits] 
- `float64` and `uint64` don't really communicate what is being returned, please add named return values. 
- The godoc describes the return arguments in the reverse order from how they are returned.
- GasUsage != GasLimit. Can we please consolidate on one term?

```suggestion
// EstimateGasPriceAndUsage returns the gasPrice and gasUsage for a set of messages given a priority. 
// Note: The returned gasPrice = estiamtedGasPrice * gasMultiplier (if set in TxClient).
func (client *TxClient) EstimateGasPriceAndUsage(
	ctx context.Context,
	msgs []sdktypes.Msg,
	priority gasestimation.TxPriority,
	opts ...TxOption,
) (gasPrice float64, gasUsage uint64, error) {
```

pkg/user/tx_client.go
```diff
@@ -516,6 +517,47 @@ func (client *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, op
...
+	client.mtx.Lock()
+	defer client.mtx.Unlock()
+
+	txBuilder, err := client.signer.txBuilder(msgs, opts...)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	// add at least 1utia as fee to builder as it affects gas calculation.
+	txBuilder.SetFeeAmount(sdktypes.NewCoins(sdktypes.NewCoin(appconsts.BondDenom, sdktypes.NewInt(1))))
```
rootulp, line 538: Can this comment describe how the 1utia affects gas calculation? This seems kinda suspicious

pkg/user/tx_client.go
```diff
@@ -516,6 +517,47 @@ func (client *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, op
...
+		TxPriority: priority,
+		TxBytes:    txBytes,
+	})
+	if err != nil {
+		return 0, 0, err
+	}
+
+	gasLimit := uint64(float64(resp.EstimatedGasUsed) * client.gasMultiplier)
+
+	return resp.EstimatedGasPrice, gasLimit, nil
```
rootulp, line 558: [nit] use one term to describe this. Either gasUsage or gasLimit.

rootulp (approved):
Still LGTM with a few nits

pkg/user/tx_client.go
rach-id: the estimation of a transaction that contains a specified fee is different than one that doesn't, and since it's a fixed field if I understand correctly, we add it here to have a more exact estimation.

Note: this was existing in the code, and left it based on the above understanding

pkg/user/tx_client.go
rach-id: 6dbd9fc95a9bbdac817a188c8d53688c0de9f060

pkg/user/tx_client.go
rach-id: 6dbd9fc95a9bbdac817a188c8d53688c0de9f060

rootulp (approved):

evan-forbes (approved):

pkg/user/tx_client.go
```diff
@@ -127,6 +129,15 @@ func WithDefaultAccount(name string) Option {
 	}
 }
 
+// WithEstimatorService allows a user to provide a connection to a special gas
+// estimation service to be used by the TxClient for estimating gas price
+// and usage.
+func WithEstimatorService(conn *grpc.ClientConn) Option {
```
renaynay, line 135: FOR CALLUM: 

Please note that when this service is set, we *exclusively* use this estimator service for estimation -- if it fails, there is no fallback of using the default connection (not currently implemented in this PR). 

If a user configures a separate estimator service to use, I think the failure of the service should be explicit + noticeable to the user. 

Thoughts?

pkg/user/tx_client_test.go
```diff
@@ -349,33 +385,64 @@ func setupTxClient(t *testing.T, ttlDuration time.Duration) (encoding.Config, *u
...
+	encCfg := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+
+	txClient, err := user.SetupTxClient(ctx.GoContext(), ctx.Keyring, ctx.GRPCClient, encCfg,
+		append(opts, user.WithGasMultiplier(multiplier))...)
 	require.NoError(t, err)
+
 	return encCfg, txClient, ctx
 }
 
-func (suite *TxClientTestSuite) TestGasPriceAndUsedEstimate() {
```
renaynay, line 357: I removed these tests bc they test now-deprecated endpoints. New endpoints have new coverage.

renaynay (dismissed):
Thank you @rach-id !

evan-forbes (approved):

cmwaters (dismissed):
Few minor changes but thanks for modifying this. It shouldn't have been on the signer which is designed just for signing txs

pkg/user/tx_client.go
```diff
@@ -512,6 +526,62 @@ func (client *TxClient) EstimateGas(ctx context.Context, msgs []sdktypes.Msg, op
...
+	txBytes, err := client.signer.EncodeTx(txBuilder.GetTx())
+	if err != nil {
+		return 0, 0, err
+	}
+	resp, err := client.gasEstimationClient.EstimateGasPriceAndUsage(ctx, &gasestimation.EstimateGasPriceAndUsageRequest{
+		TxPriority: priority,
+		TxBytes:    txBytes,
+	})
+	if err != nil {
+		return 0, 0, err
```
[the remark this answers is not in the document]
renaynay: done

pkg/user/signer.go
```diff
@@ -305,6 +305,7 @@ func (s *Signer) getSignatureV2(sequence uint64, pubKey cryptotypes.PubKey, sign
 // QueryGasPrice takes a priority and an app gRPC client.
 // Returns the current network gas price corresponding to the provided priority.
 // More on the gas price estimation can be found in docs/architecture/adr-023-gas-used-and-gas-price-estimation.md
+// Deprecated: use TxClient.EstimateGasPrice
```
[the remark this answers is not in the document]
renaynay: @rach-id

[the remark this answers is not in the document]
rach-id: https://github.com/celestiaorg/celestia-app/issues/4481

pkg/user/tx_client_test.go
```diff
@@ -255,12 +264,36 @@ func TestEvictions(t *testing.T) {
 	require.Equal(t, seqBeforeEviction, seqAfterEviction)
 }
 
-func (suite *TxClientTestSuite) TestGasEstimation() {
+// TestSeparateGasEstimatorService ensures that if the WithEstimatorService
+// option is provided to the tx client, the separate gas estimator service is
+// used to estimate gas price and usage instead of the default connection.
+func TestSeparateGasEstimatorService(t *testing.T) {
```
cmwaters, line 270:
```suggestion
func TestWithEstimatorService(t *testing.T) {
```
nit to make it clearer

cmwaters (dismissed):
Just to confirm, the client still can optionally provide a multiplier which multiplies the gas limit response?

Ideally, no multiplier is needed since you're trusting the endpoint that provided the estimation that it is correct. The multiplier should be added to the server side

cmwaters (approved):
👍

evan-forbes (approved):

Issue #4359, opened 2025-02-27
Increase confidence in unit testing by using expected errors


damiannolan:
### Summary

Many unit tests in the repository use golang table tests with: `expError bool` and with this condition assert the presence of an error. 

This can be leaky as when touching implementation or test code related to the tests, subtle problems can slip through - where the original expectation of the test is not covered, however failure in a different way is still flagged as a passing test. 

### Problem Definition

--- 

### Proposal

Changing the test structure to use `expError error` gives stronger guarantees on testing. The test can assert the presence of a specific error using `require.ErrorIs(t, err, tc.expError)`. When testing a happy path flow we simply set `expError` to `nil`.

I would recommend backlogging this issue and tackling it incrementally where possible.

PR #4465, opened 2025-03-27
fix: MinGasPFB decorator check for nested authz messages 

rootulp:
Previously the MinGasPFB decorator was not looking inside authz `MsgExec`s.

### DIFF
--- a/x/blob/ante/ante_test.go
+++ b/x/blob/ante/ante_test.go
@@ -2,6 +2,7 @@ package ante_test
 
 import (
 	"fmt"
+	"math"
 	"testing"
 
 	"github.com/celestiaorg/celestia-app/v3/app"
@@ -12,6 +13,8 @@ import (
 	blob "github.com/celestiaorg/celestia-app/v3/x/blob/types"
 	"github.com/celestiaorg/go-square/v2/share"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/x/authz"
 	"github.com/stretchr/testify/require"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	"github.com/tendermint/tendermint/proto/tendermint/version"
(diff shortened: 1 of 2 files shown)


x/blob/ante/ante_test.go
```diff
@@ -141,6 +145,48 @@ func TestPFBAnteHandler(t *testing.T) {
...
 
+// TestMinGasPFBDecoratorWithMsgExec tests that the MinGasPFBDecorator rejects a
+// MsgExec that contains a MsgExec with a MsgPayForBlob where the MsgPayForBlob
+// gas cost is greater than the tx's gas limit.
+func TestMinGasPFBDecoratorWithMsgExec(t *testing.T) {
+	anteHandler := ante.NewMinGasPFBDecorator(mockBlobKeeper{})
+	txConfig := encoding.MakeConfig(app.ModuleEncodingRegisters...).TxConfig
+
+	// Create a context with a gas meter with a high gas limit
+	gasLimit := uint64(10000)
```
evan-forbes, line 156: doesn't a normal send usually use 70k?

x/blob/ante/ante_test.go
```diff
@@ -141,6 +145,48 @@ func TestPFBAnteHandler(t *testing.T) {
...
+	require.ErrorIs(t, err, sdkerrors.ErrInsufficientFee)
+
+	// Create a MsgExec that wraps a MsgExec that contains a MsgPayForBlobs with a huge gas cost
+	nestedMsgExec := authz.NewMsgExec(sdk.AccAddress{}, []sdk.Msg{&msgExec})
+	require.NoError(t, txBuilder.SetMsgs(&nestedMsgExec))
+	tx = txBuilder.GetTx()
+
+	// Run the ante handler
+	_, err = anteHandler.AnteHandle(ctx, tx, false, mockNext)
+	require.Error(t, err)
```
evan-forbes, line 187: can we check that we're getting the error we expect here just out of good habit? https://github.com/celestiaorg/celestia-app/issues/4359

evan-forbes (approved):
I don't think we'll have to gate this by height in order to blocksync, but maybe?

rach-id (approved):

x/blob/ante/ante_test.go
rootulp: Great catch.

x/blob/ante/ante_test.go
rootulp: TBH I have no idea. Just found a random PFB on Celenium that uses 10 million gas so will increase this 

Ref: https://celenium.io/tx/6aa170e9f68fda9382cd35559117579042e6eab56df567ac78c1079e34c7707c?tab=messages

x/blob/ante/ante_test.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/4465/commits/9059f172185d9b44ade3c0d82148a7ad1c14bef9

x/blob/ante/ante_test.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/4465/commits/9059f172185d9b44ade3c0d82148a7ad1c14bef9

evan-forbes (approved):

rach-id (approved):

PR #4474, opened 2025-03-28
feat: enable ability to start grpc server in standalone mode

chatton:
## Overview

~~NOTE: temporarily pinning to my fork with these changes https://github.com/celestiaorg/cosmos-sdk/pull/439~~

This PR enables the ability start the gRPC server in standalone mode. A v3 binary with this functionality will be required for the multiplexer to work.

The logic is largely taken from the `startInProcess` function which starts the gRPC and API servers based on flags.

Testing I performed the following steps.

1. Build this binary 

```bash
make build
```
```
--> Updating go.mod
--> Building build/celestia-appd`
```

2. Compress this binary as a V3 celestia app in the multiplexer

```bash
COPYFILE_DISABLE=1 tar -cvzf internal/embedding/celestia-app_Darwin_v3_arm64.tar.gz ~/checkouts/celestia-forks/celestia-app/build/celestia-appd
```

3. Install the multiplexer (it will embed this binary based on the path)
```bash
make install-multiplexer # this happens in 01builders/celestia-app:sdk-v0.50.x
```

4. Run init script
```bash
MULTIPLEXER=true ./local_devnet/scripts/init.sh
```

5. Start

```bash
celestia-appd start --force-no-bbr --log_level=debug
```

6. Verify grpc service is running (note: in a separate process)

```bash
sudo lsof -i -nP | grep LISTEN | grep celestia
```

```
celestia- 5217        chatton   22u  IPv4 0x6605a6217c0ef7a8      0t0    TCP 127.0.0.1:26657 (LISTEN)
celestia- 5217        chatton   23u  IPv4 0xbc5901eb1fec9829      0t0    TCP 127.0.0.1:6060 (LISTEN)
celestia- 5217        chatton   25u  IPv6 0x735eb6234d226144      0t0    TCP *:26656 (LISTEN)
celestia- 5219        chatton   14u  IPv4 0xf8e00f91393a779c      0t0    TCP 127.0.0.1:1317 (LISTEN)
celestia- 5219        chatton   17u  IPv4 0xab216207f49e3ad0      0t0    TCP 127.0.0.1:9090 (LISTEN)
celestia- 5219        chatton   18u  IPv6 0x55152ef388302569      0t0    TCP *:26658 (LISTEN)
```

previously, the 9090 (grpc) and 1317 (abi) services were not running.


go.mod
```diff
@@ -256,7 +256,7 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.28.0-sdk-v0.46.16
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.28.0-sdk-v0.46.16.0.20250331072412-0673d7c9b833
```
chatton, line 259: once v3.x tax is built with this version of the sdk, it should be fully possible to perform a signalling upgrade from an embedded v3 to a native v4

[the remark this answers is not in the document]
chatton: Not sure what the standard process is here, this pin would change on the next tag of the sdk

chatton:
looking into failing test now

cmd/celestia-appd/cmd/start.go
```diff
@@ -232,6 +232,107 @@ func startStandAlone(ctx *server.Context, appCreator srvrtypes.AppCreator) error
...
+					grpc.MaxCallSendMsgSize(maxSendMsgSize),
+				),
+			)
+			if err != nil {
+				return err
+			}
+
+			clientCtx = clientCtx.WithGRPCClient(grpcClient)
+			ctx.Logger.Debug("grpc client assigned to client context", "target", config.GRPC.Address)
+		}
```
rootulp, line 288: [nit] is it possible to refactor this block so that there's less layers of indentation? Something like:

```go
if config.GRPC.Enable {
// setup GRPC
}

if config.API.Enable {
// setup API
}
```

Currently, there are two `if config.GRPC.Enable` blocks which seems repetitive and not immediately obvious why one needs to occur inside the `if.config.API.Enable` block

go.mod
rootulp, line 259: Just created a release for you: https://github.com/celestiaorg/cosmos-sdk/releases/tag/v1.28.1-sdk-v0.46.16

```suggestion
	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.28.1-sdk-v0.46.16
```

rootulp (commented):
Overall LGTM, ty! One nit and one suggestion

chatton:
@rootul thanks for suggestions! Based on your feedback I broke up existing logic in both functions to handle the 3 distinct cases. I ran tests locally and everything passed, and also verified that building with the new binary works for the embedded use case.

- GRPC enabled
- GRPC web enabled
- API enabled

I don't believe there is actually a dependency on API for the GRPC  flag to be enabled.

cmd/celestia-appd/cmd/start.go
```diff
@@ -493,18 +481,88 @@ func startInProcess(ctx *server.Context, clientCtx client.Context, appCreator sr
 			_ = tmNode.Stop()
 			_ = app.Close()
 		}
-
-		if apiSrv != nil {
-			_ = apiSrv.Close()
-		}
```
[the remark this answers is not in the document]
chatton: this should be okay since the tmNode is still cleaned up before the api server is closed do to the last in first out nature of defer statements.

evan-forbes (approved):
lgtm, I think we just need to fix the simple linting errors

rootulp (dismissed):
LGTM after lint passes. Thanks for refactoring to remove the layer of indentation.

rootulp (approved):

PR #4487, opened 2025-03-31
chore: fix `find` issue with `-prune` in proto search

famouswizard:
## Overview

I encountered an issue where the `-prune` option in the `find` command wasn't working as expected. It prevented the search in the `./cosmos` directory, but the `-name '*.proto'` still applied to all files, causing unexpected results.  

To fix this, I replaced `-prune` with `! -path`, which now correctly excludes the `./cosmos` directory from the search. This should prevent unnecessary files from being included in the search for `.proto` files.

### DIFF
--- a/scripts/protocgen.sh
+++ b/scripts/protocgen.sh
@@ -12,7 +12,7 @@ protoc_gen_gocosmos() {
 protoc_gen_gocosmos
 
 cd proto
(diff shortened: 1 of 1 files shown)


rach-id:
not relevant and doesn't add any value

PR #4488, opened 2025-03-31
Align proto field types and fix formatting in mint and query modules

Marcofann:
1. Fixed formatting of the `previous_block_time` field in `Minter`:
  - Moved `[(gogoproto.stdtime) = true]` inline with the field declaration for proper syntax and readability.

2. Updated data types in `query.proto`:
  - Changed `bytes` → `string` for `inflation_rate` and `annual_provisions` fields in response messages to match the definition in the `Minter` struct.
  - This ensures consistency in field types across `Minter` and `Query*Response` messages using `cosmos.Dec`.


rach-id:
not relevant and doesn't provide any value

PR #4489, opened 2025-03-31
feat: embed a gas multipler on the server's gas estimate

cmwaters:
When using `Simulate`, the node can occasionally return a value that would be less than the gas required to submit the transaction at a slightly different time. To lean on the conservative side, this PR adds a 1.1 multiplier to whatever gas is returned by `Simulate`

### DIFF
--- a/app/grpc/gasestimation/gas_estimator.go
+++ b/app/grpc/gasestimation/gas_estimator.go
@@ -17,6 +17,10 @@ import (
 	gogogrpc "github.com/gogo/protobuf/grpc"
 )
 
+// gasMultiplier is the multiplier for the gas limit. It's used to account for the fact that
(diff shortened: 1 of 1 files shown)


rach-id (commented):
please update the `TestEstimateGasUsed` to take this multiplier in consideration.

Also, if we're using this multiplier here, should we remove it from the tx client?

cmwaters:
> Also, if we're using this multiplier here, should we remove it from the tx client?

Yes @renaynay was going to remove it from the other PR. I think she might cherry-pick this commit into that PR as well

evan-forbes (approved):
:+1:

cmwaters:
Closing as the commit is cherry-picked to https://github.com/celestiaorg/celestia-app/pull/4459

PR #4493, opened 2025-04-01
feat!: enforce max tx size in process proposal

### LINKED ISSUE #4087: Set max tx size cap in ProcessProposal
## Description

In v4, we should enforce max tx size cap in ProcessProposal.

## Acceptance Criteria

- [ ] The check is enforced and thoroughly tested in integration tests with expected outputs and logs
- [ ] Reconsider the current max tx size ante handler

ninabarbakadze:
## Overview

Fixes #4087 

## Testing logs

Log from the process proposal test confirming that large transaction was rejected
Screenshot 2025-04-01 at 18 49 42

## Docs

I removed the old and useless ante handler therefore I removed the mention of it in the Ante Handler v3 doc.

## Misc

Do we want to mention this in the meta cip? @evan-forbes @rootulp


app/test/process_proposal_test.go
```diff
@@ -328,6 +328,23 @@ func TestProcessProposal(t *testing.T) {
...
+				// set the blob size to maxTxSize so that the raw transaction size will exceeds the max tx size limit
+				blob, err := share.NewBlob(ns1, bytes.Repeat([]byte{1}, maxTxSize), appconsts.DefaultShareVersion, nil)
+				require.NoError(t, err)
+				rawTx, _, err := signer.CreatePayForBlobs(accounts[0], []*share.Blob{blob}, user.SetGasLimit(100000), user.SetFee(100000))
+				require.NoError(t, err)
+				// override the last valid blob tx with large one that exceeds the max tx size limit
+				// proposal block should be rejected
+				d.Txs[2] = rawTx
+			},
+			appVersion:     v3.Version,
```
rootulp, line 345: [blocking] this shouldn't apply to v3 because it's breaking. It should only apply to v4.

specs/src/ante_handler_v3.md
```diff
@@ -3,7 +3,6 @@
 The AnteHandler chains together several decorators to ensure the following criteria are met for app version 3:
 
 - The tx does not contain any messages that are unsupported by the current app version. See `MsgVersioningGateKeeper`.
-- The tx size is not larger than the application's configured versioned constant [MaxTxSize](https://github.com/celestiaorg/celestia-app/blob/8ba82c1b872b7f5686d9bb91b93a0442223d7bb2/pkg/appconsts/v3/app_consts.go#L9).
```
rootulp, line 6: We need a separate PR that targets v3.x and removes the max tx size ante handler from that branch

rootulp (commented):
[blocking] since this is a change to process proposal, it's breaking and should only be included in main and for app version v4.

evan-forbes (commented):
using the exising mempool mechanism is good, but we want to ensure that validators don't accidently change that value and then start creating invalid blocks

to make sure we're not making similar mistakes to last time, and since this is a restriction on a user facing function, I think having an integration tests using the rpc/mempool/prepare/process proposal via `testnode` is worthwhile. The test should include a config that would otherwise allow large txs in the mempool to ensure that we're blocking that transaction even if the config is too large.

app/errors/errors.go
```diff
@@ -11,5 +11,5 @@ const AppErrorsCodespace = "app"
 // Application error codes start from 11000 to avoid conflicts with other modules
 var (
 	// ErrTxExceedsMaxSize is returned when a transaction size exceeds the maximum allowed limit
-	ErrTxExceedsMaxSize = errors.Register(AppErrorsCodespace, 11142, "transaction size exceeds maximum allowed limit")
+	ErrTxExceedsMaxSize = errors.Register(AppErrorsCodespace, 11142, "tx too large")
```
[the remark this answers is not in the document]
ninabarbakadze: yeah they do. I just thought that the original message was extremely verbose. You're right tho.

specs/src/ante_handler_v3.md
ninabarbakadze: https://github.com/celestiaorg/celestia-app/issues/4554

ninabarbakadze:
> using the exising mempool mechanism is good, but we want to ensure that validators don't accidently change that value and then start creating invalid blocks
> 
> to make sure we're not making similar mistakes to last time, and since this is a restriction on a user facing function, I think having an integration tests using the rpc/mempool/prepare/process proposal via `testnode` is worthwhile. The test should include a config that would otherwise allow large txs in the mempool to ensure that we're blocking that transaction even if the config is too large.

Added the integration test using testnode with a config that sets MaxTxSize larger than what's enforced by the app. Transactions will get rejected in CheckTx and not hit Prepare/Process. Should I try manipulating the context so it skips CheckTx or is this sufficient?

ninabarbakadze:
> ## Overview
> Fixes #4087
> 
> ## Testing logs
> Log from the process proposal test confirming that large transaction was rejected Screenshot 2025-04-01 at 18 49 42
> 
> ## Docs
> I removed the old and useless ante handler therefore I removed the mention of it in the Ante Handler v3 doc.
> 
> ## Misc
> Do we want to mention this in the meta cip? @evan-forbes @rootulp

This specific test case in test-race is failing on my branch and multiple other active PRs. Everything is working locally so might also be a flake.

Screenshot 2025-04-14 at 12 49 25

specs/src/ante_handler_v3.md
rootulp: [question] are we actually going to do https://github.com/celestiaorg/celestia-app/issues/4554? If not, then why remove this line from the v3 specs?

rootulp: Ohh is this because there is no ante_handler_v4.md yet?

rootulp (dismissed):
LGTM, thanks!

app/ante/max_tx_size.go
```diff
@@ -1,28 +0,0 @@
...
-// AnteHandle implements the AnteHandler interface. It ensures that tx size is under application's configured threshold.
-func (d MaxTxSizeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
-	currentTxSize := len(ctx.TxBytes())
-	maxTxSize := appconsts.DefaultMaxTxSize
-	if currentTxSize > maxTxSize {
-		bytesOverLimit := currentTxSize - maxTxSize
-		return ctx, fmt.Errorf("tx size %d bytes is larger than the application's configured threshold of %d bytes. Please reduce the size by %d bytes", currentTxSize, maxTxSize, bytesOverLimit)
-	}
-	return next(ctx, tx, simulate)
-}
```
evan-forbes, line 28: why do we want to delete this?

app/test/max_tx_size_integration_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	accounts := testnode.RandomAccounts(10)
+	// Set the max block size to 8MB
+	cparams := testnode.DefaultConsensusParams()
+	cparams.Block.MaxBytes = 8_388_608 // 8MiB
+	cfg := testnode.DefaultConfig().WithFundedAccounts(accounts...).WithConsensusParams(cparams)
+
+	// Set the max tx bytes to 3MiB in the node's mempool
+	mempoolMaxTxBytes := appv4.MaxTxSize + 1_048_576 // 2MiB + 1MiB
```
evan-forbes, line 31: [blocking]

apologies I don't think I was clear enough in the last review. I don't think we should be relying only on this config, if validators change this then they will allow txs in their mempool that could never make it in a block. so I think we still need the antehandler.

ninabarbakadze: In this test, I'm increasing the MaxTxBytes in node's mempool config to be larger than app's threshold and we see that txs get rejected with the expected log and error code.

evan-forbes (approved):

app/test/max_tx_size_integration_test.go
```diff
@@ -0,0 +1,68 @@
...
+// by the application even if the validator node's local mempool config allows them.
+func TestTxsOverMaxTxSizeGetRejected(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping max tx size integration test in short mode.")
+	}
+
+	accounts := testnode.RandomAccounts(10)
+	// Set the max block size to 8MB
+	cparams := testnode.DefaultConsensusParams()
+	cparams.Block.MaxBytes = 8_388_608 // 8MiB
```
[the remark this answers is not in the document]
ninabarbakadze: https://github.com/celestiaorg/celestia-app/issues/4585

rootulp (approved):
Still LGTM even though there are a few unresolved non blocking comments

specs/src/ante_handler_v3.md
cmwaters: Yeah I think this should stay because we're not changing it but perhaps we need a v4 section to add that the capped size is now enforced by consensus

cmwaters (approved):

PR #4498, opened 2025-04-02
lint: enable mirror to avoiding unnecessary []byte/string conversion

argentpapa:


rach-id (commented):
change doesnt seem to be necessary as it's targeting an old module. closing

PR #4502, opened 2025-04-02
Update mocha-block-sync.sh script message for better clarity

vtjl10:
A typo in the text where "Settings" was used instead of "Setting." Since it refers to a singular action, I corrected it to "Setting."


rach-id:
not enough value

PR #4504, opened 2025-04-03
fix(app/errors)!: `IsInsufficientMinGasPrice `--> `IsInsufficientFee`

renaynay:
This is BREAKING but AFAIK node is the only user of this error (as it was created for node's tx submission code).

I want to change the name for it to be clearer what we're actually checking.

Also `ParseInsufficientMinGasPrice` is now unused in celestia-node so you can consider removing.

### DIFF
--- a/app/errors/insufficient_gas_price_test.go
+++ b/app/errors/insufficient_gas_price_test.go
@@ -5,6 +5,13 @@ import (
 	"testing"
 
 	"cosmossdk.io/errors"
+	"github.com/celestiaorg/go-square/v2/share"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	"github.com/stretchr/testify/require"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+
 	"github.com/celestiaorg/celestia-app/v3/app"
 	"github.com/celestiaorg/celestia-app/v3/app/encoding"
 	apperr "github.com/celestiaorg/celestia-app/v3/app/errors"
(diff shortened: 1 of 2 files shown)


rach-id (commented):
why not do it in a backwards compatible way? define `IsInsufficientFee` and add the `Deprecated` doc to `IsInsufficientMinGasPrice`? This way we can roll it out faster

app/errors/insufficient_gas_price_test.go
```diff
@@ -56,7 +57,7 @@ func TestInsufficientMinGasPriceIntegration(t *testing.T) {
 	require.NoError(t, err)
 
 	_, err = anteHandler(ctx, sdkTx, false)
-	require.True(t, apperr.IsInsufficientMinGasPrice(err))
+	require.True(t, apperr.IsInsufficientFee(err))
```
rootulp, line 60: [nit] this test could be renamed `TestInsufficientMinGasPriceIntegration` -> `TestInsufficientFee`

app/errors/insufficient_gas_price_test.go
```diff
@@ -132,7 +133,7 @@ func TestInsufficientMinGasPriceTable(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			require.Equal(t, tc.isInsufficientMinGasPriceErr, apperr.IsInsufficientMinGasPrice(tc.err))
+			require.Equal(t, tc.isInsufficientMinGasPriceErr, apperr.IsInsufficientFee(tc.err))
```
rootulp, line 136: [nit] the test case struct

```
isInsufficientMinGasPriceErr => isInsufficientFeeErr
```

rootulp:
@renaynay I applied a variation of this PR to v3.x: https://github.com/celestiaorg/celestia-app/pull/4510 so that celestia-node can adopt the new function name before the v4.x celestia-app release.

rootulp:
Nope. We want to keep this PR targeting `main` because it contains the breaking change.

evan-forbes (approved):

ninabarbakadze (approved):

evan-forbes (approved):

rootulp (approved):

PR #4510, opened 2025-04-04
refactor: deprecate IsInsufficientMinGasPrice

rootulp:
Motivation https://github.com/celestiaorg/celestia-app/pull/4504# makes a breaking change to `main` which will be released in v4.x. It would be nice if deprecate the previous function name on v3.x so that celestia-node can upgrade to a v3.x release with the new function.

cc: @renaynay


renaynay (approved):
Thank you @rootulp

ninabarbakadze (approved):

PR #4511, opened 2025-04-04
fix Update data_structures.md

MonkeyKing44:
fix old to a new one


rach-id:
no need for this change as there is already a redirection.

PR #4526, opened 2025-04-10
test: add a core block API foolproof test

rach-id:
## Overview

Adds a test to make sure the core block API never gets forgotten again.

Question: I guess this should be backported to v4.x too, no?


rach-id:
Yes, because main wasn't bumped to the latest cosmos-sdk release. I left questions in the PR description to see whether to bump main too, or make this PR only target v3, or maybe v3 and v4? or maybe main, v3 and v4 😸

rootulp:
> I see that the new cosmos-sdk bump was only done to v3.x and not main. why is that? and should this PR be directed to v3.x only?

celestia-app main uses cosmos-sdk v0.50.x
celestia-app v3.x uses cosmos-sdk v0.46.x

I [asked](https://github.com/celestiaorg/cosmos-sdk/pull/452#issuecomment-2789910321) if we needed a variation of Cian's fix on v0.50.x but seems like no.

rootulp (approved):
Great work on adding this. Thanks!

evan-forbes (approved):

Issue #4523, opened 2025-04-08
Remove redundant height conditionals in x/blob ante decorators


damiannolan:
```
// TODO: fix hack that forces the max square size for the first height to
// 64. This is due to our fork of the sdk not initializing state before
// BeginBlock of the first block. This is remedied in versions of the sdk
// and comet that have full support of PrepareProposal, although
// celestia-app does not currently use those. see this PR for more details
// https://github.com/cosmos/cosmos-sdk/pull/14505
```

refs: 
- https://github.com/celestiaorg/celestia-app/pull/4400#discussion_r2031123447
- https://github.com/celestiaorg/celestia-app/pull/4400#discussion_r2031124448

PR #4539, opened 2025-04-11
chore: resolve TODO for block height 1

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4523


ninabarbakadze (approved):

rootulp:
Looks like merge conflicts

rach-id (approved):
does it make sense to add a test for these cases? 

We can spin up a testnode with a very long block time, ensuring it doesn't advance to heights beyond 1, and test if we can get these values?

also, to be sure this doesn't break with any sdk upgrade while we don't have the todo to remember to check it.

rootulp:
> We can spin up a testnode with a very long block time, ensuring it doesn't advance to heights beyond 1, and test if we can get these values?

We won't hit it on supported networks (Arabica, Mocha, Mainnet) because they're already past block height 1. 

If we do hit it on a local devnet then we can invest more time into writing a test / fixing it but ideally it just works now that we've upgraded cosmos SDK.

damiannolan (approved):
🙏🏻

Issue #4513, opened 2025-04-07
Delete test/util blobstreamModuleName (stale/legacy)


damiannolan:
Just need to route out whatever coin minting is going on with this module name in the test code. See `test/util/common.go` for usages

> I will create an issue to delete this in future. It seems there is some usage below with minting coins for this as well that need to be pulled out

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-app/pull/4400#discussion_r2031097610_

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/4575/files

PR #4543, opened 2025-04-11
chore: remove blobstream module name

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4513

### DIFF
--- a/test/util/common.go
+++ b/test/util/common.go
@@ -46,9 +46,6 @@ import (
 	"github.com/celestiaorg/celestia-app/v4/x/blob"
 )
 
-// TODO: This probably should be deleted.
-var blobstreamModuleName = "blobstream"
-
 var (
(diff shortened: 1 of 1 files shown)


rach-id (dismissed):
A ton of this code can be deleted because it's only used in the blobstream module testing. Maybe, even if we still can't just delete the whole module, we delete its testing suite to reduce the things we need to maintain, and work that comes with it especially when it comes to bumping sdk and related dependencies.

rootulp:
> Maybe, even if we still can't just delete the whole module

x/blobstream no longer exists on `main` https://github.com/celestiaorg/celestia-app/tree/main/x

> we delete its testing suite to reduce the things we need to maintain

Down to delete more code. Do you mind creating an issue with what can/should be deleted now that x/blobstream is no longer on main?

Issue #4496, opened 2025-04-02
Remove antehandlers from `main` that don't apply to app version 4


rootulp:
## Context

Blocked on https://github.com/celestiaorg/celestia-app/pull/4400 

## Problem

`main` no longer needs to support historical app versions because the multiplexer will be used to delegate syncing of historical blocks to a v3.x release. This means that `main` no longer needs to support ante handlers that are deprecated.

## Proposal

Remove this antehandler on `main`

```go
// Ensure that the tx's total blob size is <= the max blob size.
// Only applies to app version == 1.
blobante.NewMaxTotalBlobSizeDecorator(blobKeeper),
```

PR #4545, opened 2025-04-11
chore: remove antehandlers that do not apply to v4

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4496


rach-id (dismissed):

app/ante/ante.go
```diff
@@ -70,12 +69,8 @@ func NewAnteHandler(
 		// Contract: must be called after all decorators that consume gas.
 		// Note: does not consume gas from the gas meter.
 		blobante.NewMinGasPFBDecorator(blobKeeper),
-		// Ensure that the tx's total blob size is <= the max blob size.
-		// Only applies to app version == 1.
-		blobante.NewMaxTotalBlobSizeDecorator(blobKeeper),
```
damiannolan, line 75: nice 🤝

damiannolan (approved):
Beautiful red diffs!

evan-forbes (dismissed):

app/test/big_blob_test.go
```diff
@@ -54,40 +53,6 @@ func (s *BigBlobSuite) SetupSuite() {
...
-			subCtx, cancel := context.WithTimeout(s.cctx.GoContext(), 30*time.Second)
-			defer cancel()
-			res, err := txClient.SubmitPayForBlob(subCtx, []*share.Blob{tc.blob}, user.SetGasLimitAndGasPrice(1e9, appconsts.DefaultMinGasPrice))
-			require.Error(t, err)
-			require.Nil(t, res)
-			code := err.(*user.BroadcastTxError).Code
-			require.Equal(t, tc.want, code, err.Error())
-		})
-	}
-}
```
evan-forbes, line 89: why remove this test?

don't we want to keep ensuring that too large of txs cannot be submitted?

app/test/check_tx_test.go
```diff
@@ -173,33 +173,21 @@ func TestCheckTx(t *testing.T) {
...
-			checkType: abci.CheckTxType_New,
-			getTx: func() []byte {
-				signer := signers[9]
-				_, blobs := blobfactory.RandMsgPayForBlobsWithSigner(random.New(), signer.Account(accs[9]).Address().String(), 2_000_000, 1)
-				tx, _, err := signer.CreatePayForBlobs(accs[9], blobs, opts...)
-				require.NoError(t, err)
-				return tx
-			},
-			expectedABCICode: blobtypes.ErrTotalBlobSizeTooLarge.ABCICode(),
-		},
```
evan-forbes, line 194: same here, I think we just want to change the errors

evan-forbes (commented):
I don't think we should remove all of these tests, but otherwise lgtm

app/test/big_blob_test.go
rootulp: `ErrTotalBlobSizeTooLarge ` isn't thrown in v4. It was thrown by the ante handler that was removed in this PR but that ante handler only applies to app version v1.

rootulp: `ErrTxExceedsMaxSize` is checked by another test csae in this suite.

evan-forbes: I see, thanks for clarifying!

evan-forbes (approved):

rootulp:
We want to remove the tests that were asserting for the old error `ErrTotalBlobSizeTooLarge` that the deprecated ante handler was using. 

There are existing tests for `ErrTxExceedsMaxSize` which is the error that process proposal returns now so we don't need to modify the old tests.

damiannolan (approved):

Issue #4552, opened 2025-04-14
Ensure the default binary not named multiplexer


chatton:
I don't know if we want the default to actually be the multiplexer. Users shouldn't need to know what a multiplexer is - they should just build the binary as they did normally at it have the same behaviour as before i.e. single binary syncs

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/pull/4400#discussion_r2039648269_

We can make sure that the fact that the default binary will be a "multiplexer" does not cause confusion in the naming conventions and image names. 

Perhaps `make build` builds the multiplexer, and there is a special make target for building the native app, something along the lines of `make build-without-multiplexer`

rootulp:
Can acceptance criteria of this issue also include documenting the new `make` commands for the non multiplexer version?

1. Create a PR to update https://github.com/celestiaorg/docs
1. Update README to describe `make install-without-multiplexer`

Note these can be done in a follow-up issue:
1. Draft docs to include in the v4.0.0-arabica release notes
1. Share these docs with Luka from POPs so that he can include them in announcements to the community

Issue #4551, opened 2025-04-14
Move muiltiplexer (nova) into celestia-app repo


chatton:
The multiplexer will be moved into a separate go.mod in this directory, and use a local pin in the celestia-app go.mod.

The majority of the work has been done [here](https://github.com/01builders/celestia-app/pull/85) by @julienrbrt 

Note: there will be a few small conflicts with the E2E test being worked on [here](https://github.com/01builders/celestia-app/pull/82) so it might be worth doing the embedded file name change in the this PR if it will be merged first.

rootulp:
nit: can we call it multiplexer in this repo? Seems like a more self-descriptive name than nova.

chatton:
> nit: can we call it multiplexer in this repo? Seems like a more self-descriptive name than nova.

Ah yes that is the idea, in the PR it's being named `multiplexer`, `nova` as a name shouldn't appear anywhere in the repo when it's merged.

PR #4553, opened 2025-04-14
feat: add multiplexer

julienrbrt:
Closes: https://github.com/celestiaorg/celestia-app/issues/4551
Closes: https://github.com/celestiaorg/celestia-app/issues/4552

Supersede https://github.com/01builders/celestia-app/pull/85

Adds the multiplexer from https://github.com/01builders/nova.

Unfortunately using --no-squash doesn't work because of LFS object in history.
I would need to rewrite history in 01builders/nova to remove LFS, but it would break all commit signatures, so won't do it.

git subtree add --prefix=multiplexer novabinary main --squash


internal/embedding/celestia-app_darwin_v3_amd64.tar.gz
julienrbrt, line 1: We do not want to commit those ones on git, so we can have a placeholder

multiplexer/appd/run.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// Wait for the process to exit
+	_, err = proc.Wait()
+	if err != nil {
+		log.Printf("Error waiting for process to exit: %v", err)
+	}
+
+	a.pid = AppdStopped
+	return nil
+}
```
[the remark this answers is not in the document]
julienrbrt: this is why we skip the error, we really need to make sure it exits.

multiplexer/abci/multiplexer.go
```diff
@@ -0,0 +1,662 @@
...
+
+	switch m.activeVersion.ABCIVersion {
+	case ABCIClientVersion1:
+		return NewRemoteABCIClientV1(m.conn, m.chainID), nil
+	case ABCIClientVersion2:
+		return NewRemoteABCIClientV2(m.conn), nil
+	}
+
+	return nil, fmt.Errorf("unknown ABCI client version %d", m.activeVersion.ABCIVersion)
+}
```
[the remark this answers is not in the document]
julienrbrt: should be only one frequent start

multiplexer/abci/versions.go
```diff
@@ -0,0 +1,100 @@
...
+
+	for _, version := range v {
+		if version.AppVersion == appVersion {
+			return version, nil
+		}
+	}
+
+	// return the lowest version if the exact version is not found.
+	return lowestVersion, nil
+}
```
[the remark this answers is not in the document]
julienrbrt: good feedback however we do not want to do that as it makes the UX unclear. Additionally if we were to do that we would need to use the closest highest one, and that snippet doesn't do that.

multiplexer/abci/abci.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	resp, err := app.Commit()
+	if err != nil {
+		return nil, fmt.Errorf("failed to commit: %w", err)
+	}
+
+	// after a successful commit, we start using the app version specified in FinalizeBlock.
+	m.appVersion = m.nextAppVersion
+
```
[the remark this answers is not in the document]
julienrbrt: ABCI is called thread safe by comet, this isn't needed.

Makefile
```diff
@@ -19,39 +19,49 @@ ldflags = -X github.com/cosmos/cosmos-sdk/version.Name=celestia-app \
...
+install: check-bbr
+	@echo "--> Download embedded binaries for v3"
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Darwin_arm64.tar.gz -O internal/embedding/celestia-app_darwin_v3_arm64.tar.gz
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Linux_arm64.tar.gz -O internal/embedding/celestia-app_linux_v3_arm64.tar.gz
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Darwin_x86_64.tar.gz -O internal/embedding/celestia-app_darwin_v3_amd64.tar.gz
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Linux_x86_64.tar.gz -O internal/embedding/celestia-app_linux_v3_amd64.tar.gz
 	@echo "--> Installing celestia-appd with multiplexer support"
 	@go install $(BUILD_FLAGS_MULTIPLEXER) ./cmd/celestia-appd
-.PHONY: install-multiplexer
+.PHONY: install
```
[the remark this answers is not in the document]
julienrbrt: yes, it will be done in a follow-up https://github.com/celestiaorg/celestia-app/issues/4565

Makefile
```diff
@@ -48,7 +50,13 @@ install: check-bbr
 .PHONY: install
 
 ## install-multiplexer: Build and install the multiplexer version of celestia-appd into the $GOPATH/bin directory.
+## TODO: Improve logic here and in goreleaser to make it future proof and less expensive.
 install-multiplexer: check-bbr
+	@echo "--> Download embedded binaries for v3"
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Darwin_arm64.tar.gz -O internal/embedding/celestia-app_darwin_v3_arm64.tar.gz
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Linux_arm64.tar.gz -O internal/embedding/celestia-app_linux_v3_arm64.tar.gz
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Darwin_x86_64.tar.gz -O internal/embedding/celestia-app_darwin_v3_amd64.tar.gz
+	wget https://github.com/celestiaorg/celestia-app/releases/download/$(CELESTIA_V3_VERSION)/celestia-app_Linux_x86_64.tar.gz -O internal/embedding/celestia-app_linux_v3_amd64.tar.gz
```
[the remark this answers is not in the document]
julienrbrt: https://github.com/celestiaorg/celestia-app/pull/4553#discussion_r2041878662

multiplexer/appd/internal/testdata/data_test.go
```diff
@@ -0,0 +1,56 @@
+package testdata
```
chatton, line 1: can delete

multiplexer/abci/versions.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Default flags for standalone apps.
+	return append(args,
+		"--grpc.enable",
+		"--api.enable",
+		"--api.swagger=false",
+		"--with-tendermint=false",
+		"--transport=grpc",
+	)
```
damiannolan, line 82: Should `--rpc.grpc_laddr=tcp://0.0.0.0:{port}` be a default flag for a standalone app? this will enable the comet grpc server needed for block api.

I just don't like the idea of hardcoding a port number

multiplexer/abci/multiplexer.go
```diff
@@ -0,0 +1,662 @@
...
+	// any returned errors are logged.
+	cleanupFns []func() error
+	// ctx is the context which is passed to the comet, grpc and api server starting functions.
+	ctx context.Context
+	// g is the waitgroup to which the comet, grpc and api server init functions are added to.
+	g *errgroup.Group
+}
+
+// NewVersions returns a list of versions sorted by app version.
+func NewVersions(v ...Version) (Versions, error) {
```
damiannolan, line 92: should we move to versions.go?

multiplexer/abci/multiplexer.go
```diff
@@ -0,0 +1,662 @@
...
+
+	// startApp starts the underlying application, either native or embedded.
+	if err := m.startApp(); err != nil {
+		return err
+	}
+
+	if m.isGrpcOnly() {
+		m.logger.Info("starting node in gRPC only mode; CometBFT is disabled")
+		m.svrCfg.GRPC.Enable = true
+	} else {
```
damiannolan, line 155: this case is always confusing to me...maybe we should add a comment or something. this is for starting an app in standalone mode, but with no comet underneath? is this ever something we want or need?

multiplexer/abci/versions.go
damiannolan: we can at least error in the multiplexer if `rpc.GRPCListenAddr` is empty

multiplexer/go.mod
```diff
@@ -0,0 +1,201 @@
+module github.com/celestiaorg/celestia-app/multiplexer
```
damiannolan, line 1: idk if `module github.com/celestiaorg/celestia-app/mux` is more "go idiomatic", but it doesn't really matter

multiplexer/abci/versions.go
julienrbrt: done https://github.com/celestiaorg/celestia-app/pull/4553/commits/9f0f918f5f0baa08b8c9254949838196ad03dc2c

multiplexer/abci/multiplexer.go
julienrbrt: yes, it is useful when willing to just query state, you download the db, start in grpc only and you are done.

multiplexer/abci/multiplexer.go
rootulp: Seems like a good idea because this function isn't a method on `Multiplexer` and this file is already large

multiplexer/abci/multiplexer.go
```diff
@@ -0,0 +1,662 @@
...
+		if m.svrCfg.API.Enable {
+			metrics, err := startTelemetry(m.svrCfg)
+			if err != nil {
+				return err
+			}
+
+			if err := m.startAPIServer(grpcServer, metrics); err != nil {
+				return err
+			}
+		}
```
rootulp, line 212: [question] why does this only happen inside the `if m.svrCfg.GRPC.Enable {` block? Is it possible to enable API and not GRPC? If not, should we error early if a user passes the CLI flags: `--api.enable true --grpc.enable false`?

multiplexer/abci/multiplexer.go
```diff
@@ -0,0 +1,662 @@
...
+
+// startCmtNode initializes and starts a CometBFT node, sets up cleanup tasks, and assigns it to the Multiplexer instance.
+func (m *Multiplexer) startCmtNode() error {
+	cfg := m.svrCtx.Config
+	nodeKey, err := p2p.LoadOrGenNodeKey(cfg.NodeKeyFile())
+	if err != nil {
+		return err
+	}
+
+	// no latest app set means an embedded app is being used.
```
rootulp, line 576:
```suggestion
	// no native app set means an embedded app is being used.
```

multiplexer/abci/remote_v1.go
```diff
@@ -0,0 +1,636 @@
...
+	cryptov1 "github.com/tendermint/tendermint/proto/tendermint/crypto"
+	typesv1 "github.com/tendermint/tendermint/proto/tendermint/types"
+	versionv1 "github.com/tendermint/tendermint/proto/tendermint/version"
+)
+
+type RemoteABCIClientV1 struct {
+	abciv1.ABCIApplicationClient
+
+	// retainLastHeight is the height is set in finalize block
+	// and returned in commit
```
rootulp, line 22:
```suggestion
	// commitRetainLastHeight is the height that is set in finalize block
	// and returned in commit
```

multiplexer/abci/remote_v1.go
```diff
@@ -0,0 +1,636 @@
...
+// NewRemoteABCIClientV1 returns a new ABCI Client (using ABCI v1).
+// The client behaves like Tendermint for the server side (the application side).
+func NewRemoteABCIClientV1(conn *grpc.ClientConn, chainID string) *RemoteABCIClientV1 {
+	return &RemoteABCIClientV1{
+		ABCIApplicationClient: abciv1.NewABCIApplicationClient(conn),
+		chainID:               chainID,
+	}
+}
+
+// ApplySnapshotChunk implements abciv2.ABCI
```
rootulp, line 39: [question] why does remote abci client v1 accept requests of type abciv2 and return responses of type abciv2?

multiplexer/abci/remote_v1.go
```diff
@@ -0,0 +1,636 @@
...
+
+// Commit implements abciv2.ABCI
+func (a *RemoteABCIClientV1) Commit() (*abciv2.ResponseCommit, error) {
+	return &abciv2.ResponseCommit{
+		RetainHeight: a.commitRetainLastHeight,
+	}, nil
+}
+
+// FinalizeBlock implements abciv2.ABCI
+func (a *RemoteABCIClientV1) FinalizeBlock(req *abciv2.RequestFinalizeBlock) (*abciv2.ResponseFinalizeBlock, error) {
```
rootulp, line 91: I would've expected ABCI v1 to support EndBlock, DeliverTx and not support FinalizeBlock

multiplexer/go.mod
rootulp: [uber nit] I prefer `multiplexer` as the package name over `mux`

rootulp (dismissed):
Overall LGTM. I had a few questions on why the remove_v1.go file has methods that use types from ABCIv2 (i.e. CometBFT) instead of Tendermint

multiplexer/abci/remote_v1.go
julienrbrt: So the app (v3) is using ABCI 1, but consensus is always ABCI 2. In remote v1 we basically map consensus abci calls to the proper abci v1 flow for the app.

multiplexer/abci/abci.go
```diff
@@ -0,0 +1,157 @@
...
+	return app.Query(ctx, req)
+}
+
+func (m *Multiplexer) VerifyVoteExtension(_ context.Context, req *abci.RequestVerifyVoteExtension) (*abci.ResponseVerifyVoteExtension, error) {
+	app, err := m.getApp()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get app for version %d: %w", m.appVersion, err)
+	}
+	return app.VerifyVoteExtension(req)
+}
```
evan-forbes, line 157: similar question to https://github.com/celestiaorg/celestia-app/pull/4553/files#r2042761731

I think I'm misunderstanding something, as it seems like we shouldn't be supporting verify vote extension for tendermint

multiplexer/abci/remote_v1.go
```diff
@@ -0,0 +1,636 @@
...
+
+	"google.golang.org/grpc"
+
+	abciv2 "github.com/cometbft/cometbft/abci/types"
+	cryptov2 "github.com/cometbft/cometbft/proto/tendermint/crypto"
+	typesv2 "github.com/cometbft/cometbft/proto/tendermint/types"
+	abciv1 "github.com/tendermint/tendermint/abci/types"
+	cryptov1 "github.com/tendermint/tendermint/proto/tendermint/crypto"
+	typesv1 "github.com/tendermint/tendermint/proto/tendermint/types"
+	versionv1 "github.com/tendermint/tendermint/proto/tendermint/version"
```
evan-forbes, line 15: [random comment]
the switch to cometbft in the go mod vs using tendermint actually worked out so that we can import two different versions that aren't v1 v2 etc

evan-forbes (commented):
I see the flow to everything, and only had one similar question to Rootul

otherwise LGTM

multiplexer/appd/internal/testdata/data_linux_amd64.go
```diff
@@ -0,0 +1,6 @@
+package testdata
+
+import _ "embed"
+
+//go:embed celestia-app_Linux_x86_64.tar.gz
```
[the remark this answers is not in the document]
julienrbrt: I think we only need them for CI, so i wouldn't remove it.

docker/Dockerfile
```diff
@@ -32,6 +32,9 @@ RUN apk update && apk add --no-cache \
     musl-dev
 WORKDIR /celestia-app
 
+# multiplexer needs to be present due to local pin.
```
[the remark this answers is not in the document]
julienrbrt: cc @chatton

multiplexer/README.md
```diff
@@ -0,0 +1,148 @@
...
+This repository contains an improved Cosmos SDK Upgrading system: the multiplexer (codename: the `multiplexer`).
+Sync from genesis or upgrade a Cosmos SDK chain seamlessly for the node operators, without binary switch.
+
+## Features
+
+- **Automatic Upgrading**: The system automatically upgrades the chain to the latest version of the software after an `AppVersion` change.
+- **No binary switch required**: At the upgrade height, no binary switch is required, a huge improvement compared to solely relying on the `x/upgrade` module.
+- **One binary**: The system uses only one binary for all upgrades, which simplifies  **syncing from genesis**, the upgrade process and reduces the risk of errors.
+- **Passthrough mode**: The system can be used in passthrough mode, where a user can interact with an older binary directly.
+- **Speed of native**: `Multiplexer` is designed to minimize overhead and is only used for the duration of the upgrade process. Once the upgrade is complete, the system use the native app.
```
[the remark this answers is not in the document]
julienrbrt: This means there's no more the small grpc overhead of out of process app. I would keep this tbh.

multiplexer/appd/internal/testdata/data_linux_amd64.go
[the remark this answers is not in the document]
chatton: Can leave it for now if you want, some of the CI tests will be removed in the E2E upgrade being added, so I can address it then maybe if it doesn't make sense to have anymore.

docker/Dockerfile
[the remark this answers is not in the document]
chatton: in the go.mod there is a localpin to the multiplexer, i.e. using whatever is in the multiplexer directory. This is required in order to build the multiplexer image from source of any given branch.

In order to leverage docker layer caching, the go mod is added and go mod download is executed before the source code gets added so subsequent builds of the image are faster. But go mod download requires the multiplexer source to be there.

multiplexer/abci/abci.go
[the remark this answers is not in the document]
julienrbrt: Yes, they are in remote v1

julienrbrt: Maybe I need to add a diagram. Basically the multiplexer interacts with comet always using ABCI 2.0 (as we used comet 0.38). It would have done the reverse if we'd use 0.34.

Basically we have comet 0.38 (ABCI 2.0) -> multiplexer (remote v1 / v2) -> app v4 (v2)
                                                                                                                               -> app v3 (v1)
                                                                                                                                -> ...

multiplexer/abci/remote_v1.go
julienrbrt: definitely! we would have had to copy the protos otherwise

go.mod
```diff
@@ -2,12 +2,15 @@ module github.com/celestiaorg/celestia-app/v4
 
 go 1.23.6
 
+// TODO: tag multiplexer as v0.1.0 after merge.
+replace github.com/celestiaorg/celestia-app/multiplexer => ./multiplexer
```
chatton, line 6: @julienrbrt is the idea to pin to a specific version, or just have a local replace in the go.mod?

Pros/cons to both approaches, local pin reduces operational overhead of additional tags needing to be required in order to test changes.

Depending on pin makes it so you don't need to copy source code in the Dockerfile and that it's clear which version of the multiplexer you are using.

I kind of saw the multiplexer as just being a sub-component of the existing celestia-app, and that there wasn't really a need to have it tagged and released independently if there is just a single project using it.

cc @rootulp

julienrbrt: I thought it would be a separate go module yeah, but if it isn't i can change.

chatton: we can do local pin for now anyway and do follow up, with the tag 👍

multiplexer/abci/abci.go
evan-forbes: > the multiplexer interacts with comet always using ABCI 2.0

ahh ok I see thanks!

evan-forbes (approved):

rootulp (approved):
Still LGTM

chatton (commented):
We need to add another flag to the start versions (not changed in this PR)

I made these changes in the E2E test

```golang
	versions, err := abci.NewVersions(abci.Version{
		Appd:        v3,
		ABCIVersion: abci.ABCIClientVersion1,
		AppVersion:  3,
		StartArgs: []string{
			"--grpc.enable",
			"--grpc.address=0.0.0.0:9090", // ensure the grpc address is accessible from hosts such as txsim. (not just localhost)
			"--api.enable=true",
			"--api.swagger=false",
			"--with-tendermint=false",
			"--transport=grpc",
			"--address=0.0.0.0:26658",
		},
	})
```

go.mod
rootulp: Great question. Local pin _seems_ preferable based on the OP so inclined to stick with that if we don't need to release it separately.

Issue #4566, opened 2025-04-14
Update benchmarks in app for v4 (sdk v0.50) support


damiannolan:
The benchmarks currently do not build and run correctly for v4 (sdk v0.50).
They should be updated to support the changes in #4400.

Right now, the issue is silently hidden due to build flags, and benchmarks do not run in CI.

PR #4567, opened 2025-04-14
refactor: update benchmarks for app/v4

damiannolan:
## Overview

Gives love to benchmarks.

- Adds a single build flag `benchmarks` instead of two different ones
- Adjusts the make target to actually run benchmarks
- Adjusts tests to account for `deliverTx` -> `finalizeBlock`
- Fixes some test structure and compilation errors which exists pre-upgrade
- Refactors PFB half second test to use a binary search based approach.

Ideally the results.md file could be replaced by some prettified markdown output from the benchmark tests

closes: #4566

Note if you use vscode its worth adding to your `.vscode/settings.json`:
```json
{
    "gopls": {
        "buildFlags": [
            "-tags=benchmarks"
        ]
    },
    "go.testFlags": [
        "-tags=benchmarks"
    ]
}
```


app/benchmarks/benchmark_pfb_test.go
```diff
@@ -42,11 +42,7 @@ func BenchmarkCheckTx_PFB_Multi(b *testing.B) {
 		{blobSize: 400_000},
 		{blobSize: 500_000},
 		{blobSize: 1_000_000},
-		{blobSize: 2_000_000},
-		{blobSize: 3_000_000},
-		{blobSize: 4_000_000},
-		{blobSize: 5_000_000},
-		{blobSize: 6_000_000},
+		{blobSize: 2_000_000}, // maxTxSize is capped at 2MiB in checkTx
```
rach-id, line 45: these test cases were added if you want to manually bump the max tx size. I'm fine with removing them or keeping them

rach-id (approved):
utAck 👍

Issue #4574, opened 2025-04-14
Delete the blobstream related testsuite


rach-id:
Now that the blobstream module is no longer on main, we can also delete the test utils that are only used in it 

![Image](https://github.com/user-attachments/assets/e4ea24ea-31c5-4d66-b137-e617cfaff9d0)

_Originally posted by @rach-id in https://github.com/celestiaorg/celestia-app/pull/4543#pullrequestreview-2765703090_

PR #4575, opened 2025-04-14
chore!: remove blobstream related testing utils

rach-id:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/4574


rootulp (approved):
Awesome! Thanks for the clean up.

[uber nit] but looks like we're removing an exported const `DataCommitmentBlocksLimit` which one could argue is breaking. I'm indifferent on whether we mark this as breaking (i.e. `chore!:` vs `chore:`) because I'm guessing no one actually uses it

rach-id:
Updated :+1:

evan-forbes (approved):
:axe:

PR #4582, opened 2025-04-15
chore: check command status in BBR enable/disable targets

yarikbratashchuk:
Improve error handling in BBR-related Makefile targets

During onboarding, I identified that the BBR enable/disable commands in the Makefile were not checking the status of subcommands before proceeding to the next step. This resulted in misleading output where "BBR has been enabled" would display even when earlier commands in the sequence had failed (e.g., "modprobe: command not found").

This PR modifies the enable-bbr and disable-bbr targets to use proper command chaining with conditional execution (&&) and error handling (||), ensuring that success messages are only displayed when all required commands complete successfully. If any command in the chain fails, an appropriate error message is shown instead.

Tested it manually.


rach-id (commented):
is this only adding the `Failed to enable BBR. Please check error messages above.` message when enabling bbr doesn't work? or does something else I'm missing?

rootulp (commented):
Can you please populate the PR description with the motivation for this PR? It can be as small as "Closes #issue-number" if the relevant context exists in a GH issue. 

If there isn't a GH issue for this work then can you please include enough context for reviewers to understand:
1. What's the motivation?
2. How was the PR tested?

yarikbratashchuk:
> is this only adding the `Failed to enable BBR. Please check error messages above.` message when enabling bbr doesn't work? or does something else I'm missing?

yes, it's just for better dev experience

evan-forbes (approved):

rootulp (approved):

PR #4593, opened 2025-04-16
refactor: use one encoding.MakeConfig func

damiannolan:
## Overview

Avoid any potential footguns of having two encoding config constructors.


app/modules.go
```diff
@@ -55,32 +56,38 @@ import (
...
-	stakingModule{},
-	mintModule{},
+	circuitModule{},
+	consensus.AppModuleBasic{},
 	distribution.AppModuleBasic{},
+	evidence.AppModuleBasic{},
+	feegrantmodule.AppModuleBasic{},
+	genutil.AppModuleBasic{},
 	govModule{},
+	mintModule{},
```
damiannolan, line 71: should this be moved down to be grouped with celestia modules? I think this is custom, correct?

evan-forbes (dismissed):

app/modules.go
evan-forbes: yeah technically

rootulp: yea the x/mint module is technically Celestia specific b/c it hard-codes the inflation schedule

damiannolan: moved in https://github.com/celestiaorg/celestia-app/pull/4593/commits/53c09508d9cacfa6d55af9ef834aff58c1a2f054 and addressed linter complaints

rootulp (approved):
Thanks

evan-forbes (approved):

PR #4595, opened 2025-04-16
chore: return error if GetCommitment is invoked with an empty data availability header

rootulp:
Addresses [Slack thread](https://celestia-team.slack.com/archives/C02PFFCGMNW/p1744816819161269) from @MSevey


MSevey:
Having the `dah` validate on the exported method is good. 
`calculateCommitmentPaths` should probably have a zero square size check since that is ultimately where the panic happens. This would protect any other current or future methods that call `calculateCommitmentPaths`.

Issue #4549, opened 2025-04-14
Specify --v2-upgrade-height for embedded v3 binary.


chatton:
Follow up to main v4 PR.

We need to provide a way to specify the upgrade height [here](https://github.com/celestiaorg/celestia-app/blob/4be4619a63907bd66208d8030f16745e087ec568/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go#L39C31-L40C1)

It should be able to be specified dynamically for testing purposes, and default to the actual height at which the v2 upgrade happened.

We can do this with a build time argument using `-ldflags`.

julienrbrt:
It is unclear for me which path you want to go. I'm still leaning towards ldflags as well and let it being settable at build. 
What is your proposal?

rootulp:
What are the available paths? I don't know what the available options are.

When we implemented this for v3.x we did https://github.com/celestiaorg/celestia-app/pull/4091 so one option could be to re-apply that PR on `main`. I don't know understand why that PR was removed from `main` to begin with.

chatton:
if they are baked into the binary, we can use ldflags as an option to override for testing, but leave blank as the v3 binary has logic internally to handle this and we don't need to do anything if I understand correctly.

rootulp:
If that's the only option, that works too

julienrbrt:
I was looking at https://github.com/celestiaorg/celestia-app/pull/4091, and if this was included in a v3, i believe it will take precedence, so we only need the ldflag for the testing (where you'd be setting the flag manually)

rootulp:
You're right, it looks like https://github.com/celestiaorg/celestia-app/pull/4091 merged to main but never got included in v3.x. I can backport it to v3.x or we can use the ldflags that you mentioned.

julienrbrt:
> You're right, it looks like [#4091](https://github.com/celestiaorg/celestia-app/pull/4091) merged to main but never got included in v3.x. I can backport it to v3.x or we can use the ldflags that you mentioned.

I think having this as fallback in v3 is great. I've just opened https://github.com/celestiaorg/celestia-app/pull/4597, but with the above backported for v3, we would only need to set it while testing. When not set it would just use your PR

rootulp:
Okay, I'll backport https://github.com/celestiaorg/celestia-app/pull/4091 to v3.x

PR #4597, opened 2025-04-16
build: set v2 upgrade height at build time

julienrbrt:
## Overview

Closes: https://github.com/celestiaorg/celestia-app/issues/4549

It is pretty hard to tests, so I haven't tested it yet. I will fork the repo and try to tag something. Hopefully it triggers CI there and I am able to verify goreleaser created a binary with the correct version.

I'm more leaning towards going with a mix of both (ref https://github.com/celestiaorg/celestia-app/issues/4549#issuecomment-2810695347), as it wouldn't affect the goreleaser, as we can assume v3 have the correct default version.


.github/workflows/goreleaser.yml
```diff
@@ -18,10 +18,20 @@ jobs:
...
       - name: Create .release-env file
         run: |-
           echo 'GITHUB_TOKEN=${{secrets.GORELEASER_ACCESS_TOKEN}}' >> .release-env
+          # extract the chain name from the tag
+          TAG_NAME="${{ github.event.release.tag_name }}"
+          if [[ "$TAG_NAME" == *"-mocha"* ]]; then
+            echo 'CHAIN_NAME=mocha' >> .release-env
+          elif [[ "$TAG_NAME" == *"-arabica"* ]]; then
+            echo 'CHAIN_NAME=arabica' >> .release-env
+          else
```
rach-id, line 31: [note]
this will treat `-rc` releases also as mainnet releases. is this something we want?

julienrbrt: I'd rather not want that whole goreleaser logic tbh, but rather https://github.com/celestiaorg/celestia-app/issues/4549#issuecomment-2810695347

evan-forbes (dismissed):

rootulp (commented):
I think this PR can be simplified b/c we're backporting the default upgrade heights to v3.x in https://github.com/celestiaorg/celestia-app/pull/4607

so perhaps this PR only needs to support setting a custom v2 upgrade height for test chain ID(s)

julienrbrt:
> I think this PR can be simplified b/c we're backporting the default upgrade heights to v3.x in #4607
> 
> so perhaps this PR only needs to support setting a custom v2 upgrade height for test chain ID(s)

Yes! awesome, i'll modify

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -14,6 +14,13 @@ import (
 	embedding "github.com/celestiaorg/celestia-app/v4/internal/embedding"
 )
 
+// v2UpgradeHeight is the block height at which the v2 upgrade occurred.
+// this can be overridden at build time using ldflags:
+// -ldflags="-X 'github.com/celestiaorg/celestia-app/v4/cmd/celestia-appd/cmd.v2UpgradeHeight=1751707'" for arabica
+// -ldflags="-X 'github.com/celestiaorg/celestia-app/v4/cmd/celestia-appd/cmd.v2UpgradeHeight=2585031'" for mocha
+// -ldflags="-X 'github.com/celestiaorg/celestia-app/v4/cmd/celestia-appd/cmd.v2UpgradeHeight=2371495'" for mainnet
+var v2UpgradeHeight = "0"
```
rootulp, line 22: If the user doesn't override this value then `0` will be plumbed into `--v2-upgrade-height` which will take precedence over the defaults: https://github.com/celestiaorg/celestia-app/pull/4607 so we need to invoke v3 binary without `--v2-upgrade-height` if v2UpgradeHeight is `0`

julienrbrt: Gotcha. I see the pr sets it as default indeed and doesn't override it (SetDefault instead of Set). Was there a reason for that?

rootulp: Hmm I think I used `SetDefault` so that any value specified by user via `--v2-upgrade-height` would take precedence.

Not strongly opinionated on how we do it. We can modify v3.x to always use the hard-coded upgrade heights for the supported networks if that helps this PR.

julienrbrt: Nah, it's easier to skip here. That's fine. I'll update in a bit

julienrbrt: fixed

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -26,18 +33,22 @@ func modifyRootCommand(rootCommand *cobra.Command) {
 		panic(err)
 	}
 
+	var extraArgs []string
+	if v2UpgradeHeight > "0" {
```
chatton, line 37: doing a `>` here looks like a foot gun, should this just be `if v2UpgradeHeight != ""` and default it to `""`?

julienrbrt: it would have lexicographical worked but I do agree that it looks like one. i can change yeah

rootulp (approved):

evan-forbes (approved):

PR #4579, opened 2025-04-15
feat: enable block api proxy when running in standalone mode

damiannolan:
## Overview

Enables the usage of the core block api when running the app in standalone mode.

PR #4602, opened 2025-04-17
feat: more robust multiplexer image for e2e upgrade test

chatton:
## Overview

This PR updates the multiplexer image with an image built from https://github.com/celestiaorg/celestia-app/pull/4579

This will be required for the followup v3 -> v4 upgrade test to pass, as the image will be built with the correct architecture on merge to main.

The workflow is also updated to pass the correct build arguments when building the multiplexer image.

Passing of these build args is now enforced from within the Dockerfile at build time to ensure it is not possible to build a multiplexer with an embedded binary.


Makefile
```diff
@@ -2,6 +2,8 @@ VERSION := $(shell echo $(shell git describe --tags 2>/dev/null || git log -1 --
 COMMIT := $(shell git rev-parse --short HEAD)
 DOCKER := $(shell which docker)
 PROJECTNAME=$(shell basename "$(PWD)")
+GOOS ?= linux
+GOARCH ?= amd64
```
damiannolan, line 6: maybe these vars should be prefixed with `DOCKER`, otherwise its a little misleading why they are hardcoded to linux/amd64

evan-forbes (approved):

rootulp (approved):

PR #4603, opened 2025-04-17
ci: build v3 docker images with multiple platforms

chatton:
## Overview

In order to build and test images in the v3 line on non amd64 architectures, we can update the workflows to build multi arch images.

The current 0.5.0 workflow builds [just for amd64](https://github.com/celestiaorg/.github/blob/45533fc35847a8568deb0d4cae1851a2872f3651/.github/workflows/reusable_dockerfile_pipeline.yml#L311) while the newer workflow builds for both [`amd64` and `arm64`](https://github.com/celestiaorg/.github/blob/v0.6.1/.github/workflows/reusable_dockerfile_pipeline.yml#L305)

### DIFF
--- a/.github/workflows/docker-build-publish.yml
+++ b/.github/workflows/docker-build-publish.yml
@@ -21,7 +21,7 @@ jobs:
     permissions:
       contents: write
       packages: write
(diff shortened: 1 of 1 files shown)


rach-id (approved):

evan-forbes (approved):

rootulp (approved):
Thanks

PR #4608, opened 2025-04-17
fix: wire hyperlane module accounts in app

damiannolan:
## Overview

Fixes a bug where sending an outbound hyperlane transfer would fail due to no module accounts created in app.
ref: https://github.com/bcp-innovations/hyperlane-cosmos/blob/main/x/warp/keeper/logic_collateral.go#L23


app/app.go
```diff
@@ -137,6 +137,8 @@ var maccPerms = map[string][]string{
 	stakingtypes.NotBondedPoolName: {authtypes.Burner, authtypes.Staking},
 	ibctransfertypes.ModuleName:    {authtypes.Minter, authtypes.Burner},
 	icatypes.ModuleName:            nil,
+	hyperlanetypes.ModuleName:      nil,
+	warptypes.ModuleName:           {authtypes.Minter, authtypes.Burner},
```
damiannolan, line 141: This was missed in module wiring as its only documented for depinject usage

rootulp (approved):

cmwaters (approved):
Good catch

rach-id (approved):

rootulp (approved):

PR #4611, opened 2025-04-17
refactor!: remove `GetCommitment`

### LINKED ISSUE #4606: De-duplicate `GetCommitment`
Thanks 🙏 

I'm wondering where this is used? is it used by any downstream repos? If not, we can delete it with the upcoming breaking release. 

Currently, we have an implementation of this in go-square, so if this is a duplicate, let's get rid of it.

_Originally posted by @rach-id in https://github.com/celestiaorg/celestia-app/pull/4595#pullrequestreview-2775125420_

Himess:
## Overview

Removes the unused `GetCommitment` function and its test.

This function is no longer used in production and has effectively been superseded by the more robust `CreateCommitment` implementation in [`go-square`](https://github.com/celestiaorg/go-square/blob/main/inclusion/commitment.go).

Although `CreateCommitment` resides in a separate repository, both `go-square` and `celestia-app` are part of the Celestia codebase, and `go-square` is already a dependency of `celestia-app`. Duplicating the same logic across both repositories increases maintenance overhead and the risk of inconsistency.

This cleanup aligns with the plan described in #4606 as part of preparations for an upcoming breaking release.

**Usage check:**

The following search confirms that `GetCommitment` is only used within the files being removed:

image

Closes #4606


rootulp (approved):
LGTM if CI passes. I just checked and https://github.com/celestiaorg/go-square/blob/a1064ea59fc2e33b775315d08c561af50704edc6/inclusion/commitment.go#L17 doesn't accept a DataAvailabityHeader argument so we don't need to apply https://github.com/celestiaorg/celestia-app/pull/4595 to go-square

rach-id (approved):
Node also seem not to be using it. So we should be fine to remove

PR #4627, opened 2025-04-22
refactor(benchmarks): align metric name to `block_size(mb)`

reject-i:
BenchmarkProcessProposal_MsgSend_8MB emitted `block_size_(mb)` (extra underscore), creating a duplicate metric and splitting dashboard/alert data. Renamed to `block_size(mb)` for consistent aggregation across all benchmarks.


rach-id (approved):

rach-id:
This is not related to grafana, this is for the benchmarks we have in here: https://github.com/celestiaorg/celestia-app/tree/main/app/benchmarks

This PR just removes the extra `_` as the other benchmarks don't use it

rootulp (approved):

Issue #4578, opened 2025-04-15
Feature Request: Implement `Option` for `TxClient` to take additional core endpoints


renaynay:
Right now, `TxClient` explicitly takes a `conn` in the constructor and uses it as the sole query + submission endpoint.

`TxClient` should still require at least one `conn` explicitly as part of the constructor as `TxClient` cannot operate without at least 1 core endpoint, but we should also introduce an `Option` to allow setting additional core endpoints on the `TxClient` for it to use in case of failure / latency of the default connection.

Instead of having a single `grpc` field on the `TxClient`, it should have a `conns` field that is `[]*grpc.ClientConn`.

Implement an `Option`

```
func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
// set multiple core endpoints on the `TxClient`
  return func(c *TxClient) {
    c.conns = append(c.conns, conns...)
  }
}

```

## Submission 

For the first iteration of the feature, the `TxClient` should use the first (up to three) connections for every *submission* operation. `TxClient` should, in parallel, submit transactions to (up to three) connections and wait for whichever connection responds successfully first, exiting the other active routines.

If all routines fail within the given context deadline, return failure to user. Do not retry.

Since user can specify arbitrary amount of connections via `WithAdditionalCoreEndpoints`, we should pick active set of maximum 3 connections to use for submission. Use the default connection + the first 2 connections added via the `WithAdditionalCoreEndpoints` .

The reason for the max 3 connection limit for the first iteration is to avoid implementing logic as it relates to rotating the active connection set.

In the future, we will allow rotation of the active connection set, so I would not constrain the Option to only allow a set of 2 additional connections, but we should clarify the behaviour via docs that for now, only the first 2 additional core endpoints will be used.

## In the future
_(Note that this section should only be considered as reference for future work on this feature.)_

Eventually, we will introduce some more sophisticated ranking system for the connections (kind of like peer scoring, but for connection latency + availability) and logic for switching the active set of connections `TxClient` is using. However, for the first iteration, the above should work sufficiently fine.

We should not perform any additional sophisticated retries here until we have further information about how we want to design retry behaviour which requires more user research.

In the future, we can define a set of criteria (informed by metrics) that will dictate active connection set rotation, for example:

- a connection or service-specific failure, or
- latency of a request above a specified limit ( set by the ctx deadline of the request **by the caller** )

[...]

PR #4628, opened 2025-04-22
feat(user): Add support for multiple core endpoints to TxClient

yarikbratashchuk:
This PR introduces support for multiple gRPC core endpoints within the `TxClient` to improve transaction submission reliability.

It implements the first phase described in issue #4578, allowing the client to broadcast transactions concurrently to up to three endpoints (the primary connection + the first two provided via the new `WithAdditionalCoreEndpoints` option) and returning the result from the first successful response.

**Key changes:**
- Replaced `grpc *grpc.ClientConn` field with `conns []*grpc.ClientConn` in the `TxClient` struct to store multiple connections. The `grpc` field is kept temporarily for compatibility with methods not yet updated.
- Introduced a new option `WithAdditionalCoreEndpoints([]*grpc.ClientConn)` to allow users to provide additional endpoints beyond the primary one required by the constructor.
- Modified `NewTxClient` to initialize `conns` with the primary connection and append others provided via the new option.
- Added an internal `broadcastMulti` method:
    - Selects up to 3 connections (primary + first 2 additional) for submission.
    - Uses an `errgroup` and cancellable context to broadcast the transaction concurrently to the selected endpoints.
    - Returns the response from the first endpoint that successfully broadcasts the transaction (`BroadcastMode_BROADCAST_MODE_SYNC` success).
    - Immediately cancels the context for other in-flight broadcast attempts upon first success by returning `context.Cancelled` error.
    - Collects and aggregates errors if all attempts fail or if the context deadline is exceeded.
    - Handles potential race conditions between goroutine completion, context cancellation, and channel closing.
- Updated non-broadcast methods (`ConfirmTx`, `checkAccountLoaded`, gas estimation helpers) to use the primary connection (`conns[0]`) for now (mentioned this as a point in the follow up [#4651](https://github.com/celestiaorg/celestia-app/issues/4651).
- Added a new unit test suite `TestMultiConnBroadcast ` specifically for the multi-connection broadcast logic, utilizing mock gRPC servers.

Closes https://github.com/celestiaorg/celestia-app/issues/4578


pkg/user/tx_client.go
```diff
@@ -142,6 +153,8 @@ type TxClient struct {
 	signer   *Signer
 	registry codectypes.InterfaceRegistry
 	grpc     *grpc.ClientConn
```
rootulp, line 155: [nit] should this field be deprecated?

```suggestion
	// grpc is the primary GRPC connection.
	// Deprecated: use conns[0] instead.
	grpc     *grpc.ClientConn
```

pkg/user/tx_client.go
```diff
@@ -379,40 +393,175 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
-		return nil, err
+
+	broadcastCtx, cancel := context.WithCancel(ctx)
+	defer cancel() // Ensure cancel is called eventually if not explicitly done earlier
+
+	g, childCtx := errgroup.WithContext(broadcastCtx)
+	var successfulBroadcast sync.Once // Ensure only one success is processed
+
+	for _, conn := range broadcastConns {
+		conn := conn // Capture range variable
```
rootulp, line 450: golangci-lint is complaining here

```suggestion
```

pkg/user/tx_client.go
```diff
@@ -379,40 +393,175 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
+		g.Go(func() error {
+			broadcast(conn, childCtx)
+			// Returning nil because errors are handled via errCh.
+			// The errgroup context handles cancellation signals.
+			return nil
+		})
+	}
```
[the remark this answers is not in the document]
rootulp: Does this suggestion have merit?

pkg/user/tx_client.go
```diff
@@ -379,40 +393,175 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
+				signer:    signer,
+				timestamp: time.Now(),
+			}
+
+			incErr := client.signer.IncrementSequence(signer)
+			if incErr != nil {
+				fmt.Printf("CRITICAL: Failed to increment sequence for signer %s after successful broadcast: %v\n", signer, incErr)
+				// TODO: Wrap the increment error and return it?
+				// finalErr = fmt.Errorf("increment sequencing failed after successful broadcast: %w", incErr)
+				// firstResp = nil
```
rootulp, line 530: [blocking] commented code is a red flag. If the commented code isn't applicable, can it be removed and included in a follow-up issue created to address this TODO?

Ref: https://kentcdodds.com/blog/please-dont-commit-commented-out-code

pkg/user/tx_client.go
```diff
@@ -379,40 +393,175 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
+			}
+		})
+		// Return the successful response
+		return firstResp, nil // We prioritize returning the successful response
+	}
+
+	// If no success, collect actual errors from the channel (excluding cancellation)
+	close(errCh)
+	var broadcastErrs []string // Renamed to avoid confusion with finalErr
```
rootulp, line 539: golangci-lint is complaining here

pkg/user/tx_client_test.go
```diff
@@ -28,6 +31,7 @@ import (
 	"github.com/celestiaorg/celestia-app/v4/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v4/pkg/user"
 	"github.com/celestiaorg/celestia-app/v4/test/util/blobfactory"
+	grpctest "github.com/celestiaorg/celestia-app/v4/test/util/grpctest"
```
rootulp, line 34: golangci-lint is complaining

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,343 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+	// Assert
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+	require.Equal(t, uint32(abci.CodeTypeOK), resp.Code)
+	require.Equal(t, "HASH1", resp.TxHash)
+	require.Equal(t, int32(1), mockSvc1.Invocations.Load(), "Mock1 (winner) should be invoked once")
+	// We cannot reliably assert the invocation count for cancelled mocks, as cancellation might
+	// happen before the mock method is even called in its goroutine.
+	// require.Equal(t, int32(1), mockSvc2.Invocations.Load(), "Mock2 should be invoked once")
+	// require.Equal(t, int32(1), mockSvc3.Invocations.Load(), "Mock3 should be invoked once")
```
rootulp, line 577:
```suggestion
```

rootulp (commented):
- Thanks for the clear PR description
- Lint is failing in many places. Can you please address golangci-lint issues then re-request reviews?
- Is there any merit in the coderabbit comments to refactor the concurrent code?

rootulp:
> The grpc field is kept temporarily for compatibility with methods not yet updated.

Is there a follow-up issue to track removing the `grpc` field and updating the methods to use `conns`?

evan-forbes (commented):
I'll defer to the author and just to be clear am not blocking on the final approach, but naively it seems like we don't have to change the broadcast function and could instead keep the entire concurrency logic in a separate method.

the main benefit coming from readability and maintenance. Instead of always having to take this approach, we still have the broadcast method. Instead of mixing concurrency logic with broadcast logic, we keep them separate.

pkg/user/tx_client.go
```diff
@@ -379,40 +393,175 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
-	txClient := sdktx.NewServiceClient(client.grpc)
-	resp, err := txClient.BroadcastTx(
-		ctx,
-		&sdktx.BroadcastTxRequest{
-			Mode:    sdktx.BroadcastMode_BROADCAST_MODE_SYNC,
-			TxBytes: txBytes,
-		},
+	numConns := len(client.conns)
+	if numConns == 0 {
+		return nil, errors.New("no connections available in TxClient")
```
renaynay, line 398: this check should not exist here - it should exist in constructor. `TxClient` should fail to build if there is not at least 1 grpc connection provided to it.

pkg/user/tx_client.go
```diff
@@ -379,40 +393,175 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
+	}
+
+	// Select up to 3 connections for broadcasting: primary + first 2 additional
+	broadcastConns := client.conns
+	if numConns > 3 {
+		broadcastConns = client.conns[:3]
+	}
```
renaynay, line 405: same with this logic, wrong place to have it. Move to constructor please

pkg/user/tx_client.go
```diff
@@ -132,6 +133,16 @@ func WithEstimatorService(conn *grpc.ClientConn) Option {
 	}
 }
 
+// WithAdditionalCoreEndpoints adds additional core endpoints to the TxClient.
+// For transaction submission, the client will attempt to use the primary endpoint
+// and the first two additional endpoints provided via this option.
+func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
+	// set multiple core endpoints on the `TxClient`
+	return func(c *TxClient) {
+		c.conns = append(c.conns, conns...)
```
renaynay, line 142: IMO here, you should already trim conn set so that len(conns) <= 3 

log.Debug a warning that says "provided %d additional core endpoints, can only use 2"

pkg/user/tx_client.go
renaynay: yeah no dangling code pls

renaynay (changes requested):
My initial review -- before I can continue looking through the code, can you please spend some time combing through the broadcast logic here and doing your best to simplify the code? There's a bunch of LOC there that are unnecessary and look generated by AI (to me). Lint is also failing all over the place. Can you please try to clean it up (including the test suite which IMO is more verbose than necessary and borderline impossible to read through the test cases). Something like a table test would work best here IMO. Also several comments in the code that are unhelpful that should be removed.

ninabarbakadze (commented):
is this PR is still in progress can you please convert it to draft?

yarikbratashchuk:
> is this PR is still in progress can you please convert it to draft?

I wanted to, but for some reason can't

evan-forbes (dismissed):

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+	s.Require().NoError(err)
+	s.accAddr = sdk.AccAddress(pubKey.Address())
+
+	s.signer, err = user.NewSigner(s.kr, s.encCfg.TxConfig, BroadcastTestChainID, user.NewAccount(s.account, 0, 0))
+	s.Require().NoError(err)
+}
+
+func TestBroadcastTestSuite(t *testing.T) {
+	suite.Run(t, new(BroadcastTestSuite))
+}
```
evan-forbes, line 485: upon merging, imo we should write a flup issue that tries to address this test as this feels like a kind of insane amount of tech debt to test really simple functionality

pkg/user/tx_client.go
```diff
@@ -132,6 +134,20 @@ func WithEstimatorService(conn *grpc.ClientConn) Option {
...
+// For transaction submission, the client will attempt to use the primary endpoint
+// and the first two additional endpoints provided via this option.
+func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
+	// set 2 additional core endpoints on the `TxClient`
+	if len(conns) > 2 {
+		conns = conns[:2]
+	}
+
+	return func(c *TxClient) {
+		c.conns = append(c.conns, conns...)
```
rootulp, line 147: [not blocking] It seems like a developer could call

```go
.WithAdditionalCoreEndpoints([conn1, conn2])
.WithAdditionalCoreEndpoints([conn3, conn4])
```

and get around the 2 endpoint restriction 🤔

pkg/user/tx_client.go
```diff
@@ -428,7 +531,9 @@ func (client *TxClient) pruneTxTracker() {
 // hash. It will continually loop until the context is cancelled, the tx is found or an error
 // is encountered.
 func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxResponse, error) {
-	txClient := tx.NewTxClient(client.grpc)
+	// TODO (#4578): Implement logic to potentially use other connections if the primary fails.
```
rootulp, line 534: This PR "closes #4578" so it would be nice to open a new issue to track this TODO and any other follow-ups that were created by this PR. Also helpful to track any remaining action items from #4578

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+type broadcastTestCase struct {
+	name string
+	// Function to set up mock servers and return their handlers + connections
+	setupMocks        func(t *testing.T) ([]*grpctest.MockTxService, []*grpc.ClientConn, []func())
+	setupContext      func() (context.Context, context.CancelFunc)
+	txOpts            []user.TxOption
+	expectedSuccess   bool
+	expectedHash      string // Only checked if expectedSuccess is true
+	expectedErrSubstr string // Only checked if expectedSuccess is false
+	expectedErrType   error  // Optional: Specific error type check (e.g., context.DeadlineExceeded)
```
rootulp, line 516: [question] Why can't this use a single `expectedError error` field and instead needs an expectedErrSubstr and expectedErrType?

pkg/user/tx_client.go
```diff
@@ -390,6 +412,11 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
 	if err != nil {
 		return nil, err
 	}
+	if resp.TxResponse == nil {
+		// This case should ideally not happen with BroadcastMode_SYNC if err is nil,
+		// but handle defensively.
+		return nil, fmt.Errorf("broadcast via %s returned nil TxResponse despite nil error", conn.Target())
+	}
```
rootulp, line 419: [question] what motivated this addition? Is it actually possible to have a nil TxResponse and a nil err?

pkg/user/tx_client.go
```diff
@@ -399,20 +427,98 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+	if firstResp, ok := <-respCh; ok {
+		// save the sequence and signer of the transaction in the local txTracker
+		// before the sequence is incremented
+		client.txTracker[firstResp.TxHash] = txInfo{
+			sequence:  client.signer.accounts[signer].Sequence(),
+			signer:    signer,
+			timestamp: time.Now(),
+		}
+
+		// after the transaction has been submitted, we can increment the
```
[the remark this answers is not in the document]
rootulp: Why was this comment resolved? It seems like a valid suggestion.

pkg/user/tx_client.go
```diff
@@ -399,20 +426,96 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+		// sequence of the signer
+		incErr := client.signer.IncrementSequence(signer)
+		if incErr != nil {
+			return nil, fmt.Errorf("increment sequencing: %w", incErr)
+		}
```
rootulp, line 488: [nit] incErr seems like an odd name. Can this be renamed to `err`?
[nit] sequence not sequencing

Proposed refactor

```suggestion
		if err := client.signer.IncrementSequence(signer); err != nil {
			return nil, fmt.Errorf("increment sequence: %w", err)
		}
```

pkg/user/tx_client.go
```diff
@@ -399,20 +426,96 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+
+	// If the context finished (DeadlineExceeded or Canceled)
+	if ctxErr := ctx.Err(); ctxErr != nil { // Check original context
+		errPrefix := "broadcast context cancelled"
+		if errors.Is(ctxErr, context.DeadlineExceeded) {
+			errPrefix = "broadcast context deadline exceeded"
+		}
+		// Wrap the original context error directly
+		return nil, fmt.Errorf("%s: %w", errPrefix, ctxErr)
+	}
```
rootulp, line 511: [question] why is ctxErr guaranteed to be DeadlineExceeded or Cancelled?

Proposed refactor which seems clearer to me:

```suggestion
	if ctxErr := ctx.Err(); ctxErr != nil {
		if errors.Is(ctxErr, context.Canceled) {
			return nil, fmt.Errorf("%s: %w", "broadcast context cancelled", ctxErr)
		}
		if errors.Is(ctxErr, context.DeadlineExceeded) {
			return nil, fmt.Errorf("%s: %w", "broadcast context deadline exceeded", ctxErr)
		}
	}
```

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+				require.Nil(t, resp)
+				if tc.expectedErrSubstr != "" {
+					require.Contains(t, err.Error(), tc.expectedErrSubstr)
+				}
+				if tc.expectedErrType != nil {
+					// Use require.ErrorIs for checking specific error values like context errors
+					if errors.Is(tc.expectedErrType, context.DeadlineExceeded) || errors.Is(tc.expectedErrType, context.Canceled) {
+						require.True(t, errors.Is(err, tc.expectedErrType), "Expected error chain to contain %v, but got %v", tc.expectedErrType, err)
+					} else {
+						// Use require.ErrorAs for checking error types (structs/interfaces)
```
rootulp, line 795: This comment seems very AI generated and not useful for readers of the test

```suggestion
```

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+				require.Equal(t, abci.CodeTypeOK, resp.Code)
+				require.Equal(t, tc.expectedHash, resp.TxHash)
+			} else {
+				require.Error(t, err)
+				require.Nil(t, resp)
+				if tc.expectedErrSubstr != "" {
+					require.Contains(t, err.Error(), tc.expectedErrSubstr)
+				}
+				if tc.expectedErrType != nil {
+					// Use require.ErrorIs for checking specific error values like context errors
```
rootulp, line 791: This comment isn't helpful for readers of this test

```suggestion
```

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+				}
+				conn1, stop1 := grpctest.StartMockServer(t, mockSvc1)
+				return []*grpctest.MockTxService{mockSvc1}, []*grpc.ClientConn{conn1}, []func(){stop1}
+			},
+			setupContext: func() (context.Context, context.CancelFunc) {
+				return context.Background(), func() {}
+			},
+			txOpts:            defaultOpts,
+			expectedSuccess:   false,
+			expectedErrSubstr: "insufficient funds", // Check for the RawLog in the error
```
rootulp, line 760:
```suggestion
			expectedErrSubstr: "insufficient funds",
```

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+			checkMockInvocations: func(t *testing.T, mocks []*grpctest.MockTxService) {
+				require.Len(t, mocks, 4)
+				// broadcastTx internally selects only up to 3 connections.
+				require.Equal(t, int32(1), mocks[0].Invocations.Load())
+				require.Equal(t, int32(1), mocks[1].Invocations.Load())
+				require.Equal(t, int32(1), mocks[2].Invocations.Load())
+				require.Equal(t, int32(0), mocks[3].Invocations.Load(), "Mock4 should not be invoked")
+			},
+		},
+		// Add more cases as needed (e.g., non-zero TxResponse code failure)
```
rootulp, line 743:
```suggestion
```

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+			},
+			setupContext: func() (context.Context, context.CancelFunc) {
+				return context.Background(), func() {}
+			},
+			txOpts:          defaultOpts,
+			expectedSuccess: true,
+			expectedHash:    "HASH_MT3",
+			checkMockInvocations: func(t *testing.T, mocks []*grpctest.MockTxService) {
+				require.Len(t, mocks, 4)
+				// broadcastTx internally selects only up to 3 connections.
```
rootulp, line 736:
```suggestion
```

pkg/user/tx_client_test.go
```diff
@@ -447,3 +451,358 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+					[]*grpc.ClientConn{conn1, conn2, conn3},
+					[]func(){stop1, stop2, stop3}
+			},
+			setupContext: func() (context.Context, context.CancelFunc) {
+				// Short deadline
+				return context.WithTimeout(context.Background(), 50*time.Millisecond)
+			},
+			txOpts:          defaultOpts,
+			expectedSuccess: false,
+			expectedErrType: context.DeadlineExceeded, // Check for specific error type
```
rootulp, line 665:
```suggestion
			expectedErrType: context.DeadlineExceeded,
```

pkg/user/tx_client.go
```diff
@@ -399,20 +426,96 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+			if childCtx.Err() != nil {
+				if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+					log.Debug().Err(err).Str("target", conn.Target()).Msg("Broadcast gRPC call failed after context finished")
+				}
+				return nil
+			}
+
+			if err != nil {
+				if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+					errCh <- err
```
rootulp, line 454: [question] why can't context cancelled or context deadline errors go to the error channel?

rootulp (commented):
I'm finding it really difficult to follow the concurrent code and tests. 

Is it possible to leverage any existing code or libraries to submit 3 parallel requests and return the first one that succeeds?

pkg/user/tx_client.go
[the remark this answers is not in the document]
yarikbratashchuk: no, lock is called on the calling function

pkg/user/tx_client.go
yarikbratashchuk: I'll update #4651

pkg/user/tx_client.go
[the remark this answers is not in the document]
renaynay: Better to protect a shared resource as `TxClient`'s methods can be called concurrently.

pkg/user/tx_client.go
renaynay: I actually don't see why this shouldn't also be included in the scope of the `broadcastMulti` call. I would have just parallelised the top-level call (in `SubmitPayForBlob` and waited for the first one to return `core.TxStatusCommitted`. 

What could be a potential problem with this idea? cc @rootulp

pkg/user/tx_client.go
rootulp: Thanks for explaining. Seems safe to keep.

pkg/user/tx_client.go
```diff
@@ -23,6 +23,8 @@ import (
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types/proposal"
+	"github.com/rs/zerolog/log"
```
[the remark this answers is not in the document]
rootulp: On main we only use `github.com/rs/zerolog/log` in txsim so you're right that dependency is a bit unusual. 

@yarikbratashchuk Is there a particular reason you used zerolog in this PR? If not, it seems possible to refactor to use `cosmossdk.io/log` or even `fmt.Printf` statements b/c I only see two usages of log in this file.

[not blocking] from my perspective

pkg/user/tx_client.go
rootulp: No objections especially if it makes it easier to understand the concurrent code

pkg/user/tx_client.go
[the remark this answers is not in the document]
yarikbratashchuk: It's protected in the `BroadcastTx`

pkg/user/tx_client.go
```diff
@@ -207,6 +210,11 @@ func NewTxClient(
 		opt(txClient)
 	}
 
+	numConns := len(txClient.conns)
+	if numConns == 0 {
+		return nil, errors.New("no connections provided to TxClient")
+	}
+
```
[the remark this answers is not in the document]
yarikbratashchuk: yeah :)

rootulp:
I don't think this is ready for re-review b/c CI is failing.

pkg/user/tx_client.go
```diff
@@ -132,6 +133,20 @@ func WithEstimatorService(conn *grpc.ClientConn) Option {
...
 }
 
+// WithAdditionalCoreEndpoints adds additional core endpoints to the TxClient.
+// For transaction submission, the client will attempt to use the primary endpoint
+// and the first two additional endpoints provided via this option.
+func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
+	// set 2 additional core endpoints on the `TxClient`
+	if len(conns) > 2 {
+		conns = conns[:2]
+	}
```
renaynay, line 143: since we have sanity check in ctor, I think it's fine to omit this logic.

test/util/grpctest/mocks.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		if conn != nil {
+			connCloseErr := conn.Close()
+			if connCloseErr != nil {
+				t.Logf("Error closing connection: %v", connCloseErr)
+			}
+		}
+	}
+
+	return conn, stop
```
renaynay, line 107: why can't we t.Cleanup this?

test/util/grpctest/mocks.go
```diff
@@ -0,0 +1,108 @@
...
+// MockTxService allows controlling the behavior of BroadcastTx calls.
+type MockTxService struct {
+	sdktx.UnimplementedServiceServer // Embed the unimplemented server
+
+	BroadcastHandler func(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error)
+	Invocations      atomic.Int32 // Use atomic for potential concurrent access if needed
+}
+
+func (m *MockTxService) BroadcastTx(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error) {
+	m.Invocations.Add(1)
```
renaynay, line 27: are you using this for something?

test/util/grpctest/mocks.go
```diff
@@ -0,0 +1,108 @@
...
+func (m *MockTxService) TxDecodeAmino(context.Context, *sdktx.TxDecodeAminoRequest) (*sdktx.TxDecodeAminoResponse, error) {
+	return nil, errors.New("TxDecodeAmino not implemented in mock")
+}
+
+// StartMockServer starts a mock gRPC server with the given MockTxService using a TCP listener.
+func StartMockServer(t *testing.T, service *MockTxService) (*grpc.ClientConn, func()) {
+	t.Helper()
+
+	// Use a real TCP listener on a random port
+	lis, err := net.Listen("tcp", "localhost:0")
```
renaynay, line 72: there is a utility function `GetFreePort` in testnode pkg, please use

test/util/grpctest/mocks.go
```diff
@@ -0,0 +1,108 @@
...
+	require.NoError(t, err)
+
+	s := grpc.NewServer()
+	sdktx.RegisterServiceServer(s, service)
+
+	go func() {
+		if err := s.Serve(lis); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
+			// Log error instead of Printf for better test output
+			t.Logf("Mock server error: %v", err)
```
renaynay, line 81: this needs to fatal out here, we can't continue.

test/util/grpctest/mocks.go
```diff
@@ -0,0 +1,108 @@
...
+		closeErr := lis.Close()
+		if closeErr != nil && !errors.Is(closeErr, net.ErrClosed) && !strings.Contains(closeErr.Error(), "use of closed network connection") {
+			t.Logf("Error closing listener: %v", closeErr)
+		}
+		if conn != nil {
+			connCloseErr := conn.Close()
+			if connCloseErr != nil {
+				t.Logf("Error closing connection: %v", connCloseErr)
+			}
+		}
```
renaynay, line 104: we don't need such granular err handling logic here for tests - lets' just return error to t when doing t.Cleanup if err happens. No need to analyse it.

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+					BroadcastHandler: func(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error) {
+						return &sdktx.BroadcastTxResponse{TxResponse: &sdk.TxResponse{Code: abci.CodeTypeOK, TxHash: "HASH1"}}, nil
+					},
+				}
+				conn1, stop1 := grpctest.StartMockServer(t, mockSvc1)
+				return []*grpctest.MockTxService{mockSvc1}, []*grpc.ClientConn{conn1}, []func(){stop1}
+			},
+			expectedErr: nil,
+		},
+		{ // Secondary Success
```
renaynay, line 489: using sleeps like this (ms) will definitely flake in CI - another way you can test is by allowing the second server to return immediately (or with small delay), and the rest with longer delays.

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+				}}
+				mockSvc3 := &grpctest.MockTxService{BroadcastHandler: func(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error) {
+					return nil, errMock3
+				}}
+				conn1, stop1 := grpctest.StartMockServer(t, mockSvc1)
+				conn2, stop2 := grpctest.StartMockServer(t, mockSvc2)
+				conn3, stop3 := grpctest.StartMockServer(t, mockSvc3)
+				return []*grpctest.MockTxService{mockSvc1, mockSvc2, mockSvc3}, []*grpc.ClientConn{conn1, conn2, conn3}, []func(){stop1, stop2, stop3}
+			},
+			expectedErr: errMock1, // Expect the first error returned (or any, depends on race)
```
renaynay, line 533: Just return one common error here so we don't flake on race conditions.

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+		},
+		{ // Context Deadline
+			setupMocks: func(t *testing.T) ([]*grpctest.MockTxService, []*grpc.ClientConn, []func()) {
+				mockHandler := func(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error) {
+					select {
+					case <-time.After(200 * time.Millisecond):
```
renaynay, line 539: set a short deadline in the caller's ctx and wait til after the deadline, then return ctx err.

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+				mockSvc2 := &grpctest.MockTxService{BroadcastHandler: func(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error) {
+					return &sdktx.BroadcastTxResponse{TxResponse: &sdk.TxResponse{Code: abci.CodeTypeOK, TxHash: "HASH_LT3"}}, nil
+				}}
+				conn1, stop1 := grpctest.StartMockServer(t, mockSvc1)
+				conn2, stop2 := grpctest.StartMockServer(t, mockSvc2)
+				return []*grpctest.MockTxService{mockSvc1, mockSvc2}, []*grpc.ClientConn{conn1, conn2}, []func(){stop1, stop2}
+			},
+			expectedErr: nil,
+		},
+		{ // More Than Three Conns (Success - only first 3 used)
```
renaynay, line 569: we don't need this full blown test case, we trim conns in ctor.

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+			},
+			expectedErr: &user.BroadcastTxError{}, // Expect specific error type
+		},
+	}
+
+	for i, tc := range testCases {
+		name := fmt.Sprintf("BroadcastTestCase%d", i) // Simple naming
+		t.Run(name, func(t *testing.T) {
+			_, conns, stops := tc.setupMocks(t)
+			for _, stop := range stops {
```
renaynay, line 609: use t.Cleanup in setupMocks

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+			}
+			defer cancel()
+
+			resp, err := tempTxClient.BroadcastTx(ctx, []sdk.Msg{defaultMsg}, defaultOpts...)
+
+			if tc.expectedErr == nil {
+				require.NoError(t, err, "Expected success, but got error: %v", err)
+				require.NotNil(t, resp, "Expected non-nil response on success")
+				require.Equal(t, abci.CodeTypeOK, resp.Code, "Expected CodeTypeOK on success")
+			} else {
```
renaynay, line 644: please don't nest this logic, just move on to next case if tx.expectedErr == nil

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+							isCorrectError = true
+						}
+					}
+				}
+
+				require.True(t, isCorrectError, "Error mismatch. Expected error matching type/value '%T(%v)', but received '%T(%v)'", tc.expectedErr, tc.expectedErr, err, err)
+			}
+		})
+	}
+}
```
renaynay, line 684: Please clean up this error checking logic - it is too complicated and impossible to read. Please first un-nest and then reduce LOC

pkg/user/tx_client_test.go
```diff
@@ -447,3 +452,241 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
+
+func isGrpcDeadlineError(err error) bool {
```
renaynay, line 686: is this actually necessary? Does the error get wrapped? Regardless, I believe errors.Is checks target error for equivalency (unwrapping). Not sure you actually need this.

pkg/user/tx_client.go
```diff
@@ -415,6 +441,54 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+
+	groupErr := g.Wait()
+
+	close(respCh)
+	close(errCh)
+
+	if firstResp, ok := <-respCh; ok {
+		for err := range errCh {
+			fmt.Printf("Ignored broadcast error after success: %v", err)
```
renaynay, line 475: No prints please in code

renaynay, line 475: I would avoid doing anything with other errors right now - we will implement metrics later that will make this more useful than just dumping debug log.

pkg/user/tx_client.go
```diff
@@ -415,6 +441,54 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+
+			select {
+			case respCh <- resp:
+			case <-childCtx.Done():
+			}
+			return nil
+		})
+	}
+
+	groupErr := g.Wait()
```
renaynay, line 468: here we wait for *all* requests to return, we do not cancel other in-flight requests once a success response is read. This logic is missing.

renaynay (commented):
Looking a bit better, still more changes needed and one piece of logic missing.

pkg/user/tx_client.go
yarikbratashchuk: no, as the first one returns the group is cancelled and they return quickly

pkg/user/tx_client_test.go
yarikbratashchuk: ah yeah

pkg/user/tx_client_test.go
yarikbratashchuk: errors.Is didn't work for me 
will double check the wrapping options

test/util/grpctest/mocks.go
yarikbratashchuk: yeah, it's used in broadcast test as a mock endpoint

pkg/user/tx_client.go
renaynay: @yarikbratashchuk Where do explicitly cancel out the other in-flight routines here? `g.Wait()` blocks until *all* routines launched in that loop you have above have exited. You do not cancel the childCtx here, so I'm confused where the quick return happens?  

Also I don't understand what you mean by respCh being buffered - the buffer is there only to be able for sender to write to ch without being blocked on reader not being ready to read yet, so I'm confused what you mean by "we don't wait before it's read" ?

test/util/grpctest/mocks.go
renaynay: I meant specifically tracking the Invocations -- where do you use this?

test/util/grpctest/mocks.go
renaynay: why panic and why not t.Fatal?

test/util/grpctest/mocks.go
```diff
@@ -93,7 +97,7 @@ func StartMockServer(t *testing.T, service *MockTxService) (*grpc.ClientConn, fu
 	stop := func() {
```
renaynay, line 97: by t.Cleanup, I meant that here instead of returning the function, you can do t.Cleanup func right here in line.

pkg/user/tx_client.go
yarikbratashchuk: @renaynay on success we return err which cancels childCtx withing the errgroup.

test/util/grpctest/mocks.go
yarikbratashchuk: non test goroutine

pkg/user/tx_client_test.go
yarikbratashchuk: done

pkg/user/tx_client_test.go
yarikbratashchuk: done

pkg/user/tx_client.go
```diff
@@ -399,20 +426,96 @@ func (client *TxClient) broadcastTx(ctx context.Context, txBytes []byte, signer
...
+
+	close(respCh)
+	close(errCh)
+
+	if firstResp, ok := <-respCh; ok {
```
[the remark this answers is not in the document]
yarikbratashchuk: done

evan-forbes (approved):
if there are still flups, imo we should write them down as issues and merge. We'll likely want ajustments after dogfooding, so we can handle them then

pkg/user/tx_client.go
renaynay: Okay I see what the misunderstanding is here. This was quite confusing logic at first and not very explicit because you opted to use errgroup to take advantage of fail fast option, but inversely for a success case 😅  You are returning `context.Canceled` which is a non-nil error upon success case and signalling to errgroup to cancel childCtx which will be caught in other in-flight routines via select -- yes the logic technically works, but is difficult to read as errgroups should rather be used for operations where you want to parallelise some work but one error from one worker should cancel out operations on rest of groups + better err handling for that generally. 

Here, I would've opted for sync.WaitGroup and explicitly done the cancellation of ctx for other in-flight routines so it doesn't read like a success case --> err (then having to remember that err group cancels childCtx upon first err, whereas operations that did not finish in timeout return nil error rather than ctx Err). 

Tl;dr both options work - sync.WaitGroup would read better here IMO but i'm fine with keeping errgroup as long as the ctx cancellation is more explicit.

evan-forbes (approved):

renaynay (commented):
the changes lgtm but tests are failing: https://github.com/celestiaorg/celestia-app/actions/runs/14794484429/job/41538438460?pr=4628#step:4:189

evan-forbes:
can we merge this @renaynay ?

renaynay (approved):
Yeah 👍🏻 clear to merge @evan-forbes 

md lint failing tho but I don't think if it's bc of this PR.

evan-forbes (approved):

PR #4630, opened 2025-04-22
chore: fix `single-node.sh`

rootulp:
Previously, the `./scripts/single-node.sh` scripts failed for a couple reasons:
1. The genesis file gets created via celestia-app v4 but the celestia-app that started in the script was v3. v4 genesis files use `encoding/json` which serializes int64s as numbers in JSON (`1`). v3 genesis files use `cmtjson` which serializes int64s as strings in JSON (`"1"`).
2. A `--rpc.grpc_laddr tcp://0.0.0.0:9098` needs to be provided because there is currently no default. Will address in a follow-up.

## Testing

```
./scripts/single-node.sh
```

works


rootulp:
New error unlocked

```
3:18PM INF Using ABCI remote connection abci_version=ABCIClientVersion1 chain_id=test maximum_app_version=3 module=server
3:18PM ERR failed to start api server:
Error: error reading GenesisDoc at /Users/rootulp/.celestia-app/config/genesis.json: invalid 64-bit integer encoding "1", expected string
2025/04/22 15:18:47 Process finished with error: exit status 1
```

rootulp:
New error unlocked

```
Error: failed to start multiplexer: invalid listening address localhost:9090 (use fully formed addresses, including the tcp:// or unix:// prefix)
```

scripts/single-node.sh
```diff
@@ -103,6 +106,7 @@ startCelestiaApp() {
     --grpc.enable \
     --grpc-web.enable \
     --timeout-commit 1s \
+    --rpc.grpc_laddr tcp://0.0.0.0:9098 \
```
rach-id, line 109: grpc is no longer 9090?

scripts/single-node.sh
```diff
@@ -63,6 +63,9 @@ createGenesis() {
     # If you encounter: `sed: -I or -i may not be used with stdin` on MacOS you can mitigate by installing gnu-sed
     # https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d?permalink_comment_id=3082272#gistcomment-3082272
 
+    # Override the default genesis.json consensus params version from 0 to 4.
+    sed -i'.bak' 's#"app": "0"#"app": "4"#g' "${APP_HOME}"/config/genesis.json
```
rach-id, line 67: shouldn't this be changed in the app where we set the default to be the latest version?

rach-id (commented):
does it make sense to also update the other scripts?

scripts/single-node.sh
rootulp: It could be. Unfortunately the genesis file creation uses code from our SDK fork and celestia-core fork so it wasn't easy to make that change. I think we'd have to update the DefaultConsensusParams in celestia-core _and_ cosmos-sdk and then we'd have to do that repeatedly every time we bumped app versions. Alternatively we could override it in default_overrides.go but I don't think that will impact the genesis file.

scripts/single-node.sh
rootulp: GRPC is still 9090 but there is another service running on 9098 that apparently just supports `/broadcast_tx_commit`

damiannolan: The service being exposed on 9098 here is the cometbft grpc api. It exposes both broadcast api and more importantly, the block api introduced by celestia-core and consumed by celestia-node. 

It only truly needs to be turned on when running celestia-app through the multiplexer in standalone mode - as the app runs out of process to the celestia-core/multiplexer. In this case the app runs a proxy service to connect to the block api. There is a thread with full context in v4-protocol channel, if you have any other questions I can try answer them

tac0turtle (approved):
just ran into this, thanks for fixing

PR #4634, opened 2025-04-23
build: pin to v0.1.0 of the multiplexer and remove local pin

chatton:
## Overview

Remove dependency on local pin of multiplexer, pin to v0.1.0

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,9 +32,6 @@ RUN apk update && apk add --no-cache \
     musl-dev
 WORKDIR /celestia-app
 
-# multiplexer needs to be present due to local pin.
-COPY multiplexer multiplexer
-
 # cache go module dependencies
 COPY go.mod go.sum ./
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -46,7 +43,7 @@ RUN uname -a &&\
     CGO_ENABLED=${CGO_ENABLED} GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
     OVERRIDE_MAX_SQUARE_SIZE=${MAX_SQUARE_SIZE} \
     OVERRIDE_UPGRADE_HEIGHT_DELAY=${UPGRADE_HEIGHT_DELAY} \
-    make build
+    make build-standalone
```
chatton, line 46: since `build` now builds the multiplexer, we actually want to build the standalone in this Dockerfile

evan-forbes (approved):

Issue #4629, opened 2025-04-22
`Error: must set grpc listen address in config or by flag`


rootulp:
## Context

While testing https://github.com/celestiaorg/celestia-app/issues/4622 I discovered that ./scripts/single-node.sh still doesn't work. I assumed it would work after https://github.com/celestiaorg/celestia-app/pull/4570

## Problem

```
$ ./scripts/single-node.sh
celestia-app version: 3.2.1-rc1-180-g6d451cf2a
celestia-app home: /Users/rootulp/.celestia-app
celestia-app genesis file: /Users/rootulp/.celestia-app/config/genesis.json

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Initializing validator and node config files...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661
Starting celestia-app...
Error: must set grpc listen address in config or by flag
```

## Proposal

1. Improve the error message so that it clarifies the RPC GRPC listen address needs to be set
2. Set a default RPC GRPC listen address

rootulp:
```
$ cat ~/.celestia-app/config/config.toml | grep grpc
# Mechanism to connect to the ABCI application: socket | grpc
grpc_laddr = ""
grpc_max_open_connections = 900
# Does not include gRPC connections. See grpc_max_open_connections
```

Odd that the default config file doesn't populate grpc_laddr even though the help docs claim that it does

```
$ celestia-appd start --help

      --grpc.address string                             the gRPC server address to listen on (default "localhost:9090")
```

rootulp:
Wow this error is misleading because it's referring to `--rpc.grpc_laddr` and not `grpc_laddr`:

Ref: https://github.com/celestiaorg/celestia-app/blob/6d451cf2a65b7bd3a9db80954cbd629d97deb87f/multiplexer/cmd/start.go#L79-L81

rootulp:
1. The error message should be improved
2. We should modify the default config to use the default RPC GRPC address

PR #4637, opened 2025-04-23
chore: set default RPC GRPC listen address

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4629

1. Sets a default RPC GRPC listen address
2. Improves the error message

FYI: the default will only manifest if a user re-generates their config.toml via `celestia-appd init ...`


multiplexer/cmd/start.go
damiannolan, line 1: this will now require a new tag, right?

app/default_overrides.go
```diff
@@ -257,6 +257,7 @@ func DefaultConsensusConfig() *tmcfg.Config {
 	// Set broadcast timeout to be 50 seconds in order to avoid timeouts for long block times
 	cfg.RPC.TimeoutBroadcastTxCommit = 50 * time.Second
 	cfg.RPC.MaxBodyBytes = int64(8388608) // 8 MiB
+	cfg.RPC.GRPCListenAddress = "tcp://0.0.0.0:9098"
```
damiannolan, line 260: as a heads up, this will only take affect when you do `celestia-appd init` (creating a new config file with this), right?

If you have an existing state directory e.g. `~/.celestia-appd/` and your `config.toml` defines this as an empty string, I think this will get overridden by `interceptConfigs()` -> https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/server/util.go#L274-L279

damiannolan (approved):

evan-forbes (approved):

PR #4656, opened 2025-04-24
chore: option to provide home

tac0turtle:
## Overview

This pr adds an option to provide a home directory to the three scripts (single-node, single-node-upgrades and state-sync)

To use it run the script as normal and provide a home directory of your choice.


rach-id (approved):
utAck

evan-forbes (approved):

PR #4665, opened 2025-04-25
chore: remove embedded binaries

### LINKED ISSUE #4663: Remove embedded v3 binaries from version control
## Problem

https://github.com/celestiaorg/celestia-app/tree/main/internal/embedding contains two `*.tar.gz` which likely shouldn't be in version control

## Proposal

1. Remove https://github.com/celestiaorg/celestia-app/blob/main/internal/embedding/celestia-app_linux_v3_amd64.tar.gz
2. Remove https://github.com/celestiaorg/celestia-app/blob/main/internal/embedding/celestia-app_linux_v3_arm64.tar.gz
3. Ensure these don't get accidentally included in version control by adding them to gitignore

cc: @julienrbrt

rootulp:
Closes #4663 

We already have [this](https://github.com/rootulp/celestia-app/blob/a0ecb69da093588dad398643d764f02f90f9523e/.gitignore#L25-L26) in gitignore so the binaries should be ignored from now on.

## Testing

```
make install
```

Confirmed clean `git status`


julienrbrt (commented):
They shouldn't be deleted. They fulfill go embed requirements but are placeholder files replaced at make install 


EDIt: weird that they weren't present for all architectures. The other placeholders must have been removed in other PR. Anyway, I do not see CI complain so I suppose it is fine.

rootulp:
> EDIt: weird that they weren't present for all architectures. 

Agreed, that is weird.

julienrbrt:
> > EDIt: weird that they weren't present for all architectures. 
> 
> 
> 
> Agreed, that is weird. Sounds like we have two options:
> 
> 
> 
> 1. Remove these two placeholders
> 
> 2. Add back the other two placeholders
> 
> 
> 
> If they're not necessary, I'd rather remove them b/c I've seen them changed in git status every time I run `make install`

I thought a go build without the existing files would fail. (I'm not in front of my laptop so cannot check locally) unless CI is downloading them too I was expecting full red in CI. Anyway I think removing them makes sense, as it will fail at go build instead of runtime, as indeed you should effectively always have them.

julienrbrt:
Unless CI doesn't run tests with the multiplexer build tag ?

julienrbrt (approved):

evan-forbes (approved):

adlerjohn:
I'm getting an error running `make build` because of this.

```console
--> Updating go.mod
internal/embedding/data_linux_amd64.go:7:12: pattern celestia-app_linux_v3_amd64.tar.gz: no matching files found
make: *** [Makefile:48: build] Error 1
```

rootulp:
That's odd, I just retried and `make build` works on my machine

```
$ make build
--> Updating go.mod
```

but it fails in CI: https://github.com/celestiaorg/celestia-app/actions/runs/14674339653/job/41187721197#step:5:418

That's not a required check but we should add a required check for `make build` and revert this PR until we figure out why we need the linux placeholders and not the darwin ones.

rootulp:
Created https://github.com/celestiaorg/celestia-app/pull/4669 and https://github.com/celestiaorg/celestia-app/issues/4668

adlerjohn:
It's possible it worked on your machine because you already had the files there from previously, even if they aren't checked in to git?

rootulp:
Yes that's it.

```
$ rm -rf internal/embedding/*.tar.gz
$ make build
--> Updating go.mod
internal/embedding/data_darwin_arm64.go:7:12: pattern celestia-app_darwin_v3_arm64.tar.gz: no matching files found
make: *** [build] Error 1
```

Issue #4532, opened 2025-04-11
Refine CELESTIA_HOME usage


Wondertan:
### Problem
* `CELESTIA_HOME` conflicts with celestia-node's `CELESTIA_HOME`
* `CELESTIA_HOME` is used as `CELESTIA_HOME/.celestia-app` rather then as override for `~/.celestia-app`

### Proposal
* Rename `CELESTIA_HOME` to `CELESTIA_APP_HOME`
* Ensure `CELESTIA_APP_HOME` overrides `~/.celestia-app`, aligning with `--home` flag (cc @rach-id )

Follow up from https://github.com/celestiaorg/celestia-node/issues/4207#issuecomment-2796276455

Susmita-Dey:
Hello @Wondertan Please assign it to me. I'll make a PR after that.

rootulp:
> 1. Rename CELESTIA_HOME to CELESTIA_APP_HOME

If we do this, we should deprecate it on the v3.x branch . In v4.x we can remove the old env variable. This means you'll need two PRs:

1. A PR targeting v3.x that deprecates the existing env variable and introduces the new one. If the old env variable is used, log a warning.
2. A PR targeting v4.x that only uses the new env variable

Susmita-Dey:
Hey @Wondertan May I know when you're gonna launch v4? Asking this bcoz if v4 would be launched very soon then there won't be any of adding it to v3 as it will get deprecated

Wondertan:
No worries, it wont get deprecated even if version number increments.

Susmita-Dey:
> No worries, it wont get deprecated even if version number increments.

@Wondertan @rach-id I just need your guidance in which file shall I make the change as I see there are many files containing `CELESTIA_HOME`. Also, need clearance on **Ensure CELESTIA_APP_HOME overrides ~/.celestia-app, aligning with --home flag**

Wondertan:
I have no context on such details, so deferring to @rach-id

rach-id:
As stated in this comment, https://github.com/celestiaorg/celestia-app/issues/4532#issuecomment-2797295805, you will need two PRs, one to deprecate and make the change. The other to remove and only rely on the new environment variable.

The changes will generally go here: https://github.com/celestiaorg/celestia-app/blob/main/app/init.go

[...]

PR #4666, opened 2025-04-26
fix: `CELESTIA_HOME` for app v4

adlerjohn:
Ref: #4532

Per https://github.com/celestiaorg/celestia-app/issues/4532#issuecomment-2832582855, #4400 actually changed (breaking) the behavior of `CELESTIA_HOME` to be the directory directly. This PR renames the environment variable to `CELESTIA_APP_HOME` but does not further change the behavior.

## Summary by CodeRabbit

- **Chores**
  - Updated all references of the environment variable from `CELESTIA_HOME` to `CELESTIA_APP_HOME` across documentation, Dockerfiles, scripts, and configuration files.
  - Adjusted default directory paths and working directories to use `.celestia-appd` for improved consistency.
  - Refined documentation to reflect these changes and clarified environment variable usage.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -54,7 +54,7 @@ FROM ${RUNTIME_IMAGE} AS runtime
 # Ref: https://github.com/hexops/dockerfile/blob/main/README.md#do-not-use-a-uid-below-10000
 ARG UID=10001
 ARG USER_NAME=celestia
-ENV CELESTIA_HOME=/home/${USER_NAME}
+ENV CELESTIA_APP_HOME=/home/${USER_NAME}/.celestia-appd
 # hadolint ignore=DL3018
 RUN apk update && apk add --no-cache \
     bash \
@@ -63,7 +63,7 @@ RUN apk update && apk add --no-cache \
     && adduser ${USER_NAME} \
     -D \
     -g ${USER_NAME} \
-    -h ${CELESTIA_HOME} \
+    -h ${CELESTIA_APP_HOME} \
     -s /sbin/nologin \
     -u ${UID}
 # Copy the celestia-appd binary from the builder into the final image.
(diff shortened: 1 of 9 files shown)


local_devnet/docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
       - core0
     environment:
       - MONIKER=core1
-      - CELESTIA_HOME=/opt
+      - CELESTIA_APP_HOME=/opt
```
adlerjohn, line 42: It's unclear to me why `CELESTIA_HOME` was `/opt` in these files. That sounds like the install directory, not the directory to place data.

rach-id: if I remember correctly, at the day of creating these scripts, a couple years ago, I had some issues with alpine and the home directory and used `/opt`. Can definitly be changed

adlerjohn: Since you're familiar with this script, can you make an issue to follow-up on this? I'll leave it as it is in the PR since it's orthogonal.

rach-id: https://github.com/celestiaorg/celestia-app/issues/4670

rootulp (approved):

ninabarbakadze (approved):

rootulp (approved):

docker/Dockerfile
```diff
@@ -54,7 +54,7 @@ FROM ${RUNTIME_IMAGE} AS runtime
 # Ref: https://github.com/hexops/dockerfile/blob/main/README.md#do-not-use-a-uid-below-10000
 ARG UID=10001
 ARG USER_NAME=celestia
-ENV CELESTIA_HOME=/home/${USER_NAME}
+ENV CELESTIA_APP_HOME=/home/${USER_NAME}/.celestia-appd
```
damiannolan, line 57: This doesn't align with https://github.com/celestiaorg/celestia-app/blob/main/app/init.go#L18 which is a little confusing.

```diff
- celestia-appd
+ celestia-app
```

adlerjohn: Ohhh good catch.

adlerjohn: Can you fix @damiannolan? All instances of `.celestia-appd` actually

PR #4667, opened 2025-04-26
fix: `CELESTIA_HOME` for app v3

adlerjohn:
Ref: #4532

This PR adds the environment variable `CELESTIA_APP_HOME`, behaving as #4532 prescribes. It does not change the behavior of `CELESTIA_HOME` if used without `CELESTIA_APP_HOME`. When used together, `CELESTIA_APP_HOME` takes precedence (since `CELESTIA_HOME` can still be set for `celestia-node`).

### DIFF
--- a/app/init_test.go
+++ b/app/init_test.go
@@ -17,7 +17,7 @@ func Test_getDefaultNodeHome(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			name:            "want .celestia-app when userHome is empty and celestiaHomeOld is empty",
+			name:            "want .celestia-app when userHome is empty and celestiaHome/Old are empty",
 			userHome:        "",
 			celestiaHome:    "",
 			celestiaHomeOld: "",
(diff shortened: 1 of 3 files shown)


app/init_test.go
```diff
@@ -8,42 +8,75 @@ import (
...
+		want            string
 	}
 
 	testCases := []testCase{
 		{
-			name:         "want .celestia-app when userHome is empty and celestiaHome is empty",
-			userHome:     "",
-			celestiaHome: "",
-			want:         ".celestia-app",
+			name:            "want .celestia-app when userHome is empty and celestiaHome/Old are empty",
```
rootulp, line 20:
```suggestion
			name:            "want .celestia-app when userHome is empty and celestiaHomeOld are empty",
```

app/init_test.go
```diff
@@ -8,42 +8,75 @@ import (
...
+			celestiaHome:    "",
+			celestiaHomeOld: "celestia-home",
+			want:            "celestia-home/.celestia-app",
 		},
 		{
-			name:         "want user-home/.celestia-app when userHome is not empty and celestiaHome is empty",
-			userHome:     "user-home",
-			celestiaHome: "",
-			want:         "user-home/.celestia-app",
+			name:            "want user-home/.celestia-app when userHome is not empty and celestiaHome/Old are empty",
```
rootulp, line 34:
```suggestion
			name:            "want user-home/.celestia-app when userHome is not empty and celestiaHomeOld are empty",
```

rootulp (approved):
Thanks for adding the unit tests and marking `CELESTIA_HOME` as deprecated.

app/init_test.go
adlerjohn: Oh I meant this was celestiaHome and celestiHomeOld

rootulp: Oh got it, thanks

rootulp (approved):
Still LGTM

evan-forbes (approved):

adlerjohn:
@damiannolan could you also fix instances of `.celestia-appd` here?

PR #4589, opened 2025-04-16
test: add knuu upgrade test for v3 -> v4

chatton:
## Overview

This PR adds:

- An upgrade test from v3 -> v4 (using multiplexer)
- removes older upgrade tests. (they can be run from older branches if required.)
- Adds functionality to override images/tags being used. Initially for plugging in my own images for testing but I think it's actually a useful feature so decided to leave it in the PR.
- Added PFM wrapper with no-op migrations so the migration succeeds.
- Modified the test code to treat Genesis as raw bytes instead of a struct. This was required as we initially write a v3 genesis file (the importing struct is a v4 type, and so the contents cannot be marshalled and used directly )
- Added some utility functions to help with modifying the bytes + unit tests for them.
- Added static v3 generated genesis file as test fixture for this test.

I have tests passing with custom images

NOTE: currently seeing DNS issues in the k8s cluster, tests are failing intermittently with DNS resolution errors. If this is not a blocker happy to look into this in a follow up. (if code changes are required)

As images aren't building, I built images on my fork, running this command will allow you to run the test locally provided the KUBECONFIG is configured correctly. This image uses the embedded `v3.10.0-rc0` binary.

```sh
make test-e2e test=MajorUpgradeToV4 tag=cc703af registry=ghcr.io/chatton
```

Here is the full output of a passing test. Note: it look a long time due to lots of DNS errors in txsim, it crashed about 6 times before succeeding

```
make test-e2e test=MajorUpgradeToV4 tag=cc703af registry=ghcr.io/chatton
--> Running end to end tests
IMAGE_TAG=cc703af TEST=MajorUpgradeToV4 DOCKER_REGISTRY=ghcr.io/chatton go run ./test/e2e
test-e2e2025/04/23 09:42:25 === RUN MajorUpgradeToV4
INFO[2025-04-23T09:42:25+01:00]log/logger.go:41 Current log level                             env_log_level=LOG_LEVEL log_level=info
test-e2e2025/04/23 09:42:44 Knuu initialized with scope majorupgradetov4-20250423-094225
test-e2e2025/04/23 09:42:44 Creating testnet
test-e2e2025/04/23 09:42:44 Creating genesis nodes
test-e2e2025/04/23 09:42:44 Creating txsim
test-e2e2025/04/23 09:42:44 setting image for tx client name txsim image ghcr.io/chatton/txsim:cc703af
[...]

PR #4669, opened 2025-04-26
revert: "chore: remove embedded binaries"

rootulp:
Reverts celestiaorg/celestia-app#4665 because `make build` needs these placeholder files on Linux machines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ go.work.sum
 .envrc
 # test artifacts
 app/test/bad_block_*
+# do not include downloaded embedded binaries
+internal/embedding/**
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,3 @@ go.work.sum
 .envrc
 # test artifacts
 app/test/bad_block_*
-# do not include downloaded embedded binaries
-# internal/embedding/**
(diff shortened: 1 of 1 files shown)


tac0turtle (approved):

adlerjohn:
> weird that they weren't present for all architectures. The other placeholders must have been removed in other PR.
https://github.com/celestiaorg/celestia-app/pull/4665#pullrequestreview-2794440712

Can you also re-add the missing files for `darwin`? It looks like #4589 removed them.

I'm getting the error

```console
internal/embedding/data_darwin_arm64.go:7:12: pattern celestia-app_darwin_v3_arm64.tar.gz: no matching files found
make: *** [build] Error 1
```

when running `make build` on my mac.

rootulp:
TBH I don't understand why these placeholders need to exist and I think we should still explore removing them as a follow-up.

rootulp:
If we do https://github.com/celestiaorg/celestia-app/pull/4672 then we don't need this.

chatton:
it fails if the files do not exist as the default `make build/install` now builds in multiplexer mode. If the binaries do not exist in those paths, then you **cannot** embed them. Placeholders will mean that it will build, but will only function with the latest (v4) app.

And requests to older versions (1-3) will fail.

Issue #4677, opened 2025-04-28
pkg/inclusion: please delete calculateCommitmentPaths as it is not used anywhere


odeke-em:
The code for calculateCommitmentPaths is only used in tests and it is now defunct, please delete it.

I found a divide-by-zero problem in it and it is distracting to have code that's susceptible to some bugs left in
```shell
panic: runtime error: integer divide by zero [recovered, reraised]

goroutine 98 [running]:
testing.tRunner.func1.2({0x11372420, 0x11a9cf10})
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1741 +0x21c
testing.tRunner.func1()
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1744 +0x35b
panic({0x11372420?, 0x11a9cf10?})
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/panic.go:783 +0x132
github.com/celestiaorg/go-square/v2/inclusion.SubTreeWidth(0xc0000aa720?, 0x10a2033d?)
	/Users/emmanuelodeke/go/pkg/mod/github.com/celestiaorg/go-square/v2@v2.2.0/inclusion/blob_share_commitment_rules.go:84 +0xa5
github.com/celestiaorg/go-square/v2/inclusion.NextShareIndex(0x0, 0x8?, 0xc00026a388?)
	/Users/emmanuelodeke/go/pkg/mod/github.com/celestiaorg/go-square/v2@v2.2.0/inclusion/blob_share_commitment_rules.go:36 +0x1e
github.com/celestiaorg/celestia-app/v4/pkg/inclusion.calculateCommitmentPaths(0x1, 0x8?, 0x0, 0x0)
	/Users/emmanuelodeke/go/src/github.com/celestiaorg/celestia-app/pkg/inclusion/paths.go:21 +0x4f
```

rootulp:
> and it is now defunct

Do you know when / why it is defunct? IIRC a lot of the logic for `pkg/inclusion` was moved to https://github.com/celestiaorg/go-square but `calculateCommitmentPaths` doesn't exist there so I'm curious on the details for why it is defunct.

evan-forbes:
it was only ever needed to generate concise proofs of inclusion to a PFB share commitment, or non-deterministic square construction.

[...]

PR #4678, opened 2025-04-28
chore: delete calculateCommitmentPaths

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4677

I haven't audited pkg/inclusion so it's likely there are other functions that can be deleted.


evan-forbes (approved):
:saluting_face: 

writing this was fun, I'm also glad we don't need it after zk proving the square

ninabarbakadze (approved):

PR #4702, opened 2025-04-30
fix: update the default bond denom in default_overrides

chatton:
When running `celestia-appd init` the following is present in the gov params

```json
"expedited_min_deposit": [
      {
        "denom": "stake",
        "amount": "50000000"
      }
    ],
```

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -203,6 +203,9 @@ func (govModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
 	genState.Params.MaxDepositPeriod = &oneWeek
 	genState.Params.VotingPeriod = &oneWeek
 
+	// the ExpeditedMinDeposit list will always be of size one as set in NewGenesisState in the gov module.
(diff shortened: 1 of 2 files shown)


app/default_overrides.go
```diff
@@ -203,6 +203,9 @@ func (govModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
 	genState.Params.MaxDepositPeriod = &oneWeek
 	genState.Params.VotingPeriod = &oneWeek
 
+	// the ExpeditedMinDeposit list will always be of size one as set in NewGenesisState in the gov module.
```
rootulp, line 206: [not blocking] it would be nice to include in `default_overrides_test.go` a test case for this override so we don't accidentally revert it 😬

rootulp (approved):

rach-id (dismissed):

rootulp (approved):

evan-forbes (approved):

rootulp (approved):

Issue #4709, opened 2025-04-30
docker-build celestia-app-multiplexer fails on main


rootulp:
## Problem

https://github.com/celestiaorg/celestia-app/actions/runs/14760517755/job/41439821961 is failing on `main`

chatton:
The newly added `download_v3_binaries` is causing issues since the Dockerfile uses that Maketarget, but the multiplexer pulls it's binaries from a different container and shouldn't be downloading them at built time.

PR #4710, opened 2025-04-30
fix: docker build for multiplexer

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4709

## Testing

```
make build-docker-multiplexer
```

```
$ make build
--> Updating go.mod
--> Downloading celestia-app v3.9.0-rc0 binary

$ make build DOWNLOAD=false
--> Updating go.mod

$ make build DOWNLOAD=true
--> Updating go.mod
--> Downloading celestia-app v3.9.0-rc0 binary
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION := $(shell echo $(shell git describe --tags 2>/dev/null || git log -1 --format='%h') | sed 's/^v//')
+VERSION := $(shell echo $(shell git describe --tags --always --match "v*") | sed 's/^v//')
 COMMIT := $(shell git rev-parse --short HEAD)
 DOCKER := $(shell which docker)
 PROJECTNAME=$(shell basename "$(PWD)")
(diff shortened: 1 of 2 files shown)


evan-forbes (dismissed):

Makefile
```diff
@@ -48,6 +48,14 @@ build: mod download-v3-binaries
 	@go build $(BUILD_FLAGS_MULTIPLEXER) -o build/celestia-appd ./cmd/celestia-appd
 .PHONY: build
 
+## build-skip-download: Build the celestia-appd binary into the ./build directory and skip downloading the v3 binaries.
+build-skip-download: mod
+	@cd ./cmd/celestia-appd
+	@mkdir -p build/
+	@echo "--> Building build/celestia-appd with multiplexer enabled"
+	@go build $(BUILD_FLAGS_MULTIPLEXER) -o build/celestia-appd ./cmd/celestia-appd
+.PHONY: build-skip-download
```
chatton, line 57: rather than an additional target, it might cleaner to have

```sh
make build download=false
```

and have it default to true for running locally. WDYT?

rootulp: yea that works and seems preferable. Will update to your suggestion.

evan-forbes (dismissed):

Makefile
```diff
@@ -40,8 +40,12 @@ build-standalone: mod
 	@go build $(BUILD_FLAGS) -o build/ ./cmd/celestia-appd
 .PHONY: build-standalone
 
+download ?= true
```
adlerjohn, line 43: Makefile parameters are conventionally upper case: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html

> and reserving upper case for parameters that control implicit rules or for parameters that the user should override with command options

rootulp: Thanks! Will fix

rootulp: https://github.com/celestiaorg/celestia-app/pull/4710/commits/81b59830d18c202b1783c33baf894630886507d6

adlerjohn (approved):
Hmm so this looks like it'll work, but I'm not sure it's idiomatic Make. From e.g. https://web.mit.edu/gnu/doc/html/make_7.html, you can conditionally include lines or not. So I think the idiomatic way is to include a line that downloads the binaries, else not.

rootulp:
I tried to make it more idiomatic by using a Make conditional instead of a shell conditional

adlerjohn (approved):
utack

evan-forbes (approved):

Issue #4713, opened 2025-04-30
`failed to start native app: can not open trace writer for empty file`


rootulp:
## Problem

On current main https://github.com/celestiaorg/celestia-app/commit/656b5dc285af3e28beded7c040dfb55ab59b5041

```
$ ./scripts/single-node.sh
celestia-app version: multiplexer/v0.2.0-20-g656b5dc28
celestia-app home: /Users/rootulp/.celestia-app
celestia-app genesis file: /Users/rootulp/.celestia-app/config/genesis.json

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Initializing validator and node config files...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661
Starting celestia-app...
Error: failed to start multiplexer: failed to start native app: can not open trace writer for empty file
```

rootulp:
Looks like my nit on Cian's PR introduced a regression.

chatton:
ah it was relying on the `err` value being `nil` based on it being initialized as a named return value.

PR #4714, opened 2025-04-30
fix: can not open trace writer for empty file

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4713

## Testing

```shell
./scripts/single-node.sh
```

works now.

## Follow ups
- [ ] Create another multiplexer release
- [ ] Bump multiplexer release in go.mod

### DIFF
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -401,9 +401,11 @@ func openDB(rootDir string, backendType db.BackendType) (db.DB, error) {
 	return db.NewDB("application", backendType, dataDir)
 }
 
+// openTraceWriter opens a trace writer for the given file.
+// If the file is empty, it returns no writer and no error.
 func openTraceWriter(traceWriterFile string) (w io.WriteCloser, err error) {
 	if traceWriterFile == "" {
-		return w, fmt.Errorf("can not open trace writer for empty file")
+		return w, err
 	}
 	return os.OpenFile(
 		traceWriterFile,


evan-forbes (dismissed):

multiplexer/abci/multiplexer.go
```diff
@@ -401,9 +401,11 @@ func openDB(rootDir string, backendType db.BackendType) (db.DB, error) {
 	return db.NewDB("application", backendType, dataDir)
 }
 
+// openTraceWriter opens a trace writer for the given file.
+// If the file is empty, it returns no writer and no error.
 func openTraceWriter(traceWriterFile string) (w io.WriteCloser, err error) {
 	if traceWriterFile == "" {
-		return w, fmt.Errorf("can not open trace writer for empty file")
+		return w, err
```
odeke-em, line 408: Why not be explicit by returning `nil`?

odeke-em:
Also kindly please update the commit message to something more descriptive otherwise I was a little confused as to the fix. Perhaps
```shell
fix(multiplexer/abci): return nil error if openTraceWriter should return an empty file
```

which gives a clearer idea of what the fix is about.

multiplexer/abci/multiplexer.go
rootulp: That's what I asked for but openTraceWriter is invoked with an empty file and then this function should be a no-op instead of erroring

Ref: https://github.com/celestiaorg/celestia-app/issues/4713#issuecomment-2843010276

rootulp: Sure that also works

rootulp, line 408:
```suggestion
		return w, nil
```

odeke-em: Also you can add a simple regression test to ensure that we never have that problem regress so

```go
func TestOpenTraceWriterWithEmptyFileDoesNotError(t *testing.T) {
    in := new(bytes.Buffer)
    w, err := openTraceWriter("")
    require.NoError(t, err)
    require.Equal(t, w, in)
}
```

odeke-em: @rootulp kindly take a look at this test for a regression lock to prevent it from ever creeping back in.

evan-forbes (approved):

odeke-em:
@rootulp I had added a suggestion for a regression test, perhaps

```go
func TestOpenTraceWriterWithEmptyFileDoesNotError(t *testing.T) {
    in := new(io.PipeWriter)
    w, err := openTraceWriter("")
    require.NoError(t, err)
    require.Equal(t, w, in)
}
```

adlerjohn (approved):

adlerjohn:
@rootulp can you make an issue to add the regression test above?

Issue #4695, opened 2025-04-29
Why is multiplexer a separate `go.mod`?


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/4694

## Question

What are the pros/cons of isolating the multiplexer to it's own go.mod? @julienrbrt 

AFAIK one con is that the separate `go.mod` means there are extra steps needed to get code changes to multiplexer used in production. Specifically:
1. Create a new multiplexer release
2. Bump to new multiplexer version in celestia-app `go.mod`
3. Create a new celestia-app release

One pro is that external consumers can import multiplexer without using celestia-app but that doesn't seem like a use case we really support or care about.

julienrbrt:
Given that at most one release of celestia-app uses the multiplexer, then having its seperate go.mod isn't required. It does mean tagging/keeping the multiplexer in all release branches|

rootulp:
Sorry I don't understand. Won't all future releases of celestia-app use the multiplexer? I'm wondering why we need to tag it separately as opposed to including it directly inside the celestia-app Go package. IIUC then including it in celestia-app means we can eliminate steps 1 and 2 in the OP.

chatton:
I think it's technically not required at all to have a separate go mod. The multiplexer itself likely isn't usable by any other project as it depends on `celestia-core`. 

Since we can build the app with or without the multiplexer via build tags, I don't think it makes a difference if it is in its own go mod or not.

rootulp:
Given we don't expect any other project to use the multiplexer, seems like less friction to include the multiplexer in celestia-app's go.mod to eliminate those two extra steps but not critical to address immediately.

julienrbrt:
> Sorry I don't understand. Won't all future releases of celestia-app use the multiplexer? I'm wondering why we need to tag it separately as opposed to including it directly inside the celestia-app Go package. IIUC then including it in celestia-app means we can eliminate steps 1 and 2 in the OP. 

They will, but only the multiplexer 
package of the latest version will be used (as if you remove the go.mod, the logic will essentially be in v4, v5 and all afterwards celestia app go module). While if you keep it, there won't be such switch.

rootulp:
I assume we'll need the latest multiplexer in v4, v5, and onwards.

rootulp:
[...]

Issue #4676, opened 2025-04-28
Update linting workflow to lint multiplexer module


chatton:
The linting of the multiplexer module is not being considered in workflows at the moment. Workflows should be updated to call `make lint-fix` in the `multiplexer` module.

ref https://github.com/celestiaorg/celestia-app/pull/4675#issue-3024376537

PR #4736, opened 2025-05-02
refactor(multiplexer): remove separate go.mod

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4695
Closes https://github.com/celestiaorg/celestia-app/issues/4676 because multiplexer is included in celestia-app so the golangci-lint workflow that runs over celestia-app finds issues in multiplexer
Closes https://github.com/celestiaorg/celestia-app/issues/4737, looks like `test-race` was not running on the multiplexer tests before


evan-forbes (approved):
utACK

PR #4739, opened 2025-05-02
fix(app): gov param overrides for expedited min deposit

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4742

We need to fix the expedited min deposit params in two places:
1. genesis.json
2. x/gov module migration

## Testing

1. Ran `./scripts/single-node-upgrades.sh`
1. Wait until upgrade happens

Then the expedited min deposit uses the correct denom `utia`

```
$ celestia-appd query gov params
params:
  burn_vote_veto: true
  expedited_min_deposit:
  - amount: "50000000000"
    denom: utia
```


app/default_overrides.go
julienrbrt, line 1: ref: https://github.com/celestiaorg/cosmos-sdk/pull/585#pullrequestreview-2817284030

can we add 

```go
func init() {
  sdk.DefaultBondDenom = appconst.BondDenom
}

rootulp: Does this code change also impact the app upgrade handler? If not, where is that?

rootulp: I can't find any docs in Cosmos SDK about setting `DefaultBondDenom` via Algolia or any Markdown files with it.

Ref: https://docs.cosmos.network/

But it does seem to work. I still need to make a change to override the expedited min deposit amount in the migration.

damiannolan (dismissed):
ACK. Think we might need to run `make lint-fix`, but lgtm!

damiannolan (approved):

PR #4743, opened 2025-05-03
docs: fix typo

reject-i:
arbitrary - arbitrarily


ninabarbakadze (commented):
I don't think this change adds any value. I suggest you look through the good first issues if you'd like to contribute.

Issue #4740, opened 2025-05-02
`./scripts/single-node-upgrades.sh` doesn't work


rootulp:
## Context

I can no longer test a local v3 -> v4 upgrade

## Problem

```
$ make install
Checking if BBR is enabled...
sysctl: unknown oid 'net.ipv4.tcp_congestion_control'
WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README.
--> Downloading celestia-app v3.10.0-rc1 binary
Skipping download because expected version already downloaded: celestia-app_darwin_v3_arm64.tar.gz
--> Installing celestia-appd with multiplexer support
$ ./scripts/single-node-upgrades.sh
celestia-app version: 4.0.0-rc2-1-g83e4322d0
celestia-app home: /Users/rootulp/.celestia-app
celestia-app genesis file: /Users/rootulp/.celestia-app/config/genesis.json

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Initializing validator and node config files...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Starting celestia-app...
3:30PM INF initializing multiplexer app_version=3 chain_id=test module=server
3:30PM INF initialized remote app client address=127.0.0.1:36658 module=server
3:30PM INF starting comet node module=server
3:30PM INF No default value exists for the v2 upgrade height when the chainID is
3:30PM INF starting ABCI without Tendermint
3:30PM INF service start impl=multiAppConn module=proxy msg="Starting multiAppConn service"
3:30PM INF service start connection=query impl=localClient module=abci-client msg="Starting localClient service"
3:30PM INF service start connection=snapshot impl=localClient module=abci-client msg="Starting localClient service"
3:30PM INF service start connection=mempool impl=localClient module=abci-client msg="Starting localClient service"
3:30PM INF service start connection=consensus impl=localClient module=abci-client msg="Starting localClient service"
3:30PM INF service start impl=EventBus module=events msg="Starting EventBus service"
3:30PM INF service start impl=PubSub module=pubsub msg="Starting PubSub service"
3:30PM INF service start impl=IndexerService module=txindex msg="Starting IndexerService service"
3:30PM INF starting API server... module=api-server
3:30PM INF serve module=api-server msg={}
[...]

PR #4752, opened 2025-05-06
fix: single node upgrades

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4740

## Testing

```
...
9:04PM INF applying upgrade "v4" at height: 15 module=x/upgrade
```

### DIFF
--- a/multiplexer/abci/remote_v1.go
+++ b/multiplexer/abci/remote_v1.go
@@ -173,8 +173,7 @@ func (a *RemoteABCIClientV1) FinalizeBlock(req *abciv2.RequestFinalizeBlock) (*a
 	}
 	events = append(events, abciEventV1ToV2(endBlockResp.Events...)...)
 
-	// convert tx results
-	txResults := make([]*abciv2.ExecTxResult, 0, len(commitBlockResps))
+	txResults := make([]*abciv2.ExecTxResult, len(commitBlockResps))
 	for i, commitBlockResp := range commitBlockResps {
 		txResults[i] = &abciv2.ExecTxResult{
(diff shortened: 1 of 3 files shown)


multiplexer/abci/remote_v1.go
```diff
@@ -173,8 +173,7 @@ func (a *RemoteABCIClientV1) FinalizeBlock(req *abciv2.RequestFinalizeBlock) (*a
 	}
 	events = append(events, abciEventV1ToV2(endBlockResp.Events...)...)
 
-	// convert tx results
-	txResults := make([]*abciv2.ExecTxResult, 0, len(commitBlockResps))
```
rootulp, line 177: This fixes an issue I found _after_ the script started working. I introduced it recently. The length of the slice should not be set to `0`, it should be set to `len(commitBlockResps)`

evan-forbes (approved):

multiplexer/abci/remote_v1.go
chatton: nice catch, why was this working before? I would expect a panic on indexing at `0`.

chatton: oh that was the linter change that you mentioned in the description, yeah that's a pretty weird one the linter made that change 🤔

ninabarbakadze (approved):

multiplexer/abci/remote_v1.go
rootulp: Linter didn't make the change, it just flagged that I needed to use an allocated list and then I made the bug when trying to address the linter.

Issue #4642, opened 2025-04-23
Multiplexer starts with app version 1 even if chain should start on app version 3


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/4640

## Problem

The multiplexer doesn't load the genesis.json and use it to determine the app version correctly. It starts with app version 1 even if the genesis.json is on app version 3.

The multiplexer logs:

```
state.Version.Consensus.App: 1
2:45PM INF initializing multiplexer app_version=1 chain_id=test module=server
```

The genesis.json:
```json
  "consensus_params": {
    "version": {
      "app_version": "3"
    }
```

Looks like [this](https://github.com/rootulp/celestia-app/blob/5d682e24f4c571b1d4becbbe72f0f505decfc828/multiplexer/cmd/start.go#L61) does not work as expected.

cc: @julienrbrt

rootulp:
Yep confirmed that was the case.

julienrbrt:
On an exported genesis? It reads the consensus params from comet, not from the app state. That one is formatted as consensus.version.app.

rootulp:
A celestia-app v3 binary generates a genesis.json (via `celestia-appd init`) with `consensus.version.app_version`. I think that's also the case for a v1 and v2 binary. 

I think only v4 binary generates a genesis.json with `consensus.version.app`

julienrbrt:
Okay, then we probably need to have a fallback to the tendermint load state instead of comet if that field changed between tendermint/comet versions: https://github.com/celestiaorg/celestia-app/blob/main/multiplexer/cmd/start.go#L53-L67

rootulp:
One problem is that [this line ](https://github.com/rootulp/celestia-app/blob/e05d59bba68799a4d577928749075c18f8c95000/multiplexer/cmd/start.go#L62) does not throw an error for a v3 genesis.json. Instead it returns a state that is partially filled out.

```
[...]

PR #4758, opened 2025-05-06
fix(multiplexer): fallback to tendermint state libs

julienrbrt:
Closes: https://github.com/celestiaorg/celestia-app/issues/4642#issuecomment-2827550809

Fallback to tendermint state libs instead of cometbft based on genesis _version_.

### DIFF
--- a/multiplexer/internal/utils.go
+++ b/multiplexer/internal/utils.go
@@ -1,10 +1,63 @@
 package internal
 
 import (
+	"encoding/json"
+	"errors"
+	"os"
+
 	cmtcfg "github.com/cometbft/cometbft/config"
 	cmttypes "github.com/cometbft/cometbft/types"
 	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
 )
+
+type GenesisVersion int
+
+const (
+	GenesisVersion1 GenesisVersion = iota
+	GenesisVersion2
+)
+
+type genesisDocv1 struct {
+	ConsensusParams struct {
+		Version struct {
+			AppVersion string `json:"app_version"`
+		} `json:"version"`
+	} `json:"consensus_params"`
+}
+
+type genesisDocv2 struct {
+	Consensus struct {
+		Params struct {
+			Version struct {
+				App string `json:"app"`
+			} `json:"version"`
+		} `json:"params"`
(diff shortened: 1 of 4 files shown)


rootulp:
How was this tested? You can manually try with a genesis.json created from celestia-app v3 init for example, you can run this script after removing [these lines](https://github.com/celestiaorg/celestia-app/blob/33298ccb942e0ea1e9704f2f2dbdddcb3edc7562/scripts/single-node-upgrades.sh#L85-L87).

multiplexer/internal/utils.go
```diff
@@ -1,11 +1,66 @@
...
+			return GenesisVersion1, nil
+		}
+	}
+
+	var v2 genesisDocv2
+	if err := json.Unmarshal(genDoc, &v2); err == nil {
+		if v2.Consensus.Params.Version.App != "" {
+			return GenesisVersion2, nil
+		}
+	}
```
evan-forbes, line 59: [not blocking]
could optoinally add a test :innocent:  (could also just be a flup issue, the airdrop farmers can handle it lol)

evan-forbes (approved):
:+1: utack

rootulp:
I just manually tested and confirmed `./scripts/single-node-upgrades.sh` works without the HACKHACK so I removed it and pushed to this PR.

rootulp (approved):
Thanks for addressing so quickly! I pushed two commits to:
1. Remove a HACKHACK in the `./scripts/single-node-upgrade.sh` that is no longer necessary after this fix
2. Add unit tests to address @evan-forbes 's comment

multiplexer/internal/testdata/genesis.v4.json
```diff
@@ -0,0 +1,372 @@
...
+        "max_age_duration": "172800000000000",
+        "max_bytes": "1048576"
+      },
+      "validator": {
+        "pub_key_types": [
+          "ed25519"
+        ]
+      },
+      "version": {
+        "app": "4"
```
damiannolan, line 365: how did you generate this genesis file?

rootulp: `./scripts/single-node.sh` then copied it here

PR #4760, opened 2025-05-06
fix: use standard appDirectory where applicable

damiannolan:
## Overview

Fixes a regression introduced in https://github.com/celestiaorg/celestia-app/pull/4666

Should use the standard `appDirectory` const: https://github.com/celestiaorg/celestia-app/blob/main/app/init.go#L18


rootulp (approved):
At one point in time, I think it was `./celestia-appd` but `./celestia-app` seems preferable

Issue #4726, opened 2025-05-01
local_devnet is broken on `main`


rootulp:
## Context

On current `main` the local devnet is broken

## Problem

```
cd local_devnet
docker compose build
docker compose up -d
```

The 4 validators don't start because this error

```
2025-05-01 11:35:12             }': signature verification failed; please verify account number (0) and chain-id (test): (unable to verify single signer signature): unauthorized
2025-05-01 11:35:12 
2025-05-01 11:35:12 goroutine 1 [running]:
2025-05-01 11:35:12 github.com/cosmos/cosmos-sdk/x/genutil.AppModule.InitGenesis({{_}, {_, _}, {_, _}, {_, _}, {_, _}}, {{0x428bcb8, ...}, ...}, ...)
2025-05-01 11:35:12     /go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.29.1-sdk-v0.50.12/x/genutil/module.go:108 +0x130
2025-05-01 11:35:12 github.com/cosmos/cosmos-sdk/types/module.(*Manager).InitGenesis(_, {{0x428bcb8, 0x6a72c80}, {0x42ae420, 0x400209ae00}, {{0x0, 0x0}, {0x40026242d8, 0x4}, 0x0, ...}, ...}, ...)
2025-05-01 11:35:12     /go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.29.1-sdk-v0.50.12/types/module/module.go:508 +0x658
2025-05-01 11:35:12 github.com/celestiaorg/celestia-app/v4/app.(*App).InitChainer(_, {{0x428bcb8, 0x6a72c80}, {0x42ae420, 0x400209ae00}, {{0x0, 0x0}, {0x40026242d8, 0x4}, 0x0, ...}, ...}, ...)
2025-05-01 11:35:12     /celestia-app/app/app.go:596 +0x1d4
2025-05-01 11:35:12 github.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).InitChain(0x40011ce248, 0x4001fe1420)
2025-05-01 11:35:12     /go/pkg/mod/github.com/celestiaorg/cosmos-sdk@v1.29.1-sdk-v0.50.12/baseapp/abci.go:111 +0x580
2025-05-01 11:35:12 github.com/cosmos/cosmos-sdk/server.cometABCIWrapper.InitChain(...)
[...]

Issue #4763, opened 2025-05-06
Multiplexer doesn't gracefully shut down on CTRL + C


rootulp:
## Context

On v4.0.0-rc3

## Problem

I hit `CTRL + C` multiple times but I can't exit the multiplexer.

```logs
3:26PM INF finalized block block_app_hash=B6CC5814257136259DE475FF546BB5D161AA94589BE91AE23DDFB2580D44ADB5 height=29 module=state num_txs_res=0 num_val_updates=0
3:26PM INF executed block app_hash=B6CC5814257136259DE475FF546BB5D161AA94589BE91AE23DDFB2580D44ADB5 height=29 module=state
3:26PM INF committed state block_app_hash=A7180E3EC2C034BC98D5877232C40806A17FD19E2A16B485EAB02C507CD99A61 height=29 module=state
^C3:26PM ERR canceled grpc subscription. retrying module=rpc
3:26PM INF service stop impl="P2P Switch" module=p2p msg="Stopping P2P Switch service"
^[[A
3:26PM ERR Failed to subscribe to new blocks. retrying err="service is shutting down" module=rpc retry_number=1
^C^C^C^C^C^C3:26PM ERR Failed to subscribe to new blocks. retrying err="service is shutting down" module=rpc retry_number=2


^C^C3:26PM ERR Failed to subscribe to new blocks. retrying err="service is shutting down" module=rpc retry_number=3
^C^C^C^C^C3:26PM ERR Failed to subscribe to new blocks. retrying err="service is shutting down" module=rpc retry_number=4

^C
^C3:26PM ERR Failed to subscribe to new blocks. retrying err="service is shutting down" module=rpc retry_number=5

^C
^C







^C
^C
```

julienrbrt:
Interesting, I cannot reproduce and linux aarch64.

rootulp:
I also can't reliably repro. It only started manifesting for me recently via `$ ./scripts/single-node.sh`

chatton:
tried to reproduce this and was also not able to. Maybe we can close this issue and keep an eye out for it?

mindstyle85:
ive managed to get it, albeit a bit different messages in a systemd setup on `4.0.0-rc4-28-g1d232de1`


```
pops@Ubuntu-2404-noble-amd64-base:~$ sudo journalctl -u celestia-appd.service -f
[...]

PR #4764, opened 2025-05-06
test(scripts): add missing genesis

julienrbrt:
found when trying to replicate https://github.com/celestiaorg/celestia-app/issues/4763#issuecomment-2855796623.
running ` ./local_devnet/scripts/init.sh` would fail on hyperlane genesis unmarshal.

the script doesn't work nice because of the renaming of local_devnet chain id.
`MULTIPLEXER=true  ./local_devnet/scripts/init.sh` works fine tho.

### DIFF
--- a/multiplexer/cmd/start.go
+++ b/multiplexer/cmd/start.go
@@ -87,6 +87,9 @@ func getState(cfg *cmtcfg.Config) (state.State, error) {
 				return tmtypes.GenesisDocFromFile(cfg.GenesisFile())
 			},
 		)
+		if err != nil {
+			return state.State{}, err
+		}
 
 		// we only fill the app version and the chain id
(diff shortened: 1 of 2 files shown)


multiplexer/cmd/start.go
julienrbrt, line 1: unrelated to the other fix, but the err check was too low and this could have got us a nil pointer

evan-forbes (approved):

julienrbrt:
> Related: [#4726 (comment)](https://github.com/celestiaorg/celestia-app/issues/4726#issuecomment-2852995323)
> 
> Maybe this resolves #4726 too?

It won't solve #4726 no, because the chain id has been changed.

rootulp (approved):

ninabarbakadze (approved):

PR #4767, opened 2025-05-07
fix: ensure core block api runs through multilplexer native app

damiannolan:
## Overview

Due to the multiplexer copying start command functionality from the sdk, I discovered that running native apps (for example, v4) will not expose the core BlockAPI when through the multiplexer, i.e. the changes in https://github.com/celestiaorg/cosmos-sdk/pull/455 do not take effect which spin up the core BlockAPI in-process.

The BlockAPI is however available right now when:
- Running embedded apps (for example, v3) through the multiplexer. 
- Running a v4 standalone app without the multiplexer.

The replication of start command code being distributed and copied amongst different places is extremely error prone to manage, and quite frankly, its spaghetti.

### DIFF
--- a/docker/multiplexer.Dockerfile
+++ b/docker/multiplexer.Dockerfile
@@ -16,7 +16,7 @@ ARG MAX_SQUARE_SIZE
 ARG UPGRADE_HEIGHT_DELAY
 # the docker registry used for the embedded v3 binary.
 ARG CELESTIA_APP_REPOSITORY=ghcr.io/celestiaorg/celestia-app-standalone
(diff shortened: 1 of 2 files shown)


chatton:
Can you update the multiplexer.Dockerfile to depend on `ghcr.io/celestiaorg/celestia-app-standalone:v3.10.0-rc2` this image exists and should allow for a multiplexer build to succeed.

chatton (approved):
LGTM, nice catch. Hopefully we can get some full E2E tests soon that will catch stuff like this before manual QA

evan-forbes (approved):

Issue #4769, opened 2025-05-07
`celestia-appd init` creates genesis.json with `consensus.params.version.app: "0"`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/b0f41ea6fbf74d959a01f027310d2c5937a0a537/scripts/single-node.sh#L72-L73

## Problem

Currently `celestia-appd init foo` generates this genesis.json

```json
  "consensus": {
    "params": {
      "version": {
        "app": "0"
      },
```

which is unexpected and won't work

## Proposal

1. Fix `celestia-appd init` so that it populates `consensus.params.version.app: "4"`

rootulp:
The call trace goes from app -> sdk -> core and populates version with 0 because [this line](https://github.com/celestiaorg/celestia-core/blob/88dae3c1d03e6225c23076c1d2a3cce7ef58f79d/types/params.go#L123). We can resolve this quickly with

```diff
func DefaultVersionParams() VersionParams {
	return VersionParams{
-		App: 0,
+		App: 4, 
	}
}
```

but this diff won't work when app upgrades to app v5.

PR #4771, opened 2025-05-07
fix: `celestia-appd init` creates a genesis.json with latest app version

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4769 by doing the longer term fix. After copying the initCmd from cosmos-sdk I made a few quick refactors:
1. Remove the default denom flag because a user should never override the bond denom for Celestia genesis files
2. Rename variables for clarity
3. Rename import aliases for clarity

## Testing

```
make install
./scripts/single-node.sh
```

still works


julienrbrt (dismissed):
lgtm!

evan-forbes (approved):

julienrbrt (approved):

Issue #4773, opened 2025-05-08
Error logs on v3.x do not include underlying error


rootulp:
## Context

While working on https://github.com/celestiaorg/celestia-app/issues/4772 

## Problem

I discovered that some of the error logs on celestia-app v3.x do not log the error correctly. For example

This code: 
```go
	foo := "foo"
	ctx.Logger.Info("Starting in process node...: ", foo)
```

will emit this log:

```
10:49PM INF Starting in process node...:
```

so we have a few error logs that won't print the error https://github.com/celestiaorg/celestia-app/blob/de42a9e72d5c864853325c549476e84c7d7a717b/cmd/celestia-appd/cmd/start.go#L259 

## Proposal

Modify the error logs to actually include the error

rootulp:
I couldn't find instances of this pattern on `main` so seems isolated to one file on v3.x.

rootulp:
Closed by https://github.com/celestiaorg/celestia-app/pull/4774

PR #4774, opened 2025-05-08
fix(cmd): include error in logs

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4773

IMO it's a huge foot gun that it is possible to invoke `ctx.Logger.Error("msg", key)` without a corresponding value. 

If reviewers want, I'm also open to replacing all of the modified lines with this format which seems like the original intention:

```
ctx.Logger.Error(fmt.Sprintf("failed to start api server: %v", err))
```


evan-forbes (approved):
also agree that starting key values for logs after the first message is weird and a :foot: :gun:

chatton (approved):
😱 that's spooky, thanks for fixing.

Issue #4759, opened 2025-05-06
docker-multiplexer-build fails


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/actions/runs/14856617888/job/41711373999?pr=4755#step:9:279

## Problem

```
------
 > [linux/arm64 internal] load metadata for ghcr.io/celestiaorg/celestia-app-standalone:v3.10.0-rc0:
------
------
 > [linux/arm64 internal] load metadata for docker.io/library/alpine:3.19:
------
multiplexer.Dockerfile:28
--------------------
  26 |     # See https://github.com/hadolint/hadolint/issues/339
  27 |     # hadolint ignore=DL3006
  28 | >>> FROM --platform=$BUILDPLATFORM ${BUILDER_IMAGE} AS builder
  29 |     
  30 |     # must be specified for this build step in order for propagation of values.
--------------------
multiplexer.Dockerfile:22
--------------------
  20 |     
  21 |     # Stage 1: this base image contains already released binaries which can be embedded in the multiplexer.
  22 | >>> FROM ${CELESTIA_APP_REPOSITORY}:${CELESTIA_VERSION} AS base
  23 |     
  24 |     # Stage 2: Build the celestia-appd binary inside a builder image that will be discarded later.
--------------------
ERROR: failed to solve: ghcr.io/celestiaorg/celestia-app-standalone:v3.10.0-rc0: failed to resolve source metadata for ghcr.io/celestiaorg/celestia-app-standalone:v3.10.0-rc0: ghcr.io/celestiaorg/celestia-app-standalone:v3.10.0-rc0: not found
```

cc: @chatton

chatton:
will be fixed by #4767

rootulp:
Thanks! Closing b/c https://github.com/celestiaorg/celestia-app/pull/4767 merged.

PR #4778, opened 2025-05-08
test: basic E2E MsgPayForBlobs test

chatton:
## Overview

~This PR introduces a new E2E test which deploys a celestia-app, txsim and a bridge node in docker and ensures that the bridge node syncs headers.~

~This is still using my personal repo [here](https://github.com/chatton/celestia-test) which can be moved into the celestiaorg in the before or after merging this PR.~

~A nice follow up to this test would be to do something which uses the `BlockApi` to ensure we don't accidentally break compatibility.~

Little bit of scope creep on this PR, but it does the following.
- deploys celestia-app
- deploys a bridge-node
- submits a MsgPayForBlobs
- deploys a light-node
- queries blob data from light node

This should ensure the core components are all functioning together. In a follow up, I will update the CI to pass a celestia-app image built on the PR itself to this test.


rootulp:
Odd that docker-multiplexer-build CI failed even though this PR appears up to date with main.  https://github.com/celestiaorg/celestia-app/issues/4759 is closed b/c https://github.com/celestiaorg/celestia-app/pull/4767 merged.

.github/workflows/test-e2e.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         testcase:
           - TestE2ESimple
           - TestCelestiaChainStateSync
+          - TestE2ENodeSimple
```
rootulp, line 20: [uber nit][optional] `Simple` makes me conflate this with `TestE2ESimple`. To disambiguate, consider renaming to `TestE2EBridgeNode`

test/docker-e2e/e2e_test.go
```diff
@@ -53,30 +53,41 @@ func (s *CelestiaTestSuite) CreateCelestiaChainProvider(appVersion string) celes
...
+			Bech32Prefix:  "celestia",
+			Denom:         "utia",
+			CoinType:      "118",
+			GasPrices:     "0.025utia",
+			GasAdjustment: 1.3,
+			ModifyGenesis: func(config celestiadockertypes.Config, bytes []byte) ([]byte, error) {
+				return maps.SetField(bytes, "consensus.params.version.app", appVersion)
+			},
+			EncodingConfig:      &enc,
+			AdditionalStartArgs: []string{"--force-no-bbr", "--grpc.enable", "--grpc.address", "0.0.0.0:9090", "--rpc.grpc_laddr=tcp://0.0.0.0:9099"},
```
rootulp, line 80: [nit] does the RCP GRPC listen address need to differ from the default `9098`? https://github.com/celestiaorg/celestia-app/blob/6a2e7513dbb2c11808c1e8c99d35532d4d946aa1/app/default_overrides.go#L272

```suggestion
			AdditionalStartArgs: []string{"--force-no-bbr", "--grpc.enable", "--grpc.address", "0.0.0.0:9090", "--rpc.grpc_laddr=tcp://0.0.0.0:9098"},
```

rootulp (approved):
Awesome! Thanks for adding this

evan-forbes (commented):
this is dope!!

if we purposefully included a breaking change, then would this test will fail until node implemenets a fix? if so, should we add this in node instead?

chatton:
> Odd that docker-multiplexer-build CI failed even though this PR appears up to date with main. #4759 is closed b/c #4767 merged.

Ah, the base branch is the other branch that adds E2Es, so likely I haven't rebased on main there 👀

.github/workflows/test-e2e.yml
chatton: yes good call, will make this change.

test/docker-e2e/e2e_test.go
chatton: it shouldn't actually matter which port is used here, but no harm being consistent with the default.

evan-forbes (dismissed):
thanks again!

imo, we should keep this not-blocking, and then play it by hear when it is breaking. most changes shouldn't break, but for times we change the square format etc when we expect breaks, we're out of luck until we bump node with those changes.

rootulp (approved):

chatton:
converting to draft, will bump to the new celestiaorg repo and re-open

chatton:
I'll close this PR for now as there is going to be a bit of work done now on app/node to add some more tests and this has gotten a bit stale, but can still be used as a reference.

PR #4779, opened 2025-05-08
chore: fix minor typos and syntax issues in startup scripts

mutestt:
This PR makes small but meaningful improvements across several scripts:

- **Fixed a comment typo**  where "the some" was corrected to "some".
- **Corrected arithmetic syntax**, replacing `let CNT=CNT+1` with the more portable `CNT=$((CNT + 1))`.
- **Fixed a typo in output**,  "Settings seeds" → "Setting seeds".

### DIFF
--- a/scripts/localnet-blocks-test.sh
+++ b/scripts/localnet-blocks-test.sh
@@ -49,7 +49,7 @@ while [ ${CNT} -lt $ITER ]; do
     docker restart ${rand_container} &>/dev/null &
   fi
 
-  let CNT=CNT+1
+  CNT=$((CNT + 1))
   sleep $SLEEP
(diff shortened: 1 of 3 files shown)


scripts/localnet-blocks-test.sh
```diff
@@ -49,7 +49,7 @@ while [ ${CNT} -lt $ITER ]; do
     docker restart ${rand_container} &>/dev/null &
   fi
 
-  let CNT=CNT+1
+  CNT=$((CNT + 1))
```
evan-forbes, line 52: no drive by PRs can fix code that isn't actually broken. if this is broken open an issue to explain why

PR #4785, opened 2025-05-09
feat: tool to load network and measure latency

cmwaters:
Adds a new tool that can load a network and measure the latency. Full description in the README.md


evan-forbes (approved):
Might want to mention that the tool might not be maintained. If we don't use it often, I think it should be okay to delete rather than maintain. 

Personally I think a script that digests trace data is more useful since it can provide finer grained data over dissemination.

evan-forbes (approved):

tools/latency-monitor/main.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/celestiaorg/celestia-app/v4/app"
+	"github.com/celestiaorg/celestia-app/v4/app/encoding"
+	"github.com/celestiaorg/celestia-app/v4/pkg/user"
+)
+
+const (
+	defaultEndpoint     = "localhost:9090"
+	defaultKeyringDir   = "~/.celestia-app"
+	defaultSubmitRate   = 1.0    // KB per second
+	defaultBlobSize     = 1      // bytes
```
rootulp, line 30: I think the units here are wrong because later on this is used like so

```
				blobData := make([]byte, blobSize*1024)
```

```suggestion
	defaultBlobSize     = 1      // kibibytes
```

rootulp (approved):

PR #4790, opened 2025-05-12
chore: remove default gas price in tx client

cmwaters:
I believe this should have been removed as part of the switch to using the gas estimation services


ninabarbakadze (approved):
lgtm

evan-forbes (approved):
I'm missing context on what this solves functionally or if this is just cleaning up unused methods

rootulp:
I'm also missing some historical context but it looks like https://github.com/celestiaorg/celestia-app/pull/4459 modified the TxClient to use gas estimation endpoints so I assume we no longer need the `defaultGasPrice` that this PR removes.

rootulp (approved):

PR #4796, opened 2025-05-12
feat: add geographically distributed network generation tool

evan-forbes:
adds a tool to spin up a celestia-app network using talis

### DIFF
--- a/pkg/appconsts/versioned_consts.go
+++ b/pkg/appconsts/versioned_consts.go
@@ -4,42 +4,54 @@ import (
 	"strings"
 	"time"
 
-	appv4 "github.com/celestiaorg/celestia-app/v4/pkg/appconsts/v4" //nolint:review
+	"github.com/celestiaorg/celestia-app/v4/pkg/appconsts/v4"
 )
 
 const (
-	LatestVersion = appv4.Version
+	LatestVersion = v4.Version
 )
 
 var (
-	SquareSizeUpperBound      = appv4.SquareSizeUpperBound
-	TxSizeCostPerByte         = appv4.TxSizeCostPerByte
-	GasPerBlobByte            = appv4.GasPerBlobByte
-	Version                   = appv4.Version
-	UpgradeHeightDelay        = appv4.UpgradeHeightDelay
-	MaxTxSize                 = appv4.MaxTxSize
-	SubtreeRootThreshold      = appv4.SubtreeRootThreshold
-	TimeoutCommit             = appv4.TimeoutCommit
-	TimeoutPropose            = appv4.TimeoutPropose
+	SquareSizeUpperBound      = v4.SquareSizeUpperBound
 	TalisSquareSizeUpperBound = 512
+	TxSizeCostPerByte         = v4.TxSizeCostPerByte
+	GasPerBlobByte            = v4.GasPerBlobByte
+	Version                   = v4.Version
+	MaxTxSize                 = v4.MaxTxSize
+	SubtreeRootThreshold      = v4.SubtreeRootThreshold
+	TimeoutCommit             = v4.TimeoutCommit
+	TimeoutPropose            = v4.TimeoutPropose
+
+	TestUpgradeHeightDelay    = v4.TestUpgradeHeightDelay
+	ArabicaUpgradeHeightDelay = v4.ArabicaUpgradeHeightDelay
+	MochaUpgradeHeightDelay   = v4.MochaUpgradeHeightDelay
+	MainnetUpgradeHeightDelay = v4.MainnetUpgradeHeightDelay
+	// Deprecated: Use MainnetUpgradeHeightDelay instead.
+	UpgradeHeightDelay = v4.MainnetUpgradeHeightDelay
 )
 
 func GetTimeoutCommit(_ uint64) time.Duration {
-	return appv4.TimeoutCommit
+	return v4.TimeoutCommit
 }
 
 func GetSquareSizeUpperBound(chainID string) int {
 	if strings.Contains(chainID, TalisChainID) {
 		return TalisSquareSizeUpperBound
 	}
-	return appv4.SquareSizeUpperBound
+	return v4.SquareSizeUpperBound
 }
 
 // GetUpgradeHeightDelay returns the delay in blocks after a quorum has been
 // reached that the chain should upgrade to the new version.
 func GetUpgradeHeightDelay(chainID string) int64 {
 	if chainID == TestChainID {
-		return 3
+		return v4.TestUpgradeHeightDelay
 	}
-	return appv4.UpgradeHeightDelay
+	if chainID == ArabicaChainID {
+		return v4.ArabicaUpgradeHeightDelay
+	}
+	if chainID == MochaChainID {
+		return v4.MochaUpgradeHeightDelay
+	}
+	return v4.MainnetUpgradeHeightDelay
 }
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -23,6 +23,14 @@ pull_request_rules:
       backport:
         branches:
           - v3.x
+  - name: backport patches to v4.x branch
+    conditions:
+      - base=main
+      - label=backport:v4.x
+    actions:
+      backport:
+        branches:
+          - v4.x
   - name: forward-port patches to main branch (v1.x)
     conditions:
       - base=v1.x
--- a/pkg/appconsts/v4/app_consts.go
+++ b/pkg/appconsts/v4/app_consts.go
@@ -20,8 +20,22 @@ const (
 	MaxTxSize            int    = 2097152 // 2 MiB in bytes
 	TimeoutPropose              = time.Millisecond * 3500
 	TimeoutCommit               = time.Millisecond * 4200
-	// UpgradeHeightDelay is the number of blocks after a quorum has been
-	// reached that the chain should upgrade to the new version. Assuming a block
-	// interval of 6 seconds, this is 7 days.
-	UpgradeHeightDelay = int64(7 * 24 * 60 * 60 / 6) // 7 days * 24 hours * 60 minutes * 60 seconds / 6 seconds per block = 100,800 blocks.
+
+	// TestUpgradeHeightDelay is the number of blocks that chain-id "test" waits
+	// after a MsgTryUpgrade to activate the next version.
+	TestUpgradeHeightDelay = int64(3)
+	// ArabicaUpgradeHeightDelay is the number of blocks that Arabica waits
+	// after a MsgTryUpgrade to activate the next version. Assuming a block
+	// interval of 6 seconds, this is 1 day.
+	ArabicaUpgradeHeightDelay = int64(14_400)
+	// MochaUpgradeHeightDelay is the number of blocks that Mocha waits
+	// after a MsgTryUpgrade to activate the next version. Assuming a block
+	// interval of 6 seconds, this is 2 days.
+	MochaUpgradeHeightDelay = int64(28_800)
+	// MainnetUpgradeHeightDelay is the number of blocks that Mainnet waits
+	// after a MsgTryUpgrade to activate the next version. Assuming a block
+	// interval of 6 seconds, this is 7 day.
+	MainnetUpgradeHeightDelay = int64(100_800)
+	// Deprecated: Use MainnetUpgradeHeightDelay instead.
+	UpgradeHeightDelay = MainnetUpgradeHeightDelay
 )
--- a/pkg/appconsts/v4/app_consts_test.go
+++ b/pkg/appconsts/v4/app_consts_test.go
@@ -1,0 +1,25 @@
+package v4
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConsts(t *testing.T) {
+	t.Run("TestUpgradeHeightDelay should be 3", func(t *testing.T) {
+		require.Equal(t, int64(3), TestUpgradeHeightDelay)
+	})
+	t.Run("ArabicaUpgradeHeightDelay should be 1 day of 6 second blocks", func(t *testing.T) {
+		require.Equal(t, int64(1*60*60*24/6), ArabicaUpgradeHeightDelay)
+		require.Equal(t, int64(14_400), ArabicaUpgradeHeightDelay)
+	})
+	t.Run("MochaUpgradeHeightDelay should be 2 days of 6 second blocks", func(t *testing.T) {
+		require.Equal(t, int64(2*60*60*24/6), MochaUpgradeHeightDelay)
+		require.Equal(t, int64(28_800), MochaUpgradeHeightDelay)
+	})
+	t.Run("MainnetUpgradeHeightDelay should be 7 days of 6 second blocks", func(t *testing.T) {
+		require.Equal(t, int64(7*60*60*24/6), MainnetUpgradeHeightDelay)
+		require.Equal(t, int64(100_800), MainnetUpgradeHeightDelay)
+	})
+}
--- a/pkg/appconsts/versioned_consts_test.go
+++ b/pkg/appconsts/versioned_consts_test.go
@@ -7,28 +7,38 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestUpgradeHeightDelay(t *testing.T) {
+func TestGetUpgradeHeightDelay(t *testing.T) {
 	tests := []struct {
 		name                       string
 		chainID                    string
 		expectedUpgradeHeightDelay int64
 	}{
 		{
-			name:                       "the upgrade delay for chainID 'test' should be 3",
+			name:                       "the upgrade delay for chainID test",
 			chainID:                    appconsts.TestChainID,
-			expectedUpgradeHeightDelay: 3,
+			expectedUpgradeHeightDelay: appconsts.TestUpgradeHeightDelay,
 		},
 		{
-			name:                       "the upgrade delay should be latest value",
-			chainID:                    "arabica-11",
-			expectedUpgradeHeightDelay: appconsts.UpgradeHeightDelay,
+			name:                       "the upgrade delay for arabica",
+			chainID:                    appconsts.ArabicaChainID,
+			expectedUpgradeHeightDelay: appconsts.ArabicaUpgradeHeightDelay,
+		},
+		{
+			name:                       "the upgrade delay for mocha",
+			chainID:                    appconsts.MochaChainID,
+			expectedUpgradeHeightDelay: appconsts.MochaUpgradeHeightDelay,
+		},
+		{
+			name:                       "the upgrade delay for mainnet",
+			chainID:                    appconsts.MainnetChainID,
+			expectedUpgradeHeightDelay: appconsts.MainnetUpgradeHeightDelay,
 		},
 	}
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			actual := appconsts.GetUpgradeHeightDelay(tc.chainID)
-			require.Equal(t, tc.expectedUpgradeHeightDelay, actual)
+			got := appconsts.GetUpgradeHeightDelay(tc.chainID)
+			require.Equal(t, tc.expectedUpgradeHeightDelay, got)
 		})
 	}
 }


evan-forbes:
> What's the thinking behind using a CLI tool instead of running a go script (or go test like Knuu)?

The cli strikes a blance between automation and control.

entry points such as `make run-something` could definitely be added ontop, but the cli ux here is based on my own experience with experiments. 

Having a single command inherently means losing control over what is running. That's fine if we're spinning up a new network each time, but that's very slow. In my experience, I found that being able to quickly changes things in arbitrary ways enables much faster iteration, and if enough experimentation is done, a deeper understanding of how these protocols actually run IRL.

For example, keeping the same instances up and only pushing a new binary, genesis, or config is really fast, but that requires separating out those commands as that occurs at different times. In other instances, after collecting data I was curious about some other detail. Instead of waiting half an hour to run a new experiment, I could quickly add additional introspection to a single node and answer my question without even restarting the experiment.

Lastly, since there are so many moving pieces in a large geographically distributed network, random things will go wrong at a rate that essentially forces the experimenter to be in the loop to determine if the thing they changed caused a problem, or the cloud provider is just having issues.

multiplexer/internal/utils.go
```diff
@@ -60,14 +60,7 @@ func GetGenesisVersion(genesisPath string) (GenesisVersion, error) {
...
 
-	var v2 genesisDocv2
-	if err := json.Unmarshal(genDoc, &v2); err == nil {
-		if v2.Consensus.Params.Version.App != "" {
-			return GenesisVersion2, nil
-		}
-	}
-
-	return 0, errors.New("failed to determine genesis version")
+	return GenesisVersion2, nil
```
evan-forbes, line 51: note to self, revisit this as we might be able to delete other things if we limit this check

evan-forbes:
best place to start reviewing would be https://github.com/celestiaorg/celestia-app/blob/648fa07ab01832726191927afe1e0201088cae28/tools/talis/README.md

tools/talis/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+Note that this doesn't install binaries in the `$GOPATH/bin`, so you must specify the path when creating the payload with the `genesis` subcommand and `-a` (`--app-binary` path) and `-t` (`--txsim-binary` path) flags. See `genesis` subcomand usage below.
+
+## Usage
+
+if the relevant binaries are installed via go, and the celestia-app repo is
+downloaded, then the talis defaults should work. Your `$GOPATH` is used to copy the scripts from this repo to the payload, along with default locations for the binaries.
+
+### init
+```sh
```
rach-id: [nit]
```suggestion
### init

```sh
```

tools/talis/README.md
```diff
@@ -0,0 +1,200 @@
...
+# initializes the repo w/ editable scripts and configs
+talis init -c <chain-id> -e <experiment>
+```
+This will initiallize the directory that contains directory structure used for conducting an experiment.
```
rach-id, line 60:
```suggestion

This will initiallize the directory that contains directory structure used for conducting an experiment.
```

tools/talis/README.md
```diff
@@ -0,0 +1,200 @@
...
+{
+  "validators": [],
+  "chain_id": "talis-test-3",
+  "experiment": "test-3",
+  "ssh_pub_key_path": "/home/HOSTNAME/.ssh/id_ed25519.pub",
+  "ssh_key_name": "HOSTNAME",
+  "digitalocean_token": "pulled from env var if available",
+  "s3_config": {
+    "region": "pulled from env var if available",
+    "access_key_id": "pulled from env var if available,
```
rach-id:
```suggestion
    "access_key_id": "pulled from env var if available",
```

tools/talis/README.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+### add
+
+```sh
+# adds specific nodes to the config (see flags for further configuration)
+talis add  -t <node-type> -c <count>
+```
+
+If we call
```
rach-id:
```suggestion
If we call:
```

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+var (
+	valCount   = atomic.Uint32{}
+	nodeCount  = atomic.Uint32{}
+	lightCount = atomic.Uint32{}
+)
+
+// NodeName returns the name of the node based on its type and index. The
+// name is in the format "<node_type>-<index>". For example, "validator-0" or
+// "bridge-1". Index is a global counter that is incremented for each node created.
+func NodeName(nodeType NodeType) string {
```
rach-id, line 31: so calling `NodeName(Validator)` ten times would result in `valCount == 10`? can we separate this functionality by having a dedicated function for incrementing the number of nodes and this one only formatting the arguments to create a valid name? 

Otherwise, let's update the go doc to reflect that this is updating `valCount` which each invocation and not `index`.

tools/talis/docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Use the official Go builder image
+FROM golang:1.24 AS builder
```
rach-id, line 2: why use the builder tag here? doesn't look to be a multistaged build

tools/talis/scripts/upload_traces.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+export DEBIAN_FRONTEND=noninteractive
+export NEEDRESTART_MODE=a
+
+
+#export GOOGLE_APPLICATION_CREDENTIALS="/root/payload/congest-remote-key-gbq.json"
+apt install -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" apt-transport-https ca-certificates gnupg curl -y
+#curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg
+#echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+# sudo apt-get update -y && sudo apt-get install google-cloud-cli -y
```
rach-id: are these meant to be deleted?

tools/talis/scripts/upload_traces.sh
```diff
@@ -0,0 +1,54 @@
...
+# Set environment variables
+PROJECT_ID="numeric-mile-433416-e9"
+DATASET_ID="traces"
+
+CHAIN_ID=$CHAIN_ID
+
+LOCAL_DIR="/root/.celestia-app/data/traces"
+
+# gcloud auth activate-service-account --key-file="/root/payload/congest-remote-key-gbq.json" -q
+# gcloud config set project numeric-mile-433416-e9 -q
```
rach-id: same

tools/talis/scripts/upload_traces.sh
```diff
@@ -0,0 +1,54 @@
...
+# Parse the first part of the hostname
+nodeID=$(echo $hostname | awk -F'-' '{print $1 "-" $2}')
+
+source_dir="/root/.celestia-app/data/traces"
+logs_path="/root/logs"
+
+# clean the data by removing the last line
+find $source_dir -type f -name "*.jsonl" -exec sed -i '$d' {} \;
+
+#!/bin/bash
```
rach-id: this line looks unnecessary here

```suggestion
```

tools/talis/scripts/vars.sh
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
rach-id, line 1: should this file be deleted?

rach-id (approved):
left some nits. would love to try using it soon

tools/talis/digital_ocean.go
```diff
@@ -0,0 +1,397 @@
...
+		keys, resp, err := client.Keys.List(ctx, opt)
+		if err != nil {
+			return godo.Key{}, fmt.Errorf("failed to list SSH keys: %w", err)
+		}
+
+		for _, key := range keys {
+			// Compare the trimmed public key string
+			if strings.TrimSpace(key.PublicKey) == publicKey {
+				return key, nil
+			}
```
rach-id, line 79: this will fail to locate the key if we trim the last part of it specifying the hostname. So, I would prefer if we change this to only compare the first two parts of the keys:

```
curve pub_key hostname
```

so we only compare curve and pub_key and not also the hostname.

Otherwise, not do this and just use the `ssh_key_name` to query the exact account.

tools/talis/README.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+Note that this doesn't install binaries in the `$GOPATH/bin`, so you must specify the path when creating the payload with the `genesis` subcommand and `-a` (`--app-binary` path) and `-t` (`--txsim-binary` path) flags. See `genesis` subcomand usage below.
+
+## Usage
+
+if the relevant binaries are installed via go, and the celestia-app repo is
+downloaded, then the talis defaults should work. Your `$GOPATH` is used to copy the scripts from this repo to the payload, along with default locations for the binaries.
+
+### init
```
rach-id, line 53: A section specifying all env varialbes and what to put in them would be helpful:

```
TALIS_SSH_KEY_NAME
TALIS_SSH_KEY_PATH
DIGITALOCEAN_TOKEN
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_DEFAULT_REGION
AWS_S3_BUCKET
```

tools/talis/docker/Dockerfile
rach-id: nvm, it's used in the makefile

tools/talis/docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN mkdir -p /out
+
+WORKDIR /src
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+RUN CGO_ENABLED=0 go build -tags="ledger" -ldflags="${LDFLAGS}" -o /out/txsim ./test/cmd/txsim \
+ && CGO_ENABLED=0 GOOS=linux go build -tags="ledger" -ldflags="${LDFLAGS}" -o /out/celestia-appd ./cmd/celestia-appd
```
rach-id: to support targeting amd64 architecture even if building on an arm machine.

```suggestion
RUN GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 go build -tags="ledger" -ldflags="${LDFLAGS}" -o /out/txsim ./test/cmd/txsim \
 && GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=0 GOOS=linux go build -tags="ledger" -ldflags="${LDFLAGS}" -o /out/celestia-appd ./cmd/celestia-appd
```

Makefile
```diff
@@ -294,6 +295,32 @@ txsim-build-docker:
...
+	DOCKER_BUILDKIT=0 docker build \
+	  --file tools/talis/docker/Dockerfile \
+	  --target builder \
+	  --build-arg LDFLAGS="$(LLDFLAGS)" \
+	  --tag talis-builder:latest \
+	  .
+	mkdir -p build
+	docker create --name tmp talis-builder:latest
+	docker cp tmp:/out/. build/
+	docker rm tmp
```
rach-id: to target amd64 architecture even if running on any other architecture:

```suggestion
	docker build \
	  --file tools/talis/docker/Dockerfile \
	  --target builder \
	  --platform linux/amd64 \
	  --build-arg LDFLAGS="$(LLDFLAGS)" \
	  --build-arg GOOS=linux \
          --build-arg GOARCH=amd64 \
	  --tag talis-builder:latest \
	  .
	mkdir -p build
	docker create --platform linux/amd64 --name tmp talis-builder:latest
	docker cp tmp:/out/. build/
	docker rm tmp
```

tools/talis/README.md
```diff
@@ -0,0 +1,200 @@
...
+talis up
+```
+
+### genesis
+
+Before we can start the network, we need to create a payload that contains everything each instance needs to actually start the network. This includes all the required keys, configs, genesis.json, and startup scripts. The `--square-size` flag will change the `GovMaxSquareSize`.
+
+```sh
+# creates the payload for the network. This contains all addresses, configs, binaries (from your local GOPATH if not specified), genesis.json, and startup scripts. The `--square-size` flag will change the `GovMaxSquareSize`
+talis genesis -s 128 -a /home/$HOSTNAME/go/src/github.com/celestiaorg/celestia-app/build/celestia-appd -t /home/$HOSTNAME/go/src/github.com/celestiaorg/celestia-app/build/txsim
```
rach-id, line 156: would be good to specify that the `-a` and `-t` are the directories output by `make build-talis-bins`

tools/talis/deployment.go
```diff
@@ -0,0 +1,256 @@
...
+	var cfgPath string
+	var SSHKeyPath string
+
+	cmd := &cobra.Command{
+		Use:   "deploy",
+		Short: "Uses the config to spin up a distributed network",
+		Long:  "Initialize the Talis network with the provided configuration.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			tarPath := filepath.Join(rootDir, "payload.tar.gz")
+			tarCmd := exec.Command("tar", "-czf", tarPath, "-C", rootDir, "payload")
```
rach-id, line 85: [proposition]
~The payload is generally 400-500mb. For decent home internet, it takes time to upload to all servers. What do you think of switching to `xz` compression which will take 10sec more to compress but end up with > 50% compression of the payload~

This requires changing the scripts too, let's stick with gzip compression of 9:

```suggestion
			tarCmd := exec.Command("tar", "--gzip", "--options", "gzip:compression-level=9", "-cf", tarPath, "-C", rootDir, "payload")
```

tools/talis/txsim.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+
+			if len(cfg.Validators) == 0 {
+				return fmt.Errorf("no validators found in config")
+			}
+
+			resolvedSSHKeyPath := resolveValue(SSHKeyPath, EnvVarSSHKeyPath, strings.ReplaceAll(cfg.SSHPubKeyPath, ".pub", ""))
+
+			txsimScript := fmt.Sprintf(
+				"txsim .celestia-app/ --blob %d --blob-amounts %d --blob-sizes %d-%d --grpc-endpoint localhost:9091 --feegrant",
```
rach-id:
```suggestion
				"txsim .celestia-app/ --blob %d --blob-amounts %d --blob-sizes %d-%d --grpc-endpoint localhost:9091 --feegrant > txsim.log",
```

tools/talis/scripts/vars.sh
evan-forbes: iirc, this file just serves as something to copy and run for all nodes

currently we'll use to to export aws vars etc, but the user can write things here and they will get exported on each machine

we do need do document that at least, but I'm also ok with switching the mech as well

Makefile
```diff
@@ -88,6 +88,7 @@ mod:
 	@go mod tidy
 	@echo "--> Updating go.mod in ./test/interchain"
 	@(cd ./test/interchain && go mod tidy)
+	@(cd ./test/docker-e2e && go mod tidy)
```
rootulp:
```suggestion
	@echo "--> Updating go.mod in ./test/docker-e2e"
	@(cd ./test/docker-e2e && go mod tidy)
```

Makefile
```diff
@@ -294,6 +295,35 @@ txsim-build-docker:
 	docker build -t ghcr.io/celestiaorg/txsim -f docker/txsim/Dockerfile  .
 .PHONY: txsim-build-docker
 
+VERSION := $(shell git describe --tags --dirty --always)
+COMMIT  := $(shell git rev-parse HEAD)
```
rootulp: [blocking] Why are these repeated? They seem identical to the top of the file

Makefile
```diff
@@ -294,6 +295,35 @@ txsim-build-docker:
...
+
+# build the -ldflags string here
+LLDFLAGS := \
+  -X github.com/cosmos/cosmos-sdk/version.Name=celestia-app \
+  -X github.com/cosmos/cosmos-sdk/version.AppName=celestia-appd \
+  -X github.com/cosmos/cosmos-sdk/version.Version=$(VERSION) \
+  -X github.com/cosmos/cosmos-sdk/version.Commit=$(COMMIT) \
+  -X github.com/celestiaorg/celestia-app/v4/cmd/celestia-appd/cmd.v2UpgradeHeight=0
```
rootulp: [question] why name this LLDFLAGS instead of LDFLAGS? Why does it need to repeat the flags above?

Makefile
```diff
@@ -294,6 +295,35 @@ txsim-build-docker:
...
+
+build-talis-bins:
```
rootulp: [blocking] need a doc that shows up in `make help` output

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/celestiaorg/celestia-app/v4
 
 go 1.23.6
 
+toolchain go1.24.3
```
rootulp: [question] can we reconcile these two versions and bump the `go ` version to 1.24.3 if we need the toolchain bumped to 1.24.3?

Also if we do this, can we bump the README?

multiplexer/internal/utils.go
rootulp: I preferred the more explicit error if it couldn't determine the genesis version. Why remove that?

rootulp (commented):
Only reviewed the first couple files b/c I don't have time to review the rest

Makefile
rach-id: we could replace it with a dynamic one: 

```
mod:
	@echo "Running go mod tidy in all modules"
	@find . -type f -name 'go.mod' -execdir go mod tidy \;
```

tools/talis/deployment.go
```diff
@@ -0,0 +1,256 @@
...
+	cmd.Flags().StringVarP(&SSHKeyName, "ssh-key-name", "n", "", "name for the SSH key")
+
+	return cmd
+}
+
+// resolveValue selects a value based on priority: flag > env > config
+func resolveValue(flagVal, envKey, configVal string) string {
+	if flagVal != "" {
+		return flagVal
+	}
```
rach-id, line 251: this will always be the case if we have a default value set (which I guess is the case). We should instead check: `flag.Changed(...)` to see if the user changed it from the default value or not

tools/talis/deployment.go
```diff
@@ -0,0 +1,256 @@
...
+
+			if len(cfg.Validators) == 0 {
+				return fmt.Errorf("no validators found in config")
+			}
+
+			// overwrite the config values if flags or env vars are set
+			// flag > env > config
+			cfg.SSHKeyName = resolveValue(SSHKeyName, EnvVarSSHKeyName, cfg.SSHKeyName)
+			cfg.SSHPubKeyPath = resolveValue(SSHPubKeyPath, EnvVarSSHKeyPath, cfg.SSHPubKeyPath)
+			cfg.DigitalOceanToken = resolveValue(DOAPIToken, EnvVarDigitalOceanToken, cfg.DigitalOceanToken)
```
rach-id, line 224: `DOAPIToken` is not taken as a flag apparently

Makefile
evan-forbes: apologies, I think this snuck in with some merge thing or the when adding the docker binary builder too quickly

tools/talis/digital_ocean.go
evan-forbes: the problem with exact account name is that linode doesn't have account names, where as the actual key can be found in both.

will open a flup as I need to test to make sure it works with DO and could take a bit

tools/talis/deployment.go
evan-forbes: we can add a flup, TIL about flag.Changed!

multiplexer/internal/utils.go
evan-forbes: the genesis parsing logic would produce false negatives so had to be removed to get a v4 genesis to work (at least for me when creating this. perhaps it has since been fixed)

Makefile
evan-forbes: f363f9e47cd9fa0b94cd04c020ff25b8d1947d46

go.mod
evan-forbes: removed, the tooling is a bit annoying as it just auto added these

Makefile
evan-forbes: removed in the feedback commit

Makefile
evan-forbes: nice, updated to the dynamic one

tools/talis/deployment.go
evan-forbes: added in the feedback commit

tools/talis/README.md
evan-forbes: updated in the feedback commit

tools/talis/README.md
evan-forbes: #5018

tools/talis/digital_ocean.go
evan-forbes: #5019

tools/talis/README.md
evan-forbes: added a flup

tools/talis/config.go
evan-forbes: > Index is a global counter that is incremented for each node created.

do you want the var name specifically mentioned? imo if someone is curious on which is getting updated, they can go to the function to compare. agree we should have a function that just creates the tag, and another that counts

evan-forbes:
pinging @ninabarbakadze for review since you've just used it

tools/talis/config.go
rach-id: true 👍  That's fine, we can leave it how it is

Makefile
rootulp: I'm still seeing the undynamic one so I think a commit hasn't been pushed.

Makefile
rootulp: Still seem present so I think the feedback commit didn't get pushed.

Makefile
evan-forbes: apologies, I had two many clones open in parallel and something got lost

should be fixed in https://github.com/celestiaorg/celestia-app/pull/4796/commits/935a99cbb2ebcd08fe5494ac6dd99c1e2966b7ab

we can't use the dynamic one without also updating all the docker containers or something as they fail when calling `make mod`, so reverted to just using a hardcoded one for the 3 different go mods

rootulp: Sounds good to me. Resolving.

Makefile
```diff
@@ -294,6 +296,25 @@ txsim-build-docker:
...
+	  --build-arg LDFLAGS="$(ldflags)" \
+	  --build-arg GOOS=linux \
+          --build-arg GOARCH=amd64 \
+	  --tag talis-builder:latest \
+	  .
+	mkdir -p build
+	docker create --platform linux/amd64 --name tmp talis-builder:latest
+	docker cp tmp:/out/. build/
+	docker rm tmp
+
```
rootulp:
```suggestion
```

tools/talis/README.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+Note that this doesn't install binaries in the `$GOPATH/bin`, so you must specify the path when creating the payload with the `genesis` subcommand and `-a` (`--app-binary` path) and `-t` (`--txsim-binary` path) flags. See `genesis` subcomand usage below.
+
+## Usage
+
+if the relevant binaries are installed via go, and the celestia-app repo is
+downloaded, then the talis defaults should work. Your `$GOPATH` is used to copy the scripts from this repo to the payload, along with default locations for the binaries.
+
+### init
```
ninabarbakadze, line 53: `CopyTalisScripts` should be using more relative paths which is what i ran into first.

also, we should have a guide explaining prerequisites like DO setup etc so when a new engineer tries to run it all info is in one spot.

tools/talis/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```sh
+# adds specific nodes to the config (see flags for further configuration)
+talis add  -t <node-type> -c <count>
+```
+
+If we call:
+
+```
+talis add -t validator -t 1
```
ninabarbakadze:
```suggestion
talis add -t validator -c 1
```

pkg/appconsts/versioned_consts.go
```diff
@@ -1,29 +1,38 @@
...
 )
 
 func GetTimeoutCommit(_ uint64) time.Duration {
-	return v4.TimeoutCommit
+	return appv4.TimeoutCommit
+}
+
+func GetSquareSizeUpperBound(chainID string) int {
+	if strings.Contains(chainID, TalisChainID) {
+		return 512
```
rootulp: [nit] should this use

```suggestion
		return TalisSquareSizeUpperBound
```

or we can delete the exported constant `TalisSquareSizeUpperBound`

test/util/genesis/legacy_genesis.go
```diff
@@ -78,6 +79,18 @@ func DocumentLegacyBytes(
...
+	if err != nil {
+		return nil, fmt.Errorf("adding app_version field: %w", err)
+	}
+
+	// remove the version from the new location.
+	bz, err = maps.RemoveField(bz, "consensus_params.version.app")
+	if err != nil {
+		return nil, fmt.Errorf("removing version.app field: %w", err)
+	}
+
```
rootulp: [no change needed] seems copied from https://github.com/celestiaorg/celestia-app/pull/4823 so we can remove from this PR diff if we merge that PR first.

pkg/appconsts/versioned_consts.go
```diff
@@ -1,29 +1,38 @@
...
+	TimeoutPropose            = appv4.TimeoutPropose
+	TalisSquareSizeUpperBound = 512
 )
 
 func GetTimeoutCommit(_ uint64) time.Duration {
-	return v4.TimeoutCommit
+	return appv4.TimeoutCommit
+}
+
+func GetSquareSizeUpperBound(chainID string) int {
```
rootulp, line 37: [uber nit] could use a unit test. I can push a commit to this branch after the review.

multiplexer/internal/utils.go
rootulp: Do you have an example genesis.json where it would fail? I believe you but I'm worried that this will now silently return genesis version v2 when it can't deserialize a genesis.json.

In hindsight, the PR that introduced these conditionals should've included a unit test that tried deserializing an invalid genesis.json and asserted that the error was returned.

tools/talis/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+make build-talis-bins
+```
+
+Note that this doesn't install binaries in the `$GOPATH/bin`, so you must specify the path when creating the payload with the `genesis` subcommand and `-a` (`--app-binary` path) and `-t` (`--txsim-binary` path) flags. See `genesis` subcomand usage below.
+
+## Usage
+
+if the relevant binaries are installed via go, and the celestia-app repo is
```
rootulp:
```suggestion
If the relevant binaries are installed via go, and the celestia-app repo is
```

tools/talis/README.md
```diff
@@ -0,0 +1,202 @@
...
+### genesis
+
+Before we can start the network, we need to create a payload that contains everything each instance needs to actually start the network. This includes all the required keys, configs, genesis.json, and startup scripts. The `--square-size` flag will change the `GovMaxSquareSize`. By default, the binaries in the $GOPATH/bin will be used, however if specific binaries are needed (likely unless you are running some flavor of debian), use the -a (-a, --app-binary) and -t (-t, --txsim-binary) flags.
+
+```sh
+# creates the payload for the network. This contains all addresses, configs, binaries (from your local GOPATH if not specified), genesis.json, and startup scripts. The `--square-size` flag will change the `GovMaxSquareSize`
+talis genesis -s 128 -a /home/$HOSTNAME/go/src/github.com/celestiaorg/celestia-app/build/celestia-appd -t /home/$HOSTNAME/go/src/github.com/celestiaorg/celestia-app/build/txsim
+```
+
+Keep in mind that we can still edit anything in the payload before deploying the network. After creating it is when that would occur.
```
rootulp:
```suggestion
Keep in mind that we can still edit anything in the payload before deploying the network. 
```

tools/talis/add.go
```diff
@@ -0,0 +1,55 @@
...
+			return cfg.Save(rootDir)
+		},
+	}
+
+	cmd.Flags().StringVarP(&rootDir, "directory", "d", ".", "root directory in which to initialize")
+	cmd.Flags().IntVarP(&count, "count", "c", 0, "Name of the node")
+	_ = cmd.MarkFlagRequired("count")
+	cmd.Flags().StringVarP(&nodeType, "type", "t", "", "Type of the node (validator, bridge, light)")
+	_ = cmd.MarkFlagRequired("type")
+	cmd.Flags().StringP("provider", "p", "", "Provider for the node (digitalocean, linode)")
```
rootulp: [question] is linode actually supported? On line 30, it seems to always use DigitalOcean.

```suggestion
	cmd.Flags().StringP("provider", "p", "", "Provider for the node (digitalocean)")
```

tools/talis/add.go
```diff
@@ -0,0 +1,55 @@
...
+			default:
+				return fmt.Errorf("unknown node type %q", nodeType)
+			}
+
+			return cfg.Save(rootDir)
+		},
+	}
+
+	cmd.Flags().StringVarP(&rootDir, "directory", "d", ".", "root directory in which to initialize")
+	cmd.Flags().IntVarP(&count, "count", "c", 0, "Name of the node")
```
rootulp: [blocking] the help comment looks incorrect.

```suggestion
	cmd.Flags().IntVarP(&count, "count", "c", 0, "Number of nodes to deploy")
```

tools/talis/client.go
```diff
@@ -0,0 +1,105 @@
...
+	}, nil
+}
+
+func (c *Client) Up(ctx context.Context) error {
+	insts := make([]Instance, 0)
+	for _, v := range c.cfg.Validators {
+		if v.Provider != DigitalOcean {
+			log.Println("unexpectedly skipping instance since only DO is supported", v.Name, "in region", v.Region)
+			continue
+		}
```
rootulp, line 57: Can we remove this conditional because it looks like only DigitalOcean is supported and Linode isn't supported.

```suggestion
```

tools/talis/add.go
```diff
@@ -0,0 +1,55 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg, err := LoadConfig(rootDir)
+			if err != nil {
+				return fmt.Errorf("failed to load config %q: %w", rootDir, err)
+			}
+
+			switch nodeType {
+			case "validator":
+				for i := 0; i < count; i++ {
+					cfg = cfg.WithDigitalOceanValidator(region)
```
rootulp, line 30: [question] where does this get set? It appears to always be "" because line 15

```
	var (
		rootDir  string
		count    int
		nodeType string
		region   string
	)
```

If the flag doesn't work, let's hard-code random region and remove the flag.

tools/talis/add.go
```diff
@@ -0,0 +1,55 @@
...
+		},
+	}
+
+	cmd.Flags().StringVarP(&rootDir, "directory", "d", ".", "root directory in which to initialize")
+	cmd.Flags().IntVarP(&count, "count", "c", 0, "Name of the node")
+	_ = cmd.MarkFlagRequired("count")
+	cmd.Flags().StringVarP(&nodeType, "type", "t", "", "Type of the node (validator, bridge, light)")
+	_ = cmd.MarkFlagRequired("type")
+	cmd.Flags().StringP("provider", "p", "", "Provider for the node (digitalocean, linode)")
+	cmd.Flags().StringVarP(&region, "region", "r", "random", "the region to deploy the instance in (random if blank)")
```
rootulp, line 52: Does the `region` flag actually work? If not can it be removed?

tools/talis/client.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if v.Region == "" || v.Region == RandomRegion {
+			v.Region = RandomDORegion()
+		}
```
rootulp, line 61: This seems like it could be simplified b/c AFAICT the region will always be random because the CLI flag for region doesn't appear wired up.

```suggestion
		v.Region = RandomDORegion()
```

tools/talis/client.go
```diff
@@ -0,0 +1,105 @@
...
+	return err
+}
+
+func (c *Client) Down(ctx context.Context) error {
+	insts := make([]Instance, 0)
+	for _, v := range c.cfg.Validators {
+		if v.Provider != DigitalOcean {
+			log.Println("unexpectedly skipping instance since only DO is supported", v.Name, "in region", v.Region)
+			continue
+		}
```
rootulp, line 92: If only Digital Ocean is supported, this can be simplified

```suggestion
```

tools/talis/client.go
```diff
@@ -0,0 +1,105 @@
...
+		if v.Region == "" || v.Region == RandomRegion {
+			v.Region = RandomDORegion()
+		}
```
rootulp, line 95:
```suggestion
		v.Region = RandomDORegion()
```

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+
+var (
+	valCount   = atomic.Uint32{}
+	nodeCount  = atomic.Uint32{}
+	lightCount = atomic.Uint32{}
+)
+
+// NodeName returns the name of the node based on its type and index. The
+// name is in the format "<node_type>-<index>". For example, "validator-0" or
+// "bridge-1". Index is a global counter that is incremented for each node created.
```
rootulp, line 30: This comment appears misleading because there are separate counters for each node type

```suggestion
// "bridge-1". 
```

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+
+// Provider simply marks the provider the instance config should target.
+type Provider string
+
+const (
+	// DO represents DigitalOcean as a provider.
+	DigitalOcean Provider = "digitalocean"
+	// Linode represents Linode as a provider.
+	Linode Provider = "linode"
+)
```
rootulp, line 54: [nit] IMO could be simplified to drop support for Linode

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+// provider, region, and name. It also contains a list of tags that are
+// attached to the instance.
+type Instance struct {
+	NodeType NodeType `json:"node_type"`
+	// PublicIP is the public IP address of the instance.
+	PublicIP string `json:"public_ip"`
+	// PrivateIP is the private IP address of the instance.
+	PrivateIP string `json:"private_ip"`
+	// Provider is the provider of the instance. For example, "digitalocean" or
+	// "aws".
```
rootulp, line 67: Is aws a valid provider?

```suggestion
	// Provider is the provider of the instance. For example: digitalocean.
```

[nit] if there's only one provider, could we remove this field from the struct?

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+	Provider Provider `json:"provider"`
+	// Slug is a provider speicific string that determines what type of instance
+	// the node is ran on.
+	Slug string `json:"slug"`
+	// Region is the region in which the instance is created. For example,
+	// "nyc1" for DigitalOcean or "us-east-1" for AWS.
```
rootulp, line 73: AWS doesn't seem supported.

```suggestion
	// "nyc1" for DigitalOcean.
```

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+	// instances. This is used to identify the SSH key in the provider's
+	// dashboard. If it's not already kept by the provider, the key will be
+	// added.
+	SSHKeyName string `json:"ssh_key_name"`
+	// DigtialOceanToken is used to authenticate with DigitalOcean. It can be
+	// provided via an env var or flag.
+	DigitalOceanToken string `json:"digitalocean_token,omitempty"`
+	// LinodeToken is used to authenticate with Linode. It can be provided via
+	// an env var or flag.
+	LinodeToken string `json:"linode_token,omitempty"`
```
rootulp, line 123: If Linode isn't supproted

```suggestion
```

tools/talis/config.go
```diff
@@ -0,0 +1,246 @@
...
+
+func (cfg Config) WithDigitalOceanToken(token string) Config {
+	cfg.DigitalOceanToken = token
+	return cfg
+}
+
+func (cfg Config) WithLinodeToken(token string) Config {
+	cfg.LinodeToken = token
+	return cfg
+}
```
rootulp, line 163: If Linode isn't supported

```suggestion
```

tools/talis/digital_ocean.go
```diff
@@ -0,0 +1,397 @@
...
+		close(results)
+	}()
+
+	var created []Instance
+	for res := range results {
+		if res.err != nil {
+			fmt.Printf("❌ %s failed after %v %v\n", res.inst.Name, res.timeRequired, res.err)
+		} else {
+			created = append(created, res.inst)
+			fmt.Printf("✅ %s is up (public=%s) in %v\n",
```
rootulp, line 177: [uber nit][not blocking] I think emojis can show up incorrectly in certain Terminal emulators so I would try to avoid using them in logs and print statements.

tools/talis/init.go
```diff
@@ -0,0 +1,212 @@
...
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("git clone failed: %w", err)
+		}
+
+		src = filepath.Join(tmp, "tools", "talis", "scripts")
+	}
+
+	// 4) copy directory tree including subdirectories
```
rootulp:
```suggestion
	// copy directory tree including subdirectories
```

rootulp (approved):
I reviewed all the parts that touch real celestia-app. I skimmed the tools/talis code because seems partially AI generated and if it works, seems fine to ship. A few optional refactors to make it easier to maintain:
- Remove unsupported code. For example it looks like `linode`, and `aws` aren't actually supported providers so we could drop support from them.
- Remove unsupported node types like bridge and light.
- Use long flags instead of short flags in docs for readability. For example: prefer `--nodes` and `--count` instead of `-n` and `-c`
- Avoid usage of emojis in logs.

We could feed that context into AI for future code revisions.

[note] as discussed in sprint planning, we want to avoid including this in v4.x for now. Two options to resolve:
1. Create a v4.x branch 
2. Merge this to a non-main branch

multiplexer/internal/utils.go
evan-forbes: yeah any v4 genesis

>  this will now silently return genesis version v2 when it can't deserialize a genesis.json.

then the normal genesis deserialization will fail and return a more appropriate error. The only thing we need to do with this function is check if its an old genesis, not also confirm that its a new genesis

tools/talis/README.md
evan-forbes: good flup

tools/talis/client.go
evan-forbes: linode isn't supported yet, so we could but we will need to soon for 100+ node tests

tools/talis/client.go
evan-forbes: I'll fix so it uses the region good catch

tools/talis/client.go
evan-forbes: technically yeah, but similar to above comment we will need linode support

tools/talis/client.go
evan-forbes: same answer as above

tools/talis/config.go
evan-forbes: same answers as above

tools/talis/config.go
evan-forbes: same answer as above

tools/talis/config.go
evan-forbes: we might not support aws, but this was added on purpose to illustrate that these are different slugs per provider and everyone recognizes use-east-1 and what it means

tools/talis/config.go
evan-forbes: same answer as above

tools/talis/config.go
evan-forbes: same answer as above

tools/talis/digital_ocean.go
evan-forbes: they can use updated terminals or continue to hate fun imo

tools/talis/README.md
ninabarbakadze: I fixed `CopyTalisScripts` because I needed the fix for running it inside of a DO droplet

tools/talis/deployment.go
```diff
@@ -0,0 +1,259 @@
...
+	var SSHKeyPath string
+
+	cmd := &cobra.Command{
+		Use:   "deploy",
+		Short: "Uses the config to spin up a distributed network",
+		Long:  "Initialize the Talis network with the provided configuration.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			tarPath := filepath.Join(rootDir, "payload.tar.gz")
+			log.Printf("Compressing payload to %s\n", tarPath)
+			tarCmd := exec.Command("tar", "--xz", "--options", "xz:compression-level=7,xz:threads=0", "-cf", tarPath, "-C", rootDir, "payload")
```
ninabarbakadze: ran into issues here when running talis from inside of a droplet. will investigate and push a fix

tools/talis/init.go
```diff
@@ -131,10 +131,12 @@ func initDirs(rootDir string) error {
...
 func CopyTalisScripts(destDir string, srcRoot string) error {
-	// todo: fix import path
-	const importPath = "celestiaorg/celestia-app/tools/talis/scripts"
+	wd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get current directory: %w", err)
+	}
 
-	src := filepath.Join(srcRoot, "src", importPath)
+	src := filepath.Join(wd, "tools/talis/scripts")
```
evan-forbes: @ninabarbakadze does this assume one is running talis from celestia-app? the purpose of this is to copy scripts from the celestia-app repo

evan-forbes: reverted for now, as I don't think we should be determining the script location from the working directory. the scripts should be copied from the local copy of the celestia-app repo

https://github.com/celestiaorg/celestia-app/pull/4796/commits/27ef73704f8f1841c04976ed365f7e5d6b4449a3

evan-forbes:
created v4.x from main so merging this to main cc @rootulp

tools/talis/deployment.go
evan-forbes: did you define the gopath in the droplet? if it doesn't find stuff locally it should download them from git, but that likely can't work until we merge to main

rach-id (dismissed):

evan-forbes:
@ninabarbakadze see https://github.com/celestiaorg/celestia-app/blob/27ef73704f8f1841c04976ed365f7e5d6b4449a3/tools/talis/init.go#L93

and just removed the "celestiaorg" from the copytalisscripts func https://github.com/celestiaorg/celestia-app/pull/4796/commits/7825cecc9cbcbd351ceff53baea079d85735087c

rootulp:
> created v4.x from main so merging this to main cc @rootulp

Just flagging that this leads to extra overhead because now every (non-breaking) PR that targets `main` will need to be backported `v4.x`.

On second thought, this PR isn't breaking so do we really need to avoid including it in v4.x? If we include it in v4.x releases then we can delete the `v4.x` branch for now and don't need 2 PRs for every other PR (one that targets `main` and one that backports to `v4.x`).

evan-forbes:
> this PR isn't breaking so do we really need to avoid including it in v4.x?

I'll defer to you on this, and am ok with deleting the existing v4.x branch

evan-forbes:
need to moorge

if we have a flup let's write an issue and merge this one instead of adding more stuff please :innocent:

rach-id (approved):

Issue #4798, opened 2025-05-12
Goreleaser only creates standalone binaries


rootulp:
## Context

https://github.com/rootulp/celestia-app/releases/tag/v4.0.0-rc4

## Problem

Prebuilt binaries that are attached to releases are the _standalone_ variant and not the _multiplexer_ variant.

## Proposal

Modify GoReleaser to include the multiplexer variant in addition to the standalone variant.

chatton:
@rootulp Is the plan to manually attach the multiplexer binaries to the first release? If this is the case we won't need a warning and this can be handled automatically in goreleaser later.

rootulp:
Plan A: fix GoReleaser so we don't have to manually attach binaries
Plan B: If we don't fix GoReleaser in time then we can manually attach binaries

IMO we should keep the warning until we fix GoReleaser because the manual approach is unreliable and error-prone. Also we haven't done it for v4.0.0-arabica or v4.0.0-mocha 😓

PR #4799, opened 2025-05-13
docs: add note about multiplexer support

rootulp:
Add a warning to README until https://github.com/celestiaorg/celestia-app/issues/4798 is resolved.


evan-forbes (approved):
:eyes: 

just to make sure we're tracking this, do we have an issue?

Issue #4800, opened 2025-05-13
Arabica block sync does not work


rootulp:
## Context

As a sanity check before creating v4.0.0-arabica, I tried block syncing Arabica with a celestia-app binary from `main`

## Problem

```
$ ./scripts/arabica-block-sync.sh
...
11:17AM ERR Stopping peer for error err="reactor validation error: wrong Block.Header.Version. Expected {11 0}, got {11 1}" module=p2p peer="Peer{MConn{51.158.58.72:26656} 00d577159b2eb1f524ef9c37cb389c020a2c38d2 out}"
...
```

Looks like another issue with the multiplexer. In this case CometBFT looks like it is on app version 0 when it should be on app version 1

cc: @julienrbrt

rootulp:
The CometBFT consensus state is incorrect. I just reverted https://github.com/celestiaorg/celestia-app/pull/4758 locally but same error persists.

rootulp:
Diff of arabica's genesis.json and ./scripts/single-node-all-upgrades.sh: https://www.diffchecker.com/EobQa6r4/

rootulp:
1. Start Multiplexer
    1. Start app
        1. Start embedded app
    1. Start comet

After both embedded app and comet are started
1. Comet performs a handshake with app. 
    1. Comet calls `Info` on app
    1. App responds with app version `0`
    1. Comet calls `InitChain` on app
    1. App responds with empty consensus params <- seems like bug
    1. If consensus params are not empty, Comet updates state [here](https://github.com/celestiaorg/celestia-core/blob/65b0ab2c71b02193de49786b9e14699dcc3dced9/consensus/replay.go#L367)

rootulp:
I just checked out the latest `v3.x` branch, ran `make install && ./scripts/arabica-block-sync.sh`

```
InitChain response: {<nil> [{{0x140009cc210} 100000000}] [227 176 196 66 152 252 28 20 154 251 244 200 153 111 185 36 39 174 65 228 100 155 147 76 164 149 153 27 120 82 184 85] {10s 11s}}
2:23PM INF Completed ABCI Handshake - CometBFT and App are synced appHash="���B��\x1c\x14���șo�$'�A�d��L���\x1bxR�U" appHeight=0 appVersion=1 module=consensus
```

rootulp:
The state store reload is causing this.

```
before state store load state.Version.Consensus.App 1
after state store load state.Version.Consensus.App 0
```

Ref:
[...]

PR #4804, opened 2025-05-13
fix(multiplexer): arabica block sync

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4800

The multiplexer was converting a `nil` consensus params into a zeroed out consensus params. celestia-core used the zeroed out consensus params to overwrite the app version in state 😭 

This fix modifies the multiplexer to return `nil` consensus params if InitChain returned `nil` consensus params to match the behavior on v3.x.

## Testing

```
make install
./scripts/arabica-block-sync.sh
```

starts block syncing from genesis


evan-forbes (approved):
👀

PR #4822, opened 2025-05-15
feat(pkg/user): Add non-existent accounts to Signer's account map for lazy loading

### LINKED ISSUE #4803: feat(pkg/user): Signer should still add account to its account map even if it doesn't exist (yet)
Since `TxClient` lazy loads the account during tx submission (via `checkAccountLoaded`), I don't see why we can't just add the account to the `Signer` upon setup even if it does not yet exist. We loop over and skip it on L228 of tx_client.go but **it would be nice if we could somehow still add the account to the Signer such that it can try to lazy load it later if `AccountByAddress` is called.**

avorylli:
Fixes #4803 by:
- Modified SetupTxClient to add accounts with zero values if not in state
- Updated checkAccountLoaded to retry querying for zero-value accounts
- Enhanced AccountByAddress to lookup addresses in keyring when not found

Allows proper lazy-loading of accounts that don't exist during node startup.

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -144,18 +144,18 @@ func (s *Signer) Account(name string) *Account {
 	return s.accounts[name]
 }
 
-// AccountByAddress returns the account associated with the given address
-func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
+// AccountByAddress returns the account associated with the given address or an error if not found
+func (s *Signer) AccountByAddress(address sdktypes.AccAddress) (*Account, error) {
 	addrStr, err := s.addressCodec.BytesToString(address)
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("failed to convert address bytes to string: %w", err)
 	}
 
 	accountName, exists := s.addressToAccountMap[addrStr]
 	if !exists {
 		record, err := s.keys.KeyByAddress(address)
 		if err != nil {
-			return nil
+			return nil, fmt.Errorf("account with address %s not found in keyring: %w", addrStr, err)
 		}
 
 		acc := &Account{
@@ -167,16 +167,16 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
 
 		pk, err := record.GetPubKey()
 		if err != nil {
-			return nil
+			return nil, fmt.Errorf("failed to get public key for account %s: %w", record.Name, err)
 		}
 		acc.pubKey = pk
 
 		s.accounts[record.Name] = acc
 		s.addressToAccountMap[addrStr] = record.Name
 
-		return acc
+		return acc, nil
 	}
-	return s.accounts[accountName]
+	return s.accounts[accountName], nil
 }
 
 func (s *Signer) Accounts() []*Account {
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -153,7 +153,31 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
 
 	accountName, exists := s.addressToAccountMap[addrStr]
 	if !exists {
-		return nil
+		// Try to find this address in the keyring
+		record, err := s.keys.KeyByAddress(address)
+		if err != nil {
+			return nil
+		}
+
+		// We found a keyring record for this address
+		// Add it to our accounts map with zeroed account/sequence for lazy loading later
+		acc := &Account{
+			name:          record.Name,
+			address:       address,
+			accountNumber: 0,
+			sequence:      0,
+		}
+
+		// Get the pubkey if possible
+		if pk, err := record.GetPubKey(); err == nil {
+			acc.pubKey = pk
+		}
+
+		// Add to our maps
+		s.accounts[record.Name] = acc
+		s.addressToAccountMap[addrStr] = record.Name
+
+		return acc
 	}
 	return s.accounts[accountName]
 }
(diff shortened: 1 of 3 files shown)


pkg/user/signer.go
```diff
@@ -153,7 +153,31 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
 
 	accountName, exists := s.addressToAccountMap[addrStr]
 	if !exists {
-		return nil
+		// Try to find this address in the keyring
```
evan-forbes: nit
```suggestion
```

pkg/user/signer.go
```diff
@@ -153,7 +153,31 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
...
+		// We found a keyring record for this address
+		// Add it to our accounts map with zeroed account/sequence for lazy loading later
+		acc := &Account{
+			name:          record.Name,
+			address:       address,
+			accountNumber: 0,
+			sequence:      0,
+		}
+
+		// Get the pubkey if possible
```
evan-forbes: nit

```suggestion
```

pkg/user/signer.go
```diff
@@ -153,7 +153,31 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
...
+		}
+
+		// We found a keyring record for this address
+		// Add it to our accounts map with zeroed account/sequence for lazy loading later
+		acc := &Account{
+			name:          record.Name,
+			address:       address,
+			accountNumber: 0,
+			sequence:      0,
+		}
```
evan-forbes:
```suggestion
		acc := &Account{
			name:          record.Name,
			address:       address,
			accountNumber: 0,
			sequence:      0,
		}
```

pkg/user/signer.go
```diff
@@ -153,7 +153,31 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
...
+
+		// Get the pubkey if possible
+		if pk, err := record.GetPubKey(); err == nil {
+			acc.pubKey = pk
+		}
+
+		// Add to our maps
```
evan-forbes: nit

the code is simple enough to be self explanatory, so we can remove these comment as llms have made them tacky

pkg/user/signer.go
```diff
@@ -153,7 +153,28 @@ func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
...
+		acc := &Account{
+			name:          record.Name,
+			address:       address,
+			accountNumber: 0,
+			sequence:      0,
+		}
+
+		pk, err := record.GetPubKey()
+		if err != nil {
+			return nil
```
rootulp, line 170: [blocking][question] why does this silently ignore an error?

avorylli: > [blocking][question] why does this silently ignore an error?

@avorylli 

I think it doesn't silently ignore the error. It intentionally returns nil when it can't get the public key, as this indicates we can't fully populate the account.

rootulp: It is confusing to me that this function can return `nil` instead of a valid address especially because that's not communicated in the function signature. I would've expected

```
func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
```

to always return a valid account. If it is possible for this function to _not_ return an account then the function signature should be modified to

```
func (s *Signer) AccountByAddress(address sdktypes.AccAddress) (*Account, error) {
```

so that if `nil` is returned for the account, `err` can contain some information on _why_ the account was not returned.

evan-forbes: not having a public key is an error imo as well @avorylli. we should return an error

pkg/user/tx_client.go
```diff
@@ -683,7 +686,11 @@ func (client *TxClient) Account(name string) *Account {
 func (client *TxClient) AccountByAddress(address sdktypes.AccAddress) *Account {
 	client.mtx.Lock()
 	defer client.mtx.Unlock()
-	return client.signer.AccountByAddress(address)
+	acc, err := client.signer.AccountByAddress(address)
+	if err != nil {
+		return nil
+	}
+	return acc
 }
```
cmwaters, line 694: I think we should also propagate the error here as well

pkg/user/tx_client.go
```diff
@@ -242,7 +242,10 @@ func SetupTxClient(
 		}
 		accNum, seqNum, err := QueryAccount(ctx, conn, encCfg.InterfaceRegistry, addr)
 		if err != nil {
-			// skip over the accounts that don't exist in state
+			// Instead of skipping, add the account with zero values
+			// This allows the account to be added to the Signer's account map
+			// and potentially be lazy-loaded later when needed
+			accounts = append(accounts, NewAccount(record.Name, 0, 0))
```
cmwaters, line 248: IMO, I don't think we should add partially filled accounts here. It just requires too much additional logic. Either an account is full added (with account number and sequence number) or not at all (and thus can be lazily loaded during `checkAccountLoaded`)

pkg/user/signer.go
```diff
@@ -144,18 +144,39 @@ func (s *Signer) Account(name string) *Account {
 	return s.accounts[name]
```
cmwaters, line 144: Shouldn't this function also check for new keys added? Why just `AccountByAddress`?

cmwaters (changes requested):
I don't think we should have partially completed accounts

rootulp:
Closing in favor of https://github.com/celestiaorg/celestia-app/pull/4854

Issue #4780, opened 2025-05-08
`kv store with key KVStoreKey{0x14001565b00, signal} has not been registered in stores`


rootulp:
## Context

On current main

## Problem

./scripts/single-node-upgrades.sh

```
Error: rpc error: code = Unknown desc = kv store with key KVStoreKey{0x14001565b00, signal} has not been registered in stores: panic
```

rootulp:
When I last fixed this in https://github.com/celestiaorg/celestia-app/pull/4752

```
3:10PM INF ABCI Handshake App Info hash= height=0 module=consensus protocol-version=0 software-version=3.10.0-rc1
3:10PM INF ABCI Replay Blocks appHeight=0 module=consensus stateHeight=0 storeHeight=0
3:10PM INF mounting KV stores for app version 3
```

and then on https://github.com/celestiaorg/celestia-app/pull/4758 there was a regression to

```
3:13PM INF ABCI Handshake App Info hash= height=0 module=consensus protocol-version=0 software-version=3.10.0-rc2
3:13PM INF ABCI Replay Blocks appHeight=0 module=consensus stateHeight=0 storeHeight=0
3:13PM INF mounting KV stores for app version 1
3:13PM INF initializing blockchain state from genesis.json
```

rootulp:
Added more logs

```
9:12PM INF set default app version req.ConsensusParams={"block":{"max_bytes":1974272,"max_gas":-1},"evidence":{"max_age_duration":1814400000000000,"max_age_num_blocks":120961,"max_bytes":1048576},"validator":{"pub_key_types":["ed25519"]},"version":{}}
9:12PM INF consensus params app version is 0 so setting app version to 1
9:12PM INF InitChain mounting keys and initializing app version app version=1
9:12PM INF mounting KV stores for app version 1
```

Discovered that the problem is because v3.x `setDefaultAppVersion`

rootulp:
Will try to pair debug this with @julienrbrt tomorrow. In the meantime, I'll hack around it by updating the `single-node-upgrade.sh` script to upgrade from app version 1 -> 2 -> 3 -> 4.

rootulp:
This is a bug but it only manifest for genesis.json of v3 which won't happen on any production networks so less urgent to resolve.

damiannolan:
> This is a bug but it only manifest for genesis.json of v3 which won't happen on any production networks so less urgent to resolve.

[...]

PR #4823, opened 2025-05-15
fix(test/util): legacy genesis should use `app_version`

### LINKED ISSUE #4814: test-e2e times out
## Context

Last commit on main

https://github.com/celestiaorg/celestia-app/actions/runs/15008167375/job/42171702365

## Problem

test-e2e times out after running for 6 hours

chatton:
## Overview

The genesis file needs to be populated with the correct values expected by v3, we can't use the type directly as it has been modified so we need to update the bz directly.

closes #4814


evan-forbes:
did we want to merge this or is this added elsewhere?

chatton:
Ah yeah sorry forgot about this one 😅 , I think this logic is still correct but it's not being hit anywhere 'cause it was only in the v4 upgrade tests. I think we can still merge it since we'll want to have those upgrade tests anyway and this will make it a bit more obvious what the v3 genesis should look like.

test/util/genesis/legacy_genesis.go
```diff
@@ -78,6 +79,18 @@ func DocumentLegacyBytes(
...
+	bz, err = maps.SetField(bz, "consensus_params.version.app_version", strconv.Itoa(int(cp.Version.App)))
+	if err != nil {
+		return nil, fmt.Errorf("adding app_version field: %w", err)
+	}
+
+	// remove the version from the new location.
+	bz, err = maps.RemoveField(bz, "consensus_params.version.app")
+	if err != nil {
+		return nil, fmt.Errorf("removing version.app field: %w", err)
+	}
```
rootulp, line 92: Thanks. 

[no change needed] It was a huge pain that Cosmos SDK v0.50.x broke the format of the genesis file. This has come up a few times. Most recently still in:
- https://github.com/celestiaorg/celestia-app/issues/4780
- https://github.com/celestiaorg/celestia-app/pull/5010

rootulp (approved):
LGTM if CI passes. Out of curiosity, which tests were failing prior to this PR?

chatton:
> LGTM if CI passes. Out of curiosity, which tests were failing prior to this PR?
> 
> Update: looks like `test-e2e` but I think they've been replaced by test-docker-e2e and the upgrade tests aren't running in CI. 

currently no tests are actually using this since the knuu ones are removed, the main reason to merge this is to highlight the expected state of the genesis file for future upgrade tests using tastora.

test/util/genesis/legacy_genesis.go
```diff
@@ -78,6 +79,18 @@ func DocumentLegacyBytes(
...
 	}
 
+	// the version field used to be at this consensus_params.version.app_version
+	bz, err = maps.SetField(bz, "consensus_params.version.app_version", strconv.Itoa(int(cp.Version.App)))
+	if err != nil {
+		return nil, fmt.Errorf("adding app_version field: %w", err)
+	}
+
+	// remove the version from the new location.
+	bz, err = maps.RemoveField(bz, "consensus_params.version.app")
```
rach-id, line 89: to confirm my understanding, the `app_version` field was renamed to `app` and since this test might start using a v3 app version, we need to use the old name, right?

if so, shouldn't we keep both in case the test was started using v4 app?

rootulp: If the test was started using a v4 app then the multiplexer should be able to read genesis files that use the old format.

chatton: this code path is just for a v3 genesis, there is a separate codepath for v4, the legacy_genesis.go file was just added to handle older versions.

PR #4825, opened 2025-05-15
feat(cmd): Add command-line flags for Pyroscope profiling configuration

### LINKED ISSUE #4431: `--pyroscope-url` doesn't work
## Problem

Niky tried running Pyrscope with celestia-app and discovered that the CLI flag `--pyroscope-url` doesn't work

## Proposal

1. Reproduce his error
2. Fix the `--pyroscope-url` flag

hawkadrian:
## Overview

Fixes #4431

Adds three flags to configure Pyroscope without editing config.toml:
- `--pyroscope-url` - Sets Pyroscope server URL
- `--pyroscope-trace` - Enables tracing
- `--pyroscope-profile-types` - Configures profile types

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -36,6 +36,15 @@ const (
 
 	// TimeoutCommit is a flag that can be used to override the timeout_commit.
 	TimeoutCommitFlag = "timeout-commit"
+
+	// PyroscopeURLFlag is a flag to specify the URL for pyroscope profiling.
+	PyroscopeURLFlag = "pyroscope-url"
+
+	// PyroscopeTraceFlag is a flag to specify whether to enable pyroscope tracing.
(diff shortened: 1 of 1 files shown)


cmd/celestia-appd/cmd/root.go
```diff
@@ -36,6 +36,15 @@ const (
 
 	// TimeoutCommit is a flag that can be used to override the timeout_commit.
 	TimeoutCommitFlag = "timeout-commit"
+
+	// PyroscopeURLFlag is a flag to specify the URL for pyroscope profiling.
+	PyroscopeURLFlag = "pyroscope-url"
```
rootulp, line 41: There is already a pyroscope-url flag as reported in `celestia-appd start --help`:

```
      --pyroscope-url string                            URL of the Pyroscope instance to use for continuous profiling. If not specified, profiling will not be enabled
```

Before we add another flag, we should investigate why the existing flag doesn't work as expected.

PR #4835, opened 2025-05-19
fix: use correct test in ci workflow

chatton:
## Overview

🤦 use correct makefile target in CI, it was trying to run the Knuu tests with the names of the new docker tests. I also added a `continue-on-error` in the test matrix so each test will not be cancelled if others fail.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -10,7 +10,7 @@ on:
       - go.sum
 
 jobs:
-  test-e2e-docker:
+  test-docker-e2e:
     runs-on: ubuntu-latest
     # if one test fails, continue running the rest.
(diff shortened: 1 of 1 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -12,6 +12,8 @@ on:
 jobs:
   test-e2e-docker:
```
rootulp, line 13: [uber nit] it would be nice if this matched the make command. 

```suggestion
  test-docker-e2e:
```

or we modify the make command to match the workflow name

rootulp (dismissed):

rootulp (approved):

evan-forbes (approved):

Issue #4838, opened 2025-05-19
Every `celestia-appd` invocation creates a `/tmp/appd-v3-*` folder on Linux machines


rootulp:
## Context

Spun out of https://github.com/celestiaorg/celestia-app/issues/4832

On a linux machine. This doesn't reproduce on MacOS

## Problem

```
git checkout main
make install
celestia-appd version
4.0.0-mocha
```

Every subsequent `celestia-appd` invocation will create a tmp directory

```
root@rootulp:~/celestia-app# ls -al /tmp
total 4
drwxrwxrwt 27 root    root     540 May 19 18:40 .
drwxr-xr-x 19 root    root    4096 May 19 17:31 ..
drwxrwxrwt  2 root    root      40 May 19 17:31 .ICE-unix
drwxrwxrwt  2 root    root      40 May 19 17:31 .X11-unix
drwxrwxrwt  2 root    root      40 May 19 17:31 .XIM-unix
drwxrwxrwt  2 root    root      40 May 19 17:31 .font-unix
drwx------  2 root    root     100 May 19 17:59 appd-v3-1847296730
drwx------  2 newuser newuser  100 May 19 18:26 appd-v3-186757921
drwx------  2 newuser newuser  100 May 19 18:26 appd-v3-1974010179
drwx------  2 newuser newuser  100 May 19 18:26 appd-v3-2215971290
drwx------  2 root    root     100 May 19 17:59 appd-v3-2221767587
drwx------  2 root    root     100 May 19 18:40 appd-v3-2372263224
drwx------  2 root    root     100 May 19 17:59 appd-v3-2538645352
drwx------  2 root    root     100 May 19 18:01 appd-v3-3197384248
drwx------  2 root    root     100 May 19 17:59 appd-v3-3603797138
drwx------  2 root    root     100 May 19 17:59 appd-v3-3648348306
drwx------  2 root    root     100 May 19 17:59 appd-v3-4273400294
drwx------  2 root    root     100 May 19 17:59 appd-v3-598101303
drwx------  2 root    root     100 May 19 17:58 appd-v3-894917415
drwx------  2 root    root     100 May 19 17:59 appd-v3-956192267
drwxr-xr-x  2 root    root      40 May 19 18:01 celestia-test
drwx------  2 root    root      40 May 19 17:31 snap-private-tmp
[...]

PR #4842, opened 2025-05-20
fix(multiplexer): clean up tmp directories

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4838

Previously every `celestia-appd` invocation created a tmp directory and didn't clean it up. With this PR, the tmp directory gets cleaned up.

## Testing

```shell
$ ls  /var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/ | grep appd-v3-* | wc -l
    2289

$ rm -rf /var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/appd-v3-*

$ ls  /var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/ | grep appd-v3-* | wc -l
       0
       
$ make install

$ celestia-appd foo
Error: unknown command "foo" for "celestia-appd"
Run 'celestia-appd --help' for usage.

$ ls  /var/folders/y0/dd92_x8x4tlf397xstgwfz_c0000gn/T/ | grep appd-v3-* | wc -l
       0
```

### DIFF
--- a/multiplexer/appd/run.go
+++ b/multiplexer/appd/run.go
@@ -48,9 +48,7 @@ func New(name string, bin []byte, cfg ...CfgOption) (*Appd, error) {
 	}
 
 	defer func() {
-		if err != nil {
-			cleanup()
-		}
+		cleanup()
 	}()
 
 	// extract all files from the tar archive to the temp directory


tac0turtle (approved):

julienrbrt (approved):

damiannolan (approved):

cmwaters (approved):
🙏

ninabarbakadze (approved):
joining the party

chatton (approved):
This could just be `defer os.RemoveAll(tmpDir)` but it's functionally the same 🤷

julienrbrt:
> Ironic that this PR got so many approvals considering it introduced a regression that totally breaks the multiplexer. It was _extremely_ unexpected behavior that cleaning up a tmp directory would break the multiplexer.
> 
> Why don't we have any unit tests that verify the correct behavior of the multiplexer? @julienrbrt

Yeah, we all forgot that defer runs at the end of the function 😅.

To be fair, this behavior **_was_** tested and hardened in nova: https://github.com/01builders/nova/blob/360d472/appd/run_test.go#L12-L35.
So this change would have been catched there as well.
We actually removed that test to not need including in the repo git lfs or a binary. After this regression that went unnoticed, we should consider just pushing a binary to git, even-though it increase the repo size by 20mb (or push it to lfs, but that requires some CI changes + enough bandwith on the github side).

julienrbrt:
I do admit that the description of how the load works was a bit light in the readme, and the optimization for saving some space by not decompressing at every run should have been tracked somewhere.

> A chain defines a list of versions per AppVersion up until the last one. The Appd field a slice of []byte which holds a compressed archive of the binary (tar.gz). This archive is uncompressed and extracted to a temporary directory, which is then used to start the embedded binary.

https://github.com/celestiaorg/celestia-app/tree/v4.0.1-mocha/multiplexer#how-it-works

PR #4843, opened 2025-05-20
test: move E2E tests to run on each PR

chatton:
## Overview

This PR moves the docker E2E tests to be run on each PR. For we run each on a separate host, but we can move them to a dedicated self hosted runner and run them all in parallel there.

The celestia-app used is built from the code on any given PR, so we should find out if things break a bit sooner.  It looks like even with the regular GH runners they are still well under 10 mins. And less than the existing `go test` job.

A self hosted runner will help a lot in terms of layer caching (go mod download step will be the same 90%+ of the time)


rootulp (approved):

rootulp:
> A self hosted runner will help a lot in terms of layer caching (go mod download step will be the same 90%+ of the time)

If you end up getting blocked on pursuing this because of GitHub permissions then I have some repo admin permissions that may help unblock. Otherwise, John Adler should have org wide admin permissions and can likely give you the permissions to unblock.

evan-forbes (approved):

PR #4845, opened 2025-05-20
fix(e2e): proper timeout handling in initial height polling loop

mutestt:
Previously, the `break` inside the `select` block for `heightTimeoutCtx.Done()` only exited the `select`, not the surrounding `for` loop. As a result, if no block was produced within the timeout window, the test would continue looping, repeatedly logging `"Timed out waiting for initial height"`.

###  Fix
The loop is now labeled (`InitialHeightLoop:`), and `break InitialHeightLoop` is used to exit the polling loop cleanly on timeout. This ensures the test exits as expected and fails deterministically if block production is stuck.

### DIFF
--- a/test/docker-e2e/e2e_state_sync_test.go
+++ b/test/docker-e2e/e2e_state_sync_test.go
@@ -83,11 +83,12 @@ func (s *CelestiaTestSuite) TestStateSync() {
 	defer heightCancel()
 
 	// Check immediately first, then on ticker intervals
+	InitialHeightLoop:
 	for {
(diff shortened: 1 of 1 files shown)


rootulp:
> Previously, the break inside the select block for heightTimeoutCtx.Done() only exited the select, not the surrounding for loop. As a result, if no block was produced within the timeout window, the test would continue looping, repeatedly logging "Timed out waiting for initial height".

Can you point to a failing test or CI run of where this occurred? I'm skeptical of PRs that don't have associated issues.

PR #4851, opened 2025-05-20
fix(multiplexer): stop polluting `/tmp` with multiple v3.x binaries

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4867

## Description
- Uncompress v3.x binaries to a deterministic directory: `{NODE_HOME}/bin/v3.10.0-arabica`. 
- Skip uncompressing if a binary already exists for that version.
- Extracted helper functions from the `New` in run.go because it wasn't easy to comprehend or modify.
- Removes multiplexer `CfgOption`s because they were never used.
- Removes the `Appd.cleanup` field because it never worked.

## Testing

Works for default node home
```shell
$ ./scripts/single-node.sh 
# still works

$ ls ~/.celestia-app/bin
v3.10.0-arabica
```

Works for a custom node home directory. The script uses `--home ~/foo`
```shell
$ ./scripts/single-node.sh ~/foo
# still works

$ ls ~/foo/bin
v3.10.0-arabica
```


rootulp:
TODO: need to refactor to stop relying on the `.embed_version` file because if the binary is invoked from a different location, the path to the `.embed_version` will be wrong. One idea is to plumb the v3 version in as a flag similar to how we plumb the git commit into the binary to show up in the `celestia-appd version` output.

rootulp:
Encountering a weird error

```
$ ./scripts/single-node.sh
celestia-app version: Decompressing binary for version v3.10.0-arabica
Creating directory: /tmp/celestia-app/v3.10.0-arabica
Directory created: /tmp/celestia-app/v3.10.0-arabica
4.0.1-arabica-16-g0ed8d712c
celestia-app home: /Users/rootulp/.celestia-app
celestia-app genesis file: /Users/rootulp/.celestia-app/config/genesis.json

Do you want to delete existing /Users/rootulp/.celestia-app and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/.celestia-app...
Adding a new key to the keyring...
Adding genesis account...
Binary already decompressed for version v3.10.0-arabica
Error: failed to get address from Keyring: Binary already decompressed for version v3.10.0-arabica
celestia1x2l77rgs2mu403y7j908fgqt8stak6hy3nmv9c.info: key not found [celestiaorg/cosmos-sdk@v1.29.1-sdk-v0.50.12/crypto/keyring/keyring.go:521]
```

internal/embedding/data.go
```diff
@@ -7,14 +7,19 @@ import (
 	"runtime"
 )
 
-// CelestiaAppV3 returns the compressed platform specific Celestia binary.
-func CelestiaAppV3() ([]byte, error) {
+// NOTE: This version must be updated at the same time as the version in the
+// Makefile.
+const CELESTIA_V3_VERSION = "v3.10.0-arabica"
```
damiannolan, line 12: mega nit, consts in golang usually follow camel casing

internal/embedding/data_test.go
```diff
@@ -6,27 +6,28 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
```
damiannolan, line 9: curious why change from require to assert? they are basically the same anyways

multiplexer/appd/run.go
```diff
@@ -12,129 +12,54 @@ import (
...
+	stderr io.Writer
+	stdout io.Writer
 }
 
-// New takes a binary and untar it in a temporary directory.
-func New(name string, bin []byte, cfg ...CfgOption) (*Appd, error) {
-	if len(bin) == 0 {
-		return nil, fmt.Errorf("no binary data available: ensure `bin` is not empty")
+// New returns a new Appd instance.
+func New(version string, compressedBinary []byte) (*Appd, error) {
```
damiannolan, line 35: this function looks a lot better, nice one

multiplexer/appd/run.go
```diff
@@ -203,7 +119,131 @@ func (a *Appd) Stop() error {
...
+func (a *Appd) CreateExecCommand(args ...string) *exec.Cmd {
+	cmd := exec.Command(a.path, args...)
+	cmd.Stdin = a.stdin
+	cmd.Stdout = a.stdout
+	cmd.Stderr = a.stderr
+	return cmd
+}
+
+// getPathToBinary returns the path to the celestia-appd binary for the given version.
+func getPathToBinary(version string) (binaryPath string, err error) {
```
damiannolan, line 137: feel free to ignore/dismiss, but just want to flag use of named return args. Personally find they can sometimes be error prone

multiplexer/appd/run.go
```diff
@@ -203,7 +119,131 @@ func (a *Appd) Stop() error {
...
+
+	// look for the executable binary in the extracted files
+	err = filepath.Walk(baseDirectory, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.IsDir() && info.Mode()&0o111 != 0 {
+			binaryPath = path
+			return filepath.SkipAll // Found it, stop searching
```
damiannolan, line 148: TIL this is magic :D 

> SkipAll is used as a return value from [WalkDirFunc](https://pkg.go.dev/io/fs#WalkDirFunc) to indicate that all remaining files and directories are to be skipped. It is not returned as an error by any function.

https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/path/filepath/path.go;l=429-431

damiannolan (dismissed):
Much needed cleanup! Thank you @rootulp, nice work! 🙏🏻

internal/embedding/data.go
```diff
@@ -7,14 +7,19 @@ import (
...
+
+// CelestiaAppV3 returns the compressed platform specific Celestia binary and
+// the version.
+func CelestiaAppV3() (version string, compressedBinary []byte, err error) {
 	// Check if we actually have binary data
 	if len(v3binaryCompressed) == 0 {
```
chatton, line 18: just thinking now, it should be possible to have a version of the multiplexer with no embedded binaries if the binaries exist in the correct directory beforehand. 

I don't think we need to change anything now, but this could be useful for testing different binaries without constantly downloading and re-building / re-embedding all the time.

chatton (dismissed):
Looks great! Tested locally and everything seems to be working fine.

Left mostly nits, will leave it up to you if you think they're worth resolving.

multiplexer/appd/run.go
```diff
@@ -203,7 +119,131 @@ func (a *Appd) Stop() error {
...
+		return fmt.Errorf("binary validation failed (%s): %w\nOutput: %s", pathToBinary, err, string(testOutput))
+	}
+	return nil
+}
+
+// getDirectoryForCelestiaAppBinaries returns the directory where all
+// decompressed celestia-app binaries are stored. One directory exists per
+// version.
+func getDirectoryForCelestiaAppBinaries() string {
+	return filepath.Join(os.Getenv("HOME"), ".celestia-app", "bin")
```
chatton, line 243: is this always correct? This should be the cli specified home flag right? I think there is an issue with that now also.

evan-forbes (approved):
agree with the potential flups, but otherwise we should definitely merge

internal/embedding/data.go
rootulp: Agreed. Oops, I copied this over from the Makefile but forgot to fix the casing.

internal/embedding/data_test.go
rootulp: Require stops the execution of the test immediately so IMO it should be used for things in the test set up.

Assert doesn't stop the execution of the test so IMO it should be used for the assertions at the end of the test. This way a developer can see multiple test assertion failures (i.e. from different test cases)

multiplexer/appd/run.go
rootulp: Fair enough. Other developers on core/app also share your opinon.

I like named return arguments because they convey in the function signature _what_ is being returned. Without named return arguments, the godoc has to describe the return argument. My preference is for code to be self-documenting instead of relying on godocs which can go stale. In this instance the return arguments are evident from the function name so I'm open to removing them. IMO Named return arguments are extremely beneficial in instances with multiple arguments, [example](https://github.com/rootulp/celestia-app/blob/e9d28e5ab9c662d5088f267e919c51909878f442/internal/embedding/data.go#L16).

I dislike [naked returns](https://go.dev/tour/basics/7) but we can use named return arguments without naked returns. We even disable naked returns via linter [here](https://github.com/celestiaorg/celestia-app/blob/fdd9673f4d049aeef7b13be4ff04cc8b4e36a6ee/.golangci.yml#L23-L26)

rootulp: 3b7a02825468f50999967df7069fe7fbd0a154ff

multiplexer/appd/run.go
rootulp: That seems like a huge anti-pattern in the Go standard library to use errors for control flow 😞

multiplexer/appd/run.go
```diff
@@ -203,7 +119,131 @@ func (a *Appd) Stop() error {
...
+		}
+		f.Close()
+	}
+
+	return nil
+}
+
+// isBinaryAlreadyDecompressed returns true if the binary for the given version
+// has already been decompressed.
+func isBinaryAlreadyDecompressed(version string) bool {
```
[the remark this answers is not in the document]
rootulp: ae5b96ef274d9b9a6e51abbe57bd7b36a5e3355a

internal/embedding/data_test.go
```diff
@@ -6,27 +6,28 @@ import (
...
+		expectedVersion string
+		expectedError   error
 	}{
 		{
 			name: "valid binary data",
 			modifyFn: func() {
 				v3binaryCompressed = realData
 			},
-			expectedError: nil,
+			expectedVersion: "v3.10.0-arabica",
```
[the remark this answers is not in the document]
rootulp: c495855ad144289752134e1ec683e18b5c318c46

multiplexer/appd/run.go
```diff
@@ -12,129 +12,54 @@ import (
...
-		return nil, fmt.Errorf("no binary data available: ensure `bin` is not empty")
+// New returns a new Appd instance.
+func New(version string, compressedBinary []byte) (*Appd, error) {
+	if len(compressedBinary) == 0 {
+		return nil, fmt.Errorf("no compressed binary available for version %s", version)
 	}
 
-	// untar the binary.
-	gzr, err := gzip.NewReader(bytes.NewReader(bin))
+	err := decompressBinary(version, compressedBinary)
```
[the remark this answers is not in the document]
rootulp: 3196f461b3777acd0dc101523a60a9c71a1bceaf

multiplexer/appd/run.go
```diff
@@ -12,129 +12,54 @@ import (
...
-		}
-
-		if !info.IsDir() && info.Mode()&0o111 != 0 {
-			binaryPath = path
-			return filepath.SkipAll // Found it, stop searching
-		}
-		return nil
-	})
-
+	err = verifyBinaryIsExecutable(pathToBinary)
```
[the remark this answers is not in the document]
rootulp: 3c287ac34f18cfddbf54854f08e50c4d24f5ff76

multiplexer/appd/run.go
rootulp: I couldn't get the `--home` value plumbed here because the multiplexer gets created and added to the root command.

I will refactor this to use the CELESTIA_APP_HOME environment variable though.

Ref: https://github.com/celestiaorg/celestia-app?tab=readme-ov-file#environment-variables

multiplexer/appd/init.go
```diff
@@ -0,0 +1,32 @@
...
+func init() {
+	clienthelpers.EnvPrefix = envPrefix
+
+	var err error
+	nodeHome, err = clienthelpers.GetNodeHomeDirectory(appDirectory)
+	if err != nil {
+		// The userHome is not set in Vercel's Go runtime so log a warning but don't panic.
+		fmt.Printf("Warning userHome err: %s\n", err)
+	}
+}
```
rootulp, line 32: [note for reviewers] this was copied from app/init.go and sets `nodeHome` to the user specified `--home` value or the default if unspecified.

multiplexer/appd/run.go
rootulp: Got the `--home` value plumbed through.

multiplexer/appd/run.go
```diff
@@ -12,123 +12,52 @@ import (
...
+)
 
-// Appd represents the appd binary
+// Appd represents a celestia-appd binary.
 type Appd struct {
-	pid            int
-	path           string
-	stdin          io.Reader
-	stderr, stdout io.Writer
-	cleanup        func()
```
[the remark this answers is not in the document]
rootulp: I don't understand the proposal in that comment but that thread seems resolvable because cleanup was removed in this PR.

internal/embedding/data.go
```diff
@@ -7,14 +7,19 @@ import (
 	"runtime"
 )
 
-// CelestiaAppV3 returns the compressed platform specific Celestia binary.
-func CelestiaAppV3() ([]byte, error) {
+// NOTE: This version must be updated at the same time as the version in the
+// Makefile.
+const v3Version = "v3.10.0-arabica"
```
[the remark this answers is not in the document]
rootulp: I'd rather not override it with ldflags for now because it's more cumbersome to modify and test. In hindsight I wish we also left the `--v2-upgrade-height` param as passable to v4.x and let it proxy to v3.x instead of passing that via ldflags.

PR #4854, opened 2025-05-21
fix: pre populate addressToAccountMap in tx client

### LINKED ISSUE #4803: feat(pkg/user): Signer should still add account to its account map even if it doesn't exist (yet)

chatton:
## Overview

closes #4803

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -58,7 +58,35 @@ func NewSigner(keys keyring.Keyring, encCfg client.TxConfig, chainID string, acc
 		}
 	}
 
+	// pre-populate the address to account map.
+	// if the accounts do not yet exist, they will be lazily loaded during
+	// checkAccountLoaded.
+	if err := s.populateAddressToAccountMap(keys); err != nil {
+		return nil, fmt.Errorf("populating address to account map: %w", err)
+	}
+
 	return s, nil
+}
+
+// populateAddressToAccountMap retrieves keys from the keyring and maps their addresses to account names in the signer.
+func (s *Signer) populateAddressToAccountMap(kr keyring.Keyring) error {
+	records, err := kr.List()
+	if err != nil {
+		return fmt.Errorf("retrieving keys from keyring: %w", err)
+	}
+
+	for _, r := range records {
+		addr, err := r.GetAddress()
+		if err != nil {
+			return fmt.Errorf("getting address for record: %w", err)
+		}
+		addrStr, err := s.addressCodec.BytesToString(addr)
+		if err != nil {
+			return fmt.Errorf("converting address to string: %w", err)
+		}
+		s.addressToAccountMap[addrStr] = r.Name
+	}
+	return nil
 }
 
 // CreateTx forms a transaction from the provided messages and signs it.
--- a/pkg/user/tx_client.go
+++ b/pkg/user/tx_client.go
@@ -680,9 +680,16 @@ func (client *TxClient) Account(name string) *Account {
 	return acc.Copy()
 }
 
-func (client *TxClient) AccountByAddress(address sdktypes.AccAddress) *Account {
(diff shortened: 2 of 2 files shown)


chatton:
@rootulp this wasn't in the list for v4 but I think it probably should be, without this the txclient logic is a bit broken for **new** accounts.

rootulp (approved):

evan-forbes (approved):

rootulp:
The last 4 test-race failures are because `TestPrepareProposalCappingNumberOfMessages` times out after 15 minutes. It looks like that test does use a signer [here](https://github.com/celestiaorg/celestia-app/blob/19c13819973ce192a0a3ddab52fcc108063206b7/app/test/prepare_proposal_test.go#L271) so we have two options to resolve:
1. Investigate deeper on why this PR makes it slower to run that test
2. Skip `TestPrepareProposalCappingNumberOfMessages` in race mode by adding it to the skip line [here](https://github.com/celestiaorg/celestia-app/blob/19c13819973ce192a0a3ddab52fcc108063206b7/Makefile#L257)

pkg/user/signer.go
```diff
@@ -146,16 +174,22 @@ func (s *Signer) Account(name string) *Account {
...
 func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
+	accountName := s.accountNameByAddress(address)
+	return s.accounts[accountName]
+}
+
+// accountNameByAddress returns the account name associated with the given address
+func (s *Signer) accountNameByAddress(address sdktypes.AccAddress) string {
 	addrStr, err := s.addressCodec.BytesToString(address)
 	if err != nil {
-		return nil
```
rach-id, line 151: [nit]
shouldn't we log something here instead of failing silently?

pkg/user/tx_client.go
```diff
@@ -680,9 +680,21 @@ func (client *TxClient) Account(name string) *Account {
...
 	client.mtx.Lock()
 	defer client.mtx.Unlock()
+
+	accountName := client.signer.accountNameByAddress(address)
+	if accountName == "" {
+		return nil
+	}
+
+	if err := client.checkAccountLoaded(ctx, accountName); err != nil {
+		return nil
```
rach-id, line 674: same

rach-id (dismissed):

rootulp:
This is going to keep failing test-race until my last comment is addressed.

pkg/user/signer.go
chatton: here I was just following the existing pattern of returning nil on failures like below with the `checkAccountLoaded` check.

chatton: it will end up failing here with the account not found if there is a problem. 

```go
// AccountByAddress returns the account associated with the given address
func (s *Signer) AccountByAddress(address sdktypes.AccAddress) *Account {
	accountName := s.accountNameByAddress(address)
	return s.accounts[accountName]
}
```

I think it might be nice to change the method signature to return an error but maybe that could be a follow up?

chatton:
> This is going to keep failing test-race until my last comment is addressed.

sorry @rootulp I missed this, will add to the ignored list for now if that is okay, I am happy to dig into the race condition failure a bit more deeply but maybe as a follow up

rootulp:
No worries on miss, agreed it makes sense to skip test-race for now, thanks for addressing!

pkg/user/signer.go
rach-id: sounds good to me 👍

evan-forbes (approved):

Issue #4704, opened 2025-04-30
[zk-execution-ism] Set up Hyperlane ZK Execution ISM Module


Manav-Aggarwal:
Create a standalone Cosmos SDK Module which implements a custom Hyperlane ISM which extends the core functionality of the existing module, `x/core` located in https://github.com/bcp-innovations/hyperlane-cosmos. In the custom module, import the core Keeper just like how the bank Keeper is imported.

The module should be in the feature branch called: `feature/zk-execution-ism`

The ISM should implement the following interface:

```
type HyperlaneInterchainSecurityModule interface {
    proto.Message
    ModuleType() uint8
    GetId() (util.HexAddress, error)
    Verify(ctx context.Context, metadata []byte, message util.HyperlaneMessage) (bool, error)
}

```

Module Type: INTERCHAIN_SECURITY_MODULE_TYPE_ZK_EXECUTION 

 ZK ISM Structure

```
// ZkExecutionISM implements a zero-knowledge proof based ISM
type ZkExecutionISM struct {
    Id                        util.HexAddress  `protobuf:"bytes,1,opt,name=id,proto3,customtype=github.com/bcp-innovations/hyperlane-cosmos/util.HexAddress" json:"id,omitempty"`
    Owner                     string           `protobuf:"bytes,2,opt,name=owner,proto3" json:"owner,omitempty"`
    StateRoot                 []byte           `protobuf:"bytes,3,opt,name=state_root,json=stateRoot,proto3" json:"state_root,omitempty"`
    Height                     uint             `protobuf:"varint,4,opt,name=height,proto3" json:"height,omitempty"`
    StateTransitionVerifierKey []byte           `protobuf:"bytes,5,opt,name=state_transition_verifier_key,json=stateTransitionVerifierKey,proto3" json:"state_transition_verifier_key,omitempty"`
    StateMembershipVerifierKey []byte           `protobuf:"bytes,6,opt,name=state_membership_verifier_key,json=stateMembershipVerifierKey,proto3" json:"state_membership_verifier_key,omitempty"`
}

```

Add this note on adding a new ISM as a comment for now:
- The relayer needs to know how to fetch metadata of ZK ISM. 
- The CCIP Read ISM generalizes the query for metadata so if the ZK ISM implements the interface of CCIP Read, then we won’t need to modify the relayer.

cmwaters:
> Module Type: INTERCHAIN_SECURITY_MODULE_TYPE_ZK_EXECUTION

[...]

PR #4878, opened 2025-05-26
feat: add zk execution ism module

damiannolan:
## Overview

Adds the initial module structure for zk execution ism module.

- Hyperlane ISM wiring and interface impls
- Initial MsgServer / QueryServer
- SDK module, genesis and CLIs with app wiring

closes: #4704


damiannolan:
I will extend this test (https://github.com/celestiaorg/celestia-app/blob/main/test/interop/hyperlane_test.go) to validate the wiring of the ism module. But as I understand it, it is wired correctly with the router.

app/modules.go
```diff
@@ -205,5 +209,6 @@ func allStoreKeys() []string {
 		circuittypes.StoreKey,     // added in v4
 		hyperlanetypes.ModuleName, // added in v4
 		warptypes.ModuleName,      // added in v4
+		zkismtypes.StoreKey,
```
damiannolan, line 212: This will be needed as an added store key in a `StoreUpgrades` for the v5 upgrade handler

proto/celestia/zkism/v1/tx.proto
```diff
@@ -0,0 +1,36 @@
...
+message MsgCreateZKExecutionISM {
+  option (cosmos.msg.v1.signer) = "creator";
+
+  // creator is the message sender.
+  string creator = 1;
+  // the zk state transition verifier key of the sp1 prover client
+  bytes state_transition_verifier_key = 2;
+  // the zk state membership verifier key of the sp1 prover client
+  bytes state_membership_verifier_key = 3;
+}
```
damiannolan, line 28: Do we need to provide an initial root of trust here, and associated height?

app/modules.go
damiannolan: https://github.com/celestiaorg/celestia-app/issues/4884

Issue #4847, opened 2025-05-20
Invalid memory address or nil pointer dereference\ncaused by transaction:\n*types.MsgPayForBlobs


pippokr:
### Summary of Bug

Upgrading from v4.0.0-arabica to v.4.0.1-arabica

### Version

v.4.0.1-arabica

### Steps to Reproduce

can be observed after upgrade in arabica

damiannolan:
Looks like the nil pointer is coming from `pubKey` here: https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/x/auth/tx/adapter.go#L63

How was the transaction signed and submitted?

damiannolan:
To avoid panics this func should really error if the sig is nil: https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/x/auth/tx/builder.go#L337

rootulp:
@pippokr is that log unique to v4.0.1-arabica or was it present in v4.0.0-arabica too? The code Damian linked doesn't appear modified in between v4.0.0-arabica and v4.0.1-arabica so I would expect it to impact previous releases. It also looks like it doesn't prevent the consensus node from proceeding so it could a be a non-issue.

damiannolan:
To me it seems potentially like a client side issue. Other than looking a bit scary with a stacktrace in the logs, it doesn't really cause an issue on chain as the recovery middleware catches it.

But imho, the sdk should try its best to prevent you from creating a malformed tx.

rootulp:
- @pippokr did you submit any txs to this Arabica node or do you think the logs were caused by some other user submitting txs to the Arabica node? If we have info on how the tx was created, it will be easier to reproduce the issue.
- Is this a regression in behavior between SDK v0.46.x and v0.50x? 
- If that tx is submitted on celestia-app v3 does it also result in similar panic logs? 
- Any other changes needed in SDK besides:

> To avoid panics this func should really error if the sig is nil: https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/x/auth/tx/builder.go#L337

rootulp:
A few observations:
- The namespace on Arabica that the user was submitting to is `CeroA`


Pretty printed stack trace

```
3:03PM ERR panic recovered in runTx err=
    "recovered: runtime error: invalid memory address or nil pointer dereference
     caused by transaction:
     *types.MsgPayForBlobs{
[...]

PR #4879, opened 2025-05-26
test: sig verification panic on nil pub key

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4847

### DIFF
--- a/docs/release-notes/release-notes.md
+++ b/docs/release-notes/release-notes.md
@@ -35,7 +35,7 @@ grpc_laddr = "tcp://127.0.0.1:9098"
 
 #### IAVL v1 Migration
 
-Celestia-app v4 uses IAVL v1 for better performance. When upgrading from v3, the migration happens lazily over time. If you'd like to avoid the lazy migration, you can perform a fresh state sync so that your node uses IAVL v1 exclusively.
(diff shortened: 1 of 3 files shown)


rootulp:
Now I'm hitting this error: https://github.com/celestiaorg/cosmos-sdk/blob/4217638e2a6413e54604a68f2f215a48b9e1e05e/x/auth/ante/sigverify.go#L285

evan-forbes (approved):

rach-id (approved):
so the plan is to wait and see if the cosmos-sdk's upstream changes this to stop panicking in this case and pull their changes? 

what about syncing? if I understand correctly, when syncing with the new sdk, this panics and it shows in the logs, should we document somewhere that these are expected logs (for node operators not to bombard us with questions about this?)

rootulp:
> so the plan is to wait and see if the cosmos-sdk's upstream changes this to stop panicking in this case and pull their changes?

Yes.

> what about syncing? if I understand correctly, when syncing with the new sdk, this panics and it shows in the logs, should we document somewhere that these are expected logs (for node operators not to bombard us with questions about this?)

I don't think the logs will occur during syncing because a tx with a nil pubkey in the signature data shouldn't be included in any blocks. The log will only appear for new txs that enter a node's mempool. We could still document it somewhere

Issue #4722, opened 2025-05-01
Add Metadata structure for ZK Execution ISM


Manav-Aggarwal:
The ISM metadata is an off-chain metadata provided by a relayer, specific to the ZK Execution ISM.

Implement this Metadata structure and a constructor for it from a byte array.

```
// ZkExecutionISMMetadata contains the ZK proof and verification data
type ZkExecutionISMMetadata struct {
    ProofType   uint8    // Type of ZK Proof System used. Default: Groth16
    Proof       []byte   // The ZK proof bytes
    PublicInputs [][]byte // Public inputs to the proof verification
    MerkleProofs [][]byte // The Merkle Proof bytes
}

// NewZkExecutionISMMetadata parses raw metadata into a structured format
func NewZkExecutionISMMetadata(metadata []byte) (ZkExecutionISMMetadata, error) {
    // Parse the metadata format
    // [0] - Type of the ZK Proof System used
    // [1:5] - Size of the ZK Proof, N, if it exists.
    // [5:N+5] - ZK proof 
    // [N+5:N+9] - Number of public inputs, k
    // [N+9:N+9+k] - Public inputs
    // [N+9+k:] - Merkle Proofs
    
    // Implementation details...
    return ZkExecutionISMMetadata{}, nil
}
```

damiannolan:
Public inputs contains the data required to construct the `PublicWitness`, correct? This is the public outputs from the zk circuit computed off-chain.

We will need to parse these in order to store the state root and the height on the ism.

damiannolan:
closed by #4880

PR #4880, opened 2025-05-26
feat: add zk execution ism structured metadata 

damiannolan:
## Overview

Follows the metadata format as per #4722 

```
// [0] - Type of the ZK Proof System used
// [1:5] - Size of the ZK Proof, N, if it exists.
// [5:N+5] - ZK proof 
// [N+5:N+9] - Number of public inputs, k
// [N+9:N+9+k] - Public inputs
// [N+9+k:] - Merkle Proofs
```

closes: #4722

### DIFF
--- a/x/zkism/types/metadata_test.go
+++ b/x/zkism/types/metadata_test.go
@@ -37,7 +37,7 @@ func TestNewZkExecutionISMMetadata(t *testing.T) {
 		return b
 	}
 
-	tests := []struct {
+	testcases := []struct {
 		name     string
 		metadata []byte
 		expError error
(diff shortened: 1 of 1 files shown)


rootulp (approved):
LGTM. Great unit tests 👍 

Extremely optional idea that we don't have to pursue: we may want to version the metadata format with a leading byte that is the version. The format included in this PR could be version 0 of the format. If we end up needing new fields, then we could bump the version to 1. TBH this seems unlikely and if needed, we can likely make the breaking change before we ship this so I don't think we _need_ to add versioning.

rootulp:
> Great consideration, I think its a great idea. I think we can also support it through the ProofType uint8 at the beginning of the byte slice

Good point, we could add a new proof type like `ProofTypeGroth16V2` so I'm open to skipping versioning.

PR #4881, opened 2025-05-26
fix: prevent sequence mismatch in `TxClient` during evictions

### LINKED ISSUE #4784: account sequence mismatch not handled in `TxClient.SubmitPayForBlob`
### Summary of Bug

I've implemented a minimal transaction client that imports `celestia-app` as a package and uses the [`TxClient.SubmitPayForBlob`](https://github.com/celestiaorg/celestia-app/blob/v3.8.1/pkg/user/tx_client.go#L251) method to locally sign and submit blob transactions.

Each instance of the client runs with a unique keyring, but all communicate with a single centralized celestia-app bridge node.

The implementation generally works well, but occasionally I encounter the following error:

```bash
broadcast tx error: account sequence mismatch: expected 8901, got 8900: incorrect account sequence
```

When this happens, the only reliable fix seems to be restarting the client (looks to be some kind of a cache issue?)

### Version

v3.8.1

### Steps to Reproduce

N/A

yarikbratashchuk:
**Closes #4784**

Fixed sequence mismatch errors that occurred when transactions were evicted from the mempool. The issue arose because the Cosmos SDK's checkState maintained incremented sequence numbers even after mempool eviction, causing temporary desynchronization between client and chain state.

The fix adds transaction commitment verification in `handleEvictions()` to ensure sequence numbers are only rolled back when transactions are truly evicted and not committed. This maintains proper sequence synchronization even during mempool evictions.


pkg/user/tx_client.go
```diff
@@ -569,6 +578,40 @@ func (client *TxClient) handleEvictions(txHash string) error {
...
+		if strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "tx does not exist") {
+			if i < maxRetries-1 {
+				time.Sleep(retryDelay)
+				continue
+			}
+			return false, nil
+		}
+
+		return false, err
+	}
```
rootulp, line 610: The nesting is confusing here. It's more comment to check if `err != nil`. Here's a refactor that makes this easier to follow.

Also I think we can eliminate this whole conditional

```
			if i < maxRetries-1 {
				time.Sleep(retryDelay)
				continue
			}
			return false, nil
```

Here's a refactor

```go
func (client *TxClient) isTransactionCommitted(txHash string) (bool, error) {
	serviceClient := sdktx.NewServiceClient(client.conns[0])

	maxRetries := 10
	retryDelay := 500 * time.Millisecond

	for i := 0; i < maxRetries; i++ {
		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)

		_, err := serviceClient.GetTx(ctx, &sdktx.GetTxRequest{Hash: txHash})
		cancel()

		if err != nil {
			if strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "tx does not exist") {
				time.Sleep(retryDelay)
				continue
			}
			// The tx wasn't found and it isn't a "not found" error.
			return false, err
		}
		// The tx was found, so it was committed
		return true, nil
	}
	return false, nil
}
```

rootulp (commented):
Thanks for the unit test. It would also increase confidence if we could get Conduit to run this and confirm they no longer hit https://github.com/celestiaorg/celestia-app/issues/4784

rootulp:
I'm not re-reviewing this PR because there are comments that haven't been resolved or even responded to yet.

pkg/user/tx_client.go
```diff
@@ -555,11 +555,20 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
 func (client *TxClient) handleEvictions(txHash string) error {
 	client.mtx.Lock()
 	defer client.mtx.Unlock()
+
+	// Before rolling back sequence, verify the transaction wasn't actually committed
+	// This prevents false positive evictions where tx was included in a block
+	if committed, err := client.isTransactionCommitted(txHash); err == nil && committed {
+		delete(client.txTracker, txHash)
```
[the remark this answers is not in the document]
yarikbratashchuk: > Also there is a helper deleteFromTxTracker that does this in a thread safe way.

not needed here, as we're already holding a mutex lock in the handleEviction by itself

> why does this delete the tx from the txTracker?

Transactions should be removed immediately when they reach a terminal state, the 10-minute rule is for automatic cleanup of stale entries of pending txs.

yarikbratashchuk:
@rootulp addressed your comments

pkg/user/tx_client.go
```diff
@@ -566,6 +575,35 @@ func (client *TxClient) handleEvictions(txHash string) error {
...
+	serviceClient := sdktx.NewServiceClient(client.conns[0])
+
+	maxRetries := 10
+	retryDelay := 500 * time.Millisecond
+
+	for i := 0; i < maxRetries; i++ {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+
+		_, err := serviceClient.GetTx(ctx, &sdktx.GetTxRequest{Hash: txHash})
+		cancel()
```
evan-forbes, line 593: per idea from @ninabarbakadze , I think we should just be using TxStatus here, since that's very explicit in that the tx was confirmed, where here seems too implicit to be safe indefinitely

pkg/user/sequence_mismatch_eviction_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/celestiaorg/go-square/v2/share"
+
+	"github.com/celestiaorg/celestia-app/v4/pkg/user"
+	"github.com/celestiaorg/celestia-app/v4/test/util/blobfactory"
+	"github.com/celestiaorg/celestia-app/v4/test/util/random"
+)
+
+// TestSequenceMismatchEviction tests that the TxClient correctly handles evictions without
+// causing sequence number desynchronization. This test verifies the fix for issue #4784.
+func TestSequenceMismatchEviction(t *testing.T) {
```
ninabarbakadze, line 18: is there a way to improve readability here? the loop with nested if else statements becomes a bit hard to read

pkg/user/sequence_mismatch_eviction_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	account := txClient.Account(testAccount)
+	currentSequence := account.Sequence()
+
+	t.Logf("Verifying no sequence mismatch with current sequence %d", currentSequence)
+
+	testBlob := blobfactory.ManyRandBlobs(random.New(), 1000)[0]
+	_, err := txClient.SubmitPayForBlob(ctx, []*share.Blob{testBlob})
+
+	if err != nil {
```
ninabarbakadze, line 82: ditto

pkg/user/tx_client.go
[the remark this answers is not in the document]
ninabarbakadze: deletion still needs to be thread safe

pkg/user/tx_client.go
ninabarbakadze: How is this different from what's happening here? https://github.com/celestiaorg/celestia-app/blob/9629e004407f9b3b5916e09e1a1b6b36189bbd19/pkg/user/tx_client.go#L520

pkg/user/tx_client.go
```diff
@@ -566,6 +575,35 @@ func (client *TxClient) handleEvictions(txHash string) error {
...
+			}
+			// The tx wasn't found and it isn't a "not found" error.
+			return false, err
+		}
+		// The tx was found, so it was committed
+		return true, nil
+	}
+	return false, nil
+}
+
```
ninabarbakadze: I'd like to avoid nested if statements especially in production code if possible

ninabarbakadze (commented):
A few questions and comments re code clarity and readability. Also, would you mind compressing pr descriptions in size? They're usually helpful but most of it seems like issue description which is better suited for GH issues themselves. One small paragraph as a description for any extra information should be sufficient.

pkg/user/tx_client.go
yarikbratashchuk: i hear you

ninabarbakadze: nice

pkg/user/tx_client.go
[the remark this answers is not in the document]
yarikbratashchuk: it is thread safe, we're holding a mutex already

pkg/user/sequence_mismatch_eviction_test.go
yarikbratashchuk: Even with `TxStatus` returning "Committed", there can be timing gaps where the tx is committed but not yet indexed/retrievable by other parts of the system. `GetTx` ensures both commitment AND retrievability, which is what the function name promises.

cmwaters (commented):
So if I understand this correctly, with this change, if a node pronounces a transaction evicted, the client proceeds to check for inclusion in a block for the following 5 seconds (10 queries every 500milliseconds). If it isn't included it then rolls back the sequence number. Is that correct? 

In `ConfirmTx` why not just continue to poll like we do normally (which has a check for committed txs) and ignore the evictions case. In fact, what I was going to propose is that for an eviction, the client should try resubmit the transaction but perhaps we can hold off on that complexity for a separate PR

yarikbratashchuk:
> So if I understand this correctly, with this change, if a node pronounces a transaction evicted, the client proceeds to check for inclusion in a block for the following 5 seconds (10 queries every 500milliseconds). If it isn't included it then rolls back the sequence number. Is that correct?

correct

>In ConfirmTx why not just continue to poll like we do normally (which has a check for committed txs) and ignore the evictions case. In fact, what I was going to propose is that for an eviction, the client should try resubmit the transaction but perhaps we can hold off on that complexity for a separate PR

tried using TxStatus insteadd of GetTx but it doesn't fix the problem (test gets flaky, i answered this same question [above](https://github.com/celestiaorg/celestia-app/pull/4881#discussion_r2132315937))

ninabarbakadze (commented):
assigned myself and I'm taking over this work cc @evan-forbes @cmwaters

ninabarbakadze:
This PR doesn't actually address the root of the issue and even if we were to keep polling the consensus node for confirmation all the subsequent transactions will get rejected in the meantime with nonce errors because we wouldn't have rolled back the nonce in time if the transaction was truly evicted. 

Better path forwards is to address this more holistically and that work can be tracked here https://github.com/celestiaorg/celestia-app/issues/3768#issuecomment-2985152534

PR #4895, opened 2025-05-29
chore(deps): upgrade to v3.10.1-arabica

rootulp:
Upgrade to v3.10.1-arabica which enables node operators to downgrade from v4.x to v3.x _prior_ to the v4 activation height.


rootulp:
Draft PR until the v3.10.1-arabica binary is attached by Goreleaser and then need to retry CI

evan-forbes (approved):

tac0turtle (approved):

Issue #4891, opened 2025-05-28
Return an error if user invokes `celestia-appd init --chain-id` with a public network's chain ID


rootulp:
## Problem

A user tried to create a genesis file for Mocha via

```
celestia-appd init --chain-id mocha-4 test 
```

and they should have instead downloaded the Mocha genesis file

## Proposal

Modify `celestia-appd init` to throw an error if the user provides a chain-id that belongs to a public testnet. Instruct the user to download the genesis file instead of creating a new one

samuraikillers:
Hi @rootulp ,
I would like to work on this issue.

Few doubts I had.
1) How do we verify if the chain-id that user provides belongs to a public testnet or not? Should we create array with list of these public testnet  and verify from there? or is there any other way also?

rootulp:
> Should we use pkg/appconsts/chain_ids.go to verify if the chain-id is among them?

Yes. The public networks are Arabica, Mocha, Mainnet [here](https://github.com/celestiaorg/celestia-app/blob/6c97e9838d7ab96fb7899bd1d24c52a2ad50893d/pkg/appconsts/chain_ids.go#L4-L6).

Note: it should still be possible for a user to create a genesis.json using the chain-id `test`.

evan-forbes:
maybe instead of throwing an error, we could automatically try to download the genesis since that's presumably what the user wants to do.

we just need to notify them that this is occurring in the very slim edge case they are attempting to fork mocha-4 in a weird way.

rootulp:
> maybe instead of throwing an error, we could automatically try to download the genesis since that's presumably what the user wants to do.

Good point, we don't want to error because `celestia-appd init` is used to create all the config files not just the genesis.json. I'll update the issue description.

IMO:
- [sufficient to close this issue] Modify `celestia-appd init <public chain id>` to log a warning saying this is a public network and the generated genesis.json does not match the genesis.json for this network. Please download the genesis.json for the network with `celestia-appd download-genesis ...`
- [nice to have] actually download the genesis.json for that network

PR #4896, opened 2025-05-29
feat: prevent init command from using public network chain-IDs

samuraikillers:
This PR resolves issue #4891


evan-forbes:
imo we should not throw an error as that is potentially confusing to the user. Instead we can use the information that the user is trying to start a node on a public network, thus we should download the genesis for that network automatically 

https://github.com/celestiaorg/celestia-app/issues/4891#issuecomment-2921590680

evan-forbes (changes requested):

rach-id:
@samuraikillers yes

evan-forbes:
> Is this the desired approach?

yeah

evan-forbes:
we can reopen if we want to pursue the different approach, but otherwise closing this PR and documenting the modified approach in the original issue

Issue #4723, opened 2025-05-01
[zk-execution-ism] Add Proof verification Logic


Manav-Aggarwal:
Implement the following methods that'll be called the `verify` method in the ZK Execution ISM struct:

1. `verifyZKProof`: Verify a ZK proof to update the ISM's state root and height
2. `verifyMerkleProofs`: Verify merkle inclusion proofs against the current state root

// verifyMerkleProofs verifies merkle inclusion proofs against the current state root
func (m *ZkExecutionISM) verifyMerkleProofs(metadata ZkExecutionISMMetadata, message util.HyperlaneMessage) (bool, error) {
    // Leverage the existing membership proof verification logic from IBC's verifyMembership
    // Adapt the verification logic to the Hyperlane message format
        
    vk := getMembershipVerifierKey(m.StateMembershipVerifierKey)
    merkleProofs := parseMerkleProofs(metadata.MerkleProofs)
    verified, err := m.verifyMerkleProofs(merkleProofs, vk)
    return verified, err
}

damiannolan:
We should change `merkleProofs [][]byte` in metadata to be `merkleProof []byte` and length prefix it.
For the first iteration we will be using an mpt proof for state membership which will be a proof returned from `eth_getProof()` API.

We should be unmarshalling the proof to this structure: https://github.com/celestiaorg/celestia-zkevm-ibc-demo/blob/main/ibc/lightclients/groth16/client_state.go#L32-L48

damiannolan:
Part one of this issue is implemented and merged. Part two requires a decision on what will be used for proving state membership. 

There is a couple of different designs being discussed:
1. ZK state membership proof using sp1 circuit. This requires a circuit design and benchmark in order to evaluate viability.
2. MPT proof against EVM state root. This requires a custom hyperlane post-dispatch hook in order to store hyperlane message IDs in a eth storage slot and prove against the state root using a proof returned from `eth_getProof`. The post-dispatch hook has been prototyped as part of https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/29
3. Merkle inclusion proof against a subroot. This requires an evaluation of the existing incremental merkle tree structure used by the merkle tree post-dispatch hook available as part of the core hyperlane contracts.

damiannolan:
Closing as this is now superseded by #5804 and #5805. 

[...]

PR #4922, opened 2025-06-04
feat: add gnark proof verification in zk ism module

damiannolan:
## Overview

Builds off the work done by @cmwaters in https://github.com/cmwaters/sp1-groth16-go-verifier/tree/main/verifier.

Part 1 of 
- https://github.com/celestiaorg/celestia-app/issues/4723

When circuits are rewired to support plugging in trusted data then we can take a `proof.bin` file and drop it in `testdata` for a full integration test which would be great!

### DIFF
--- a/x/zkism/types/ism.go
+++ b/x/zkism/types/ism.go
@@ -68,7 +68,7 @@ func (ism *ZKExecutionISM) verifyZKProof(metadata ZkExecutionISMMetadata) (bool,
 		return false, err
 	}
 
-	if err := ism.verifyPublicInputs(metadata.PublicInputs); err != nil {
+	if err := ism.validatePublicInputs(metadata.PublicInputs); err != nil {
 		return false, err
 	}
 
@@ -96,11 +96,11 @@ func (ism *ZKExecutionISM) verifyZKProof(metadata ZkExecutionISMMetadata) (bool,
 	return true, nil
 }
 
-// TODO: verify public inputs with trusted ism/celestia data
+// TODO: validate public inputs with trusted ism/celestia data
 // - trusted state root (DONE)
 // - trusted height (optional?)
 // - celestia header hash(es)
-func (ism *ZKExecutionISM) verifyPublicInputs(inputs PublicInputs) error {
+func (ism *ZKExecutionISM) validatePublicInputs(inputs PublicInputs) error {
 	if !bytes.Equal(inputs.TrustedStateRoot[:], ism.StateRoot) {
 		return fmt.Errorf("cannot trust public inputs trusted state root: expected %x, but got %x", ism.StateRoot, inputs.TrustedStateRoot)
 	}


x/zkism/types/ism.go
```diff
@@ -46,8 +51,51 @@ func (ism *ZKExecutionISM) Verify(ctx context.Context, metadata []byte, message
...
+	if err != nil {
+		return false, err
+	}
+
+	vk, err := groth16.NewVerifyingKey(ism.StateTransitionVerifierKey)
+	if err != nil {
+		return false, err
+	}
+
+	// TODO: We should store the sp1 verifier key on the ism.
```
damiannolan: We need to store both the groth16 verifier key and the sp1 verifier key on the ism, is that correct? cc. @cmwaters

x/zkism/types/ism.go
```diff
@@ -46,8 +51,51 @@ func (ism *ZKExecutionISM) Verify(ctx context.Context, metadata []byte, message
...
+	vkHashBigInt := new(big.Int).SetBytes(vkHash[:])
+
+	// TODO: Find out and validate what exactly are the public inputs, how they should be received and decoded.
+	// The public inputs returned from the evm aggregration sp1 program are currently defined here:
+	// https://github.com/celestiaorg/celestia-zkevm-ibc-demo/blob/main/provers/blevm/common/src/lib.rs#L16
+	//
+	// We must be able to parse the public input bytes in order to update the ism height and state root.
+	// We should be able to just read n bytes from a single public inputs byte slice returned from the prover service.
+	// This can be done as part of metadata parsing.
+	// For now just take index 0 here to satisfy compiler and put code in place.
```
damiannolan: The initial ism metadata structure suggested `publicInputs [][]byte`. 
From my understanding the prover service should return the following [here](https://github.com/celestiaorg/celestia-zkevm-ibc-demo/blob/main/provers/evm-prover/src/main.rs#L243-L246)

```rust
let response = ProveStateTransitionResponse {
    proof: aggregation_output.proof.bytes(),
    public_values: aggregation_output.proof.public_values.to_vec(),
};
```

This will give us a single byte slice of `public_values` which we can then parse/unmarshal in the go code, allowing us to update the height and state root on the ism. cc. @cmwaters

damiannolan (commented):
Left some questions about encoding of public values and keys. 

There is some discrepancies with the existing code within the ibc demo. [This code](https://github.com/celestiaorg/celestia-zkevm-ibc-demo/blob/main/ibc/lightclients/groth16/client_state.go#L321-L330) seems to suggest that the trusted state root must be part of the public inputs?

x/zkism/types/ism.go
cmwaters: Yes, more precisely we should store the bn254 hash of the sp1 verifier key. The relayer or whoever initiates the module should provide the hashed version directly (which would be used as the first public input). We should have it stored in bytes but the actual verification will need to convert it to a big.Int and then a fr.Element

damiannolan: Understood, the conversion code is in place here already. But we will need to update the ism protobuf definition to include the additional field

x/zkism/types/ism.go
cmwaters: Ok there's actually two sets of public values, the public values for the groth16 and SP1. This looks like the public values for the groth16 circuit (there should be only two inputs). The first one is the hash of the SP1 key which should not change and should be stored and provided by the ISM module. The second is the hash of the public inputs of the SP1 circuit. The metadata should actually provide a subset of the public values of the SP1 circuit

damiannolan: Yes. Isn't that subset of values of the SP1 circuit defined in what is included in the aggregate circuit output? (i.e. here: https://github.com/celestiaorg/celestia-zkevm-ibc-demo/blob/main/provers/blevm/blevm-aggregator/src/main.rs#L49-L58).

The SP1 circuit is run N times based on how many blocks we skip/aggregate. Those SP1 circuit proofs and pub inputs are fed into the aggregator circuit (groth16), along with verifier key and verified there. Thus we only need to verify the groth16 circuit on-chain. Is my understanding here correct?

damiannolan:
> In the `Verify` method you will also need to take the mailbox merkle root and pass that to the `verifyMerkleProofs` method.

Yeah, will come later. Want to scope this PR to state transition proofs only

x/zkism/types/ism.go
damiannolan: We still need to provide some part of a trusted anchor (from on chain state) as part of the public inputs/outputs from the program. This is what I was eluding to in https://github.com/celestiaorg/celestia-app/pull/4922#pullrequestreview-2895985032

damiannolan: My take is that the sp1 and groth16 programs inputs and outputs should be adjusted such that we take the trusted state root from the on-chain ism and pass that into the evm block prover program as `trusted_state_root`.
We then compare this against the current block state root in the sp1 program. We can then propagate the new state root and trusted state root in the program outputs. 

This way the groth16 aggregation program can take the `trusted_state_root` from `inputs[0]` and the `new_state_root` from `outputs[n]` and apply them in the groth16 aggregation program outputs. 

We can then parse the public inputs we receive here within the ism verification code, and plug in our trusted state root from on-chain data. We can also include the trusted height as well, if we really need to. 
Does this make sense? 

This would be the first step of diffs we need to make, as well as the changes required to aggregation program outputs (obviously all of that rust code needs to be refactored and cleaned up).

```diff
diff --git a/provers/blevm/blevm-prover/src/prover.rs b/provers/blevm/blevm-prover/src/prover.rs
index de11d9d..12b3dd3 100644
--- a/provers/blevm/blevm-prover/src/prover.rs
+++ b/provers/blevm/blevm-prover/src/prover.rs
@@ -6,6 +6,7 @@ use celestia_types::{
 };
 use celestia_types::{AppVersion, Blob, Commitment};
 use eq_common::KeccakInclusionToDataRootProofInput;
+use ethers::types::H256;
 use rsp_client_executor::io::EthClientExecutorInput;
 use serde::{Deserialize, Serialize};
 use sha3::{Digest, Keccak256};
@@ -49,6 +50,7 @@ pub struct AggregationOutput {
 /// Input data for block proving
 #[derive(Clone)]
 pub struct BlockProverInput {
+    pub trusted_state_root: H256,
     pub inclusion_height: u64,
     pub client_executor_input: Vec<u8>,
     pub rollup_block: Vec<u8>,
@@ -252,6 +254,7 @@ impl BlockProver {
         stdin.write(&header.header.hash());
         stdin.write_vec(data_hash_bytes);
         stdin.write(&data_hash_proof);
+        stdin.write(&input.trusted_state_root);
         Ok(stdin)
     }
 
diff --git a/provers/blevm/blevm/src/main.rs b/provers/blevm/blevm/src/main.rs
index 7dab23e..e323944 100644
--- a/provers/blevm/blevm/src/main.rs
+++ b/provers/blevm/blevm/src/main.rs
@@ -28,6 +28,13 @@ pub fn main() {
     let proof_data_hash_to_celestia_hash: Proof<TmSha2Hasher> = sp1_zkvm::io::read();
     println!("cycle-tracker-end: deserialize input");
 
+    println!("cycle-tracker-start: assert trusted state root");
+    let trusted_state_root = sp1_zkvm::io::read_vec();
+    if client_executor_input.current_block.state_root != trusted_state_root {
+        panic!("current block state root does not match trusted state root");
+    }
+    println!("cycle-tracker-end: assert trusted state root");
+
```

damiannolan: We discussed this morning synchronously that this is the case, and the zk programs should be adjusted. I'll resolve this thread for now and continue with follow up issues. Marking this PR as r4r

x/zkism/types/ism.go
```diff
@@ -46,11 +52,62 @@ func (ism *ZKExecutionISM) Verify(ctx context.Context, metadata []byte, message
...
+// - trusted state root (DONE)
+// - trusted height (optional?)
+// - celestia header hash(es)
+func (ism *ZKExecutionISM) verifyPublicInputs(inputs PublicInputs) error {
+	if !bytes.Equal(inputs.TrustedStateRoot[:], ism.StateRoot) {
+		return fmt.Errorf("cannot trust public inputs trusted state root: expected %x, but got %x", ism.StateRoot, inputs.TrustedStateRoot)
+	}
+
+	return nil
+}
```
damiannolan: We can extend this function later with more assertions on public inputs from the comments.

x/zkism/types/metadata.go
```diff
@@ -97,3 +107,122 @@ func NewZkExecutionISMMetadata(metadata []byte) (ZkExecutionISMMetadata, error)
...
+		return nil, fmt.Errorf("write NewHeight: %w", err)
+	}
+
+	return buf.Bytes(), nil
+}
+
+// Unmarshal decodes a bincode-serialized PublicInputs struct.
+// This function expects the input byte slice to be encoded using Rust bincode's
+// default configuration: (little-endian, fixed-width integers, length-prefixed slices).
+func (pi *PublicInputs) Unmarshal(data []byte) error {
```
damiannolan, line 171: The bincode marshalling and unmarshalling is a bit verbse but such is life. But afaict it needs to be done to check validity of pub inputs and to pass the correct encoding format the proof expects

rootulp (approved):
LGTM, great work!

x/zkism/types/ism.go
```diff
@@ -46,11 +52,62 @@ func (ism *ZKExecutionISM) Verify(ctx context.Context, metadata []byte, message
...
+// - trusted state root (DONE)
+// - trusted height (optional?)
+// - celestia header hash(es)
+func (ism *ZKExecutionISM) validatePublicInputs(inputs PublicInputs) error {
+	if !bytes.Equal(inputs.TrustedStateRoot[:], ism.StateRoot) {
+		return fmt.Errorf("cannot trust public inputs trusted state root: expected %x, but got %x", ism.StateRoot, inputs.TrustedStateRoot)
+	}
+
+	return nil
+}
```
damiannolan, line 109: I'll create an issue for this

Issue #4926, opened 2025-06-05
v4.0.2-mocha error about path to genesis


antony-everstake:
### Summary of Bug

After upgrading from `3.8.1-mocha` to `v4.0.2-mocha `I've had an error about genesis path
```
ERR failed to start api server err="couldn't read GenesisDoc file: open /home/celestia/.celestia-app/config/genesis.json: no such file or directory"
Error: couldn't read GenesisDoc file: open /home/celestia/.celestia-app/config/genesis.json: no such file or directory
2025/06/05 05:53:42 Process finished with error: exit status 1
```
Our home folder for celestia-appd not default. Command for start celestia-appd` celestia-appd --home <$path_to_celestia_folder> start`

Looks like the `v4.0.2-mocha` version of celestia-appd ignore genesis path in config.toml.

### Version

v4.0.2

### Steps to Reproduce

Start celestia-appd using a different path than the default one to home folder.
`celestia-appd --home <$path_to_celestia_folder> start`

mindstyle85:
we have observed one other having this issue and somehow it was resolved when he replaced the genesis file

rootulp:
I couldn't reproduce with the single-node.sh script. 

```
$ mkdir ~/foo
$ ./scripts/single-node.sh ~/foo
celestia-app version: 4.0.2-arabica-9-g984786703
celestia-app home: /Users/rootulp/foo
celestia-app genesis file: /Users/rootulp/foo/config/genesis.json

Do you want to delete existing /Users/rootulp/foo and start a new local testnet? [y/n]
y
Deleting /Users/rootulp/foo...
Initializing validator and node config files...
Adding a new key to the keyring...
Adding genesis account...
Creating a genesis tx...
Collecting genesis txs...
Tracing is set up with the ability to pull traced data from the node on the address http://127.0.0.1:26661
Starting celestia-app...
11:39AM INF P2P Node ID ID=f14ee398f82411f8650bf37701f7d466a755d5ee file=/Users/rootulp/foo/config/node_key.json module=p2p
11:39AM INF Adding persistent peers addrs=[] module=p2p
11:39AM INF Adding unconditional peer ids ids=[] module=p2p
11:39AM INF Add our address to book addr=f14ee398f82411f8650bf37701f7d466a755d5ee@0.0.0.0:26656 book=/Users/rootulp/foo/config/addrbook.json module=p2p
[...]

PR #4933, opened 2025-06-05
fix(multiplexer): preserve home args before `start` command

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4926

Previously the multiplexer removed the first two arguments passed to celestia-appd. This is problematic for Everstake who start celestia-appd like this

```
celestia-appd --home /foo start
```

This PR fixes the issue by filtering the os.Args to remove `celestia-appd` and `start` but still preserve the additional flags regardless of the order they're provided.

### DIFF
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -226,7 +226,7 @@ func (m *Multiplexer) startApp() error {
 	}
 
 	if currentVersion.Appd.Pid() == appd.AppdStopped {
-		programArgs := getProgramArgs(os.Args)
+		programArgs := removeStart(os.Args)
 
 		// start an embedded app.
 		m.logger.Debug("starting embedded app", "app_version", currentVersion.AppVersion, "args", currentVersion.GetStartArgs(programArgs))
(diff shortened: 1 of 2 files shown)


multiplexer/abci/multiplexer.go
```diff
@@ -226,10 +226,7 @@ func (m *Multiplexer) startApp() error {
 	}
 
 	if currentVersion.Appd.Pid() == appd.AppdStopped {
-		programArgs := os.Args
-		if len(os.Args) > 2 {
-			programArgs = os.Args[2:] // remove 'appd start' args
```
rootulp, line 231: this was the problem. This assumed that the first two os.Args were:

```
celestia-appd start
```

but that's not always the case as Everstake pointed out.

julienrbrt (commented):
Nice catch, but we should still skip os.Args[0] as you can rename the binary to whatever. 
Nice catch to path find the first "start" argument instead of assuming it is second.

multiplexer/abci/multiplexer.go
```diff
@@ -226,10 +226,7 @@ func (m *Multiplexer) startApp() error {
...
-		}
+		programArgs := getProgramArgs(os.Args)
```
julienrbrt: I'd rename this to removeStartArg and pass os.Args[1:]

rootulp:
> Nice catch, but we should still skip os.Args[0] as you can rename the binary to whatever.

True. Seems like an edge case that is unlikely to happen but worth addressing so I'll fix shortly

julienrbrt (approved):
lgtm! There always be the edge case that you must not pass a flag that contains "start" but that is very unlikely to happen. 

Someone using `celestia-app start --home start` deserves a bug 🤣

ninabarbakadze (approved):

PR #4937, opened 2025-06-05
docs(README): server architecture

rootulp:
During the adoption onsite, @jcstein and I were confused about all the servers exposed by celestia-app. I promised to write a blurb about it and this is a first attempt at that.

I could see this being improved upon in the future by:
- describe which server gets hit when a user submits a blob
- describe which server needs to be used to connect to a bridge node / light node
- include a diagram

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ make bbr-enable
 
 ### Environment variables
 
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):
nice

PR #4954, opened 2025-06-08
chore: revise codeowners

rootulp:
Currently PRs in celestia-app have reviewers automatically added that aren't actually expected to review the PR. This is because the celestia-core Github team now includes members (Tomasz, Manav, Yarrick) that don't work on celestia-app. 

I'm 100% open to modifying the list of CODEOWNERS but it should reflect the people who are expected to review PRs so that the default reviewers for each PR notifies the correct people. 

Related: I don't think we need the auto request review GH workflow if we list out the CODEOWNERS. I'm basing this on celestia-node b/c they don't use the auto request review workflow:
- https://github.com/celestiaorg/celestia-node/blob/main/.github/CODEOWNERS
- https://github.com/celestiaorg/celestia-node/tree/main/.github


.github/auto_request_review.yml
rach-id, line 1: if this is to be deleted, then you should delete this file also: https://github.com/celestiaorg/celestia-app/blob/main/.github/workflows/pr-review-requester.yml

But thinking about it, relying on the code owners file would request 4-5 reviewers per PR which not ideal IMO. When you see multiple others need to review a PR, you be like: the others can do it. However, using the auto requester, we can only specify 2-3 per PR which makes them targeted to review. And if the PR require another person's perspective, they can add them as reviewers manually.

rootulp: Good catch on deleting the Github workflow.

I don't like the current auto request reviewer configuration because it is requesting ppl to review who don't actually review celestia-app PRs. I'm okay with reviving it if we specify a list of CODEOWNERS and PR auto request reviewers who actually review PRs they are tagged to review on

rach-id (approved):

evan-forbes (approved):

Issue #4950, opened 2025-06-06
`Error: cannot marshal response block_id` while block syncing Mocha


rootulp:
## Context

I'm trying to debug another v4 issue: https://github.com/celestiaorg/celestia-app/issues/4932

## Problem

In one tab, I started block syncing a Mocha node

```
$ ./scripts/block-sync.sh --network mocha
```

In another I tried querying the latest block

```
$ celestia-appd query consensus comet block-latest
[...]

PR #4958, opened 2025-06-09
fix(app): disable `query consensus comet` CLI command

julienrbrt:
Closes: https://github.com/celestiaorg/celestia-app/issues/4950
ref: https://github.com/celestiaorg/celestia-app/issues/4950#issuecomment-2955569603 for the rationale.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -770,7 +770,10 @@ func (app *App) AutoCliOpts() autocli.AppOptions {
 
 	moduleOptions := runtimeservices.ExtractAutoCLIOptions(app.ModuleManager.Modules)
 
-	// skip the comet consensus module autocli commands for v4.0.0
+	// Disable the comet consensus module autocli commands for v4.x.x because
+	// there is a bug in the unmarshalling of blocks.
+	//
+	// https://github.com/celestiaorg/celestia-app/issues/4950
 	consensusModuleOptions := moduleOptions[consensustypes.ModuleName]
 	consensusModuleOptions.Query.SubCommands = nil
 	moduleOptions[consensustypes.ModuleName] = consensusModuleOptions
--- a/app/app.go
+++ b/app/app.go
@@ -768,9 +768,17 @@ func (app *App) AutoCliOpts() autocli.AppOptions {
 		}
 	}
 
+	moduleOptions := runtimeservices.ExtractAutoCLIOptions(app.ModuleManager.Modules)
+
+	// skip the comet consensus module autocli commands for v4.0.0
+	// to be re-added in v4.1.0 when the system is full on v0.50
+	consensusModuleOptions := moduleOptions[consensustypes.ModuleName]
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -768,9 +768,17 @@ func (app *App) AutoCliOpts() autocli.AppOptions {
 		}
 	}
 
+	moduleOptions := runtimeservices.ExtractAutoCLIOptions(app.ModuleManager.Modules)
+
+	// skip the comet consensus module autocli commands for v4.0.0
+	// to be re-added in v4.1.0 when the system is full on v0.50
```
rootulp: I don't understand why it is safe to re-add this in v4.1.0.

The multiplexer needs to be able to handle block syncing historical blocks from genesis. If we re-add this in v4.1.0 what happens when a v4.1.0 multiplexer block syncs through v3.x blocks and a user queries:

```
$ celestia-appd query consensus comet block-latest
```

rootulp (commented):
[question] why is there a submodule for celestia-core in this PR?

app/app.go
julienrbrt: The query works, this is just the query command (cli) that has an issue because of AutoCLI. This command has been created in v0.50, so it is a new command.
You can query via grpc or grpc gateway, as usual. Syncing will work as well.

julienrbrt:
> [question] why is there a submodule for celestia-core in this PR?

My bad. It's gone now.

app/app.go
rootulp: Why is it broken on v4.x.x and why is it safe to add back in v4.1.0?

rootulp (changes requested):

app/app.go
evan-forbes: agree, some clarity here would be helpful :slightly_smiling_face: . I'm missing how this fixes the issue permanently. 

is the issue peristent if both the client and server are running v4 or does it occur then as well? 

if this is a permanent problem, assuming the feature is not important (querying the block from the cli seems like it falls in that category), then I'd be okay returning an error for v3 blocks but ofc would prefer fixing the root issue

rootulp: [blocking] let's remove this line because we shouldn't add the command back in v4.1.x.

We should only add it back when we fix the unmarshaling issue in the CLI which we're not prioritizing.

```suggestion
```

app/app.go
```diff
@@ -768,9 +768,16 @@ func (app *App) AutoCliOpts() autocli.AppOptions {
 		}
 	}
 
+	moduleOptions := runtimeservices.ExtractAutoCLIOptions(app.ModuleManager.Modules)
+
+	// skip the comet consensus module autocli commands for v4.0.0
```
rootulp:
```suggestion
	// Disable the comet consensus module autocli commands for v4.x.x because
	// there is a bug in the unmarshalling of blocks.
	//
	// https://github.com/celestiaorg/celestia-app/issues/4950
```

Issue #4968, opened 2025-06-10
Multiplexer v3.x embedded binary doesn't respect log level


rootulp:
## Problem

Multiplexer embedded v3.x binary doesn't appear to respect the log level specified via CLI or config. I ran this:

```
celestia-appd start --log-to-file celestia-app.log --log_level info
```

and my config.toml looks like this

```toml
# Output level for logging, including package level options
log_level = "info"
```

but in the log file I'm seeing debug logs. Example:

```
3:05AM DBG Read PacketMsg conn=MConn{88.198.70.23:11656} module=p2p packet={"channel_id":32,"eof":true,"data":"OgkI5vaQAxgBICs="} peer=17a1fd70609562ecb7cd10599e0a7d566d8b8699@88.198.70.23:11656
```

This is on the Mocha DevOps validator

julienrbrt:
~~Flags aren't passed down to the multiplexer: https://github.com/celestiaorg/celestia-app/blob/858f9383f11156a4a254a09046668d9586723f75/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go#L41-L52~~
~~This is related to https://github.com/celestiaorg/celestia-app/issues/4932 as well.
I have opened a PR to fix in on v4 however.~~

~~I am still thinking on how to properly pass down flags to the embedded binary, because you cannot actually pass down all the flags, as some may exist in the current but not on the latest embedded  (this is the reason it was hardcoded). This may require a slight redesign.~~

julienrbrt:
Following up on the message above, the flags are correctly passed down via `GetStartArgs` on the multiplexer (directly using os.Args). This means when a flag is added on v4 start that do not exist on an embedded binary this may be a problem, but this isn't a problem today.

A design such as the following may be required later, but today it isn't needed.

```
$ celestia-appd start --force-no-bbr --embedded-args-file ~/downloads/embedded_args.json
```

```json
{
    "v3": [
        "--grpc.address=localhost:50051"
    ],
    "vX": [
        "--grpc.address=localhost:50051"
    ]
}
```

julienrbrt:
It mainly looks like we have two sort of loggers, it is shown by running:

```sh
celestia-appd start --log-to-file celestia-app.log --log_level info --force-no-bbr --log_format json
```

The file will contain a mix of normal (debug) log lines and json log lines

julienrbrt:
[...]

PR #4970, opened 2025-06-10
fix(cmd): pass down log options

julienrbrt:
Partially fixes: https://github.com/celestiaorg/celestia-app/issues/4968
ref: https://github.com/celestiaorg/celestia-app/issues/4968#issuecomment-2958215472

The log options weren't as well passed down when using the `--log-to-file` flag in the app command on v4.
Another PR comes for the fix in v3 as well.

### DIFF
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -85,15 +85,7 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
-			if command.Flags().Changed(FlagLogToFile) {
-				// optionally log to file by replacing the default logger with a file logger
-				err = replaceLogger(command)
-				if err != nil {
-					return err
-				}
-			}
-
-			return nil
+			return replaceLogger(command)
 		},
 		SilenceUsage: true,
 	}
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -184,6 +184,9 @@ func replaceLogger(cmd *cobra.Command) error {
 	}
 
 	sctx := server.GetServerContextFromCmd(cmd)
-	sctx.Logger = log.NewLogger(kitlog.NewSyncWriter(file))
+	sctx.Logger, err = server.CreateSDKLogger(sctx, kitlog.NewSyncWriter(file))
+	if err != nil {
+		return fmt.Errorf("failed to create logger: %w", err)
+	}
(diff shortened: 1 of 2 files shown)


rootulp (commented):
How was this PR tested? Can you include in the PR description manual testing steps if the PR doesn't contain unit tests.

cmd/celestia-appd/cmd/root.go
```diff
@@ -85,15 +85,7 @@ func NewRootCmd() *cobra.Command {
...
-			if command.Flags().Changed(FlagLogToFile) {
-				// optionally log to file by replacing the default logger with a file logger
-				err = replaceLogger(command)
-				if err != nil {
-					return err
-				}
-			}
-
-			return nil
+			return replaceLogger(command)
```
evan-forbes, line 88: why do we want to replace the logger if the user doesn't set the flag?

julienrbrt: this doesn't not replace it, `replaceLogger` had some logic to exit early if the flag wasn't changed. So this just simplify the code instead of doing it twice.

julienrbrt:
> How was this PR tested? Can you include in the PR description manual testing steps if the PR doesn't contain unit tests.

This was tested by running a v4 local node (using the local node script).
Before this PR: log level aren't respected
After this PR: they are.

evan-forbes (approved):
utack

multiplexer/abci/multiplexer.go
```diff
@@ -521,10 +521,7 @@ func (m *Multiplexer) startEmbeddedApp(version Version) error {
 		}
 
 		// start the new app
-		programArgs := os.Args
-		if len(os.Args) > 2 {
-			programArgs = os.Args[2:] // Remove 'appd start' args
-		}
```
rootulp, line 527: [note for other reviewers] this is a drive by fix for a separate issue.

https://github.com/celestiaorg/celestia-app/pull/4933 fixed it in one location. I didn't know there was another location.

rootulp:
> This was tested by running a v4 local node (using the local node script).
> Before this PR: log level aren't respected
> After this PR: they are.

Thanks! Just confirmed by trying different combinations of on `main` and this PR.

```
$ celestia-appd start --force-no-bbr --log-to-file foo.log --log_level debug

$ celestia-appd start --force-no-bbr --log-to-file foo.log --log_level info
```

Odd that the default log level used on `main` was debug when writing to a file but this seems to resolve.

PR #4972, opened 2025-06-10
docs(specs): Fix Decimal type spec link to current location

GarmashAlex:
## Overview

Replaced the outdated reference to the Decimal type specification in types.proto with the correct, up-to-date link:
https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/data_structures.md#decimal
This ensures the documentation comment points to an existing and relevant section in the Celestia specs.


rootulp (approved):
Thanks. We can also prob delete this file

PR #4974, opened 2025-06-10
refactor(multiplexer)!: use file for multiplexer embedded binary args

julienrbrt:
ref: https://github.com/celestiaorg/celestia-app/issues/4968#issuecomment-2959303763

This came up earlier than I thought because of the `--log_no_color` flag.
So doing this improvement already.

```sh
$ celestia-appd start --force-no-bbr --log-to-file=celestia-app.log --embedded-args-file ~/downloads/embedded_args.json 
```

```json
{
    "v3": [
        "--force-no-bbr",
        "--v2-upgrade-height=2585031",
        "--log-to-file=celestia-app.log"
    ]
}
```


julienrbrt:
I am not sure if i like this, but that is a solution for the above.

multiplexer/abci/versions.go
```diff
@@ -76,19 +76,19 @@ func (v Versions) ShouldUseLatestApp(appVersion uint64) bool {
...
 	// Default flags for standalone apps.
-	return append(args,
+	return []string{
 		"--grpc.enable",
 		"--api.enable",
 		"--api.swagger=false",
 		"--with-tendermint=false",
 		"--transport=grpc",
-	)
+	}
```
rootulp: If I understand correctly this default list will never be used because 

```

	versions, err := abci.NewVersions(abci.Version{
		Appd:        appdV3,
		ABCIVersion: abci.ABCIClientVersion1,
		AppVersion:  3,
		StartArgs: append([]string{
			"--grpc.enable",
			"--api.enable",
			"--api.swagger=false",
			"--with-tendermint=false",
			"--transport=grpc",
		}, embeddedArgs.V3...),
	})
```

so IMO we could remove the default list.

rootulp (commented):
Code changes seem fine. This is a breaking change for node operators so I would appreciate if the PR added a section to https://github.com/celestiaorg/celestia-app/blob/main/docs/release-notes/release-notes.md#node-operators-v400 to explain how to provide flags to the v3 embedded binary.

multiplexer/abci/versions.go
julienrbrt: shall we always keep the default list, and make the wiring in new versions simpler?
the multiplexer **needs** those to work, so i feel like we should let the user decide?

rootulp: My suggestion was to remove the default list entirely and retain the list of start arguments the `abci.NewVersions` code snippet. Seems clearer to have `abci.NewVersions` explicitly list all the start arguments and remove the default list.

I'm open to alternative ideas.

julienrbrt: my counter proposal was this: https://github.com/celestiaorg/celestia-app/pull/4974/commits/fa00806f6189570586a61afb47d9bab8b5193b69

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -14,13 +20,6 @@ import (
 	embedding "github.com/celestiaorg/celestia-app/v4/internal/embedding"
 )
 
-// v2UpgradeHeight is the block height at which the v2 upgrade occurred.
```
julienrbrt, line 17: This still needs to be set here eventually, so im thinking of bringing the chainID parsing logic from v3 here instead?

multiplexer/abci/versions.go
rootulp: What happens if the user specifies a config file any of the default flags? For example:

```
{
    "v3": [
        "--grpc.enable=false"
    ]
}
```

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
rootulp: That works. Sounds like we would un-deprecate the `--v2-upgrade-height` flag but instead of a user passing it on command line, they would pass it via the new file.

julienrbrt: yeah, so i totally removed it from v4 to avoid confusion.

multiplexer/abci/versions.go
julienrbrt: Nothing more than what happens today if a user passes `--grpc.enable=false` to the command directly.
I believe flags.Parse / viper parses the latest one.

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
julienrbrt: letting the user pass it manually via the file may be better than having default, especially that v3 already has default based on chain-id 👍🏾

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -33,22 +32,41 @@ func modifyRootCommand(rootCommand *cobra.Command) {
...
 	}
 
-	var extraArgs []string
-	if v2UpgradeHeight != "" {
-		extraArgs = append(extraArgs, "--v2-upgrade-height="+v2UpgradeHeight)
+	var (
+		embeddedArgsFile string
+		embeddedArgs     EmbeddedArgs
+	)
+	// get the home directory from the flag
```
rootulp, line 39: [blocking] This comment is misleading because the code is getting the embedded arg flag not the home flag. Instead of correcting the comment, consider extracting a function with a descriptive name to eliminate the comment entirely.

```go
	args := os.Args
	embeddedArgsFile = getEmbeddedArgsFile(args)
```

and later on

```go
func getEmbeddedArgsFile(args []string) string {
	for i := 0; i < len(args); i++ {
		if args[i] == fmt.Sprintf("--%s", FlagEmbeddedArgsFile) && i+1 < len(args) {
			return filepath.Clean(args[i+1])
		} else if strings.HasPrefix(args[i], fmt.Sprintf("--%s=", FlagEmbeddedArgsFile)) {
			return filepath.Clean(args[i][7:])
		}
	}
	return ""
}
```

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -33,22 +32,41 @@ func modifyRootCommand(rootCommand *cobra.Command) {
...
+	args := os.Args
+	for i := 0; i < len(args); i++ {
+		if args[i] == fmt.Sprintf("--%s", FlagEmbeddedArgsFile) && i+1 < len(args) {
+			embeddedArgsFile = filepath.Clean(args[i+1])
+		} else if strings.HasPrefix(args[i], fmt.Sprintf("--%s=", FlagEmbeddedArgsFile)) {
+			embeddedArgsFile = filepath.Clean(args[i][7:])
```
rootulp, line 45: This code makes the assumption that we'll never change the string inside FlagEmbeddedArgsFile. Can this please be made more robust? Ideally it should not use a hard-coded index like `7`.

```go
		} else if strings.HasPrefix(args[i], fmt.Sprintf("--%s=", FlagEmbeddedArgsFile)) {
			trimmed := strings.TrimPrefix(args[i], fmt.Sprintf("--%s=", FlagEmbeddedArgsFile))
			return filepath.Clean(trimmed)
		}
```


[blocking] needs unit tests

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -33,22 +32,41 @@ func modifyRootCommand(rootCommand *cobra.Command) {
...
+		}
+	}
+
+	if len(embeddedArgsFile) > 0 {
+		f, err := os.ReadFile(embeddedArgsFile)
+		if os.IsNotExist(err) {
+			panic(fmt.Sprintf("file %s does not exist: %s", err.Error()))
```
rootulp, line 52: There are two formatting directives in this string but only one argument.

rootulp (commented):
This needs docs in release-notes because this is a user-facing breaking change to v4.x.x

rootulp:
Closing because stale and merge conflicts.

Issue #4932, opened 2025-06-05
v4.0.2-mocha: grpc call doesn't use RPC define in config


Galadrin:
### Summary of Bug

Upgrade to v4.0.2-mocha broke GRPC query for node not using localhost or inet_any address.

GRPC request are forwarded to RPC without using configured address in config.toml

### Version

commit: 3ebd4cae
cosmos_sdk_version: v1.29.1-sdk-v0.50.12
go: go version go1.23.6 linux/amd64
name: celestia-app
server_name: celestia-appd
version: 4.0.2-mocha

### Steps to Reproduce

set configuration as follow :
```
[rpc]

# TCP or UNIX socket address for the RPC server to listen on
laddr = "tcp://a.b.c.d:26657"

grpc_laddr = "tcp://a.b.c.d:9092"
```
where `a.b.c.d` is not localhost or 0.0.0.0 (can be public IP, VPN address, IPv6...)

query consensus over grpc
```
~/go/bin/celestia-appd query consensus comet block-latest --grpc-addr a.b.c.d:9090 --grpc-insecure
Error: rpc error: code = Unknown desc = post failed: Post "http://localhost:26657": dial tcp 127.0.0.1:26657: connect: connection refused
```

rootulp:
Thanks for reporting.

Investigation notes:
- ~~Requests to the CometBFT GRPC server should route to the CometBFT RPC server. But according to the config file the CometBFT GRPC server only supports one endpoint so I'm not sure if this request `celestia-appd query consensus comet block-latest` actually uses the CometBFT GRPC server.~~ 
- These help docs should specify _which_ GRPC endpoint the command expects. Does it want the CometBFT GRPC server or Cosmos SDK GRPC server? So I improved the docs in https://github.com/celestiaorg/cosmos-sdk/issues/639
  ```
  $ celestia-appd query consensus comet block-latest --help
  Query for the latest committed block
  
  Usage:
    celestia-appd query consensus comet block-latest [flags]
  
  Flags:
        --grpc-addr string         the gRPC endpoint to use for this chain
        --grpc-insecure            allow gRPC over insecure channels, if not the server must use TLS
        --node string              <host>:<port> to CometBFT RPC interface for this chain (default "tcp://localhost:26657")
  ```

- I think I reproed on `main` with
  ```
    celestia-appd start \
      --home "${APP_HOME}" \
      --api.enable \
      --grpc.enable \
      --grpc-web.enable \
[...]

PR #4975, opened 2025-06-11
fix: pass env and sigkill to embedded app

### LINKED ISSUE #4763: Multiplexer doesn't gracefully shut down on CTRL + C

evan-forbes:
## Overview

fixes #4763 
maybe also https://github.com/celestiaorg/celestia-app/issues/4932 

we might need to set the `cmd.Dir` as well

### DIFF
--- a/multiplexer/appd/run.go
+++ b/multiplexer/appd/run.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"syscall"
 )
 
 const (
@@ -65,10 +66,11 @@ func New(version string, compressedBinary []byte) (*Appd, error) {
 func (a *Appd) Start(args ...string) error {
 	cmd := exec.Command(a.path, append([]string{"start"}, args...)...)
 
-	// Set up I/O
 	cmd.Stdin = a.stdin
 	cmd.Stdout = a.stdout
 	cmd.Stderr = a.stderr
+
+	cmd = setupCmd(cmd)
 
 	if err := cmd.Start(); err != nil {
 		return fmt.Errorf("failed to start %s: %w", a.path, err)
@@ -85,6 +87,17 @@ func (a *Appd) Start(args ...string) error {
 	}()
 
 	return nil
+}
+
+func setupCmd(cmd *exec.Cmd) *exec.Cmd {
+	cmd.Env = os.Environ()
+
+	// ensure child and its sub-children form their own process group
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
+
+	return cmd
 }
 
 // Stop terminates the running appd process if it exists.
@@ -125,6 +138,7 @@ func (a *Appd) Pid() int {
 // CreateExecCommand creates an exec.Cmd for the appd binary.
 func (a *Appd) CreateExecCommand(args ...string) *exec.Cmd {
 	cmd := exec.Command(a.path, args...)
+	cmd = setupCmd(cmd)
 	cmd.Stdin = a.stdin
 	cmd.Stdout = a.stdout
 	cmd.Stderr = a.stderr
@@ -230,6 +244,7 @@ func isBinaryDecompressed(version string) bool {
 
 func verifyBinaryIsExecutable(pathToBinary string) error {
 	testCmd := exec.Command(pathToBinary, "--help")
+	testCmd = setupCmd(testCmd)
 	testOutput, err := testCmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("binary validation failed (%s): %w\nOutput: %s", pathToBinary, err, string(testOutput))
--- a/multiplexer/appd/run_test.go
+++ b/multiplexer/appd/run_test.go
@@ -3,10 +3,15 @@
 package appd
 
 import (
+	"bufio"
 	"bytes"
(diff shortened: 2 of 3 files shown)


multiplexer/appd/run.go
```diff
@@ -87,6 +89,17 @@ func (a *Appd) Start(args ...string) error {
...
+func setupCmd(cmd *exec.Cmd) *exec.Cmd {
+	cmd.Env = os.Environ()
+
+	// ensure child and its sub-children form their own process group
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
+
+	return cmd
+}
```
evan-forbes: we might also need to set the `cmd.Dir` but seemed unlikely

evan-forbes:
I haven't manually tested this, so we shouldn't merge until then. This is ready for review tho

multiplexer/appd/run.go
```diff
@@ -87,6 +89,17 @@ func (a *Appd) Start(args ...string) error {
 	return nil
 }
 
+func setupCmd(cmd *exec.Cmd) *exec.Cmd {
+	cmd.Env = os.Environ()
+
+	// ensure child and its sub-children form their own process group
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
```
rach-id: IMO we should look where we're getting the `signal.Notify` and overriding its behavior. This line seems to be making sub-processes have their own process groups, which means `SIGINT` will not be propagated to all sub-processes -> running `ctrl-c` will only kill the parent process. If I understand correctly, we want the opposite of this behavior.

The test is passing because we're sending `-pgid`, which intentionally sends the SIGINT to all process groups related to the parent. But ctrl-c doesn't do that, so I don't think this would work.

An alternative would be to catch the SIGINT signal and re-propagate it using the `-pgid`, but I'm not sure how that would look.

evan-forbes: that makes sense, thanks!

I just and couldn't replicate the issue with either approach while using the embedded app :thinking:

rootulp: The original issue https://github.com/celestiaorg/celestia-app/issues/4763 doesn't reliably repro for me either but given how many times I've started and stopped the multiplexer, it has come up more than a handful of times.

multiplexer/appd/run_test.go
```diff
@@ -104,3 +106,23 @@ func createMockExecutable(t *testing.T, bashCommand string) string {
...
+func TestSetupCmd(t *testing.T) {
+	t.Run("EnvVarPropagation", func(t *testing.T) {
+		key, val := "TEST_ENV_VAR", "42"
+		require.NoError(t, os.Setenv(key, val))
+		defer os.Unsetenv(key)
+
+		// prepare the command: wrap the var in quotes so we know it came through
+		cmd := exec.Command("bash", "-c", fmt.Sprintf("echo -n \"\\\"$%s\\\"\"", key))
+
+		// now invoke your function under test
```
rootulp, line 119:
```suggestion
```

rootulp (approved):
I don't see how this resolves https://github.com/celestiaorg/celestia-app/issues/4932 because I could reproduce that issue without using any environment variables.

Regardless, this does seem like it fixes another issue where environment variables aren't propagated to the v3 embedded app.

rach-id:
> Regardless, this does seem like it fixes another issue where environment variables aren't propagated to the v3 embedded app.

Was this tested using a v3 binary and verified that the env variables reach it? if I understand correctly, all env variables should be inherited by processes by default. I guess we need to dig deeper where the multiplexer is changing `os` related settings to find the root of the issue

rootulp:
Let's hold off on merging until we're confident it fixes the issue in the OP. I'm not sure it does. If it fixes a separate issue then we need a new GH issue for that problem.

PR #4984, opened 2025-06-12
refactor(consensus-breaking): mv circuit breaker to top of ante stack

damiannolan:
## Overview

Moves the circuit breaker decorator to the top of the ante stack.

cc. @evan-forbes


evan-forbes (approved):
technically (however very unlikely afaiu) consensus breaking so we could add a ! in the title

rootulp (approved):

Issue #4981, opened 2025-06-12
Add bullet proof block construction


evan-forbes:
It's possible for users to submit many transactions from the same account. If the last one is smaller than the second to last, there's a chance the second to last will get pruned after filtering in prepareProposal

This results in the proposer included an invalid block.

There could be hacks around this, such as limiting the block size to ~7.5MB when filtering, however the most optimal bullet proof solution would likely involve moving the `square.Build` function to the application.

cmwaters:
> would likely involve moving the `square.Build` function to the application.

Isn't square.Build already in the application?

Issue #4824, opened 2025-05-15
Error: writing square: square size 16 is too small to fit all blobs


evan-forbes:
validator's have observed a bug in mocha that caused a few nodes to panic

> writing square: square size 16 is too small to fit all blobs

This occured during `PrepareProposal`, therefore its possible that this isn't a non-determinism issue.

from a very helpful discord validator:
```
[...]

Issue #4921, opened 2025-06-04
Add multi-version installer script or cli


evan-forbes:
Instead of bumping all larger release branches every time we make a lower major release (ie release v3.10.1 then bump all v4 branches) we could rely on an installer that checks for the latest git tags and either installs from source or downloads the relevant binaries from github. 

the goal would be to avoid having to automate PRs that do this for us or manually doing them ourselves each time similar to:
- https://github.com/celestiaorg/celestia-app/pull/4919
- https://github.com/celestiaorg/celestia-app/pull/4918

cc @adlerjohn for the original idea to using a installer similar to rustup etc

We could also add a flag that makes the process interactive, where the user could select which versions they want to download. There could also be a flag to prune the old binaries.

I could see this being beneficial as we get more and more binaries and some users want to archive sync.

adlerjohn:
Strongly in favor of this approach.

Issue #4721, opened 2025-05-01
Remove `CELESTIA_V3_VERSION` in Makefile


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/9d9a325a9bb3a85fd4e556939fa00acb90ecb551/Makefile#L27

## Problem

Currently we have to manually modify that line, and cut a new v4.x.x release each time we create a new v3.x.x release.

## Proposal

Consider removing the hard-coded v3.x version. Automate the release process so that if a new v3.x.x release is created, the latest v4.x.x release fetches it.

rootulp:
Note: there is also a hard-coded line in docker/multiplexer.Dockerfile.

```
ARG CELESTIA_VERSION="v3.10.0-rc2"
```

chatton:
The `ARG CELESTIA_VERSION="v3.10.0-rc2"` can be overridden in the image building workflow as a docker build argument, I'm not sure what the desired flow should be here. I guess we want to make sure that this is pointing to a non-rc build before we have a final v4 build.

rootulp:
> I'm not sure what the desired flow should be here

Sorry I underspecified this issue. Originally I thought it was a maintenance burden to update the `Makefile` with a new celestia-app v3.x.x release but this approach has some benefits. The major benefit is we can control the roll-out of v3.x.x releases rather than deploying them to all networks immediately (i.e. if the v4.x binary always fetched the latest v3.x.x release)

Given that, we can close this as won't do.

PR #4985, opened 2025-06-12
fix!: check if tx is still valid before including it in a proposal

### LINKED ISSUE #4961: Investigate why Mocha block times spiked to 204 seconds
## Context

Conduit reported that a few blocks on Mocha took much longer than expected block time (~6 seconds)

## Problem

- https://mocha.celenium.io/block/6562897 67 seconds
- https://mocha.celenium.io/block/6562898 204 seconds
- https://mocha.celenium.io/block/6562899 29 seconds
- https://mocha.celenium.io/block/6564028 188 seconds

## Proposal

1. Investigate the issue
2. Respond to Conduit

evan-forbes:
## Overview

fixes #4961 
fixes https://github.com/celestiaorg/celestia-app/issues/4981
maybe also #4824 :crossed_fingers: 

~don't merge until go-square is released and this go.mod is updated~

### DIFF
--- a/app/ante/ante.go
+++ b/app/ante/ante.go
@@ -33,6 +33,8 @@ func NewAnteHandler(
 		// Set up the context with a gas meter.
 		// Must be called before gas consumption occurs in any other decorator.
 		ante.NewSetUpContextDecorator(),
+		// Ensure that the tx does not contain any messages that are disabled by the circuit breaker.
+		circuitante.NewCircuitBreakerDecorator(circuitkeeper),
 		// Ensure the tx does not contain any extension options.
 		ante.NewExtensionOptionsDecorator(nil),
 		// Ensure the tx passes ValidateBasic.
@@ -77,8 +79,6 @@ func NewAnteHandler(
 		ante.NewIncrementSequenceDecorator(accountKeeper),
 		// Ensure that the tx is not an IBC packet or update message that has already been processed.
 		ibcante.NewRedundantRelayDecorator(channelKeeper),
-		// Ensure that the tx does not contain any messages that are disabled by the circuit breaker.
-		circuitante.NewCircuitBreakerDecorator(circuitkeeper),
 	)
 }
 
--- a/cmd/celestia-appd/cmd/root.go
+++ b/cmd/celestia-appd/cmd/root.go
@@ -85,15 +85,7 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
-			if command.Flags().Changed(FlagLogToFile) {
-				// optionally log to file by replacing the default logger with a file logger
-				err = replaceLogger(command)
-				if err != nil {
-					return err
-				}
-			}
-
-			return nil
+			return replaceLogger(command)
 		},
 		SilenceUsage: true,
 	}
@@ -184,6 +176,9 @@ func replaceLogger(cmd *cobra.Command) error {
 	}
 
 	sctx := server.GetServerContextFromCmd(cmd)
-	sctx.Logger = log.NewLogger(kitlog.NewSyncWriter(file))
+	sctx.Logger, err = server.CreateSDKLogger(sctx, kitlog.NewSyncWriter(file))
+	if err != nil {
+		return fmt.Errorf("failed to create logger: %w", err)
+	}
 	return server.SetCmdServerContext(cmd, sctx)
 }
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -521,10 +521,7 @@ func (m *Multiplexer) startEmbeddedApp(version Version) error {
 		}
 
 		// start the new app
-		programArgs := os.Args
-		if len(os.Args) > 2 {
-			programArgs = os.Args[2:] // Remove 'appd start' args
-		}
+		programArgs := removeStart(os.Args)
 
 		m.logger.Info("Starting app for version", "app_version", version.AppVersion, "args", version.GetStartArgs(programArgs))
 		if err := version.Appd.Start(version.GetStartArgs(programArgs)...); err != nil {


.goreleaser.yaml
```diff
@@ -9,10 +9,10 @@ before:
...
     # these hooks get run once. We want to explicitly download every embedded binary so we can build the multiplexer
     # for every arch.
-    - ./scripts/download_v3_binary.sh celestia-app_Darwin_arm64.tar.gz celestia-app_darwin_v3_arm64.tar.gz v3.10.1-arabica
-    - ./scripts/download_v3_binary.sh celestia-app_Linux_arm64.tar.gz celestia-app_linux_v3_arm64.tar.gz v3.10.1-arabica
-    - ./scripts/download_v3_binary.sh celestia-app_Darwin_x86_64.tar.gz celestia-app_darwin_v3_amd64.tar.gz v3.10.1-arabica
-    - ./scripts/download_v3_binary.sh celestia-app_Linux_x86_64.tar.gz celestia-app_linux_v3_amd64.tar.gz v3.10.1-arabica
+    - ./scripts/download_v3_binary.sh celestia-app_Darwin_arm64.tar.gz celestia-app_darwin_v3_arm64.tar.gz v3.10.1-mocha
+    - ./scripts/download_v3_binary.sh celestia-app_Linux_arm64.tar.gz celestia-app_linux_v3_arm64.tar.gz v3.10.1-mocha
+    - ./scripts/download_v3_binary.sh celestia-app_Darwin_x86_64.tar.gz celestia-app_darwin_v3_amd64.tar.gz v3.10.1-mocha
+    - ./scripts/download_v3_binary.sh celestia-app_Linux_x86_64.tar.gz celestia-app_linux_v3_amd64.tar.gz v3.10.1-mocha
```
evan-forbes: I originally forked v4.0.2-release, which has this change and main doesn't, however we should bump it on main anyway

app/validate_txs.go
```diff
@@ -31,46 +63,40 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []*tx.BlobTx) {
...
-// transactions that return an error.
-//
-// Side-effect: arranges all normal transactions before all blob transactions.
-func FilterTxs(logger log.Logger, ctx sdk.Context, handler sdk.AnteHandler, txConfig client.TxConfig, txs [][]byte) [][]byte {
-	normalTxs, blobTxs := separateTxs(txConfig, txs)
-	normalTxs, ctx = filterStdTxs(logger, txConfig.TxDecoder(), ctx, handler, normalTxs)
-	blobTxs, _ = filterBlobTxs(logger, txConfig.TxDecoder(), ctx, handler, blobTxs)
-	return append(normalTxs, encodeBlobTxs(blobTxs)...)
-}
+func (fb *FilteredBuilder) Fill(txs [][]byte) [][]byte {
```
evan-forbes: the logic here is almost identical to the two functions we had before

however now it includes adding and reverting txs to the square.Builder

internal/embedding/data.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // NOTE: This version must be updated at the same time as the version in the
 // Makefile.
-const v3Version = "v3.10.1-arabica"
+const v3Version = "v3.10.1-mocha"
```
cmwaters: Is there some way to better track this. I feel like this is something we're going to forget unless there's some automation

evan-forbes: #4921

test/docker-e2e/go.sum
```diff
@@ -350,8 +350,8 @@ github.com/celestiaorg/cosmos-sdk v1.29.2-sdk-v0.50.12 h1:Q5Y9mZymbQP9qUr/l5IPcg
 github.com/celestiaorg/cosmos-sdk v1.29.2-sdk-v0.50.12/go.mod h1:clLEg7hWK0iJfiO1Vp71tcQ0vGbFzVsEcLxxnRobIZM=
 github.com/celestiaorg/cosmos-sdk/x/upgrade v0.1.0 h1:+i3G5mP/kPgFEn83EEXGly29QDin2Gvdt0kgpmw/vTg=
 github.com/celestiaorg/cosmos-sdk/x/upgrade v0.1.0/go.mod h1:T4K9O18zQNKNpt4YvTL3lcUt4aKOEU05ZIFWVdQi3Ak=
-github.com/celestiaorg/go-square/v2 v2.2.0 h1:zJnUxCYc65S8FgUfVpyG/osDcsnjzo/JSXw/Uwn8zp4=
-github.com/celestiaorg/go-square/v2 v2.2.0/go.mod h1:j8kQUqJLYtcvCQMQV6QjEhUdaF7rBTXF74g8LbkR0Co=
+github.com/celestiaorg/go-square/v2 v2.2.1-0.20250612221901-9fc8d4790a54 h1:yyBbRo3LAiu0Pl2ffQmTyXkQLGYvB2UsZwkOHzXDRLA=
+github.com/celestiaorg/go-square/v2 v2.2.1-0.20250612221901-9fc8d4790a54/go.mod h1:6M2txj0j6dkoE+cgwyG0EqrEPhbZpM2R1lsWEopMIBc=
```
evan-forbes: note to self to revert before merging

test/util/malicious/out_of_order_prepare.go
```diff
@@ -46,11 +46,23 @@ func (a *App) OutOfOrderPrepareProposal(req *abci.RequestPrepareProposal) (*abci
...
+	if err != nil {
+		panic(err)
+	}
+
+	txs := fb.Fill(req.Txs)
 
 	// build the square from the set of valid and prioritised transactions.
 	// The txs returned are the ones used in the square and block
-	dataSquare, txs, err := Build(txs, appconsts.LatestVersion, a.MaxEffectiveSquareSize(sdkCtx), OutOfOrderExport)
+	dataSquare, err := fb.Build()
```
evan-forbes: really hope we can just delete this package as maintaining two copies of things is cursed #4990

evan-forbes:
There's a weird bug in the `malicious` package that I'm waiting to debug as I think we can just delete the whole package, which would be :fire: . The rest are passing though so I'm marking this ready for review. ref #4990 

blocked on merging until we get a go-square release.

app/validate_txs.go
```diff
@@ -8,11 +8,44 @@ import (
...
 
+	square "github.com/celestiaorg/go-square/v2"
 	"github.com/celestiaorg/go-square/v2/tx"
 
 	"github.com/celestiaorg/celestia-app/v4/pkg/appconsts"
 )
 
+// FilteredBuilder filters txs and blobs using a copy of the state and tx validity
+// rules before adding it the square.
+type FilteredBuilder struct {
```
damiannolan: maybe we could call this `FilteredSquareBuilder` to be more specific about what the type is used for, I think it would help the reader

I would also suggest moving the code in this file (`validate_txs.go` -> `square.go` or something similar). I think `square_size.go` exists in this pkg with a single small func which could maybe be moved and live next to all things square 🟦 (although, I think its a method on the app)

app/validate_txs.go
```diff
@@ -8,11 +8,44 @@ import (
...
+	logger   log.Logger
+	ctx      sdk.Context
```
damiannolan: I think we could remove these as fields on the struct and pass `ctx` to `Fill(ctx, req.Txs)`, and any other funcs as necessary

Also could derive a logger from ctx on the fly using something like:
```
logger := ctx.Logger().With("app/filtered-square-builder")
```

app/test/prepare_proposal_test.go
```diff
@@ -30,6 +33,74 @@ import (
...
+
+		prepareResponse, err := testApp.PrepareProposal(&abci.RequestPrepareProposal{
+			Txs:    txs,
+			Height: height,
+			Time:   time.Now(),
+		})
+		require.NoError(t, err)
+		require.Equal(t, 7, len(prepareResponse.Txs))
+
+		fmt.Println("using height", height)
```
damiannolan: using `fmt` might flood test output unncessarily

```suggestion
		t.Logf("using height %d\n", height)
```

damiannolan (commented):
lgtm! I just left some nitty comments about code org and structuring which I think might help non-familiar devs

internal/embedding/data.go
rootulp: Related: https://github.com/celestiaorg/celestia-app/issues/4721

app/test/prepare_proposal_test.go
rootulp: Also this log doesn't seem helpful after the test was written

```suggestion
```

app/test/prepare_proposal_test.go
```diff
@@ -30,6 +33,74 @@ import (
 	blobtypes "github.com/celestiaorg/celestia-app/v4/x/blob/types"
 )
 
+func TestPrepareProposalValidConstruction(t *testing.T) {
+	// Reproduces https://github.com/celestiaorg/celestia-app/issues/4961
+	t.Run("prepare proposal creates a proposal that process proposal throws an account sequence mismatch", func(t *testing.T) {
```
rootulp: [nit] now that the test passes 🎊 

```suggestion
	t.Run("prepare proposal creates a proposal that process proposal accepts", func(t *testing.T) {
```

app/test/prepare_proposal_test.go
```diff
@@ -30,6 +33,74 @@ import (
...
+		txs := createBlobTxs(t, testApp, encConf, kr, accounts)
+		require.Equal(t, 9, len(txs))
+
+		prepareResponse, err := testApp.PrepareProposal(&abci.RequestPrepareProposal{
+			Txs:    txs,
+			Height: height,
+			Time:   time.Now(),
+		})
+		require.NoError(t, err)
+		require.Equal(t, 7, len(prepareResponse.Txs))
```
rootulp: [optional]

```suggestion
		require.NoError(t, err)
		// The filtered builder in prepare proposal should have dropped the last two txs.
		// The filtered builder should have dropped the second to last tx because it was too large to fit in the square.
		// The filtered builder should have dropped the last tx because the nonce for it was invalidated by dropping the second to last tx.
		require.Equal(t, 7, len(prepareResponse.Txs))
```

app/validate_txs.go
```diff
@@ -8,11 +8,44 @@ import (
...
+
+func NewFilteredBuilder(
+	logger log.Logger,
+	ctx sdk.Context,
+	handler sdk.AnteHandler,
+	txConfig client.TxConfig,
+	maxSquareSize,
+	subtreeRootThreshold int,
+) (*FilteredBuilder, error) {
+	builder, err := square.NewBuilder(maxSquareSize, subtreeRootThreshold)
```
rootulp, line 30: [nit] it can be tricky to debug functions that return a non nil value AND an error. IMO for new code, we should return value or error but never both.

```suggestion
	builder, err := square.NewBuilder(maxSquareSize, subtreeRootThreshold)
	if err != nil {
		return nil, err
	}
```

app/validate_txs.go
```diff
@@ -31,101 +64,106 @@ func separateTxs(_ client.TxConfig, rawTxs [][]byte) ([][]byte, []*tx.BlobTx) {
...
 		// Set the tx size on the context before calling the AnteHandler
-		ctx = ctx.WithTxBytes(tx)
+		fb.ctx = fb.ctx.WithTxBytes(tx)
 
 		msgTypes := msgTypes(sdkTx)
 		if nonPFBMessageCount+len(sdkTx.GetMsgs()) > appconsts.MaxNonPFBMessages {
-			logger.Debug("skipping tx because the max non PFB message count was reached", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()))
+			fb.logger.Debug("skipping tx because the max non PFB message count was reached", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()))
 			continue
 		}
```
rootulp: [blocking][question] the builder already appended the tx on line 85 but here we're skipping running the handler on it and we haven't reverted it from the builder. I would expect this to call

```go
		if nonPFBMessageCount+len(sdkTx.GetMsgs()) > appconsts.MaxNonPFBMessages {
			fb.logger.Debug("skipping tx because the max non PFB message count was reached", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()))
			err = fb.builder.RevertLastTx()
			if err != nil {
				fb.logger.Error("reverting last transaction", "error", err)
			}
			continue
		}
```

rootulp: Alternatively this bug isn't present in the blob txs section below because appending happened after this message count check so we could move it down like so:

```go
		msgTypes := msgTypes(sdkTx)
		if nonPFBMessageCount+len(sdkTx.GetMsgs()) > appconsts.MaxNonPFBMessages {
			logger.Debug("skipping tx because the max non PFB message count was reached", "tx", tmbytes.HexBytes(coretypes.Tx(tx).Hash()))
			continue
		}


		if !fb.builder.AppendTx(tx) {
			continue
		}
```

app/validate_txs.go
rootulp: d2f5be16e8733e6b10dd6deef3b8c134e8138b04

app/validate_txs.go
rootulp: Renamed the struct to `FilteredSquareBuilder` and the file to `filtered_square_builder.go`.

In a separate follow-up PR we could move the one method in square.go to app.go because IMO methods on structs should be defined in the same file as the struct even though `app.go` is already enormous.

app/filtered_square_builder.go
```diff
@@ -8,124 +8,147 @@ import (
...
+			fsb.logger.Error(
 				"filtering already checked blob transaction", "tx", tmbytes.HexBytes(coretypes.Tx(tx.Tx).Hash()), "error", err,
 			)
 			telemetry.IncrCounter(1, "prepare_proposal", "invalid_blob_txs")
+			err = fsb.builder.RevertLastBlobTx()
+			if err != nil {
+				fsb.logger.Error("reverting last blob transaction failed", "error", err)
+			}
 			continue
 		}
```
rootulp: [blocking] there is a bug here where we don't decrement `pfbMessageCount` even though we're reverting the last blobTx

test/util/malicious/out_of_order_prepare.go
```diff
@@ -46,11 +46,21 @@ func (a *App) OutOfOrderPrepareProposal(req *abci.RequestPrepareProposal) (*abci
...
+	if err != nil {
+		panic(err)
+	}
+
+	txs := fb.Fill(sdkCtx, req.Txs)
 
 	// build the square from the set of valid and prioritised transactions.
 	// The txs returned are the ones used in the square and block
-	dataSquare, txs, err := Build(txs, appconsts.LatestVersion, a.MaxEffectiveSquareSize(sdkCtx), OutOfOrderExport)
+	dataSquare, err := fb.Build()
```
rootulp: [blocking] this no longer returns the txs that are used so the txs that are returned in `ResponsePrepareProposal` are incorrect.

rootulp: Disregard. `fb.Build()` doesn't prune any txs so the txs from above are still valid.

app/validate_txs.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/4985/commits/2ed3b9670130569d2e15cd00669de15f4280cd97

rootulp:
I think we can drop the `!` from the PR title b/c doesn't seem consensus breaking.

evan-forbes:
okay, just fixed the malicious test in https://github.com/celestiaorg/celestia-app/pull/4985/commits/d8723ad771cf70919316611a44234a06402ff57f :facepalm: so we should be good to merge!

> I think we can drop the ! from the PR title b/c doesn't seem consensus breaking.

per sync, I think this is technically (but not meaningfully) api breaking since we deleted `FilterTxs`. imo, it doesn't make sense to go through the normal deprecation process since it is only used by this repo

rootulp (approved):

rootulp:
> However, I am missing where the transactions get rejected if they're invalid? I'm referring to the case where two transactions are from the same account, if the former gets rejected, and the latter has a higher nonce, the latter should also be rejected. Where is that happening?

It is happening inside the ante handler (`handler`)

Issue #4982, opened 2025-06-12
PubKey is included in signature only if the sequence number is 0


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/0a7154538e6b47ca806e9f4b3fd9ff1581213099/pkg/user/signer.go#L299-L301

## Question

Why only include the pubkey in the signature if the sequence is 0? This is causing an error in one of my in-progress unit tests and I'd like the signature to always contain the pubkey regardless of the sequence number.

cmwaters:
The first transaction by an address (i.e. sequence at 0) must include the public key so the state machine is able to verify the signature of the transaction. The public key can be included in later transactions but it isn't needed because the state machine has already persisted the public key

rootulp:
Related: https://github.com/cosmos/cosmos-sdk/issues/24849

There must be some code path that I failed to inspect because I observed that a signature without a public key on SDK v0.46.x would fail the sig verification decorator and panic on SDK v0.50.x.

rootulp:
Note this bug is only present on the signer in v3. On v4, the signature always contains the PubKey. See
- https://github.com/celestiaorg/celestia-app/blob/fff1649853640894f4b57ee176172aeb033ec302/pkg/user/signer.go#L296
- https://github.com/celestiaorg/celestia-app/blob/fff1649853640894f4b57ee176172aeb033ec302/pkg/user/signer.go#L313

rootulp:
Resolved on v3.x by merging https://github.com/celestiaorg/celestia-app/pull/4987

PR #4987, opened 2025-06-12
fix(signer): always include pubkey in signature

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4982

### DIFF
--- a/pkg/user/signer.go
+++ b/pkg/user/signer.go
@@ -295,9 +295,7 @@ func (s *Signer) getSignatureV2(sequence uint64, pubKey cryptotypes.PubKey, sign
 			Signature: signature,
 		},
 		Sequence: sequence,
-	}
-	if sequence == 0 {
-		sigV2.PubKey = pubKey
+		PubKey:   pubKey,
 	}
 	return sigV2
 }
--- a/pkg/user/signer_test.go
+++ b/pkg/user/signer_test.go
@@ -1,0 +1,62 @@
+package user_test
+
+import (
+	"testing"
+
+	"github.com/celestiaorg/celestia-app/v3/app"
+	"github.com/celestiaorg/celestia-app/v3/app/encoding"
+	"github.com/celestiaorg/celestia-app/v3/pkg/appconsts"
+	"github.com/celestiaorg/celestia-app/v3/pkg/user"
+	testutil "github.com/celestiaorg/celestia-app/v3/test/util"
+	"github.com/celestiaorg/celestia-app/v3/test/util/testfactory"
+	"github.com/celestiaorg/celestia-app/v3/test/util/testnode"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSigner(t *testing.T) {
+	account := "test"
+	testApp, kr := testutil.SetupTestAppWithGenesisValSetAndMaxSquareSize(app.DefaultConsensusParams(), 128, account)
+	addr := testfactory.GetAddress(kr, account)
+	enc := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	acc := testutil.DirectQueryAccount(testApp, addr)
+
+	signer, err := user.NewSigner(kr, enc.TxConfig, testutil.ChainID, appconsts.LatestVersion, user.NewAccount(account, acc.GetAccountNumber(), acc.GetSequence()))
+	require.NoError(t, err)
+	msg := banktypes.NewMsgSend(
+		addr,
+		testnode.RandomAddress().(sdk.AccAddress),
+		sdk.NewCoins(sdk.NewInt64Coin(appconsts.BondDenom, 10)),
+	)
+
+	t.Run("should create a tx with a signature that contains a pubkey", func(t *testing.T) {
+		tx, err := signer.CreateTx([]sdk.Msg{msg})
+		require.NoError(t, err)
+		require.NotNil(t, tx)
+
+		decodedTx, err := signer.DecodeTx(tx)
+		require.NoError(t, err)
+
+		sigs, err := decodedTx.GetSignaturesV2()
+		require.NoError(t, err)
+		require.NotNil(t, sigs[0].PubKey)
+		require.Equal(t, uint64(0), sigs[0].Sequence)
+
+	})
+	t.Run("should create a tx with a signature that contains a pubkey even if sequence is > 0", func(t *testing.T) {
+		signer.IncrementSequence(account)
+
+		tx, err := signer.CreateTx([]sdk.Msg{msg})
+		require.NoError(t, err)
+		require.NotNil(t, tx)
+
+		decodedTx, err := signer.DecodeTx(tx)
+		require.NoError(t, err)
+
+		sigs, err := decodedTx.GetSignaturesV2()
+		require.NoError(t, err)
+		require.NotNil(t, sigs[0].PubKey)
+		require.Equal(t, uint64(1), sigs[0].Sequence)
+	})
+}


rootulp:
I don't understand why CI is failing

rootulp:
Fixed two CI checks in separate PRs:
- https://github.com/celestiaorg/celestia-app/pull/4994
- https://github.com/celestiaorg/celestia-app/pull/4995

evan-forbes (commented):
imo we should remove it from v4 :sweat_smile: 

its not a ton of wasted block space (or more important to the user gas usage), but it still is wasted when its already stored in the state.

rootulp:
> imo we should remove it from v4 😅

We can't. Cosmos SDK v0.50.x now has a codepath that `panic`s if the PubKey isn't included in the signature. See:
- https://github.com/celestiaorg/celestia-app/issues/4847
- https://github.com/cosmos/cosmos-sdk/issues/24849

evan-forbes:
posted in the other thread https://github.com/cosmos/cosmos-sdk/issues/24849#issuecomment-2977145228


the state machine definitely allows for us to not include the public key, that might be a premature optimization, especially for pfbs, so I'm not opposed to merging based on that alone. However since the state machine does support not having a public key, I want to make sure everyone is on the same page as to what the sdk's abilities actually are and the problems we're trying to fix. 

If its true that we only need the public key in the first tx, then this feels like a hack rather than fixing what is actually the problem is

rootulp:
> the state machine definitely allows for us to not include the public key

I'd like to see a unit test that verifies this claim on v4.x.

evan-forbes (approved):
disappointing that this slipped into the sdk when the entire rest of the x/auth statemachine was designed not to force the inclusion of the public key in every tx

might as well remove the public key from the state to get some benefit out of this

evan-forbes:
fwiw, we technically don't need this on v3, however I like the idea of being consistent and merging this

this test passes on v3 but fails on v4 (with a modified txclient to not include the pubkey)

<details>
  <summary>test</summary>

```
package app_test

import (
	"sync"
	"testing"

	"cosmossdk.io/math"
	sdktx "github.com/cosmos/cosmos-sdk/types/tx"
	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/suite"
	abci "github.com/tendermint/tendermint/abci/types"
	cmtrand "github.com/tendermint/tendermint/libs/rand"

	"github.com/celestiaorg/celestia-app/v3/app"
	"github.com/celestiaorg/celestia-app/v3/app/encoding"
	"github.com/celestiaorg/celestia-app/v3/pkg/user"
	"github.com/celestiaorg/celestia-app/v3/test/util/blobfactory"
	"github.com/celestiaorg/celestia-app/v3/test/util/testfactory"
	"github.com/celestiaorg/celestia-app/v3/test/util/testnode"
)

func TestDebugPubTest(t *testing.T) {
	if testing.Short() {
		t.Skip("skipping SDK integration test in short mode.")
	}
	suite.Run(t, new(DebugPubTest))
}

type DebugPubTest struct {
	suite.Suite

	accounts []string
	cfg      *testnode.Config
	cctx     testnode.Context
	ecfg     encoding.Config

	mut            sync.Mutex
	accountCounter int
}

func (s *DebugPubTest) SetupSuite() {
	t := s.T()
	t.Log("setting up integration test suite")

	accounts := make([]string, 35)
	for i := 0; i < len(accounts); i++ {
		accounts[i] = cmtrand.Str(9)
	}

	s.cfg = testnode.DefaultConfig().WithFundedAccounts(accounts...)
	s.cctx, _, _ = testnode.NewNetwork(t, s.cfg)
	s.accounts = accounts
}

func (s *DebugPubTest) getValidatorName() string {
	return s.cfg.Genesis.Validators()[0].Name
}

func (s *DebugPubTest) getValidatorAccount() sdk.ValAddress {
	record, err := s.cfg.Genesis.Keyring().Key(s.getValidatorName())
	s.Require().NoError(err)
	address, err := record.GetAddress()
	s.Require().NoError(err)
	return sdk.ValAddress(address)
}

func (s *DebugPubTest) TestDebugPubKey() {
	t := s.T()
	type test struct {
		name         string
		msgFunc      func() (msgs []sdk.Msg, signer string)
		expectedCode uint32
	}
	tests := []test{
		{
			name: "send 1 utia",
				account1, account2 := s.accounts[0], s.accounts[1]
				msgSend := banktypes.NewMsgSend(
					testfactory.GetAddress(s.cctx.Keyring, account1),
					testfactory.GetAddress(s.cctx.Keyring, account2),
					sdk.NewCoins(sdk.NewCoin(app.BondDenom, math.NewInt(1))),
				)
				return []sdk.Msg{msgSend}, account1
			},
			expectedCode: abci.CodeTypeOK,
		},
		{
			name: "send 1 utia again, this time w/o the pubkey",
				account1, account2 := s.accounts[0], s.accounts[1]
				msgSend := banktypes.NewMsgSend(
					testfactory.GetAddress(s.cctx.Keyring, account1),
					testfactory.GetAddress(s.cctx.Keyring, account2),
					sdk.NewCoins(sdk.NewCoin(app.BondDenom, math.NewInt(1))),
				)
				return []sdk.Msg{msgSend}, account1
			},
			expectedCode: abci.CodeTypeOK,
		},
	}

	// sign and submit the transactions
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			serviceClient := sdktx.NewServiceClient(s.cctx.GRPCClient)
			msgs, signer := tt.msgFunc()
			txClient, err := user.SetupTxClient(s.cctx.GoContext(), s.cctx.Keyring, s.cctx.GRPCClient, s.ecfg, user.WithDefaultAccount(signer))
			require.NoError(t, err)
			res, err := txClient.SubmitTx(s.cctx.GoContext(), msgs, blobfactory.DefaultTxOpts()...)
			if tt.expectedCode != abci.CodeTypeOK {
				require.Error(t, err)
				require.Nil(t, res)

				switch txError := err.(type) {
				case *user.ExecutionError:
					txResp, err := serviceClient.GetTx(s.cctx.GoContext(), &sdktx.GetTxRequest{Hash: txError.TxHash})
					require.NoError(t, err)
					assert.Equal(t, tt.expectedCode, txError.Code, txResp.TxResponse.RawLog)
				case *user.BroadcastTxError:
					assert.Equal(t, tt.expectedCode, txError.Code, txError.ErrorLog)
				}
			} else {
				require.NoError(t, err)
				require.NotNil(t, res)
				getTxResp, err := serviceClient.GetTx(s.cctx.GoContext(), &sdktx.GetTxRequest{Hash: res.TxHash})
				require.NoError(t, err)
				assert.Equal(t, tt.expectedCode, res.Code, getTxResp.TxResponse.RawLog)
			}
		})
	}
}
```
</details>

rootulp:
Thanks for diving deep and following up on https://github.com/cosmos/cosmos-sdk/issues/24849. I agree we don't need this PR on v3.x.

If we want the signer to behave consistently for v3.x and v4.x then we can merge this.

app/test/consistent_apphash_test.go
```diff
@@ -70,6 +70,7 @@ func DefaultTxOpts() []user.TxOption {
 // and compares it against a previously generated hash from the same set of transactions.
 // App hashes across different commits should be consistent.
 func TestConsistentAppHash(t *testing.T) {
+	t.Skip("skipping consistent app hash test because including the pubkey in the signature broke this test. It was removed on main already.")
```
cmwaters, line 73: We will need to update the hashes we assert since now the consumption of gas and the balances in state are different. Can this be followed up in a separate issue (or is this already done on main?)

rootulp: Sorry my skip message wasn't clear. `main` deleted this test entirely.

We likely still want this unit test on v3.x so I shouldn't have skipped this test. Can address in a follow up PR or preferably in this PR

rootulp:
Okay this test is already borked because the expected app hash and data root on v3.x do not match the expected app hash and data root on v1.x. Created https://github.com/celestiaorg/celestia-app/issues/5002

app/test/consistent_apphash_test.go
rootulp: Created a follow up because the test appear broken even prior to this PR.

https://github.com/celestiaorg/celestia-app/issues/5003

cmwaters:
Was this forward ported to `main`?

rootulp:
It was already fixed on main [here](https://github.com/celestiaorg/celestia-app/blob/2fd381b63ff0284339476eb8c243a3a6ee753dd9/pkg/user/signer.go#L306-L333) via the v4 mega PR: https://github.com/celestiaorg/celestia-app/pull/4400

See https://github.com/celestiaorg/celestia-app/issues/4982#issuecomment-2967845793

PR #4976, opened 2025-06-11
[WIP] test: account sequence mismatch

rootulp:
Adds a unit test to reproduce the `account sequence mismatch` errors observed in https://github.com/celestiaorg/celestia-app/issues/4961#issuecomment-2956994502

PR #4988, opened 2025-06-12
feat: blob tx decoder

rootulp:
Spun out from https://github.com/celestiaorg/celestia-app/pull/4976

The default tx decoder can't decode blob txs.


rootulp:
CI is failing

rach-id:
a go mod tidy is required. also, is the test CI supposed to fail?

rach-id (approved):

PR #4998, opened 2025-06-16
feat(tools): allow disabling metrics reporting for blob submission latency tool

renaynay:
Allow this tool to be used purely to load a network - no need for latency reports if `--disable-metrics` is passed.

### DIFF
--- a/tools/latency-monitor/main.go
+++ b/tools/latency-monitor/main.go
@@ -41,11 +41,12 @@ type txResult struct {
 
 func main() {
 	var (
-		endpoint     = flag.String("grpc-endpoint", defaultEndpoint, "gRPC endpoint to connect to")
-		keyringDir   = flag.String("keyring-dir", defaultKeyringDir, "Directory containing the keyring")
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):
why not use txsim to load the network?

renaynay:
@evan-forbes I was told by @cmwaters that this tool was more reliable for network loading.

evan-forbes:
@renaynay fwiw I think this tool should not be expected to be as reliable since it uses a single account to submit many different txs. That's much more difficult to fill a block since all txs have to be ordered. If we just want to fill a block, then txsim is more reliable since it spins up many different accounts, and txs don't need to be ordered. txsim is what we use in load bearing tests as well

maybe I'm missing something @cmwaters ?

cmwaters:
I created this tool because txsim wasn't initially designed for load testing. It was designed to simulate a range of different transactions. Instead it gets used for load testing which for the most part I think is overly complicated. You should just be able to wrap the TxClient and have a much simpler tool for people to use. 

Perhaps we could allow multiple accounts if we're concerned about ordering but I do think that's something we should just fix. We wouldn't want to require sequencers to have to split their sequencer across multiple accounts.

PR #5000, opened 2025-06-17
chore(ci): update runner

### LINKED ISSUE #4869: Update goreleaser to use self hosted runner
The go releaser task has been having issues on the standard github runners. Once we have a self hosted runner set up we can run that task on it.

Ref: https://github.com/celestiaorg/celestia-app/issues/4711

sysrex:
Fixes #4869 
- updates the runner to be set to ubuntu-8 github managed, non shared runner

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   goreleaser-check:
     if: github.event_name == 'release'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-8
     steps:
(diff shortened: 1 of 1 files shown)


rach-id (approved):
Question: is this a dedicated runner for go-releaser in celestia-app? or for all celestia CI jobs and we need to update the other repos as well?

sysrex:
> Question: is this a dedicated runner for go-releaser in celestia-app? or for all celestia CI jobs and we need to update the other repos as well?

We are testing with go releaser and then if all goes well we are considering rolling it out. Will PR the rest when the time comes.

evan-forbes (approved):

rootulp (approved):
🤞

PR #4996, opened 2025-06-13
fix!: check if tx is still valid before including it in a proposal

rootulp:
Backports https://github.com/celestiaorg/celestia-app/pull/4985 to v3.x

PR #5007, opened 2025-06-18
fix!: check if tx is still valid before including it in a proposal

rootulp:
After an async discussion we decided to backport https://github.com/celestiaorg/celestia-app/pull/4985 and release v3.8.2 directly to Mainnet. 

## Testing

- [x] state sync this binary to Mainnet

## FLUPs

Note: v3.8.2 will skip testnets because Arabica and Mocha are already on v4.x.

- [x] Create v3.8.2-rc0
- [x] Ask DevOps to upgrade Mainnet nodes
- [x] Wait a day
- [x] Create v3.8.2
- [x] Ask POPs to announce

### DIFF
--- a/app/filtered_square_builder.go
+++ b/app/filtered_square_builder.go
@@ -1,0 +1,191 @@
+package app
+
+import (
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	tmbytes "github.com/tendermint/tendermint/libs/bytes"
+	coretypes "github.com/tendermint/tendermint/types"
+
+	square "github.com/celestiaorg/go-square/v2"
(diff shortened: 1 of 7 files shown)


cmwaters:
Do we also have to backport this to the v3.x branch (for the v4 with multuiplexer version)?

rootulp:
Yes. Backported to v3.x in https://github.com/celestiaorg/celestia-app/pull/4996

evan-forbes (approved):
don't need to change things now, but wdyt in the future of creating a v3.8.x branch instead of a release branch for each version (or I think this is what we're doing currently with v4.0.4-release etc) ?

rootulp:
Even though it leads to a lot of branches, I prefer a release branch per patch version so that we don't accidentally release something to Mainnet before it goes on testnets. It is possible to avoid if we're careful with release branches per minor version but seems easier to track with a release branch per patch version.

PR #5024, opened 2025-06-20
feat: add genesis download functionality for known chains

### LINKED ISSUE #4891: Return an error if user invokes `celestia-appd init --chain-id` with a public network's chain ID

deblasis:
## Overview

This PR aims at addressing #4891 

Specifically:
- Add genesis download with SHA-256 verification for known chains (https://github.com/celestiaorg/celestia-app/issues/4891#issuecomment-2921590680) 
- New download-genesis CLI command
- Auto-download during init for public chains with warnings
- Support celestia, mocha-4, arabica-10, arabica-11 networks

Fixes #4891

image


app/download_genesis.go
```diff
@@ -0,0 +1,116 @@
...
+
+// chainIDToSha256 is a map of chainID to the SHA-256 hash of the genesis file for that chain ID.
+// To add a new chain-id, download the genesis file from the networks repo and compute the SHA-256 hash.
+// Add the chain-id and hash to this map.
+var chainIDToSha256 = map[string]string{
+	"celestia":   "9727aac9bbfb021ce7fc695a92f901986421283a891b89e0af97bc9fad187793",
+	"mocha-4":    "0846b99099271b240b638a94e17a6301423b5e4047f6558df543d6e91db7e575",
+	"arabica-10": "fad0a187669f7a2c11bb07f9dc27140d66d2448b7193e186312713856f28e3e1",
+	"arabica-11": "77605cee57ce545b1be22402110d4baacac837bdc7fc3f5c74020abf9a08810f",
+}
```
deblasis: These are the chains and hashes previously in cmd/celestia-appd/cmd/download_genesis.go

I noticed a bit of redundancy with the chainIds defined in /app/chain_ids.go but this approach seemed more fit for purpose due to the sha verification. Potentially we could use the constants from chain_ids.go in here as well but the constants consider only the latest instance (eg: **arabica-10** is missing).

cmd/celestia-appd/cmd/init.go
```diff
@@ -105,6 +105,24 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
 				return fmt.Errorf("genesis.json file already exists: %v", genesisFile)
 			}
 
+			if app.IsKnownChainID(chainID) {
+				fmt.Println("Warning: You are initializing a public chain. This is a very rare edge case and will likely result in a fork. Please be sure you know what you are doing.")
```
rach-id: I suggest updating this message to prompt the user to confirm whether they want to download the known genesis. If not, then a message should state that they should use a different chainID for local environments.

app/download_genesis.go
rach-id, line 1: can you add some unit tests to the new functions added to this file?

cmd/celestia-appd/cmd/init.go
deblasis: Sounds good!

image
image

Addressed in 0e622fefb18714a6ac22c733ff5feca9742e111b

app/download_genesis.go
deblasis: I essentially refactored/moved the existing functions but since they are now public, it makes sense to add some tests.

Let's see if you like 46a37e174b6b42aadf58ccf7f6a78e69d858e339 :)

deblasis:
@rach-id 👋 I have addressed the comments and I believe it's ready for another lookie. 🫡

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := DownloadGenesis(tt.chainID, tt.outputFile)
+
+			if tt.wantErr {
+				if err == nil {
+					t.Errorf("DownloadGenesis() expected error but got none")
+					return
+				}
```
rach-id: could be simplified to `require.NoError(t, err)`

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+				if tt.errContains != "" && !strings.Contains(err.Error(), tt.errContains) {
+					t.Errorf("DownloadGenesis() error = %v, should contain %s", err, tt.errContains)
+				}
```
rach-id: same, could be simplified to require.Contains

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+			} else {
+				if err != nil {
+					t.Errorf("DownloadGenesis() unexpected error = %v", err)
+					return
+				}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+				if tt.errContains != "" && !strings.Contains(err.Error(), tt.errContains) {
+					t.Errorf("computeSha256() error = %v, should contain %s", err, tt.errContains)
+				}
+			} else {
+				if err != nil {
+					t.Errorf("computeSha256() unexpected error = %v", err)
+					return
+				}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create test file
+			testFile := filepath.Join(tempDir, fmt.Sprintf("test_%s.json", tt.name))
+			err := os.WriteFile(testFile, []byte(tt.content), 0644)
+			if err != nil {
+				t.Fatalf("Failed to create test file: %v", err)
+			}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+				}
+
+				// Verify hash is correct
+				hasher := sha256.New()
+				hasher.Write([]byte(tt.content))
+				expectedHash := hex.EncodeToString(hasher.Sum(nil))
+
+				if hash != expectedHash {
+					t.Errorf("computeSha256() = %s, want %s", hash, expectedHash)
+				}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+			}
+		})
+	}
+
+	// Test with non-existent file
+	t.Run("non-existent file", func(t *testing.T) {
+		_, err := computeSha256("/non/existent/file.json")
+		if err == nil {
+			t.Errorf("computeSha256() expected error for non-existent file but got none")
+		}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+
+			if tt.wantErr {
+				if err == nil {
+					t.Errorf("downloadFile() expected error but got none")
+				}
+			} else {
+				if err != nil {
+					t.Errorf("downloadFile() unexpected error = %v", err)
+					return
+				}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+
+				// Verify file content
+				content, err := os.ReadFile(tt.filepath)
+				if err != nil {
+					t.Errorf("Failed to read downloaded file: %v", err)
+					return
+				}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,395 @@
...
+
+				if string(content) != tt.expectContent {
+					t.Errorf("downloadFile() content = %s, want %s", string(content), tt.expectContent)
+				}
```
rach-id: same

app/download_genesis.go
```diff
@@ -0,0 +1,119 @@
...
+func DownloadGenesis(chainID, outputFile string) error {
+	if !IsKnownChainID(chainID) {
+		return fmt.Errorf("unknown chain-id: %s. Must be: %s", chainID, ChainIDs())
+	}
+
+	fmt.Printf("Downloading genesis file for %s to %s\n", chainID, outputFile)
+
+	url := fmt.Sprintf("%s/%s/genesis.json", genesisBaseURL, chainID)
+	if err := downloadFile(outputFile, url); err != nil {
+		return fmt.Errorf("error downloading / persisting the genesis file: %s", err)
```
rach-id:
```suggestion
		return fmt.Errorf("error downloading / persisting the genesis file: %w", err)
```

app/download_genesis.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	fmt.Printf("Downloaded genesis file for %s to %s\n", chainID, outputFile)
+
+	// Compute SHA-256 hash of the downloaded file
+	hash, err := computeSha256(outputFile)
+	if err != nil {
+		return fmt.Errorf("error computing sha256 hash: %s", err)
```
rach-id:
```suggestion
		return fmt.Errorf("error computing sha256 hash: %w", err)
```

deblasis:
> thanks for addressing the feedback so quickly. left some code improvements

Sorry for the silly mistakes. I wrapped the errors and used testify to simplify the tests. 

Also I fixed a test in the **cmd** package that I had missed (I caused the failure with my refactoring, now it's OK but I'd consider removing **cmd/celestia-appd/cmd/download_genesis_test.go** alltogether since it's now pretty dumb and it only wraps logic actually being tested in the new files in **app**).

image

Edit: I removed cmd/celestia-appd/cmd/download_genesis_test.go in the end, executive decision :) happy to restore if it sounds more appropriate to do so.

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+			expected: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := IsKnownChainID(tt.chainID)
+			if result != tt.expected {
+				t.Errorf("IsKnownChainID(%s) = %v, want %v", tt.chainID, result, tt.expected)
+			}
```
rach-id: use `assert.Equal(...)`

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+
+func TestChainIDs(t *testing.T) {
+	result := ChainIDs()
+
+	// Check that all known chain IDs are included
+	expectedChains := []string{"celestia", "mocha-4", "arabica-10", "arabica-11"}
+	for _, chainID := range expectedChains {
+		if !strings.Contains(result, chainID) {
+			t.Errorf("ChainIDs() = %s, should contain %s", result, chainID)
+		}
```
rach-id: use `assert.True(strings.Contains(result, chainID), msg)`

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	// Check that result is comma-separated
+	if !strings.Contains(result, ",") {
+		t.Errorf("ChainIDs() = %s, should be comma-separated", result)
+	}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+			expected: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := GetChainIDOrDefault(tt.args)
+			if result != tt.expected {
+				t.Errorf("GetChainIDOrDefault(%v) = %s, want %s", tt.args, result, tt.expected)
+			}
```
rach-id: same

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+				require.ErrorContains(t, err, tt.errContains)
+			} else {
+				require.NoError(t, err)
+
+				// Verify hash is correct
+				hasher := sha256.New()
+				hasher.Write([]byte(tt.content))
+				expectedHash := hex.EncodeToString(hasher.Sum(nil))
+
+				require.Equal(t, expectedHash, hash)
```
rach-id:
```suggestion
				assert.Equal(t, expectedHash, hash)
```

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+			if tt.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+
+				// Verify file content
+				content, err := os.ReadFile(tt.filepath)
+				require.NoError(t, err)
+
+				require.Equal(t, tt.expectContent, string(content))
```
rach-id:
```suggestion
				assert.Equal(t, tt.expectContent, string(content))
```

app/download_genesis_test.go
```diff
@@ -0,0 +1,364 @@
...
+			errContains: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := downloadFile(tt.filepath, tt.url)
+
+			if tt.wantErr {
+				require.Error(t, err)
```
rach-id:
```suggestion
				assert.Error(t, err)
```

app/download_genesis.go
rach-id, line 1: why btw did you move these to a new package/file? why not leave them where they were?

deblasis: I felt like the logic was quite complex and domain specific and normally I tend to keep my `cmd` packages focused on managing the CLI, outsourcing complexity in the relevant packages.

As soon as I saw the chains and the SHAs of the genesis files in the cmd package I wanted to relocate that code elsewhere if that makes sense.

app/download_genesis_test.go
deblasis: TL;DR I'd use require.Equal here.

Agreed that we should use testify here BUT, what makes you prefer **assert** vs **require** here and in the other comments from this last batch?

Personally I like the fail-fast approach. If a single assertion is not matching expectation, the test is red and we move on. 

I saw that you guys use both in tests here and there, I cannot distil a consistent pattern of usage. 
Please help me understand/learn from your train of thought.

I mean, I see the benefit of having all the failed assertions being reported at once for each test but require sounds simpler (you don't have to remember to differentiate between the require/assert), and safer (you avoid panics by checking with require the objects first and then their properties).

app/download_genesis_test.go
deblasis: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2164882595

app/download_genesis_test.go
deblasis: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2164882595

app/download_genesis_test.go
deblasis: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2164882595

app/download_genesis_test.go
deblasis: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2164882595

app/download_genesis_test.go
deblasis: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2164882595

app/download_genesis_test.go
deblasis: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2164882595

app/download_genesis_test.go
rach-id: personally, I use `require` when there are other steps depending on the statement to be true. For example, if I have `result, err := someFunc()`, I'll do:

```
require.NoError(t, err)
require.NotNil(t, result)
assert.Equal(t, len(some_result), len(result))
assert.Contains(t, result, some_value)
...
```

If there is an error, none of the below statements are meaningful to execute, so we can stop.  But, if `assert.Equal(t, len(some_result), len(result))` is not valid, it would be more meaningful to know if that slice contains some value, maybe it contains some and not all values, etc, which gives more insight on where the issue is. If I run `require.Equal(t, len(some_result), len(result))`, then the only information I know is the two slices are not equally sized.

It's a personal preference, feel free to ignore

rach-id: nvm, just saw the changes

app/download_genesis_test.go
rach-id: same https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2165927543

app/download_genesis_test.go
rach-id: https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2165927543 same for the rest

app/download_genesis.go
rach-id: Makes sense, but I think in this case, it's better to keep them where they were. It would reduce the diff size and make it easier to review.

rach-id: Once you revert the files to their original place, this PR will be ready to review by others

app/download_genesis.go
```diff
@@ -0,0 +1,119 @@
...
+var genesisBaseURL = "https://raw.githubusercontent.com/celestiaorg/networks/master"
+
+// chainIDToSha256 is a map of chainID to the SHA-256 hash of the genesis file for that chain ID.
+// To add a new chain-id, download the genesis file from the networks repo and compute the SHA-256 hash.
+// Add the chain-id and hash to this map.
+var chainIDToSha256 = map[string]string{
+	"celestia":   "9727aac9bbfb021ce7fc695a92f901986421283a891b89e0af97bc9fad187793",
+	"mocha-4":    "0846b99099271b240b638a94e17a6301423b5e4047f6558df543d6e91db7e575",
+	"arabica-10": "fad0a187669f7a2c11bb07f9dc27140d66d2448b7193e186312713856f28e3e1",
+	"arabica-11": "77605cee57ce545b1be22402110d4baacac837bdc7fc3f5c74020abf9a08810f",
```
rach-id: where did you get these hashes btw? 

mocha-4 doesn't seem to be the correct hash: https://github.com/celestiaorg/networks/blob/master/mocha-4/genesis_hash.txt

deblasis: I took them from here:

https://github.com/celestiaorg/celestia-app/blob/e4f57db0a67a326ec9cdced4ccce6a5a595252b3/cmd/celestia-appd/cmd/download_genesis.go#L19-L24

app/download_genesis_test.go
deblasis: I see, thank you for taking the time to elaborate more 👍

deblasis:
I moved the functionality back into the `cmd` package even if I don't fully agree. 

Domain logic in the cmd package could cause duplication and one example is the fact that in order to know which are the PublicNetworks, we have multiple way to do it and therefore we'll have to maintain both.

One is in `chain_ids.go`
image

The other is in `download_genesis.go`
image

image

As a newcomer to this codebase looking at the issue #4891 I immediately noticed this https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2158820605 

I am all for decentralization 😄  but when it comes to code, if there's a truth, I'd like it to come from a single place.

Similar problem with the hash of the genesis that you flagged (here https://github.com/celestiaorg/celestia-app/pull/5024#discussion_r2165938399) mismatches another place I wasn't even aware of, we should try to minimize these problems as much as possible.
These situation cause confusion and ultimately waste energy (the reviewers' as well of course) needlessly IMHO. 

It's nice to review a smaller PR but sometimes a slightly bigger one can pay off dividends in the long run.

rach-id (dismissed):
utAck

deblasis:
Sorry to bother you again @rach-id 🙈
 I noticed that I forgot to lint and also there was a typo I introduced while moving the code that made a test fail. It should be all sorted now.

rach-id:
@deblasis if you check the CI logs, some tests intentionally initialize a network using the known chains. Let's add a flag that answers the `Warning: You are initializing a public chain...` message without requiring to type `y`.

deblasis:
> @deblasis if you check the CI logs, some tests intentionally initialize a network using the known chains. Let's add a flag that answers the `Warning: You are initializing a public chain...` message without requiring to type `y`.

Sounds good! Right away, sir

deblasis:
@rach-id: If I understood correctly how Tastora works, this PR https://github.com/celestiaorg/tastora/pull/55 should be enough to handle the new flag and allow E2E tests to pass with a nice green

cmd/celestia-appd/cmd/download_genesis.go
```diff
@@ -13,6 +13,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// genesisBaseURL is the base URL for downloading genesis files. Can be overridden in tests.
+var genesisBaseURL = "https://raw.githubusercontent.com/celestiaorg/networks/master"
```
rach-id, line 17:
```suggestion
const genesisBaseURL = "https://raw.githubusercontent.com/celestiaorg/networks/master"
```

cmd/celestia-appd/cmd/init.go
```diff
@@ -104,6 +106,45 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
...
+					fmt.Println("Auto-confirming genesis download due to --yes flag")
+				} else {
+					inBuf := bufio.NewReader(cmd.InOrStdin())
+					var err error
+					confirm, err = input.GetConfirmation("Do you want to download the known genesis for this chain?", inBuf, cmd.OutOrStdout())
+					if err != nil {
+						return err
+					}
+				}
+
```
rach-id: could be simplified to:
```suggestion
				confirm, _ := cmd.Flags().GetBool(FlagYes)

				if !confirm {
					inBuf := bufio.NewReader(cmd.InOrStdin())
					var err error
					confirm, err = input.GetConfirmation("Do you want to download the known genesis for this chain?", inBuf, cmd.OutOrStdout())
					if err != nil {
						return err
					}
				}

```

cmd/celestia-appd/cmd/init.go
```diff
@@ -148,6 +189,7 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
 	cmd.Flags().Bool(FlagRecover, false, "provide seed phrase to recover existing key instead of creating")
 	cmd.Flags().String(flags.FlagChainID, "", "genesis file chain-id, if left blank will be randomly created")
 	cmd.Flags().Int64(flags.FlagInitHeight, 1, "specify the initial block height at genesis")
+	cmd.Flags().BoolP(FlagYes, "y", false, "automatically answer yes to prompts")
```
rach-id:
```suggestion
	cmd.Flags().BoolP(FlagYes, "y", false, "automatically answer yes to genesis download for known chainID")
```

rach-id:
also, since the change in the other PR is just a single line, and it would make the CI green for this PR. Please move that change to this PR.

cmd/celestia-appd/cmd/download_genesis.go
deblasis: Can't do, we override the url in tests so that we can use the testserver.

image

cmd/celestia-appd/cmd/init.go
deblasis: nice

deblasis:
> also, since the change in the other PR is just a single line, and it would make the CI green for this PR. Please move that change to this PR.

The PR https://github.com/celestiaorg/tastora/pull/55 is on another repository, I can't bring it here 😄 
Unless you know another way to pass in the flag to the init command in `tastora` from here I didn't notice.

deblasis:
@rach-id I guess we have a Chicken and Egg problem. Which one came first? 🤣 

Essentially: 

- in this PR we need https://github.com/celestiaorg/tastora/pull/55 to be merged to have successful E2E tests
- in https://github.com/celestiaorg/tastora/pull/55 `tastora` needs the `celestia-app` binary to support the new `--yes` flag

I mean, these are my first contributions and I am already having fun!

I guess we need to coordinate.

I suggest we proceed with PR review and to make it essentially ready to be merged, same with the PR in the other repository, then when we are ready we should merge with a commit message that skips the CI run on both repositories: `[skip ci]`.

Once both PRs are merged we should run CI on both and expect all green, otherwise we should fix/revert the merges and try again.

Of course this is just the way I would do it, but I might be totally wrong.

deblasis:
I think I have a better idea, we can detect if we are running in CI and avoid using flags.

Github automatically injects the environment variable `CI=true` we could rely on that

image

Source: https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/

Here I simulated how it would work, same behaviour as with the `--yes` flag but `tastora` won't complain about unknown flags:

image


I just pushed the change in bf86ef92497aa082f49f900ba74877986f1aaa83

deblasis:
I looked into the way you guys test things, pretty cool stuff. I realized that the environment variable is not passing thought because in k8s it's a completely separate environment so in 4d30823cad47347e4db0c1e459687d208ecb809b I set it myself.

Unless I missed something else, this should be finally it.

rach-id:
I'm beginning to lean towards not having an interactive yes. And just download by default the genesis for known chains. However, I would like @rootulp opinion on this.

rootulp:
Agreed on not making it interactive.

deblasis:
Hi guys.

The latest change worked! Using the environment variable to toggle non-interactivity worked.

This means that there's no need for the other PR https://github.com/celestiaorg/tastora/pull/55 (I just closed it)

## Receipts

### Before
image
> The tests were failing waiting for the user to interact with it

### After
image
> the `Init` downloads the genesis 🤘 (we don't see logs because the test doesn't capture stdout inside the github action, perhaps one improvement here could be to add log entries when running in CI so that we know what is happening when not interactive)

## Test failure
The tests failed because previously the `celestia` chain was a brand new one in tests since there was no download, now we start from an existing genesis and, **I haven't looked into the root cause**, it fails validation because of that. 
Regardless, I have the same failure locally so the fact that's interactive or not is irrelevant to the latest e2e test failure:

image

The easiest fix here would be to initialize a non-public chain in tests, `celestia-test` for example. That's another one-liner change.

## Question
I am obviously going to get rid of interactivity if you are all on the same page, but I am VERY curious about the **why** you backtracked on what sounded like a legit requirement. 👀

rach-id:
@deblasis Thanks a lot for sticking around, get involved in discussions, and apply review feedback so quickly 🙏 

In fact, it's not really removing a requirement, we're testing different approaches of how it would look like. After thinking about it, I think it does make sense to keep the interactive setup and add the `--yes` flag, while renaming it to `--confirm-known-chain` or similar to be specific instead of a generic `--yes` that can collide/require breaking changes in the future.

Then, to make the tests work, can you replace the chain ID in here:

https://github.com/celestiaorg/celestia-app/blob/585d2066bbec9fa896ff4d533ddd38783d43d157/test/docker-e2e/e2e_test.go#L68

with this one: 

https://github.com/celestiaorg/celestia-app/blob/585d2066bbec9fa896ff4d533ddd38783d43d157/pkg/appconsts/chain_ids.go#L8

And since this is not an edge case IMO, it's legitimate to want to init a known chainID. I propose to also update the message from:

```
Warning: You are initializing a public chain. This is a very rare edge case and will likely result in a fork. Please be sure you know what you are doing.
```

to:

```
Warning: You are initializing a public chain. This should only be done if you intend to join an existing public network.
```

deblasis:
> @deblasis Thanks a lot for sticking around, get involved in discussions, and apply review feedback so quickly 🙏

You are welcome @rach-id 

This was fun!

cmd/celestia-appd/cmd/init.go
```diff
@@ -31,6 +31,8 @@ const (
 	FlagOverwrite = "overwrite"
 	// FlagRecover defines a flag to initialize the private validator key from a specific seed.
 	FlagRecover = "recover"
+	// FlagConfirmKnownChain defines a flag that automatically answers yes to prompts for known/public chainIDs
```
rach-id: [nit]
```suggestion
	// FlagConfirmKnownChain defines a flag that automatically answers yes to prompts for known/public chainIDs.
```

evan-forbes (dismissed):

cmd/celestia-appd/cmd/download_genesis.go
```diff
@@ -84,10 +85,19 @@ func isKnownChainID(chainID string) bool {
...
 func chainIDs() string {
 	return strings.Join(getKeys(chainIDToSha256), ", ")
 }
 
+// getChainIDOrDefault returns the chainID from the provided arguments or defaults to "celestia".
+func getChainIDOrDefault(args []string) string {
+	if len(args) == 1 {
+		return args[0]
+	}
+	return "celestia"
```
rootulp, line 98: [no change needed] given this was present prior to the PR but seems like we could refactor to use appconsts.

```suggestion
	return appconsts.MainnetChainID
```

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+		{
+			name:     "unknown chain ID",
+			chainID:  "unknown-chain",
+			expected: false,
+		},
+		{
+			name:     "empty chain ID",
+			chainID:  "",
+			expected: false,
+		},
```
rootulp, line 43: [nit] this is missing at least two test cases:
1. arabica-10
2. arabica-11

the arabica-10 test case previously existed also I don't think the names add much so IMO would be easier to maintain to remove the names from the test cases.

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+
+	// Check that all known chain IDs are included
+	expectedChains := []string{"celestia", "mocha-4", "arabica-10", "arabica-11"}
+	for _, chainID := range expectedChains {
+		require.Contains(t, result, chainID)
+	}
+
+	// Check that result is comma-separated
+	require.Contains(t, result, ",")
+}
```
rootulp, line 65: 👍

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+
+func TestDownloadGenesis(t *testing.T) {
+	tempDir := t.TempDir()
+
+	// Content that matches the expected hash for celestia
+	correctContent := `{"genesis_time":"2023-01-01T00:00:00Z","chain_id":"celestia"}`
+	correctHasher := sha256.New()
+	correctHasher.Write([]byte(correctContent))
+	correctHash := hex.EncodeToString(correctHasher.Sum(nil))
```
rootulp, line 110: I'm so confused. Why does this result in the correctHash? The hash should be over the entire `genesis.json` not just this small part.

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+		genesisBaseURL = originalBaseURL
+	}()
+
+	// Temporarily modify chainIDToSha256 for testing
+	originalChainIDToSha256 := chainIDToSha256
+	chainIDToSha256 = map[string]string{
+		"celestia":   correctHash,                             // Correct hash - should pass
+		"mocha-4":    "this_is_definitely_not_the_right_hash", // Wrong hash - should fail
+		"arabica-10": correctHash,                             // Hash won't matter since server returns error
+	}
```
rootulp, line 149: this seems like overkill for the test. We haven't had regressions associated with this code before so I don't think maintaining this test is worth it. IMO we could delete `TestDownloadGenesis` entirely.

Since it is mocking:
- the genesis hash
- the HTTP server
- the chainIdToSha256 map

I don't think it's adding much useful test coverage

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+}
+
+func TestComputeSha256(t *testing.T) {
+	tempDir := t.TempDir()
+
+	tests := []struct {
+		name        string
+		content     string
+		wantErr     bool
+		errContains string
```
rootulp: this field is never used so it should be removed

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+	}{
+		{
+			name:    "valid file",
+			content: "test content",
+			wantErr: false,
```
rootulp: wantErr is never set to true so it likely also should be removed or a test case should be added that uses it

cmd/celestia-appd/cmd/download_genesis_test.go
```diff
@@ -1,27 +1,357 @@
...
+		case "/not-found":
+			w.WriteHeader(http.StatusNotFound)
+			_, _ = w.Write([]byte("Not Found"))
+		default:
+			w.WriteHeader(http.StatusInternalServerError)
+			_, _ = w.Write([]byte("Internal Server Error"))
+		}
+	}))
+	defer server.Close()
+
```
rootulp, line 284: IMO this test is unnecessary because it is mocking an HTTP server.

cmd/celestia-appd/cmd/init.go
```diff
@@ -31,6 +31,8 @@ const (
 	FlagOverwrite = "overwrite"
 	// FlagRecover defines a flag to initialize the private validator key from a specific seed.
 	FlagRecover = "recover"
+	// FlagConfirmKnownChain defines a flag that automatically answers yes to prompts for known/public chainIDs.
+	FlagConfirmKnownChain = "confirm-known-chain"
```
rootulp, line 35: this flag confuses me. Can we please remove it?

cmd/celestia-appd/cmd/init.go
```diff
@@ -104,6 +106,41 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
 				return fmt.Errorf("genesis.json file already exists: %v", genesisFile)
 			}
 
+			if isKnownChainID(chainID) {
+				fmt.Println("Warning: You are initializing a public chain. This should only be done if you intend to join an existing public network.")
```
rootulp, line 110: This shouldn't log a warning because it is expected behavior.

cmd/celestia-appd/cmd/init.go
```diff
@@ -104,6 +106,41 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
...
+
+				confirm, _ := cmd.Flags().GetBool(FlagConfirmKnownChain)
```
rootulp, line 112: We don't need a separate confirm flag for this.

cmd/celestia-appd/cmd/init.go
```diff
@@ -104,6 +106,41 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
...
+
+				if !confirm {
+					inBuf := bufio.NewReader(cmd.InOrStdin())
+					var err error
+					confirm, err = input.GetConfirmation("Do you want to download the known genesis for this chain?", inBuf, cmd.OutOrStdout())
```
rootulp, line 117: Don't prompt the user, just download the genesis file.

test/docker-e2e/e2e_test.go
```diff
@@ -65,7 +66,7 @@ func (s *CelestiaTestSuite) CreateDockerProvider(opts ...ConfigOption) celestiat
 			Version:       getCelestiaTag(),
 			NumValidators: &numValidators,
 			NumFullNodes:  &numFullNodes,
-			ChainID:       "celestia",
+			ChainID:       appconsts.TestChainID,
```
rootulp, line 69: [question] why did this change in this PR?

rach-id: using the chainID "celestia" confuses whether this test is connecting to mainnet or a local network. Also, if you want to init a genesis with a known chain, the confirm known chain needs to be true.

cmd/celestia-appd/cmd/init.go
rach-id: My personal opinion was to have a confirmation whether a user wants to download the genesis, or they mistakenly tried to init a chain with a known chainID. And since this confirmation exists, having a flag to bypass this confirmation is useful.

I wouldn't like it if I run a binary and it starts downloading things without asking me 😆

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: The point of this test is to cover the return value of `isKnownChainID` of the various cases: known -> true, unknown -> false, empty -> false, not to cover all the known chains (which would be just maintenance overhead without any added value IMHO).

To make the test future-proof and deterministic, we could override and defer restore the underlying map inside the test and strictly test the behaviour, without having to rely on known IDs. 

We did something similar in another test in this suite.

How does this sound?

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: TL;DR: These lines just say: I am expecting this made up content and that content has this hash, then in the test cases we return the correct one or not to test the expectations.

The testserver does the rest using the URL as a switch for the logic.

The string in `correctContent` is irrelevant, it "looks" like part of a genesis file but it could have been anything really. Having the whole genesis file here as a string or as a testcase file would have been overkill since we are testing the hash verification logic.

Does this clarify your doubts? Do you think it's unclear to the point that we should improve the code/comments?

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: These tests aim at documenting the behaviour around hash validation and other edge cases, see also the comment above.

I don't think these cases were covered before.

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: It's testing the behaviour of `downloadFile` and it's doing it without hitting the network, that's best practice.
What is that you don't like about this specifically?

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: I started the implementation thinking I could mock the errors easily but I thought it was overkill but then I forgot that I had already added the "non-existent file" test (`computeSha256` essentially can only fail because of io/filesystem related issues)

Addressed in 8534439

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: Addressed in 8534439

cmd/celestia-appd/cmd/download_genesis.go
deblasis: Creating an issue to track this: https://github.com/celestiaorg/celestia-app/issues/5121 (maybe Copilot can give it a shot :smile: )

test/docker-e2e/e2e_test.go
rootulp: Thanks. Resolving.

cmd/celestia-appd/cmd/init.go
rootulp: Confirmation prompts are a UX inconvenience because a node operator can't copy + paste a series of commands into their terminal. IMO if a confirmation prompt is necessary then it would be preferable for the binary to not auto download anything for known chain-ids and instead instruct the node operator to do

```
celestia-appd init celestia
celestia-appd download-genesis
```

instead of

```
celestia-appd init celestia
<wait>
y
```

cmd/celestia-appd/cmd/download_genesis_test.go
rootulp: I think it's unnecessary to test made up hashes. I don't think these test cases add much value.

cmd/celestia-appd/cmd/download_genesis_test.go
rootulp: I do not recommend restoring the underlying map inside the test. Instead I recommend adding a test case for all the known chain-ids: arabica-10 and arabica-11.

cmd/celestia-appd/cmd/init.go
```diff
@@ -104,6 +106,41 @@ func initCmd(basicManager module.BasicManager, defaultNodeHome string) *cobra.Co
...
+					if err != nil {
+						return err
+					}
+				}
+
+				if !confirm {
+					fmt.Println("Genesis download cancelled. For local development, please use a different chain-id (e.g., 'test-chain-1').")
+					return nil
+				}
```
rootulp, line 126: Can this be merged into the conditional block above?

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: I appreciate the feedback, but I respectfully disagree. Let me try again discussing the testing approach here and my philosophy in general. 

The current test already covers the core functionality adequately by testing:

- Known chain ID → true
- Unknown chain ID → false
- Empty chain ID → false

Adding all specific chain IDs would create several issues:

1. **Coupling**: The test becomes tightly coupled to the current data set. When new chains are added (which seems frequent given arabica-10 vs arabica-11), we'd need to update tests purely for data changes, not logic changes.
2. **Testing data vs. behavior**: `isKnownChainID` is a simple map lookup function. We should test that the lookup logic works correctly, not validate the completeness of our data set.
3. **Maintenance overhead**: Every chain addition/removal becomes a test change, creating noise in the test suite without testing additional functionality.

The mock approach I suggested would:

- Test the actual function behavior **deterministically**
- Remain stable regardless of chain ID changes, making it essentially future-proof
- Focus on the algorithm rather than current data
- Follow testing best practices of isolating units under test

If we want to ensure all current chains are covered that's another story, we could add a separate test that validates the completeness of `chainIDToSha256`, keeping our unit test focused on behavior.

What are your thoughts on this distinction between testing behavior vs. testing data completeness?

cmd/celestia-appd/cmd/download_genesis_test.go
deblasis: I understand the concern about using mock data, but I'd like to respectfully rephrase the testing perspective behind this approach.

This test is designed as a unit test that verifies the hash verification logic and error handling behavior of `downloadGenesis`. The specific content being hashed is intentionally irrelevant - what we're testing is:

1. **Hash computation accuracy** - Does the SHA-256 calculation work correctly?
2. Hash **comparison log**ic - Does the function properly detect mismatches?
3. **Error handling** - Are server errors, file system errors, and unknown chain IDs handled properly?
4. **Workflow integrity** - Does the download → hash → verify sequence work as expected?

Using real genesis files would actually make this a **worse** test IMHO because:

- What makes a genesis file real? That can always change.
- Having a real genesis file and hashing that is equivalent, since we are testing **behaviour**, to hashing an "hello world" string or a slimmed down genesis like it is right now

The current approach follows established unit testing principles: fast, deterministic, isolated, and focused. The mock data gives us complete control over test scenarios, allowing us to verify that our integrity checking mechanism works correctly under all conditions.
I
The value here is ensuring our download and verification logic is bulletproof, regardless of what content we're actually processing.

After reading this, can you please elaborate more on 

> I think it's unnecessary to test made up hashes. I don't think these test cases add much value.

Maybe I am not seeing the forest from the trees 🤷‍♂️ but I am always willing to learn something new

cmd/celestia-appd/cmd/download_genesis_test.go
rootulp: > The test becomes tightly coupled to the current data set. When new chains are added (which seems frequent given arabica-10 vs arabica-11), we'd need to update tests purely for data changes, not logic changes.

We don't add chain IDs frequently so IMO it would be more beneficial to have exhaustive test coverage of all the known chain IDs. I think it's good idea to add a test case every time we add a chain ID to the map.

rootulp:
Overall, thank you for persisting through several rounds of feedback. It is my fault the original issue underspecified some details that only became apparent after iterations on this PR. Given our team doesn't have consensus on details for the feature (e.g. should the automatic download require a confirmation prompt?), it is unfair to expect you to drive this forward.

Given it isn't a productive use of time to debate the rationale for some of the unit tests in this PR, I'm inclined to close for now.

deblasis:
The beauty of open source lies in finding consensus (blockchain pun intended) and iterating through feedback. I was happy to proceed with the team's decision, whether or not it included the interactivity flag. 

Asking questions is in my nature, otherwise, I’d be an LLM that just writes code and executes commands until the PR is merged. 🤣 

I think you are wrong. Debating the rationale behind unit tests is a productive use of time, especially when there’s disagreement and you should expect pushback. It’s how we learn and improve. It's a mistake to assume we know everything because we do not.

Thank you all, it’s been a valuable experience, and I enjoyed contributing. 🫡

PR #5025, opened 2025-06-20
feat(test/cmd/txsim): Allow predefined set of namespaces for blob submission

renaynay:
This PR adds a flag to allow a set of one or more predefined namespaces for txsim to use for blob submission. If more than one namespace is provided, it will choose a namespace at random for blob submission. 

```
--namespace <28 byte hex encoded V0 string> --namespace <28 byte hex encoded V0 string>
```

There is no way to currently ensure every defined namespace lands in every block (if 2 or more are provided), but we can add a feature later that disables the randomisation of `blobsPerPFB` per block.

### DIFF
--- a/test/cmd/txsim/cli.go
+++ b/test/cmd/txsim/cli.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/celestiaorg/go-square/v2/share"
 	"math/rand"
 	"os"
 	"os/signal"
@@ -45,6 +47,7 @@ var (
 	blobShareVersion                                  int
 	gasLimit                                          uint64
 	gasPrice                                          float64
+	namespaces                                        []string
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


evan-forbes (dismissed):
utack

test/cmd/txsim/cli.go
```diff
@@ -2,8 +2,10 @@ package main
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/celestiaorg/go-square/v2/share"
```
rootulp: golangci-lint is complaining.

TBH I dislike this lint rule so I would also be okay with disabling the custom settings [here](https://github.com/celestiaorg/celestia-app/blob/ea39814e10c87ae883b8eedce0a020d59060ef9f/.golangci.yml#L44-L50).

test/txsim/blob.go
```diff
@@ -40,10 +40,10 @@ func NewBlobSequence(sizes, blobsPerPFB Range) *BlobSequence {
 	}
 }
 
-// WithNamespace provides the option of fixing a predefined namespace for
+// WithNamespaces provides the option of fixing a set of predefined namespaces for
 // all blobs.
-func (s *BlobSequence) WithNamespace(namespace share.Namespace) *BlobSequence {
-	s.namespace = namespace
+func (s *BlobSequence) WithNamespaces(namespaces []share.Namespace) *BlobSequence {
```
rootulp, line 45: [no change needed] seems technically API breaking because exported method renamed but I couldn't find any usage outside this repo so seems fine

rootulp (approved):
LGTM. Lint is complaining so I'll push a commit to fix it.

test/cmd/txsim/cli.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/5025/commits/b406161a8223b280d450a644df71ce1b50728131

rootulp (approved):

test/txsim/blob.go
renaynay: Yeah it's dangling so i figured nbd

rootulp:
Overriding second approval b/c Evan previously approved and the only change was a lint fix.

PR #3425, opened 2024-05-02
fix: downgrade ibc and retract v1.8.0

evan-forbes:
## Overview

downgrades ibc and retracts v1.8.0

PR #5035, opened 2025-06-23
chore: fix Tx endpoint via updating to CometBFT v0.38 protos

### LINKED ISSUE #4925: v4.0.2-mocha API returns errors on some requests
### Summary of Bug

In the new version, we can't see transaction details via the API.   
For example: https://celestia-testnet-api.itrocket.net/cosmos/tx/v1beta1/txs/8715C3C7FC1583908AA40D60B84DB3C9021AD8E2F11A301D1C461849857D4042
```
{
  "code": 3,
  "message": "error in json rpc client, with http response metadata: (Status: 200 OK, Protocol HTTP/1.1). error unmarshalling result: illegal base64 data at input byte 4: invalid request",
  "details": [
  ]
}
```

### Version

```
cosmos_sdk_version: v1.29.1-sdk-v0.50.12
go: go version go1.24.1 linux/amd64
name: celestia-app
server_name: celestia-appd
version: 4.0.2-mocha
```

### Steps to Reproduce

https://celestia-testnet-api.itrocket.net/cosmos/tx/v1beta1/txs/8715C3C7FC1583908AA40D60B84DB3C9021AD8E2F11A301D1C461849857D4042

evan-forbes:
## Overview

closes #4925 

this fix is annoying to test as we need to have a v3 embedded app running on v0.38 to hit

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -254,10 +254,10 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.30.0-sdk-v0.46.16
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.30.0-sdk-v0.46.16.0.20250623184849-41564030103e // v1.30.0-sdk-v0.46.16
 	// Replace IBC with celestiaorg fork which includes fixes for security vulnerabilities.
 	github.com/cosmos/ibc-go/v6 => github.com/celestiaorg/ibc-go/v6 v6.2.4
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
-	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.52.1-tm-v0.34.35
+	github.com/tendermint/tendermint => github.com/celestiaorg/celestia-core v1.52.1-tm-v0.34.35.0.20250623165146-701d1edca3d5 // v1.52.1-tm-v0.34.35
 )
--- a/go.sum
+++ b/go.sum
@@ -318,10 +318,10 @@ github.com/celestiaorg/bittwister v0.0.0-20231213180407-65cdbaf5b8c7 h1:nxplQi8w
 github.com/celestiaorg/bittwister v0.0.0-20231213180407-65cdbaf5b8c7/go.mod h1:1EF5MfOxVf0WC51Gb7pJ6bcZxnXKNAf9pqWtjgPBAYc=
 github.com/celestiaorg/blobstream-contracts/v3 v3.1.0 h1:h1Y4V3EMQ2mFmNtWt2sIhZIuyASInj1a9ExI8xOsTOw=
 github.com/celestiaorg/blobstream-contracts/v3 v3.1.0/go.mod h1:x4DKyfKOSv1ZJM9NwV+Pw01kH2CD7N5zTFclXIVJ6GQ=
-github.com/celestiaorg/celestia-core v1.52.1-tm-v0.34.35 h1:N+sJFu0h8okPkRjgQh/2iN4pi3atrR9a/vkXLwEKXfw=
-github.com/celestiaorg/celestia-core v1.52.1-tm-v0.34.35/go.mod h1:SI38xqZZ4ccoAxszUJqsJ/a5rOkzQRijzHQQlLKkyUc=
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -254,10 +254,10 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.30.0-sdk-v0.46.16
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.30.0-sdk-v0.46.16.0.20250623184849-41564030103e // v1.30.0-sdk-v0.46.16
```
evan-forbes: ofc, we need to bump to official releases before merging

evan-forbes:
manually confirmed due to needing a v3 client to query a v4 node, although we should aim to have tests for all queries should we upgrade comet again.

v3 client v4 server
```
celestia-appd q tx d2af9828bebf2696803346fba22b6b1d2bfb17c529ec884c6ee8079827ab3025
Error: error in json rpc client, with http response metadata: (Status: 200 OK, Protocol HTTP/1.1). error unmarshalling result: illegal base64 data at input byte 4
```

this branch and a v4 server

```code: 0
codespace: ""
data: 122A0A282F63656C65737469612E626C6F622E76312E4D7367506179466F72426C6F6273526573706F6E7365
events:
- attributes:
  - index: false
    key: spender
    value: celestia1q779acczk0ffp0wtmurt9ey45s8ka2h9qsw7hd
...fz5u2lcJn/BOwU6q5O/ad8hAhmtpwTKvzBDnxKC0hAAHooqaVuTMo6Z4JFBKp6mdv1ApVrif6mIkbcKha4Pxww==
txhash: D2AF9828BEBF2696803346FBA22B6B1D2BFB17C529EC884C6EE8079827AB3025```

**Comment by evan-forbes** [comment]
the last thing to test with this is to ensure it doesn't mess with relayers. I don't think it should, since we didn't change anything in ibc and mocha is already using these protos in v0.38, but we likely need to check any way before mainnet

imo we can merge an unblock mocha

**Comment by evan-forbes** [comment]
worst case scenario, we just have to create a temporary fork of hermes/go-relayer to also have the string instead of bytes in a events

File: go.mod [hunk 2]
```diff
@@ -254,10 +253,10 @@ require (
 )
 
 replace (
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.30.0-sdk-v0.46.16
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.31.0-sdk-v0.46.16
 	// Replace IBC with celestiaorg fork which includes fixes for security vulnerabilities.
-	github.com/cosmos/ibc-go/v6 => github.com/celestiaorg/ibc-go/v6 v6.2.4
+	github.com/cosmos/ibc-go/v6 => github.com/celestiaorg/ibc-go/v6 v6.3.0
```
rootulp, line 258: 😨  IBC minor versions are state machine breaking so this version number is a bit scary. Recall that during the Kazbegi onsite we had to downgrade ibc-go in https://github.com/celestiaorg/celestia-app/pull/3425

There aren't any other consumers of celestiaorg/ibc-go so proposal to:
1. Delete v6.3.0 release
2. Create v6.2.6 release
3. Use that here

That way no one confuses this release with upstream ibc-go v6.3.0 release.

evan-forbes: our fork of IBC won't make breaking changes in minor versions tho. that being said it does make sense that we could have likely gotten away with doing v6.2.6

rach-id: the dependency bump in ibc-go seems to also bump core/cosmos-sdk. I guess it would be safer to run a test

rach-id (approved):
approving to unblock, but ideally this change gets tested in a realistic network

evan-forbes:
> but ideally this change gets tested in a realistic network

the multiplexer currently makes this pointlessly difficult as it will download and use the binary that was hardcoded in it instead of letting the user choose. Because of that we hit snags with tastora / multiplexer and didn't want to wait to fix those. Not 100% sure what is meant by a real network here, but it was manually tested on mocha and the issue can also be tested by running a v4 node and querying the rest tx endpoint and a v3 client.

Ideally we do have rpc tests between upgraded versions that update comet. That should have been a requirement before merging v4 or the multiplexer.

go.mod
evan-forbes: @rach-id the only reason to bump IBC here was to fix compiler issues in tests. nothing else was changed. this was to fix the breaking changes in the core in protos which why we had to fix them everywhere else before bumping here.

rach-id: That makes sense, thanks for clarifying :pray:

Issue #4962, opened 2025-06-09
grpc: received message larger than max when calling ProcessProposal


evan-forbes:
validators on mocha noticed the below error
```
[...]

PR #5044, opened 2025-06-24
debug: grpc max msg size

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/4962

I used a script and txsim to generate blocks with > 4 MiB of blob data but couldn't repro the error in the original issue.

```
./scripts/v3-big-blocks.sh
```

```
txsim --key-path ~/.celestia-app/ --grpc-endpoint localhost:9090 --seed 1234 --poll-time 1s --blob 5 --blob-sizes 1000000
```

Wait until a block has 5 x 1 MiB blobs included

```
3:32PM INF finalizing commit of block hash=7069E79DE869A347DA0F80353C09E029EECBA7860938659B3BF221183DFC70C3 height=20 module=consensus num_txs=5 root=864D52D27724EB19A3C0ECA207805C08B50F9C748EDDB5BE53E7FF14E12D8736
```


rootulp:
Doesn't need to be merged.

Issue #4992, opened 2025-06-13
Stop panicking in PrepareProposal unless absolutely necessary


evan-forbes:
We could return errors, but currently we're panicking. There's not a lot of benefit to panicking since it only increases attack surface or if the panics are caught then they should just be an error.

rootulp:
On v3.x, `PrepareProposal` only had one return argument so it was not possible to return an error. https://github.com/celestiaorg/celestia-app/blob/39292d9f8a64f6849169088e4e8b69016f777524/app/prepare_proposal.go#L24

On v4.x, `PrepareProposalHandler` has a second return argument for errors so it is possible to return an error instead of `panic`ing 
https://github.com/celestiaorg/celestia-app/blob/e2c559507c9369ce74becca2bdeddb4cdb61406e/app/prepare_proposal.go#L20

rootulp:
@MozirDmitriy has a PR to close this: https://github.com/celestiaorg/celestia-app/pull/5460

PR #5047, opened 2025-06-25
fix: stop panicking in PrepareProposalHandler

MozirDmitriy:
## Overview

Refactored the PrepareProposalHandler function to return errors instead of panicking when encountering failures during proposal preparation.
This change improves robustness and reduces the attack surface by ensuring that recoverable errors are properly propagated to the caller, rather than causing the node to crash.
Error logging is preserved where appropriate.
No changes were made to the core logic or the function signature.
This addresses issue #4992.

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -40,7 +40,7 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
 		appconsts.SubtreeRootThreshold,
 	)
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 
 	txs := fsb.Fill(ctx, req.Txs)
@@ -48,7 +48,7 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
 	// Build the square from the set of valid and prioritised transactions.
 	dataSquare, err := fsb.Build()
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 
 	// Erasure encode the data square to create the extended data square (eds).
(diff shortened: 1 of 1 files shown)


app/prepare_proposal.go
```diff
@@ -57,13 +57,13 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
...
-		panic(err)
+		return nil, err
 	}
 
 	dah, err := da.NewDataAvailabilityHeader(eds)
 	if err != nil {
 		app.Logger().Error("failure to create new data availability header", "error", err.Error())
-		panic(err)
+		return nil, err
 	}
```
rootulp: [optional] we can do this in a follow-up PR but IMO it makes sense to include the extra info that we're currently logging in the returned error. So one option would be:

```go
dah, err := da.NewDataAvailabilityHeader(eds)
if err != nil {
    wrapped := fmt.Errorf("error creating new data availability header with eds %+v: %w", eds, err)
    app.Logger().Error(wrapped)
    return nil, wrapped
}
```

rootulp (dismissed):
Thanks for the PR, overall LGTM. Two things that could improve this PR but we can also do them in follow-up PRs:
1. [strongly encouraged] write unit tests for each of the cases that can now return errors and assert that an error is indeed returned instead of a panic.
2. These errors lack surrounding context even though we log additional context before some of the errors. Proposal to refactor so that the returned error contains the information from the logs.

rach-id (dismissed):

app/prepare_proposal.go
```diff
@@ -40,30 +41,32 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
...
 	// Erasure encode the data square to create the extended data square (eds).
 	// Note: uses the nmt wrapper to construct the tree. See
 	// pkg/wrapper/nmt_wrapper.go for more information.
 	eds, err := da.ExtendShares(share.ToBytes(dataSquare))
 	if err != nil {
-		app.Logger().Error("failure to erasure the data square while creating a proposal block", "error", err.Error())
-		panic(err)
+		msg := fmt.Sprintf("failure to erasure the data square while creating a proposal block: %v", err)
+		app.Logger().Error(msg)
+		return nil, fmt.Errorf(msg)
```
rootulp, line 62: golangci-lint is failing here

evan-forbes:
will close this for now @MozirDmitriy as it doesn't seem to be worked on but please feel free to reopen and update. If we're still panicking on main we can change here and merge

Issue #5057, opened 2025-06-25
`DEPRECATED:  archives.builds  should not be used anymore, check https://goreleaser.com/deprecations#archivesbuilds for more info`


rootulp:
## Context

[v4.0.5-rc0](https://github.com/celestiaorg/celestia-app/releases/tag/v4.0.5-rc0) does not have prebuilt binaries attached.

## Problem

https://github.com/celestiaorg/celestia-app/actions/runs/15881977632/job/44784459579#step:7:65

PR #5058, opened 2025-06-25
fix: goreleaser config by replacing deprecated fields

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5057

## Testing

```
make goreleaser-check
```

passes


rootulp:
CI isn't going to give us any new info because goreleaser-check doesn't run on PRs. It probably should though.

PR #5059, opened 2025-06-25
fix(multiplexer): shut down on CTRL + C

### LINKED ISSUE #4763: Multiplexer doesn't gracefully shut down on CTRL + C

evan-forbes:
stops the multiplexer from hitting its mutex twice, which will happen if comet calls an abci method _during_ cleanup.

takes a while to hit, but if you decrease the block time it gets hit more. Its still pretty difficult to hit ,which made it kinda hard to debug

a full fix would probably require a larger refactor. closing comet before taking the mutex is maybe another solution, but this is fast and I know it works. if it didn't take so long to hit I might try other approaches

closes #4763


multiplexer/abci/multiplexer.go
```diff
@@ -52,6 +53,9 @@ const (
 type Multiplexer struct {
 	logger log.Logger
 	mu     sync.Mutex
+	// done is a hack to prevent the multiplexer from catching the above mutex twice and avoiding mu.TryLock
+	// as that could result in an accidental crash.
+	done atomic.Bool
```
evan-forbes, line 58: per comments, this is a hack and a longer term solution likely involves a refactor. I could have used TryLock on the mutex, but I was worried that that was not explicit enough and maybe could be hit if we accidentally call the mutex twice

multiplexer/abci/multiplexer.go
```diff
@@ -450,9 +454,12 @@ func openTraceWriter(traceWriterFile string) (w io.WriteCloser, err error) {
 
 // getApp gets the appropriate app based on the latest application version.
 func (m *Multiplexer) getApp() (servertypes.ABCI, error) {
```
tac0turtle, line 456: it seems like everywhere getApp is called we have a context. did you try passing the context here for cancellation? it may be hard to verify on all flows as im not sure comet uses contexts correctly for shutdown?

cmwaters (approved):
Makes sense. There's no case for restarting the multiplexer right? i.e. where you'd need to set the done from true to false again?

evan-forbes:
> There's no case for restarting the multiplexer right?

not after we start the cleanup I don't think no

PR #5061, opened 2025-06-26
fix: use go-square to unmarshal blob tx instead of tendermint

cmwaters:
Noticed that we were using an outdated version of decoding the blob tx (which now resides in go-square instead of tendermint)

### DIFF
--- a/app/encoding/blob_tx_decoder.go
+++ b/app/encoding/blob_tx_decoder.go
@@ -2,12 +2,16 @@ package encoding
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	coretypes "github.com/tendermint/tendermint/types"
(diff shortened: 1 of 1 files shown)


rach-id (commented):
makes me wonder if we're depending on old code somewhere else... We need to find a way not to have duplicate code, and if we do, we should do something for this not to happen again

rach-id (approved):

rootulp:
> using an outdated version of decoding the blob tx

If [this](https://github.com/celestiaorg/celestia-core/blob/a97fbf8982a73683e18fddcb2bf57974e0493798/types/tx.go#L323-L343) is outdated then why isn't it deprecated? 

> We need to find a way not to have duplicate code, and if we do, we should do something for this not to happen again

Deprecate and eventually delete the code that isn't intended to be used.

PR #5066, opened 2025-06-26
ci: allow backporting to v4.x

rootulp:
I would prefer if we didn't need this yet and instead consolidated `main` and `v4.x`

Ref: https://github.com/celestiaorg/celestia-app/pull/4796#issuecomment-3008632798

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -23,6 +23,14 @@ pull_request_rules:
       backport:
         branches:
           - v3.x
+  - name: backport patches to v4.x branch
+    conditions:
+      - base=main
+      - label=backport:v4.x
+    actions:
+      backport:
+        branches:
+          - v4.x
   - name: forward-port patches to main branch (v1.x)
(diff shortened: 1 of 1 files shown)


rach-id (approved):
Agree. Talis addition is non breaking. Unless we want to depend also on latest core, but I dont think thats beneficial at the moment.
Anyway, having a mergify rule is fine and will be handy whenever we cut v4.x. So im approving this PR

evan-forbes:
we have to have this at some point, and since we're wrapping up with v4 bugs its worth the tradeoff imo to not worry about accidently including something or have a new release branch system

rootulp:
> we have to have this at some point, and since we're wrapping up with v4 bugs its worth the tradeoff imo to not worry about accidently including something

IMO we should include everything in v4 _except_ breaking changes that are strictly for v5 but not worth arguing about it. I'll just start backporting every PR from now on.

PR #5110, opened 2025-06-30
chore: fix talis ssh and genesis config

rach-id:
## Overview

Just a couple small fixes for talis:

1. compares only the curve and the key parts of the ssh key and not the hostname too
2. apply the genesis mods when exporting to allow setting the square size

### DIFF
--- a/tools/talis/README.md
+++ b/tools/talis/README.md
@@ -158,6 +158,8 @@ talis genesis -s 128 -a /home/$HOSTNAME/go/src/github.com/celestiaorg/celestia-a
 
 Keep in mind that we can still edit anything in the payload before deploying the network.
 
+Note: When increasing the genesis square size, ensure you also increase the `SquareSizeUpperBound` constant to allow blocks to be created at the new size.
+
 ### deploy
 
 This step is when the network is actually started. The payload is uploaded to each instance in the network directly from the user's machine. After delivering the payload, the start script is executed in a tmux session called "app" on each machine.
--- a/tools/talis/digital_ocean.go
+++ b/tools/talis/digital_ocean.go
@@ -61,7 +61,11 @@ func DOClient(token string) *godo.Client {
 // GetDOSSHKeyMeta checks if the provided raw SSH public key is registered in DigitalOcean
 // and returns its ID and Name. If not found, returns an error instructing to upload the key.
 func GetDOSSHKeyMeta(ctx context.Context, client *godo.Client, publicKey string) (godo.Key, error) {
-	publicKey = strings.TrimSpace(publicKey)
+	pubKeySplit := strings.Split(publicKey, " ")
+	if len(pubKeySplit) <= 1 {
+		return godo.Key{}, fmt.Errorf("invalid public key format")
+	}
+	publicKey = strings.Join(pubKeySplit[:2], "")
 
 	// Pagination options
 	opt := &godo.ListOptions{PerPage: 200}
@@ -73,8 +77,9 @@ func GetDOSSHKeyMeta(ctx context.Context, client *godo.Client, publicKey string)
 		}
 
 		for _, key := range keys {
-			// Compare the trimmed public key string
-			if strings.TrimSpace(key.PublicKey) == publicKey {
+			// only compare the first two parts of the public key. The third part is the host
+			// which can be ignored.
+			if strings.Join(strings.Split(key.PublicKey, " ")[:2], "") == publicKey {
 				return key, nil
 			}
 		}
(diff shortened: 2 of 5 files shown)


tools/talis/digital_ocean.go
```diff
@@ -73,8 +77,9 @@ func GetDOSSHKeyMeta(ctx context.Context, client *godo.Client, publicKey string)
 		}
 
 		for _, key := range keys {
-			// Compare the trimmed public key string
-			if strings.TrimSpace(key.PublicKey) == publicKey {
+			// only compare the first two parts of the public key. The third part is the host
+			// which can be ignored.
+			if strings.Join(strings.Split(key.PublicKey, " ")[:2], "") == publicKey {
```
rach-id, line 82: it's fine to assume the ssh keys from DO will always contain at least 2 fields. This is because they're validated before they're added.

ninabarbakadze (dismissed):
utAck

tools/talis/init.go
```diff
@@ -113,8 +109,8 @@ func initCmd() *cobra.Command {
 func DefaultConfigProfile(cfg *cmtconfig.Config, tables []string) *cmtconfig.Config {
 	cfg.Instrumentation.TracingTables = strings.Join(tables, ",")
 	cfg.Instrumentation.TraceType = "local"
-	cfg.P2P.SendRate = 100000000
-	cfg.P2P.RecvRate = 110000000
+	cfg.P2P.SendRate = 400_000_000
+	cfg.P2P.RecvRate = 410_000_000
```
evan-forbes: understand changing this since it was 400 above, but maybe we should keep at max 100 instead? 

 ~1Gbps is seems pretty easy to dos by itself, and 400 seems like it would mainly make that DoS vector worse while not really improving throughput since we're always sending data to and from multiple peers

tools/talis/genesis.go
```diff
@@ -101,7 +101,7 @@ func generateCmd() *cobra.Command {
 
 	cmd.Flags().StringVarP(&chainID, chainIDFlag, "c", "", "Override the chainID in the config")
 	cmd.Flags().StringVarP(&rootDir, rootDirFlag, "d", ".", "root directory in which to initialize (default is the current directory)")
-	cmd.Flags().IntVarP(&squareSize, "ods-size", "s", appconsts.TalisSquareSizeUpperBound, "The size of the ODS for the network")
+	cmd.Flags().IntVarP(&squareSize, "ods-size", "s", appconsts.TalisSquareSizeUpperBound, "The size of the ODS for the network (make sure to also build a celestia-app binary with a greater SquareSizeUpperBound)")
```
rootulp, line 104: [no change needed] Now that we have a `main` and `v4.x` branch, we can actually bump `SquareSizeUpperBound` on main

tools/talis/init.go
rootulp: [nit] once we agree on the desired SendRate and RecvRate, can we make the numbers readable via:

```go
	cfg.P2P.SendRate = 1 * gigabyte
	cfg.P2P.RecvRate = 1 * gigabyte
```

tools/talis/README.md
```diff
@@ -158,6 +158,8 @@ talis genesis -s 128 -a /home/$HOSTNAME/go/src/github.com/celestiaorg/celestia-a
 
 Keep in mind that we can still edit anything in the payload before deploying the network.
 
+Note: When increasing the genesis square size, ensure you also increase the `SquareSizeUpperBound` constant to allow blocks to be created at the new size.
```
rootulp, line 167: [optional] and also `MaxBytes`

Ref: https://github.com/rootulp/celestia-app/blob/768ef5da532acc9fc3ea305557c0765875520dcb/app/default_overrides.go#L248

rootulp (dismissed):
LGTM but the comment about SendRate / RecvRate from Evan may block this

tools/talis/init.go
rach-id: 8377ad57d2509959637edc8090427f079df0bcd9

tools/talis/genesis.go
rach-id: should we also bump for v4 to follow main? or as you said, let's delay until necessary to avoid needing to backport sutff to v4?

tools/talis/README.md
rach-id: if the genesis has 256 as max square size, does MaxBytes needs to change? I thought it's dynamic depending on what's in the gov parameter?

rootulp: `MaxBytes` needs to be increased either in the genesis file or by governance after chain start. For example https://www.mintscan.io/celestia/proposals/5 increases the square size and the max bytes.

rach-id: aaah, makes sense 👍 That's already handled by talis automatically: https://github.com/rootulp/celestia-app/blob/768ef5da532acc9fc3ea305557c0765875520dcb/tools/talis/network.go#L56

Issue #5134, opened 2025-07-01
Decide on height or time based TTLs


ninabarbakadze:
## Description 

Explore having a user specified TTL (either height based or time based).

ninabarbakadze:
Height based I'm thinking @cmwaters since we recently removed time based ttls from the mempool it feels like height based ttls will sync up with consensus nodes better

evan-forbes:
height based also makes sense to me since its universal for all nodes at the tip of the chain

ninabarbakadze:
I'm closing this as I already started implementation for height based ttls

Issue #5133, opened 2025-07-01
Replay evicted transactions in tx client


ninabarbakadze:
## Description

Finish the prototype — [PR #4791](https://github.com/celestiaorg/celestia-app/pull/4791)  - which is designed to address the following scenario:

In some cases of sequence mismatches, the expected nonce is lower than what the user provided, suggesting that earlier transactions were evicted. The prototype above updates txInfo to retain the txBytes, allowing the TxClient to replay those transactions to the consensus node in situations where it may not have them all.

evan-forbes:
can we close this since #5301 is merged?

ninabarbakadze:
The title of the issue was a bit outdated but the issue is still relevant. Specifically this part is what I'm working on currently.

_This next phase builds on that by preparing for block-based TTLs, where users can specify a height-based expiry for their transactions. Once TTLs are supported, if a transaction hasn't landed in a block by the specified expiry height, it will be rejected and the tx client can safely resubmit a fresh transaction with the correct nonce. By tracking submitted transactions in the TxTracker, we ensure that replay only occurs after TTL expiry, which gives us replay protection and clearer lifecycle management for transactions on the client side. This should significantly reduce nonce mismatch related errors._

evan-forbes:
can we get an update here?

ninabarbakadze:
I don't have any updates here the feature is mostly done we just need to evaluate as a group whether we still want this change or not. We deprioritised this two weeks ago because we had other more important work in progress so we could revisit it during our upcoming sync.

evan-forbes:
> I don't have any updates here the feature is mostly done we just need to evaluate as a group whether we still want this change or not

is this all the stuff above in OP comment or just the TTLs

ninabarbakadze:
We never reached consensus on this. Moving it to unplanned for now.

PR #5139, opened 2025-07-02
fix: tx client evictions bug

### LINKED ISSUE #4784: account sequence mismatch not handled in `TxClient.SubmitPayForBlob`

### LINKED ISSUE #5132: Handle rejections and evictions in tx client
## Description

We've had a number of bugs reported like [this one](https://github.com/celestiaorg/celestia-app/issues/4784)  caused by a mismatch in transaction status across consensus nodes. Specifically, one node may return an eviction status for a transaction, while another node proposes it and includes it in a block.

In such cases, we roll back the nonce in the signer based on the eviction status. However, once the transaction is included in a block, the on-chain nonce increases becoming higher than what we’re tracking locally. As a result, all subsequent transactions submitted by the user through the tx client will be rejected, since the local nonce is now lower than expected. To recover, users must restart the LN.

## To remedy this ASAP

1. Only roll back the nonce when handling rejections knowing that this tx won't get included by another node.
2. Continue polling for evicted transactions because they should eventually get proposed by another node.
3. Replay transactions once we are sure that they got evicted [feat: handle sequence mismatches by replaying transactions](https://github.com/celestiaorg/celestia-app/pull/4791) (might not need it in the initial fix @cmwaters)

ninabarbakadze:
## Overview

- Bumps core to the patch that included [mempool fix](https://github.com/celestiaorg/celestia-core/commit/92b3a75f187b2ecfc7cee225630b9e1ba68b14ab)
- Addresses the initial painful bug users where hitting where transaction was evicted by one node and we'd roll back the nonce effectively stalling the tx client #4784 

Fixes #4784
Fixes #5132

Current design only rolls back the nonce when transaction was rejected in recheck. If we get eviction from a node we will resubmit the transaction and continue polling for confirmation. Nodes that already have this transaction bytes will just discard it. Tx will eventually be committed. 

This is an initial fix so users can actually avoid restarting the nodes. Follow-up work is adding user specified TTLs and replaying transactions #5134 #5133

Opens #5273


cmwaters:
Just FYI, while nina is out, I will look to in my spare time polish this into a place where it's ready to be reviewed

ninabarbakadze:
Testing this is blocked on https://github.com/celestiaorg/celestia-core/issues/2195 because sometimes the mempool state differs from the network state. When a transaction gets evicted from the mempool, the mempool state may still retain the transaction, causing nonce mismatch errors where the expected nonce is higher than the current tx's nonce.

ninabarbakadze:
I just thought about a scenario where if user submits 10 transactions with nonces ranging from 1-10  first 3 get confirmed then 2 get rejected and network nonce is 3 but tx tracker has subsequent transactions saved as nonce 6 onwards. If a transaction gets evicted after rejections we’d hit a situation where we can only submit transactions with nonce 4 but locally in tx tracker have transactions saved as nonce 6 that are also signed with this nonce so if we were to shift the nonce we'd be effectively resigning these transactions. This error would be very confusing for the user.

This should be a pretty rare edge case though because transaction would have to get rejected in recheck and a subsequent one evicted for this to get hit.

pkg/user/tx_client_test.go
```diff
@@ -228,39 +230,101 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
...
+		require.Equal(t, res.Code, abci.CodeTypeOK)
+		// They should be removed from the tx tracker after confirmation
+		_, _, exists := txClient.GetTxFromTxTracker(resp.TxHash)
+		require.False(t, exists)
 	}
 
-	seqAfterEviction := sender.Sequence()
-	require.Equal(t, seqBeforeEviction, seqAfterEviction)
+	// At least 8 txs should have been evicted and resubmitted
+	require.GreaterOrEqual(t, len(evictedTxHashes), 8)
```
rootulp, line 312: [question] why 8?

rootulp (dismissed):

pkg/user/tx_client.go
```diff
@@ -404,17 +405,40 @@ func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn,
...
 	if err := client.signer.IncrementSequence(signer); err != nil {
 		return nil, fmt.Errorf("increment sequencing: %w", err)
 	}
+
+	return resp.TxResponse, nil
+}
+
+// broadcastTxAfterEviction resubmits a transaction that was evicted from the mempool.
+// Unlike the initial broadcast, it doesn't increment the signer's sequence number.
+func (client *TxClient) broadcastTxAfterEviction(ctx context.Context, conn *grpc.ClientConn, txBytes []byte, signer string) (*sdktypes.TxResponse, error) {
```
cmwaters: It looks like you could consolidate this with the `broadcastTx` function since they have a high degree of overlap.

Suggestion:
- rename this method to `broadcastTx`
- rename the other method (what was `broadcastTx`) to `broadcastTxAndIncrementSequence`

pkg/user/tx_client.go
```diff
@@ -518,37 +542,37 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+			if !exists {
+				return nil, fmt.Errorf("tx: %s not found in tx client txTracker; likely failed during broadcast", txHash)
+			}
+			// Reset sequence to the rejected tx's sequence to enable resubmission
+			// of subsequent transactions.
+			if err := client.signer.SetSequence(signer, sequence); err != nil {
+				return nil, fmt.Errorf("setting sequence: %w", err)
+			}
+			client.deleteFromTxTracker(txHash)
+			return nil, fmt.Errorf("tx with hash %s was rejected by the node", txHash)
```
cmwaters, line 550: Should this be a broadcastTxError?

pkg/user/tx_client_test.go
```diff
@@ -228,39 +230,101 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
...
 	sender := txClient.Signer().Account(txClient.DefaultAccountName())
-	msg := bank.NewMsgSend(sender.Address(), testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(params.BondDenom, 10)))
-	var seqBeforeEviction uint64
-	// Loop five times until the tx is evicted
-	for i := 0; i < 5; i++ {
-		seqBeforeEviction = sender.Sequence()
-		resp, err := txClient.BroadcastTx(ctx.GoContext(), []sdk.Msg{msg}, fee, gas)
+	seqBeforeSubmission := sender.Sequence()
+	blobs := blobfactory.ManyRandBlobs(random.New(), 2, 2)
+	resp, err := txClient.BroadcastPayForBlob(ctx.GoContext(), blobs, fee, gas, user.SetTimeoutHeight(timeoutHeight))
```
cmwaters, line 245: Given that the timeout height is 1 and we are already past height 1 by the time the txClient submits the transaction, I am surprised that this transactions doesn't immediately fail

cmwaters: You can also just use `.SubmitPayForBlob` and remove the WaitForBlocks and the ConfirmTx part

pkg/user/tx_client_test.go
cmwaters: Wouldn't it be 9 are evicted since only 1 of the 10 initial transactions can fit into a block (and the rest get evicted)

cmwaters: Or is it because two blocks are made in the time it takes to submit all the transactions?

pkg/user/tx_client.go
ninabarbakadze: imo broadcastTxError should only be returned during broadcast

pkg/user/tx_client_test.go
ninabarbakadze: I wanted to split it up before broadcast gives you a response, which I used to extract the transaction hash that I used later to check if it was removed from tx tracker or not.

pkg/user/tx_client_test.go
ninabarbakadze:
```
blockHeight 1
STATUS PENDING
blockHeight 2
STATUS PENDING
blockHeight 3
STATUS EVICTED
blockHeight 4
STATUS EVICTED
blockHeight 5
STATUS EVICTED
blockHeight 6
STATUS EVICTED
blockHeight 7
STATUS EVICTED
blockHeight 8
STATUS EVICTED
blockHeight 9
STATUS EVICTED
blockHeight 10
STATUS EVICTED
```
2 txs make it in initially

pkg/user/tx_client.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/5139/commits/6da59b1bf793a87c70068d1625486a5b80972f25

rootulp (approved):

pkg/user/tx_client.go
cmwaters: So this just doesn't have an error type?

cmwaters: If the transaction were to be rejected straight away we would return a broadcast error. How come if it gets rejected later it's not a broadcast error

ninabarbakadze:
> Looks good! just want to resolve the error type before merging

The problem with broadcast error is that during rejection error code is 0 and log is empty in TxStatus response. This kind of defeats the purpose of it

pkg/user/tx_client_test.go
ninabarbakadze: https://github.com/celestiaorg/celestia-core/issues/2224

PR #5141, opened 2025-07-03
refactor: update E2E tests to use new builder pattern

chatton:
## Overview

This PR updates the existing docker e2e tests to move the configuration of the chain into this repo. The canonical way to configure a celestia-app chain in docker for testing can live here and be imported by node.

We generate raw bytes for the genesis file using the existing functions in this repo, and provide those directly to tastora where they are used along with the provided keyring.

In this PR I just created a few standalone functions that depend on the existing testnode types as that was all that was required. 

based on the second action item in [this doc](https://docs.google.com/document/d/1wr6Ktk0_chjtiT2fQNyYrXgWzo2aodkEiTLPD3rm2X4/edit?tab=t.0)


test/docker-e2e/dockerchain/config.go
```diff
@@ -0,0 +1,90 @@
+package dockerchain
```
chatton, line 1: analogous to testnode, embeds the testnode.Config and extends it with Docker specific config.

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/cometbft/cometbft/privval"
+	servercfg "github.com/cosmos/cosmos-sdk/server/config"
+	"github.com/cosmos/cosmos-sdk/types/module/testutil"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+// NewCelestiaChainBuilder constructs a new ChainBuilder configured for a Celestia instance with predefined parameters.
+func NewCelestiaChainBuilder(t *testing.T, cfg *Config) *tastoradockertypes.ChainBuilder {
```
chatton, line 23: the idea with this function is that it would provide the "default" way that a celestia chain is built, from the node repo, the E2E tests would call this function.

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return tastoradockertypes.NewChainBuilder(t).
+		WithName("celestia"). // just influences home directory on the host.
+		WithChainID(cfg.Genesis.ChainID).
+		WithDockerClient(cfg.DockerClient).
+		WithDockerNetworkID(cfg.DockerNetworkID).
+		WithImage(tastoradockertypes.NewDockerImage(cfg.Image, cfg.Tag, "10001:10001")).
+		WithAdditionalStartArgs("--force-no-bbr", "--grpc.enable", "--grpc.address", "0.0.0.0:9090", "--rpc.grpc_laddr=tcp://0.0.0.0:9099").
+		WithEncodingConfig(&encodingConfig).
+		WithPostInit(getPostInitModifications("0.025utia")...).
```
chatton, line 61: post init functions let us modify anything on the node filesystem before start is called, so basically allows full control over what all the config files look like.

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,119 @@
...
+		WithNodes(vals...).
+		WithGenesis(genesisBz)
+}
+
+// getPostInitModifications returns a slice of functions to modify configuration files of a ChainNode post-initialization.
+func getPostInitModifications(gasPrices string) []func(context.Context, *tastoradockertypes.ChainNode) error {
+	var fns []func(context.Context, *tastoradockertypes.ChainNode) error
+
+	fns = append(fns, func(ctx context.Context, node *tastoradockertypes.ChainNode) error {
+		return config.Modify(ctx, node, "config/config.toml", func(cfg *cometcfg.Config) {
```
chatton, line 71: the new generic function that allows typed modification of arbitrary files.

test/docker-e2e/e2e_simple_test.go
```diff
@@ -72,4 +80,40 @@ func (s *CelestiaTestSuite) TestE2ESimple() {
 			t.Fatalf("Timed out waiting for %d transactions", requiredTxs)
 		}
 	}
+
+	testBankSend(t, celestia, cfg)
+}
+
+// testBankSend performs a basic bank send using txClient.
+func testBankSend(t *testing.T, chain *tastoradockertypes.Chain, cfg *dockerchain.Config) {
```
chatton, line 93: this is kind of just a PoC fn to verify the txClient can broadcast transactions properly. It can be removed in the future as more robust mechanisms are put in place in E2Es.

test/docker-e2e/e2e_test.go
```diff
@@ -40,78 +37,33 @@ type CelestiaTestSuite struct {
...
 		Binds: []string{chain.GetVolumeName() + ":/celestia-home"},
 	}
 
+	hostGRPCPort := strings.Split(chain.GetGRPCAddress(), ":")[1]
+
 	args := []string{
 		"/bin/txsim",
 		"--key-path", "/celestia-home",
-		"--grpc-endpoint", chain.GetGRPCAddress(),
+		"--grpc-endpoint", "host.docker.internal:" + hostGRPCPort,
```
chatton, line 66: not 100% sure exactly what was going on here, but the host name wasn't resolving correctly without using the special DNS entry.

test/docker-e2e/dockerchain/config.go
```diff
@@ -0,0 +1,90 @@
...
+	Image           string
+	Tag             string
+	DockerClient    *client.Client
+	DockerNetworkID string
+}
+
+// DefaultConfig returns a configured instance of Config with a custom genesis and validators.
+func DefaultConfig(client *client.Client, network string) *Config {
+	tnCfg := testnode.DefaultConfig()
+	// default + 2 extra validators.
```
chatton, line 28: maybe we want the default to be 1?  Doesn't seem to effect runtime too dramatically

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,145 @@
...
+	"testing"
+	"time"
+)
+
+// NewCelestiaChainBuilder constructs a new ChainBuilder configured for a Celestia instance with predefined parameters.
+func NewCelestiaChainBuilder(t *testing.T, cfg *Config) *tastoradockertypes.ChainBuilder {
+	genesisBz, err := cfg.Genesis.ExportBytes()
+	require.NoError(t, err, "failed to export genesis bytes")
+
+	// TODO: why do I need to do this?
```
chatton, line 27: I think something is broken with the bytes that are exported currently

test/docker-e2e/dockerchain/config.go
rootulp: I don't have a strong preference. 1 validator seems fine for most tests

test/docker-e2e/dockerchain/config.go
```diff
@@ -0,0 +1,90 @@
...
+	tnCfg.Genesis = tnCfg.Genesis.
+		WithChainID("test").
```
rootulp: [optional] import this from https://github.com/rootulp/celestia-app/blob/bf41b7dd886ee1d6c7e035ced3ac9fad6eea0224/pkg/appconsts/chain_ids.go#L8

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,145 @@
...
+	// fails with 2025-07-03 14:57:42 Error: failed to get current app state: failed to determine genesis version
```
rootulp, line 28: this sounds like this multiplexer utility: https://github.com/celestiaorg/celestia-app/blob/6b7fb8b684d00b91fce51c83eaf2967171874990/multiplexer/internal/utils.go#L69

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,145 @@
...
+			Build()
+	}
+
+	return tastoradockertypes.NewChainBuilder(t).
+		WithName("celestia"). // just influences home directory on the host.
+		WithChainID(cfg.Genesis.ChainID).
+		WithDockerClient(cfg.DockerClient).
+		WithDockerNetworkID(cfg.DockerNetworkID).
+		WithImage(tastoradockertypes.NewDockerImage(cfg.Image, cfg.Tag, "10001:10001")).
+		WithAdditionalStartArgs("--force-no-bbr", "--grpc.enable", "--grpc.address", "0.0.0.0:9090", "--rpc.grpc_laddr=tcp://0.0.0.0:9099").
```
rootulp: [not blocking] Besides `--force-no-bbr` are the other flags configurable via config.toml or app.toml? That _seems_ cleaner.

rootulp: Ahh it seems possible with the `getPostInitModifications` function below

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,145 @@
...
+		ctx,
+		cfg.Genesis.Keyring(),
+		cn.GrpcConn,
+		encCfg,
+	)
+}
+
+// NodeConfigBuilders returns a list of ChainNodeConfigBuilder and any error if one occurs.
+// this function populates a default set of ChainNodeConfigBuilder based on the provided config.
+// this handles the population of private key bytes, record name and keyring.
```
rootulp:
```suggestion
// This function populates a default set of ChainNodeConfigBuilder based on the provided config.
// This handles the population of private key bytes, record name and keyring.
```

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -0,0 +1,145 @@
...
+//
+// If any custom modifications are required to any individual validator, this function can be used
+// and can be used to modify the ChainNodeConfigBuilder.
```
rootulp:
```suggestion
// If any custom modifications are required to any individual validator, this function can be used
// to modify the ChainNodeConfigBuilder.
```

test/docker-e2e/dockerchain/config.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"github.com/celestiaorg/celestia-app/v5/test/util/genesis"
+	"github.com/celestiaorg/celestia-app/v5/test/util/testnode"
+	"github.com/moby/moby/client"
+	"os"
+)
+
+const (
+	multiplexerImage   = "ghcr.io/celestiaorg/celestia-app"
+	defaultCelestiaTag = "v4.0.7-mocha"
```
rootulp, line 13: [follow up] this is definitely fine for now but it may be nice to automatically bump this version tag as we cut more releases. Currently there are a lot of hard-coded v3.x versions in the codebase which is a maintenance burden when we're cutting releases so I'm worried about this hard-coded version becoming similar.

test/docker-e2e/dockerchain/testchain.go
chatton: is it failing 'cause the genesis bytes are incorrect or the multiplexer is looking for the wrong thing? I can look into this further.

test/docker-e2e/dockerchain/testchain.go
chatton: yeah good point, we can actually just put anything in the default config override, these were hanging around from before.

test/docker-e2e/dockerchain/config.go
chatton: yeah for sure, in the tests maybe we don't actually even want a default 🤔 , we will likely always be testing a specific version. The default might just be "the one built locally"

test/util/genesis/accounts.go
```diff
@@ -143,3 +145,21 @@ func (v *Validator) GenTx(ecfg encoding.Config, kr keyring.Keyring, chainID stri
 
 	return txBuilder.GetTx(), nil
 }
+
+// PrivateKeyBytes returns the contents of the priv_validator_key.json file.
+func (v *Validator) PrivateKeyBytes() ([]byte, error) {
```
evan-forbes: not blocking, optional

At first glance, `PrivateKeyBytes` would make me assume that these are the raw bytes of the private key, but as stated in the comment this is a special json format that also includes the address and the public key. therefore, imo a name that distinguishes between the two would be slightly more optimal

evan-forbes (approved):
:+1:

test/util/genesis/accounts.go
chatton: maybe we can actually just return the struct directly at the test can marshal it

Issue #5082, opened 2025-06-27
Weird TestSquareSizeIntegrationTest flake


evan-forbes:
noticed a new flake

```sh
--- FAIL: TestSquareSizeIntegrationTest (50.90s)
    square_size_test.go:47: setting up square size integration test
    buffer.go:261: server.go:88: <nil> INF starting gRPC server... address=127.0.0.1:36635
    --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound (48.03s)
        --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/default (3.29s)
            square_size_test.go:172: 
                	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:172
                	            				/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:110
                	Error:      	Received unexpected error:
                	            	rpc error: code = NotFound desc = tx not found: E6D84E119DD26E80AB17DFB783F763FEA0363248F496B21A1761C42C0ED69E01
                	Test:       	TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/default
        --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/max_bytes_constrains_square_size (8.27s)
            square_size_test.go:181: 
                	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:181
                	            				/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:110
[...]

Issue #3585, opened 2024-06-18
Flakiness in TestSquareSizeIntegrationTest


staheri14:
# Problem
`TestSquareSizeIntegrationTest` is flakey.
[Link](https://github.com/celestiaorg/celestia-app/actions/runs/9569869274/job/26383358195?pr=3583) to the logs.
```
--- FAIL: TestSquareSizeIntegrationTest (61.04s)
    square_size_test.go:46: setting up square size integration test
    --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound (49.82s)
        --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/gov_square_size_==_hardcoded_max (16.26s)
            square_size_test.go:137: 
                	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:1[37](https://github.com/celestiaorg/celestia-app/actions/runs/9569869274/job/26383358195?pr=3583#step:4:38)
                	Error:      	Not equal: 
                	            	expected: 128
                	            	actual  : 64
                	Test:       	TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/gov_square_size_==_hardcoded_max
    network.go:46: tearing down testnode
```

# Acceptance Criteria
- [ ] Resolve the flakiness.

staheri14:
Another instance of flakiness if this test https://github.com/celestiaorg/celestia-app/actions/runs/9584852073/job/26429395522?pr=3591

rootulp:
One idea to resolve the flakiness is to refactor the test to use testapp instead of a testnode. That way the test can invoke `BeginBlock`, `DeliverTx` with enough txs to fill a 128 x 128 square, `EndBlock` and make assertions. IMO the flakiness is caused by the current test using a testnode which runs txsim. Since txsim can't control which blocks its transactions get included in, there is no guarantee that enough txs land in one block to cause a 128 x 128 square.

rootulp:
This flake has been fixed by multiple commits:
1. **632577eab** - Fixed race conditions in parameter changes and block validation.
2. **a14a3f0bd** (#6504) - Fixed a block fetching bug (loop using `&latestHeight` instead of `&i`) and timing issue (checking blocks from wrong window). Captured start/end heights around the wait period and added proper block wait after parameter confirmation.

The test is currently stable and passing.

PR #5169, opened 2025-07-07
test: split up go test into 4 separate jobs

chatton:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,9 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        part: ["aa", "ab", "ac", "ad"]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
 
@@ -25,11 +28,24 @@ jobs:
         with:
           go-version-file: "go.mod"
 
+      - name: Create package list
+        run: go list ./... > pkgs.txt
+
+      - name: Split packages into parts
+        run: split -n l/4 pkgs.txt pkgs.txt.part.
+
       - name: Run tests
-        run: make test
+        run: |
+          # run 1/4 of the tests in each job.
+          export PACKAGES=$(cat pkgs.txt.part.${{ matrix.part }} | tr '\n' ' ')
+          echo "Running tests for packages: $PACKAGES"
+          make test PACKAGES="$PACKAGES"
 
   test-coverage:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        part: ["aa", "ab", "ac", "ad"]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
 
@@ -37,13 +53,23 @@ jobs:
         with:
           go-version-file: "go.mod"
 
-      - name: Generate coverage.txt
-        run: make test-coverage
+      - name: Create package list
+        run: go list ./... > pkgs.txt
 
-      - name: Upload coverage.txt
+      - name: Split packages into parts
+        run: split -n l/4 pkgs.txt pkgs.txt.part.
+
+      - name: Generate coverage for part ${{ matrix.part }}
+        run: |
+          export PACKAGES=$(cat pkgs.txt.part.${{ matrix.part }} | tr '\n' ' ')
+          echo "Running coverage for packages: $PACKAGES"
+          go test -timeout 30m -coverprofile=coverage-${{ matrix.part }}.out -covermode=atomic $PACKAGES
+
+      - name: Upload coverage part ${{ matrix.part }}
         uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 #v5.4.3
         with:
-          files: ./coverage.txt
+          files: ./coverage-${{ matrix.part }}.out
+          flags: part-${{ matrix.part }}
 
   test-race:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,11 @@ lint-fix: fmt
 ## test: Run tests.
 test:
 	@echo "--> Running tests"
-	@go test -timeout 30m ./...
+	@if [ -z "$(PACKAGES)" ]; then \
+		go test -timeout 30m ./...; \
+	else \
+		go test -timeout 30m $(PACKAGES); \
+	fi
 .PHONY: test
 
 ## test-short: Run tests in short mode.


chatton:
this approach helps, but there are still some packages that just take a long time to test

- github.com/celestiaorg/celestia-app/v4/app/test	489.587s
- github.com/celestiaorg/celestia-app/v4/test/txsim	193.260s
- github.com/celestiaorg/celestia-app/v4/pkg/user	144.728s

We can look into why that is the case

rootulp:
- Can you please push this to a fork to verify that it decreases overall test runtime? I'm only in favor of splitting if it significantly reduces test run time because it increases complexity, number of required checks to enable / disable in GitHub settings, number of workflows that need to be retried when a test flakes, etc. Also now `make test` is not representative of what runs in CI so it can be more difficult to repro a CI failure on a maintainer's local machine.
-  Previously celestia-app split test files but it wasn't significantly faster than consolidating them. See: 
    - https://github.com/celestiaorg/celestia-app/issues/1439 
    - https://github.com/celestiaorg/celestia-app/pull/1798
- One alternative idea is to split `make test` into two `make test` and `make test-flaky` groups, one that runs the 100% reliable tests and one that runs tests that sometimes flake.

chatton:
@rootulp yeah after running it a few times I see that the long running app/test package still eats up a lot of time, I'll try and figure out something to actually reduce time without making things convoluted 👍

.github/workflows/test.yml
```diff
@@ -80,6 +118,7 @@ jobs:
 
       - name: Run multiplexer tests
         run: make test-multiplexer
+        timeout-minutes: 15
```
chatton: accidental inclusion will remove

.github/workflows/test.yml
```diff
@@ -26,7 +41,30 @@ jobs:
...
+            make test PACKAGES="$(go list ./... | grep -v 'app/test$' | grep -v 'test/txsim$' | grep -v 'pkg/user$' | tr '\n' ' ')"
+          elif [ -n "${{ matrix.pattern }}" ]; then
+            # run specific test patterns
+            go test -timeout 30m -run "${{ matrix.pattern }}" ${{ matrix.package }}
+          else
+            # run specific package
+            make test PACKAGES="${{ matrix.package }}"
+          fi
+
+  test:
```
chatton, line 56: trick to make one required task depend on a matrix job

.github/workflows/test.yml
```diff
@@ -16,8 +16,23 @@ jobs:
         run: make test-short
         timeout-minutes: 10
 
-  test:
+  go-test:
```
rootulp, line 19: [question] what was the motivation to rename this to `go-test`?

.github/workflows/test.yml
```diff
@@ -16,8 +16,23 @@ jobs:
...
+          - package: "unit"
+          # split the heaviest ones
+          - package: "github.com/celestiaorg/celestia-app/v5/app/test"
+            pattern: "TestIntegration.*"
+          - package: "github.com/celestiaorg/celestia-app/v5/app/test"
+            pattern: "TestBigBlob.*|TestStandardSDK.*"
+          - package: "github.com/celestiaorg/celestia-app/v5/app/test"
+            pattern: "TestSquareSize.*|TestPriority.*"
+          - package: "github.com/celestiaorg/celestia-app/v5/test/txsim"
+          - package: "github.com/celestiaorg/celestia-app/v5/pkg/user"
```
rootulp, line 35: I'm concerned that the tests run in CI will now behave slightly differently from running them locally via `make test` which could lead to failures that are only reproducible in one environment and not the other.

Is it possible to split `make test` into multiple commands and then have different CI workers run the different `make test` commands? For example:

```
make test <- runs all unit tests
make test-integration <- runs all tests that use a testnode
```

rootulp: Another option is to delete the tests: `TestIntegration`, `TestBigBlob`, `TestStandardSDK`, `TestSquareSize`, `TestPriority` and re-write them without using a testnode (or whichever component is taking a long time to run).

FWIW those tests also seem to be the most flakey

rootulp (commented):
Thanks for working on this. The approach here seems sound.

I'd rather we rewrite or remove the tests that take a long time to run b/c I don't think they're worth maintaining in their current state. 

I'm not strictly opposed to the solution in this PR but I'm concerned it may lead to tougher to locally replicate test failures. Also I would prefer a simple CI workflow that doesn't use a matrix but that's personal preference. If other maintainers feel strongly, can merge this as-is.

.github/workflows/test.yml
rootulp: Oh disregard. Likely because there is a new `test` block below

.github/workflows/test.yml
chatton: If these tests are flakey / unreliable what we could try doing is experimenting with converting the flakey ones into full E2Es and see if that improves things

We would then have the reliable tests using in memory nodes and unit tests which run quickly running under `make test`.

If there is still an issue in run time, I like your suggestion of a separate make target, potentially something like 

`make test-integration GROUP=1` and use this locally and in CI to ensure the same batch of tests is being run without needing to manually think about which subset of tests to choose.

Taking a look at a few tests, I don't think there's anything stopping us from converting 
 - TestBlobExceedsMaxTxSize
 - TestSquareSizeUpperBound
 - TestStandardSDK

And plenty others, into E2E tests with Tastora. And ensuring sufficient unit tests on the underlying functions that get called in `make test` if any are missing.

Whether or not these tests as E2Es is valuable is another question, but I figured I would throw it out there!

evan-forbes: > Another option is to delete the tests

I'm not inherently opposed, but fwiw in the past some of the integration tests were added because some basic functionality regressed, such as vesting module stuff in the sdk, or delegating one time iirc

if we limit the scope of those tests more, ofc that would also be good

> If these tests are flakey / unreliable what we could try doing is experimenting with converting the flakey ones into full E2Es and see if that improves things. ... Whether or not these tests as E2Es is valuable is another question, but I figured I would throw it out there!

again, not inherently against this however I am generally skeptical of moving things to a different framework etc being a permanent solution, as it feels like we're just avoiding the problem rather than solving it. However, I definitely could be missing the reasons for testnode is so flakey and taking forever! in theory, aren't we running the same code w/ each if we have a single validator?

perhaps we could also spend some cycles making those faster and more reliable? so far, we have had decent luck with making it more reliable. 

I haven't check testnode's block times recently, its possible there was a reversion somewhere and block times increased or something

chatton: Definitely agree with the point about jumping to a different framework not really solving anything long term. 

If people are okay with this PR as a temporary solution to get CI a little faster, we could then dedicate some time to dig into some of the testnode tests and see if we can make the faster/more reliable

We can make more intentional decisions about which tests should be docker E2Es, and which should use testnode.

rootulp: Confirmed they are flakey:
- https://github.com/celestiaorg/celestia-app/issues/2942
- https://github.com/celestiaorg/celestia-app/issues/3585
- https://github.com/celestiaorg/celestia-app/issues/5082
- https://github.com/celestiaorg/celestia-app/issues/5198

I'm on board with converting them to E2E tests if that is less flakey. TBH I don't think we even need all of these to be converted, some can likely be deleted. 

IMO next steps would be to:
- Investigate all tests that use testnode to understand what each test is actually meant to cover.
- Refactor as much as possible of those tests into unit tests.
- Evaluate if the remaining still needs tests and if yes, convert to e2e tastora tests.

This seems like a considerable amount of effort and doesn't immediately resolve the long test time in CI so as a shorter term solution I'm okay with merging this PR or adding `t.Skip()` to the tests that take too long

evan-forbes (approved):

PR #5172, opened 2025-07-07
feat: support the talis reset command

rach-id:
## Overview

Supports resetting the nodes to redeploy the network without having to tear down the DO instances


rootulp (approved):

tools/talis/reset.go
```diff
@@ -0,0 +1,82 @@
...
+
+			resolvedKey := resolveValue(SSHKeyPath, EnvVarSSHKeyPath, strings.ReplaceAll(cfg.SSHPubKeyPath, ".pub", ""))
+
+			// Filter validators if specific ones were requested
+			targetValidators := cfg.Validators
+			if len(validators) > 0 {
+				targetValidators = make([]Instance, 0)
+				for _, v := range cfg.Validators {
+					for _, requested := range validators {
+						if strings.Contains(v.Name, requested) {
```
evan-forbes, line 43: I know we currently don't support multiple tests as we don't add the chain-id due to a bug, but it would good imo not to build on that bug here by also checking the chain-id

evan-forbes (commented):
Nice!

tools/talis/reset.go
rach-id: Once we support multiple experiments, we can change this to follow a pattern: `validator-index-chainID` and not accept anything else. But before that, adding this here will only block this. I'll reference this comment in the issue and the copilot PR to make the changes

PR #5186, opened 2025-07-08
fix: incorrect assertion in pruning test

sky-coderay:
Fixed a logical error in the pruning test where the assertion was comparing the number of pruned transactions with the difference between initial count and pruned count. The correct assertion should compare the number of transactions that should remain after pruning with the actual remaining count.

### DIFF
--- a/pkg/user/pruning_test.go
+++ b/pkg/user/pruning_test.go
@@ -45,6 +45,6 @@ func TestPruningInTxTracker(t *testing.T) {
 	txClient.pruneTxTracker()
 	// Prunes the transactions that are 10 minutes old
 	// 5 transactions will be pruned
-	require.Equal(t, txsToBePruned, txTrackerBeforePruning-txsToBePruned)
+	require.Equal(t, txsNotReadyToBePruned, txTrackerBeforePruning-txsToBePruned)
 	require.Equal(t, len(txClient.txTracker), txsNotReadyToBePruned)
 }


pkg/user/pruning_test.go
```diff
@@ -45,6 +45,6 @@ func TestPruningInTxTracker(t *testing.T) {
 	txClient.pruneTxTracker()
 	// Prunes the transactions that are 10 minutes old
 	// 5 transactions will be pruned
-	require.Equal(t, txsToBePruned, txTrackerBeforePruning-txsToBePruned)
+	require.Equal(t, txsNotReadyToBePruned, txTrackerBeforePruning-txsToBePruned)
```
rach-id, line 48: hum, why wasn't this failing before? if they're equal, we should update the test to have different numbers to make sure we're not missing anything

futreall: the test wasn't failing because txsToBePruned = txsNotReadyToBePruned = 5, and txTrackerBeforePruning-txsToBePruned = 10-5 = 5
I suggest fixing the assertion and using different values (like 7 and 3) for more robust testing

rach-id: good idea, can you do that please?

futreall: sure

sky-coderay: I'm late:):) thanks for the help, I'll close the PR then

Issue #5175, opened 2025-07-07
Modify min gas price in the app.toml to at least 0.004 on startup of the new release


cmwaters:
(only if it was 0.002 - the previous default).

rootulp:
Instead of the binary modifying the config file directly can we:
1. Modify the default for newly generated `app.toml` files
2. Ask node operators to update their `app.toml` in the v5 release notes

PR #5187, opened 2025-07-08
feat: update min gas price and add a prehook to update the default

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/5175

I have added prehooks which can be run prior to the start command. A new pre hook reads the min gas price. If it is the same as the old default, it updates it, else it does nothing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	github.com/tendermint/tendermint v0.0.0-00010101000000-000000000000
 	golang.org/x/exp v0.0.0-20250305212735-054e65f0b394
 	golang.org/x/oauth2 v0.28.0
-	golang.org/x/sync v0.15.0
+	golang.org/x/sync v0.16.0
 	google.golang.org/genproto/googleapis/api v0.0.0-20250324211829-b45e905df463
 	google.golang.org/grpc v1.73.0
 	google.golang.org/protobuf v1.36.6
--- a/go.sum
+++ b/go.sum
@@ -1379,8 +1379,8 @@ golang.org/x/sync v0.0.0-20210220032951-036812b2e83c/go.mod h1:RxMgew5VJxzue5/jJ
 golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20220929204114-8fcdb60fdcc0/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
-golang.org/x/sync v0.15.0 h1:KWH3jNZsfyT6xfAfKiz6MRNmd46ByHDYaZ7KSkCtdW8=
-golang.org/x/sync v0.15.0/go.mod h1:1dzgHSNfp02xaA81J2MS99Qcpr2w7fw1gpm99rleRqA=
+golang.org/x/sync v0.16.0 h1:ycBJEhp9p4vXvUZNszeOq0kGTPghopOL8q0fq3vstxw=
+golang.org/x/sync v0.16.0/go.mod h1:1dzgHSNfp02xaA81J2MS99Qcpr2w7fw1gpm99rleRqA=
 golang.org/x/sys v0.0.0-20180823144017-11551d06cbcc/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20180830151530-49385e6e1522/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20180905080454-ebe1bf3edb33/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
--- a/test/docker-e2e/go.mod
+++ b/test/docker-e2e/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/celestiaorg/tastora v0.1.1
 	github.com/cometbft/cometbft v0.38.17
 	github.com/cosmos/cosmos-sdk v0.50.13
-	github.com/docker/docker v28.3.0+incompatible
+	github.com/docker/docker v28.3.2+incompatible
 	github.com/moby/moby v27.5.1+incompatible
 	github.com/stretchr/testify v1.10.0
 	go.uber.org/zap v1.27.0
@@ -249,7 +249,7 @@ require (
 	golang.org/x/exp v0.0.0-20250305212735-054e65f0b394 // indirect
 	golang.org/x/net v0.38.0 // indirect
 	golang.org/x/oauth2 v0.28.0 // indirect
-	golang.org/x/sync v0.15.0 // indirect
+	golang.org/x/sync v0.16.0 // indirect
 	golang.org/x/sys v0.31.0 // indirect
 	golang.org/x/term v0.30.0 // indirect
 	golang.org/x/text v0.23.0 // indirect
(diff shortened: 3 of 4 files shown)


cmd/celestia-appd/cmd/gas_price_updater.go
```diff
@@ -0,0 +1,70 @@
...
+
+		// Write the updated configuration back to the file
+		if err := viper.WriteConfig(); err != nil {
+			return fmt.Errorf("failed to write updated app.toml: %w", err)
+		}
+
+		// Log the update
+		logMessage := fmt.Sprintf("Updated minimum gas prices in %s: %s -> %s",
+			appConfigPath, oldDefaultMinGasPrice, minGasPrice)
+		fmt.Println(logMessage)
```
evan-forbes: Not blocking

Do we have access to some logger to keep the messages looking consistent

cmd/celestia-appd/cmd/pre_start_hooks.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+type PreStartHook func(cmd *cobra.Command) error
+
+// addPreStartHooks finds the start command and adds pre-start hooks using Cobra's PreRunE
+func addPreStartHooks(rootCmd *cobra.Command, hooks ...PreStartHook) error {
```
evan-forbes, line 14: not blocking opinion

imo would be slightly more explicit if we could make the point that we're modifying things clearer

so simply returning the thing we're modifying even though we don't need to due to them being pointers

evan-forbes (approved):

cmd/celestia-appd/cmd/gas_price_updater.go
```diff
@@ -0,0 +1,70 @@
...
+// are set to the old default value and updates them if necessary.
+func checkAndUpdateMinGasPrices(cmd *cobra.Command) error {
+	// Get the home directory from the command
+	homeDir, err := cmd.Flags().GetString(flags.FlagHome)
+	if err != nil {
+		return fmt.Errorf("failed to get home directory: %w", err)
+	}
+	if homeDir == "" {
+		homeDir = app.NodeHome
+	}
```
rootulp: [optional] this package imports app which has NodeHome so I think it should be possible to remove these lines and instead use `NodeHome` directly

Ref: https://github.com/celestiaorg/celestia-app/blob/dcded42f669298c549b10d31bf31a364cbe240a4/app/init.go#L22-L24

cmd/celestia-appd/cmd/gas_price_updater_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckAndUpdateMinGasPrices(t *testing.T) {
```
rootulp, line 16: Great unit test 👍

cmd/celestia-appd/cmd/gas_price_updater.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/celestiaorg/celestia-app/v5/app"
+	"github.com/celestiaorg/celestia-app/v5/app/params"
+	"github.com/celestiaorg/celestia-app/v5/pkg/appconsts"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const (
+	// Old default minimum gas price that needs to be updated
```
rach-id:
```suggestion
	// oldDefaultMinGasPrice Old default minimum gas price that needs to be updated
```

cmd/celestia-appd/cmd/gas_price_updater.go
```diff
@@ -0,0 +1,70 @@
+package cmd
```
rach-id, line 1: is this the direction we're taking? updating all the fields, that need changes, using the binary instead of asking validator operators to run a `sed`?

cmd/celestia-appd/cmd/gas_price_updater.go
```diff
@@ -0,0 +1,70 @@
...
+	viper.SetConfigFile(appConfigPath)
+	if err := viper.ReadInConfig(); err != nil {
+		return fmt.Errorf("failed to read app.toml: %w", err)
+	}
+
+	// Get the current minimum gas prices
+	currentMinGasPrices := viper.GetString("minimum-gas-prices")
+
+	// Check if it matches the old default
+	if strings.TrimSpace(currentMinGasPrices) == oldDefaultMinGasPrice {
```
rach-id, line 42: what about a validator that manually updated his gas price to `0.0025` or any number between `0.002` and `0.004`? maybe we should in here for any number in this range and change it

cmd/celestia-appd/cmd/root.go
```diff
@@ -129,20 +129,9 @@ func initRootCommand(rootCommand *cobra.Command, capp *app.App) {
...
-
-	// Add the BBR check to the start command
-	startCmd.RunE = func(cmd *cobra.Command, args []string) error {
-		if err := checkBBR(cmd); err != nil {
-			return err
-		}
-
-		return startCmdRunE(cmd, args)
+	// Add hooks run prior to the start command
+	if err := addPreStartHooks(rootCommand, checkAndUpdateMinGasPrices, checkBBR); err != nil {
```
rootulp, line 133: Thanks for the refactor, seems cleaner than it was before. 

I was going to suggest adding these pre start hooks to the `start` command directly but unfortunately the start command gets added to the root command via code in Cosmos SDK so we'd have to make the modification in celestiaorg/cosmos-sdk which isn't as nice as making the change in celestia-app.

The other [extremely optional] idea is to copy the start command from Cosmos SDK to celestia-app and then add the PreStartHooks directly in that file (e.g. `start.go`) but this has other downsides.

Ref: https://github.com/celestiaorg/cosmos-sdk/blob/5259747ebf054c2148032202d946945a2d1896c7/server/start.go#L147

rach-id (commented):
[nit]
seems like too many code comments. might be cleaner to keep only relevant ones.

rootulp (approved):

cmd/celestia-appd/cmd/root.go
cmwaters: Yeah there was a time I wanted to bring the start (and init) commands over to celestia-app not just for this but to also make the start up function consistent in our integration tests

cmd/celestia-appd/cmd/gas_price_updater.go
cmwaters: As I'm not a fan of changing people's config, my rationale was to keep the surface area of the change as minimal as possible. Thus I only wanted to modify it if it was exactly the default. If it's 0.0025, it means the node operator changed their config, since this is more of a soft way of raising the gas price, I didn't want to interfere there. It's possible that people also have their gas price set to 0.001. If we want to raise the floor we need to increase the network wide minimum gas price

rach-id: > possible that people also have their gas price set to 0.001

if they set it below the network minimum 0.002, it means they're accepting in the mempool transactions that can't be included in any block, right?

cmd/celestia-appd/cmd/gas_price_updater.go
cmwaters: Mus proposed having a binary change to update it (I gather under the assumption that operators have a tendency to bump binaries more than modifying the config)

cmd/celestia-appd/cmd/gas_price_updater.go
cmwaters: makes sense. I'll see if I can change it

cmd/celestia-appd/cmd/pre_start_hooks.go
cmwaters: I was leaning on the word "add" to imply that it was modifying the command. I can also return the command struct but I'm not sure what use that would be

cmd/celestia-appd/cmd/gas_price_updater.go
cmwaters: The network minimum is 0.000001

evan-forbes (approved):

cmwaters:
Need to fix these tests - not sure how they broke

rootulp (approved):
Still LGTM

PR #5188, opened 2025-07-08
fix: pruning test assertion and improve test robustness

futreall:
ref: https://github.com/celestiaorg/celestia-app/pull/5186

## Overview

Fixed a logical error in the pruning test where the assertion was incorrectly comparing the number of pruned transactions with the difference between initial count and pruned count. The correct assertion should compare the number of transactions that should remain after pruning with the actual remaining count.

Additionally, updated the test to use different values for pruned (7) and remaining (3) transactions instead of equal values (5 and 5) to make the test more robust and prevent accidental passing due to value coincidence.

### DIFF
--- a/pkg/user/pruning_test.go
+++ b/pkg/user/pruning_test.go
@@ -18,8 +18,8 @@ func TestPruningInTxTracker(t *testing.T) {
 	var txsToBePruned int
 	var txsNotReadyToBePruned int
 	for i := 0; i < numTransactions; i++ {
-		// 5 transactions will be pruned
-		if i%2 == 0 {
+		// 7 transactions will be pruned
+		if i < 7 {
 			txClient.txTracker["tx"+fmt.Sprint(i)] = txInfo{
 				signer:   "signer" + fmt.Sprint(i),
 				sequence: uint64(i),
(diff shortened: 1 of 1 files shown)


rootulp:
What issue does this close? The link the PR description is to a closed PR.

futreall:
Hey @rootulp you can see the comments of the closed issue, https://github.com/celestiaorg/celestia-app/pull/5186#discussion_r2192355222, if you still have questions, I will answer, thanks

rootulp:
If it wasn't failing before, why do we need to change the numbers? This change doesn't seem necessary to me so closing for now.

If you're interested in contributing, please tackle a [good first issue](https://github.com/celestiaorg/celestia-app/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22) or create an issue to get feedback from maintainers before creating a PR.

futreall:
I think there's still a real bug in the assertion logic - it's comparing txsToBePruned with txTrackerBeforePruning-txsToBePruned when it should compare txsNotReadyToBePruned. The test only passes because both happen to be 5. Happy to drop the number changes if you prefer, but the assertion itself needs fixing

rach-id (approved):

Issue #4808, opened 2025-05-14
Can't state sync Arabica


rootulp:
## Context

I'm trying to verify state sync works on Arabica. It fails due to

```
10:13PM ERR Can't verify err="failed to obtain the header at height #5931091: context deadline exceeded" module=light
```

## Problem

```
./scripts/arabica.sh

celestia-app home: /home/rootulp/.celestia-app
celestia-app version: 4.0.0-rc6-4-g376725ac

Are you sure you want to delete: /home/rootulp/.celestia-app? [y/n] y
Deleting /home/rootulp/.celestia-app...
Initializing config files...
Setting seeds in config.toml...
Latest height: 5933233
Block height: 5931233
Trust hash: 874167D50A4D2B0FDFF0017FEE285C740007B06AF92F725FE739DD3E4319EC65
Enabling state sync in config.toml...
Downloading genesis file...
Downloading genesis file for arabica-11 to /home/rootulp/.celestia-app/config/genesis.json
Downloaded genesis file for arabica-11 to /home/rootulp/.celestia-app/config/genesis.json
SHA-256 hash verified for arabica-11
Starting celestia-appd...
10:12PM INF initializing multiplexer app_version=1 chain_id=arabica-11 module=server
10:12PM INF initialized remote app client address=127.0.0.1:36658 module=server
10:12PM INF starting comet node module=server
10:12PM INF service start impl=multiAppConn module=proxy msg="Starting multiAppConn service"
10:12PM INF service start connection=query impl=localClient module=abci-client msg="Starting localClient service"
10:12PM INF service start connection=snapshot impl=localClient module=abci-client msg="Starting localClient service"
10:12PM INF service start connection=mempool impl=localClient module=abci-client msg="Starting localClient service"
10:12PM INF service start connection=consensus impl=localClient module=abci-client msg="Starting localClient service"
10:12PM INF service start impl=EventBus module=events msg="Starting EventBus service"
10:12PM INF service start impl=PubSub module=pubsub msg="Starting PubSub service"
10:12PM INF Version info abci=2.0.0 block=11 commit_hash= module=server p2p=8 tendermint_version=0.38.17
10:12PM INF This node is not a validator addr=B0B33F509A55A67441644C26485C7C41906EE621 module=consensus pubKey=PubKeyEd25519{E960AB9FAA3CECE0C92D31A3D1477ADF5D824C8B073D4932E1AA386ECB69FB95}
[...]

PR #4807, opened 2025-05-14
fix: state sync

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4806

Issue #5098, opened 2025-06-27
Can't state sync on Mocha


rootulp:
## Context

I'm trying to reproduce another issue.

## Problem

I can't state sync on Mocha. I see logs of the form

```
2:04PM ERR rejecting invalid snapshot err="snapshot has 0 chunk hashes, but 25 chunks: invalid snapshot metadata" format=2 height=6855000
2:04PM INF Snapshot rejected format=2 hash=C9448BD4F20CFB7B9D8D8288D0BEBF663A7DCBF8E1FFF35046DEB3E10AC7A47B height=6855000 module=statesync
2:04PM INF VerifyHeader hash=414022AEFCD63D714FEA8846FCD3AB0D44B54242F86A7B3AB55EC065CDFE4F58 height=6850001 module=light
2:04PM INF error from light block request from primary, removing... error="error in json rpc client, with http response metadata: (Status: 429 Too Many Requests, Protocol HTTP/1.1). error unmarshalling: unexpected end of JSON input" height=6854955 module=light primary=http{https://celestia-testnet-rpc.itrocket.net:443}
2:04PM ERR Can't verify err="failed to obtain the header at height #6854955: no witnesses connected. please reset light client" module=light
2:04PM INF failed to fetch and verify app hash err="failed to obtain the header at height #6854955: no witnesses connected. please reset light client" module=statesync
2:04PM INF Snapshot rejected format=2 hash=6CF7F297AEF29849C116FA631C45B124BFC4E91626B751B6F7B2E883678D4AF9 height=6850000 module=statesync
2:04PM INF VerifyHeader hash=D8C323B770B8432E26CEB9B755B945ED99A2967C9C48C98844B07DCD41A35B4A height=4514501 module=light
2:04PM INF error from light block request from primary, removing... error="error in json rpc client, with http response metadata: (Status: 429 Too Many Requests, Protocol HTTP/1.1). error unmarshalling: unexpected end of JSON input" height=6854955 module=light primary=http{https://celestia-testnet-rpc.itrocket.net:443}
2:04PM ERR error on light block request from witness, removing... error="error in json rpc client, with http response metadata: (Status: 429 Too Many Requests, Protocol HTTP/1.1). error unmarshalling: unexpected end of JSON input" module=light primary=http{https://celestia-testnet-rpc.itrocket.net:443}
2:04PM ERR failed to remove witnesses err="no witnesses connected. please reset light client" module=light witnessesToRemove=[0]
[...]

PR #5193, opened 2025-07-09
fix: use the app version from the offer snapshot request

evan-forbes:
## Overview

uses the app version passed by comet when validating a snapshot to use

closes #4807 #5098 likely #4808

### DIFF
--- a/multiplexer/abci/abci.go
+++ b/multiplexer/abci/abci.go
@@ -117,6 +117,15 @@ func (m *Multiplexer) LoadSnapshotChunk(_ context.Context, req *abci.RequestLoad
 }
 
 func (m *Multiplexer) OfferSnapshot(_ context.Context, req *abci.RequestOfferSnapshot) (*abci.ResponseOfferSnapshot, error) {
+	_, err := m.versions.GetForAppVersion(m.appVersion)
+	if err != nil {
+		return nil, fmt.Errorf("no app version for snapshot: version %d", req.AppVersion)
+	}
+
+	m.mu.Lock()
+	m.appVersion = req.AppVersion
+	m.mu.Lock()
+
 	app, err := m.getApp()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get app for version %d: %w", m.appVersion, err)


multiplexer/abci/abci.go
```diff
@@ -117,6 +117,15 @@ func (m *Multiplexer) LoadSnapshotChunk(_ context.Context, req *abci.RequestLoad
...
 
 func (m *Multiplexer) OfferSnapshot(_ context.Context, req *abci.RequestOfferSnapshot) (*abci.ResponseOfferSnapshot, error) {
+	_, err := m.versions.GetForAppVersion(req.AppVersion)
+	if err != nil {
+		return nil, fmt.Errorf("no app version for snapshot: version %d", req.AppVersion)
+	}
+
+	m.mu.Lock()
+	m.appVersion = req.AppVersion
+	m.mu.Lock()
```
evan-forbes: I think this should be fine, but very open to improvements

rootulp:
Unlikely to close https://github.com/celestiaorg/celestia-app/issues/4808 because there is no RPC endpoint available for Arabica. See https://github.com/celestiaorg/devops/issues/1272#issuecomment-3049071305

Also https://github.com/celestiaorg/celestia-app/pull/4807 is a closed PR.

evan-forbes:
hmm still broken, but we at least get different errors.

this doesn't appear to be shutting down the embedded app correctly, which could be related to how we're just blindly changing the app version here.

fwiw the standalone app will state sync

cmwaters (approved):
This makes sense to me. Although can we double check that someone can't do anything nefarious like send a snapshot of version 100 and make the node crash. We should check first that the app version of the snapshot matches the app version of the header that the light client has synced to

cmwaters:
I think I know how to handle this new error:

```
Error: 130
2:21PM ERR State sync failed err="snapshot restoration failed: failed to offer snapshot: failed to get app for version 4: failed to start latest app: failed to initialize database: resource temporarily unavailable" module=statesync
2025/07/09 14:21:26 Process finished with error: exit status 1
```

we need to set the initial app version (should be v4) like:

```
func init() {
	baseapp.InitialAppVersion = appconsts.Version
}
```

cmwaters:
Ok fixed that and got another error:

```
2:39PM ERR State sync failed err="snapshot restoration failed: failed to offer snapshot: failed to get app for version 4: failed to start latest app: failed to initialize database: resource temporarily unavailable" module=statesync
```

cmwaters:
I guess we have to close down the remote app when switching to the native one perhaps

cmwaters:
It works with https://github.com/celestiaorg/celestia-app/pull/5195 cherry-picked on top 🎉

app/app.go
```diff
@@ -142,6 +142,10 @@ var (
 	_ ibctesting.TestingApp   = (*App)(nil)
 )
 
+func init() {
+	baseapp.InitialAppVersion = appconsts.Version
+}
```
evan-forbes: I'm missing why we want to do this, could you elaborate?

evan-forbes: for example, what if we state syncing mainnet cc @cmwaters

rootulp: It may no longer be needed after https://github.com/celestiaorg/celestia-app/pull/5195 based on https://github.com/celestiaorg/celestia-app/pull/5193#issuecomment-3052463265

Seems sus to me so would like to remove if possible.

evan-forbes:
can confirm this works on mocha using the normal multiplexer, also if someone passes us a snapshot that has a weird app version the getApp will error.


so I think we're good to merge and have a flup #5204

app/app.go
evan-forbes: was this just to fix a bug? doesn't seem to hurt anything

rootulp: I'm still extremely suspect of including this. 

Does state sync work without it?

If we need it, let's put it here so all the side-effect hacky `init()` logic is in one place

https://github.com/celestiaorg/celestia-app/blob/dcfacdcc32802ed93bd33edf6726e089bea43cb8/app/init.go#L26-L35

evan-forbes: just tested w/o and worked fine, so removed in c22426178c595890f8367b0451e302f3fd7192a9

rootulp: Awesome thanks

rootulp:
I cherry-picked this to v4.x and was able to state sync to Mocha and Mainnet.

rootulp:
>  also if someone passes us a snapshot that has a weird app version the getApp will error.

It would be really nice to verify this with a unit test. The multiplexer abci file currently has 0 unit tests 😢 

Update: I started working on a unit test. The error handling in the multiplexer is wild 💀

rootulp (approved):

cmwaters:
I am noticing that state sync on mocha is still failing for me. I get the following error:

```
11:46AM INF Offering snapshot to ABCI app format=3 hash=0AD78F8D405DA7BD00F60B9668A567DD78793F70EB2806B088795731478F299A height=7062000 module=statesync
11:46AM INF Using ABCI remote connection abci_version=ABCIClientVersion1 chain_id=mocha-4 maximum_app_version=3 module=server
11:46AM INF starting embedded app abci_version=ABCIClientVersion2 app_version=4 module=server
11:46AM INF stopping embedded app for version active_app_version=3 module=server
Error: 130
2025/07/10 11:46:01 Process finished with error: exit status 1
11:46AM INF Starting app for version app_version=4 args=["--force-no-bbr","--grpc.enable","--api.enable","--api.swagger=false","--with-tendermint=false","--transport=grpc","--v2-upgrade-height=0"] module=server
Flag --v2-upgrade-height has been deprecated, This flag is deprecated and was only useful prior to v4.
11:46AM INF starting ABCI without CometBFT module=server
11:46AM INF App cannot be started without CometBFT when using the multiplexer. module=server
```


This is when I download the following binaries as part of make install: 

```
❯ make install 
Checking if BBR is enabled...
sysctl: unknown oid 'net.ipv4.tcp_congestion_control'
WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README.
--> Downloading celestia-app v3.10.3 binary
--> Downloading celestia-app v4.0.7-mocha binary
Skipping download because expected version already downloaded: celestia-app_darwin_v4_arm64.tar.gz
--> Installing celestia-appd with multiplexer support
```

It looks like the multiplexer closes the v3 remote app, tries to start the v4 app but it fails because it's not starting without cometbft. I'll look into it

cmwaters:
fwiw, v4.x does work when state syncing on mocha

rootulp:
Ahh so was it failing on `main` but works on `v4.x`?

evan-forbes:
> Ahh so was it failing on main but works on v4.x?

I only ever tried on v4.x. I'm not super surprised main doesn't work since the multiplexer is untested for handling more than two versions (v3, v4, and now v5)

Issue #5194, opened 2025-07-09
`failed to get app for version 4: failed to start latest app: failed to initialize database: resource temporarily unavailable`


rootulp:
## Context

While working on another issue I tried running the single-node-upgrades script and discovered a bug

## Problem

```
$ ./scripts/single-node-upgrades.sh

[...]

PR #5195, opened 2025-07-09
fix(multiplexer): wait for v3 app to stop before starting v4

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5194

The problem was that v3 wasn't fully stopped before v4 started. v4 tried to access the db before v3 fully released it. The fix is to wait for v3 to fully stop in the `Stop` command.

## Testing

```
...
8:33PM INF stopping embedded app for version active_app_version=3 module=server
Error: 130
2025/07/08 20:33:12 Process finished with error: exit status 1
8:33PM INF using latest app app_version=4 module=server
8:33PM INF Upgrading IAVL storage for faster queries + execution on live state. This may take a while commit=436F6D6D697449447B5B5D3A307D module=server store_key="KVStoreKey{0x14002bc4fc0, circuit}" version=36
8:33PM INF Upgrading IAVL storage for faster queries + execution on live state. This may take a while commit=436F6D6D697449447B5B5D3A307D module=server store_key="KVStoreKey{0x14002bc4fb0, consensus}" version=36
8:33PM INF Upgrading IAVL storage for faster queries + execution on live state. This may take a while commit=436F6D6D697449447B5B5D3A307D module=server store_key="KVStoreKey{0x14002bc4fd0, hyperlane}" version=36
8:33PM INF Upgrading IAVL storage for faster queries + execution on live state. This may take a while commit=436F6D6D697449447B5B5D3A307D module=server store_key="KVStoreKey{0x14002bc4fa0, minfee}" version=36
8:33PM INF Upgrading IAVL storage for faster queries + execution on live state. This may take a while commit=436F6D6D697449447B5B5D3A307D module=server store_key="KVStoreKey{0x14002bc4fe0, warp}" version=36
8:33PM ERR failed to get consensus params err="collections: not found: key 'no_key' of type github.com/cosmos/gogoproto/tendermint.types.ConsensusParams" module=server
8:33PM ERR failed to get consensus params err="collections: not found: key 'no_key' of type github.com/cosmos/gogoproto/tendermint.types.ConsensusParams" module=server
8:33PM INF starting gRPC server... address=localhost:9090 module=grpc-server
8:33PM INF starting API server... address=tcp://localhost:1317 module=api-server
...
```

### DIFF
--- a/multiplexer/appd/run.go
+++ b/multiplexer/appd/run.go
@@ -100,14 +100,13 @@ func (a *Appd) IsStopped() bool {
 	return false
 }
 
-// Stop terminates the running appd process if it exists and waits for it to fully exit.
+// Stop interrupts and then kills the running appd process if it exists and
+// waits for it to fully exit. If the process is not running, it returns nil.
 func (a *Appd) Stop() error {
 	if a.cmd == nil {
-		log.Printf("Can not stop appd because cmd is nil")
 		return nil
 	}
 	if a.cmd.Process == nil {
-		log.Printf("Can not stop appd because process is nil")
 		return nil
 	}
 
--- a/multiplexer/appd/run.go
+++ b/multiplexer/appd/run.go
@@ -103,10 +103,12 @@ func (a *Appd) IsStopped() bool {
 // Stop terminates the running appd process if it exists and waits for it to fully exit.
 func (a *Appd) Stop() error {
 	if a.cmd == nil {
-		return fmt.Errorf("can not stop appd because cmd is nil")
+		log.Printf("Can not stop appd because cmd is nil")
+		return nil
 	}
 	if a.cmd.Process == nil {
-		return fmt.Errorf("can not stop appd because process is nil")
+		log.Printf("Can not stop appd because process is nil")
+		return nil
 	}
 
 	err := a.cmd.Process.Signal(os.Interrupt)
--- a/multiplexer/appd/run.go
+++ b/multiplexer/appd/run.go
@@ -30,6 +30,7 @@ type Appd struct {
 	stdin  io.Reader
 	stderr io.Writer
 	stdout io.Writer
+	cmd    *exec.Cmd
 }
 
 // New returns a new Appd instance.
@@ -84,23 +85,14 @@ func (a *Appd) Start(args ...string) error {
 	}
 
 	a.pid = cmd.Process.Pid
-
-	go func() {
-		// This waits whether process exits naturally or is killed by Stop()
-		if err := cmd.Wait(); err != nil {
-			log.Printf("Process finished with error: %v\n", err)
-		} else {
-			log.Printf("Process finished with no error\n")
-		}
-		a.pid = AppdStopped // Always reset PID when process ends
-	}()
+	a.cmd = cmd
 
 	return nil
 }
 
-// Stop terminates the running appd process if it exists.
+// Stop terminates the running appd process if it exists and waits for it to fully exit.
 func (a *Appd) Stop() error {
-	if a.pid == AppdStopped {
+	if a.pid == AppdStopped || a.cmd == nil {
 		return nil
 	}
 
@@ -118,6 +110,14 @@ func (a *Appd) Stop() error {
 		}
 	}
 
+	// Wait for the process to actually exit
+	if err := a.cmd.Wait(); err != nil {
+		log.Printf("Process finished with error: %v\n", err)
+	} else {
+		log.Printf("Process finished with no error\n")
+	}
+
+	a.pid = AppdStopped
 	return nil
 }
 
--- a/multiplexer/appd/run.go
+++ b/multiplexer/appd/run.go
@@ -13,24 +13,18 @@ import (
 	"syscall"
 )
 
-const (
-	// AppdStopped is the process ID of the celestia-appd binary when it is not running.
-	AppdStopped = -1
-)
-
 // Appd represents a celestia-appd binary.
 type Appd struct {
 	// version is the version of the celestia-appd binary.
 	// Example: "v3.10.0-arabica"
 	version string
-	// pid is the process ID of the celestia-appd binary.
-	pid int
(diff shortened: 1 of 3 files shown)


multiplexer/appd/run.go
```diff
@@ -84,23 +85,14 @@ func (a *Appd) Start(args ...string) error {
 	}
 
 	a.pid = cmd.Process.Pid
-
-	go func() {
```
rach-id, line 88: Does it make sense to keep both waits? one in Start and one in Stop for safety? This would ensure `a.pid` to have the right value every time. We could add a check if `err != nil && errors.Contains("Wait was already called", err)` to make sure we catch all cases

rach-id (dismissed):
utAck

multiplexer/appd/run.go
rootulp: Sorry for silly questions:
- How does preserving a cmd.Wait() command in `Start` add safety? Seems complicated and unncessary to me.

> This would ensure a.pid to have the right value every time.

One idea to further refactor is to remove `pid` as a field from Appd and instead always use the pid from inside a.cmd

rootulp: Given reactions, I'll implement the refactor.

multiplexer/appd/run.go
```diff
@@ -13,23 +13,18 @@ import (
 	"syscall"
 )
 
-const (
-	// AppdStopped is the process ID of the celestia-appd binary when it is not running.
-	AppdStopped = -1
-)
```
rootulp, line 19: this seems hacky and not worth exporting.

multiplexer/appd/run.go
```diff
@@ -82,50 +76,56 @@ func (a *Appd) Start(args ...string) error {
...
-	go func() {
-		// This waits whether process exits naturally or is killed by Stop()
-		if err := cmd.Wait(); err != nil {
-			log.Printf("Process finished with error: %v\n", err)
-		} else {
-			log.Printf("Process finished with no error\n")
-		}
-		a.pid = AppdStopped // Always reset PID when process ends
-	}()
+func (a *Appd) IsStopped() bool {
```
rootulp, line 87: Implemented this so that callers don't need to do:

```
		if version.Appd.Pid() == appd.AppdStopped {
```

multiplexer/appd/run_test.go
```diff
@@ -40,48 +40,85 @@ func TestCreateExecCommand(t *testing.T) {
...
-	err := appdInstance.Start()
-	require.NoError(t, err, "Start should not return an error")
-
-	// Ensure PID is set
-	require.Greater(t, appdInstance.Pid(), 0, "Process PID should be greater than 0")
-
-	// Stop the process after the test
-	err = appdInstance.Stop()
-	require.NoError(t, err, "Stop should terminate the process")
+func TestStart(t *testing.T) {
```
rootulp, line 43: Refactored unit tests

multiplexer/appd/run.go
```diff
@@ -82,50 +76,56 @@ func (a *Appd) Start(args ...string) error {
...
+	return false
 }
 
-// Stop terminates the running appd process if it exists.
+// Stop terminates the running appd process if it exists and waits for it to fully exit.
 func (a *Appd) Stop() error {
-	if a.pid == AppdStopped {
-		return nil
+	if a.cmd == nil {
+		return fmt.Errorf("can not stop appd because cmd is nil")
```
cmwaters: the previous behaviour was not to error if `IsStopped` was true i.e. if you call Stop on an already stopped multiplexer so we should be careful this doesn't break anything.

cmwaters (approved):
Honestly, I'm not so familiar with exec.Cmd so I trust that we don't have any gaps where we think a process is stopped and it hasn't. 

Otherwise this and the unit tests looks good

evan-forbes (dismissed):

multiplexer/appd/run.go
```diff
@@ -82,50 +76,56 @@ func (a *Appd) Start(args ...string) error {
...
-		if err := cmd.Wait(); err != nil {
-			log.Printf("Process finished with error: %v\n", err)
-		} else {
-			log.Printf("Process finished with no error\n")
-		}
-		a.pid = AppdStopped // Always reset PID when process ends
-	}()
+func (a *Appd) IsStopped() bool {
+	// Never started or failed to start
+	if a.cmd == nil || a.cmd.Process == nil {
```
julienrbrt, line 89: Once we actually stop, does this get nil? Shouldn't we check the Process.Pid to be sure it is stopped?

multiplexer/appd/run.go
rootulp: Agreed the previous behavior was not to error if Stop was called on a stopped appd. I can revert this change, even though it seems clearer, I agree it could break something.

multiplexer/appd/run.go
rootulp: > Once we actually stop, does this get nil?

No this never sets `cmd` to nil because the cmd contains `Process` which has information about whether it was stopped (specifically in `cmd.Process.ProcessState`)

multiplexer/appd/run.go
rootulp: Reverted in https://github.com/celestiaorg/celestia-app/pull/5195/commits/c319b08d70642ce1fb6b96021dadfffd4af8aa4a

multiplexer/appd/run.go
rootulp: > Shouldn't we check the Process.Pid to be sure it is stopped?

We don't need to check Process.Pid because `Process.ProcessState` has the information we care about. When it is present, the process has stopped. See

```go
	// ProcessState contains information about an exited process.
	// If the process was started successfully, Wait or Run will
	// populate its ProcessState when the command completes.
	ProcessState *os.ProcessState
```

rootulp:
Merging this b/c previously got approvals from Callum and I just pushed a commit to address his feedback.

PR #5113, opened 2025-06-30
feat(test): add e2e upgrade test with Tastora

mojtaba-esk:
# Implement End-to-End Upgrade Testing using Tastora Framework

## Overview
This PR introduces an end-to-end upgrade testing framework for Celestia app that validates network upgrades while ensuring all data availability functionality remains intact throughout the process.

### 🎯 **Core Implementation**
- **E2E Upgrade Test**: Full upgrade simulation from v4.0.0-rc6 → v4.1.0-dev with 5-phase validation
- **Configuration Management**: Extracted all hardcoded constants to top-level definitions with environment variable support
- **Multi-Validator Testing**: 3-validator network setup for realistic consensus testing

### **5-Phase Validation Process**
1. **🚀 Network Setup**: Multi-validator network initialization with base version
2. **📊 Pre-Upgrade Validation**: State capture + PayForBlobs
3. **⚡ Upgrade Execution**: Signal/governance-based upgrade triggering with progress monitoring
4. **✅ Post-Upgrade Validation**: State consistency + continued block production verification
5. **🔍 Functional Testing**: End-to-end DA feature validation + API endpoint testing

### **Celestia-Specific Validations**
- **PayForBlobs Transaction Flow**: Pre and post-upgrade blob submission testing
- **Data Availability Features**: Share organization, namespace isolation, commitments, light client APIs
- **Network Consensus**: Multi-validator consistency throughout upgrade process
- **API Endpoints**: gRPC and REST endpoint functionality validation

## Configuration

All test parameters are now configurable via environment variables:

```bash
# Version configuration
BASE_VERSION=v4.0.0-rc6
TARGET_VERSION=v4.1.0-dev

# Test parameters  
VALIDATOR_COUNT=3
UPGRADE_TIMEOUT=15m
PRE_UPGRADE_BLOCKS=20
POST_UPGRADE_BLOCKS=20
UPGRADE_METHOD=signal  # or governance
```

## Test Results

✅ __262.7s successful test execution__

- 3 validators maintained consensus throughout
- Block production: continuous (height 4 → 51)
- All DA features functional pre/post upgrade
- Zero test failures

## Files Changed

- `test/docker-e2e/e2e_upgrade_test.go`: Complete upgrade test implementation
- Constants extracted and properly organized
- TOML configuration fixes applied

## Usage

```bash
# Run with defaults
./test/docker-e2e/run_upgrade_test.sh

# Run with custom configuration
[...]

PR #5197, opened 2025-07-09
test: Add block sync test

### LINKED ISSUE #5185: Test that a node can blocksync from a height on v4 to the latest height on main (v5)
### Overview

An E2E test which ensures that with an image built from a given PR, a new full node can be added and block syncs correctly.
### Cadence

This test should run on every PR.

### Test Flow

- Build an image with code from PR
- Deploy celestia-app
- Deploy TxSim to spam transactions
- Add a new full node configured with block sync
- Assert it catches up.

chatton:
## Overview

dependant on https://github.com/celestiaorg/celestia-app/pull/5141

closes #5185


test/docker-e2e/e2e_block_sync_test.go
```diff
@@ -0,0 +1,117 @@
...
+					cfg.StateSync.Enable = false
+
+					// configure block sync
+					cfg.BlockSync.Version = "v0"
+
+					// configure p2p for better connectivity
+					cfg.P2P.AddrBookStrict = false
+					cfg.P2P.AllowDuplicateIP = true
+					cfg.P2P.MaxNumInboundPeers = 100
+					cfg.P2P.MaxNumOutboundPeers = 100
```
cmwaters: minor thing, but I wouldn't change anything from the default settings. Block sync should be able to work without having to change the defaults. It also looks like you add all peers as persistent peers anyway.

test/docker-e2e/e2e_block_sync_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+					// set persistent peers to connect to existing nodes
+					cfg.P2P.PersistentPeers = peerList
```
cmwaters, line 87: Given that persistent peers or conditional peers are supposed to be used for sentry nodes (and that we're also looking to deprecate supporting sentries), in the future, I think we should just use seed nodes (or whatever mechanism we expect users to start nodes with). For now, I don't mind leaving this as is

cmwaters (approved):
Just a few minor nits but otherwise looks good

test/docker-e2e/e2e_block_sync_test.go
chatton: I'll strip out as much additional config as I can. It should be relatively straight forward to make any modifications to configs we want via this fn.

.github/workflows/test-nightly.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   test-state-sync-mocha:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-8
```
rootulp: [question] goreleaser also runs on ubuntu-8. Should we be concerned about resource contention and/or GoReleaser running slower b/c this test will be using the instance?

Ref: https://github.com/rootulp/celestia-app/blob/e3861acd63e153eb7f74cec7a443c2624391ab1a/.github/workflows/goreleaser.yml#L11

test/docker-e2e/e2e_block_sync_test.go
```diff
@@ -0,0 +1,109 @@
...
+	s.Require().Equal("fn", blockSyncNode.GetType(), "expected block sync node to be a full node")
+
+	blockSyncClient, err := blockSyncNode.GetRPCClient()
+	s.Require().NoError(err)
+
+	err = s.WaitForSync(ctx, blockSyncClient, blockSyncTimeout, func(info rpctypes.SyncInfo) bool {
+		return !info.CatchingUp && info.LatestBlockHeight >= latestHeight
+	})
+
+	s.Require().NoError(err, "failed to wait for block sync node to catch up")
```
rootulp, line 107: [question] in a follow-up test or in this test, would it make sense for the chain to have a few blocks on app version 1, 2, 3, and 4?

I think it is uncommon to hit regressions with block syncing if all the blocks are on the latest version and the block sync'ing node starts on the latest version.

However, I think this test may have greater coverage and help us catch future regressions if it block synced through blocks with all historical app versions.

rootulp (dismissed):
LGTM thanks. Two optional questions.

test/docker-e2e/e2e_block_sync_test.go
chatton: Yes I think we can maybe have  a more elaborate test in which we perform incremental upgrades and make sure we have blocks at each version. The upgrade logic is being handled as its own test in https://github.com/celestiaorg/celestia-app/pull/5113 when that is merged we can make a hybrid of the two, and potentially remove this one or just run it less frequently if there aren't regressions with same version block syncs that often

.github/workflows/test-nightly.yml
chatton: yes you may be right, would need to look into this a bit more, this test is actually quite fast so I think we can leave it as `ubuntu-latest` for now. 

I think we may need to handle state cleanup per-workflow and I'm actually unsure how it behaves when you have workflows in different files that may be triggered at the same time 🤔 

Thanks for pointing it out!

test/docker-e2e/e2e_block_sync_test.go
chatton: https://github.com/celestiaorg/celestia-app/issues/5261

test/docker-e2e/e2e_block_sync_test.go
chatton: https://github.com/celestiaorg/celestia-app/issues/5261

rootulp (approved):
Still LGTM

PR #5206, opened 2025-07-09
ci: enable forwardport to main from v4.x

rootulp:
Motivation: https://github.com/celestiaorg/celestia-app/issues/5194#issuecomment-3053790327

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -31,17 +31,8 @@ pull_request_rules:
       backport:
         branches:
           - v4.x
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -31,17 +31,8 @@ pull_request_rules:
       backport:
         branches:
           - v4.x
-  - name: forward-port patches to main branch (v1.x)
+  - name: forward-port patches to main branch
     conditions:
-      - base=v1.x
```
rootulp, line 36: We don't need this `base=` condition

PR #5208, opened 2025-07-09
feat: support mainnet staking distribution

rach-id:
## Overview

A nice to have to create a network similar to that of Celestia mainnet. It takes the staking amounts from the Celestia mainnet and uses them to generate the genesis.

### DIFF
--- a/tools/talis/genesis.go
+++ b/tools/talis/genesis.go
@@ -106,7 +106,7 @@ func generateCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&appBinaryPath, "app-binary", "a", filepath.Join(gopath, "celestia-appd"), "app binary to include in the payload (assumes the binary is installed")
 	cmd.Flags().StringVarP(&nodeBinaryPath, "node-binary", "n", filepath.Join(gopath, "celestia"), "node binary to include in the payload (assumes the binary is installed")
 	cmd.Flags().StringVarP(&txsimBinaryPath, "txsim-binary", "t", filepath.Join(gopath, "txsim"), "txsim binary to include in the payload (assumes the binary is installed)")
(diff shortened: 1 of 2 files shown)


tools/talis/genesis.go
```diff
@@ -146,6 +153,34 @@ func createPayload(ips []Instance, chainID, ppath string, squareSize int, mods .
 	return nil
 }
 
+// mainnetVotingPowers contains the current Celestia mainnet staking distribution for more realistic tests.
+var mainnetVotingPowers []int
```
rootulp, line 157: [question] why is this defined in global scope if it is only assigned to inside `getMainnetStake` and only used inside `getMainnetStake`? Can it be moved inside that function?

tools/talis/genesis.go
```diff
@@ -146,6 +153,34 @@ func createPayload(ips []Instance, chainID, ppath string, squareSize int, mods .
...
+
+func getMainnetStake(index int) int64 {
+	if index < 0 {
+		return 0
+	}
+	if len(mainnetVotingPowers) == 0 {
+		// these figures reflect the exact staking values on 09/07/25.
+		mainnetVotingPowers = []int{
```
rootulp, line 165: [question] why not define this at the top of the function?

rootulp (approved):

tools/talis/genesis.go
rach-id: to only assign it a single time if needed, like a lazy initialization. it's probably premature, but didn't want this to be initialized on each run of the command if it's not going to be used. If many objects start being initialized without being used, that's extra ram usage and even slower execution (happened to me before, and it was because of a huge number of testing objects being initialized unnecessarily)

tools/talis/genesis.go
rach-id: to avoid reinitializing it every time. This gets initialized only once, if the mainnet staking distribution is needed, and gets used afterward.

evan-forbes (approved):

PR #5214, opened 2025-07-10
test: OfferSnapshot should error for unsupported app versions

rootulp:
Motivation: https://github.com/celestiaorg/celestia-app/pull/5193#issuecomment-3055012772

### DIFF
--- a/multiplexer/abci/abci_test.go
+++ b/multiplexer/abci/abci_test.go
@@ -1,0 +1,82 @@
+//go:build multiplexer
+
+package abci
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	"cosmossdk.io/log"
+	embedding "github.com/celestiaorg/celestia-app/v5/internal/embedding"
+	"github.com/celestiaorg/celestia-app/v5/multiplexer/appd"
+	abci "github.com/cometbft/cometbft/abci/types"
+	db "github.com/cosmos/cosmos-db"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/server"
+	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
+	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOfferSnapshot(t *testing.T) {
+	serverContext := server.NewDefaultContext()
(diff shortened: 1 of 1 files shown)


rootulp:
This works locally but not in CI

```
$ make test-multiplexer
--> Downloading celestia-app v3.10.3 binary
--> Downloading celestia-app v4.0.7-mocha binary
Skipping download because expected version already downloaded: celestia-app_darwin_v4_arm64.tar.gz
--> Running multiplexer tests
ok  	github.com/celestiaorg/celestia-app/v5/multiplexer/abci	1.012s
ok  	github.com/celestiaorg/celestia-app/v5/multiplexer/appd	4.142s
ok  	github.com/celestiaorg/celestia-app/v5/multiplexer/cmd	1.142s
ok  	github.com/celestiaorg/celestia-app/v5/multiplexer/internal	0.313s
```

https://github.com/celestiaorg/celestia-app/actions/runs/16184672201/job/45687903596?pr=5214#step:4:15

evan-forbes:
I was going off this test https://github.com/rootulp/celestia-app/blob/c6406238f55ac719aded22b00a7daaa8bcace898/multiplexer/abci/versions_test.go#L57-L68

but makes sense to add a test further up the stack.

the multiplexer should return the lowest known version instead of throwing an error, which imo is bad as it will just result in weird errors for unsupported versions when state syncing or doing other non-sense. 

When _not_ state syncing, comet will halt if with an explicit error for an unexpected app version being committed, so ideally state sync should do the same. We could probably just check in comet, if the state sync snapshot version is not the app version last found in the state or during `abci.Info()`, then reject the snapshot and continue to look

rootulp:
> I was going off this test https://github.com/rootulp/celestia-app/blob/c6406238f55ac719aded22b00a7daaa8bcace898/multiplexer/abci/versions_test.go#L57-L68

What's jank is even if that function returns an error, it doesn't get propagated in the way I would expect [here](https://github.com/celestiaorg/celestia-app/blob/dcfacdcc32802ed93bd33edf6726e089bea43cb8/multiplexer/abci/multiplexer.go#L426-L450)

rootulp:
> the multiplexer should return the lowest known version instead of throwing an error, which imo is bad as it will just result in weird errors for unsupported versions when state syncing or doing other non-sense.

+100, spinning this out as a new issue: https://github.com/celestiaorg/celestia-app/issues/5223

PR #5218, opened 2025-07-10
docs: fix comment

jingchanglu:
## Overview

 fix inconsistent function name in comment


evan-forbes (approved):

evan-forbes:
linters kinda suck if they can't catch these. even if the function is not exported, it should just always be enforced

rootulp (approved):

PR #5219, opened 2025-07-10
test: adding test to perform state sync with mocha

### LINKED ISSUE #5204: Fix the state sync test
we have a statesync e2e test, but it clearly did not catch the broken state sync #5098 

we should fix that so that it at least catches the old issue

chatton:
## Overview

Test which deploys a full node wired up to the mocha network and performs a state sync

As the condition to hit this issue was different from the existing state sync test, I've added a new test to sync w/mocha. The existing state sync test is deploying a fresh v4 chain and state syncing from there on a local test network.

reproduced #5098

verifies fix in #5193

closes #5204


.github/workflows/test.yml
```diff
@@ -90,6 +90,7 @@ jobs:
         testcase:
           - TestE2ESimple
           - TestStateSync
+          - TestStateMocha
```
chatton: we probably don't need to run this on every PR 🤔 , leaving it here now so it runs in CI.

test/docker-e2e/networks/config.go
```diff
@@ -0,0 +1,23 @@
...
+// Config holds the configuration for connecting to an existing live chain
+type Config struct {
+	Name    string
+	ChainID string
+	RPCs    []string
+	Seeds   string
+}
+
+// NewMochaConfig returns a Config for the mocha testnet
+func NewMochaConfig() *Config {
```
chatton, line 12: with these configs, you can wire up with an existing network, so you can deploy full nodes alongside them to test syncing functionality.

test/docker-e2e/networks/networks.go
```diff
@@ -0,0 +1,71 @@
...
+		WithGenesis(genesisBz)
+}
+
+// NewClient creates a new RPC client for connecting to the specified chain.
+func NewClient(rpc string) (*rpchttp.HTTP, error) {
+	return rpchttp.New(rpc, "/websocket")
+}
+
+// downloadGenesis downloads the genesis file for the given chain ID from the celestia networks repo
+func downloadGenesis(chainID string) ([]byte, error) {
```
chatton, line 58: same thing that happens in `celestia-appd download-genesis` however as the test pipes in the genesis bytes to the chain, it seems cleaner to do in this way instead of calling `download-genesis` via cli on the nodes themselves.

test/docker-e2e/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	cosmossdk.io/math v1.5.3
 	github.com/celestiaorg/celestia-app/v5 v5.0.0
 	github.com/celestiaorg/go-square/v2 v2.3.0
-	github.com/celestiaorg/tastora v0.1.0
+	github.com/celestiaorg/tastora v0.1.1-0.20250711073508-a5d093a7e063
```
chatton: this includes one fix which makes tastora not wait for sync with a hard coded timeout. This will make its way into v0.1.1 as soon as the PR is approved

ref: https://github.com/celestiaorg/tastora/issues/69

.github/workflows/test.yml
```diff
@@ -128,6 +128,7 @@ jobs:
         testcase:
           - TestE2ESimple
           - TestStateSync
+          - TestStateSyncMocha
```
evan-forbes: is this ran each PR for each push? how long does this test take?

imo we should add this to a nightly etc as I don't see the reason to add that much load to the network

evan-forbes: it says it finished in 9m? how is that possible?

edit: we're not blocksyncing https://github.com/celestiaorg/celestia-app/pull/5219/files#r2200414750 however we are still downloading multiple gigabytes worth of state from the network. I'm not sure I expect this to work that quickly each time

evan-forbes: also state sync on main shouldn't be working on mocha atm

see https://github.com/celestiaorg/celestia-app/pull/5193#issuecomment-3056646082 and https://github.com/celestiaorg/celestia-app/pull/5228 which just merged

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -177,3 +178,116 @@ func (s *CelestiaTestSuite) TestStateSync() {
...
+		}
+
+		t.Logf("State sync node status: Height=%d, CatchingUp=%t", status.SyncInfo.LatestBlockHeight, status.SyncInfo.CatchingUp)
+
+		// for mocha sync, we consider success when the node is no longer catching up
+		// and has synced to a reasonable height
+		if !status.SyncInfo.CatchingUp && status.SyncInfo.LatestBlockHeight > trustHeight {
+			t.Logf("State sync from mocha successful! Node caught up to height %d", status.SyncInfo.LatestBlockHeight)
+			break
+		}
```
evan-forbes, line 284: ahh I see smart

.github/workflows/test.yml
chatton: Can definitely move it into a nightly build and have it do a slack alert or something on failure.

Looking at those issues, it looks like the issue was in the download-binaries make target which is not used in the `multiplexer.Dockerfile` which just [pulls the binary from the released images](https://github.com/celestiaorg/celestia-app/blob/8007ea0d4e32ec2f668c8cb5385214dfd7c41893/docker/multiplexer.Dockerfile#L27-L28).

I tested it locally a few times and it does seem to work with images built after https://github.com/celestiaorg/celestia-app/pull/5193

chatton: block syncing code should be very similar, but we would just need to modify timeouts and definitely not run that on PRs 😅

chatton: Built an image locally and ran again and dumped the container logs [here](https://pastebin.com/PWtbZXYE)

evan-forbes:
> Can definitely move it into a nightly build and have it do a slack alert or something on failure.

I'd advocate for this, as I'm not sure that we should always block on this test due to introducing temporary incompatibilities, and due to downloading multiple gigabytes of state each push seems like an unnecessary load to put on mocha. alternatively we could run on merging to main / v4.x / v5.x

ofc, we don't need the slack or whatever bot in place to merge.

it is a _very_ useful test that will save a ton of time, thanks again!

chatton:
@evan-forbes moved it into a nightly test file so it won't run on this PR anymore. We can do the slack integration in a follow up if you're okay with that, we can manually check the cron runs and make sure it's passing and then add the notifications to slack.

EDIT: just noticed your suggestion for merges to v4/v5, I think either is fine, on merge is probably better as you'll get quicker feedback that something potentially just broke.

chatton:
I'll merge this one as is once tests are passing, then in a follow up we can re-consider cadence and add slack integration

rootulp (commented):
LGTM, thanks a ton for adding this!

Issue #5222, opened 2025-07-10
state sync error on mocha with main (v5) 


cmwaters:
On `main`

I am noticing that state sync on mocha is still failing for me. I get the following error:

```
11:46AM INF Offering snapshot to ABCI app format=3 hash=0AD78F8D405DA7BD00F60B9668A567DD78793F70EB2806B088795731478F299A height=7062000 module=statesync
11:46AM INF Using ABCI remote connection abci_version=ABCIClientVersion1 chain_id=mocha-4 maximum_app_version=3 module=server
11:46AM INF starting embedded app abci_version=ABCIClientVersion2 app_version=4 module=server
11:46AM INF stopping embedded app for version active_app_version=3 module=server
Error: 130
2025/07/10 11:46:01 Process finished with error: exit status 1
11:46AM INF Starting app for version app_version=4 args=["--force-no-bbr","--grpc.enable","--api.enable","--api.swagger=false","--with-tendermint=false","--transport=grpc","--v2-upgrade-height=0"] module=server
Flag --v2-upgrade-height has been deprecated, This flag is deprecated and was only useful prior to v4.
11:46AM INF starting ABCI without CometBFT module=server
11:46AM INF App cannot be started without CometBFT when using the multiplexer. module=server
```

This is when I download the following binaries as part of make install: 

```
❯ make install 
Checking if BBR is enabled...
sysctl: unknown oid 'net.ipv4.tcp_congestion_control'
WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README.
--> Downloading celestia-app v3.10.3 binary
Skipping download because expected version already downloaded: celestia-app_darwin_v3_arm64.tar.gz
--> Downloading celestia-app v4.0.7-mocha binary
Skipping download because expected version already downloaded: celestia-app_darwin_v4_arm64.tar.gz
--> Installing celestia-appd with multiplexer support
```

It looks like the multiplexer closes the v3 remote app, tries to start the v4 app but it fails because it's not starting without cometbft. I'll look into it

_Originally posted by @cmwaters in https://github.com/celestiaorg/celestia-app/issues/5193#issuecomment-3056646082_

cmwaters:
We seem to be hitting this:

https://github.com/celestiaorg/celestia-app/blob/2030a67d984a5e29817a07b957910a014f5c8865/multiplexer/cmd/main.go#L23

But the app is supposed to start without comet bft - it's already running

[...]

PR #5228, opened 2025-07-11
fix(statesync): download the standalone v4 binary instead

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/5222

Because I was foolish and ignorant, in my earlier v5 PR, I downloaded the multiplexer binary of v4 instead of the standalone one. When running v5, it recognised it was v4 and tried to run the v4 multiplexer. We had multiplexers within multiplexers

image

Fortunately, this was an easy fix. I've tested manually and it seems to be working.


evan-forbes (approved):
ahhhh lol

PR #5244, opened 2025-07-14
fix: apply v2 upgrade height to v3 binary only

cmwaters:
I was observing this message in the logs:

```
Flag --v2-upgrade-height has been deprecated, This flag is deprecated and was only useful prior to v4.
```

This was coming about because, the v2UpgradeHeight flag was being set to both the v3 and v4 binary when it only needs to be set for v3

### DIFF
--- a/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
+++ b/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
@@ -49,9 +49,9 @@ func modifyRootCommand(rootCommand *cobra.Command) {
 		panic(err)
 	}
 
-	var v3Args []string
+	v3Args := defaultArgs
 	if v2UpgradeHeight != "" {
-		v3Args = append(defaultArgs, "--v2-upgrade-height="+v2UpgradeHeight)
+		v3Args = append(v3Args, "--v2-upgrade-height="+v2UpgradeHeight)
 	}
 
 	versions, err := abci.NewVersions(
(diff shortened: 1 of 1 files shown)


cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
```diff
@@ -49,22 +49,22 @@ func modifyRootCommand(rootCommand *cobra.Command) {
 		panic(err)
 	}
 
-	var extraArgs []string
+	var v3Args []string
 	if v2UpgradeHeight != "" {
-		extraArgs = append(extraArgs, "--v2-upgrade-height="+v2UpgradeHeight)
+		v3Args = append(defaultArgs, "--v2-upgrade-height="+v2UpgradeHeight)
```
rach-id: I think this is the correct way to do it. Otherwise, if `v2UpgradeHeight == ""`, then `v3Args` will be empty
```suggestion
	v3Args := defaultArgs
	if v2UpgradeHeight != "" {
		v3Args = append(v3Args, "--v2-upgrade-height="+v2UpgradeHeight)
```

rach-id (changes requested):

cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
cmwaters: True, good catch

rach-id (approved):

PR #5246, opened 2025-07-14
test: mocha block sync test

chatton:
## Overview

based on the other block sync test, so merge after https://github.com/celestiaorg/celestia-app/pull/5197

This PR adds a test which performs a block sync with mocha.

It will run for days, so needs to be run on a self hosted runner. It cannot be fully tested until merging to main.

I ran locally, and it does deploy a full node and starts block syncing.

~TODO: need to include this fix https://github.com/celestiaorg/tastora/pull/72 before merging~

### DIFF
--- a/.github/workflows/test-mocha-block-sync.yml
+++ b/.github/workflows/test-mocha-block-sync.yml
@@ -2,7 +2,6 @@ name: Mocha Block Sync Test
 
 on:
   # manual trigger only
-  pull_request: # temp:
   workflow_dispatch:
     inputs:
       tag:


test/docker-e2e/e2e_block_sync_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Build(),
+		).
+		Build(ctx)
+
+	s.Require().NoError(err, "failed to create chain")
+
+	t.Log("Starting mocha block sync node")
+
+	// use a timeout context for starting the chain
+	startCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
chatton, line 164: I noticed the full node seems to take a bit longer than usual to start, not the full 10 mins, but this should avoid any flakiness in the test.

evan-forbes:
any updates on this?

I don't think its a super high priority or anything, just trying to prune PRs

chatton:
Ah, sorry @evan-forbes this is definitely stale, I think the consensus was that this would be overkill, there are other issues to work through block sync at different app versions which should address this better than a mocha block sync test. We can close this one.

PR #5251, opened 2025-07-14
test: Basic Full Stack E2E Test

### LINKED ISSUE #5227: Basic Full Stack E2E Test
### Overview

An E2E test which ensures basic core functionality throughout the entire stack. This will catch any issues in app wiring, configuration, breakage of core functionality from within celestia-app

My proposal would be to use a static tag of celestia-node and have the celestia-app image built on every PR. This would provide some base line confidence that the entire stack functions with any given changes to app.

### Cadence

This test would run on every PR.
### Test Flow

- Build an image with code from PR
- Deploy celestia-app
- Deploy DA network with light/full/bridge nodes
	- This would use a fixed tag, the latest stable release
- Test will submit PFBs
- Assert blobs can be retrieved from the light node.

chatton:
## Overview

This test performs a basic full stack test which deploys app, node and uses txclient to broadcast PFBs.

The node image can stay relatively static (bump as required) the app image is built on every tag/pr. This test should flag if anything like app / wiring or anything that effects core functionality breaks.

If it is expected that the test breaks on a given PR, we can ignore the failure as these E2Es are not marked as required.

closes #5227

### DIFF
--- a/test/util/genesis/genesis.go
+++ b/test/util/genesis/genesis.go
@@ -137,6 +137,7 @@ func (g *Genesis) WithKeyring(kr keyring.Keyring) *Genesis {
 // WithAppVersion sets the application version for the genesis configuration and returns the updated Genesis instance.
 func (g *Genesis) WithAppVersion(appVersion uint64) *Genesis {
 	g.appVersion = appVersion
+	g.ConsensusParams.Version.App = appVersion
 	return g
 }
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,6 +128,7 @@ jobs:
         testcase:
           - TestE2ESimple
           - TestStateSync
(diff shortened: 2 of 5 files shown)


test/util/genesis/genesis.go
```diff
@@ -137,6 +137,7 @@ func (g *Genesis) WithKeyring(kr keyring.Keyring) *Genesis {
 // WithAppVersion sets the application version for the genesis configuration and returns the updated Genesis instance.
 func (g *Genesis) WithAppVersion(appVersion uint64) *Genesis {
 	g.appVersion = appVersion
+	g.ConsensusParams.Version.App = appVersion
```
chatton, line 140: previously, since we bumped to v5, it was using a v5 in consensus params which node was not happy about 😅

rootulp (approved):

PR #5266, opened 2025-07-16
fix: update go mod to v2.3.1 of go-square

cmwaters:
Just waiting for CI to pass and I will cut the final release of go-square


cmwaters:
All the checks look good here so I will cut v2.3.1

PR #5268, opened 2025-07-16
fix(txsim): pollTime CLI flag should override environment variable

Galoretka:
Previously, the logic for setting the pollTime parameter did not properly prioritize the CLI flag over the TXSIM_POLL environment variable. If both were set, the environment variable could override the flag, which is not the expected behavior for CLI tools.

This commit updates the logic so that the --poll-time flag always takes precedence. The value from the TXSIM_POLL environment variable is only used if the flag is not explicitly set (i.e., when pollTime is still at its default value). This ensures consistent and predictable configuration behavior for users.


test/cmd/txsim/cli.go
```diff
@@ -177,7 +177,8 @@ account that can act as the master account. The command runs until all sequences
 				}
 			}
 
-			if os.Getenv(TxsimPoll) != "" && pollTime != user.DefaultPollTime {
+			// set pollTime: flag has priority, then env var, then default
+			if pollTime == user.DefaultPollTime && os.Getenv(TxsimPoll) != "" {
```
rootulp: LGTM but can you please include a unit test or include in the PR description how you manually verified this change?

Galoretka: > LGTM but can you please include a unit test or include in the PR description how you manually verified this change?

Work is done, hope code looks good

test/cmd/txsim/cli.go
```diff
@@ -177,11 +177,10 @@ account that can act as the master account. The command runs until all sequences
...
 
-			if os.Getenv(TxsimPoll) != "" && pollTime != user.DefaultPollTime {
-				pollTime, err = time.ParseDuration(os.Getenv(TxsimPoll))
-				if err != nil {
-					return fmt.Errorf("parsing poll time: %w", err)
-				}
+			// set pollTime: flag has priority, then env var, then default
+			pollTime, err = getPollTime(pollTime, os.Getenv(TxsimPoll), user.DefaultPollTime)
+			if err != nil {
+				return fmt.Errorf("parsing poll time: %w", err)
```
rootulp:
```suggestion
				return fmt.Errorf("get poll time: %w", err)
```

test/cmd/txsim/cli_test.go
```diff
@@ -90,6 +90,44 @@ func TestTxsimDefaultKeypath(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestGetPollTime(t *testing.T) {
```
rootulp, line 93: I'm going to push some commits to refactor this unit test

test/cmd/txsim/cli.go
```diff
@@ -177,11 +177,10 @@ account that can act as the master account. The command runs until all sequences
 				}
 			}
 
-			if os.Getenv(TxsimPoll) != "" && pollTime != user.DefaultPollTime {
-				pollTime, err = time.ParseDuration(os.Getenv(TxsimPoll))
-				if err != nil {
-					return fmt.Errorf("parsing poll time: %w", err)
-				}
+			// set pollTime: flag has priority, then env var, then default
```
rootulp:
```suggestion
```

test/cmd/txsim/cli.go
```diff
@@ -296,3 +295,19 @@ func parseUpgradeSchedule(schedule string) (map[int64]uint64, error) {
 	}
 	return scheduleMap, nil
 }
+
+// getPollTime returns the correct pollTime value based on CLI flag, environment variable, and default.
```
rootulp:
```suggestion
// getPollTime returns the pollTime value based on CLI flag, environment variable, and default.
```

test/cmd/txsim/cli_test.go
rootulp: I can't push any commits so I'll just leave suggestions.

test/cmd/txsim/cli_test.go
```diff
@@ -90,6 +90,44 @@ func TestTxsimDefaultKeypath(t *testing.T) {
...
+	if poll != flagPoll {
+		t.Errorf("expected flag poll time, got %v", poll)
+	}
+
+	// 4. Invalid env value
+	_, err = getPollTime(defaultPoll, "notaduration", defaultPoll)
+	if err == nil {
+		t.Error("expected error for invalid env value, got nil")
+	}
+}
```
rootulp:
```suggestion
func TestGetPollTime(t *testing.T) {
	defaultPoll := 10 * time.Second
	flagPoll := 30 * time.Second

	t.Run("should return default", func(t *testing.T) {
		got, err := getPollTime(defaultPoll, "", defaultPoll)
		require.NoError(t, err)
		require.Equal(t, defaultPoll, got)
	})

	t.Run("should return env", func(t *testing.T) {
		got, err := getPollTime(defaultPoll, "20s", defaultPoll)
		require.NoError(t, err)
		require.Equal(t, 20*time.Second, got)
	})

	t.Run("should return flag", func(t *testing.T) {
		got, err := getPollTime(flagPoll, "20s", defaultPoll)
		require.NoError(t, err)
		require.Equal(t, flagPoll, got)
	})

	t.Run("should return error for invalid env", func(t *testing.T) {
		_, err := getPollTime(defaultPoll, "notaduration", defaultPoll)
		require.Error(t, err)
		require.ErrorContains(t, err, "parsing poll time from env")
	})
}
```

test/cmd/txsim/cli.go
```diff
@@ -296,3 +295,19 @@ func parseUpgradeSchedule(schedule string) (map[int64]uint64, error) {
...
+// Priority: flag > env > default.
+func getPollTime(flagValue time.Duration, envValue string, defaultValue time.Duration) (time.Duration, error) {
+	if flagValue != defaultValue {
+		return flagValue, nil
+	}
+	if envValue != "" {
+		val, err := time.ParseDuration(envValue)
+		if err != nil {
+			return 0, err
```
rootulp:
```suggestion
			return 0, fmt.Errorf("parsing poll time from env: %w", err)
```

rootulp (dismissed):

evan-forbes (dismissed):

evan-forbes:
utack

@Galoretka out of curiousity what you are using txsim for or how did you come across this?

test/cmd/txsim/cli.go
```diff
@@ -296,3 +294,19 @@ func parseUpgradeSchedule(schedule string) (map[int64]uint64, error) {
 	}
 	return scheduleMap, nil
 }
+
+// getPollTime returns the pollTime value based on CLI flag, environment variable, and default.
+// Priority: flag > env > default.
+func getPollTime(flagValue time.Duration, envValue string, defaultValue time.Duration) (time.Duration, error) {
+	if flagValue != defaultValue {
```
rach-id: this can be simplified using `cmd.Flags().Changed()` which would return true if the flag has been changed

Galoretka: > this can be simplified using `cmd.Flags().Changed()` which would return true if the flag has been changed

yes

Galoretka:
> Code changes LGTM. @Galoretka are you able to try testing this manually to verify it works as expected?

Yes, tested, 

'go run test_poll_time_priority.go'
Testing poll time priority logic...
Priority: flag > env > default

Test 1: Default value (no flag, no env)
Result: 3s

Test 2: Environment variable only
Result: 5s

Test 3: Flag only
Result: 10s

Test 4: Flag and env (flag should take priority)
Result: 15s

Test 5: Invalid environment variable
Error: parsing poll time from env: time: invalid duration "invalid"

Test 6: Flag with same value as default (should still use flag)
Result: 3s

Manual testing completed!

Galoretka:
> Code changes LGTM. @Galoretka are you able to try testing this manually to verify it works as expected?

should I show the code of tests?

Galoretka:
Comparison between old and new approaches
==========================================

Scenario: User sets --poll-time=3s (same as default)
Flag value: 3s
Env value: 5s
Default value: 3s

Old approach (comparing with default):
Result: 5s

New approach (using flag.Changed()):
Result: 3s

Key difference:
- Old approach: Cannot distinguish between 'user set flag to default value' and 'user didn't set flag'   
- New approach: Correctly identifies when user explicitly set the flag, regardless of value

This is why the new approach is better:
1. It respects user intent - if they explicitly set a flag, it should be used
3. It's more explicit and less error-prone

Also comparison

PR #5275, opened 2025-07-17
fix: tx client overrides fee to 1utia if set

cmwaters:
There's a subtle bug I only came across when I set the tx client tests to have a non zero min fee. Essentially, when a user submits a transaction with a fee and no gas limit, we override the fee when we go to estimate the gas


pkg/user/tx_client.go
```diff
@@ -623,10 +626,8 @@ func (client *TxClient) EstimateGasPrice(ctx context.Context, priority gasestima
 	return resp.EstimatedGasPrice, nil
 }
 
+// estimateGas simulates the transaction, calculating the amount of gas that was
 func (client *TxClient) estimateGas(ctx context.Context, txBuilder client.TxBuilder) (uint64, error) {
-	// add at least 1utia as fee to builder as it affects gas calculation.
-	txBuilder.SetFeeAmount(sdktypes.NewCoins(sdktypes.NewCoin(appconsts.BondDenom, sdkmath.NewInt(1))))
```
cmwaters, line 628: This was being called when creating a transaction and was overriding the value that the user had set

ninabarbakadze (commented):
nice catch

pkg/user/tx_client.go
```diff
@@ -623,10 +626,8 @@ func (client *TxClient) EstimateGasPrice(ctx context.Context, priority gasestima
 	return resp.EstimatedGasPrice, nil
 }
 
+// estimateGas simulates the transaction, calculating the amount of gas that was
```
rootulp: incomplete sentence

```suggestion
// estimateGas returns an estimate for the gas used by this tx.
```

rootulp (approved):

pkg/user/tx_client.go
rach-id: This line seems to just have been moved from here to the calling function?

Also, if I understand correctly, having 1 or the value that the user set is the same for the estimation since that number is treated the same. Gas is not increased/decreased by setting a higher/lower fee amount. So, I think it's added here to always have a fee when calling `estimateGas` and avoid having users who don't know that gas affects the calculation to have invalid estimations

cmwaters: that's fine but it doesn't check if a fee is already set and thus overrides the value. As this is a private function, I've moved the check that their is at least some fee set to the two methods that call this function

rootulp (approved):

Issue #4986, opened 2025-06-12
Enable synthetic tokens for Hyperlane and remove IBC token filter


cmwaters:
In v4, we have disabled the creation of synthetic tokens on Celestia (thus only TIA can be on Celestia's state machine).

To support the usage of Celestia as a routing layer for Non TIA assets to Celestia Rollups we would like to advocate for the enabling of synthetic tokens. The next step is to write a CIP. Following a CIP, enabling synthetic tokens can be done by simply modifying app/app.go as follows:

```
app.WarpKeeper = warpkeeper.NewKeeper(
    encodingConfig.Codec,
    encodingConfig.AddressCodec,
    runtime.NewKVStoreService(keys[warptypes.ModuleName]),
    govModuleAddr,
    app.BankKeeper,
    &app.HyperlaneKeeper,
    []int32{int32(warptypes.HYP_TOKEN_TYPE_COLLATERAL), int32(warptypes.HYP_TOKEN_TYPE_SYNTHETIC)}, // Add SYNTHETIC here
)
```

Additionally, we should remove the token filter from IBC:

https://github.com/celestiaorg/celestia-app/blob/c1f2a4c5c773f20c25043f98f4b8759b603ce825/app/app.go#L357

Manav-Aggarwal:
CIP here: https://github.com/celestiaorg/CIPs/pull/325

PR #5276, opened 2025-07-17
feat: Remove IBC and Hyperlane token filter

Manav-Aggarwal:
## Overview

This PR removes the token filter middleware from the IBC transfer stack and enables synthetic token support in the Hyperlane module. These changes allow Celestia to accept inbound IBC transfers of non-native tokens and support bidirectional Hyperlane token transfers.

### 🚫 Token Filter Removal
- **Removed token filter middleware** from the IBC transfer stack in `app/app.go`
- **Eliminated module imports** and references to `x/tokenfilter`
- **Updated transfer stack** to only include Packet Forwarding Middleware and Transfer modules
- **Removed token filter documentation** from `specs/src/state_machine_modules_v5.md`

### 🔄 Synthetic Token Support
- **Enabled synthetic tokens** in Hyperlane keeper configuration
- **Updated allowed token types** to include both `HYP_TOKEN_TYPE_COLLATERAL` and `HYP_TOKEN_TYPE_SYNTHETIC`
- **Allows bidirectional token transfers** through Hyperlane protocol

### 🧪 Test Suite Updates
- **Enhanced IBC tests** to verify inbound transfers now work correctly
- **Restructured Hyperlane tests** to mirror IBC test patterns:
  - `TestHyperlaneOutboundTransfer`: Tests Celestia → Simapp → Celestia transfers
  - `TestHyperlaneInboundTransfer`: Tests Simapp → Celestia → Simapp transfers
- **Added comprehensive round-trip verification** for both test suites
- **Updated test documentation** to reflect new functionality

### 📚 Documentation Updates
- **Removed token filter references** from setup documentation
- **Updated test comments** to reflect new behavior
- **Added comprehensive test headers** explaining functionality

### Before
- ❌ Inbound IBC transfers of non-native tokens were rejected
- ❌ Only collateral tokens supported in Hyperlane

### After
- ✅ Inbound IBC transfers of non-native tokens are accepted
- ✅ Full synthetic token support in Hyperlane
- ✅ Comprehensive test coverage for round-trip transfers

## Breaking Changes
- **IBC behavior change**: Non-native tokens can now be transferred to Celestia via IBC
- **Hyperlane functionality expansion**: Synthetic tokens are now supported

## Testing
- ✅ All existing tests updated and passing
- ✅ New round-trip test scenarios added
- ✅ Comprehensive verification of balance integrity
- ✅ Both IBC and Hyperlane test suites fully functional

Closes: https://github.com/celestiaorg/celestia-app/issues/4986

### DIFF
--- a/test/interop/hyperlane_test.go
+++ b/test/interop/hyperlane_test.go
@@ -61,7 +61,7 @@ func (s *HyperlaneTestSuite) GetSimapp(chain *ibctesting.TestChain) *SimApp {
 	return app
 }
 
-// TestHyperlaneTransfer tests the full hyperlane transfer flow from celestia to simapp and back.
+// TestHyperlaneOutboundTransfer tests the full hyperlane transfer flow from celestia to simapp and back.
 // It sends utia tokens from celestia to simapp, then sends them back to celestia,
 // verifying the round-trip maintains balance integrity.
 func (s *HyperlaneTestSuite) TestHyperlaneOutboundTransfer() {


app/app.go
```diff
@@ -351,9 +350,6 @@ func New(
 		packetforwardkeeper.DefaultForwardTransferPacketTimeoutTimestamp, // forward timeout
 	)
 
-	// Token filter wraps packet forward middleware and is thus the first module in the transfer stack.
-	transferStack = tokenfilter.NewIBCMiddleware(transferStack)
```
rootulp, line 355: [blocking]
1. Please remove x/tokenfilter module if it is no longer needed
2. Please update specs to remove this line: https://github.com/celestiaorg/celestia-app/blob/6928c85dc740a858bf072e31cbb2048e7bbae88b/specs/src/state_machine_modules_v5.md?plain=1#L11

app/app.go
```diff
@@ -390,7 +386,7 @@ func New(
 		govModuleAddr,
 		app.BankKeeper,
 		&app.HyperlaneKeeper,
-		[]int32{int32(warptypes.HYP_TOKEN_TYPE_COLLATERAL)},
+		[]int32{int32(warptypes.HYP_TOKEN_TYPE_COLLATERAL), int32(warptypes.HYP_TOKEN_TYPE_SYNTHETIC)},
```
rootulp, line 388: [question] should we add a unit test that confirms we can bridge synthetic tokens?

app/app.go
Manav-Aggarwal: done, thanks!

app/app.go
Manav-Aggarwal: It looks like how the synthetic tokens were tested earlier was via tests in the `test/interop` directory.

How about we modify this method in `test/interop`: https://github.com/celestiaorg/celestia-app/blob/6928c85dc740a858bf072e31cbb2048e7bbae88b/test/interop/hyperlane_test.go#L134

to check that synthetic tokens are enabled, would that be sufficient?

Manav-Aggarwal:
> Should we also add an IBC test for a non-TIA asset?
> 
> The CIP claims that this test will be added: https://github.com/celestiaorg/CIPs/pull/325/files#diff-4ee2106d31b7a5b9ae741ea601f2c02c56b40f1953f94bae161b13915248b62bR52

It looks like how the token filter was tested earlier was via a test in the `test/interop` directory.

How about we modify this method in `test/interop`: https://github.com/celestiaorg/celestia-app/blob/6928c85dc740a858bf072e31cbb2048e7bbae88b/test/interop/tokenfilter_test.go#L103

to check that token filter is disabled, would that be sufficient?

Manav-Aggarwal:
I was reading the test: https://github.com/celestiaorg/celestia-app/blob/117b77cc511850eb24ac02472f0b18aec99a76cd/test/interop/tokenfilter_test.go#L103 but seems like it's incorrect, shouldn't this line https://github.com/celestiaorg/celestia-app/blob/117b77cc511850eb24ac02472f0b18aec99a76cd/test/interop/tokenfilter_test.go#L127 be getting the balance from CelestiaApp instead of Simapp since Celestia is chainA here or am I missing something?

Saw this was first added in https://github.com/celestiaorg/celestia-app/commit/953860c4b8c3730377ba860ce7a6360bd855ddda#diff-4225e23ad8131367f390fcbb324d9b5bae525290574114577c3f074c02143ad6 by @cmwaters 

I tried running this test after removing the token filter but it still passed when it shouldn't, which prompted this.

Either way, if the inbound transfer is rejected from Simapp to Celestia, the balance on Simapp should still be greater than zero right? Seems like it is 0 though weirdly

cmwaters:
> be getting the balance from CelestiaApp instead of Simapp since Celestia is chainA here or am I missing something?

Yeah that seems right that it should be asserting that the balance of the celestia address is 0 not the simapp address

app/app.go
cmwaters: Yup that would be good

cmwaters (commented):
Looks good!
Can we: 
- rename tokenfilter_test to something like ibc_test
- modify the hyperlane test to ensure synthetic tokens are enabled
- fix the tokenfilter test (looks like it might not have been testing the filtering correctly in the first place) and ensure that foreign assets can go two and from Celestia

Manav-Aggarwal:
> Looks good! Can we:
> 
> * rename tokenfilter_test to something like ibc_test
> * modify the hyperlane test to ensure synthetic tokens are enabled
> * fix the tokenfilter test (looks like it might not have been testing the filtering correctly in the first place) and ensure that foreign assets can go to and from Celestia

Thanks! Incorporated all these changes, feel free to re-review

test/interop/hyperlane_test.go
```diff
@@ -59,12 +61,18 @@ func (s *HyperlaneTestSuite) GetSimapp(chain *ibctesting.TestChain) *SimApp {
 	return app
 }
 
-func (s *HyperlaneTestSuite) TestHyperlaneTransfer() {
+// TestHyperlaneTransfer tests the full hyperlane transfer flow from celestia to simapp and back.
```
rootulp:
```suggestion
// TestHyperlaneOutboundTransfer tests the full hyperlane transfer flow from celestia to simapp and back.
```

rootulp (approved):
Thanks for including the tests

rootulp (approved):

evan-forbes (approved):

PR #5302, opened 2025-07-23
feat: add default block based ttl and resign expired txs

ninabarbakadze:

### DIFF
--- a/pkg/user/tx_client.go
+++ b/pkg/user/tx_client.go
@@ -36,6 +36,7 @@ import (
 
 const (
 	DefaultPollTime          = 3 * time.Second
+	DefaultTTL               = 5
 	txTrackerPruningInterval = 10 * time.Minute
 )
 
@@ -48,6 +49,7 @@ type txInfo struct {
 	signer    string
 	timestamp time.Time
 	txBytes   []byte
+	ttlHeight uint64 // Block height at which transaction expires
 }
 
 // TxResponse is a response from the chain after
@@ -134,6 +136,15 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
 	}
 }
 
+// WithTTL sets a block height-based TTL for the transaction.
+// If the transaction hasn't been included in a block by the specified height,
+// it will be considered expired and can be safely resubmitted.
+func WithTTL(height uint64) Option {
+	return func(c *TxClient) {
+		c.ttlHeight = height
+	}
+}
+
 // TxClient is an abstraction for building, signing, and broadcasting Celestia transactions
 // It supports multiple accounts. If none is specified, it will
 // try to use the default account.
@@ -147,12 +158,16 @@ type TxClient struct {
 	conns []*grpc.ClientConn
 	// how often to poll the network for confirmation of a transaction
 	pollTime time.Duration
+	// sets the TTL for the transaction
+	ttlHeight uint64
 	// sets the default account with which to submit transactions
 	defaultAccount string
 	defaultAddress sdktypes.AccAddress
 	// txTracker maps the tx hash to the Sequence and signer of the transaction
 	// that was submitted to the chain
-	txTracker           map[string]txInfo
+	txTracker map[string]txInfo
+	// txBySignerSequence maps the signer to the sequence of the transaction
+	txBySignerSequence  map[string]map[uint64]string
 	gasEstimationClient gasestimation.GasEstimatorClient
 }
 
@@ -182,9 +197,11 @@ func NewTxClient(
 		registry:            registry,
 		conns:               []*grpc.ClientConn{conn},
 		pollTime:            DefaultPollTime,
+		ttlHeight:           DefaultTTL,
 		defaultAccount:      records[0].Name,
 		defaultAddress:      addr,
 		txTracker:           make(map[string]txInfo),
+		txBySignerSequence:  make(map[string]map[uint64]string),
 		cdc:                 cdc,
 		gasEstimationClient: gasestimation.NewGasEstimatorClient(conn),
 	}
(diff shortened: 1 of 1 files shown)


ninabarbakadze (commented):
blocked on being able to access the failure log for rejected transactions because in core we currently only check transactions in rejected cache and return tx status without error code or logs.

pkg/user/tx_client.go
```diff
@@ -539,13 +587,86 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+						// confirm the resubmitted transaction using the NEW hash
+						confirmResp, err := client.ConfirmTx(ctx, resp.TxHash)
+						if err != nil {
+							return nil, fmt.Errorf("failed to confirm resubmitted tx: %w", err)
+						}
+
+						return confirmResp, nil
+					}
+				}
+			}
```
evan-forbes: could we possible clean up this logic using the typical techniques?

I could be wrong, but looks like we avoid 4 layers nested if statements by exiting early with negative if statements and deduping logic

pkg/user/tx_client_test.go
```diff
@@ -231,38 +231,93 @@ func (suite *TxClientTestSuite) TestConfirmTx() {
 }
 
 func TestRejections(t *testing.T) {
-	ttlNumBlocks := int64(5)
-	_, txClient, ctx := setupTxClient(t, ttlNumBlocks, appconsts.DefaultMaxBytes)
+	// t.Run("should reset the signer nonce after tx is rejected with no subsequent txs to resubmit", func(t *testing.T) {
```
evan-forbes: why comment this out? can we delete if not needed?

ninabarbakadze:
> apologies, reviewed then noticed it was a draft

No worries! It should be close to done now. I will need to look over it but feedback is welcome at this point. 

Important to note: We can only merge this after we've released the recent cache changes we made on v0.38.x in core and bumped app

pkg/user/tx_client.go
```diff
@@ -116,6 +120,15 @@ func WithDefaultAccount(name string) Option {
 	}
 }
 
+// WithDefaultTTLNumBlocks sets a block height-based TTL for the transaction.
+// If the transaction hasn't been included in a block by the specified height,
+// it will be considered expired and can be safely resubmitted.
```
cmwaters, line 125:
```suggestion
// WithDefaultTTLNumBlocks sets a height-based TTL for the transaction.
// If the transaction hasn't been included in a block by the specified height,
// it will be considered expired and can be safely resigned.
```

pkg/user/tx_client.go
```diff
@@ -293,17 +309,45 @@ func (client *TxClient) BroadcastPayForBlobWithAccount(ctx context.Context, acco
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Do not override user set timeout height for the transaction
+	if txTimeoutHeight > 0 {
+		timeOutHeight = txTimeoutHeight
+	}
+
+	opts = append([]TxOption{SetGasLimit(gasLimit), SetFee(fee), SetTimeoutHeight(timeOutHeight)}, opts...)
```
cmwaters, line 332:
```suggestion
	latestHeight, err := client.getLatestBlockHeight(ctx)
	if err != nil {
		return nil, err
	}
	timeOutHeight := uint64(latestHeight) + client.ttlNumBlocks

	opts = append([]TxOption{SetGasLimit(gasLimit), SetFee(fee), SetTimeoutHeight(timeOutHeight)}, opts...)
```
Since you're prepending it, if a user has it set, it will override it

pkg/user/tx_client.go
```diff
@@ -546,8 +603,54 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+				// If network sequence is greater than cached sequence, a tx was confirmed
+				// and the client is not aware of it.
+				if networkSequence > sequence {
+					return nil, fmt.Errorf("tx with hash %s was rejected by the node with log: %s", txHash, resp.Error)
+				}
+
+				// Set sequence to the network sequence to resign and resubmit the transaction
+				if err := client.signer.SetSequence(signer, networkSequence); err != nil {
+					return nil, fmt.Errorf("setting sequence: %w", err)
+				}
```
cmwaters: This won't work, if a transaction is rejected and there are several subsequent transactions, each of those subsequent ones will all get set with the same network sequence and thus only the first thereafter will be accepted, the rest will be immediately rejected in CheckTx

cmwaters (changes requested):

pkg/user/tx_client.go
ninabarbakadze: 🥲🥲🥲🥲🥲🥲🥲🥲🥲🥲

I see it now the reason it worked in my tests was that i was confirming them 1 by 1

I'll work on better design

ninabarbakadze:
Putting my thoughts in this doc until I come up with a solution that can anticipate nonces and resign transactions in a way that won't get them rejected

https://www.notion.so/celestiaorg/Handle-rejections-in-ReCheck-2482a70c951580e2aee2c1a128728973?source=copy_link

ninabarbakadze:
> Putting my thoughts in this doc until I come up with a solution that can anticipate nonces and resign transactions in a way that won't get them rejected
> 
> https://www.notion.so/celestiaorg/Handle-rejections-in-ReCheck-2482a70c951580e2aee2c1a128728973?source=copy_link

Currently prototyping this design

ninabarbakadze:
Designing a resubmission engine that consistently passes all pressure testing scenarios has proven non trivial. The current implementation is still hitting nonce errors. It’s difficult to pinpoint exactly which edge cases are being hit in these failures so further investigation is needed to identify and account for all scenarios before we can fully resolve them.

Another potential avenue to explore would be having a sort of a local mempool in the tx client that so that when we get nonce failures we will resubmit transactions that were rejected with that nonce and currently are waiting in local pool to be resubmitted.

ninabarbakadze:
Closing this PR as this work remains unplanned. We can revisit this in case we decide to add this feature in TxClient v2

Issue #5174, opened 2025-07-07
check that consensus nodes should not provide a fee estimate that is less than their local minimum


cmwaters:
Double check that changes in the local min fee are reflected in the server side of the gas fee estimator i.e. consensus nodes should not provide a fee estimate that is less than their local minimum

Issue #5257, opened 2025-07-15
Gas estimator relies on local min gas price only


cmwaters:
The current design of the gas estimator will always ensure every estimate is above the local min gas price (i.e. will be accepted by that node). We don't however incorporate the network min gas price into this floor. This means that if we were to raise the min gas price, it would not be reflected in the new gas price estimations

PR #5252, opened 2025-07-14
fix: gas estimator should provide estimate greater than local min gas price

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/5174

PR #5304, opened 2025-07-23
fix: ensure gas estimator provides a price greater than the network and local min gas price

cmwaters:
Closes: #5174, #5257

Supecedes: #5252


app/grpc/gasestimation/gas_estimator.go
```diff
@@ -122,14 +128,29 @@ func (s *gasEstimatorServer) estimateGasPrice(ctx context.Context, priority TxPr
 	if err != nil {
 		return 0, err
 	}
+
+	// Get the network minimum gas price from the min fee module
+	minGasPrice, err := s.minGasPriceFn()
+	if err != nil {
+		// If we can't get the network min gas price, fall back to default
+		minGasPrice = appconsts.DefaultMinGasPrice
```
cmwaters: not too sure if we should just return an error to the user here instead of using the DefaultMinGasPrice as it might be the wrong value

test/util/genesis/genesis.go
```diff
@@ -141,6 +143,12 @@ func (g *Genesis) WithAppVersion(appVersion uint64) *Genesis {
 	return g
 }
 
+// WithGasPrice sets the gas price of the genesis.
+func (g *Genesis) WithGasPrice(gasPrice float64) *Genesis {
```
cmwaters, line 147: New functionality to set the gas price for the gentxs (this was failing when I set a high minimum network price in genesis)

app/grpc/gasestimation/gas_estimator.go
rootulp: Seems reasonable to propagate the error

rootulp (dismissed):

rootulp (approved):

rootulp (approved):

PR #5311, opened 2025-07-23
fix(ci): multiplexer dockerfile

rootulp:
I resolved a merge conflict incorrectly in https://github.com/celestiaorg/celestia-app/pull/5299 so now the multiplexer Dockerfile doesn't work because an argument isn't named correctly.


rootulp:
Going to merge b/c this should fix CI on other in-flight PRs.

PR #5315, opened 2025-07-24
feat!: implement CIP 31 by increasing the upper bound square size and tx size limit

evan-forbes:
## Overview

there were quite a few tests that needed to be debugged or slightly refactored. Also, apparently testnode wasn't using all of the default overrides for configs, so I changed that as well.

There were also some notable test changes. I removed a few duplicate tests and test cases, included some for checking if the tx was too big and the SquareSizeIntegration test. Submitting 128MB worth of blobs for that test is hilarious and will likely cause a ton of issues in CI so I just limited it to 64 x 64. a good flup is to change that test into a unit test for prepare/process proposal and worry less about the mempool or param changes etc. 

NOT YET READY for review, as there are probably still some CI failures I want to get to


app/default_overrides.go
```diff
@@ -264,15 +264,18 @@ func DefaultEvidenceParams() *tmproto.EvidenceParams {
...
+	cfg.RPC.MaxBodyBytes = mempoolSize
 	cfg.RPC.GRPCListenAddress = "tcp://127.0.0.1:9098"
 
 	cfg.Mempool.TTLNumBlocks = 12
 	cfg.Mempool.TTLDuration = 75 * time.Second
-	cfg.Mempool.MaxTxBytes = 2 * mebibyte
-	cfg.Mempool.MaxTxsBytes = 80 * mebibyte
+	cfg.Mempool.MaxTxBytes = appconsts.MaxTxSize
+	cfg.Mempool.MaxTxsBytes = mempoolSize
 	cfg.Mempool.Type = tmcfg.MempoolTypePriority
```
evan-forbes: note to self to create a flup to update the release notes to inform operators that they likely need to update their mempools

app/test/square_size_test.go
```diff
@@ -73,34 +75,38 @@ func (s *SquareSizeIntegrationTest) TestSquareSizeUpperBound() {
...
 			maxBytes:         appconsts.DefaultMaxBytes,
-			expMaxSquareSize: appconsts.DefaultGovMaxSquareSize,
-		},
-		{
-			name:             "gov square size == hardcoded max",
-			govMaxSquareSize: appconsts.SquareSizeUpperBound,
-			maxBytes:         appconsts.DefaultUpperBoundMaxBytes,
-			expMaxSquareSize: appconsts.SquareSizeUpperBound,
+			expMaxSquareSize: 64,
 		},
```
evan-forbes, line 83: see op comment, felt this test was redundant and outdated a bit now that we have huge blonks. we should flup by changing to one or more unit tests

app/test/square_size_test.go
```diff
@@ -73,34 +75,38 @@ func (s *SquareSizeIntegrationTest) TestSquareSizeUpperBound() {
...
 	go func() {
-		seqs := txsim.NewBlobSequence(txsim.NewRange(100_000, 100_000), txsim.NewRange(1, 1)).Clone(100)
-		opts := txsim.DefaultOptions().WithSeed(rand.Int63()).WithPollTime(time.Second).SuppressLogs()
-		errCh <- txsim.Run(ctx, s.grpcAddr, s.cctx.Keyring, s.enc, opts, seqs...)
+		// this for loop is a hack to restart txsim on failures
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
```
evan-forbes, line 95: hacky, as a flup we should enabled txsim to handle failures gracefully instead of returning after 1

see other comments for converting this test to one or more unit tests

app/test/square_size_test.go
```diff
@@ -174,28 +191,50 @@ func (s *SquareSizeIntegrationTest) SetupBlockSizeParams(t *testing.T, squareSiz
...
-	msgVote := govv1.NewMsgVote(testfactory.GetAddress(s.cctx.Keyring, testnode.DefaultValidatorAccountName), propResp.Proposals[0].Id, govv1.OptionYes, "")
-	res, err = txClient.SubmitTx(s.cctx.GoContext(), []sdk.Msg{msgVote}, blobfactory.DefaultTxOpts()...)
-	require.NoError(t, err)
-	require.Equal(t, abci.CodeTypeOK, res.Code)
+	// try to query a few times until the voting period has passed
+	for i := 0; i < 20; i++ {
+		// query the proposal to get the id
+		govQueryClient := govv1.NewQueryClient(s.cctx.GRPCClient)
+		propResp, err := govQueryClient.Proposals(s.cctx.GoContext(), &govv1.QueryProposalsRequest{ProposalStatus: govv1.StatusVotingPeriod})
+		require.NoError(t, err)
```
evan-forbes, line 201: noticed this queries were flakey, so trying to fix that by querying a bunch instead of trying to time it once

pkg/appconsts/app_consts.go
```diff
@@ -3,9 +3,10 @@ package appconsts
 import "time"
 
 const (
+	// Version is the current application version.
 	Version uint64 = 5
 	// SquareSizeUpperBound imposes an upper bound on the max effective square size.
-	SquareSizeUpperBound int = 128
+	SquareSizeUpperBound int = 512
```
evan-forbes, line 9: Image

pkg/user/tx_client_test.go
```diff
@@ -499,7 +500,14 @@ func setupEstimatorService(t *testing.T) *mockEstimatorServer {
...
 
-	conn, err := grpc.NewClient(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	conn, err := grpc.NewClient(
+		addr,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallSendMsgSize(math.MaxInt32),
+			grpc.MaxCallRecvMsgSize(math.MaxInt32),
+		),
+	)
```
evan-forbes, line 510: 💀

test/txsim/blob.go
```diff
@@ -55,6 +56,11 @@ func (s *BlobSequence) WithShareVersion(version uint8) *BlobSequence {
 	return s
 }
 
+func (s *BlobSequence) WithGasPrice(gasPrice float64) *BlobSequence {
+	s.gasPrice = gasPrice
+	return s
+}
+
```
evan-forbes, line 63: used this in tests to set a lower gas price than other normal txs, maybe not needed

test/txsim/run_test.go
```diff
@@ -192,7 +193,13 @@ func TestTxSimUpgrade(t *testing.T) {
...
 	require.NoError(t, err)
 
-	conn, err := grpc.NewClient(grpcAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	conn, err := grpc.NewClient(grpcAddr,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallSendMsgSize(math.MaxInt32),
+			grpc.MaxCallRecvMsgSize(math.MaxInt32),
+		),
+	)
```
evan-forbes, line 202: why the default is 4MB is beyond me, what is this, 1998?

app/test/integration_test.go
```diff
@@ -63,44 +64,16 @@ func (s *IntegrationTestSuite) SetupSuite() {
...
 			s.cctx.Keyring,
 			c.GRPCClient,
-			50*kibibyte,
+			mebibyte-1,
 			8,
 			true,
-			s.accounts[40:120],
+			s.accounts[:60],
 		)
 	}
```
evan-forbes, line 78: I think we were a bit overzealous or rushed with a few of these tests before mainnet and should refactor them

app/test/max_tx_size_integration_test.go
```diff
@@ -1,67 +0,0 @@
...
-	"github.com/celestiaorg/celestia-app/v5/test/util/blobfactory"
-	"github.com/celestiaorg/celestia-app/v5/test/util/random"
-	"github.com/celestiaorg/celestia-app/v5/test/util/testnode"
-	"github.com/stretchr/testify/require"
-)
-
-// TestTxsOverMaxTxSizeGetRejected tests that transactions over the max tx size get rejected
-// by the application even if the validator node's local mempool config allows them.
-func TestTxsOverMaxTxSizeGetRejected(t *testing.T) {
-	if testing.Short() {
```
evan-forbes, line 19: we have the "BigBlobTest" test in this directory that does the same thing so we're good to delete this

pkg/appconsts/app_consts.go
rach-id, line 9: this means we no longer need the talis specific one. Or we could increase it to 1024

app/default_overrides.go
```diff
@@ -264,15 +264,18 @@ func DefaultEvidenceParams() *tmproto.EvidenceParams {
...
+	// this value should be the same as the largest possible response. In this case, that's
+	// likely Unconfirmed txs for a full mempool
+	cfg.RPC.MaxBodyBytes = mempoolSize
 	cfg.RPC.GRPCListenAddress = "tcp://127.0.0.1:9098"
 
 	cfg.Mempool.TTLNumBlocks = 12
 	cfg.Mempool.TTLDuration = 75 * time.Second
-	cfg.Mempool.MaxTxBytes = 2 * mebibyte
-	cfg.Mempool.MaxTxsBytes = 80 * mebibyte
+	cfg.Mempool.MaxTxBytes = appconsts.MaxTxSize
```
cmwaters, line 277: Can we make an upgrading note for node operators to increate their MaxTxBytes to 8MB?

app/test/square_size_test.go
cmwaters: I thought txsim continued until all sequences had errored (and not after the first)

pkg/appconsts/app_consts.go
```diff
@@ -17,7 +18,7 @@ const (
 	SubtreeRootThreshold int    = 64
 	TxSizeCostPerByte    uint64 = 10
 	GasPerBlobByte       uint32 = 8
-	MaxTxSize            int    = 2097152 // 2 MiB in bytes
+	MaxTxSize            int    = 8_388_608 // 8 MiB in bytes
 	TimeoutPropose              = time.Millisecond * 3500
 	TimeoutCommit               = time.Millisecond * 4200
```
cmwaters, line 23: Do these need to be adjusted at all or just stay the same

cmwaters (approved):
LGTM - just a few minor remarks

Not too sure if there is already a PR, but we should also update the MaxBytes in celestia-core from 100MB (iirc) to 128MB

app/default_overrides.go
evan-forbes: defo, see the flups from the op #5316

pkg/appconsts/app_consts.go
evan-forbes: we might need to yeah, alongside the send and recieve rates #5318

evan-forbes:
> celestia-core from 100MB (iirc) to 128MB

correct! it's already at ~130MB due to talis testing requiring that

> I thought txsim continued until all sequences had errored (and not after the first)

same but I kept seeing empty blocks after a single failure

pkg/appconsts/app_consts.go
evan-forbes: yeah noticed that as well. I deleted the talis one here, and figured we'd just add it back if needed. We'd also have to increase the max block size in comet before being able to go over 128MB

app/default_overrides.go
```diff
@@ -264,15 +264,18 @@ func DefaultEvidenceParams() *tmproto.EvidenceParams {
 
 func DefaultConsensusConfig() *tmcfg.Config {
 	cfg := tmcfg.DefaultConfig()
+	mempoolSize := int64(appconsts.DefaultUpperBoundMaxBytes) * 3
```
ninabarbakadze, line 267: could be a stupid question but i don't understand why * 3 specifically?

app/default_overrides.go
```diff
@@ -301,6 +304,7 @@ func DefaultAppConfig() *serverconfig.Config {
 	cfg.StateSync.SnapshotInterval = 1500
 	cfg.StateSync.SnapshotKeepRecent = 2
 	cfg.MinGasPrices = fmt.Sprintf("%v%s", appconsts.DefaultMinGasPrice, params.BondDenom)
-	cfg.GRPC.MaxRecvMsgSize = 20 * mebibyte
+	cfg.GRPC.MaxRecvMsgSize = appconsts.DefaultUpperBoundMaxBytes * 2
+	cfg.GRPC.MaxSendMsgSize = appconsts.DefaultUpperBoundMaxBytes * 2
```
ninabarbakadze, line 308: why was this updated?

app/default_overrides_test.go
```diff
@@ -62,7 +63,7 @@ func TestDefaultConsensusConfig(t *testing.T) {
 	t.Run("RPC overrides", func(t *testing.T) {
 		want := tmcfg.DefaultRPCConfig()
 		want.TimeoutBroadcastTxCommit = 50 * time.Second
-		want.MaxBodyBytes = int64(8388608) // 8 MiB
+		want.MaxBodyBytes = int64(appconsts.DefaultUpperBoundMaxBytes) * 3
```
ninabarbakadze, line 66: we could set `int64(appconsts.DefaultUpperBoundMaxBytes) * 3` as `mempoolSize` for better readability

app/default_overrides.go
evan-forbes: I made it up 😅

The mempool must be at least the size of the block, in the past we did 5

being 3x a 128MB block seemed reasonable, but we might want to go back to 5x

app/default_overrides.go
evan-forbes: cause large queries broke it

app/default_overrides_test.go
evan-forbes: sg #5324

PR #5323, opened 2025-07-24
fix(x/blob): event for update blob params

rnkrtt:
## Overview

Fix incorrect message type reference in EventTypeUpdateBlobParams variable. The variable was incorrectly using EventPayForBlobs{} instead of the proper EventUpdateBlobParams{} struct. This ensures that each event type variable correctly references its corresponding event struct type


x/blob/types/events.go
```diff
@@ -6,7 +6,7 @@ import (
 
 var (
 	EventTypePayForBlob       = proto.MessageName(&EventPayForBlobs{})
-	EventTypeUpdateBlobParams = proto.MessageName(&EventPayForBlobs{})
+	EventTypeUpdateBlobParams = proto.MessageName(&EventUpdateBlobParams{})
```
rach-id, line 9: luckily, this event is not used anywhere and I think we can delete it. what do you think? @rootulp

rootulp: Agreed this event is unused. Open to this PR or deleting it entirely.

```suggestion
```

rootulp (dismissed):

rnkrtt:
@rootulp done, thanks

rootulp:
golangci-lint is complaining. I'll push a commit

rootulp:
Looks like I can't push a commit.

rnkrtt:
@rootulp thank you for the corrections and guidance, and sorry for the messy commit history

x/blob/keeper/keeper.go
```diff
@@ -81,7 +81,10 @@ func (k Keeper) UpdateBlobParams(goCtx context.Context, msg *types.MsgUpdateBlob
 
 	// Emit an event indicating successful parameter update.
 	if err := ctx.EventManager().EmitTypedEvent(
-		types.NewUpdateBlobParamsEvent(msg.Authority, msg.Params),
```
rootulp: oh huh, I'm not sure how Rachid and I missed this usage. Sorry to flip back and forth but it seems like we should've just merged your original PR without deleting the function.

rnkrtt: do I need to make any additional changes ?

rach-id: Aaah, it's not used at all. The other one could be used by consumers in implementations like:
```
for _, e := range res.TxResult.GetEvents() {
				if e.Type == types.EventTypePayForBlob {
					signer := e.GetAttributes()[0].GetValue()
					_, err = sdk.AccAddressFromBech32(signer)
					require.NoError(err)
					blob, err := hex.DecodeString(tc.args[1])
					require.NoError(err)
					blobSize, err := strconv.ParseInt(e.GetAttributes()[1].GetValue(), 10, 64)
					require.NoError(err)
					require.Equal(len(blob), int(blobSize))
				}
			}
```

But since there is no test for this one, it's not used

rach-id: so this line most definitely shouldn't be deleted. Also, the other, it's arguable whether we want to keep it while unused for consumers to import, or they can define theirs. I don't see much value in importing this unless someone is already using it. Anyway, deleting it is a breaking change, so probably as rootul said, we should just fix the bug

rnkrtt:
@rootulp @rach-id in the end, are we going back to the first commit? We're fixing the function to use the correct type `EventUpdateBlobParams` instead of `EventPayForBlobs`
If so, can I open a new PR with this change so that it would be clean?

rach-id:
@rnkrtt you can make the change in this PR, that's fine

rootulp (approved):

Issue #4841, opened 2025-05-19
Decrease the unbonding period to 14 days and 1 hour


evan-forbes:
per other discussions, it has been suggested to decrease the unbonding period at some point. The soonest this could be done is v5, afaiu there is no exact period of time that is universally agreed upon, but the current period is 3 weeks

evan-forbes:
@nashqueue what value do we want here? is this value different from the sampling period (changing to 1 week iirc)

rootulp:
Apologies if this is already planned but this issue warrants a CIP IMO so perhaps the CIP can elaborate on the proposed value and rationale.

nashqueue:
The only consideration is that it should be higher than the trusting period as per https://forum.celestia.org/t/frame-by-frame-unpacking-celestia-s-windows-and-parameters/2001#p-3439-unbonding-period-must-exceed-the-weak-subjectivity-3 

These changes might break existing IBC connection to other chains so a migration path should be thought out as well

evan-forbes:
forwarding CIP and #5008 here https://github.com/celestiaorg/CIPs/blob/main/cips/cip-037.md

evan-forbes:
I'll start this tmrw

rootulp:
Blocked on https://github.com/celestiaorg/celestia-app/issues/5289

rootulp:
Updating this issue:
- Unblocked because https://github.com/celestiaorg/celestia-app/issues/5289 merged.
- This targets v6 because v5 was an emergency release to fix IBC.
- The purpose of this issue is to implement CIP-37

PR #5332, opened 2025-07-25
feat!: implement CIP-37

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/4841
Implements https://github.com/celestiaorg/CIPs/blob/main/cips/cip-037.md
- Makes the evidence params unmodifiable by governance
- The staking params are already unmodifiable by governance

## Testing

I queried parameters before and after the upgrade using

```
```

### Before

```
$ celestia-appd query staking params
params:
...
  unbonding_time: 504h0m0s
```

```
$ celestia-appd query consensus params
params:
...
  evidence:
    max_age_duration: 48h0m0s
    max_age_num_blocks: "100000"
    max_bytes: "1048576"
```

### After

```
$ celestia-appd query staking params
params:
...
  unbonding_time: 337h0m0s
```

```
$ celestia-appd query consensus params
params:
...
  evidence:
    max_age_duration: 337h0m0s
    max_age_num_blocks: "242640"
    max_bytes: "1048576"
```


evan-forbes (approved):
dope!

evan-forbes:
the squareSize test might have actually broken this time. weird error

```
--- FAIL: TestSquareSizeIntegrationTest (2.64s)
    square_size_test.go:46: setting up square size integration test
    --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound (1.01s)
        --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/max_bytes_constrains_square_size (0.01s)
            square_size_test.go:181: 
                	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:181
                	            				/home/runner/work/celestia-app/celestia-app/app/test/square_size_test.go:116
                	            	broadcast tx error: hash CC60370FA81DCA5E78732F213815A42085E00FE56937F712711FA1646C7A5925 error: invalid evidence parameters: unauthorized code: 4
                	Test:       	TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/max_bytes_constrains_square_size
    network.go:50: tearing down testnode
FAIL
FAIL	github.com/celestiaorg/celestia-app/v5/app/test	2.747s
FAIL
```

app/param_filters.go
```diff
@@ -74,5 +74,10 @@ func consensusParamFilter(msg sdk.Msg) error {
 		return errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid validator parameters")
 	}
 
+	evidenceParams := coretypes.DefaultConsensusParams().ToProto().Evidence
+	if !updateParams.Evidence.Equal(evidenceParams) {
+		return errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid evidence parameters")
+	}
```
evan-forbes: this appears to be the thing breaking the test, since I don't think the existing params are the default maybe?

we're attempting to set the evidence params to what they are onchain in the squareSize test, but that's not matching the default, therefore that's not allowing any param changes

rootulp: thanks for digging. I'll follow up

rootulp:
The square size integration test is so flaky. It failed before I merged in main: https://github.com/celestiaorg/celestia-app/actions/runs/16760489767/job/47454202969#step:4:73

Now it passes in CI and is passing locally.

app/test/square_size_test.go
```diff
@@ -47,6 +47,7 @@ func (s *SquareSizeIntegrationTest) SetupSuite() {
 
 	s.enc = encoding.MakeConfig(app.ModuleEncodingRegisters...)
 	cfg := testnode.DefaultConfig().
+		WithConsensusParams(app.DefaultConsensusParams()).
```
evan-forbes: testnode should use the default if its not

app/param_filters.go
```diff
@@ -74,5 +74,10 @@ func consensusParamFilter(msg sdk.Msg) error {
 		return errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid validator parameters")
 	}
 
+	evidenceParams := DefaultEvidenceParams()
+	if !updateParams.Evidence.Equal(evidenceParams) {
+		return errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid evidence parameters")
+	}
```
evan-forbes: bit of a nit

I don't like the idea of making a function called "Default..." consensus breaking. feels like a footgun. It's kinda nice as we can just change things in one spot, but idk

wdyt?

app/default_overrides.go
```diff
@@ -252,13 +252,10 @@ func DefaultBlockParams() *tmproto.BlockParams {
 // DefaultEvidenceParams returns a default EvidenceParams with a MaxAge
 // determined using a goal block time.
 func DefaultEvidenceParams() *tmproto.EvidenceParams {
```
evan-forbes: maybe we just need to update the docs here and mention changing a value here is now consensus breaking

evan-forbes (dismissed):

app/test/square_size_test.go
rootulp: This can prob be reverted.

app/default_overrides.go
rootulp: CIP-37 made the evidence params unmodifiable by governance so I think what you're asking for is to update the comment to say that a consensus breaking version bump is needed to modify any of the evidence params. Can do.

app/param_filters.go
rootulp: The function is not consensus breaking. A testnode can be created with different evidence parameters. 

The param filter here just prevents the evidence params from being modified via governance. So I'm not sure what the action item is here but maybe you want this rename: `DefaultEvidenceParams()` => `EvidenceParams()`?

app/test/square_size_test.go
rootulp: 26b6f55f95f54ba77e476d49258336284ca484ff

app/default_overrides.go
rootulp: 3fc3a8409c0fbdb0dafd080a99bec7ec2c679929

rootulp:
The 3 failing CI checks appear unrelated to this PR. Created issues for the e2e test flakes.

app/param_filters.go
evan-forbes: if we change a value in the Default consensus param function in v6.1 relative to v6.0, then this filter will be different no? the evidence params will no longer be equal and this will fail

this is part of the reason why the square size integration test failed previously (yes its flakey af, but it was failing consistently with the same error indicating that evidence params were no longer equal to the defaults) afaiu

evan-forbes: > A testnode can be created with different evidence parameters.

yeah! then what happens when we do that and try to change the maxBlockBytes? if that passes, then I'm incorrect and this isn't consensus breaking

evan-forbes (dismissed):
we're removing the liveness check if that's flakey holy 🤬

ofc approving as its a minor naming/comment wording thing, but 

> The function is not consensus breaking. A testnode can be created with different evidence parameters.

if we change the consensus params in testnode, then try to submit a consensus params change tx (such as in the squareSizeIntegrationTest, we get this failure

```
--- FAIL: TestSquareSizeIntegrationTest (2.57s)
    square_size_test.go:46: setting up square size integration test
    --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound (1.01s)
        --- FAIL: TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/max_bytes_constrains_square_size (0.01s)
            square_size_test.go:184:
                	Error Trace:	/home/evan/src/clones/celestiaorg/celestia-app/celestia-app/app/test/square_size_test.go:184
                	            				/home/evan/src/clones/celestiaorg/celestia-app/celestia-app/app/test/square_size_test.go:119
                	            	broadcast tx error: hash E99F979953BD01A0AF8D2BB204D1B08F500DC823AA46B7E48997916CC43794B2 error: invalid evidence parameters: unauthorized code: 4
                	Test:       	TestSquareSizeIntegrationTest/TestSquareSizeUpperBound/max_bytes_constrains_square_size
```

therefore we know for certain that a change to `DefaultEvidenceParams` is consensus breaking. Which makes sense, since we're using it make a comparison when filtering messages in the state machine.

this is purley a feeble attempt to avoid footguns, as I don't like that we're having to remember not to change those functions on accident

app/default_overrides.go
```diff
@@ -249,16 +249,14 @@ func DefaultBlockParams() *tmproto.BlockParams {
 	}
 }
 
-// DefaultEvidenceParams returns a default EvidenceParams with a MaxAge
-// determined using a goal block time.
+// DefaultEvidenceParams returns the evidence params defined in CIP-37. The
+// evidence parameters are not modifiable by governance so a consensus breaking
+// release is needed to modify them.
```
evan-forbes:
```suggestion
// DefaultEvidenceParams returns the evidence params defined in CIP-37. The
// evidence parameters are not modifiable by governance so a consensus breaking
// release is needed to modify them or parameters returned from this function.
```

app/param_filters.go
rootulp: >this is part of the reason why the square size integration test failed previously (yes its flakey af, but it was failing consistently with the same error indicating that evidence params were no longer equal to the defaults) afaiu

The square size integration test was failing because it attempted to modify fields in the consensus params and also set the evidence params to values that differ from the CIP-37 values.

> yeah! then what happens when we do that and try to change the maxBlockBytes? if that passes, then I'm incorrect and this isn't consensus breaking

I'll try testing it. I expect that it will be possible to change the `maxBlockBytes` as long as the evidence params remain untouched.

app/default_overrides.go
rootulp: In this sentence "them" refers to the evidence parameters. So I'm not sure why "or parameters returned from this function" is necessary.

Maybe this helps?

```suggestion
// DefaultEvidenceParams returns the evidence params defined in CIP-37. The
// evidence parameters are not modifiable by governance so a consensus breaking
// release is needed to modify the evidence parameters.
```

app/param_filters.go
rootulp: Sorry on re-read I more understand the concern. If I create a testnode with different evidence params from the defaults, it isn't possible to set `maxBlockBytes` unless I also set the evidence params to `DefaultEvidenceParams()`

rootulp: > I don't like the idea of making a function called "Default..." consensus breaking.

Seems like one mitigation is to remove the `Default` prefix. Can do.

app/default_overrides.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/5332/commits/0c077a88d8832efd0be4be4d3a4816791c1dae73

app/param_filters.go
rootulp: Please let me know if you think https://github.com/celestiaorg/celestia-app/pull/5332/commits/0c077a88d8832efd0be4be4d3a4816791c1dae73 resolves this thead.

evan-forbes (dismissed):

app/default_overrides.go
evan-forbes: the commit posted solves the issue imo thanks!

the reasoning for adding that last bit originally was just to be extra explicit. perhaps a warning would have been better to simply indicate that we use this function in the state machine.

> The evidence parameters are not modifiable by governance so a consensus breaking release is needed to modify them

to me, this could be interpreted simply as a statement that a hardfork is needed to change the evidence params, not that if you change something in this function it will introduce a consensus breaking change that could halt the chain the next time someone tries to increase the block size.

app/param_filters.go
```diff
@@ -74,5 +74,10 @@ func consensusParamFilter(msg sdk.Msg) error {
 		return errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid validator parameters")
 	}
 
+	evidenceParams := EvidenceParams()
+	if !updateParams.Evidence.Equal(evidenceParams) {
+		return errors.Wrapf(sdkerrors.ErrUnauthorized, "invalid evidence parameters")
+	}
```
cmwaters, line 80: non blocking: could updateParams.Evidence also be empty (implying no change to the params)

app/param_filters_test.go
```diff
@@ -197,6 +197,17 @@ func TestConsensusParamFilter(t *testing.T) {
...
+		{
+			name: "invalid case: non-default evidence params",
+			msg: &consensustypes.MsgUpdateParams{
+				Authority: "authority",
+				Block:     coretypes.DefaultConsensusParams().ToProto().Block,
+				Evidence:  &tmproto.EvidenceParams{MaxAgeNumBlocks: 1, MaxAgeDuration: time.Hour, MaxBytes: 1000000},
+				Validator: coretypes.DefaultConsensusParams().ToProto().Validator,
+				Abci:      coretypes.DefaultConsensusParams().ToProto().Abci,
+			},
+			expectedErr: sdkerrors.ErrUnauthorized,
```
cmwaters: nit: we could check the error is specifically because the evidence params is incorrect and not something else

app/upgrades.go
```diff
@@ -92,3 +104,47 @@ func (app App) RegisterUpgradeHandlers() {
...
+
+	err = app.StakingKeeper.SetParams(ctx, params)
+	if err != nil {
+		sdkCtx.Logger().Error("failed to set staking params", "error", err)
+		return err
+	}
+	return nil
+}
+
+func (app App) OverrideEvidenceParams(ctx context.Context) error {
```
cmwaters: nit: Would prefer terminology such as "setting". I think "overriding" sounds like it should normally be the old value but we're temporarily changing it

pkg/appconsts/initial_consts.go
```diff
@@ -24,7 +24,9 @@ const (
 	// DefaultUnbondingTime is the default time a validator must wait
 	// to unbond in a proof of stake system. Any validator within this
 	// time can be subject to slashing under conditions of misbehavior.
-	DefaultUnbondingTime = 3 * 7 * 24 * time.Hour
+	//
+	// Modified from 3 weeks to 14 days + 1 hour in CIP-037.
+	DefaultUnbondingTime = 337 * time.Hour // (14 days + 1 hour)
```
cmwaters:
```suggestion
	UnbondingTime = 337 * time.Hour // (14 days + 1 hour)
```
To be more correct, we could call it the `UnbondingTime` and move it to the version constants, not here in the initial constants

cmwaters (dismissed):
Just had a bunch of non-blocking nits, but semantically these changes look good.

It would be nice to add querying of the unbonding period before and after the upgrade in the Tastura e2e test

app/param_filters.go
rootulp: Hmm good question. Currently MsgUpdateParams has a field for Evidence that is a pointer so it could be nil implying no change to the params but it would fail here because this diff forces the MsgUpdateParams to have evidence params that exactly match those in CIP-37.

```
// MsgUpdateParams is the Msg/UpdateParams request type.
type MsgUpdateParams struct {
	// authority is the address that controls the module (defaults to x/gov unless overwritten).
	Authority string `protobuf:"bytes,1,opt,name=authority,proto3" json:"authority,omitempty"`
	// params defines the x/consensus parameters to update.
	// VersionsParams is not included in this Msg because it is tracked
	// separarately in x/upgrade.
	//
	// NOTE: All parameters must be supplied.
	Block     *types.BlockParams     `protobuf:"bytes,2,opt,name=block,proto3" json:"block,omitempty"`
	Evidence  *types.EvidenceParams  `protobuf:"bytes,3,opt,name=evidence,proto3" json:"evidence,omitempty"`
	Validator *types.ValidatorParams `protobuf:"bytes,4,opt,name=validator,proto3" json:"validator,omitempty"`
	// Since: cosmos-sdk 0.50
	Abci *types.ABCIParams `protobuf:"bytes,5,opt,name=abci,proto3" json:"abci,omitempty"`
}
```

rootulp: Although I'm not sure if it's even possible to submit a MsgUpdateParams with a nil Evidence in the rest of the state machine. I think it might fail some `ValidateBasic` check elsewhere but haven't investigated. 

Seems small enough to force any submitter to include the exact EvidenceParams unmodified. If you feel differently, please create a follow-up issue and we can tackle it post v6.

app/param_filters_test.go
rootulp: Good idea. Fixed in 503c80e48be95dd50a2bf05f4675cb439ce692dc

app/upgrades.go
rootulp: Agreed. Fixed in 8c4b9a9ed4d0fe5d06a9cd888bcfc6ba7bbe08b0

pkg/appconsts/initial_consts.go
rootulp: Resolved in 1110a96645e8155c6ea1575b806e3cad4a4f9487 and 429c9c5da5096e1c5f4bc01d43727c2c9e19a013

Moved to app_consts.go instead of versioned_consts.go because versioned_consts.go just has getters.

cmwaters (approved):
👍 

Great work

Issue #5325, opened 2025-07-24
Stop using `share.ContinuationSparseShareContentSize` for constants


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/0641738ad3a7c79bc9a351e94b84e78989cbc688/pkg/appconsts/initial_consts.go#L36

https://github.com/celestiaorg/celestia-app/blob/0641738ad3a7c79bc9a351e94b84e78989cbc688/pkg/appconsts/initial_consts.go#L16-L18

## Problem

I don't understand why these need to use the const `share.ContinuationSparseShareContentSize`. 

It's also extremely misleading because `MaxBytes` was modified via governance on all networks to be 8 MiB which differs from the current constant.

## Proposal

It would be much clearer if DefaultUpperBoundMaxBytes and DefaultMaxBytes did not depend on `share.ContinuationSparseShareContentSize`. Instead they should be a whole number of MiB which matches the value we plan on using in a governance proposal on networks after v5 is deployed.

evan-forbes:
good point, yeah I think we're just being too precise here w/ using ContinuationSparseShareContentSize

overflowing isn't bad in PrepareProposal, so we should be good to increase

PR #5335, opened 2025-07-25
refactor(appconsts): use explicit number of mebibytes for max bytes

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5325


rootulp:
Test seems broken and flaky. On `main` I'm getting a different error

```
                	            	tx execution failed with code 3: failed to execute message; message index: 0: 0: inactive proposal
```

pkg/appconsts/initial_consts.go
```diff
@@ -2,20 +2,19 @@ package appconsts
...
 const (
+	mebibyte = 1_048_576 // bytes
 	// DefaultGovMaxSquareSize is the default value for the governance modifiable
 	// max square size.
 	DefaultGovMaxSquareSize = 256
 
 	// DefaultMaxBytes is the default value for the governance modifiable
 	// maximum number of bytes allowed in a valid block.
-	DefaultMaxBytes = DefaultGovMaxSquareSize * DefaultGovMaxSquareSize * share.ContinuationSparseShareContentSize
+	DefaultMaxBytes = 32 * mebibyte
```
rach-id, line 17: The previous value was `31588352` and not 32mib, is this something we want to bump?

pkg/appconsts/initial_consts.go
```diff
@@ -31,6 +30,6 @@ const (
 	// included in a block if they specify a gas price lower than this.
 	// Only applies to app version >= 2
 	DefaultNetworkMinGasPrice = 0.000001 // utia
-)
 
-var DefaultUpperBoundMaxBytes = SquareSizeUpperBound * SquareSizeUpperBound * share.ContinuationSparseShareContentSize
```
rach-id, line 36: same, this used to be `126353408`

pkg/appconsts/initial_consts.go
rootulp: Yes because the previous value isn't easy to remember or communicate. `31588352` is really specific but AFAIK these constants don't have to be that specific or dependent on `share.ContinuationSparseShareContentSize`

rootulp: FWIW this value was modified via governance to 8 mebibyte and it hasn't been an issue.

evan-forbes (approved):

PR #5343, opened 2025-07-28
fix!: register upgrade handler

evan-forbes:
## Overview

cherry picks the upgrade handler changes to v5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ BUILD_FLAGS_MULTIPLEXER := -tags "ledger multiplexer" -ldflags '$(ldflags)'
 # .goreleaser.yaml
 # docker/multiplexer.Dockerfile
 CELESTIA_V3_VERSION := v3.10.5
+CELESTIA_V4_VERSION := v4.0.10
 
 ## help: Get more info on make commands.
 help: Makefile
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -7,13 +7,17 @@ version: 2
 before:
   hooks:
     - go mod tidy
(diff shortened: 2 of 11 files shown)


rootulp:
I think we have to bump all the import statements to use 

```
github.com/celestiaorg/celestia-app/v5
```

and the go.mod `module` directive

evan-forbes:
> I think we have to bump all the import statements to use

yeah I was in the process of backporting those and we likely still need a few other random things. sorting through main now

evan-forbes:
the markdown filter is expected to fail since we didn't merge e6e1b2be76703ee1208034087ca9940e91d64339 yet

went through all the diffs and we should be good. there doesn't appear to be anything that shouldn't be imported via cherry-picking.

tests still take forever on this branch as we don't have fixes reducing the block time in testnode, splitting up the tests, or removing a bunch of redundant integration tests

I'm going to branch off of this branch and get the tastora test to work. We can do that for #5344 as well or just wait for this to merge

rootulp (approved):
LGTM, I didn't try it manually though

Makefile
```diff
@@ -27,6 +27,7 @@ BUILD_FLAGS_MULTIPLEXER := -tags "ledger multiplexer" -ldflags '$(ldflags)'
 # .goreleaser.yaml
 # docker/multiplexer.Dockerfile
 CELESTIA_V3_VERSION := v3.10.5
+CELESTIA_V4_VERSION := v4.0.10
```
cmwaters, line 30: Can update to v4.1.0-arabica now

cmwaters (approved):
LGTM

Tested state sync on mocha manually and single-node script

Issue #5340, opened 2025-07-28
IBC failing after v4 – [module does not own channel capability]


qubelabsio:
IBC transactions from Osmosis to Celestia (channel-6994 → channel-2) are failing with the following error on-chain:

> failed to execute message; message index: 0: module does not own channel capability: channel capability not found

Logs (Hermes v1.12.0):

```
Jul 28 14:40:01 Ubuntu-2204-jammy-amd64-base hermes[800383]: 2025-07-28T14:40:01.158182Z ERROR ThreadId(27) worker.batch{chain=osmosis-1}:supervisor.handle_batch{chain=osmosis-1}:supervisor.process_batch{chain=osmosis-1}:worker.packet{src_chain=osmosis-1 src_port=transfer src_channel=channel-6994 dst_chain=celestia}:relay{odata=cleared/005533d8 ->Destination @1-40769599; len=2}:send_messages_and_wait_check_tx{chain=celestia tracking_id=cleared/005533d8}:send_tx_with_account_sequence_retry{chain=celestia account.sequence=109750}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 1: could not retrieve module from port-id: capabilities/ports/transfer/channels/channel-2: capability not found [cosmos/ibc-go/modules/capability@v1.0.1/keeper/keeper.go:452] with gas used: '427402'", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "6680865"} }
[...]

PR #5344, opened 2025-07-28
fix: add capability module

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5340

## Testing

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -102,6 +102,7 @@ import (
 	"github.com/cosmos/ibc-apps/middleware/packet-forward-middleware/v8/packetforward"
 	packetforwardkeeper "github.com/cosmos/ibc-apps/middleware/packet-forward-middleware/v8/packetforward/keeper"
 	packetforwardtypes "github.com/cosmos/ibc-apps/middleware/packet-forward-middleware/v8/packetforward/types"
+	"github.com/cosmos/ibc-go/modules/capability"
 	capabilitykeeper "github.com/cosmos/ibc-go/modules/capability/keeper"
 	capabilitytypes "github.com/cosmos/ibc-go/modules/capability/types"
 	icahost "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/host"
@@ -413,6 +414,7 @@ func New(
 		authzmodule.NewAppModule(encodingConfig.Codec, app.AuthzKeeper, app.AccountKeeper, app.BankKeeper, encodingConfig.InterfaceRegistry),
 		consensus.NewAppModule(encodingConfig.Codec, app.ConsensusKeeper),
 		ibcModule{ibc.NewAppModule(app.IBCKeeper)},
+		capability.NewAppModule(encodingConfig.Codec, *app.CapabilityKeeper, true),
 		transfer.NewAppModule(app.TransferKeeper),
 		blob.NewAppModule(encodingConfig.Codec, app.BlobKeeper),
 		signal.NewAppModule(app.SignalKeeper),


rootulp:
My local devnet testing may be flawed because I was able to bridge utia back and forth using existing celestia-app v4.0.10 which has the bug.

One hypothesis for why I can bridge utia in and out on v4.0.10 is that the capability module bug only applies to connections that were created _before_ the v4 activation height and in my testing I created the connection after the v4 activation height.

rootulp:
Yep confirmed that I can actually reproduce the error only when I start the chain on v3, create a connection, send tokens on it (works), upgrade to v4, then try to send tokens on that connection (fails):

```
2025-07-29T04:33:24.127090Z ERROR ThreadId(15) send_messages_and_wait_commit{chain=test tracking_id=ft-transfer}:send_tx_with_account_sequence_retry{chain=test account.sequence=8}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 0: module does not own channel capability: channel capability not found [cosmos/ibc-go/v8@v8.7.0/modules/apps/transfer/keeper/relay.go:78] with gas used: '56597'", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "138"} }
2025-07-29T04:33:24.127129Z ERROR ThreadId(15) send_messages_and_wait_commit{chain=test tracking_id=ft-transfer}:send_tx_with_account_sequence_retry{chain=test account.sequence=8}: gas estimation failed or encountered another unrecoverable error error=gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 0: module does not own channel capability: channel capability not found [cosmos/ibc-go/v8@v8.7.0/modules/apps/transfer/keeper/relay.go:78] with gas used: '56597'", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "138"} }
ERROR transfer error: failed while submitting the Transfer message to chain test: gRPC call `send_tx_simulate` failed with status: status: Unknown, message: "failed to execute message; message index: 0: module does not own channel capability: channel capability not found [cosmos/ibc-go/v8@v8.7.0/modules/apps/transfer/keeper/relay.go:78] with gas used: '56597'", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "138"} }
```

rootulp:
Confirmed that I can send tokens on a connection created on v3 after the upgrade to v4 if I use a celestia binary built from this PR.

cmwaters:
> One hypothesis for why I can bridge utia in and out on v4.0.10 is that the capability module bug only applies to connections that were created before the v4 activation height and in my testing I created the connection after the v4 activation height.

That's interesting. So people could still move funds if they wanted to, they would just need to create new connections

scripts/hermes.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+# This script is used to test IBC between Celestia and IBC-0 (a Gaia chain managed via Gaia Manager).
+# Follow the prerequisites and tutorial at https://hermes.informal.systems/tutorials/index.html
+#
+# Steps:
+# 1. Run celestia-app via ./scripts/single-node.sh or ./scripts/single-node-upgrades.sh
+# 2. Run Gaia Manager via ./bin/gm start
```
cmwaters, line 8: nit: would be nice to have a link to this or at least how to get the binary

cmwaters (approved):
utAck

rach-id (approved):

PR #5351, opened 2025-07-29
chore: update upgrade script to v5

cmwaters:


cmwaters:
We should probably forward port this to main

evan-forbes (approved):

PR #5376, opened 2025-07-30
docs(release-guide): explain release branch

rootulp:


docs/maintainers/release-guide.md
```diff
@@ -2,26 +2,29 @@
...
     1. If the release candidate is not satisfactory, go back to step 1
     1. If the release candidate is satisfactory, create an official release
 
 ## Notes
 
-1. The v3.x.x binary used by the multiplexer is hard-coded in the Makefile (until [#4721](https://github.com/celestiaorg/celestia-app/issues/4721) is resolved). In order to include a code change to the v3.x branch in celestia-app v4.x, you must:
-    1. Create a release of celestia-app v3.x.x
-    1. Modify the Makefile to point to the new version
-    1. Create a release of celestia-app v4.x.x
+The versions of previous binaries are hard-coded at multiple places in celestia-app due to the multiplexer (until https://github.com/celestiaorg/celestia-app/issues/4921 is resolved). In order to include code changes to historical binaries, you need to make a release of that code change and bump the hard-coded versions.
```
rach-id, line 13: FLUP that copilot can do is list where the versions are hardcoded to make sure none is missed

rach-id (approved):

docs/maintainers/release-guide.md
```diff
@@ -2,26 +2,29 @@
...
+
+## Release branch
+
+1. Navigate to <https://github.com/celestiaorg/celestia-app/branches>.
+2. Click **New Branch**.
+3. Create a release branch. Example name: `v4.1.0-release`. Example source: `v4.x.x`.
```
cmwaters: Do we do an x.x now? Is that necessary? I don't see us backporting a patch release to another minor version i.e. say we have a 4.0.1 and a 4.1.0 and we want to release a patch we wouldn't cut v4.1.1 **and** 4.0.2. We would just do the former, so I don't see why we need branches for each minor version (just the release branch)

docs/maintainers/release-guide.md
rootulp: They're listed in Makefile https://github.com/celestiaorg/celestia-app/blob/1f5041c2cfc897e4d95a2e6290941fa5cfc4caac/Makefile#L27-L33

If we include them in this doc, they may become stale but not opposed.

docs/maintainers/release-guide.md
rootulp: Oh sorry, you're right we don't have an x.x now. There is only:
- v4.x
- v5.x

so I'll modify the Example source:

```suggestion
3. Create a release branch. Example name: `v4.1.0-release`. Example source: `v4.x`.
```

cmwaters (approved):
Looks good barring the tests you seem to have broken by changing the markdown document :)

Issue #5388, opened 2025-07-31
Add a query endpoint to make it easier to see which validators have not yet signalled


cmwaters:
This can make it easier to work out who needs following up with

PR #5389, opened 2025-07-31
feat: query validators who have not signalled

cmwaters:
Closes: https://github.com/celestiaorg/celestia-app/issues/5388


cmwaters:
As this is a query endpoint it's not a breaking change and can be added to v5.x. I would also forward port it to main

evan-forbes (dismissed):
Nice, we can flup adding cli toobfor added convenience

rootulp (dismissed):
Code changes LGTM. 

[optional][follow up] It may be helpful to sort the missing validator monikers in descending order of voting power.

Agreed with Evan that a CLI command to query this endpoint also seems helpful as a follow up.

evan-forbes (approved):

PR #5398, opened 2025-08-01
chore(x/minfee): remove legacy params file

mutestt:
This pull request removes the `legacy_params.go` file from the `x/minfee/types` package.

The file contains a `TODO` comment indicating that it can be removed once the migration from the `x/params` module to self-managed module parameters is complete. This migration was finalized in Cosmos SDK v0.47, and `celestia-app` is on a more recent version, making this file and its contents obsolete.

A search of the codebase confirms that this file is no longer in use. This change removes dead code, simplifying the module and completing the action requested in the `TODO`.


evan-forbes:
this will be closed with the broken conflicts at some point

also I don't see a huge benefit in removing the file, maybe we can just remove that todo

mutestt:
@rootulp 
- Removed outdated TODO comment from `x/minfee/types/legacy_params.go` as per review feedback.

damiannolan:
> also I don't see a huge benefit in removing the file, maybe we can just remove that todo

Since we migrated away from x/params in v4, we should be able to remove x/params and the surrounding code for KeyTables, Subspaces..etc. Now that we are on v5 and soon v6, in theory it should be possible to lift x/params entirely from the codebase.

My recommendation would be to remove all of that code rather than just this one file, and also run upgrade tests and verify everything is working as expected. I think it would be best to create a dedicated issue to track that work.

PR #5399, opened 2025-08-01
chore: Address legacy TODOs in blob and specs

mutestt:
This pull request performs some code cleanup by addressing two old `TODO` comments related to legacy features.

x/blob/keeper/params.go:
- The function `SetParamsLegacy` has been removed. A `TODO` in the code indicated that this function was for past migrations and is no longer needed. A code search confirms it is not being used anywhere.

specs/src/proto/wire.proto:
- A `Deprecated` comment has been added to the `Decimal` message. This clarifies that the message is obsolete, as decided in issue #120, and helps prevent it from being used in new code.

### DIFF
--- a/specs/src/proto/wire.proto
+++ b/specs/src/proto/wire.proto
@@ -8,6 +8,9 @@ message TransactionFee {
   uint64 tip_rate = 1;
 }
 
+// Deprecated: This message is obsolete as per the decision in issue #120 to use
+// a 'string' type for decimals. It is kept for legacy purposes and will be
+// removed in a future refactor.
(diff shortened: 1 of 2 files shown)


evan-forbes:
this PR will be closed w/o further progress

not sure its worth resolving these todos since there's a lot of context lost over the years since they were written. we can probably just remove the todos

evan-forbes:
actually there aren't really todos this PR is addressing entirely and seems redundant to the other PR from @mutestt  so closing now

if I'm incorrect here or you had something else in mind feel free to reopen @mutestt

PR #5401, opened 2025-08-01
test: add IBC test

### LINKED ISSUE #5387: Tastora IBC Test
### Overview

Ensure IBC functionality works correctly with Tastora

### Test Case

- Deploy App vN and a Second Chain
- Establish Connection/Channel with the Hermes relayer
- Send tokens back and forth (can also verify token filter works here)
- Upgrade to N + 1
- Send tokens back and forth on same channel
- Create new channel and Send tokens back and forth.

Dependant on https://github.com/celestiaorg/tastora/issues/77

chatton:
## Overview

This PR adds an IBC test verifying outbound ibc transfers (it can be extended in a followup to send them back, I left it out for now as the PR was already growing quite a bit)

We bump to the newest version of tastora which has support for ibc and hermes.

The test introduced performs the following steps

- deploys 2 chains
- deploys hermes
- create an IBC connection and channel
- performs a token transfer
- upgrades from app version N-1 to N
- performs token transfer over the same channel
- creates a new connection and channel after upgrade
- performs token transfer over a new channel

A follow up PR can make sure the tokens can be sent back over the channels and are unwound into the native denom.

There is one hack/cavet with how the second chain is being deployed.

Because the global SDK config is sealed and the bech32 prefix is "celestia", some SDK functions which depend on the global type explode when dealing with a chain that has a different prefix e.g. "cosmos", what I have done for now is created a modified version of the dummy simapp from the ibc-go repo and changed the prefix to "celestia" to work around this. (it lives at `ghcr.io/chatton/ibc-go-simd:v8.5.0` build and pushed locally from [this branch](https://github.com/chatton/ibc-go/tree/release/v8.5.x)) 

It currently is sitting in my ghcr registry, this can either be moved into the celestiaorg ghcr registry, or, ideally, we find a workaround to enable different bech32 prefixes on chains.

This would likely involve "unsealing" or conditionally sealing the global config in celestia-app which doesn't seam ideal either.

This test still successfully verifies ibc transfers since it technically doesn't matter if the bech32 prefix is the same or not.

closes #5387

### DIFF
--- a/test/docker-e2e/go.mod
+++ b/test/docker-e2e/go.mod
@@ -271,8 +271,8 @@ require (
 replace (
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
 	github.com/celestiaorg/celestia-app/v6 => ../..
-	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v1.57.2-tm-v0.38.17
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.29.4-sdk-v0.50.14
+	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v1.57.3-tm-v0.38.17
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.29.5-sdk-v0.50.14
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
 	// celestia-core(v0.34.x): used for multiplexing abci v1 requests
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -31,6 +31,14 @@ pull_request_rules:
       backport:
         branches:
           - v4.x
+  - name: backport patches to v5.x branch
+    conditions:
+      - base=main
+      - label=backport:v5.x
+    actions:
+      backport:
+        branches:
+          - v5.x
   - name: forward-port patches to main branch
     conditions:
       - label=forwardport:main
--- a/.github/workflows/dependabot-tidy.yml
+++ b/.github/workflows/dependabot-tidy.yml
@@ -1,0 +1,50 @@
+name: dependabot-tidy
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - '**/go.mod'
+      - '**/go.sum'
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  tidy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 #v5.5.0
+        with:
+          go-version-file: "go.mod"
+
+      - name: Run go mod tidy in all modules
+        run: |
+          find . -name 'go.mod' -execdir sh -c '
+            echo "Running go mod tidy in $(pwd)"
+            go mod tidy
+          ' \;
+
+      - name: Commit and push tidy changes (only Dependabot PRs)
+        if: github.event_name == 'pull_request' && github.actor == 'dependabot[bot]'
+        run: |
+          git config user.name "github-actions"
+          git config user.email "github-actions@github.com"
+
+
+          git config --global credential.helper store
+          echo "https://oauth2:${{ secrets.DEPENDABOT_TIDY_TOKEN }}@github.com" > ~/.git-credentials
+
+          git add .
+
+          if git diff --cached --quiet; then
+            echo "No changes to commit after go mod tidy."
+            exit 0
+          fi
+
+          git commit -m "chore: run go mod tidy on all modules"
+          git push --force origin "HEAD:${{ github.head_ref }}"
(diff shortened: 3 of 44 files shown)


test/docker-e2e/dockerchain/testchain.go
```diff
@@ -51,15 +52,22 @@ func NewCelestiaChainBuilder(t *testing.T, cfg *Config) *tastoradockertypes.Chai
 			Build()
 	}
 
+	// register the first wallet as the faucet wallet
+	addr, err := records[0].GetAddress()
+	require.NoError(t, err, "failed to get address from keyring record")
+	// Create the wallet with all required fields
+	faucetWallet := tastoradockertypes.NewWallet(addr, addr.String(), "celestia", records[0].Name)
```
chatton, line 59: the implementation in tastora uses the faucet wallet to fund relayer wallets, we  can just use the first validator for this purpose.

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -51,15 +52,22 @@ func NewCelestiaChainBuilder(t *testing.T, cfg *Config) *tastoradockertypes.Chai
...
+
 	return tastoradockertypes.NewChainBuilder(t).
 		WithName("celestia"). // just influences home directory on the host.
 		WithChainID(cfg.Genesis.ChainID).
 		WithDockerClient(cfg.DockerClient).
 		WithDockerNetworkID(cfg.DockerNetworkID).
-		WithImage(tastoradockertypes.NewDockerImage(cfg.Image, cfg.Tag, "10001:10001")).
+		WithImage(tastoracontainertypes.NewImage(cfg.Image, cfg.Tag, "10001:10001")).
```
chatton, line 66: some structural changes in tastora to split out the original mono-package

test/docker-e2e/e2e_ibc_test.go
```diff
@@ -0,0 +1,347 @@
...
+// newSimappChainBuilder builds a standard simapp chain without token filters
+func newSimappChainBuilder(t *testing.T, cfg *dockerchain.Config) *tastoradockertypes.ChainBuilder {
+	encodingConfig := testutil.MakeTestEncodingConfig(app.ModuleEncodingRegisters...)
+	return tastoradockertypes.NewChainBuilder(t).
+		WithEncodingConfig(&encodingConfig).
+		WithName("simapp").
+		WithChainID("chain-b").
+		// TODO: this is a custom built simapp that has the bech32prefix as "celestia" as a workaround for the global
+		// SDK config not being usable when 2 chains have a different beck32 preix (e.g. "celestia" and "cosmos" ) because it is sealed.
+		WithImage(tastoracontainertypes.NewImage("ghcr.io/chatton/ibc-go-simd", "v8.5.0", "1000:1000")).
```
chatton, line 338: this is the main thing that needs to be resolved before merging.

If we are happy to merge this for now we can create an issue to immediately move this image somewhere else or figure out a solution to the global sdk config.

test/docker-e2e/e2e_ibc_test.go
```diff
@@ -0,0 +1,347 @@
...
+// This reuses the existing upgrade logic from e2e_upgrade_test.go
+func (s *CelestiaTestSuite) upgradeChain(ctx context.Context, chain tastoratypes.Chain, targetAppVersion uint64) {
+	validatorNode := chain.GetNodes()[0]
+	cfg := s.celestiaCfg
+	kr := cfg.Genesis.Keyring()
+
+	records, err := kr.List()
+	s.Require().NoError(err, "failed to list keyring records")
+
+	upgradeHeight := s.signalAndGetUpgradeHeight(ctx, chain, validatorNode, cfg, records, targetAppVersion)
```
chatton, line 264: re-uses upgrade function from the upgrade tests.

test/docker-e2e/e2e_ibc_test.go
```diff
@@ -0,0 +1,347 @@
...
+	s.transferTokens(ctx, chainA, chainB, channel, "utia", 100000)
+}
+
+// transferTokens performs an IBC token transfer from Celestia to simapp
+func (s *CelestiaTestSuite) transferTokens(ctx context.Context, sourceChain, destChain tastoratypes.Chain, channel ibc.Channel, denom string, amount int64) {
+	sourceWallet := sourceChain.GetFaucetWallet()
+	destWallet := destChain.GetFaucetWallet()
+	ibcTransfer := s.createIBCTransferMsg(sourceWallet, destWallet, channel, denom, amount)
+
+	// Submit transaction and verify results - always use TxClient since we only transfer from Celestia
```
chatton, line 200: note: when we want to send a msg transfer back, we can't use the tx client as there are some hard coded assumptions we are dealing the celestia chain and it does not work generically with any other chains. Part of the reason to implement only transfer from celestia->chainb for now

test/docker-e2e/e2e_ibc_test.go
evan-forbes: LOL

evan-forbes (approved):
not blocking question

I'm not sure on the best way to index these tests with respect to upgrades. Yes we definitely want to test IBC, and there is an edge case where IBC breaks after upgrading but not when starting a chain from scratch. We are testing both here afaiu, but if we want to test all functionality after upgrading then maybe it makes sense to do all of that in the upgrade test and not upgrade in each e2e test 

🤷 still unsure

test/docker-e2e/e2e_ibc_test.go
chatton: that's a good way to sum up the situation :D

test/docker-e2e/go.mod
```diff
@@ -270,11 +270,11 @@ require (
 
 replace (
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
+	// goleveldb: canonical version
```
rootulp: [nit] the comment got moved to an incorrect place. IMO seems safe to remove this comment b/c it doesn't help maintainers

```suggestion
```

rootulp (approved):

PR #5412, opened 2025-08-01
test(e2e): validator liveliness check

### LINKED ISSUE #5380: Check for block production and halting in e2e tests
spun out of [this comment ](https://github.com/celestiaorg/celestia-app/issues/5220#issuecomment-3109070187)

if possible, before finishing the tastora tests, we should check if all validators produced blocks for each version that validator used in the test, and we also need to check if any node in the network halted. If either of these things are false, then the test should fail. 

Note: Depending on the number of block producers, there could be some flakiness if not enough blocks are produced per version. Therefore, the number of block the network produces at each version should be at least 3x that of the number of potential block producers. Assuming even voting power, we should then expect each validator to have at least 3 chances to produce a block.

mojtaba-esk:
Closes #5380 

**Note**
The assertion is added to `TestCelestiaAppUpgrade` test and if needed we can add it to the other tests as well.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -9,18 +9,18 @@ before:
     - go mod tidy
     # These hooks get run once. We want to explicitly download every embedded binary so we can build the multiplexer
     # for every architecture. These versions must be updated at the same time as the version in the Makefile.
-    - ./scripts/download_binary.sh celestia-app_Darwin_arm64.tar.gz celestia-app_darwin_v3_arm64.tar.gz v3.10.5
-    - ./scripts/download_binary.sh celestia-app_Linux_arm64.tar.gz celestia-app_linux_v3_arm64.tar.gz v3.10.5
-    - ./scripts/download_binary.sh celestia-app_Darwin_x86_64.tar.gz celestia-app_darwin_v3_amd64.tar.gz v3.10.5
-    - ./scripts/download_binary.sh celestia-app_Linux_x86_64.tar.gz celestia-app_linux_v3_amd64.tar.gz v3.10.5
-    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_arm64.tar.gz celestia-app_darwin_v4_arm64.tar.gz v4.1.0-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Linux_arm64.tar.gz celestia-app_linux_v4_arm64.tar.gz v4.1.0-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_x86_64.tar.gz celestia-app_darwin_v4_amd64.tar.gz v4.1.0-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Linux_x86_64.tar.gz celestia-app_linux_v4_amd64.tar.gz v4.1.0-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_arm64.tar.gz celestia-app_darwin_v5_arm64.tar.gz v5.0.1-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Linux_arm64.tar.gz celestia-app_linux_v5_arm64.tar.gz v5.0.1-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_x86_64.tar.gz celestia-app_darwin_v5_amd64.tar.gz v5.0.1-arabica
-    - ./scripts/download_binary.sh celestia-app-standalone_Linux_x86_64.tar.gz celestia-app_linux_v5_amd64.tar.gz v5.0.1-arabica
+    - ./scripts/download_binary.sh celestia-app_Darwin_arm64.tar.gz celestia-app_darwin_v3_arm64.tar.gz v3.10.6
+    - ./scripts/download_binary.sh celestia-app_Linux_arm64.tar.gz celestia-app_linux_v3_arm64.tar.gz v3.10.6
+    - ./scripts/download_binary.sh celestia-app_Darwin_x86_64.tar.gz celestia-app_darwin_v3_amd64.tar.gz v3.10.6
+    - ./scripts/download_binary.sh celestia-app_Linux_x86_64.tar.gz celestia-app_linux_v3_amd64.tar.gz v3.10.6
+    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_arm64.tar.gz celestia-app_darwin_v4_arm64.tar.gz v4.1.0
+    - ./scripts/download_binary.sh celestia-app-standalone_Linux_arm64.tar.gz celestia-app_linux_v4_arm64.tar.gz v4.1.0
+    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_x86_64.tar.gz celestia-app_darwin_v4_amd64.tar.gz v4.1.0
+    - ./scripts/download_binary.sh celestia-app-standalone_Linux_x86_64.tar.gz celestia-app_linux_v4_amd64.tar.gz v4.1.0
+    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_arm64.tar.gz celestia-app_darwin_v5_arm64.tar.gz v5.0.1
+    - ./scripts/download_binary.sh celestia-app-standalone_Linux_arm64.tar.gz celestia-app_linux_v5_arm64.tar.gz v5.0.1
+    - ./scripts/download_binary.sh celestia-app-standalone_Darwin_x86_64.tar.gz celestia-app_darwin_v5_amd64.tar.gz v5.0.1
+    - ./scripts/download_binary.sh celestia-app-standalone_Linux_x86_64.tar.gz celestia-app_linux_v5_amd64.tar.gz v5.0.1
 
 builds:
   - id: darwin-amd64-multiplexer
--- a/Makefile
+++ b/Makefile
@@ -28,9 +28,9 @@ BUILD_FLAGS_MULTIPLEXER := -tags "ledger multiplexer" -ldflags '$(ldflags)'
 # internal/embedding/data.go
 # .goreleaser.yaml
 # docker/multiplexer.Dockerfile
-CELESTIA_V3_VERSION := v3.10.5
-CELESTIA_V4_VERSION := v4.1.0-arabica
-CELESTIA_V5_VERSION := v5.0.1-arabica
+CELESTIA_V3_VERSION := v3.10.6
+CELESTIA_V4_VERSION := v4.1.0
+CELESTIA_V5_VERSION := v5.0.1
 
 ## help: Get more info on make commands.
 help: Makefile
@@ -52,7 +52,7 @@ build: mod
 ifeq ($(DOWNLOAD),true)
 	@$(MAKE) download-v3-binaries
 	@$(MAKE) download-v4-binaries
-	@$(MAKE) download-v5-binaries	
+	@$(MAKE) download-v5-binaries
 endif
 	@mkdir -p build/
 	@echo "--> Building build/celestia-appd with multiplexer enabled"
--- a/docker/multiplexer.Dockerfile
+++ b/docker/multiplexer.Dockerfile
@@ -18,9 +18,9 @@ ARG UPGRADE_HEIGHT_DELAY
 ARG CELESTIA_APP_REPOSITORY=ghcr.io/celestiaorg/celestia-app-standalone
 # NOTE: This version must be updated at the same time as the version in the
 # Makefile.
-ARG CELESTIA_VERSION_V3="v3.10.5"
-ARG CELESTIA_VERSION_V4="v4.1.0-arabica"
-ARG CELESTIA_VERSION_V5="v5.0.1-arabica"
+ARG CELESTIA_VERSION_V3="v3.10.6"
+ARG CELESTIA_VERSION_V4="v4.1.0"
+ARG CELESTIA_VERSION_V5="v5.0.1"
 
 # Stage 1: this base image contains already released v3 binaries which can be embedded in the multiplexer.
 FROM ${CELESTIA_APP_REPOSITORY}:${CELESTIA_VERSION_V3} AS base-v3
--- a/docs/architecture/adr-015-namespace-id-size.md
+++ b/docs/architecture/adr-015-namespace-id-size.md
@@ -155,7 +155,7 @@ If the namespace size is increased, whenever a SHA256 invocation takes place (e.
 | 33                     | 197                                                                                                                                   |
 
 The data size is calculated as follows:
-`SHA256(domain_sep || left_min || left_max || left_data_hash || right_min || right_max || right_data_hash)` = 1 + namespaceSize + namespaceSize + 32 + namespaceSize + namespaceSize + 32 which simplifies to (4 \* namespaceSIze) + (2 \* 32) + 1.
+`SHA256(domain_sep || left_min || left_max || left_data_hash || right_min || right_max || right_data_hash)` = 1 + namespaceSize + namespaceSize + 32 + namespaceSize + namespaceSize + 32 which simplifies to (4 \* namespaceSize) + (2 \* 32) + 1.
 
 Hashing has a high performance impact in ZK contexts. Since a larger namespace size results in a larger data size for the SHA256 operation, increasing the namespace size has a negative performance impact on the cost to perform the SHA256 operation in a ZK context.
 
--- a/docs/architecture/adr-019-strict-inflation-schedule.md
+++ b/docs/architecture/adr-019-strict-inflation-schedule.md
@@ -117,7 +117,7 @@ Note: we should use DecCoin to avoid rounding errors. See <https://github.com/os
 
 Q: What is the skew between `BlocksPerYear` on popular Cosmos SDK chains and the actual number of blocks per year?
 
-Cosmos Hub (a.k.a Gaia) sets `BlocksPerYear` to 4,360,000 [here](https://github.com/cosmos/gaia/blob/8a522e98a2863205cf02fb97f8ad27d091670b9d/docs/governance/current-parameters.json#L86). Cosmos Hub has a block time 6.353 ms. Numia data doesn't have complete data for 2021 so we'll examine 2022:
+Cosmos Hub (a.k.a Gaia) sets `BlocksPerYear` to 4,360,000 [here](https://github.com/cosmos/gaia/blob/8a522e98a2863205cf02fb97f8ad27d091670b9d/docs/governance/current-parameters.json#L86). Cosmos Hub has a block time 6.353 s. Numia data doesn't have complete data for 2021 so we'll examine 2022:
 
 | Year | BlocksPerYear | Actual # of Blocks | Skew  |
 |------|---------------|--------------------|-------|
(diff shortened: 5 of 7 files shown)


test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+// stalled below endHeight.
+//
+// Call this at the *end* of an E2E test once you're sure endHeight is final.
+func (s *CelestiaTestSuite) AssertHealthy(
+	ctx context.Context,
+	chain tastoratypes.Chain,
+	startHeight, endHeight int64,
+) error {
+	if endHeight <= startHeight {
+		return fmt.Errorf("invalid height range %d → %d", startHeight, endHeight)
```
rootulp: [nit] 

```suggestion
		return fmt.Errorf("invalid height range %d to %d", startHeight, endHeight)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+	}
+
+	s.T().Logf("Checking validator health from height %d to %d", startHeight, endHeight)
+
+	// choose the first node as our RPC reader
+	reader, err := chain.GetNodes()[0].GetRPCClient()
+	if err != nil {
+		return fmt.Errorf("reader RPC: %w", err)
```
rootulp:
```suggestion
		return fmt.Errorf("failed to get RPC client: %w", err)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+	validators, err := reader.Validators(ctx, &endHeight, nil, nil)
+	if err != nil {
+		return fmt.Errorf("validators query: %w", err)
+	}
+
+	s.T().Logf("Checking %d validators for proposer activity", len(validators.Validators))
+	for _, val := range validators.Validators {
+		if _, ok := proposers[val.Address.String()]; !ok {
+			return fmt.Errorf("validator %s never proposed between %d-%d",
+				val.Address.String(), startHeight, endHeight)
```
rootulp:
```suggestion
			return fmt.Errorf("validator %s never proposed block from %d to %d",
				val.Address.String(), startHeight, endHeight)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+		}
+	}
+
+	// 3. ensure no node halted below endHeight
+	for i, n := range chain.GetNodes() {
+		c, err := n.GetRPCClient()
+		if err != nil {
+			return fmt.Errorf("node %d RPC: %w", i, err)
```
rootulp:
```suggestion
			return fmt.Errorf("failed to get RPC client for index %d: %w", i, err)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+		}
+		status, err := c.Status(ctx)
+		if err != nil {
+			return fmt.Errorf("node %d status: %w", i, err)
```
rootulp:
```suggestion
			return fmt.Errorf("failed to get node %d status: %w", i, err)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+		}
+		if status.SyncInfo.LatestBlockHeight < endHeight {
+			return fmt.Errorf("node %d halted at %d (expected ≥ %d)",
+				i, status.SyncInfo.LatestBlockHeight, endHeight)
+		}
+	}
+
+	s.T().Logf("All validators healthy: proposed blocks and no nodes halted")
```
rootulp:
```suggestion
	s.T().Logf("All validators proposed at least one blocks and no validator halted")
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -97,15 +97,15 @@ func (s *CelestiaTestSuite) runUpgradeTest(ImageTag string, baseAppVersion, targ
 	// Signal for upgrade and get the upgrade height
 	upgradeHeight := s.signalAndGetUpgradeHeight(ctx, chain, validatorNode, cfg, records, targetAppVersion)
 
-	// Get current height
+	// Record current height - we'll use it later for health assertions
```
rootulp:
```suggestion
	// Record start height - we'll use it later for health assertions
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -97,15 +97,15 @@ func (s *CelestiaTestSuite) runUpgradeTest(ImageTag string, baseAppVersion, targ
...
 	status, err := rpcClient.Status(ctx)
 	s.Require().NoError(err, "failed to get node status")
-	currentHeight := status.SyncInfo.LatestBlockHeight
+	startHeight := status.SyncInfo.LatestBlockHeight
 
-	s.T().Logf("Current height: %d, Upgrade height: %d", currentHeight, upgradeHeight)
+	s.T().Logf("Current height: %d, Upgrade height: %d", startHeight, upgradeHeight)
```
rootulp: [blocking] the log is now confusing because this variable name was modified but the log line wasn't

```suggestion
	s.T().Logf("Start height: %d, Upgrade height: %d", startHeight, upgradeHeight)
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -119,6 +119,17 @@ func (s *CelestiaTestSuite) runUpgradeTest(ImageTag string, baseAppVersion, targ
 	// Sanity check: Test bank send after upgrade
 	s.T().Log("Testing bank send functionality after upgrade")
 	testBankSend(s.T(), chain, cfg)
+
+	// Validators health check
+	finalStatus, err := rpcClient.Status(ctx)
+	s.Require().NoError(err, "failed to get final status")
+	endHeight := finalStatus.SyncInfo.LatestBlockHeight
+
+	s.T().Logf("Asserting validator health for blocks %d → %d", startHeight, endHeight)
```
rootulp:
```suggestion
	s.T().Logf("Asserting validator health for blocks %d to %d", startHeight, endHeight)
```

rootulp (commented):
Overall seems good but:
- the error messages could be improved
- the logs could be improved

rootulp (commented):
This seems to only call AssertHealthy in the e2e upgrade test but the second criteria of the original issue is

> Call this helper in t.Cleanup of each E2E test (opt-out flag available).

so should the AssertHealthy be added to more e2e tests?

mojtaba-esk:
> This seems to only call AssertHealthy in the e2e upgrade test but the second criteria of the original issue is
> 
> > Call this helper in t.Cleanup of each E2E test (opt-out flag available).
> 
> so should the AssertHealthy be added to more e2e tests?

I think it adds value if we add it to the rest of tastora based tests. what do you and @evan-forbes think?

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
 		}
 	}
 }
+
+// AssertHealthy checks that (a) every validator proposed at least one block
+// between startHeight (exclusive) and endHeight (inclusive) and (b) no node
+// stalled below endHeight.
+//
+// Call this at the *end* of an E2E test once you're sure endHeight is final.
+func (s *CelestiaTestSuite) AssertHealthy(
```
evan-forbes: not blocking

we could be more specific that "healthy" which could mean a large number of things. For example "CheckLiveness" makes a lot of sense given that's what we're looking for here

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+	startHeight, endHeight int64,
+) error {
+	if endHeight <= startHeight {
+		return fmt.Errorf("invalid height range %d to %d", startHeight, endHeight)
+	}
+
+	s.T().Logf("Checking validator health from height %d to %d", startHeight, endHeight)
+
+	// choose the first node as our RPC reader
+	reader, err := chain.GetNodes()[0].GetRPCClient()
```
evan-forbes: not blocking

its rare for me to care about var names, but reader is a non-standard name for a client since reader usually means that something is like an io.Reader. this will throw off 90% of go devs so it should really be changed imo. the client does not fulfill the io.Reader interface nor should it be used as such

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +171,72 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+	if err != nil {
+		return fmt.Errorf("failed to get RPC client: %w", err)
+	}
+
+	// 1. gather proposer addresses per block
+	proposers := make(map[string]struct{})
+	for h := startHeight + 1; h <= endHeight; h++ {
+		blockRes, err := reader.Block(ctx, &h)
```
evan-forbes: not blocking

this is downloading the entire chain more or less

I think we can use headers here

evan-forbes (commented):
per original issue, we likely want to do this for most if not all test. If this cannot be applied to all tests, then being explicit with the test is likely better than having an implicit assertion

test/docker-e2e/e2e_test.go
mojtaba-esk: oh right, good catch, for larger tests this wont be efficient. will fix it.

test/docker-e2e/e2e_test.go
```diff
@@ -22,7 +23,22 @@ import (
...
+
+	// Liveness check constants
+	defaultBlocksPerValidator = 3                    // Minimum blocks each validator should propose for liveness validation
+	validatorAddrTruncateLen  = 8                    // Length to truncate validator addresses for display
+	validatorNameFormat       = "validator_%s"       // Format string for validator display names
+	nodeStatusFormat          = "node_%d(height_%d)" // Format string for node status messages
+
+	// Node type constants (matching tastora framework's string representations)
+	validatorNodeTypeString = "val" // String representation of validator nodes
+	fullNodeTypeString      = "fn"  // String representation of full nodes
```
rootulp, line 35: [nit]

- it's a code smell to include the type of the variable in the variable name. Is it possible to remove `String` from these variable names?
- IMO `fn` is an overloaded term. When I see it, I think "function". Can we use the un-abbreviated strings: "validator" and "full-node" here? 
- I've tried to use "full-node" less because it can easily be confused with a celestia-node (a.k.a DA) full node. The terminology I prefer for celestia-app: all nodes are consensus nodes. Some consensus nodes are validators.

test/docker-e2e/e2e_test.go
```diff
@@ -22,7 +23,22 @@ import (
...
+
+	// Log message formats
+	waitingForBlocksLogFormat    = "Waiting for %d more blocks to meet minimum requirement (%d produced, %d required for %d validators × %d blocks each)"
```
rootulp, line 38: This log message is used in one location. Can it be inlined instead of creating a constant for it?

test/docker-e2e/e2e_test.go
```diff
@@ -22,7 +23,22 @@ import (
...
+	blockRequirementMetLogFormat = "Minimum block requirement already met: %d blocks ≥ %d required (%d validators × %d blocks each)"
```
rootulp, line 39: This log message is used in one location. Can it be inlined?

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +187,256 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+
+	if err := s.validateAllValidatorsProposed(startValidators, endValidators, proposers, startHeight, endHeight); err != nil {
+		return err
+	}
+
+	if err := s.validateNodesNotHalted(ctx, chain, endHeight); err != nil {
+		return err
+	}
+
+	s.T().Logf("✅ Liveness check passed: all validators proposed blocks and all nodes reached height %d", endHeight)
```
rootulp, line 229: Please avoid using emojis in logs.

```suggestion
	s.T().Logf("Liveness check passed: all validators proposed blocks and all nodes reached height %d", endHeight)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +187,256 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+	validators, err := rpcClient.Validators(ctx, &currentHeight, nil, nil)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get validators: %w", err)
+	}
+	numValidators := len(validators.Validators)
+	minBlocksRequired := int64(numValidators * defaultBlocksPerValidator)
+
+	if blocksProduced >= minBlocksRequired {
+		s.T().Logf(blockRequirementMetLogFormat,
+			blocksProduced, minBlocksRequired, numValidators, defaultBlocksPerValidator)
```
rootulp, line 258: [uber nit] I find it more difficult to read code when there are random newlines introduced. This applies to many other instances in this PR

```suggestion
		s.T().Logf(blockRequirementMetLogFormat, blocksProduced, minBlocksRequired, numValidators, defaultBlocksPerValidator)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +187,256 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+		// Check blocks after startHeight (exclusive) through endHeight (inclusive)
+		minHeight = startHeight + 1
+		maxHeight = endHeight
+	)
+
+	for minHeight <= maxHeight {
+		// BlockchainInfo returns headers in descending order and may limit the range
+		blockchainInfo, err := rpcClient.BlockchainInfo(ctx, minHeight, maxHeight)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch block headers for range %d-%d: %w", minHeight, maxHeight, err)
```
rootulp, line 321: For consistency with other logs in this PR

```suggestion
			return nil, fmt.Errorf("failed to fetch block headers for range %d to %d: %w", minHeight, maxHeight, err)
```

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +187,256 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+		}
+
+		if len(blockchainInfo.BlockMetas) == 0 {
+			return nil, fmt.Errorf("no block headers returned for range %d-%d", minHeight, maxHeight)
```
rootulp, line 325:
```suggestion
			return nil, fmt.Errorf("no block headers returned for range %d to %d", minHeight, maxHeight)
```

test/docker-e2e/e2e_test.go
```diff
@@ -22,7 +23,22 @@ import (
 const (
 	txsimImage = "ghcr.io/celestiaorg/txsim"
 	txSimTag   = "v4.0.10-arabica"
-	homeDir    = "/var/cosmos-chain/celestia"
+
+	// Liveness check constants
+	defaultBlocksPerValidator = 3                    // Minimum blocks each validator should propose for liveness validation
+	validatorAddrTruncateLen  = 8                    // Length to truncate validator addresses for display
```
rootulp, line 29: [question] Why truncate the address to 8 characters? Seems possible to print the entire address.

test/docker-e2e/e2e_test.go
```diff
@@ -22,7 +23,22 @@ import (
...
+	validatorNameFormat       = "validator_%s"       // Format string for validator display names
+	nodeStatusFormat          = "node_%d(height_%d)" // Format string for node status messages
```
rootulp, line 31: [question] these format strings seem like a code smell. If they're used in format strings then can they be inlined? Without more context, I don't know what the %s or %d mean in these logs so it's more confusing to separate this from the actual usage.

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +187,256 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+	var (
+		allValidators  = make(map[string]struct{})
+		validatorNames = make(map[string]string) // for better error reporting
+	)
+
+	// Add start validators
+	for _, val := range startValidators.Validators {
+		addr := val.Address.String()
+		allValidators[addr] = struct{}{}
+		validatorNames[addr] = fmt.Sprintf(validatorNameFormat, addr[:validatorAddrTruncateLen]) // shortened for readability
```
rootulp, line 379: [nit] can we eliminate the separate `validatorNames` map by just using the entire address in `allValidators`? If yes that seem desirable to reduce complexity (maintaining multiple maps) and unnecessary constants (`validatorAddrTruncateLen`)

test/docker-e2e/e2e_test.go
```diff
@@ -171,3 +187,256 @@ func (s *CelestiaTestSuite) WaitForSync(ctx context.Context, statusClient rpccli
...
+// especially when added during the test (e.g., state sync nodes, full nodes).
+func (s *CelestiaTestSuite) validateNodesNotHalted(
+	ctx context.Context,
+	chain tastoratypes.Chain,
+	endHeight int64,
+) error {
+	var haltedNodes []string
+	for i, n := range chain.GetNodes() {
+		// Only check validator nodes for height consistency
+		if n.GetType() != validatorNodeTypeString {
```
rootulp, line 419: [question] can we export the output of `GetType` from tastora and import it here? Or even better, export a method from the Tastora node type like `.IsValidator()` so that it can be used here?

Seems brittle to re-define a constant like `validatorNodeTypeString` here

rootulp (approved):

evan-forbes:
@mojtaba-esk can you open a flup PR or issue to handle remaining feedback? there are still a log of things that could be cleaned up and I don't want to lose that context

tbc, please don't open a single issue per comment. it could be a PR to address the remaining feedback or an issue that keeps track of all comments

test/docker-e2e/e2e_test.go
mojtaba-esk: True. A followup issue is created to fix it: https://github.com/celestiaorg/tastora/issues/91 and I will be opening a PR for it. After a Tastora release, we can do a refactor in the e2e tests.

test/docker-e2e/e2e_test.go
mojtaba-esk: unfortunately the output of `GetType` was string and the actual constant was not exported. So I am fixing it in a Tastora PR.

test/docker-e2e/e2e_test.go
mojtaba-esk: just to avoid verbose logs, I can put it back

test/docker-e2e/e2e_test.go
mojtaba-esk: technically we can do that, it is only used to format error messages, but we can eliminate it entirely and just use the raw addresses directly.

PR #5438, opened 2025-08-04
fix: cross-platform `sed` syntax for better compatibility

voronor:
## Overview
noticed the `sed -i'.bak'` syntax was causing issues on GNU/Linux, since it expects a space after `-i` when providing a backup extension.
replaced it with the more portable `-i.bak`, which works consistently across both macOS and Linux.

p.s.everything else stays the same - just fixing the compatibility hiccup.

### DIFF
--- a/scripts/single-node.sh
+++ b/scripts/single-node.sh
@@ -67,19 +67,19 @@ createGenesis() {
         > /dev/null 2>&1 # Hide output to reduce terminal noise
 
     # Override the default RPC server listening address
-    sed -i'.bak' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' "${APP_HOME}"/config/config.toml
+    sed -i.bak 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' "${APP_HOME}"/config/config.toml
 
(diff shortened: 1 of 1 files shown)


rootulp (commented):
Based on ChatGPT, it seems like `sed -i'.bak'` should work too but I'm not opposed to this PR if it works. Any chance you tested on both Linux and MacOS? I can test on MacOS if you haven't.

Screenshot 2025-08-04 at 2 27 13 PM

voronor:
[rootulp](https://github.com/rootulp), ty for the feedback!
yes, I tested this change on both Linux and macOS. on Linux, `sed -i'.bak'` caused issues because GNU sed expects a space after `-i` when specifying a backup extension, while `-i.bak` works consistently on both platforms.

evan-forbes (approved):
works for me on linux

I think the old script also worked for me as I used it to debug something a month back or so

voronor:
[evan-forbes](https://github.com/evan-forbes), ty for checking!
yeah, the old syntax can work on some Linux setups, but it depends on sed version and shell.
the `-i.bak` form is just safer and works consistently on both macOS and Linux without issues.

just making sure it runs smoothly everywhere!

PR #5441, opened 2025-08-05
feat: add a cli sub command to migrate configs for specific versions

### LINKED ISSUE #5366: Add a cli command to edit config for v6
we should add a cli command that will modify the user's config for them to update all required params for v6.

this includes peer count, MaxTxSize, send and receive rates, min gas price

the command should ideally also log which values it changes

evan-forbes:
## Overview

adds a quick cli command to migrate. Might be over-engineered, but also adds a way to keep track of configs / migrations per version. tbh, not sure that's needed.

closes #5366 but note that I'm not sure this includes all the configs that we need to modify

### DIFF
--- a/cmd/celestia-appd/cmd/migrate_config.go
+++ b/cmd/celestia-appd/cmd/migrate_config.go
@@ -1,0 +1,152 @@
+package cmd
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/cometbft/cometbft/config"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/celestiaorg/celestia-app/v6/app"
+)
+
+// ConfigMigrator defines a function type for applying version-specific migrations
+type ConfigMigrator func(*config.Config, *serverconfig.Config) (*config.Config, *serverconfig.Config)
+
+// migrationRegistry maps version strings to their corresponding migration functions
+var migrationRegistry = map[string]ConfigMigrator{
+	"v6": applyV6Migrations,
+}
+
+// migrateConfigCmd returns the migrate-config command that migrates
+// configuration files based on the target version.
+func migrateConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "migrate-config [version]",
+		Short:   "Migrate configuration files to a target version",
+		Long:    "Migrate configuration files (config.toml and app.toml) to be compatible with a target application version.",
+		Example: "celestia-appd migrate-config v6 --home ~/.celestia-app",
+		Args:    cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			homeDir, err := cmd.Flags().GetString(flags.FlagHome)
+			if err != nil {
+				return err
+			}
+
+			targetVersion := args[0]
+			return migrateConfig(homeDir, targetVersion)
+		},
+	}
+
+	cmd.Flags().String(flags.FlagHome, app.NodeHome, "The application home directory")
+	return cmd
+}
+
+// migrateConfig performs the actual migration of configuration files.
+func migrateConfig(homeDir, targetVersion string) error {
+	fmt.Printf("Migrating configuration files to version %s...\n", targetVersion)
+
+	migrator, exists := migrationRegistry[targetVersion]
+	if !exists {
+		return fmt.Errorf("unsupported target version: %s. Supported versions: %v", targetVersion, getSupportedVersions())
+	}
+
+	configDir := filepath.Join(homeDir, "config")
+	cometConfigPath := filepath.Join(configDir, "config.toml")
+	appConfigPath := filepath.Join(configDir, "app.toml")
+
+	cometConfig, err := loadCometBFTConfig(cometConfigPath, homeDir)
+	if err != nil {
+		return fmt.Errorf("failed to load CometBFT config: %w", err)
+	}
+
+	serverConfig, err := loadServerConfig(appConfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to load server config: %w", err)
+	}
+
+	fmt.Printf("Loaded configs successfully. Applying %s migrations...\n", targetVersion)
+
+	cometConfig, serverConfig = migrator(cometConfig, serverConfig)
+
+	config.WriteConfigFile(cometConfigPath, cometConfig)
+	serverconfig.WriteConfigFile(appConfigPath, serverConfig)
+
+	fmt.Printf("Successfully migrated configuration files to version %s\n", targetVersion)
+	return nil
+}
+
+// loadCometBFTConfig loads the CometBFT configuration from config.toml
+func loadCometBFTConfig(configPath, homeDir string) (*config.Config, error) {
+	cfg := app.DefaultConsensusConfig()
+	cfg.SetRoot(homeDir)
+
+	v := viper.New()
+	v.SetConfigFile(configPath)
+	v.SetConfigType("toml")
+
+	if err := v.ReadInConfig(); err != nil {
+		return nil, fmt.Errorf("failed to read config file %s: %w", configPath, err)
+	}
+
+	if err := v.Unmarshal(cfg); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+
+	return cfg, nil
+}
+
+// loadServerConfig loads the Cosmos SDK server configuration from app.toml
+func loadServerConfig(configPath string) (*serverconfig.Config, error) {
+	cfg := app.DefaultAppConfig()
+
+	v := viper.New()
+	v.SetConfigFile(configPath)
+	v.SetConfigType("toml")
+
+	if err := v.ReadInConfig(); err != nil {
+		return nil, fmt.Errorf("failed to read config file %s: %w", configPath, err)
+	}
+
+	if err := v.Unmarshal(cfg); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+
+	return cfg, nil
+}
+
+// getSupportedVersions returns a list of supported migration versions
+func getSupportedVersions() []string {
+	versions := make([]string, 0, len(migrationRegistry))
+	for version := range migrationRegistry {
+		versions = append(versions, version)
+	}
+	return versions
+}
+
+// applyV6Migrations applies configuration changes needed for v6
+func applyV6Migrations(cmtCfg *config.Config, appCfg *serverconfig.Config) (*config.Config, *serverconfig.Config) {
+	fmt.Println("Applying v6 configuration migrations...")
+
+	defaultCfg := app.DefaultConsensusConfig()
+	cmtCfg.RPC.MaxBodyBytes = defaultCfg.RPC.MaxBodyBytes
+
+	cmtCfg.Mempool.TTLNumBlocks = defaultCfg.Mempool.TTLNumBlocks
+	cmtCfg.Mempool.TTLDuration = defaultCfg.Mempool.TTLDuration
+	cmtCfg.Mempool.MaxTxBytes = defaultCfg.Mempool.MaxTxBytes
+	cmtCfg.Mempool.MaxTxsBytes = defaultCfg.Mempool.MaxTxsBytes
+	cmtCfg.Mempool.Type = defaultCfg.Mempool.Type
+	cmtCfg.Mempool.MaxGossipDelay = defaultCfg.Mempool.MaxGossipDelay
+
+	cmtCfg.P2P.SendRate = defaultCfg.P2P.SendRate
+	cmtCfg.P2P.RecvRate = defaultCfg.P2P.RecvRate
+
+	defaultAppCfg := app.DefaultAppConfig()
+	appCfg.MinGasPrices = defaultAppCfg.MinGasPrices
+	appCfg.GRPC.MaxRecvMsgSize = defaultAppCfg.GRPC.MaxRecvMsgSize
+
+	return cmtCfg, appCfg
+}
--- a/cmd/celestia-appd/cmd/migrate_config.go
+++ b/cmd/celestia-appd/cmd/migrate_config.go
@@ -13,6 +13,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/celestiaorg/celestia-app/v6/app"
+	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
 )
 
 // ConfigMigrator defines a function type for applying version-specific migrations
@@ -27,11 +28,11 @@ var migrationRegistry = map[string]ConfigMigrator{
 // configuration files based on the target version.
 func migrateConfigCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "migrate-config [version]",
+		Use:     "migrate-config",
 		Short:   "Migrate configuration files to a target version",
 		Long:    "Migrate configuration files (config.toml and app.toml) to be compatible with a target application version.",
-		Example: "celestia-appd migrate-config v6 --home ~/.celestia-app",
-		Args:    cobra.ExactArgs(1),
+		Example: "celestia-appd migrate-config --home ~/.celestia-app\ncelestia-appd migrate-config --version v5 --home ~/.celestia-app",
+		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			homeDir, err := cmd.Flags().GetString(flags.FlagHome)
 			if err != nil {
@@ -43,13 +44,17 @@ func migrateConfigCmd() *cobra.Command {
 				return err
 			}
 
-			targetVersion := args[0]
+			targetVersion, err := cmd.Flags().GetString("version")
+			if err != nil {
+				return err
+			}
 			return migrateConfig(homeDir, targetVersion, backup)
 		},
 	}
 
 	cmd.Flags().String(flags.FlagHome, app.NodeHome, "The application home directory")
 	cmd.Flags().Bool("backup", false, "Create backups of config files before migrating")
+	cmd.Flags().String("version", fmt.Sprintf("v%d", appconsts.Version), "Target version for migration")
 	return cmd
 }
 
--- a/cmd/celestia-appd/cmd/migrate_config.go
+++ b/cmd/celestia-appd/cmd/migrate_config.go
@@ -2,7 +2,9 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/cometbft/cometbft/config"
 	"github.com/cosmos/cosmos-sdk/client/flags"
@@ -36,17 +38,23 @@ func migrateConfigCmd() *cobra.Command {
 				return err
 			}
 
+			backup, err := cmd.Flags().GetBool("backup")
+			if err != nil {
+				return err
+			}
+
 			targetVersion := args[0]
-			return migrateConfig(homeDir, targetVersion)
+			return migrateConfig(homeDir, targetVersion, backup)
 		},
 	}
 
 	cmd.Flags().String(flags.FlagHome, app.NodeHome, "The application home directory")
+	cmd.Flags().Bool("backup", false, "Create backups of config files before migrating")
 	return cmd
 }
 
 // migrateConfig performs the actual migration of configuration files.
-func migrateConfig(homeDir, targetVersion string) error {
+func migrateConfig(homeDir, targetVersion string, backup bool) error {
 	fmt.Printf("Migrating configuration files to version %s...\n", targetVersion)
 
 	migrator, exists := migrationRegistry[targetVersion]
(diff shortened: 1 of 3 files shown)


cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,213 @@
...
+	cmtCfg.Mempool.MaxTxsBytes = defaultCfg.Mempool.MaxTxsBytes
+	cmtCfg.Mempool.Type = defaultCfg.Mempool.Type
+	cmtCfg.Mempool.MaxGossipDelay = defaultCfg.Mempool.MaxGossipDelay
+
+	cmtCfg.P2P.SendRate = defaultCfg.P2P.SendRate
+	cmtCfg.P2P.RecvRate = defaultCfg.P2P.RecvRate
+
+	defaultAppCfg := app.DefaultAppConfig()
+	appCfg.MinGasPrices = defaultAppCfg.MinGasPrices
+	appCfg.GRPC.MaxRecvMsgSize = defaultAppCfg.GRPC.MaxRecvMsgSize
```
evan-forbes, line 211: got as many of these as I could think of, but we need to either go over again later and not close the op issue or open a flup

I had a function that got entirely too clever with reflection to optionally print out the struct tag if the value was changed, but figured that was overkill and deleted it. Definitely an option though if we want to print out if the value changed or not.

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,213 @@
...
+// migrationRegistry maps version strings to their corresponding migration functions
+var migrationRegistry = map[string]ConfigMigrator{
+	"v6": applyV6Migrations,
+}
+
+// migrateConfigCmd returns the migrate-config command that migrates
+// configuration files based on the target version.
+func migrateConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "migrate-config",
```
rootulp: [question] out of curiosity, why wasn't it possible to use the existing command? Maybe because that command only does app.toml and this command seems to do config.toml and app.toml.

```
$ celestia-appd config migrate --help
Migrate the contents of the Cosmos SDK app configuration (app.toml) to the specified version.
The output is written in-place unless --stdout is provided.
In case of any error in updating the file, no output is written.

Usage:
  celestia-appd config migrate [target-version] <app-toml-path> (options) [flags]```
```

[follow up] should we remove the command `$ celestia-appd config migrate` to avoid node operator confusion. They could easily mix these up

```
$ celestia-appd config migrate
$ celestia-appd migrate-config
```

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,213 @@
...
+		Short:   "Migrate configuration files to a target version",
+		Long:    "Migrate configuration files (config.toml and app.toml) to be compatible with a target application version.",
+		Example: "celestia-appd migrate-config --home ~/.celestia-app\ncelestia-appd migrate-config --version v5 --home ~/.celestia-app",
```
rootulp: [blocking] why does the example command use v5 if the only supported target is v6?

```suggestion
		Example: "celestia-appd migrate-config --home ~/.celestia-app\ncelestia-appd migrate-config --version v6 --home ~/.celestia-app",
```

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,213 @@
...
+			targetVersion, err := cmd.Flags().GetString("version")
+			if err != nil {
+				return err
+			}
+			return migrateConfig(homeDir, targetVersion, backup)
+		},
+	}
+
+	cmd.Flags().String(flags.FlagHome, app.NodeHome, "The application home directory")
+	cmd.Flags().Bool("backup", false, "Create backups of config files before migrating")
```
rootulp: [optional] can we remove the backup flag option and _always_ create backups? Given the default is false, I don't think anyone will actually use the backup option and it seems helpful if this command always left intact backup copies of the existing files.

or alternatively set the default backup flag to true?

cmd/celestia-appd/cmd/migrate_config_test.go
```diff
@@ -0,0 +1,144 @@
...
+			setupTestConfigFiles(t, configDir)
+
+			// Run migration
+			err := migrateConfig(tempDir, tt.version, false)
+
+			if tt.expectError {
+				require.Error(t, err)
+				if tt.errorContains != "" {
+					assert.Contains(t, err.Error(), tt.errorContains)
+				}
```
rootulp: [nit] there's only one test case that expects an error and it actually has errorContains

```suggestion
				assert.Contains(t, err.Error(), tt.errorContains)
```

cmd/celestia-appd/cmd/migrate_config_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		cfg.API.Enable = !cfg.API.Enable
+		originalAPIEnable := cfg.API.Enable
+
+		serverconfig.WriteConfigFile(appConfigPath, cfg)
+
+		// Load again and verify change persisted
+		reloadedCfg, err := loadServerConfig(appConfigPath)
+		require.NoError(t, err)
+		assert.Equal(t, originalAPIEnable, reloadedCfg.API.Enable)
```
rootulp: [blocking] I'm confused by this test. I would expect the opposite of this assertion. Something like

```suggestion
		assert.Equal(t, !originalAPIEnable, reloadedCfg.API.Enable)
```

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: yeah it doesn't cover most of the config changes we want and and it also covers things we don't. imo its also over engineered (even more than this one lol).

for example, look at this awful protobuf plan thing used to modify a toml

https://github.com/celestiaorg/cosmos-sdk/blob/c8e39a151d5be3fc45d7830b93b2295b9d6376ff/tools/confix/upgrade.go#L41-L44

ain't nobody got time for that

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: its just an example. I didn't update it after I noticed it as I didn't expect us to update this with every release

I'll happily change this to use the current appversion, however in the future I don't think its worth blocking on things like this.

cmd/celestia-appd/cmd/migrate_config_test.go
evan-forbes: hmmm good catch

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: sure we can definitely change the default to true. I like keeping the option to disable it due to permission or other issues.

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: > in the future I don't think its worth blocking on things like this.

tbc, could still be a nit, but blocking seems excessive unless its absolutely crucial that the example be always up to date

evan-forbes: also see that the default for the flag is already pointing to the current app version

rootulp: Yea fair shouldn't have been blocking. Can address in a follow up. I actually mentioned that in the overall review comment.

rootulp: > tbc, could still be a nit, but blocking seems excessive unless its absolutely crucial that the example be always up to date

The example doesn't need to be up to date but if I understand correctly the current example will result in an error because v5 isn't supported by this tool.

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// migrateConfigCmd returns the migrate-config command that migrates
+// configuration files based on the target version.
+func migrateConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "migrate-config",
+		Short:   "Migrate configuration files to a target version",
+		Long:    "Migrate configuration files (config.toml and app.toml) to be compatible with a target application version.",
+		Example: fmt.Sprintf("celestia-appd migrate-config --home ~/.celestia-app\ncelestia-appd migrate-config --version v%d --home ~/.celestia-app", appconsts.Version),
```
rootulp: [nit] This example can break if we bump `appconsts.Version` to 7 and don't include v7 in the `migrationRegistry` so IMO it would've been sufficient to use v6 here.

```suggestion
		Example: "celestia-appd migrate-config --home ~/.celestia-app\ncelestia-appd migrate-config --version v6 --home ~/.celestia-app",
```

rootulp (dismissed):

cmd/celestia-appd/cmd/migrate_config_test.go
evan-forbes: the wording was just weird and has been fixed. that was the updated value and we should expect the written config to be the same as it https://github.com/celestiaorg/celestia-app/pull/5441/commits/a25655017b1fe337edec44bb8f25baad2a69e32a

cmd/celestia-appd/cmd/migrate_config_test.go
evan-forbes: fixed in https://github.com/celestiaorg/celestia-app/pull/5441/commits/312d140161a89419096cea6ba7d1329dfed5dd6d

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: good point, and its fair to block on examples not working when reasonable, apologies I didn't see that as a requirement at first

not for this one as its reasonable to expect this one to work, but for other changes I don't think it should be expected that they work.

there are also a bunch that can't work, for example almost all of our tx subcommands ofc cannot work

```
celestia-appd tx staking delegate cosmosvalopers1l2rsakp388kuv9k8qzq6lrm9taddae7fpx59wm 1000stake --from mykey
```

```
celestia-appd tx blob pay-for-blob 0x00010203040506070809 0x48656c6c6f2c20576f726c6421 \
        --chain-id private \ <-- nothing uses this chain-id
        --from validator \
        --keyring-backend test \
        --fees 21000utia \
        --yes
```

```
Where proposal.json contains:

{
  "title": "Test Proposal",
  "description": "My awesome proposal",
  "type": "Text",
  "deposit": "10test"
}

```

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: https://github.com/celestiaorg/celestia-app/commit/a25655017b1fe337edec44bb8f25baad2a69e32a

cmd/celestia-appd/cmd/migrate_config.go
rootulp: Including broken examples in our CLI will cause operator confusion. We should make those examples work.

evan-forbes: my point here is just that some inherently require data or arguments to be filled in correctly. Ie a validators address. those cannot be included in the cli, so inherently we cannot always have examples working

when possible, like the one above, I agree that we should try to get them working :slightly_smiling_face:

cmd/celestia-appd/cmd/migrate_config.go
julienrbrt: Overriding that variable with your plan should be sufficient: https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/tools/confix/migrations.go#L26

You could remove all other configuration migration as well.
Where do you see protobuf? This is fully relying on toml parsing using this lib https://github.com/creachadair/tomledit

The default plan behavior is a simple diff check between current and target config (in the data folder).

tac0turtle: confix seems to have everything that is needed, also config.toml usage, https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/tools/confix/migrations.go#L16. curious what you found to be missing or over engineered?

evan-forbes: > curious what you found to be missing or over engineered?

perhaps I just don't have enough context of the problem, but we're just needing to update a few values. Presumably, there's a way to keep the constants in the application and define the "plan" etc here. I didn't think taking the time to figure that out was worth it when we could just load the config, change something, then save it again.

I wrote the above code in the same amount of time it took to write this comment, which was shorter than the time it takes to understand how to make a plan and modify both configs :shrug: 

I probably should have dug deeper into the code to figure out how, but I'm tired.

confix probably supports backing up old configs, but that was another thing I'd have to go find where as implementing that functionality is trivial.

are there examples in osmosis or gaia etc I could look at? I searched there after looking at the confix docs but didn't find anything so decided to cut my loses. the docs make it seem like this is strictly for migrating for sdk breaks in the config, not a general purpose tool one can use to make their own changes.

> Overriding that variable with your plan should be sufficient:

by overriding, I'm assuming there's a way to do that in celestia-app so we can use our consts etc, but I couldn't find examples in the docs or other app repos. there's no mention of comet configs either, so hopefully you understand me missing this ability.

> Where do you see protobuf? This is fully relying on toml parsing using this lib

good point! I incorrectly saw the message Plan and assumed there were two, one for upgrades and one here, but as you pointed out there's none in confix which is nice.

### tldr

overall, I don't want to overthink optionally updating a config. If this works (it does for me locally and in tests) then I'm in favor of keeping this.

rootulp (dismissed):

rootulp:
[follow up] in the v6 release notes we should instruct node operators to invoke this tool.

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+				return err
+			}
+
+			return updateConfig(homeDir, targetVersion, backup)
+		},
+	}
+
+	cmd.Flags().String(flags.FlagHome, app.NodeHome, "The application home directory")
+	cmd.Flags().Bool("backup", true, "Create backups of config files before updating them")
+	cmd.Flags().String("app-version", fmt.Sprintf("v%d", appconsts.Version), "Target version for config changes")
```
evan-forbes, line 57: had to change this flag as there is overlap somewhere

by default this value would have been v0.38.17 instead of v6

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+
+// updateRegistry maps version strings to their corresponding update functions
+var updateRegistry = map[string]ConfigUpdater{
+	"v6": applyV6Config,
+}
+
+// updateConfigCmd returns the update-config command that updates
+// configuration files based on the target version.
+func updateConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
```
evan-forbes, line 29: just renamed everything from migrate to update as that's more accurate to what is going on. We're not migrating a config, (ie from a old format to a new one) we're just updating the values in it.

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+		Use:     "update-config",
+		Short:   "Update configuration values to be that of a specific app version",
+		Long:    "Update configuration files (config.toml and app.toml) to be compatible with a specific app version.",
+		Example: "celestia-appd update-config --home ~/.celestia-app\ncelestia-appd update-config --version v6 --home ~/.celestia-app",
+		Args:    cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			homeDir, err := cmd.Flags().GetString(flags.FlagHome)
```
cmwaters, line 36: I think there is an app.NodeHome that you can use instead. At least that's what I did with the min gas migration if you want to have a look at it. That way users don't have to provide it if it's in the default location

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// updateConfigCmd returns the update-config command that updates
+// configuration files based on the target version.
+func updateConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "update-config",
+		Short:   "Update configuration values to be that of a specific app version",
+		Long:    "Update configuration files (config.toml and app.toml) to be compatible with a specific app version.",
+		Example: "celestia-appd update-config --home ~/.celestia-app\ncelestia-appd update-config --version v6 --home ~/.celestia-app",
```
cmwaters, line 33: Need to update example to app-version

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// ConfigUpdater defines a function type for applying version-specific updates
+type ConfigUpdater func(*config.Config, *serverconfig.Config) (*config.Config, *serverconfig.Config)
+
+// updateRegistry maps version strings to their corresponding update functions
+var updateRegistry = map[string]ConfigUpdater{
+	"v6": applyV6Config,
```
cmwaters, line 23: For signalling we don't prepend the v. It's a slight nit but I would like to be consistent with whether to prepend the v or not

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+// applyV6Configs applies configuration changes needed for v6
+func applyV6Config(cmtCfg *config.Config, appCfg *serverconfig.Config) (*config.Config, *serverconfig.Config) {
+	fmt.Println("Applying v6 updates to configs...")
+
+	defaultCfg := app.DefaultConsensusConfig()
+	cmtCfg.RPC.MaxBodyBytes = defaultCfg.RPC.MaxBodyBytes
+
+	cmtCfg.Mempool.TTLNumBlocks = defaultCfg.Mempool.TTLNumBlocks
+	cmtCfg.Mempool.TTLDuration = defaultCfg.Mempool.TTLDuration
+	cmtCfg.Mempool.MaxTxBytes = defaultCfg.Mempool.MaxTxBytes
```
cmwaters, line 201: Weren't we going to remove this? (And have it hardcoded in the binary instead)

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+	cmtCfg.Mempool.MaxTxsBytes = defaultCfg.Mempool.MaxTxsBytes
+	cmtCfg.Mempool.Type = defaultCfg.Mempool.Type
+	cmtCfg.Mempool.MaxGossipDelay = defaultCfg.Mempool.MaxGossipDelay
+
+	cmtCfg.P2P.SendRate = defaultCfg.P2P.SendRate
+	cmtCfg.P2P.RecvRate = defaultCfg.P2P.RecvRate
+
+	defaultAppCfg := app.DefaultAppConfig()
+	appCfg.MinGasPrices = defaultAppCfg.MinGasPrices
```
cmwaters, line 210: My plan was to set this value to an empty string "" - to indicate that it wasn't set and to use the hardcoded default.

For now you can leave it as is and I will submit a PR on top of yours to make the right edits

cmd/celestia-appd/cmd/migrate_config.go
```diff
@@ -0,0 +1,214 @@
...
+	serverConfig, err := loadServerConfig(appConfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to load server config: %w", err)
+	}
+
+	fmt.Printf("Loaded configs successfully. Applying %s config updates...\n", targetVersion)
+
+	cometConfig, serverConfig = updater(cometConfig, serverConfig)
+
+	config.WriteConfigFile(cometConfigPath, cometConfig)
```
cmwaters, line 94: Does this keep all the comments in the toml file in tact or does it remove it?

cmwaters (approved):
Small nits but I think this is well done overall.

Need to remember to add notes everywhere for operators

I will submit the min gas price changes in a separate PR (and may build off yours if this isn't merged to main yet)

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: > Does this keep all the comments in the toml file in tact or does it remove it?

keeps the standard comments/template. It will remove any user comments I think however. the good news is by default we're backing things up if users need to compare the old

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: my understanding was that we were deprecating the configs https://github.com/celestiaorg/celestia-core/pull/2293 and modifying them (why not if we're overriding them) but we can also override them. 
 
 which do we want to override? the two TTLs and MaxTx size?

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: to what?

not the v?

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: did you see the defaults?

that's what we use afaiu, maybe you're referring to something else?

https://github.com/celestiaorg/celestia-app/blob/e0247d7f44bfd32f11a18c8638d4935026f7b7b5/cmd/celestia-appd/cmd/migrate_config.go#L55

cmd/celestia-appd/cmd/migrate_config.go
evan-forbes: makes sense

will change in a flup

evan-forbes:
merging but submitting a flup for the v6 vs 6 thing

cmd/celestia-appd/cmd/migrate_config.go
cmwaters: Isn't the flag "app-version" not "version"

Issue #5400, opened 2025-08-01
ICA broken after v4/v5 upgrades


joe-bowman:
### Summary of Bug

After the v5 upgrade, we can now create transfer channels successfully, however ICA channels can not be created.

Hermes gets a panic `ica/host params are not set in store` (from `ibc-go/v8/modules/apps/27-interchain-accounts/host/keeper.Keeper.GetParams()`) on handling a `MsgChanOpenTry`.

Full stack:

```
[...]

PR #5442, opened 2025-08-05
fix!: set ICA params during v6 upgrade handler

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5400

###  `./scripts/single-node-upgrade.sh`

On v5
```
$ celestia-appd query ica host params
Error: rpc error: code = Unknown desc = unknown query path: unknown request
```

On v6
```
$ celestia-appd query ica host params
Error: rpc error: code = Unknown desc = unknown query path: unknown request
$ celestia-appd query ica host params
allow_messages:
- /ibc.applications.transfer.v1.MsgTransfer
- /cosmos.bank.v1beta1.MsgSend
- /cosmos.staking.v1beta1.MsgDelegate
- /cosmos.staking.v1beta1.MsgBeginRedelegate
- /cosmos.staking.v1beta1.MsgUndelegate
- /cosmos.staking.v1beta1.MsgCancelUnbondingDelegation
- /cosmos.distribution.v1beta1.MsgSetWithdrawAddress
- /cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward
- /cosmos.distribution.v1beta1.MsgFundCommunityPool
- /cosmos.gov.v1.MsgVote
- /cosmos.feegrant.v1beta1.MsgGrantAllowance
- /cosmos.feegrant.v1beta1.MsgRevokeAllowance
host_enabled: true
```

### `./scripts/single-node-all-upgrades.sh`

On v6

```
$ celestia-appd query ica host params
Error: rpc error: code = Unknown desc = unknown query path: unknown request
$ celestia-appd query ica host params
allow_messages:
- /ibc.applications.transfer.v1.MsgTransfer
- /cosmos.bank.v1beta1.MsgSend
- /cosmos.staking.v1beta1.MsgDelegate
- /cosmos.staking.v1beta1.MsgBeginRedelegate
- /cosmos.staking.v1beta1.MsgUndelegate
- /cosmos.staking.v1beta1.MsgCancelUnbondingDelegation
- /cosmos.distribution.v1beta1.MsgSetWithdrawAddress
- /cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward
- /cosmos.distribution.v1beta1.MsgFundCommunityPool
- /cosmos.gov.v1.MsgVote
- /cosmos.feegrant.v1beta1.MsgGrantAllowance
- /cosmos.feegrant.v1beta1.MsgRevokeAllowance
host_enabled: true
```

### DIFF
--- a/app/upgrades_test.go
+++ b/app/upgrades_test.go
@@ -1,27 +1,55 @@
 package app_test
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
 	"cosmossdk.io/log"
+	upgradetypes "cosmossdk.io/x/upgrade/types"
 	"github.com/celestiaorg/celestia-app/v6/app"
+	"github.com/celestiaorg/celestia-app/v6/test/util"
 	"github.com/celestiaorg/celestia-app/v6/test/util/testfactory"
 	tmdb "github.com/cosmos/cosmos-db"
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/stretchr/testify/require"
 )
 
-func TestRegisterUpgradeHandlers(t *testing.T) {
-	logger := log.NewNopLogger()
-	db := tmdb.NewMemDB()
-	traceStore := &NoopWriter{}
-	timeoutCommit := time.Second
-	appOptions := NoopAppOptions{}
+func TestUpgrades(t *testing.T) {
+	t.Run("app.New() should register a v6 upgrade handler", func(t *testing.T) {
+		logger := log.NewNopLogger()
+		db := tmdb.NewMemDB()
+		traceStore := &NoopWriter{}
+		timeoutCommit := time.Second
+		appOptions := NoopAppOptions{}
 
-	// app.New() should invoke RegisterUpgradeHandlers.
-	testApp := app.New(logger, db, traceStore, timeoutCommit, appOptions, baseapp.SetChainID(testfactory.ChainID))
+		testApp := app.New(logger, db, traceStore, timeoutCommit, appOptions, baseapp.SetChainID(testfactory.ChainID))
 
-	require.False(t, testApp.UpgradeKeeper.HasHandler("v5"))
-	require.True(t, testApp.UpgradeKeeper.HasHandler("v6"))
+		require.False(t, testApp.UpgradeKeeper.HasHandler("v5"))
+		require.True(t, testApp.UpgradeKeeper.HasHandler("v6"))
+	})
+}
+
+func TestApplyUpgrade(t *testing.T) {
+	t.Run("ICA host params should have an allowlist of messages before and after ApplyUpgrade", func(t *testing.T) {
+		testApp, _, _ := util.NewTestAppWithGenesisSet(app.DefaultConsensusParams())
+		require.True(t, testApp.UpgradeKeeper.HasHandler("v6"))
+		plan := upgradetypes.Plan{
+			Name:   "v6",
+			Time:   time.Now(),
+			Height: 1,
+			Info:   "info",
+		}
+
+		ctx := testApp.NewContext(false)
+		params := testApp.ICAHostKeeper.GetParams(ctx)
+		fmt.Printf("params before upgrade: %+v\n", params)
+
+		err := testApp.UpgradeKeeper.ApplyUpgrade(ctx, plan)
+		require.NoError(t, err)
+
+		ctx = testApp.NewContext(false)
+		params = testApp.ICAHostKeeper.GetParams(ctx)
+		fmt.Printf("params after upgrade: %+v\n", params)
+	})
 }
--- a/app/upgrades_test.go
+++ b/app/upgrades_test.go
@@ -1,7 +1,6 @@
 package app_test
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
@@ -45,7 +44,6 @@ func TestApplyUpgrade(t *testing.T) {
 		params := testApp.ICAHostKeeper.GetParams(ctx)
 		require.True(t, params.HostEnabled)
 		require.Equal(t, params.AllowMessages, app.IcaAllowMessages())
-		fmt.Printf("params before upgrade: %+v\n", params)
 
 		err := testApp.UpgradeKeeper.ApplyUpgrade(ctx, plan)
 		require.NoError(t, err)
@@ -54,6 +52,5 @@ func TestApplyUpgrade(t *testing.T) {
 		params = testApp.ICAHostKeeper.GetParams(ctx)
 		require.True(t, params.HostEnabled)
 		require.Equal(t, params.AllowMessages, app.IcaAllowMessages())
-		fmt.Printf("params after upgrade: %+v\n", params)
 	})
 }
--- a/app/upgrades_test.go
+++ b/app/upgrades_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/celestiaorg/celestia-app/v6/test/util/testfactory"
 	tmdb "github.com/cosmos/cosmos-db"
 	"github.com/cosmos/cosmos-sdk/baseapp"
+	icahosttypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/host/types"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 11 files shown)


rootulp:
Doesn't work as expected because after running `./scripts/single-node-all-upgrades.sh` I see

```
$ celestia-appd query  interchain-accounts host params
allow_messages:
- '*'
host_enabled: true
```

rootulp:
Update: for some reason on my local devnet, the param in the x/params store is incorrect

```
$ celestia-appd query params subspace icahost AllowMessages
param:
  key: AllowMessages
  subspace: icahost
  value: '["*"]'
```

rootulp:
If I just start a local devnet on v3.x then it works correctly

```
gco v3.x
make install
./scripts/single-node.sh

$ celestia-appd query params subspace icahost AllowMessages
key: AllowMessages
subspace: icahost
value: '["/ibc.applications.transfer.v1.MsgTransfer","/cosmos.bank.v1beta1.MsgSend","/cosmos.staking.v1beta1.MsgDelegate","/cosmos.staking.v1beta1.MsgBeginRedelegate","/cosmos.staking.v1beta1.MsgUndelegate","/cosmos.staking.v1beta1.MsgCancelUnbondingDelegation","/cosmos.distribution.v1beta1.MsgSetWithdrawAddress","/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward","/cosmos.distribution.v1beta1.MsgFundCommunityPool","/cosmos.gov.v1.MsgVote","/cosmos.feegrant.v1beta1.MsgGrantAllowance","/cosmos.feegrant.v1beta1.MsgRevokeAllowance"]'
```

rootulp:
The migration handler is overwriting the ICA host params to use the defaults when it should actually set them to the params from x/params module. When the migrator attempts to get the params from the x/params subspace, it fails to find them.

rootulp:
I resolved by not using the migration logic included in ibc-go and instead setting the parameters explicitly to the values defined in CIP-14.

app/upgrades_test.go
```diff
@@ -0,0 +1,56 @@
...
+		require.True(t, params.HostEnabled)
+		require.Equal(t, params.AllowMessages, app.IcaAllowMessages())
+
+		err := testApp.UpgradeKeeper.ApplyUpgrade(ctx, plan)
+		require.NoError(t, err)
+
+		ctx = testApp.NewContext(false)
+		params = testApp.ICAHostKeeper.GetParams(ctx)
+		require.True(t, params.HostEnabled)
+		require.Equal(t, params.AllowMessages, app.IcaAllowMessages())
```
evan-forbes: blocking question

will this test run the upgrade or do we need to change the default consensus params to v5 first?

https://github.com/rootulp/celestia-app/blob/cfd9c4112c5b1b94d1dcbbe8607c37d452eed44c/app/default_overrides.go#L238

evan-forbes: nice test tho

rootulp: This test runs the upgrade because it executes the fix (`setICAHostParams`)

evan-forbes: wait but aren't the params the same before and after the fix? maybe I'm reading the test wrong?

evan-forbes: if it requires running the multiplexer to run v5 then upgrade we can just test in the upgrade e2e test imo

or maybe we could remove the params, apply the upgrade, then check to see if they're set if we want a unit test?

rootulp: The params are the same before and after.

This doesn't repro what's happening on Mainnet b/c we expect the params to be unset on Mainnet (and/or the module to not be registered) before the upgrade.

> if it requires running the multiplexer to run v5 then upgrade we can just test in the upgrade e2e test imo

Agreed. Tracking in https://github.com/celestiaorg/celestia-app/issues/5452

> or maybe we could remove the params, apply the upgrade, then check to see if they're set if we want a unit test?

I can remove or modify this unit test considering it's causing more confusing than helping. It was helpful at development time because the IBC param migrator does not work as expected so we needed to write our own.

rootulp: Refactored the test in [562ee1b](https://github.com/celestiaorg/celestia-app/pull/5442/commits/562ee1b7d8b7cd6a3aa6849306f7ad68a79758bd)

scripts/single-node-all-upgrades.sh
```diff
@@ -147,52 +136,36 @@ upgradeToV3AndV4() {
...
-        echo "Current version: $current_version, waiting for version 4..."
+        echo "Current version: $current_version, waiting for version 2..."
         sleep 1
     done
+
+    # Perform upgrades to versions 3, 4, 5, and 6
+    performUpgrade 3
+    performUpgrade 4
+    performUpgrade 5
+    performUpgrade 6
```
evan-forbes, line 165: we don't have a v6 yet afiu, so does the upgrade to v6 work? https://github.com/rootulp/celestia-app/blob/562ee1b7d8b7cd6a3aa6849306f7ad68a79758bd/internal/embedding/data.go#L10-L16

evan-forbes (approved):

scripts/single-node-all-upgrades.sh
rootulp: main is v6: https://github.com/celestiaorg/celestia-app/blob/d8bd2676fdb06dbef783e26ec9c76e1cdcf606eb/go.mod#L1

evan-forbes: how do we upgrade though if we don't have a version in the multiplexer with v6

evan-forbes: https://github.com/celestiaorg/celestia-app/blob/d8bd2676fdb06dbef783e26ec9c76e1cdcf606eb/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go#L67-L86

perhaps I'm just misunderstanding how the upgrades work?

evan-forbes: ahh the native app covers v6 

which now I don't like that its handled w/o a "version"

rootulp: It is confusing but yes the native app is v6 so it doesn't need an abci.Version in that list. It may be clearer to see on v4.x where the native app (v4) isn't in the abci.Version list: https://github.com/celestiaorg/celestia-app/blob/25910206ff1b7850cc05076094703b7122832014/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go#L39-L50

scripts/single-node-all-upgrades.sh
```diff
@@ -107,21 +105,12 @@ startCelestiaApp() {
...
-        current_version=$(celestia-appd status | jq -r '.node_info.protocol_version.app')
-        if [ "$current_version" = "2" ]; then
-            echo "App version 2 detected, proceeding with v3 upgrade..."
-            break
-        fi
-        echo "Current version: $current_version, waiting for version 2..."
-        sleep 1
-    done
+# Function to perform upgrade to a specific version
+performUpgrade() {
```
damiannolan, line 109: is it expected that the user running this script does some sanity checks after the upgrade succeeds? 

there's no query based assertions built into the script, right?

damiannolan (approved):
This looks good to me (verified that ICA is indeed added back to module manager).

Was the automatic migration skipped in v4, due to the module missing from the manager? And now it needs to be done manually?

scripts/single-node-all-upgrades.sh
rootulp: Yea no queries baked into the script. I usually run this script in one terminal and then in a separate terminal run queries to debug whatever I'm working on.

rootulp:
AFAICT we had to manually invoke the ICA param migrator which I tried to do in the early commits on this PR.

But the upstream ICA param migrator wasn't setting the correct params on the icahost sub store. Instead of using the Celestia-specific params, it was using the upstream defaults. I think it could be related to https://github.com/celestiaorg/celestia-app/pull/5506#discussion_r2267079384 because it wasn't fetching the Celestia-specific params from the x/params substore.

So after Cian's PR merges, we could try to revert part of this change and use the upstream param migrator. Although I haven't tested that idea yet.

Issue #4355, opened 2025-02-24
feat(tx_client): Remove unnecessary full block fetch in tx client initialization


walldiss:
### Summary

Currently, the tx client retrieves the entire block via `GetLatestBlock` during initialization, even though it only needs the `ChainID` and `appVersion`.  

### Problem Definition

As blocks grow larger, this approach becomes inefficient.

### Proposal

The required parameters (`ChainID` and `appVersion`) are already available on the client side and can be passed to the initialization function directly, eliminating the need for a full block request.

evan-forbes:
we could also use the header method as well if the chain-id and app version are not provided

ultraviolet10:
Gm! May I pick up this issue? Helps me dive a little deeper into the codebase 🙂

PR #5449, opened 2025-08-05
perf: optimize TxClient initialization by removing GetLatestBlock call

ultraviolet10:
## Overview

Modified `SetupTxClient()` function signature to accept `chainID` and `appVersion` parameters.


pkg/user/tx_client.go
```diff
@@ -208,18 +207,10 @@ func SetupTxClient(
 	keys keyring.Keyring,
 	conn *grpc.ClientConn,
 	encCfg encoding.Config,
+	chainID string, // Pass chainID directly
+	appVersion string, // Pass appVersion directly
```
evan-forbes, line 211: this is a breaking change

the better approach here is to pass as options then only avoid the rpc call if all data is needed. I'm not sure that we can do it even then

evan-forbes (commented):
is this targetting a specific issue?

if not, please see our contribution guidlines on drive by PRs. we love external contributions, but their source must be from actual problematic usage, not asking claude or codex to find something to fix

pkg/user/tx_client.go
evan-forbes: if this is actually problematic @ultraviolet10 , then please open an issue and we'd suggest to fix via options instead of breaking other user's

ultraviolet10: there is an issue! https://github.com/celestiaorg/celestia-app/issues/4355

I kept getting github's unicorn on this page once I opened the PR, thought I'd wait for a bit before trying to access this page again 😓 Wasn't able to link the issue for that reason

Sincere apologies for not trying more promptly, I was caught up in some other stuff (dinner 😅)

Please lmk if I should open another PR in that case, since you've provided some updated context on this

Issue #4886, opened 2025-05-27
Dependabot PRs break docker-e2e tests


rootulp:
## Context

docker-e2e tests are failing on `main` because go mod tidy is needed. I fixed recently in https://github.com/celestiaorg/celestia-app/pull/4875 and then a dependabot PR https://github.com/celestiaorg/celestia-app/pull/4877 broke it again.

## Problem

Dependabot PRs only run `go mod tidy` in the root-level go.mod but the docker-e2e package go mod also needs a `go mod tidy` after most dependabot PRs

## Proposal

1. Ideally fix dependabot so that PRs that make go.mod modifications run go mod tidy in the root and in test/docker-e2e`
2. Consider making docker-e2e tests *required* checks so that PRs can't merge if they break them

rootulp:
Closing as a duplicate of https://github.com/celestiaorg/celestia-app/issues/4929

PR #5455, opened 2025-08-05
fix: try to fix dependabot PRs

evan-forbes:
idk if this will work, but a bunch of dependabot PRs are failing due to not calling go mod tidy in our sub directories that have their own go mods, which results in failures for those tests

see #4886 #5427

### DIFF
--- a/.github/workflows/dependabot-go-tidy.yml
+++ b/.github/workflows/dependabot-go-tidy.yml
@@ -1,0 +1,48 @@
+name: Dependabot Go Mod Tidy
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  go-mod-tidy:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-go-tidy.yml
```diff
@@ -0,0 +1,48 @@
+name: Dependabot Go Mod Tidy
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  go-mod-tidy:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
```
evan-forbes, line 10: this is 100% claude generated, and while I reviewed I'm skeptical this will work as intended since iirc we can't make commits

PR #5460, opened 2025-08-06
fix: stop panicking in PrepareProposalHandler

MozirDmitriy:
Closes https://github.com/celestiaorg/celestia-app/issues/4992
Reopen #5047 
fixed golanci lint, also fixed comment to match the current logic

### DIFF
--- a/app/prepare_proposal.go
+++ b/app/prepare_proposal.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


MozirDmitriy:
"[strongly encouraged] write unit tests for each of the cases that can now return errors and assert that an error is indeed returned instead of a panic."
If I should, i will work on it

rootulp (approved):
LGTM

evan-forbes (approved):

PR #5468, opened 2025-08-07
test: fix missing error check in block sync test

Snezhkko:
Add error check for BuildInternalPeerAddressList call in TestBlockSync

The function was returning an error that wasn't being checked, which could
cause the test to continue with invalid peer configuration. This fix ensures
proper error handling and aligns with the pattern used in other e2e tests.


ninabarbakadze (approved):

evan-forbes (approved):
would have expected the linter to pick that up

PR #5469, opened 2025-08-07
chore: remove broken url in specs/src/proto/types.proto

queryfast:
## Overview

fix broken url in specs/src/proto/types.proto


queryfast:
> I think you need to run `make proto-gen` from the root of the project to update all the proto files

Thanks for your guidance.  Executed.

queryfast:
image

specs/src/proto/types.proto
```diff
@@ -40,14 +40,14 @@ message AvailableDataHeader {
 
 // Protobuf definitions for the contents of state elements
 
-// https://github.com/celestiaorg/celestia-specs/blob/master/specs/networking.md#account
+// https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/networking.md#account
```
evan-forbes: we can just delete these links as that repo is no longer used and archived.

idek if we even use these protos somewhere

tools/talis/config.go
```diff
@@ -125,7 +125,7 @@ type Config struct {
 	LinodeToken string `json:"linode_token,omitempty"`
 	// S3Config is used to configure the S3 bucket that will be used to store
 	// traces, logs, and other data.
-	S3Config S3Config `json:"s3_config,omitempty"`
+	S3Config S3Config `json:"s3_config"`
```
evan-forbes: seems very unrelated, and I don't think this makes sense

why was this changed in the first place?!

specs/src/proto/types.proto
queryfast: Deleted. Please review again.

tools/talis/config.go
queryfast: I also have some doubts; it wasn't caused by make proto-gen. I have rolled it back.

queryfast:
image

I checked the error messages, and they seem unrelated to my changes. Do I need to address them in this PR?

rootulp:
No. If mardown link check is failing for links unrelated to this PR you can ignore.

rootulp (approved):

rach-id (approved):

PR #5472, opened 2025-08-07
refactor(multiplexer): improve error message

rnkrtt:
## Overview

Replace m.nativeApp with version.AppVersion in error message on line 500 of multiplexer.go. The %d format specifier requires a numeric value, but m.nativeApp is an application object

### DIFF
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -497,7 +497,7 @@ func (m *Multiplexer) startEmbeddedApp(version Version) error {
 		}
 
 		if version.Appd.IsStopped() {
-			return fmt.Errorf("app for version %d failed to start", version.AppVersion)
+			return fmt.Errorf("app for version %d stopped", version.AppVersion)
 		}
 
 		m.activeVersion = version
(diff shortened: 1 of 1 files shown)


evan-forbes (dismissed):

evan-forbes:
@rnkrtt please no more driveby PRs, see contribution guidelines

multiplexer/abci/multiplexer.go
```diff
@@ -497,7 +497,7 @@ func (m *Multiplexer) startEmbeddedApp(version Version) error {
 		}
 
 		if version.Appd.IsStopped() {
-			return fmt.Errorf("app for version %d failed to start", m.nativeApp)
+			return fmt.Errorf("app for version %d failed to start", version.AppVersion)
```
rootulp: this seems fine but the original error message is misleading. IMO this would've been clearer

```suggestion
			return fmt.Errorf("app for version %d stopped", version.AppVersion)
```

because the command above starts it and this just checks if it stopped

rootulp (approved):

rootulp (approved):

PR #5456, opened 2025-08-06
chore: add PFB tx to upgrade test

mojtaba-esk:
Closes #5450

PR #5473, opened 2025-08-07
chore: removed redundant code from PFB tx upgrade test

mojtaba-esk:
This is a followup PR for #5456 to remove redundant code from the test.

context: https://github.com/celestiaorg/celestia-app/pull/5456#discussion_r2257871844

### DIFF
--- a/test/docker-e2e/e2e_simple_test.go
+++ b/test/docker-e2e/e2e_simple_test.go
@@ -166,12 +166,5 @@ func testPFBSubmission(t *testing.T, chain *tastoradockertypes.Chain, cfg *docke
 	require.NoError(t, err, "failed to submit PFB transaction")
 	require.Equal(t, uint32(0), txResp.Code, "PFB transaction failed with code %d", txResp.Code)
 
-	t.Logf("PFB transaction submitted! TxHash: %s, Height: %d", txResp.TxHash, txResp.Height)
-
-	t.Logf("Confirming transaction inclusion on-chain...")
-	confirmedTx, err := txClient.ConfirmTx(ctx, txResp.TxHash)
-	require.NoError(t, err, "failed to confirm transaction inclusion")
-	require.Equal(t, uint32(0), confirmedTx.Code, "transaction execution failed with code %d", confirmedTx.Code)
-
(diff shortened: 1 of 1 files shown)


test/docker-e2e/e2e_simple_test.go
```diff
@@ -166,12 +166,5 @@ func testPFBSubmission(t *testing.T, chain *tastoradockertypes.Chain, cfg *docke
...
 
-	t.Logf("PFB transaction submitted! TxHash: %s, Height: %d", txResp.TxHash, txResp.Height)
-
-	t.Logf("Confirming transaction inclusion on-chain...")
-	confirmedTx, err := txClient.ConfirmTx(ctx, txResp.TxHash)
-	require.NoError(t, err, "failed to confirm transaction inclusion")
-	require.Equal(t, uint32(0), confirmedTx.Code, "transaction execution failed with code %d", confirmedTx.Code)
-
-	t.Logf("Transaction confirmed on-chain! TxHash: %s, Height: %d", confirmedTx.TxHash, confirmedTx.Height)
+	t.Logf("PFB transaction included on-chain. TxHash: %s, Height: %d", txResp.TxHash, txResp.Height)
```
evan-forbes, line 169: this isn't redundant @mojtaba-esk we need to check that the tx got committed (confirmed) as opposed to only included in the mempool (first code == 0 check)

we should close imo

evan-forbes (changes requested):
why is this redundant

test/docker-e2e/e2e_simple_test.go
evan-forbes: maybe I'm missing something? otherwise I don't think we should merge

mojtaba-esk: `SubmitPayForBlob` already calls the tx confirm

```go
	resp, err := client.BroadcastPayForBlob(ctx, blobs, opts...)
	if err != nil {
		return nil, err
	}

	return client.ConfirmTx(ctx, resp.TxHash)
}
```

and then in the test code we evaluate the resp code:

```go
require.Equal(t, uint32(0), txResp.Code, "PFB transaction failed with code %d", txResp.Code)
```

so that should be sufficient.

evan-forbes: ahh I see! ok makes sense thanks for explaining

I should have clicked on the context in op

Issue #5272, opened 2025-07-17
v4.0.10-mocha reports version as v4.0.10-arabica


rootulp:
## Problem

```
$ gco v4.0.10-mocha
HEAD is now at 117b77cc5 fix: update go mod to v2.3.1 of go-square (backport #5266) (#5269)
$ make install
Checking if BBR is enabled...
sysctl: unknown oid 'net.ipv4.tcp_congestion_control'
WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README.
--> Downloading celestia-app v3.10.4 binary
Skipping download because expected version already downloaded: celestia-app_darwin_v3_arm64.tar.gz
--> Installing celestia-appd with multiplexer support
$ celestia-appd version
4.0.10-arabica
```

## Proposal

The version should've reported v4.0.10-mocha

cmwaters:
It's because the makefile just sets the version as:

> git describe --tags --always --match "v*"

Which just takes the first tag that matches the "v". Since we use the same commit for multiple tags, this command has no idea if it's arabica, mocha or mainnet. It might be better to take the latest tag (i.e. better to say mainnet when you're checking out arabica, then the other way around)

rootulp:
`git describe --tags` does return the latest tag that points to a commit. From `git describe --help`:

```
The command finds the most recent tag that is reachable from a commit.
```

and it looks like `git describe --tags` has remained in the `Makefile` even after these relevant PRs:

1. https://github.com/celestiaorg/celestia-app/pull/4094
2. https://github.com/celestiaorg/celestia-app/pull/4720

so I'm not sure when the regression was introduced. One thing worth noting is that I created v4.0.10-arabica and v4.0.10-mocha in quick succession but historically, I usually wait some time between cutting releases.

anhductn2001:
After some testing, I found that the idea of automatically detecting which tag was checked out (when multiple tags point to the same commit) is unfortunately not reliable. Git doesn’t preserve the “tag context” in these cases — it just checks out the commit, and git will return the first matching tag (usually not the one you expect).

Because of this limitation, I’d like to propose a simpler and more reliable approach:
We add an optional argument to the build command, allowing the user to specify the intended tag or version manually if they want precision.

If the argument is provided (e.g. make build VERSION=mocha), we inject it.
If not, we fall back to a default

That way, the build process stays flexible and avoids relying on Git’s ambiguous behavior in this case. Let me know what you think!

[...]

PR #5475, opened 2025-08-07
feat: add VERSION argument support for make build and install commands

anhductn2001:
Close: #5272 

This small PR adds support for specifying custom VERSION via command line argument when building and installing celestia-appd, allowing users to build binaries with specific network versions (e.g., `4.0.10-mocha`, `4.0.10-arabica`).

Through research and testing, it was determined that automatically detecting the current tag is not feasible (limitation of git) for a commit containing multiple tags.

## Results
```
celestia-app git:(v4.0.10) ✗ VERSION=4.0.10-arabica make install
Checking if BBR is enabled...
sysctl: unknown oid 'net.ipv4.tcp_congestion_control'
WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage se
ction in the README.
--> Downloading celestia-app v3.10.4 binary
--> Installing celestia-appd with multiplexer support

celestia-app git:(v4.0.10) ✗ celestia-appd version
4.0.10-arabica

celestia-app git:(v4.0.10) ✗ VERSION=4.0.10-mocha make install
Checking if BBR is enabled...
sysctl: unknown oid 'net.ipv4.tcp_congestion_control'
WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README.
--> Downloading celestia-app v3.10.4 binary
--> Installing celestia-appd with multiplexer support

celestia-app git:(v4.0.10) ✗ celestia-appd version
4.0.10-mocha
```


anhductn2001:
So do you want me to add the precedence rule in case VERSION is not provided?

rootulp:
I'm not sure. As you pointed out, the precedence rule doesn't work in all cases so it would be nice if there was a solution that always produced the correct `--version` output. I'm open to modifying how we create releases (currently via Github releases which appear to create lightweight git tags) if creating a non-lightweight tag resolves.

anhductn2001:
Ok, when multiple tags point to the same commit, it’s impossible to reliably choose which tag to use when building locally from that commit.

One possible solution is to store the version/tag inside the repo at release time.
Since we can't push directly to main, we could:

1. Create a release branch (e.g. release/v4.0.10-mocha).
2. Write the version to a file (e.g. VERSION) and commit it.
3. Tag the branch with VERSION commit.

With this approach, anyone checking out the tag locally will see the correct version from the file, avoiding mismatches during local `make build` or `make install`.

```
git checkout -b release/v4.0.10-mocha <commit-hash>
echo "v4.0.10-mocha" > VERSION
git add VERSION
git commit -m "Set version to v4.0.10-mocha"
git tag v4.0.10-mocha
git push origin release/v4.0.10-mocha
git push origin v4.0.10-mocha
```

rootulp:
That's a reasonable approach. Although if we push a commit to the release branch in between releases then the `version` command will work because we'll only have one tag per commit. So the commit that we push doesn't even need to populate the `VERSION` file, it can just be a no-op commit or a one line change to the README.

anhductn2001:
yea, I just gave an example of adding 1 line to the VERSION file, we can totally change something else or not change anything (but it will be a bit meaningless)

anhductn2001:
Have we made a decision yet? Or are you still looking for something simpler? @rootulp

rootulp:
I'd like to avoid persisting the `VERSION` in this repo given an additional commit will create a unique commit for the tag which also resolves. Open to other ideas on how to get more reliable version info from the binary when multiple commits exist on the same lightweight git tag.

rootulp:
No longer seems necessary. We merged a different PR that returns the version without a network suffix if multiple tags point to the same commit.

Issue #5476, opened 2025-08-07
Multiplexer refactor wishlist


evan-forbes:
keeping track of potential features/improvements we could refactor the multiplexer around:

- move the concept of embedded vs native to appd
    - We could for example specify in the version if something is native vs embedded in the version itself and avoid all of the logic that is in the multiplexer
- remove the third copy of the app version from the multiplexer
    - currently the multiplexer has a third copy of the app version that it uses to route getApp calls. This is not a requirement, instead we could have a single application that is swapped when the multiplexer see's an upgrade
    - create the correct version of the app upon ABCI calls. 
        - we half do this for statesync already, but upon InitChain() or Info(), we can perform the same steps the application does to determine the app version, then create that version of the application, then proceed with the abci call. 
- get rid of the build flags, instead pass supported versions that are being used via flag
    - imo using build flags to route around using the multiplexer are annoying and we can simply get rid of the standalone
- untangle cli cmd initialization
    - the multiplexer is initializing a comet node, which is added to a hook to the start command which is doing a lot of logic and wrapping the application, but its not the application its actually the multiplexer, which is passed again to comet (?) which passes it a DB I think? unless its an embedded app in which case we don't pass a db, idk tbh its even more speghetti code than the normal init process which we already didn't need. We should start from scratch and just initialize things here. we don't need out of process comet support for now as the multiplexer is handling that so just stop supporting that
- we could also refactor the entire multiplxer to be inprocess
    - this requires changing cosmos-sdk v0.50 -> v2 of the cosmos-sdk so go modules can actually route between different versions of the sdk

PR #5482, opened 2025-08-07
refactor(multiplexer): upgrade directly after the commit

### LINKED ISSUE #5480: Upgrade the application as soon as possible
currently, after the application upgrades and the multiplexer changes its internal appversion, the upgrade to the application does not occur. The upgrade instead occurs at the next .getApp() call, which maybe (doubt it) be the source of bugs such as #5406.

instead, as soon as the multiplexer recognizes that there is a version change, it can swap the app while holding a mutex to prevent accidental calls to the application at random points during the upgrade process.

evan-forbes:
closes #5480

I don't think this fixes anything, but maybe? also seems nice to upgrade during the timeout commit instead of on checkTx, prepareProposal, or processProposal.

### DIFF
--- a/multiplexer/abci/abci.go
+++ b/multiplexer/abci/abci.go
@@ -52,15 +52,11 @@ func (m *Multiplexer) Commit(context.Context, *abci.RequestCommit) (*abci.Respon
 
 	// after a successful commit, start using the app version specified in FinalizeBlock. If
 	// there is an upgrade, perform that now.
-	m.mu.Lock()
-	defer m.mu.Unlock()
 	oldAppVersion := m.appVersion
 	m.appVersion = m.nextAppVersion
 	if oldAppVersion != m.nextAppVersion {
-		// this effectively performs the upgrade immediately instead of waiting until the next call to getApp. the
-		// unsafe version is used in order to call the mutex earlier and guatantee that we are not calling other abci
-		// methods or calls to the application while upgrading.
-		_, err = m.unsafeGetApp()
+		// this effectively performs the upgrade immediately instead of waiting until the next call to getApp.
+		_, err = m.getApp()
 		if err != nil {
 			return nil, fmt.Errorf("multiplexer failed upgrade: %w", err)
 		}
--- a/multiplexer/abci/abci.go
+++ b/multiplexer/abci/abci.go
@@ -50,8 +50,21 @@ func (m *Multiplexer) Commit(context.Context, *abci.RequestCommit) (*abci.Respon
 		return nil, fmt.Errorf("failed to commit: %w", err)
 	}
 
-	// after a successful commit, we start using the app version specified in FinalizeBlock.
+	// after a successful commit, start using the app version specified in FinalizeBlock. If
+	// there is an upgrade, perform that now.
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	oldAppVersion := m.appVersion
 	m.appVersion = m.nextAppVersion
+	if oldAppVersion != m.nextAppVersion {
+		// this effectively performs the upgrade immediately instead of waiting until the next call to getApp. the
+		// unsafe version is used in order to call the mutex earlier and guatantee that we are not calling other abci
+		// methods or calls to the application while upgrading.
+		_, err = m.unsafeGetApp()
+		if err != nil {
+			return nil, fmt.Errorf("multiplexer failed upgrade: %w", err)
+		}
+	}
 
 	return resp, nil
 }
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -421,6 +421,12 @@ func openTraceWriter(traceWriterFile string) (w io.WriteCloser, err error) {
 func (m *Multiplexer) getApp() (servertypes.ABCI, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	return m.unsafeGetApp()
+}
+
+// unsafeGetApp gets or creates an application for the current app version. It is unsafe due to not
+// holding the mutex.
+func (m *Multiplexer) unsafeGetApp() (servertypes.ABCI, error) {
 	m.logger.Debug("getting app", "app_version", m.appVersion, "next_app_version", m.nextAppVersion)
 
 	// get the appropriate version for the latest app version.
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -421,12 +421,6 @@ func openTraceWriter(traceWriterFile string) (w io.WriteCloser, err error) {
 func (m *Multiplexer) getApp() (servertypes.ABCI, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	return m.unsafeGetApp()
-}
-
-// unsafeGetApp gets or creates an application for the current app version. It is unsafe due to not
-// holding the mutex.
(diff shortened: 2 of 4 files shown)


multiplexer/abci/abci.go
```diff
@@ -50,8 +50,21 @@ func (m *Multiplexer) Commit(context.Context, *abci.RequestCommit) (*abci.Respon
...
+	// there is an upgrade, perform that now.
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	oldAppVersion := m.appVersion
 	m.appVersion = m.nextAppVersion
+	if oldAppVersion != m.nextAppVersion {
+		// this effectively performs the upgrade immediately instead of waiting until the next call to getApp. the
+		// unsafe version is used in order to call the mutex earlier and guatantee that we are not calling other abci
+		// methods or calls to the application while upgrading.
+		_, err = m.unsafeGetApp()
```
evan-forbes: we don't need to hold the mtx here, since comet should be holding the mutex for all ABCI calls. However in the future we could remove that and still have this property if we hold it here.

cmwaters (commented):
When does the migrations actually happen. Is it immediately when the node starts up or only in finalize block of the next block?

evan-forbes:
> When does the migrations actually happen.

good point! the upgrade is scheduled for the next block, and the migration occurs during the "PreBlocker" on finalize there.

we're still good to swap binaries here I believe (otherwise we're switching on Prepare/ProcessProposal at the latest. However it does seem like we definitely want to be calling the migrations directly after switching binaries!! which is currently not happening. I could see that being a pretty big foot gun, where we're unable to produce a block or vote on a block because that requires state access that requires some migration to work properly.

multiplexer/abci/multiplexer.go
```diff
@@ -421,6 +421,12 @@ func openTraceWriter(traceWriterFile string) (w io.WriteCloser, err error) {
 func (m *Multiplexer) getApp() (servertypes.ABCI, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	return m.unsafeGetApp()
+}
+
+// unsafeGetApp gets or creates an application for the current app version. It is unsafe due to not
+// holding the mutex.
+func (m *Multiplexer) unsafeGetApp() (servertypes.ABCI, error) {
```
rootulp: Is it possible to refactor the mutex handling so that there isn't an `unsafeGetApp()`?

I don't understand when `unsafeGetApp()` should be used vs the presumably safe `getApp()`.

multiplexer/abci/abci.go
rootulp: I don't follow the multiplexer mutex logic. Is it possible to implement tests alongside or prior to this refactor to confirm that their are no accidental data races or deadlocks in the multiplexer mutex logic?

multiplexer/abci/multiplexer.go
evan-forbes: typically this is done via the exported method having the mutex, and the unexported not using the mutex. currently the unexported is being used everywhere but we can change that

the unsafe is purely needed when one wants to retrieve the application while still holding the mutex. That requires not taking the mutex in the first place.

I'll add some better docs and use the more standard
```
GetApp (holds mutex)
getApp (doesn't hold mutex)
```

evan-forbes:
> I don't follow the multiplexer mutex logic. Is it possible to implement tests alongside or prior to this refactor to confirm that their are no accidental data races or deadlocks in the multiplexer mutex logic?

the only diff here is that we're holding the mutex while we change the appversion and then calling getApp. We could change the appversion w/o taking the mutex, then getApp like normal if that makes more sense. This is not needed atm, since comet takes a mutex on all abci calls (thus also calls to the multiplexer).

like all out of process upgrades, due to the multiplexer's out of process management, we don't really test the multiplexer outside of docker which makes it a lot harder ofc since we are not using the race detector there. If we could mock it we probably could, but that requires a refactor in how the multiplexer managements embedded vs native binaries. I've suggested this change in #5476 

for now to get this change through, I'll simply not make changes w/ unsafe or mutexes. We'll simply call getApp after there is an upgrade.

multiplexer/abci/multiplexer.go
evan-forbes: scratch that, this mutex fix is not required, it was just trying to avoid footguns in the future.

I'll remove for now

multiplexer/abci/abci.go
```diff
@@ -50,8 +50,17 @@ func (m *Multiplexer) Commit(context.Context, *abci.RequestCommit) (*abci.Respon
...
+	// there is an upgrade, perform that now.
+	oldAppVersion := m.appVersion
 	m.appVersion = m.nextAppVersion
+	if oldAppVersion != m.nextAppVersion {
+		// this effectively performs the upgrade immediately instead of waiting until the next call to getApp.
+		_, err = m.getApp()
+		if err != nil {
+			return nil, fmt.Errorf("multiplexer failed upgrade: %w", err)
+		}
+	}
```
evan-forbes, line 63: I reverted to keeping things closer to what we do now @rootulp @cmwaters 

that means not holding the mutex when change app versions, only when switching the binary on the getApp call

the other way was working fine, just untested and there's not an easy way to test atm

evan-forbes: we have #5476 to keep track of some ideas for refactoring the multiplexer, including ways to make it easier to test even w/ out of process

multiplexer/abci/abci.go
```diff
@@ -50,8 +50,17 @@ func (m *Multiplexer) Commit(context.Context, *abci.RequestCommit) (*abci.Respon
 		return nil, fmt.Errorf("failed to commit: %w", err)
 	}
 
-	// after a successful commit, we start using the app version specified in FinalizeBlock.
+	// after a successful commit, start using the app version specified in FinalizeBlock. If
+	// there is an upgrade, perform that now.
```
rootulp, line 54: [optional]

```suggestion
	// After a successful commit, start using the app version specified in FinalizeBlock. If
	// there is an app version upgrade, perform that now.
```

multiplexer/abci/abci.go
rootulp, line 63: [optional] refactor

```suggestion
	if m.appVersion != m.nextAppVersion {
		m.appVersion = m.nextAppVersion
		// Invoke getApp for its side effect which handles stopping the binary for the previous app version and starting the binary for the next app version.
		_, err = m.getApp()
		if err != nil {
			return nil, fmt.Errorf("failed to get app: %w", err)
		}
	}
```

rootulp: [no change needed] The logic inside `getApp` is crazy confusing. That function is poorly named because nothing about `getApp` conveys that it may trigger a change to the embedded binary on invocation. One idea for refactor is to extract the logic that handles stopping and starting binaries and invoke that here.

And then `getApp` can always just return the current app and not do anything related to binary management.

evan-forbes: > One idea for refactor is to extract the logic that handles stopping and starting binaries and invoke that here.

yes! :raised_hands:  definitely. keeping track of this in #5476

evan-forbes: also feel free to add things there. It's truly a list of potential things, not things that we are doing so any idea can be added

evan-forbes: merged for now as I have other PRs based on this branch (:see_no_evil:) but agree this is nicer

PR #5485, opened 2025-08-08
ci: run go mod tidy on dependabot PRs

rach-id:
Runs go mod tidy on dependabot PRs to update all the modules.

### DIFF
--- a/.github/workflows/dependabot-tidy.yml
+++ b/.github/workflows/dependabot-tidy.yml
@@ -1,4 +1,4 @@
-name: Tidy All Go Modules
+name: dependabot-tidy
 
 on:
   pull_request:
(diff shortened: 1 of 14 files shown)


.github/workflows/dependabot-tidy.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Commit and push tidy changes (only Dependabot PRs)
+        if: github.event_name == 'pull_request' && github.actor == 'dependabot[bot]'
+        run: |
+          git config user.name "github-actions"
+          git config user.email "github-actions@github.com"
+
+
+          git config --global credential.helper store
+          echo "https://oauth2:${{ secrets.DEPENDABOT_TIDY_TOKEN }}@github.com" > ~/.git-credentials
```
rach-id, line 40: I tested this and it seems to work.

all that's left is to generate a new token with only enough permissions to push code, set it as a secret called `DEPENDABOT_TIDY_TOKEN` then we merge.

rootulp (approved):
Thanks

PR #5506, opened 2025-08-11
test: Add test for ICA

### LINKED ISSUE #5452: Tastora test for ICA
## Context

We broke ICA in v4.

## Problem

There are no automated tests that confirm ICA works.

## Proposal

Tastora gained IBC support with https://github.com/celestiaorg/celestia-app/pull/5401

Consider adding a new Tastora test that covers ICA.

chatton:
## Overview

WIP: need to clean up a lot of this code, however the test is functional and ICA seems to be working correctly again with the changes to app.go

closes #5452


app/app.go
```diff
@@ -758,9 +758,9 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(distrtypes.ModuleName)
 	paramsKeeper.Subspace(slashingtypes.ModuleName)
 	paramsKeeper.Subspace(govtypes.ModuleName)
-	paramsKeeper.Subspace(ibctransfertypes.ModuleName)
+	paramsKeeper.Subspace(ibctransfertypes.ModuleName).WithKeyTable(ibctransfertypes.ParamKeyTable())
```
chatton, line 760: this is what was causing the icahost module to not correctly bind capabilities, I tested with a locally built image and everything works. Without it, I was getting these errors from hermes

```
message: "failed to execute message; message index: 1: could not retrieve module from port-id: ports/icahost: capability not found [cosmos/ibc-go/modules/capability@v1.0.1/keeper/keeper.go:452] with gas used: '117491'", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "66"} }
```

app/app.go
```diff
@@ -758,9 +758,9 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
...
 	paramsKeeper.Subspace(ibcexported.ModuleName)
-	paramsKeeper.Subspace(icahosttypes.SubModuleName)
+	paramsKeeper.Subspace(icahosttypes.SubModuleName).WithKeyTable(icahosttypes.ParamKeyTable())
```
rootulp, line 762: [no change needed] I wonder how this interacts with https://github.com/celestiaorg/celestia-app/pull/5442

For context: I noticed that the ICA host params on v6 defaulted to the upstream ICA host params (a.k.a wildcard for AllowMessages) and I had to explicitly set them in the v5 -> v6 upgrade handler. But I wonder if that migration is not needed after this `.WithKeyTable` change.

Do you know why the icahosttypes and ibctransfertypes param subspaces need the `.WithKeyTable` lines but the other subspaces don't need them?

rootulp (commented):
Thanks so much for adding this test!

test/docker-e2e/e2e_ibc_test.go
```diff
@@ -21,13 +21,113 @@ import (
...
+	hermes          *relayer.Hermes
+	connection      ibc.Connection
+	transferChannel ibc.Channel
+
+	// Transaction clients
+	celestiaTxClient *user.TxClient
+}
+
+// setupIBCInfrastructure sets up IBC test infrastructure with specified app version
+func (s *IBCTestSuite) setupIBCInfrastructure(appVersion uint64) {
```
chatton, line 56: this is the logic from the regular ibc test pulled out so it can be re-used in test setup

app/app.go
chatton: I think your change will still be required, the WithKeyTable change makes it so that the icahost module doesn't fail silently (fails to claim capability)

I tried running

```
$ celestia-appd query params subspace icahost AllowMessages
param:
  key: AllowMessages
  subspace: icahost
``` 

It looks like here we have the key present, but it has no value, so I would imagine it defaults to the `*`

However we should double check that both changes work together 👀 maybe we can extend the test to query the allow messages. 

I can do this in a follow up maybe as the PR is quite large as is.

Issue #5392, opened 2025-07-31
Include documentation around release grpc default changes for docker containers


evan-forbes:
between v3 and v4, the docker container changed the default listen address, but this was not documented well enough for folks to find

also from a sync
> the [grpc] flag enabled is ignored in the latest versions because grpc is enabled mandatorily

if there are maditory setting, those should also be documented and ideally also logged to routinely inform operators that some setting was overwritten. grpc being manditory also sounds like a bug, so that should be investigated.

rootulp:
The multiplexer always enables GRPC for the embedded apps (see [here](https://github.com/celestiaorg/celestia-app/blob/55581ae3afe4b21c838203cfaa1c59b0109eaa79/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go#L23)) so I agree this should've been documented in the [release notes](https://github.com/celestiaorg/celestia-app/blob/main/docs/release-notes/release-notes.md#v400).

@julienrbrt 
- Why is GRPC always enabled for embedded binaries? I assume so that the multiplexer can route requests to the cosmos-sdk GRPC server running in the embedded binaries. But is that necessary if the user doesn't want the GRPC server enabled? 
- If a user configures their `config.toml` to disable GRPC or passes the `--grpc.enable=false` flag to the multiplexer then the v3 embedded binary will still start with the GRPC server enabled which seems like a bug.

rootulp:
> between v3 and v4, the docker container changed the default listen address, but this was not documented well enough for folks to find

I dug through the Conduit thread to figure out what this means. This is specifically about the default app.toml GRPC address going from `0.0.0.0` to `localhost`. 

On v3.x the default `app.toml`.

```
[grpc]

# Enable defines if the gRPC server should be enabled.
enable = false

# Address defines the gRPC server address to bind to.
address = "0.0.0.0:9090"
```

On v4.x the default `app.toml`

```
[grpc]

# Enable defines if the gRPC server should be enabled.
enable = false

# Address defines the gRPC server address to bind to.
address = "localhost:9090"
```

Ref:
- https://github.com/celestiaorg/cosmos-sdk/blob/5d65bd52fd6df191cbf707c0d0175355c14e2422/server/config/config.go#L23-L24
- https://github.com/celestiaorg/cosmos-sdk/blob/1dafc072cea3b44fe6823b2bfd0239594fc75ef7/server/config/config.go#L22-L23

[...]

PR #5507, opened 2025-08-11
fix(multiplexer): stop forcing grpc and api for embedded binaries

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5392

cc: @julienrbrt


rootulp:
The upgrade e2e test flaked so retrying but considering the rest passed, seems safe.

cmwaters:
Just double checking here: but if grpc is set to enable in the config, it will be enabled when you run the multiplexer (even if it's using an earlier version i.e. v5)

rootulp:
> Just double checking here: but if grpc is set to enable in the config, it will be enabled when you run the multiplexer (even if it's using an earlier version i.e. v5)

If a user is using the v5.0.1 multiplexer, regardless of what is in the config, grpc will be enabled on the embedded binary because of the flag removed in this PR.

After this PR, GRPC will be enabled on the embedded binary if a user has enabled the GRPC in config or passed the `--grpc.enable` flag to the multiplexer. If the GRPC in config is false and there is no `--grpc.enable` flag passed to the multiplexer, GRPC won't be enabled on the embedded binary.

evan-forbes (approved):

PR #5518, opened 2025-08-12
feat: set the default min gas price to an empty string

cmwaters:
Depends on https://github.com/celestiaorg/cosmos-sdk/pull/662

This also will mean that the update-config command will set the users min gas price to an empty string only if it was using the older min gas price

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"encoding/json"
-	"fmt"
 	"time"
 
 	"cosmossdk.io/math"
@@ -301,7 +300,10 @@ func DefaultAppConfig() *serverconfig.Config {
 	// snapshots to nodes that state sync
 	cfg.StateSync.SnapshotInterval = 1500
 	cfg.StateSync.SnapshotKeepRecent = 2
-	cfg.MinGasPrices = fmt.Sprintf("%v%s", appconsts.DefaultMinGasPrice, params.BondDenom)
+	// this is set to an empty string. As an empty string, the binary will use
+	// the hardcoded default gas price. To override this, the user must set the
+	// minimum gas prices in the app.toml file.
+	cfg.MinGasPrices = ""
 	cfg.GRPC.MaxRecvMsgSize = appconsts.DefaultUpperBoundMaxBytes * 2
 	cfg.GRPC.MaxSendMsgSize = appconsts.DefaultUpperBoundMaxBytes * 2
 	return cfg
--- a/cmd/celestia-appd/cmd/migrate_config.go
+++ b/cmd/celestia-appd/cmd/migrate_config.go
@@ -30,7 +30,7 @@ func updateConfigCmd() *cobra.Command {
 		Use:     "update-config",
 		Short:   "Update configuration values to be that of a specific app version",
 		Long:    "Update configuration files (config.toml and app.toml) to be compatible with a specific app version.",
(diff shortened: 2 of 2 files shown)


cmwaters:
I've manually tested the `update-config` command to check that it works as expected

cmd/celestia-appd/cmd/update_config.go
```diff
@@ -30,7 +30,7 @@ func updateConfigCmd() *cobra.Command {
 		Use:     "update-config",
 		Short:   "Update configuration values to be that of a specific app version",
 		Long:    "Update configuration files (config.toml and app.toml) to be compatible with a specific app version.",
-		Example: "celestia-appd update-config --home ~/.celestia-app\ncelestia-appd update-config --version 6 --home ~/.celestia-app",
+		Example: "celestia-appd update-config --home ~/.celestia-app\ncelestia-appd update-config --app-version v6 --home ~/.celestia-app --backup false",
```
rootulp: [question] I would've expected `--app-version 6` and not `--app-version v6`, any chance you tried this locally?

cmd/celestia-appd/cmd/update_config_test.go
```diff
@@ -140,3 +142,29 @@ func verifyUpdatedConfigs(t *testing.T, configDir, version string) {
 		assert.Equal(t, expectedServerConfig.MinGasPrices, serverConfig.MinGasPrices)
 	}
 }
+
+func TestUpdateMinGasPrices(t *testing.T) {
+	t.Run("legacy default min gas price is unset to empty string", func(t *testing.T) {
```
rootulp, line 147: [uber nit] I could see "unset" being fine too

```suggestion
	t.Run("legacy default min gas price is set to empty string", func(t *testing.T) {
```

cmd/celestia-appd/cmd/update_config_test.go
```diff
@@ -140,3 +142,29 @@ func verifyUpdatedConfigs(t *testing.T, configDir, version string) {
...
+		// Setup configs with legacy default min gas price
+		cmtCfg := app.DefaultConsensusConfig()
+		appCfg := app.DefaultAppConfig()
+		// Set legacy default min gas price explicitly
+		appCfg.MinGasPrices = fmt.Sprintf("%v%s", appconsts.LegacyDefaultMinGasPrice, appconsts.BondDenom)
+
+		_, updatedAppCfg := applyV6Config(cmtCfg, appCfg)
+
+		assert.Equal(t, "", updatedAppCfg.MinGasPrices, "MinGasPrices should be unset to empty string if legacy default")
```
rootulp, line 156:
```suggestion
		assert.Equal(t, "", updatedAppCfg.MinGasPrices, "MinGasPrices should be set to empty string if legacy default")
```

rootulp (dismissed):

cmd/celestia-appd/cmd/update_config.go
cmwaters: Let me double check but I think it has to be v6. Iirc correctly I made a similar comment when the update-config command was being implemented that it would be more intuitive to use just 6 (like we do when signalling for example)

cmd/celestia-appd/cmd/update_config_test.go
cmwaters: Yeah unset was to imply that it was previously had a number but now its ""

cmd/celestia-appd/cmd/update_config.go
cmwaters: Ah your totally right we did change it to just be 6. I've updated the PR

PR #5522, opened 2025-08-12
fix(ante): add nil check in FormatTx to prevent panic

oooLowNeoNooo:
Add nil transaction check in FormatTx function to prevent panic

- Added nil check for tx parameter in FormatTx function
- Returns appropriate message when tx is nil instead of panicking
- Added test case to verify nil transaction handling
- Fixes potential runtime panic when HandlePanicDecorator processes nil transactions


evan-forbes:
we have issues open if you would like to contribute @oooLowNeoNooo 

asking some llm to find and fix something doesn't work and isn't useful

see contrib guidelines

PR #5523, opened 2025-08-12
chore: bump to the unofficial commits for comet v0.39

evan-forbes:
## Overview

bumps to core v0.39

ofc, must actually cut the release before merging


app/test/gas_estimation_test.go
```diff
@@ -97,7 +97,7 @@ func TestEstimateGasPrice(t *testing.T) {
 	// price, then test the gas estimator API.
 	accountNames := testfactory.GenerateAccounts(10)
 	cfg := testnode.DefaultConfig().WithFundedAccounts(accountNames...).
-		WithTimeoutCommit(20 * time.Second) // to have all transactions in the mempool without being included in a block
+		WithTimeoutCommit(1700 * time.Millisecond) // to have all transactions in the mempool without being included in a block
```
rach-id: changed this because for some reason, during the execution of some other tests, the values of timeout commit change in memory and trigger the pre-emptive block building.

evan-forbes: from a sync:

this seems crazy as some tests pass when using the flood mempool

it also makes a lot of sense because the failures only occur when the tests are ran together

evan-forbes: we're also seeing the same failure occur in the cosmos-sdk

evan-forbes: there is an issue with how comet does pre-emptive block building although I don't think this is root cause. 

currently the timeout commit being used in premeptive block building is using what is in the commit and not what is being passed from the application. The weird thing is that I think this should be the same value in this case, since the testnode is using the configured value to pass back from the application.

evan-forbes: https://github.com/celestiaorg/celestia-core/pull/2324

rach-id: now that we will revert pre-emptive block building, we should revert these changes.

go.mod
```diff
@@ -272,8 +272,9 @@ replace (
 	cosmossdk.io/api => github.com/celestiaorg/cosmos-sdk/api v0.7.6
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
 	// TODO: update to a v0.39.x release after that is created from celestia-core main branch.
-	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v1.57.3-tm-v0.38.17
-	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.29.6-sdk-v0.50.14
+	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.2
+	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v1.29.6-sdk-v0.50.14.0.20250812210754-fcb7d156613b // v1.29.6-sdk-v0.50.14
+	github.com/cosmos/ibc-go/v8 => github.com/celestiaorg/ibc-go/v8 v8.0.0-20250812222333-dbbce7262310
```
rootulp, line 277: can we use official releases for these?

Issue #5435, opened 2025-08-04
Reduce Inflation as per CIP 41


cmwaters:
This is a tracking issue of [CIP 41](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-041.md) - which reduces inflation in x/mint to 2.5%

rootulp:
I assume [this requirement](https://github.com/celestiaorg/CIPs/blob/f388a5e986830b1546bee192eff0ba02243e4c19/cips/cip-041.md?plain=1#L111) means that we must do two things:
1. In the v6 upgrade handler, override the MinCommisionRate parameter to 10%
2. In the v6 upgrade handler, iterate over all validators and modify their commission rate to 10% if they were previously less than 10% 

Separately we may consider making `MinComissionRate` non governance modifiable because if a subsequent parameter change proposal modifies the `MinComissionRate` up or down, it will only impact new validators.

rootulp:
To investigate:
- On a local testnet, start a chain with a validator at 5% commission rate, modify the min commission rate from 5% to 10% and see if validator gets auto migrated or we need a custom migration handler.
- If we need a custom migration handler, investigate Osmosis

evan-forbes:
ahhh found it https://github.com/osmosis-labs/osmosis/blob/d31c9b386cd6e7f00eea6334df5a1d04c0507f31/app/upgrades/v3/forks.go#L30-L59

osmosis continues to be goats at organizing a sdk based repo

rootulp:
Note if we do an auto migration to modify the commission rate for existing validators from 5% to 10%, I expect some will fail to update because they have a max change rate of 1%.

Ref: https://celenium.io/validators?status=active&page=1

rootulp:
> On a local testnet, start a chain with a validator at 5% commission rate, modify the min commission rate from 5% to 10% and see if validator gets auto migrated or we need a custom migration handler.

Done in https://github.com/celestiaorg/celestia-app/pull/5526#issuecomment-3185449228. We do need the migration logic. Or we remove the migration logic and do this manually:

1. Wait until v6 activates on-chain
2. Submit a governance proposal to modify MinCommissionRate to 10%
3. [Each validator] submit an edit-validator message to increase their commission rate if it was below that

cc: @cmwaters

PR #5526, opened 2025-08-13
feat!: implement CIP-41

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5435

## Description
1. Decrease inflation
2. Increase MinComissionRate from 5% to 10%

## Notes
The upgrade handler in this PR will attempt to update the commission rate for validators to the new min commission rate if the validator's commission rate was previously below that. However, the update will fail for certain validators (example: [SmartStake](https://celenium.io/validator/82827?tab=Delegators)) who have set a max rate that is lower than the new min rate. I also expect it to fail for validators with a small `MaxChangeRate`.

## Follow ups
1. https://github.com/celestiaorg/celestia-app/issues/5528

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	sdkmath "cosmossdk.io/math"
 	upgradetypes "cosmossdk.io/x/upgrade/types"
 	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
 	blobtypes "github.com/celestiaorg/celestia-app/v6/x/blob/types"
@@ -203,6 +204,9 @@ func (a App) SetMinCommisionRate(ctx context.Context) error {
 	return nil
 }
 
+// UpdateValidatorCommissionRates iterates over all validators and increases
+// their commission rate and max commission rate if they are below the new
+// minimum commission rate.
 func (a App) UpdateValidatorCommissionRates(ctx context.Context) error {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 
@@ -213,17 +217,12 @@ func (a App) UpdateValidatorCommissionRates(ctx context.Context) error {
 	}
 
 	for _, validator := range validators {
-		if validator.Commission.Rate.GTE(appconsts.MinCommissionRate) {
-			sdkCtx.Logger().Debug("validator commission rate is greater than or equal to the minimum commission rate, skipping", "validator", validator.GetOperator())
+		if validator.Commission.Rate.GTE(appconsts.MinCommissionRate) && validator.Commission.MaxRate.GTE(appconsts.MinCommissionRate) {
+			sdkCtx.Logger().Debug("validator commission rate and max commission rate are already greater than or equal to the minimum commission rate", "validator", validator.GetOperator())
 			continue
 		}
-
-		sdkCtx.Logger().Info("updating validator commission rate", "validator", validator.GetOperator(), "old-commission", validator.Commission.Rate, "new-commission", appconsts.MinCommissionRate)
-		commission, err := a.StakingKeeper.UpdateValidatorCommission(ctx, validator, appconsts.MinCommissionRate)
-		if err != nil {
-			sdkCtx.Logger().Error("failed to update validator commission rate", "error", err)
-			continue
-		}
+		rate := getMax(validator.Commission.Rate, appconsts.MinCommissionRate)
+		maxRate := getMax(validator.Commission.MaxRate, appconsts.MinCommissionRate)
 
 		valAddr, err := sdk.ValAddressFromBech32(validator.GetOperator())
 		if err != nil {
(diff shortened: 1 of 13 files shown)


ValarDragon:
LGTM!

cmwaters:
I am not familiar with the minimum commission rate logic. If we increase that to 10% won't that enforce that all validators (including existing ones with previous commission rates lower than 10%), have at least 10%? My initial understanding was that the bump to the minimum commission rate would come as a governance proposal (and nothing else was needed)

app/upgrades.go
```diff
@@ -170,3 +182,64 @@ func (a App) setICAHostParams(ctx context.Context) error {
...
+
+	params, err := a.StakingKeeper.GetParams(ctx)
+	if err != nil {
+		sdkCtx.Logger().Error("failed to get staking params", "error", err)
+		return err
+	}
+
+	params.MinCommissionRate = appconsts.MinCommissionRate
+
+	fmt.Printf("Setting the staking params min commission rate to %v.\n", appconsts.MinCommissionRate)
```
evan-forbes: maybe a logger would be more appropriate here?

app/upgrades_test.go
```diff
@@ -64,4 +67,74 @@ func TestApplyUpgrade(t *testing.T) {
...
+			Name:   "v6",
+			Time:   time.Now(),
+			Height: 1,
+			Info:   "info",
+		}
+		// Set the block time to 25 hours ahead of the genesis block to ensure
+		// the commission rate can be updated. If the block time is within 24
+		// hours of the genesis block, the commission rate will fail to update
+		// due to ErrCommissionUpdateTime.
+		ctx = testApp.NewContext(false).WithBlockTime(util.GenesisTime.Add(time.Hour * 25))
```
evan-forbes, line 130: TIL

evan-forbes (dismissed):
nice 👍

app/upgrades.go
rootulp: oops agreed, will fix

rootulp:
> I am not familiar with the minimum commission rate logic. If we increase that to 10% won't that enforce that all validators (including existing ones with previous commission rates lower than 10%), have at least 10%? 

No because there is no logic to modify existing validators commission rates [here](https://github.com/celestiaorg/cosmos-sdk/blob/fe3dd41a9a2a6bcbf87b797518ce650f3e452c80/x/staking/keeper/msg_server.go#L603-L619).

Update: I'll work on a local script to verify my assumption.

> My initial understanding was that the bump to the minimum commission rate would come as a governance proposal (and nothing else was needed)

We can remove the SetMinCommisionRate from this PR and instead submit a governance proposal to modify it. But then all validators with commission rates below the new minimum will have to submit edit validator txs to modify their commission rate.

app/upgrades.go
rootulp: 1c4d030f383f715b3013e3b9ccad7b483c0f8aa4

rootulp:
ugh mint integration test suite is failing. Will fix

rootulp:
> Update: I'll work on a local script to verify my assumption.

Cosmos SDK broke the legacy gov param proposals because now a type is required. When I include the `type: "Test"` which seems like the only valid proposal type, I get

```
raw_log: 'failed to execute message; message index: 0: gov: no handler exists for proposal type'
```

I was referencing:
- https://gist.github.com/rootulp/fcf5160a6506dc23a228a90d68e356bd
- https://github.com/celestiaorg/celestia-app/blob/main/docs/maintainers/modify-param.md

rootulp:
I included a script in this PR to demonstrate that modifying the governance params doesn't impact the commission rate on existing validators.

After the script runs, the staking params are updated but the validator's commission remains at 5%

```
Querying staking params...
params:
  bond_denom: utia
  historical_entries: 10000
  max_entries: 7
  max_validators: 100
  min_commission_rate: "0.100000000000000000"
  unbonding_time: 337h0m0s
Querying staking validators...
pagination:
  total: "1"
validators:
- commission:
    commission_rates:
      max_change_rate: "1.000000000000000000"
      max_rate: "1.000000000000000000"
      rate: "0.050000000000000000"
    update_time: "2025-08-13T19:50:29.665801Z"
  consensus_pubkey:
    type: /cosmos.crypto.ed25519.PubKey
    value: i5Vzuus1hK/6ZLBOpHw0+tbDLu/Hgf5RQMz7tiJ9VrI=
  delegator_shares: "5000000000.000000000000000000"
  description:
    moniker: test
  min_self_delegation: "1"
  operator_address: celestiavaloper18mx278u37h8vtppnlp00zvrsqmx78s0yflg3zk
  status: BOND_STATUS_BONDED
  tokens: "5000000000"
  unbonding_time: "1970-01-01T00:00:00Z"
```

rootulp:
> However, the update will fail for certain validators (example: [SmartStake](https://celenium.io/validator/82827?tab=Delegators)) who have set a max rate that is lower than the new min rate. I also expect it to fail for validators with a small `MaxChangeRate`.

I have a few ideas to make the auto migration more robust:
1. If a validator's max commission rate is lower than the new min commission rate, increase it to the min commission rate
2. Bypass the max rate change check so that all validator's get modified regardless of their max rate change

cmwaters:
> We can remove the SetMinCommisionRate from this PR and instead submit a governance proposal to modify it. But then all validators with commission rates below the new minimum will have to submit edit validator txs to modify their commission rate.

Ah so `SetMinCommissionRate` only works for new validators. That's no good. I agree with your proposal. Let's skip the governance proposal and do everything in the migration:
- Set min commission to 10%
- Increase everyone's commission to at least 10%
- (Possibly) modify some validators max commission to 10% (if it's below)

I can update the CIP

evan-forbes:
> I have a few ideas to make the auto migration more robust:
If a validator's max commission rate is lower than the new min commission rate, increase it to the min commission rate
Bypass the max rate change check so that all validator's get modified regardless of their max rate change

sgtm. makes sense to modify the max rate as well.

rootulp:
> Set min commission to 10%
> Increase everyone's commission to at least 10%
> (Possibly) modify some validators max commission to 10% (if it's below)

Agreed and already done in this PR.

PR #5534, opened 2025-08-13
fix: tests that changing to cat broke

### LINKED ISSUE #5498: Three integration tests failed after switching to CAT
after switching to a default of CAT in #5490, we had to revert 3 tests back to using the priority mempool in order to get them to pass. We also observed a panic in one test. We need to ensure that those tests can run the TxClient w/ CAT and not fail.

those tests were the
- TestBigBlobSuite (saw a panic here, but didn't look into why. it could have just been the test and not txclient)

```
--- FAIL: TestBigBlobSuite (1.73s)
    --- FAIL: TestBigBlobSuite/TestBlobExceedsMaxTxSize (0.12s)
        --- FAIL: TestBigBlobSuite/TestBlobExceedsMaxTxSize/1_MiB_over_the_max_blob_size (0.06s)
            iface.go:275: test panicked: interface conversion: error is *status.Error, not *user.BroadcastTxError
                goroutine 819 [running]:
                runtime/debug.Stack()
                	/opt/hostedtoolcache/go/1.24.2/x64/src/runtime/debug/stack.go:26 +0x5e
                github.com/stretchr/testify/suite.failOnPanic(0xc002962380, {0x3ea0d40, 0xc002a2eab0})
                	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:89 +0x37
                github.com/stretchr/testify/suite.recoverAndFailOnPanic(0xc002962380)
                	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:83 +0x35
                panic({0x3ea0d40?, 0xc002a2eab0?})
                	/opt/hostedtoolcache/go/1.24.2/x64/src/runtime/panic.go:792 +0x132
                github.com/celestiaorg/celestia-app/v6/app/test_test.(*BigBlobSuite).TestBlobExceedsMaxTxSize.func1()
                	/home/runner/work/celestia-app/celestia-app/app/test/big_blob_test.go:76 +0x294
                github.com/stretchr/testify/suite.(*Suite).Run.func1(0xc002962380)
                	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:115 +0x169
                testing.tRunner(0xc002962380, 0xc00282be40)
                	/opt/hostedtoolcache/go/1.24.2/x64/src/testing/testing.go:1792 +0xf4
                created by testing.(*T).Run in goroutine 812
                	/opt/hostedtoolcache/go/1.24.2/x64/src/testing/testing.go:1851 +0x413
```

- TestIntegrationTestSuite

```
--- FAIL: TestIntegrationTestSuite (47.16s)
    --- FAIL: TestIntegrationTestSuite/TestUnwrappedPFBRejection (0.11s)
        integration_test.go:159: 
            	Error Trace:	/home/runner/work/celestia-app/celestia-app/app/test/integration_test.go:159
            	            	application rejected transaction with code 11131 (Log: no blobs provided)
            	Test:       	TestIntegrationTestSuite/TestUnwrappedPFBRejection
    network.go:50: tearing down testnode
FAIL
FAIL	github.com/celestiaorg/celestia-app/v6/app/test	47.263s
```

- TestEvictions

```
--- FAIL: TestEvictions (22.80s)
    tx_client_test.go:305: 
        	Error Trace:	/home/runner/work/celestia-app/celestia-app/pkg/user/tx_client_test.go:305
        	            	resubmission for evicted tx with hash 77DC528588B261253F2A79890E701F5D1F3A3E4E4A83B8CF72DAC7082FA02F9E failed: rpc error: code = Unknown desc = application rejected transaction with code 32 (Log: account sequence mismatch, expected 10, got 3: incorrect account sequence)
        	Test:       	TestEvictions
    network.go:50: tearing down testnode
```

presumably this is due to the different errors emitted by CAT and Priority cc @ninabarbakadze

ninabarbakadze:
## Overview

Fixes #5498 

Opens

### DIFF
--- a/app/test/big_blob_test.go
+++ b/app/test/big_blob_test.go
@@ -2,6 +2,8 @@ package app_test
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -11,7 +13,6 @@ import (
 	"github.com/celestiaorg/celestia-app/v6/test/util/testfactory"
 	"github.com/celestiaorg/celestia-app/v6/test/util/testnode"
 	"github.com/celestiaorg/go-square/v2/share"
-	tmcfg "github.com/cometbft/cometbft/config"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
@@ -38,8 +39,6 @@ func (s *BigBlobSuite) SetupSuite() {
 	cfg := testnode.DefaultConfig().WithFundedAccounts(s.accounts...)
 	// purposefully bypass the configurable mempool check
 	cfg.TmConfig.Mempool.MaxTxBytes = appconsts.MaxTxSize * 2
-	// Use priority mempool for consistent error behavior
-	cfg.TmConfig.Mempool.Type = tmcfg.MempoolTypePriority
 
 	cctx, _, _ := testnode.NewNetwork(t, cfg)
 	s.cctx = cctx
@@ -76,8 +75,8 @@ func (s *BigBlobSuite) TestBlobExceedsMaxTxSize() {
 			res, err := txClient.SubmitPayForBlob(subCtx, []*share.Blob{tc.blob}, user.SetGasLimitAndGasPrice(1e9, appconsts.DefaultMinGasPrice))
 			require.Error(t, err)
 			require.Nil(t, res)
-			code := err.(*user.BroadcastTxError).Code
-			require.Equal(t, tc.expectedCode, code, err.Error(), tc.expectedErr)
+			require.True(t, strings.Contains(err.Error(), tc.expectedErr))
+			require.True(t, strings.Contains(err.Error(), fmt.Sprintf("%d", tc.expectedCode)))
 		})
 	}
 }
--- a/app/test/integration_test.go
+++ b/app/test/integration_test.go
@@ -3,6 +3,8 @@ package app_test
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -18,7 +20,6 @@ import (
 	square "github.com/celestiaorg/go-square/v2"
 	"github.com/celestiaorg/go-square/v2/share"
 	abci "github.com/cometbft/cometbft/abci/types"
-	tmconfig "github.com/cometbft/cometbft/config"
 	rpcclient "github.com/cometbft/cometbft/rpc/client"
 	coretypes "github.com/cometbft/cometbft/types"
 	"github.com/cosmos/cosmos-sdk/client"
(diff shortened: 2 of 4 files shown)


app/test/big_blob_test.go
```diff
@@ -76,8 +75,8 @@ func (s *BigBlobSuite) TestBlobExceedsMaxTxSize() {
 			res, err := txClient.SubmitPayForBlob(subCtx, []*share.Blob{tc.blob}, user.SetGasLimitAndGasPrice(1e9, appconsts.DefaultMinGasPrice))
 			require.Error(t, err)
 			require.Nil(t, res)
-			code := err.(*user.BroadcastTxError).Code
-			require.Equal(t, tc.expectedCode, code, err.Error(), tc.expectedErr)
+			require.True(t, strings.Contains(err.Error(), tc.expectedErr))
+			require.True(t, strings.Contains(err.Error(), fmt.Sprintf("%d", tc.expectedCode)))
```
cmwaters: Is the error no longer of type `BroadcastTxError`?

app/test/integration_test.go
```diff
@@ -158,9 +156,10 @@ func (s *IntegrationTestSuite) TestUnwrappedPFBRejection() {
 	btx, isBlob := coretypes.UnmarshalBlobTx(blobTx[0])
 	require.True(t, isBlob)
 
-	res, err := s.cctx.BroadcastTxSync(btx.Tx)
-	require.NoError(t, err)
-	require.Equal(t, blobtypes.ErrNoBlobs.ABCICode(), res.Code)
+	_, err := s.cctx.BroadcastTxSync(btx.Tx)
```
cmwaters: Is res supposed to be nil? Do we want to check that?

ninabarbakadze: Res is nil because CAT returns errors now for app rejections

app/test/big_blob_test.go
ninabarbakadze: `BroadcastTxError` was set based on the response if response code was no `0`. CAT just throws an error.

app/test/integration_test.go
cmwaters: Yeah can we assert that

app/test/big_blob_test.go
cmwaters: Hmm okay so maybe we should just be focusing on the solution which is to match the error throwing between the two mempool designs

app/test/integration_test.go
evan-forbes: apologies as this was likely previously discussed and answered but I forgot. Why do we want to return an error if the tx was rejected? my understanding is that errors here are reserved to client errors (such as the server is not responsive) where application errors (such as an invalid tx that gets rejected) go in the result

cmwaters: The priority mempool followed that logic. The CAT pool differed in that anything that was not the tx being submitted is an error regardless if it's from the server is not responsive or if the server deems your request invalid (because you don't have enough funds for example)

evan-forbes:
will approve after we also change https://github.com/celestiaorg/celestia-app/blob/e248d537180954e10c96c5c3f21728fb0b9eb7e0/pkg/user/tx_client_test.go#L439

or open a flup ofc

ninabarbakadze:
> will approve after we also change
> 
> https://github.com/celestiaorg/celestia-app/blob/e248d537180954e10c96c5c3f21728fb0b9eb7e0/pkg/user/tx_client_test.go#L439
> 
> or open a flup ofc

should be ready now

evan-forbes (approved):

Issue #5447, opened 2025-08-05
Modify state sync test to test different app versions


rootulp:
## Context

The e2e state sync test uses Tastora to verify that a node can state sync.

https://github.com/celestiaorg/celestia-app/blob/main/test/docker-e2e/e2e_state_sync_test.go

https://www.notion.so/celestiaorg/Testing-and-QA-for-V6-Checklist-2462a70c9515809cafddf7edd4dc867d?source=copy_link#2462a70c9515808c80cfcb4965714f8c

## Problem

During a QA v6 call we discussed a few things that should block the v6 release

## Proposal

- [ ] Verify that the v6 multiplexer binary can produce v5 snapshots when the app version is v5
- [ ] Verify that the v6 multiplexer binary can produce v6 snapshots when the app version is v6
- [ ] Verify that the v6 multiplexer binary can state sync using a v5 snapshot when the app version is v5
- [ ] Verify that the v6 multiplexer binary can state sync using a v6 snapshot when the app version is v6
- [ ] Verify that the test uses state sync and not block sync

rootulp:
Related to https://github.com/celestiaorg/celestia-app/issues/5204

evan-forbes:
imo the plan of attack here would be to:

- fix the state sync test so that it definitively state syncs #5204 
    - this can be done by prune block data prematurely, which is actually blocked by https://github.com/celestiaorg/cosmos-sdk/pull/448. Here we would ensure at least the first block or two no longer exists.
    - after block data has been pruned, actually changing that config to no longer prune. this step  allows the test to avoid having to state sync too quickly while also pruning data quickly. Without this we'd need to let the test run for a few minutes before pruning data.
    - creating a snapshot for every height (simple config change)
- state sync using the embedded binary
- upgrade
- erase the ./data directory (or an unsafe reset) on the state sync'd node and state sync again from the same height by simply restarting  

this exact scenario is a bit exact, but I believe it covers all of the above in the most minimal way possible

fwiw it also acts as a smoke test for block syncing past the upgrade, as that is inherently required

rootulp:
Moving a conversation from Slack. The OP

> - [ ] Verify that the v6 multiplexer binary can produce v5 snapshots when the app version is v5

Done in https://github.com/celestiaorg/celestia-app/pull/5545

> - [ ] Verify that the v6 multiplexer binary can produce v6 snapshots when the app version is v6

[...]

Issue #5385, opened 2025-07-31
Verify all validators are healthy after e2e test


mojtaba-esk:
**Goal**  
Guarantee every E2E test checks for proposer rotation and node liveness.

**Tasks**  
1. Implement `AssertHealthy(chain, startHeight, endHeight)` that  
   - Scans block proposers; verifies every validator proposed ≥ once.  
   - Confirms `/status` does not report halts.  
2. Call this helper in `t.Cleanup` of each E2E test (opt-out flag available).

evan-forbes:
this seems redundant to #5380

rootulp:
The title of this issue doesn't make sense to me. Can it be renamed to something like: "Verify all validators are healthy after e2e test"?

mojtaba-esk:
> this seems redundant to [#5380](https://github.com/celestiaorg/celestia-app/issues/5380)

did not know that one exist. closing this one.

PR #5545, opened 2025-08-15
feat (test-e2e): state sync test while upgrading

### LINKED ISSUE #5382: E2E Test – State-Sync Across Mixed App Versions
**Goal**  
Verify that a full node can *state-sync* to the latest height when the chain history spans multiple app versions.

**Tasks**  
1. Spin up a v4 chain (4 validators) with tx-traffic for 20 blocks.  
2. Initiate on-chain upgrade → v5, let it finalize, produce 50 more blocks.  
3. Launch a new full node with `StateSync.Enable=true`.  
4. Assert:  
   - Node finishes state-sync (`/status` → catching_up=false).  
   - `ABCIInfo.app_version == 5`.  
   - Basic bank-send succeeds from the synced node.  
5. Add helper `WaitForStateSync(chain, node)` inside the test if missing.  
6. Run nightly (not PR-blocking).

**Note**
Using multiple versions must be discussed i.e. from v2 --> v5 or just v4 -> v5 is sufficient

mojtaba-esk:
Closes #5382

### DIFF
--- a/test/docker-e2e/e2e_state_sync_test.go
+++ b/test/docker-e2e/e2e_state_sync_test.go
@@ -4,14 +4,22 @@ import (
 	"celestiaorg/celestia-app/test/docker-e2e/dockerchain"
 	"celestiaorg/celestia-app/test/docker-e2e/networks"
 	"context"
+	"fmt"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
+	sdkmath "cosmossdk.io/math"
+	"github.com/celestiaorg/celestia-app/v6/pkg/user"
+	"github.com/celestiaorg/celestia-app/v6/test/util/genesis"
+	tastoradockertypes "github.com/celestiaorg/tastora/framework/docker"
 	"github.com/celestiaorg/tastora/framework/testutil/config"
 	cometcfg "github.com/cometbft/cometbft/config"
 	rpctypes "github.com/cometbft/cometbft/rpc/core/types"
 	servercfg "github.com/cosmos/cosmos-sdk/server/config"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 
 	celestiadockertypes "github.com/celestiaorg/tastora/framework/docker"
 	addressutil "github.com/celestiaorg/tastora/framework/testutil/address"
@@ -23,6 +31,109 @@ const (
 	stateSyncTrustHeightOffset = 5
 	stateSyncTimeout           = 10 * time.Minute
 )
+
+// keepAliveTransactionGenerator continuously sends small bank transfer transactions
+// to keep the chain producing blocks with changing state. It runs in a goroutine
+// and can be stopped by cancelling the provided context.
+// This version is resilient to connection drops during node restarts (e.g., upgrades).
+func (s *CelestiaTestSuite) keepAliveTransactionGenerator(ctx context.Context, wg *sync.WaitGroup, chain *tastoradockertypes.Chain, cfg *dockerchain.Config, intervalSeconds int) {
+	defer wg.Done()
+
+	// Create a unique recipient wallet for keep-alive transactions
+	recipientWalletName := fmt.Sprintf("keepalive-recipient-%d", time.Now().UnixNano())
+	wallet, err := chain.CreateWallet(ctx, recipientWalletName)
+	if err != nil {
+		s.T().Logf("Failed to create keep-alive recipient wallet: %v", err)
+		return
+	}
+
+	recipientAddr, err := sdk.AccAddressFromBech32(wallet.GetFormattedAddress())
+	if err != nil {
+		s.T().Logf("Failed to parse keep-alive recipient address: %v", err)
+		return
+	}
+
+	s.T().Logf("Keep-alive transaction generator started, recipient: %s", recipientAddr.String())
+
+	ticker := time.NewTicker(time.Duration(intervalSeconds) * time.Second)
+	defer ticker.Stop()
+
+	txCounter := 0
+	var txClient *user.TxClient
+	var fromAddr sdk.AccAddress
+
+	// Helper function to establish or re-establish connection
+	setupConnection := func() bool {
+		var err error
+		txClient, err = dockerchain.SetupTxClient(ctx, chain.GetNodes()[0].(*tastoradockertypes.ChainNode), cfg)
+		if err != nil {
+			s.T().Logf("Failed to setup TxClient for keep-alive transactions: %v", err)
+			return false
+		}
+		fromAddr = txClient.DefaultAddress()
+		s.T().Logf("Keep-alive transaction generator connected: %s -> %s", fromAddr.String(), recipientAddr.String())
+		return true
+	}
+
+	// Initial connection setup
+	if !setupConnection() {
+		s.T().Logf("Keep-alive transaction generator failed to establish initial connection")
+		return
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			s.T().Logf("Keep-alive transaction generator stopped after sending %d transactions", txCounter)
+			return
+		case <-ticker.C:
+			// Send a small bank transfer (1000 utia = 0.001 TIA)
+			sendAmount := sdk.NewCoins(sdk.NewCoin("utia", sdkmath.NewInt(1000)))
+			msg := banktypes.NewMsgSend(fromAddr, recipientAddr, sendAmount)
+
+			// Retry logic with connection re-establishment
+			maxRetries := 3
+			retryDelay := 2 * time.Second
+			success := false
+
+			for attempt := 0; attempt < maxRetries && !success; attempt++ {
+				// Submit transaction with appropriate gas and fee
+				_, err := txClient.SubmitTx(ctx, []sdk.Msg{msg}, user.SetGasLimit(200000), user.SetFee(5000))
+				if err != nil {
+					// Check if it's a connection error
+					if strings.Contains(err.Error(), "connection") || strings.Contains(err.Error(), "rpc error") || strings.Contains(err.Error(), "Unavailable") {
+						s.T().Logf("Keep-alive transaction %d connection error (attempt %d/%d): %v", txCounter+1, attempt+1, maxRetries, err)
+
+						// Wait before retrying and attempt to re-establish connection
+						if attempt < maxRetries-1 {
+							time.Sleep(retryDelay)
+							if setupConnection() {
+								// Update the message with the new fromAddr after reconnection
+								msg = banktypes.NewMsgSend(fromAddr, recipientAddr, sendAmount)
+								// Double the retry delay for next attempt
+								retryDelay *= 2
+							}
+						}
+					} else {
+						// Non-connection error, log and move on
+						s.T().Logf("Keep-alive transaction %d failed (non-connection error): %v", txCounter+1, err)
+						break
+					}
+				} else {
+					success = true
+					txCounter++
+					if txCounter%10 == 0 {
+						s.T().Logf("Keep-alive transaction generator: sent %d transactions", txCounter)
+					}
+				}
+			}
+
+			if !success && maxRetries > 0 {
+				s.T().Logf("Keep-alive transaction %d failed after %d retries", txCounter+1, maxRetries)
+			}
+		}
+	}
+}
 
 // validatorStateSyncAppOverrides modifies the app.toml to configure state sync snapshots for the given node.
 func validatorStateSyncAppOverrides(ctx context.Context, node *celestiadockertypes.ChainNode) error {
@@ -212,4 +323,231 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
 	})
 
 	s.Require().NoError(err, "failed to wait for state sync to complete")
+}
+
+// TestStateSyncAcrossMixedAppVersions tests that a full node can state-sync to the latest height
+// when the chain history spans multiple app versions (v4 → v5).
+//
+// Test steps:
+// 1. Start a v4 chain with 4 validators and tx traffic for 20 blocks
+// 2. Perform on-chain upgrade v4 → v5, finalize, and produce 50 more blocks
+// 3. Launch a new full node with StateSync.Enable=true
+// 4. Assert: node finishes state-sync, app_version == 5, basic bank-send succeeds
+func (s *CelestiaTestSuite) TestStateSyncAcrossMixedAppVersions() {
+	t := s.T()
+	if testing.Short() {
+		t.Skip("skipping state sync across mixed app versions test in short mode")
+	}
+
+	ctx := context.TODO()
+
+	// Get the Docker image tag for the test
+	tag, err := dockerchain.GetCelestiaTagStrict()
+	s.Require().NoError(err)
+
+	// Phase 1: Start v4 chain with 4 validators
+	t.Log("Phase 1: Starting v4 chain with 4 validators")
+	cfg := dockerchain.DefaultConfig(s.client, s.network).WithTag(tag)
+
+	// Configure genesis for v4 with 4 validators total (default has 2, add 2 more)
+	cfg.Genesis = cfg.Genesis.
+		WithAppVersion(4).
+		WithValidators(
+			genesis.NewDefaultValidator("validator3"),
+			genesis.NewDefaultValidator("validator4"),
+		)
+
+	chain, err := dockerchain.NewCelestiaChainBuilder(s.T(), cfg).
+		WithPostInit(validatorStateSyncAppOverrides). // Enable snapshots
+		Build(ctx)
+	s.Require().NoError(err, "failed to build chain")
+
+	t.Cleanup(func() {
+		if err := chain.Stop(ctx); err != nil {
+			t.Logf("Error stopping chain: %v", err)
+		}
+	})
+
+	err = chain.Start(ctx)
+	s.Require().NoError(err, "failed to start chain")
+
+	// Verify chain is running and get initial state
+	height, err := chain.Height(ctx)
+	s.Require().NoError(err, "failed to get chain height")
+	s.Require().Greater(height, int64(0), "chain height should be greater than 0")
+
+	validatorNode := chain.GetNodes()[0]
+	rpcClient, err := validatorNode.GetRPCClient()
+	s.Require().NoError(err, "failed to get RPC client")
+
+	// Verify we're starting with app version 4
+	abciInfo, err := rpcClient.ABCIInfo(ctx)
+	s.Require().NoError(err, "failed to fetch ABCI info")
+	s.Require().Equal(uint64(4), abciInfo.Response.GetAppVersion(), "should start with app version 4")
+
+	// Wait for chain to be ready by waiting for blocks to be produced
+	t.Log("Waiting for chain to be ready")
+	initialHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get initial height")
+	s.Require().Greater(initialHeight, int64(0), "initial height should be greater than 0")
+
+	// Wait for a few blocks to ensure chain is stable
+	s.Require().NoError(wait.ForBlocks(ctx, 3, chain), "failed to wait for chain stability")
+
+	// Start transaction simulation for network activity
+	s.CreateTxSim(ctx, chain)
+
+	// Produce 20 blocks with transaction activity
+	t.Log("Producing 20 blocks with transaction activity")
+	stableHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get stable height")
+
+	targetHeight := stableHeight + 20
+	s.Require().NoError(wait.ForBlocks(ctx, int(targetHeight-stableHeight), chain),
+		"failed to wait for 20 blocks")
+
+	t.Logf("Successfully produced %d blocks at app version 4", targetHeight-stableHeight)
+
+	// Test functionality before upgrade to ensure gRPC is working
+	t.Log("Testing functionality before upgrade")
+
+	// Wait a bit more for gRPC to be fully ready (TxSim is already connected, so gRPC is working)
+	time.Sleep(10 * time.Second)
+
+	testBankSend(s.T(), chain, cfg)
+	testPFBSubmission(s.T(), chain, cfg)
+
+	// Phase 2: Perform v4 → v5 upgrade
+	t.Log("Phase 2: Performing v4 → v5 upgrade")
+
+	kr := cfg.Genesis.Keyring()
+	records, err := kr.List()
+	s.Require().NoError(err, "failed to list keyring records")
+	s.Require().Len(records, len(chain.GetNodes()), "number of accounts should match number of nodes")
+
+	// Signal and execute upgrade from v4 to v5 (following working upgrade test pattern)
+	upgradeHeight := s.signalAndGetUpgradeHeight(ctx, chain, validatorNode, cfg, records, 5)
+
+	// Start keep-alive transaction generator to ensure chain doesn't stall during upgrade
+	upgradeCtx, upgradeCancel := context.WithCancel(ctx)
+	var upgradeWg sync.WaitGroup
+	upgradeWg.Add(1)
+	go s.keepAliveTransactionGenerator(upgradeCtx, &upgradeWg, chain, cfg, 3) // Send tx every 3 seconds
+	t.Log("Started keep-alive transaction generator for upgrade period")
+
+	// Wait for upgrade to complete
+	currentHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get current height")
+
+	blocksToWait := int(upgradeHeight-currentHeight) + 2
+	t.Logf("Waiting for %d blocks to reach upgrade height %d", blocksToWait, upgradeHeight)
+	s.Require().NoError(wait.ForBlocks(ctx, blocksToWait, chain),
+		"failed to wait for upgrade completion")
+
+	// Stop the upgrade keep-alive generator
+	upgradeCancel()
+	upgradeWg.Wait()
+	t.Log("Stopped keep-alive transaction generator for upgrade period")
+
+	// Verify upgrade completed successfully
+	abciInfo, err = rpcClient.ABCIInfo(ctx)
+	s.Require().NoError(err, "failed to fetch ABCI info after upgrade")
+	s.Require().Equal(uint64(5), abciInfo.Response.GetAppVersion(), "app version should be 5 after upgrade")
+
+	// Start keep-alive transaction generator for post-upgrade block production
+	postUpgradeCtx, postUpgradeCancel := context.WithCancel(ctx)
+	var postUpgradeWg sync.WaitGroup
+	postUpgradeWg.Add(1)
+	go s.keepAliveTransactionGenerator(postUpgradeCtx, &postUpgradeWg, chain, cfg, 2) // Send tx every 2 seconds
+	t.Log("Started keep-alive transaction generator for post-upgrade block production")
+
+	// Produce 50 more blocks at v5
+	t.Log("Producing 50 more blocks at app version 5")
+	s.Require().NoError(wait.ForBlocks(ctx, 50, chain),
+		"failed to wait for 50 blocks after upgrade")
+
+	// Stop the post-upgrade keep-alive generator
+	postUpgradeCancel()
+	postUpgradeWg.Wait()
+	t.Log("Stopped keep-alive transaction generator for post-upgrade block production")
+
+	finalHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get final height")
+	t.Logf("Chain now at height %d with mixed version history (v4: blocks 1-%d, v5: blocks %d-%d)",
+		finalHeight, upgradeHeight-1, upgradeHeight, finalHeight)
+
+	// Phase 3: Launch state sync node
+	t.Log("Phase 3: Launching new full node with state sync enabled")
+
+	// Calculate state sync parameters from the v5 portion of the chain
+	latestHeight := finalHeight
+	trustHeight := latestHeight - stateSyncTrustHeightOffset
+	s.Require().Greater(trustHeight, upgradeHeight,
+		"trust height %d should be in v5 portion (after upgrade height %d)", trustHeight, upgradeHeight)
+
+	trustBlock, err := rpcClient.Block(ctx, &trustHeight)
+	s.Require().NoError(err, "failed to get block at trust height %d", trustHeight)
+
+	trustHash := trustBlock.BlockID.Hash.String()
+	rpcServers, err := addressutil.BuildInternalRPCAddressList(ctx, chain.GetNodes())
+	s.Require().NoError(err, "failed to build RPC address list")
+
+	t.Logf("State sync parameters: trust_height=%d, trust_hash=%s", trustHeight, trustHash)
+
+	// Add state sync node
+	err = chain.AddNode(ctx,
+		celestiadockertypes.NewChainNodeConfigBuilder().
+			WithNodeType(celestiadockertypes.FullNodeType).
+			WithPostInit(func(ctx context.Context, node *celestiadockertypes.ChainNode) error {
+				return config.Modify(ctx, node, "config/config.toml", func(cfg *cometcfg.Config) {
+					cfg.StateSync.Enable = true
+					cfg.StateSync.TrustHeight = trustHeight
+					cfg.StateSync.TrustHash = trustHash
+					cfg.StateSync.RPCServers = strings.Split(rpcServers, ",")
+				})
+			}).
+			Build(),
+	)
+	s.Require().NoError(err, "failed to add state sync node")
+
+	allNodes := chain.GetNodes()
+	stateSyncNode := allNodes[len(allNodes)-1]
+	s.Require().Equal("fn", stateSyncNode.GetType(), "expected state sync node to be a full node")
+
+	stateSyncClient, err := stateSyncNode.GetRPCClient()
+	s.Require().NoError(err, "failed to get state sync client")
+
+	// Phase 4: Wait for state sync and validate
+	t.Log("Phase 4: Waiting for state sync completion and validation")
+
+	err = s.WaitForSync(ctx, stateSyncClient, stateSyncTimeout, func(info rpctypes.SyncInfo) bool {
+		return !info.CatchingUp && info.LatestBlockHeight >= latestHeight
+	})
+	s.Require().NoError(err, "failed to wait for state sync to complete")
+
+	// Validate: /status shows catching_up=false
+	status, err := stateSyncClient.Status(ctx)
+	s.Require().NoError(err, "failed to get status from state sync node")
+	s.Require().False(status.SyncInfo.CatchingUp, "state sync node should not be catching up")
+	t.Logf("✓ State sync node status: catching_up=%t, height=%d",
+		status.SyncInfo.CatchingUp, status.SyncInfo.LatestBlockHeight)
+
+	// Validate: ABCIInfo.app_version == 5
+	syncedAbciInfo, err := stateSyncClient.ABCIInfo(ctx)
+	s.Require().NoError(err, "failed to fetch ABCI info from state sync node")
+	s.Require().Equal(uint64(5), syncedAbciInfo.Response.GetAppVersion(),
+		"state sync node should have app version 5")
+	t.Logf("✓ State sync node app version: %d", syncedAbciInfo.Response.GetAppVersion())
+
+	// Validate: Basic bank send succeeds from synced node
+	t.Log("Testing bank send transaction from state sync node")
+	testBankSend(s.T(), chain, cfg)
+	t.Log("✓ Bank send transaction succeeded")
+
+	// Final liveness check
+	t.Log("Performing final liveness check")
+	s.Require().NoError(s.CheckLiveness(ctx, chain), "liveness check failed")
+	t.Log("✓ Liveness check passed")
+
+	t.Log("State sync across mixed app versions test completed successfully!")
 }
--- a/.github/workflows/test-nightly.yml
+++ b/.github/workflows/test-nightly.yml
@@ -47,3 +47,37 @@ jobs:
         run: make test-docker-e2e test=TestStateSyncMocha
 
       # TODO: add slack notification on failure
+
+  test-state-sync-mixed-versions:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+
+      - name: Set up Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 #v5.5.0
+        with:
+          go-version-file: "go.mod"
+
+      - name: Generate image tag
+        id: tag
+        run: |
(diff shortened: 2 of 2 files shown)


test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	if testing.Short() {
+		t.Skip("skipping state sync compatibility across upgrade test in short mode")
+	}
+
+	ctx := context.Background()
+
+	tag, err := dockerchain.GetCelestiaTagStrict()
+	s.Require().NoError(err)
+
+	t.Log("Phase 1: Starting v4 chain with 4 validators")
```
rootulp: [nit] v4 will be incorrect when we bump appconsts.Version. Can this log instead use `appconsts.Version - 1`?

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	cfg := dockerchain.DefaultConfig(s.client, s.network).WithTag(tag)
+
+	var (
+		baseAppVersion   = appconsts.Version - 1
+		targetAppVersion = appconsts.Version
+	)
+	t.Logf("Starting chain with app version %d -> %d", baseAppVersion, targetAppVersion)
```
rootulp: [nit] this log is misleading because the chain can't start with app version "4 -> 5". Maybe

```suggestion
	t.Logf("Starting chain with app version %d and will upgrade to %d.", baseAppVersion, targetAppVersion)
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+// checkSyncMetrics queries Prometheus metrics to determine sync method
+func checkSyncMetrics(t *testing.T, node *tastoradockertypes.ChainNode) (stateSync bool, blockSync bool, err error) {
+	ctx := context.Background()
+
+	// Try multiple approaches to get metrics
+	endpoints := []string{
+		"http://localhost:26660/metrics", // Default prometheus port
+		"http://127.0.0.1:26660/metrics", // Alternative localhost
+		"http://localhost:26657/metrics", // Alternative CometBFT port
+		"http://127.0.0.1:26657/metrics", // Alternative CometBFT port
```
rootulp: [question] are these really necessary? I assume tastora spins up the node with a reliable Prometheus endpoint so seems odd that we need to try all these.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+		t.Logf("Detected block sync: incremental progression pattern in %v", heightHistory)
+		return false, true
+	}
+
+	t.Logf("Detected state sync: no incremental pattern in %v", heightHistory)
+	return true, false
+}
+
+// verifySyncMethod validates that state sync was used and not block sync
+func verifySyncMethod(t *testing.T, heightHistory []int64, stateSyncNode *tastoradockertypes.ChainNode) {
```
rootulp: [nit] this function name isn't self describing. Perhaps

`verifySyncMethod` => `verifyStateSync` or `verifyStateSyncOccurred`

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	t.Logf("Height progression during sync: %v", heightHistory)
+
+	// Analyze height progression pattern
+	usedStateSync, usedBlockSync := analyzeSyncMethod(t, heightHistory)
+
+	// FAIL if block sync was detected
+	if usedBlockSync {
+		t.Fatalf("CRITICAL FAILURE: Node used BLOCK SYNC instead of STATE SYNC! "+
+			"This test requires state sync. Height progression: %v", heightHistory)
```
rootulp: IMO the all caps aren't necessary

```suggestion
		t.Fatalf("Failed because state sync node used block sync instead of state sync. Height progression: %v", heightHistory)
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	}
+
+	// FAIL if we couldn't determine the sync method
```
rootulp: This comment is misleading because it seems like the test did determine the sync method, it just happened to not be state sync

```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	if !usedStateSync {
+		t.Fatalf("Could not confirm state sync was used. "+
+			"Height progression: %v", heightHistory)
```
rootulp: Inline

```suggestion
		t.Fatalf("Could not confirm state sync was used. Height progression: %v", heightHistory)
```

Also what is height progression? How does it differ from height history?

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,423 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+}
+
+// validatorStateSyncProducerOverrides configures validators to produce state sync snapshots.
+func validatorStateSyncProducerOverrides(ctx context.Context, node *tastoradockertypes.ChainNode) error {
+	return config.Modify(ctx, node, "config/app.toml", func(cfg *servercfg.Config) {
+		// configure frequent state sync snapshots for new nodes to consume
+		cfg.StateSync.SnapshotInterval = 5
+		cfg.StateSync.SnapshotKeepRecent = 3
+	})
+}
```
rootulp: [question] why are both of these necessary? They seem duplicative.

rootulp (commented):
[question] is it possible to remove all the code that analyzes block history for incremental blocks vs a jump in blocks?

The metrics approach to determine if the node used state sync or block sync seems clearer and easier to maintain.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: nice catch! sure will fix it

mojtaba-esk:
> [question] is it possible to remove all the code that analyzes block history for incremental blocks vs a jump in blocks?
> 
> The metrics approach to determine if the node used state sync or block sync seems clearer and easier to maintain.

That was added to have more certainty on the sync method. Are you sure metric would be enough? if so and it is always available, sure I will remove the history check so the code will be lighter.

rootulp:
> Are you sure metric would be enough?

I'm not sure because I haven't tested myself but I would assume that metrics are always emitted. If metrics aren't reliably emitted then that seems like a bug.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: you're right, it is overengineered. will fix it.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: thanks `verifyStateSync` seems more clear.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: They're mostly duplicative. Height progression analysis ([0, N] vs [0,1,2,3...]) is reliable enough to detect state sync vs block sync. The Prometheus metrics check was added as extra validation, but it can fail if the endpoint has issues or somehow is not available. Also the height progression thing can have edge cases, but mostly more reliable than metrics as the metrics tell us what CometBFT intended, not what actually happened from the users perspective. So I would keep both verification methods.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
 
 	s.Require().NoError(err, "failed to wait for state sync to complete")
 }
+
+// TestStateSyncCompatibilityAcrossUpgrade verifies that a full node can state-sync to the latest height
+// when the chain history includes an on-chain upgrade.
```
rootulp: [nit] there is no off-chain upgrade so specifying "on-chain" here isn't necessary.

```suggestion
// when the chain history includes an app version upgrade.
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	testBankSend(s.T(), chain, cfg)
+	testPFBSubmission(s.T(), chain, cfg)
+
+	t.Log("Phase 2: Performing upgrade")
+	upgradeHeight := s.performUpgrade(ctx, chain, cfg, targetAppVersion)
+
+	finalHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get final height")
+	t.Logf("Chain now at height %d with mixed version history (%d: blocks 1-%d, %d: blocks %d)",
```
rootulp: [nit] this log is pretty confusing to read. It could be clearer

```suggestion
	t.Logf("Blocks 1 to %d have app version %d", upgradeHeight-1, baseAppVersion)
	t.Logf("Blocks %d to %d have app version %d", upgradeHeight, finalHeight, targetAppVersion)
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+		finalHeight, baseAppVersion, upgradeHeight-1, targetAppVersion, finalHeight)
+
+	t.Log("Validators configured with state sync snapshots every 5 blocks")
```
rootulp: [nit] this log isn't near the config for 5 blocks so it's highly likely this log won't be updated if/when the state sync config changes from 5 blocks to something else. IMO this log doesn't add any value so it could be deleted entirely. If you want to keep it, can it be relocated near where the 5 block config is located?

```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+				return configureStateSyncClient(ctx, node, strings.Split(rpcServers, ","), trustHeight, trustHash)
+			}).
+			Build(),
+	)
+	s.Require().NoError(err, "failed to add state sync node")
+
+	var (
+		allNodes      = chain.GetNodes()
+		stateSyncNode = allNodes[len(allNodes)-1]
+	)
```
rootulp: [nit]

```suggestion
	allNodes      := chain.GetNodes()
	stateSyncNode := allNodes[len(allNodes)-1]
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	t.Log("Validators configured with state sync snapshots every 5 blocks")
+
+	t.Log("Phase 3: Launching new full node with state sync enabled")
+
+	// Calculate state sync parameters from the targetAppVersion portion of the chain
+	var (
+		latestHeight = finalHeight
+		trustHeight  = latestHeight - stateSyncTrustHeightOffset
+	)
```
rootulp: [nit]

```suggestion
	latestHeight := finalHeight
	trustHeight  := latestHeight - stateSyncTrustHeightOffset
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	var (
+		allNodes      = chain.GetNodes()
+		stateSyncNode = allNodes[len(allNodes)-1]
+	)
+	s.Require().Equal(tastoratypes.NodeTypeConsensusFull, stateSyncNode.GetType(), "expected state sync node to be a full node")
+
+	stateSyncClient, err := stateSyncNode.GetRPCClient()
+	s.Require().NoError(err, "failed to get state sync client")
+
+	t.Log("Phase 4: Waiting for state sync completion and validation")
```
rootulp:
```suggestion
	t.Log("Waiting for node to state sync...")
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	// get fresh chain height for accurate comparison (chain continued producing blocks during state sync)
+	currentChainHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get current chain height")
+
+	// Verify that the node actually performed state sync by checking it reached current height
+	nodeHeight := status.SyncInfo.LatestBlockHeight
+	s.Require().GreaterOrEqual(nodeHeight, currentChainHeight-5,
+		"state sync node should be within 5 blocks of the current chain tip")
+	s.Require().LessOrEqual(nodeHeight, currentChainHeight+2,
+		"state sync node should not be significantly ahead of the chain")
```
rootulp: [question] why is the + or - not the same. Seems plausible for these to both be + or - 1 or 2 blocks.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	t.Logf("State sync successful: node reached height %d (current chain tip: %d)",
+		nodeHeight, currentChainHeight)
+
+	// Validate: ABCIInfo.app_version == targetAppVersion
```
rootulp:
```suggestion
	// Verify ABCIInfo.app_version == targetAppVersion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	syncedAbciInfo, err := stateSyncClient.ABCIInfo(ctx)
+	s.Require().NoError(err, "failed to fetch ABCI info from state sync node")
+	s.Require().Equal(targetAppVersion, syncedAbciInfo.Response.GetAppVersion(),
+		"state sync node should have app version %d", targetAppVersion)
+	t.Logf("State sync node app version: %d", syncedAbciInfo.Response.GetAppVersion())
+
+	// Validate: Basic bank send succeeds from synced node
+	t.Log("Testing bank send transaction from state sync node")
+	testBankSend(s.T(), chain, cfg)
+	t.Log("Bank send transaction succeeded")
```
rootulp: IMO this isn't necessary

```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	// Final liveness check
+	t.Log("Performing final liveness check")
+	s.Require().NoError(s.CheckLiveness(ctx, chain), "liveness check failed")
+	t.Log("Liveness check passed")
```
rootulp: [question] why is this needed?

```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+}
+
+// checkSyncMetrics queries Prometheus metrics to determine sync method
+func checkSyncMetrics(t *testing.T, node *tastoradockertypes.ChainNode) (stateSync bool, blockSync bool, err error) {
+	var (
+		ctx      = context.Background()
+		endpoint = "http://localhost:26657/metrics"
+		cmd      = []string{"curl", "-s", "--connect-timeout", "5", endpoint}
+	)
```
rootulp: [nit] these `var()` blocks only seem idiomatic when they're outside function bodies.

[nit] prefer long form flags in code. Short form flags can be used on the CLI when readability isn't a requirement.

```suggestion
	ctx      := context.Background()
	endpoint := "http://localhost:26657/metrics"
	cmd      := []string{"curl", "--silent", "--connect-timeout", "5", endpoint}
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	stdout, stderr, execErr := node.Exec(ctx, cmd, nil)
+
+	if execErr != nil {
+		return false, false, fmt.Errorf("failed to fetch metrics from %s: %v, stderr: %s", endpoint, execErr, string(stderr))
+	}
+
+	metrics := string(stdout)
+	if len(metrics) < 10 {
+		return false, false, fmt.Errorf("empty or invalid metrics response from %s", endpoint)
+	}
```
rootulp: [question] why do there need to be 10 results? Will this return an error if 9 metrics are returned from the endpoint? That seems incorrect.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+		t.Logf("Prometheus metrics: Block sync completed (blocksync_syncing=0)")
+		return false, true, nil
+	}
+
+	t.Logf("Prometheus metrics available but no sync metrics found")
+	return false, false, nil
+}
+
+// analyzeSyncMethod determines if state sync or block sync was used based on height progression
+func analyzeSyncMethod(t *testing.T, heightHistory []int64) (usedStateSync bool, usedBlockSync bool) {
```
rootulp: [blocking] why is this still needed? I thought the outcome of the previous review was to just use the metrics and not look at heights.

test/docker-e2e/e2e_state_sync_test.go
rootulp: My question was about these lines:

```
// validatorStateSyncAppOverrides modifies the app.toml to configure state sync snapshots for the given node.
func validatorStateSyncAppOverrides(ctx context.Context, node *tastoradockertypes.ChainNode) error {
	return config.Modify(ctx, node, "config/app.toml", func(cfg *servercfg.Config) {
		cfg.StateSync.SnapshotInterval = 5
	})
}

// validatorStateSyncProducerOverrides configures validators to produce state sync snapshots.
func validatorStateSyncProducerOverrides(ctx context.Context, node *tastoradockertypes.ChainNode) error {
	return config.Modify(ctx, node, "config/app.toml", func(cfg *servercfg.Config) {
		// configure frequent state sync snapshots for new nodes to consume
		cfg.StateSync.SnapshotInterval = 5
		cfg.StateSync.SnapshotKeepRecent = 3
	})
}
```

Why do we need two different configs that only differ based on `SnapshotKeepRecent = 2` vs `3`? Please delete one.

rootulp: I'm strongly against the existing height progression code because it seems extremely difficult to maintain.

> The Prometheus metrics check was added as extra validation, but it can fail if the endpoint has issues or somehow is not available.

Why would the endpoint have issues or be unavailable? If that method has issues and the issue isn't unique to tastora we should create an issue, fix metrics for state sync, and then use it in this PR. 

> but mostly more reliable than metrics as the metrics tell us what CometBFT intended, not what actually happened from the users perspective.

Is this true? I don't expect CometBFT to emit metrics claiming that state sync is occurring if it doesn't actually state sync.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +210,408 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	if testing.Short() {
+		t.Skip("skipping state sync compatibility across upgrade test in short mode")
+	}
+
+	ctx := context.Background()
+
+	tag, err := dockerchain.GetCelestiaTagStrict()
+	s.Require().NoError(err)
+
+	t.Log("Phase 1: Starting the chain with 3 validators")
```
rootulp: why does this use 3 validators instead of 1?

Ref: https://github.com/celestiaorg/celestia-app/issues/5382#issuecomment-3139925176

mojtaba-esk: This is the default behaviour of Tastora. I think we can configure it to use only one. let me try.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: we can simply remove it.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: :)

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: ok tried it and found out that state sync requires at least 2 RPC servers for the client to verify blocks through cross-validation. So let's keep it as it is.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: that -5 is for network delays and the time it takes for the state sync node to catch up after syncing and the +2 is to prevent test flakiness from timing/ race conditions. We could remove +2 and use the exact match to catch potential bugs. Not sure which way is the best. WDYT?

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: it is just a sanity check, we could remove it if it is not needed.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: We added this to all tastora based e2e tests to make sure the chain worked properly and not halted and participated in block proposal. ref: #5385

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: True, since curl will return an error if something is not right and then we look for string matches in the output and if it is empty it will be already returning an error, this check seems unnecessary and could be removed entirely.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: Can we simply rely on the metrics? as it is just what cometBFT reports not what actually happened. If that's enough, we can remove it.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: ok will remove it then.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: saw your other comment, will remove it.

test/docker-e2e/e2e_state_sync_test.go
chatton: [nit] we should probably make the log output the actual number of validators instead of "3" since we may accidentally be using a different number. This can come from a part of the `cfg` below I believe. Maybe the len of the records.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,276 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	t.Logf("State sync node app version: %d", syncedAbciInfo.Response.GetAppVersion())
+
+	// Final liveness check
+	t.Log("Performing final liveness check")
+	s.Require().NoError(s.CheckLiveness(ctx, chain), "liveness check failed")
+	t.Log("Liveness check passed")
+}
+
+// checkSyncMetrics queries Prometheus metrics to determine sync method
+func checkSyncMetrics(t *testing.T, node *tastoradockertypes.ChainNode) (stateSync bool, blockSync bool, err error) {
```
chatton: This is a great way to verify that are state syncing instead of block syncing given the oddities of the sdk options. My only suggestion here would be to check the metrics using native go code instead of using curl, e.g. something like 

```golang
      // Get the metrics endpoint URL - would need to get the actual exposed port
      endpoint := "http://localhost:26657/metrics"

      // Create HTTP request
      req, err := http.NewRequestWithContext(ctx, "GET", endpoint, nil)
      if err != nil {
          return false, false, fmt.Errorf("failed to create request: %w", err)
      }
```

If something is blocking this approach I'm happy to have this as a follow up issue since it is functionally doing the same thing, it might even be nice to have that feature in tastora to extract metrics 👀 

I think ideally we exec into the containers as little as possible and interact with the exposed APIs from within the test when possible

mojtaba-esk: Thanks @chatton, yes there is the metrics port limitation, right now the only working solution is to use `curl`, once Tastora have the port issue resolved in a flup, we can update this part of the code to directly fetch it.

mojtaba-esk:
> The test looks great overall! I left a few comments but one overall suggestion for test structuring is to put each of the larger pieces of the test under its own `t.Run`, basically each of the "phases" in the test could be a subtest to make the `go test` output more legible instead of reading through the logs all at one level.
> 
> The test is not actually be run in ci though as far as I can tell, it needs to be added [here](https://github.com/celestiaorg/celestia-app/blob/32f59198fa129c8eaddf9d09155d0c173e8e0554/.github/workflows/test.yml#L128)

Let's have this as a followup.

.github/workflows/test.yml
```diff
@@ -135,6 +135,7 @@ jobs:
             entrypoint: TestIBCTestSuite
           - testcase: TestICA
             entrypoint: TestIBCTestSuite
+          - testcase: TestStateSyncCompatibilityAcrossUpgrade
```
rootulp: [optional] can we shorten to something like

```suggestion
          - testcase: TestStateSyncWithAppUpgrade
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	s.Require().NoError(err, "failed to fetch ABCI info")
+	s.Require().Equal(baseAppVersion, abciInfo.Response.GetAppVersion(), "should start with app version %v", baseAppVersion)
+
+	s.CreateTxSim(ctx, chain)
+
+	t.Log("Producing 20 blocks with transaction activity")
+	s.Require().NoError(wait.ForBlocks(ctx, 20, chain), "failed to wait for 20 blocks")
+
+	t.Log("Testing functionality before upgrade")
+
```
rootulp:
```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	t.Logf("Blocks 1 to %d have app version %d", upgradeHeight-1, baseAppVersion)
+	t.Logf("Blocks %d to %d have app version %d", upgradeHeight, finalHeight, targetAppVersion)
+
+	t.Log("Phase 3: Launching new full node with state sync enabled")
+
+	// Calculate state sync parameters from the targetAppVersion portion of the chain
+	latestHeight := finalHeight
+	trustHeight := latestHeight - stateSyncTrustHeightOffset
+	s.Require().Greater(trustHeight, upgradeHeight,
+		"trust height %d should be in %d portion (after upgrade height %d)", trustHeight, targetAppVersion, upgradeHeight)
```
rootulp: [discussion] I think it would actually be more useful for the state syncing node to use a snapshot from _before_ the upgrade height. If both the snapshot producer and snapshot consumer are on the latest version then this test is almost identical to the existing state sync e2e test. The point of upgrading during the test is to verify that state sync works across the app upgrade boundary.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	allNodes := chain.GetNodes()
+	stateSyncNode := allNodes[len(allNodes)-1]
+	s.Require().Equal(tastoratypes.NodeTypeConsensusFull, stateSyncNode.GetType(), "expected state sync node to be a full node")
+
+	stateSyncClient, err := stateSyncNode.GetRPCClient()
+	s.Require().NoError(err, "failed to get state sync client")
+
+	t.Log("Waiting for node to state sync...")
+	// Wait for sync to complete
```
rootulp: self explanatory by the log above

```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+					t.Logf("State sync confirmed: applied %s snapshot chunks", countStr)
+					return true, false, nil
+				}
+			}
+		}
+	}
+
+	// No evidence of state sync found
+	return false, false, nil
+}
```
rootulp: [question] why does this return blockSync as a return argument if it is always `false`?

test/docker-e2e/e2e_state_sync_test.go
rootulp: The function name and return arguments make this a bit confusing to understand. Consider renames:

```go
// getSyncMethod returns the sync method (stateSync or blockSync) used by the node based on Prometheus metrics.
func getSyncMethod(t *testing.T, node *tastoradockertypes.ChainNode) (stateSyncUsed bool, blockSyncUsed bool, err error) {
```

test/docker-e2e/e2e_state_sync_test.go
rootulp: Ref: https://github.com/celestiaorg/celestia-app/issues/5447

> Verify that the v6 multiplexer binary can state sync using a v5 snapshot when the app version is v5

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	if metricBlockSync {
+		t.Fatal("FAILED: Prometheus metrics show BLOCK SYNC was used instead of state sync!")
+	}
+
+	if !metricStateSync {
+		t.Fatal("FAILED: Could not confirm state sync was used via Prometheus metrics!")
+	}
+
+	t.Log("SUCCESS: Prometheus metrics confirm state sync was used")
```
rootulp: [nit] why are some of these all caps and some aren't? These types of minor inconsistencies make this code less pleasing to read.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	validatorNode := chain.GetNodes()[0]
+	kr := cfg.Genesis.Keyring()
+	records, err := kr.List()
+	s.Require().NoError(err, "failed to list keyring records")
+	s.Require().Len(records, len(chain.GetNodes()), "number of accounts should match number of nodes")
+
+	upgradeHeight := s.signalAndGetUpgradeHeight(ctx, chain, validatorNode, cfg, records, targetAppVersion)
+
+	// Wait for upgrade to complete
```
rootulp: This comment doesn't apply to the line following it.
```suggestion
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+	rpcClient, err := validatorNode.GetRPCClient()
+	s.Require().NoError(err, "failed to get RPC client")
+
+	currentHeight, err := s.GetLatestBlockHeight(ctx, rpcClient)
+	s.Require().NoError(err, "failed to get current height")
+
+	blocksToWait := int(upgradeHeight-currentHeight) + 2
+	t.Logf("Waiting for %d blocks to reach upgrade height %d", blocksToWait, upgradeHeight)
+	s.Require().NoError(wait.ForBlocks(ctx, blocksToWait, chain),
+		"failed to wait for upgrade completion")
```
rootulp: [nit] why arbitrary new lines?
```suggestion
	s.Require().NoError(wait.ForBlocks(ctx, blocksToWait, chain), "failed to wait for upgrade completion")
```

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+
+	if !metricStateSync {
+		t.Fatal("FAILED: Could not confirm state sync was used via Prometheus metrics!")
+	}
+
+	t.Log("SUCCESS: Prometheus metrics confirm state sync was used")
+}
+
+// performUpgrade executes the upgrade to the target app version
+func (s *CelestiaTestSuite) performUpgrade(ctx context.Context, chain tastoratypes.Chain, cfg *dockerchain.Config, targetAppVersion uint64) int64 {
```
rootulp: [nit] nothing about this function signature communicates that the return argument is the upgrade height.

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
...
+		cfg.Telemetry.Enabled = true
+	})
+	if err != nil {
+		return err
+	}
+
+	return config.Modify(ctx, node, "config/config.toml", func(cfg *cometcfg.Config) {
+		cfg.Instrumentation.Prometheus = true
+		cfg.Instrumentation.PrometheusListenAddr = "0.0.0.0:26660"
+	})
```
rootulp: [question] why does Prometheus need to be enabled on the state sync producer?

test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -214,3 +209,303 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
 
 	s.Require().NoError(err, "failed to wait for state sync to complete")
 }
+
+// TestStateSyncCompatibilityAcrossUpgrade verifies that a full node can state-sync to the latest height
+// when the chain history includes an app version upgrade.
+func (s *CelestiaTestSuite) TestStateSyncCompatibilityAcrossUpgrade() {
```
rootulp: [nit] can we extract this to it's own file?

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: ah this is a leftover from the previous code where we used to check the sync method and it used to return which method was used. will fix it asap

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: nice catch. right it has to be enabled only on the client. this must have left here during the debugging process. will remove it.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: sure, that could make it cleaner.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: Great point! totally agreed. The test should use a snapshot from before the upgrade height (when the chain was running the old app version) and have the state sync node consume that snapshot while running the new app version. This would actually test state sync across the upgrade boundary, which is the core purpose of this test. will fix it.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: good point, the return value could have a name to reflect that.

test/docker-e2e/e2e_state_sync_test.go
mojtaba-esk: Thanks, fixed now

test/docker-e2e/e2e_state_sync_upgrade_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	t.Log("Producing 20 blocks with transaction activity")
+	s.Require().NoError(wait.ForBlocks(ctx, 20, chain), "failed to wait for 20 blocks")
+
+	t.Log("Testing functionality before upgrade")
+	testBankSend(s.T(), chain, cfg)
+	testPFBSubmission(s.T(), chain, cfg)
+
+	t.Log("Phase 2: Waiting for snapshot creation before upgrade")
+	// Wait for a snapshot to be created at the current app version (5)
```
rootulp, line 78: Here we're using "current" but the variable name is "base" app version

```suggestion
	// Wait for a snapshot to be created at the base app version
```

test/docker-e2e/e2e_state_sync_upgrade_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	trustBlock, err := rpcClient.Block(ctx, &trustHeight)
+	s.Require().NoError(err, "failed to get block at trust height %d", trustHeight)
+
+	trustHash := trustBlock.BlockID.Hash.String()
+	rpcServers, err := addressutil.BuildInternalRPCAddressList(ctx, chain.GetNodes())
+	s.Require().NoError(err, "failed to build RPC address list")
+
+	t.Logf("State sync parameters: trust_height=%d, trust_hash=%s", trustHeight, trustHash)
+	t.Logf("Cross-version state sync: snapshot from app version %d, consumer on app version %d", baseAppVersion, targetAppVersion)
```
rootulp, line 108: [nit] technically the consumer isn't on the target app version yet. It's a fresh node on the latest binary version but the app version should be the base app version after state sync completes, then it will block sync to the latest height and only after syncing through the upgrade height will it be on the target app version.

Seems safe to delete this log

```suggestion
```

test/docker-e2e/e2e_state_sync_upgrade_test.go
```diff
@@ -0,0 +1,328 @@
...
+	t.Logf("State sync successful: node reached height %d (current chain tip: %d)", nodeHeight, currentChainHeight)
+
+	// Verify ABCIInfo.app_version == targetAppVersion
+	syncedAbciInfo, err := stateSyncClient.ABCIInfo(ctx)
+	s.Require().NoError(err, "failed to fetch ABCI info from state sync node")
+	s.Require().Equal(targetAppVersion, syncedAbciInfo.Response.GetAppVersion(), "state sync node should have app version %d", targetAppVersion)
+	t.Logf("State sync node app version: %d", syncedAbciInfo.Response.GetAppVersion())
+
+	// Verify cross-version state sync success
+	t.Logf("Success: Cross-version state sync completed - consumed snapshot from app version %d with node running app version %d", baseAppVersion, targetAppVersion)
```
rootulp, line 166: [nit] again the state syncing node was running the latest binary version not the target app version. 

It is fair to say that the state syncing node eventually is on targetAppVersion at the end of this test but it's probably more informative to include the binary version here.

Note: you'll need to define binaryVersion

```suggestion
	t.Logf("Success: Cross-version state sync completed - consumed snapshot from app version %d with node running binary version %d", baseAppVersion, binaryVersion)
```

test/docker-e2e/e2e_state_sync_upgrade_test.go
rootulp: also (5) is likely to become obsolete when we next bump the app version

test/docker-e2e/e2e_state_sync_upgrade_test.go
mojtaba-esk: Sorry Saw this message late, will fix them in a followup

mojtaba-esk: The followup PR is here: #5758

PR #5561, opened 2025-08-18
refactor: skip BBR check on non-Linux machines

### LINKED ISSUE #3853: Expand on BBR instructions
Might be good to specify on which systems these commands work.

_Originally posted by @rach-id in https://github.com/celestiaorg/celestia-app/pull/3851#discussion_r1751344713_

Additional ideas:
1. Provide instructions to enable BBR on as many platforms / operating systems as possible.
2. Skip the BBR check on platforms / operating systems that don't support BBR so that users don't need to provide `--force-no-bbr`.

Fibonacci747:
Closes #3853 

## Testing

On MacOS

```
$ make check-bbr
Checking if BBR is enabled...
BBR is not available on non-Linux systems.
$ make enable-bbr
Configuring system to use BBR...
BBR is not available on non-Linux systems.
```

On Linux

```shell
root@rootulp-scaleway:~/celestia-app# make enable-bbr
Configuring system to use BBR...
BBR is not enabled. Configuring BBR...
tcp_bbr
net.core.default_qdisc=fq
net.ipv4.tcp_congestion_control=bbr
net.core.default_qdisc = fq
net.ipv4.tcp_congestion_control = bbr
BBR has been enabled.

root@rootulp-scaleway:~/celestia-app# make check-bbr
Checking if BBR is enabled...
BBR is enabled.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,14 +60,14 @@ endif
 .PHONY: build
 
 ## install-standalone: Build and install the celestia-appd binary into the $GOPATH/bin directory. This target does not install the multiplexer.
-install-standalone: check-bbr
+install-standalone:
 	@echo "--> Installing celestia-appd"
 	@go install $(BUILD_FLAGS_STANDALONE) ./cmd/celestia-appd
 .PHONY: install-standalone
 
 ## install: Build and install the multiplexer version of celestia-appd into the $GOPATH/bin directory.
 # TODO: Improve logic here and in goreleaser to make it future proof and less expensive.
-install: check-bbr download-v3-binaries download-v4-binaries download-v5-binaries
+install: download-v3-binaries download-v4-binaries download-v5-binaries
 	@echo "--> Installing celestia-appd with multiplexer support"
 	@go install $(BUILD_FLAGS_MULTIPLEXER) ./cmd/celestia-appd
 .PHONY: install
@@ -421,7 +421,10 @@ goreleaser: prebuilt-binary
 ## check-bbr: Internal command to check if BBR congestion control is enabled on the system.
 check-bbr:
 	@echo "Checking if BBR is enabled..."
-	@if [ "$$(sysctl net.ipv4.tcp_congestion_control | awk '{print $$3}')" != "bbr" ]; then \
+	@if [ "$$(uname -s)" != "Linux" ]; then \
+		echo "BBR is not available on non-Linux systems."; \
+		exit 0; \
+	elif [ "$$(sysctl net.ipv4.tcp_congestion_control | awk '{print $$3}')" != "bbr" ]; then \
 		echo "WARNING: BBR is not enabled. Please enable BBR for optimal performance. Call make enable-bbr or see Usage section in the README."; \
 	else \
 		echo "BBR is enabled."; \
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -84,14 +84,22 @@ See <https://docs.celestia.org/how-to-guides/celestia-app> for more information.
...
-# If BBR is not enabled then enable it.
+# If BBR is not enabled then enable it (Linux only).
 make bbr-enable
 ```
 
+Platform notes:
+- Linux: Requires kernel >= 4.9. The `bbr-check` / `bbr-enable` targets use `modprobe` and `sysctl` to enable and persist BBR.
+- macOS: System-wide BBR is not available. No action is required; the node runs with the default congestion control.
+- Windows: System-wide BBR is not available. No action is required; the node runs with the default congestion control. If using WSL2, BBR can be enabled inside the Linux environment as above.
+- *BSD: Availability varies by distribution and version. If unsupported, the check is skipped automatically.
```
rootulp: Why *BSD?

Given the code change elsewhere in this file, it looks like the check-bbr command silently skips if the runtime OS isn't Linux so I would expect this to say something similar to the macOS and Windows sections.

Fibonacci747: > Why *BSD?
> 
> Given the code change elsewhere in this file, it looks like the check-bbr command silently skips if the runtime OS isn't Linux so I would expect this to say something similar to the macOS and Windows sections.

Corrected

rootulp (approved):

Makefile
```diff
@@ -60,14 +60,14 @@ endif
 .PHONY: build
 
 ## install-standalone: Build and install the celestia-appd binary into the $GOPATH/bin directory. This target does not install the multiplexer.
-install-standalone: check-bbr
```
rach-id, line 63: why is `check-bbr` not called anymore?

scripts/arabica.sh
```diff
@@ -57,4 +57,4 @@ echo "Downloading genesis file..."
 celestia-appd download-genesis ${CHAIN_ID}
 
 echo "Starting celestia-appd..."
-celestia-appd start --v2-upgrade-height 1751707 --force-no-bbr
+celestia-appd start --v2-upgrade-height 1751707
```
rach-id, line 60: I think this is still useful to have because even if your OS supports BBR, not necessarily you want it to be enabled. Similar to other scripts too.

What do you think @rootulp?

rootulp: Hmm that's fair. I figured if a user was on Linux they would configure BBR and then not see anymore warnings. I can revert this from the scripts if you prefer.

Makefile
rootulp: IMO it is unnecessary friction for the Makefile commands to also check-bbr because it is already checked inside the binary.

rootulp: Also previously `check-bbr` was only run for install commands but not build commands which seems inconsistent.

scripts/arabica.sh
rach-id: i'll defer to your decision on this

rootulp: I'm on Mac so it doesn't impact me but if you're on Linux then I can revert

rach-id: no need then

Issue #5444, opened 2025-08-05
Optimize process proposal via building the block in parallel


rach-id:
As described in https://github.com/celestiaorg/celestia-app/issues/5396#issuecomment-3149292218, we could parallelize the block building in process proposal to improve performance.

rach-id:
Update: 

The assumption on this was that parallelizing the block construction and the transaction validation could yield performance improvements. Apparently, it's not the case given how heavy the block construction is, which saturates all CPU power.

@ninabarbakadze implemented this: https://github.com/celestiaorg/celestia-app/pull/5571 and we ran benchmarks on 128mb blocks and got the following results:

- before:
```
process_proposal:
  Count: 694
  Min:   0.267 ms
  Max:   3721.759 ms
  Avg:   1125.152 ms
```

<img width="1117" height="546" alt="Image" src="https://github.com/user-attachments/assets/e5599112-864f-4b0b-9e6b-7707fd068087" />

- after:
```
process_proposal:
  Count: 478
  Min:   0.320 ms
  Max:   2722.078 ms
  Avg:   1077.379 ms
```

<img width="837" height="402" alt="Image" src="https://github.com/user-attachments/assets/7a36ba1e-aa0e-48eb-ba19-45ace3772ff7" />


This means that any further parallelization wouldn't impact performance as much.

PR #5571, opened 2025-08-20
feat: parallelize tx validation and square construction

ninabarbakadze:

### DIFF
--- a/app/process_proposal.go
+++ b/app/process_proposal.go
@@ -23,6 +23,13 @@ import (
 
 const rejectedPropBlockLog = "Rejected proposal block:"
 
+// squareResult holds the result of square construction
+type squareResult struct {
+	dataSquare square.Square
+	dah        da.DataAvailabilityHeader
+	err        error
+}
+
 func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcessProposal) (resp *abci.ResponseProcessProposal, err error) {
 	defer telemetry.MeasureSince(time.Now(), "process_proposal")
 	// In the case of a panic resulting from an unexpected condition, it is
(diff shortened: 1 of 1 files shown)


app/process_proposal.go
```diff
@@ -23,6 +23,13 @@ import (
 
 const rejectedPropBlockLog = "Rejected proposal block:"
 
+// squareResult holds the result of square construction
```
rach-id: I don't think we need this. The error can be logged and then we could simply return whether an error happened or not

rach-id (commented):
Great start 👍 

I think we can improve this to also support not finishing the processing in a go routine if the other found an error. Also, you could only have a single go routine while the other uses the main thread to run. But I don't have any heavy preference over this.

In my opinion, given that transaction validation is iterative, we could rely on a go context or some signaling to signal whether the square generation failed at any step.
But also since the square generation doesn't use any contexts, I think we can check whether the transaction validation failed at every heavy step using some signaling. 

The way I think about it, both the square generation and the transaction validation should stop if any of them fail at any step. Because, if we leave for example the square generation running even if a tx is invalid, that's wasted time.

I'll defer to you on how to structure the implementation.

ninabarbakadze:
> Great start 👍
> 
> I think we can improve this to also support not finishing the processing in a go routine if the other found an error. Also, you could only have a single go routine while the other uses the main thread to run. But I don't have any heavy preference over this.
> 
> In my opinion, given that transaction validation is iterative, we could rely on a go context or some signaling to signal whether the square generation failed at any step. But also since the square generation doesn't use any contexts, I think we can check whether the transaction validation failed at every heavy step using some signaling.
> 
> The way I think about it, both the square generation and the transaction validation should stop if any of them fail at any step. Because, if we leave for example the square generation running even if a tx is invalid, that's wasted time.
> 
> I'll defer to you on how to structure the implementation.

Thanks for the feedback! Changed it to use error groups and if one routine fails the other one will get cancelled. Let's see if this makes a difference in the experiment tomorrow.

app/process_proposal.go
rach-id, line 1: The defer on line 32 could probably be deleted. Also, the handler could be moved inside the go routine.

app/process_proposal.go
```diff
@@ -54,112 +55,163 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+		if err := gctx.Err(); err != nil {
+			return err
 		}
 
-		blobTx, isBlobTx, err := blobtx.UnmarshalBlobTx(rawTx)
-		if isBlobTx {
+		// iterate over all txs and ensure that all blobTxs are valid, PFBs are correctly signed, non
+		// blobTxs have no PFBs present and all txs are less than or equal to the max tx size limit
+		for idx, rawTx := range req.Txs {
+
```
rach-id, line 78:
```suggestion
```

app/process_proposal.go
```diff
@@ -54,112 +55,163 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+	g.Go(func() (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				logInvalidPropBlock(app.Logger(), ctx.BlockHeader(), fmt.Sprintf("caught panic: %v", r))
+				telemetry.IncrCounter(1, "process_proposal", "panics")
+				err = fmt.Errorf("panic in transaction validation: %v", r)
+			}
+		}()
+
+		// check if context is cancelled
```
rach-id, line 70: this I guess is unnecessary since it will be executed almost immediately after the function starts

app/process_proposal.go
```diff
@@ -54,112 +55,163 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+		defer func() {
+			if r := recover(); r != nil {
+				logInvalidPropBlock(app.Logger(), ctx.BlockHeader(), fmt.Sprintf("caught panic: %v", r))
+				telemetry.IncrCounter(1, "process_proposal", "panics")
+				err = fmt.Errorf("panic in square construction: %v", r)
+			}
+		}()
+
+		// check if context is cancelled
+		if gctx.Err() != nil {
```
rach-id, line 164: same

app/process_proposal.go
```diff
@@ -54,112 +55,163 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+			logInvalidPropBlockError(app.Logger(), blockHeader, "failure to erasure the data square", err)
+			return err
+		}
 
-	eds, err := da.ExtendShares(share.ToBytes(dataSquare))
-	if err != nil {
-		logInvalidPropBlockError(app.Logger(), blockHeader, "failure to erasure the data square", err)
-		return reject(), nil
-	}
+		// check if context is cancelled
```
rach-id, line 190:
```suggestion
		// check if context is canceled
```

rach-id:
Closing as per: https://github.com/celestiaorg/celestia-app/issues/5444#issuecomment-3210561513

Issue #5521, opened 2025-08-12
After digging around delegation to understand how things work, I think this change should fix it. I would sleep a little easier if I knew there was a test that actually calls BeginRedelegation and Undelegation and then WithdrawDelegationRewards (i.e. an actual integration test)


evan-forbes:
After digging around delegation to understand how things work, I think this change should fix it. I would sleep a little easier if I knew there was a test that actually calls BeginRedelegation and Undelegation and then WithdrawDelegationRewards (i.e. an actual integration test)

_Originally posted by @cmwaters in https://github.com/celestiaorg/cosmos-sdk/pull/655#pullrequestreview-3110490488_

add integration test in the std sdk test to withdraw rewards w/ no delegation. we only need to check if there was not a failure after redelegating.

Issue #5381, opened 2025-07-31
Can not claim existing reward after redelegating to a new validator or undelegating


cmwaters:
From Keplr:

I found that when a user performs a maximum undelegation or redelegation, the existing delegation is removed but the rewards remain.
In this case, since rewards still exist, the user attempts to claim rewards from the validator. However, because there is no longer any delegation to that validator, an error occurs.
From what I briefly checked, this part of the code —
https://github.com/celestiaorg/cosmos-sdk/blob/1dafc072cea3b44fe6823b2bfd0239594fc75ef7/x/distribution/keeper/keeper.go#L136 —
seems to take into account the case where there is no delegation.
However, this method —
https://github.com/celestiaorg/cosmos-sdk/blob/1dafc072cea3b44fe6823b2bfd0239594fc75ef7/x/staking/keeper/delegation.go#L31 —
returns an error if there is no delegation, so an error is actually returned earlier at this line:
https://github.com/celestiaorg/cosmos-sdk/blob/1dafc072cea3b44fe6823b2bfd0239594fc75ef7/x/distribution/keeper/keeper.go#L132.
As a result, the logic intended to handle the absence of delegation is never reached.

edit:

for those looking for a fix:
- ideally the user claims rewards before redelegating/undelegating
- if that cannot be done, then the user simply needs to delegate to that validator again with _any_ amount ( 1 utia for example ) before claiming rewards.

a fix is scheduled to be included in the v6 (matcha) hardfork (september or october 2025)

cmwaters:
@tac0turtle could you look into this when you get a moment

cmwaters:
<img width="2448" height="954" alt="Image" src="https://github.com/user-attachments/assets/be6ce2ce-8b18-4c60-96b7-380711b689ef" />

Some other users are noticing this as well

evan-forbes:
I generated an integration test that I believe confirms the issue

```go
// TestDelegationRewardsAfterFullUndelegation tests the scenario where:
// 1. A user delegates to a validator
// 2. Earns some rewards
// 3. Undelegates entirely
// 4. Attempts to claim rewards and expects an error
func TestDelegationRewardsAfterFullUndelegation(t *testing.T) {
	// Set up testnode with funded accounts
	accounts := testnode.RandomAccounts(2)
	config := testnode.DefaultConfig().WithFundedAccounts(accounts...)
	cctx, _, _ := testnode.NewNetwork(t, config)

[...]

PR #5580, opened 2025-08-21
fix!: claim rewards after undelegating

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/5381 by upgrading to https://github.com/celestiaorg/cosmos-sdk/releases/tag/v0.51.0
Closes https://github.com/celestiaorg/celestia-app/issues/5521

## Testing
1. Unit test
2. `single-node.sh` + the `test-claim.sh` script in this PR


rootulp:
Converting to draft because I want to add a test to verify this fixes the issue.

evan-forbes (dismissed):

rootulp:
Currently the test fails here

```
--- FAIL: TestClaimRewardsAfterFullUndelegation (14.76s)
    /Users/rootulp/git/rootulp/celestiaorg/celestia-app/app/test/claim_test.go:87: Rewards before undelegation: 0.031303234515000000utia
    /Users/rootulp/git/rootulp/celestiaorg/celestia-app/app/test/claim_test.go:126: 
        	Error Trace:	/Users/rootulp/git/rootulp/celestiaorg/celestia-app/app/test/claim_test.go:126
        	            	tx execution failed with code 5: failed to execute message; message index: 0: no delegation distribution info
        	Test:       	TestClaimRewardsAfterFullUndelegation
    /Users/rootulp/git/rootulp/celestiaorg/celestia-app/app/test/network.go:50: tearing down testnode
FAIL
FAIL	github.com/celestiaorg/celestia-app/v6/app/test	15.469s
```

I have low confidence on the fix in cosmos-sdk.

Failing here: https://github.com/celestiaorg/cosmos-sdk/blob/20c9f61bed6175d6e4405b2862ef44d15cf74391/x/distribution/keeper/keeper.go#L138

Which was introduced in https://github.com/celestiaorg/cosmos-sdk/pull/437

rootulp:
I was able to claim rewards with the script so seems like it does work. Maybe the test is broken.

app/test/claim_test.go
```diff
@@ -0,0 +1,190 @@
...
+	delegationAmount := math.NewInt(1_000_000_000_000) // 1,000,000 TIA
+	amount := types.NewCoin(appconsts.BondDenom, delegationAmount)
+
+	delegateToValidator(t, &cctx, txClient, delegatorAddress, validatorAddress, amount)
+	verifyDelegationExists(t, &cctx, stakingClient, delegatorAddress, validatorAddress, delegationAmount)
+	verifyRewardsExist(t, &cctx, distributionClient, delegatorAddress, validatorAddress)
+
+	undelegate(t, &cctx, txClient, delegatorAddress, validatorAddress, amount)
+	verifyDelegationDoesNotExist(t, &cctx, stakingClient, delegatorAddress, validatorAddress)
+	verifyDelegationRewardsDoNotExist(t, &cctx, distributionClient, delegatorAddress, validatorAddress)
```
cmwaters, line 50: Shouldn't the query still show outstanding rewards that can be claimed?

app/test/claim_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	delegatorAddress := getDelegatorAddress(t, &cctx, accounts)
+	validatorAddress := getValidatorAddress(t, &cctx)
+
+	delegationAmount := math.NewInt(1_000_000_000_000) // 1,000,000 TIA
+	amount := types.NewCoin(appconsts.BondDenom, delegationAmount)
+
+	delegateToValidator(t, &cctx, txClient, delegatorAddress, validatorAddress, amount)
+	verifyDelegationExists(t, &cctx, stakingClient, delegatorAddress, validatorAddress, delegationAmount)
+	verifyRewardsExist(t, &cctx, distributionClient, delegatorAddress, validatorAddress)
```
cmwaters, line 46: Would we need to wait at least 1 block to make sure that there are rewards that have accumulated?

cmwaters (dismissed):
LGTM 👍  - just have a few outstanding questions

evan-forbes (dismissed):

evan-forbes:
same question as cmwaters on the waiting for the block. seems like we would need / want to add that step to keep the test robust even assuming really fast block times.

app/test/claim_test.go
rootulp: Rewards have accumulated without explicitly waiting for one block. Likely because the tx client submits txs and waits for it to be included so at least one block has occurred before this verify line is executed.

app/test/claim_test.go
rootulp: Yea. Seems confusing. I'll investigate as a follow-up.

PR #5582, opened 2025-08-21
feat: resign nonce rejected txs during broadcast

### LINKED ISSUE #5454: Consider checking for nonce mismatches during broadcast
## Description

Currently, we already hit nonce mismatches during broadcast. We should consider adding sequence mismatch handling logic there too.

ninabarbakadze:
## Overview

Fixes #5454


pkg/user/tx_client.go
```diff
@@ -416,6 +418,17 @@ func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn,
 		return nil, err
 	}
 	if resp.TxResponse.Code != abci.CodeTypeOK {
+		if apperrors.IsNonceMismatchCode(resp.TxResponse.Code) {
```
ninabarbakadze, line 444: do we want to limit retries or retry until we get it right and tx client gets unblocked?

cmwaters: I think it's fine. It should also stop if the user cancels the context or hits their deadline

pkg/user/tx_client.go
```diff
@@ -427,6 +463,71 @@ func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn,
...
+		}
+	}
+
+	broadcastTxResp, err := s.broadcastTx(ctx, s.conns[0], newTxBytes, signer)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add to tx tracker.
+	s.trackTransaction(signer, broadcastTxResp.TxHash, newTxBytes)
```
cmwaters, line 527: Do we want to remove the previously tracked transaction (else this will keep appending everytime it loops)

pkg/user/tx_client_test.go
```diff
@@ -109,6 +123,10 @@ func (suite *TxClientTestSuite) TestSubmitTx() {
 	msg := bank.NewMsgSend(addr, testnode.RandomAddress().(sdk.AccAddress), sdk.NewCoins(sdk.NewInt64Coin(params.BondDenom, 10)))
 
 	t.Run("submit tx without provided fee and gas limit", func(t *testing.T) {
+		// This should fail in gas estimation and recover
+		seqBeforeBroadcast := suite.txClient.Signer().Account(suite.txClient.DefaultAccountName()).Sequence()
+		err := suite.txClient.Signer().SetSequence(suite.txClient.DefaultAccountName(), seqBeforeBroadcast+1)
+		require.NoError(t, err)
```
cmwaters, line 129: Are we setting an incorrect sequence number and seeing if this recovers? Shouldn't it be it's own test?

cmwaters (commented):
In terms of code structure, I think it would have been better to introduce this trusted mode (and only use this new code path if trusted mode is set) and have it as the default. Then later on we can add a non trusting mode. This also makes it a little more explicit the trust assumptions they have with the consensus node

pkg/user/tx_client.go
```diff
@@ -427,6 +463,71 @@ func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn,
...
+		txBuilder.SetFeeAmount(fee)
+	}
+	if gas := tx.GetGas(); gas > 0 {
+		txBuilder.SetGasLimit(gas)
+	}
+
+	signer, _, err := s.signer.signTransaction(txBuilder)
+	if err != nil {
+		return nil, fmt.Errorf("resigning transaction: %w", err)
+	}
```
evan-forbes, line 506: not blocking

this logic looks familar, are we repeating it elsewhere?

pkg/user/tx_client.go
```diff
@@ -356,7 +359,29 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
+				return nil, err
+			}
+
+			// Handle the sequence mismatch path by setting the sequence to the expected sequence
+			// and retrying gas estimation.
+			parsedErr := extractSequenceError(err.Error())
+
+			expectedSequence, err := apperrors.ParseExpectedSequence(parsedErr)
+			if err != nil {
+				return nil, fmt.Errorf("parsing sequence mismatch: %w. RawLog: %s", err, err)
```
evan-forbes, line 373: not blocking

parsing strings to determine the correct sequence seems like a brittle and hacky way to get the sequence. 

is this because we don't have a way to query the sequence from the mempool?

evan-forbes:
> This also makes it a little more explicit the trust assumptions they have with the consensus node

wait what's the trust tradeoff w/ this approach? just that we're querying the consensus node for the latest nonce instead of relying on our own? I'm assuming that the reason this is more attackable is due to the mempool app hash inherently not being signed over? or maybe there is a discussion elsewhere describing why we need this?

is the light client in node choosing which consensus node to submit txs to?

pkg/user/tx_client.go
```diff
@@ -356,7 +359,29 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
...
+
+			if err = client.signer.SetSequence(account, expectedSequence); err != nil {
+				return nil, fmt.Errorf("setting sequence: %w", err)
+			}
+
+			// Retry gas estimation with the corrected sequence.
+			gasLimit, err = client.estimateGas(ctx, txBuilder)
+			if err != nil {
+				return nil, fmt.Errorf("retrying gas estimation: %w", err)
+			}
```
rootulp, line 384: [question][not blocking] why does the new expected sequence impact gas estimation? I would expect gas estimation to return the same result. If that's true then we can remove the invocation to re-estimate gas.

pkg/user/tx_client.go
```diff
@@ -356,7 +359,29 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
 		}
 		gasLimit, err = client.estimateGas(ctx, txBuilder)
 		if err != nil {
-			return nil, err
+			// If not a sequence mismatch, return the error.
+			if !strings.Contains(err.Error(), sdkerrors.ErrWrongSequence.Error()) {
```
rootulp, line 363: [question] why compare the strings instead of something like

```go
			if !errors.Is(err, sdkerrors.ErrWrongSequence) {
				return nil, err
			}
```

pkg/user/tx_client_test.go
rootulp, line 129: Why should this fail in gas estimation? I would expect a sequence mismatch error to not impact gas estimation.

pkg/user/tx_client_test.go
```diff
@@ -118,6 +136,10 @@ func (suite *TxClientTestSuite) TestSubmitTx() {
 	})
 
 	t.Run("submit tx with provided gas limit", func(t *testing.T) {
+		// This should fail in gas estimation and recover
+		seqBeforeBroadcast := suite.txClient.Signer().Account(suite.txClient.DefaultAccountName()).Sequence()
+		err := suite.txClient.Signer().SetSequence(suite.txClient.DefaultAccountName(), seqBeforeBroadcast+1)
+		require.NoError(t, err)
```
rootulp, line 142: Why should this fail in gas estimation? I would expect a sequence mismatch error to not impact gas estimation.

pkg/user/tx_client_test.go
```diff
@@ -148,6 +170,20 @@ func (suite *TxClientTestSuite) TestSubmitTx() {
...
+		resp, err := suite.txClient.BroadcastTx(suite.ctx.GoContext(), []sdk.Msg{msg}, feeOption, gasLimitOption)
+		require.NoError(t, err)
+		require.Equal(t, abci.CodeTypeOK, resp.Code)
+
+		// does not block further submissions
+		resp, err = suite.txClient.BroadcastTx(suite.ctx.GoContext(), []sdk.Msg{msg}, feeOption, gasLimitOption)
+		require.NoError(t, err)
+		require.Equal(t, abci.CodeTypeOK, resp.Code)
+	})
```
rootulp, line 186: Is there a unit test that verifies that the tx is re-broadcast with the correct nonce? I don't see that assertion in this test.

pkg/user/tx_client.go
ninabarbakadze: > is this because we don't have a way to query the sequence from the mempool?

Yes

pkg/user/tx_client.go
ninabarbakadze: We unmarshall tx bytes in `CheckTx` which is similar and also the way we rebuild transactions is similar to some of the logic in broadcast functions. I borrowed this logic from an old branch which could be why it's familiar.

pkg/user/tx_client_test.go
ninabarbakadze: > Are we setting an incorrect sequence number and seeing if this recovers?

yes

> Shouldn't it be it's own test?

Why? We're testing broadcast functionality.

pkg/user/tx_client.go
ninabarbakadze: Yep good idea. It'd eventually get purged when it hits ttl but why not already remove them here to not risk running out of memory

pkg/user/tx_client.go
ninabarbakadze: It's a good question. I was hitting nonce failures during gas estimation before it'd even get to broadcast. I think it's because when we simulate a transaction we sign it with the current signer sequence. Adjusting the signer sequence fixes the gas estimation nonce error.

cmwaters:
> wait what's the trust tradeoff w/ this approach? just that we're querying the consensus node for the latest nonce instead of relying on our own? I'm assuming that the reason this is more attackable is due to the mempool app hash inherently not being signed over? or maybe there is a discussion elsewhere describing why we need this?

We're trusting the consensus node for the nonces. It could lie and make us to sign the same message several times breaking replay protection guarantees. 

The previous approach was to not do that but to just track nonces locally only.

> is the light client in node choosing which consensus node to submit txs to?

Yes

pkg/user/tx_client_test.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/5582#discussion_r2309680274

pkg/user/tx_client.go
ninabarbakadze: `errors.Is` was unable to find `sdkerrors.ErrWrongSequence` in the error tree I think because it's wrapped in grpc error.

pkg/user/tx_client_test.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/5582#discussion_r2309680274

pkg/user/tx_client_test.go
ninabarbakadze: That assertion is on line 180. The flow is:
- Manipulated signer sequence to be off by 5
- Submitted the transaction which was signed with incorrect nonce and should get resigned with the expected sequence after nonce error.
- We make sure that the transaction was successfully submitted with `CodeTypeOK`

evan-forbes:
> We're trusting the consensus node for the nonces. It could lie and make us to sign the same message several times breaking replay protection guarantees.
The previous approach was to not do that but to just track nonces locally only.
is the light client in node choosing which consensus node to submit txs to?
Yes

useful info, thanks! if the light node has the option to choose then I think we're good there as they can switch if this is happening to them. imo if we can get past nonce mismatches without having to handle each situation there perfectly this seems like a great tradeoff

pkg/user/tx_client.go
evan-forbes: icky but nice hack I guess

evan-forbes (approved):
maybe some flups left, but imo let's try this out asap

pkg/user/tx_client.go
cmwaters: I wonder what branch of state estimate gas uses? Is it checktxstate? Do they also increment the nonce

cmwaters: > I was hitting nonce failures during gas estimation before it'd even get to broadcast.

Was this systematic. It seems rather hacky to have to always check what the error is to know the actual sequence that needs to be used

cmwaters: Is there a way to recreate the nonce error in the estimation?

pkg/user/tx_client.go
ninabarbakadze: Actually, this is not an issue because transactions only get added to tx tracker after successful CheckTx

PR #5605, opened 2025-08-26
docs: add spec for first fibre message

evan-forbes:
## Overview

adds spec for the mechanism for fibre clients to get fibre provider ip addresses.

almost good for review besides me wanting to change the single string to a bounded slice of strings


x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+}
+```
+
+**Validation Rules:**
+- `validator_address` must be a valid validator operator address
+- `ip_address` must be non-empty and ≤ 45 characters
+- Signer must match the validator address
+- Validator must be in the active set
+
+### MsgRemoveFibreProviderInfo
```
rach-id, line 65: 1. Who calls this?
2. Also, how to make sure validators in the active set actually register their IP address? 
3. And how to guarantee the validator is actually exposing that IP and is accessible?

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+}
+```
+
+### Store Keys
+
+- `0x01 | ValAddress -> ProtocolBuffer(FibreProviderInfo)`: Maps validator address to fibre provider info
+
+## Messages
+
+### MsgSetFibreProviderInfo
```
rach-id, line 48: How to update the validator address in case the validator changed addresses?

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+4. [Events](#events)
+5. [Queries](#queries)
+6. [Parameters](#parameters)
+7. [Client](#client)
+
+## Concepts
+
+### Fibre Service Provider
+
+Every validator in the active set is a Fibre Service Provider (FSPs). Each FSP register's their service address to the celestia-app state. Fibre clients encode data and send unique chunks to each FSP. In return, each FSP signs over a commitment to that data, indicating that they have downloaded it, verified that the encoding is uniquely decodable, and will serve that data upon request for at least the service period.
```
rach-id: What if they register their ip address but its not accessible?

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+  // validator_address is the operator address of the validator
+  string validator_address = 1;
+  // ip_address is the IP address for the fibre service (max 45 characters for IPv6)
+  string ip_address = 2;
+}
+```
+
+**Validation Rules:**
+- `validator_address` must be a valid validator operator address
+- `ip_address` must be non-empty and ≤ 45 characters
```
rach-id: Why 45 characters long? Do we allow DNS? If so, that can be of any length

x/fibre/spec.md
evan-forbes: > Who calls this?

anyone. perhaps this is a premature optimization, but I just wanted some mechanism to delete state. Currently iirc the sdk doens't have a way to delete inactive validators so maybe we don't need this.

> Also, how to make sure validators in the active set actually register their IP address?

the only way to do this afaiu is social consensus unless you want to start to jail folks for not posting something. we can add that later but currently is not technically a problem therefore don't need to for v1 imo

> And how to guarantee the validator is actually exposing that IP and is accessible?

good q, once again there is only one way afaiu and that's social consensus

x/fibre/spec.md
evan-forbes: we just call this message again

x/fibre/spec.md
rach-id: interesting. So if we assume social consensus will define this. Does it make sense to have the list of IP addresses be added to state via a trusted entity? Then whenever anything changes, that trusted entity updates the entries. 

The way I see it, we just need a way to publish Validator IPs in a verifiable way. So we might even think of just posting them in the celestia docs without adding this module.

Sorry if my questions are late the discussion

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+### MsgSetFibreProviderInfo
+
+Allows a validator to set or update their fibre provider information.
+
+```protobuf
+message MsgSetFibreProviderInfo {
+  // validator_address is the operator address of the validator
+  string validator_address = 1;
+  // ip_address is the IP address for the fibre service (max 45 characters for IPv6)
+  string ip_address = 2;
```
walldiss: What is validator_address?

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+Query fibre provider information for all validators in the active set.
+
+**Request:**
+```protobuf
+message QueryAllActiveFibreProvidersRequest {
+  // pagination defines optional pagination parameters
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
+```
```
walldiss: Imo we can start without pagination. If there are 100 validators response size should not be that bad to require pagination. And it complicates logic of the client

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+  // pagination defines the pagination response
+  cosmos.base.query.v1beta1.PageResponse pagination = 2;
+}
+
+message ActiveFibreProvider {
+  // validator_address is the operator address of the validator
+  string validator_address = 1;
+  // info contains the fibre provider information
+  FibreProviderInfo info = 2;
+}
```
walldiss: Need pub key of validator to verify signatures. And also need to add voting power of val. Fubre client would need to ensure 66% of stake has voted on blob

x/fibre/spec.md
```diff
@@ -0,0 +1,221 @@
...
+  // info contains the fibre provider information
+  FibreProviderInfo info = 1;
+  // found indicates if the validator has registered info
+  bool found = 2;
+}
+```
+
+### QueryAllActiveFibreProviders
+
+Query fibre provider information for all validators in the active set.
```
walldiss, line 133: Ideally there should be another message to query diff from previous requested state of ValSet. Similar git fetch. Request would need to point to previous state hash / block heigh and server will be able to server just diff if there is any. This would save quite some bandwidth on periodic updates of ValSet on client (for both client and server)

x/fibre/spec.md
evan-forbes: > So if we assume social consensus will define this. Does it make sense to have the list of IP addresses be added to state via a trusted entity? 

I'm missing the connection between social consensus for validators actually performing a service and a trusted entity posting the set. For example, we currently rely on social consensus to pick validators that produce uncensored blocks. Why would it be okay to rely on a centralized service to determine that for light clients?

> The way I see it, we just need a way to publish Validator IPs in a verifiable way

It is true that validators could post their addresses via any decentralized mechanism since light clients know the validator set and therefore can verify that they are communicating with.

Either we have the ability to update the IPs or they are static forever. If they are not static forever, then light clients need a way to sync the latest. Doing this in a bullet proof way over a p2p network would be difficult, but doable. We'd likely end up replicating a lot of the functionality of a decentralized state machine, as there is some state (the validator's ips) that can only be modified by some deterministic rules, such as the respective validator is the only one that can update their IP.

x/fibre/spec.md
evan-forbes: I changed it to 90 actually so I need to update this. we can't have any length due to state and tx size being bounded. better to just set an explicit limit imo.

originally 45 is the max length of an ipv6 address https://stackoverflow.com/questions/166132/maximum-length-of-the-textual-representation-of-an-ipv6-address#166157

was also thinking about changing this to a []string (bounded in length say 10 elements)

evan-forbes: and yeah there's nothing stopping anyone from using DNS

x/fibre/spec.md
evan-forbes: valopaddress is the address in the state that the validator uses to submit txs to the statemachine. so to first register as a validator or update their existing info such as name etc. for example

`celestiavaloper1jkuw8rxxrsgn9pq009987kzelkp46cgcaa7lhj`

https://celestia.explorers.guru/validator/celestiavaloper1jkuw8rxxrsgn9pq009987kzelkp46cgcaa7lhj

each has one, and this key is expected to be secure.

x/fibre/spec.md
evan-forbes: sure we can remove this

by default however not providing any parameters will provide the entire set

x/fibre/spec.md
evan-forbes: this message is strictly for a validator to update the state for their FibreProviderInfo, why do we need the public key or voting power for that? the state machine already has those things.

the state machine already knows the active set, so can verify simply by comparing the valop address and the auth portion of the state machine verifying that the signature is that of the `GetSigners`

there might be some missing context here for how the cosmos-sdk works, likely faster to sync as somethings are not intuitive

x/fibre/spec.md
evan-forbes: > Ideally there should be another message to query diff from previous requested state of ValSet.

doing this is actually a very large change, as we'd need to enforce old copies are kept in state or that all rpc nodes have past copies of the state.

naively, optimizing for bandwidth atm seems like a premature optimization since these nodes will be uploading and downloading large amount of data to and from validators. imo, it seems like the diff can be managed effectively by re-querying for specific unknown validator's FibreProviderInfo as they join the set since the light client is already tracking the validator set to verify the header, or simply by re-querying this endpoint to get the latest set.

x/fibre/spec.md
evan-forbes: to hopefully make this simpler, I've switched to entirely using the validator consensus address (instead of their operator address) in the state and in this message that updates the state. This is the address of the consensus key, which can be derived by light client's from the validator's public key, which they should have already.

They should also already have voting power etc from the validator set

evan-forbes: if there is still confusion I'm definitely happy to sync!

x/fibre/spec.md
```diff
@@ -0,0 +1,209 @@
+# Fibre Module Specification
+
+## Abstract
+
+The `x/fibre` module enables validators in the active set to register and manage their fibre service provider information.
```
rootulp, line 5: > in the active set

Why only the active set? It likely doesn't happen often but an active validator can be replaced by a bonded validator. If the bonded validator can't register their fibre service provider information ahead of time then they won't service Fibre requests when they become active and there will be some downtime before they register.

x/fibre/spec.md
```diff
@@ -0,0 +1,209 @@
...
+
+```protobuf
+message MsgSetFibreProviderInfo {
+  // ip_addresses are the IP addresses for the fibre service (max 5 addresses, each ≤ 90 characters)
+  repeated string ip_addresses = 1;
+}
+```
+
+**Validation Rules:**
+- `validator_address` must be a valid validator consensus address
```
rootulp, line 60: [question] this isn't a field in the protobuf message above. Should it be?

rootulp (approved):
LGTM, great work

x/fibre/spec.md
evan-forbes: yeah good point this seems like a premature optimization for state growth 

we can just get rid of that requirement and the deletion mechanism for now in all likely hood

x/fibre/spec.md
```diff
@@ -0,0 +1,209 @@
...
+message MsgRemoveFibreProviderInfo {
+  // validator_consensus_address is the consensus address of the validator to remove
+  string validator_consensus_address = 1;
+}
+```
+
+**Validation Rules:**
+- `validator_consensus_address` must be a valid validator consensus address
+- Validator must NOT be in the active set
+- Provider info must exist for the validator
```
rootulp, line 79: One idea is to amend the validation rules to allow only the validator to delete their fibre provider info and allow it to be submitted even if the validator is no longer active.

```suggestion
**Validation Rules:**
- `validator_consensus_address` must be a valid validator consensus address
- Signer must match the validator operator address for the matching consensus validator address
- Provider info must exist for the validator
```

cmwaters:
Non urgent: Can we move this to the spec repo

cmwaters:
Has been moved here: https://github.com/celestiaorg/fibre-da-spec/pull/4

I think we can close it here

Issue #5587, opened 2025-08-22
Use releases of cosmos-sdk and ibc-go


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/blob/1bade4f79e6e5de9d5a241d4c47dbe38b3b6ab86/go.mod#L275-L276

## Problem

We should use releases of these dependencies on `main` and especially before we create v6.0.0-rc0

rootulp:
This is resolved on main and v6.0.0-rc0.

- https://github.com/celestiaorg/celestia-app/blob/b90a2e2227c23f3acd261e204983cf343363e2b6/go.mod#L275-L276
- https://github.com/celestiaorg/celestia-app/blob/7dc5b1a44f76e0ac82f596f8b35e3a314de81d77/go.mod#L275-L276

PR #5611, opened 2025-08-26
chore: bump ibc to official tag

evan-forbes:
## Overview

bumps to official ibc tag

part of #5587


rootulp (approved):
Worked for me locally

PR #5613, opened 2025-08-27
feat: cache CheckTx validation to skip commitment generation in ProcessProposal

### LINKED ISSUE #5607: Cache txs in CheckTx
## Description

Cache transactions after they were checked in `CheckTx` so we don't have to run tx validation logic in `ProcessProposal` again. This should save ~500ms.

ninabarbakadze:
## Overview

Fixes #5607


app/tx_cache.go
```diff
@@ -0,0 +1,74 @@
...
+
+// NewTxValidationCache creates a new transaction validation cache
+func NewTxValidationCache(ttl time.Duration) *TxValidationCache {
+	return &TxValidationCache{
+		cache: make(map[string]*bool),
+	}
+}
+
+// getTxKey generates a deterministic key for a transaction
+func (c *TxValidationCache) getTxKey(tx []byte) string {
```
ninabarbakadze: I wonder if hashing would be too expensive in the long term. We could pass the keys from comet since I think we already hash them for the mempool anyway.

app/app.go
```diff
@@ -220,11 +224,12 @@ func New(
...
-		keys:          keys,
-		tkeys:         tkeys,
-		memKeys:       memKeys,
-		timeoutCommit: timeoutCommit,
+		BaseApp:           baseApp,
+		keys:              keys,
+		tkeys:             tkeys,
+		memKeys:           memKeys,
+		timeoutCommit:     timeoutCommit,
+		txValidationCache: NewTxValidationCache(5 * time.Minute), // Cache for 5 minutes
```
rach-id: what do you think of setting the TTL equal to that of the mempool?

app/process_proposal.go
```diff
@@ -110,15 +115,26 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
+		// commitment verification since it was already validated. Otherwise, fall back to full validation.
 		// - there is one PFB
 		// - that each blob has a valid namespace
 		// - that the sizes match
 		// - that the namespaces match between blob and PFB
 		// - that the share commitment is correct
-		if err := blobtypes.ValidateBlobTx(app.encodingConfig.TxConfig, blobTx, appconsts.SubtreeRootThreshold, appconsts.Version); err != nil {
-			logInvalidPropBlockError(app.Logger(), blockHeader, fmt.Sprintf("invalid blob tx %d", idx), err)
-			return reject(), nil
+		if valid := app.txValidationCache.Get(rawTx); valid && isBlobTx {
```
rach-id: since this is only caching the share commitment validation, should we rename the cache to `ComitmentValidationCache` or similar?

app/process_proposal.go
```diff
@@ -25,6 +25,11 @@ const rejectedPropBlockLog = "Rejected proposal block:"
 
 func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcessProposal) (resp *abci.ResponseProcessProposal, err error) {
 	defer telemetry.MeasureSince(time.Now(), "process_proposal")
+
+	// Clear the cache after ProcessProposal completes (whether accept or reject)
+	// to prevent memory growth and ensure fresh cache for the next block proposal
+	defer app.txValidationCache.Clear()
```
rach-id: I don't think we should be clearing in process proposal. If a block makes it to process proposal, it doesn't mean it will be finalized. So I think we should clear the cache in the finalizeBlock and only for transactions that were included.

app/tx_cache.go
```diff
@@ -0,0 +1,74 @@
...
+	"crypto/sha256"
+	"sync"
+	"time"
+)
+
+// TxValidationCache caches the results of expensive transaction validation
+// to avoid repeating the same work in ProcessProposal that was already done in CheckTx.
+type TxValidationCache struct {
+	mu    sync.RWMutex
+	cache map[string]*bool
```
rach-id: why is this using a pointer to bool?

app/tx_cache.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Clear removes all entries from the cache
+func (c *TxValidationCache) Clear() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.cache = make(map[string]*bool)
+}
+
+// Cleanup removes expired entries from the cache
+func (c *TxValidationCache) Cleanup() {
```
rach-id: I think we should add a method that only deletes certain keys. Also, I think it would be useful to benchmark how much time it takes to cleanup ~500 entry from the map. This would give us an idea of the added cost when running finalize block. We might even want to run the cleanup in parallel to finalize block.

Also, given the cleanup will take a certain period of time, it might block CheckTx and have an impact on the mempool since checkTx will take way longer during the deletion phase. I'm not sure which is the best solution but a benchmark of how much time it takes to delete entries would be useful

app/app.go
ninabarbakadze: I was just playing around with it. In the end I didn't feel like having a TTL made sense so I removed it. Do you think we should reconsider?

app/tx_cache.go
ninabarbakadze: previously I had a struct so it's leftover. will remove.

app/tx_cache.go
ninabarbakadze: I'm not sure I understand why we'd not completely clear the cache after `finalizeBlock`?

tzdybal: I think because we might not include all txs in a block and want the cache for next block.

I would love to see some benchmarks too.

app/tx_cache.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"crypto/sha256"
+	"sync"
+)
+
+// TxValidationCache caches the results of expensive transaction validation
+// to avoid repeating the same work in ProcessProposal that was already done in CheckTx.
+type TxValidationCache struct {
+	mu    sync.RWMutex
+	cache map[string]*bool
```
tzdybal: This cache seems like a perfect candidate for `sync.Map` to me

app/tx_cache.go
ninabarbakadze: > I think because we might not include all txs in a block and want the cache for next block.

Apologies for being off with this. Somehow I was thinking about this as process proposal specific state where in reality it's basically the mempool state.

app/app.go
```diff
@@ -526,6 +529,23 @@ func (app *App) Info(req *abci.RequestInfo) (*abci.ResponseInfo, error) {
...
+// FinalizeBlock implements the abci interface. It overrides baseapp's FinalizeBlock method, essentially becoming a decorator
+// in order to add transaction pruning logic after normal finalize block processing.
+func (app *App) FinalizeBlock(req *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
+	// Call the normal BaseApp FinalizeBlock first
+	res, err := app.BaseApp.FinalizeBlock(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// go through all the transactions that are getting executed and prune the tx tracker
```
ninabarbakadze:
```suggestion
	// Go through all the transactions that are getting executed and prune the tx tracker
```

app/app.go
```diff
@@ -191,9 +191,11 @@ type App struct {
 	BasicManager  module.BasicManager
 	ModuleManager *module.Manager
 	configurator  module.Configurator
-	// blockTime is used to override the default TimeoutHeightDelay. This is
-	// useful for testing purposes and should not be used on public networks
-	// (Arabica, Mocha, or Mainnet Beta).
+
```
ninabarbakadze:
```suggestion
```

app/app.go
```diff
@@ -191,9 +191,11 @@ type App struct {
...
+	// txValidationCache caches expensive blob transaction validation results
+	// from CheckTx to be reused in ProcessProposal.
+	txValidationCache *TxValidationCache
+
```
ninabarbakadze:
```suggestion
```

app/tx_cache_bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+		cache.Set(tx)
+	}
+}
+
+// BenchmarkTxValidationCache_Operations benchmarks cache operations
+func BenchmarkTxValidationCache_Operations(b *testing.B) {
+	// this is the avg size of `Tx` part of `blobTx` which consists commitments from blob shares,
+	// we hash it to use as key for transaction cache,
+	// I got after checking 30000 blocks,the max was 4450
+	txSize := 1130
```
ninabarbakadze, line 31: how did you choose this size?

app/validate_blob_tx_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/celestiaorg/go-square/v3/share"
+	"github.com/celestiaorg/go-square/v3/tx"
+	abci "github.com/cometbft/cometbft/abci/types"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateBlobTxWithCache(t *testing.T) {
```
ninabarbakadze: was there not an existing test suite for testing blob validation? I was thinking we could add to it maybe?

app/app.go
```diff
@@ -191,9 +191,11 @@ type App struct {
 	BasicManager  module.BasicManager
 	ModuleManager *module.Manager
 	configurator  module.Configurator
-	// blockTime is used to override the default TimeoutHeightDelay. This is
-	// useful for testing purposes and should not be used on public networks
-	// (Arabica, Mocha, or Mainnet Beta).
+
+	// txValidationCache caches expensive blob transaction validation results
```
rach-id: [nit]
expensive might be misleading here with the fees
```suggestion
	// txValidationCache caches blob transaction validation results
```

app/app.go
```diff
@@ -526,6 +529,23 @@ func (app *App) Info(req *abci.RequestInfo) (*abci.ResponseInfo, error) {
...
+	// Call the normal BaseApp FinalizeBlock first
+	res, err := app.BaseApp.FinalizeBlock(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// go through all the transactions that are getting executed and prune the tx tracker
+	for _, tx := range req.Txs {
+		app.txValidationCache.RemoveTransaction(tx)
+	}
```
rach-id, line 541: [non blocking]
would it make sense to run these 2 parallel? just in case removing the transactions from the cache is heavy. Otherwise it's fine

app/tx_cache_bench_test.go
rach-id, line 31: Can we do 8mb transactions in 128mb block? we're optimising for the highest load, and mainnet is currently very under utilised

app/tx_cache_bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, tx := range txs {
+		cache.Set(tx)
+	}
+}
+
+// BenchmarkTxValidationCache_Operations benchmarks cache operations
+func BenchmarkTxValidationCache_Operations(b *testing.B) {
+	// this is the avg size of `Tx` part of `blobTx` which consists commitments from blob shares,
+	// we hash it to use as key for transaction cache,
+	// I got after checking 30000 blocks,the max was 4450
```
rach-id: this comment is good for reviewers but I guess we can delete when we want to merge:

```suggestion
```

ninabarbakadze (commented):
I can't approve since I'm the author but implementation looks sound to me 👍🏼

app/tx_cache_bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+// populateCache fills the cache with the given transactions
+func populateCache(cache *TxValidationCache, txs [][]byte) {
+	for _, tx := range txs {
+		cache.Set(tx)
+	}
+}
+
+// BenchmarkTxValidationCache_Operations benchmarks cache operations
+func BenchmarkTxValidationCache_Operations(b *testing.B) {
+	// this is the avg size of `Tx` part of `blobTx` which consists commitments from blob shares,
```
rach-id:
```suggestion
	// this is the avg size of `Tx` part of `blobTx` which consists of commitments from blob shares,
```

app/tx_cache_bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+	// we hash it to use as key for transaction cache,
+	// I got after checking 30000 blocks,the max was 4450
+	txSize := 1130
+	testCases := []struct {
+		name   string
+		numTxs int
+	}{
+		// this is the max num I got after checking 30000 blocks
```
rach-id:
```suggestion
```

rach-id (dismissed):
LGTM

left some nits

app/validate_blob_tx_test.go
mcrakhman: There was, but because the ValidateBlobTxWithCache uses the app I thought that it would be ok to have it there. But now I am also thinking if we could somehow make it one place. Lemme get back to you

app/tx_cache_bench_test.go
mcrakhman: basically I checked 30000 real blocks from mainnet and took the average from the Tx part (which is the commitment) - see comment above the txSize

mcrakhman: This is a "header" size where the pfb messages are marshaled, the blobs are saved separately

ninabarbakadze (commented):
on another thought: do we have talis experiment results for this? @rach-id @mcrakhman

app/app.go
```diff
@@ -526,6 +529,23 @@ func (app *App) Info(req *abci.RequestInfo) (*abci.ResponseInfo, error) {
 	return res, nil
 }
 
+// FinalizeBlock implements the abci interface. It overrides baseapp's FinalizeBlock method, essentially becoming a decorator
+// in order to add transaction pruning logic after normal finalize block processing.
+func (app *App) FinalizeBlock(req *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
```
rootulp, line 531: [no change needed] I'm surprised there wasn't already a FinalizeBlock method on app 🤔

app/tx_cache.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+// getTxKey generates a deterministic key for a transaction
+func (c *TxValidationCache) getTxKey(tx []byte) string {
+	hash := sha256.Sum256(tx)
+	return string(hash[:])
+}
+
+// Exists retrieves a validation result from the cache
```
rootulp:
```suggestion
// Exists retrieves whether the tx exists in the cache
```

app/check_tx.go
```diff
@@ -54,6 +54,8 @@ func (app *App) CheckTx(req *abci.RequestCheckTx) (*abci.ResponseCheckTx, error)
 		if err != nil {
 			return responseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false), err
 		}
+		// Cache the validation result for ProcessProposal to reuse
+		app.txValidationCache.Set(btx.Tx)
```
rootulp: [question] how does this save the validation result? It looks like it just saves the tx and not the result of ValidateBlobTx.

Perhaps the tx validation cache only stores keys of txs that are valid? If yes, can that be clarified in a comment on the tx validation cache?

app/tx_cache.go
```diff
@@ -0,0 +1,55 @@
...
+	key := c.getTxKey(tx)
+	_, exists = c.cache.Load(key)
+	return exists
+}
+
+// Set stores a validation result in the cache
+func (c *TxValidationCache) Set(tx []byte) {
+	key := c.getTxKey(tx)
+	c.cache.Store(key, struct{}{})
+}
```
rootulp: [question] why isn't the value in the map a `bool` or some other type that indicates the result of validation?

app/tx_cache_bench_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+// BenchmarkTxValidationCache_Operations benchmarks cache operations
+func BenchmarkTxValidationCache_Operations(b *testing.B) {
+	// this is the avg size of `Tx` part of `blobTx` which consists commitments from blob shares,
+	// we hash it to use as key for transaction cache,
+	// I got after checking 30000 blocks,the max was 4450
+	txSize := 1130
+	testCases := []struct {
+		name   string
+		numTxs int
```
rootulp: [nit] this could clarify that it is the number of blob txs. This suggestion can't be applied directly because it implies a rename in the test.

```suggestion
		numBlobTxs int
```

app/tx_cache_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+func TestTxValidationCache_RemoveTransactionNonExistent(t *testing.T) {
+	cache := NewTxValidationCache()
+	tx := []byte("tx1")
+	nonExistentTx := []byte("non existent")
+
+	cache.Set(tx)
+	assert.Equal(t, 1, cache.Size())
+
+	cache.RemoveTransaction(nonExistentTx)
```
rootulp, line 101: [no change needed] I would expect this to return an error and then the caller can decide whether to propagate the error, log a warning, or ignore it entirely.

app/validate_blob_tx_test.go
```diff
@@ -0,0 +1,146 @@
...
+		require.True(t, isBlobTx)
+
+		resp, err := testApp.CheckTx(&abci.RequestCheckTx{
+			Type: abci.CheckTxType_New,
+			Tx:   blobTxBytes,
+		})
+		require.NoError(t, err)
+		require.Equal(t, abci.CodeTypeOK, resp.Code)
+
+		fromCache, err := testApp.ValidateBlobTxWithCache(blobTx)
```
rootulp: [no change needed] ahh this is why ValidateBlobTxWithCache returns two arguments. I was wondering why `fromCache` was returned as the first argument when it isn't used in production code but it seems used in tests.

rach-id:
I can run something later

rootulp (approved):
LGTM. My only question is for the tx validation cache which seems to be a map of tx keys that have successfully passed validation because the values are `struct{}{}`. If so, perhaps the cache can have a comment describing that keys in the map are valid and keys that aren't present are either invalid or haven't been validated yet.

app/app.go
ninabarbakadze: Me too! We were just using the default SDK one

app/tx_cache.go
ninabarbakadze: AFAIK this way is more memory efficient. If validation is successful -> add tx in the cache.

mcrakhman: I think it is better not to save the validation result. My reasoning was the following, if the tx is invalid we will return the error, but not save in cache. So in theory the tx can be resubmitted again with correct data on process proposal for example. Even if it is not the case validation of the tx now is relatively fast due to perf optimisations, so it is ok to only cache the happy path.

app/check_tx.go
mcrakhman: Yes, we only store if the tx is valid, that is correct, I will indicate this in comments, thanks

app/validate_blob_tx_test.go
mcrakhman: Yes, exactly, this is needed for testing, because we can't otherwise know for sure if we used cache or not, and therefore we can't check different business logic

app/tx_cache.go
mcrakhman: Renamed this to just `TxCache`

app/app.go
mcrakhman: in general we have 20 txs in cache, so it should be very fast

app/validate_blob_tx_test.go
mcrakhman: Moved to blob_tx_test.go, thanks

app/app.go
rach-id: why 20? worst case is 200

PR #5616, opened 2025-08-28
Fix(dead link): Removed dead link in `Update adr-018-network-upgrades.md`

letmehateu:
## Overview

There is a dead link in `adr-018-network-upgrades.md`. I removed it and fix a little a text


rootulp (approved):
What you have done is fine. We could have also replaced the broken link with a permalink to x/circuit from a commit where Cosmos SDK still has it.

PR #5619, opened 2025-08-28
refactor: avoid duplicate filepath.Ext call in payforblob validation

dennsikl:
## Overview

- Store filepath.Ext(path) result in 'ext' variable
- Use 'ext' variable in both condition check and error message
- Reduces function calls from 2 to 1 per validation

### DIFF
--- a/x/blob/client/cli/payforblob.go
+++ b/x/blob/client/cli/payforblob.go
@@ -85,8 +85,8 @@ The blob must be a hex encoded string of non-zero length.
 			}
 
 			if path != "" {
-				if filepath.Ext(path) != FileInputExtension {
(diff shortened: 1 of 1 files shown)


rootulp (approved):
IMO safe to close or merge. Not a big difference either way.

ninabarbakadze:
> IMO safe to close or merge. Not a big difference either way.

Closing as it makes no significant difference and is not worth trying to get CI to pass. 

@dennsikl If you'd like to make a meaningful contribution please take a look at good first issues https://github.com/celestiaorg/celestia-app/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

PR #5626, opened 2025-08-28
chore: parallel validate blob txs in processProposal

evan-forbes:
## Overview

we'll need this even after cacheing, as there is the case where the node didn't download the txs in the mempool, therefore wouldn't have the txs cached.

meaningful decrease as well, although not perfect. 8 CPU (16 threads) results in an 5-6x speedup for relatively full blocks

```
BenchmarkValidateBlobTxsComparison/10_tx_8MiB_sequential-16            	       6	 166898472 ns/op
BenchmarkValidateBlobTxsComparison/10_tx_8MiB_parallel-16              	      37	  32007726 ns/op
BenchmarkValidateBlobTxsComparison/100_tx_1MiB_sequential-16           	      10	 106334685 ns/op
BenchmarkValidateBlobTxsComparison/100_tx_1MiB_parallel-16             	      54	  19500342 ns/op
```

### DIFF
--- a/x/blob/types/blob_tx.go
+++ b/x/blob/types/blob_tx.go
@@ -2,7 +2,7 @@ package types
 
 import (
 	"bytes"
-	context "context"
+	"context"
 	"runtime"
 	"slices"
 


x/blob/types/blob_tx_test.go
```diff
@@ -260,3 +262,278 @@ func TestValidateBlobTx(t *testing.T) {
...
+				assert.NoError(t, parallelErr, "parallel validation should not error")
+			}
+
+			if (serialErr == nil) != (parallelErr == nil) {
+				t.Errorf("validation results differ: serial error = %v, parallel error = %v", serialErr, parallelErr)
+			}
+
+			if serialErr != nil && parallelErr != nil {
+				assert.Equal(t, serialErr.Error(), parallelErr.Error(), "error messages should match")
+			}
```
evan-forbes, line 398: this is the main test that makes me feel comfortable shipping this beyond just theoretically it should be fine

evan-forbes:
cherry picked https://github.com/celestiaorg/celestia-core/pull/2379/commits/d8f4f412159ac90137ac0c887d50795480361a07 to fix the linter

rach-id:
@evan-forbes just to confirm, are you willing to continue this or  someone can pick it up?

evan-forbes:
ahh sorry missed this @rach-id 

would appreciate someone taking it over, but iirc this was mostly good to go

rach-id:
Cool, lets check whether we need it anyway given the parallelisation we're doing when calculating the commitment

evan-forbes:
I forget our decision here, iirc we thought about closing this since we are already verifying the commitment in parallel?

rach-id:
> we thought about closing this since we are already verifying the commitment in parallel?

since the work is almost already done, we could run a benchmark and see if this provides any additional benefits. Otherwise, yes, we can close

evan-forbes:
I'll just close for now but if we time everything again and see this is still taking a while we can reopen

PR #5635, opened 2025-09-01
docs: improve comment about sequence behavior in tx tracker test

ANtutov:
Fix misleading comment in assertTxInTxTracker function.

The comment incorrectly stated "sequence does not increase" but the test
actually verifies that sequence is properly incremented after broadcast.


rootulp (approved):

rootulp:
Has a merge conflict.

PR #5639, opened 2025-09-03
chore!: bump core v0.39.3

evan-forbes:
bumps core to use new dynamic timeouts, prevent panics, performance improvements, and bug fixes


pkg/user/tx_client_test.go
```diff
@@ -302,6 +305,15 @@ func TestEvictions(t *testing.T) {
...
 		res, err := txClient.ConfirmTx(ctx.GoContext(), resp.TxHash)
+		// the txs can get rejected if one is evicted with execution code 32 (sequence mismatch). Therefore,
+		// we need to skip those.
+		if err != nil {
+			if strings.Contains(err.Error(), "rejected") {
+				continue
+			}
+			// purposefully fail as this is an unexpected error
+			require.NoError(t, err)
+		}
```
evan-forbes, line 314: please review @ninabarbakadze 

I kept hitting flakes with this test. this assumption seemed wrong, but I might be missing something. it seems like we can get a rejected tx, and therefore an error, if one of the txs gets evicted but one in the future doesn't. If it doesn't then the sequence will not match.

pkg/appconsts/app_consts.go
```diff
@@ -19,12 +19,18 @@ const (
...
+	GasPerBlobByte        uint32 = 8
+	MaxTxSize             int    = 8_388_608 // 8 MiB in bytes
+	TimeoutPropose               = time.Millisecond * 5000
+	TimeoutProposeDelta          = time.Millisecond * 2000
+	TimeoutPrevote               = time.Millisecond * 1000
+	TimeoutPrevoteDelta          = time.Millisecond * 5000
+	TimeoutPrecommit             = time.Millisecond * 1000
+	TimeoutPrecommitDelta        = time.Millisecond * 5000
+	TimeoutCommit                = time.Millisecond * 200
+	TimeoutHeightDelay           = time.Millisecond * 6000
```
evan-forbes: @rach-id what timeouts do we want here for 6s 128MB and 32MB blocks?

pkg/user/tx_client_test.go
cmwaters: I was seeing the same thing on a different branch that I was working on. I think this has to do with this issue [here](https://github.com/celestiaorg/celestia-app/issues/5655). 

This here basically bypasses the test by allowing it to reject. I think we should do this for now to have the test passing and address this as a FLUP

app/app.go
```diff
@@ -220,11 +220,11 @@ func New(
...
-		BaseApp:       baseApp,
-		keys:          keys,
-		tkeys:         tkeys,
-		memKeys:       memKeys,
-		timeoutCommit: timeoutCommit,
+		BaseApp:   baseApp,
+		keys:      keys,
+		tkeys:     tkeys,
+		memKeys:   memKeys,
+		blockTime: timeoutCommit,
```
cmwaters: Can we rename the timeoutCommit in the constructor as well (for consistency)

app/app.go
```diff
@@ -190,10 +190,10 @@ type App struct {
 	BasicManager  module.BasicManager
 	ModuleManager *module.Manager
 	configurator  module.Configurator
-	// timeoutCommit is used to override the default timeoutCommit. This is
+	// blockTime is used to override the default TimeoutHeightDelay. This is
```
cmwaters, line 193: Is TimeoutHeightDelay the same as timeout commit in function in that it is the time following from the commit with which to wait before starting the next height?

pkg/appconsts/app_consts.go
cmwaters: What is `TimeoutHeightDelay` used for here? I don't see it called anywhere. What happened to DelayedPrecommitTimeout?

app/app.go
```diff
@@ -811,17 +807,30 @@ func (app *App) NewProposalContext(header tmproto.Header) sdk.Context {
...
+func (app *App) TimeoutInfo() abci.TimeoutInfo {
+	return abci.TimeoutInfo{
+		TimeoutPropose:          appconsts.TimeoutPropose,
+		TimeoutProposeDelta:     appconsts.TimeoutProposeDelta,
+		TimeoutCommit:           appconsts.TimeoutCommit,
+		TimeoutPrevote:          appconsts.TimeoutPrevote,
+		TimeoutPrevoteDelta:     appconsts.TimeoutPrevoteDelta,
+		TimeoutPrecommit:        appconsts.TimeoutPrecommit,
+		TimeoutPrecommitDelta:   appconsts.TimeoutPrecommitDelta,
+		DelayedPrecommitTimeout: app.blockTime,
```
cmwaters: I'm a bit confused here. The DelayedPrecommitTimeout isn't necessary the block time, it's the minimum time before nodes precommit

cmwaters (commented):
I am a little confused with the different terminology around timing. Everything else looks good but I think this could be cleaned up

pkg/appconsts/app_consts.go
rach-id: for 32MB blocks, I think we can be fine with:
- propose: 5s
- propose delta: 1s. should be enough given we propose 128mb blocks in ~2s
- prevote: 1.2s. process proposal should take ~0.5s at most. Then we can leave the rest to receive the votes in time
- prevote delta: 500ms
- precommit: 1s. just to get the precommit votes, nothing more
- precommit delta: 500ms 
- commit: could be set even to 1ns. We could set it for now to 100ms or 200ms
- timeout height delay: I think we should set it to 5.5s instead of 6. This is because the block time = timeout height delay + (finalize block + save block and other small operations which would take ~0.5s). So 5.5 + 0.5 = ~6s block time.

For 128mb block do we need them now? or we can leave it until we're done with the optimisations?

rach-id: Also, should I run a 32mb network with 95 validator test to see if these timeouts work?

cmwaters: Would the 128MB timeouts be different than the 32MB timeouts?

pkg/user/tx_client_test.go
evan-forbes: >This here basically bypasses the test by allowing it to reject

out of  curiousity, could the scenario that I described occur?

> it seems like we can get a rejected tx, and therefore an error, if one of the txs gets evicted but one in the future doesn't

pkg/appconsts/app_consts.go
evan-forbes: @rach-id and I spoke, and now we think we should error on the side of caution with the timeouts now that we have the delayed commits, in combination with keeping the delta's small.

so long timeout propose (8-9s) and prevote/precommit (3s) and roughly 6s for the delayprecommit. 

When the network is not healthy, we will have longer block times, however fewer multiple rounds due to missing the proposal. When the network is healthy, we'll close to 6s blocks. When a validator is down, we will have a longer wait for the propose, but we won't wait for the delay for the second round, so just produce a block as fast as possible.

seems like a great set of tradeoffs imo but curious what @cmwaters thinks

app/app.go
evan-forbes: good point, at once point I think we used the term "target block time", would that work here? 

I can also just use the DelayedPrecommitTimeout.

pkg/user/tx_client_test.go
cmwaters: hmm I would expect a FIFO like behaviour to evicting so I wouldn't expect a later transaction to be evicted before an earlier one (which would then cause the mismatch)

My theory is there's an inherent incongruence with priority and sequences which especially comes out when you have multiple transactions from the same signer in a block.

app/app.go
cmwaters: If possible, I would prefer the same term used throughout

pkg/appconsts/app_consts.go
cmwaters: The tradeoff iiuc is tighter timing but more frequent multi-round blocks vs relaxed timing and less frequent multi-round blocks.

I'm not sure on the right balance so I'll let you guys decide. I think having multiple rounds occur occasionally isn't necessary the worst because it may highlight which operators are skimping out on the required resources. But generally I'd want to make sure its not too frequent. I think 1 in 1000 is acceptable

app/app.go
evan-forbes: cool sounds good did in https://github.com/celestiaorg/celestia-app/pull/5639/commits/b8e2364290b10e071a2070063c6d870a63eb5bf9

I think the reasoning to try and change the names here was misguided, and instead I should have tried to push for renaming that constant in core. 

1) all other timeouts use a different format `Timeout...` instead of `...Timeout`
2) We already have TimeoutCommit, TimeoutPrecommit, TimeoutPrecommitDelta. If we added TimeoutDelayedPrecommit, that would be okay, but idk how someone not deeply familar with the changes is supposed to understand that they need to change this time in order to adjust the block time

app/app.go
evan-forbes: see other comment, done in https://github.com/celestiaorg/celestia-app/pull/5639/commits/b8e2364290b10e071a2070063c6d870a63eb5bf9

app/app.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5639/commits/b8e2364290b10e071a2070063c6d870a63eb5bf9

app/app.go
cmwaters: it could have been `MinimumPrecommitTime`. As Rachid alluded to, this is a proxy to block time, the actual time is 0.5s slower on average. I think `TargetBlockTime` is also acceptable. I brought this up mainly because now we are using different named variables to mean the same thing

pkg/appconsts/app_consts.go
```diff
@@ -19,12 +19,18 @@ const (
...
+	GasPerBlobByte          uint32 = 8
+	MaxTxSize               int    = 8_388_608 // 8 MiB in bytes
+	TimeoutPropose                 = time.Millisecond * 8500
+	TimeoutProposeDelta            = time.Millisecond * 500
+	TimeoutPrevote                 = time.Millisecond * 3000
+	TimeoutPrevoteDelta            = time.Millisecond * 500
+	TimeoutPrecommit               = time.Millisecond * 3000
+	TimeoutPrecommitDelta          = time.Millisecond * 500
+	TimeoutCommit                  = time.Millisecond * 100
+	DelayedPrecommitTimeout        = time.Millisecond * 5200
```
evan-forbes: pls see new times for final review @rach-id and @cmwaters 

we can adjust these up until the mocha release, @rach-id was going to run a few more experiments to get estimates

I'm less sure on the 5200, but it would align w/ current block times for empty blocks while allowing the max amount of time

evan-forbes:
okay, should finally be ready for final reivew

app/app.go
rootulp: I support renaming `app.blockTime` to `app.delayedPrecommitTimeout` for consistency.

app/test/square_size_test.go
```diff
@@ -48,8 +48,8 @@ func (s *SquareSizeIntegrationTest) SetupSuite() {
 	s.enc = encoding.MakeConfig(app.ModuleEncodingRegisters...)
 	cfg := testnode.DefaultConfig().
 		WithModifiers(genesis.ImmediateProposals(s.enc.Codec)).
-		WithTimeoutCommit(time.Millisecond * 500). // long timeout commit to provide time for submitting txs
-		WithFundedAccounts("txsim")                // add a specific txsim account
+		WithDelayedPrecommitTimeout(time.Millisecond * 500). // long timeout commit to provide time for submitting txs
```
rootulp:
```suggestion
		WithDelayedPrecommitTimeout(time.Millisecond * 500). // long time to provide time for submitting txs
```

cmd/celestia-appd/cmd/root.go
```diff
@@ -34,7 +34,11 @@ const (
 	UpgradeHeightFlag = "v2-upgrade-height"
 
 	// TimeoutCommit is a flag that can be used to override the timeout_commit.
+	// Deprecated: Use BlockTimeFlag instead.
```
rootulp:
```suggestion
	// Deprecated: Use DelayedPrecommitTimeoutFlag instead.
```

cmd/celestia-appd/cmd/root.go
```diff
@@ -144,6 +148,11 @@ func addStartFlags(startCmd *cobra.Command) {
 	}
 
 	startCmd.Flags().Duration(TimeoutCommitFlag, 0, "Override the application configured timeout_commit. Note: only for testing purposes.")
+	if err := startCmd.Flags().MarkDeprecated(TimeoutCommitFlag, "Use --block-time instead."); err != nil {
```
rootulp:
```suggestion
	if err := startCmd.Flags().MarkDeprecated(TimeoutCommitFlag, "Use --delayed-precommit-timeout instead."); err != nil {
```

go.mod
```diff
@@ -271,7 +271,7 @@ replace (
 	cosmossdk.io/api => github.com/celestiaorg/cosmos-sdk/api v0.7.6
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
 	// TODO: update to a v0.39.x release after that is created from celestia-core main branch.
```
rootulp:
```suggestion
```

test/util/testnode/config.go
```diff
@@ -25,7 +25,10 @@ const (
 	DefaultValidatorAccountName = "validator"
 	DefaultInitialBalance       = genesis.DefaultInitialBalance
 	// TimeoutCommit is a flag that can be used to override the timeout_commit.
+	// Deprecated: Use BlockTimeFlag instead.
 	TimeoutCommitFlag = "timeout-commit"
+	// BlockTimeFlag is a flag that can be used to override the DelayedPrecommitTimeout.
+	BlockTimeFlag = "block-time"
```
rootulp: this is different in celestia-app. I think the `block-time` flag was renamed to delayed precommit timeout flag.

rootulp (commented):
LGTM with a few nits. 

1. I think `app.blockTime` can be replaced with `app.delayedPrecommitTimeout`
2. A few instances still use `blockTime` (flags) when I think they meant to use the delayed precommit flag

app/app.go
evan-forbes: ahh sorry missed that one, thanks

cmwaters (dismissed):
Looks good. I would echo Rootul's point of having either block time or delayed precommit time consistently through the code base

cmd/celestia-appd/cmd/root.go
evan-forbes: fixed in https://github.com/celestiaorg/celestia-app/pull/5639/commits/1c8cd007ebd004eb449f2385693a4d550f2049f7

evan-forbes:
> Looks good. I would echo Rootul's point of having either block time or delayed precommit time consistently through the code base

agreed, already in https://github.com/celestiaorg/celestia-app/pull/5639/commits/1c8cd007ebd004eb449f2385693a4d550f2049f7

test/util/testnode/config.go
evan-forbes: good catch https://github.com/celestiaorg/celestia-app/pull/5639/commits/1c8cd007ebd004eb449f2385693a4d550f2049f7

app/app.go
```diff
@@ -190,19 +190,20 @@ type App struct {
 	BasicManager  module.BasicManager
 	ModuleManager *module.Manager
 	configurator  module.Configurator
-	// timeoutCommit is used to override the default timeoutCommit. This is
+	// blockTime is used to override the default TimeoutHeightDelay. This is
```
rootulp, line 193:
```suggestion
	// delayedPrecommitTimeout is used to override the default delayedPrecommitTimeout. This is
```

test/util/testnode/config.go
```diff
@@ -25,7 +25,10 @@ const (
 	DefaultValidatorAccountName = "validator"
 	DefaultInitialBalance       = genesis.DefaultInitialBalance
 	// TimeoutCommit is a flag that can be used to override the timeout_commit.
+	// Deprecated: Use BlockTimeFlag instead.
```
rootulp, line 28:
```suggestion
	// Deprecated: Use DelayedPrecommitTimeout instead.
```

test/util/testnode/config.go
```diff
@@ -84,9 +87,20 @@ func (c *Config) WithSuppressLogs(sl bool) *Config {
 }
 
 // WithTimeoutCommit sets the timeout commit in the cometBFT config and returns
-// the Config.
+// the Config. For backward compatibility, it also sets the app's block time.
+// Deprecated: Use WithBlockTime instead.
```
rootulp, line 91:
```suggestion
// Deprecated: Use WithDelayedPrecommitTimeout instead.
```

test/util/testnode/config.go
```diff
@@ -165,11 +179,19 @@ func DefaultAppCreator(opts ...AppCreationOptions) srvtypes.AppCreator {
 		baseAppOptions := server.DefaultBaseappOptions(appOptions)
 		baseAppOptions = append(baseAppOptions, baseapp.SetMinGasPrices(fmt.Sprintf("%v%v", appconsts.DefaultMinGasPrice, appconsts.BondDenom)))
 
+		// Check for the new --block-time flag first, then fall back to deprecated --timeout-commit
```
rootulp, line 182:
```suggestion
		// Check for the new --delayed-precommit-timeout flag first, then fall back to deprecated --timeout-commit
```

test/util/testnode/config.go
```diff
@@ -165,11 +179,19 @@ func DefaultAppCreator(opts ...AppCreationOptions) srvtypes.AppCreator {
...
+		} else if timeoutCommitFromFlag := appOptions.Get(TimeoutCommitFlag); timeoutCommitFromFlag != nil {
+			blockTime = timeoutCommitFromFlag.(time.Duration)
+		}
+
 		app := app.New(
 			log.NewNopLogger(),
 			dbm.NewMemDB(),
 			nil, // trace store
-			appOptions.Get(TimeoutCommitFlag).(time.Duration), // timeout commit
+			blockTime,
```
rootulp, line 194: `blockTime` => `delayedPrecommitTimeout`

rootulp (approved):

rach-id (approved):
Let's :ship:

PR #5659, opened 2025-09-03
chore: check if a port is available using UDP

rach-id:
## Overview

This makes sure ports collision issues don't happen, or at least happen less often on macOS.

In fact, when we bind to 0 to get a free port:
- For UDP ports, once we close the listener, the port is free immediately.
- In TCP:
    - On macOS, 60s is enforced on that port before anyone can bind to it. So if the node is started within 60s or probably some constant, we're not able to bind to it
     - On Linux, that constant can be different depending on OS and also can be tweaked

So this switches to listening on UDP to make sure we can bind to the port directly after.

### DIFF
--- a/test/util/testnode/utils.go
+++ b/test/util/testnode/utils.go
@@ -135,7 +135,7 @@ func MustGetFreePort() int {
 
 // isPortAvailable checks if a port is available by attempting to listen on it.
 func isPortAvailable(port int) bool {
-	l, err := net.Listen("udp", fmt.Sprintf(":%d", port))
+	l, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
 	if err != nil {
 		return false
 	}


evan-forbes (dismissed):

test/util/testnode/utils.go
```diff
@@ -120,7 +120,7 @@ func GetFreePort() (int, error) {
 
 // isPortAvailable checks if a port is available by attempting to listen on it.
 func isPortAvailable(port int) bool {
```
rach-id: We might even start using this by default and using 0 instead of having that global port this. I guess the issue we had was this and not the race condition that happens between checking if a port is available vs actually using it.

rootulp:
CI is failing

evan-forbes:
>In fact, when we bind to 0 to get a free port:

going off our past experience of using GetFreePort which uses port 0, are we confident that it will fix the issues we kept seeing with flakey tests?

test/util/testnode/utils.go
```diff
@@ -105,17 +105,32 @@ func GenerateAccounts(count int) []string {
...
+	if err != nil {
+		panic(err)
+	}
+
+	l, err := net.ListenUDP("udp", a)
+	if err != nil {
+		panic(err)
+	}
+	defer l.Close()
+	return l.LocalAddr().(*net.UDPAddr).Port
```
cmwaters, line 133: Why can't this just call GetFreePort and panic on error?

rach-id: we could, it's just that using it here: 
image

would mean handling the error 😆  didn't want to add extra `if err != nil` 5 times. 

if you prefer, I can switch to that, I'm fine with any

cmwaters: No I meant that  it looks like this

```
func MustGetFreePort int {
   port, err := GetFreePort()
   if err != nil {
       panic(err)
   }
   return port
}
```

rather than rewriting the function

cmwaters (approved):
utAck

test/util/testnode/utils.go
rach-id: aaah, makes sense. If you want me to change, I can open a PR or ask copilot to do it

PR #5660, opened 2025-09-03
docs(app): Correct comment in nonce mismatch integration test

YangLi0x0:
## Overview

The comment for the nonce mismatch test was misleading. It mentioned the `DeductFeeDecorator` when the test actually uses the `SigVerificationDecorator`.

This commit corrects the comment to accurately reflect the code under test.


YangLi0x0:
> Seems correct b/c the test creates a `NewSigVerificationDecorator`

The comment appears to be a copy-paste error from the app/errors/insufficient_gas_price_test.go , where it is correct

Issue #5614, opened 2025-08-27
Some archive nodes return error on balance query `version mismatch on immutable IAVL tree`


rootulp:
## Context

https://celestia-team.slack.com/archives/C07TJQPUDJP/p1756218104909649

## Problem

QuickNodes has an archive node. When they query a few block heights, they get an error

```shell
$ curl -X GET "https://docs-demo.celestia-mainnet.quiknode.pro/cosmos/bank/v1beta1/balances/celestia1cwhxmw9kaz72nhdsuqvk6ljf9dzes8kg7ysg9d" -H "x-cosmos-block-height: 2748392"
{"code":2, "message":"codespace sdk code 18: invalid request: failed to load state at height 2748392; version mismatch on immutable IAVL tree; version does not exist. Version has either been pruned, or is for a future block height (latest height: 7194538)", "details":[]}%
```

### Details

For block `1,000,000`, all providers return the same response:

```
docs-demo.celestia-mainnet.quiknode.pro
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
api-celestia.alphab.ai
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
api.archive.celestia.cumulo.com.es
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
celestia-mainnet-api.itrocket.net
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
celestia.api.kjnodes.com
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
api-celestia.mzonder.com
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
api.celestia.pops.one
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
celestia-mainnet-api.denodes.xyz
{"balances":[{"denom":"utia","amount":"28312"}],"pagination":{"next_key":null,"total":"1"}}
```

For block `2,748,392`, some providers return an error:

```
docs-demo.celestia-mainnet.quiknode.pro
[...]

PR #5661, opened 2025-09-03
debug: attempt to reproduce archive query bug

rootulp:
Attempt to reproduce https://github.com/celestiaorg/celestia-app/issues/5614


rootulp:
So far I can query all historical blocks on old app versions so I haven't reproduced it yet.

Issue #4945, opened 2025-06-06
[zk-execution-ism] Validate public inputs against trusted state


damiannolan:
See https://github.com/celestiaorg/celestia-app/pull/4922#discussion_r2128016892. 

The public inputs which we receive from the relayer and decode as part of ism metadata should be validated against trusted state.

The code currently checks the trusted state root, trusted height, namespace and pubkey:

```golang
func (ism *ZKExecutionISM) validatePublicInputs(inputs PublicInputs) error {
	if !bytes.Equal(inputs.TrustedStateRoot[:], ism.StateRoot) {
		return fmt.Errorf("cannot trust public inputs trusted state root: expected %x, but got %x", ism.StateRoot, inputs.TrustedStateRoot)
	}

	if inputs.TrustedHeight != ism.Height {
		return fmt.Errorf("cannot trust public inputs trusted height: expected %d, but got %d", ism.Height, inputs.TrustedHeight)
	}

	if !bytes.Equal(inputs.Namespace[:], ism.Namespace) {
		return fmt.Errorf("cannot trust public inputs namespace: expected %x, but got %x", ism.Namespace, inputs.Namespace)
	}

	if !bytes.Equal(inputs.PublicKey[:], ism.PublicKey) {
		return fmt.Errorf("cannot trust public inputs public key: expected %x, but got %x", ism.PublicKey, inputs.PublicKey)
	}

	return nil
}
```

We _must_ also validate the celestia header hash. We commit to the latest celestia header hash in the outputs of the aggregation/range circuit, and have chained our way up to this. 
We can access a trusted header hash which is available through the x/staking keeper in `HistoricalInfo`.
We lookup the historical info for a given height, and access the header. We hash it and compare it against the value from the proof values.

damiannolan:
This is a little nuanced as we are slightly limited by the API of the ISM via Hyperlane:

```golang
func (ism ZKExecutionISM) Verify(ctx context.Context, metadata []byte, message hyperlane.Message)
```

Essentially what we need to do in order to validate the proof value against the trusted celestia header is the following, where we have direct access to `HistoricalInfo` via the staking keeper, to lookup a trusted celestia header hash (on-chain) for a given height.

```diff
func (k *Keeper) Verify(ctx context.Context, ismId util.HexAddress, metadata []byte, message util.HyperlaneMessage) (bool, error) {
	zkExecutionISM, err := k.isms.Get(ctx, ismId.GetInternalId())
	if err != nil {
[...]

PR #5662, opened 2025-09-04
feat: validate zk proof public values against trusted state

damiannolan:
## Overview

closes: #4945 

This PR deviates from the initial plan to consume `HistoricalInfo` headers from x/staking and hash them on the fly, but due to the headers stored in x/staking containing only a subset of their original values (see: https://github.com/cosmos/cosmos-sdk/issues/25316), this means we need to track headers hashes directly within this module.

Header hashes are tracked via `BeginBlock` in the x/zkism module. The module maintains a limit of header hashes (`MaxHeaderHashes`) and prunes accordingly (defaulting to 50k header hashes initially, this can be changed as we please). 

The parameter `MaxHeaderHashes` can be updated using the module authority (default x/gov)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -564,7 +564,7 @@ func (app *App) EndBlocker(ctx sdk.Context) (sdk.EndBlock, error) {
 				return sdk.EndBlock{}, fmt.Errorf("failed to schedule upgrade: %v", err)
 			}
 
-			if err := app.UpgradeKeeper.DumpUpgradeInfoToDisk(upgrade.UpgradeHeight+1, plan); err != nil {
+			if err := app.UpgradeKeeper.DumpUpgradeInfoToDisk(plan.Height, plan); err != nil {
 				return sdk.EndBlock{}, fmt.Errorf("failed to dump upgrade info to disk: %v", err)
 			}
 
--- a/x/zkism/keeper/abci.go
+++ b/x/zkism/keeper/abci.go
@@ -25,7 +25,7 @@ func (k *Keeper) StoreHeaderHash(goCtx context.Context) error {
 		return err
 	}
 
-	// Prune store to ensure we only have parameter-defined entries.
+	// Prune store to ensure the number of entries in store does not exceed the max header hashes parameter.
 	// In most cases, this will involve removing a single entry.
 	// In the rare scenario when the entries gets reduced to a lower value k'
 	// from the original value k. k - k' entries must be deleted from the store.


x/zkism/keeper/keeper.go
```diff
@@ -45,17 +53,80 @@ func NewKeeper(cdc codec.Codec, storeService corestore.KVStoreService, hyperlane
...
 // Verify implements hyperlane util.InterchainSecurityModule.
 func (k *Keeper) Verify(ctx context.Context, ismId util.HexAddress, metadata []byte, message util.HyperlaneMessage) (bool, error) {
-	zkExecutionISM, err := k.isms.Get(ctx, ismId.GetInternalId())
+	ism, err := k.isms.Get(ctx, ismId.GetInternalId())
 	if err != nil {
 		return false, err
 	}
 
-	return zkExecutionISM.Verify(ctx, metadata, message)
+	meta, err := types.NewZkExecutionISMMetadata(metadata)
```
damiannolan, line 79: this will likely need further change to support different consumers/circuits

x/zkism/keeper/abci.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// Prune store to ensure we only have parameter-defined entries.
+	// In most cases, this will involve removing a single entry.
+	// In the rare scenario when the entries gets reduced to a lower value k'
+	// from the original value k. k - k' entries must be deleted from the store.
+	// Since the entries to be deleted are always in a continuous range, we can iterate
+	// over the historical entries starting from the most recent version to be pruned
+	// and then return at the first empty entry.
+	for i := ctx.BlockHeight() - int64(numEntries); i >= 0; i-- {
```
blasrodri, line 35: should the numEntries be limited? To avoid having a very long loop?

damiannolan: The loop counter is decrementing, and the loop will `break` below when we lookup and get `collections.ErrNotFound`.

We start at current height - `numEntries` so that we always retain `numEntries` worth of header hashes. 

E.g. imagine numEntires is 10000. The current height 10001, we start at 1, lookup and prune at height 1, and then add 10001 below on L53.

As mentioned in the godoc, in most cases we will always be pruning a single entry and then breaking on the next iteration of the loop. Unless the `numEntires` has been lowered, in which case we need to prune `k - k'`

damiannolan: The number of entries is an on-chain parameter, and controlled by validators via governance. I've chosen to maintain 50k header hashes as a reasonable estimate in this PR, but we can decide what works best when this module goes to mainnet via CIP

damiannolan:
With respect to failing e2es. I applied the diffs in https://github.com/celestiaorg/celestia-app/pull/5724 to the v5.x branch of celestia-app and built a standalone docker image which I then applied the following diff to this branch:

```diff
diff --git a/docker/multiplexer.Dockerfile b/docker/multiplexer.Dockerfile
index 205c0a85..dd4e8931 100644
--- a/docker/multiplexer.Dockerfile
+++ b/docker/multiplexer.Dockerfile
@@ -20,7 +20,7 @@ ARG CELESTIA_APP_REPOSITORY=ghcr.io/celestiaorg/celestia-app-standalone
 # Makefile.
 ARG CELESTIA_VERSION_V3="v3.10.6"
 ARG CELESTIA_VERSION_V4="v4.1.0"
-ARG CELESTIA_VERSION_V5="v5.0.4-rc0"
+ARG CELESTIA_VERSION_V5="bb1b85de"
```

And ran the ibc test which performs an upgrade and does a bunch of ibc functionality:
```
make test-docker-e2e test=TestIBC entrypoint=TestIBCTestSuite
```

The tests pass ✅ 
```
--- PASS: TestIBCTestSuite (394.15s)
    --- PASS: TestIBCTestSuite/TestIBC (393.84s)
        --- PASS: TestIBCTestSuite/TestIBC/setup_ibc_infrastructure (166.33s)
        --- PASS: TestIBCTestSuite/TestIBC/start_relayer (2.15s)
        --- PASS: TestIBCTestSuite/TestIBC/initial_transfers (28.03s)
        --- PASS: TestIBCTestSuite/TestIBC/upgrade_app_version (46.19s)
        --- PASS: TestIBCTestSuite/TestIBC/existing_channel_transfers (28.06s)
        --- PASS: TestIBCTestSuite/TestIBC/new_connection_and_channel (64.80s)
        --- PASS: TestIBCTestSuite/TestIBC/new_channel_transfers (26.03s)
PASS
ok  	celestiaorg/celestia-app/test/docker-e2e	394.935s
```

Note, this will require the fix to be backported to have successful tests in CI on this branch.

x/zkism/keeper/abci.go
blasrodri: thanks

x/zkism/keeper/abci.go
```diff
@@ -0,0 +1,54 @@
...
+		_, err := k.GetHeaderHash(ctx, uint64(i))
+		if err != nil {
+			if errors.Is(err, collections.ErrNotFound) {
+				break
+			}
+			return err
+		}
+
+		if err = k.headers.Remove(ctx, uint64(i)); err != nil {
+			return err
```
blasrodri, line 45: what happens in this case? Could there be a case of inconsistent state? Where not all the entries are pruned?
Isn't it better to use `cacheCtx, writeCache := sdkCtx.CacheContext()` and then commit at the end of the loop in case there were no errors?

app/app.go
```diff
@@ -564,7 +564,7 @@ func (app *App) EndBlocker(ctx sdk.Context) (sdk.EndBlock, error) {
 				return sdk.EndBlock{}, fmt.Errorf("failed to schedule upgrade: %v", err)
 			}
 
-			if err := app.UpgradeKeeper.DumpUpgradeInfoToDisk(upgrade.UpgradeHeight, plan); err != nil {
+			if err := app.UpgradeKeeper.DumpUpgradeInfoToDisk(upgrade.UpgradeHeight+1, plan); err != nil {
```
rootulp: [optional] if you think this is clearer

```suggestion
			if err := app.UpgradeKeeper.DumpUpgradeInfoToDisk(plan.Height, plan); err != nil {
```

x/zkism/keeper/abci.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+
+	if numEntries == 0 {
+		return nil
+	}
```
rootulp, line 51: [optional] it seems possible to move this _before_ the `for` loop in order to return from the function body earlier.

x/zkism/keeper/abci.go
```diff
@@ -0,0 +1,54 @@
...
+// StoreHeaderHash stores the header hash keyed by block height and prunes the oldest entry
+func (k *Keeper) StoreHeaderHash(goCtx context.Context) error {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	numEntries, err := k.GetMaxHeaderHashes(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Prune store to ensure we only have parameter-defined entries.
```
rootulp: [optional] re-wording b/c parameter-defined wasn't immediately obvious. After reading the rest of the PR I realized parameter-defined == the parameter for max header hashes

```suggestion
	// Prune store to ensure the number of entries in store does not exceed the max header hashes parameter.
```

x/zkism/keeper/abci.go
rootulp, line 35: [no change needed] I ran through an example to verify this for loop's correctness and I think it seems sound.

1. Set MaxHeaderHashes = 100
2. Assume blocks 1 through 100 are produced and all are stored.
3. Assume block 101 is produced and header hash 1 is pruned.
4. Now set MaxHeaderHashes = 50
5. On the next block height, this loop will iterate from block height (102 - 50 = 52) down to 0 and prune all those hashes. The result is leaving hashes 52 to 102 in state.
6. Now set MaxHeaderHashes = 100
7. On the next block height, this loop will iterate from block height (103 - 100 = 3) down to 0 and no-op because those 3 blocks have already been pruned.

app/app.go
damiannolan: 100%, thank you for spotting this. This is a relic from debugging that issue :D

x/zkism/keeper/abci.go
damiannolan: if this was moved to before the `for` loop and we decreased the number of entries from e.g. 10k to 0, then we would maintain 10k stale entries which are never pruned and num entries would be inconsistent to actual state

x/zkism/keeper/abci.go
damiannolan: Generally speaking, no. We don't really use the branched multistore with cached context for basic store ops in cosmos sdk.

Furthermore, if we hit an error on the iavl tree store we will panic before it can bubble up the stack. https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.51.x-celestia/store/iavl/store.go#L222

x/zkism/keeper/abci.go
rootulp: oh wow, good point! Resolving.

PR #5677, opened 2025-09-06
docs: improve test comments

JohnBonny:
## Overview

Update comments to reference the correct variable 'ctx' instead of 
'suite.ctx'. The test function uses a local 'ctx' variable, not a 
test suite context.

Changes:
- Line 96: `set suite.ctx with TxBytes manually` → `set ctx with TxBytes manually`
- Line 124: `Set suite.ctx with smaller simulated TxBytes manually` → `Set ctx with smaller simulated TxBytes manually`

### DIFF
--- a/app/ante/tx_size_gas_test.go
+++ b/app/ante/tx_size_gas_test.go
@@ -120,6 +120,7 @@ func TestConsumeGasForTxSize(t *testing.T) {
 			// require that simulated tx is smaller than tx with signatures
 			require.True(t, len(simTxBytes) < len(txBytes), "simulated tx still has signatures")
 
+			// Set ctx with smaller simulated TxBytes manually
 			ctx = ctx.WithTxBytes(simTxBytes).WithExecMode(sdk.ExecModeSimulate)
 
 			beforeSimGas := ctx.GasMeter().GasConsumed()


app/ante/tx_size_gas_test.go
```diff
@@ -93,7 +93,7 @@ func TestConsumeGasForTxSize(t *testing.T) {
 			txSizeCostPerByte := appconsts.TxSizeCostPerByte
 			expectedGas := storetypes.Gas(len(txBytes)) * txSizeCostPerByte
 
-			// set suite.ctx with TxBytes manually
+			// set ctx with TxBytes manually
```
rootulp: The line of code is self-describing so IMO this entire comment can be deleted.

```suggestion
```

rootulp: The line of code is self-describing so IMO this entire comment can be deleted.

```suggestion
```

rootulp (dismissed):

JohnBonny:
Hey @rootulp ! I noticed a GitHub interface duplication - you pointed to the same line (Line 96) twice suggesting to remove the comment. I thought you meant to remove both line 96 and line 124, so I fixed both.

If I'm wrong, please let me know and I'll restore the comment on line 124.

rootulp:
My comment was meant just for line 96 because I think your revised comment on line 124 was fine.

app/ante/tx_size_gas_test.go
```diff
@@ -121,7 +120,6 @@ func TestConsumeGasForTxSize(t *testing.T) {
 			// require that simulated tx is smaller than tx with signatures
 			require.True(t, len(simTxBytes) < len(txBytes), "simulated tx still has signatures")
 
-			// Set suite.ctx with smaller simulated TxBytes manually
 			ctx = ctx.WithTxBytes(simTxBytes).WithExecMode(sdk.ExecModeSimulate)
```
rootulp, line 124:
```suggestion
			// Set ctx with smaller simulated TxBytes manually
			ctx = ctx.WithTxBytes(simTxBytes).WithExecMode(sdk.ExecModeSimulate)
```

PR #5690, opened 2025-09-08
feat: resign during broadcast flup introducing trusted mode

ninabarbakadze:
## Overview

Flups of https://github.com/celestiaorg/celestia-app/pull/5582

### DIFF
--- a/pkg/user/tx_client.go
+++ b/pkg/user/tx_client.go
@@ -40,6 +40,7 @@ import (
 const (
 	DefaultPollTime          = 3 * time.Second
 	txTrackerPruningInterval = 10 * time.Minute
+	defaultTrustedMode       = true
 )
 
 type Option func(client *TxClient)
@@ -157,9 +158,16 @@ type TxClient struct {
 	// that was submitted to the chain
 	txTracker           map[string]txInfo
 	gasEstimationClient gasestimation.GasEstimatorClient
+	// trustedMode indicates whether to trust the consensus node for account information.
+	// When true (default), the client trusts the consensus node to provide accurate
+	// account numbers and sequences without additional verification.
+	// When false, additional verification steps would be performed (to be implemented).
+	trustedMode bool
 }
 
-// NewTxClient returns a new signer using the provided keyring
+// NewTxClient returns a new TxClient using the provided keyring.
+// By default, the client operates in trusted mode, which assumes the consensus
+// node is honest.
 func NewTxClient(
 	cdc codec.Codec,
 	signer *Signer,
@@ -190,6 +198,11 @@ func NewTxClient(
 		txTracker:           make(map[string]txInfo),
 		cdc:                 cdc,
 		gasEstimationClient: gasestimation.NewGasEstimatorClient(conn),
+		// By default, the client operates in trusted mode,
+		// which trusts the consensus node to provide accurate account information
+		// without additional verification.
+		// It is advised to use the consensus endpoint that you trust.
+		trustedMode: defaultTrustedMode, // Default to trusted mode is true
 	}
 
 	for _, opt := range options {
@@ -205,7 +218,7 @@ func NewTxClient(
 }
 
 // SetupTxClient uses the underlying grpc connection to populate the chainID, accountNumber and sequence number of all
-// the accounts in the keyring.
+// the accounts in the keyring. By default, the client operates in trusted mode use the consensus endpoint that you trust.
 func SetupTxClient(
 	ctx context.Context,
 	keys keyring.Keyring,
@@ -328,6 +341,10 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
 	// pruning has to be done in broadcast, since users
 	// might not always call ConfirmTx().
 	client.pruneTxTracker()
+
+	if !client.trustedMode {
+		return nil, fmt.Errorf("non-trusted mode not yet implemented")
+	}
 
 	account, err := client.getAccountNameFromMsgs(msgs)
 	if err != nil {
@@ -523,8 +540,6 @@ func (s *TxClient) retryBroadcastingTx(ctx context.Context, txBytes []byte) (*sd
 		return nil, err
 	}
 
-	// Add to tx tracker.
-	s.trackTransaction(signer, broadcastTxResp.TxHash, newTxBytes)
 	return broadcastTxResp, nil
 }
 
@@ -786,6 +801,11 @@ func (client *TxClient) DefaultAddress() sdktypes.AccAddress {
 
 func (client *TxClient) DefaultAccountName() string { return client.defaultAccount }
 
+// IsTrustedMode returns whether the client is operating in trusted mode.
+func (client *TxClient) IsTrustedMode() bool {
+	return client.trustedMode
+}
+
 func (client *TxClient) checkAccountLoaded(ctx context.Context, account string) error {
 	if _, exists := client.signer.accounts[account]; exists {
 		return nil
@@ -798,7 +818,15 @@ func (client *TxClient) checkAccountLoaded(ctx context.Context, account string) 
 	if err != nil {
 		return fmt.Errorf("retrieving address from keyring: %w", err)
 	}
-	// FIXME: have a less trusting way of getting the account number and sequence
+
+	if !client.trustedMode {
+		// Non-trusted mode would implement additional verification here
+		// TODO: implement non-trusted account verification
+		return fmt.Errorf("non-trusted mode not yet implemented")
+	}
+
+	// In trusted mode, we trust the consensus node to provide accurate account information
+	// without additional verification. This assumes the node is honest and not compromised.
 	accNum, sequence, err := QueryAccount(ctx, client.conns[0], client.registry, addr)
 	if err != nil {
 		return fmt.Errorf("querying account %s: %w", account, err)


pkg/user/tx_client.go
```diff
@@ -329,6 +342,10 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
 	// might not always call ConfirmTx().
 	client.pruneTxTracker()
 
+	if !client.trustedMode {
+		return nil, fmt.Errorf("non-trusted mode not yet implemented")
+	}
```
cmwaters, line 347: In theory, you should still be able to broadcast a transaction, it just wouldn't be resubmitted if the nonce was incorrect, it would instead just error back to the user.

pkg/user/tx_client.go
```diff
@@ -190,6 +198,11 @@ func NewTxClient(
 		txTracker:           make(map[string]txInfo),
 		cdc:                 cdc,
 		gasEstimationClient: gasestimation.NewGasEstimatorClient(conn),
+		// By default, the client operates in trusted mode,
+		// which trusts the consensus node to provide accurate account information
+		// without additional verification.
+		// It is advised to use the consensus endpoint that you trust.
+		trustedMode: defaultTrustedMode, // Default to trusted mode is true
```
cmwaters, line 205: given this is a bool, I would call it:
```suggestion
		trusted: true, // Default to trusting the consensus node
```
instead of trustedMode

pkg/user/tx_client.go
```diff
@@ -786,6 +801,11 @@ func (client *TxClient) DefaultAddress() sdktypes.AccAddress {
 
 func (client *TxClient) DefaultAccountName() string { return client.defaultAccount }
 
+// IsTrustedMode returns whether the client is operating in trusted mode.
+func (client *TxClient) IsTrustedMode() bool {
+	return client.trustedMode
+}
```
cmwaters, line 807: I think it reads better if it's like this: 
```suggestion
func (client *TxClient) IsTrusting() bool {
	return client.trustedMode
}
```

pkg/user/tx_client.go
```diff
@@ -798,7 +818,15 @@ func (client *TxClient) checkAccountLoaded(ctx context.Context, account string)
 	if err != nil {
 		return fmt.Errorf("retrieving address from keyring: %w", err)
 	}
-	// FIXME: have a less trusting way of getting the account number and sequence
+
+	if !client.trustedMode {
+		// Non-trusted mode would implement additional verification here
+		// TODO: implement non-trusted account verification
+		return fmt.Errorf("non-trusted mode not yet implemented")
```
cmwaters, line 825: This should return a more helpful error like:
```suggestion
		return fmt.Errorf("no info for account %s found", account)
```
You could also check this earlier on line 183

cmwaters (commented):
Good initial start. 

If we want to support the non-trusted format, I would add an Option that allows you to change it to not trusting

pkg/user/tx_client.go
```diff
@@ -157,9 +158,16 @@ type TxClient struct {
 	// that was submitted to the chain
 	txTracker           map[string]txInfo
 	gasEstimationClient gasestimation.GasEstimatorClient
+	// trustedMode indicates whether to trust the consensus node for account information.
+	// When true (default), the client trusts the consensus node to provide accurate
+	// account numbers and sequences without additional verification.
+	// When false, additional verification steps would be performed (to be implemented).
```
rootulp, line 164: Do we need a follow-up issue to track implementing the additional verification steps?

```suggestion
	// When false, additional verification steps need to be performed (to be implemented).
```

pkg/user/tx_client.go
```diff
@@ -190,6 +198,11 @@ func NewTxClient(
 		txTracker:           make(map[string]txInfo),
 		cdc:                 cdc,
 		gasEstimationClient: gasestimation.NewGasEstimatorClient(conn),
+		// By default, the client operates in trusted mode,
+		// which trusts the consensus node to provide accurate account information
+		// without additional verification.
+		// It is advised to use the consensus endpoint that you trust.
```
rootulp, line 204: [nit] this comment won't be read by end-users and won't surface in API docs. The target audience for this code comment should be maintainers. 

IMO this information belongs in the CLI or wherever a user is providing the consensus node endpoint.

```suggestion
```

pkg/user/tx_client.go
```diff
@@ -157,9 +158,16 @@ type TxClient struct {
...
+	// When true (default), the client trusts the consensus node to provide accurate
+	// account numbers and sequences without additional verification.
+	// When false, additional verification steps would be performed (to be implemented).
+	trustedMode bool
 }
 
-// NewTxClient returns a new signer using the provided keyring
+// NewTxClient returns a new TxClient using the provided keyring.
+// By default, the client operates in trusted mode, which assumes the consensus
+// node is honest.
```
rootulp, line 170: [nit] it's not really by default because it's forced. There is no way to override the default.

```suggestion
// The client operates in trusted mode, which assumes the consensus
// node is honest.
```

ninabarbakadze:
> Good initial start.
> 
> If we want to support the non-trusted format, I would add an Option that allows you to change it to not trusting

Given that the not-trusted option is not really usable I didn't think giving that option was really valuable.

pkg/user/tx_client.go
```diff
@@ -205,7 +218,7 @@ func NewTxClient(
 }
 
 // SetupTxClient uses the underlying grpc connection to populate the chainID, accountNumber and sequence number of all
-// the accounts in the keyring.
+// the accounts in the keyring. By default, the client operates in trusted mode use the consensus endpoint that you trust.
```
rootulp, line 221: Again not a default because no option to override the trusted mode.

```suggestion
// the accounts in the keyring. Use a GRPC connection that you trust.
```

pkg/user/tx_client.go
rootulp: Why is this check necessary? It doesn't seem possible to set trustedMode to `false`.

cmwaters:
> Given that the not-trusted option is not really usable I didn't think giving that option was really valuable.

Why is it not usable? It's probably not advised for high frequency transaction submission but it should work for low frequency

ninabarbakadze:
> > Given that the not-trusted option is not really usable I didn't think giving that option was really valuable.
> 
> Why is it not usable? It's probably not advised for high frequency transaction submission but it should work for low frequency

Because from what I can tell most of tx client users have high transaction submission frequency. If I was a user and there were trusted and not trusted modes I would go with not trusted expecting it to work. We can keep the non-trusted way but document it very clearly that it's not the best for high frequency submissions.

cmwaters:
> > > Given that the not-trusted option is not really usable I didn't think giving that option was really valuable.
> > 
> > 
> > Why is it not usable? It's probably not advised for high frequency transaction submission but it should work for low frequency
> 
> Because from what I can tell most of tx client users have high transaction submission frequency. If I was a user and there were trusted and not trusted modes I would go with not trusted expecting it to work. We can keep the non-trusted way but document it very clearly that it's not the best for high frequency submissions.

Well the default is trusted and I assume most people won't deviate from the default. I agree though that I would add a warning that the trustless mode is not suited for high frequency submissions (i.e. multiple blobs per block)

pkg/user/tx_client.go
evan-forbes, line 164: is the thing blocking this just deeper integration with the light clients so we can verify proofs?

how does setting that to untrusted (my understanding of what we were doing before) fix that? are we trusting the consensus node for the initial value?

cmwaters:
> I think I have a misunderstanding of what is meant by untrusted. In theory, if we're verifying proofs to state and following the headers, then how would a trusted version be more suitable for high frequency? or is this just that we don't have proofs to mempool state (we don't have the ability to query mempool state anyway)?

Trusted means you rely on a third party for replay protection. Untrusted means you don't. If you added proofs to the trusted version then that would come along way to making it untrusted. Of course this wouldn't fully work because trusted relies on a view of the mempool state which is difficult and probably impractical to prove.

> if we don't have untrusted mode implemented, then imo we can just not merge this and include it there.

The original implementation of the TxClient was untrusted/trustless. Only when the ability to query the accounts for missing account info and parsing the errors from the consensus node do we start to trust what the consensus node is telling us in a way that can break replay protection.

> without very clear docs on what we mean by trusted and the consequences of that then adding the option only adds confusion.

Regardless of whether we support a trustless option or not, we should be far more explicit in the past that the tx client depends on a trusted node.

I get that trustless is hard and that's why I haven't been opposed to just going fully trusted and perhaps the long term dream is one day it can be trustless else what's the point of having light nodes if everyone needs to run full nodes anyway. 

I am happy if the decision we want to make is to go full trusted in which case this PR doesn't seem to be necessary

evan-forbes:
> Trusted means you rely on a third party for replay protection. Untrusted means you don't.

>  I haven't been opposed to just going fully trusted...

> I am happy if the decision we want to make is to go full trusted in which case this PR doesn't seem to be necessary

maybe striking the balance is a product decision then? if the design can be simplified by trusting, then perhaps its time to revisit the assumption that untrusted is needed. 

>  ...and perhaps the long term dream is one day it can be trustless else what's the point of having light nodes if everyone needs to run full nodes anyway.

[not important questions for my own understanding]

what's the worst attack that can occur in this scenario? 

even the "untrusted" is trusting a gas price estimate correct? and that the untrusted txclient is trusting rpc node is providing honest errors from the mempool as there are no proofs for those. also, are we actually verifying proofs against the app hash? if not, then do we get any meaningful different between "trusted" and "untrusted" currently?

ninabarbakadze:
Sooooo are we ok to close this PR? @cmwaters @evan-forbes 

Personally, I don't think we need trusted and not trusted modes until we have both implemented and can actually be switched between in a manner that's functional and doesn't block users.

cmwaters:
> what's the worst attack that can occur in this scenario?

A malicious consensus node could drain all the funds of an account by taking their one blob they want to submit and making them sign it 10,000 times exhausting all the funds in the account.

> even the "untrusted" is trusting a gas price estimate correct? and that the untrusted txclient is trusting rpc node is providing honest errors from the mempool as there are no proofs for those. also, are we actually verifying proofs against the app hash? if not, then do we get any meaningful different between "trusted" and "untrusted" currently?

You're totally right about the gas price (I think we have some upper bounds for additional protection but a malicious estimator could make a user overpay). At least here I think it's more intuitive what the trust assumptions are.

Regarding mempools. You're also right that there are no proofs. A malicious consensus node could return an error on a perfectly valid transaction or go quiet on an invalid one. That's bad but afaict, it doesn't touch a users funds. There's different levels of trusting and trusting a consensus node for replay protection is probably one of the more dangerous.

In summary, **my personal view is that having a fully trusting approach undermines the existence of light nodes and a core value of Celestia**. I also believe it's possible to have a world of light nodes with far fewer trust assumptions than what is status quo now. It just takes a bit of engineering

In saying those views, **I was also aware (and complicit) from nina's previous PR that in order to have better UX for now, we're going to sacrifice the trust assumptions and I think that's the right choice.**

For this PR, we can close this for now if we're not ready to start on lowering trust assumption but I think this should be a high priority for both protocol teams (perhaps just under the throughput goal) going forward. **Users should be able to reliably and securely submit and retrieve data without needing to run a consensus node themselves.**

ninabarbakadze:
> > what's the worst attack that can occur in this scenario?
> 
> A malicious consensus node could drain all the funds of an account by taking their one blob they want to submit and making them sign it 10,000 times exhausting all the funds in the account.
> 
> > even the "untrusted" is trusting a gas price estimate correct? and that the untrusted txclient is trusting rpc node is providing honest errors from the mempool as there are no proofs for those. also, are we actually verifying proofs against the app hash? if not, then do we get any meaningful different between "trusted" and "untrusted" currently?
> 
> You're totally right about the gas price (I think we have some upper bounds for additional protection but a malicious estimator could make a user overpay). At least here I think it's more intuitive what the trust assumptions are.
> 
> Regarding mempools. You're also right that there are no proofs. A malicious consensus node could return an error on a perfectly valid transaction or go quiet on an invalid one. That's bad but afaict, it doesn't touch a users funds. There's different levels of trusting and trusting a consensus node for replay protection is probably one of the more dangerous.
> 
> In summary, **my personal view is that having a fully trusting approach undermines the existence of light nodes and a core value of Celestia**. I also believe it's possible to have a world of light nodes with far fewer trust assumptions than what is status quo now. It just takes a bit of engineering
> 
> In saying those views, **I was also aware (and complicit) from nina's previous PR that in order to have better UX for now, we're going to sacrifice the trust assumptions and I think that's the right choice.**
> 
> For this PR, we can close this for now if we're not ready to start on lowering trust assumption but I think this should be a high priority for both protocol teams (perhaps just under the throughput goal) going forward. **Users should be able to reliably and securely submit and retrieve data without needing to run a consensus node themselves.**

Closing this PR and tagging @evan-forbes for context. 

All in all I agree with both of your statements and I think eventually once we ship FibreDA and/or other pressing workstreams if we decide to commit to this higher engineering effort can go into rethinking tx submissions in general from both core/app and node teams because it will be a big undertaking to prove all the paths that are currently trusted. Someone just needs to lead this either @cmwaters or myself.

PR #5675, opened 2025-09-05
refactor: remove unused params

rootulp:
On `main`, my editor surfaces these unused parameters as lint warnings so I removed them in this PR.

PR #5694, opened 2025-09-08
fix: unused params regression

rootulp:
CI on main has been failing since https://github.com/celestiaorg/celestia-app/pull/5675/files merged. The PR that merged right before it started using the `signer` in that method: https://github.com/celestiaorg/celestia-app/pull/5582/files


rootulp:
Overriding to fix CI on `main`

PR #5700, opened 2025-09-08
docs(txsim): correct Run function documentation to reflect actual gRPC-only usage

sashass1315:
Remove incorrect RPC endpoint requirement from Run function documentation.
Function only accepts grpcEndpoint parameter and uses gRPC for all operations.
RPC is only used in tests for validation, not in main function logic.


sashass1315:
> LGTM based on the arguments to the function. @sashass1315 did you verify that there's no other way to provide an RPC endpoint to txsim? I looked at git history and this function hasn't been modified recently so perhaps the comment is super stale.

The comment is indeed stale - likely copied from another function or from an earlier version that used both endpoints. The current implementation is gRPC-only.

PR #5701, opened 2025-09-09
fix: add missing nextAnteHandler function in ante_test.go

lidanserj:
Add missing nextAnteHandler function that was referenced but not defined in TestSigVerificationDecorator, causing compilation error.

This follows the same pattern already used in params_test.go in the same package, where nextAnteHandler is defined as a simple test helper function that returns the context without modifications.


lidanserj:
my mistake, closing the PR

PR #5706, opened 2025-09-09
chore(ci): update Github Actions runner to dedicated one

### LINKED ISSUE #4870: Update docker E2E tests to use self hosted runner
We are currently running the docker E2E tests in a test matrix using the standard github runners, we should be able to move these to just run in parallel on a single self hosted runner.

In the go tests under `docker-e2e` we can add `t.Parallel()` in the test definitions e.g. [here](https://github.com/celestiaorg/celestia-app/blob/7ed27698bd9780aafe67621419f8839bb405efc3/test/docker-e2e/e2e_simple_test.go#L11).

For `t.Parallel()` to work, we will need to move [`SetupSuite`](https://github.com/celestiaorg/celestia-app/blob/a77a2d07e80bd64fb8da51767cef72dde357c9c8/test/docker-e2e/e2e_test.go#L40-L44) into a `SetupTest`. Otherwise tests will re-use (and teardown) the same docker network.

In the [workflow](https://github.com/celestiaorg/celestia-app/blob/a77a2d07e80bd64fb8da51767cef72dde357c9c8/.github/workflows/test.yml#L72-L80) we can update it to just run go test in the `docker-e2e` directory.

sysrex:
fixes #4870

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
         timeout-minutes: 10
 
   go-test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-8
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


chatton (approved):
👍 thank you!

Issue #5702, opened 2025-09-09
misc(pkg/user): Allow caller to pass span through context into TxClient hotpaths


renaynay:
Celestia-node uses otel traces inside of our codebase. We are also the primary user of `TxClient`. We would like the ability to pass a span into certain `TxClient` hot paths such that we can record events that happen inside of `TxClient` and have more **granular visibility** in there. **We will be able to use these traces to debug individual cases of long tx submission time, along with other potential issues.** 

_This is a minimal change and would not impact usability for other callers who do not utilise otel spans, otel will produce a noop span in the case it cannot find it in the given context. It is also minimal in that it does not require celestia-app to fully adopt otel tracing architecture and caters in a lightweight manner to users who already have otel tracing architecture integrated without impacting usability for those who do not._

For example, grab the span from the context inside of `SubmitPayForBlob` and add some events with relevant attributes there.

```go
// SubmitPayForBlob forms a transaction from the provided blobs, signs it, and submits it to the chain.
// TxOptions may be provided to set the fee and gas limit.
func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error) {
	span := trace.SpanFromContext(ctx)

	resp, err := client.BroadcastPayForBlob(ctx, blobs, opts...)
	if err != nil {
		return nil, err
	}
	span.AddEvent("txclient: broadcasted PFB", trace.WithAttributes(attribute.Int("num blobs", len(blobs))))

	return client.ConfirmTx(ctx, resp.TxHash)
}
```

And then we grab the same span from inside of `ConfirmTx` (the most interesting path for coverage)

```go
func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxResponse, error) {
	txClient := tx.NewTxClient(client.conns[0])

	pollTicker := time.NewTicker(client.pollTime)
	defer pollTicker.Stop()

	span := trace.SpanFromContext(ctx)
...
```

and cover the more interesting cases (e.g. add an event when we have an eviction and attempt to resubmit, add an event when tx status is finally something other than `pending`, etc.)

renaynay:
Please also add traces to the following paths in #5707: 

- [ ] AccountByAddress
- [ ] EstimateGasPrice
- [ ] EstimateGasPriceAndUsage

Apply the same principles as for the events recorded in the SubmitPFB path.

renaynay:
[...]

PR #5707, opened 2025-09-09
feat: add open telemetry traces

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/5702


pkg/user/tx_client.go
```diff
@@ -254,11 +256,24 @@ func SetupTxClient(
...
+		// 	attribute.Int("num_blobs", len(blobs)),
+		// 	attribute.String("error", err.Error()),
+		// ))
 		return nil, err
 	}
 
+	span.AddEvent("txclient: broadcasted PFB", trace.WithAttributes(
+		attribute.Int("num_blobs", len(blobs)),
+		attribute.String("tx_hash", resp.TxHash),
+		attribute.Float64("elapsed_seconds", time.Since(startTime).Seconds()),
```
renaynay: no times needed bc span inherently does this

pkg/user/tx_client.go
```diff
@@ -609,17 +632,29 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
 			// Continue polling if the transaction is still pending
 			select {
 			case <-ctx.Done():
+				span.AddEvent("txclient: context cancelled during confirmation", trace.WithAttributes(
```
renaynay: we don't need this one

pkg/user/tx_client.go
```diff
@@ -628,19 +663,41 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
 				client.deleteFromTxTracker(txHash)
 				return nil, executionErr
 			}
+			// if it took more than 7 seconds to confirm, log it
+			if elapsed > 7*time.Second {
```
renaynay: we don't need this for traces, just add the event itself bc we'll see the time it took for it to get confirmed eventually. We would need smth like this more for metrics but we can discuss that later.

pkg/user/tx_client.go
```diff
@@ -628,19 +663,41 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
+				attribute.String("tx_hash", txHash),
+				attribute.Float64("elapsed_seconds", time.Since(startTime).Seconds()),
+			))
 			_, signer, exists := client.GetTxFromTxTracker(txHash)
 			if !exists {
 				return nil, fmt.Errorf("tx: %s not found in txTracker; likely failed during broadcast", txHash)
 			}
 			// Resubmit straight away in the event of eviction and keep polling until tx is committed
 			_, err := client.broadcastTx(ctx, client.conns[0], client.txTracker[txHash].txBytes, signer)
 			if err != nil {
```
renaynay, line 699: let's add one more event here in case of successful resubmission

renaynay (commented):
we don't need to add attributes that are individual to the tx itself here (like tx hash, height, etc) because they don't really do much for us in the context of understanding patterns across traces.

let's stick to minimal attributes like num blobs and if an error occurs, u can interpolate it as a string in the event message itself like

`span.AddEvent(fmt.Sprintf("error happened during blah blah : %w", err))`

ninabarbakadze:
> we don't need to add attributes that are individual to the tx itself here (like tx hash, height, etc) because they don't really do much for us in the context of understanding patterns across traces.
> 
> let's stick to minimal attributes like num blobs and if an error occurs, u can interpolate it as a string in the event message itself like
> 
> `span.AddEvent(fmt.Sprintf("error happened during blah blah : %w", err))`

Addressed your comments. Please take another look.

renaynay:
Screenshot 2025-09-22 at 17 32 24

added some of my own traces in here to make sure it works but yes it works. 

I will add some more requests for traces but i'll just push directly to this branch @ninabarbakadze

pkg/user/tx_client.go
```diff
@@ -684,6 +732,7 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
 			client.deleteFromTxTracker(txHash)
 			return nil, fmt.Errorf("tx with hash %s was rejected by the node with execution code %d", txHash, resp.ExecutionCode)
 		default:
+			span.RecordError(fmt.Errorf("txclient/ConfirmTx: unknown tx: %s", txHash))
```
rootulp: [question] this case seems like it will occur when the tx status doesn't match any of the above cases. So perhaps it's more accurate to say


```suggestion
			span.RecordError(fmt.Errorf("txclient/ConfirmTx: unknown tx status for tx: %s", txHash))
```

pkg/user/tx_client.go
```diff
@@ -809,12 +878,16 @@ func (client *TxClient) AccountByAddress(ctx context.Context, address sdktypes.A
 	client.mtx.Lock()
 	defer client.mtx.Unlock()
 
+	span := trace.SpanFromContext(ctx)
+
 	accountName := client.signer.accountNameByAddress(address)
 	if accountName == "" {
+		span.AddEvent("txclient/AccountByAddress: account not found for address")
```
rootulp: [question] should we include the address in the event to aid debugging if this gets hit?

rootulp (approved):
LGTM

pkg/user/tx_client.go
```diff
@@ -882,7 +883,7 @@ func (client *TxClient) AccountByAddress(ctx context.Context, address sdktypes.A
 
 	accountName := client.signer.accountNameByAddress(address)
 	if accountName == "" {
-		span.AddEvent("txclient/AccountByAddress: account not found for address")
+		span.AddEvent(fmt.Sprintf("txclient/AccountByAddress: account not found for accountName: %s", accountName))
```
rootulp: This event will always contain an empty account name. Instead of including the account name, it should include the address. Something like this:

```go
		span.AddEvent(fmt.Sprintf("txclient/AccountByAddress: account not found for address: %s", address))
```

pkg/user/tx_client.go
```diff
@@ -883,7 +882,9 @@ func (client *TxClient) AccountByAddress(ctx context.Context, address sdktypes.A
 
 	accountName := client.signer.accountNameByAddress(address)
 	if accountName == "" {
-		span.AddEvent(fmt.Sprintf("txclient/AccountByAddress: account not found for accountName: %s", accountName))
+		span.AddEvent(
+			fmt.Sprintf("txclient/AccountByAddress: account not found for accountName: %s", address.String()),
```
```suggestion
			fmt.Sprintf("txclient/AccountByAddress: account not found for address: %s", address.String()),
```

rootulp (dismissed):

rootulp:
Note this has a merge conflict so I don't think it will auto-merge

ninabarbakadze:
> Note this has a merge conflict so I don't think it will auto-merge

Resolved! maybe take another look @renaynay to make sure everything is as you left it since there were some structural changes so I had to move the code around a little bit

pkg/user/tx_client.go
```diff
@@ -434,9 +455,11 @@ func (client *TxClient) submitToSingleConnection(ctx context.Context, txBytes []
 		// Retry with updated sequence
 		retryTxBytes, err := client.resignTransactionWithNewSequence(txBytes)
 		if err != nil {
+			span.RecordError(fmt.Errorf("txclient/retryBroadcastingTx: rebroadcast error: %w", err))
 			return nil, err
 		}
 
+		span.AddEvent("txclient/retryBroadcastingTx: successfully rebroadcasted tx after sequence mismatch")
```
rootulp: [question] `retryBroadcastingTx` doesn't match a method call above. Did it mean to say `resignTransactionWithNewSequence`?

rootulp (approved):

pkg/user/tx_client.go
ninabarbakadze: fixed

rootulp (dismissed):

renaynay (dismissed):

evan-forbes (approved):

PR #5708, opened 2025-09-09
docs(pkg/user): fix misleading comment

galaparm:
## Overview

- Fix incorrect comment in assertTxInTxTracker function that claimed sequence does not increase, when it actually validates sequence increases by 1
- Remove leftover refactoring comment "Changed from error to bool" in broadcastTestCase struct

### DIFF
--- a/pkg/user/tx_client_test.go
+++ b/pkg/user/tx_client_test.go
@@ -492,7 +492,7 @@ func wasRemovedFromTxTracker(txHash string, txClient *user.TxClient) bool {
 	return !exists && seq == 0 && signer == ""
 }
 
-// asserts that a tx was indexed in the tx tracker and that the sequence increases by 1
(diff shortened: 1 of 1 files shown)


pkg/user/tx_client_test.go
```diff
@@ -492,7 +492,7 @@ func wasRemovedFromTxTracker(txHash string, txClient *user.TxClient) bool {
 	return !exists && seq == 0 && signer == ""
 }
 
-// asserts that a tx was indexed in the tx tracker and that the sequence does not increase
+// asserts that a tx was indexed in the tx tracker and that the sequence increases by 1
```
rootulp:
```suggestion
// assertTxInTxTracker verifies that a tx was indexed in the tx tracker and that the sequence increases by one after broadcast.
```

rootulp (dismissed):

rootulp (approved):

Issue #5697, opened 2025-09-08
Submit governance proposal to increase max block size to 32 MiB on Arabica


rootulp:
## Context

V6 activates on Arabica at https://arabica.celenium.io/block/8105605

## Problem

We need to submit governance proposals to increase from 8 MiB blocks to 32 MiB blocks.

## Proposal

1. Prepare a governance proposal to modify the params from 8 MiB to 32 MiB
2. [blocked] After v6 activates on Arabica, submit the governance proposal
3. [blocked] Vote yes on the proposal from all the Arabica validators
4. After the proposal passes, verify that the max block size increased to 32 MiB

rootulp:
For reference, the proposal to increase from 2 MiB to 8 MiB on Mainnet: https://www.mintscan.io/celestia/proposals/5

Based on that, I think we need to increase the `GovMaxSquareSize` and `BlockParams.max_bytes`.

Helpful links:
- https://github.com/celestiaorg/celestia-app/blob/main/docs/maintainers/increase-square-size.md
- https://github.com/celestiaorg/celestia-app/blob/main/docs/maintainers/modify-param.md

rootulp:
Confirmed that those two params need to be modified on Arabica. Here's how to query their current values

```shell
ssh arabica-validator-1
su celestia

celestia@validator-1:/root$ celestia-appd query blob params
params:
  gas_per_blob_byte: 8
  gov_max_square_size: "128"

celestia@validator-1:/root$ celestia-appd query consensus params
params:
  block:
    max_bytes: "8388608"
    max_gas: "-1"
  evidence:
    max_age_duration: 337h0m0s
    max_age_num_blocks: "242640"
    max_bytes: "1048576"
  validator:
    pub_key_types:
    - ed25519
  version:
    app: "6"
```

rootulp:
I can't submit a legacy gov proposal because: https://github.com/celestiaorg/celestia-app/issues/5537

```
celestia@validator-1:~$ celestia-appd tx gov submit-legacy-proposal param-change gov-proposal.json --from $FROM --fees $FEES --gas $GAS --gas-adjustment $GAS_ADJUSTMENT
Error: failed to parse proposal: proposal type is required
```

where gov-proposal.json is:

```json
{
  "type": "/cosmos.params.v1beta1.ParameterChangeProposal",
  "title": "Increase the GovMaxSquareSize to 256 and block.MaxBytes to 32 MiB to support 32 MiB blocks",
[...]

Issue #5537, opened 2025-08-13
Legacy gov parameter proposals no longer work


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/5526#issuecomment-3184580103

## Problem

It looks like legacy gov parameter proposals no longer work. I'm hitting this on the `cip-41` branch which is based on `main`

```
raw_log: 'failed to execute message; message index: 0: gov: no handler exists for proposal type'
```

cmwaters:
Was it removed when we migrated to v0.50.x-celestia?

Does that mean we should get rid of the antehandler (I recall we had one for legacy proposals)?

rootulp:
> Was it removed when we migrated to v0.50.x-celestia?

If it was then I don't think it was intentional or it was another breaking change in Cosmos SDK that went totally under the radar. If legacy gov proposals aren't expected to work we should:
- Call it out in release notes. 
- Remove CLI for it `celestia-appd tx gov submit-legacy-proposal`
- Update [this guide](https://github.com/celestiaorg/celestia-app/blob/main/docs/maintainers/modify-param.md) to use the non-legacy proposal

PR #5712, opened 2025-09-09
docs: update modify-param to use new gov proposal

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/5697

### DIFF
--- a/docs/maintainers/modify-param.md
+++ b/docs/maintainers/modify-param.md
@@ -1,48 +1,104 @@
 # Modify Param
 
-This doc will guide you through the process of modifying a parameter via governance.
+This doc contains steps to modify a parameter via governance. There are two ways to modify a parameter:
+
+1. Legacy gov proposal (`submit-legacy-proposal`)
+2. New gov proposal (`submit-proposal`)
+
+This guide uses the new gov proposal format.
 
 ## Prerequisites
 
+Verify the current parameter values
+
 ```shell
-# Verify the current parameter value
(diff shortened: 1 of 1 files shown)


docs/maintainers/modify-param.md
```diff
@@ -1,48 +1,104 @@
...
+This doc contains steps to modify a parameter via governance. There are two ways to modify a parameter:
+
+1. Legacy gov proposal (`submit-legacy-proposal`)
+2. New gov proposal (`submit-proposal`)
+
+This guide uses the new gov proposal format.
 
 ## Prerequisites
 
+Verify the current parameter values
```
rach-id: [nit]
```suggestion
Verify the current parameter values:
```

docs/maintainers/modify-param.md
```diff
@@ -1,48 +1,104 @@
 # Modify Param
 
-This doc will guide you through the process of modifying a parameter via governance.
+This doc contains steps to modify a parameter via governance. There are two ways to modify a parameter:
+
+1. Legacy gov proposal (`submit-legacy-proposal`)
```
rach-id, line 5: so this was removed from v6, that's why we're replacing it? or we still can make legacy proposals?

rootulp: According to @julienrbrt  we can still make legacy gov proposals but I wasn't able to figure out how to do so via the CLI. Will follow-up on investigating the legacy gov proposal issue in https://github.com/celestiaorg/celestia-app/issues/5537

PR #5720, opened 2025-09-11
docs: fix comment

timberrrrrr:
## Overview
Correct the comment for NewTxClient function to accurately reflect that it returns a *TxClient rather than a signer. The function signature shows it returns (*TxClient, error), not a signer.


pkg/user/tx_client.go
```diff
@@ -159,7 +159,7 @@ type TxClient struct {
 	gasEstimationClient gasestimation.GasEstimatorClient
 }
 
-// NewTxClient returns a new signer using the provided keyring
+// NewTxClient returns a new TxClient using the provided keyring
```
rootulp: a keyring isn't provided as an argument so can that please be fixed? One option:

```suggestion
// NewTxClient returns a new TxClient
```

Issue #4348, opened 2025-02-19
txsim nice to haves: block filling


zatarx:
## Context
             > @rootulp Could you elaborate on having an extra command for blobs? You mentioned filling up the blocks, but I didn't quite get that. Should it dynamically produce a PFB number, blob amount per PFB and blob sizes based on some network parameters and then spam the chain?

Yes. I want a way to more easily fill blocks to their max capacity. Currently Celestia blocks can contain 8 MiB of blob data but each individual blob is limited to ~2 MiB. This means that a user needs to submit 4 x ~2 MiB (or slightly smaller) blobs in order to fill a block. Even when I tried doing that via txsim, I wasn't able to consistently fill blocks so there may be other issues needed to accomplish my goal. You can skip this and/or extract it to a new issue so that it doesn't block the work in this PR.

## Problem
It's difficult to actually fill blocks on a testnet. It would be nice if txsim had a command to fill blocks so that users don't have to fiddle with the flags --blob, --blob-amounts, --blob-sizes

Extends #4223 https://github.com/celestiaorg/celestia-app/issues/4340#issuecomment-2665590195

zatarx:
@rootulp , I wanna introduce a flag `--fill-block n` to try to fill the n blocks by generating random PFBs. I draw inspiration from here https://github.com/celestiaorg/celestia-app/blob/main/app/benchmarks/results.md. I also wanna introduce a mandatory flag `--rpc-endpoint` specifically if this flag is provided. Need it for consensus_params to acquire most recent block max_bytes value. 

I was wondering if you wanna alter/add any additional context to this issue? Feel free to it assign me. If you think implementation notes/ADR is better here, lmk

rootulp:
Thanks for volunteering! I have a suggestion: can `--fill-block` be provided without a parameter in order to try filling blocks indefinitely. 

> I also wanna introduce a mandatory flag --rpc-endpoint specifically if this flag is provided.

I thought txsim already needed an RPC endpoint in order to submit txs. If it doesn't already have a consensus node endpoint, how does it know which chain to submit txs to?

zatarx:
[...]

PR #5721, opened 2025-09-11
feat(txsim): fill-blocks subcommand

zatarx:
Resolves https://github.com/celestiaorg/celestia-app/issues/4348 

## Overview
* added `fill-blocks` subcommand with optional `--share-overhead` flag to reduce usable share size by accounting for protocol metadata
* dev-tested locally with 32 MiB block size, achieving 8_339_456 bytes blob size by setting --share-overhead=1 (manually achieved best blob size was 8_388_000 where txs would still be committed)

## TODO
* run linter


test/cmd/txsim/cli.go
```diff
@@ -14,23 +14,27 @@ import (
...
-	TxsimKeypath       = "TXSIM_KEYPATH"
-	TxsimMasterAccName = "TXSIM_MASTER_ACC_NAME"
-	TxsimMnemonic      = "TXSIM_MNEMONIC"
+	TxsimGRPC              = "TXSIM_GRPC"
+	TxsimSeed              = "TXSIM_SEED"
+	TxsimPoll              = "TXSIM_POLL"
+	TxsimKeypath           = "TXSIM_KEYPATH"
+	TxsimMasterAccName     = "TXSIM_MASTER_ACC_NAME"
+	TxsimMnemonic          = "TXSIM_MNEMONIC"
+	FillBlocksBlobTxAmount = 1
```
rootulp: This isn't an environment variable so it doesn't belong in this const() block. Or the comment above the const block needs to be updated.

test/cmd/txsim/cli.go
```diff
@@ -219,11 +224,67 @@ account that can act as the master account. The command runs until all sequences
...
+
+	fillBlocksCmd := &cobra.Command{
+		Use:   "fill-blocks",
+		Short: "Fill blocks to their max capacity",
+		RunE:  fillBlocksExecute,
+		Long: `
+The subcommand is designed to simplify filling blocks to near-max capacity by submitting a single blob transaction.
+It targets a blob payload size of approximately (maxBytes / 4) - txOverhead, where maxBytes is the block size limit 
+is calculated using max square size, and txOverhead accounts for protocol-level overhead — 
+such as namespace size, share info, signer bytes, blob headers`,
```
rootulp: This command likely won't work as expected because on Mainnet the current max blob size is 2 MiB but the max block size is 8 MiB.

test/cmd/txsim/cli.go
```diff
@@ -219,11 +224,67 @@ account that can act as the master account. The command runs until all sequences
...
+	blobParamsResp, err := blobQueryClient.Params(cmd.Context(), &blobtypes.QueryParamsRequest{})
+	if err != nil {
+		return fmt.Errorf("Could not get blob params due to error: %v", err)
+	}
+
+	govMaxSquareSize := blobParamsResp.Params.GovMaxSquareSize
+	hardMaxSquareSize := appconsts.SquareSizeUpperBound // use getter instead???
+	maxSquareSize := min(int(govMaxSquareSize), hardMaxSquareSize)
+
+	// NOTE: Division by 4 accounts for 2d erasure coding schema
```
rootulp: This comment doesn't apply the line immediately after so it should be moved above the line it applies to

zatarx:
@rootulp , Thanks for your feedback. 
Correct, this will only submit 1 blob within a sequence.

This is my understanding so far - please correct me if I'm wrong. Sorry I'm still learning. 

1. Each blob expands by the factor of 4 (due to the erasure coding scheme) when stored in the block, i.e. if we have total blob size of slightly under 2MiB it is stored as 8MiB in the block. We can't exceed the total bytes size of blob within the PFB which is 2MiB for each block.
2. What I noticed when testing locally is when I break the rule above, I get code 21. I will look into this further to know for sure what this is.

rootulp:
> Each blob expands by the factor of 4 (due to the erasure coding scheme) when stored in the block

Each blob does get erasure coded but when I mention the block capacity (currently 8 MiB) that is for the original data square, not the erasure encoded data square. So to fill an 8 MiB block with 2 MiB blobs, we need to submit 4 x 2 MiB blobs and get them included in one block.

Ref: https://gist.github.com/rootulp/8a0083332e6ad6fd3f3f9121de3a956e

zatarx:
Gotcha, what's your concern when it comes to submitting 1 blob of ~8MiB vs 4 of 2Mib within the same tx? Was reading more on here https://github.com/celestiaorg/celestia-node/issues/3356#issuecomment-2471477558. Looks like blob size check was dropped. I will read up on this some more.

rootulp:
The max tx size was increased from 2 MiB to 8 MiB in https://github.com/celestiaorg/CIPs/blob/main/cips/cip-038.md but that is in celestia-app v6 which isn't on all networks yet.

zatarx:
Arabica run (additional `hash` field was manually added for testing). Blob size is close to 8MiB. Submitted from `celestia1dekz5qswtc3teqd6mpuwgs2xh9ugq8vhgug7nm`
```
{"level":"info","height":8205677,"address":"celestia194ydlm9xm6q09m83nhdmsvj7g9u3c7wng7xz78","hash":"412B486DE25129B82B686202AD5243B31F8D2EF54FC3D11AA3127C1DE148505E","blobs count":"1","total byte size of blobs":7897088,"time":"2025-09-15T19:29:31-07:00","message":"tx committed"}
{"level":"info","height":8205679,"address":"celestia194ydlm9xm6q09m83nhdmsvj7g9u3c7wng7xz78","hash":"7EF760E8CD14C51DA97AAC6C97467F0205BFF92537C5CEB409E3DCF2F252BB16","blobs count":"1","total byte size of blobs":7897088,"time":"2025-09-15T19:29:44-07:00","message":"tx committed"}
{"level":"info","height":8205681,"address":"celestia194ydlm9xm6q09m83nhdmsvj7g9u3c7wng7xz78","hash":"0D371A6B0BD730CF4CA753E28BD2950D283194B356747F73BEAC2E62BDCACB0C","blobs count":"1","total byte size of blobs":7897088,"time":"2025-09-15T19:29:56-07:00","message":"tx committed"}
```

Mocha run. Here max blocks size is 8MiB. ATM by default, 1 blob gets submitted per tx of max blob size of roughly 2MiB (~1.88MiB). Submitted from `celestia1s3y0tz68e458kj3rxfuesypyvzykda5nmkr6sn`
``` 
{"level":"info","height":8054046,"address":"celestia1rd6z7qtmnnxnnm85y94yw6lq44wwz6xcdnrad7","hash":"BD7ECA97565C44DA18A8613A726C68DB62260B4FD22B0D4195DC780BA57F4F1D","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T20:31:30-07:00","message":"tx committed"}
{"level":"info","height":8054047,"address":"celestia1rd6z7qtmnnxnnm85y94yw6lq44wwz6xcdnrad7","hash":"5144DC75ACCDB83F8376FA8DB414A09081AFB8D6D1AF84CEEAD8E86E7C70489A","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T20:31:48-07:00","message":"tx committed"}
...

{"level":"info","height":8054379,"address":"celestia12dcg0px70qjdjvq9aczcvzmlycdn3hccl0nmkh","hash":"F0E775459B37D6F96036EB4EB7975E43B3F6D8B073CF2A561206C69249A55905","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T21:11:24-07:00","message":"tx committed"}
{"level":"info","height":8054380,"address":"celestia12dcg0px70qjdjvq9aczcvzmlycdn3hccl0nmkh","hash":"5F181F64A1DC1E7FC6CCCB90FC4D537FD3AC2175895967929EEA5FAD5A6009CE","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T21:11:41-07:00","message":"tx committed"}
{"level":"info","height":8054383,"address":"celestia12dcg0px70qjdjvq9aczcvzmlycdn3hccl0nmkh","hash":"2DF138370814F7B4A26CE876430A330560169C8BB10C3B4CF132819C9CE03920","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T21:12:00-07:00","message":"tx committed"}
{"level":"info","height":8054386,"address":"celestia12dcg0px70qjdjvq9aczcvzmlycdn3hccl0nmkh","hash":"7F24035423D45366E9D64FF6411F7B168CD4921C3941CAD89CBD0AA7B6D6CD21","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T21:12:30-07:00","message":"tx committed"}
{"level":"info","height":8054389,"address":"celestia12dcg0px70qjdjvq9aczcvzmlycdn3hccl0nmkh","hash":"A1D6336D886EC42B1C8F8CD2AFF46A22D6C1F274FDF0D06B7D296DA63D678147","blobs count":"1","total byte size of blobs":1974272,"time":"2025-09-15T21:12:45-07:00","message":"tx committed"}

```

test/cmd/txsim/cli.go
zatarx: fixed

zatarx:
My bad, was rebinding the remote locally and accidentally deleted the remote. Restored now.

rootulp:
Seems like it doesn't work as expected on Mocha yet.

zatarx:
> Seems like it doesn't work as expected on Mocha yet.

just to confirm, can I think of the block as filled if its `square_size` * `square_size` * `share_size` = `max_block_size`? I get `square_size` with `celestia-appd query block 8075154 --home ~/.celestia-mocha`. In this case, square_size is 128

rootulp:
Yes. Also https://gist.github.com/rootulp/8a0083332e6ad6fd3f3f9121de3a956e might help. Basically max block size is the blob capacity in that table.

zatarx:
> Seems like it doesn't work as expected on Mocha yet.

I tried submitting txs where blob size would be slightly over 2MiB and they all failed (did with one and multiple blobs). I also checked square sizes for all the blocks on mocha that were included in the log, and they have `square_size` of 128, which is indicative of 8MiB block size.

My bad for circling over this again, but if we have a max square size of 128 returned in the block, it means that the original square size is 64, and so max tx size should 64 * 64 * 512 = 2MiB for blobs or slightly less due to share overhead. I looked at the gist, but if we have 128 as original size, it means that the ESS will be 256, which will lead to 32MiB block size. In case of Mocha, the OSS should be 64, that way ESS is 128 => 8MiB block size.

I have gas to submit a ~2MiB tx, but once I go over the said limit, I get an error every time on Mocha. What do you think I'm doing wrong here?

rootulp:
> I tried submitting txs where blob size would be slightly over 2MiB and they all failed (did with one and multiple blobs).

Currently, Mocha and Mainnet have a max tx size of 2 MiB. That will change when v6 activates on those networks.

> I also checked square sizes for all the blocks on mocha that were included in the log, and they have square_size of 128, which is indicative of 8MiB block size.

Correct. Just to clarify, in celestia-app and Celenium, `square_size` describes the original square size. Ref: https://celenium.io/block/7582149?tab=transactions is a block with original square size of 64 so blob capacity is ~2 MiB. Ref: https://gist.github.com/rootulp/8a0083332e6ad6fd3f3f9121de3a956e. A square size of 128 has blob capacity of 8 MiB.

> My bad for circling over this again, but if we have a max square size of 128 returned in the block, it means that the original square size is 64

No. A max square size of 128 means that the max original square size is 128.

> I have gas to submit a ~2MiB tx, but once I go over the said limit, I get an error every time on Mocha. What do you think I'm doing wrong here?

Each tx has to be slightly less than 2 MiB. If we need to fill an 8 MiB block on Mocha then we need 4 x PFB txs where each PFB tx is slightly less than 2 MiB (e.g. 1.8 MiB).

zatarx:
Ahh, thanks a lot for clearing it up. All along, I was pursuing the thesis that we could submit 1 tx and have it fill the entire block.

zatarx:
@rootulp would like your feedback on a slightly different approach.

By default, the tx amount (`--blob` flag value) approaches the minimum, and `blobSize` approaches `maxTxSize`. 
However, txsim users should be able to customize the desired tx amount to better optimize blockspace usage — 
especially when multiple users are submitting transactions concurrently.

For example, if someone is running txsim while other users are also submitting txs, 
defaulting to `blobSize == maxTxSize` can result in the txsim transaction not fitting in the current block. 
If there isn’t enough other activity to fill the block, this can lead to underutilized space. 
For instance, a new block might be 80% full, but the next txsim transaction requires 25% of the block space, 
which causes it to be deferred to the next block.

To avoid this, I propose to introduce `--tx-size` (`blobSize` + `txOverhead`) < `maxTxSize`, or use `--blob` as part of `fill-blocks` command. 
This increases the total number of transactions, making more efficient use of blockspace.

zatarx:
update: added appconst.MaxTxSize to cap tx size with share-overhead for users to configure as needed. `--blob` can be used to increase the amount of concurrent tx submissions, that way I now acquire cases when 2-3 transactions are submitted in the same block. Tested with `--blob` 16 on mocha
```
{"level":"info","height":8330924,"address":"celestia1ac30y7nav5e427jg53yhflnjrfet7nlza2vzxf","hash":"A0E9F97CE7C8136F9FF3177FC54EB6B37EA35D0CA4E21FA873E676DB0B81A52B","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:16-07:00","message":"tx committed"}
{"level":"info","height":8330925,"address":"celestia1kvavlqnvn67p24zx70w5u90axhx7v442dmkc5d","hash":"D4BEE6396218F3F155EA213D6B50D859774385521194A64029809D2222BFC075","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:16-07:00","message":"tx committed"}
{"level":"info","height":8330926,"address":"celestia1clccsaun8ehakcfvgamvng9mlhengsm43gc0sj","hash":"4A9F96C5D77DAD234F5BFE19B32B2CEE9E939C2B0C7B0ED1090C92CF3594D2F1","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:16-07:00","message":"tx committed"}
{"level":"info","height":8330926,"address":"celestia1yjw7eqd8ecjwjrkm009qre66hmd3k5gw800xw6","hash":"8F5DCC5D2D0E3C905A60565A563E2BF3E342C7397CCB77800FA56E4E83370B95","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:16-07:00","message":"tx committed"}
{"level":"info","height":8330927,"address":"celestia1hg2p339hmx0w8ewf70s5lvlnvx2c4uch75yvwv","hash":"D365F98F293920442F63C62DA608B2D87632A7289CBAB26ABE5383E9304D82D5","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:16-07:00","message":"tx committed"}
{"level":"info","height":8330928,"address":"celestia1qddn0mhdm64nsa7ppn9att0s8a5uf485rk2llp","hash":"D8E4B5F0D0E70EB1B10033148809361DB9B6D922CB6F7B61C40E8C95EAA7EB34","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:16-07:00","message":"tx committed"}
{"level":"info","height":8330932,"address":"celestia1xjyxmh7wtvcsqlzjqqkfg7nvcgh4tupnxmcptp","hash":"7CACB55D696F88BAA430617E20D860455887B499C42BF5BA6F60EEC0AF997AB0","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330932,"address":"celestia1g47x9wnm96y7lqnvreexvm6c8mdgznpcr5ndk7","hash":"820F732C0D46B763F33087C5648452F256CAAC6E5A6C7ED55BA75740A022CBA4","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330929,"address":"celestia1s5ux6nhfr7tqnapnmhfp8duekrpn9n5qegnsn9","hash":"A1B603C53A8BAF24C2E6DA553EB3288C0D1BA0F9E9C861072F19A01DA74109E2","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330933,"address":"celestia1njmu3flf97y5mcjyawxj9ze8njs505xxp4l05d","hash":"7A83488037F80EE643215663D7B7BC4601A39AB4A6513D3F97BFDC66975CD53F","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330930,"address":"celestia1jwdxr23ks25g0xm8mhpr5zx0803w3t8w73h77q","hash":"B888B716BFD191DD2ACAE677209F3FC4B3DE807C3C6EE2C98601236FF0773B6D","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330931,"address":"celestia1nzjm88qe2klhk0umdx8mrv08vhgxw2kxn2qw5k","hash":"ED4354D4E4EC613A71C05BA58E84B1D45D2D6422D1BBF2E5D15481D28FA2BCAA","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330934,"address":"celestia1n8gegt2r5a0hghtrr7jn7kal9nnvepj3ypekuf","hash":"983513F0AE05347F56AAA473F8E122064C8086659D63C290F6FE6FB4DB75F790","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330935,"address":"celestia10hhxyy6m3lqehj2n0kkdr2y9wgqahc6lhg940m","hash":"E989CB0C7E82A8C55CA90E93C1949111ED5FAD9EC6412D49BF5323622A381D2C","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:22-07:00","message":"tx committed"}
{"level":"info","height":8330936,"address":"celestia1xrd8atttla53hs2z4cmrh4mvvj93vzzdpyhjr7","hash":"5A2E81F82539779995A14863F199F59C42C037414A5328E2542321AE1A77622C","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:16:47-07:00","message":"tx committed"}
{"level":"info","height":8330938,"address":"celestia1kvavlqnvn67p24zx70w5u90axhx7v442dmkc5d","hash":"7EAFA6627F02DF1E9238389853A8556E78E08007729F4E243A03EDB9A2E24B11","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:17:29-07:00","message":"tx committed"}
{"level":"info","height":8330939,"address":"celestia1ac30y7nav5e427jg53yhflnjrfet7nlza2vzxf","hash":"BF5BD511BFC6A929468FCDE74D5B6695D5CC4D63A40CC12EA7CAAF19301AE9F4","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:17:29-07:00","message":"tx committed"}
{"level":"info","height":8330937,"address":"celestia1yjyx45j0rt9g86ac7pn6rqkj9fd0r86lsn2j9z","hash":"24991D7B6D2AED8E35AB1E1DD614E65194A9650A741DE051A96BFB1C672C2946","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:17:29-07:00","message":"tx committed"}
{"level":"info","height":8330940,"address":"celestia1hg2p339hmx0w8ewf70s5lvlnvx2c4uch75yvwv","hash":"1958F81AA9844933E8ED92CD06A2872EFC4176FA187A8099A588E70C55400736","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:17:29-07:00","message":"tx committed"}
{"level":"info","height":8330940,"address":"celestia1yjw7eqd8ecjwjrkm009qre66hmd3k5gw800xw6","hash":"A9456E146048C878B2C1BBD45F2AD5063173B0CC02604057E454696DE1ABB199","blobs count":"1","total byte size of blobs":1605632,"time":"2025-10-08T17:17:29-07:00","message":"tx committed"}
```

rootulp:
> To avoid this, I propose to introduce --tx-size (blobSize + txOverhead) < maxTxSize, or use --blob as part of fill-blocks command.

I'd rather there be fewer CLI flags. The context behind the original issue is that it's complicated and confusing to use txsim currently to fill blocks. Ideally there's one command that a user doesn't need to configure that fills blocks on a network. The user of this `fill-blocks` command shouldn't have to do math or provide values for a bunch of CLI flags.

zatarx:
Gotcha. ATM `--blob` sets the desired amount of sequences submitted concurrently at a time (in `fill-blocks` it's `maxBytes` / `appconsts.MaxTxSize` by default), and `--share-overhead` is set to 30 to account for share encoding which can also be changed if user needs bigger/smaller tx size. I tested it on mocha (most recent test result) and those blocks where you see 2 submissions had `square_size` of 128.

I guess the default `--blob` value can be increased by some constant to increase the chance of 2 or more txs getting included in the same block to maximize the block `square_size`. What are your thoughts on this?

rootulp:
Yea I'm open to the `fill-blocks` command using a different `--blob` amount if it increases the reliability of filling blocks.

zatarx:
@rootulp, feel free to review. 
I'm thinking of potentially decreasing `--share-overhead` to reduce blob size cause it performs better when it comes to filling blocks.

Measured on 30 blocks for each test:
1. When we have `--share-overhead=30` (default value), we get half of the blocks with 64 OSS and the other half is 128.
2. However, when we it's 50, we get 20 out of 30 blocks full.

test/cmd/txsim/cli.go
```diff
@@ -33,6 +36,10 @@ const (
 	TxsimMnemonic      = "TXSIM_MNEMONIC"
 )
 
+const (
+	FillBlocksDefaultBlobAmt = 16
```
rootulp, line 40: please include a comment describing this constant

test/cmd/txsim/cli.go
```diff
@@ -47,6 +54,7 @@ var (
 	gasLimit                                          uint64
 	gasPrice                                          float64
 	namespaces                                        []string
+	perShareOverheadSize                              int
```
rootulp, line 57: [question] what does this mean? Can this please include a comment explaining it?

test/cmd/txsim/cli.go
```diff
@@ -219,11 +227,65 @@ account that can act as the master account. The command runs until all sequences
...
+
+	fillBlocksCmd := &cobra.Command{
+		Use:   "fill-blocks",
+		Short: "Fill blocks to their max capacity",
+		RunE:  fillBlocksExecute,
+		Long: `
+Fills blocks to near max capacity by submitting batches of transactions.
+Each tx targets a blob size close to maxTxSize - txOverhead by default.
+Batch size (--blob) and share overhead (--share-overhead) can be configured based on the
+desired test load.
```
rootulp, line 240: [question] do these need to be configured for `fill-blocks`? Ideally this command just filled blocks to max capacity and there was no fiddling with extra arguments.

```suggestion
```

test/cmd/txsim/cli.go
```diff
@@ -219,11 +227,65 @@ account that can act as the master account. The command runs until all sequences
...
+`,
+		PostRunE: cmd.RunE,
+	}
+
+	fillBlocksCmd.Flags().IntVar(&perShareOverheadSize, "share-overhead", share.ShareSize-share.ContinuationSparseShareContentSize,
+		"Per-share byte metadata overhead in bytes",
+	)
```
rootulp, line 247: Can this be non-configurable. The motivation for a `fill-blocks` command is that it's confusing to specify so many arguments to fill blocks.

If a user doesn't need to modify this to fill blocks, let's remove it as a param flag.

test/cmd/txsim/cli.go
```diff
@@ -219,11 +227,65 @@ account that can act as the master account. The command runs until all sequences
...
+	govMaxSquareSize := blobParamsResp.Params.GovMaxSquareSize
+	maxSquareSize := min(int(govMaxSquareSize), appconsts.SquareSizeUpperBound)
+
+	maxSquareShares := (maxSquareSize * maxSquareSize)
+	maxBytes := maxSquareShares * share.ShareSize
+
+	if blob == 0 {
+		blob = FillBlocksDefaultBlobAmt
+	}
+	blobSizes = strconv.Itoa(appconsts.MaxTxSize - maxSquareShares*perShareOverheadSize)
```
rootulp, line 274: [blocking] this strategy will only work when the max tx size is the max block bytes which isn't true on any network.

Arabica:
- Max tx size: 8 MiB
- Max block size: 32 MiB

Mocha
- Max tx size: 8 MiB
- Max block size: 32 MiB

Mainnet
- Max tx size: 2 MiB
- Max block size: 8 Mi

test/cmd/txsim/cli.go
zatarx: * it represents protocol metadata bytes size that gets deducted from overall tx size to enable erasure coding down the line (ex. it's a bytes size delta when max tx size is 2MiB tx, and practical max is ~1.88MiB). In other words, the main idea is to deduct `perShareOverheadSize` amount of bytes from each share (512 bytes) in the original square. 

* on it.

test/cmd/txsim/cli.go
zatarx: You are correct, thanks for pointing it out.
Makes no sense this way, it should rely completely on maxTxSize and nothing else.

Future patched strategy:
```
	blobSizes = strconv.Itoa(appconsts.MaxTxSize - appconsts.MaxTxSize / shareSize * perShareOverheadSize)
```

test/cmd/txsim/cli.go
zatarx: my only concern here is if user accidentally runs it on mainnet and runs out of gas funds. Shouldn't it be configurable then? I guess I could set a gas limit as an option.

Otherwise it can be non-configurable.

test/cmd/txsim/cli.go
zatarx: the default values are meant to be optimal for filling blocks, but it depends on the `max tx size : max block size` ratio. If in most cases it's 1:4 and not more, than `--blob` value can be non-configurable. However, it it gets 1:(16+) on other chains, it may be worth increasing.

test/cmd/txsim/cli.go
rootulp: Agreed the practical max on Mainnet is currently ~1.88 MiB and the actual max tx size is 2 MiB but the max tx size varies based on app version. Is it possible to eliminate the `perShareOverheadSize` variable? It would be nice if this command didn't depend on implementation details that are likely to change.

test/cmd/txsim/cli.go
rootulp: Why does this param impact gas funds?

zatarx: When you are running txsim, you need an account to pay for gas as you need to submit txs. The bigger the `--blob n` value, the more gas txsim will have to spend to spin up n txs (each round n txs/sequences get submitted concurrently in a loop).

test/cmd/txsim/cli.go
zatarx: How about I name it txOverhead? So far, I need to have some sort of variable that I can use to adjust the tx size.

If I hardcode it, there's a scenario where if block gets 80% filled, and our tx `i` is too big, the block will not be fully filled. Although OSS in this case will likely be maxed.

test/cmd/txsim/cli.go
rootulp: I suggest trying to remove this param if possible

PR #5723, opened 2025-09-11
perf: optimize hex encoding in DataAvailabilityHeader.String()

lidanserj:
## Overview

Improves performance by using specialized hex encoding functions instead of generic string formatting. The hex.EncodeToString() function is specifically designed for byte-to-hex conversion and avoids the overhead of reflection and format string parsing used by fmt.Sprintf().

The behavior is identical in all cases: both approaches produce the same uppercase hexadecimal string representation of the hash.

- Reduces memory allocations during string conversion
- Improves performance for hash string representation
- Maintains exact same output format and behavior

### DIFF
--- a/pkg/da/data_availability_header.go
+++ b/pkg/da/data_availability_header.go
@@ -2,9 +2,11 @@ package da
 
 import (
 	"bytes"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math"
+	"strings"
 
 	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
 	"github.com/celestiaorg/celestia-app/v6/pkg/wrapper"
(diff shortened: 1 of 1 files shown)


pkg/da/data_availability_header.go
```diff
@@ -79,7 +81,7 @@ func (dah *DataAvailabilityHeader) String() string {
 	if dah == nil {
 		return "<nil DAHeader>"
 	}
-	return fmt.Sprintf("%X", dah.Hash())
```
rootulp, line 82: 1. Can you please add a test case for this String method so that reviewers have confidence this refactor doesn't change behavior?
2. Can you please benchmark the change before and after and include that in the PR description? Without benchmarks, it's not easy for reviewers to determine if this PR has the desired effect.

lidanserj: Closing this PR - benchmarks prove this degrades performance.
Thanks for catching this with your benchmark requirement!

PR #5370, opened 2025-07-30
feat!: update to v6

cmwaters:

PR #5724, opened 2025-09-11
fix: use correct upgrade height when dumping info to disk

damiannolan:
## Overview

The issue here was surfaced from e2e test failures seen on the zk execution ism feature branch, when adding a new module. 
The x/upgrade handler would produce this panic trace:
```
2025-09-09 11:14:35 10:14AM ERR CONSENSUS FAILURE!!! err="failed to load latest version: version of store zkism mismatch root store's version; expected 44 got 0; new stores should be added using StoreUpgrades" module=consensus stack="goroutine 375 [running]:\nruntime/debug.Stack()\n\t/usr/local/go/src/runtime/debug/stack.go:26 +0x5e\ngithub.com/cometbft/cometbft/consensus.(*State).receiveRoutine.func2()\n\t/go/pkg/mod/github.com/celestiaorg/celestia-core@v0.39.2/consensus/state.go:856 +0x46\npanic({0x428a220?, 0xc000690240?})\n\t/usr/local/go/src/runtime/panic.go:792 +0x132\ngithub.com/celestiaorg/celestia-app/v6/app.New({0x53a46c0, 0xc001f7cff0}, {0x53c34e0, 0xc00243a9e8}, {0x0, 0x0}, 0x0, {0x5335b00, 0xc003338300}, {0xc006eabc00, ...})\n\t/celestia-app/app/app.go:507 +0x713f\ngithub.com/celestiaorg/celestia-app/v6/cmd/celestia-appd/cmd.NewAppServer({0x53a46c0, 0xc001f7cff0}, {0x53c34e0, 0xc00243a9e8}, {0x0, 0x0}, {0x5335b00, 0xc003338300})\n\t/celestia-app/cmd/celestia-appd/cmd/app_server.go:15 +0xda\ngithub.com/celestiaorg/celestia-app/v6/multiplexer/abci.(*Multiplexer).startNativeApp(0xc002a1b208)\n\t/celestia-app/multiplexer/abci/multiplexer.go:388 
```

This was due to the app's store loader not being set for the correct height, and thus we do not enter the conditional here and add the module's store key: https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.51.x-celestia/x/upgrade/types/storeloader.go#L14-L17

The x/signal upgrade height was being used for dumping the x/upgrade `UpgradeInfo` to disk. When the `UpgradeInfo` was read subsequently by the new app's upgrade handler, it was using the x/signal upgrade height for setting store loaders (which is off by one).

**This fix needs to be backported to any version which will have subsequent versions adding, updating or removing new module store keys. As the bug itself will present itself running in the `EndBlocker` handler of the embedded multiplexer binary running the "pre-upgrade" app version.**

Shoutout to @chatton for pairing on uncovering this issue and fixing it.

Why it worked on v3 -> v4: https://github.com/celestiaorg/celestia-app/blob/v3.x/app/app.go#L501
I demonstrated the fix on the feature/zk-execution-ism branch to resolve my issue here: https://github.com/celestiaorg/celestia-app/pull/5662#issuecomment-3280735686

Co-authored-by: Cian Hatton <github.qpeyb@simplelogin.fr>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   test-short:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
@@ -17,7 +17,7 @@ jobs:
         timeout-minutes: 10
 
   go-test:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -54,7 +54,7 @@ jobs:
           fi
 
   test:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     needs: go-test
     if: always()
     steps:
@@ -67,7 +67,7 @@ jobs:
           echo "All tests passed"
 
   test-coverage:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
@@ -84,7 +84,7 @@ jobs:
           files: ./coverage.txt
 
   test-race:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
@@ -96,7 +96,7 @@ jobs:
         run: make test-race
 
   test-fuzz:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
@@ -108,7 +108,7 @@ jobs:
         run: make test-fuzz
 
   test-multiplexer:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
--- a/app/errors/nonce_mismatch_test.go
+++ b/app/errors/nonce_mismatch_test.go
@@ -19,7 +19,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// This will detect any changes to the DeductFeeDecorator which may cause a
+// This will detect any changes to the SigVerificationDecorator which may cause a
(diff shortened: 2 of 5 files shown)


app/app.go
```diff
@@ -541,26 +541,27 @@ func (app *App) EndBlocker(ctx sdk.Context) (sdk.EndBlock, error) {
...
-				Height: upgrade.UpgradeHeight + 1, // next block is performing the upgrade.
+				Name:   fmt.Sprintf("v%d", signalUpgrade.AppVersion),
+				Height: upgradeHeight,
 			}
 
 			if err := app.UpgradeKeeper.ScheduleUpgrade(ctx, plan); err != nil {
 				return sdk.EndBlock{}, fmt.Errorf("failed to schedule upgrade: %v", err)
 			}
 
-			if err := app.UpgradeKeeper.DumpUpgradeInfoToDisk(upgrade.UpgradeHeight, plan); err != nil {
```
damiannolan, line 559: This is the bug. Here `upgrade.UpgradeHeight` is the x/signal upgrade, and not the x/upgrade plan height.

app/upgrades.go
```diff
@@ -110,7 +110,7 @@ func (app App) RegisterUpgradeHandlers() {
 
 			sdkCtx.Logger().Info("finished to upgrade", "upgrade-name", upgradeName, "duration-sec", time.Since(start).Seconds())
 
-			return fromVM, nil
+			return app.ModuleManager.RunMigrations(ctx, app.configurator, fromVM)
```
damiannolan, line 113: This was removed in this PR https://github.com/celestiaorg/celestia-app/pull/5370/files#diff-aeec7ddb4cf8b7e9e6e57625770d369cce5f377cc350b42c3aca0427bba8dfbcL123

We should keep this in regardless of if there is any automatic module migrations being run or not, as it will no-op for the latter. 
In the event of adding a new module, this function will also call the `InitGenesis` of that module (which is needed for the work on the feat branch I was doing)

evan-forbes: is this consensus breaking to call this now? I don't think so, correct?

damiannolan: Kind of a weird one - since this is just the upgrade handler. But I would say that all nodes should share this same code for an upgrade, otherwise you could actually have diverging state. 
If nodes are already on v6 (upgraded), then I would say no, not consensus breaking per se. The situation we wouldn't want is this code and the previous code being live in a multiplexer build running the embedded application (pre-upgrade) - then there is a large chance to app hash - but technically only if there's module migrations to run or modules added or removed (which I think there is not lol).

Sorry for the wall of text but, just wanted to be clear!

chatton:
e2e test failures unrelated, due to issue with self hosted runners atm

PR #5726, opened 2025-09-11
chore: set timeout precommit to 1ms and the delayed precommit to 5850ms

rach-id:
## Overview

This PR updates the timeout commit to 1ms and the delayed precommit to 5850ms. The reasoning behind:
- 200ms is wasted time waiting when finalizing the block. It's better to just go ahead and start the new round.
- setting the timeout commit to 1ms allows further optimizations to finalize the block, saving the block, etc, without having to make a breaking change to change them. So if one day we were able to have superfast finalization, no need to be stuck with 200ms and require a network upgrade.
- setting the delayed precommit timeout to 5850ms allows having exactly 6s blocks both for empty blocks and 32mb blocks.

This is from a Talis experiment using the above parameters:

image

image

*Welcome nuclear-watch precision, exactly 6 seconds block times 🕙*

So, it would be good to have these changes merged ASAP so that they make it to mocha.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   test-short:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
@@ -17,7 +17,7 @@ jobs:
         timeout-minutes: 10
 
   go-test:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -54,7 +54,7 @@ jobs:
           fi
 
   test:
-    runs-on: ubuntu-8
+    runs-on: ubuntu-latest
     needs: go-test
     if: always()
     steps:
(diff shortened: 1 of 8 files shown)


evan-forbes:
> Was DelayedPrecommitTimeout introduced recently? I'm wondering why we didn't use it earlier when we tried targeting 6 second blocks in previous upgrades.

yeah we introduced it in the arabica release due to needing it for 20MB/s. we didn't use it earlier due to not wanting to mess with timing related things simply due to being burned in the past and not deeming consistently block times as worth the risk. Instead we just did a bit of work during the timeout commit. then that introduced bugs and didn't pipeline as much as we could

pkg/appconsts/app_consts.go
```diff
@@ -29,8 +29,8 @@ const (
 	TimeoutPrevoteDelta            = time.Millisecond * 500
 	TimeoutPrecommit               = time.Millisecond * 3000
 	TimeoutPrecommitDelta          = time.Millisecond * 500
-	TimeoutCommit                  = time.Millisecond * 100
-	DelayedPrecommitTimeout        = time.Millisecond * 5700
+	TimeoutCommit                  = time.Millisecond
+	DelayedPrecommitTimeout        = time.Millisecond * 5850
```
evan-forbes, line 33: originally this was set for 128MB blocks instead of 32MB

with that under consideration, do we still want to increase this?

rach-id: yes, it's fine to increase given we decreased the timeout commit. Actually, 128mb blocks will either be 6s or more. So, it's fine.

rach-id:
> Is this technically breaking? If yes: do we need the ! in PR title? How will we deploy this to Arabica considering Arabica already upgraded to v6?

It's not breaking. It's just some nodes will start waiting for new blocks/proposing a 100ms earlier and they will precommit 150ms later. The timeout precommit is 3s, so it's fine, not breaking at all

PR #5730, opened 2025-09-12
test(talis): improve droplets list

tzdybal:
Small fix of pagination.

Move header printing outside of the loop for clarity. Adjust the placement of total instance count.


tools/talis/deployment.go
```diff
@@ -415,13 +415,17 @@ func listCmd() *cobra.Command {
 			}
 
 			opts := &godo.ListOptions{}
+
+			cnt := 0
```
rootulp, line 422: [extreme nit] I would've preferred the variable name `count := 0`

evan-forbes: normally I don't comment on var names but agree here as inferring the correct vowls and incorrect one is way to close lolol

PR #5741, opened 2025-09-13
docs: fix incomplete comment for canClimbRight function

galaparm:
## Overview

Add missing condition check (depth > minDepth) to canClimbRight function comment. The function checks both position evenness and minimum depth constraint, but the comment only mentioned the position check

### DIFF
--- a/pkg/inclusion/paths.go
+++ b/pkg/inclusion/paths.go
@@ -30,7 +30,8 @@ func (c coord) climb() coord {
 
 // canClimbRight uses the current position to calculate the direction of the next
 // climb. Returns true if the next climb is right (if the position (index) is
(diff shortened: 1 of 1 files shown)


pkg/inclusion/paths.go
```diff
@@ -30,7 +30,8 @@ func (c coord) climb() coord {
 
 // canClimbRight uses the current position to calculate the direction of the next
 // climb. Returns true if the next climb is right (if the position (index) is
-// even). please see depth and position example map in docs for coord.
+// even) AND the current depth is greater than minDepth. please see depth and
```
rootulp, line 33: fails golangci-lint

PR #5743, opened 2025-09-14
proto: fix comment in MsgUpdateBlobParams

denevertt:
## Overview

Changed `client parameters` to `blob parameters` to correctly reflect that MsgUpdateBlobParams is used to update blob module parameters, not client parameters.


evan-forbes:
closing as this is broken

if you want to reopen after its fixed maybe but pls contribution guidelines

PR #5744, opened 2025-09-14
fix: remove duplicate RPC endpoint in mocha testnet config

lidanserj:
## Overview

Replace duplicate itrocket.net RPC with publicnode.com alternative  to provide redundancy and improve fault tolerance for mocha testnet configuration in docker e2e tests.

Or should I delete it?


evan-forbes:
this is to test state sync, not be secure so it makes sense to use 1

see the contribution guidlines around driveby PRs. if you have a question on if something is an issue then pls open an issue

PR #5776, opened 2025-09-16
feat: ability to submit PFBs in parallel

### LINKED ISSUE #4787: Add the ability for the TxClient to handle parallel tx submission for the user
relying on the sequence coordination amongst the TxClient, mempool, committed, and nodes in the network is often a delicate operation. For users who don't require sequenced submission of blobs, the TxClient has the ability to manage multiple accounts under the hood.

We could extend that by managing feegrant to keep funds in a single account, and by adding a queue. The idea being that a user could blindly submit as many blobs as they want, specify the number of concurrent submitters, then the TxClient would continually submit txs in parallel. 

This has the downside of bloating the state machine, as it would be more efficient to use something like "nonce lanes". However, this approach doesn't require changes to the state machine.

evan-forbes:
## Overview

this PR essentially does the same trick that txsim uses to be able to completely bypass nonce mismatches. It works slightly differently here, instead of having a "sequence" which is a single account that submits specific types of transactions, this is a set of workers that process a queue of transactions.

The tradeoff here is that transactions are not guaranteed to be included in the order that the user submits them, hence the different method. This is equivalent of removing the nonce. The user has always had this ability, this just makes it a lot easier for them as they don't have to think about feegranter or managing multiple accounts.

we could also refactor txsim in the future using this code.

Closes #4787 
Closes https://github.com/celestiaorg/celestia-app/issues/5792


pkg/user/tx_client.go
```diff
@@ -547,7 +569,7 @@ func (client *TxClient) broadcastMulti(ctx context.Context, txBytes []byte, sign
 		go func(conn *grpc.ClientConn) {
 			defer wg.Done()
 
-			resp, err := client.broadcastTxAndIncrementSequence(ctx, conn, txBytes, signer)
+			resp, err := client.broadcastTx(ctx, conn, txBytes, signer)
```
evan-forbes: we were increasing the nonce for each connection that we submitted to!!

this was clearly very broken, and the mock tests we have for this clearly don't check nonce / signature

imo, we should either fix this here or remove entirely.

pkg/user/tx_client.go
```diff
@@ -568,7 +590,13 @@ func (client *TxClient) broadcastMulti(ctx context.Context, txBytes []byte, sign
...
 
 	// Return first successful response, if any
-	if resp, ok := <-respCh; ok {
+	if resp, ok := <-respCh; ok && resp != nil {
+		client.trackTransaction(signer, resp.TxHash, txBytes)
+
+		if err := client.signer.IncrementSequence(signer); err != nil {
+			return nil, fmt.Errorf("increment sequencing: %w", err)
+		}
+
```
evan-forbes: same as above, this fix is purely realted to fixing the increase of the nonce per each connection we were submitting to

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+	}
+}
+
+// processJob handles the actual transaction submission for a job
+func (w *TxWorker) processJob(job *SubmissionJob) {
+	ctx := context.Background()
+
+	// Use the worker's dedicated account to submit the transaction
+	txResponse, err := w.client.SubmitPayForBlobWithAccount(ctx, w.accountName, job.Blobs, job.Options...)
```
evan-forbes: in the future we can do any transaction and not only PFBs, but wanted to try and limit scope here

happy to increase scope here if needed

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,381 @@
...
+		default:
+			close(job.ResultC)
+		}
+	}
+}
+
+// SubmitPayForBlobParallel submits a transaction for parallel processing and returns immediately with a job ID.
+// The result can be retrieved later using GetSubmissionResult.
+// Returns an error only if the parallel pool is not configured or if job submission fails.
+func (client *TxClient) SubmitPayForBlobParallel(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (chan *SubmissionResult, error) {
```
evan-forbes: this will be the entry point for celestia-node cc @walldiss @renaynay 

would you all prefer a channel be returned per PFB or a single channel that all results go through?

the end goal is to add an option to the submit blob api that will add the blob to the queue here and one of the worker accounts will pick it up and submit it.

pkg/user/tx_client.go
```diff
@@ -137,6 +141,27 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
 	}
 }
 
+// WithTxWorkers enables parallel transaction submission with the specified number of worker accounts.
+// Each worker will use a separate account for transaction submission.
+// If workerAccounts is nil or empty, worker accounts will be automatically created and initialized.
```
rootulp: This doesn't seem true based on the body of the function. Maybe its happening elsewhere (?). If not, can remove this line of godoc.

```suggestion
```

pkg/user/tx_client.go
rootulp: > we were increasing the nonce for each connection that we submitted to!!

+1 seems broken

pkg/user/tx_client.go
```diff
@@ -856,6 +884,28 @@ func (client *TxClient) Signer() *Signer {
...
+func (p *ParallelTxPool) Workers() []*TxWorker {
+	return p.workers
+}
+
+// IsStarted returns whether the parallel pool is started
+func (p *ParallelTxPool) IsStarted() bool {
+	p.mtx.RLock()
+	defer p.mtx.RUnlock()
+	return p.started
+}
```
rootulp: [optional] For code organization, I like it when methods are defined in the same file as where the type is defined. In this case, I think these methods could be moved to the file that defines the `ParallelTxPool` type.

rootulp (approved):
LGTM. 

[optional] the txsim nonce incrementing for each connection seems like a bug that we could fix outside this PR if this doesn't get enough approvals or merge quickly.

pkg/user/tx_client.go
evan-forbes: oops agree

original everything was in tx_clinet and I moved stuff to a new file, will move these as well 9a3a09029817813bf94eee78284228f50989ac08

rootulp (approved):

pkg/user/tx_client.go
renaynay: good find.

walldiss:
Idea with auto generation of accounts makes total sense and I think this is nice feature for users. I think this feature would need doc with explanation of how to use new API with examples.  

I have few questions that more UX focused and some are just to clarify design choices:
- Would user need to manually store generated accounts after first auto-generation? (restart scenario)
- How user will keep track of created accounts? How user can close them if those accounts no longer needed?
- What would be the way to migrate such accounts to another node? Would it require manual export?
---
- Would this work with signed blobs? (BLobV1)
---
- How many accounts will be "enough"? Would be nice to provide guidance to users 
- Do you plan any feedback / telemetry around worker pool being exhausted?

walldiss (approved):
Questions above are non-blocking. Feature looks good!

cmwaters:
I haven't looked into the implementation too closely but the approach is sound and is a neat way of doing unordered transactions without having to implement something like sequence lanes.

I would however pause going through with rolling out this feature so we can:
- First, more thoroughly validate the need across our main users (@renaynay)
- Second, quantify the throughput difference between this approach and the current one
- Third, consider how the design would compare if it resided on the node side. Also consider whether there are other solutions we'd want to support (like the tx lifetimes that Solana does)

If Rise are interested in this, perhaps we can ask them to use a branch and use that as an opportunity to get feedback

cmwaters:
I can answer some of the above questions:

> Would user need to manually store generated accounts after first auto-generation? (restart scenario)

No, the accounts would be stored in the keyring (by default under a specific naming scheme) such that upon restart, we can loop through the keyring and see what accounts are "worker accounts". If I'm not wrong, I think there is a slight bug whereby we should remove those account if we fail to actually set up the fee grant with them

> How user will keep track of created accounts? How user can close them if those accounts no longer needed?

I think, using the above construction, it wouldn't be too hard to have a go method that returns all the currently set up accounts. You can't remove accounts from state but you could revoke the fee grant and delete the keys. This isn't supported in the original PR but neither is it too important.

> What would be the way to migrate such accounts to another node? Would it require manual export?

You would need to migrate the keyring over. All the necessary information is there.

> Would this work with signed blobs? (BLobV1)

Yes, it could also work with normal transactions as well

> How many accounts will be "enough"? Would be nice to provide guidance to users

That is roughly a function of how many transactions they would like to submit per block. If they want to submit two transactions a block, then three accounts would be sufficient (as my recommendation)

> Do you plan any feedback / telemetry around worker pool being exhausted?

As in when the queue becomes fills up faster than the workers can submit blobs. It could be added later on. I think teams will know roughly the frequency they want to submit and can allocate workers accordingly.

pkg/user/mock_tx_server_test.go
```diff
@@ -24,9 +24,23 @@ type mockTxServer struct {
...
 func (m *mockTxServer) TxStatus(ctx context.Context, req *tx.TxStatusRequest) (*tx.TxStatusResponse, error) {
+	return m.txStatusHandler(ctx, req)
+}
+
+func (m *mockTxServer) BroadcastTx(ctx context.Context, req *sdktx.BroadcastTxRequest) (*sdktx.BroadcastTxResponse, error) {
+	return m.broadcastHandler(ctx, req)
+}
+
+// defaultTxStatusHandler implements the original default behavior for TxStatus
+func (m *mockTxServer) defaultTxStatusHandler(ctx context.Context, req *tx.TxStatusRequest) (*tx.TxStatusResponse, error) {
```
evan-forbes, line 47: this change is not purely a refactor of the new mocks, this is required as the previous mocks were not flexible enough to suit these tests

pkg/user/account.go
```diff
@@ -81,3 +82,21 @@ func QueryAccount(ctx context.Context, conn *grpc.ClientConn, registry codectype
 	accNum, seqNum = acc.GetAccountNumber(), acc.GetSequence()
 	return accNum, seqNum, nil
 }
+
+// QueryAccountBalance fetches the balance of a specific denomination for an account.
+func QueryAccountBalance(ctx context.Context, conn *grpc.ClientConn, registry codectypes.InterfaceRegistry, address types.AccAddress, denom string) (types.Coin, error) {
```
evan-forbes: used to determine if we need to send a single utia to an account to create it or if it already exists

pkg/user/e2e_test.go
```diff
@@ -76,3 +78,121 @@ func TestConcurrentTxSubmission(t *testing.T) {
...
+
+	t.Log("Successfully verified parallel pool shutdown behavior")
+
+	// Part 3: Test restart scenario - create new client and verify existing accounts are reused
+	t.Log("Testing restart scenario with existing worker accounts...")
+
+	// Create a new TxClient using the same keyring (simulating restart)
+	// Use the same default account as the first client to avoid using a worker account as default
+	originalDefaultAccount := txClient.DefaultAccountName()
+	txClient2, err := user.SetupTxClient(ctx.GoContext(), ctx.Keyring, ctx.GRPCClient, encCfg, user.WithTxWorkersNoInit(numWorkers*3), user.WithDefaultAccount(originalDefaultAccount))
```
evan-forbes: here's where we test re-initialize the txclient and add more workers

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,531 @@
...
+	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
+)
+
+// SubmissionJob represents a transaction submission task for parallel processing
+type SubmissionJob struct {
+	ID       string
+	Blobs    []*share.Blob
+	Options  []TxOption
+	Ctx      context.Context
+	ResultsC chan SubmissionResult
```
evan-forbes, line 24: so I tried using a single channel in node, but this was way easier to add in node since we can treat SubmitPFB there essentially the same

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,531 @@
...
+	Signer     string
+	TxResponse *TxResponse
+	Error      error
+}
+
+// ParallelTxPool manages parallel transaction submission
+type ParallelTxPool struct {
+	mtx         sync.RWMutex
+	client      *TxClient
+	jobQueue    chan *SubmissionJob
```
evan-forbes, line 37: random comment, but the cool thing about using a queue and single worker is that if users need to get txs included in order, that is as close as get

still needs a flup if we want to provide guarantees around it, but generally should work well since all workers wait to submit a new tx until after they get a confirmation

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,531 @@
...
+	ID       string
+	Blobs    []*share.Blob
+	Options  []TxOption
+	Ctx      context.Context
+	ResultsC chan SubmissionResult
+}
+
+// SubmissionResult contains the result of a parallel transaction submission
+type SubmissionResult struct {
+	Signer     string
```
evan-forbes, line 29: node doesn't use this , but we are returning the signer in the result

we can also include the blob commitments in a flup (was a larger change to TxClient)

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,531 @@
...
+
+	select {
+	case <-jobCtx.Done():
+		job.ResultsC <- SubmissionResult{Signer: w.address, Error: jobCtx.Err()}
+		return
+	case <-w.ctx.Done():
+		job.ResultsC <- SubmissionResult{Signer: w.address, Error: w.ctx.Err()}
+		return
+	default:
+	}
```
evan-forbes, line 207: wasn't sure if this is necessary or not, but we will cancel if the ctx passed to SubmitPFB is cancelled or if the total txclient is cancelled.

evan-forbes (commented):
there are likely more scenarios where the feegrant could mess up. for example if one or more of the worker grants get revoked. 

However we have the major cases working (initialization, restarting, adding more workers), so imo we can flup the rest.

pkg/user/tx_client.go
```diff
@@ -138,6 +142,44 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
...
+
+	return func(c *TxClient) {
+		pool := newParallelTxPool(c, numWorkers, true)
+		c.parallelPool = pool
+	}
+}
+
+// WithTxWorkersNoInit enables parallel transaction submission without automatic initialization.
+// InitializeWorkerAccounts must be called manually when using this option.
+func WithTxWorkersNoInit(numWorkers int) Option {
```
evan-forbes: also forgot to mention that we can initialize automatically or manually later using the method

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,528 @@
...
+		pool.workers[i] = worker
+	}
+
+	return pool
+}
+
+// NewParallelTxPool creates a new parallel transaction submission pool.
+func NewParallelTxPool(client *TxClient, numWorkers int, initialize bool) *ParallelTxPool {
+	return newParallelTxPool(client, numWorkers, initialize)
+}
```
rootulp: [silly question] this public method seems like a no-op wrapper around the private method. Could we eliminate these lines and make the private method public?

evan-forbes: yeah good point we could and it would save a few lines. perhaps even better is to just not export it.

will update if there is other feedback

evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5776/commits/4c02643ce5f19a279ee00bd2f667c1fc971fb3f9

pkg/user/e2e_test.go
```diff
@@ -1,78 +1,136 @@
...
 	"github.com/celestiaorg/celestia-app/v6/test/util/blobfactory"
 	"github.com/celestiaorg/celestia-app/v6/test/util/random"
 	"github.com/celestiaorg/celestia-app/v6/test/util/testnode"
 	"github.com/celestiaorg/go-square/v3/share"
-	"github.com/cometbft/cometbft/config"
 	"github.com/stretchr/testify/require"
 )
 
-func TestConcurrentTxSubmission(t *testing.T) {
+func TestParallelTxSubmission(t *testing.T) {
```
evan-forbes, line 20: this test is redundant now that we are only submitting blobs one at a time. Therefore I'm using the shorter test that also covers restarting txclient / increasing worker count.

evan-forbes: also it took a minute to execute on my machine

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+
+// SubmissionResult contains the result of a parallel transaction submission
+type SubmissionResult struct {
+	Signer     string
+	TxResponse *TxResponse
+	Error      error
+}
+
+// txQueue manages parallel transaction submission
+type txQueue struct {
```
evan-forbes, line 35: per request, all unnecessary stuff was unexported. also renamed this since we use it for most tx submission now.

pkg/user/tx_client.go
```diff
@@ -251,18 +289,60 @@ func SetupTxClient(
...
+	return result.TxResponse, nil
+}
+
+// SubmitJob submits a job to the tx queue for parallel processing
+func (client *TxClient) SubmitJob(job *SubmissionJob) {
+	if client.txQueue != nil {
+		client.txQueue.submitJob(job)
+	} else {
+		job.ResultsC <- SubmissionResult{Error: errors.New("tx queue not configured")}
+	}
```
evan-forbes, line 346: Left this as an easter egg for power users who want to insert their own txs directly into the queue w/o blocking on submitPFB

this allows for using a single results channel for instance

pkg/user/tx_client.go
```diff
@@ -251,18 +289,60 @@ func SetupTxClient(
...
 
 // SubmitPayForBlob forms a transaction from the provided blobs, signs it, and submits it to the chain.
 // TxOptions may be provided to set the fee and gas limit.
+// This method uses the tx queue infrastructure and blocks until the transaction is confirmed.
 func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error) {
-	resp, err := client.BroadcastPayForBlob(ctx, blobs, opts...)
-	if err != nil {
-		return nil, err
+	if client.txQueue == nil {
+		return nil, errors.New("tx queue not configured")
```
evan-forbes, line 312: NOTE: per request, there is no long a SubmitPFBParallel method, now there is only a SubmitPFB method. We are forcing users to use the queue when submitting blobs via this method.

In order to get around this, they need to call BroadcastPayForBlobWithAccount.

cc @cmwaters @renaynay

pkg/user/account.go
renaynay: codec is unnecessary param

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	"github.com/celestiaorg/celestia-app/v6/pkg/appconsts"
+	"github.com/celestiaorg/go-square/v3/share"
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	bank "github.com/cosmos/cosmos-sdk/x/bank/types"
+)
+
+// SubmissionJob represents a transaction submission task for parallel processing
+type SubmissionJob struct {
```
renaynay, line 20: all of these can be private

renaynay: including the fields

pkg/user/tx_client.go
```diff
@@ -885,6 +966,53 @@ func (client *TxClient) Signer() *Signer {
 	return client.signer
 }
 
+// StartTxQueue starts the tx queue if not already started
+func (client *TxClient) StartTxQueue(ctx context.Context) error {
```
renaynay: unused - can be removed - start is managed by construct + setup of `txClient`

pkg/user/tx_client.go
```diff
@@ -885,6 +966,53 @@ func (client *TxClient) Signer() *Signer {
...
+
+// StopTxQueue stops the tx queue if running
+func (client *TxClient) StopTxQueue() {
+	if client.txQueue != nil {
+		client.txQueue.stop()
+	}
+}
+
+// IsTxQueueStarted returns whether the tx queue is started
+func (client *TxClient) IsTxQueueStarted() bool {
```
renaynay: private

pkg/user/tx_client.go
```diff
@@ -885,6 +966,53 @@ func (client *TxClient) Signer() *Signer {
...
+	if client.txQueue == nil {
+		return false
+	}
+	return client.txQueue.isStarted()
+}
+
+// TxQueueWorkerCount returns the number of workers in the tx queue
+func (client *TxClient) TxQueueWorkerCount() int {
```
renaynay, line 1065: all queue-related methods pls private

pkg/user/tx_client.go
```diff
@@ -885,6 +966,53 @@ func (client *TxClient) Signer() *Signer {
...
+	if client.txQueue == nil {
+		return 0
+	}
+	return len(client.txQueue.workers)
+}
+
+// TxQueueWorkerAddress returns the address for the worker at the given index
+func (client *TxClient) TxQueueWorkerAddress(index int) string {
```
renaynay, line 1073: also these methods are test-related only so not even sure they deserve to exist as client and its fields are directly accessible by tests anyway

pkg/user/tx_client.go
```diff
@@ -861,7 +941,8 @@ func (client *TxClient) getAccountNameFromMsgs(msgs []sdktypes.Msg) (string, err
 	return record.Name, nil
 }
 
-// trackTransaction tracks a transaction in the tx client's local tx tracker.
+// trackTransaction tracks a transaction without acquiring the mutex.
+// This should only be called when the caller already holds the mutex.
```
renaynay, line 1014: nit: would be easier + cleaner to just lock inside here instead of on caller-side.

pkg/user/tx_client.go
```diff
@@ -570,7 +650,7 @@ func (client *TxClient) submitToMultipleConnections(ctx context.Context, txBytes
 	close(errCh)
 
 	// Return first successful response, if any
-	if resp, ok := <-respCh; ok {
+	if resp, ok := <-respCh; ok && resp != nil {
```
renaynay, line 678: curious -- how did you come up on the possibility that resp could be nil even if sent through respCh?

pkg/user/e2e_test.go
```diff
@@ -1,78 +1,136 @@
...
+	close(errCh)
+
+	// Check for any errors
+	for err := range errCh {
+		require.NoError(t, err)
 	}
+
+	t.Logf("Successfully submitted %d parallel transactions", numJobs)
+
+	// Part 2: Test shutdown behavior and verify pool can be stopped properly
```
ninabarbakadze, line 75: mb I'm missing something but can these be t.Run() instead of logs+comments?

pkg/user/tx_client.go
renaynay: What is the benefit of bypassing SubmitPFB here? I would privatise this please

renaynay: and even better, just remove entirely as it's a wrapper method for txqueue submission anyway

pkg/user/tx_client.go
renaynay: User would need to actually call `SubmitPayForBlobWithAccount`, but yes, this is the only method that would bypass the blocking queue entirely.

pkg/user/tx_client.go
```diff
@@ -251,18 +289,60 @@ func SetupTxClient(
...
 }
 
 // SubmitPayForBlob forms a transaction from the provided blobs, signs it, and submits it to the chain.
 // TxOptions may be provided to set the fee and gas limit.
+// This method uses the tx queue infrastructure and blocks until the transaction is confirmed.
 func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error) {
-	resp, err := client.BroadcastPayForBlob(ctx, blobs, opts...)
-	if err != nil {
-		return nil, err
+	if client.txQueue == nil {
```
renaynay, line 311: How can this be possible though? I thought now we should have `txQueue` constructed by default in `txClient` -- hitting this case here would indicate borked construction of `txClient`

pkg/user/tx_client.go
```diff
@@ -251,18 +289,60 @@ func SetupTxClient(
...
+	job := &SubmissionJob{
+		Blobs:    blobs,
+		Options:  opts,
+		Ctx:      ctx,
+		ResultsC: resultsC,
+	}
+
+	client.SubmitJob(job)
+
+	// Block waiting for the result
```
renaynay, line 331: select here and wait on ctx deadline exceeded as well pls 🙏🏻

pkg/user/tx_client.go
```diff
@@ -251,18 +289,60 @@ func SetupTxClient(
...
+	if client.txQueue == nil {
+		return nil, errors.New("tx queue not configured")
 	}
 
-	return client.ConfirmTx(ctx, resp.TxHash)
+	if !client.txQueue.started.Load() {
+		return nil, errors.New("tx queue not started")
+	}
+
+	resultsC := make(chan SubmissionResult, 1)
```
renaynay, line 319: who responsible for closing this?

pkg/user/tx_client.go
```diff
@@ -199,9 +236,10 @@ func NewTxClient(
 		opt(txClient)
 	}
 
-	// Sanity check to ensure we don't have more than 3 connections
-	if len(txClient.conns) > 3 {
-		txClient.conns = txClient.conns[:3]
+	// Always create a tx queue with at least 1 worker (the default account)
+	// unless already configured by WithTxWorkers option
+	if txClient.txQueue == nil {
```
renaynay, line 244: how would it be not nil here?

pkg/user/tx_client.go
```diff
@@ -251,18 +289,60 @@ func SetupTxClient(
...
 	}
 
-	return NewTxClient(encCfg.Codec, signer, conn, encCfg.InterfaceRegistry, options...)
+	txClient, err := NewTxClient(encCfg.Codec, signer, conn, encCfg.InterfaceRegistry, options...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Always start the tx queue
+	if txClient.txQueue != nil {
```
renaynay: same here, we should fail out if txQueue is somehow missing - i'm even fine with letting it panic if calling start on nil txQueue as constructor should always construct + attach it by default.

pkg/user/parallel_tx_submission.go
renaynay: yea let's privatise SubmissionResult as it's unnecessary outside of `txClient` and we'll deal with the actual tx response stuff in the FLUP

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+		}
+	}
+}
+
+// processJob handles the actual transaction submission for a job
+func (w *txWorker) processJob(job *SubmissionJob, workerCtx context.Context) {
+	jobCtx := job.Ctx
+	if jobCtx == nil {
+		jobCtx = context.Background()
+	}
```
renaynay, line 197:
```suggestion
```

jobCtx should never be nil, ctx required from caller

pkg/user/tx_client.go
ninabarbakadze, line 244: [question] do we want to do this even for single blob submissions?

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+
+	select {
+	case <-jobCtx.Done():
+		job.ResultsC <- SubmissionResult{Signer: w.address, Error: jobCtx.Err()}
+		return
+	case <-workerCtx.Done():
+		job.ResultsC <- SubmissionResult{Signer: w.address, Error: workerCtx.Err()}
+		return
+	default:
```
renaynay, line 206: we don't need select here, you can check for ctx errors from both job + worker ctx at top of method and short-circuit out of job processing. 

ctx errors will be caught by select in `SubmitPayForBlob` and propagated back to user anyway.

pkg/user/tx_client.go
ninabarbakadze, line 311: [question] how would we hit this if we are always starting the queue during initialization

also, maybe we can use IsTxQueueStarted method

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+
+	// Send result back through the job-specific results channel
+	job.ResultsC <- result
+}
+
+// InitializeWorkerAccounts creates and initializes all worker accounts for parallel submission.
+// It creates the accounts in the keyring if they don't exist, funds them with a small balance,
+// and sets up fee grants so the main account pays for transaction fees.
+// This method should be called after TxClient creation but before parallel submissions.
+func (client *TxClient) InitializeWorkerAccounts(ctx context.Context) error {
```
renaynay: also private pls

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	if client.txQueue == nil {
```
renaynay: same problem here - this shouldn't happen

pkg/user/tx_client.go
ninabarbakadze: we could maybe call StartTxQueue here

pkg/user/tx_client.go
ninabarbakadze, line 1073: [question] why do we need this?

pkg/user/tx_client.go
```diff
@@ -885,6 +966,53 @@ func (client *TxClient) Signer() *Signer {
...
+// StartTxQueue starts the tx queue if not already started
+func (client *TxClient) StartTxQueue(ctx context.Context) error {
+	if client.txQueue == nil {
+		return errors.New("tx queue not configured")
+	}
+	return client.txQueue.start(ctx)
+}
+
+// StopTxQueue stops the tx queue if running
+func (client *TxClient) StopTxQueue() {
```
renaynay: Not necessary - stop is controlled by top-level context cancellation passed in by user in `SetupTxClient` -- all worker routines will exit on ctx cancellation so a manual call to `stop` is not necessary. 

`txQueue` could also listen for top level context cancellation and change its initialisation/started status to stopped.

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	Error      error
+}
+
+// txQueue manages parallel transaction submission
+type txQueue struct {
+	mtx         sync.RWMutex
+	client      *TxClient
+	jobQueue    chan *SubmissionJob
+	workers     []*txWorker
+	started     atomic.Bool
```
renaynay, line 39: Nit - I would prefer if there were one `started/initialised` bool rather than two. 

`txQueue` should not be considered ready for use if it is not `initialised` so let's just use that one please and `started` doesn't really have a use then.

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	job.ResultsC <- result
+}
+
+// InitializeWorkerAccounts creates and initializes all worker accounts for parallel submission.
+// It creates the accounts in the keyring if they don't exist, funds them with a small balance,
+// and sets up fee grants so the main account pays for transaction fees.
+// This method should be called after TxClient creation but before parallel submissions.
+func (client *TxClient) InitializeWorkerAccounts(ctx context.Context) error {
+	if client.txQueue == nil {
+		return errors.New("tx queue not configured - use WithTxWorkers option")
```
ninabarbakadze: return errors.New("tx queue not configured - use WithTxWorkers option")
```suggestion
		return errors.New("tx queue not configured, use WithTxWorkers option")
```

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	}
+
+	// No work required if we've already initialized all workers.
+	if client.txQueue.initialized.Load() {
```
ninabarbakadze: same here I think tx client exposes methods on this

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	// Skip the first worker (index 0) as it always uses the existing signer account
+	var workersToInit []*txWorker
+	var workersToLoad []*txWorker
+	for i, worker := range client.txQueue.workers {
+		if i == 0 {
+			// Skip first worker - it uses existing account
+			continue
+		}
+
+		// Check if account exists in signer
```
ninabarbakadze, line 255: CheckAccountLoaded from tx client should work here I think

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+// SubmissionResult contains the result of a parallel transaction submission
+type SubmissionResult struct {
+	Signer     string
+	TxResponse *TxResponse
+	Error      error
+}
+
+// txQueue manages parallel transaction submission
+type txQueue struct {
+	mtx         sync.RWMutex
```
renaynay: also don't really know the value of this mutex.

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+			msgs = append(msgs, feegrantMsg)
+			totalGasLimit += FeegrantGasLimit
+		}
+	}
+
+	return msgs, totalGasLimit, nil
+}
+
+// loadWorkerAccount loads an existing account from keyring into the signer
+func (client *TxClient) loadWorkerAccount(worker *txWorker) error {
```
ninabarbakadze, line 340: CheckAccountLoaded could be good here too?

renaynay (commented):
API is good now, just some general comments around the `txQueue`

My biggest comment is to omit all methods not directly called by anything in code (so ones that are only used by tests) as they add unnecessary bulk. 

Also `txQueue` lifecycle management comments

Thank you @evan-forbes ! 🙏🏻

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+
+	msgs := make([]sdktypes.Msg, 0, len(workers)*2) // Each worker needs up to 2 msgs: send + feegrant
+	totalGasLimit := uint64(0)
+
+	masterAddress := client.defaultAddress
+
+	// Create send messages for funding accounts that need funding
+	for _, worker := range workers {
+		// Get worker address
+		record, err := client.signer.keys.Key(worker.accountName)
```
ninabarbakadze, line 429: client.Signer.Account() could probably work here since the worker seems to be loaded into signer

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	// Add the worker accounts to the signer
+	for _, worker := range workers {
+		record, err := client.signer.keys.Key(worker.accountName)
+		if err != nil {
+			return fmt.Errorf("failed to get worker account %s from keyring: %w", worker.accountName, err)
+		}
+
+		workerAddress, err := record.GetAddress()
+		if err != nil {
+			return fmt.Errorf("failed to get address for worker account %s: %w", worker.accountName, err)
```
ninabarbakadze, line 484: Similar logic seems to be written twice in the function maybe we can reduce duplication

pkg/user/parallel_tx_submission.go
evan-forbes: I wanted to provide a way to directly insert things into the queue.

imo we should keep these.

pkg/user/tx_client.go
evan-forbes: cause the multi connection thing originally fixed in this PR throws nil if one connection fails

pkg/user/parallel_tx_submission.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5776#discussion_r2389712674

pkg/user/tx_client.go
```diff
@@ -140,6 +150,31 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
 	}
 }
 
+// WithTxWorkers enables parallel transaction submission with the specified number of worker accounts.
+// Worker accounts are automatically generated with hardcoded names unless numWorkers is 1, in which case
+// the existing default account is used.
+// Workers are initialized automatically when SetupTxClient is called.
+func WithTxWorkers(numWorkers int) Option {
```
renaynay, line 159: please move to tx_options.go 🙏🏻

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,497 @@
...
+	Signer     string
+	TxResponse *TxResponse
+	Error      error
+}
+
+// txQueue manages parallel transaction submission
+type txQueue struct {
+	mtx         sync.RWMutex
+	client      *TxClient
+	jobQueue    chan *SubmissionJob
```
renaynay, line 37: jobQueue needs close when lifecycle stopped

pkg/user/tx_client.go
evan-forbes: resolving as this is unchanged in this PR I just added this comment as that was causing me issues debugging

pkg/user/tx_client.go
```diff
@@ -140,6 +150,31 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
...
+
+	return func(c *TxClient) {
+		pool := newTxQueue(c, numWorkers)
+		c.txQueue = pool
+	}
+}
+
+// WithParallelQueueSize sets the buffer size for the parallel submission job queue.
+// Default is 100 if not specified.
+func WithParallelQueueSize(size int) Option {
```
renaynay, line 172: I prefer not exposing this for now - also I think default 100 is a lot to buffer - that's 1 tx every 60ms. Should we pick a more sane value?

renaynay (commented):
sry some forgotten ones

pkg/user/parallel_tx_submission.go
evan-forbes: flup https://github.com/celestiaorg/celestia-app/issues/5907

pkg/user/tx_client.go
evan-forbes: this is a developer focused struct, so I wanted access to this for debugging etc

pkg/user/tx_client.go
evan-forbes: I don't think so since we will also use this in place of txsim to spam a chain.

I think there is more value in keeping this option considering a too low value will halt. 

I'll defer to you all on what is acceptable for light clients, but then you all can set the default in node with this option.

pkg/user/tx_client.go
evan-forbes: those are options for txs such as gas etc, I think we should keep these here.

pkg/user/tx_client.go
evan-forbes: yes, this was feedback from @renaynay where we should only have a single api endpoint

pkg/user/e2e_test.go
evan-forbes: no we're purposefully creating a new txqueue to test for re-initialization / restarting works

pkg/user/tx_client.go
evan-forbes: aren't we using start below? or maybe this got updated since then?

the if nil is just defensive, I didn't see value in removing it so kept it

pkg/user/tx_client.go
evan-forbes: ? aren't we doing that below this line? maybe this updated since you viewed it not sure

pkg/user/parallel_tx_submission.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5776/commits/aa0eb99f48ad7ec2b1023aa4012f009ab36c766f

pkg/user/parallel_tx_submission.go
evan-forbes: not sure on this one anymore since the api was updated

pkg/user/parallel_tx_submission.go
evan-forbes: removed good call it was vestigial

pkg/user/tx_client.go
evan-forbes: these make tests a lot easier so I just renamed them with a Test on the end

added to a flup as well https://github.com/celestiaorg/celestia-app/issues/5907

pkg/user/parallel_tx_submission.go
evan-forbes: yeah true these are likely vestigial since the refactor, but I didn't see the harm in keeping them

pkg/user/parallel_tx_submission.go
evan-forbes: this is a developer focused struct so I think we should keep this exported

added to a flup anyway https://github.com/celestiaorg/celestia-app/issues/5907

pkg/user/tx_client.go
evan-forbes: added a testing method in its place as this makes it easier to test

evan-forbes: also added as a flup

pkg/user/tx_client.go
evan-forbes: same response as below https://github.com/celestiaorg/celestia-app/pull/5776#discussion_r2391453597

Wondertan:
Is this only for PFBs or for any tx? 

My current read is that this is only for PFBs. It would be great to have support for PFFs too, by allowing any sdk.Msg to be sent in this way.  PFFs are not that far as it may seem, and we don't want to bottleneck the fibre by nounce sequentially as well, including for benchmarking

evan-forbes:
@Wondertan this is only for PFBs atm to limit scope, but we have a flup https://github.com/celestiaorg/celestia-app/issues/5902

eventually it will be for both :+1:

pkg/user/tx_client.go
evan-forbes: added a close above to be explicit, but previously just the garbage collector

pkg/user/tx_client.go
evan-forbes: upon a second review we shouldn't actually check that here as it gets checked here https://github.com/celestiaorg/celestia-app/blob/90ce66d1e69921a1f11445583a9d96673fcd69ad/pkg/user/parallel_tx_submission.go#L164-L168

pkg/user/tx_client.go
evan-forbes: so that we can have a txsim equivalent without having to wrap a SubmitPayForBlob and then recombine all the channels

imo we should still keep this

pkg/user/tx_client.go
```diff
@@ -162,6 +197,8 @@ type TxClient struct {
 	// that was submitted to the chain
 	txTracker           map[string]txInfo
 	gasEstimationClient gasestimation.GasEstimatorClient
+	// txQueue manages parallel transaction submission when enabled
```
rootulp, line 200: [nit] it seems like the txQueue is always created in `NewTxClient` so it doesn't seem possible to _disable_
```suggestion
	// txQueue manages parallel transaction submission
```

pkg/user/tx_client.go
```diff
@@ -142,6 +152,31 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
...
+	}
+
+	return func(c *TxClient) {
+		pool := newTxQueue(c, numWorkers)
+		c.txQueue = pool
+	}
+}
+
+// WithParallelQueueSize sets the buffer size for the parallel submission job queue.
+// Default is 100 if not specified.
```
rootulp, line 171: The function body doesn't have a default of 100 so I don't understand that comment.


```suggestion
```

pkg/user/tx_client.go
```diff
@@ -142,6 +152,31 @@ func WithAdditionalCoreEndpoints(conns []*grpc.ClientConn) Option {
 	}
 }
 
+// WithTxWorkers enables parallel transaction submission with the specified number of worker accounts.
+// Worker accounts are automatically generated with hardcoded names unless numWorkers is 1, in which case
+// the existing default account is used.
+// Workers are initialized automatically when SetupTxClient is called.
```
rootulp, line 158: > // Workers are initialized automatically when SetupTxClient is called.

I don't see where that happens in `SetupTxClient`. Should this line be removed?


```suggestion
```

pkg/user/account.go
```diff
@@ -81,3 +82,21 @@ func QueryAccount(ctx context.Context, conn *grpc.ClientConn, registry codectype
 	accNum, seqNum = acc.GetAccountNumber(), acc.GetSequence()
 	return accNum, seqNum, nil
 }
+
+// QueryAccountBalance fetches the balance of a specific denomination for an account.
+func QueryAccountBalance(ctx context.Context, conn *grpc.ClientConn, _ codectypes.InterfaceRegistry, address types.AccAddress, denom string) (types.Coin, error) {
```
rootulp, line 87: [question] does the third argument need to be provided to this function if it is unused?


```suggestion
func QueryAccountBalance(ctx context.Context, conn *grpc.ClientConn, address types.AccAddress, denom string) (types.Coin, error) {
```

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,505 @@
...
+}
+
+const (
+	defaultParallelQueueSize = 100
+)
+
+func newTxQueue(client *TxClient, numWorkers int) *txQueue {
+	pool := &txQueue{
+		client:   client,
+		jobQueue: make(chan *SubmissionJob, defaultParallelQueueSize),
```
rootulp, line 62: [nit] it looks like `defaultParallelQueueSize` can't be overridden so IMO we could remove the `default` from the name.

In other words, rename: `defaultParallelQueueSize` to `parallelQueueSize`

pkg/user/parallel_tx_submission_test.go
```diff
@@ -0,0 +1,520 @@
...
+		require.Equal(t, abci.CodeTypeOK, result.TxResponse.Code)
+	}
+
+	// Verify that multiple workers were used by checking sequence numbers
+	var totalSequence uint64
+	for i := 0; i < client.TxQueueWorkerCount(); i++ {
+		accountName := client.TxQueueWorkerAccountName(i)
+		totalSequence += client.Signer().Account(accountName).Sequence()
+	}
+	require.Equal(t, uint64(len(results)), totalSequence)
```
rootulp, line 340: [no change needed] I don't really understand why the totalSequence should match the `uint64(len(results))` 🤷 

Does this assume that all the worker accounts start with sequence `0`? Seems like a safe assumption but I suppose this test would break if the default account (or a worker account) submits one transaction before this test is run.

rootulp (approved):

pkg/user/tx_client.go
rootulp: Maybe this comment meant to say that the WithTxWorkers option should be provided as an option to the SetupTxClient invocation?

cmwaters:
@evan-forbes here, we use the existence of keys as a way of determining whether the fee grant has been set up. To be properly atomic, we should remove the keys if there are certain failures in the initialisation process. If we don't, the tx client may assume a worker account is initialized when it actually isn't. 

Given this is merged, can we perhaps open an issue to follow up with this.

pkg/user/parallel_tx_submission.go
```diff
@@ -0,0 +1,505 @@
...
+	for i := range numWorkers {
+		var accountName, address string
+
+		if i == 0 {
+			// First worker uses the current default account
+			accountName = client.DefaultAccountName()
+			address = client.DefaultAddress().String()
+		} else {
+			// Additional workers use generated account names
+			accountName = fmt.Sprintf("parallel-worker-%d", i)
```
cmwaters, line 76: Can this be a public method or constant? I think we should make it possible for users to be able to use this to get the address of their account from their keyring

evan-forbes: see the GetWorkerAddress method

PR #5778, opened 2025-09-17
Reduce nmt allocs in process proposal in square root calculation

mcrakhman:
The changes implement a tree pooling optimization for NMTs to improve performance of `ProcessProposal`. A fixed-size pool of reusable NMT instances is created with pre-allocated buffers to reduce memory allocations. The pool is integrated into calculating roots of the square, allowing trees and their buffers to be acquired and released for reuse rather than creating new instances for each operation.


app/process_proposal.go
```diff
@@ -142,7 +142,7 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 		return reject(), nil
 	}
 
-	eds, err := da.ExtendShares(share.ToBytes(dataSquare))
+	eds, err := da.ExtendSharesWithTreePool(share.ToBytes(dataSquare), app.TreePool().GetTreePool(uint64(dataSquare.Size())))
```
rach-id: we do this in prepare proposal as well, can we call this there too?

rach-id: also, I think it would make sense to do this only for certain block sizes like >= 128mb or so. I'm afraid the overhead of calling this for small blocks can add an extra overhead instead of speeding things up. Do we have numbers for each block size?

app/process_proposal.go
rach-id, line 1: also to confirm are using the `FastRoot` in:
```
dah, err := da.NewDataAvailabilityHeader(eds)
```
?

mcrakhman: Correct

app/process_proposal.go
mcrakhman: I don't think we should have too large of an overhead (but we should test of course):

The allocations happen here
```go
for i := 0; i < size; i++ {
		tree := NewErasuredNamespacedMerkleTree(squareSize, 0, opts...)
		treePtr := &tree
		treePtr.pool = pool
		entrySize := share.ShareSize + share.NamespaceSize
		treePtr.buffer = make([]byte, 2*squareSize*uint64(entrySize))
		pool.availableNMTs <- treePtr
	}
```	
we have 4 * num cpu resuable nmts ~= 40
For each nmt we have:
- entrySize = 512 + 29
- squareSize = 512
So we alloc for each nmt 2 * 512 * 541 bytes, then multiply by 40, because we have 40 of them in the buffer = 2 * 512 * 541 * 40 = 22159360 (22mb)

For all other square sizes we should have roughly same amount, because 128 = 64 + 32 + 16 + 8 + 4 + 2 + 1 + 1. Also in theory we can reuse the buffer for the largest square for all other squares if needed

app/benchmarks/gas_estimator_test.go
```diff
@@ -84,8 +84,10 @@ func BenchmarkGasPriceEstimation(b *testing.B) {
 
 func generateRandomPFBs(b *testing.B, signer *user.Signer, account string, numberOfTransactions int, blobSize int) []types.Tx {
 	rand := random.New()
-	gasLimit := blobtypes.DefaultEstimateGas([]uint32{uint32(blobSize)})
 	blobs := blobfactory.ManyBlobs(rand, []share.Namespace{share.RandomBlobNamespace()}, []int{blobSize})
+	tempMsg, err := blobtypes.NewMsgPayForBlobs(signer.Account(account).Address().String(), appconsts.Version, blobs...)
+	require.NoError(b, err)
+	gasLimit := blobtypes.DefaultEstimateGas(tempMsg)
```
evan-forbes, line 90: why did we need to add this for this PR?

mcrakhman: The benchmarks that I used to test speed were broken and I just fixed them and forgot that I added it

app/process_proposal.go
rach-id: what if we create the buffers for square size 1, then 2, then 4, then 8, ..., then 256, would any of them be reusable for 512? and would they all stay in memory or we delete them at some point?

app/app.go
```diff
@@ -193,6 +194,7 @@ type App struct {
 	// blockTime is used to override the default TimeoutHeightDelay. This is
 	// useful for testing purposes and should not be used on public networks
 	// (Arabica, Mocha, or Mainnet Beta).
+	processProposalTreePool *wrapper.TreePoolProvider
```
rach-id: maybe change the name given this is also used in prepare proposal

app/app.go
```diff
@@ -480,6 +482,8 @@ func New(
 		&app.CircuitKeeper,
 		app.GovParamFilters(),
 	))
+	app.processProposalTreePool = wrapper.NewTreePoolProvider()
+	app.processProposalTreePool.PreallocatePool(uint64(appconsts.SquareSizeUpperBound))
```
rach-id: :+1:

app/prepare_proposal.go
```diff
@@ -54,7 +54,7 @@ func (app *App) PrepareProposalHandler(ctx sdk.Context, req *abci.RequestPrepare
 	// Erasure encode the data square to create the extended data square (eds).
 	// Note: uses the nmt wrapper to construct the tree. See
 	// pkg/wrapper/nmt_wrapper.go for more information.
-	eds, err := da.ExtendShares(share.ToBytes(dataSquare))
+	eds, err := da.ExtendSharesWithTreePool(share.ToBytes(dataSquare), app.TreePool().GetTreePool(uint64(dataSquare.Size())))
```
rach-id: so technically, regardless of the provided data square size, this will always return 512 trees, right?

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,215 @@
...
+		return tree
+	}
+}
+
+var _ rsmt2d.Tree = &bufferedTree{}
+
+// bufferedTree is a wrapper around NamespaceMerkleTree with buffer pooling support
+// for efficient memory management in high-throughput scenarios.
+type bufferedTree struct {
+	squareSize uint64 // note: this refers to the width of the original square before erasure-coded
```
rach-id:
```suggestion
	squareSize uint64 // note: this refers to the width of the original square before erasure encoding
```

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,215 @@
...
+	buffer     []byte         // Pre-allocated buffer for share data
+}
+
+// newBufferedTree creates a new bufferedTree with pre-allocated buffer and pool reference
+func newBufferedTree(squareSize uint64, axisIndex uint, pool *fixedTreePool, options ...nmt.Option) *bufferedTree {
+	if squareSize == 0 {
+		panic("cannot create a bufferedTree of squareSize == 0")
+	}
+	if pool == nil {
+		panic("cannot create a bufferedTree of pool == nil")
```
rach-id: if the pool is nil, can we create it instead of panicking?

pkg/wrapper/buffered_tree_test.go
```diff
@@ -0,0 +1,309 @@
...
+			squareSize: 4,
+		},
+		{
+			name:       "medium square 16x16",
+			squareSize: 16,
+		},
+		{
+			name:       "large square 64x64",
+			squareSize: 64,
+		},
```
rach-id: can we add 128, 256 and 512 if possible

rach-id (commented):
LGTM

pkg/wrapper/buffered_tree.go
mcrakhman: I think it is ok to leave it like this, because the constructor is private, so it should never be constructed without a pool and if it is - this is a critical problem in our implementation and we will catch it in tests

pkg/wrapper/buffered_tree.go
mcrakhman: Will fix it with the other comments

app/prepare_proposal.go
mcrakhman: No it has "number of parallel compute root operations" trees in the pool, which are reused. To clarify we have a fixed size tree pool, each tree has buffers equal to row of the largest square (but if their size is still lower than it extends automatically).

pkg/wrapper/buffered_tree_test.go
mcrakhman: Done in `bc2a07557411653f92dc7ffccb7c730726e60a09`

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// DefaultPreallocatedTreePool creates a new TreePool with a default pool size
+func DefaultPreallocatedTreePool(squareSize uint) *TreePool {
+	return NewTreePool(squareSize, runtime.NumCPU()*4)
+}
+
+// NewTreePool creates a new TreePool with the specified configuration.
+func NewTreePool(initSquareSize uint, poolSize int, opts ...nmt.Option) *TreePool {
+	pool := &TreePool{
```
rach-id: does it make sense to add a zero pool size case where we return an error or set a default? Just to foolproof it

mcrakhman: Yes, we can probably just use a default value for that for simplicity

rach-id: can we do it? because otherwise, the other methods might block

go.mod
```diff
@@ -24,8 +24,8 @@ require (
 	github.com/bcp-innovations/hyperlane-cosmos v1.0.1
 	github.com/celestiaorg/go-square/v2 v2.3.3
 	github.com/celestiaorg/go-square/v3 v3.0.1
-	github.com/celestiaorg/nmt v0.24.1
-	github.com/celestiaorg/rsmt2d v0.15.0
+	github.com/celestiaorg/nmt v0.24.2-0.20250918183924-34ea794437b6
+	github.com/celestiaorg/rsmt2d v0.15.1-0.20250919110934-5677c0cb7614
```
evan-forbes: converting to a draft for now as to avoid accidentally merging

evan-forbes: we can make this not a draft as soon we cut tags for both of these

I don't think we'll technically need to bump core before hand, however its likely good practice as well

app/app.go
```diff
@@ -691,6 +695,11 @@ func (app *App) GetSubspace(moduleName string) paramstypes.Subspace {
 	return subspace
 }
 
+// TreePool returns the instance used for managing a pool of Merkle trees.
+func (app *App) TreePool() *wrapper.TreePool {
+	return app.nmtTreePool
```
rootulp: [uber nit] is it possible to rename this method NmtTreePool or the field in the app to `app.treePool`. It seems inconsistent that they're not identical.

pkg/da/data_availability_header_test.go
```diff
@@ -53,14 +55,38 @@ func TestMinDataAvailabilityHeader(t *testing.T) {
...
-	dahv3 := MinDataAvailabilityHeader()
-	shareV2 := sharev2.ToBytes(sharev2.TailPaddingShares(appconsts.MinShareCount))
-	eds, err := ExtendShares(shareV2)
-	require.NoError(t, err)
-	dahV2, err := NewDataAvailabilityHeader(eds)
-	require.NoError(t, err)
-	require.Equal(t, dahv3.hash, dahV2.hash)
+	for _, extendShares := range []extendFunc{
+		extendSharesWithPool,
+		ExtendShares,
```
rootulp, line 86: [flup] this is a great test for backwards compatibility but I wonder if we need to continue supporting `ExtendShares` given `extendSharesWithPool` seems like a perf optimized version.

In a follow-up issue or PR should we explore removing `ExtendShares`?

pkg/da/data_availability_header_test.go
```diff
@@ -276,15 +332,20 @@ func TestConstructEDS_Versions(t *testing.T) {
...
-			eds, err := ConstructEDS(shares, appVersion, maxSquareSize)
-			if appVersion == 0 {
-				require.Error(t, err)
-				require.Nil(t, eds)
-			} else {
-				require.NoError(t, err)
-				require.NotNil(t, eds)
+			for _, constructEDS := range []constructFunc{
+				constructEDSWithPool,
+				ConstructEDS,
```
rootulp, line 343: [question][follow up] same question regarding removing `ConstructEDS` now that we have a perf optimized `constructEDSWithPool`

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,188 @@
...
+
+// newResizeableBufferTree creates a new resizeableBufferTree with pre-allocated buffer and pool reference.
+func newResizeableBufferTree(maxSquareSize uint, axisIndex uint, pool *TreePool, options ...nmt.Option) *resizeableBufferTree {
+	// this should never happen (we also check this with tests), because this is a private
+	if maxSquareSize == 0 {
+		panic("cannot create a resizeableBufferTree of maxSquareSize == 0")
+	}
+	if pool == nil {
+		panic("cannot create a resizeableBufferTree of pool == nil")
+	}
```
rootulp, line 107: [question] instead of `panic`ing can this function return an error argument instead? That way callers can handle what to do with the error (usually propagate it up)

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,188 @@
...
+	options = append(options, nmt.NamespaceIDSize(share.NamespaceSize))
+	options = append(options, nmt.IgnoreMaxNamespace(true), nmt.ReuseBuffers(true))
+	tree := nmt.New(appconsts.NewBaseHashFunc(), options...)
+	namespaceSize := share.NamespaceSize
+	entrySize := share.ShareSize + namespaceSize
+	return &resizeableBufferTree{
+		squareSize:      maxSquareSize,
+		maxSquareSize:   maxSquareSize,
```
rootulp, line 115: [question] do we need both of these fields if `squareSize` == `maxSquareSize`? It seems like they're both changed at the same time in the resize function.

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+}
+
+// Push adds share data to the tree using the pre-allocated buffer to avoid memory allocation.
+func (t *resizeableBufferTree) Push(data []byte) error {
+	if t.axisIndex+1 > 2*t.squareSize || t.shareIndex+1 > 2*t.squareSize {
+		return fmt.Errorf("pushed past predetermined square size: boundary at %d index at %d %d", 2*t.squareSize, t.axisIndex, t.shareIndex)
+	}
+	if len(data) < t.namespaceSize {
+		return fmt.Errorf("data is too short to contain namespace ID")
```
rootulp: [nit] this may have been copied from elsewhere but the namespace includes a namespace ID + namespace version. Namespace size is for the whole thing.

```suggestion
		return fmt.Errorf("data is too short to contain namespace")
```

https://celestiaorg.github.io/celestia-app/namespace.html

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	err := t.tree.Push(nidAndData)
+	if err != nil {
+		return err
+	}
+	t.incrementShareIndex()
+	return nil
+}
+
+// Root calculates the tree root using FastRoot and releases the tree back to the pool.
```
rootulp: [question] what is FastRoot?

rootulp (commented):
Overall looks good. Agreed it would be nice to bump the nmt and rsmt2d versions in this PR before merging so we should ideally get those PRs merged and released in the libraries.

pkg/wrapper/buffered_tree.go
mcrakhman: No FastRoot anymore, I rewrote the architecture etc, I will fix the comment :-)

pkg/wrapper/buffered_tree.go
mcrakhman: Yeah we can, I thought that we should prevent these conditions ever from happening, because ideally the app shouldn't be working if the pool is not functioning. But returning error is more idiomatic

pkg/da/data_availability_header_test.go
mcrakhman: `ExtendShares` is used to calculate shares in `NewShareInclusionProof`, so I would not touch it for now at least.

pkg/da/data_availability_header_test.go
mcrakhman: `ConstructEDS` is used in tool `check-eds-construction/main.go`. Maybe we can remove this in a follow up PR after we run this for a bit, just to be on the safe side?

pkg/wrapper/buffered_tree.go
mcrakhman: No, they are not, because `squareSize` reflects the current row size for the tree, whereas `maxSquareSize` reflects max row size which buffer can support

app/benchmarks/gas_estimator_test.go
rach-id: will this be reverted?

pkg/da/data_availability_header.go
```diff
@@ -91,6 +91,35 @@ func ConstructEDS(txs [][]byte, appVersion uint64, maxSquareSize int) (*rsmt2d.E
...
 
+// ConstructEDSWithTreePool constructs an ExtendedDataSquare from the given transactions and app version,
+// it uses treePool to optimize allocations.
+// If maxSquareSize is less than 0, it will use the upper bound square size for the given app version.
+func ConstructEDSWithTreePool(txs [][]byte, appVersion uint64, maxSquareSize int, treePool *wrapper.TreePool) (*rsmt2d.ExtendedDataSquare, error) {
+	switch appVersion {
+	case 0:
+		return nil, fmt.Errorf("app version cannot be 0")
+	case 1, 2, 3, 4, 5: // versions 1-5 are all compatible with v2 of the square package
+		if maxSquareSize < 0 {
```
rach-id, line 102: is `maxSquareSize == 0` a valid square size?

pkg/da/data_availability_header.go
```diff
@@ -91,6 +91,35 @@ func ConstructEDS(txs [][]byte, appVersion uint64, maxSquareSize int) (*rsmt2d.E
...
+			maxSquareSize = v5.SquareSizeUpperBound
+		}
+		// all versions 5 and below have the same parameters and algorithm
+		square, err := squarev2.Construct(txs, maxSquareSize, v5.SubtreeRootThreshold)
+		if err != nil {
+			return nil, err
+		}
+		return ExtendSharesWithTreePool(sharev2.ToBytes(square), treePool)
+	default: // assume all other versions are compatible with v3 of the square package
+		if maxSquareSize < 0 {
```
rach-id, line 112: same

pkg/da/data_availability_header_test.go
rach-id: let's deprecate `ExtendShares` now, and replace it with `ExtendSharesWithPool` in the next breaking release

pkg/da/data_availability_header_test.go
rach-id: same, let's deprecate in this PR and remove in the next major release

pkg/wrapper/buffered_tree.go
```diff
@@ -0,0 +1,195 @@
...
+	shareIndex      uint
+	namespaceSize   int
+	parityNamespace []byte
+	pool            *TreePool // reference to the pool this tree belongs to
+	buffer          []byte    // Pre-allocated buffer for share data
+	bufferEntrySize int
+}
+
+// newResizeableBufferTree creates a new resizeableBufferTree with pre-allocated buffer and pool reference.
+func newResizeableBufferTree(maxSquareSize uint, axisIndex uint, pool *TreePool, options ...nmt.Option) (*resizeableBufferTree, error) {
```
rach-id, line 100: probably should be `newResizableBufferTree`

go.mod
```diff
@@ -24,8 +24,8 @@ require (
 	github.com/bcp-innovations/hyperlane-cosmos v1.0.1
 	github.com/celestiaorg/go-square/v2 v2.3.3
 	github.com/celestiaorg/go-square/v3 v3.0.1
-	github.com/celestiaorg/nmt v0.24.1
-	github.com/celestiaorg/rsmt2d v0.15.0
+	github.com/celestiaorg/nmt v0.24.2
+	github.com/celestiaorg/rsmt2d v0.15.1
```
rach-id, line 28: probably these releases should have been minor version releases instead of bug fixes since they improved performance. nbd if not changed tho

rach-id (dismissed):
LGTM. Left some nits that can also be handled as followups

rach-id:
Even tho 1 approval is enough, let's hold off from merging until the other reviewers who have unresolved comments also approve.

go.mod
rootulp: I'm not sure whether `perf:` qualifies as a minor or patch release. I'm okay with either because I think conventional commits doesn't specify. 

Ref: https://www.conventionalcommits.org/en/v1.0.0/#summary

Personally I thought `perf:` would be a minor release but I recall Sevey claiming that anything non `feat:` should be a patch release.

rootulp (dismissed):
LGTM

[flup] one thing that may improve confidence is a test that runs process proposal for historical blocks and verifies the EDS generated on Mainnet matches the EDS generated via this PR (specifically `ConstructEDSWithTreePool`).

evan-forbes:
opened #5855 per rec from Rootul, which should provide supreme confidence in this never occurring

Issue #5671, opened 2025-09-05
flake: TestCelestiaTestSuite/TestE2EFullStackPFB


evan-forbes:
```
=== RUN   TestCelestiaTestSuite/TestE2EFullStackPFB
    logger.go:146: 2025-09-05T14:51:45.443Z	INFO	Exec	{"validator": true, "i": 0, "chain_id": "test", "test": "TestCelestiaTestSuite/TestE2EFullStackPFB", "image": "ghcr.io/celestiaorg/celestia-app:0b78d8e3", "test_name": "TestCelestiaTestSuite/TestE2EFullStackPFB", "command": "celestia-appd init test-val-0-TestCelestiaTestSuite_TestE2EFullStackPFB --chain-id test --home /var/cosmos-chain/celestia", "env,": [], "hostname": "TestCelestiaTestSuite_TestE2EFullStackPFB-ktqvnf", "container": "TestCelestiaTestSuite_TestE2EFullStackPFB-ktqvnf"}
    logger.go:146: 2025-09-05T14:51:45.443Z	INFO	Exec	{"validator": true, "i": 1, "chain_id": "test", "test": "TestCelestiaTestSuite/TestE2EFullStackPFB", "image": "ghcr.io/celestiaorg/celestia-app:0b78d8e3", "test_name": "TestCelestiaTestSuite/TestE2EFullStackPFB", "command": "celestia-appd init test-val-1-TestCelestiaTestSuite_TestE2EFullStackPFB --chain-id test --home /var/cosmos-chain/celestia", "env,": [], "hostname": "TestCelestiaTestSuite_TestE2EFullStackPFB-ordygx", "container": "TestCelestiaTestSuite_TestE2EFullStackPFB-ordygx"}
[...]

Issue #5531, opened 2025-08-13
flake: TestCelestiaTestSuite/TestStateSync


evan-forbes:
noticed this flake in the e2e test

```
        	Error Trace:	/home/runner/work/celestia-app/celestia-app/test/docker-e2e/e2e_state_sync_test.go:50
        	Error:      	Received unexpected error:
        	            	Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint test-val-0-TestCelestiaTestSuite_TestStateSync (5ddf51d7f225576179e87eefc2b816cd5be95fedf3bf36950eccc935252a9c36): failed to listen on TCP socket: address already in use
        	Test:       	TestCelestiaTestSuite/TestStateSync
        	Messages:   	failed to start chain
--- FAIL: TestCelestiaTestSuite (16.57s)
    --- FAIL: TestCelestiaTestSuite/TestStateSync (16.52s)
```

for testnode we still get these occasionally, however one thing that at least made them less frequent was using a deterministic and threadsafe port determination function

rootulp:
This is a Docker networking issue (`failed to listen on TCP socket: address already in use`) in the Docker E2E test infrastructure. Same class of issue as #5858 and #5671.

Note: For the non-Docker testnode tests, this class of port binding issue was fixed by commit **0833ebbde** which introduced deterministic port allocation and commit **493251a2e** which added retry logic. The Docker E2E tests use a different networking model (Docker container port mappings) where these fixes don't apply.

Issue #5477, opened 2025-08-07
TestStateSync failed due to `failed to initialize chain nodes: set volume owner: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit.`


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/actions/runs/16807518826/job/47603794907

## Problem

```
=== RUN   TestCelestiaTestSuite/TestStateSync
    e2e_state_sync_test.go:47: 
        	Error Trace:	/home/runner/work/celestia-app/celestia-app/test/docker-e2e/e2e_state_sync_test.go:47
        	Error:      	Received unexpected error:
        	            	failed to initialize chain nodes: set volume owner: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
        	Test:       	TestCelestiaTestSuite/TestStateSync
        	Messages:   	failed to get chain
```

evan-forbes:
😭

rootulp:
This is an infrastructure issue (Docker Hub unauthenticated pull rate limit), not a code bug. The error `toomanyrequests: You have reached your unauthenticated pull rate limit` is a CI environment configuration issue that should be addressed by authenticating Docker pulls in CI workflows. This is not reproducible as a code flake.

PR #5779, opened 2025-09-17
fix: update tastora to latest tag to fix flaky tests

mojtaba-esk:
**Fix flaky tests by updating tastora framework**

Updates tastora to latest version which includes a mutex fix for port allocation race conditions that were causing `TestE2EFullStackPFB` to fail randomly. This fix might resolve some other flaky tests as well.


rootulp:
https://github.com/celestiaorg/celestia-app/issues/5671 https://github.com/celestiaorg/celestia-app/issues/5531 seem port related.

https://github.com/celestiaorg/celestia-app/issues/5477 doesn't seem port related so why do you expect v0.4.1 to fix?

@mojtaba-esk Also should we close those issues with this PR (i.e. "Closes #x")

Issue #4690, opened 2025-04-29
Investigate removing go-square v1 from celestia-app v3.x


rootulp:
## Context

[Sync on app v4 for node](https://docs.google.com/document/d/1BPIhNc2gqeV0Xx5Hz6S_4chG2QUHcnqw_AqCNoxZyo8/edit?usp=sharing)

## Problem

celestia-app v3.x has [these lines](https://github.com/celestiaorg/celestia-app/blob/1d32f5a9b2f2f2a59ce28ffc7ff71f9a5f854347/app/process_proposal.go#L150-L158) that are absent from celestia-node. Since celestia-node can sync through historical blocks (i.e. v1 and v2 blocks) without using go-square v1, it seems plausible that those lines are not needed on celestia-app v3.x

## Prerequisites

1. https://github.com/celestiaorg/celestia-app/issues/4689

## Proposal

1. Remove those lines
2. Verify that the above unit test continues to pass

PR #5781, opened 2025-09-17
chore: create construct EDS function to be used in process proposal and by node

cmwaters:
This is a proposed version whereby a single constructor method for the EDS can handle various versions.

This can be reviewed for design comments but go-square will need to have this [PR](https://github.com/celestiaorg/go-square/pull/194) merged and a new release cut that this PR should update to


pkg/da/data_availability_header.go
```diff
@@ -62,9 +65,39 @@ func NewDataAvailabilityHeader(eds *rsmt2d.ExtendedDataSquare) (DataAvailability
 	return dah, nil
 }
 
+// ConstructEDS constructs an ExtendedDataSquare from the given transactions and app version.
+// If maxSquareSize is less than 0, it will use the upper bound square size for the given app version.
+func ConstructEDS(txs [][]byte, appVersion uint64, maxSquareSize int) (*rsmt2d.ExtendedDataSquare, error) {
+	if appVersion == 0 {
+		return nil, fmt.Errorf("app version cannot be 0")
+	}
+	if appVersion <= v5.Version {
```
renaynay, line 74: maybe use a switch here rather? Looks cleaner

pkg/da/data_availability_header.go
```diff
@@ -62,9 +65,39 @@ func NewDataAvailabilityHeader(eds *rsmt2d.ExtendedDataSquare) (DataAvailability
...
+		if maxSquareSize < 0 {
+			maxSquareSize = appconsts.SquareSizeUpperBound
+		}
+		square, err := squarev3.Construct(txs, maxSquareSize, appconsts.SubtreeRootThreshold)
+		if err != nil {
+			return nil, err
+		}
+		return ExtendShares(sharev3.ToBytes(square))
+	}
+	return nil, fmt.Errorf("unsupported app version: %d", appVersion)
```
evan-forbes, line 95: ofc not blocking

we've had this discussion a bunch of different times, so I won't repeat the args here, but imo we should just assume this version of go-square is being used. version protection is already handled in both comet and node before this code gets reached.

test/docker-e2e/go.mod
```diff
@@ -77,6 +77,7 @@ require (
 	github.com/bits-and-blooms/bitset v1.17.0 // indirect
 	github.com/bytedance/sonic v1.14.0 // indirect
 	github.com/bytedance/sonic/loader v0.3.0 // indirect
+	github.com/celestiaorg/go-square/v2 v2.3.3-rc0 // indirect
```
rootulp: We can probably create v2.3.3 official release now

pkg/da/data_availability_header.go
```diff
@@ -62,9 +65,39 @@ func NewDataAvailabilityHeader(eds *rsmt2d.ExtendedDataSquare) (DataAvailability
...
+func ConstructEDS(txs [][]byte, appVersion uint64, maxSquareSize int) (*rsmt2d.ExtendedDataSquare, error) {
+	if appVersion == 0 {
+		return nil, fmt.Errorf("app version cannot be 0")
+	}
+	if appVersion <= v5.Version {
+		if maxSquareSize < 0 {
+			maxSquareSize = v5.SquareSizeUpperBound
+		}
+		// all versions 5 and below have the same parameters and algorithm
+		square, err := squarev2.Construct(txs, maxSquareSize, v5.SubtreeRootThreshold)
```
rootulp, line 79: Is this true? I thought app version 3 which introduced authored blobs had a different construct method which led to https://github.com/celestiaorg/celestia-app/blob/4dd92890ac0a50d163627053d6e0323c9b5ba6a9/app/process_proposal.go#L140-L158

on re-review there is also https://github.com/celestiaorg/celestia-app/issues/4690 so maybe we don't need cases for the historical app versions and go-versions.

pkg/da/data_availability_header_test.go
```diff
@@ -239,6 +252,70 @@ func TestSquareSize(t *testing.T) {
 	}
 }
 
+func TestConstructEDS_Versions(t *testing.T) {
+	for version := uint64(0); version <= appconsts.Version*uint64(2); version++ {
```
rootulp: [nit] we probably only need to go 1 above the current app version. Otherwise, half the iterations in this loop will be error cases.

```suggestion
	for version := uint64(0); version <= appconsts.Version+uint64(1); version++ {
```

rootulp (approved):
LGTM but it would be confidence building to have test cases that verify ConstructEDS works for all historical app versions, in particular app versions 1 and 2 which uses go-square v1 and had a potentially different construct function.

test/docker-e2e/go.mod
rootulp: Created: https://github.com/celestiaorg/go-square/releases/tag/v2.3.3 will bump to it here

rootulp: https://github.com/celestiaorg/celestia-app/pull/5781/commits/a411a9ac36d99db7288a71d94dd06dbeb26f0037

pkg/da/data_availability_header_test.go
rootulp: 4ba726fa7741855826c8c320ca48bcd4b6dd0cb6

pkg/da/data_availability_header.go
cmwaters: The addition of authored blobs is additive I believe meaning that V2 should be able to produce the same square as V1 square for a block with v0 share version transactions only

app/process_proposal.go
```diff
@@ -130,24 +128,18 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
...
 	}
 
 	// Assert that the square size stated by the proposer is correct
-	if uint64(dataSquare.Size()) != req.SquareSize {
+	if uint64(eds.Width()) != req.SquareSize*2 {
 		logInvalidPropBlock(app.Logger(), blockHeader, "proposed square size differs from calculated square size")
 		return reject(), nil
 	}
 
-	eds, err := da.ExtendShares(share.ToBytes(dataSquare))
```
renaynay, line 145: What was the point of this check before ? Just to ensure that constructed square can be decomposed back into shares, and then extended again into square? But there's no equality check

pkg/da/data_availability_header.go
renaynay: Wait @cmwaters wasn't the whole problem with the broken v2 that it did not encode authored blobs correctly and that's why we need v3?

pkg/da/data_availability_header.go
```diff
@@ -193,13 +226,13 @@ func MinDataAvailabilityHeader() DataAvailabilityHeader {
...
 func MinShares() [][]byte {
-	return share.ToBytes(EmptySquareShares())
+	return sharev3.ToBytes(EmptySquareShares())
 }
 
 // EmptySquareShares is a copy of the function defined in the square package to avoid
 // a circular dependency. TODO deduplicate
-func EmptySquareShares() []share.Share {
-	return share.TailPaddingShares(appconsts.MinShareCount)
+func EmptySquareShares() []sharev3.Share {
```
renaynay, line 234: To be sure: there's no difference between squarev3 empty square shares and squarev2 empty square shares?

pkg/da/data_availability_header_test.go
```diff
@@ -31,6 +34,16 @@ func TestMinDataAvailabilityHeader(t *testing.T) {
 	require.NoError(t, dah.ValidateBasic())
 }
 
+func TestMinDataAvailabilityHeaderVersioning(t *testing.T) {
```
renaynay, line 37:
```suggestion
func TestMinDataAvailabilityHeaderBackwardsCompatibility(t *testing.T) {
```

pkg/da/data_availability_header.go
renaynay: Ah I guess that's what the backwards compat test is for 👍🏻  all g

pkg/da/data_availability_header_test.go
```diff
@@ -31,6 +34,16 @@ func TestMinDataAvailabilityHeader(t *testing.T) {
...
+	dah := MinDataAvailabilityHeader()
```
renaynay, line 38: Nit: would call this dahV3 just to be clear what's being tested.

pkg/da/data_availability_header.go
```diff
@@ -62,9 +65,39 @@ func NewDataAvailabilityHeader(eds *rsmt2d.ExtendedDataSquare) (DataAvailability
 	return dah, nil
 }
 
+// ConstructEDS constructs an ExtendedDataSquare from the given transactions and app version.
+// If maxSquareSize is less than 0, it will use the upper bound square size for the given app version.
+func ConstructEDS(txs [][]byte, appVersion uint64, maxSquareSize int) (*rsmt2d.ExtendedDataSquare, error) {
```
renaynay, line 70: Okay so for celestia-node's call to this function, we should always pass -1 for `maxSquareSize` such that this function sets the defaults for us. 👍🏻

renaynay (approved):
Thank you @cmwaters 

Left some optional nits but func signature LGTM

pkg/da/data_availability_header.go
cmwaters: I don't think it was about encoding but about estimating the amount of space taken by authored blob shares.

V1 and V2 compatibility is a different question

pkg/da/data_availability_header.go
cmwaters: Yeah I wanted to make sure

pkg/da/data_availability_header.go
cmwaters: Yes

pkg/da/data_availability_header.go
cmwaters: I had a switch before and I didn't like it but I also don't like this

app/process_proposal.go
cmwaters: Are you talking about `ExtendShares`? That was part of constructing the DAH

cmwaters:
@renaynay I plan to address your nits in a follow up PR

evan-forbes (commented):
apologies, forgot to approve

Issue #5755, opened 2025-09-15
Transactions might accidentally get resigned during evictions


ninabarbakadze:
## Description

Currently, the consensus node we're talking to might evict a transaction from its local mempool(which does not mean global eviction) and our way of working around it is resubmitting transaction bytes with the same signature. https://github.com/celestiaorg/celestia-app/blob/bd124cef269e98e80a9e93cd45160c82fe4a510e/pkg/user/tx_client.go#L639

When I was pressure testing TxClient I noticed a strange pattern where an evicted transaction might get sequence mismatch during eviction. This is bad for two reasons:
- Just because this particular node evicted transaction does not mean it won't get proposed by another node.
- Resigning evicted transaction with new nonce will give it a new hash.

This means that we're breaking replay protection and these blobs will most likely get included twice.

## Proposed solution

Keep polling for a period of one minute(could be a different number) and see if it eventually gets included, otherwise return evicted to the user.

evan-forbes:
wait, don't we need to wait one block for the state to reset in the mempool ? or do we already do that?

ninabarbakadze:
During evictions? If a transaction was broadcasted and then got evicted it means mempool state was already updated.

evan-forbes:
> If a transaction was broadcasted and then got evicted it means mempool state was already updated.

what if the eviction is not due to recheck, but size constraints. afaiu the cosmos-sdk doesn't have a way to revert the state it must wait until the next height.

> This means that we're breaking replay protection and these blobs will most likely get included twice.

if we sign over two different sequences for the same blob then yeah that makes sense. now that we are reverting that in #5783, just to clarify you are saying there are no cases where we want to sign over a new sequence?

can we test to ensure that the txclient doesn't halt after seeing a mismatch? or do we know why are folks still having this issue then?

ninabarbakadze:
> what if the eviction is not due to recheck, but size constraints. afaiu the cosmos-sdk doesn't have a way to revert the state it must wait until the next height.

So you think that sequence mismatch is happening because mempool state is not getting updated by the time we try to resubmit?

> if we sign over two different sequences for the same blob then yeah that makes sense. now that we are reverting that in https://github.com/celestiaorg/celestia-app/pull/5783, just to clarify you are saying there are no cases where we want to sign over a new sequence?

We can sign over transaction in two scenarios:
- Tx is just getting submitted has not been added to node's mempool and we're getting sequence mismatch broadcast errors. We can safely resign here.
[...]

PR #5783, opened 2025-09-17
fix: tx client evictions bug

ninabarbakadze:

### DIFF
--- a/pkg/user/tx_client.go
+++ b/pkg/user/tx_client.go
@@ -411,7 +411,7 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
 }
 
 func (client *TxClient) broadcastTxAndIncrementSequence(ctx context.Context, conn *grpc.ClientConn, txBytes []byte, signer string) (*sdktypes.TxResponse, error) {
-	resp, err := client.broadcastTx(ctx, conn, txBytes, signer)
+	resp, err := client.broadcastTxWithRetry(ctx, conn, txBytes, signer)
 	if err != nil {
 		return nil, err
 	}
@@ -431,7 +431,8 @@ func (client *TxClient) broadcastTxAndIncrementSequence(ctx context.Context, con
 
 // broadcastTx resubmits a transaction that was evicted from the mempool.
 // Unlike the initial broadcast, it doesn't increment the signer's sequence number.
-func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn, txBytes []byte, signer string) (*sdktypes.TxResponse, error) {
+// broadcastTx broadcasts a transaction to the chain and returns the response.
+func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn, txBytes []byte) (*sdktypes.TxResponse, error) {
 	txClient := sdktx.NewServiceClient(conn)
 	resp, err := txClient.BroadcastTx(
 		ctx,
@@ -444,17 +445,6 @@ func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn, 
 		return nil, err
 	}
 	if resp.TxResponse.Code != abci.CodeTypeOK {
-		if apperrors.IsNonceMismatchCode(resp.TxResponse.Code) {
-			expectedSequence, err := apperrors.ParseExpectedSequence(resp.TxResponse.RawLog)
-			if err != nil {
-				return nil, fmt.Errorf("error parsing sequence mismatch: %w. RawLog: %s", err, resp.TxResponse.RawLog)
-			}
-			if err := client.signer.SetSequence(signer, expectedSequence); err != nil {
-				return nil, fmt.Errorf("setting sequence: %w", err)
-			}
-
-			return client.retryBroadcastingTx(ctx, txBytes)
-		}
 		broadcastTxErr := &BroadcastTxError{
 			TxHash:   resp.TxResponse.TxHash,
 			Code:     resp.TxResponse.Code,
@@ -464,6 +454,27 @@ func (client *TxClient) broadcastTx(ctx context.Context, conn *grpc.ClientConn, 
 	}
 
 	return resp.TxResponse, nil
+}
+
+// broadcastTxWithRetry broadcasts a transaction and retries if it encounters a sequence mismatch.
+func (client *TxClient) broadcastTxWithRetry(ctx context.Context, conn *grpc.ClientConn, txBytes []byte, signer string) (*sdktypes.TxResponse, error) {
+	resp, err := client.broadcastTx(ctx, conn, txBytes)
+	if err != nil {
+		broadcastTxErr, ok := err.(*BroadcastTxError)
+		if !ok || !apperrors.IsNonceMismatchCode(broadcastTxErr.Code) {
+			return nil, err
+		}
+		expectedSequence, err := apperrors.ParseExpectedSequence(broadcastTxErr.ErrorLog)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing sequence mismatch: %w. ErrorLog: %s", err, broadcastTxErr.ErrorLog)
+		}
+		if err := client.signer.SetSequence(signer, expectedSequence); err != nil {
+			return nil, fmt.Errorf("setting sequence: %w", err)
+		}
+		return client.retryBroadcastingTx(ctx, txBytes)
+	}
+
+	return resp, nil
 }
 
 // retryBroadcastingTx creates a new transaction by copying over an existing transaction but creates a new signature with the
@@ -521,7 +532,7 @@ func (s *TxClient) retryBroadcastingTx(ctx context.Context, txBytes []byte) (*sd
 		}
 	}
 
-	broadcastTxResp, err := s.broadcastTx(ctx, s.conns[0], newTxBytes, signer)
+	broadcastTxResp, err := s.broadcastTxWithRetry(ctx, s.conns[0], newTxBytes, signer)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/user/tx_client_test.go
```diff
@@ -391,6 +391,68 @@ func TestEvictions(t *testing.T) {
...
+		expectedTxHash := "test-tx-hash-123" // This matches what our mock server returns
+
+		// Happy path: PENDING -> EVICTED -> COMMITTED
+		responseSequences := map[string][]*tx.TxStatusResponse{
+			expectedTxHash: {
+				{Status: core.TxStatusPending},                                  // 1st call
+				{Status: core.TxStatusEvicted},                                  // 2nd call
+				{Status: core.TxStatusEvicted},                                  // 3rd call
+				{Status: core.TxStatusEvicted},                                  // 4th call
+				{Status: core.TxStatusCommitted, Height: 102, ExecutionCode: 0}, // 3rd call
```
evan-forbes: the comments seem off but also dont add much info imo
```suggestion
				{Status: core.TxStatusPending},
				{Status: core.TxStatusEvicted},
				{Status: core.TxStatusEvicted},
				{Status: core.TxStatusEvicted},
				{Status: core.TxStatusCommitted, Height: 102, ExecutionCode: 0},
```

pkg/user/tx_client_test.go
```diff
@@ -391,6 +391,68 @@ func TestEvictions(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, txInfo.Status, core.TxStatusCommitted)
 	}
+
+	t.Run("gets nonce mismatch during eviction and gets confirmed in the meantime", func(t *testing.T) {
+		expectedTxHash := "test-tx-hash-123" // This matches what our mock server returns
+
+		// Happy path: PENDING -> EVICTED -> COMMITTED
```
evan-forbes, line 398: happy relative to what?

pkg/user/tx_client_test.go
rootulp: Why is this the 3rd call and not the 5th call?
```suggestion
				{Status: core.TxStatusCommitted, Height: 102, ExecutionCode: 0}, // 5th call
```

pkg/user/tx_client.go
```diff
@@ -631,14 +644,27 @@ func (client *TxClient) ConfirmTx(ctx context.Context, txHash string) (*TxRespon
...
-			// Resubmit straight away in the event of eviction and keep polling until tx is committed
-			_, err := client.broadcastTx(ctx, client.conns[0], client.txTracker[txHash].txBytes, signer)
+
+			if evictionPollTimeStart != nil {
+				// Eviction timer is running, no need to resubmit again
+				break
+			}
+
+			// If we're not already tracking eviction timeout, try to resubmit
+			_, err := client.broadcastTx(ctx, client.conns[0], client.txTracker[txHash].txBytes)
```
rootulp, line 658: [question] If the tx was just evicted from the mempool, why should the tx client try to resubmit it? The proposed solution in https://github.com/celestiaorg/celestia-app/issues/5755 is to wait for one minute before returning evicted to the user. I don't see where/why the tx client should retry submitting the tx if it was just evicted.

rootulp (commented):
I don't understand the original issue enough to see how this resolves.

At a very high level, I don't understand why the tx client needs to parse the expected sequence from a broadcast tx error. Instead of doing that, is it possible for the tx client to query the most recent nonce observed on-chain _before_ submitting the tx?

pkg/user/mock_tx_server_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"google.golang.org/grpc/test/bufconn"
+)
+
+// mockTxServer implements both gRPC ServiceServer and TxServer interfaces for mocking broadcast and tx status responses
+type mockTxServer struct {
+	sdktx.UnimplementedServiceServer
+	tx.UnimplementedTxServer
+	txStatusResponses   map[string][]*tx.TxStatusResponse // txHash with sequence of responses
+	txStatusCallCounts  map[string]int                    // txHash with number of TxStatus calls made
+	broadcastCallCounts map[string]int                    // txHash with number of BroadcastTx calls made
```
evan-forbes, line 26: I don't dislike this, but aren't we're using other mocks for tests in this package?

we need to fix here or flup imo

pkg/user/tx_client.go
```diff
@@ -40,6 +40,7 @@ import (
 const (
 	DefaultPollTime          = 3 * time.Second
 	txTrackerPruningInterval = 10 * time.Minute
+	evictionPollTime         = 1 * time.Minute
```
evan-forbes, line 43: can we add a quick comment here for what this polling value actually means and how it differs from the other poll time?

pkg/user/tx_client.go
```diff
@@ -410,8 +411,10 @@ func (client *TxClient) BroadcastTx(ctx context.Context, msgs []sdktypes.Msg, op
 	return client.broadcastTxAndIncrementSequence(ctx, client.conns[0], txBytes, account)
 }
 
+// broadcastTxAndIncrementSequence submits a transaction to the chain and increments the sequence of the signer.
+// It will retry the transaction if it encounters a sequence mismatch.
 func (client *TxClient) broadcastTxAndIncrementSequence(ctx context.Context, conn *grpc.ClientConn, txBytes []byte, signer string) (*sdktypes.TxResponse, error) {
-	resp, err := client.broadcastTx(ctx, conn, txBytes, signer)
+	resp, err := client.broadcastTxWithRetry(ctx, conn, txBytes, signer)
```
evan-forbes, line 417: so we're potentially hitting an incorrect sequence here, but then increasing?

what happens if the retry also fails?

evan-forbes (commented):
have a few blocking comments/questions

please also remove the check on lines 366-382 of tx_client as those don't make any sense and I'm pretty sure are impossible to hit

pkg/user/tx_client_test.go
ninabarbakadze: by happy path I mean the intended success outcome relative to an eviction case. This is the lifecycle:
pending -> evicted -> resubmission attempt -> nonce mismatch -> confirmed.

So while eviction and nonce mismatch aren’t happy in the general sense this is the best-case resolution within that flow because the tx ultimately gets confirmed.

pkg/user/mock_tx_server_test.go
ninabarbakadze: Do you mean if we use other mocks in this package? Yes, `TestMultiConnBroadcast` also uses mocks for grpc responses and I could potentially unify both mock setups. The logic in that test looks kind of convoluted and hard to read so I didn't want to get into that but it will be cleaner if I mock once for both tests.

pkg/user/tx_client.go
ninabarbakadze: Good question! When we receive an eviction response, it just means that a particular node has removed the transaction from its mempool. However that transaction may still exist in the mempools of other nodes, and it’s sort of expected that another node will eventually propose it and it'll get committed.

If we were to tell the user that their transaction was evicted, they might assume that they need to resubmit it. In practice, that would mean re-signing and re-broadcasting the transaction, producing a new signature and hash. If the original transaction later gets confirmed this would effectively result in a double spend.

On the other hand, resubmitting the same transaction bytes does not change the hash it just makes the transaction reappear in the node’s mempool.

Now, regarding the bug fix:
The issue we addressed was that evicted transactions were being mistakenly re-signed during re-broadcast. This happened because the we might hit sequence mismatch error durign re-submission of the eivcted tx and attempted to re-sign with the “expected” sequence number. If that re-signing occurred it would produce a different transaction hash and introduce the same double-spending problem described above.

We recently added logic to handle expected sequence parsing and re-signing correctly, and the fix ensures that this mechanism does not trigger in the case of evictions. That prevents accidental re-signing and avoids the associated risks.

LMK if you'd like me to further expand on anything.

pkg/user/tx_client_test.go
ninabarbakadze: I was moving the lines around and the comments got out of order. Will fix.

pkg/user/tx_client.go
ninabarbakadze: During evictions we do not want to increase the sequence at all. If we get broadcast error during eviction resubmission we will try to wait and see if another node proposes this transaction during the period of max 1 minute. If the tx response is still evicted we will just exit the function and return eviction to the user.

pkg/user/tx_client.go
rootulp: Thanks for expanding!

>This happened because the we might hit sequence mismatch error durign re-submission of the eivcted tx and attempted to re-sign with the “expected” sequence number.

This approach seems fragile. [flup] I wonder if it's possible to eliminate the re-signing with "expected" sequence number which seems like it would remove the type of bug this PR is addressing.

pkg/user/mock_tx_server_test.go
evan-forbes: we can definitely delete the other one, but there should be only one for our own sanity
 
 totally fine to flup

pkg/user/tx_client.go
evan-forbes: is there any chance or place that we will continually retry indefinitely?

evan-forbes (approved):

pkg/user/tx_client.go
ninabarbakadze: Currently parsing the expected sequence is the only way to actually know what the mempool state is/what sequence it expects because we can query the network sequence but no way of querying mempool sequence.

ninabarbakadze:
I'm merging this and I'll address some comments in flup @evan-forbes 

edit: Actually, I'd want Callum to take a look too since he has a lot of context on tx client.

pkg/user/tx_client.go
ninabarbakadze: Consensus node could be being malicious giving us sequence errors to try and DoS the tx client but once a transaction gets signed with right sequence and passes check tx it will get into the mempool and this will return. Otherwise user can also cancel the context and the function will exit.

PR #5794, opened 2025-09-18
feat: add all of the fibre protos

evan-forbes:
## Overview

just add all the protos in one go as this is essentially copy and pasting from the spec and then generating them


evan-forbes:
still need to clean up one file and ~~then it should be ready for review~~

pls don't review

proto/celestia/fibre/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+// EventRequestWithdrawalFromEscrow is emitted when a withdrawal is requested from an escrow account.
+message EventRequestWithdrawalFromEscrow {
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  cosmos.base.v1beta1.Coin amount = 2 [(gogoproto.nullable) = false];
+  google.protobuf.Timestamp available_at = 3 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
+}
+
+// EventProcessWithdrawal is emitted when a withdrawal is processed.
+message EventProcessWithdrawal {
+  string processor = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
cmwaters: Who is the processor here? I thought this was done automatically in `EndBlock`

proto/celestia/fibre/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+  string signer = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  cosmos.base.v1beta1.Coin amount = 3 [(gogoproto.nullable) = false];
+}
+
+// EventPayForFibre is emitted when a payment for fibre is processed.
+message EventPayForFibre {
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string escrow_signer = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
cmwaters: Aren't these one and the same thing?

proto/celestia/fibre/v1/event.proto
```diff
@@ -0,0 +1,44 @@
...
+  bytes namespace = 3;
+  uint32 validator_count = 4;
```
cmwaters: Why is this added? Why not something like commitment?

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,51 @@
...
+  // signer is the owner of the escrow account to charge
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // namespace is the namespace the blob is associated with. share version must be 2.
+  bytes namespace = 2;
+  // blob_size is the size of the blob in bytes
+  uint32 blob_size = 3;
+  // commitment is the hash of the row root and the Random Linear Combination (RLC) root
+  bytes commitment = 4;
+  // share_version is the version of the row format
+  uint32 share_version = 5;
```
cmwaters: I thought this was called `row_version`. I feel like it should at least have a name that differentiates itself from the existing `share_version`

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,51 @@
...
+  // signer is the address that owns the escrow account this withdrawal is for
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // amount is the amount to be withdrawn
+  cosmos.base.v1beta1.Coin amount = 2 [(gogoproto.nullable) = false];
+  // requested_at is the timestamp when withdrawal was requested
+  google.protobuf.Timestamp requested_at = 3 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
+}
+
+// PaymentPromise contains the commitment and payment details for a fibre blob.
+message PaymentPromise {
```
cmwaters, line 29: I believe the `validator_height` is missing here

proto/celestia/fibre/v1/query.proto
```diff
@@ -0,0 +1,100 @@
...
+
+  // PendingWithdrawals queries pending withdrawals for an escrow account.
+  rpc PendingWithdrawals(QueryPendingWithdrawalsRequest) returns (QueryPendingWithdrawalsResponse) {
+    option (google.api.http).get = "/fibre/v1/pending-withdrawals/{signer}";
+  }
+
+  // ProcessedPromise queries whether a promise has been processed.
+  rpc ProcessedPromise(QueryProcessedPromiseRequest) returns (QueryProcessedPromiseResponse) {
+    option (google.api.http).get = "/fibre/v1/processed-promise/{promise_hash}";
+  }
```
cmwaters: nit: Can we try to consistently call them `PaymentPromise`s - even if it is a little more verbose. I think the fact that it is for a payment is important context

proto/celestia/fibre/v1/query.proto
```diff
@@ -0,0 +1,100 @@
...
+// QueryEscrowAccountRequest is the request type for the Query/EscrowAccount RPC method.
+message QueryEscrowAccountRequest {
+  string signer = 1;
+}
+
+// QueryEscrowAccountResponse is the response type for the Query/EscrowAccount RPC method.
+message QueryEscrowAccountResponse {
+  EscrowAccount escrow_account = 1;
+  bool found = 2;
+}
```
cmwaters, line 63: Don't we want the `balance` and the `available_balance` here

proto/celestia/fibre/v1/tx.proto
```diff
@@ -0,0 +1,120 @@
...
+  // PayForFibre contains the original payment promise with validator signatures
+  rpc PayForFibre(MsgPayForFibre) returns (MsgPayForFibreResponse) {
+    option (google.api.http) = {
+      post: "/fibre/v1/pay-for-fibre",
+      body: "*"
+    };
+  }
+
+  // PaymentTimeout processes a payment promise after the timeout period
+  rpc PaymentTimeout(MsgPaymentTimeout) returns (MsgPaymentTimeoutResponse) {
```
cmwaters: nit: all other methods have some verb: "Pay", "Request", "Deposit". Added one here for consistency
```suggestion
  rpc ProcessPaymentTimeout(MsgProcessPaymentTimeout) returns (MsgProcessPaymentTimeoutResponse) {
```

proto/celestia/fibre/v1/tx.proto
```diff
@@ -0,0 +1,120 @@
...
+// MsgUpdateFibreParams defines the sdk.Msg type to update the fibre parameters.
+message MsgUpdateFibreParams {
+  option (cosmos.msg.v1.signer) = "authority";
+  // authority is the address of the governance account.
+  string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // params defines the fibre parameters to update.
+  //
+  // NOTE: All parameters must be supplied.
+  Params params = 2 [(gogoproto.nullable) = false];
+}
```
cmwaters, line 114: Did we make a decision on whether we want these parameters to be governance modifiable or only changeable via a hard fork

cmwaters (commented):
Good start! Have a few nits and I think there are a few fields that are extraneous or missing 

Nit: when it comes to the `PaymentPromise`, I'd rather it be explicitly called "payment promise" than just "promise". Maybe I'm reading into it but I can imagine readers asking a promise for what?

proto/celestia/fibre/v1/fibre.proto
rootulp: [nit] Can the field be called `height` if it's just block height? 

Ref: https://github.com/celestiaorg/fibre-da-spec/pull/3#discussion_r2373348329

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,51 @@
...
+  // signer is the owner of the escrow account to charge
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // namespace is the namespace the blob is associated with. share version must be 2.
+  bytes namespace = 2;
+  // blob_size is the size of the blob in bytes
+  uint32 blob_size = 3;
+  // commitment is the hash of the row root and the Random Linear Combination (RLC) root
+  bytes commitment = 4;
```
rootulp, line 37: [question] are these two hashes concatenated? Or how is the commitment composed of these two roots?

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,51 @@
...
+  // share_version is the version of the row format
+  uint32 share_version = 5;
+  // creation_timestamp is the timestamp when this promise was created. This
+  // is critical for determining which validators sign the commitment and
+  // determining when service stops for this blob.
+  google.protobuf.Timestamp creation_timestamp = 6 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
```
rootulp, line 43: [question] I'm a little confused on why there is a `creation_timestamp`. Can a user create a PaymentPromise off-chain at T1, then wait 100 blocks until T2 and then submit it on-chain?

If yes, should this field be populated with T1 or T2?

I also don't follow how this field helps for determining which validators sign the commitment, because I thought that was based on the `validator_height` field.

proto/celestia/fibre/v1/genesis.proto
```diff
@@ -0,0 +1,23 @@
...
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/fibre/types";
+
+// GenesisState defines the fibre module's genesis state.
+message GenesisState {
+  Params params = 1 [(gogoproto.nullable) = false];
+  repeated EscrowAccount escrow_accounts = 2 [(gogoproto.nullable) = false];
+  repeated PendingWithdrawal pending_withdrawals = 3 [(gogoproto.nullable) = false];
+  repeated ProcessedPromiseEntry processed_promises = 4 [(gogoproto.nullable) = false];
```
rootulp: [question] why do these need to be in genesis state? I assume they'll all be set to an empty slice.

proto/celestia/fibre/v1/query.proto
```diff
@@ -0,0 +1,100 @@
...
+
+// QueryProcessedPromiseRequest is the request type for the Query/ProcessedPromise RPC method.
+message QueryProcessedPromiseRequest {
+  bytes promise_hash = 1;
+}
+
+// QueryProcessedPromiseResponse is the response type for the Query/ProcessedPromise RPC method.
+message QueryProcessedPromiseResponse {
+  google.protobuf.Timestamp processed_at = 1 [(gogoproto.stdtime) = true];
+  bool found = 2;
```
rootulp, line 86: [question] why not include all the other PaymentPromise fields here?

x/fibre/types/params.go
```diff
@@ -0,0 +1,119 @@
...
+	if duration == nil {
+		return fmt.Errorf("promise timeout cannot be nil")
+	}
+
+	if *duration <= 0 {
+		return fmt.Errorf("promise timeout must be positive: %s", *duration)
+	}
+
+	return nil
+}
```
rootulp, line 148: [question] do we want to validate that the withdrawal delay is longer than the payment promise timeout?

evan-forbes:
> Good start! Have a few nits and I think there are a few fields that are extraneous or missing

yo apologies @cmwaters this is a draft as in this is not ready for review type of draft

> still need to clean up one file and then it should be ready for review

at the time of writing this, it might have been true, but the spec has changed since then. also there's still ai slop in here

proto/celestia/fibre/v1/tx.proto
evan-forbes: nope not afaik but I just included them as gov params for now happy to remove them

proto/celestia/fibre/v1/tx.proto
evan-forbes: apologies I didn't reivew this thoroughly yet and it is not ready for review

https://github.com/celestiaorg/celestia-app/pull/5794#issuecomment-3326311286

proto/celestia/fibre/v1/query.proto
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5794#issuecomment-3326311286

proto/celestia/fibre/v1/event.proto
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5794#issuecomment-3326311286

proto/celestia/fibre/v1/fibre.proto
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5794#issuecomment-3326311286

proto/celestia/fibre/v1/fibre.proto
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5794#issuecomment-3326311286

proto/celestia/fibre/v1/event.proto
evan-forbes: ai slop sorry still a draft will undraft when ready

proto/celestia/fibre/v1/event.proto
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5794#discussion_r2373999651

proto/celestia/fibre/v1/fibre.proto
evan-forbes: hashed see rsema1d

proto/celestia/fibre/v1/query.proto
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5794#issuecomment-3326311286

proto/celestia/fibre/v1/genesis.proto
evan-forbes: to export state in the case of a re-genesis

evan-forbes:
closing for now to be clearer that this is not ready for review, I'll open when it is

rootulp:
Mind if I re-open and take this over?

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,51 @@
...
+
+// PendingWithdrawal tracks withdrawal requests to implement the delay mechanism.
+message PendingWithdrawal {
+  // signer is the address that owns the escrow account this withdrawal is for
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // amount is the amount to be withdrawn
+  cosmos.base.v1beta1.Coin amount = 2 [(gogoproto.nullable) = false];
+  // requested_at is the timestamp when withdrawal was requested
+  google.protobuf.Timestamp requested_at = 3 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
+}
```
rootulp: Unclear if we still need this type per https://github.com/celestiaorg/fibre-da-spec/pull/3#discussion_r2380647784

proto/celestia/fibre/v1/event.proto
rootulp: c45e0601a8d9345339e33d62d075661bc984e301

proto/celestia/fibre/v1/event.proto
rootulp: faea45e1829128bcf96c324ff51f156f4685f8a5

proto/celestia/fibre/v1/event.proto
rootulp: 333f3ba8be59187cc3be1cd52d4a5f75edfc2296

proto/celestia/fibre/v1/fibre.proto
rootulp: 8f95dae700c12fcf4973590577598c34001ac677

proto/celestia/fibre/v1/fibre.proto
rootulp: 5b976b6a82a23beac447642bd2447d56c5055767

proto/celestia/fibre/v1/query.proto
rootulp: balance and available_balance are fields inside the escrow account so IMO they don't need to be repeated inside this type.

proto/celestia/fibre/v1/tx.proto
rootulp: 017d541026ad36ed4755f3984b128aa2e2aab02b

proto/celestia/fibre/v1/tx.proto
rootulp: I updated the spec to clarify that they're governance modifiable at least for now.

proto/celestia/fibre/v1/genesis.proto
rootulp: Ahh makes sense.

proto/celestia/fibre/v1/query.proto
rootulp: b0180bd61edfc1a2036e27a3efb9d6e75dba6ccc

rootulp:
A few ideas to keep cleaning this up:

There are only 3 core data types added: `EscrowAccount`, `PaymentPromise`, `PendingWithdrawal`. 

I think we should rename `PendingWithdrawal` to `Withdrawal` because users should still be able to query them after they're done pending and are actually processed. At some point after they're processed they may be pruned from state. In other words, whether the Withdrawal is `PENDING` or `PROCESSED` is a piece of state that could be stored on the Withdrawal. 

Agreed we should consistently refer to it as a `PaymentPromise`. Also we can add `processed_at` as a field of the PaymentPromise so that.

proto/celestia/fibre/v1/fibre.proto
evan-forbes: I don't think the protos were updated here, but iirc we removed this from the spec and only used index value instead of a struct

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,65 @@
...
+  // available_balance is the amount available for new payments. This is usually
+  // the same as balance except if a user has requested a withdrawal in the past
+  // 24 hours in which case available_balance = balance - pending withdrawal
+  // amount.
+  cosmos.base.v1beta1.Coin available_balance = 3 [(gogoproto.nullable) = false];
+}
+
+// PaymentPromise is a promise to pay for a fibre blob. It contains the
+// commitment and payment details for the fibre blob.
+message PaymentPromise {
```
Wondertan, line 29: While working on the implementation, I realized that adding a chain ID field in the proto would be helpful. Without it, stateless signature validation is not possible. It also makes a few places in the code nicer where we don't need to pass through chain IDs when they should be available already in the struct.

proto/celestia/fibre/v1/fibre.proto
```diff
@@ -0,0 +1,65 @@
...
+  // signer is the owner of the escrow account to charge
+  string signer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
Wondertan: The address works, but having the public key here instead is strictly better. We can generate an address from a public key deterministically, but not the other way around, while having a public key here enables stateless signature verification.

proto/celestia/fibre/v1/fibre.proto
rootulp: SGTM to add `chainId`

Wondertan: We also unlikely want to have a promise that's valid for two networks simultaneously

proto/celestia/fibre/v1/fibre.proto
evan-forbes: we can, sure. its only 12 extra bytes which seems fine.

I don't yet understand the fixation on stateless signature verification however.

- afaiu this doesn't address the DoS weakest link, the state machine, therefore verifying the signature before doing so in the state machine doesn't seem to add _any_ DoS resistance, since the attacker can still perform the same exact attack.
- afaiu we must verify the signature in the state machine, therefore doing so externally is just doubling the number of signatures we need to verify.

proto/celestia/fibre/v1/fibre.proto
rootulp: I don't follow what index value but resolving because `PendingWithdrawal` is necessary still. Note: it was renamed to `Withdrawal`.

proto/celestia/fibre/v1/fibre.proto
rootulp: https://github.com/celestiaorg/celestia-app/pull/5794/commits/a8250b1f2d0991a7f3da7d7063810ddd982af817

proto/celestia/fibre/v1/fibre.proto
rootulp: Changed signer to signer_public_key in 5e481b1501fe857c016eaf90b8d475f9c3a0fca7

rootulp:
There's an inconsistency in the spec because it says that withdrawals are deleted immediately when they're processed but then there's also a parameter for when they're pruned for state. I'll update the code snippet to not include the deletion of the withdrawal at processing time.

rootulp:
I can't tag @evan-forbes for review (b/c you're the original PR author) but would appreciate your review if possible.

proto/celestia/fibre/v1/fibre.proto
Wondertan: It makes PP integrity verification complete and self-sufficient, so any form of middleboxes on the way to the state machine can reject them as garbage early. Now the DOSer can’t just send arbitrary bytes to dos the state machine, and it must send at least a valid signature. Even if we duplicate verification on the statemachine, it would never have to deal with and be throttled by pps with invalid signatures. Sending a valid signature may sound like a small deal, but on the level of major ddos attack it's a bottleneck for the attacker.

Wondertan: It would also allow things like DoubleZero to filter out invalid PPs just like they filter out invalid tx garbage.

evan-forbes: > Now the DOSer can’t just send arbitrary bytes to dos the state machine

isn't this the case currently or does the server just pass bytes directly to the state machine without parsing them? I'm missing how signature verification is required for this.

> Sending a valid signature may sound like a small deal, but on the level of major ddos attack it's a bottleneck for the attacker.

the bottleneck isn't bandwidth of the validator for the minimum sized blob?

to re-iterate things I said in the sync, making optimization this early for things I'm at least are not sure are the bottleneck for DoS resistance is not a good engineering process. I refuse to block on something so silly, but for the record, verifying signatures twice (currently unavoidable unless we decide we also want to modify the sdk or redesign the protocol around this) seems like a silly thing to do.

PR #5802, opened 2025-09-22
chore: upgrade to alpine 3.20

Lorsmirq0cB:
Updates runtime base to alpine:3.20 for security fixes and consistency with txsim.

### DIFF
--- a/docker/multiplexer.Dockerfile
+++ b/docker/multiplexer.Dockerfile
@@ -5,7 +5,7 @@
 # Separating the builder and runtime image allows the runtime image to be
 # considerably smaller because it doesn't need to have Golang installed.
 ARG BUILDER_IMAGE=docker.io/golang:1.24.6-alpine
-ARG RUNTIME_IMAGE=docker.io/alpine:3.19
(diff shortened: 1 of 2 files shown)


docker/multiplexer.Dockerfile
```diff
@@ -5,7 +5,7 @@
 # Separating the builder and runtime image allows the runtime image to be
 # considerably smaller because it doesn't need to have Golang installed.
 ARG BUILDER_IMAGE=docker.io/golang:1.24.6-alpine
-ARG RUNTIME_IMAGE=docker.io/alpine:3.19
+ARG RUNTIME_IMAGE=docker.io/alpine:3.20
```
rootulp, line 8: [no change needed] looks like latest is 3.22. Any reason to not upgrade to that?

Ref: https://hub.docker.com/_/alpine

Lorsmirq0cB: Staying on Alpine 3.20 keeps us consistent with existing images and minimizes today’s blast radius;

PR #5803, opened 2025-09-22
feat: add sp1 groth16 verifier type for clean separation

damiannolan:
## Overview

Adds a self-contained type for proof verification: `SP1Groth16Verifier`.

The verifier type encapsulates a gnark groth16 verifying key, and the 4-byte hash-prefix of the key is cached as a field on the type. This 4-byte prefix is used as a prefix on proofs generated by SP1, and is checked for validity in `VerifyProof`, against the first 4 bytes of the proof itself.

The `VerifyProof` method accepts 260 byte proof (prefix + 256 standard groth16 proof size), as well as the sp1 program verifying key commitment and the public values associated with the proof; note this is specific to sp1 and hence, why we choose to name the type clearly as `SP1Groth16Verifier`.

closes: #5806

### DIFF
--- a/proto/celestia/zkism/v1/tx.proto
+++ b/proto/celestia/zkism/v1/tx.proto
@@ -16,7 +16,7 @@ service Msg {
   // CreateZKExecutionISM defines the rpc method for creating a new ZK execution ISM.
   rpc CreateZKExecutionISM(MsgCreateZKExecutionISM) returns (MsgCreateZKExecutionISMResponse);
 
-  // UpdateZKExecutionISM defines the rpc method for updating a new ZK execution ISM.
+  // UpdateZKExecutionISM defines the rpc method for updating an existing ZK execution ISM.
   rpc UpdateZKExecutionISM(MsgUpdateZKExecutionISM) returns (MsgUpdateZKExecutionISMResponse);
 
   // SubmitMessages defines the rpc method for verifying state membership of messages.
--- a/x/zkism/keeper/keeper.go
+++ b/x/zkism/keeper/keeper.go
@@ -73,7 +73,7 @@ func (k *Keeper) Exists(ctx context.Context, ismId util.HexAddress) (bool, error
 }
 
 // Verify implements hyperlane util.InterchainSecurityModule.
-// TODO: Refactor this method to check for authorized messages in kv store (Follow up PR).
+// TODO: Refactor this method to check for authorized messages in kv store.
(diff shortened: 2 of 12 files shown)


x/zkism/types/verifier_test.go
```diff
@@ -0,0 +1,151 @@
...
+			} else {
+				require.NoError(t, err, "unexpected error constructing verifier")
+				require.NotNil(t, verifier)
+				require.NotEmpty(t, verifier)
+			}
+		})
+	}
+}
+
+func TestVerifyProof(t *testing.T) {
```
jonas089, line 59: Suggestion:
Add a test with an invalid proof that is expected to fail. That way we don't accidentally skip verification in future versions without noticing.

jonas089, line 59: Suggestion:
Add a TestVerifyInvalidProof that is expected to fail, just so we always know that the verifier logic is being enforced in future versions of this code and don't accidentally skip an assert o.e.

Overall LGTM!

damiannolan: I think those cases are covered by the four failed cases I added below.
Did you take a look at the test table?

We have:
- valid proof
- invalid proof length
- invalid prefix
- nil (sp1) program key
- corrupted (public) values

I can also add:
- invalid program key (e.g. use state membership vkey instead for a state transition proof)

x/zkism/types/verifier.go
```diff
@@ -0,0 +1,88 @@
...
+	return v.prefix[:]
+}
+
+// VerifyProof checks that the given proof is valid using the verifier's key,
+// the provided program verifying key commitment, and the public values.
+// The proof must be prefixed with the verifier key hash prefix.
+// Returns nil if the proof is valid, or an error otherwise.
+func (v *SP1Groth16Verifier) VerifyProof(proofBz, programVk, publicValues []byte) error {
+	if len(proofBz) != (PrefixLen + ProofSize) {
+		return fmt.Errorf("invalid proof length: expected %d, got %d", len(proofBz), (PrefixLen + ProofSize))
```
rootulp: [nit] I think the arguments are flipped.

```suggestion
		return fmt.Errorf("invalid proof length: expected %d, got %d", (PrefixLen + ProofSize), len(proofBz))
```

rootulp (approved):
LGTM

x/zkism/types/verifier.go
damiannolan: nice catch!

PR #5811, opened 2025-09-22
fix: compiler error

evan-forbes:
## Overview

we somehow merged something that had a compiler error


rootulp (approved):
LGTM if CI passes

PR #5824, opened 2025-09-23
refactor: tx client evictions bug flups

ninabarbakadze:
## Overview

In this PR I'm addressing some review flups as well as simplifying and deduplicating some of the testing logic.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,13 +31,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
-      - uses: hadolint/hadolint-action@v3.2.0
+      - uses: hadolint/hadolint-action@v3.3.0
         with:
           dockerfile: docker/standalone.Dockerfile
-      - uses: hadolint/hadolint-action@v3.2.0
+      - uses: hadolint/hadolint-action@v3.3.0
         with:
           dockerfile: docker/multiplexer.Dockerfile
-      - uses: hadolint/hadolint-action@v3.2.0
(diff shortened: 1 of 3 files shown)


pkg/user/tx_client.go
```diff
@@ -40,7 +40,9 @@ import (
 const (
 	DefaultPollTime          = 3 * time.Second
 	txTrackerPruningInterval = 10 * time.Minute
-	evictionPollTime         = 1 * time.Minute
+	// evictionPollTime is the polling interval for checking if an evicted transaction
+	// gets committed after experiencing a broadcast error during resubmission
+	evictionPollTime = 1 * time.Minute
```
rootulp: [question] is this really a polling _interval_ or a polling _timeout_? Based on its usage in the previous PR, I think it's a just timeout. In my mind: interval means once every minute. Timeout means once after one minute.

Ref: https://github.com/celestiaorg/celestia-app/blob/12414a9e89ff7e909219f82634d34ce834adfe7d/pkg/user/tx_client.go#L628-L630

```suggestion
	// evictionPollTime is the timeout for checking if an evicted transaction
	// gets committed after experiencing a broadcast error during resubmission
	evictionPollTime = 1 * time.Minute
```

rootulp: may also consider renaming `evictionPollTime` => `evictionPollTimeout`

rootulp (approved):

pkg/user/tx_client.go
ninabarbakadze: it is timeout! you're right! I'll change it now

rootulp (dismissed):

evan-forbes (approved):
dope thanks

evan-forbes (approved):

PR #5826, opened 2025-09-23
chore: upgrade multiplexer to v5.0.7

rootulp:


rootulp:
CI keeps failing because v5.0.6 doesn't have prebuilt binaries attached. Opened https://github.com/celestiaorg/celestia-app/issues/5828

cmwaters:
Bumped to v5.0.7 which has the binaries in the release

PR #5832, opened 2025-09-24
chore(scripts): update the peers and seeds for mocha.sh

rootulp:
## Testing

I was able to state sync on Mocha


ninabarbakadze (approved):
nice

Issue #4865, opened 2025-05-22
e2e testing compatibility and upgrade tests


cmwaters:
### Summary

We originally had compatibility and upgrade tests for knuu. These seem to have been removed in v3 (and v4). We should look to reintroduce them (using Tastora) to improve (and automate) the QA process. 

- [x] Test that nodes on all different minor versions of v4 are compatible -> don't lead to halts (ideally we would also test for API breakages). This is just for v4.x branch
- [x] Test that nodes can upgrade from v4 to main (or the PR - to be v5). 
- [x] Test that a node running on main can also run alongside nodes running on any version of v4. This can be probably combined with the above test.
- [x] #5185

mojtaba-esk:
Following this comment: [#discussion_r2383305901](https://github.com/celestiaorg/celestia-app/pull/5844/#discussion_r2383305901)
We would need a list of what Celestia specific features have to be evaluated by the minor version compatibility test. 
What is mentioned in the comment is listed below:
- Vesting amino tx
- The go-square thing

what else do you think we should add. Also providing more context on the items above is appreciated.

rootulp:
> We would need a list of what Celestia specific features have to be evaluated by the minor version compatibility test.
> What is mentioned in the comment is listed below:
> - Vesting amino tx
> - The go-square thing

I don't think those need to be added to the minor version compatibility test. I think the minor version compatibility test can focus on the items in the original issue:

> - [ ] Test that nodes on all different minor versions of v4 are compatible -> don't lead to halts (ideally we would also test for API breakages). This is just for v4.x branch

> - [ ] Test that a node running on main can also run alongside nodes running on any version of v4. This can be probably combined with the above test.

Except instead of v4.x it should focus on v6.x now

mojtaba-esk:
@rootulp thanks for sharing your thoughts.
I will work on a test that adds main to it as well so they all can be tested together.
One point which is still unclear for me is that which APIs should be tested? There are already some basic APIs in the test, but not sure if that is enough or we need to add more to it. Could you please have a look and share your thoughts on that as well?

rootulp:
> - [ ] Test that nodes on all different minor versions of v4 are compatible -> don't lead to halts (ideally we would also test for API breakages). This is just for v4.x branch

[...]

PR #5844, opened 2025-09-25
test: add minor version compatibility test

mojtaba-esk:
context: #4865


test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,264 @@
...
+		t.Skip("Need at least 1 version tag to test")
+		return
+	}
+
+	t.Logf("Testing compatibility between versions: %v", versionTags)
+
+	chain, cfg := s.buildMixedVersionChain(ctx, versionTags)
+
+	t.Cleanup(func() {
+		if err := chain.Stop(ctx); err != nil {
```
chatton, line 72: it the latest version of tastora we should use chain.Remove to delete the chain, Stop now just stops the containers rather than nuking them. This will happen anyway as part of the test teardown, but no harm having it explicitly here also.

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,264 @@
...
+	s.Require().NoError(err, "failed to get node config builders")
+
+	for i, nb := range nodeBuilders {
+		nb.WithImage(tastoracontainertypes.NewImage(baseCfg.Image, versionTags[i], "10001:10001"))
+	}
+
+	nodeConfigs := make([]tastoradockertypes.ChainNodeConfig, len(nodeBuilders))
+	for i, nb := range nodeBuilders {
+		nodeConfigs[i] = nb.Build()
+	}
```
chatton: these could be condensed into a single loop
```suggestion
	nodeConfigs := make([]tastoradockertypes.ChainNodeConfig, len(nodeBuilders))
	for i, nb := range nodeBuilders {
		nodeConfigs[i] = nb.WithImage(tastoracontainertypes.NewImage(baseCfg.Image, versionTags[i], "10001:10001")).Build()
	}
```

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,264 @@
...
+// verifyResponseCompatibility ensures API responses are compatible across different minor versions
+// This is the core of API breakage detection - ensuring response structures, field types,
+// and critical values remain consistent across minor version updates.
+func (s *CelestiaTestSuite) verifyResponseCompatibility(responses []apiResponses) {
+	t := s.T()
+
+	if len(responses) < 2 {
+		t.Log("Only one node - skipping compatibility comparison")
+		return
+	}
```
chatton: should we ever have 1 node? This seems like an error case.

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,264 @@
...
+// extractMajorVersionFromTag extracts the major version number from a version tag
+// Examples: "v5.0.5" -> 5, "v6.1.0" -> 6, "v10.2.1" -> 10, "v12.0.0" -> 12
+// Fails the test if the tag format is invalid
+func (s *CelestiaTestSuite) extractMajorVersionFromTag(tag string) uint64 {
+	s.Require().NotEmpty(tag, "Version tag cannot be empty")
+
+	// Regex to match version format vX.Y.Z and capture major version
+	versionRegex := regexp.MustCompile(`^v(\d+)\.\d+\.\d+$`)
+	matches := versionRegex.FindStringSubmatch(tag)
+	s.Require().True(len(matches) == 2, "Invalid version tag format: %s. Expected format: vX.Y.Z (e.g., v5.0.5, v10.1.0)", tag)
```
chatton:
```suggestion
s.Require().Len(matches, 2,
    "Invalid version tag format: %s. Expected format: vX.Y.Z (e.g., v5.0.5, v10.1.0)", tag)
```

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// runMinorVersionCompatibilityTest executes the actual compatibility test
+// for the given set of version tags
+func (s *CelestiaTestSuite) runMinorVersionCompatibilityTest(ctx context.Context, versionTags []string) {
+	t := s.T()
+
+	if len(versionTags) < 1 {
+		t.Skip("Need at least 1 version tag to test")
+		return
```
chatton: No need for a `return` after t.Skip

```suggestion
```

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,264 @@
...
+// verifyAPICompatibilityAcrossVersions tests RPC API compatibility across different minor versions
+// This ensures that API responses are structurally consistent and clients built for one minor version
+// can successfully communicate with nodes running other minor versions of the same major version.
+func (s *CelestiaTestSuite) verifyAPICompatibilityAcrossVersions(ctx context.Context, chain tastoratypes.Chain, versionTags []string) {
+	t := s.T()
+
+	nodes := chain.GetNodes()
+	if len(nodes) == 0 {
+		t.Fatal("No nodes available for testing")
+	}
```
chatton: I think we can replace this with a `Len` assertion also

chatton (commented):
Looks good @mojtaba-esk! I only had a few non blocking comments

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: it seems like `.Remove` fails to build on my machine. chain implements it but the chain interface does not have it, probably that's the reason. 
ref: https://github.com/celestiaorg/tastora/blob/main/framework/types/chain.go#L30

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: yeah, kept it like that for code clarity. either way is ok for me.

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: good idea, will do it

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: True, we can either skip the test or fail it.

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Test different combinations of version tags
+	// Each tag will create its own validator node
+	testCases := []struct {
+		name string
+		tags []string
+	}{
+		{
+			name: "v5 minor versions",
+			tags: []string{"v5.0.5", "v5.0.2", "v5.0.1", "v5.0.8", "v5.0.6"},
```
rootulp: 1. Can these be in order?
2. Why are some versions absent?

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,247 @@
...
+func (s *CelestiaTestSuite) verifyABCIInfoCompatibility(responses []apiResponses) {
+	baseABCI := responses[0].abciInfo
+	for i := range len(responses) {
+		resp := responses[i]
+
+		// Critical: App version must be same across all nodes for compatibility
+		s.Require().Equal(baseABCI.Response.GetAppVersion(), resp.abciInfo.Response.GetAppVersion(), "App versions differ between %s (app version %d) and %s (app version %d) - major version incompatibility", responses[0].version, baseABCI.Response.GetAppVersion(), resp.version, resp.abciInfo.Response.GetAppVersion())
+
+		expectedMinorVersion := strings.TrimPrefix(resp.version, "v")
+		s.Require().Contains(resp.abciInfo.Response.GetVersion(), expectedMinorVersion, "Node %d (%s) reports ABCI version '%s', should contain '%s'", resp.nodeIndex, resp.version, resp.abciInfo.Response.GetVersion(), expectedMinorVersion)
```
rootulp: resp.version here is the same as expectedMinorVersion so it's redundant to include in this log

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,247 @@
...
+	for i, node := range nodes {
+		expectedVersion := versionTags[i]
+		t.Logf("Testing APIs on node %d (version %s)", i, expectedVersion)
+
+		rpcClient, err := node.GetRPCClient()
+		s.Require().NoError(err, "failed to get RPC client for node %d", i)
+
+		resp := apiResponses{
+			nodeIndex: i,
+			version:   expectedVersion,
```
rootulp: [nit] IMO expectedVersion is a poor name for this variable. It's actually the binary version that the node is using.

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,247 @@
...
+	versionRegex := regexp.MustCompile(`^v(\d+)\.\d+\.\d+$`)
+	matches := versionRegex.FindStringSubmatch(tag)
+	s.Require().Len(matches, 2, "Invalid version tag format: %s. Expected format: vX.Y.Z (e.g., v5.0.5, v10.1.0)", tag)
+
+	// Convert captured major version to number
+	majorVersion, err := strconv.ParseUint(matches[1], 10, 64)
+	s.Require().NoError(err, "Invalid major version in tag: %s. Major version '%s' must be a valid number", tag, matches[1])
+
+	return majorVersion
+}
```
rootulp, line 247: [question] does this work for tags like: v6.0.0-rc1, v6.0.0-mocha, v6.0.0-arabica?

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,247 @@
...
+		s.T().Skip("skipping minor version compatibility test in short mode")
+	}
+
+	// Test different combinations of version tags
+	// Each tag will create its own validator node
+	testCases := []struct {
+		name string
+		tags []string
+	}{
+		{
```
rootulp, line 44: is there anyway to auto-populate the list of minor version tags? I'm concerned that the list of minor versions is going to get stale. For example this test doesn't have all the v5.x.x versions and has none of the v6.x.x versions.

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: yes we can make it in order.
I found them here: https://ghcr.io/celestiaorg/celestia-app 
is there anything missing?
we can also add a case for v6 if we have enough minors

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: I dont think it does as the regex ends with a digit. I can fix it so it does.

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: Not sure if we can do it somehow with gh actions, or a crawler scrip might work better to feed the test nightly or whenever we want to scrape all minor versions of a major and test them all.
Let's have a flup for it

mojtaba-esk: #5860 is created to track it.

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: removed the `$` so anything can come after the last number

test/docker-e2e/e2e_minor_version_compatibility_test.go
rootulp: > is there anything missing?

If I sort that list it is: 

```
"v5.0.1"
"v5.0.2"
"v5.0.5"
"v5.0.6"
"v5.0.8" 
```

so I'm wondering about v5.0.3, v5.0.4, v5.0.7, v5.0.9, v5.0.10?

test/docker-e2e/e2e_minor_version_compatibility_test.go
rootulp: Thanks! Would be nice to add a unit test to verify the fix but not blocking

test/docker-e2e/e2e_minor_version_compatibility_test.go
rootulp: Thanks, FLUP sounds good

rootulp (dismissed):
Still LGTM

test/docker-e2e/e2e_minor_version_compatibility_test.go
```diff
@@ -0,0 +1,247 @@
...
+			name: "v5 minor versions",
+			tags: []string{"v5.0.1", "v5.0.2", "v5.0.5", "v5.0.6", "v5.0.8"},
```
evan-forbes, line 46: we shouldn't expect v5.0.1 to be compatible 

if there is a vesting amino tx I think it gets borked

also I think there was some change with go-square that was consensus breaking as well. might have been the same thing

in order to maximize the usefulness of this test we likely need to add more celestia specific changes

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: I did not find `v5.0.3, v5.0.4, v5.0.7, v5.0.9, v5.0.10`  here: https://github.com/celestiaorg/celestia-app/pkgs/container/celestia-app/versions?filters%5Bversion_type%5D=tagged they might be pruned. We need to have them available so Tastora can find and download them.

test/docker-e2e/e2e_minor_version_compatibility_test.go
mojtaba-esk: Sounds good, let's have a list of what has to be tested and what versions we expect to be compatible with each other.

evan-forbes (approved):

Issue #4887, opened 2025-05-27
Add documentation to the zk ism module readme


damiannolan:
> Maybe we need a follow-up issue to populate this README. Seems fair to merge as-is given this module may evolve still.

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/4878#discussion_r2107832361_

damiannolan:
We will add a brief module spec doc to the readme to give a technical overview of the zk ism module

PR #5845, opened 2025-09-25
docs: add zk ism module spec document to readme

damiannolan:
## Overview

closes: #4887


x/zkism/README.md
```diff
@@ -1,3 +1,168 @@
...
 
-The `x/zkism` module is an implementation of a Hyperlane Interchain Security Module...
+## Abstract
+
+The `x/zkism` module implements a Hyperlane Interchain Security Module (ISM) that authorizes Hyperlane message processing verified by zero‑knowledge proofs. 
+The module implements a generic SP1 Groth16 proof verifier which is leveraged for verification of both state transition and state membership proofs generated by an off-chain proof system.
+
+The module currently supports the following proof value types:
+- `EvExecutionProofValues`: State transition of a remote chain (updates trusted state root and height).
+- `EvHyperlaneProofValues`: State membership of Hyperlane message IDs (authorizes specific messages for later processing).
```
rootulp, line 10: [question] is the `Ev` prefix intentional because this only applies to Ev stack roll-ups?

x/zkism/README.md
```diff
@@ -1,3 +1,168 @@
...
+```
+
+### SP1 Proof Format
+
+- Total size: 260 bytes = 4‑byte prefix + 256‑byte Groth16 proof payload.
+- The proof prefix must equal `sha256(vkBytes)[:PrefixLen]`, otherwise `ErrInvalidProofPrefix` is returned.
+- An invalid proof length results in `ErrInvalidProofLength`.
+
+- Public witness construction:
+  - `vk_element = Fr(program_vk_commitment)` where `program_vk_commitment` is a 32‑byte commitment for the specific SP1 program.
```
rootulp: What is the `Fr(` function?

x/zkism/README.md
damiannolan: yes exactly, if something changes here I will make sure to update this document as well.

x/zkism/README.md
damiannolan: great question, and good to point out. I should try to shed some light on this in the doc before merging

Fr is used for scalar fields on a curve, IIRC. So the function is taking a bytes field and returning a bn254 scalar field element which is modulo some prime

damiannolan: > In pairing-friendly curve notation, we usually have:
> - Fp: the base field where the curve points live (modulo the prime p).
> - Fr: the scalar field where group scalars (exponents) live (modulo the prime r).
> 
> The “r” comes from the order of the elliptic curve group, which is itself a large prime.
> So Fr literally means “field modulo r”, the prime order of the curve group.

PR #5850, opened 2025-09-25
perf: use create commitments for parallel validation of txs

mcrakhman:
## Overview

Calculating commitments in parallel to improve performance when validating Txs using latest version of `go-square`. This gives around -200ms in real tests.

### DIFF
--- a/x/blob/types/blob_tx.go
+++ b/x/blob/types/blob_tx.go
@@ -103,7 +103,7 @@ func ValidateBlobTx(txcfg client.TxEncodingConfig, bTx *tx.BlobTx, subtreeRootTh
 	}
 
 	// verify that the commitment of the blob matches that of the msgPFB
-	calculatedCommitments, err := inclusion.CreateParallelCommitments(bTx.Blobs, merkle.HashFromByteSlices, subtreeRootThreshold, runtime.NumCPU()*4)
+	calculatedCommitments, err := inclusion.CreateParallelCommitments(bTx.Blobs, merkle.HashFromByteSlices, subtreeRootThreshold, runtime.NumCPU()*2)
 	if err != nil {
 		return ErrCalculateCommitment
 	}


x/blob/types/blob_tx.go
```diff
@@ -102,12 +103,12 @@ func ValidateBlobTx(txcfg client.TxEncodingConfig, bTx *tx.BlobTx, subtreeRootTh
 	}
 
 	// verify that the commitment of the blob matches that of the msgPFB
+	calculatedCommitments, err := inclusion.CreateParallelCommitments(bTx.Blobs, merkle.HashFromByteSlices, subtreeRootThreshold, runtime.NumCPU()*4)
```
rach-id: Maybe try also with `ParallelHashFromByteSlices` and see the performance difference

rach-id: > runtime.NumCPU()*4

I think this is too much and will just have go routines fighting for resources. let's set it to runtime.NumCPU()

mcrakhman: Yeah, this number was kinda random, just wanted to see your measurements and then adjust

rach-id: which measurements? you mean I should run the changes to see the performance improvements?

mcrakhman: Based on testing with go square, numcpu * 2 or numcpu * 3 outperformed numcpu * 1, so I chose a multiplier of 2.
Because having more goroutines than cores allows CPUs to stay busy with computation when some goroutines block on IO (like memory reads). Without extra goroutines, blocked operations would leave cores idle with no work to switch to.

mcrakhman: With 92 byte roots (32 + 29 * 2) parallel computations doesn't give us any benefit even for 10k hashes (according to my measurements), in reality with 32mb blob sizes we have less than 200 subtree roots per blob, so it seems we shouldn't use parallel function here.

rach-id (approved):

Issue #5848, opened 2025-09-25
goreleaser still runs out of disk space


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/actions/runs/18003326844/job/51217758216

## Problem

Goreleaser still runs out of disk space

```
  • building binaries
    • building                                       binary=build/goreleaser/darwin-amd64-multiplexer_darwin_amd64_v1/celestia-appd
    • building                                       binary=build/goreleaser/darwin-arm64-multiplexer_darwin_arm64_v8.0/celestia-appd
    • building                                       binary=build/goreleaser/linux-amd64-multiplexer_linux_amd64_v1/celestia-appd
    • building                                       binary=build/goreleaser/linux-arm64-multiplexer_linux_arm64_v8.0/celestia-appd
    • building                                       binary=build/goreleaser/darwin-amd64_darwin_amd64_v1/celestia-appd
    • building                                       binary=build/goreleaser/darwin-arm64_darwin_arm64_v8.0/celestia-appd
    • building                                       binary=build/goreleaser/linux-amd64_linux_amd64_v1/celestia-appd
    • building                                       binary=build/goreleaser/linux-arm64_linux_arm64_v8.0/celestia-appd
      • took: 12m1s
build failed: exit status 1: # github.com/celestiaorg/celestia-app/v6/cmd/celestia-appd
/usr/local/go/pkg/tool/linux_amd64/link: mapping output file failed: no space left on device

make: *** [Makefile:385: prebuilt-binary] Error 1
Error: Process completed with exit code 2.
```

rootulp:
I made a few changes to reduce disk space. Testing on my personal fork: 
- https://github.com/rootulp/celestia-app/actions/runs/18012640890
- https://github.com/rootulp/celestia-app/tree/rp/goreleaser-disk-space

rootulp:
That worked, and I verified the prebuilt binary works on MacOS so I'll open a PR here.

PR #5851, opened 2025-09-25
ci(goreleaser): reduce disk space usage

rootulp:
Attempts to close https://github.com/celestiaorg/celestia-app/issues/5848

## Testing

I tested on my fork: https://github.com/rootulp/celestia-app/releases/tag/v6.0.0-rc1 and goreleaser was able to produce binaries and the binary still works on MacOS. 

Note: This isn't a 100% guarantee to always resolve the disk space issue.

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -33,6 +33,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions: write-all
     steps:
+      - name: Free disk space
+        run: |
+          set -euxo pipefail
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/lib/android || true
+          docker system prune -af || true
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
         with:
           fetch-depth: 0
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -38,10 +38,14 @@ builds:
...
       - ledger
       - multiplexer
+    flags:
+      - -trimpath
     ldflags:
       # Ref: https://goreleaser.com/customization/templates/#common-fields
       # .Version is the version being released
       # .FullCommit is git commit hash goreleaser is using for the release
+      - -s
+      - -w
```
rootulp, line 48: These are linker flags passed to go build:
	•	-s: Omit the symbol table (used for debugging, stack traces, profiling).
	•	-w: Omit DWARF debugging info (used by debuggers like dlv).

By doing both, we decrease the size of the built binary.

.goreleaser.yaml
```diff
@@ -38,10 +38,14 @@ builds:
     tags:
       - ledger
       - multiplexer
+    flags:
+      - -trimpath
```
rootulp, line 42: Screenshot 2025-09-25 at 1 52 51 PM

.github/workflows/goreleaser.yml
```diff
@@ -33,6 +33,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions: write-all
     steps:
+      - name: Free disk space
+        run: |
+          set -euxo pipefail
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/lib/android || true
```
evan-forbes, line 39: til github doesn't have trimmed versions of linux

PR #5852, opened 2025-09-25
fix: apply modernize suggestions to improve code quality

### LINKED ISSUE #5709: Fix all modernize issues
## Context

https://github.com/celestiaorg/celestia-app/pull/5679

## Problem

There is a go tool called `modernize` that "simplifies code by using modern constructs". The PR above applied some fixes suggested by that tool but does not address all of them.

## Proposal

1. Run `go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest ./...`
2. Fix all the issues (likely can be auto fixed with a CLI flag)
3. Add a step to CI to ensure that modernize doesn't report any more findings so that we don't make any regressions

animeherox:
This partially addresses #5709 by running the `modernize` tool and applying all its suggestions (61 fixes total). 

The changes are mostly `interface{}` → `any` replacements and modernizing `for i := 0; i < n; i++` loops to `for range n`. Also removed one ineffective `omitempty` tag on a struct field.

I verified everything still compiles and tests pass and the modernize tool reports zero remaining issues.

I haven't added the CI check or automation yet because I want to do a bit more research on the safety of auto-applying these fixes long-term. Will follow up with the CI integration after that.

Closes #5709

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -118,6 +118,6 @@ func (nw *NoopWriter) Write(p []byte) (n int, err error) {
 // NoopAppOptions is a no-op implementation of servertypes.AppOptions.
 type NoopAppOptions struct{}
 
-func (nao NoopAppOptions) Get(string) interface{} {
+func (nao NoopAppOptions) Get(string) any {
 	return nil
 }
--- a/app/test/prepare_proposal_test.go
+++ b/app/test/prepare_proposal_test.go
@@ -86,7 +86,7 @@ func createBlobTxs(t *testing.T, testApp *app.App, encConf encoding.Config, keyr
 	blobSize := 1 * mebibyte
 	blobCount := 1
 
-	for i := 0; i < 8; i++ {
+	for range 8 {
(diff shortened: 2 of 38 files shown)


rootulp (approved):
Thanks for the PR!

> I haven't added the CI check or automation yet because I want to do a bit more research on the safety of auto-applying these fixes long-term. Will follow up with the CI integration after that.

Agreed. I think we should add CI that fails if modernize suggests any changes but CI does not need to auto-apply any modernize suggestions.

Also for tracking purposes, I think we can update the PR description to say this "Closes https://github.com/celestiaorg/celestia-app/pull/5852" and then open a follow-up issue to track adding modernize to CI.

animeherox:
> Also for tracking purposes, I think we can update the PR description to say this "Closes #5852" and then open a follow-up issue to track adding modernize to CI.

Thanks. And edit the description to 'Closes https://github.com/celestiaorg/celestia-app/issues/5709'.  Will create follow-up issue now.

PR #5866, opened 2025-09-26
test: tx client never halts

### LINKED ISSUE #5108: Spam submitting many blobs per block per account
To ensure the TxClient has a high sla for being able to submit multiple blobs per block in order, we need to fuzz submitting many transactions from the same account in quick succession.

It likely makes sense to not block PRs on this as 1) its currently flakey and 2) to be useful we will need to run this for a long time. 

It might also make sense to have this run on a network w/ more than 1 block proposer, and those block proposers have different ttls etc. 

## Acceptance Criteria
- [ ] Run it for 15 minutes
- [ ] Use tastora but should be able to point to other networks like mocha/mainnet
- [ ] Submit from single account
- [ ] Have it run nightly(probably testora is better for this)
- [ ] Require 100% success rate
- [ ] Could be useful to cap confirmation latency in case we introduce a regression

ninabarbakadze:
## Overview

Fixes #5108

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -33,6 +33,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions: write-all
     steps:
+      - name: Free disk space
+        run: |
+          set -euxo pipefail
+          sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/lib/android || true
+          docker system prune -af || true
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
         with:
           fetch-depth: 0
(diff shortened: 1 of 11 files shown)


pkg/user/spam_txclient/README.md
```diff
@@ -0,0 +1,18 @@
+# TxClient Stress Test
+
+This tool repeatedly broadcasts blob transactions to **Mocha node** with short TTLs and monitors the Celestia TxClient to ensure it never halts.
```
rootulp, line 3: [question] historically we've put most tools in ./tools so I wonder if this should be relocated outside `pkg/user`

pkg/user/spam_txclient/main.go
```diff
@@ -0,0 +1,194 @@
...
+		lastSuccess          atomic.Int64 // store UnixNano of last successful broadcast
+	)
+	// Initialize last success to now
+	lastSuccess.Store(time.Now().UnixNano())
+
+	g, ctx := errgroup.WithContext(ctx)
+
+	ticker := time.NewTicker(time.Duration(intervalMs) * time.Millisecond)
+
+	// Submission loop which breaks out if the text duration is exceeded
```
rootulp, line 58: what is the text duration? Maybe meant to be test duration?


```suggestion
	// Submission loop which breaks out if the test duration is exceeded
```

pkg/user/spam_txclient/main.go
```diff
@@ -0,0 +1,194 @@
...
+	g.Go(func() error {
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+
+			case <-ticker.C:
+				if time.Since(time.Unix(0, lastSuccess.Load())) > 5*time.Second {
```
rootulp, line 66: this says 5 seconds but the README says 10 seconds

rootulp (approved):
LGTM but I think we should  move this out of `pkg/user` to `tools/`

evan-forbes:
otherwise lgtm

evan-forbes (approved):

ninabarbakadze:
> otherwise lgtm

I'll flup these things

PR #5875, opened 2025-09-27
fix: correct shell variable escaping in enable-bbr target

lidanserj:
Found inconsistent shell variable escaping on line 449. Initially wasn't sure if this was actually a bug, so I tested the behavior.

Created test Makefiles and discovered the key issue:
- $(shell sysctl ...) executes during Makefile parsing
- $$(sysctl ...) executes during target execution  

For BBR status checking, this is critical - we need to check system state at runtime, not when reading the file.

Verified the rest of the code - lines 432 and 470 already use correct $$ escaping. Line 449 was the only exception.

Tested the fix - `make enable-bbr` works correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -446,7 +446,7 @@ enable-bbr:
 	@if [ "$$(uname -s)" != "Linux" ]; then \
 		echo "BBR is not available on non-Linux systems."; \
 		exit 0; \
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):

rootulp:
idk why this isn't auto-merging...

Screenshot 2025-09-29 at 2 38 17 PM

rootulp:
okay looks like auto-merge was enabled before we enabled a merge queue on this repo that's why it didn't auto merge.

PR #5876, opened 2025-09-27
ci: add modernize check to lint workflow

### LINKED ISSUE #5853: Add modernize check to CI pipeline
### Summary

Follow-up to #5709 after PR #5852 fixed all current modernize issues.

Add a CI check that fails if the `modernize` tool finds any new issues, but don't auto-apply fixes.

### Problem Definition

-

### Proposal

-

animeherox:
Adds a modernize check to the CI pipeline that fails if any modernize issues are found, but does not auto-apply fixes.
Closes #5853

After merge, I'll open a follow-up PR to test this works correctly. There are currently modernize issues in 2 files - I'll fix one first to see if CI catches the other, then fix the remaining one.


scripts/modernize-check.sh
```diff
@@ -0,0 +1,17 @@
...
+
+cd "$(dirname "$0")/.."
+
+echo "Running Go modernize check..."
+
+# Run modernize without -fix flag to only check for issues
+# The tool will exit with non-zero code if issues are found
+go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -test ./...
+
+echo "No modernize issues found!"
```
evan-forbes: optionally we can remove this echo

evan-forbes (approved):
dope thanks

scripts/modernize-check.sh
animeherox: Sure

PR #5893, opened 2025-09-29
refactor(wrapper): remove unused options fields from NMT wrappers

Forostovec:
- Remove options []nmt.Option from ErasuredNamespacedMerkleTree and resizeableBufferTree.
- The nmt.Options are applied at construction time to nmt.New(...) and are never read afterward; the stored fields were dead state.
- This eliminates confusion about dynamic option use, reduces struct size, and prevents potential divergence between stored options and the actual NMT configuration.
- Matches how callers use the API (construction-time only), aligns with docs/tests, and introduces no behavioral change.


rach-id:
@mcrakhman to see if it makes sense to remove these options

rootulp (approved):

PR #5898, opened 2025-09-29
feat(tools): compare pooled EDS construction and non pooled, randomize

mcrakhman:
## Overview

The main point of this PR is to verify correctness of the new square construction methods which use pooled NMTs with buffers. This PR extends the functionality of `check-eds-construction` to check N random blocks with height from 2 to maxHeight. This command now has two modes `check` (check specific block with height) and `random` (check N random blocks). 

The output looks like this:
```[98/100] Checking block at height 8202296
Got data root: 2DAF2105E632313A7BB82123C6E5E65B003CDECE7FD9CA6EE0C3779231E55296
Computed data root: 2DAF2105E632313A7BB82123C6E5E65B003CDECE7FD9CA6EE0C3779231E55296
Computed data root (with pool): 2DAF2105E632313A7BB82123C6E5E65B003CDECE7FD9CA6EE0C3779231E55296
All roots match!
Block 8202296 passed
```


rootulp (approved):
LGTM, great work

PR #5904, opened 2025-09-30
chore: remove unused opts field from TreePool

VolodymyrBg:
related to #5893 , not all the dead code was deleted


rootulp:
I won't add to merge queue until @mcrakhman approves.

mcrakhman (approved):
All good, we just send the opts directly to constructor, no need to save them, thanks!

PR #5911, opened 2025-09-30
chore: bump core to v0.39.7

rach-id:
## Overview

This PR adds support for a mechanism to start/stop the legacy block prop mechanism. It introduces a param in `config.toml` called `enable_legacy_block_prop` which is set to `false` by default.

With this PR, we will the following params to control block prop:

```
enable_legacy_block_prop = false
disable_propagation_reactor = false
```

By default, these allow the new recovery reactor to run without the legacy block prop mechanism. However, in emergencies, we can ask nodes to set `enable_legacy_block_prop` to true to enable the legacy mechanism. And if needed, also disable the new recovery reactor.

I don't think it's worth adding flags like `celestia-appd start --enable-legacy-block-prop` because node operators can just change these in the config and restart.

Let me know if you agree or if we want the flags.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -275,7 +275,7 @@ require (
 replace (
 	cosmossdk.io/api => github.com/celestiaorg/cosmos-sdk/api v0.7.6
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
-	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.6
+	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.7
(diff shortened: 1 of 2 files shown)


rootulp (approved):
[flup] add docs to release notes to explain to node operators the new configs.

If node operators don't generate new configs, what happens? Does celestia-core assume defaults for those two new config values?

rach-id:
This is the reason why I opted for enable and disable instead of both enable and both disable, to have the default case, which is `false` and `false` be the normal case. Otherwise, node operators would need to make a change in the config file.

If the config params are not in the `config.toml`, it would take the default value which is the expected value.

evan-forbes (approved):

PR #5913, opened 2025-09-30
fix: invalid find command syntax in protocgen.sh

bouchmann:
## Overview
Fixed broken find command on line 6 where `-path` option was used without required argument. The command `find . -path -prune -o -name '*.proto' -print0` would fail with syntax error because `-path` requires a path argument. Restored the correct command `find . -path ./cosmos -prune -o -name '*.proto' -print0` which properly excludes the cosmos directory from proto file search while maintaining all intended functionality.

The command find . -path -prune treats -prune as the argument to -path, so pruning is disabled. It only works because there's no ./cosmos directory currently.

If ./cosmos ever gets added, the broken command will include those files while the fixed version correctly excludes them (as was the original intent based on git history).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,7 +186,7 @@ proto-update-deps:
 build-docker-standalone:
 	@echo "--> Building Docker image"
 	$(DOCKER) build -t celestiaorg/celestia-app -f docker/standalone.Dockerfile .
-.PHONY: build-docker
+.PHONY: build-docker-standalone
 
 ## docker-build: Build the celestia-appd docker image from the current branch. Requires docker.
 docker-build: build-docker-multiplexer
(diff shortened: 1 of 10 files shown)


rach-id:
Seems to be working just fine:
```
$ find . -path -prune -o -name '*.proto' -print0

./proto/celestia/core/v1/proof/proof.proto./proto/celestia/core/v1/da/data_availability_header.proto./proto/celestia/core/v1/tx/tx.proto./proto/celestia/core/v1/gas_estimation/gas_estimator.proto./proto/celestia/mint/v1/query.proto./proto/celestia/mint/v1/genesis.proto./proto/celestia/mint/v1/mint.proto./proto/celestia/minfee/v1/tx.proto./proto/celestia/minfee/v1/query.proto./proto/celestia/minfee/v1/event.proto./proto/celestia/minfee/v1/genesis.proto./proto/celestia/minfee/v1/params.proto./proto/celestia/blob/v1/tx.proto./proto/celestia/blob/v1/query.proto./proto/celestia/blob/v1/event.proto./proto/celestia/blob/v1/genesis.proto./proto/celestia/blob/v1/params.proto./proto/celestia/signal/v1/tx.proto./proto/celestia/signal/v1/upgrade.proto./proto/celestia/signal/v1/query.proto%
```

closing

bouchmann:
Hey @rach-id. You're right it works now, but the syntax is incorrect. 

The command `find . -path -prune` treats `-prune` as the argument to `-path`, so pruning is disabled. It only works because there's no `./cosmos` directory currently.

If `./cosmos` ever gets added, the broken command will include those files while the fixed version correctly excludes them (as was the original intent based on git history).

The fix prevents future issues when cosmos directory exists.

rach-id:
@bouchmann for reference, the PR changes are different to the description

rach-id (approved):

PR #5914, opened 2025-09-30
fix(makefile): correct .PHONY declaration for build-docker-standalone…

animeherox:
Looks like the `.PHONY` declaration for `build-docker-standalone` was mistakenly pointing to `build-docker` (which doesn't exist). Fixed it to match the actual target name.


rootulp (approved):
thank you

PR #5916, opened 2025-10-01
docs: ethereum address instead of public key for identification

mdqst:


rach-id:
The QGB module is deprecated. No need to change anything in it

PR #5917, opened 2025-10-01
docs: fix Makefile portability by using `$(shell pwd)`

0xwitty:
## Overview

replaced `pwd` with `$(shell pwd)` in the Makefile to make it more portable across different shells.
this should prevent issues when running the Makefile in environments where `pwd` behaves differently.

### DIFF
--- a/docs/maintainers/prebuilt-binaries.md
+++ b/docs/maintainers/prebuilt-binaries.md
@@ -27,7 +27,7 @@ Prebuilt binaries are attached to each release via [GoReleaser](https://goreleas
             -e CGO_ENABLED=1 \
             --env-file .release-env \
             -v /var/run/docker.sock:/var/run/docker.sock \
-            -v `pwd`:/go/src/$(PACKAGE_NAME) \
+            -v $(shell pwd):/go/src/$(PACKAGE_NAME) \
             -w /go/src/$(PACKAGE_NAME) \
             ghcr.io/goreleaser/goreleaser-cross:${GOLANG_CROSS_VERSION} \
     -       release --clean


rach-id:
> this should prevent issues when running the Makefile in environments where pwd behaves differently

when should we expect this to behave differently?

0xwitty:
[rach-id](https://github.com/rach-id), ty for the question! The main reason is consistency: in a Makefile, `pwd` is executed by whatever shell `make` happens to use (`/bin/sh`, which can be `bash`, `dash`, etc.). Different shells (and their built-in `pwd`) sometimes return slightly different results — for example, resolving symlinks differently.
Using `$(shell pwd)` makes `make` evaluate it once itself and substitute the result directly, so the path is stable across shells, CI environments, and different platforms.

rach-id:
This will be run in a makefile, so there shouldn't be a difference, no?

rootulp:
> [rach-id](https://github.com/rach-id), ty for the question! The main reason is consistency: in a Makefile, pwd is executed by whatever shell make happens to use (/bin/sh, which can be bash, dash, etc.). Different shells (and their built-in pwd) sometimes return slightly different results — for example, resolving symlinks differently.
Using $(shell pwd) makes make evaluate it once itself and substitute the result directly, so the path is stable across shells, CI environments, and different platforms.

I'm confused, which shell is "shell that `make` happens to use" and how can that differ from `$(shell pwd)`? Did you encounter this as an actual problem on your machine or is this a hypothetical problem.

0xwitty:
[rootulp](https://github.com/rootulp) the real difference is timing: `pwd` runs every time the recipe executes, while `$(shell pwd)` is expanded once when parsing the Makefile. I haven’t hit a concrete issue - it was more of a preventive change to keep the path stable across environments.

rootulp:
Seems like a non-issue for now.

PR #5921, opened 2025-10-01
fix: remove incorrect nolint:unused annotation from defaultBroadcastHandler

lolkdsanyaw:
Function is actually used in mock server setup and parallel tests


rootulp (approved):

rootulp:
IMO not worth backporting but safe to merge.

Issue #5922, opened 2025-10-01
Fill in the signer when using blob share version v1 if nil


evan-forbes:
currently we error if the signer is nil, but the txclient knows the signer and the pfb is not yet signed, therefore it could add it.

this likely means that we want to return the commitment as well

ninabarbakadze:
IIUC do you mean the signer first calls `NewMsgPayForBlobs` it passes signer address but the transaction is not yet signed with this address ([happens on line after this one](https://github.com/celestiaorg/celestia-app/blob/8025e182ffee867c1564c9d98a69accef8ebd0db/pkg/user/signer.go#L139)) then we could just use the passed signer address instead of it failing validation.

ninabarbakadze:
When you populate a blob setting a signer is required https://github.com/celestiaorg/go-square/blob/ce051a4cf82df52b1429a744da4bf3e11ec161ee/share/blob.go#L40

evan-forbes:
hmm maybe we don't need the signer in the blob type? or are we using that anywhere the thinking behind this issue is that we aren't modifying that in the blob which might be problematic

```go
// Blob (stands for binary large object) is a core type that represents data
// to be submitted to the Celestia network alongside an accompanying namespace
// and optional signer (for proving the signer of the blob)
type Blob struct {
	namespace    Namespace
	data         []byte
	shareVersion uint8
	signer       []byte
}
```

https://github.com/celestiaorg/celestia-app/blob/3cc70d97cf4097b3e1412b651469260d46c0f9ac/pkg/user/tx_client.go#L380-L394

if its not then I guess we can close this

ninabarbakadze:
It seems like having blob signer match the signer of the transaction is one of the share version validation rules.

```go
// ValidateBlobShareVersion validates any share version specific rules
func ValidateBlobShareVersion(signer sdk.AccAddress, blobs ...*share.Blob) error {
	for _, blob := range blobs {
		if blob.ShareVersion() == share.ShareVersionOne && !bytes.Equal(blob.Signer(), []byte(signer)) {
			return ErrInvalidBlobSigner.Wrapf("blob signer %X does not match msgPFB signer %X", blob.Signer(), signer)
		}
	}
	return nil
}
```

I don't really understand what exactly we're solving for here. Is it that blob signer might not match tx signer? Is it that blob signer might be nil?

renaynay:
[...]

PR #5925, opened 2025-10-01
fix(x/blob/types): Ensure ValidateBlobShareVersion also checks for nil signer

renaynay:
If ShareVersion == ShareVersionOne, signer field in blob **must** be populated && signer passed in must be non-nil.


x/blob/types/payforblob.go
rootulp, line 221: [uber nit] for clarity, I would've preferred two separate checks with different error messages but definitely not blocking

```go
// ValidateBlobShareVersion validates any share version specific rules
func ValidateBlobShareVersion(signer sdk.AccAddress, blobs ...*share.Blob) error {
	for _, blob := range blobs {
		if blob.ShareVersion() == share.ShareVersionOne {
			if signer == nil {
				return ErrInvalidBlobSigner.Wrapf("MsgPayForBlobs signer is nil")
			}
			if !bytes.Equal(blob.Signer(), signer) {
				return ErrInvalidBlobSigner.Wrapf("blob signer %X does not match MsgPayForBlobs signer %X", blob.Signer(), signer)
			}
		}
	}
	return nil
}
```

rootulp (approved):

evan-forbes:
@ninabarbakadze do you mind fixing the linter and mergin?

also we should mention that this closes #5922

ninabarbakadze:
> @ninabarbakadze do you mind fixing the linter and mergin?
> 
> also we should mention that this closes #5922

It doesn't close that issue but I will merge it.

renaynay:
@evan-forbes this PR doesn't address https://github.com/celestiaorg/celestia-app/issues/5922

I'm also skeptical that we should do https://github.com/celestiaorg/celestia-app/issues/5922

ninabarbakadze:
> @evan-forbes this PR doesn't address #5922
> 
> I'm also skeptical that we should do #5922

We could just do it inside the tx client because we know that during parallel submission the signer address on the blob won't match the worker's address

rootulp:
Is this consensus breaking?

evan-forbes:
> Is this consensus breaking?

yeah was just thinking the same thing. ~~definitely seems consensus breaking~~ isn't called by validate basic

evan-forbes:
> I'm also skeptical that we should do https://github.com/celestiaorg/celestia-app/issues/5922

> We could just do it inside the tx client because we know that during parallel submission the signer address on the blob won't match the worker's address

correct, this is meant to be done inside the tx client as its the only thing that knows which account is submitting the blob. 

Either we do this, or we don't allow share version 1 to be used with queued txs either is acceptable. I doubt anyone intends to use authored blobs that way, however it seems like a theoretically viable use case

PR #5934, opened 2025-10-01
chore: rebase main on v6.x

evan-forbes:
we forgot to backport stuff, but none of it was breaking therefore the thinking here is just to backport everything back on to v6

we don't need to review each loc imo since we just rebased, but we do need to review which commits and ensure there weren't breaking things there we forgot.

then we can make more consistent use of the backport tag


rootulp (dismissed):

ninabarbakadze (dismissed):

evan-forbes:
apologies, I wanted to not squash merge so had to rebase again (and therefore force push) as we merged the v0.39.7 bump of core to v6.x which was included in this PR as well.

evan-forbes:
closing in favor #5972

Issue #5907, opened 2025-09-30
FLUP: txQueue and parallel blob submission flups


evan-forbes:
- one start vs initialized bool https://github.com/celestiaorg/celestia-app/pull/5776#discussion_r2390670258

rach-id:
All the PRs seem to be merged, probably we can close

PR #5935, opened 2025-10-01
refactor: revert changes to SubmitPFB and add new method SubmitPFBInQueue instead

### LINKED ISSUE #5924: Revert SubmitPayForBlob using the queue and create a new method instead
we want to avoid breaking users as much as possible therefore we should use a new method to use the txQueue instead of forcing it w/ SubmitPFB

evan-forbes:
## Overview

closes #5924


evan-forbes:
will follow this up with #5907

pkg/user/tx_client.go
```diff
@@ -337,13 +328,28 @@ func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*share.Blo
...
+func (client *TxClient) QueueBlob(ctx context.Context, resultC chan SubmissionResult, blobs []*share.Blob, opts ...TxOption) {
+	if client.txQueue == nil {
+		resultC <- SubmissionResult{Error: errors.New("tx queue not configured")}
+		return
+	}
+
+	if !client.txQueue.started.Load() {
+		resultC <- SubmissionResult{Error: errors.New("tx queue not started")}
+		return
+	}
```
renaynay, line 343: [non-blocking nit]: can combine these statements with one error -- misconfiguration / unstarted is basically same thing from txclient setup standpoint

pkg/user/tx_client.go
```diff
@@ -306,27 +305,19 @@ func SetupTxClient(
...
-	if !client.txQueue.started.Load() {
-		return nil, errors.New("tx queue not started")
-	}
+	return client.SubmitPayForBlobWithAccount(ctx, client.defaultAccount, blobs, opts...)
+}
 
+// SubmitPayForBlobInQueue submits blobs to the parallel transaction queue and blocks until confirmed.
+// TxOptions may be provided to set the fee and gas limit. This method uses the tx queue infrastructure
+// for parallel submission.
+func (client *TxClient) SubmitPayForBlobInQueue(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error) {
```
renaynay: naming nit


```suggestion
func (client *TxClient) SubmitPayForBlobToQueue(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error) {
```

renaynay (dismissed):

pkg/user/tx_client.go
```diff
@@ -306,27 +305,19 @@ func SetupTxClient(
...
+// This method broadcasts the transaction and waits for confirmation using the default account.
 func (client *TxClient) SubmitPayForBlob(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error) {
-	if client.txQueue == nil {
-		return nil, errors.New("tx queue not configured")
-	}
-
-	if !client.txQueue.started.Load() {
-		return nil, errors.New("tx queue not started")
-	}
+	return client.SubmitPayForBlobWithAccount(ctx, client.defaultAccount, blobs, opts...)
```
ninabarbakadze, line 310: I know we already discussed this but i might have missed something. Why did we decide to keep SubmitPFB as is? Since IIUC `SubmitPayForBlobInQueue` with 1 worker should work the same way.

evan-forbes: > SubmitPayForBlobInQueue with 1 worker should work the same way.

the diff is just that this won't submit more than one blob at a time when called  multiple times as all txs enter the queue. for example if we had 10 transactions  via SubmitPFB, they would be included potentially in a single block. While rare it does occur occasionally. Where as queue guarantees this takes at minimum 10 blocks

pkg/user/tx_client.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5935/commits/3cc70d97cf4097b3e1412b651469260d46c0f9ac

pkg/user/tx_client.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/5935/commits/3cc70d97cf4097b3e1412b651469260d46c0f9ac

PR #5937, opened 2025-10-01
fix(app): remove redundant address conversion in BlockedAddresses()

galaparm:
## Issue
`BlockedAddresses()` was double-converting module addresses. The function iterated over `ModuleAccountAddrs()` results (already string addresses) and incorrectly re-applied `NewModuleAddress().String()` conversion.

## Fix
- Remove redundant `authtypes.NewModuleAddress(acc).String()` call
- Use `acc` directly since it's already a formatted address string

Tested both versions:
- **Before**: Generated malformed addresses like `cosmos1cosmos1govmoduleaddressmoduleaddress`
- **After**: Correct addresses like `cosmos1govmoduleaddress` ✅

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -625,7 +625,7 @@ func (app *App) ModuleAccountAddrs() map[string]bool {
 func (app *App) BlockedAddresses() map[string]bool {
 	modAccAddrs := make(map[string]bool)
 	for acc := range app.ModuleAccountAddrs() {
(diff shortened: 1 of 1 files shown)


ninabarbakadze (approved):

rootulp (changes requested):
I don't believe the PR description.

rootulp:
Closing unless there's strong evidence this is a fix for a real issue.

PR #5938, opened 2025-10-01
fix: correct misleading comments in UpgradeSequence Init and validator selection

GarmashAlex:
The UpgradeSequence comments were inaccurate:
- Init was labeled as a “no-op”, but it actually allocates a funding account used to submit MsgTryUpgrade.
- The code comment claimed a “random validator” is chosen as an “authority”, but the implementation deterministically selects the next validator who hasn’t signalled yet, and does not assign any special authority.


ninabarbakadze (commented):
This is not a very high value improvement therefore i'll be closing this PR. For future contributions please refer to the good first issues section

Issue #5274, opened 2025-07-17
Write a spec for the new tx client logic


ninabarbakadze:
## Description

There's a lot of work being done on tx client to improve the UX of tx submission. That work should be documented in the specs.

PR #5939, opened 2025-10-01
docs: tx client spec

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/5274


specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
+## Transaction Client
+
+## Abstract
+
+The Transaction Client (TxClient) provides a high-level abstraction for constructing, signing, broadcasting, and confirming transactions on Celestia chain. It handles sequence mismatches, mempool evictions, rejections and provides parallel transaction submission via worker accounts. The client is built around two reliable submission strategies: ordered (sequential) and unordered (parallel) processing.
```
evan-forbes: > The client is built around two reliable submission strategies: ordered (sequential) and unordered (parallel) processing.

currently do we support guaranteed ordered at all? or is this a future scenario? 

is there a scenario where even when waiting to confirm a tx we error and skip that tx, submitting the next one?

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Important**: These are the only reliable ways to submit transactions. Other patterns (like multiple transactions from one account without confirmation) will cause sequence mismatches and failures.
```
evan-forbes: > Other patterns (like multiple transactions from one account without confirmation) will cause sequence mismatches and failures.

is this is only a current thing or is this something we're fixing in the future. if the former then it might be worth mentioning "Currently these are the only reliable ways"

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Usage Modes
+
+1. **Ordered Submission:** (Low Throughput, Sequential Processing, Ordering guarantees):
+    - **Single Account**: Uses one account for all transactions
+    - **Sequential Processing**: One transaction per block, confirmed before next submission
```
evan-forbes: currently we one per two blocks so maybe we can just say "Max one tx per block"

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+    - **Ordering Guarantees**: Transactions are processed in exact submission order
```
evan-forbes: I'm not sure we guarantee this currently or if we have discussed that we need this guarantee.

If we need this guarantee then I think we need to disable resigning, and resubmit on a loop until each tx gets confirmed one by one. currently is seems like there could be a failure or a timeout and we'd proceed to the next tx in the queue.

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+    - **Use Case**: Applications requiring strict transaction ordering
```
evan-forbes: also we should mention that users who require that all their blobs be signed by the same account.

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+2. **Unordered Submission:** (High Throughput, Parallel Processing, No Ordering)
```
evan-forbes: nit

Parallel Submission > Processing imo

we might use processing elsewhere but submission still seems more precise

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+- **Signing**: Signs transactions using account private keys
+  - `SignTx(msgs []sdktypes.Msg, opts ...TxOption) (authsigning.Tx, string, uint64, error)`
+- **Encoding/Decoding**: Converts transactions to bytes and vice versa
+  - `EncodeTx(tx sdktypes.Tx) ([]byte, error)`
+  - `DecodeTx(txBytes []byte) (authsigning.Tx, error)`
+- **Account Management**: Tracks account sequences and manages multiple signers
+  - `IncrementSequence(accountName string) error`
+  - `SetSequence(accountName string, seq uint64) error`
+
+**How TxClient Uses Signer**: The TxClient delegates all transaction creation, signing, and encoding operations to the Signer, which handles sequence management, multi-account support, and blob transaction processing.
```
evan-forbes: not blocking

not sure this needs to be included in the spec. the reason being that I'm not sure if api for the signer isn't really important or an implementation detail.

if it does then do we also need to include all details around signing such as sign bytes etc?

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+The worker queue is critical for unordered submission mode, enabling concurrent transaction processing without sequence contention.
+
+**Key Functionality**:
+
+- **Job Queue**: Holds submission jobs (`SubmissionJob`) containing blobs, options, and response channels
+- **Workers**:
+  - Worker 0 = primary account
+  - Workers >0 = additional accounts created and fee-granted by primary
+- **Parallelism**: Allows concurrent submission of multiple transactions without sequence contention
+- **Fee Management**: Non-primary workers use fee grants so all transaction fees are paid by the primary account
```
evan-forbes: not blocking

this feels out of order

why talk about parallel submission before talking about normal submission?

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+**Gas Estimator APIs**:
+
+- `EstimateGasPriceAndUsage(ctx context.Context, msgs []sdktypes.Msg, priority gasestimation.TxPriority, opts ...TxOption) (gasPrice float64, gasUsed uint64, err error)`
+- `EstimateGasPrice(ctx context.Context, priority gasestimation.TxPriority) (float64, error)`
+
+#### Confirmation Loop
+
+Handles transaction resolution and recovery logic using tx data from `TxTracker`.
+
+**Key Functionality**: The confirmation loop polls the chain until transactions are resolved as `Committed`, `Rejected`, `Evicted`, or `Unknown`.
```
evan-forbes: not blocking

it might be better to include a tx submission flow as this off by itself seems weird and difficult to keep track of for implementers

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+- **ExecutionError**: Returned when transaction is included but execution fails
+
+  ```go
+  // ExecutionError occurs when transaction execution fails
+  type ExecutionError struct {
+      TxHash   string // Transaction hash
+      Code     uint32 // Execution result code
+      ErrorLog string // Execution error details
+  }
+  ```
```
evan-forbes: imo we can refactor this to describe what the txclient needs to do when it hits these errors as that seems like critical behavior

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+The submission phase handles transaction creation, job queuing, and worker assignment. The TxClient provides five main submission APIs with different execution paths:
+
+#### SubmitTx
+
+```go
+func (client *TxClient) SubmitTx(ctx context.Context, msgs []sdktypes.Msg, opts ...TxOption) (*TxResponse, error)
+```
+
+- **Path**: Direct processing, no worker queue
+- **Input**: Cosmos SDK messages
```
evan-forbes:
```suggestion
```

we can see the args

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+   - Call gas estimation service to determine gas usage and price
+   - If estimation fails due to a sequence mismatch, update local sequence to expected and retry
+3. **Fee Setting**: Fee is computed as `ceil(minGasPrice * gasLimit)` unless explicitly set by the user
+4. **Broadcast**: Transaction is encoded and submitted via gRPC:
+   - **Single-endpoint mode**: Retries on mismatch until success or failure
+   - **Multi-endpoint mode**: Broadcasts concurrently to multiple endpoints and accepts the first successful response
+5. **Tracker Entry**: On successful broadcast, a record is added to the transaction tracker containing (signer, sequence, txBytes, timestamp) so the transaction can be resubmitted or rolled back if needed
+6. **Sequence Increment**: After a successful broadcast, the local sequence for that account is incremented
+
+⚠️ **Note**: This approach assumes the TxClient is connected to a trusted consensus node in order to maintain replay protection.
```
evan-forbes: not blocking

I find this sentence confusing, since ofc the state machine ensure that txs cannot be replayed, just that the user could get the same blob included multiple times.

if we're really worried about this, then we just limit the number of resigning, and the txclient can detect when it continually increases the sequence for the same blob over and over and over again

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+**Process:**
+
+1. **Build & Sign**: Transaction is built and signed using the Signer
+2. **Gas Estimation**:
+   - If no gas provided, simulate with minimal fee
+   - Call gas estimation service to determine gas usage and price
+   - If estimation fails due to a sequence mismatch, update local sequence to expected and retry
+3. **Fee Setting**: Fee is computed as `ceil(minGasPrice * gasLimit)` unless explicitly set by the user
+4. **Broadcast**: Transaction is encoded and submitted via gRPC:
+   - **Single-endpoint mode**: Retries on mismatch until success or failure
```
evan-forbes: feels weird to include the retry on mismatch logic here as a subtle comment when this requires more discussion imo

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+   - **Multi-endpoint mode**: Broadcasts concurrently to multiple endpoints and accepts the first successful response
```
evan-forbes: why do we discuss resigning in the above but not here

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+## Confirmation
+
+After broadcast, the TxClient continuously polls the chain for transaction status until resolution. This is the tx confrimation life cycle:
+
+- **Pending**: Transaction in the mempool but not yet included -> continue polling
+- **Committed**: Transaction included in a block
+  - If execution succeeded (code = OK): return success
+  - If execution failed: return an execution error and remove from tx tracker
+- **Rejected**:
+  - Transaction explicitly refused by the node during `ReCheck()`(This usually happens in the mempool after a block is committed).
```
evan-forbes:
```suggestion
  - Transaction explicitly refused by the node during `ReCheck()`(This happens in the mempool after a block is committed).
```

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+- Trusted Node: The client depends on a trusted consensus node for account state, sequence numbers, and gas estimation (no proof verification).
+
+- Sequence Consistency: Currently we rely on consensus node to provide us with the correct sequence in case of sequence mismatch errors. If we are connected to a malicious node this could break replay protection.
+
+- Eviction Behavior: Evictions are local to a node. A transaction evicted by one node may still be proposed by another, so users must not re-sign the transaction immediately to avoid double-spending.
+
+- Submission Modes: Only sequential (single account) or parallel (worker accounts with fee grants) are safe; other patterns cause sequence mismatches.
+
+- Resilience: At least one gRPC endpoint must remain live and connected for reliable broadcast and confirmation.
```
evan-forbes: not blocking

I think we can remove this as it feels out of place

my current understanding is that were care that the txclient is the same across the rust and go nodes, so this should be focused entirely on describing the tx flow.

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+The TxClient implementation can be found in the `celestia-app` repo in the `pkg/user` directory:
+
+**Core TxClient code is located in:**
+
+- [`pkg/user/tx_client.go`](https://github.com/celestiaorg/celestia-app/blob/main/pkg/user/tx_client.go) - Transaction submission, broadcasting, and confirmation logic
+- [`pkg/user/parallel_tx_submission.go`](https://github.com/celestiaorg/celestia-app/blob/main/pkg/user/parallel_tx_submission.go) - Parallel worker management and job queuing
+- [`pkg/user/signer.go`](https://github.com/celestiaorg/celestia-app/blob/main/pkg/user/signer.go) - Transaction signing and account management
+- [`pkg/user/account.go`](https://github.com/celestiaorg/celestia-app/blob/main/pkg/user/account.go) - Account state management
+- [`pkg/user/tx_options.go`](https://github.com/celestiaorg/celestia-app/blob/main/pkg/user/tx_options.go) - Transaction configuration options
```
evan-forbes: not blocking

imo we can delete

evan-forbes (commented):
good start

generally I think we can narrow the focus of this document to strictly focus on the transaction submission flow and order the entire document around that. 

Assuming that an implementer has something that can create and send a valid PFBs via grpc, then they should be able to recreate a working version of the txclient in a different language. I'm not sure that an implementer would be able to gleam enough details here do that.

When do we resubmit? when to we resign? how should the local sequence state get updated? What do we do upon each failure? when to we start the txstatus query routine?

personal opinion, but currently it feels like the document has sections for things in a kind of arbitrary way.

specs/src/tx_client.md
ninabarbakadze: The way I see this is that guaranteed ordering depends on how you consume these functions which ties into your second question too. 

> is there a scenario where even when waiting to confirm a tx we error and skip that tx, submitting the next one?

The users could broadcast and not wait for confirm at all and keep broadcasting. This logic is split up so it's possible to use tx client in many different ways. We just need to define well how it's meant to be used _currently_.

specs/src/tx_client.md
ninabarbakadze: Yes, I will mention more explicitly that this is current limitation. Everything that I have added here is my current understanding of how this work right now and tried to not speculate too much about what we will enable in the future.

specs/src/tx_client.md
ninabarbakadze: I think my intention here was to say that from a single signer there can't be more than two txs in a block.

specs/src/tx_client.md
ninabarbakadze: In sequential mode ordering is guaranteed because only one transaction is ever submitted at a time. client increments the local sequence after broadcast but it does not proceed with the next submission until the previous transaction has been confirmed. This is what ensures strict ordering (tx1 → confirm → tx2 → confirm → tx3 …). Sequence mismatches and re-signing simply correct local state; they don’t break ordering since no other transactions are in flight.

ninabarbakadze: at least this is my understanding. could be wrong.

specs/src/tx_client.md
ninabarbakadze: I decided to include this because users have complained in the past about being able to set their own sequence and demonstrating that there's a signer API for this was useful for them. 

Also, @renaynay wanted this included. Happy to omit if it's not useful.

specs/src/tx_client.md
ninabarbakadze: I did not sort it in any specific order. Will re-sort.

specs/src/tx_client.md
ninabarbakadze: I think I agree. I will remove this section entirely.

specs/src/tx_client.md
ninabarbakadze: I'll remove all inputs since I'm referencing implementation.

specs/src/tx_client.md
ninabarbakadze: Number of resigning should not be a problem if you are connected to an honest node. I think even if we limit number of times we resign a malicious node can still get the same blobs included twice. 

I can rephrase or omit this.

specs/src/tx_client.md
ninabarbakadze: I thought we wanted to write tx client spec in general for us + the users which is why I tried to follow the template and try to describe it more holistically. If we wanted just a submission spec I can remove most of these things or create a separate spec for it.

specs/src/tx_client.md
ninabarbakadze: `  - **Single-endpoint mode**: Retries on mismatch until success or failure`

Sequence mismatch would have made it explicit. Will update. 

edit: I just saw the diff more explicitly and you're right. Will emphasize.

ninabarbakadze:
> good start, some blocking comments, but most are not blocking
> 
> generally I think we can narrow the focus of this document to strictly focus on the transaction submission flow and order the entire document around that.
> 
> Assuming that an implementer has something that can create and send a valid PFBs via grpc, then they should be able to recreate a working version of the txclient in a different language. I'm not sure that an implementer would be able to gleam enough details here do that.
> 
> When do we resubmit? when to we resign? how should the local sequence state get updated? What do we do upon each failure? when to we start the txstatus query routine?
> 
> personal opinion, but currently it feels like the document has sections for things in a kind of arbitrary way.

Sure, we can narrow it down. 

I think Confirm section answers a few of your questions. 

> When do we resubmit?

 Confirmation
- **Evicted**: Transaction dropped from mempool (e.g. low fees).
  - The client resubmits using txBytes stored in the tracker
  - If resubmission fails again (e.g. mismatch), an eviction timeout window (1 minute) begins before reporting failure

> When do we resign? How should the local sequence state get updated?

These are the three places where we currently resign/update local state

  1.  If estimation fails due to a sequence mismatch, update local sequence to expected and retry
  2. . **Broadcast**: Transaction is encoded and submitted via gRPC:
   - **Single-endpoint mode**: Retries on mismatch until success or failure  -- **this can definitely be more explicit**
  3. 6. **Sequence Increment**: After a successful broadcast, the local sequence for that account is incremented

> When do we start the txstatus query routine?

 Confirmation

After broadcast, the TxClient continuously polls the chain for transaction status until resolution. This is the tx confrimation life cycle:


> What do we do upon each failure?

I can definitely expand on this much more. I see how I only scratched the surface with error types.

specs/src/tx_client.md
renaynay: signer api is useful to document - it exposes internals of account(s) used by `TxClient`

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
+## Transaction Client
+
+## Abstract
+
+The Transaction Client (TxClient) provides a high-level abstraction for constructing, signing, broadcasting, and confirming transactions on Celestia chain. It handles sequence mismatches, mempool evictions, rejections and provides parallel transaction submission via worker accounts. The client is built around two reliable submission strategies: ordered (sequential) and unordered (parallel) processing.
+
+**Important**: These are the only reliable ways to submit transactions. Other patterns (like multiple transactions from one account without confirmation) will cause sequence mismatches and failures.
+
+### Usage Modes
```
renaynay: are we not going to mention here that we also provide API for submitting directly to mempool (unqueued)? I would consider that the 3rd (less desirable) usage mode prone to squence contention

specs/src/tx_client.md
renaynay: Could also mention that the workers get unlimited feegrant from master account.

renaynay: and that these accounts are reused between runtimes (if parallel tx submission is enabled)

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+- **Parallelism**: Allows concurrent submission of multiple transactions without sequence contention
+- **Fee Management**: Non-primary workers use fee grants so all transaction fees are paid by the primary account
+
+### Gas Estimator
+
+The gas estimation service is an external dependency that provides accurate gas and fee calculations for transaction submission.
+
+**Key Functionality**:
+
+TxClient calls the gas estimation service during broadcast to ensure transactions have appropriate gas limits and fees, handling sequence mismatches automatically through re-signing with corrected sequences.
```
renaynay, line 15: mention that gas estimation is only necessary when user does not set gas limit or gasprice in options -- meaning it is the default, but shouldn't occur if tx opts already specify gasprice/gas limit

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+- **Process**: PayForBlob creation -> Broadcast -> Confirmation
+- **Usage**: Only for sequential submissions using the default account
+
+#### SubmitPayForBlobToQueue
+
+```go
+func (client *TxClient) SubmitPayForBlobToQueue(ctx context.Context, blobs []*share.Blob, opts ...TxOption) (*TxResponse, error)
+```
+
+- **Path**: Uses worker queue for parallel processing
```
renaynay: or single worker if sequental ordering (1 worker) is specified

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+Broadcasting steps are universal across SDK transactions and blob transactions.
+
+**Process:**
+
+1. **Build & Sign**: Transaction is built and signed using the Signer
+2. **Gas Estimation**:
+   - If no gas provided, simulate with minimal fee
+   - Call gas estimation service to determine gas usage and price
+   - If estimation fails due to a sequence mismatch, update local sequence to expected and retry
```
renaynay: There's still some debate around whether we should actually be doing this (updating local sequence with what consensus node hands back). I'm not sure if we should specify this behaviour rn 

cc @cmwaters

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+- **Committed**: Transaction included in a block
+  - If execution succeeded (code = OK): return success
+  - If execution failed: return an execution error and remove from tx tracker
+- **Rejected**:
+  - Transaction explicitly refused by the node during `ReCheck()`(This happens in the mempool after a block is committed).
+  - The client rolls back the sequence to the rejected transaction's sequence
+  - Removes the `TxTracker` entry
+  - Returns error
+- **Evicted**: Transaction dropped from mempool (e.g. low fees).
+  - The client resubmits using txBytes stored in the tracker
```
renaynay: do we talk about concept of TTL? We should probably specify that as it's the desired behaviour here as well (replay protection) ?

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+Request: EstimateGasPriceAndUsageRequest { TxBytes, TxPriority }
+
+Response: EstimatedGasUsed, EstimatedGasPrice
+```
+
+## Assumptions and Considerations
+
+- Trusted Node: The client depends on a trusted consensus node for account state, sequence numbers, and gas estimation (no proof verification).
+
+- Sequence Consistency: Currently we rely on consensus node to provide us with the correct sequence in case of sequence mismatch errors. If we are connected to a malicious node this could break replay protection.
```
renaynay: ah ok i see you mention it here, that's good

specs/src/tx_client.md
```diff
@@ -0,0 +1,273 @@
...
+
+- Eviction Behavior: Evictions are local to a node. A transaction evicted by one node may still be proposed by another, so users must not re-sign the transaction immediately to avoid double-spending.
+
+- Submission Modes: Only sequential (single account) or parallel (worker accounts with fee grants) are safe; other patterns cause sequence mismatches.
```
renaynay:
```suggestion
- Submission Modes: Only sequential (single account) or parallel (worker accounts with fee grants) are safe; other patterns may lead to sequence mismatches.
```

renaynay (commented):
pretty good and comprehensive overview

specs/src/tx_client.md
ninabarbakadze: We have not yet settled on what we want TTLs to look like. I'd only add them to the spec once we actually decide.

specs/src/tx_client.md
ninabarbakadze: I think it's very confusing when you get sequence mismatch during simulation so until this behaviour is resolved i think we should mention it as this is the only way to bypass sequence mismatch during estimation

specs/src/tx_client.md
evan-forbes: makes sense, do you think its worth changing?

specs/src/tx_client.md
evan-forbes: > because only one transaction is ever submitted at a time.

sequential meaning queued? as in we only submit a single tx at a time?

what if the first tx fails, then we immediately submit the second in the queue and it passes. if that can occur, then to me that counts as ordering is not guaranteed

specs/src/tx_client.md
evan-forbes: > I decided to include this because users have complained in the past about being able to set their own sequence and demonstrating that there's a signer API for this was useful for them.

_currently_ users do not have the ability in txclient to specify a sequence directly. if we want that in the future then I'm fine with adding it here and adding a flup to add it.

maybe we can mention that is not required for the txclient that the signer have these methods?

for example, the rust txclient does not have these exact methods and that seems fine. my understanding of the scope of this document a spec for the txclient

specs/src/tx_client.md
ninabarbakadze: > sequential meaning queued? as in we only submit a single tx at a time?

yes

> what if the first tx fails, then we immediately submit the second in the queue and it passes. if that can occur, then to me that counts as ordering is not guaranteed

You submit tx1: create a fee grant from A to B, and then tx2: B uses the fee grant from A to pay for a transaction. If tx1 fails (meaning the fee grant was never created), then tx2 should also fail since it cannot cover fees without the grant.

If the user wants ordering but a transaction fails due to a state machine error then we can’t guarantee ordering in that case. Ordering guarantees just mean: if transactions follow the state machine rules and the user has sufficient funds for fees, then those transactions should be included in the chain in the order they were submitted.

That’s my current understanding, but I could be wrong so it would be good to clarify what we actually mean by ordering.

specs/src/tx_client.md
evan-forbes: assumptions and usage docs should go in a readme or something similar imo

spec is for implementooooors

specs/src/tx_client.md
ninabarbakadze: > currently users do not have the ability in txclient to specify a sequence directly. if we want that in the future then I'm fine with adding it here and adding a flup to add it.

They do actually

```go
txClient.Signer().SetSequence()
```

> maybe we can mention that is not required for the txclient that the signer have these methods?

Sure! I'm indifferent about signer interface I included it because our tx client does depend on the signer for account management, signing etc but of course this can look different in a different tx client implementation and does not need to necessarily follow the same interface.

I think we should also have a spec for our own tx client implementation and something super generalised like just the tx flow because many of our users and engineers feel ambiguity around it.

specs/src/tx_client.md
ninabarbakadze: The implementooors should probably be aware of these assumptions as I don't feel like they're very intuitive/obvious to people who do not know internals of the SDK/Tendermint flow.

I was just following our spec template :D https://github.com/celestiaorg/celestia-app/blob/main/specs/template.md

I'm gonna apply the comments and remove some stuff then we can see if we want to remove more.

specs/src/tx_client.md
ninabarbakadze: I feel like if you're submitting directly to RPC then you don't need tx client. Happy to add if you think it's useful.

specs/src/tx_client.md
evan-forbes: > If the user wants ordering but a transaction fails due to a state machine error then we can’t guarantee ordering in that case

>  then those transactions should be included

I'm talking about blobs here. you're definitely correct that the failures are less likely due to the state machine playing less of a roll, but afaiu there is no mechanism to always loop until a tx is confirmed before moving on to the next one. If that's true then it seems like there is always a chance for failure, and thus ordering is not guaranteed. We should not say that txs will be submitted in order as that is not something users can build on. from my perspective, there are no sequential ordering guarantees in any scenario using TxClient.

specs/src/tx_client.md
evan-forbes: > They do actually

why have users been asking for this property, is this something that node needs to expose and we just don't expose it there?

evan-forbes: I just want to make it clear that this interface is currently not a requirement to implement the tx client.

if we want to make it a requirement, then that's a different story but I'm not aware of that decision if it was made

specs/src/tx_client.md
```diff
@@ -0,0 +1,254 @@
...
+```go
+type txInfo struct {
+    sequence  uint64    // Account sequence at submission time
+    signer    string    // Account name that signed the transaction
+    timestamp time.Time // Submission timestamp
+    txBytes   []byte    // Raw transaction bytes for resubmission
+}
+```
+
+The transaction tracker is a critical component that enables the TxClient to handle network failures and mempool evictions. After successfully broadcasting a transaction, the client stores essential metadata locally to enable different recovery mechanisms.
```
evan-forbes: not blocking

potential simplification

```suggestion
The transaction tracker stores metadata locally to enable recovery mechanisms for evictions and rejection.
```

separate from the above

I see that we use the txtracker later in the spec, but if we wanted to minimize the spec here we could just say there is a local copy of the sequence per account. or maybe just hand wave all the user state in the "signer" or smt. Then below we could say when that sequence gets updated with "... local sequence MUST be incremented." etc

its definitely not the end of the world to include this, but I'm under the impression that this is a spec and not an ADR, so this purposefully leaves out as many of the implementation details or docs as possible.

specs/src/tx_client.md
```diff
@@ -0,0 +1,254 @@
...
+    ```
+
+- **QueueBlob**: Direct job submission to worker queue.
+  - **Process**: Job queued directly → Worker assignment → PayForBlob creation → Broadcast → Confirmation
+
+    ```go
+    func (client *TxClient) QueueBlob(ctx context.Context, resultC chan SubmissionResult, blobs []*share.Blob, opts ...TxOption)
+    ```
+
+### Submission API Summary
```
evan-forbes:
```suggestion
```

specs/src/tx_client.md
```diff
@@ -0,0 +1,254 @@
...
+Request: TxStatusRequest { TxId }
+Response: TxStatusResponse { Status, Height, ExecutionCode, Error }
+```
+
+### Gas Estimation
+
+```go
+Request: EstimateGasPriceAndUsageRequest { TxBytes, TxPriority }
+Response: EstimatedGasUsed, EstimatedGasPrice
+```
```
evan-forbes, line 200: not blocking

don't we cover this on lines 43

specs/src/tx_client.md
evan-forbes: I just don't like "breaks replay protection"

this can be misinterpreted imo. ofc, there's a **huge** difference mechanically between signing the same blob repeatedly and being able to repeatedly execute the exact same transaction. 

also, we still haven't discussed this as a group if this is something that is worth being concerned about. **there's no such thing as "untrusted"**, or avoiding this beyond running one's own full node, properly supporting calling TxStatus from multiple nodes, or limiting the number of sequential resigns.

I'd be more okay with something along the lines of "currently there is no requirement in the spec for limiting the number of resigning".

specs/src/tx_client.md
evan-forbes: touche

evan-forbes (approved):
if we added a diagram, that could be useful as well. I asked claude to generate options, and while it certainly could be improved on maybe cool to add in the future

```mermaid
sequenceDiagram
    participant Client as TxClient
    participant Signer as Signer
    participant Node as Consensus Node

    Note over Client,Node: Broadcast Phase
    Client->>Client: Create MsgPayForBlobs
    Client->>Node: Estimate gas
    Node-->>Client: Gas estimate
    Client->>Signer: Sign transaction
    Signer-->>Client: Signed txBytes

    Client->>Node: BroadcastTx(txBytes)
    alt Success
        Node-->>Client: TxHash
        Client->>Signer: Increment sequence
    else Sequence Mismatch
        Node-->>Client: Expected sequence error
        Client->>Signer: Update sequence
        Client->>Signer: Resign transaction
        Signer-->>Client: New txBytes
        Client->>Node: BroadcastTx(newTxBytes) [RETRY]
        Node-->>Client: TxHash
        Client->>Signer: Increment sequence
    else Other Error
        Node-->>Client: Error
        Client-->>Client: EXIT with error
    end

    Note over Client,Node: Confirmation Phase
    loop Poll every 3s
        Client->>Node: TxStatus(txHash)
        alt Pending
            Node-->>Client: Status=Pending
        else Committed OK
            Node-->>Client: Status=Committed, Code=0
            Client-->>Client: EXIT success
        else Committed Error
            Node-->>Client: Status=Committed, Code!=0
            Client-->>Client: EXIT with execution error
        else Evicted
            Node-->>Client: Status=Evicted
            Client->>Node: BroadcastTx(txBytes) [RETRY]
            alt Retry Success
                Node-->>Client: Success
            else Retry Failed
                Node-->>Client: Error
                Note over Client: Start 1min timeout timer
            end
        else Rejected
            Node-->>Client: Status=Rejected
            Client->>Signer: Reset sequence to rejected tx
            Client-->>Client: EXIT with rejection error
        end
    end
```

I think there is enough info to implement implementers can ask questions let's either merge or ask eiger to review here imo

evan-forbes:
cc @zvolin @oblique for review

specs/src/tx_client.md
ninabarbakadze: This is trying to break down communication format. I'm happy to remove though.

specs/src/tx_client.md
ninabarbakadze: > afaiu there is no mechanism to always loop until a tx is confirmed before moving on to the next one

This needs to be done on the user side. Do not submit subsequent until you get confirmation. 

I removed the ordering guarantees part.

specs/src/tx_client.md
ninabarbakadze: I removed it

specs/src/tx_client.md
ninabarbakadze: Sometimes claude in cursor decides to cook in places where I miss :D my bad for not removing overly verbose garbage.

zvolin: I agree, `TxTracker` is an implementation detail, the spec should just say that the `sequence` `hash` and `bytes` of a transaction need to be kept around for different fallback scenarios

specs/src/tx_client.md
```diff
@@ -0,0 +1,253 @@
...
+
+                 ```go
+                  type BroadcastTxError struct {
+                  TxHash   string // Transaction hash
+                  Code     uint32 // Error code from node
+                  ErrorLog string // Detailed error message
+                }
+                ```
+
+6. **Tracker Entry**: On success, record (`signer`, `sequence`, `txBytes`, `timestamp`) in TxTracker. Entries older than 10 minutes are pruned automatically.
```
zvolin: another tx tracker reference

specs/src/tx_client.md
```diff
@@ -0,0 +1,253 @@
...
+    type ExecutionError struct {
+        TxHash   string // Transaction hash
+        Code     uint32 // Execution result code
+        ErrorLog string // Execution error details
+    }
+    ```
+
+- **Rejected**: Transaction was explicitly refused by the node during `ReCheck()` (after each block commit).
+  - Retrieve tx sequence from `TxTracker`
+  - Roll back the sequence to the retrieved tx sequence
```
zvolin: should this also happen if we get error related to sequence? I saw `TxRejected` errors with `WrongSequence`, does it still make sense to revert the local sequence in such case?

specs/src/tx_client.md
```diff
@@ -0,0 +1,253 @@
...
+- **Evicted**: Transaction dropped from mempool (low fees, mempool full).
+  - The client resubmits using txBytes stored in the tracker.
+  - If resubmission fails (e.g. sequence mismatch):
+    - **Do not resign with a new sequence**. Eviction isn't final, another node might still propose this tx and resigning risks duplicate inclusion.
+    - Start a 1-minute eviction timeout window. If the tx is not confirmed within timeout period, construct and return an eviction error. (no specific error type)
+  - If resubmission is successful:
+    - The same tx is re-added to the node's mempool
+    - Continue polling until the updated status is resolved
+
+- **Unknown/Not Found**: Tx is neither evicted nor rejected or confirmed. Return error to the user that tx is unknown.
```
zvolin, line 173: when does it happen? I don't think it is possible outcome if the node that is queried for tx status is the same node that received the broadcast. But if that appears to be the case, shouldn't transaction be treated as if it never happened and thus have sequence rollback too?

specs/src/tx_client.md
zvolin: maybe it should be re-signed and re-submitted on sequence errors?

specs/src/tx_client.md
```diff
@@ -0,0 +1,253 @@
...
+  - Workers >0 = additional fee-granted accounts
+- **High Throughput**: Multiple transactions can be submitted and processed per block concurrently, without sequence contention.
+- **No Ordering Guarantees**: Execution order is nondeterministic; workers process transactions as they become available.
+- **Fee Grants**: All fees are covered by the primary account, so workers operate without individual balance management.
+- **Account Reuse**: Worker accounts are persisted across runtimes when parallel submission is enabled.
+- **Use Case**: Best suited for applications that prioritize throughput over ordering.
+
+**note:** If initialized with only one worker, this mode behaves identically to Sequential Submission (one tx per block).
+
+**APIs**:
```
zvolin, line 83: is the `APIs` part of the spec bonding? If so, then I'd say it should be less specific, on rust side having things like `resultC` would feel weird

specs/src/tx_client.md
zvolin: also, timestamp sounds like an implementation detail of tx tracker too

zvolin:
Also cc @fl0rek

specs/src/tx_client.md
zvolin: maybe the api's section could say something like "Go APIs reference:"

specs/src/tx_client.md
ninabarbakadze: This happens if we're confirming a hash that does not exist in any of the other caches in the mempool which currently are rejected, evicted, committed, mempool(pending).

> I don't think it is possible outcome if the node that is queried for tx status is the same node that received the broadcast.

Why not? You could pass any `hash` into Confirm(). What do you do with a hash that node is not aware of? What other response would you have in place for such scenarios.

> shouldn't transaction be treated as if it never happened and thus have sequence rollback too?

If it's unknown how can you be sure that it was even broadcasted? I suppose you can check if the hash is tracked by the client.

ninabarbakadze: This is only because Confirm can be called independently.

specs/src/tx_client.md
ninabarbakadze: > I saw TxRejected errors with WrongSequence, does it still make sense to revert the local sequence in such case?

This should not really happen now that we have sequential submissions where you basically wait for tx confirmation before you submit another one with the same address. 

Sequence error rejections happen because you submitted a bunch of txs from the same address in parallel and one of them got rejected for a different reason which got all the subsequent ones to have sequence mismatch.

specs/src/tx_client.md
zvolin: let's narrow this conversation to the hash of a transaction that was successfully broadcasted, because what you say above is also an implementation detail of Go client to me. Is the unknown status possible to be returned in such case? And if it is, shouldn't the sequence be reverted?

ninabarbakadze: > transaction that was successfully broadcasted

If it was successfully broadcasted and not confirmed during the window of that hash being available in those caches then it will be unknown because it got pruned in the meantime. I can't tell you exactly what the pruning interval is there. It should not really happen especially if you call Confirm after broadcasting.

> shouldn't the sequence be reverted?

I think if you can reliably verify that it was broadcasted by the tx client then sure.

When it comes to sequence management locally that can be improved a lot but transactions currently get resigned with correct sequence if broadcasting fails with sequence mismatch(local sequence fell behind, etc).

> implementation detail of Go client to me

we're just parsing a response from the core RPC endpoint. Isn't it similar in rust?

specs/src/tx_client.md
ninabarbakadze: I think I misinterpreted this comment. Do you mean just submitting transactions without waiting?

ninabarbakadze:
> Also cc @fl0rek

@zvolin @evan-forbes I tried to remove more implementation details. I'm loosely mentioning tracker because it's the easiest way for me to explain that transaction metadata and info gets written somewhere. What kind of naming and structure you choose in rust client is obviously down to your own design preference.

specs/src/tx_client.md
zvolin: > > implementation detail of Go client to me
>
> we're just parsing a response from the core RPC endpoint. Isn't it similar in rust?

We call cofirm tx internally, so user doesn't call it with wrong hashes. We also revert the sequence on unknown, was just wondering if it should make it to the spec there

specs/src/tx_client.md
zvolin: wondering if it should still be outlined there, it can still happen if someone shares the same private key between `TxClient` and wallet or so

specs/src/tx_client.md
ninabarbakadze: So ConfirmTx is not part of the API itself like it's a private function?

ninabarbakadze: > We also revert the sequence on unknown, was just wondering if it should make it to the spec there

I am okay with this. As I said it'd require checking if that tx was actually broadcasted by the client. If the team agrees with it I'm happy to add @evan-forbes

specs/src/tx_client.md
ninabarbakadze: There are two kinds of sequence mismatches and it seems like we're blurring the lines here a little bit.

> wondering if it should still be outlined there, it can still happen if someone shares the same private key between TxClient and wallet or so

In this scenario we'd get sequence mismatch in `CheckTx` which happens right when you broadcast a transaction and we have sequence error handling there.

The sequence error we were speaking about here is when existing txs in the mempool become invalid after `ReCheck` because we have multiple pending txs from the same signer. This was a design flaw on the client UX side because a single signer should only have one transaction pending at a time because of the sequential nature of nonces.

zvolin: okie, I'm fine with resolving this

zvolin: I'm wondering, as I'm adding resigning and retransmission on rust side.

> Sequence error rejections happen because you submitted a bunch of txs from the same address in parallel and one of them got rejected for a different reason

I have a test where I submit 100 blobs of different sizes in parallel from a single account. I cannot see any other errors but rejected transactions with the wrong sequence

zvolin: the test for reference
https://github.com/zvolin/lumina/blob/feat/grpc-parallel-submit/grpc/src/client.rs#L1022

specs/src/tx_client.md
zvolin: just to note, on rust side we went with re-querying the account instead of matching a regex. Probably less responsive / slower approach, but felt more correct

zvolin:
I'm good starting there. Just curious in some threads trying to see whole picture. No blocks from me there

ninabarbakadze:
> I'm good starting there. Just curious in some threads trying to see whole picture. No blocks from me there

If I missed anything or something is not clear please tag me anywhere and I'll be happy to answer.

specs/src/tx_client.md
ninabarbakadze: I don't think it's bonding. I'll defer to @evan-forbes for final say though.

PR #5940, opened 2025-10-01
fix(tool): check earliest height when checking random blocks 

mcrakhman:
## Overview

This PR fixes the problem when the `check-eds-construction` tool didn't check the earliest height and could request the blocks before such height.


rach-id (commented):
This seems to still be checking random blocks, does it make sense to have a sequential version that goes over all heights? so that we run in a server and be 100% sure all is good.

evan-forbes (approved):

PR #5974, opened 2025-10-06
fix: override v1 blob signer with tx signer

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/5922

### DIFF
--- a/app/process_proposal.go
+++ b/app/process_proposal.go
@@ -121,10 +121,9 @@ func (app *App) ProcessProposalHandler(ctx sdk.Context, req *abci.RequestProcess
 			return reject(), nil
 		}
 
-		// validated the PFB signature
 		ctx, err = handler(ctx, sdkTx, false)
 		if err != nil {
-			logInvalidPropBlockError(app.Logger(), blockHeader, "invalid PFB signature", err)
+			logInvalidPropBlockError(app.Logger(), blockHeader, "ante handler validation failed", err)
 			return reject(), nil
 		}
 
--- a/docs/maintainers/modify-param.md
+++ b/docs/maintainers/modify-param.md
@@ -37,45 +37,41 @@ params:
 
 1. Create a proposal.json file with contents:
 
-    ```json
+```json
+{
+  "messages": [
     {
-    "messages": [
-        {
-        "@type": "/celestiaorg.celestia_app.blob.MsgUpdateParams",
-        "authority": "celestia10d07y265gmmuvt4z0w9aw880jnsr700jtgz4v7",
-        "params": {
-            "gas_per_blob_byte": 8,
-            "gov_max_square_size": 256
-        }
-        },
-        {
-        "@type": "/cosmos.consensus.v1.MsgUpdateParams",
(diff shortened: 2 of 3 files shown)


pkg/user/tx_client.go
```diff
@@ -395,6 +395,18 @@ func (client *TxClient) BroadcastPayForBlobWithAccount(ctx context.Context, acco
...
+	// Fill in the signer for v1 blobs to match the transaction signer
+	for i, blob := range blobs {
+		if blob.ShareVersion() == share.ShareVersionOne && !bytes.Equal(blob.Signer(), acc.Address()) {
+			newBlob, err := share.NewV1Blob(blob.Namespace(), blob.Data(), acc.Address())
+			if err != nil {
+				return nil, fmt.Errorf("creating v1 blob with filled signer: %w", err)
+			}
+			blobs[i] = newBlob
+		}
+	}
```
evan-forbes: wdyt about only doing this with parallel submission? 

this definitely works as well, but might be less intuitive to the user. if they mess up, then they get an error and can figure out how to fix. With this if they mess up, it gets fixed quietly

pkg/user/tx_client_test.go
```diff
@@ -847,3 +871,29 @@ func (suite *TxClientTestSuite) TestSequenceIncrementOnlyOnceInMultiConnBroadcas
...
 	require.Equal(t, multiConnClient.DefaultAccountName(), trackedSigner, "Tracked signer should match")
 }
+
+func (suite *TxClientTestSuite) TestBlobSigner() {
+	t := suite.T()
+
+	blob, err := share.NewV1Blob(share.RandomBlobNamespace(), random.Bytes(100), testnode.RandomAddress().(sdk.AccAddress))
+	require.NoError(t, err)
+
+	// not let's submit this pfb with default account
```
rootulp:
```suggestion
	// now let's submit this pfb with default account
```

pkg/user/tx_client.go
```diff
@@ -395,6 +395,18 @@ func (client *TxClient) BroadcastPayForBlobWithAccount(ctx context.Context, acco
 		return nil, fmt.Errorf("account %s not found", accountName)
 	}
 	signer := acc.Address().String()
+
+	// Fill in the signer for v1 blobs to match the transaction signer
```
rootulp: [question][no change needed] given the caller of this function knows about the `blobs` and the `accountName`, I'm wondering why it isn't correctly creating `blobs` with the signer for the `accountName`?

In other words, is it possible to remove these lines which seem more like a fix _after the fact_ and instead fix the signer wherever `blobs` get created?

pkg/user/tx_client.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/5974/commits/13232c44696727f936665f475dc37ddf21530145

pkg/user/tx_client.go
ninabarbakadze: This became a problem because the blob signer no longer matches the transaction signer. The blob is already passed into the transaction client with a populated signer field which is normal behavior. However, with parallel submission we need to ensure that this signer is overridden with the worker’s address.

PR #5975, opened 2025-10-06
Remove misleading startHeight logs and clarify CheckLiveness range

phrwlk:
This change removes unused and misleading startHeight variables and logs in e2e_simple_test.go and e2e_full_stack_pfb_test.go, updates the liveness log in e2e_upgrade_test.go to avoid implying a start-height-based check, and clarifies in e2e_test.go that CheckLiveness validates from height 1 to the latest height.

Why this is needed:
- CheckLiveness internally uses startHeight=1 and fetches proposers from [1, end], so logging “from height %d” with a locally computed startHeight is inaccurate.
- The startHeight variables in the simple and full-stack tests were effectively dead code, used only for misleading logs.
- Correcting the logs and documenting the actual range prevents confusion during test runs and improves maintainability and diagnostic clarity.

### DIFF
--- a/test/docker-e2e/e2e_simple_test.go
+++ b/test/docker-e2e/e2e_simple_test.go
@@ -51,14 +51,12 @@ func (s *CelestiaTestSuite) TestE2ESimple() {
 	s.CreateTxSim(ctx, celestia)
 
 	// Record start height for liveness check
-	startHeight, err := celestia.Height(ctx)
-	s.Require().NoError(err, "failed to get start height")
 
 	assertTransactionsIncluded(ctx, t, celestia)
 
 	testBankSend(t, celestia, cfg)
 
-	s.T().Logf("Checking validator liveness from height %d", startHeight)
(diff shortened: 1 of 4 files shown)


test/docker-e2e/e2e_simple_test.go
```diff
@@ -51,14 +51,12 @@ func (s *CelestiaTestSuite) TestE2ESimple() {
 	s.CreateTxSim(ctx, celestia)
 
 	// Record start height for liveness check
```
rootulp: this line should also be removed

```suggestion
```

rootulp (changes requested):
LGTM after one more comment is removed.

@mojtaba-esk can you please review this PR too? I think these dead logs were missed while addressing PR feedback.

rootulp (approved):

evan-forbes (commented):
changes like this are not worth reviewing and tbh I think we should auto close these

check the contribution guidelines @phrwlk

PR #5981, opened 2025-10-07
feat: add tool to query params for a given height

### LINKED ISSUE #5977: Write a go script that scrapes the parameters
users need to find parameters for historical and future states. We should be able do this by writing a script that points to a state archive node and queries for all state related to the parameters for that app version.

to close this issue we can just post the script here or merge it to the scripts directory

evan-forbes:
## Overview

not sure if we need to merge this, but it per user req we should add a tool that makes it easier to query parameters at a given height. This does that, at least for cosmos-sdk v0.50 based app versions.

closes #5977

### DIFF
--- a/tools/query-params/main.go
+++ b/tools/query-params/main.go
@@ -5,11 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


evan-forbes:
going to close but if someone is curious they should be able to use this as a reference / point some llm at

PR #5986, opened 2025-10-07
fix: move entrypoint scripts from /opt to home directory

### LINKED ISSUE #4670: move the local_devnet celestia-app files from /opt to home
It's unclear to me why `CELESTIA_HOME` was `/opt` in these files. That sounds like the install directory, not the directory to place data.

_Originally posted by @adlerjohn in https://github.com/celestiaorg/celestia-app/pull/4666#discussion_r2061589832_

DeVikingMark:
Fixes #4670

Moves entrypoint scripts from `/opt` to user home directory in Docker containers.

**Changes:**
- Updated `docker/standalone.Dockerfile`
- Updated `docker/multiplexer.Dockerfile` 
- Updated `docker/txsim/Dockerfile`

**Before:** Scripts copied to `/opt/entrypoint.sh`
**After:** Scripts copied to `${CELESTIA_APP_HOME}/entrypoint.sh`


rootulp (commented):
How was this PR tested? Can you please try building Docker images with this change, running them, and verifying the entrypoint.sh script is still invoked correctly? Perhaps also manually inspecting the produced Docker image to ensure the entrypoint.sh script is moved to the correct place.

DeVikingMark:
Hi @rootulp! Thanks for the comment. I've tested the changes, and it seems that everything works correctly.
## docker build -t celestia-node-test .
image

image

DeVikingMark:
Hi @rootulp! Thanks for the comment. I've tested the changes, and HERE'S what I got.
# docker build -t celestia-node-test .
image

# docker run --rm celestia-node-test:latest ls -la /home/celestia/entrypoint.sh
image

# docker run --rm celestia-node-test:latest celestia
image
image

# docker run --rm celestia-node-test:latest env | grep -E "(NODE_TYPE|P2P_NETWORK|CELESTIA_HOME)"
image

rootulp:
The CLI output is for celestia-node which is different from celestia-app.

DeVikingMark:
# 1. Building Docker images
## docker build -f docker/standalone.Dockerfile -t celestia-app-standalone .
image

## docker build -f docker/multiplexer.Dockerfile -t celestia-app-multiplexer .
image
image

## docker build -f docker/txsim/Dockerfile -t celestia-app-txsim .
image

# 2. Checking entrypoint.sh placement
## docker run --rm --entrypoint="" celestia-app-standalone ls -la /home/celestia/.celestia-app/entrypoint.sh
image

## docker build -f docker/multiplexer.Dockerfile -t celestia-app-multiplexer .
image

## docker run --rm --entrypoint="" celestia-app-txsim ls -la /home/celestia/.celestia-app/entrypoint.sh
image

# 3. Verifying entrypoint functionality
## docker run --rm celestia-app-standalone start --help
image
image
image

## docker run --rm celestia-app-multiplexer start --help
image
image
image

## docker run --rm celestia-app-txsim --help
image

rootulp:
@DeVikingMark why does this PR modify the Dockerfiles if the motivating issue is only for local_devnet: https://github.com/celestiaorg/celestia-app/issues/4670

DeVikingMark:
Hey @rootulp, sorry for the late answer,

While #4670 specifically mentions local_devnet, the underlying issue was that entrypoint scripts were being copied to /opt/entrypoint.sh, which requires root permissions.

The /opt directory requires root permissions, which isn't great for security. Moving everything to the user's home directory (${CELESTIA_APP_HOME}) is cleaner and more secure.

So I figured if we're fixing this pattern, we might as well fix it everywhere - standalone, multiplexer, and txsim containers.

Does that make sense?

rootulp:
That makes sense but I'm hesitant to include this without getting user feedback from node operators that use the existing Docker images. I don't want to break them unnecessarily.

rach-id:
I guess a first feedback should come from the devops team. cc @sysrex

sysrex (changes requested):
why would the entrypoint reside in the home folder? this has always been in /opt just like for node https://github.com/celestiaorg/celestia-node/blob/main/Dockerfile#L69

- also most installations will mount `/home/celestia/.celestia-app` on additional disks for obvious reasons, which means this can have unintended consequences.

DeVikingMark:
@sysrex thanks for the feedback! I've updated the pullrequest to apply changes only to local_devnet

docker/txsim/Dockerfile
```diff
@@ -47,11 +47,11 @@ RUN apk update && apk add --no-cache \
 # Copy in the txsim binary
 COPY --from=builder /celestia-app/build/txsim /bin/txsim
 
-COPY --chown=${USER_NAME}:${USER_NAME} docker/txsim/entrypoint.sh /opt/entrypoint.sh
+COPY --chown=${USER_NAME}:${USER_NAME} docker/txsim/entrypoint.sh ${CELESTIA_APP_HOME}/entrypoint.sh
```
rootulp, line 50: >  I've updated the pullrequest to apply changes only to local_devnet

txsim can be used outside of local_devnet

rootulp: I've re-read the original issue and I still don't understand the concrete problem and proposal. I think we need more clarity in https://github.com/celestiaorg/celestia-app/issues/4670 before we try implementing it.

rootulp:
Also now the PR description is out of date. IMO let's close this for now and ask for clarity in the original issue.

Issue #5835, opened 2025-09-24
Implement x/fibre module


rootulp:
## Context

https://github.com/celestiaorg/fibre-da-spec/pull/3

## Steps

- [x] Implement protos for x/fibre (.proto files, generated .pb.go files): https://github.com/celestiaorg/celestia-app/pull/5794
- [ ] Implement stateless validation `ValidateBasic` for all the types (including unit tests): https://github.com/celestiaorg/celestia-app/pull/5999
- [ ] Implement keepers for state related changes (including unit tests)
- [ ] Implement QueryServer
- [ ] Implement MsgServer + message handlers (including unit tests + events)
- [ ] Implement CLI commands

rootulp:
Moving to another repo.

PR #5999, opened 2025-10-08
feat(x/fibre): `ValidateBasic` methods

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/5835

To-do:
- [ ] Verify that this actually implements all the stateless validation described in the spec


rootulp:
Moving to another repo.

PR #6003, opened 2025-10-09
refactor(x/blob): remove unused BlobKeeper dependency from MinGasPFBDecorator

viktorking7:
## Description
Removes unused `BlobKeeper` dependency from `MinGasPFBDecorator` to simplify the decorator's API and reduce unnecessary coupling.

## Changes
- Remove unused `k BlobKeeper` field from `MinGasPFBDecorator` struct
- Update `NewMinGasPFBDecorator()` constructor to take no parameters
- Update all call sites in `app/ante/ante.go` and test files

## Rationale
The `BlobKeeper` dependency was never used in the decorator's logic but created false coupling and complicated the constructor signature. This change:
- Reduces API surface area
- Eliminates unnecessary dependency injection
- Maintains identical behavior while improving code clarity

### DIFF
--- a/x/blob/ante/ante_test.go
+++ b/x/blob/ante/ante_test.go
@@ -114,7 +114,7 @@ func TestPFBAnteHandler(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			anteHandler := ante.NewMinGasPFBDecorator(mockBlobKeeper{})
+            anteHandler := ante.NewMinGasPFBDecorator()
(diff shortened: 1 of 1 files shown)


viktorking7:
> this is a breaking change and a driveby
> 
> you or your llm is not wrong, but see contrib guidelines. also see the linter etc

@evan-forbes I can fix the lint

evan-forbes:
it's still a breaking change and drivebys aren't welcome

We get a billion of these PRs. See contribution guidelines. they don't fix anything consequential and aren't worth anyone's time or even the inference tbh

PR #6006, opened 2025-10-09
chore: remove duplicate minfee DefaultNetworkMinGasPrice init

menzudilat:
Cleaned up x/minfee legacy params: dropped the duplicate init of DefaultNetworkMinGasPrice (same logic lives in params.go) and removed the now-unused appconsts import. Behaviour stays the same, less dead code to maintain.


evan-forbes:
this is a driveby, and potetially a clean up, but I don't want to do the due dilligence to actually verify something that adds very little to no value to the project

see the contribution guidelines @menzudilat

Issue #2519, opened 2023-09-18
Improve `TestPrepareProposalConsistency`


evan-forbes:
The `TestPrepareProposalConsistency` test uses randomly sized transactions to test if PrepareProposal will ever create a proposal block that ProcessProposal rejects. While other tests are responsible for ensuring that a transaction that gets through CheckTx, gets through PrepareProposal, and then ProcessProposal, this test can still be improved upon by being more exact with how many transactions are expected to be included in the block.

https://github.com/celestiaorg/celestia-app/blob/b59e876cd15ea9f17608d1c2ca865a86151b7e5f/app/test/fuzz_abci_test.go#L144

ThanhNhann:
Hey @evan-forbes, can I handle this issue?

exp7l:
Greetings! @evan-forbes @rootulp 
Looks like this is still open. Can I work on it?

exp7l:
Looking

rootulp:
Closing this as won't do because it's not immediately clear what this issue is asking for. See https://github.com/celestiaorg/celestia-app/pull/3087#issuecomment-1932966792

cc: @evan-forbes please re-open with acceptance criteria if we still want this.

evan-forbes:
>  this test can still be improved upon by being more exact with how many transactions are expected to be included in the block.

I think this is the core piece, we aren't actually checking how many transactions are being included, only that one blobTx did. This means that the rest of the blob txs could be filtered out. Ideally we would make this more restrictive. This might involve writing a different or doing some analysis for what conservative % of txs should be expected to get through.

https://github.com/celestiaorg/celestia-app/blob/949a2d579619512e9e56eea10538fe847f12306d/app/test/fuzz_abci_test.go#L149-L155

rawbytedev:
Hi
Looks like this is still open. I would like to work on it

rawbytedev:
@evan-forbes 
> This means that the rest of the blob txs could be filtered out. Ideally we would make this more restrictive. This might involve writing a different or doing some analysis for what conservative % of txs should be expected to get through.

Would that mean that not all random blob are valid? But we are expecting at least 1 blob to be included 
So the improvement should be to set a min rate of blob that should be included in block for test to be valid

rawbytedev:
# This the result of test:
https://github.com/rawbytedev/celestia-app/actions/runs/18431780761/job/52520034678
the first test gives over 10 %
but the second drops to 7%:
[...]

PR #6012, opened 2025-10-11
test: Improve TestPrepareProposalConsistency

rawbytedev:
# Proposal
Improve TestPrepareProposalConsistency by setting a min rate(%) of blob to be included in block
#2519


evan-forbes (commented):
this isn't addressing the issue, its just adding redundant test

its also super sloppy and doesn't use the existing tooling

appreciate that you tried to address an issue, but this needs so much work that I think its okay to close for now.

PR #6013, opened 2025-10-11
refactor!: modify the latency monitor to submit blobs sequentially at a specific delay

evan-forbes:
this is low prio change, but this simplifies the latency monitor and uses cobra. if we don't want to change this that's fine I can keep a local copy for my own usage.

instead of working backwards from a rate of KiB/s and not having control over the time between submissions, this submits blobs of a given size with specified delay. This is really helpful for testing sequential submission. 

it is still keeping track of the amount of time it takes to submit each transaction, which is also helpful.

### DIFF
--- a/tools/latency-monitor/main.go
+++ b/tools/latency-monitor/main.go
@@ -52,44 +52,46 @@ type txResult struct {
 }
 
 func main() {
-	if err := rootCmd.Execute(); err != nil {
+	if err := newRootCmd().Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		//nolint:gocritic
 		os.Exit(1)
 	}
 }
 
-var rootCmd = &cobra.Command{
-	Use:   "latency-monitor",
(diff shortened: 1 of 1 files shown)


tools/latency-monitor/main.go
```diff
@@ -35,54 +46,68 @@ type txResult struct {
...
-
-	if err := monitorLatency(ctx, *endpoint, *keyringDir, *submitRate, *blobSize, *namespaceStr, *disableMetrics); err != nil {
+	if err := rootCmd.Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		//nolint:gocritic
 		os.Exit(1)
 	}
 }
 
+var rootCmd = &cobra.Command{
```
evan-forbes: using cobra over go's standard lib flag parsing etc saves lives so I switched to using it.

no more messing up the terminal with prints

ninabarbakadze (approved):

PR #6014, opened 2025-10-12
fix(user): fixes bug in setup where partial setup (loading accts but not funding) prevents future starts

renaynay:
This PR fixes a bug in `TxClient` setup where partial (failled or ctx cancelled) setup prevents future starts. This is bc some accounts would be created + added to keyring, but didn't get funded. Future starts expect if account exists --> means it's already funded. 

Ref https://github.com/celestiaorg/celestia-node/issues/4629

Also cleans up some unnecessary logic in `txQueue` with start/stop/initialisation determination, which IMO is unnecessary.

### DIFF
--- a/multiplexer/abci/multiplexer.go
+++ b/multiplexer/abci/multiplexer.go
@@ -78,7 +78,7 @@ type Multiplexer struct {
 	conn *grpc.ClientConn
 	// ctx is the context which is passed to the comet, grpc and api server starting functions.
 	ctx context.Context
-	// g is the waitgroup to which the comet, grpc and api server init functions are added to.
+	// g is the errgroup to which the gRPC server, API server, block event listener, and signal handler are added to.
 	g *errgroup.Group
 	// traceWriter is the trace writer for the multiplexer.
 	traceWriter io.WriteCloser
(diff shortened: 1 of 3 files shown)


pkg/user/parallel_tx_submission.go
```diff
@@ -231,63 +217,41 @@ func (w *txWorker) processJob(job *SubmissionJob, workerCtx context.Context) {
...
-	// Skip the first worker (index 0) as it always uses the existing signer account
-	var workersToInit []*txWorker
-	var workersToLoad []*txWorker
+	needFunding := make([]*txWorker, 0)
 	for i, worker := range p.workers {
 		if i == 0 {
 			// Skip first worker - it uses existing account
 			continue
 		}
+		if _, exists := p.client.signer.accounts[worker.accountName]; exists {
```
renaynay, line 231: this is unnecessary and I'd like to avoid this, but it allows test to pass, otherwise would have to extend mock grpc server

renaynay:
FYI @evan-forbes I made a test in node to catch this: https://github.com/celestiaorg/celestia-node/pull/4634

PR #6022, opened 2025-10-13
chore: fix the format of the comment

stemlaud:
## Overview

 fix the format of the comment


rach-id:
I think the change is valid given the existing comment is not correctly indented

PR #6027, opened 2025-10-13
test(e2e): validate v5 to v6 params and unify expectations

### LINKED ISSUE #5987: e2e major upgrade test v5 -> v6
## Context

https://www.notion.so/celestiaorg/Testing-and-QA-for-V6-Checklist-2462a70c9515809cafddf7edd4dc867d#2462a70c9515808c80cfcb4965714f8c

## Problem

The e2e major upgrade test currently doesn't cover all the acceptance criteria we discussed. Ref: https://github.com/celestiaorg/celestia-app/blob/923a4dd423f0911ecf5ba05d587fc9d232cadabb/test/docker-e2e/e2e_upgrade_test.go#L21

## Proposal

Modify to the v5 -> v6 upgrade test to check:
- [ ]  Decreasing inflation
- [ ]  Params
    - [ ]  Decrease unbonding period
    - [ ]  Evidence params
    - [ ]  ICA host params
    - [ ]  Min commission rate

mojtaba-esk:
Closes #5987

### DIFF
--- a/tools/latency-monitor/README.md
+++ b/tools/latency-monitor/README.md
@@ -4,12 +4,13 @@ A tool for monitoring and measuring transaction latency in Celestia networks. Th
 
 ## Features
 
-- Configurable transaction submission rate (in KB/s)
-- Adjustable blob size
+- Configurable submission delay between transactions
+- Random blob sizes (up to a specified maximum)
 - Custom namespace support
 - Real-time transaction monitoring
+- Tracks both successful and failed confirmations
 - Detailed latency statistics including mean and standard deviation
-- CSV export of all transaction results
+- CSV export of all transaction results with failure tracking
 
 ## Prerequisites
 
@@ -25,11 +26,15 @@ A tool for monitoring and measuring transaction latency in Celestia networks. Th
 
 ### Available Flags
 
-- `-grpc-endpoint`: gRPC endpoint to connect to (default: "localhost:9090")
-- `-keyring-dir`: Directory containing the keyring (default: "~/.celestia-app")
-- `-submit-rate`: Data submission rate in KB/sec (default: 1.0)
-- `-blob-size`: Size of blob data in bytes (default: 1024)
-- `-namespace`: Namespace for blob submission (default: "test")
+| Flag | Shorthand | Default | Description |
+|------|-----------|---------|-------------|
+| `--grpc-endpoint` | `-e` | `localhost:9090` | gRPC endpoint to connect to |
+| `--keyring-dir` | `-k` | `~/.celestia-app` | Directory containing the keyring |
+| `--account` | `-a` | _(first account)_ | Account name to use from keyring |
+| `--blob-size` | `-b` | `1024` | Maximum blob size in bytes (blobs will be random size from 1 to this value) |
+| `--submission-delay` | `-d` | `4000ms` | Delay between transaction submissions |
+| `--namespace` | `-n` | `test` | Namespace for blob submission |
+| `--disable-metrics` | `-m` | `false` | Disable metrics collection |
 
 ### Example
 
@@ -37,25 +42,42 @@ A tool for monitoring and measuring transaction latency in Celestia networks. Th
 # Run with default settings (from the root directory)
 go run ./tools/latency-monitor
 
-# Run with custom settings
-go run ./tools/latency-monitor -grpc-endpoint=localhost:9090 -submit-rate=2.0 -blob-size=2048 -namespace=custom
+# Run with custom settings (long flags)
+go run ./tools/latency-monitor --grpc-endpoint localhost:9090 --submission-delay 200ms --blob-size 2048 --namespace custom
+
+# Run with custom settings (short flags)
+go run ./tools/latency-monitor -e localhost:9090 -d 200ms -b 2048 -n custom
+
+# Use a specific account from keyring
+go run ./tools/latency-monitor -a validator
+
+# View help
+go run ./tools/latency-monitor --help
 ```
 
 ## Output
 
 The tool provides:
 
-1. Real-time updates every 10 seconds showing the number of transactions submitted
+1. Real-time logging for each transaction:
+   - `[SUBMIT]` when a blob is broadcast (with tx hash, size, and timestamp)
+   - `[CONFIRM]` when a transaction is confirmed (with tx hash, height, latency, code, and timestamp)
+   - `[FAILED]` when confirmation fails (with tx hash and error)
+   - Status updates every 10 seconds showing the number of transactions submitted
 2. A CSV file (`latency_results.csv`) containing:
    - Submit time
    - Commit time
    - Latency (in milliseconds)
    - Transaction hash
+   - Block height
    - Transaction code
+   - Failed status (true/false)
+   - Error message (if failed)
 3. Final statistics including:
    - Total number of transactions
-   - Average latency
-   - Standard deviation
+   - Success/failure counts and percentages
+   - Average latency (for successful transactions)
+   - Standard deviation (for successful transactions)
 
 ## Stopping the Tool
 
--- a/tools/latency-monitor/main.go
+++ b/tools/latency-monitor/main.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/csv"
-	"flag"
 	"fmt"
 	"math"
+	mathrand "math/rand"
 	"os"
 	"os/signal"
 	"sync"
@@ -17,16 +17,27 @@ import (
 	"github.com/celestiaorg/celestia-app/v6/pkg/user"
 	"github.com/celestiaorg/go-square/v3/share"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
 const (
-	defaultEndpoint     = "localhost:9090"
-	defaultKeyringDir   = "~/.celestia-app"
-	defaultSubmitRate   = 1.0    // KB per second
-	defaultBlobSize     = 1      // KB
-	defaultNamespaceStr = "test" // default namespace for blobs
+	defaultEndpoint        = "localhost:9090"
+	defaultKeyringDir      = "~/.celestia-app"
+	defaultBlobSize        = 1024                    // bytes
+	defaultNamespaceStr    = "test"                  // default namespace for blobs
+	defaultSubmissionDelay = 4000 * time.Millisecond // delay between submissions
+)
+
+var (
+	endpoint        string
+	keyringDir      string
+	accountName     string
+	blobSize        int
+	namespaceStr    string
+	disableMetrics  bool
+	submissionDelay time.Duration
 )
 
 type txResult struct {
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -137,6 +137,7 @@ jobs:
             entrypoint: TestIBCTestSuite
           - testcase: TestStateSyncWithAppUpgrade
           - testcase: TestMinorVersionCompatibility
+          - testcase: TestCelestiaAppV5ToV6UpgradeValidation
```
rootulp: Why is this a new test? Can we add the validation after the existing major version upgrade test? I think that's `TestCelestiaAppUpgrade` but it can be renamed to include major version in there

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -2,45 +2,48 @@ package docker_e2e
...
 	"github.com/celestiaorg/tastora/framework/testutil/wait"
 	tastoratypes "github.com/celestiaorg/tastora/framework/types"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	"celestiaorg/celestia-app/test/docker-e2e/dockerchain"
 )
 
-// TestCelestiaAppUpgrade tests app version upgrade using the signaling mechanism.
-func (s *CelestiaTestSuite) TestCelestiaAppUpgrade() {
```
rootulp, line 20: [blocking][question] why does this delete `TestCelestiaAppUpgrade`. And if it does, shouldn't `test-docker-e2e` be updated to remove it?

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -2,45 +2,48 @@ package docker_e2e
...
-		baseAppVersion   uint64
-		targetAppVersion uint64
-	}{
-		{
-			baseAppVersion:   5,
-			targetAppVersion: 6,
-		},
-	}
+	UnbondingTimeV5Hours = 337 // Current codebase default (CIP-037 already applied)
+	UnbondingTimeV6Hours = 337 // Same as v5 (no change in this upgrade test)
```
rootulp: [blocking] the unbodning time in v5 shouldn't be the same as v6. This looks like AI generated and not manually reviewed.

Ref: https://github.com/celestiaorg/CIPs/blob/main/cips/cip-037.md

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -2,45 +2,48 @@ package docker_e2e
...
-	for _, tc := range tt {
-		s.Run(fmt.Sprintf("upgrade from v%d to v%d", tc.baseAppVersion, tc.targetAppVersion), func() {
-			s.runUpgradeTest(tag, tc.baseAppVersion, tc.targetAppVersion)
-		})
-	}
-}
+	MinCommissionRateV5 = "0.05" // 5%
+	MinCommissionRateV6 = "0.10" // 10%
+
+	EvidenceMaxAgeV5Hours = 337 // Current codebase default
```
rootulp: [blocking] this should be 507

Ref: https://github.com/celestiaorg/CIPs/blob/main/cips/cip-037.md

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -2,45 +2,48 @@ package docker_e2e
...
+	EvidenceMaxAgeV6Hours = 337 // Same as v5 (no change in this upgrade test)
+
+	EvidenceMaxAgeV5Blocks = 242640 // Current codebase default
```
rootulp: [blocking] this should be 120,960

Ref: https://github.com/celestiaorg/CIPs/blob/main/cips/cip-037.md

.github/workflows/test.yml
mojtaba-esk: `TestCelestiaAppUpgrade` is too generic while `TestCelestiaAppV5ToV6UpgradeValidation` refers exactly to `v5` to `v6` and validating parameters before and after upgrade.

test/docker-e2e/e2e_upgrade_test.go
mojtaba-esk: It is actually renamed to better reflect what we evaluate before and after v5 to v6 upgrade, here it shows it as delete and then add a new one. and yes we should update the `test-docker-e2e`

test/docker-e2e/e2e_upgrade_test.go
mojtaba-esk: oh I see, I thought it happened before v5. will fix it. thanks for the note

mojtaba-esk: and yes AI helped in finding the values in the right CIPs

mojtaba-esk:
> I only reviewed part of this but this isn't ready for human review. There's obviously AI generated content that hasn't been reviewed by a human yet. Please review the PR yourself before asking for reviews.

The code was definitely reviewed by myself, corrected some parts, iterated over it and ran a couple of times to make sure things are alright. AI helped in the research about the values in CIPs and that's where it goes inaccurate. Those 3 values are updated, but the tests fail:

```sh
...
    e2e_upgrade_test.go:386: 
                Error Trace:    /home/moji/celestia/celestia-app/test/docker-e2e/e2e_upgrade_test.go:386
                                                        /home/moji/celestia/celestia-app/test/docker-e2e/e2e_upgrade_test.go:315
                                                        /home/moji/celestia/celestia-app/test/docker-e2e/e2e_upgrade_test.go:116
                                                        /home/moji/go/pkg/mod/github.com/stretchr/testify@v1.11.1/suite/suite.go:196
                Error:          Not equal: 
                                expected: 504h0m0s
                                actual  : 337h0m0s
                Test:           TestCelestiaTestSuite/TestCelestiaAppV5ToV6UpgradeValidation
                Messages:       v5 unbonding time mismatch: expected 504h0m0s (504 hours), got 337h0m0s
...
```

Did a bit of investigation and found out here in [this commit](https://github.com/celestiaorg/celestia-app/commit/9991db103ca94336f11064dccc404a7b49cfb10f) where CIP037 is applied. It is applied to the entire binary, correct me if I am wrong, but it seems like the app version does not play a role here. Therefore the entire test will be meaningless unless we use a binary swap test which runs an old binary v5.x.x then run a newer => v6.0.0 and evaluate params.

Can you please tell me how to proceed?

mojtaba-esk:
Ok after some investigations found out that the latest binary already includes CIP-037 defaults, so starting a chain at app version 5 would not show legacy values by default but if we explicitly preset the legacy (pre–CIP-037) params in the v5 genesis  and then upgrade to v6 the new values should be set. So the code is updated to do so.

.github/workflows/test.yml
rootulp: Can we please shorten it to `TestCelestiaAppV5ToV6`

- Upgrade is implied by `V5ToV6`
- Validation occurs in all tests so it's adding any value in the test name.

rootulp: [blocking] `TestCelestiaAppUpgrade` is still in this list and it shouldn't be.

Makefile
```diff
@@ -301,7 +301,7 @@ test-docker-e2e-upgrade:
 	@echo "--> Building celestia-appd docker image (tag $(CELESTIA_TAG))"
 	@DOCKER_BUILDKIT=0 docker build --build-arg BUILDPLATFORM=linux/amd64 --build-arg TARGETOS=linux --build-arg TARGETARCH=amd64 -t "ghcr.io/celestiaorg/celestia-app:$(CELESTIA_TAG)" . -f docker/multiplexer.Dockerfile
 	@echo "--> Running upgrade test"
-	cd test/docker-e2e && go test -v -run ^TestCelestiaTestSuite/TestCelestiaAppUpgrade$$ -count=1 ./... -timeout 15m
+	cd test/docker-e2e && go test -v -run ^TestCelestiaTestSuite/TestAllUpgrades$$ -count=1 ./... -timeout 15m
```
rootulp, line 304: [nit] IMO we should clarify which upgrade test this is running b/c this PR is changing it from running the singular app upgrade test to all upgrades. The comments and echo statements just say "upgrade test" but thats not specific enough considering there are now multiple upgrade tests.

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -82,6 +86,53 @@ func (s *CelestiaTestSuite) TestAllUpgrades() {
 	}
 }
 
+// TestCelestiaAppV5ToV6UpgradeValidation validates that parameters are correctly applied across the v5 to v6 upgrade,
+// including changes introduced by CIP-042 as well as earlier CIPs and pre-v5 defaults.
```
rootulp: How does this test verify pre-v5 defaults? IMO this is clearer:

```suggestion
// including changes introduced by CIP-042.
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -82,6 +86,53 @@ func (s *CelestiaTestSuite) TestAllUpgrades() {
...
+func (s *CelestiaTestSuite) TestCelestiaAppV5ToV6UpgradeValidation() {
```
rootulp: [nit]

```suggestion
func (s *CelestiaTestSuite) TestCelestiaAppV5ToV6() {
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -82,6 +86,53 @@ func (s *CelestiaTestSuite) TestAllUpgrades() {
...
+	if testing.Short() {
+		s.T().Skip("skipping v5 to v6 upgrade validation test in short mode")
```
rootulp:
```suggestion
		s.T().Skip("skipping v5 to v6 test in short mode")
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -256,3 +304,192 @@ func getSignalQueryClient(node tastoratypes.ChainNode) (signaltypes.QueryClient,
 	}
 	return nil, nil, fmt.Errorf("GRPC connection is nil")
 }
+
+// getICAHostQueryClient returns an icahosttypes.QueryClient for the provided node.
+// If the node is a docker ChainNode with a live *grpc.ClientConn, it is reused.
+// Returns an error if no gRPC connection is available.
+func getICAHostQueryClient(node tastoratypes.ChainNode) (icahosttypes.QueryClient, func(), error) {
```
rootulp: [blocking] what is the second return argument here for? In the happy path it's an empty function, in the failure path it's nil. It seems completely useless.

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -256,3 +304,192 @@ func getSignalQueryClient(node tastoratypes.ChainNode) (signaltypes.QueryClient,
...
+	if dcNode, ok := node.(*tastoradockertypes.ChainNode); ok && dcNode.GrpcConn != nil {
+		return icahosttypes.NewQueryClient(dcNode.GrpcConn), func() {}, nil
+	}
+	return nil, nil, fmt.Errorf("GRPC connection is nil")
+}
+
+// validateParameters validates that all parameters match expected values for the given app version
+func (s *CelestiaTestSuite) validateParameters(ctx context.Context, node tastoratypes.ChainNode, version uint64) {
```
rootulp: [nit]

```suggestion
func (s *CelestiaTestSuite) validateParameters(ctx context.Context, node tastoratypes.ChainNode, appVersion uint64) {
```

test/docker-e2e/e2e_upgrade_test.go
```diff
@@ -256,3 +304,192 @@ func getSignalQueryClient(node tastoratypes.ChainNode) (signaltypes.QueryClient,
...
+		return
+	}
+
+	s.validateInflationRate(ctx, node, InflationRateV6, AppVersionV6)
+	s.validateUnbondingTime(ctx, node, UnbondingTimeV6Hours, AppVersionV6)
+	s.validateMinCommissionRate(ctx, node, MinCommissionRateV6, AppVersionV6)
+	s.validateEvidenceParams(ctx, node, EvidenceMaxAgeV6Hours, EvidenceMaxAgeV6Blocks, AppVersionV6)
+	// Check ICA host params only on v6: v5 doesn't expose the icahost gRPC query service;
+	// the v6 upgrade applies these params per CIP-14.
+	s.validateICAHostParams(ctx, node, true, app.IcaAllowMessages(), AppVersionV6)
```
rootulp: [blocking] IMO this should be gated behind a conditional for future proofing

```suggestion
	if version == AppVersionV6 {
		s.validateInflationRate(ctx, node, InflationRateV6, AppVersionV6)
		s.validateUnbondingTime(ctx, node, UnbondingTimeV6Hours, AppVersionV6)
		s.validateMinCommissionRate(ctx, node, MinCommissionRateV6, AppVersionV6)
		s.validateEvidenceParams(ctx, node, EvidenceMaxAgeV6Hours, EvidenceMaxAgeV6Blocks, AppVersionV6)
		// Check ICA host params only on v6: v5 doesn't expose the icahost gRPC query service;
		// the v6 upgrade applies these params per CIP-14.
		s.validateICAHostParams(ctx, node, true, app.IcaAllowMessages(), AppVersionV6)
	}
```

.github/workflows/test.yml
mojtaba-esk: nice catch, removed

Makefile
mojtaba-esk: Good point. yes that has to be specified which test to run the singular one that tests the params which is offered by this PR or All versions Upgrades.
We can also add another test to test upgrade from `currentVersion -1` to the `currentVersion` for this command.

test/docker-e2e/e2e_upgrade_test.go
mojtaba-esk: it is a placeholder for a cleanup callback. If this helper ever dials its own gRPC connection (if needed in future), callers can `defer cleanup()` to release it. right now `dcNode.GrpcConn` is reused, so there's nothing to close.
If it is blocking, we can remove it for now and if in future it is needed, we will do a refactor on the caller(s)

rootulp: I don't see how it's a placeholder for a cleanup, the func does nothing. We should remove the placeholder second return argument.

rootulp (approved):

PR #6028, opened 2025-10-14
fix: create missing intermediate maps for SetField

lolkdsanyaw:
Updated SetField to auto-create missing intermediate maps when inserting nested JSON fields, kept RemoveField’s error semantics unchanged, added unit tests covering both behaviors, and ran go test ./internal/utils

### DIFF
--- a/internal/utils/maps.go
+++ b/internal/utils/maps.go
@@ -43,9 +43,16 @@ func setOrDeleteNestedField(doc map[string]any, path string, value any) error {
 			return nil
 		}
 
-		next, ok := current[key].(map[string]any)
+		nextValue, exists := current[key]
+		next, ok := nextValue.(map[string]any)
 		if !ok {
-			return fmt.Errorf("invalid path: %s is not a map", strings.Join(keys[:i+1], "."))
+			// create missing intermediate map when inserting new fields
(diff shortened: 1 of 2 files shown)


rach-id:
@lolkdsanyaw what's the motivation behind this PR?

lolkdsanyaw:
> @lolkdsanyaw what's the motivation behind this PR?

@rach-id SetField used to fail when the path introduced a new nested branch because intermediate maps were missing. This patch makes it create those maps on demand so genesis-edit and similar tooling stop erroring, and the new tests capture the fixed behavior.

rach-id:
Where did it fail before? Which commands made it fail?

lolkdsanyaw:
I discovered this issue during code review of setOrDeleteNestedField - it couldn't handle missing intermediate paths, causing 'invalid path: ... is not a map' errors. This affects legacy genesis export in DocumentLegacyBytes when setting consensus_params.version.app_version. I added test cases to demonstrate the problem and validate the fix. To reproduce the original issue: revert the changes in maps.go and run go test ./internal/utils - the new test cases will fail.

rach-id:
This seems like a drive by PR and the issue is not reproducible in any possible way. Closing.

If you want to contribute, check good first issues. Drive-by PRs will always be closed.

PR #6029, opened 2025-10-14
fix(cmd): add HTTP status validation in genesis download

kilavvy:
## Summary

Add HTTP status code validation in `downloadFile()` to prevent saving error response bodies as genesis files.

## Problem

The `download-genesis` command would save error pages (404/500 HTML responses) to the genesis file without validation, causing confusing "sha256 hash mismatch" errors later instead of clear download failures.

## Solution

- Add `resp.StatusCode != http.StatusOK` check before file creation
- Return descriptive error with actual HTTP status for early failure detection
- Prevents corrupted genesis files from being written to disk

### DIFF
--- a/cmd/celestia-appd/cmd/init.go
+++ b/cmd/celestia-appd/cmd/init.go
@@ -187,5 +187,6 @@ func displayInfo(info printInfo) error {
 func getConsensusParams() *comettypes.ConsensusParams {
 	params := comettypes.DefaultConsensusParams()
 	params.Version.App = appconsts.Version
+	params.Block.MaxBytes = appconsts.DefaultMaxBytes
 	return params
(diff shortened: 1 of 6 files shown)


rootulp:
Thanks! Out of curiosity, did you encounter this as a real problem? Or what motivated this code change?

rootulp (dismissed):
Code change LGTM but I haven't tested it locally

kilavvy:
> Code change LGTM but I haven't tested it locally

@rootulp fixed gci formatting.

ninabarbakadze (commented):
lint is broken pls fix

kilavvy:
@rootulp fixed lint locally, but failed again

PR #6031, opened 2025-10-15
chore: make the nightly test sequential

mojtaba-esk:
This PR attempts to reduce flakes in the nightly tests where the test chain sometimes fails to start on shared GitHub runners. It tries to run them sequentially and adds some retrials.

### DIFF
--- a/.github/copilot-setup-steps.yml
+++ b/.github/copilot-setup-steps.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: "go.mod"
 
--- a/tools/throughput/main.go
+++ b/tools/throughput/main.go
@@ -92,32 +92,29 @@ func Run() error {
 
 	metrics := NewBlockMetrics(100) // Keep track of last 100 blocks
 
-	newBlocksSub, err := client.Subscribe(ctx, "blocks-watcher", "tm.event = 'NewBlock'")
+	newHeightSub, err := client.Subscribe(ctx, "heights-watcher", "tm.event = 'NewBlockHeader'")
 	if err != nil {
-		return fmt.Errorf("failed to subscribe to new blocks: %w", err)
+		return fmt.Errorf("failed to subscribe to new block headers: %w", err)
 	}
 	defer func() {
-		err := client.Unsubscribe(ctx, "blocks-watcher", "tm.event = 'NewBlock'")
+		err := client.Unsubscribe(ctx, "heights-watcher", "tm.event = 'NewBlockHeader'")
 		if err != nil {
 			fmt.Printf("error while unsubscribing: %s", err.Error())
 		}
 	}()
 
-	fmt.Println("Listening for new blocks...")
+	fmt.Println("Listening for new heights...")
 
 	for {
 		select {
-		case event := <-newBlocksSub:
-			blockEvent := event.Data.(types.EventDataNewBlock)
-			block := blockEvent.Block
+		case event := <-newHeightSub:
+			heightEvent := event.Data.(types.EventDataNewBlockHeader)
+			header := heightEvent.Header
 
 			// Calculate block size (including transactions)
-			blockSize := 0
-			for _, tx := range block.Txs {
-				blockSize += len(tx)
-			}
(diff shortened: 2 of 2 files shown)


tty47 (approved):

mojtaba-esk:
The test ran manually and it is passing: https://github.com/celestiaorg/celestia-app/actions/runs/18524032734 
but we have to see it in action if it keeps passing in the next days after getting this PR merged or will be flaky again.

rootulp:
It failed on it's first attempt: https://github.com/celestiaorg/celestia-app/actions/runs/18524032734/job/52790478156 and only succeeded on its second.

.github/workflows/test-nightly.yml
```diff
@@ -36,17 +40,21 @@ jobs:
           fi
 
       - name: Build Docker image locally
-        if: ${{ !inputs.tag }} # only build an image if no tag is specified
+        if: ${{ inputs.tag == '' }} # only build an image if no tag is specified
```
rootulp, line 43: why did this change?

.github/workflows/test-nightly.yml
```diff
@@ -36,17 +40,21 @@ jobs:
...
         run: docker build -t "ghcr.io/celestiaorg/celestia-app:${{ steps.tag.outputs.value }}" . -f docker/multiplexer.Dockerfile
 
-      - name: Run TestStateSyncMocha
+      - name: Run TestStateSyncMocha (with retry)
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 55
```
rootulp: do we really expect the Mocha state sync to take 55 minutes? Seems to me like this should be max ~10 minutes.

.github/workflows/test-nightly.yml
```diff
@@ -68,15 +76,18 @@ jobs:
...
       - name: Build Docker image locally
-        if: ${{ !inputs.tag }} # only build an image if no tag is specified
+        if: ${{ inputs.tag == '' }} # only build an image if no tag is specified
         run: docker build -t "ghcr.io/celestiaorg/celestia-app:${{ steps.tag.outputs.value }}" . -f docker/multiplexer.Dockerfile
 
-      - name: Run TestAllUpgrades
+      - name: Run TestAllUpgrades (with retry)
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 55
```
rootulp: same, instead of putting 55 minutes here, we should analyze historical successful test runs and find a rough average add a slight buffer and use that as the timeout.

.github/workflows/test-nightly.yml
mojtaba-esk: b/c gh Actions treats inputs as strings, using `!` on a string is unreliable. The explicit check `inputs.tag == ''` is more robust way to detect if no tag is provided.

.github/workflows/test-nightly.yml
mojtaba-esk: Not really, it usually takes less than 10 mins. here I thought it might be waiting for the resources as it was the main cause of flakiness. Higher timeout might help reducing it. We could try some number around 20 mins as well.

.github/workflows/test-nightly.yml
mojtaba-esk: Good idea let me go for it

mojtaba-esk: The stateSync took usually around 10 mins, last time 11 mins and 04 seconds, so let's change it to 20 and the AllUpgrades test took around 15 to 16 mins so 20 should be a safe threshold for both

rootulp:
Why do you expect running the tests sequentially to reduce flakes?

.github/workflows/test-nightly.yml
mojtaba-esk: I tried another time to run it manually when the failed test wanted to retry, due to existence of some config files from the failed test, could not retry. So the retrial was removed and therefore that timeout was removed as well. Therefore we rely on the timeout in the makefile as it is shorter than the timeout on the job in this ci conf file.

mojtaba-esk:
> Why do you expect running the tests sequentially to reduce flakes?

bc the gh runner is very limited in resource and when running concurrently, the tests cannot get enough resources to start and therefore they either timeout or fail to start. This might help.

rootulp (approved):

PR #6037, opened 2025-10-16
fix: Add Trusted Celestia Root and Height to ZKISM and Message Types (using Protobuf)

jonas089:
Context:

I was not 100% sure but I revisited the code and found why it is not safe to skip any DA blocks (even those with 0 transactions in them). 

We assert that the header matches the input from the ClientExecutor (EV-RETH) after applying the transactions to the trusted state. So a malicious actor could hide transactions and supply an invalid executor input to our circuit, convincing the verifier that this is a valid state transition. 

Exclusion proofs are absolutely necessary for the soundness of this prover.
[@Damian](https://celestia-team.slack.com/team/U08BJBR105A) If that's true, then we also have to update the ZKISM. 

Because the ZKISM currently only asserts that the new evolve height is > the trusted one. But it does not assert that the range prover used the correct Celestia start height (it could just skip some DA blocks with transactions in them). (edit

### DIFF
--- a/x/zkism/keeper/events.go
+++ b/x/zkism/keeper/events.go
@@ -20,6 +20,8 @@ func EmitCreateISMEvent(ctx sdk.Context, ism types.ZKExecutionISM) error {
 	return ctx.EventManager().EmitTypedEvent(&types.EventCreateZKExecutionISM{
 		Id:                  ism.Id,
 		Owner:               ism.Owner,
+		CelestiaHeaderHash:  encodeHex(ism.CelestiaHeaderHash),
+		CelestiaHeight:      ism.CelestiaHeight,
 		StateRoot:           encodeHex(ism.StateRoot),
 		Height:              ism.Height,
 		Namespace:           namespace.String(),
@@ -33,9 +35,11 @@ func EmitCreateISMEvent(ctx sdk.Context, ism types.ZKExecutionISM) error {
 // EmitUpdateISMEvent emits a typed event to signal updating of an ism.
 func EmitUpdateISMEvent(ctx sdk.Context, ism types.ZKExecutionISM) error {
 	return ctx.EventManager().EmitTypedEvent(&types.EventUpdateZKExecutionISM{
-		Id:        ism.Id,
-		StateRoot: encodeHex(ism.StateRoot),
-		Height:    ism.Height,
+		Id:                 ism.Id,
+		StateRoot:          encodeHex(ism.StateRoot),
+		Height:             ism.Height,
+		CelestiaHeaderHash: encodeHex(ism.CelestiaHeaderHash),
+		CelestiaHeight:     ism.CelestiaHeight,
 	})
 }
 


x/zkism/keeper/keeper.go
```diff
@@ -120,6 +120,21 @@ func (k *Keeper) validatePublicValues(ctx context.Context, height uint64, ism ty
...
+	if len(publicValues.PrevCelestiaHeaderHash[:]) != 32 {
+		return errorsmod.Wrapf(types.ErrInvalidStateRoot, "expected 32 bytes, got %d", len(publicValues.PrevCelestiaHeaderHash))
+	}
+
+	var celestiaIsmRoot [32]byte
+	copy(celestiaIsmRoot[:], ism.CelestiaStateRoot)
+
+	if publicValues.PrevCelestiaHeaderHash != celestiaIsmRoot {
+		return errorsmod.Wrapf(types.ErrInvalidStateRoot, "expected %x, got %x", ism.CelestiaStateRoot, publicValues.PrevCelestiaHeaderHash)
+	}
```
damiannolan: Let's replace this with a `if !bytes.Equal() { ... }` check, that is functionality equivalent and aligns with the existing code.

x/zkism/keeper/msg_server.go
```diff
@@ -74,13 +77,15 @@ func (m msgServer) UpdateZKExecutionISM(ctx context.Context, msg *types.MsgUpdat
 	if err != nil {
 		return nil, err
 	}
-
+	fmt.Println("checkpoint 1")
```
damiannolan: we can remove

x/zkism/keeper/msg_server_test.go
```diff
@@ -57,10 +63,14 @@ func (suite *KeeperTestSuite) TestCreateZKExecutionISM() {
 }
 
 func (suite *KeeperTestSuite) TestUpdateZKExecutionISM() {
-	trustedRoot, err := hex.DecodeString("af50a407e7a9fcba29c46ad31e7690bae4e951e3810e5b898eda29d3d3e92dbe")
+	trustedRoot, err := hex.DecodeString("340623e91b5a72c72a9214357cbc99a6b59ef23e5069ac5354a3a1619af0d568")
+	suite.Require().NoError(err)
+	trustedCelestiaHash, err := hex.DecodeString("0a02e7b488766f5ba73f8b44d96e97e27ca61580050e4a798bb664216876aa44")
+	suite.Require().NoError(err)
+	trustedCelestiaHeight := uint64(29)
 	suite.Require().NoError(err)
```
damiannolan, line 71: additional error check is redundant

x/zkism/types/public_values.go
```diff
@@ -11,8 +11,14 @@ import (
...
 // EvExecutionPublicValues defines the set of public values associated with a proof generated by a ZK program.
 type EvExecutionPublicValues struct {
+	// PrevCelestiaHeaderHash is the hash of the celestia header at which previous_height is available.
+	PrevCelestiaHeaderHash [32]byte
+	// PreviousCelestiaHeight is the height of the celestia header at which previous_height is available.
+	PrevCelestiaHeight uint64
 	// CelestiaHeaderHash is the hash of the celestia header at which new_height is available.
 	CelestiaHeaderHash [32]byte
+	// NewCelestiaHeight is the height of the celestia header at which new_height is available.
+	NewCelestiaHeight uint64
```
damiannolan, line 21: Including this now in public values means we can omit it in the `MsgUpdateZKExecutionISM` message.

We would use this height to look up the celestia header hash for this header on-chain, and validate it against the `CelestiaHeaderHash` above.

proto/celestia/zkism/v1/types.proto
```diff
@@ -22,16 +22,20 @@ message ZKExecutionISM {
   bytes state_root = 3;
   // latest tracked height of the remote chain
   uint64 height = 4;
+  // trusted celestia root
+  bytes celestia_state_root = 5;
```
damiannolan: A far as I understand from this PR, the value being used for this field is not actually the state root, but rather the hash of the entire header which includes both the state root and data availability root..etc

We should update this field name to reflect that accurately!

proto/celestia/zkism/v1/tx.proto
```diff
@@ -77,10 +80,14 @@ message MsgUpdateZKExecutionISM {
...
 // MsgUpdateZKExecutionISMResponse is the response type for UpdateZKExecutionISM.
 message MsgUpdateZKExecutionISMResponse {
-  // updated ism trusted state root (hex-encoded)
+  // state commitment root of the remote chain
   string state_root = 1;
-  // updated ism trusted height
+  // latest tracked height of the remote chain
   uint64 height = 2;
+  // trusted celestia root
+  string celestia_state_root = 3;
```
damiannolan: same comment on state root

proto/celestia/zkism/v1/tx.proto
```diff
@@ -29,23 +29,26 @@ service Msg {
...
   option (cosmos.msg.v1.signer) = "creator";
-
   // creator is the message sender.
   string creator = 1;
   // state commitment root of the remote chain
   bytes state_root = 2;
   // latest tracked height of the remote chain
   uint64 height = 3;
+  // trusted celestia root
+  bytes celestia_state_root = 4;
```
damiannolan: ditto

proto/celestia/zkism/v1/events.proto
```diff
@@ -18,20 +18,24 @@ message EventCreateZKExecutionISM {
   ];
   // the owner or creator of the ism
   string owner = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
-  // state commitment root of the remote chain (hex encoded)
+  // state commitment root of the remote chain
   string state_root = 3;
   // latest tracked height of the remote chain
   uint64 height = 4;
+  // trusted celestia root
+  string celestia_state_root = 5;
```
damiannolan: ditto

build-local.sh
damiannolan, line 1: We should remove this file.

You can build docker images in celestia-app by going to the Actions tab on GitHub and using the workflow dispatch on the "Docker build and push" workflow

proto/celestia/zkism/v1/types.proto
jonas089: yes I was planning to do that too, already updated it in the other PR (prover repo)

damiannolan: Ah nice one! 🙏🏻

x/zkism/types/public_values.go
damiannolan: Opened https://github.com/celestiaorg/celestia-app/issues/6068 which can be done in a follow up.

x/zkism/keeper/events.go
damiannolan, line 1: I added the new fields to events emitted here which was missing 🙏🏻

jonas089: Thanks for helping me finish this :D

PR #6039, opened 2025-10-16
refactor: preallocate slice capacity in Versions.Sorted()

animeherox:
Preallocate slice capacity when we know the final size to avoid unnecessary reallocations during append operations.


evan-forbes (commented):
this is a valid change but we have to stop letting drive by PRs in if we want them to stop

see contribution guidlines @animeherox

animeherox:
> this is a valid change but we have to stop letting drive by PRs in if we want them to stop
> 
> see contribution guidlines @animeherox

Got it, Will check the guidlines( not sure if I understand the means drive by PRs TBH)

PR #6040, opened 2025-10-17
feat: sequence aware mempool gossiping

evan-forbes:
## Overview

this PR adds the bump to comet that completes the sequence aware mempool.

blocked by official releases from core and sdk

### DIFF
--- a/tools/talis/deployment.go
+++ b/tools/talis/deployment.go
@@ -54,6 +54,28 @@ func upCmd() *cobra.Command {
 				return fmt.Errorf("failed to create client: %w", err)
 			}
 
+			// Check for running experiments
+			runningDroplets, err := checkForRunningExperiments(cmd.Context(), client.do)
+			if err != nil {
+				log.Printf("⚠️  Warning: failed to check for running experiments: %v", err)
+			}
+			if len(runningDroplets) > 0 {
+				log.Println("⚠️  Warning: Found another talis experiment running!")
+				log.Println("\nRunning validators:")
+				for _, d := range runningDroplets {
+					publicIP := ""
+					for _, net := range d.Networks.V4 {
+						if net.Type == "public" {
+							publicIP = net.IPAddress
+							break
+						}
+					}
+					log.Printf("  - %s (Region: %s, IP: %s, Created: %s)", d.Name, d.Region.Slug, publicIP, d.Created)
+				}
+				log.Printf("\nTotal: %d instance(s)\n", len(runningDroplets))
+				return fmt.Errorf("existing talis experiments are running")
+			}
+
 			if err := client.Up(cmd.Context(), workers); err != nil {
 				return fmt.Errorf("failed to spin up network: %w", err)
 			}
--- a/tools/talis/digital_ocean.go
+++ b/tools/talis/digital_ocean.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"math/rand"
 	"net/http"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -409,4 +410,22 @@ func hasAllTags(candidate, want []string) bool {
 		}
 	}
 	return true
+}
+
+// checkForRunningExperiments checks if there are any talis instances running.
+// It returns a list of all talis droplets.
+func checkForRunningExperiments(ctx context.Context, client *godo.Client) ([]godo.Droplet, error) {
+	droplets, err := listAllDroplets(ctx, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list droplets: %w", err)
+	}
+
+	var talisDroplets []godo.Droplet
+	for _, d := range droplets {
+		if slices.Contains(d.Tags, "talis") {
+			talisDroplets = append(talisDroplets, d)
+		}
+	}
+
+	return talisDroplets, nil
 }


rach-id:
@evan-forbes conflicts

evan-forbes:
> conflicts

totally, just was waiting to do anything to this PR as it needed to use official tags

test/util/testnode/query_sequence_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/celestiaorg/celestia-app/v6/test/util/testfactory"
+	blobtypes "github.com/celestiaorg/celestia-app/v6/x/blob/types"
+	"github.com/celestiaorg/go-square/v3/share"
+	abci "github.com/cometbft/cometbft/abci/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestQuerySequence(t *testing.T) {
```
evan-forbes, line 22: this seems like a weird place to keep this, however its actually needed so that we can get access to the unexported cometBFT node and query the sequence from that

If we're going to to an integration test with the mempool, then imo we need to do that.

tools/latency-monitor/main.go
```diff
@@ -227,25 +243,28 @@ func monitorLatency(
...
+				results = append(results, txResult{
+					submitTime: submitTime,
+					commitTime: commitTime,
+					latency:    latency,
+					txHash:     confirmed.TxHash,
+					code:       confirmed.Code,
+					height:     confirmed.Height,
+					failed:     false,
+					errorMsg:   "",
+				})
```
evan-forbes, line 264: this was just a bug with the latency monitor and how it counted stuff

tools/latency-monitor/main.go
```diff
@@ -79,14 +82,15 @@ between submission and commitment, providing detailed latency statistics.`,
...
+			return monitorLatency(ctx, endpoint, keyringDir, accountName, blobSize, minBlobSize, namespaceStr, disableMetrics, submissionDelay)
 		},
 	}
 
 	cmd.Flags().StringVarP(&endpoint, "grpc-endpoint", "e", defaultEndpoint, "gRPC endpoint to connect to")
 	cmd.Flags().StringVarP(&keyringDir, "keyring-dir", "k", defaultKeyringDir, "Directory containing the keyring")
 	cmd.Flags().StringVarP(&accountName, "account", "a", "", "Account name to use from keyring (defaults to first account)")
-	cmd.Flags().IntVarP(&blobSize, "blob-size", "b", defaultBlobSize, "Maximum size of blob data in bytes (actual size will be random up to this value)")
+	cmd.Flags().IntVarP(&blobSize, "blob-size", "b", defaultBlobSize, "Maximum size of blob data in bytes (actual size will be random between this value and the minimum)")
+	cmd.Flags().IntVarP(&minBlobSize, "blob-size-min", "z", defaultMinBlobSize, "Minimum size of blob data in bytes (actual size will be random between this value and the maximum)")
```
evan-forbes, line 93: could have probably isolated these changes, however they were required to test the new mempool so :shrug: 

just needed a way to have random blobs within a range

random blobs are actually needed for testing. previously non-random blobs didn't have the priority rounding issue. this ensure we're covering that while also keeping blobs big

app/app.go
```diff
@@ -196,6 +197,10 @@ type App struct {
 	// treePool used for ProcessProposal and PrepareProposal to optimize root calculation allocs
 	treePool                *wrapper.TreePool
 	delayedPrecommitTimeout time.Duration
+	// checkStateMu protects concurrent access to BaseApp's checkState (mempool state).
+	// This prevents data races between Commit updating checkState and QuerySequence
+	// reading it via CheckState().
+	checkStateMu *sync.RWMutex
```
evan-forbes, line 203: had to add this as we only exposed the state in the sdk, we didn't actually manage it and the mutex there is only held to return the struct

we should evaluate mutex usage in state more holistically, but keeping the mutex here is better anyway for now imo. the reason being that we can see when its held, and keep it mempool specific

longer term we could even only rely on this mutex for the mempool state, which would allow us to remove the ABCI mutex

multiplexer/abci/abci.go
```diff
@@ -181,6 +181,24 @@ func (m *Multiplexer) VerifyVoteExtension(_ context.Context, req *abci.RequestVe
 	return app.VerifyVoteExtension(req)
 }
 
+func (m *Multiplexer) QuerySequence(ctx context.Context, req *abci.RequestQuerySequence) (*abci.ResponseQuerySequence, error) {
+	// Return noop (sequence 0) for app version <= 4
+	if m.appVersion <= 4 {
+		return &abci.ResponseQuerySequence{Sequence: 0}, nil
+	}
```
evan-forbes, line 188: here's the defensive noop for not operating this method for every version of the application that doesn't need to run it.

we'd probably be fine anyway, but might as well add it

rootulp: Since the defensive loop applies to <= 4, I infer that this PR needs to be backported to v5.x.

multiplexer/abci/abci.go
```diff
@@ -181,6 +181,24 @@ func (m *Multiplexer) VerifyVoteExtension(_ context.Context, req *abci.RequestVe
...
+
+	app, err := m.getApp()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get app for version %d: %w", m.appVersion, err)
+	}
+	resp, err := app.QuerySequence(ctx, req)
+	if err != nil {
+		// If the app doesn't implement QuerySequence, return a noop response
+		return &abci.ResponseQuerySequence{Sequence: 0}, nil
+	}
```
rootulp: [question] here we're silently discarding the error which may make debugging more difficult. Why not propagate the error?

test/util/testnode/query_sequence_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	enc := encoding.MakeConfig(app.ModuleEncodingRegisters...)
+	blobGenState := blobtypes.DefaultGenesis()
+	blobGenState.Params.GovMaxSquareSize = uint64(appconsts.SquareSizeUpperBound)
+
+	// Create a config with a longer block time (3 seconds) to allow time to query
+	// the sequence before the transaction is included in a block
+	cfg := DefaultConfig().
+		WithFundedAccounts(s.accounts...).
+		WithModifiers(genesis.SetBlobParams(enc.Codec, blobGenState.Params))
```
rootulp: I may be missing something but the comment doesn't describe the code. It looks like the block time is the default b/c it's not overriden

multiplexer/abci/remote_v1.go
```diff
@@ -470,6 +470,13 @@ func (a *RemoteABCIClientV1) ExtendVote(ctx context.Context, req *abciv2.Request
 	return &abciv2.ResponseExtendVote{}, nil
 }
 
+// QuerySequence implements abciv2.ABCI
+func (a *RemoteABCIClientV1) QuerySequence(_ context.Context, _ *abciv2.RequestQuerySequence) (*abciv2.ResponseQuerySequence, error) {
+	return &abciv2.ResponseQuerySequence{
+		Sequence: 0,
+	}, nil
```
rootulp: [question] will these lines ever be invoked? I assume no because this conditional

```
	// Return noop (sequence 0) for app version <= 4
	if m.appVersion <= 4 {
		return &abci.ResponseQuerySequence{Sequence: 0}, nil
	}
```

in which case, I guess this could return an error instead:

```suggestion
	return nil, fmt.Errorf("RemoteABCIClientV1 does not support QuerySequence")
```

not strictly necessary, I just think errors may help us find issues faster than returning a no-op `0`

rootulp (dismissed):

multiplexer/abci/abci.go
evan-forbes: > Since the defensive loop applies to <= 4, I infer that this PR needs to be backported to v5.x.

yeah more or less. we ofc first need a version of the sdk that supports this which is what I'm working on now, but its just taking while since we haven't updated that branch in while

test/util/testnode/query_sequence_test.go
evan-forbes: ahh good catch I forgot I updated this during a refactor. Its not the end of the world and passes still since the checkTxstate will get updated upon committing a new block, but ideally we query between blocks

https://github.com/celestiaorg/celestia-app/pull/6040/commits/76b1d7e8652a539cc20a350bbdbac72da6cd5962

multiplexer/abci/remote_v1.go
evan-forbes: hmmm yeah good point https://github.com/celestiaorg/celestia-app/pull/6040/commits/3a06c55396731714f6590ac32214930131b794fe

multiplexer/abci/abci.go
evan-forbes: this is a good point and I tried changing it, but this causes a lot of tests to fail in testnode as this method is getting called somehwere before

I didn't feel like refactoring testnode just to handle this so I reverted it. happy to open a flup though

https://github.com/celestiaorg/celestia-app/pull/6040/commits/0dbe8fc7765a8eaead2a617c03d225fbb10239d4

ninabarbakadze (approved):
lgtm

PR #6064, opened 2025-10-20
chore: add the CPU requirements benchmark docker image

rach-id:
## Overview

Adds a docker image for the benchmarking tool so it's easy for validators to run


evan-forbes (approved):

rach-id:
> Or just manually push a Docker image if we don't expect it to change.

Already pushed the image. It's not something that we want to change or maintain, so it's fine to be without CI

PR #6077, opened 2025-10-24
test: Improve PrepareProposalConsistency

rawbytedev:
Added more checks to `PrepareProposalConsistency`:
- All sendTxs must be included in block
- No unknow Txs should be added to block
- Txs in Block must come from the generated Txs set

I'm still thinking of way to solve #2519 
all I could do is propose so improvement for it

@evan-forbes Thanks for previous review and honest opinion, I'll avoid doing same mistakes in future PR

### DIFF
--- a/app/test/fuzz_abci_test.go
+++ b/app/test/fuzz_abci_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/celestiaorg/celestia-app/v6/pkg/user"
 	testutil "github.com/celestiaorg/celestia-app/v6/test/util"
 	"github.com/celestiaorg/celestia-app/v6/test/util/random"
+	"github.com/cometbft/cometbft/crypto/tmhash"
 	"github.com/celestiaorg/go-square/v3/share"
 	abci "github.com/cometbft/cometbft/abci/types"
 	coretypes "github.com/cometbft/cometbft/types"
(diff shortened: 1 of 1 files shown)


app/test/fuzz_abci_test.go
```diff
@@ -133,6 +134,23 @@ func TestPrepareProposalConsistency(t *testing.T) {
...
+						require.Contains(t, respSet, k, "missing SendTxs from PrepareProposal")
+					}
+					// check that no extras txs were added to proposal
+					for _, txs := range resp.Txs {
+						h := tmhash.Sum(txs)
+						var k [32]byte
+						copy(k[:], h)
+						require.Contains(t, txsSet, k, "proposal contains unknown tx")
+					}
+
```
evan-forbes, line 153: why add this though

this protects against nothing in this test afaiu

evan-forbes:
closing as I see this as doa and don't want to waste other reviewers time

thanks for continuing to think about #2519

maybe try a different issue. the previous ai bot failed at https://github.com/celestiaorg/celestia-app/pull/6076 maybe you could improve that via testing it and minimizing the info needed in that document

app/test/fuzz_abci_test.go
rawbytedev: @evan-forbes 
First loop (sendTx inclusion): Previously, the test only checked that the total number of transactions in the proposal was greater than the number of send transactions. This indirectly suggested that at least one blob was included. That was not a strong assurance. Now, the loop clearly confirms that every send transaction submitted is included in the proposal. This requirement is straightforward: no send transaction can be dropped without notice.

Second loop (no unknowns): This loop makes sure that every transaction in the proposal originates from the original input set. In other words, PrepareProposal cannot create or change transactions; the proposal must include only a subset of what was provided. This prevents the accidental or intentional addition of "unknown" transactions. 

Loop 1: Guarantees that all send transactions remain in the proposal.
Loop 2: Guarantees that the proposal contains only transactions from the input set, stopping unknown or fabricated transactions.

PR #6078, opened 2025-10-26
feat: Add generic load-bytes support to txsim

radik878:
## Overview

This PR introduces a generic load capability to txsim that allows users to generate a target amount of network load by specifying a total blob byte budget. The core now tracks submitted blob payload bytes across all sequences and cancels execution once the configured threshold is reached. The CLI exposes this via a new --load-bytes flag and enforces its use alongside --blob, as the metric applies to blob payloads. This provides a simple, repeatable way to create transaction load without scripting multiple runs and aligns with the linked issue’s proposal to simplify tx load generation.

### DIFF
--- a/test/cmd/txsim/cli.go
+++ b/test/cmd/txsim/cli.go
@@ -47,6 +47,7 @@ var (
 	gasLimit                                          uint64
 	gasPrice                                          float64
 	namespaces                                        []string
+	loadBytes                                         int64
 )
 
 func main() {
@@ -110,6 +111,10 @@ account that can act as the master account. The command runs until all sequences
 
 			if stake == 0 && send == 0 && blob == 0 && upgradeSchedule == "" {
 				return errors.New("no sequences specified. Use --stake, --send, --upgrade-schedule or --blob")
+			}
+
+			if loadBytes > 0 && blob == 0 {
+				return errors.New("--load-bytes requires --blob to be set (byte counting applies to blobs)")
 			}
 
 			// setup the sequences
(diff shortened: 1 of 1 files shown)


rootulp:
> This provides a simple, repeatable way to create transaction load without scripting multiple runs and aligns with the linked issue’s proposal to simplify tx load generation.

Where is the linked issue? Can you please include that in the PR description?

test/txsim/run.go
```diff
@@ -29,6 +30,14 @@ const (
 
 var defaultTLSConfig = &tls.Config{InsecureSkipVerify: true}
 
+func opBytes(op Operation) int64 {
+	var n int64
+	for _, b := range op.Blobs {
+		n += int64(len(b.Data()))
+	}
+	return n
+}
```
rootulp, line 39: [blocking] IMO this should be a method on `Operation`.

1. Please use descriptive variables names. Concretely: `n` => `totalBytes`

test/txsim/run.go
```diff
@@ -60,8 +69,11 @@ func Run(
 		zerolog.SetGlobalLevel(zerolog.Disabled)
 	}
 
+	ctx2, cancel := context.WithCancel(ctx)
```
rootulp, line 72: [blocking] why does this need to be ctx2? How does that differ from ctx? When should developers use ctx vs ctx2 after this line?

test/txsim/run.go
```diff
@@ -96,17 +109,33 @@ func Run(
...
 
 				// Submit the messages to the chain.
-				if err := manager.Submit(ctx, ops); err != nil {
+				if err := manager.Submit(ctx2, ops); err != nil {
 					errCh <- fmt.Errorf("sequence %d: %w", seqID, err)
 					return
 				}
+
+				if opts.byteLimit > 0 {
+					sz := opBytes(ops)
```
rootulp, line 129: [blocking] please use descriptive variable names.

evan-forbes:
no drive-by @radik878 see contribution guidelines

this feature doesn't make a ton of sense and at first glance looks like its not doing what you think its doing

imo this is at the very bottom of priorities for reviewers

given the huge amount of spam we get we can just close this

Issue #6086, opened 2025-10-27
Include documentation to not use and migrate away from Horcrux


evan-forbes:
Horcrux is no longer maintained and it double signed with the bug

evan-forbes:
closing in favor of #6087

evan-forbes:
I think we can update this slightly to simply state that folks need to avoid distributing their signers and if there are other configs needed to avoid double signing then they need to use that.

it is still no longer officially maintained but seems fine for now, generally use at your own risk. it does offer functionality that simply isn't offered elsewhere so if users do require this functionality then we'll need to support.

PR #6088, opened 2025-10-28
chore(docs): document kms disclaimer

rach-id:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/6086#issuecomment-3456165957


rach-id:
Once we agree on the wording here, I will add the same disclaimer in our fork of horcrux

rach-id:
Also related and would use a review: https://github.com/celestiaorg/horcrux/pull/6

PR #6090, opened 2025-10-28
fix: pass the sequence and address back from checktx

evan-forbes:
@ninabarbakadze and I noticed that the txs were being passed to recheck out of order, and the reason why this occurs is that we aren't actually passing the tx sequence or signer, so the mempool doens't know how to organize them.

while gossiping, query sequence is actually checking this, however during recheck we're not.

I'm still not sure why its not a problem for larger txs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -277,7 +277,7 @@ require (
 replace (
 	cosmossdk.io/api => github.com/celestiaorg/cosmos-sdk/api v0.7.6
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
-	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v1.54.0-tm-v0.38.17.0.20251030040926-dba3fcd65678 // v0.39.10
+	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.12
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v0.51.4
 	github.com/cosmos/ibc-go/v8 => github.com/celestiaorg/ibc-go/v8 v8.7.2
 	// Use ledger-cosmos-go v0.16.0 because v0.15.0 causes "hidapi: unknown failure"
--- a/go.sum
+++ b/go.sum
@@ -781,8 +781,8 @@ github.com/bytedance/sonic v1.14.1/go.mod h1:gi6uhQLMbTdeP0muCnrjHLeCUPyb70ujhnN
 github.com/bytedance/sonic/loader v0.3.0 h1:dskwH8edlzNMctoruo8FPTJDF3vLtDT0sXZwvZJyqeA=
 github.com/bytedance/sonic/loader v0.3.0/go.mod h1:N8A3vUdtUebEY2/VQC0MyhYeKUFosQU6FxH2JmUe6VI=
 github.com/casbin/casbin/v2 v2.1.2/go.mod h1:YcPU1XXisHhLzuxH9coDNf2FbKpjGlbCg3n9yuLkIJQ=
-github.com/celestiaorg/celestia-core v1.54.0-tm-v0.38.17.0.20251030040926-dba3fcd65678 h1:ValdxUc8MgC5Hksah7FYkJXk3HPlTR1biiPrtlUElLw=
-github.com/celestiaorg/celestia-core v1.54.0-tm-v0.38.17.0.20251030040926-dba3fcd65678/go.mod h1:GxCnrVUUJZ3dtpNlKfHw3otb5TqyOdwjXef1/pNqSzA=
+github.com/celestiaorg/celestia-core v0.39.12 h1:SvlGmZr6XmsSeruFDSyz2rQmCdAJT+6JoPZwNDPpOhU=
+github.com/celestiaorg/celestia-core v0.39.12/go.mod h1:GxCnrVUUJZ3dtpNlKfHw3otb5TqyOdwjXef1/pNqSzA=
 github.com/celestiaorg/celestia-core v1.55.0-tm-v0.34.35 h1:FREwqZwPvYsodr1AqqEIyW+VsBnwTzJNtC6NFdZX8rs=
 github.com/celestiaorg/celestia-core v1.55.0-tm-v0.34.35/go.mod h1:SI38xqZZ4ccoAxszUJqsJ/a5rOkzQRijzHQQlLKkyUc=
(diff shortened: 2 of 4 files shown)


app/check_tx.go
```diff
@@ -22,53 +24,71 @@ func (app *App) CheckTx(req *abci.RequestCheckTx) (*abci.ResponseCheckTx, error)
...
-			return responseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false), err
-		}
-		// reject transactions that have a MsgPFB but no blobs attached to the tx
-		for _, msg := range sdkTx.GetMsgs() {
-			if _, ok := msg.(*blobtypes.MsgPayForBlobs); !ok {
-				continue
-			}
+	if isBlob {
+		return app.handleBlobCheckTx(req, btx)
+	}
```
evan-forbes, line 38: it was possible to add the necessary change without refactoring like this but it is max slop since we need to perform the same handling repeatedly

pkg/user/tx_client_test.go
```diff
@@ -335,9 +337,8 @@ func TestEvictions(t *testing.T) {
 		t.Skip("skipping evictions test in short mode")
 	}
 
-	ttlNumBlocks := int64(1)
-	blocksize := int64(1048576) // 1 MiB
-	_, txClient, ctx := setupTxClient(t, ttlNumBlocks, blocksize)
+	blocksize := int64(1024 * 1024 * 2) // 2 MiB
+	_, txClient, ctx := setupTxClient(t, 1, 64, blocksize)
```
evan-forbes, line 341: see #6091 for more details

using a single MB block limit but a larger square limit results in a chain halt. 

The previous test never hit this due to the mempool not working and only getting a single tx at a time. TxClient will try keep retrying over and over until its gets included.

@ninabarbakadze is the txclient supposed to loop continuously? I forget the answer here as iirc you mentioned it shouldn't be doing that. Is this only for evictions?

pkg/user/tx_client_test.go
```diff
@@ -594,7 +604,7 @@ func setupTxClient(
 }
 
 func setupTxClientWithDefaultParams(t *testing.T, opts ...user.Option) (encoding.Config, *user.TxClient, testnode.Context) {
-	return setupTxClient(t, 0, 8388608, opts...) // no ttl and 8MiB block size
+	return setupTxClient(t, 0, 128, 8388608, opts...) // no ttl and 8MiB block size
```
evan-forbes, line 605: if we want or need a default config for tests here then imo we should just have a function that returns that and not have this function

app/check_tx.go
```diff
@@ -82,3 +102,25 @@ func responseCheckTxWithEvents(err error, gw, gu uint64, events []abci.Event, de
...
+
+	if len(sigs) == 0 {
+		return nil, 0, fmt.Errorf("tx of type %T contains no signatures", tx)
+	}
+
+	if sigs[0].PubKey == nil {
+		return nil, 0, fmt.Errorf("tx signer %d has no associated pubkey", 0)
+	}
+
+	return sigs[0].PubKey.Address().Bytes(), sigs[0].Sequence, nil
```
cmwaters, line 125: FYI my implementation used

```
signers, err := sigTx.GetSigners()
if err != nil {
	return nil, 0, err
}
return signers[0]
```

Not sure if there's a difference in formatting but I don't think it matters

app/check_tx.go
```diff
@@ -82,3 +102,25 @@ func responseCheckTxWithEvents(err error, gw, gu uint64, events []abci.Event, de
...
+		return nil, 0, err
+	}
+
+	if len(sigs) == 0 {
+		return nil, 0, fmt.Errorf("tx of type %T contains no signatures", tx)
+	}
+
+	if sigs[0].PubKey == nil {
+		return nil, 0, fmt.Errorf("tx signer %d has no associated pubkey", 0)
+	}
```
cmwaters, line 123: Do we require that all transactions have pub keys now?

app/check_tx.go
```diff
@@ -22,53 +24,71 @@ func (app *App) CheckTx(req *abci.RequestCheckTx) (*abci.ResponseCheckTx, error)
...
+
+	return app.forwardCheckTx(baseReq, sdkTx)
+}
+
+func (app *App) forwardCheckTx(req *abci.RequestCheckTx, sdkTx sdk.Tx) (*abci.ResponseCheckTx, error) {
+	signerAddr, signerSeq, err := signerDataFromTx(sdkTx)
+	if err != nil {
+		return responseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false), err
+	}
```
cmwaters: It's probably safer to have this after CheckTx so we don't have to worry about all the checks

pkg/user/tx_client_test.go
ninabarbakadze: During evictions we won’t loop nonstop. We’ll resubmit the tx bytes and if it fails we’ll wait for confirmation. If it gets evicted again, we’ll keep resubmitting until it finally goes through.

If we’re also submitting other transactions on top the evicted one will likely hit a sequence mismatch during resubmission. In that case we’ll just wait - either it gets confirmed or we can tell the user to resign. We might need to extend the confirmation window if we actually want to wait until the block TTL expires so we have that final rejection across all nodes.

pkg/user/tx_client_test.go
ninabarbakadze: we could use appconsts?

app/check_tx.go
evan-forbes: yeah this is actually an sdk v0.50 thing that super subtle and seemed unintentional.

pkg/user/tx_client_test.go
evan-forbes: > During evictions we won’t loop nonstop.

> If it gets evicted again, we’ll keep resubmitting until it finally goes through.

I think this is where my confusion stems from actually. this to me means it will loop indefinitely if we will keep resubmitting. or do you mind expanding on the difference?

app/check_tx.go
evan-forbes: https://github.com/celestiaorg/celestia-app/pull/6090/commits/fd9650ebb3b6e338e66c54b43c057ae1c83d9e16

evan-forbes: makes sense!

app/check_tx.go
evan-forbes: I was grabbing the sequence from here as well so figured might as well

that is cleaner tho

evan-forbes:
this is only blocked on creating a release then it should be good to go

ninabarbakadze (approved):

PR #6105, opened 2025-11-01
fix: use shared mebibyte constant in default overrides

prestoalvarez:
Replace the hard-coded `mebibyte` literal in `app/default_overrides.go` with a value derived from `appconsts.DefaultUpperBoundMaxBytes`, keeping configuration limits in sync with global constants.


rach-id:
SERIOUSLY!

Issue #6114, opened 2025-10-30
Do a write up about extending TxStatus to include the Signer


ninabarbakadze:
## Description

Write reasoning around why I extended the TxStatus field to include the signer field.

ninabarbakadze:
Can also transfer this issue to app wasn't sure where to open it.

evan-forbes:
is there an outcome we want to record here or can we close?

ninabarbakadze:
I wrote a small ADR the other day will push now. We don't need to merge it if we don't find it useful.

rootulp:
I assume this is closed by https://github.com/celestiaorg/celestia-app/pull/6116

PR #6116, opened 2025-11-03
docs: write ADR about extending TxStatus with signers

ninabarbakadze:
## Overview

Fixes #https://github.com/celestiaorg/celestia-app/issues/6114

@evan-forbes if you don't think this belongs here or is useful we don't need to merge


docs/architecture/adr-024-extend-tx-status-with-signers.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Positive
+
+- **Persistent Access**: Signer info is now stored with transaction metadata, allowing repeated queries for the signers.
+- **Better API**: `TxStatus` provides all fields necessary to populate `TxResponse`
+
+### Negative
+
+- **Storage Overhead**: Each transaction result now stores signer information (typically 1-5 addresses per transaction)
+- **Tech Debt** - Introduces tech debt in the API
```
rootulp: [question] what is the tech debt?

[suggestion] another negative is perhaps that this bloats the diff between upstream and our forked repos (specifically cosmos-sdk and celestia-core). And more specifically that the diff includes changes to protos.

docs/architecture/adr-024-extend-tx-status-with-signers.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Testing
+
+- Unit and integration
```
rootulp: IMO this should be removed or expanded on. 


```suggestion
```

docs/architecture/adr-024-extend-tx-status-with-signers.md
```diff
@@ -0,0 +1,57 @@
...
+
+- 2025-10-30: Initial draft (@ninabarbakadze)
+
+## Status
+
+Implemented
+
+## Context
+
+The `TxClient` must expose which account signed each transaction once it’s confirmed onchain. With parallel tx submission, multiple worker accounts can sign transactions concurrently, so users can’t assume which account signed a given one. Applications need to track which worker was responsible, and users should be able to easily see the signing address.
```
rootulp: > multiple worker accounts can sign transactions concurrently

[question] why can _multiple_ accounts sign a transaction concurrently? Does this mean a transaction can be signed by two accounts before being submitted on-chain? I assume no because that doesn't make a lot of sense if the account isn't a multi-sig.

Perhaps this meant to say that _any_ worker account can sign a transaction.

rootulp (approved):

docs/architecture/adr-024-extend-tx-status-with-signers.md
ninabarbakadze: yeah good point about bloating the diff, will add. 

> what is the tech debt?

extending an API users depend on just adds complexity and more maintenance in case we do not need this change in the future.

docs/architecture/adr-024-extend-tx-status-with-signers.md
ninabarbakadze: yes this is referring to workers and parallel tx submission. I can reword it for clarity.

ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6116/commits/169a215598112c79b66b1a6ddcc1a1ffb3a4abc7

docs/architecture/adr-024-extend-tx-status-with-signers.md
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6116/commits/169a215598112c79b66b1a6ddcc1a1ffb3a4abc7

rootulp: > extending an API users depend on just adds complexity and more maintenance

I disagree. An API change can reduce complexity and maintenance. It's not clear from this ADR specifically what the tech debt is.

ninabarbakadze: removed it https://github.com/celestiaorg/celestia-app/pull/6116/commits/35892709917a39587f33abb88a61e60d4c3c54ab

PR #6138, opened 2025-11-04
feat: add --enable-legacy-block-prop flag for start command

### LINKED ISSUE #6129: Add a flag for enabling the legacy block prop
Currently, we can enable and disable the legacy block propagation mechanism using a config. We should add flag that allows setting that to make it easier for validators to do it.

This change also includes removing setting that flag to true in `update-config` command

forkfury:
## Summary

Adds a command-line flag `--enable-legacy-block-prop` to the start command, allowing validators to enable/disable legacy block propagation without editing config files.

## Changes

- Added `--enable-legacy-block-prop` flag to the start command
- Implemented pre-start hook to override config value when flag is set
- Removed automatic `enable_legacy_block_prop = true` setting from `update-config` command

Closes #6129


rach-id:
Sorry but I don't think we will go in this direction, so I'll close this PR.

Issue #6093, opened 2025-10-29
Support additional cloud providers in Talis


rach-id:
Now that we aim to run 128mb/6s on stronger hardware, we need to support a different cloud provider than DO that provides such machines to know exactly how fast the blocks are.

rach-id:
As per sync discussion with Evan, we should support AWS to also support upcoming products that require high bandwidth.

Issue #6100, opened 2025-10-31
Multiple talis experiments at the same time


rach-id:
Currently, only a single user can run a talis experiment per cloud provider. This is limiting because multiple people are using it.

in order to support multiple experiments at the same time, the counter that initializes the validator tags will need to be updated to list the servers that are running and only assign counters for available spots. This should happen when calling `init` and also when calling `up` with a confirmation that you want to update the tags.

devlongs:
i will be happy to work on this

PR #6144, opened 2025-11-07
feat(tools): talis support google cloud provider and multiple experiments at the same time

rach-id:
## Overview

Closes https://github.com/celestiaorg/celestia-app/issues/6093 and https://github.com/celestiaorg/celestia-app/issues/6100

### DIFF
--- a/docs/release-notes/release-notes.md
+++ b/docs/release-notes/release-notes.md
@@ -24,6 +24,13 @@ Validators should build binaries in freshly created environments or download pre
 
 This release introduces a new message type that needs to be signed by KMS. **If you use Horcrux or TmKMS, you must use these versions in order to participate in consensus. If you use an alternative KMS, please reach out.**
 
+**Disclaimer:** (KMS) are third-party software. Validators are responsible for ensuring their own KMS setup is correctly configured.
+An incorrect setup may result in double signing, which can lead to slashing. So, any error log should be thoroughly investigated, even if the KMS appears to be signing normally.
+
+Validators must also ensure they maintain a fast and reliable setup. Future network upgrades will require low signing latency, so server colocation can be explored to achieve faster signatures.
+
+**Validators choosing to run KMS do so at their own risk.**
+
 #### Horcrux
 
 For horcrux, use [v3.3.3-celestia](https://github.com/celestiaorg/horcrux/releases/tag/v3.3.3-celestia). All the setups and configs remain the same.


tools/talis/config.go
```diff
@@ -119,13 +146,10 @@ type Config struct {
 	SSHKeyName string `json:"ssh_key_name"`
 	// DigitalOceanToken is used to authenticate with DigitalOcean. It can be
 	// provided via an env var or flag.
-	DigitalOceanToken string `json:"digitalocean_token,omitempty"`
-	// LinodeToken is used to authenticate with Linode. It can be provided via
-	// an env var or flag.
-	LinodeToken string `json:"linode_token,omitempty"`
```
rach-id, line 125: removed linode as no implementation was there, just couple variables here and there

tools/talis/deployment.go
```diff
@@ -433,71 +436,114 @@ func listCmd() *cobra.Command {
...
+
+func checkForRunningExperiments(ctx context.Context, cfg Config) error {
+	var hasRunningExperiments bool
+
+	if cfg.DigitalOceanToken != "" {
+		tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: cfg.DigitalOceanToken})
+		doClient := godo.NewClient(oauth2.NewClient(ctx, tokenSource))
+		running, err := checkForRunningDOExperiments(ctx, doClient, cfg.Experiment, cfg.ChainID)
+		if err != nil {
+			log.Printf("⚠️  Warning: failed to check DigitalOcean for running experiments: %v", err)
```
rach-id, line 468: kept the emojis to stay consistent with the rest of talis code :D

tools/talis/google_cloud.go
```diff
@@ -0,0 +1,757 @@
...
+
+func RandomGCZone(region string) string {
+	zones, ok := GCZones[region]
+	if !ok || len(zones) == 0 {
+		return region + "-a"
+	}
+	return zones[rand.Intn(len(zones))]
+}
+
+func ensureGCFirewallRule(ctx context.Context, project string, opts []option.ClientOption) error {
```
rach-id, line 238: GC requires creating a firewall rule otherwise the servers will have all traffic blocked

tools/talis/add.go
```diff
@@ -24,10 +25,21 @@ func addCmd() *cobra.Command {
...
 			case "validator":
 				for i := 0; i < count; i++ {
-					cfg = cfg.WithDigitalOceanValidator(region)
+					switch provider {
+					case "digitalocean":
+						cfg = cfg.WithDigitalOceanValidator(region)
+					case "googlecloud":
+						cfg = cfg.WithGoogleCloudValidator(region)
+					default:
+						return fmt.Errorf("unknown provider %q (supported: digitalocean, googlecloud)", provider)
```
rootulp, line 41: [FLUP] do we need a FLUP to add AWS or is that no longer necessary?

Ref: https://github.com/celestiaorg/celestia-app/issues/6093#issuecomment-3461596125

tools/talis/README.md
```diff
@@ -91,6 +109,7 @@ the celestia-app configs (config.toml and app.toml) can be manually edited here,
 
 Notes:
 
+- **Only use one cloud provider per experiment.** Fill out either DigitalOcean or Google Cloud fields, not both. Filling them both might end up ruining other experiments or having stuck experiments that need to be removed by hand.
```
rootulp, line 112: [extremely optional FLUP] we could add a check at the beginning of all Talis commands to see if the user specified both DO and Google Cloud. If the check finds both configured, it stops immediately and informs the user.

That way we can avoid a foot-gun where talis accidentally ruins other experiments.

rootulp (approved):
LGTM 🚢

tools/talis/add.go
rach-id: AWS said that they don't like our workloads. Google cloud is enough and will cover our upcoming testing with very high throuput and strong machines. So we should be good

tools/talis/README.md
rach-id: Its a good FLUP. I dont think we need it immediately, so Ill create an issue. This is because everyone will use DO. Google cloud will only used it needed

rach-id: https://github.com/celestiaorg/celestia-app/issues/6146

PR #6152, opened 2025-11-10
chore(docs): refer to the official tmkms release in release notes

rach-id:
## Overview

we can use the official release of tmkms


rach-id:
This PR is related and can also be reviewed: https://github.com/celestiaorg/tmkms/pull/1

I will archive that repo once that PR is merged

PR #6167, opened 2025-11-10
test(docker-e2e): PFM Multihop e2e test

### LINKED ISSUE #5988: PFM e2e test
## Context

We previously had a PFM e2e test that used interchaintest

## Problem

Interchaintest got deprecated and wasn't easy to maintain so those tests were deleted. Now we currently don't have an e2e test that covers PFM functionality.

## Proposal

Consider adding an e2e test that uses tastora to verify PFM works as expected.

mojtaba-esk:
Closes #5988

### DIFF
--- a/test/docker-e2e/e2e_pfm_test.go
+++ b/test/docker-e2e/e2e_pfm_test.go
@@ -97,19 +97,19 @@ func (s *PFMTestSuite) TestPFMMultiHop() {
 	// Create PFM memo: chain-a sends to chain-b, which forwards to chain-c
 	memoJSON := makePFMMemo(addrC.String(), s.chBToC.PortID, s.chBToC.ChannelID, nil)
 
-	initialA := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
+	initialChainABalance := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
 	denomOnB := calculateIBCDenomTrace(
 		baseDenom,
 		hop{port: s.chAToB.CounterpartyPort, channelID: s.chAToB.CounterpartyID},
 	)
-	initialBonB := s.getBalance(ctx, s.chainB, walletB.GetFormattedAddress(), denomOnB)
+	initialChainBRelayBalance := s.getBalance(ctx, s.chainB, walletB.GetFormattedAddress(), denomOnB)
 
 	denomOnC := calculateIBCDenomTrace(
 		baseDenom,
 		hop{port: s.chAToB.CounterpartyPort, channelID: s.chAToB.CounterpartyID},
 		hop{port: s.chBToC.CounterpartyPort, channelID: s.chBToC.CounterpartyID},
 	)
-	initialConC := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), denomOnC)
+	initialChainCRecipientBalance := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), denomOnC)
 
 	amt := sdkmath.NewInt(sendAmount)
 	msg := ibctransfertypes.NewMsgTransfer(
@@ -129,12 +129,12 @@ func (s *PFMTestSuite) TestPFMMultiHop() {
 	s.Require().Equal(uint32(0), resp.Code, "transfer failed, code=%d", resp.Code)
 	s.T().Logf("Transfer submitted successfully, tx hash: %s, height: %d", resp.TxHash, resp.Height)
 
-	expectedAAfter := initialA.Sub(amt).SubRaw(feePaid)
+	expectedChainABalance := initialChainABalance.Sub(amt).SubRaw(feePaid)
 	s.T().Logf("Waiting for PFM multi-hop transfer to complete...")
 
 	err = wait.ForCondition(ctx, 3*time.Minute, 10*time.Second, func() (bool, error) {
-		finalConC := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), denomOnC)
-		delta := finalConC.Sub(initialConC)
+		finalChainCRecipientBalance := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), denomOnC)
+		delta := finalChainCRecipientBalance.Sub(initialChainCRecipientBalance)
 		if !delta.Equal(amt) {
 			s.T().Logf("Chain C has not received tokens yet (got %s, expected %s)", delta.String(), amt.String())
 			return false, nil
@@ -145,11 +145,11 @@ func (s *PFMTestSuite) TestPFMMultiHop() {
 
 	s.Require().NoError(err, "PFM multi-hop transfer failed")
 
-	finalBonB := s.getBalance(ctx, s.chainB, walletB.GetFormattedAddress(), denomOnB)
-	s.Require().True(finalBonB.Equal(initialBonB), "chain-b retained funds: before=%s after=%s", initialBonB.String(), finalBonB.String())
+	finalChainBRelayBalance := s.getBalance(ctx, s.chainB, walletB.GetFormattedAddress(), denomOnB)
+	s.Require().True(finalChainBRelayBalance.Equal(initialChainBRelayBalance), "chain-b retained funds: before=%s after=%s", initialChainBRelayBalance.String(), finalChainBRelayBalance.String())
 
-	finalA := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
-	s.Require().True(expectedAAfter.Equal(finalA), "chain-a balance mismatch: expected %s got %s", expectedAAfter.String(), finalA.String())
+	finalChainABalance := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
+	s.Require().True(expectedChainABalance.Equal(finalChainABalance), "chain-a balance mismatch: expected %s got %s", expectedChainABalance.String(), finalChainABalance.String())
 
 	s.T().Logf("Verifying PFM used two-hop path (not direct A->C)")
 
@@ -158,15 +158,15 @@ func (s *PFMTestSuite) TestPFMMultiHop() {
 		hop{port: s.chAToB.CounterpartyPort, channelID: s.chAToB.CounterpartyID},
 		hop{port: s.chBToC.CounterpartyPort, channelID: s.chBToC.CounterpartyID},
 	)
-	twoHopBalance := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), twoHopDenom)
-	s.T().Logf("Two-hop path (A->B->C) balance: %s %s", twoHopBalance.String(), twoHopDenom)
+	finalTwoHopRecipientBalance := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), twoHopDenom)
+	s.T().Logf("Two-hop path (A->B->C) balance: %s %s", finalTwoHopRecipientBalance.String(), twoHopDenom)
 
 	s.assertOnlyTwoHopDenom(ctx, s.chainC, walletC.GetFormattedAddress(), twoHopDenom)
 
-	if twoHopBalance.Equal(sdkmath.NewInt(sendAmount)) {
+	if finalTwoHopRecipientBalance.Equal(sdkmath.NewInt(sendAmount)) {
 		s.T().Logf("PFM used two-hop path (A->B->C)")
 	} else {
-		s.T().Fatalf("expected %d via two-hop path, got %s", sendAmount, twoHopBalance.String())
+		s.T().Fatalf("expected %d via two-hop path, got %s", sendAmount, finalTwoHopRecipientBalance.String())
 	}
 }
 


test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+	memoJSON := makePFMMemo(addrC.String(), s.chBToC.PortID, s.chBToC.ChannelID, nil)
+
+	s.waitChannelOpen(ctx, s.chainA, s.chAToB.PortID, s.chAToB.ChannelID, s.chainB, s.chAToB.CounterpartyPort, s.chAToB.CounterpartyID)
+	s.waitChannelOpen(ctx, s.chainB, s.chBToC.PortID, s.chBToC.ChannelID, s.chainC, s.chBToC.CounterpartyPort, s.chBToC.CounterpartyID)
+
+	initialA := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
+	denomOnB := calculateIBCDenomTrace(
+		[]struct{ port, channel string }{
+			{s.chAToB.CounterpartyPort, s.chAToB.CounterpartyID},
+		},
```
chatton: this syntax is super difficult to read, I would suggest just pulling out a small named struct to group these fields.
```go
type Hop struct {
  port, channelID string
}
```

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+	s.hermes, err = relayer.NewHermes(ctx, s.client, t.Name(), s.network, 0, s.logger)
+	s.Require().NoError(err, "failed to create hermes")
+	err = s.hermes.Init(ctx, []tastoratypes.Chain{s.chainA, s.chainB, s.chainC})
+	s.Require().NoError(err, "failed to initialize hermes")
+
+	s.Require().NoError(s.hermes.CreateClients(ctx, s.chainA, s.chainB))
+	s.Require().NoError(s.hermes.CreateClients(ctx, s.chainB, s.chainC))
+
+	s.connAToB, s.chAToB = s.establishConnectionAndChannel(ctx, s.chainA, s.chainB)
+	s.connBToC, s.chBToC = s.establishConnectionAndChannel(ctx, s.chainB, s.chainC)
```
chatton, line 270: I think these should can be done in parallel, the handshake takes quite a bit of time so it might be nice to do chainA -> chainB and chainB -> chainC at the same time, similar to the chain build refactor above.

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+// setupPFMInfrastructure sets up three chains and IBC routing required for PFM.
+func (s *PFMTestSuite) setupPFMInfrastructure(ctx context.Context) {
+	t := s.T()
+
+	tag, err := dockerchain.GetCelestiaTagStrict()
+	s.Require().NoError(err, "failed to get celestia tag")
+
+	s.chainA = s.buildChain(ctx, "chain-a", tag)
+	s.chainB = s.buildChain(ctx, "chain-b", tag)
+	s.chainC = s.buildChain(ctx, "chain-c", tag)
```
chatton: we can also do these in parallel, something like

```diff
diff --git a/test/docker-e2e/e2e_pfm_test.go b/test/docker-e2e/e2e_pfm_test.go
index bd04ed77..1066cc13 100644
--- a/test/docker-e2e/e2e_pfm_test.go
+++ b/test/docker-e2e/e2e_pfm_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"golang.org/x/sync/errgroup"
 	"strings"
 	"testing"
 	"time"
@@ -246,9 +247,27 @@ func (s *PFMTestSuite) setupPFMInfrastructure(ctx context.Context) {
 	tag, err := dockerchain.GetCelestiaTagStrict()
 	s.Require().NoError(err, "failed to get celestia tag")
 
-	s.chainA = s.buildChain(ctx, "chain-a", tag)
-	s.chainB = s.buildChain(ctx, "chain-b", tag)
-	s.chainC = s.buildChain(ctx, "chain-c", tag)
+	var eg errgroup.Group
+
+	eg.Go(func() error {
+		var err error
+		s.chainA, err = s.buildChain(ctx, "chain-a", tag)
+		return err
+	})
+
+	eg.Go(func() error {
+		var err error
+		s.chainB, err = s.buildChain(ctx, "chain-b", tag)
+		return err
+	})
+
+	eg.Go(func() error {
+		var err error
+		s.chainC, err = s.buildChain(ctx, "chain-c", tag)
+		return err
+	})
+
+	s.Require().NoError(eg.Wait(), "failed to build chains")
 
 	s.hermes, err = relayer.NewHermes(ctx, s.client, t.Name(), s.network, 0, s.logger)
 	s.Require().NoError(err, "failed to create hermes")
@@ -268,7 +287,7 @@ func (s *PFMTestSuite) setupPFMInfrastructure(ctx context.Context) {
 }
 
 // buildChain creates and starts a single celestia-app chain with one validator.
-func (s *PFMTestSuite) buildChain(ctx context.Context, chainID, tag string) *cosmos.Chain {
+func (s *PFMTestSuite) buildChain(ctx context.Context, chainID, tag string) (*cosmos.Chain, error) {
 	cfg := &dockerchain.Config{
 		Config: &testnode.Config{
 			Genesis: genesis.NewDefaultGenesis().
@@ -290,10 +309,11 @@ func (s *PFMTestSuite) buildChain(ctx context.Context, chainID, tag string) *cos
 	}
 
 	chain, err := dockerchain.NewCelestiaChainBuilder(s.T(), cfg).Build(ctx)
-	s.Require().NoError(err, "failed to build chain %s", chainID)
-	s.Require().NoError(chain.Start(ctx), "failed to start chain %s", chainID)
+	if err != nil {
+		return nil, err
+	}
 
-	return chain
+	return chain, chain.Start(ctx)
 }
 
 // establishConnectionAndChannel creates a connection and ICS20 channel between two chains.

```

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+	s.Require().Equal(uint32(0), resp.Code, "transfer failed, code=%d", resp.Code)
+	s.T().Logf("Transfer submitted successfully, tx hash: %s, height: %d", resp.TxHash, resp.Height)
+
+	expectedAAfter := initialA.Sub(amt).SubRaw(feePaid)
+	s.T().Logf("Waiting for PFM multi-hop transfer to complete...")
+
+	err = wait.ForCondition(ctx, 3*time.Minute, 5*time.Second, func() (bool, error) {
+		if err := wait.ForBlocks(ctx, 5, s.chainA, s.chainB, s.chainC); err != nil {
+			return false, err
+		}
```
chatton: I don't think we need to do a wait for blocks inside a wait for condition, it shouldn't take longer than 3 mins for the transfer to complete.

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+		finalConC := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), denomOnC)
+		finalBonB := s.getBalance(ctx, s.chainB, walletB.GetFormattedAddress(), denomOnB)
+		finalA := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
+
+		conditionsMet := true
```
chatton: rather than having all assertions built into this one bool, it might be cleaner to have separate `wait.ForCondtion`s to assert the balance on each chain, we can have more descriptive error messages for each failure if we do it that way WDYT?

chatton: Or do one wait for the balance to reach the final chain, then do assertions on every chain in the middle

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+		Version:        ibctransfertypes.Version,
+	}
+
+	channel, err := s.hermes.CreateChannel(ctx, a, connection, channelOpts)
+	s.Require().NoError(err, "failed to create IBC channel")
+	return connection, channel
+}
+
+// waitChannelOpen waits for a channel to be in OPEN state on both ends.
+func (s *PFMTestSuite) waitChannelOpen(ctx context.Context, src *cosmos.Chain, srcPort, srcChan string, dst *cosmos.Chain, dstPort, dstChan string) {
```
chatton: is this required? I think when hermes creates the channel, it will be open when `hermes.CreateChannel` returns.

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+				return false, nil
+			}
+			return true, nil
+		}),
+		"channels failed to reach OPEN state",
+	)
+}
+
+// checkChannelState verifies a channel is in OPEN state.
+func (s *PFMTestSuite) checkChannelState(ctx context.Context, chain *cosmos.Chain, portID, channelID string) error {
```
chatton: we should be able to remove this also

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	if s.hermes != nil {
+		if err := s.hermes.Remove(ctx); err != nil {
+			s.T().Logf("Error stopping hermes: %v", err)
+		}
+	}
+}
+
+func (s *PFMTestSuite) getAllBalances(ctx context.Context, chain *cosmos.Chain, address string) sdk.Coins {
```
chatton, line 193: I think we can use the built in tastora fn to do this https://github.com/celestiaorg/tastora/blob/1a4acde331d5fb23bc9284b4a8afe79839868909/framework/testutil/query/query.go#L12

chatton (commented):
Looks good overall! Left some suggestions about some possible code we can remove and a few ways to speed up runtime.

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: Great idea, yes I think it can be done concurrently

mojtaba-esk: I tried to run the two handshakes in parallel, but the test failed and the reason is that both legs rely on Hermes using the same relayer account on chain-b. When the CLI fires create connection transactions at once, the account sequence diverges and Hermes aborts.

chatton: ah I see that's annoying! I think we can still do it with the chain creation at least

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: This one should work. will do it. Nice idea BTW.

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: I faced multiple errors when did not have it, instead of time.sleep I added this. I can try again to see if we can safely remove this one.

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: Good idea when the final chain has the value, I assume the other must have been already updated so we can have a cleaner code.

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: sure

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	// Create PFM memo: chain-a sends to chain-b, which forwards to chain-c
+	memoJSON := makePFMMemo(addrC.String(), s.chBToC.PortID, s.chBToC.ChannelID, nil)
+
+	initialA := s.getBalance(ctx, s.chainA, walletA.GetFormattedAddress(), baseDenom)
+	denomOnB := calculateIBCDenomTrace(
+		baseDenom,
+		hop{port: s.chAToB.CounterpartyPort, channelID: s.chAToB.CounterpartyID},
+	)
+	initialBonB := s.getBalance(ctx, s.chainB, walletB.GetFormattedAddress(), denomOnB)
```
rootulp: [question] what does Bon mean?

test/docker-e2e/e2e_pfm_test.go
```diff
@@ -0,0 +1,350 @@
...
+	s.Require().NoError(err)
+	s.Require().Equal(uint32(0), resp.Code, "transfer failed, code=%d", resp.Code)
+	s.T().Logf("Transfer submitted successfully, tx hash: %s, height: %d", resp.TxHash, resp.Height)
+
+	expectedAAfter := initialA.Sub(amt).SubRaw(feePaid)
+	s.T().Logf("Waiting for PFM multi-hop transfer to complete...")
+
+	err = wait.ForCondition(ctx, 3*time.Minute, 10*time.Second, func() (bool, error) {
+		finalConC := s.getBalance(ctx, s.chainC, walletC.GetFormattedAddress(), denomOnC)
+		delta := finalConC.Sub(initialConC)
```
rootulp: What does ConC mean? Maybe it can be renamed `finalBalanceOnC`?

Same for other terms like BonB

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: Initial balance of wallet B on chain B (denom from chain B), so later we can compare it with the final balance to ensure everything worked correctly.

test/docker-e2e/e2e_pfm_test.go
mojtaba-esk: yeah we could rename it which will be more descriptive, but it will be a longer name

mojtaba-esk: done

PR #6168, opened 2025-11-10
feat: check-version CLI command to debug archive nodes

rootulp:
Tool to help debug https://github.com/celestiaorg/celestia-app/issues/5614

## Testing

On a local node with only 3 blocks

```shell
celestia-app main $   celestia-appd debug check-version 2748392
Latest version: 3
Checking version: 2748392

Error: height 2748392 is greater than latest version 3
```

- On a [local node](https://pastebin.com/bQQMLTsb).
- On an [Arabica archive node](https://pastebin.com/9nffYBJ8).

### DIFF
--- a/cmd/celestia-appd/cmd/check_version.go
+++ b/cmd/celestia-appd/cmd/check_version.go
@@ -9,7 +9,7 @@ import (
 	storetypes "cosmossdk.io/store/types"
 	circuittypes "cosmossdk.io/x/circuit/types"
 	evidencetypes "cosmossdk.io/x/evidence/types"
-	feegrant "cosmossdk.io/x/feegrant"
+	"cosmossdk.io/x/feegrant"
 	upgradetypes "cosmossdk.io/x/upgrade/types"
 	"github.com/celestiaorg/celestia-app/v6/app"
 	blobtypes "github.com/celestiaorg/celestia-app/v6/x/blob/types"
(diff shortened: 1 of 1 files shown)


cmd/celestia-appd/cmd/check_version.go
```diff
@@ -0,0 +1,239 @@
...
+
+			// Try to get commit info for this version
+			// Note: GetCommitInfo might not be available on the interface, we'll handle it gracefully
+			var commitInfo *storetypes.CommitInfo
+			if rootStore, ok := cms.(interface {
+				GetCommitInfo(int64) (*storetypes.CommitInfo, error)
+			}); ok {
+				commitInfo, err = rootStore.GetCommitInfo(height)
+				if err != nil {
+					fmt.Printf("⚠️  Warning: Could not get commit info for version %d: %v\n", height, err)
```
rach-id, line 107: emoji 🤔 

Image

rootulp:
> But I am wondering if we will still need this functionality after this bug is fixed? or we could just give the validator a commit to build and run?

QuickNode doesn't want to run an experimental release so if I can't debug it separately then ideally we merge this and include it in an official release.

Agreed long term though we don't really need to maintain this code so can hold off on merging for now. Or merge now, debug, then delete it later.

rach-id:
I'll defer to you and the other reviewers

cmd/celestia-appd/cmd/check_version.go
rootulp: Sorry. I revise my stance. Given AI generates code with emojis, it probably isn't the best use of our time to revise it to remove the emojis.

For clarity: most of this code was AI generated.

rach-id: Thats fine, for anything thats not going to production, emojis are fun IMO

Issue #6046, opened 2025-10-17
128MB/6s certified hardware


rach-id:
Anyone who runs the tool in https://github.com/celestiaorg/celestia-app/pull/6038, they can share the result they got so that we compile a list of hardware that's compatible and make the transition for validators easier.

rach-id:
From the discussion with POPS and devops team, we have the following who are good:
- amd 7900x
- amd 7950x
- amd 7700x

The following are not:
- older EPYCs with 16 cores are not viable

mindstyle85:
Ok:

AMD 9900x
AMD 9950x
AMD 7600
AMD EPYC 9454P
AMD EPYC 9124
AMD EPYC 9555P
AMD Ryzen 9 7950X3D
AMD EPYC 4344P
AMD EPYC 4464P
13th Gen Intel(R) Core(TM) i5-13500
AMD Ryzen 7 PRO 8700GE w/ Radeon 780M Graphics
AMD EPYC 4584PX

Not ok:
AMD Ryzen 9 5950X
Dual Intel(R) Xeon(R) Silver 4210
Dual Intel Xeon Platinum 8474C
AMD EPYC 7443P 24-Core

kinrokinro:
AMD EPYC 9454P:
```
=== Performance Test Results (128MB/6s Compatibility) ===

Prepare Proposal:
  Your time: 211 ms
  Reference time: 700 ms
  Status: FASTER than reference (3.3x faster)

Process Proposal:
  Your time: 434 ms
  Reference time: 700 ms
  Status: FASTER than reference (1.6x faster)

Finalize Block:
  Your time: 129 ms
  Reference time: 400 ms
  Status: FASTER than reference (3.1x faster)

Propose Block:
  Your time: 408 ms
  Reference time: 400 ms
  Status: SLOWER than reference (1.0x slower)

Encode Block:
  Your time: 278 ms
  Reference time: 400 ms
  Status: FASTER than reference (1.4x faster)

Decode Block:
  Your time: 343 ms
  Reference time: 500 ms
  Status: FASTER than reference (1.5x faster)

=== Final Assessment ===

WARNING: Your system does NOT meet the 128MB/6s upgrade requirements!

RECOMMENDATION:
To handle the 128MB/6s upgrade, you need to upgrade your hardware to:
  - 32 CPU cores (or more)
  - CPUs that support GFNI (Galois Field New Instructions)
  - CPUs that support SHA-NI (SHA New Instructions)

[...]

PR #6172, opened 2025-11-11
chore(docs): add the hardware requirements section to the release notes

rach-id:
## Overview

This is a clean list from https://github.com/celestiaorg/celestia-app/issues/6046

### DIFF
--- a/docs/release-notes/release-notes.md
+++ b/docs/release-notes/release-notes.md
@@ -17,11 +17,12 @@ After upgrading to 128mb/6s, specific validator hardware specifications are requ
 - AMD Ryzen 9 7950X3D
 - AMD Ryzen 9 9900X
 - AMD Ryzen 7 PRO 8700GE
+- AMD Ryzen 7 3700X
+- AMD Ryzen 3900X
 - AMD 7600
 - AMD 7700X
 - AMD 7900X
 - AMD Ryzen 7900
-- Ryzen 7 3700X
 - AMD EPYC 4344P
 - AMD EPYC 4464P
 - AMD EPYC 4545P
@@ -36,12 +37,7 @@ After upgrading to 128mb/6s, specific validator hardware specifications are requ
 - 13th Gen Intel Core i5-13500
 - Core i7-1270P
 - Intel Xeon-E 2386G
-</details>
 
-<details>
-  <summary>Preferable to switch to one of the supported CPUs:</summary>
-  - AMD Ryzen 7 3700X
-  - AMD Ryzen 3900X
 </details>
 
 <details>


docs/release-notes/release-notes.md
```diff
@@ -2,6 +2,61 @@
 
 This guide provides notes for major version releases. These notes may be helpful for users when upgrading from previous major versions.
 
+## v6 - 128mb/6s
+
+This section is specific to bumping from 32mb/6s in the v6 release to 128mb/6s. If you're upgrading from v5, please check the [v6.0.0](#v600) first. And only then check this one.
```
rootulp, line 7: [uber nit] we could place this at the bottom of the `##v6.0.0` heading and then get rid of this line


```suggestion
## v6 - 128mb/6s

This section is specific to bumping from 32mb/6s in the v6 release to 128mb/6s. 
```

docs/release-notes/release-notes.md
```diff
@@ -2,6 +2,61 @@
...
+- 13th Gen Intel Core i5-13500
+- Core i7-1270P
+- Intel Xeon-E 2386G
+</details>
+
+<details>
+  <summary>Preferable to switch to one of the supported CPUs:</summary>
+  - AMD Ryzen 7 3700X
+  - AMD Ryzen 3900X
+</details>
```
rootulp: [question] To clarify, these two CPUs are partially supported but encouraged to migrate to one of the supported CPUs instead?

If so, maybe we can phrase it like this:


```suggestion
<details>
  <summary>Partially supported CPUs but please switch to one of the supported CPUs above:</summary>
  - AMD Ryzen 7 3700X
  - AMD Ryzen 3900X
</details>
```

Or to simplify, remove this list and move these two into the unsupported category to make it extremely clear.

rach-id: 03b56b1f5afc8ec74f83fdfcc4499e1375d0d6d4

docs/release-notes/release-notes.md
rach-id: The reason I put it separate is because the v6 already has a hardware requirement section and didn't want to remove it or have 2 hardware reqs sections

rootulp: thanks for pointing that out. We could update the hardware requirements on docs.celestia.org and consolidate the sections  on this page b/c 128 MiB / 6 seconds is approaching fast and it would be cumbersome for node operators to change hardware for 32 MiB / 6 seconds and then change again for 128 MiB / 6 seconds shortly after.

rach-id: The problem is I'm afraid to confuse them with multiple requirements... Also, most of them don't need to bump, so it shouldn't take much time.

I'm planning to merge this once the v6 releases passes on mainnet to avoid giving them multiple things to think about at the same time.

rootulp: > I'm afraid to confuse them with multiple requirements... 

I agree, that's why I think we should get rid of any requirements that are only true for 32 MiB / 6 seconds and consolidate the requirements in these release notes (and docs.celestia.org) to apply to 128 MiB / 6 seconds.

> I'm planning to merge this once the v6 releases passes on mainnet to avoid giving them multiple things to think about at the same time.

I don't expect node operators to re-read the v6 release notes after v6 activates.

Defer to you on this.

rach-id: > I don't expect node operators to re-read the v6 release notes after v6 activates.

100%, that's why I put it in a separate section. Anyway, I'm fine with consolidating the two, we will generally post on discord and tag the validators.

Issue #6175, opened 2025-11-12
`BlockedAddresses` are incorrect


rootulp:
## Context

https://github.com/celestiaorg/celestia-app/pull/6174

## Problem

The `BlockedAddresses` implementation is flawed. It gets module account addresses but treats them like account names.

https://github.com/celestiaorg/celestia-app/blob/013eb80f31c3751842895fd64403f95f45930bcb/app/app.go#L650-L651

The `BlockedAddresses` are used by the bank keeper to prevent sending tokens to the blocked addresses.

## Proposal

1. Fix the implementation
2. Add unit tests

Note: this change is breaking so can't be included until v7.

PR #6176, opened 2025-11-12
fix!: blocked addresses

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/6175

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -645,6 +645,11 @@ func (app *App) ModuleAccountAddrs() map[string]bool {
 }
 
 // BlockedAddresses returns a map of the app's blocked account addresses.
+// NOTE: This blocklist defines the set of addresses that are not allowed
+// to receive funds through direct and explicit actions, for example, by using a MsgSend or
+// by using a SendCoinsFromModuleToAccount execution.
+// Thus, a module account such as the Hyperlane module account can still receive escrow funds via the protocol,
+// however sending directly to the module account via x/bank is forbidden.
 func (app *App) BlockedAddresses() map[string]bool {
 	modAccAddrs := make(map[string]bool)
 	for address := range app.ModuleAccountAddrs() {


damiannolan:
Looks like maybe [@galaparm](https://github.com/galaparm) should be added as co-author on this PR?

rootulp:
> Great catch, did this cause an issue with these invariants?

AFAICT it hasn't caused any real issues so this can wait until the v7.x.x release

> Looks like maybe [@galaparm](https://github.com/galaparm) should be added as co-author on this PR?

Can do

rootulp:
@galaparm if you want to be added as a co-author, I need your GitHub-verified email.

galaparm:
@rootulp I'd really appreciate being added as a co-author. My verified email is: 
173857417+galaparm@users.noreply.github.com

Thanks a lot!

damiannolan (approved):

app/app_test.go
```diff
@@ -100,6 +102,59 @@ func TestInitChain(t *testing.T) {
...
+			"fee_collector",
+			"distribution",
+			"mint",
+			"bonded_tokens_pool",
+			"not_bonded_tokens_pool",
+			"transfer",
+			"interchainaccounts",
+			"hyperlane",
+			"warp",
+			"burn",
```
rootulp: FYI we may have to remove this if we remove the x/burn module

cc: @Manav-Aggarwal

app/app_test.go
```diff
@@ -100,6 +102,59 @@ func TestInitChain(t *testing.T) {
...
+		"distribution",
+		"gov",
+		"mint",
+		"bonded_tokens_pool",
+		"not_bonded_tokens_pool",
+		"transfer",
+		"interchainaccounts",
+		"hyperlane",
+		"warp",
+		"burn",
```
rootulp: FYI we may have to remove this if we remove the x/burn module

cc: @Manav-Aggarwal

app/app_test.go
Manav-Aggarwal: Thanks for the heads up. We should remove this since we do plan on removing the x/burn module

rootulp: Let's remove the x/burn module in a separate PR. This test will fail if I remove the "burn" line because the burn module (specifically burn module account) is still in the repo.

rootulp:
The burn module was removed so we need to remove that from this PR

app/app_test.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/6176/commits/6c594db36a9985a222e11d64ff0a014e91667e8d

app/app_test.go
rootulp: https://github.com/celestiaorg/celestia-app/pull/6176/commits/6c594db36a9985a222e11d64ff0a014e91667e8d

app/app.go
```diff
@@ -644,11 +644,11 @@ func (app *App) ModuleAccountAddrs() map[string]bool {
 	return modAccAddrs
 }
 
-// BlockedAddresses returns all the app's blocked account addresses.
+// BlockedAddresses returns a map of the app's blocked account addresses.
```
damiannolan, line 647: Maybe this can shed more light on how the block list operates for developers.

```suggestion
// BlockedAddresses returns a map of the app's blocked account addresses.
// NOTE: This blocklist defines the set of addresses that are not allowed
// to receive funds through direct and explicit actions, for example, by using a MsgSend or
// by using a SendCoinsFromModuleToAccount execution.
// Thus, a module account such as the Hyperlane module account can still receive escrow funds via the protocol,
// however sending directly to the module account via x/bank is forbidden.
```

Issue #6177, opened 2025-11-12
`wrong Block.Header.Version. Expected {11 1}, got {11 2}`


rootulp:
## Context

@rach-id identified on an Arabica archive consensus node that is block sync-ing from genesis, it fails at the v2 upgrade height.

https://celestia-team.slack.com/archives/C02PFFCGMNW/p1762959487650529

## Problem

```
2:57PM ERR Error in validation err="wrong Block.Header.Version. Expected {11 1}, got {11 2}" module=blocksync
```

The consensus node is using the correct default v2 upgrade height for Arabica

```
2:56PM INF Since the chainID is arabica-11, configuring the default v2 upgrade height to 1751707
```

## Proposal

Investigate and fix the issue

rootulp:
Logs: https://pastebin.com/2C3dbE5V

rootulp:
celestia-core didn't update the state after processing block 1751706 even though it should have according to:

https://github.com/celestiaorg/celestia-app/blob/4dd92890ac0a50d163627053d6e0323c9b5ba6a9/app/app.go#L474-L477

I'm looking at the [diff](https://www.diffchecker.com/YHWUdtzM/) for updateState between v0.34.x-celestia and v0.39.x-celestia

rootulp:
I rolled back one height 

```
devs@snapshot-node-arabica:~/celestia-app$ sudo ~/.celestia-app/bin/v3.10.6/celestia-appd rollback --home /home/celestia/.celestia-app
```

and then tried starting it again:

```
devs@snapshot-node-arabica:~/celestia-app$ sudo ./build/celestia-appd start --home /home/celestia/.celestia-app --address=tcp://127.0.0.1:26658
5:57PM INF Overriding Mempool TTLNumBlocks to minimum configured=12 minimum=36 module=server
5:57PM INF Overriding Mempool MaxGossipDelay configured_seconds=0.00006 module=server new_seconds=0.00002
5:57PM INF Overriding Mempool MaxTxsBytes to minimum configured_mib=384 minimum_mib=400 module=server
5:57PM INF initializing multiplexer app_version=1 chain_id=arabica-11 module=server
5:57PM INF initialized remote app client address=127.0.0.1:26658 module=server
5:57PM INF starting comet node module=server
5:57PM INF service start impl=multiAppConn module=proxy msg="Starting multiAppConn service"
5:57PM INF service start connection=query impl=localClient module=abci-client msg="Starting localClient service"
[...]

PR #6197, opened 2025-11-13
chore!: undeprecate v2 upgrade height

rootulp:
While working on https://github.com/celestiaorg/celestia-app/issues/6177 I realized it is cumbersome to set the `--v2-upgrade-height` via ldflags.

This change reverts the deprecation for the `--v2-upgrade-height`. If a user passes the `--v2-upgrade-height` to the multiplexer, it forwards that to the embedded binaries. 

Note for reviewers: we don't strictly need this change. 

## Testing

```
./scripts/single-node-all-upgrades.sh

...
Upgrade to version 6 complete!
```

### DIFF
--- a/scripts/single-node-all-upgrades.sh
+++ b/scripts/single-node-all-upgrades.sh
@@ -2,10 +2,6 @@
 
 # This script starts a local single node testnet on app version 1 and then
 # upgrades to app version 2, 3, 4, 5, and 6.
-#
-# Prerequisites:
-# - Modify the `Makefile` and set V2_UPGRADE_HEIGHT = 2
-# - Run `make install`
 
 set -o errexit # Stop script execution if an error is encountered
 set -o nounset # Stop script execution if an undefined variable is used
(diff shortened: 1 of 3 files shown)


ninabarbakadze (approved):

scripts/single-node-all-upgrades.sh
```diff
@@ -101,7 +108,9 @@ startCelestiaApp() {
     --api.enable \
     --grpc.enable \
     --grpc-web.enable \
-    --delayed-precommit-timeout 1s
+    --force-no-bbr \
+    --v2-upgrade-height=3 \
```
rach-id, line 112: for known chainID, this is hardcoded in the binary, right?

rach-id (commented):
This change is breaking. Not sure if we should merge it now or wait for a major release 🤔

scripts/single-node-all-upgrades.sh
rootulp: Yes for Arabica, Mocha, and Mainnet if a user doesn't specify the `--v2-upgrade-height` then it gets defaulted in the v3.x binary

rootulp:
Agreed it is breaking so will hold off on merging for now.

ninabarbakadze:
> This change is breaking. Not sure if we should merge it now or wait for a major release 🤔

afaiu we're just waiting to cut v6.x, right?

rootulp:
> afaiu we're just waiting to cut v6.x, right?

Yes. IMO we shouldn't do that until we need to merge changes that are breaking (i.e. can't be merged to main). Likely sometime after v6 activation height on Mainnet.

PR #6200, opened 2025-11-13
test: block sync through v2 upgrade height

rootulp:
Test for https://github.com/celestiaorg/celestia-app/issues/6177
Blocked on https://github.com/celestiaorg/celestia-app/pull/6200 because this test will fail until we update the e2e tests to use a release with that change.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -133,7 +133,7 @@ func (app App) SetUnbondingTime(ctx context.Context) error {
 		return err
 	}
 
-	sdkCtx.Logger().Info("Setting unbonding time to %v.", appconsts.UnbondingTime)
+	sdkCtx.Logger().Info(fmt.Sprintf("Setting unbonding time to %v.", appconsts.UnbondingTime))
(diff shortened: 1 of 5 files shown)


rootulp:
I'm stumped. I tried re-running the test a few times and the test passes when I revert the fix in https://github.com/celestiaorg/celestia-core/pull/2641

Now I'm wondering if v0.39.x and main have differences but it doesn't look like it: https://www.diffchecker.com/nsNKxtyM/

rootulp:
The test fails with celestia-app v6.2.5 but passes with celestia-app built using https://github.com/celestiaorg/celestia-core/pull/2641

but the difference between that file and main is almost nothing. So the issue exists on v0.39.x-celestia but doesn't exist on main.

rootulp:
The test passes when I use celestia-core v0.39.15.

The test fails when I use celestia-app v6.2.5. The only other change in this PR has to do with the ldflags

rootulp:
Found the fix, it's literally the cursed v2 upgrade height set via ldflags

PR #6201, opened 2025-11-14
test: EndBlock on v3.x

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/6177

Extremely optional PR. I don't plan on cutting a release of v3.x if this merges. This is just to verify that v3.x `EndBlock` does in fact return `v2` in the `EndBlock.ConsensusParamUpdates` field for the v2 upgrade.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -20,6 +20,7 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 	"github.com/tendermint/tendermint/libs/log"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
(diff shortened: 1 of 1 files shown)


rach-id (commented):
This is just a test you want to run locally, right? nothing that needs to be on the v3 branch?

rootulp:
> This is just a test you want to run locally, right? nothing that needs to be on the v3 branch?

Ran locally already, confirmed it works. Figured optional to merge just in case we need to debug this again in the future. Since we're not cutting a v3.x release after this or changing v3.x, I'm open to closing or merging.

ninabarbakadze (approved):

PR #6202, opened 2025-11-14
fix(multiplexer): v2 upgrade height

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/6177

The multiplexer was setting the v2 upgrade height to `0`. 

Short term: this PR adds a check to only append the `--v2-upgrade-height` flag if it isn't set to 0.

Long term: remove the ldflags entirely in https://github.com/celestiaorg/celestia-app/pull/6197

### DIFF
--- a/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
+++ b/cmd/celestia-appd/cmd/modify_root_command_multiplexer.go
@@ -57,7 +57,7 @@ func modifyRootCommand(rootCommand *cobra.Command) {
 	}
 
 	v3Args := defaultArgs
-	if v2UpgradeHeight != "" {
+	if v2UpgradeHeight != "" && v2UpgradeHeight != "0" {
 		v3Args = append(v3Args, "--v2-upgrade-height="+v2UpgradeHeight)
 	}
 


rach-id (approved):
Great debugging. Was this tested on Chris's node?

rootulp:
> Was this tested on Chris's node?

Sadly no. I verified this fixes the e2e test.

But I can't explain why Chris's node that was running celestia-app v3.x (without multiplexer) still can't sync through the v2 upgrade height. This change wouldn't fix that issue because this fixes the multiplexer, not the v3.x binary.

rach-id:
> But I can't explain why Chris's node that was running celestia-app v3.x (without multiplexer) still can't sync through the v2 upgrade height. This change wouldn't fix that issue because this fixes the multiplexer, not the v3.x binary.

would we be able to to try and sync a node from genesis using the v6 binary and see if it works? to confirm the issue is gone

rootulp:
Yes agreed. @pippokr can we please run this version on one of the Arabica snapshot nodes?

evan-forbes (approved):

ninabarbakadze (approved):

PR #6223, opened 2025-11-19
fix(test): improve TestStateSyncMocha reliability

mojtaba-esk:
## Problem

The `TestStateSyncMocha` test was flaky in nightly CI runs. The test tries to state-sync from the live Mocha testnet, but it kept failing because the node couldn't connect to any peers. Even though we were writing seeds and persistent peers to the config file, CometBFT wasn't picking them up reliably. This might be a Tastora setup issue under the hood. 

## Solution

Instead of writing to the config file, we now pass seeds and persistent peers directly as CLI flags (`--p2p.seeds` and `--p2p.persistent_peers`) when starting the node. CometBFT reads CLI args correctly, so this ensures the node can actually connect to the Mocha network.
Also updated the docker image to `v6.2.5` to match the live Mocha network

Now the nightly tests should run much more stable.


test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -163,20 +160,24 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
 	dockerCfg, err := networks.NewConfig(mochaConfig, s.client, s.network)
 	s.Require().NoError(err, "failed to create mocha config")
 
-	// create a mocha chain builder (no validators, just for state sync nodes)
-	mochaChain, err := networks.NewChainBuilder(s.T(), mochaConfig, dockerCfg).
+	// Pass seeds and peers via CLI flags (CometBFT reads CLI args correctly, config file had issues)
```
rootulp, line 163: what was the issue with the config file? Can you please create a separate issue to investigate that?

mojtaba-esk: Not sure what was the issue, it seemed like when starting a node, it did not read the values from the config file. probably it was overridden by something else. sure will create an issue to track it.

mojtaba-esk: Follow up: #6230

PR #6224, opened 2025-11-19
docs: write early spec for tx client v2

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/6225

### DIFF
--- a/specs/src/tx_client_v2.md
+++ b/specs/src/tx_client_v2.md
@@ -113,7 +113,6 @@ This usually means we rolled back too far due to multiple rejections. **(this sh
 
 If this happens:
 
-- Don’t roll back again
 - Pause signing
 - Check if we're in recovery mode (the confirmation loop gets in recovery mode when transactions get a sequence mismatch in ReCheck)
 - Wait for recovery mode to complete


specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
+# Tx Client v2 – Sequential Submission Queue
+
+## Abstract
+
+The biggest pain point and bottleneck preventing 100% successful submissions is a race condition caused by eviction handling in tandem with our recovery mechanism in broadcast:
+
+1. A transaction **Tx(3)** is **evicted** from the mempool.
```
rach-id, line 7: what is Tx(3)? why not start with 1?

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+2. The **Tx client does not poll** quickly enough to detect the eviction.
```
rach-id, line 8: should we just increase the polling frequency to catch this? or check if the previous transaction was evicted before sending the new one?

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+3. A new transaction **Tx(4)** is submitted while the node still expects **sequence=3**.
+4. Recovery logic **re-signs Tx(3)** and submits it.
+5. Other nodes may still have the *original* Tx(3).
+
+    This results in a **race**:
+
+- Either the **evicted original Tx(3)** is included first,
+- Or the **new re-signed Tx(3)** is included first.
+
+**Both outcomes caused failures.** This recurring issue with transaction handling has sparked discussions around a fundamental redesign of our approach.
```
rach-id, line 18: > **Both outcomes caused failures.**

why not keep track of all the resigned transactions and check all of them if they're included when we have a new block?

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+To understand why we need a new design, let's first look at how TxClient v1 currently works. In v1, transactions are:
+
+1. Submitted via explicit `Submit*` methods.
+2. Confirmed in parallel by separate confirmation logic.
+
+This is not ideal for various reasons:
+
+- **Users had to coordinate everything themselves:** when to submit, at which cadence, when to confirm, how often to poll. The TxClient should handle that automatically.
+- **Handling failures requires knowing what happened before.** For example, you shouldn't resign a tx if the previous one was evicted. Currently there's no previous state tracking.
+- **Evictions, rejections, resubmissions weren't handled consistently or deterministically** because no single component owned the account's entire transaction flow causing races.
```
rach-id, line 31: what does `deterministically` mean in this context?

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+This prompted us to re-design **Tx Client v2** around a **sequential per account submission queue**, where one component owns:
+
+- nonce/sequence management
+- submission cadence
+- TxStatus polling(confirmations)
+- engine algorithm: when to resubmit, when to resign, etc.
+
+Additionally I propose that we:
+
+Drop mempool enforced TTLs and move TTL responsibility entirely to the client/user. The node no longer expires transactions; instead, callers control TTL via setting it as tx option.
```
rach-id, line 42: What if they don't specify a TTL or provide a very low gas price and a very long ttl? we should have a fallback mechanism to avoid having transactions stuck in the mempool unnecessarily which would impact the other transactions.

Also, seems like if I make the mistake of sending a transaction with a very low gas price and a very long ttl, and the network has some activity and only including transaction that has a high gas price, my account will no longer be able to send tranactions given all of them will have a higher sequence and will not be accepted, and there is no way for me to remove my original transaction until the activity on the network slows down

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+No API breakage: TxClient V2 still wraps the v1 client maintaining full backwards compatibility as well as the parallel submission flow introduced in v1. It solely focuses on introducing new API which should be a sequential tx queue.
+
+## Protocol/Component Description
+
+### Per account sequential queue
+
+The basic idea is that every account gets its own **transaction lane**.
+
+This lane is in charge of everything for that account: taking new transactions, broadcasting them in the right order, and keeping track of what happens to them afterwards.
```
rach-id, line 52: I like this 👍

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+Along the way, it's becoming pretty clear that we'll also need **a separate map from signer + sequence → tx status** (likely its hash and last known status). Even if we poll statuses in batches we still need to know which hash belongs to which sequence.
+
+---
+
+### Internal loops (still flexible)
+
+Each lane will probably run two background loops:
+
+1. **A submitter** - picks the next queued tx, signs it, broadcasts it to the node, and moves it into pending.
+    1. can also be triggered by channels receiving jobs (ig this defeats the purpose of the queue though)
```
rach-id, line 70: what's `ig`?

rach-id (commented):
Thanks for the specs 🙏 

I added some questions to understand more given I don't have any context regarding the mempool discussion

specs/src/tx_client_v2.md
ninabarbakadze: Just arbitrary example, it could be 1 too

specs/src/tx_client_v2.md
ninabarbakadze: We could increase polling frequency but if we hit sequence mismatch when we're trying to broadcast we'd also check if previous tx was evicted and we're waiting for it to be resubmitted.

specs/src/tx_client_v2.md
ninabarbakadze: We should never be hitting this scenario where we have two transactions with the same nonce

specs/src/tx_client_v2.md
ninabarbakadze: by deterministically I meant that every time we handle evictions, rejections, etc we should have the same outcome where for example transaction is resubmitted during evictions and it never gets front ran by another transaction.

specs/src/tx_client_v2.md
ninabarbakadze: > What if they don't specify a TTL or provide a very low gas price and a very long ttl? 

Client would set it by default but I can imagine how this would fall through the cracks for submitting transactions without the client. We could make it required.

> my account will no longer be able to send tranactions given all of them will have a higher sequence and will not be accepted, and there is no way for me to remove my original transaction until the activity on the network slows down

I think default mempool TTL is 30 blocks currently so we'd hit the same problem there, no?

specs/src/tx_client_v2.md
ninabarbakadze: "I guess" I was thinking out loud in this first iteration for readers to give suggestions. I'll remove it before we merge.

specs/src/tx_client_v2.md
rach-id: > 30 blocks currently so we'd hit the same problem there, no?

if I understand correctly, we would hit it for 30 blocks ~= 3min, but if there is no TTL, we could hit it indefinitely as long as there is congestion

evan-forbes (commented):
I don't think we should merge this yet, at least as a spec. It seems like a good explanation or discusison around some of the issues and some implementation details, however I don't think this should be used as a spec for a second rust based implementation. 

for the spec @ninabarbakadze and I sync'd and concluded that roughly:

abstract: 3 sentences that describe the problem, high level guarantee, then expected outcome (of if the prototype is finished we can measure the outcome and confirm this implementation gets zero failures)

intro: continue from the high level problem introduced in the abstract and narrow down to specifics. In this case the high level guarantee we're actually providing is onchain ordering being the same as submission order.

actual spec: english as code (the client MUST do _____, the client SHOULD do _______). sequence diagrams are always helpful but ofc options 

conclusion: a few sentences around to wrap up how the spec provides the outcome in the abstract

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
+# Tx Client v2 – Sequential Submission Queue
+
+## Abstract
```
tzdybal, line 3: It's probably too obvious to us all, but I think it should be explicitly stated that this issue is related to submission from single account.

specs/src/tx_client_v2.md
tzdybal: We need TTL on the protocol level, otherwise it will be easy DoS attack vector. Attacker would be able to spam the network with Txs without TTL or with very high TTL, with missing sequence numbers, etc.

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+Drop mempool enforced TTLs and move TTL responsibility entirely to the client/user. The node no longer expires transactions; instead, callers control TTL via setting it as tx option.
+
+No API breakage: TxClient V2 still wraps the v1 client maintaining full backwards compatibility as well as the parallel submission flow introduced in v1. It solely focuses on introducing new API which should be a sequential tx queue.
+
+## Protocol/Component Description
+
+### Per account sequential queue
+
+The basic idea is that every account gets its own **transaction lane**.
```
tzdybal, line 50: I would change a wording a bit - because the main idea of 'sequencial **queue**' is 'transaction **lane**'. And things are **queued** there. 

Probably we can think of better names. My proposition:
- transaction queue
- transaction states: new/waiting, broadcasted.

Of there are no other possible, we can as well just expose boolean property **Broadcasted**.

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
+# Tx Client v2 – Sequential Submission Queue
+
+## Abstract
+
+The biggest pain point and bottleneck preventing 100% successful submissions is a race condition caused by eviction handling in tandem with our recovery mechanism in broadcast:
```
tzdybal, line 5: Consider defining/describing eviction and rejection.

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+The monitoring loop periodically checks on pending txs and sees how the chain responded.
+
+Depending on the status:
+
+- **Still pending** → keep waiting
+- **Committed** → great, return success
+- **Committed with execution error** → return error
+- **Evicted** → put it back in the queue to try again/directly resubmit since it's already signed
+- **Rejected:**
```
tzdybal, line 137: Ok, so here are the possible Tx states :+1:

IMHO it's good to just define a flat structure with all states (queued, broadcasted, commited, errored, evicted, rejected).

ninabarbakadze:
Closing this and moving context elsewhere.

PR #6234, opened 2025-11-21
docs: fix outdated versioned constant references

bouchmann:
Remove incorrect `versioned constant` terminology for TxSizeCostPerByte and GasPerBlobByte. These constants were versioned in v3-v4 but the versioning was removed in v5 (commit 573b4a5).


rach-id:
I don't think this is a necessary change or adds any value. Closing. If you want to contribute, please check good first issues

PR #6235, opened 2025-11-21
fix(test): mocha statesync flaky nightly test

mojtaba-esk:
The test was timing out because it required the node to not be catching up (`!info.CatchingUp`) after state sync completed. However, after state sync finishes, it automatically transitions to block sync to catch up to the latest chain height, and during this block sync phase, `CatchingUp` remains `true` until fully synced.

The fix removes the `!info.CatchingUp` requirement from the sync condition. The test now only verifies that state sync completed by checking the node reached `trustHeight`. We still verify that state sync was actually used (not block sync fallback) via Prometheus metrics at the end, which will fail if state sync failed and block sync was used instead.

### DIFF
--- a/test/docker-e2e/e2e_state_sync_test.go
+++ b/test/docker-e2e/e2e_state_sync_test.go
@@ -204,8 +204,11 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
 	stateSyncClient, err := fullNode.GetRPCClient()
 	s.Require().NoError(err, "failed to get state sync client")
 
+	// Wait for state sync to complete (node reaches trust height).
+	// After state sync, the node transitions to block sync to catch up to latest height,
+	// but we verify state sync was used via metrics.
 	err = s.WaitForSync(ctx, stateSyncClient, stateSyncTimeout, func(info rpctypes.SyncInfo) bool {
-		return !info.CatchingUp && info.LatestBlockHeight >= trustHeight
+		return info.LatestBlockHeight >= trustHeight
 	})
 
 	s.Require().NoError(err, "failed to wait for state sync to complete")


test/docker-e2e/e2e_state_sync_test.go
```diff
@@ -204,8 +204,11 @@ func (s *CelestiaTestSuite) TestStateSyncMocha() {
 	stateSyncClient, err := fullNode.GetRPCClient()
 	s.Require().NoError(err, "failed to get state sync client")
 
+	// Wait for state sync to complete (node reaches trust height).
+	// After state sync, the node transitions to block sync to catch up to latest height,
+	// but we verify state sync was used via metrics.
 	err = s.WaitForSync(ctx, stateSyncClient, stateSyncTimeout, func(info rpctypes.SyncInfo) bool {
-		return !info.CatchingUp && info.LatestBlockHeight >= trustHeight
+		return info.LatestBlockHeight >= trustHeight
```
evan-forbes, line 211: makes sense

what is meant by checking if state is used via metrics? Which metric? iirc this is trying to ensure that since the chain is small we use state sync correct and not simply blocksync the entire thing

if we are truly pruning the first height, then afaiu that's the only way of determining this but I definitely could be missing a metric

evan-forbes (approved):

test/docker-e2e/e2e_state_sync_test.go
rootulp: This test infers whether state sync was used by looking at metrics and searching for `apply_snapshot_chunk` which seems like a reasonable proxy

Ref: https://github.com/celestiaorg/celestia-app/blob/599b3b25dfcca5d0a514e711ff1b37b85b8903d3/test/docker-e2e/e2e_state_sync_test.go#L282-L287

mojtaba-esk: The first trial for state sync test was to use the pruning idea, but due to the underlying limitation, that did not work. So the only way was to read metrics to be certain that the sync was done using state sync and not block sync. This PR removes an extra check to avoid long waits for the node to sync to the tip of the chain thus resulting sometimes in timeouts. Now it just waits for the node to sync only up to the trust height and then checks the metrics to ensure the sync method.
Now it is merged, so we can keep an eye on it for the next couple of days to see if they fail again or not.

PR #6239, opened 2025-11-24
chore(ci): upgrade checkout to v6

rejected-l:
Update checkout action to v6. Better credential handling in containers.

https://github.com/actions/checkout/releases/tag/v6.0.0


rach-id:
We leave deps updates to dependabot.

PR #6264, opened 2025-11-26
fix(signal): correct FirstSignalKey value to actually differ from UpgradeKey

levanchverak:
Found that `FirstSignalKey` was set to `[]byte{0x000}` which is identical to `UpgradeKey` (`[]byte{0x00}`) - in Go, `0x000` equals `0x00` equals `0`.

The comment says FirstSignalKey should be a "divider" separating UpgradeKey from validator signal keys, but both had the same value. The code worked only because of an explicit `bytes.Equal` check that skipped UpgradeKey during iteration.

Changed FirstSignalKey to `[]byte{0x00, 0x00}` (2 bytes) which is:
1 lexicographically greater than UpgradeKey (1 byte)
2 lexicographically less than any 20-byte validator address

### DIFF
--- a/x/signal/types/keys.go
+++ b/x/signal/types/keys.go
@@ -7,6 +7,7 @@ var (
 
 	// FirstSignalKey is used as a divider to separate the UpgradeKey from all
 	// the keys associated with signals from validators. In practice, this key
-	// isn't expected to be set or retrieved.
-	FirstSignalKey = []byte{0x000}
+	// isn't expected to be set or retrieved. It must be lexicographically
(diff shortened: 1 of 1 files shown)


rootulp:
Even though the PR description seems correct, I don't think we should change the FirstSignalKey b/c that's breaking.

Ref: https://go.dev/play/p/CGw4NPTTClr

Relevant, code that iterates starting with the FirstSignalKey, skips over the UpgradeKey b/c they're the same: https://github.com/celestiaorg/celestia-app/commit/0bfd074807b6e9a4c882f386432f4a4ea85f3478

Issue #5927, opened 2025-10-01
Add zk ism module client CLIs


damiannolan:
## Description

The zk ism module requires client CLIs for invoking the msg server rpcs.

An out-of-date impl exists for ism creation and this can be updated.
We should also add CLIs for submission of zk proofs for the other two rpcs.

PR #6266, opened 2025-11-26
feat: Generic ISM

jonas089:
As discussed synchronously with @damiannolan, this PR alongside with https://github.com/celestiaorg/celestia-zkevm/pull/310 replaces the evolve-specific ZK ISM with a generic implementation that is re-usable for things like consensus proofs. 

Anyone can register a new ISM for their block prover or consensus client and the only requirement is that the outputs adhere to our structure: `state_len_le_u64 || state_bytes || new_state_bytes` and that `state_bytes` starts with the 32 byte trusted root that is used to verify Hyperlane message inclusion in `SubmitMessage`.

### DIFF
--- a/x/zkism/README.md
+++ b/x/zkism/README.md
@@ -20,9 +20,10 @@ The `Verify` method consumes the stored message IDs and authorizes the message f
 - Authorization Set: A transient set of message IDs. Membership proofs add message IDs; the Hyperlane router consumes them during message verification.
 
 ## State
-Users can define any State and write circuits that encapsulate a transition from State => New State. 
+
+Users can define any State and write circuits that encapsulate a transition from State => New State.
 The only constraint is that the circuit outputs must be the raw bytes of `state_length_u64_as_little_endian_bytes` || `state` || `new_state` and that
-the `state_root`, which is used to verify Hyperlane messages, is the first field in `state`, e.g. the first 32 bytes of `state` should be the state root, 
+the `state_root`, which is used to verify Hyperlane messages, is the first field in `state`, e.g. the first 32 bytes of `state` should be the state root,
 regardless of the use-case (Evolve Block Prover, ZK Consensus Client, ...). If this rule is violated, the ISM will not be able to verify Hyperlane messages
 against the incremental Tree and will only store the most recent `state`.
 
--- a/x/zkism/keeper/msg_server_test.go
+++ b/x/zkism/keeper/msg_server_test.go
@@ -131,14 +131,11 @@ func (suite *KeeperTestSuite) TestUpdateInterchainSecurityModule() {
 func (suite *KeeperTestSuite) TestSubmitMessages() {
 	trustedState, err := hex.DecodeString("fac92413c55a229e0b67ca195c32f19cfab4ba670a150215302564cf68531d9fd2b39cf65fbbea4bdbf30aea8fc46e050b8fa020234ee8223acc25da3b26f4fb4200000000000000ef0000000000000000000000000000000000000000000000000000a8045f161bf468bf4d443324b79a9a978e3de925c8069ad2316ee2324b3ca961d5941b5a68b2901c6ec9")
 	suite.Require().NoError(err)
-	// trusted root is first 32 bytes of trusted state
-	trustedRoot := trustedState[:32]
-	suite.Require().NoError(err)
 	trustedCelestiaHash, err := hex.DecodeString("777c8f454c421f9f6402f2512a273c55078eadde2ac784288791d79f510ad56e")
 	suite.Require().NoError(err)
 	trustedCelestiaHeight := uint64(35)
 
-	ism := suite.CreateTestIsm(trustedRoot, trustedCelestiaHash, trustedCelestiaHeight)
+	ism := suite.CreateTestIsm(trustedState, trustedCelestiaHash, trustedCelestiaHeight)
 	proofBz, pubValues := readStateMembershipProofData(suite.T())
 
 	var msg *types.MsgSubmitMessages


rootulp:
Reviewers can still start reviewing but FYI

- lint is failing: https://github.com/celestiaorg/celestia-app/actions/runs/19708979641/job/56464034511?pr=6266#step:5:38
- test is failing: https://github.com/celestiaorg/celestia-app/actions/runs/19708979641/job/56464034687?pr=6266#step:4:146

damiannolan:
I will prioritise reviewing this tomorrow 🙏🏻

jonas089:
> Reviewers can still start reviewing but FYI
> 
> * lint is failing: https://github.com/celestiaorg/celestia-app/actions/runs/19708979641/job/56464034511?pr=6266#step:5:38
> * test is failing: https://github.com/celestiaorg/celestia-app/actions/runs/19708979641/job/56464034687?pr=6266#step:4:146

@rootulp @damiannolan The test that is failing is failing because it requires us to re-generate the proof payloads. I'm working on it.

app/app.go
```diff
@@ -161,31 +161,31 @@ type App struct {
...
+	EvidenceKeeper                 evidencekeeper.Keeper
+	TransferKeeper                 ibctransferkeeper.Keeper
+	FeeGrantKeeper                 feegrantkeeper.Keeper
+	ICAHostKeeper                  icahostkeeper.Keeper
+	PacketForwardKeeper            *packetforwardkeeper.Keeper
+	BlobKeeper                     blobkeeper.Keeper
+	CircuitKeeper                  circuitkeeper.Keeper
+	HyperlaneKeeper                hyperlanekeeper.Keeper
+	WarpKeeper                     warpkeeper.Keeper
+	InterchainSecurityModuleKeeper *zkismkeeper.Keeper
```
damiannolan: We can just call this `IsmKeeper` or `ZkIsmKeeper`

proto/celestia/zkism/v1/events.proto
```diff
@@ -20,41 +20,27 @@ message EventCreateZKExecutionISM {
   string owner = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
   // state commitment root of the remote chain
   string state_root = 3;
```
damiannolan: It looks like this field is not being used in the implementation for event emission anymore.
Since it is included within the opaque `state` bytes, we can just omit it

proto/celestia/zkism/v1/events.proto
```diff
@@ -20,41 +20,27 @@ message EventCreateZKExecutionISM {
...
-  // trusted celestia header hash
-  string celestia_header_hash = 5;
-  // trusted celestia height
-  uint64 celestia_height = 6;
-  // the celestia namespace identifier used by the application (hex-encoded)
-  string namespace = 7;
-  // the public key of the sequencer (hex-encoded)
-  string sequencer_public_key = 8;
+  // latest state
+  bytes state = 4;
```
damiannolan: For events I recommend that we keep this as a `string` field and we hex encode as we were doing previously with fields like `state_root`

proto/celestia/zkism/v1/events.proto
```diff
@@ -20,41 +20,27 @@ message EventCreateZKExecutionISM {
...
-  // state commitment root of the remote chain
-  string state_root = 2;
-  // latest tracked height of the remote chain
-  uint64 height = 3;
-  // trusted celestia header hash
-  string celestia_header_hash = 4;
-  // trusted celestia height
-  uint64 celestia_height = 5;
+  // new state
+  bytes state = 2;
```
damiannolan: Ditto, same comment regarding string and hex encoding the bytes for events.

proto/celestia/zkism/v1/tx.proto
```diff
@@ -13,11 +13,11 @@ option go_package = "github.com/celestiaorg/celestia-app/x/zkism/types";
 service Msg {
   option (cosmos.msg.v1.service) = true;
 
-  // CreateZKExecutionISM defines the rpc method for creating a new ZK execution ISM.
-  rpc CreateZKExecutionISM(MsgCreateZKExecutionISM) returns (MsgCreateZKExecutionISMResponse);
+  // CreateInterchainSecurityModule defines the rpc method for creating a new ZK execution ISM.
```
damiannolan:
```suggestion
  // CreateInterchainSecurityModule defines the rpc method for creating a new zk ISM.
```

proto/celestia/zkism/v1/tx.proto
```diff
@@ -13,11 +13,11 @@ option go_package = "github.com/celestiaorg/celestia-app/x/zkism/types";
...
+  rpc CreateInterchainSecurityModule(MsgCreateInterchainSecurityModule) returns (MsgCreateInterchainSecurityModuleResponse);
 
-  // UpdateZKExecutionISM defines the rpc method for updating an existing ZK execution ISM.
-  rpc UpdateZKExecutionISM(MsgUpdateZKExecutionISM) returns (MsgUpdateZKExecutionISMResponse);
+  // UpdateInterchainSecurityModule defines the rpc method for updating an existing ZK execution ISM.
```
damiannolan:
```suggestion
  // UpdateInterchainSecurityModule defines the rpc method for updating an existing zk ISM.
```

proto/celestia/zkism/v1/tx.proto
```diff
@@ -26,41 +26,31 @@ service Msg {
...
-  uint64 height = 3;
-  // trusted celestia header hash
-  bytes celestia_header_hash = 4;
-  // trusted celestia height
-  uint64 celestia_height = 5;
-  // the celestia namespace identifier used by the application
-  bytes namespace = 6;
-  // the public key of the sequencer
-  bytes sequencer_public_key = 7;
+  // initial state
```
damiannolan:
```suggestion
  // initial trusted state
```

proto/celestia/zkism/v1/tx.proto
```diff
@@ -76,16 +66,10 @@ message MsgUpdateZKExecutionISM {
...
-  string state_root = 1;
-  // latest tracked height of the remote chain
-  uint64 height = 2;
-  // trusted celestia header hash
-  string celestia_header_hash = 3;
-  // trusted celestia height
-  uint64 celestia_height = 4;
+// MsgUpdateInterchainSecurityModuleResponse is the response type for UpdateInterchainSecurityModule.
+message MsgUpdateInterchainSecurityModuleResponse {
+  // current ism state
```
damiannolan:
```suggestion
  // new ism state
```

x/zkism/README.md
damiannolan, line 1: I think this document is going to need a bigger refactor after this PR is merged.
I'll create an issue for this. I'm happy to pick that work up

x/zkism/client/cli/tx.go
```diff
@@ -40,7 +40,7 @@ This CLI command requires both the StateTransitionVkey and the StateMembershipVk
 			}
 
 			// TODO: fill in the remaining fields for the CLI cmd
```
damiannolan, line 42: We have an issue for this which still needs to completed https://github.com/celestiaorg/celestia-app/issues/5927

x/zkism/keeper/events.go
```diff
@@ -4,42 +4,27 @@ import (
...
+	return ctx.EventManager().EmitTypedEvent(&types.EventCreateInterchainSecurityModule{
 		Id:                  ism.Id,
 		Owner:               ism.Owner,
-		CelestiaHeaderHash:  encodeHex(ism.CelestiaHeaderHash),
-		CelestiaHeight:      ism.CelestiaHeight,
-		StateRoot:           encodeHex(ism.StateRoot),
-		Height:              ism.Height,
-		Namespace:           namespace.String(),
-		SequencerPublicKey:  encodeHex(ism.SequencerPublicKey),
+		State:               ism.State,
```
damiannolan: when we update the field to be string again

```suggestion
		State:               encodeHex(ism.State),
```

x/zkism/keeper/events.go
```diff
@@ -4,42 +4,27 @@ import (
...
-	return ctx.EventManager().EmitTypedEvent(&types.EventUpdateZKExecutionISM{
-		Id:                 ism.Id,
-		StateRoot:          encodeHex(ism.StateRoot),
-		Height:             ism.Height,
-		CelestiaHeaderHash: encodeHex(ism.CelestiaHeaderHash),
-		CelestiaHeight:     ism.CelestiaHeight,
+func EmitUpdateISMEvent(ctx sdk.Context, ism types.InterchainSecurityModule) error {
+	return ctx.EventManager().EmitTypedEvent(&types.EventUpdateInterchainSecurityModule{
+		Id:    ism.Id,
+		State: ism.State,
```
damiannolan: And same here

```suggestion
		State: encodeHex(ism.State),
```

x/zkism/keeper/keeper.go
```diff
@@ -102,38 +102,9 @@ func (k *Keeper) Verify(ctx context.Context, ismId util.HexAddress, _ []byte, me
 	return authorized, nil
 }
 
-func (k *Keeper) validatePublicValues(ctx context.Context, ism types.ZKExecutionISM, publicValues types.EvExecutionPublicValues) error {
```
damiannolan, line 105: Nice! 🙏🏻 

Context for the reader: this is where we see the primary benefit to this PR, where we assert the trusted opaque state bytes are part of our proof values. This allows to us accept the new state bytes once the proof has been verified.

x/zkism/keeper/keeper_test.go
damiannolan, line 1: I think we can remove `publicKeyHex` and `namespaceHex` test variables here. They are causing the linter to fail since they are unused now

x/zkism/keeper/msg_server.go
```diff
@@ -68,6 +62,7 @@ func (m msgServer) UpdateZKExecutionISM(ctx context.Context, msg *types.MsgUpdat
 		return nil, errorsmod.Wrap(sdkerrors.ErrInvalidType, err.Error())
 	}
 
+	// verify that the State used in the public values is the same as the ISM state
```
damiannolan: Suggestion on wording, bit simpler and less wordy

```suggestion
	// verify the public values against the trusted ism state
```

x/zkism/types/ism.go
```diff
@@ -13,10 +13,10 @@ const (
 	InterchainSecurityModuleTypeZKExecution = 42
```
damiannolan, line 13: We should rename this field... doesn't really matter what its called, on that it doesn't collide with the uint8 numbers used by hyperlane's core ism types. https://github.com/bcp-innovations/hyperlane-cosmos/blob/main/x/core/01_interchain_security/types/types.go#L33

We could literally just call it `ModuleType` if we wanted. I'm fine with that or `ModuleTypeZkISM`

x/zkism/types/msgs.go
```diff
@@ -2,49 +2,23 @@ package types
...
-		return errorsmod.Wrapf(ErrInvalidVerifyingKey, "invalid groth16 verifying key")
-	}
-
-	if len(msg.StateTransitionVkey) != 32 {
-		return errorsmod.Wrap(ErrInvalidVerifyingKey, "program verifying key commitment must be exactly 32 bytes")
-	}
-
-	if len(msg.StateMembershipVkey) != 32 {
-		return errorsmod.Wrap(ErrInvalidVerifyingKey, "program verifying key commitment must be exactly 32 bytes")
-	}
```
damiannolan, line 41: Can you add this back? This validation shouldn't be removed

x/zkism/types/msgs.go
```diff
@@ -2,49 +2,23 @@ package types
...
-		return errorsmod.Wrapf(ErrInvalidNamespace, "failed to parse namespace from bytes: %x", msg.Namespace)
-	}
-
-	if len(msg.SequencerPublicKey) != 32 {
-		return errorsmod.Wrap(ErrInvalidSequencerKey, "public must be exactly 32 bytes")
-	}
-
-	if len(msg.StateRoot) != 32 {
-		return errorsmod.Wrap(ErrInvalidStateRoot, "state root must be exactly 32 bytes")
-	}
```
damiannolan, line 29: Since we are removing these fields in favour of `state`. I think we should enforce that the state be a minimum of 32 bytes, as that's what we expect for the root used when verifying messages.

x/zkism/types/msgs_test.go
```diff
@@ -6,13 +6,12 @@ import (
...
 	"github.com/bcp-innovations/hyperlane-cosmos/util"
 	"github.com/celestiaorg/celestia-app/v6/x/zkism/types"
-	"github.com/celestiaorg/go-square/v2/share"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/stretchr/testify/require"
 )
 
-func TestMsgCreateZKExecutionISMValidateBasic(t *testing.T) {
-	var msg *types.MsgCreateZKExecutionISM
+func TestMsgCreateInterchainSecurityModuleValidateBasic(t *testing.T) {
```
damiannolan, line 13: These test cases are failing now because of the removal of the code I mentioned in `msgs.go`

x/zkism/types/public_values.go
```diff
@@ -4,59 +4,26 @@ import (
 	"bytes"
 	"encoding/binary"
 	"encoding/hex"
-	"errors"
 	"fmt"
 	"io"
 )
 
 // EvExecutionPublicValues defines the set of public values associated with a proof generated by a ZK program.
 type EvExecutionPublicValues struct {
```
damiannolan: We can name this type `PublicValues` now I think

x/zkism/types/public_values.go
```diff
@@ -114,49 +57,24 @@ func (pi *EvExecutionPublicValues) Marshal() ([]byte, error) {
...
 	}
 
-	if buf.Len() != 0 {
-		return errors.New("trailing data after decoding")
+	// everything else becomes NewState
+	remaining, err := io.ReadAll(buf)
+	if err != nil {
+		return err
 	}
+	pi.NewState = remaining
```
damiannolan: I think we should length prefix the new state as well as it becomes much more robust with it. And its only worth the trade off of 8 bytes.

This way the format would become:
```
<len_old> | <old_state> | <len_new> | <new_state> | [optional future tail]
```

Main Benefits:
- We can append future fields without much friction
- We can ensure that there is no trailing data which gives us confidence/safety around type encoding

While we don't necessarily need to length prefix the new state today, the fact that its cheap to do could save us future headaches tomorrow.

x/zkism/types/public_values.go
```diff
@@ -114,49 +57,24 @@ func (pi *EvExecutionPublicValues) Marshal() ([]byte, error) {
...
-
-	if err := readBytes(buf, pi.NewStateRoot[:]); err != nil {
-		return fmt.Errorf("read NewStateRoot: %w", err)
-	}
-
-	if err := readBytes(buf, pi.Namespace[:]); err != nil {
-		return fmt.Errorf("read Namespace: %w", err)
+	// read length of State
+	var l uint64
+	if err := binary.Read(buf, binary.LittleEndian, &l); err != nil {
```
damiannolan, line 68: I think the linter is complaining about the `readBytes()` helper fn not being used. We can either use it in places like this or remove it.

x/zkism/types/msgs.go
jonas089: I am adding this check to validatePublicValues.

damiannolan: We should ensure that it happens in both. Like you are suggesting in `validatePublicInputs` we should ensure that the new state is indeed 32 bytes minimum. 

This is for ism creation. And the benefit of doing it in this function is that it will reject the tx from ever entering a full nodes mempool if the initial `state` is less than 32 bytes :)

damiannolan:
Protobuf break check is always going to fail on this branch so no need to worry about it. I can help try to address some of the CI here.

damiannolan:
Docker build failure is unrelated to this PR. Looks like perhaps the self-hosted runner needs some maintenance.

```
 > [linux/amd64 builder 15/15] RUN uname -a &&    CGO_ENABLED=0 GOOS=linux GOARCH=amd64     make build DOWNLOAD=false:
291.1 github.com/cosmos/cosmos-sdk/x/genutil/migrations/v046: mkdir /tmp/go-build3464629736/b2112/: no space left on device
291.1 github.com/cosmos/cosmos-sdk/x/genutil/migrations/v047: mkdir /tmp/go-build3464629736/b2113/: no space left on device
291.1 github.com/cosmos/cosmos-sdk/server/cmd: mkdir /tmp/go-build3464629736/b2114/: no space left on device
291.1 github.com/cosmos/cosmos-sdk/client/snapshot: mkdir /tmp/go-build3464629736/b2109/: no space left on device
291.2 # github.com/celestiaorg/celestia-app/v6/internal/embedding
291.2 compile: writing output: write $WORK/b1996/_pkg_.a: no space left on device
291.3 # github.com/tendermint/tendermint/proto/tendermint/types
291.3 compile: writing output: write $WORK/b2014/_pkg_.a: no space left on device
291.3 go: failed to trim cache: write /root/.cache/go-build/trim.txt: no space left on device
291.7 make: *** [Makefile:59: build] Error 1
```

build.sh
damiannolan, line 1: We can remove this file

x/zkism/keeper/keeper_test.go
```diff
@@ -45,10 +43,10 @@ func TestKeeperTestSuite(t *testing.T) {
 func (suite *KeeperTestSuite) SetupTest() {
 	testApp, _ := testutil.SetupTestAppWithGenesisValSet(app.DefaultConsensusParams())
 	suite.ctx = testApp.NewUncachedContext(false, cmtproto.Header{Version: version.Consensus{App: appconsts.Version}})
-	suite.zkISMKeeper = testApp.ZKExecutionISMKeeper
+	suite.zkISMKeeper = testApp.IsmKeeper
 }
 
-func (suite *KeeperTestSuite) CreateTestIsm(trustedRoot []byte, trustedCelestiaHash []byte, trustedCelestiaHeight uint64) types.ZKExecutionISM {
+func (suite *KeeperTestSuite) CreateTestIsm(trustedState []byte, trustedCelestiaHash []byte, trustedCelestiaHeight uint64) types.InterchainSecurityModule {
```
damiannolan, line 47: I think we'll be able to remove the `trustedCelestiaHeight` var here, but that should be part of the next PR to clean out the unneeded stuff, so I'm happy to take care of that clean up

x/zkism/keeper/msg_server_test.go
```diff
@@ -62,18 +54,18 @@ func (suite *KeeperTestSuite) TestCreateZKExecutionISM() {
 	}
 }
 
-func (suite *KeeperTestSuite) TestUpdateZKExecutionISM() {
-	trustedRoot, err := hex.DecodeString("340623e91b5a72c72a9214357cbc99a6b59ef23e5069ac5354a3a1619af0d568")
+func (suite *KeeperTestSuite) TestUpdateInterchainSecurityModule() {
+	trustedState, err := hex.DecodeString("8d29024fff03736c34ed9d8fd28e4c3efccd327a67808f68a868cc4691dd9435287fbab01019e1e327b8020ba833c2a8661cc4d1654340ad1b2b17d3af692b802400000000000000770000000000000000000000000000000000000000000000000000a8045f161bf468bf4d44825a3f4ccd1d0a13a418683c03108460623b760888c34c81fc75d6af376839bd")
```
damiannolan: What was your process for getting the hex encoded trusted state here?

Did you just generate a proof and then hex dump the trusted state?

damiannolan: It might be good to also update the `TestSubmitMessages` below to use this hex encoded state field, as its currently passing a hardcoded state root, so in that test the `state` field on the ism only actually contains the state root itself now, and none of the other encoded data.

This would require generating a new message proof as well tho, since the root has likely changed. I'm fine to leave it for now as long as its known - can address in a follow up PR.

What do you think?

x/zkism/types/public_values.go
damiannolan: Could we add this to the encoding @jonas089? I feel like I'd sleep better at night if we added this

damiannolan:
The e2e test failures are due to our feat branch being out of date. I actually fixed the root cause of this in https://github.com/celestiaorg/celestia-app/pull/5724 which was backported to the v5 line.

Once we rebase our feature branch this failures should go away, but we can do that later as it doesn't need to be done for this PR which is targeted at our feat branch anyways

x/zkism/keeper/msg_server_test.go
jonas089: For that we'd have to re-generate the payloads. Maybe we can do this in a follow up.

x/zkism/keeper/keeper_test.go
jonas089: Yes as we discussed there can be a follow up to remove the things we don't need anymore.

build.sh
jonas089: Will remove it after testing everything with the new changes.

x/zkism/types/public_values.go
jonas089: @damiannolan Sure, I can add this.

damiannolan: Thank you! 🙏🏻

x/zkism/keeper/msg_server_test.go
damiannolan: Yeah, no worries

jonas089: @damiannolan I took care of this because I had to re-generate the payloads anyways when adding the prefix for the new_state.

x/zkism/keeper/msg_server_test.go
```diff
@@ -127,22 +119,24 @@ func (suite *KeeperTestSuite) TestUpdateZKExecutionISM() {
...
 		})
 	}
 }
 
 func (suite *KeeperTestSuite) TestSubmitMessages() {
-	trustedRoot, err := hex.DecodeString("acd4fcbcd3bbf25bd2055b2125f7d361f9f58d97ad167fe35a5b7f1806f5f8ea")
+	trustedState, err := hex.DecodeString("fac92413c55a229e0b67ca195c32f19cfab4ba670a150215302564cf68531d9fd2b39cf65fbbea4bdbf30aea8fc46e050b8fa020234ee8223acc25da3b26f4fb4200000000000000ef0000000000000000000000000000000000000000000000000000a8045f161bf468bf4d443324b79a9a978e3de925c8069ad2316ee2324b3ca961d5941b5a68b2901c6ec9")
+	suite.Require().NoError(err)
+	// trusted root is first 32 bytes of trusted state
+	trustedRoot := trustedState[:32]
```
damiannolan: Can't we just pass the full `trustedState` bytes to `suite.CreateTestIsm()` now?

PR #6267, opened 2025-11-26
feat: add service for batch tx status queries

ninabarbakadze:
## Overview

Adds a service for tx status batch queries in app.

### DIFF
--- a/app/grpc/tx/server.go
+++ b/app/grpc/tx/server.go
@@ -133,7 +133,6 @@ func (s *txServer) TxStatusBatch(ctx context.Context, req *TxStatusBatchRequest)
 		responses[i] = &TxStatusResult{
 			TxHash: req.TxIds[i],
 			Status: &TxStatusResponse{
-
 				Height:        status.Result.Height,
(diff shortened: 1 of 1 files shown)


ninabarbakadze:
fixing lint 1 sec

app/grpc/tx/server.go
```diff
@@ -90,3 +90,63 @@ func (s *txServer) TxStatus(ctx context.Context, req *TxStatusRequest) (*TxStatu
 		Signers:       resTx.Signers,
 	}, nil
 }
+
+func (s *txServer) TxStatusBatch(ctx context.Context, req *TxStatusBatchRequest) (*TxStatusBatchResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "request cannot be nil")
+	}
+
+	if len(req.TxIds) > 20 {
```
mcrakhman, line 99: nit: this could be a constant

ninabarbakadze: true! I have this as a constant on the core side and wasn't sure even if I need this here.

PR #6281, opened 2025-12-01
chore: update zk ism spec doc for generic verifier refactor

damiannolan:
## Overview

Updates the zk ism module spec document.


damiannolan:
Merging this, any follow up edits can be subsequently

PR #6282, opened 2025-12-01
chore: rename public values types for state transition and inclusion proofs

damiannolan:
## Overview

Basic renaming...


damiannolan:
@jonas089 I forgot that when we renamed the public values type to just `PublicValues` that we also have the other one (which we forgot to rename).

This PR just renames both types to `StateTransitionValues` and `StateMembershipValues`.

PR #6285, opened 2025-12-04
feat: add lumina latency monitor

mcrakhman:
## Overview

As per discussions with @evan-forbes we decided to move here the latency-monitor based on lumina client to be able to test submissions closer to the real environment.


ninabarbakadze (approved):
can't really comment on rust implementation but utAck

PR #6296, opened 2025-12-08
docs: align ShareProof spec with implementation

kurahin:
Updated networking.md to align ShareProof with the implemented proto: removed obsolete isCol/position and single-share fields, documented current data, shareProofs, rowProof, namespaceId, and namespaceVersion, and clarified that proofs are built against row roots only.


ninabarbakadze (commented):
please refer to good first issues for contributions

PR #6298, opened 2025-12-08
feat: query and store validator set metadata

evan-forbes:
## Overview

I've had to recreate this a few times over the years when I wanted to identify specific validators. It's not that complicated just annoying to remember how to do this, so figured we could add it to the tools.

### DIFF
--- a/tools/query-valset/main.go
+++ b/tools/query-valset/main.go
@@ -1,0 +1,274 @@
+package main
+
+import (
+	"context"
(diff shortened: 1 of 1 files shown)


rootulp:
I'm trying to clean up old PRs. IMO if a tool works, we should :shipit: 

@evan-forbes do you have any opinions on merging vs closing?

rootulp:
Closing b/c stale.

PR #6299, opened 2025-12-09
docs: fix incorrect links and CLI commands in blob and signal modules

lolkdsanyaw:
In x/blob/README.md:
   - Fixed broken links to CreateCommitment function (pointed to non-existent file)
   - Updated with proper pkg.go.dev links to go-square package
   - Replaced missing SubmitPayForBlobs reference with actual implementation paths

In x/signal/README.md:
   - Added missing required arguments to CLI commands
   - Added two undocumented query commands (upgrade, missing-validators)
   - Organized commands into query/tx sections for better readability


ninabarbakadze (commented):
not actually fixing the links

PR #6313, opened 2025-12-15
docs: tx client v2 spec

### LINKED ISSUE #6225: Write first iteration of the spec for TxClient v2
## Description

In an effort to get everyone on the same page when it comes to redesigning the tx client I wrote a spec with high level design, could be considered more of an RFC. 

## Acceptance Criteria

- [ ] Get feedback on design
- [ ] Resume prototype to confirm that the design is sound

ninabarbakadze:
## Overview

Fixes #6225


specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
+# Tx Client v2 – Sequential Submission Queue
```
rootulp, line 1: [blocking] this is in the specs directory but it won't show up in the specs hosted on https://celestiaorg.github.io/celestia-app/ unless it's added as a page to https://github.com/celestiaorg/celestia-app/blob/main/specs/src/SUMMARY.md

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+Inside the queue we keep two kinds of transactions:
+
+- **Queued** - submitted but not broadcast yet
+- **Pending** - already broadcast and waiting for confirmation
+
+---
+
+### Broadcast and Confirmation loops
+
+Each lane runs two background loops:
```
rootulp: What is a lane? How does a lane differ from a queue?

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. **A submitter**
+    - Picks the next queued tx, signs it, broadcasts it to the node.
+    - Listens for evicted transactions and resubmits them.
```
rootulp: [no change needed] listening for evicted transactions naively like a responsibility for the "monitor". I would expect the monitor to listen for evicted txs and if it needs to resubmit a tx, add that as a task for the submitter.

specs/src/tx_client_v2.md
rootulp: Do you intend for this to be visible on the specs? If yes then please add it to the summary. If no, then let's move this to a different location (docs/architecture/)

specs/src/tx_client_v2.md
ninabarbakadze: forgot to rename should be queue

specs/src/tx_client_v2.md
ninabarbakadze: I see what you mean. I make it clearer in the monitor description how that communication happens but perhaps I'll add a few words here as well.

Monitor will be the one to find out that a transaction was evicted. It will send that tx to a channel that a submitter is listening to. Submitter should drop everything and resubmit evicted tx.

specs/src/tx_client_v2.md
ninabarbakadze: yeah I'm not 100% sure about this. We want the spec to be for other contributors to be able to replicate the implementation.

rootulp: Other contributors can reference the ADR so IMO seems possible to move there

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+Inside the queue we keep two kinds of transactions:
+
+- **Queued** - submitted but not broadcast yet
+- **Pending** - already broadcast and waiting for confirmation
+
+---
+
+### Broadcast and Confirmation loops
+
+Each queue runs two background loops:
```
mcrakhman: I would think that the number of loops should be bounded to some fixed number, because the number of signers can be arbitrary high.

E.g. we can just have let's say X (fixed number) of workers which just perform the tasks (e.g. we can have X submission workers and X confirmation workers). This will allow enough parallelism without creating too much contention.

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+
+1. **A submitter**
+    - Picks the next queued tx, signs it, broadcasts it to the node.
+    - Listens for evicted transactions and resubmits them.
+2. **A monitor**
+    Checks the statuses of pending txs and knows how to manage evictions/rejections/confirmations.
```
mcrakhman: This can actually be even one loop per node if we change a little bit the txbatch "tx_status_batch" method which we will probably be using here:
- instead of returning all the confirmed hashes for a single signer we can I guess return only one with is the highest sequence number (also we can do a binary search here).
- we can check txs from top to bottom which should be faster (because if the top one is not confirmed then so is the bottom one), we can also do a binary search here.
- we can combine multiple signers to reduce the number of requests and potentially remove the need to have parallel confirmation requests for a single node

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+
+Tx is neither evicted nor rejected or confirmed. Return error to the user that tx is unknown.
+
+### Backward compatibility
+
+Tx Client v2 wrapps Tx Client v1 maintaining full API compatibility.
+
+## Message Structure/Communication Format
+
+TBD
```
mcrakhman: Should we mention somewhere the edge cases regarding the min fee etc, where some transaction can be successfully broadcasted to one node, but then not accepted by the network due to fee being to low. Here we assume that the min fee requirement is respected by the client, but if that is not true the transaction will be retried indefinitely.

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+  - user options
+  - context
+  - a result channel
+- The queue enforces memory based backpressure:
+  - If sufficient memory is available, the job is appended to the queue
+  - If queue exceeds its memory limit, submission blocks until memory is freed up
+ User should be able to set desired queue size during client initialization.
+- Nothing gets signed yet.
+- No network operations are performed.
+- User will receive tx result asynchronously via the result channel.
```
mcrakhman: As an alternative implementation we can have only one channel per signer that the broadcaster listens, but not sure if that would be easier or not.

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+When a user submits a transaction (e.g. via `SubmitPayForBlobToQueue`):
+
+- Tx is wrapped in a job containing:
+  - transaction data
+  - user options
+  - context
+  - a result channel
+- The queue enforces memory based backpressure:
+  - If sufficient memory is available, the job is appended to the queue
+  - If queue exceeds its memory limit, submission blocks until memory is freed up
```
mcrakhman: Should it block or return an error? Because if the nodes are slow or somehow the txs are not being confirmed then we can have this thing blocking indefinitely. As a compromise it can be that we can provide some timeout and this thing would try to block for this timeout and then return an error.

specs/src/tx_client_v2.md
ninabarbakadze: afaik @evan-forbes said that this client will be designed for a system that's not congested and we expect that txs will always be accepted by the node. 

I think If tx was initially accepted into the mempool unless he node changes the min fee in the config and restarts it will keep being accepted.

specs/src/tx_client_v2.md
ninabarbakadze: this queue is initialised per signer since all txs from the same signer need to be in order. if we had per signer channel how would you store txs?

specs/src/tx_client_v2.md
ninabarbakadze: I'm not opinionated on this. Whatever we think is the most user friendly will work for me.

specs/src/tx_client_v2.md
ninabarbakadze: I'm still not understanding what optimistic confirmations really save us. It will take the same amout of time for future tx to be confirmed and batch tx status queries are only a few milliseconds. 

We could benchmark and see if this really makes a difference.

specs/src/tx_client_v2.md
ninabarbakadze: > I would think that the number of loops should be bounded to some fixed number, because the number of signers can be arbitrary high.

I doubt that number of signers actively submitting transactions simultaniously would be arbitrarily high but it doesn't hurt to cap number of workers anyways

specs/src/tx_client_v2.md
mcrakhman: I think we should just mention that we assume this and that's it, because otherwise it may seem that we didn't think about it when taking the decision, and also future us may forget about it

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+- Tx hash, sequence, and signed bytes are recorded directly on the queued tx.
+- Tx is now considered pending.
+
+If broadcast fails due to a sequence mismatch:
+
+- Submitter pauses further submissions.
+- Job remains in the queue.
+- Waits for confirmation/recovery logic to unblock submissions.
+
+Sequence mismatches during broadcast are treated as coordination signals not terminal errors. If client remains blocked for longer than a minute, then it's possible to parse expected sequence from the error message and resign as recovery mech.
```
evan-forbes: > then it's possible to parse expected sequence from the error message and resign as recovery

we might want to be a bit clearer here given the consequences of resigning.

should we ever resign in v2?

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Context & Motivation
+
+Tx Client v2 enforces a per-account sequential submission queue that serializes signing, submission, and recovery, avoiding sequence races and minimizing transaction failures.
+
+## Protocol/Component Description
+
+### Per account sequential queue
+
+Every account gets its own **transaction queue**.
```
evan-forbes: :+1:

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+- Tx is re-broadcast using the same signed tx bytes.
+- Tx is **never** resigned.
+- Resubmitted tx remains as pending in queue.
+- Recovery mode is exited.
+- Broadcasting unblocks.
+
+1. Rejected
+
+Sequence recovery:
+A rejected transaction means the node will be expecting a lower sequence for this account. To unblock submissions the client must update its local sequence to match the node.
```
evan-forbes: > A rejected transaction means the node will be expecting a lower sequence for this account. 

is this true? my understanding is that the new strategy is simply resubmit, which we should be able to do due to the queue

specs/src/tx_client_v2.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Tx is removed from the queue.
+- An error is returned to the user via the result channel.
+
+### Phase 2: Confirmation
+
+Confirmation runs asynchronously at a configurable polling interval.
+
+The monitor periodically queries the chain for the status of broadcast transactions, performing status checks in bounded batches starting from the first pending transaction.
+
```
evan-forbes: it might be worth clarifying if v2 expects to get all of the updates at once (via the new TxStatus rpc method).

this can dramatically simply the logic in that we can be confident that we know which items in the queue need to be resubmitted

evan-forbes (commented):
nice!

left a few blocking questions that after we answer can proceed

specs/src/tx_client_v2.md
ninabarbakadze: yeah I also have conflicting feelings about this. technically it should never stall but the tx client soundness depends on our assumptions on how the mempool works. in the past we have missed things and it's possible we have regressions/miss edge cases again. 

I think we could have a function on the client that resets the sequence in case it does get stuck and the users wants to reset the state. It would be a recovery mechanism so should not cause races.

specs/src/tx_client_v2.md
ninabarbakadze: Yes it is true. We can still have rejected transactions in ReCheck e.g user ran out of funds.

specs/src/tx_client_v2.md
ninabarbakadze: I will add.

specs/src/tx_client_v2.md
ninabarbakadze: btw, are you okay with us moving this to the ADR section? Or should we make this spec listed on the hosted specs.

specs/src/tx_client_v2.md
ninabarbakadze: I haven't edited the spec with this because I think we should have more discussions on this and agree on how to move forward.

specs/src/tx_client_v2.md
ninabarbakadze: I'm not sure I understand this fully but I don't want to have design suggestions go unnoticed. If you think this would be a better design, would you mind elaborating a little further?

specs/src/tx_client_v2.md
ninabarbakadze: addressed https://github.com/celestiaorg/celestia-app/pull/6313/commits/a6b61dd6eea295af8833ce8e6cdbb58ff2d857ac

specs/src/tx_client_v2.md
ninabarbakadze: I'm waiting on Evan's opinion on this. Will address and link a commit.

specs/src/tx_client_v2.md
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6313/commits/a6b61dd6eea295af8833ce8e6cdbb58ff2d857ac

specs/src/tx_client_v2.md
ninabarbakadze: addressed https://github.com/celestiaorg/celestia-app/pull/6313/commits/a6b61dd6eea295af8833ce8e6cdbb58ff2d857ac

specs/src/tx_client_v2.md
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6313/commits/a6b61dd6eea295af8833ce8e6cdbb58ff2d857ac

specs/src/tx_client_v2.md
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6313/commits/a6b61dd6eea295af8833ce8e6cdbb58ff2d857ac

specs/src/tx_client_v2.md
mcrakhman: @evan-forbes wdyt? My impression was that we can actually have a single call to one node with all the signers/sequences there (we don't have to do it, but at least we would have this option)

specs/src/tx_client_v2.md
mcrakhman: So, I started thinking about this:

First let's start with a single signer situation. 

Suppose we have a client (who uses our API) who submits many txs and waits for them. Because a client doesn't want to block the broadcast, it awaits in the separate thread/routine. Therefore he should probably wait for each of the channels in sequence, because a higher sequence can't be approved before lower sequence (e.g. have a loop where he waits for ←ch1, ←ch2 etc). That is exactly the same as having only one channel per signer, that's why I proposed that.

But then it actually seems that the story becomes different when we have multiple signers. In both of these scenarios a client submitting txs for multiple signers I assume doesn't want to wait in different routines for different signers. 

_Therefore we should maybe even have one channel for all signers where we send the latest batch approved for a signer as soon as it has been approved_.

Tbh, let's discuss this on a call maybe

specs/src/tx_client_v2.md
mcrakhman: Ok, added this as FLUP as discussed https://github.com/celestiaorg/celestia-app/issues/6328, added more suggestions

ninabarbakadze:
Addressed feedback from yesterday's call @evan-forbes @mcrakhman Let me know if you want anything simplified further.

PR #6323, opened 2025-12-17
refactor: use collections pair prefix to namespace msg ids by ism

damiannolan:
## Overview

Adds namespacing to message ids per ism by adopting a `collections.Pair` as a key for the keyset.
The pair is a tuple containing the internal ism identifier (`uint64`). This provides domain separation across the keyspace for message ids per ism.


jonas089:
@damiannolan I think we should merge the other 2 PRs and create a new tag + redeploy the testnet. This PR is good, but it can wait until the next release imo. Lmk what you think :).

PR #6360, opened 2026-01-06
chore: remove parse duration dependency

rootulp:
Closes https://github.com/celestiaorg/celestia-app/security/dependabot/112

This pull request contains changes generated by a Cursor Cloud Agent

<a href="https://cursor.com/background-agent?bcId=bc-d4cdf7f6-6694-49b4-8209-af353eaffd9a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-d4cdf7f6-6694-49b4-8209-af353eaffd9a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

PR #6362, opened 2026-01-07
feat: add ci for latency monitor rust

mcrakhman:
## Overview

Latency monitor ci for rust

### DIFF
--- a/tools/lumina-latency-monitor/Cargo.toml
+++ b/tools/lumina-latency-monitor/Cargo.toml
@@ -17,7 +17,6 @@ anyhow = "1.0"
 clap = { version = "4.5", features = ["derive", "env"] }
 csv = "1.3"
 humantime = "2.1"
-parse_duration = "2"
 thiserror = "2.0"
 
 # Keyring support
--- a/tools/lumina-latency-monitor/src/config.rs
+++ b/tools/lumina-latency-monitor/src/config.rs
@@ -104,7 +104,7 @@ pub struct ValidatedConfig {
 }
 
 pub fn validate_args(args: &Args) -> Result<ValidatedConfig> {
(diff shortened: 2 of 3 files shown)


tools/lumina-latency-monitor/Cargo.toml
```diff
@@ -17,6 +17,7 @@ anyhow = "1.0"
 clap = { version = "4.5", features = ["derive", "env"] }
 csv = "1.3"
 humantime = "2.1"
+parse_duration = "2"
```
rootulp: I think this should be removed per https://github.com/celestiaorg/celestia-app/pull/6360

```suggestion
```

tools/lumina-latency-monitor/src/config.rs
```diff
@@ -107,7 +104,7 @@ pub struct ValidatedConfig {
 }
 
 pub fn validate_args(args: &Args) -> Result<ValidatedConfig> {
-    let submission_delay = humantime::parse_duration(&args.submission_delay)
+    let submission_delay = parse_duration::parse(&args.submission_delay)
```
rootulp: I think this should be reverted so `parse_duration` isn't added as a dependency again. See https://github.com/celestiaorg/celestia-app/pull/6360 for motivation on why I removed it

```suggestion
    let submission_delay = humantime::parse_duration(&args.submission_delay)
```

rootulp (commented):
LGTM after we remove parse_duration so that we don't reintroduce the dependency that caused https://github.com/celestiaorg/celestia-app/security/dependabot/112

tools/lumina-latency-monitor/src/config.rs
mcrakhman: sorry missed it, yeah ofc will fix

PR #6363, opened 2026-01-07
refactor: cache blob hashes in ValidateBlobTx

ninabarbakadze:
## Overview

This PR adds functionality for caching blob hashes in during CheckTx

### DIFF
--- a/app/test/process_proposal_test.go
+++ b/app/test/process_proposal_test.go
@@ -339,22 +339,22 @@ func TestProcessProposal_ProposalWithInconsistentBlobTxFails(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, uint32(0), checkTxResp.Code, "CheckTx should pass: %s", checkTxResp.Log)
 
-	t.Run("Tampered blob proposal fails", func(t *testing.T) {
+	t.Run("Proposal with inconsistent blob tx fails", func(t *testing.T) {
 		// replace the blob with a different one (same namespace, different data)
-		tamperedBlob, err := share.NewBlob(ns, bytes.Repeat([]byte{0xDE, 0xAD}, 50), share.ShareVersionZero, nil)
+		inconsistentBlob, err := share.NewBlob(ns, bytes.Repeat([]byte{0xDE, 0xAD}, 50), share.ShareVersionZero, nil)
 		require.NoError(t, err)
-		tamperedBlobTxBytes, err := tx.MarshalBlobTx(blobTx.Tx, tamperedBlob)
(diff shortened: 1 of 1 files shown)


mcrakhman (approved):

app/tx_cache.go
```diff
@@ -23,17 +25,40 @@ func (c *TxCache) getTxKey(tx []byte) string {
...
+		return false
+	}
+
+	blobHash := c.getBlobsHash(blobs)
+	return cachedBlobHash == blobHash
 }
 
 // Set stores the Tx in the cache
-func (c *TxCache) Set(tx []byte) {
+func (c *TxCache) Set(tx []byte, blobs []*share.Blob) {
```
rootulp, line 46: [question] can this tx cache be used for txs that don't contain blobs? If yes, when `Set` is called for a non PFB tx, is the second parameter to this function an empty slice?

app/tx_cache.go
```diff
@@ -23,17 +25,40 @@ func (c *TxCache) getTxKey(tx []byte) string {
...
+	c.cache.Store(key, blobsHash)
+}
+
+func (c *TxCache) getBlobsHash(blobs []*share.Blob) string {
+	h := sha256.New()
+	for _, blob := range blobs {
+		h.Write(blob.Namespace().Bytes())
+		h.Write(blob.Data())
+		h.Write([]byte{blob.ShareVersion()})
+		h.Write(blob.Signer())
```
rootulp, line 58: [no change needed] these lines may become stale if the share.Blob type adds a new field. And then the hash may not be dependent on all the new fields in share.Blob.

An alternative I considered was:

```go
func (c *TxCache) getBlobsHash(blobs []*share.Blob) string {
	h := sha256.New()
	for _, blob := range blobs {
		h.Write(blob.Hash())
	}
	sum := h.Sum(nil)
	return string(sum)
}
```

and then in go-square's blob.go, adding a method:

```go
func (b *Blob) Hash() []byte {
	h := sha256.New()
	h.Write(b.Namespace().Bytes())
	h.Write(b.Data())
	h.Write([]byte{b.ShareVersion()})
	h.Write(b.Signer())
	return h.Sum(nil)
}
```

which is more likely to get updated if we add a field to the Blob struct. But this refactor requires making a change in go-square, creating a release, and bumping here so not strictly required.

app/tx_cache.go
ninabarbakadze: the cache is used to skip commitment verification and if i'm not wrong we only generate commitments for blob shares

app/tx_cache.go
ninabarbakadze: I like this. We can do it as a flup.

PR #6375, opened 2026-01-11
chore: rename test cases for clarity

gap-editor:
**Test Case Duplication:**
There are two test cases named exactly **`"min gas price is empty"`**.
**Impact:** If the second one fails, your test output will show two identical names, making it harder to distinguish which logic failed (the "good tx" path or the "bad tx" path).


gap-editor:
if this pr is good let me know and i will add changes from #6374

Issue #6215, opened 2025-11-17
Cut v6.x long lived branch and bump repo to v7


cmwaters:
As the first part of preparation for V7 we need to update the go imports and multiplexer etc ready to introduce breaking changes (like fibre DA). We should also prior to that create a v6.x long lived branch for future minor changes that target the v6 lifecycle

rootulp:
Created https://github.com/celestiaorg/celestia-app/tree/v6.x

PR #6392, opened 2026-01-12
chore: prep for v7.x

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/6215

## Testing

```shell
./scripts/single-node.sh
celestia-appd status | jq .node_info.protocol_version.app
"7"
```

This one works too:
```shell
```

## FLUPs
- fix single-node-all-upgrades.sh
- fix e2e tests so they're not flaky

### DIFF
--- a/test/docker-e2e/go.mod
+++ b/test/docker-e2e/go.mod
@@ -269,7 +269,7 @@ replace (
 	cosmossdk.io/log => github.com/celestiaorg/cosmos-sdk/log v1.1.1-0.20251116153902-f48fea92e627
 	cosmossdk.io/x/upgrade => github.com/celestiaorg/cosmos-sdk/x/upgrade v0.2.0
 	github.com/celestiaorg/celestia-app/v7 => ../..
-  github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.22
+	github.com/cometbft/cometbft => github.com/celestiaorg/celestia-core v0.39.22
 	github.com/cosmos/cosmos-sdk => github.com/celestiaorg/cosmos-sdk v0.51.8
 	github.com/cosmos/ibc-go/v8 => github.com/celestiaorg/ibc-go/v8 v8.7.2
 	// Use ledger-cosmos-go v0.16.0 because v0.15.0 causes "hidapi: unknown failure"


app/upgrades_test.go
```diff
@@ -29,254 +22,7 @@ func TestUpgrades(t *testing.T) {
...
 		testApp := app.New(logger, db, traceStore, timeoutCommit, appOptions, baseapp.SetChainID(testfactory.ChainID))
 
-		require.False(t, testApp.UpgradeKeeper.HasHandler("v5"))
-		require.True(t, testApp.UpgradeKeeper.HasHandler("v6"))
+		require.False(t, testApp.UpgradeKeeper.HasHandler("v6"))
+		require.True(t, testApp.UpgradeKeeper.HasHandler("v7"))
 	})
 }
-
-func TestApplyUpgrade(t *testing.T) {
```
rach-id, line 37: why are these deleted?

scripts/single-node-upgrades.sh
```diff
@@ -79,8 +79,8 @@ createGenesis() {
     # Override the VotingPeriod from 1 week to 1 minute
     sed -i'.bak' 's#"604800s"#"60s"#g' "${APP_HOME}"/config/genesis.json
 
-    # Override the log level to debug
-    # sed -i'.bak' 's#log_level = "info"#log_level = "debug"#g' "${APP_HOME}"/config/config.toml
+    # Override the log level to reduce noisy logs
+    sed -i.bak 's#log_level = "info"#log_level = "*:error,p2p:info,state:info"#g' "${APP_HOME}"/config/config.toml
```
rach-id, line 83: how is this different than the info level?

app/upgrades_test.go
rootulp: this one was deleted b/c it tested certain param overrides that occur during the v6 upgrade handler. Those param overrides no longer need to happen in v7.

scripts/single-node-upgrades.sh
rootulp: it's info only for p2p2 and state module, otherwise it's just error. This helps filter out noisy logs from other modules

mcrakhman (approved):

PR #6403, opened 2026-01-14
feat: improve talis latency monitor   observability

ninabarbakadze:
## Overview

- Piped latency monitor logs from each validator to grafana. Can be filtered by validator.
- Added more metrics and extended dashboards in grafana to include:
    - Submission Rate (tx/s, 1m)  
    - CheckTx Latency (s, 1m)
- Added mempool debug metrics and dashboards which is optional and can't be viewed without server side changes.

Screenshot 2026-01-14 at 17 57 12
Screenshot 2026-01-14 at 17 57 35
Screenshot 2026-01-14 at 14 31 36

### DIFF
--- a/tools/talis/scripts/promtail.sh
+++ b/tools/talis/scripts/promtail.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export HOSTNAME=$(hostname)
+PROMTAIL_CONFIG=/root/promtail-config.yml
+printf "%s" "__PROMTAIL_CONFIG_B64__" | base64 -d > "$PROMTAIL_CONFIG"
+
+if ! command -v promtail >/dev/null 2>&1; then
+  arch=$(uname -m)
+  if [ "$arch" = "x86_64" ] || [ "$arch" = "amd64" ]; then arch=amd64;
+  elif [ "$arch" = "aarch64" ] || [ "$arch" = "arm64" ]; then arch=arm64;
+  else echo "unsupported arch: $arch" >&2; exit 1; fi
+  apt-get update -y >/dev/null
+  apt-get install -y curl unzip >/dev/null
+  tmpdir=$(mktemp -d)
+  curl -fsSL -o "$tmpdir/promtail.zip" "https://github.com/grafana/loki/releases/download/v2.9.3/promtail-linux-$arch.zip"
+  unzip -o "$tmpdir/promtail.zip" -d "$tmpdir" >/dev/null
+  install -m 0755 "$tmpdir/promtail-linux-$arch" /usr/local/bin/promtail
+fi
+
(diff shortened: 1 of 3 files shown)


metrics/docker/loki/loki-config.yml
```diff
@@ -0,0 +1,30 @@
+auth_enabled: false
```
rootulp, line 1: [no change needed] it looks like loki is for collecting logs. In a separate PR we may consider renaming `metrics/` to `observability` or `telemetry` because it looks like now the parent directory has more than just metrics, it has metrics and logs.

metrics/README.md
```diff
@@ -15,6 +15,11 @@ This package provides a simple metrics stack for Consensus nodes using Prometheu
                                          │   Grafana   │
                                          │ (port 3000) │
                                          └─────────────┘
+
+┌──────────────────────┐    push logs    ┌─────────────┐
+│ latency-monitor logs │ ──────────────► │    Loki     │
+└──────────────────────┘                 │ (port 3100) │
+                                         └─────────────┘
```
rootulp, line 22: Do the Loki logs get surfaced in Grafana? If yes then it looks like there's an arrow missing from Loki to Grafana above it

tools/talis/scripts/promtail.sh
```diff
@@ -0,0 +1,25 @@
...
+  curl -fsSL -o "$tmpdir/promtail.zip" "https://github.com/grafana/loki/releases/download/v2.9.3/promtail-linux-$arch.zip"
+  unzip -o "$tmpdir/promtail.zip" -d "$tmpdir" >/dev/null
+  install -m 0755 "$tmpdir/promtail-linux-$arch" /usr/local/bin/promtail
+fi
+
+promtail -config.file="$PROMTAIL_CONFIG" -config.expand-env -server.http-listen-port=9080 > /root/promtail.log 2>&1 &
+sleep 1
+pgrep -a promtail >/dev/null 2>&1 || (echo "promtail failed to start:" >&2; tail -200 /root/promtail.log >&2; exit 1)
+
+__LATENCY_MONITOR_CMD__
```
rach-id, line 25: why the `__`?

rootulp: I think the variables get interpolated later by Go code

metrics/README.md
ninabarbakadze: yes it does. I will add the arrow in a flup.

tools/talis/scripts/promtail.sh
rach-id: Yes, just wondering about the convention

rach-id: Yes, just wondering about the convention

ninabarbakadze: it is a dev made bash convention to visually signify that the value is not an env var and just a placeholder

PR #6417, opened 2026-01-16
feat(forwarding): 2/4 - Types (address, messages, genesis)

Manav-Aggarwal:
## Summary

Part 2 of 4 for the forwarding module. This PR adds type implementations:

- **Address derivation** - `DeriveForwardingAddress()` with deterministic address generation
- **Message types** - `MsgExecuteForwarding` and `MsgUpdateForwardingParams` with `ValidateBasic()`
- **Params** - Module parameters (only `MinForwardAmount` - TIA collateral token derived from warp module)
- **Genesis** - Genesis state types
- **Codec** - `RegisterLegacyAminoCodec` for Amino JSON serialization

## Files (~600 lines)

| File | Lines | Purpose |
|------|-------|---------|
| `address.go` | 46 | Address derivation algorithm |
| `address_test.go` | 188 | Address derivation tests |
| `params.go` | 30 | Module parameters (MinForwardAmount only) |
| `msgs.go` | 75 | Message types + URL constants |
| `msgs_test.go` | 150 | Message validation tests |
| `genesis.go` | 28 | Genesis state |
| `genesis_test.go` | 70 | Genesis validation tests |
| `codec.go` | 25 | RegisterInterfaces + RegisterLegacyAminoCodec |

### Design Decisions
- **No TiaCollateralTokenId in params**: Removed - TIA collateral token is discovered at runtime from warp module
- Params only contain `MinForwardAmount` (defaults to 0 = disabled)

## Review Focus

- **Security-critical**: Address derivation in `address.go`
- `ValidateBasic()` implementations for both message types
- Test coverage for edge cases

## Test Plan

- [x] Unit tests for address derivation
- [x] Unit tests for message validation
- [x] Unit tests for genesis validation
- [x] Module builds (`go build ./x/forwarding/...`)

## PR Chain

```
feature/forwarding (base)
└── feature/forwarding-1-proto (#6416)
    └── feature/forwarding-2-types (this PR)
        └── feature/forwarding-3-keeper (next)
            └── feature/forwarding-4-tests (pending)
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/x/forwarding/types/msgs_test.go
+++ b/x/forwarding/types/msgs_test.go
@@ -10,8 +10,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func init() {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount("celestia", "celestiapub")
+}
+
 func TestMsgForward_ValidateBasic(t *testing.T) {
-	// Generate valid addresses using SDK defaults (cosmos1 prefix)
 	validSignerBytes := []byte("testsigner__________")      // 20 bytes
 	validForwardAddrBytes := []byte("forwardaddr_________") // 20 bytes
 
@@ -162,7 +166,6 @@ func TestMsgForward_ValidateBasic(t *testing.T) {
 		})
 	}
 }
-
 
 func TestMsgUpdateParamsValidateBasic(t *testing.T) {
 	validAuthorityBytes := []byte("authority___________") // 20 bytes


x/forwarding/types/address.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// Returns an error if destRecipient is not exactly RecipientLength (32) bytes.
+func DeriveForwardingAddress(destDomain uint32, destRecipient []byte) ([]byte, error) {
+	if len(destRecipient) != RecipientLength {
+		return nil, fmt.Errorf("%w: expected %d bytes, got %d", ErrInvalidRecipient, RecipientLength, len(destRecipient))
+	}
+
+	// Step 1: Encode destDomain as 32-byte big-endian (right-aligned)
+	destDomainBytes := make([]byte, 32)
+	binary.BigEndian.PutUint32(destDomainBytes[28:], destDomain)
```
rach-id: would be nice to know what the magic numbers are

x/forwarding/types/address.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Step 2: callDigest = sha256(destDomain || destRecipient)
+	callDigestPreimage := make([]byte, 32+RecipientLength)
+	copy(callDigestPreimage, destDomainBytes)
+	copy(callDigestPreimage[32:], destRecipient)
+	callDigest := sha256.Sum256(callDigestPreimage)
```
rach-id: [nit]
we could use a hash writer to avoid the manual copies here: 
```
h := sha256.New()
h.Write(stuff)
```

same for the others

x/forwarding/types/address_test.go
```diff
@@ -0,0 +1,192 @@
...
+	t.Logf("salt (sha256): %s", hex.EncodeToString(salt))
+
+	// Step 5: Use SDK's address.Module for the final derivation
+	derivedAddr := address.Module(types.ModuleName, salt)[:20]
+	t.Logf("derived address (20 bytes): %s", hex.EncodeToString(derivedAddr))
+
+	// Verify this matches the function output
+	addr, err := types.DeriveForwardingAddress(destDomain, destRecipient)
+	require.NoError(t, err)
+	require.Equal(t, derivedAddr, addr, "manual derivation should match function output")
```
rach-id: to ensure we're testing the same logic that's in the method, and to avoid updating one and forgetting the other, we could split that method into smaller methods and use them here

x/forwarding/types/address_test.go
```diff
@@ -0,0 +1,192 @@
...
+	recipient2 := hexToBytes(t, "0000000000000000000000002222222222222222222222222222222222222222")
+	addr3, err := types.DeriveForwardingAddress(1, recipient1)
+	require.NoError(t, err)
+	addr4, err := types.DeriveForwardingAddress(1, recipient2)
+	require.NoError(t, err)
+	require.NotEqual(t, addr3, addr4, "different recipients should produce different addresses")
+}
+
+// TestDeriveForwardingAddressIntermediates verifies the intermediate values in derivation.
+// This test is crucial for TypeScript SDK cross-verification.
```
rach-id, line 86: Agree 👍 . We should have some test vectors using the reference implementation for each step and use them here too to double check

x/forwarding/types/address_test.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, tc := range testVectors {
+		t.Run(tc.name, func(t *testing.T) {
+			recipient := hexToBytes(t, tc.destRecipient)
+			addr, err := types.DeriveForwardingAddress(tc.destDomain, recipient)
+			require.NoError(t, err)
+
+			t.Logf("Test Vector: %s", tc.name)
+			t.Logf("  destDomain: %d", tc.destDomain)
+			t.Logf("  destRecipient: 0x%s", tc.destRecipient)
+			t.Logf("  derivedAddressHex: %s", hex.EncodeToString(addr))
```
rach-id: This doesn't seem to be comparing the test vectors, only logging them.

x/forwarding/types/msgs_test.go
```diff
@@ -0,0 +1,241 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+// Note: GetSigners() is inferred from proto annotation (cosmos.msg.v1.signer)
+// Route() and Type() methods are deprecated in newer Cosmos SDK versions
+// These methods are no longer required for sdk.Msg interface
```
rach-id: is this comment supposed to stay?

x/forwarding/types/address.go
Manav-Aggarwal: Done\! Added constants `DomainEncodingSize` (32) and `DomainOffset` (28) with documentation explaining they're for ABI-encoding domain IDs as uint256.

x/forwarding/types/address.go
Manav-Aggarwal: Good suggestion! Refactored to use `sha256.New()` with streaming writes instead of manual buffer copies.

x/forwarding/types/address_test.go
Manav-Aggarwal: The duplication here is intentional - the test re-implements the algorithm step-by-step to verify the main function against an independent implementation. If we extracted helpers and used them in both places, we'd be testing code against itself, which defeats the purpose of cross-verification. The test should fail if the main function's implementation diverges from the documented algorithm.

x/forwarding/types/address_test.go
Manav-Aggarwal: Added hardcoded expected addresses to the test vectors. These were generated by running the derivation and can be used to verify TypeScript SDK implementations.

x/forwarding/types/address_test.go
Manav-Aggarwal: Fixed! Added `expectedAddress` field to test vectors with assertions. The test now validates against expected values instead of just logging.

x/forwarding/types/msgs_test.go
Manav-Aggarwal: Keeping this comment - it documents why we don't test `GetSigners()`, `Route()`, and `Type()` methods. `GetSigners()` is auto-generated from the proto `cosmos.msg.v1.signer` annotation, and `Route()`/`Type()` are deprecated in newer Cosmos SDK versions. This context helps future reviewers understand the intentional absence of those tests.

x/forwarding/types/msgs_test.go
```diff
@@ -0,0 +1,241 @@
...
+	"testing"
+
+	"cosmossdk.io/math"
+	"github.com/celestiaorg/celestia-app/v7/x/forwarding/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMsgForward_ValidateBasic(t *testing.T) {
+	// Generate valid addresses using SDK defaults (cosmos1 prefix)
```
rootulp: Why use cosmos1 prefix instead of celestia prefix?

x/forwarding/types/msgs_test.go
rootulp: TBH I don't think future reviewers are going to even notice that there aren't tests for `GetSigners()`, `Route()` and `Type()`. I'm fine with removing the comment (or keeping). Not a big deal either way

```suggestion
```

x/forwarding/types/msgs_test.go
Manav-Aggarwal: Switched to celestia prefix for consistency - validation works either way.

PR #6418, opened 2026-01-16
feat(forwarding): 3/4 - Keeper + Module + App Wiring

Manav-Aggarwal:
## Summary

Part 3 of 4 for the forwarding module. This PR adds core implementation:

- **Keeper** - State management, business logic, and governance authority
- **MsgServer** - `ExecuteForwarding` + `UpdateForwardingParams` handlers
- **QueryServer** - `DeriveForwardingAddress` query
- **Genesis** - Import/export handlers
- **Module** - `AppModule` implementation
- **App Wiring** - Integration in `app.go` and `modules.go`

## Files (~800 lines)

| File | Lines | Purpose |
|------|-------|---------|
| `keeper/keeper.go` | 150 | Core keeper + TIA collateral token discovery from warp |
| `keeper/msg_server.go` | 200 | ExecuteForwarding + UpdateForwardingParams |
| `keeper/query_server.go` | 75 | DeriveForwardingAddress query |
| `keeper/genesis.go` | 24 | Genesis import/export |
| `module.go` | 107 | AppModule implementation |
| `app/app.go` | +15 | Keeper initialization |
| `app/modules.go` | +7 | Module registration |

### Key Implementation Details

**TIA Collateral Token Discovery**: Instead of storing `TiaCollateralTokenId` in params, the keeper discovers TIA collateral tokens at runtime by iterating warp tokens with `OriginDenom="utia"` and checking for enrolled routers to the destination domain.

```go
func (k Keeper) findTIACollateralTokenForDomain(ctx context.Context, destDomain uint32) (HypToken, error)
```

This approach:
- Removes configuration burden (no need to manually set token ID)
- Automatically works with any TIA collateral token that has the right route
- Fails cleanly if no route exists to the destination

## Review Focus

- **ExecuteForwarding logic** in `msg_server.go`:
  - Token flow: forwardAddr → module → warp
  - Error handling and recovery paths
- **TIA collateral discovery** in `keeper.go`
- **UpdateForwardingParams** governance authorization
- Query validation before returning derived addresses
- Module initialization order

## Test Plan

- [x] Module builds (`go build ./x/forwarding/...`)
- [ ] Integration tests (PR 4)

## PR Chain

```
feature/forwarding (base)
└── feature/forwarding-1-proto (#6416)
    └── feature/forwarding-2-types (#6417)
        └── feature/forwarding-3-keeper (this PR)
            └── feature/forwarding-4-tests (next)
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/x/forwarding/keeper/keeper.go
+++ b/x/forwarding/keeper/keeper.go
@@ -1,3 +1,5 @@
+// Package keeper implements the forwarding module keeper, which manages
+// automatic cross-chain token forwarding via Hyperlane warp routes.
 package keeper
 
 import (
@@ -16,6 +18,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+// Keeper manages forwarding module state and coordinates with Hyperlane warp for cross-chain transfers.
 type Keeper struct {
 	cdc             codec.BinaryCodec
 	Schema          collections.Schema
@@ -36,6 +39,12 @@ func NewKeeper(
 	hyperlaneKeeper types.HyperlaneKeeper,
 	authority string,
 ) Keeper {
+	if accountKeeper == nil {
+		panic("accountKeeper cannot be nil")
+	}
+	if bankKeeper == nil {
+		panic("bankKeeper cannot be nil")
+	}
 	if warpKeeper == nil {
 		panic("warpKeeper cannot be nil")
 	}
@@ -69,6 +78,9 @@ func (k Keeper) GetParams(ctx context.Context) (types.Params, error) {
 }
 
 func (k Keeper) SetParams(ctx context.Context, params types.Params) error {
+	if err := params.Validate(); err != nil {
+		return err
+	}
 	return k.Params.Set(ctx, params)
 }
 
@@ -80,7 +92,7 @@ func (k Keeper) FindHypTokenByDenom(ctx context.Context, denom string, destDomai
 	case denom == "utia":
 		return k.findTIACollateralTokenForDomain(ctx, destDomain)
 	case strings.HasPrefix(denom, "hyperlane/"):
-		return k.findSyntheticToken(ctx, strings.TrimPrefix(denom, "hyperlane/"))
+		return k.getTokenById(ctx, strings.TrimPrefix(denom, "hyperlane/"))
 	default:
 		return warptypes.HypToken{}, types.ErrUnsupportedToken
 	}
@@ -89,6 +101,9 @@ func (k Keeper) FindHypTokenByDenom(ctx context.Context, denom string, destDomai
 // findTIACollateralTokenForDomain finds the TIA collateral token with a route to the destination domain.
 // It iterates all warp tokens to find one with OriginDenom="utia", TokenType=COLLATERAL,
 // and an enrolled router for the destination domain.
+//
+// Note: This is O(n) where n = number of warp tokens. Optimization would require an index
+// in the warp keeper mapping (denom, destDomain) -> tokenId, which is outside this module's scope.
 func (k Keeper) findTIACollateralTokenForDomain(ctx context.Context, destDomain uint32) (warptypes.HypToken, error) {
 	iter, err := k.warpKeeper.HypTokens.Iterate(ctx, nil)
 	if err != nil {
@@ -112,10 +127,6 @@ func (k Keeper) findTIACollateralTokenForDomain(ctx context.Context, destDomain 
 	return warptypes.HypToken{}, fmt.Errorf("%w: no TIA collateral route to domain %d", types.ErrNoWarpRoute, destDomain)
 }
 
-func (k Keeper) findSyntheticToken(ctx context.Context, tokenIdHex string) (warptypes.HypToken, error) {
-	return k.getTokenById(ctx, tokenIdHex)
-}
-
 func (k Keeper) getTokenById(ctx context.Context, tokenIdHex string) (warptypes.HypToken, error) {
 	tokenId, err := util.DecodeHexAddress(tokenIdHex)
 	if err != nil {
@@ -132,6 +143,9 @@ func (k Keeper) HasEnrolledRouter(ctx context.Context, tokenId util.HexAddress, 
 	return k.warpKeeper.EnrolledRouters.Has(ctx, collections.Join(tokenId.GetInternalId(), destDomain))
 }
 
+// ExecuteWarpTransfer executes a Hyperlane warp transfer using the pre-computed IGP fee.
+// The quotedFee must be provided by the caller (collected from the relayer in msg_server).
+// This ensures only relayer-provided funds are used for IGP fees (no module-paid fallback).
 func (k Keeper) ExecuteWarpTransfer(
 	ctx sdk.Context,
 	token warptypes.HypToken,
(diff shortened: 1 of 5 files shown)


x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+	return warptypes.HypToken{}, fmt.Errorf("%w: no TIA collateral route to domain %d", types.ErrNoWarpRoute, destDomain)
+}
+
+func (k Keeper) findSyntheticToken(ctx context.Context, tokenIdHex string) (warptypes.HypToken, error) {
+	return k.getTokenByHex(ctx, tokenIdHex)
+}
+
+func (k Keeper) getTokenByHex(ctx context.Context, tokenIdHex string) (warptypes.HypToken, error) {
```
rach-id: [nit]
Probably `getTokenById` is better

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,199 @@
...
+	if !forwardAddr.Equals(sdk.AccAddress(expectedAddr)) {
+		return nil, fmt.Errorf("%w: provided=%s derived=%s", types.ErrAddressMismatch, forwardAddr.String(), sdk.AccAddress(expectedAddr).String())
+	}
+
+	balances := m.k.bankKeeper.GetAllBalances(ctx, forwardAddr)
+	if balances.IsZero() {
+		return nil, types.ErrNoBalance
+	}
+
+	if len(balances) > types.MaxTokensPerForward {
```
rach-id, line 54: if some address has too many tokens, how to reduce them to the max?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, types.ErrTooManyTokens
+	}
+
+	moduleAddr := m.k.accountKeeper.GetModuleAddress(types.ModuleName)
+
+	params, err := m.k.GetParams(ctx)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			ctx.Logger().Warn("forwarding module params not configured, using defaults - TIA forwarding will fail until params are set")
```
rach-id, line 63: 0 means disabled, right? where are we failing when we encounter that?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	if err := m.k.bankKeeper.SendCoins(ctx, forwardAddr, moduleAddr, sdk.NewCoins(balance)); err != nil {
+		return types.NewFailureResult(balance.Denom, balance.Amount, "failed to move tokens: "+err.Error())
+	}
+
+	messageId, err := m.k.ExecuteWarpTransfer(ctx, hypToken, moduleAddr.String(), destDomain, destRecipient, balance.Amount)
+	if err != nil {
+		if recoveryErr := m.k.bankKeeper.SendCoins(ctx, moduleAddr, forwardAddr, sdk.NewCoins(balance)); recoveryErr != nil {
+			ctx.Logger().Error("CRITICAL: tokens stuck in module account after failed recovery",
```
rach-id, line 207: how to recover from this?

x/forwarding/keeper/query_server.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if len(destRecipient.Bytes()) != types.RecipientLength {
+		return nil, status.Errorf(codes.InvalidArgument, "dest_recipient must be %d bytes, got %d", types.RecipientLength, len(destRecipient.Bytes()))
+	}
+
+	// Check if there's a TIA warp route to the destination domain
+	_, err = q.k.findTIACollateralTokenForDomain(ctx, req.DestDomain)
+	if err != nil {
+		return nil, status.Errorf(codes.FailedPrecondition, "no warp route to domain %d: %v", req.DestDomain, err)
+	}
```
rach-id: This query seems to only want the derived address, why is it checking if the route exists?

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,148 @@
...
+func (k Keeper) ExecuteWarpTransfer(
+	ctx sdk.Context,
+	token warptypes.HypToken,
+	sender string,
+	destDomain uint32,
+	destRecipient util.HexAddress,
+	amount math.Int,
+) (util.HexAddress, error) {
+	gasLimit := math.ZeroInt()
+	maxFee := sdk.NewCoin("utia", math.ZeroInt())
```
rootulp: I expect this to fail so maybe this should have a // TODO because it needs to be updated later, right?

x/forwarding/keeper/msg_server.go
rootulp: +1 if an address has 21 tokens, we should process the MaxTokensPerForward as opposed to error'ing here.

If we error here then tokens are permanently locked in the forward address. If we process in batches of 20, a user can still get their funds out.

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,199 @@
...
+		if !hasRoute {
+			return types.NewFailureResult(balance.Denom, balance.Amount, "no warp route to destination domain")
+		}
+	}
+
+	if params.MinForwardAmount.IsPositive() && balance.Amount.LT(params.MinForwardAmount) {
+		return types.NewFailureResult(balance.Denom, balance.Amount, "below minimum forward amount")
+	}
+
+	if err := m.k.bankKeeper.SendCoins(ctx, forwardAddr, moduleAddr, sdk.NewCoins(balance)); err != nil {
```
rootulp, line 191: [question] I'm a little confused because I assumed tokens would be sent directly from the forward address to the destination chain + destination address.

Why do tokens need to move from the forward address to the module account address?

x/forwarding/keeper/query_server.go
rootulp: I think it's a good idea to check the route exists because if it doesn't then the user could send funds to an address and the relayer can't forward them b/c no connection exists to the destination.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Done. Renamed to `getTokenById`.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Changed to truncate to first MaxTokensPerForward tokens instead of erroring. User can call Forward again to process remaining tokens.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Added comment clarifying that 0 means disabled. `IsPositive()` returns false for zero, so the check is skipped.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Added a "Recovery from Stuck Tokens" section to SPEC.md explaining the recovery process: create the warp route, then call Forward again.

x/forwarding/keeper/query_server.go
Manav-Aggarwal: Added comment explaining the route check. It prevents users from getting addresses for non-existent routes, which would lead to stuck funds until a route is created.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Implemented fee quoting. Added `HyperlaneKeeper` dependency and now call `QuoteDispatch` to get the required fee. Module account pays from its pre-funded balance.

EDIT: Relayer pays to the module account.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Added comment explaining why we use moduleAddr. Warp's IGP charges fees from the sender address, so using moduleAddr lets the module pay fees from its pre-funded balance rather than from forwarded tokens.

EDIT: Modified this so that the Relay actually pays for this by transferring TIA to the module address instead of needing to pre-fund the module address.

PR #6419, opened 2026-01-16
feat(forwarding): 4/4 - Integration Tests

Manav-Aggarwal:
## Summary

Part 4 of 4 (final) for the forwarding module. This PR adds integration tests.

## Files (~960 lines)

| File | Lines | Purpose |
|------|-------|---------|
| `test/interop/forwarding_integration_test.go` | 961 | Integration test suite |
| `x/forwarding/keeper/keeper_test.go` | 35 | Unit test for address derivation |

## Review Focus

- Test coverage for all code paths
- Mock setup for warp keeper
- Error case handling
- TIA collateral token discovery testing (no TiaCollateralTokenId in params)

## PR Chain (Complete)

```
feature/forwarding (base)
└── #6416 - Proto + Foundation + Docs
    └── #6417 - Types
        └── #6418 - Keeper + Module
            └── this PR - Integration Tests
```

**Depends on:** #6418

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #6416, opened 2026-01-16
feat(forwarding): 1/4 - Proto + Foundation + Documentation

Manav-Aggarwal:
## Summary

This is **Part 1 of 4** PRs for the forwarding module implementation, split for easier review.

Start by reviewing the `README.md` file to understand the design of the forwarding module.

### What's in this PR:

**Proto Definitions (~250 lines)**
- `event.proto`: Typed events for token forwarding
- `genesis.proto`: Genesis state
- `params.proto`: Module parameters (`min_forward_amount` only - TIA collateral token is derived from warp module)
- `query.proto`: DeriveForwardingAddress query
- `tx.proto`: MsgExecuteForwarding + MsgUpdateForwardingParams (governance)

**Foundation Types (~100 lines)**
- `keys.go`: Store prefixes, module constants, event types (typed events only)
- `errors.go`: Error definitions (codes 2-8)
- `codec.go`: Interface registration (RegisterInterfaces only)
- `expected_keepers.go`: Keeper interfaces

**Documentation (~850 lines)**
- `README.md`: Module overview, design, and error codes table
- `CLAUDE.md`: Agent context for code review
- `RELAYER.md`: Relayer implementation guide

### Design Decisions
- **No TiaCollateralTokenId in params**: TIA collateral token is discovered at runtime by iterating warp tokens with `OriginDenom="utia"` and checking for routes to the destination domain
- `MsgUpdateForwardingParams` for governance param updates
- Typed events only (no legacy attributes)

### PR Stack

| PR | Focus | Status |
|----|-------|--------|
| **#6416** | **Proto + Foundation + Docs** | **This PR** |
| #6417 | Types (address, msgs, genesis) | Next |
| #6418 | Keeper + Module | Pending |
| #6419 | Integration Tests | Pending |

### Review Focus

- Does the design make sense?
- Is the proto API clean?
- Is the documentation complete?

## Test plan
- [x] Proto files regenerate successfully (`make proto-gen`)
- [x] Module builds (`go build ./x/forwarding/...`)

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #6427, opened 2026-01-19
feat(forwarding): implement x/forwarding module (CIP-45)

Manav-Aggarwal:
## Summary

Implements [CIP-45: Forwarding Module](https://github.com/celestiaorg/CIPs/pull/373/) - enables single-signature cross-chain transfers through Celestia via Hyperlane warp routes. Users send tokens to a deterministically derived forwarding address, and anyone can permissionlessly trigger forwarding to the committed destination.

Consolidates the following PRs:
https://github.com/celestiaorg/celestia-app/pull/6416
https://github.com/celestiaorg/celestia-app/pull/6417
https://github.com/celestiaorg/celestia-app/pull/6418 
https://github.com/celestiaorg/celestia-app/pull/6419

### What this enables
- **Single signature UX**: User signs once on source chain
- **CEX compatible**: Works with exchange withdrawals (no smart contract interaction needed)
- **Permissionless execution**: Anyone can trigger forwarding with correct parameters
- **Cryptographic binding**: Funds can only go to the committed destination

### Components Added

| Component | Description |
|-----------|-------------|
| `x/forwarding/` | New Cosmos SDK module |
| Proto definitions | Messages, queries, events, genesis |
| Types | Address derivation, message validation |
| Keeper | State management, warp token discovery |
| MsgServer | `MsgForward`, `MsgUpdateParams` |
| QueryServer | `DeriveForwardingAddress`, `QuoteForwardingFee` |
| CLI | Query and tx commands |
| Documentation | SPEC.md, RELAYER.md, CLAUDE.md |

### Key Design Decisions

**Deterministic address derivation**: Forwarding addresses are cryptographically derived from `(destDomain, destRecipient)`:
```
callDigest   = sha256(destDomain_32bytes || destRecipient)
salt         = sha256(version_byte || callDigest)
forwardAddr  = address.Module("forwarding", salt)[:20]
```

**Runtime TIA collateral discovery**: Instead of storing token IDs in params, the keeper discovers TIA collateral tokens at runtime by iterating warp tokens with `OriginDenom="utia"` and checking for routes to the destination domain. This removes configuration burden and automatically works with any properly routed TIA collateral token.

**Multi-token support with limits**: Forwards ALL token balances at an address. Partial failures allowed - one token failing doesn't block others. Max 20 tokens per call to bound gas consumption.

**Relayer-paid IGP fees**: The relayer (signer) pays both Celestia gas and Hyperlane IGP fees. Fee is charged per-token, only actual quoted fee collected (not max).

**Graceful failure handling**: Failed tokens stay at `forwardAddr` for retry. No fund loss - if a warp route doesn't exist, tokens remain recoverable once the route is created.

**Security model**: 160-bit collision resistance (same as standard Cosmos addresses). Funds cryptographically bound to committed destination - permissionless execution cannot redirect funds.

## Test Plan
- [x] Module builds successfully
- [x] Integration tests 

## Related
- CIP-43: https://github.com/celestiaorg/CIPs/pull/369/

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-app/pull/6427">
  <picture>
    Open with Devin
  </picture>
</a>


test/interop/forwarding_integration_test.go
```diff
@@ -0,0 +1,1003 @@
...
+
+	s.T().Logf("Test 2 PASSED: FindHypTokenByDenom works for TIA collateral")
+	s.T().Logf("Token type: %v", hypToken.TokenType)
+	s.T().Logf("Origin denom: %s", hypToken.OriginDenom)
+}
+
+func (s *ForwardingIntegrationTestSuite) TestFindHypTokenByDenom_Synthetic() {
+	const (
+		CelestiaDomainID = 69420
+		SimappDomainID   = 1337
```
rootulp: I thought this was refactored to use chainA and chainB not Simapp

x/forwarding/client/cli/tx.go
```diff
@@ -0,0 +1,88 @@
...
+				return fmt.Errorf("invalid dest_domain: %w", err)
+			}
+
+			maxIgpFeeStr, err := cmd.Flags().GetString("max-igp-fee")
+			if err != nil {
+				return err
+			}
+			if maxIgpFeeStr == "" {
+				return fmt.Errorf("--max-igp-fee is required")
+			}
```
rootulp: one quality of life improvement is to make `--max-igp-fee` not required. If `--max-igp-fee` is not provided, this code can query the IGP fee by using `query forwarding quote-fee` and use that as the default value

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,224 @@
...
+		warpKeeper:      warpKeeper,
+		hyperlaneKeeper: hyperlaneKeeper,
+		authority:       authority,
+	}
+
+	schema, err := sb.Build()
+	if err != nil {
+		panic(err)
+	}
+	k.Schema = schema
```
rootulp: [question] is there a reason this is done after the `k := Keeper{}` definition?

It seems possible to re-arrange so the schema is generated before and then provided when we initialize Keeper:

```diff
	schema, err := sb.Build()
	if err != nil {
		panic(err)
	}

	k := Keeper{
		cdc:             cdc,
		Params:          collections.NewItem(sb, types.ParamsPrefix, "params", codec.CollValue[types.Params](cdc)),
		accountKeeper:   accountKeeper,
		bankKeeper:      bankKeeper,
		warpKeeper:      warpKeeper,
		hyperlaneKeeper: hyperlaneKeeper,
		authority:       authority,
+		Schema: schema,
	}
```

but if the ordering here is important then maybe we should add a comment that the schema needs to be built _after_ the Keeper is created? Seems like a code smell

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	return k.Params.Set(ctx, params)
+}
+
+// FindHypTokenByDenom finds the HypToken for a given bank denom and destination domain.
+// For TIA (utia), it searches warp tokens with OriginDenom="utia" that have a route to destDomain.
+// For synthetic tokens (hyperlane/...), it looks up the token by ID.
+func (k Keeper) FindHypTokenByDenom(ctx context.Context, denom string, destDomain uint32) (warptypes.HypToken, error) {
+	switch {
+	case denom == "utia":
```
rootulp: [nit] can we use appconsts.BondDenom here?


```suggestion
	case denom == appconsts.BondDenom:
```

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	defer iter.Close()
+
+	for ; iter.Valid(); iter.Next() {
+		token, err := iter.Value()
+		if err != nil {
+			continue
+		}
+		// Find TIA collateral token with route to destination
+		if token.OriginDenom == "utia" && token.TokenType == warptypes.HYP_TOKEN_TYPE_COLLATERAL {
```
rootulp:
```suggestion
		if token.OriginDenom == appconsts.BondDenom && token.TokenType == warptypes.HYP_TOKEN_TYPE_COLLATERAL {
```

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,224 @@
...
+	// Multi-denom IGP fees not supported
+	if len(quotedFee) > 1 {
+		return sdk.Coin{}, fmt.Errorf("multi-denom IGP fees not supported")
+	}
+
+	// Use the first coin from quoted fee, or zero if no fee required
+	if len(quotedFee) > 0 {
+		return quotedFee[0], nil
+	}
+	return sdk.NewCoin("utia", math.ZeroInt()), nil
```
rootulp:
```suggestion
	return sdk.NewCoin(appconsts.BondDenom, math.ZeroInt()), nil
```

more generally search for all instances of "utia" in this PR and replace with appconsts.BondDenom

x/forwarding/keeper/keeper_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/celestiaorg/celestia-app/v7/x/forwarding/types"
+	"github.com/stretchr/testify/require"
+)
+
+// NOTE: Keeper tests require a non-nil warpKeeper which can't be easily mocked.
+// Full keeper tests are in the integration test suite at test/interop/forwarding_integration_test.go
+
+// TestDeriveForwardingAddress verifies address derivation.
+// The derivation is done via types.DeriveForwardingAddress, not a keeper method.
```
rootulp: Why is this test located in keeper_test.go? Can it be moved to the `_test.go` file where `DeriveForwardingAddress` is defined?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,259 @@
...
+		balances = balances[:types.MaxTokensPerForward]
+	}
+
+	moduleAddr := m.k.accountKeeper.GetModuleAddress(types.ModuleName)
+
+	params, err := m.k.GetParams(ctx)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			ctx.Logger().Warn("forwarding module params not configured, using defaults - TIA forwarding will fail until params are set")
+			params = types.DefaultParams()
```
rootulp: [question] why do we need this branch of the error handling? I don't expect it to be hit ever. If it's never going to be hit, can we simplify to:

```go
	params, err := m.k.GetParams(ctx)
	if err != nil {
			return nil, fmt.Errorf("failed to read module params: %w", err)
	}
```

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,259 @@
...
+		// Try to return IGP fee to relayer if token move failed
+		if quotedFee.IsPositive() {
+			if recoveryErr := m.k.bankKeeper.SendCoins(ctx, moduleAddr, signerAddr, sdk.NewCoins(quotedFee)); recoveryErr != nil {
+				// TODO(v2): Consider emitting EventIgpFeeStuck for better observability
+				ctx.Logger().Error("failed to return IGP fee to relayer after token move failure",
+					"error", recoveryErr, "fee", quotedFee)
+			}
+		}
+		return types.NewFailureResult(balance.Denom, balance.Amount, "failed to move tokens: "+err.Error())
+	}
```
rootulp: [auestion] the error handling here seems complex. Are there any risks that an attacker can drain the module account of escrowed IGP fees?

x/forwarding/keeper/query_server.go
```diff
@@ -0,0 +1,93 @@
...
+	}, nil
+}
+
+// Params returns the module parameters
+func (q queryServer) Params(ctx context.Context, req *types.QueryParamsRequest) (*types.QueryParamsResponse, error) {
+	params, err := q.k.GetParams(ctx)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			params = types.DefaultParams()
+		} else {
```
rootulp: when will this happen? I expect the the params to be set to the default params as soon as the module gets added so this error handling seems unnecessary

x/forwarding/types/address_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+			// Verify address is 20 bytes
+			require.Len(t, addr, 20, "derived address should be 20 bytes")
+
+			// Verify determinism - same inputs produce same output
+			addr2, err := types.DeriveForwardingAddress(tc.destDomain, tc.destRecipient)
+			require.NoError(t, err)
+			require.Equal(t, addr, addr2, "derivation should be deterministic")
+
+			// Log for debugging and TypeScript SDK cross-verification
```
rootulp: there aren't any Typescript tests here so may be removable

```suggestion
			// Log for debugging
```

x/forwarding/types/params.go
```diff
@@ -0,0 +1,30 @@
...
+func NewParams(minForwardAmount math.Int) Params {
+	return Params{
+		MinForwardAmount: minForwardAmount,
+	}
+}
+
+// Validate validates the set of params
+func (p Params) Validate() error {
+	if p.MinForwardAmount.IsNegative() {
+		return ErrBelowMinimum
```
rootulp: [blocking] this is the wrong error to return here

x/forwarding/CLAUDE.md
```diff
@@ -0,0 +1,52 @@
+# CLAUDE.md - Forwarding Module
```
rootulp, line 1: [question] idk if this should be included in version control long-term. Not blocking on including it now but we should decide as a team if we want CLAUDE.md docs in multiple directories.

x/forwarding/RELAYER.md
```diff
@@ -0,0 +1,344 @@
...
+### IGP Fee Edge Cases
+
+| Scenario | Fee Status | Tokens Status | Relayer Action |
+|----------|------------|---------------|----------------|
+| Warp succeeds | Consumed | Forwarded | None |
+| Warp fails (e.g., route removed mid-tx) | **Consumed** | Returned to `forwardAddr` | Retry after fixing route |
+| IGP fee denom mismatch | Not collected | Unchanged | Fix `max_igp_fee` denom |
+| Insufficient max_igp_fee | Not collected | Unchanged | Increase `max_igp_fee` |
+| Relayer has insufficient balance | Not collected | Unchanged | Fund relayer account |
+| Recovery fails (CRITICAL) | Consumed | **Stuck in module** | Contact governance |
```
rootulp: I don't think we should establish a precedence that governance rescues funds that are stuck in the x/forwarding module

x/forwarding/RELAYER.md
```diff
@@ -0,0 +1,344 @@
...
+
+Response 201:
+{
+  "forward_addr": "celestia1abc...",
+  "created_at": "2024-01-15T10:30:00Z"
+}
+
+Response 400:
+{ "error": "invalid dest_recipient format" }
+```
```
rootulp: [question] I don't really understand why this is called an "intent". 

AFAIU intents are a user's preference. Something like "I want token X on chain Y and I want the relayer to take the most efficient path there to maximize my bridged value"

On the other hand, this is a concrete forwarding from A -> Celestia -> B. The user has already defined the bridging path and the amount to bridge. As a result we could call this something like "RequestForward"

x/forwarding/README.md
```diff
@@ -0,0 +1,223 @@
+# `x/forwarding`
```
rootulp, line 1: can this file be renamed to README.md?

rootulp (approved):
I had one blocking comment but it can optionally be addressed in a follow up

x/forwarding/README.md
Manav-Aggarwal: I did have it as README.md and is what is consistent across all modules. I had renamed it to SPEC.md given Evan's comment here: https://github.com/celestiaorg/celestia-app/pull/6416#pullrequestreview-3675017363, renamed back to README.md.

x/forwarding/CLAUDE.md
Manav-Aggarwal: I'm strongly of the opinion that long-term we should have them in EVERY core director including in the top level directory. Keeping it in version control helps AI agents understand the module context when working on this codebase.

Can discuss in-person soon too.

x/forwarding/types/params.go
Manav-Aggarwal: Fixed. Changed to `fmt.Errorf("min_forward_amount cannot be negative: %s", p.MinForwardAmount)` which is semantically correct for validation errors.

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Fixed. Renamed `SimappDomainID` → `ChainADomainID` throughout the test file.

x/forwarding/client/cli/tx.go
Manav-Aggarwal: Created follow-up issue #6442 for an optional `--auto` flag. Keeping `--max-igp-fee` required for v1 since relayers typically query fees programmatically and CLI is mainly for testing/debugging.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Added a comment explaining the ordering requirement. The `collections.NewItem()` call registers the item with the SchemaBuilder, so the schema must be built AFTER all items are registered. This is the standard pattern in Cosmos SDK with collections.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Fixed. Using `appconsts.BondDenom` throughout the keeper.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Fixed.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Fixed.

x/forwarding/keeper/keeper_test.go
Manav-Aggarwal: The test you're referring to is actually in `x/forwarding/types/address_test.go`, not keeper_test.go. The keeper_test.go file tests keeper-specific functionality that uses address derivation.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Simplified. Removed the special ErrNotFound handling - if params are missing, something is seriously wrong (InitGenesis should always set them), so we fail loudly now.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: The design is secure. The module account only holds IGP fees transiently during warp execution - fees are immediately passed to the warp transfer. If a transfer fails after IGP collection, tokens are returned to forwardAddr (safe for retry) but the IGP fee stays in the module account (incentivizes relayers to verify routes first). There's no external path to drain the module account.

x/forwarding/keeper/query_server.go
Manav-Aggarwal: Simplified. Removed the special ErrNotFound handling - params should always be set at genesis.

x/forwarding/types/address_test.go
Manav-Aggarwal: Fixed. Removed the TypeScript SDK reference.

x/forwarding/RELAYER.md
Manav-Aggarwal: Good point. Changed 'Contact governance' to just 'None'

x/forwarding/RELAYER.md
Manav-Aggarwal: You're right - renamed 'intent' to 'ForwardingRequest' throughout RELAYER.md. It's a concrete forwarding request with specific parameters, not a user preference.

app/modules.go
```diff
@@ -204,5 +208,6 @@ func allStoreKeys() []string {
 		circuittypes.StoreKey,     // added in v4
 		hyperlanetypes.ModuleName, // added in v4
 		warptypes.ModuleName,      // added in v4
+		forwardingtypes.StoreKey,  // added in v7 (forwarding module)
```
damiannolan: `StoreKey` needs to be included in `StoreUpgrades` of upgrade handler

proto/celestia/forwarding/v1/params.proto
```diff
@@ -0,0 +1,20 @@
...
+import "gogoproto/gogo.proto";
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/celestiaorg/celestia-app/x/forwarding/types";
+
+// Params defines the parameters for the forwarding module.
+message Params {
+  // min_forward_amount is the minimum amount required for forwarding.
+  // Set to 0 to disable minimum threshold.
+  string min_forward_amount = 1 [
```
damiannolan: do we need the param if the default is zero? will this be used?

test/interop/forwarding_integration_test.go
```diff
@@ -0,0 +1,1003 @@
...
+			if err != nil {
+				continue
+			}
+			if eventDispatch, ok := protoMsg.(*coretypes.EventDispatch); ok {
+				return eventDispatch.Message
+			}
+		}
+	}
+	return ""
+}
```
damiannolan: This code doesn't need to exist on the suite. And it already exists in-line in the `hyperlane_test.go` file (L241-253). 

Ideally we could extract this out to a standalone func and use it in both places

test/interop/forwarding_integration_test.go
```diff
@@ -0,0 +1,1003 @@
...
+
+func (s *ForwardingIntegrationTestSuite) countDispatchEvents(events []abci.Event) int {
```
damiannolan: I guess this is also not needed as a method on the suite

x/forwarding/client/cli/query.go
```diff
@@ -0,0 +1,139 @@
...
+  celestia-appd query forwarding derive-address 42161 0x000000000000000000000000742d35cc6634c0532925a3b844bc9e7595f00000`,
+		Args: cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			destDomainStr := args[0]
+			destRecipient := args[1]
```
damiannolan, line 49: nit; This field could be sanitized in the client before it reaches the node.

x/forwarding/client/cli/tx.go
damiannolan: You can provide a default in `cmd.Flags()` below on line 84.

x/forwarding/client/cli/tx.go
```diff
@@ -0,0 +1,88 @@
...
+		Args: cobra.ExactArgs(3),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			forwardAddr := args[0]
+			destDomainStr := args[1]
+			destRecipient := args[2]
```
damiannolan, line 53: nit; could also sanitize here before reaching the node

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,226 @@
...
+	token, err := k.findTIACollateralTokenForDomain(ctx, destDomain)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+	return k.QuoteIgpFeeForToken(sdk.UnwrapSDKContext(ctx), token, destDomain)
+}
+
+// QuoteIgpFeeForToken returns the IGP fee required for a warp transfer of a specific token.
+func (k Keeper) QuoteIgpFeeForToken(ctx sdk.Context, token warptypes.HypToken, destDomain uint32) (sdk.Coin, error) {
+	gasLimit := math.ZeroInt()
```
damiannolan: Shouldn't the gas limit be taken from the value on the registered remote router for the provided domain id?

x/forwarding/keeper/keeper_test.go
damiannolan: This doesn't appear to be testing anything keeper specific.

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,253 @@
...
+	if balances.IsZero() {
+		return nil, types.ErrNoBalance
+	}
+
+	// Process up to MaxTokensPerForward tokens. User can call Forward again for remaining.
+	if len(balances) > types.MaxTokensPerForward {
+		balances = balances[:types.MaxTokensPerForward]
+	}
+
+	moduleAddr := m.k.accountKeeper.GetModuleAddress(types.ModuleName)
```
damiannolan: I think this can be a stateless action and does not need the accountKeeper DI'd into this module? 

https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.52.x-celestia/x/auth/types/permissions.go#L27

https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.52.x-celestia/x/auth/types/account.go#L157

x/forwarding/keeper/msg_server.go
damiannolan: I agree, the error handling is complex here.

x/forwarding/types/keys.go
```diff
@@ -0,0 +1,18 @@
...
+	ModuleName = "forwarding"
+	StoreKey   = ModuleName
+	ParamsKey  = "params"
+
+	// MaxTokensPerForward prevents unbounded iteration and gas exhaustion.
+	MaxTokensPerForward = 20
+
+	EventTypeTokenForwarded     = "forwarding.token_forwarded"
+	EventTypeForwardingComplete = "forwarding.forwarding_complete"
+	EventTypeTokensStuck        = "forwarding.tokens_stuck"
```
damiannolan: are these unused now that you switched to typed events?

x/forwarding/types/msgs.go
```diff
@@ -0,0 +1,80 @@
...
+	return &MsgForward{
+		Signer:        signer,
+		ForwardAddr:   forwardAddr,
+		DestDomain:    destDomain,
+		DestRecipient: destRecipient,
+		MaxIgpFee:     maxIgpFee,
+	}
+}
+
+func (msg *MsgForward) ValidateBasic() error {
```
damiannolan, line 29: I would suggest adding an interface assertion for `HasValidateBasic()` since its an optional interface now rather than on `sdk.Msg`...

x/forwarding/types/msgs.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := msg.MaxIgpFee.Validate(); err != nil {
+		return errors.Wrap(err, "invalid max_igp_fee")
+	}
+
+	return nil
+}
+
+var _ sdk.Msg = &MsgUpdateParams{}
+
+func (msg *MsgUpdateParams) ValidateBasic() error {
```
damiannolan: same here

rootulp:
Converting to draft until Damian's comments are addressed

app/modules.go
Manav-Aggarwal: Done. Added forwarding StoreKey to StoreUpgrades in app/upgrades.go for the chain upgrade.

proto/celestia/forwarding/v1/params.proto
Manav-Aggarwal: Removed. The param was a v1 placeholder that added complexity without clear benefit. Params message is now empty.

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Done. Extracted shared helpers to test/interop/helpers.go: ExtractDispatchMessage(), CountDispatchEvents(), MakeRecipient32(), RecipientToHex(). Both forwarding_integration_test.go and hyperlane_test.go now use these.

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Correct. Converted to standalone functions in test/interop/helpers.go and removed suite methods.

x/forwarding/client/cli/query.go
Manav-Aggarwal: Done. Added client-side sanitization - ensures 0x prefix before sending to node.

x/forwarding/client/cli/tx.go
Manav-Aggarwal: Done. Changed default from empty to "1000000utia" (1 TIA).

x/forwarding/client/cli/tx.go
Manav-Aggarwal: Done. Added same client-side 0x prefix sanitization here too.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Done. Now using router.Gas from GetEnrolledRouter() instead of math.ZeroInt() in both ExecuteWarpTransfer and QuoteIgpFeeForToken.

x/forwarding/keeper/keeper_test.go
Manav-Aggarwal: Correct. Moved the test to x/forwarding/types/address_test.go where it belongs - this tests address derivation logic, not keeper functionality.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Done. Removed accountKeeper from Keeper struct entirely. Now using authtypes.NewModuleAddress(types.ModuleName) for stateless module address derivation.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Okay I simplified the flow a bit so it moves the tokens first and then collect the IGP fee and only rollback if IGP fee isn't sufficient

x/forwarding/types/keys.go
Manav-Aggarwal: Yes, removed. We use typed events (EventTokenForwarded, etc.) via proto - these string constants are no longer needed.

x/forwarding/types/msgs.go
Manav-Aggarwal: Done. Added interface assertions for both sdk.Msg and sdk.HasValidateBasic.

x/forwarding/types/msgs.go
Manav-Aggarwal: Done. Added the same assertion for MsgUpdateParams.

test/interop/forwarding_integration_test.go
rootulp: Can we delete this wrapper function and invoke `CountDispatchEvents` directly?

test/interop/forwarding_integration_test.go
```diff
@@ -150,12 +102,12 @@ func (s *ForwardingIntegrationTestSuite) TestFindHypTokenByDenom_TIA() {
 
 	// Set up simapp counterparty and enroll router so TIA has a route
```
rootulp: why is simapp still present here?

test/interop/forwarding_integration_test.go
```diff
@@ -169,15 +121,15 @@ func (s *ForwardingIntegrationTestSuite) TestFindHypTokenByDenom_TIA() {
...
+		ChainADomainID   = 1337
 	)
 
 	celestiaApp := s.GetCelestiaApp(s.celestia)
 	ctx := s.celestia.GetContext()
 
 	// Set up hyperlane infrastructure on simapp (origin chain)
 	ismIDSimapp := s.SetupNoopISM(s.chainA)
-	mailboxIDSimapp := s.SetupMailBox(s.chainA, ismIDSimapp, SimappDomainID)
+	mailboxIDSimapp := s.SetupMailBox(s.chainA, ismIDSimapp, ChainADomainID)
```
rootulp: variable name still contains simapp

rootulp (commented):
1. The integration test still references SimApp in a bunch of places.
2. Do we need to update the CIP now that this module no longer has any params?

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Done. Deleted the wrapper function in commit 48cfa288. Now calling `ExtractDispatchMessage` from helpers.go directly.

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Done. Both wrapper functions (`countDispatchEvents` and `extractDispatchMessage`) are now deleted. Callers use the shared helpers directly. See commits 467f0b83 and 48cfa288.

Manav-Aggarwal:
> 1. The integration test still references SimApp in a bunch of places.
> 2. Do we need to update the CIP now that this module no longer has any params?

Done. Removed simapp references.
CIP changes: https://github.com/celestiaorg/CIPs/pull/377

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Fixed.

test/interop/forwarding_integration_test.go
Manav-Aggarwal: Fixed.

test/interop/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+// The address bytes are placed at the end (bytes 12-31).
+func MakeRecipient32(addr sdk.AccAddress) []byte {
+	recipient := make([]byte, 32)
+	copy(recipient[12:], addr.Bytes())
+	return recipient
+}
+
+// RecipientToHex converts a 32-byte recipient to a HexAddress.
+func RecipientToHex(recipient []byte) util.HexAddress {
+	hexAddr, _ := util.DecodeHexAddress("0x" + hex.EncodeToString(recipient))
```
rach-id, line 53: why is the error not handled here?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+type msgServer struct {
+	k Keeper
+}
+
+func NewMsgServerImpl(keeper Keeper) types.MsgServer {
+	return &msgServer{k: keeper}
+}
+
+// Forward forwards up to 20 tokens at forwardAddr to the committed destination.
+// Partial failures are by design: failed tokens remain for retry while others proceed.
```
rach-id: what's the possibility of having an ever increasing number of failed tokens? do we have any cleanup mechanism?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+	forwardAddr, err := sdk.AccAddressFromBech32(msg.ForwardAddr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid forward_addr %q: %w", msg.ForwardAddr, err)
+	}
+
+	destRecipient, err := util.DecodeHexAddress(msg.DestRecipient)
+	if err != nil {
+		return nil, fmt.Errorf("invalid dest_recipient hex: %w", err)
+	}
+	// Length validation is done in ValidateBasic
```
rach-id: [nit]
probably unnecessary
```suggestion
```

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	// Quote IGP fee for this token transfer
+	quotedFee, err := m.k.QuoteIgpFeeForToken(ctx, hypToken, destDomain)
+	if err != nil {
+		return types.NewFailureResult(balance.Denom, balance.Amount, fmt.Sprintf("failed to quote IGP fee: %s", err.Error()))
+	}
+
+	// Verify relayer provided sufficient max_igp_fee
+	// Only compare if quoted fee is positive and same denom
```
rach-id, line 132: and if it's not? what does that mean?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+
+	// Move tokens to module account first - if this fails, we haven't touched relayer's funds yet
+	if err := m.k.bankKeeper.SendCoins(ctx, forwardAddr, moduleAddr, sdk.NewCoins(balance)); err != nil {
+		return types.NewFailureResult(balance.Denom, balance.Amount, "failed to move tokens: "+err.Error())
+	}
+
+	// Collect IGP fee from relayer (signer) to module account
+	// Only collect if there's a positive fee to pay
```
rach-id, line 150: can the fee ever be 0?

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+	if quotedFee.IsPositive() {
+		if err := m.k.bankKeeper.SendCoins(ctx, signerAddr, moduleAddr, sdk.NewCoins(quotedFee)); err != nil {
+			// Return tokens to forward address since we can't complete the transfer
+			if recoveryErr := m.k.bankKeeper.SendCoins(ctx, moduleAddr, forwardAddr, sdk.NewCoins(balance)); recoveryErr != nil {
+				ctx.Logger().Error("CRITICAL: tokens stuck in module after IGP collection failure",
+					"denom", balance.Denom, "amount", balance.Amount.String(), "error", recoveryErr)
```
rach-id: would also make sense to log the `err` to see the original error

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+				Amount:      balance.Amount,
+				ModuleAddr:  moduleAddr.String(),
+				ForwardAddr: forwardAddr.String(),
+				Error:       fmt.Sprintf("warp: %s; recovery: %s", err.Error(), recoveryErr.Error()),
+			}); emitErr != nil {
+				ctx.Logger().Error("failed to emit EventTokensStuck", "error", emitErr)
+			}
+			return types.NewFailureResult(balance.Denom, balance.Amount, fmt.Sprintf("CRITICAL: warp failed and recovery failed - tokens stuck in module account: warp=%s recovery=%s", err.Error(), recoveryErr.Error()))
+		}
+		// Note: IGP fee is NOT returned on warp failure - it was consumed by the failed attempt.
```
rach-id, line 179: what do you mean by consumed here? it seems that the fees are stuck in the module and there is no recovery mechanism

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,223 @@
...
+			return types.NewFailureResult(balance.Denom, balance.Amount,
+				fmt.Sprintf("failed to collect IGP fee from relayer: %s", err.Error()))
+		}
+	}
+
+	messageId, err := m.k.ExecuteWarpTransfer(ctx, hypToken, moduleAddr.String(), destDomain, destRecipient, balance.Amount, quotedFee)
+	if err != nil {
+		// Warp failed - return tokens to forward address
+		if recoveryErr := m.k.bankKeeper.SendCoins(ctx, moduleAddr, forwardAddr, sdk.NewCoins(balance)); recoveryErr != nil {
+			ctx.Logger().Error("CRITICAL: tokens stuck in module account after failed recovery",
```
rach-id, line 167: same, is having no recovery mechanism for stuck tokens intended?

x/forwarding/keeper/msg_server.go
rach-id, line 1: As a follow up, we should see the processing time of this kind of messages during prepare/process/finalize proposal and also checktx, similar to what we're doing to blobs and sdk txs, to make sure a 32mb/128mb full of these forwarding messages don't massively increase resource consumption

x/forwarding/types/expected_keepers.go
```diff
@@ -0,0 +1,64 @@
...
+// This is intentional because the hyperlane-cosmos warp keeper exposes its state via public
+// collections.Map fields (HypTokens, EnrolledRouters) rather than through methods.
+// Go interfaces cannot expose struct fields, so we must use the concrete type.
+//
+// The forwarding keeper wraps these field accesses in helper methods:
+//   - FindHypTokenByDenom: wraps warpKeeper.HypTokens.Get
+//   - HasEnrolledRouter: wraps warpKeeper.EnrolledRouters.Has
+//
+// This interface documents the warp keeper capabilities we depend on for future reference.
+type WarpKeeper interface {
```
rach-id, line 27: Add a static type check here to tie both to the expected methods:

```
var _ WarpKeeper = &warpkeeper.Keeper{}
```

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,218 @@
...
+		return warptypes.HypToken{}, fmt.Errorf("failed to iterate warp tokens: %w", err)
+	}
+	defer iter.Close()
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	for ; iter.Valid(); iter.Next() {
+		token, err := iter.Value()
+		if err != nil {
+			sdkCtx.Logger().Warn("failed to read warp token during TIA collateral search", "error", err)
+			continue
```
rach-id, line 80: In which case could this happen?

x/forwarding/RELAYER.md
rach-id, line 1: Where is the relayer implemented? If it's still not implemented, how are we planning to test this? 

Also, would be good to link the relayer implementation in this document.

test/interop/helpers.go
Manav-Aggarwal: Fixed. Added a comment explaining why the error is safe to ignore - since `hex.EncodeToString` always produces a valid hex string and we prepend "0x" ourselves, `DecodeHexAddress` cannot fail in this context.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: When forwarding fails:
1. **Before token move**: Failure returns immediately, tokens stay at `forwardAddr`
2. **After token move**: Tokens are returned to `forwardAddr` via the recovery path

For any "failed" forwarding, relayers simply retry - tokens remain at `forwardAddr` waiting. The module account is only used transiently during the warp transfer execution.

There's no cleanup mechanism here, we could introduce it in a future upgrade but I see it as a non-risk, it's equivalent to just sending funds to the wrong address and getting tokens stuck.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Done. Removed the redundant comment - partial failure behavior is documented in RELAYER.md.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: If `quotedFee` is not positive (zero or negative), we skip the IGP fee comparison entirely and proceed without collecting any fee. This happens when:

1. **No IGP hook configured**: The warp route has no interchain gas paymaster
2. **Free relaying**: The IGP returns zero fee for this destination

In both cases, the relayer doesn't pay an IGP fee but still pays gas for the tx itself. The warp transfer proceeds normally - Hyperlane will deliver the message but without guaranteed gas payment on the destination (relayer covers it or it relies on alternative mechanisms).

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Yes, the fee can theoretically be zero if:

1. **No IGP hook configured** on the warp route
2. **Free delivery** to certain destinations (e.g., subsidized routes)
3. **IGP oracle returns zero** for the destination domain

This doesn't happen in practice though. When `quotedFee.IsPositive()` is false (zero), we skip the fee comparison check and don't collect any IGP fee from the relayer. The warp transfer still proceeds - Hyperlane will dispatch the message but without prepaid gas on the destination chain.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Done. Updated the log to include both `igp_error` (the original error from IGP fee collection) and `recovery_error` (the error from attempting to return tokens).

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: "Consumed" means the IGP fee was paid to the Hyperlane IGP contract/hook as part of the failed warp attempt. It's not stuck in our module - it was transferred to the Hyperlane system for the dispatch that ultimately failed.

This is intentional and documented: it incentivizes relayers to verify routes exist before submitting. If we returned IGP fees on failure, relayers could spam invalid transfers at no cost beyond gas. 

For tokens that fail warp, they're returned to `forwardAddr` (not stuck). Only in the extremely rare case where bank `SendCoins` fails during recovery would tokens be stuck in the module account - we emit `EventTokensStuck` for monitoring that case.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Yes, this is intentional for v1. Tokens that fail warp transfer are returned to the `forwardAddr` where they can be retried. The only tokens that could truly get "stuck" are in the CRITICAL case where our own `SendCoins` recovery fails (line 196-214), which should be extremely rare and would require something catastrophically wrong with chain state.

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Good suggestion. Created follow-up issue: #6447 to benchmark MsgForward processing time during prepare/process/finalize proposal and checktx, similar to blob benchmarks.

x/forwarding/types/expected_keepers.go
Manav-Aggarwal: Added a NOTE to the interface documentation explaining why we intentionally do NOT add a static type assertion here.

The `WarpKeeper` interface only documents the method-based capabilities. The concrete `*warpkeeper.Keeper` also exposes critical functionality via public struct fields (`HypTokens`, `EnrolledRouters` as `collections.Map`), which cannot be expressed in a Go interface. Since our keeper depends on BOTH the methods AND direct field access, a type assertion would fail compilation.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: This happens when iterating warp tokens and we fail to read one of them from state. It could occur if:

1. **Corrupted state**: A warp token entry was written incorrectly (shouldn't happen in normal operation)
2. **Codec mismatch**: Schema evolution issue during an upgrade
3. **Storage error**: Very unlikely but possible transient read failure

We log a warning and continue iterating rather than failing the entire operation. This allows us to find a valid TIA collateral token even if some unrelated warp tokens have issues. The logging ensures operators can investigate any corrupted entries.

x/forwarding/RELAYER.md
Manav-Aggarwal: The relayer is not yet implemented - it's planned as a separate repository. I don't think blocking merging this on relayer implementation is necessary.

For testing, this PR includes:
1. **Unit tests**: `x/forwarding/types/*_test.go` and `x/forwarding/keeper/*_test.go`
2. **Integration test**: `test/interop/forwarding_integration_test.go` - tests the full flow including warp transfer execution with a simulated Hyperlane setup

We also plan to have a tastora e2e test, see https://github.com/celestiaorg/celestia-app/issues/6430
The integration test validates the on-chain module behavior. For end-to-end testing with a real relayer, we'll need the relayer implementation.

I'll add a link to the relayer repo once it's created. Thanks for the suggestion!

x/forwarding/types/address.go
```diff
@@ -0,0 +1,56 @@
...
+	h := sha256.New()
+	h.Write(destDomainBytes)
+	h.Write(destRecipient)
+	callDigest := h.Sum(nil)
+
+	// Step 3: salt = sha256(ForwardVersion || callDigest)
+	h.Reset()
+	h.Write([]byte{ForwardVersion})
+	h.Write(callDigest)
+	salt := h.Sum(nil)
```
cmwaters, line 56: What is the reason for having this over two stages instead of one?

x/forwarding/types/address.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Step 4: Use SDK's address.Module for deterministic derivation
+	addr := address.Module(ModuleName, salt)
+
+	return addr[:20], nil
```
cmwaters: Should this be a constant?

x/forwarding/types/expected_keepers.go
rach-id: did you try it? I added it to the code and seems to build fine

x/forwarding/types/address.go
Manav-Aggarwal: The two-stage approach provides domain separation for upgradeability:

1. `callDigest = sha256(destDomain || destRecipient)` - unique forwarding target identifier  
2. `salt = sha256(ForwardVersion || callDigest)` - version prefix for future changes

If we change the derivation in v2, using ForwardVersion=2 generates different addresses without collisions. This follows the pattern in `address.Module()`.

x/forwarding/types/address.go
Manav-Aggarwal: Done. Added `CosmosAddressLen = 20` constant in 47b6f719.

x/forwarding/types/expected_keepers.go
Manav-Aggarwal: Thanks for verifying! Added the type assertion in 47b6f719

Manav-Aggarwal:
> It doesn't look like there are any e2e tests testing that the upgrade happens correctly and you can use the forwarding module afterwards

Good point. E2E upgrade testing is tracked in issue https://github.com/celestiaorg/celestia-app/issues/6430 (testaro e2e test suite). This PR includes:

- Unit tests (x/forwarding/types/*_test.go and x/forwarding/keeper/*_test.go)
- Integration test (test/interop/forwarding_integration_test.go) that validates the full flow

The upgrade-specific e2e test will validate the module works after chain upgrade.

x/forwarding/CLAUDE.md
rootulp: Sure but why does Claude get special treatment? Other devs may use Codex, Gemini, Cursor, etc. 

> I'm strongly of the opinion that long-term we should have them in EVERY core director including in the top level directory.

I'm concerned that if we allow this for all possible AI agents then we'll have # of AI agents * # of important directores = # of Markdown files. Ideally there is a consistent format for all AI agents to use. IMO that could just be README.md in every directory.

x/forwarding/types/keys.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+const (
+	ModuleName = "forwarding"
+	StoreKey   = ModuleName
```
damiannolan: Isn't the module stateless? I think this can be removed, right?

x/forwarding/types/msgs_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	"cosmossdk.io/math"
+	"github.com/celestiaorg/celestia-app/v7/x/forwarding/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
+	config := sdk.GetConfig()
```
damiannolan, line 15: the classic sdk global... 🤮

x/forwarding/types/expected_keepers.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"cosmossdk.io/math"
+	"github.com/bcp-innovations/hyperlane-cosmos/util"
+	warpkeeper "github.com/bcp-innovations/hyperlane-cosmos/x/warp/keeper"
+	warptypes "github.com/bcp-innovations/hyperlane-cosmos/x/warp/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Compile-time interface compliance check
+var _ WarpKeeper = &warpkeeper.Keeper{}
```
damiannolan: I disagree with this comment which asked to add this interface assertion https://github.com/celestiaorg/celestia-app/pull/6427#discussion_r2716034490

In my opinion it is unnecessary and not really the responsibility of this module. All this does is a dependency on the warp keeper pkg to this types pkg.

The expected interface will break the compiler in app.go if the API changes upon software upgrade, and that would be the queue to fix it. This additional assertion is redundant. 

tldr; we can remove this

x/forwarding/types/address.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/types/address"
+)
+
+const (
+	ForwardVersion  = uint8(1)
```
damiannolan: nit; may as well godoc the version as well

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,221 @@
...
+			Error:       result.Error,
+		}); err != nil {
+			ctx.Logger().Error("failed to emit EventTokenForwarded", "error", err)
+		}
+	}
+
+	return results
+}
+
+func (m msgServer) emitSummaryEvent(ctx sdk.Context, msg *types.MsgForward, results []types.ForwardingResult) {
```
damiannolan: nit; I don't think this needs to be a method on the msgServer judging from the looks of things.

Could we create `events.go` inside keeper with standalone funcs (not keeper/msgServer methods). 

We can then create some funcs in there which we can call from this file. A lot of the event emission spans multiple lines and pollutes the code a little bit, making it a little bit harder to read through quickly

x/forwarding/keeper/keeper.go
```diff
@@ -0,0 +1,218 @@
...
+	cdc             codec.BinaryCodec
+	bankKeeper      types.BankKeeper
+	warpKeeper      *warpkeeper.Keeper
+	hyperlaneKeeper types.HyperlaneKeeper
+}
+
+func NewKeeper(
+	cdc codec.BinaryCodec,
+	bankKeeper types.BankKeeper,
+	warpKeeper *warpkeeper.Keeper,
```
damiannolan: Accept the expected interface here, and not the `warpkeeper` concrete type.

x/forwarding/keeper/genesis.go
damiannolan, line 1: This doesn't look like it needs to exist. Can just noop / early return method implementations in module.go

app/upgrades.go
```diff
@@ -102,7 +104,10 @@ func (app App) RegisterUpgradeHandlers() {
 	}
 
 	if upgradeInfo.Name == upgradeName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) { //nolint:staticcheck
-		// TODO: Apply any store upgrades here.
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added: []string{forwardingtypes.StoreKey},
```
damiannolan: This can actually be removed now with the const in the types pkg I raised in another comment.
I think this is a relic from when there was params being stored in state.

x/forwarding/types/keys.go
Manav-Aggarwal: Removed StoreKey since module is stateless. Also removed the StoreUpgrades.Added entry in app/upgrades.go in 3bd9554f

x/forwarding/types/msgs_test.go
Manav-Aggarwal: 😅 Yeah, the SDK address config pattern. At least it's test-only here.

x/forwarding/types/expected_keepers.go
Manav-Aggarwal: Removed the interface assertion per your suggestion. Keeping the concrete type in keeper.go since warpkeeper.Keeper exposes state via public \`collections.Map\` fields that cannot be expressed in interfaces in 3bd9554f. cc: @rach-id

x/forwarding/types/address.go
Manav-Aggarwal: Added godoc to ForwardVersion explaining its purpose in 3bd9554f

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Created keeper/events.go with standalone event emission functions: \`EmitTokenForwardedEvent\`, \`EmitForwardingCompleteEvent\`, and \`EmitTokensStuckEvent\` in 3bd9554f

x/forwarding/keeper/genesis.go
Manav-Aggarwal: Deleted genesis.go and inlined the no-op implementations in module.go in 3bd9554f

app/upgrades.go
Manav-Aggarwal: Removed - module is stateless so no StoreUpgrades.Added needed in 3bd9554f

x/forwarding/RELAYER.md
Manav-Aggarwal: Fixed in 48cfa288 - deleted the wrapper function.

x/forwarding/RELAYER.md
Manav-Aggarwal: Fixed in 47b6f719 - renamed simapp references to chainA in hyperlane_test.go.

x/forwarding/keeper/keeper.go
Manav-Aggarwal: Addressed in 15475bfc: Created `WarpKeeperAdapter` that wraps the concrete `*warpkeeper.Keeper` and implements the expanded `WarpKeeper` interface. The interface now includes `GetHypToken`, `GetAllHypTokens`, `HasEnrolledRouter`, and `GetEnrolledRouter` methods that wrap the collections.Map field access.

This addresses the review comment about accepting the expected interface instead of the concrete type.

damiannolan: The adapter actually looks like a decent approach to this. Initially I did not realise that the reason the concrete type was imported was due to the lack of public apis offered for collection iteration etc. by the warp module. 

Happy enough with this approach 🙏🏻

app/upgrades.go
```diff
@@ -95,15 +95,6 @@ func (app App) RegisterUpgradeHandlers() {
...
 	)
-
-	upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
-	if err != nil {
-		panic(err)
-	}
-
-	if upgradeInfo.Name == upgradeName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) { //nolint:staticcheck
-		// TODO: Apply any store upgrades here.
-	}
```
damiannolan, line 106: Sorry to be a pain but could we leave this code in place as its usually needed for upgrades with do apply stores (e.g. zkism).

test/interop/forwarding_integration_test.go
damiannolan, line 1: I can't say I've reviewed this test file in great detail, if I can identify any improvements when trying to write docker e2e I can chime in.

app/upgrades.go
Manav-Aggarwal: Done in 5e523111

rootulp:
Note: golangci-lint is failing which will block merging b/c it's a required CI check. I'll still review

x/forwarding/RELAYER.md
```diff
@@ -0,0 +1,344 @@
+# Forwarding Module Relayer Specification
```
rootulp: [question] if the relayer is implemented in a separate repo (e.g. https://github.com/celestiaorg/forwarding-relayer) can this markdown file be deleted? Or can this just link to the README of the relayer?

I'm concerned the relayer implementation will deviate from what is described here and this file will go stale.

x/forwarding/keeper/msg_server.go
```diff
@@ -0,0 +1,185 @@
...
+		if err := m.k.bankKeeper.SendCoins(ctx, signerAddr, moduleAddr, sdk.NewCoins(quotedFee)); err != nil {
+			// Return tokens to forward address since we can't complete the transfer
+			if recoveryErr := m.k.bankKeeper.SendCoins(ctx, moduleAddr, forwardAddr, sdk.NewCoins(balance)); recoveryErr != nil {
+				ctx.Logger().Error("CRITICAL: tokens stuck in module after IGP collection failure",
+					"denom", balance.Denom, "amount", balance.Amount.String(), "igp_error", err, "recovery_error", recoveryErr)
+			}
+			return types.NewFailureResult(balance.Denom, balance.Amount,
+				fmt.Sprintf("failed to collect IGP fee from relayer: %s", err.Error()))
+		}
+	}
```
rootulp, line 161: I echo Damian's concern about the error handling here being complex. The "CRITICAL: " case seems unlikely but I wonder if it's possible to remove this recovery mechanism.

1. Is it possible to move funds directly from the signerAddr to the forwardAddr and bypass the forward module account? That way there are fewer `bankKeeper.SendCoins` invocations and if it fails, coins remain at the signerAddr.
2. If the above option isn't possible, can we have an AI bot try to attack this code? For example, is it ever possible for an attacker to get more funds returned to them than they sent to the module account address?

x/forwarding/RELAYER.md
Manav-Aggarwal: Addressed the relayer documentation feedback: linked to the external [forwarding-relayer docs](https://github.com/celestiaorg/forwarding-relayer/blob/master/RELAYER.md) in the README and deleted the local RELAYER.md file.

Commit: 140d956a

x/forwarding/keeper/msg_server.go
Manav-Aggarwal: Created follow-up issue: #6459

Issue #6395, opened 2026-01-13
feat: Add ZK ISM module for enabling trust-minimized Hyperlane message verification


Manav-Aggarwal:
## Summary

Add a new `x/zkism` module that implements a Hyperlane Interchain Security Module (ISM) using SP1 Groth16 zero-knowledge proofs for trust-minimized cross-chain message verification.

## Problem Definition

Traditional cross-chain bridges rely on validator sets or multisigs. This module enables cryptographic verification of remote chain state and message authenticity without trusted intermediaries.

## Proposal

The module provides:

- **MsgCreateInterchainSecurityModule** - Initialize ISM with trusted state and verifier config
- **MsgUpdateInterchainSecurityModule** - Update trusted state via ZK state transition proof
- **MsgSubmitMessages** - Authorize Hyperlane messages via ZK membership proof

Key features:
- SP1 Groth16 proof verification (BN254 curve, gnark library)
- Hyperlane router integration
- One-time message consumption for replay protection

## Reference Implementation

https://github.com/celestiaorg/celestia-app/tree/feature/zk-execution-ism/x/zkism

rootulp:
@damiannolan @Manav-Aggarwal @jonas089 is anything remaining on this issue? It seems resolved by https://github.com/celestiaorg/celestia-app/pull/6438

damiannolan:
Happy to close it out

PR #6438, opened 2026-01-20
feat: add hyperlane ism module x/zkism with generic sp1 groth16 verifier

damiannolan:
## Overview

Adds `x/zkism` module for Hyperlane ISM verification, using a generic SP1 groth16 verifier.

At the moment this PR is intended for CI and I will reach out to core/app devs for comment on whether or not they would prefer if this PR (feat-branch) was broken up into smaller PRs. cc. @rootulp 

#6395


x/zkism/keeper/msg_server.go
```diff
@@ -0,0 +1,168 @@
...
+	if !bytes.Equal(ism.State, publicValues.State) {
+		return nil, errorsmod.Wrapf(types.ErrInvalidTrustedState, "expected %x, got %x", ism.State, publicValues.State)
+	}
+
+	verifier, err := types.NewSP1Groth16Verifier(ism.Groth16Vkey)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := verifier.VerifyProof(msg.Proof, ism.StateTransitionVkey, msg.PublicValues); err != nil {
```
damiannolan, line 81: We should actually be charging additional gas here. And in the `SubmitMessages` rpc. 
Or just pass the ctx to verifier and charge in there.

Manav-Aggarwal: Let's do it, feel free to suggest something reasonable

app/modules.go
```diff
@@ -204,5 +209,6 @@ func allStoreKeys() []string {
 		circuittypes.StoreKey,     // added in v4
 		hyperlanetypes.ModuleName, // added in v4
 		warptypes.ModuleName,      // added in v4
+		zkismtypes.StoreKey,
```
rootulp: idk if these comments are helpful any more so extremely optional


```suggestion
		zkismtypes.StoreKey, // added in v7
```

x/zkism/keeper/export_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+
+	"cosmossdk.io/collections"
+	"github.com/bcp-innovations/hyperlane-cosmos/util"
+	"github.com/celestiaorg/celestia-app/v7/x/zkism/types"
+)
+
+// GetIsm is a test func used for checking existence of a message id in the messages store collection.
```
rootulp:
```suggestion
// HasMessageId is a test func used for checking existence of a message id in the messages store collection.
```

x/zkism/keeper/export_test.go
```diff
@@ -0,0 +1,29 @@
+package keeper
```
rootulp, line 1: [question] if all the files in this file are test functions, is it possible to hide them from the exported public API?

x/zkism/types/ism.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"errors"
+
+	"github.com/bcp-innovations/hyperlane-cosmos/util"
+	ismtypes "github.com/bcp-innovations/hyperlane-cosmos/x/core/01_interchain_security/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+const (
+	ModuleTypeZkISM = 42
```
rootulp: [question] what is the rationale for setting this to 42? Perhaps it's to de-conflict with any existing module type that Hyperlane has implementations for?

x/zkism/doc.go
```diff
@@ -0,0 +1,4 @@
+// The x/zkism module is a custom implementation of a Hyperlane Interchain Security Module (ISM).
+// It's purpose is to provide a means for proving execution of a remote chain through a standardised interface, the ISM.
```
rootulp:
```suggestion
// Its purpose is to provide a means for proving execution of a remote chain through a standardised interface, the ISM.
```

go.mod
```diff
@@ -29,6 +30,8 @@ require (
 	github.com/celestiaorg/rsmt2d v0.15.1
 	github.com/cometbft/cometbft v0.38.17
 	github.com/cometbft/cometbft-db v1.0.4
+	github.com/consensys/gnark v0.14.0
+	github.com/consensys/gnark-crypto v0.19.0
```
rootulp: [nit] why not the latest release?


```suggestion
	github.com/consensys/gnark-crypto v0.19.2
```

x/zkism/internal/groth16/groth16_test.go
```diff
@@ -0,0 +1,76 @@
...
+	require.NotNil(t, proof)
+
+	// sanity checks that the proof components are non-zero
+	require.False(t, proof.Ar.IsInfinity(), "Ar should not be point at infinity")
+	require.False(t, proof.Bs.IsInfinity(), "Bs should not be point at infinity")
+	require.False(t, proof.Krs.IsInfinity(), "Krs should not be point at infinity")
+}
+
+func TestHashBN254(t *testing.T) {
+	input := []byte("just trust me bro")
```
Manav-Aggarwal, line 42: lol

x/zkism/types/public_values.go
```diff
@@ -0,0 +1,161 @@
...
+		return fmt.Errorf("read MerkleTreeAddress: %w", err)
+	}
+
+	var count uint64 // read uint64 (little-endian) length prefix
+	if err := binary.Read(buf, binary.LittleEndian, &count); err != nil {
+		return fmt.Errorf("read message ids length: %w", err)
+	}
+
+	remaining := buf.Len()
+	if remaining < int(count*32) {
```
Manav-Aggarwal, line 144: This check has an integer overflow issue. `count*32` overflows in uint64 before int conversion for large values of count. For example, `count = 0x2000000000000001` causes `count*32` to wrap to 32, passing the check but then causing OOM on the subsequent make().

x/zkism/keeper/genesis.go
```diff
@@ -0,0 +1,94 @@
...
+			Id:       ism.Id,
+			Messages: msgs,
+		})
+	}
+
+	sort.Slice(genesisMessages, func(i, j int) bool {
+		return genesisMessages[i].Id.Compare(genesisMessages[j].Id) < 0
+	})
+
+	return &types.GenesisState{
```
Manav-Aggarwal, line 132: The `messageProofSubmitted` collection in the keeper is not being exported here. This means the state will be lost on chain upgrade/export, which would allow duplicate proof submissions after a chain upgrade. Should add `messageProofSubmitted` to the GenesisState proto and export/import it here?

x/zkism/keeper/genesis.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// ExportGenesis outputs the modules state for genesis exports.
+func (k *Keeper) ExportGenesis(ctx context.Context) (*types.GenesisState, error) {
+	var isms []types.InterchainSecurityModule
```
Manav-Aggarwal, line 66: The ISMs are not sorted after the Walk. While messages are sorted later (line 86-88), the ISMs array order depends on map iteration order, which is non-deterministic. This can cause hash mismatches when comparing genesis exports. Should add `sort.Slice(isms, ...)` after the Walk.

x/zkism/keeper/msg_server.go
```diff
@@ -0,0 +1,168 @@
...
+// NewMsgServerImpl creates and returns a new module MsgServer instance.
+func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
+	return &msgServer{keeper}
+}
+
+// CreateInterchainSecurityModule implements types.MsgServer.
+func (m msgServer) CreateInterchainSecurityModule(ctx context.Context, msg *types.MsgCreateInterchainSecurityModule) (*types.MsgCreateInterchainSecurityModuleResponse, error) {
+	ismId, err := m.coreKeeper.IsmRouter().GetNextSequence(ctx, types.ModuleTypeZkISM)
+	if err != nil {
+		return nil, errorsmod.Wrap(err, err.Error())
```
Manav-Aggarwal: Redundant error wrapping - `errorsmod.Wrap(err, err.Error())` wraps the error with its own message. Consider using descriptive context instead, e.g., `errorsmod.Wrap(err, "failed to get next ISM sequence")`

x/zkism/keeper/msg_server.go
```diff
@@ -0,0 +1,168 @@
...
+		Owner:               msg.Creator,
+		State:               msg.State,
+		Groth16Vkey:         msg.Groth16Vkey,
+		MerkleTreeAddress:   msg.MerkleTreeAddress,
+		StateTransitionVkey: msg.StateTransitionVkey,
+		StateMembershipVkey: msg.StateMembershipVkey,
+	}
+
+	if err := m.isms.Set(ctx, ismId.GetInternalId(), newIsm); err != nil {
+		return nil, errorsmod.Wrap(err, err.Error())
```
Manav-Aggarwal: Same redundant error wrapping pattern here - `errorsmod.Wrap(err, err.Error())`. Consider: `errorsmod.Wrap(err, "failed to store ISM")`

x/zkism/module.go
```diff
@@ -0,0 +1,118 @@
...
+	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(am.keeper))
+	types.RegisterQueryServer(cfg.QueryServer(), keeper.NewQueryServerImpl(am.keeper))
+}
+
+// DefaultGenesis returns default genesis state as raw bytes for the module.
+func (AppModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
+	return cdc.MustMarshalJSON(types.DefaultGenesis())
+}
+
+// ValidateGenesis performs genesis state validation for the zkism module.
```
Manav-Aggarwal, line 85: ValidateGenesis currently returns nil without validation. While InitGenesis catches some issues, ValidateGenesis should catch errors earlier (before InitGenesis runs). Consider implementing `GenesisState.Validate()` method and calling it here.

x/zkism/client/cli/query.go
```diff
@@ -0,0 +1,127 @@
...
+
+			res, err := queryClient.Isms(cmd.Context(), req)
+			if err != nil {
+				return err
+			}
+
+			return clientCtx.PrintProto(res)
+		},
+	}
+
```
Manav-Aggarwal, line 85: This command reads pagination with `client.ReadPageRequest(cmd.Flags())` at line 67, but does not add the pagination flags to the command. Add `flags.AddPaginationFlagsToCmd(cmd, "isms")` before the return.

x/zkism/client/cli/query.go
```diff
@@ -0,0 +1,127 @@
...
+	flags.AddQueryFlagsToCmd(cmd)
+	return cmd
```
Manav-Aggarwal, line 128: Same issue - pagination is read at line 105 but flags are not added. Add `flags.AddPaginationFlagsToCmd(cmd, "messages")` before the return.

x/zkism/keeper/export_test.go
damiannolan: This is somewhat of a trick and also done by go standard library in order to allow test functions access to unexported symbols. Since tests live in `package pkg_test`  they would otherwise not have access to unexported functions in `package pkg`. 

The `_test.go` files are only actually compiled when running `go test`. So when they cannot be used as a normal public APIs. Example: https://go.dev/src/runtime/export_test.go

go.mod
damiannolan: good catch, thank you!

x/zkism/types/ism.go
damiannolan: There is a bunch of enums defined by the hyperlane spec for ism types and they define them in their cosmos module here - https://github.com/bcp-innovations/hyperlane-cosmos/blob/main/x/core/01_interchain_security/types/types.go#L32-L45

Not all ism types are actually implemented but at least the values linked above are assigned. This affects hex address identifier generation in hyperlane cosmos. E.g. `0x726f757465725f69736d000000000000000000000000002a0000000000000001` - 2a == 42 when hex converts to decimal.

42 was chosen arbitrarily. Would you like to suggest changing it to anything else? We could also go with 255.
It would probably be good to ensure that this const is assigned `uint8`.

x/zkism/keeper/genesis.go
damiannolan: This should have been done in https://github.com/celestiaorg/celestia-app/pull/6420. Will address this.

Fwiw nobody really uses genesis exports for upgrades anymore. Unless its a disaster scenario.
Most teams use snapshots. And celestia now uses rolling upgrades with multiplexer binaries.

x/zkism/keeper/msg_server.go
damiannolan: Added in https://github.com/celestiaorg/celestia-app/pull/6448

x/zkism/client/cli/query.go
damiannolan: https://github.com/celestiaorg/celestia-app/pull/6438/commits/99a8d394be78b2cb8db544bd74cdfaa10917b293

x/zkism/client/cli/query.go
damiannolan: https://github.com/celestiaorg/celestia-app/pull/6438/commits/99a8d394be78b2cb8db544bd74cdfaa10917b293

x/zkism/keeper/msg_server.go
damiannolan: https://github.com/celestiaorg/celestia-app/pull/6438/commits/99a8d394be78b2cb8db544bd74cdfaa10917b293

x/zkism/keeper/msg_server.go
damiannolan: addressed in https://github.com/celestiaorg/celestia-app/pull/6438/commits/99a8d394be78b2cb8db544bd74cdfaa10917b293 along with pagination flags, thank you!

x/zkism/keeper/genesis.go
damiannolan: The map collection in cosmos sdk is not implemented with a go map, that would be bad, as it would have non-deterministic iteration within the state machine. This would lead to all sorts of crazy problems. 

There is actually no need to sort the messages below. I can add or remove sort for consistency with both collections, but keys are ordered lexicographically within the underlying IAVL tree so the keyset collection would traverse that in order.

x/zkism/types/public_values.go
jonas089: Since in practice we should never encounter a proof with this many messages, I think it probably makes sense to set a very high maximum for count e.g. 1 billion and return a meaningful error if someone ever tries to submit more than that.

x/zkism/module.go
damiannolan: Addressed in https://github.com/celestiaorg/celestia-app/pull/6438/commits/15a0c89b6fe85b67a8f9c0ff6c8640b6500884df 🙏🏻

x/zkism/types/public_values.go
damiannolan: It might be more practical to set a max bytes on the public values... and just return err in `ValidateBasic` if it exceeds that threshold. What do you think?

x/zkism/types/ism.go
rootulp: `42` or `255` work for me. It would help if there was a comment above this const that explained the rationale. Something like

```suggestion
	// ModuleTypeZkISM is set to 42 to avoid a conflict with any existing Hyperlane ISM defined in https://github.com/bcp-innovations/hyperlane-cosmos/blob/main/x/core/01_interchain_security/types/types.go#L32-L45
	ModuleTypeZkISM = 42
```

Agreed with the `uint8` considering that's what bc-innovations does. Extremely optional but we could consider adopting their naming scheme so: `INTERCHAIN_SECURITY_MODULE_TYPE_ZK_ISM`

x/zkism/keeper/export_test.go
rootulp: Thanks for the wisdom. That makes a lot of sense. The test functions don't render in the public Go docs: https://pkg.go.dev/runtime#pkg-functions which is what we want 👍

x/zkism/keeper/genesis.go
damiannolan: implemented in https://github.com/celestiaorg/celestia-app/pull/6438/commits/6d7a4cc14d2c21333755d8caabd041abfe244dee

x/zkism/types/ism.go
damiannolan: Addressed in https://github.com/celestiaorg/celestia-app/pull/6438/commits/724bea7893fe2cd4f47764bc8cb688fe637c0d6c.

I kept the casing here because I strongly disagree with using Java / Rust style casing for constants in Go! :D

x/zkism/types/public_values.go
Manav-Aggarwal: Agree with the `ValidateBasic` approach, keeps it as a single point of validation 👍

x/zkism/types/ism.go
```diff
@@ -10,7 +10,14 @@ import (
...
 const (
-	ModuleTypeZkISM = 42
+	// ModuleTypeZkISM defines the x/zkism module identifier.
+	// The module identifier is used to instantiate a binding between the module and the core Hyperlane router.
+	// The identifier is also encoded directly into the Hyperlane HexAddresses for ism types generated by this module.
+	//
+	// Note the uint8 42 is chosen arbitrarily and has no specific meaning. This value is outside of the range of
+	// reserved identifiers used by Hyperlane core.
+	// See: https://github.com/bcp-innovations/hyperlane-cosmos/blob/v1.0.1/x/core/01_interchain_security/types/types.go#L32-L45
+	ModuleTypeZkISM uint8 = 42
```
rootulp, line 20: 👍

Issue #6394, opened 2026-01-13
Flaky e2e tests


rootulp:
## Problem

e2e tests have been failing recentl. For example the most recent commit on main had 10 failing checks

https://github.com/celestiaorg/celestia-app/commit/9e8c67f4adfe8469745195a54092446fb2fa4913

<img width="1320" height="936" alt="Image" src="https://github.com/user-attachments/assets/656ba03a-9291-4ef4-8bf6-f088dba50a4b" />

## Proposal

It would be nice if all e2e tests passed before we create v7.0.0-rc0

PR #6443, opened 2026-01-21
fix: correct base image in multiplexer docker file

chatton:
## Overview

The multiplexer docker image was using the wrong base image to pull the v6 binary, so the "v6" embedded binary, was actually v5.

should resolve some of #6394


rootulp:
I would prefer rolling upgrade logic for the latest major version

rootulp:
Thanks for finding and fixing

Issue #6437, opened 2026-01-20
Increase the max commission rate for existing validators to 60% in the v7 upgrade handler


rootulp:
## Context

[CIP-44](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-044.md)

### Currently (v6)
- MinCommissionRate: 10%
- MaxCommissionRate (for new validators): 25% 

### Proposal (v7)
- MinCommissionRate: 20%
- MaxCommissionRate (for new validators): 60% 

## Problem

Note: v4 introduced the MaxCommissionRate (for new validators) field via [CIP-31](https://github.com/celestiaorg/CIPs/blob/main/cips/cip-031.md). However, it didn't modify any existing validators so validators can have a self-imposed max commission rate lower than that. MaxCommissionRate is not modifiable after creation.

I noticed 92 [validators](https://celenium.io/validators?status=active&page=1) have a MaxCommissionRate less than 60%. v7 currently has a change to increase the MaxCommissionRate (for new validators) to 60% but that doesn't help the 92 existing validators.

## Proposal

In the v6 to v7 upgrade handler we could increase the max commission rate of existing validators from whatever they currently have to 60%. 

Note: there is already precedent for updating commission rates via the upgrade handler. In v6 we updated the commission rate and max rate of all validators if they were lower than the v6 MinCommissionRate of 10%. See [here](https://github.com/celestiaorg/celestia-app/blob/ea0d01adefdf2c8c9e71a4ed15ccc14017228412/app/upgrades.go#L207-L210)

## Plan
- [x] Get rough consensus
- [ ] Update CIP https://github.com/celestiaorg/CIPs/pull/379
- [ ] Update implementation https://github.com/celestiaorg/celestia-app/pull/6445

rootulp:
Got rough consensus from @Bidon15 @mindstyle85 and Soph from POPs.

PR #6445, opened 2026-01-21
feat: auto migrate max commission to 60%

rootulp:
Closes https://github.com/celestiaorg/celestia-app/issues/6437

### DIFF
--- a/app/upgrades_test.go
+++ b/app/upgrades_test.go
@@ -206,7 +206,7 @@ func createValidatorWithCommission(t *testing.T, testApp *app.App, ctx sdk.Conte
 	maxRateDec, err := math.LegacyNewDecFromStr(maxRate)
 	require.NoError(t, err)
 
-	maxChangeRateDec, err := math.LegacyNewDecFromStr("0.10") // 10%
+	maxChangeRateDec := math.LegacyOneDec()
 	require.NoError(t, err)
 
 	validators, err := testApp.StakingKeeper.GetAllValidators(ctx)
@@ -242,7 +242,6 @@ func TestMaxCommissionRate(t *testing.T) {
 		// the commission rate can be updated.
 		ctx := testApp.NewContext(false).WithBlockTime(util.GenesisTime.Add(time.Hour * 25))
 
-		// Set up validator with a high max change rate to allow commission changes
 		validator := createValidatorWithCommission(t, testApp, ctx, "0.20", "1.00")
 		valAddr, err := sdk.ValAddressFromBech32(validator.GetOperator())
 		require.NoError(t, err)


app/upgrades.go
```diff
@@ -139,12 +139,12 @@ func (a App) UpdateValidatorCommissionRates(ctx context.Context) error {
 	}
 
 	for _, validator := range validators {
-		if validator.Commission.Rate.GTE(appconsts.MinCommissionRate) && validator.Commission.MaxRate.GTE(appconsts.MinCommissionRate) {
-			sdkCtx.Logger().Debug("validator commission rate and max commission rate are already greater than or equal to the minimum commission rate", "validator", validator.GetOperator())
+		if validator.Commission.Rate.GTE(appconsts.MinCommissionRate) && validator.Commission.MaxRate.GTE(stakingtypes.MaxCommissionRate) {
```
rach-id, line 142: > validator.Commission.MaxRate.GTE(stakingtypes.MaxCommissionRate)

this means the validator's max commission >= MaxCommissionRate? Probably should be the opposite

app/upgrades.go
```diff
@@ -139,12 +139,12 @@ func (a App) UpdateValidatorCommissionRates(ctx context.Context) error {
...
+			sdkCtx.Logger().Debug("validator commission rate and max commission rate are already compliant", "validator", validator.GetOperator())
 			continue
 		}
 		rate := getMax(validator.Commission.Rate, appconsts.MinCommissionRate)
-		maxRate := getMax(validator.Commission.MaxRate, appconsts.MinCommissionRate)
+		maxRate := getMax(validator.Commission.MaxRate, stakingtypes.MaxCommissionRate)
```
rach-id, line 147: probably this should be `min` not `max`?

app/upgrades_test.go
```diff
@@ -109,64 +109,60 @@ func TestApplyUpgrade(t *testing.T) {
...
+			wantRate:       "0.20",
+			wantMaxRate:    "0.60",
+			shouldUpdate:   false,
+		},
+		{
+			name:           "should not update if both are above 20% and 60%",
+			initialRate:    "0.25",
+			initialMaxRate: "0.80",
+			wantRate:       "0.25",
+			wantMaxRate:    "0.80",
```
rach-id, line 164: This should be 60%, no? The max rate is the maximum any validator can set

rach-id (changes requested):

app/upgrades.go
rootulp: if the validator's max commission is >= 60% then it's a no-op (log and continue)

If the validators max commission is < 60% then we need to modify it but that happens below

app/upgrades.go
rootulp: We want the larger value to be used as the validator's max commission

validator's current max commission -> validator's new max commission
30% -> 60%
60% -> 60%
100% -> 100%

app/upgrades_test.go
rootulp: If a validator has a max already greater than 60% then the migration shouldn't change it, so the expected value is the same as before 80%

PR #6446, opened 2026-01-21
fix: update consensus params app version before upgrade block

### LINKED ISSUE #6231: v3.x binary hits `wrong Block.Header.Version. Expected {11 1}, got {11 2}` at v2 upgrade height on Arabica
## Context

@pippokr is running a consensus snapshot node on Arabica

## Problem

The v3.x binary can't sync through the v2 upgrade height. It hits

```
9:57AM INF service start impl=MConn{195.154.218.162:26656} module=p2p msg={} peer={"id":"365b07692fdc8dd996063f970cf6788ef608ddb1","ip":"195.154.218.162","port":26656}
9:57AM ERR Error in validation err="wrong Block.Header.Version. Expected {11 1}, got {11 2}" module=blocksync
9:57AM ERR Stopping peer for error err="reactor validation error: wrong Block.Header.Version. Expected {11 1}, got {11 2}" module=p2p peer={"Data":{},"Logger":{"Logger":{}}} reactor=BlockSync
9:57AM INF service stop impl={"Logger":{"Logger":{}}} module=p2p msg={} peer={"id":"365b07692fdc8dd996063f970cf6788ef608ddb1","ip":"195.154.218.162","port":26656}
```

bit2swaz:
## Overview

Fixes #6231.

This PR fixes a critical synchronization issue where v3.x binaries fail to sync past the v2 upgrade height due to a `wrong Block.Header.Version` validation error.

### The Problem
When syncing, the node encountered the following error at the upgrade height:
`Error in validation err="wrong Block.Header.Version. Expected {11 1}, got {11 2}" module=blocksync`

This occurred because the `ConsensusParams` (specifically the App Version) were being updated in `EndBlocker` at the upgrade height (Height N). However, CometBFT validates the header of Height N using the state from Height N-1. Since the state at N-1 still contained App Version 1, validating a block header with App Version 2 failed.

### The Solution
This change shifts the update of the consensus params to **Height N-1**.

By updating the `ConsensusParams` in the `EndBlocker` of the block *prior* to the upgrade, we ensure the state is committed with the new App Version (v2) before validators construct the upgrade block.

**The corrected sequence is now:**
1.  **Height N-1:** `EndBlocker` updates consensus params to `signalUpgrade.AppVersion`.
2.  **Height N-1:** State is committed to disk with the new version.
3.  **Height N:** Validators create the block header using the new version.
4.  **Height N:** Syncing nodes validate the header against the state from N-1 (versions now match).
5.  **Height N:** `PreBlocker` runs the actual upgrade handler.

### Changes
- Adjusted `upgradeHeight` calculation to target the correct block for param updates.
- Replaced `app.SetAppVersion` with a direct update to `app.ConsensusKeeper.ParamsStore` to ensure precise control over the state commit timing relative to block validation.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -577,7 +577,7 @@ func (app *App) EndBlocker(ctx sdk.Context) (sdk.EndBlock, error) {
 		if signalUpgrade.AppVersion > currentVersion {
 			app.BaseApp.Logger().Info("upgrading app version", "current version", currentVersion, "new version", signalUpgrade.AppVersion)
 
-			upgradeHeight := signalUpgrade.UpgradeHeight + 1 // next block is performing the upgrade.
+			upgradeHeight := signalUpgrade.UpgradeHeight
 			plan := upgradetypes.Plan{
 				Name:   fmt.Sprintf("v%d", signalUpgrade.AppVersion),
 				Height: upgradeHeight,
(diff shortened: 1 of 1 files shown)


rootulp:
> This occurred because the ConsensusParams (specifically the App Version) were being updated in EndBlocker at the upgrade height (Height N). However, CometBFT validates the header of Height N using the state from Height N-1. Since the state at N-1 still contained App Version 1, validating a block header with App Version 2 failed.

This is false. The app version should be 2 and 3. This type of mistake makes me think this comment was AI generated.

rootulp:
How did you confirm this fixes the issue? We need a test for block syncing through the v3 upgrade height. We want that test to fail with existing code and pass after your fix.

PR #6475, opened 2026-01-27
feat: retain at least 3000 blocks regardless of min-retain-blocks config

### LINKED ISSUE #6388: All nodes should retain at least 3000 blocks regardless of config
## Context

https://github.com/celestiaorg/cosmos-sdk/pull/700

## Problem

It is now possible for a user to modify their config so that they no longer retain blocks within the state sync snapshot window. For example, if the user modifies the default config to set:

```
snapshot-interval = 0
snapshot-keep-recent = 0
min-retain-blocks = 1
```

## Proposal

Suggestion from Callum is to retain at least 3000 blocks regardless of what the user configures.

We likely want to deprecate the `min-retain-blocks` config or indicate to the user that the config is overriden by a hard-coded 3000 in the binary

rootulp:
Closes #6388

Add a pre-start hook that ensures nodes retain at least 3000 blocks (or the snapshot window size, whichever is larger) regardless of user configuration. This is critical for network health as it ensures nodes can sync from state sync snapshots.

Key changes:
- Add MinRetainBlocks constant (3000) to appconsts
- Create overrideMinRetainBlocks hook that:
  - Preserves 0 (prune nothing)
  - Overrides low config file values with log message
  - Errors on explicit CLI flags below minimum
  - Considers snapshot_interval × snapshot_keep_recent
- This hook is NOT bypassable unlike other config overrides


rootulp:
This shouldn't be reviewed until we have a fix for https://github.com/celestiaorg/celestia-app/issues/6476

app/default_overrides.go
```diff
@@ -300,8 +300,8 @@ func DefaultAppConfig() *serverconfig.Config {
 	// snapshots to nodes that state sync
 	cfg.StateSync.SnapshotInterval = 1500
 	cfg.StateSync.SnapshotKeepRecent = 2
-	// Set the MinRetainBlocks to 3000 blocks so that all blocks in the snapshot window are retained.
-	cfg.MinRetainBlocks = 3000
+	// Set MinRetainBlocks so that all blocks in the snapshot window are retained.
+	cfg.MinRetainBlocks = appconsts.MinRetainBlocks
```
rootulp, line 304: @claude explain this change

PR #6516, opened 2026-02-06
feat: add app side observability and gas price visualisation

ninabarbakadze:
## Overview

- [x] Add prometheus metrics to app
- [x] Track gas price in `fee.go`
- [x] Add Grafana visualisation for gas price tracking 
- [x] Add .env file generation to talis with provider specific(DO, GC) configuration
- [x] Misc talis cleanup changes


observability/docker/grafana/dashboards/celestia.json
```diff
@@ -1597,6 +1597,133 @@
...
+      "panels": [],
+      "title": "Gas Prices",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus"
+      },
+      "description": "Quantiles from gas_price_observed summary (per-tx samples). Each series is averaged across validators.",
```
rootulp: [question] it seems like gas_price_observed will be reported from all consensus nodes, not just validators.

Also why are there multiple series here that are averaged across consensus nodes? Naively it seems like there would be only one series 🤷‍♂️ 


```suggestion
      "description": "Quantiles from gas_price_observed summary (per-tx samples). Each series is averaged across consensus nodes.",
```

tools/talis/env.go
```diff
@@ -0,0 +1,105 @@
...
+			// Write .env file
+			if err := os.WriteFile(".env", []byte(envContent), 0o600); err != nil {
+				return fmt.Errorf("failed to write .env file: %w", err)
+			}
+
+			fmt.Printf("✅ Created .env template for %s\n", provider)
+			fmt.Println("\nNext steps:")
+			fmt.Println("1. Edit .env and fill in your credentials")
+			fmt.Println("2. Run: talis init -c <chain-id> -e <experiment> --with-observability --provider", provider)
+			fmt.Println("\nNote: .env file has restrictive permissions (0600) for security")
```
rootulp: This log doesn't seem necessary for users of this tool

```suggestion
```

tools/talis/env.go
```diff
@@ -0,0 +1,105 @@
...
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&provider, "provider", "p", "digitalocean", "Cloud provider (digitalocean or googlecloud)")
+
+	return cmd
+}
+
+func generateDigitalOceanEnv() string {
```
rach-id, line 57: This feels counter intuitive for me. WHat's the need for the dot env file? you would create such a file then populate it, then init talis to generate config.toml with those same fields, why not just change the fields in config.toml? why add yet another config file?

ninabarbakadze: because it tells you exactly what fields you need to populate for which provider. talis makes decisions based on what is populated in the config and a new user will likely not be aware of it. ref: https://github.com/celestiaorg/celestia-app/blob/c01e26850af6638d398c39b0b4e62b307db48cd5/tools/talis/client.go#L36

I found it convenient but if others don't that's fine we can remove it.

observability/docker/grafana/dashboards/celestia.json
ninabarbakadze: In talis all consensus nodes are validators but on mocha we would be talking to our own validator node so this won't be necessary. 

> Also why are there multiple series here that are averaged across consensus nodes?

Multiple series is just different quantiles (p50, p90, p99 IIRC for this particular one) since all nodes have these metrics enabled we will get 50 responses in a 50 node network and we just average for each quantile.

rootulp: > In talis all consensus nodes are validators but on mocha we would be talking to our own validator node so this won't be necessary.

Can a non-validator emit this metric? If yes, I think we should remove validator from the description.

> since all nodes have these metrics enabled we will get 50 responses in a 50 node network and we just average for each quantile.

It doesn't seem accurate to average p99 values across nodes because then the p99 could be skewed down much lower than the true p99.

ninabarbakadze: > Can a non-validator emit this metric?

This is for talis. All nodes will be validators outside of observability and bridge which do not emit metrics. As for mocha dashboards we will be talking to a validator node.

 > It doesn't seem accurate to average p99 values across nodes because then the p99 could be skewed down much lower than the true p99.
 
How so? The metric is emitted in ante handle's `fee.go` so all validators should be seeing the the same fees unless I'm mistaken

rootulp: > As for mocha dashboards we will be talking to a validator node.

Why can't a regular Mocha consensus node emit these metrics?

> so all validators should be seeing the the same fees unless I'm mistaken

If the original values aren't identical across nodes then the resulting average is inaccurate.

Screenshot 2026-02-10 at 6 39 27 AM

rootulp: Claude Code investigated this thread. IMO:

1. Change to use "consensus node"
2. Keep averaging but optionally explain why it's okay to average here. It's only okay because all the original values should be identical across nodes. A more accurate alternative would've been histograms but doesn't seem easy to switch so let's skip that.

## Screenshots

Screenshot 2026-02-10 at 6 45 41 AM

Screenshot 2026-02-10 at 6 48 53 AM

ninabarbakadze: > Why can't a regular Mocha consensus node emit these metrics?

It can. It just won't unless we enabled metrics for all validators which we likely won't. Currently these will work for:
- Talis: all nodes reporting these metrics are validators
- Mocha: we would have internal dashboards 

If you the purpose of this change is to underline that technically it can be reported by consensus nodes then I'm fine to do it but in practice it won't be is all I'm saying.

> A more accurate alternative would've been histograms but doesn't seem easy to switch so let's skip that.

Agreed! I tried using them but it didn't work super well. It will probably work well when we only have on mocha node reporting.

rootulp: > If you the purpose of this change is to underline that technically it can be reported by consensus nodes then I'm fine to do it but in practice it won't be is all I'm saying.

That is partially the point. 

I don't understand why consensus node operators can't or won't enable these metrics in practice. IMO celestia-app should produce metrics that can visualized by any node operator not just validators. Ideally celestia-app ships with a set of default Grafana dashboards and node operators can enable them and see them in a local Grafana instance

> It will probably work well when we only have on mocha node reporting.

Why?

ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6516/commits/9bc5fc8d33fb92266fcd896f8df5d25619770602

> Why?

because we will have one histogram instead of 50 🤷🏻‍♀️

tools/talis/env.go
rach-id: I'm not strongly against it, so I'll defer to you

ninabarbakadze: ok let's add it and if we hate it we can remove it

rootulp (approved):

Issue #6508, opened 2026-02-05
Use the gas estimation endpoint in tx client for gas price


rach-id:
Currently, we're still using the min gas price in tx client when building transactions:
https://github.com/celestiaorg/celestia-app/blob/a14a3f0bd3b2bb1980a94139c69507a1341699ab/pkg/user/tx_client.go#L435

We should update this to use the gas estimation endpoint

PR #6517, opened 2026-02-06
fix: update txclient to use gas price estimation

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/6508

### DIFF
--- a/pkg/user/tx_client_test.go
+++ b/pkg/user/tx_client_test.go
@@ -113,7 +113,6 @@ func (suite *TxClientTestSuite) TestSubmitPayForBlob() {
 
 	t.Run("submit blob with nonce error", func(t *testing.T) {
 		seqBeforeBroadcast := suite.txClient.Signer().Account(suite.txClient.DefaultAccountName()).Sequence()
-		fmt.Printf("seqBeforeBroadcast: %d\n", seqBeforeBroadcast)
 		err := suite.txClient.Signer().SetSequence(suite.txClient.DefaultAccountName(), seqBeforeBroadcast-1)
 		require.NoError(t, err)
 		resp, err := suite.txClient.SubmitPayForBlob(subCtx, blobs)


pkg/user/tx_client.go
```diff
@@ -899,22 +912,37 @@ func (client *TxClient) EstimateGasPriceAndUsage(
...
+		if err != nil {
+			return 0, 0, fmt.Errorf("re-signing with corrected sequence: %w",
+				err)
+		}
+
+		txBytes, err = client.signer.EncodeTx(txBuilder.GetTx())
+		if err != nil {
+			return 0, 0, fmt.Errorf("re-encoding tx: %w", err)
+		}
 	}
```
rootulp, line 938: [question] why does this happen in a for loop?

pkg/user/tx_client_test.go
```diff
@@ -113,7 +113,8 @@ func (suite *TxClientTestSuite) TestSubmitPayForBlob() {
 
 	t.Run("submit blob with nonce error", func(t *testing.T) {
 		seqBeforeBroadcast := suite.txClient.Signer().Account(suite.txClient.DefaultAccountName()).Sequence()
-		err := suite.txClient.Signer().SetSequence(suite.txClient.DefaultAccountName(), seqBeforeBroadcast-5)
+		fmt.Printf("seqBeforeBroadcast: %d\n", seqBeforeBroadcast)
```
rootulp: Is this residual from debugging?

```suggestion
```

If you want to preserve it, you probably want `t.Lotf("seqBeforeBroadcast: %d\n", seqBeforeBroadcast)` so this only shows up when the test fails or `go test -v`

ninabarbakadze: yes residual from debugging I will get rid of it!

pkg/user/tx_client.go
ninabarbakadze: Because when I was working on this and running experiments I could see that the mempool sequence(expected) was changing.

rootulp (approved):

PR #6560, opened 2026-02-12
chore: add support for parallel submission in latency monitor

ninabarbakadze:
## Overview

Adds the option of running multi account submission in latency monitor + talis. The only problem is that it's hard to distinguish between broadcast and confirm failures for queued transactions. For now I've tracked it as parallel submission failures but happy to change.


ninabarbakadze:
had to resolve merge conflicts

PR #6561, opened 2026-02-12
chore: parse the loki url directly from `config.json`

ninabarbakadze:
## Overview

This removes the need for having to pass --loki-url as an argument but directly parses it if present in the config.


tools/talis/latency_monitor.go
```diff
@@ -71,14 +70,21 @@ func startLatencyMonitorCmd() *cobra.Command {
...
+
+				if cfg.Observability[0].PublicIP != "" {
+					lokiURL = fmt.Sprintf("http://%s:3100", cfg.Observability[0].PublicIP)
+					fmt.Printf("Using Loki URL from observability node: %s\n", lokiURL)
+				}
 			}
 
 			latencyMonitorCmd := fmt.Sprintf(
-				"latency-monitor -k .celestia-app -a txsim -e localhost:9091 -b %d -z %d -d %s -n %s --observability-port %d 2>&1 | tee -a /root/latency-monitor-logs",
+				"stdbuf -oL latency-monitor -k .celestia-app -a txsim -e localhost:9091 -b %d -z %d -d %s -n %s --observability-port %d 2>&1 | tee -a /root/latency-monitor-logs",
```
rach-id, line 87: why this change?

ninabarbakadze: so lines are buffered immediately and we can see the logs as soon as they come in grafana

PR #6562, opened 2026-02-12
chore: talis should kill all sessions

ninabarbakadze:
## Overview

With the `&&` command I experienced the issue where if txsim session didn't exist it would not kill latency-monitor.

### DIFF
--- a/tools/talis/reset.go
+++ b/tools/talis/reset.go
@@ -53,7 +53,9 @@ func resetCmd() *cobra.Command {
 			}
 
 			cleanupScript := `
-				tmux kill-session -t app && tmux kill-session -t txsim && tmux kill-session -t latency-monitor
+				tmux kill-session -t app 2>/dev/null || true
+				tmux kill-session -t txsim 2>/dev/null || true
+				tmux kill-session -t latency-monitor 2>/dev/null || true
 				rm -rf .celestia-app logs payload payload.tar.gz /bin/celestia* /bin/txsim
 			`
 			// Run cleanup on each validator


tools/talis/reset.go
```diff
@@ -53,7 +53,9 @@ func resetCmd() *cobra.Command {
 			}
 
 			cleanupScript := `
-				tmux kill-session -t app && tmux kill-session -t txsim && tmux kill-session -t latency-monitor
+				tmux kill-session -t app 2>/dev/null || true
+				tmux kill-session -t txsim 2>/dev/null || true
+				tmux kill-session -t latency-monitor 2>/dev/null || true
```
rach-id, line 58:
```suggestion
				tmux kill-session -a
```
probably this is enough

ninabarbakadze: I know my PR title says all sessions BUT the reason I was still killing specific sessions was to maintain how it was done previously but if we don't care and we want to wipe it all then one liner is better

rach-id: if that's the case, then: 


```suggestion
				tmux kill-session -t app; tmux kill-session -t txsim; tmux kill-session -t latency-monitor
```

should be enough

ninabarbakadze: mb I'm wrong but using ; will run all commands but it can still return a non zero exit code (and print errors) if a session doesn’t exist.

with `2>/dev/null || true`  we try all kills, suppress noise, and never fail if a session is already gone.

I'm not too opinionated tho, whatever works works.

PR #6564, opened 2026-02-12
ci: use go tool golangci-lint to pin version via go.mod

rootulp:
## Summary
- Add `tool github.com/golangci/golangci-lint/v2/cmd/golangci-lint` directive to `go.mod` to pin golangci-lint v2.8.0
- Replace `golangci/golangci-lint-action` in CI with `go tool golangci-lint run --timeout 10m`
- Update `Makefile` and `multiplexer/Makefile` to use `go tool golangci-lint` instead of bare `golangci-lint`
- Remove golangci-lint from the manual install instructions in `README.md` since it's now managed via `go.mod`

This ensures CI and local `make lint` always use the same golangci-lint version, pinned in a single place (`go.mod`).

## Test plan
- [ ] Verify `go tool golangci-lint version` prints `2.8.0`
- [ ] Verify `make lint` works locally
- [ ] Verify CI lint job passes

### DIFF
--- a/test/docker-e2e/go.mod
+++ b/test/docker-e2e/go.mod
@@ -175,7 +175,7 @@ require (
 	github.com/hashicorp/go-metrics v0.5.4 // indirect
 	github.com/hashicorp/go-plugin v1.6.3 // indirect
 	github.com/hashicorp/go-safetemp v1.0.0 // indirect
-	github.com/hashicorp/go-version v1.7.0 // indirect
+	github.com/hashicorp/go-version v1.8.0 // indirect
 	github.com/hashicorp/golang-lru v1.0.2 // indirect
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -20,11 +20,7 @@ jobs:
             **/**.go
             go.mod
             go.sum
-      - uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 #v9.2.0
-        with:
-          version: v2.8.0
```
rootulp, line 25: now we no longer need to keep this version in sync in the README, CI, Makefile, etc.

Source of truth is the version in `go.mod`

rach-id (approved):

PR #6568, opened 2026-02-13
chore: remove mempool debug dashboard

ninabarbakadze:
## Overview

This dashboard is outdated because I used it for debugging mempool issues that are no longer a priority because we meet the submission KPIs therefore that work will likely be deprioritised for the foreseeable future


ninabarbakadze:
None of these metrics are being tracked in core on main and my PR adding them will likely not get merged after the KPIs. We can re-evaluate if we need these for the future tho but they're currently empty.

PR #6570, opened 2026-02-13
docs: describe min-retain-blocks override

### LINKED ISSUE #6569: Describe that minimum `MinRetainBlocks` is 3000
## Context

https://github.com/celestiaorg/celestia-app/pull/6475

## Problem

This isn't described in the v7 release notes

## Proposal

Update the v7 release notes to explain to node operators that the minimum `MinRetainBlocks` is 3000. Note: `0` is still a valid value and it's the default. `0` means prune nothing.

rootulp:
Closes #6569

## Summary

- Document the `min-retain-blocks` minimum in the v7 release notes
- Simplify the override behavior to always override 1-2999 to 3000. Previously it would override to the snapshotInterval * snapshotKeepRecent if a user configured those to be higher.

## Test plan

- [x] Existing tests updated and passing (`go test -v -run TestOverrideMinRetainBlocks ./cmd/celestia-appd/cmd/`)
- [x] `markdownlint` passes on release notes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -320,7 +320,7 @@ test-race:
 # TODO: Remove the -skip flag once the following tests no longer contain data races.
 # https://github.com/celestiaorg/celestia-app/issues/1369
 	@echo "--> Running tests in race mode"
-	@go test -timeout 15m ./... -v -race -skip "TestPrepareProposalConsistency|TestIntegrationTestSuite|TestSquareSizeIntegrationTest|TestStandardSDKIntegrationTestSuite|TestTxsimCommandFlags|TestTxsimCommandEnvVar|TestTxsimDefaultKeypath|TestMintIntegrationTestSuite|TestMaliciousTestNode|TestBigBlobSuite|TestPriorityTestSuite|TestTimeInPrepareProposalContext|TestCLITestSuite|TestTxClientTestSuite|Test_testnode|TestEvictions|TestEstimateGasUsed|TestEstimateGasPrice|TestWithEstimatorService|TestReadBlockchainHeaders|TestPrepareProposalCappingNumberOfMessages|TestGasEstimatorE2E|TestGasEstimatorE2EWithNetworkMinGasPrice|TestRejections|TestClaimRewardsAfterFullUndelegation|TestParallelTxSubmission|TestV2SubmitMethods|TestSendToSelfWithLargeFee"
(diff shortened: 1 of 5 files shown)


rach-id (commented):
why do we return an error and expect the operator to change the config? why not do the same as the other configs where we just override the value if it's invalid and output a log without stopping the node?

rootulp:
> why not do the same as the other configs where we just override the value if it's invalid and output a log without stopping the node?

This is confusing user experience. As a user, if I configured a setting in `config.toml`, I expect the binary to respect it. 

If the field is no longer configurable, we should hard-code it from the binary and remove it from the config file. Currently the mempool is still a field in the config file so I think the binary should error on start-up if a user has configured anything besides CAT. IMO it shouldn't override a user's selection.

cmwaters:
> Error on startup instead of silently overriding min-retain-blocks values below 3000, instructing node operators to update their config to 0 or >= 3000

Btw, I believe the previous behaviour here was to ignore the user if they picked a value less than the minimum (previously the evidence age) and still keep the necessary blocks. 

Allowing users to set it to 1 has the benefit that if the minimum changes, they don't have to do any updates to the config. For a user, the preference I believe is more often going to be "I just want to save the minimum amount" then "I want to save X number of recent blocks"

rach-id:
> If the field is no longer configurable, we should hard-code it from the binary and remove it from the config file. Currently the mempool is still a field in the config file so I think the binary should error on start-up if a user has configured anything besides CAT. IMO it shouldn't override a user's selection.

I'm more inclined towards reducing the users choices. IMO, we should remove these configs from the config files instead of asking users to pick the right value. The less choices operators have, the easier it is to reason about the network

cmwaters:
With pruning, users should still have the option to choose the value

rootulp:
> With pruning, users should still have the option to choose the value

Agreed. Unfortunately I think this config isn't entirely eliminate-able yet. 

>  I believe is more often going to be "I just want to save the minimum amount" then "I want to save X number of recent blocks"

Yea I can see an argument for this. The motivation for this PR is that I found it difficult to reason about all the config fields because the binary overrides the user selection for a bunch of configs. 

One idea to accommodate a user's preference for "save the minimum amount" is to add a new config like:

```toml
block-retention-strategy = "keep-all" # or "keep-minimum"
```

and eliminate the `min-retain-blocks` as a config option. That way we can change the number "keep-minimum" maps to without user intervention. Since proposal is a new breaking change and the ship has sailed for v7, I won't pursue it. 

I'll update the implementation to revert back to overriding the user's selection because both @cmwaters and @rach-id mentioned it.

PR #6574, opened 2026-02-14
fix: warn and override non-CAT mempool type

### LINKED ISSUE #6573: Deprecate and remove the non CAT mempools
## Context

1. https://github.com/celestiaorg/celestia-core/issues/1382
2. https://github.com/celestiaorg/celestia-core/issues/2802

Since v6 the default mempool has been CAT. The binary overrides the user configured mempool if it isn't CAT: https://github.com/celestiaorg/celestia-app/blob/5bd6b866a0d0b9442d605047272107b0f8e0cc02/cmd/celestia-appd/cmd/override_p2p_config.go#L69-L75

## Problem

I think it's confusing to have a config that the binary overrides. Either the config field is configurable (and the binary emits a warning / error for unsupported values) or the config field isn't configurable and should be removed from the config. Overriding the config may confuse the user b/c the value in their config file isn't actually being used by the binary. 

Instead of the overriding, I would've preferred if v6 deprecated the non CAT mempools. Concretely, in v6 if a user configures a non CAT mempool, then instead of logging this override, it should've logged a deprecation warning.

Regardless, node operators are expected to upgrade to v7 soon.

## Proposal

Since v6 overrode the config value, IMO we can proceed with removing the non CAT mempools in celestia-core. Concretely:

In v7:
1. Delete the non CAT mempools from celestia-core
2. Error on start-up if user configures a non CAT mempool
3. Remove the mempool overriding code
4. Update v7 release notes to explain that non CAT mempools are no longer supported

rootulp:
Closes #6573
Blocked on a release with https://github.com/celestiaorg/celestia-core/pull/2811

## Summary

- Replace the silent override of non-CAT mempool types with a warning log and override to CAT
- Add `overrideMempoolType` function that runs before the bypass-config-overrides flag check (non-CAT mempools are fundamentally unsupported and must not be bypassable)
- Remove the silent mempool type override from `overrideMempoolConfig`
- Update release notes to document the new behavior

A companion PR in celestia-core removes the mempool `type` field from the config.toml template so new config files no longer expose it: https://github.com/celestiaorg/celestia-core/pull/2811

## Test plan

- [x] `go test -v -run TestOverrideP2PConfig ./cmd/celestia-appd/cmd/` — all tests pass
- [x] `go test -v -run TestOverrideMempoolType ./cmd/celestia-appd/cmd/` — unit tests pass
- [x] `markdownlint docs/release-notes/release-notes.md` — no lint errors
- [ ] Verify that a node with `type = "flood"` in config.toml starts with a warning and overrides to CAT
- [ ] Verify that a node with `type = "cat"` starts normally without a warning
- [ ] Verify that `--bypass-config-overrides` does NOT bypass the mempool type override


docs/release-notes/release-notes.md
```diff
@@ -24,7 +24,12 @@ No manual action is required, but validators should be aware of this change.
 
 #### Config Changes
 
-No configuration changes are required for v7. Existing v6 configurations remain compatible.
+celestia-app now returns a startup error if the mempool type is not `cat`. Previously, non-CAT mempool types (e.g. `flood`) were silently overridden to `cat` at startup. If your `config.toml` uses a different mempool type, update the `[mempool]` section:
```
rootulp: This isn't technically true because the override would log a message.

rootulp:
FLUPs:
1. Remove the mempool config field
2. Make the default config file not have a mempool config
3. Add a warning if a user still has a mempool specified in config file

ninabarbakadze (approved):

PR #6592, opened 2026-02-16
test: add forwarding relayer to hyperlane e2e suite

damiannolan:
## Overview

Requires:
- https://github.com/celestiaorg/forwarding-relayer/pull/9
- https://github.com/celestiaorg/tastora/pull/176

Adapts the Hyperlane E2E test for `x/forwarding` to depend on a ForwardingRelayer service. 
The relayer service is deployed with an associated backend and is maintained here: https://github.com/celestiaorg/forwarding-relayer

ref: https://linear.app/celestia/issue/INTEROP-20/use-the-interop-6-in-the-interop-e2e-tests


damiannolan:
Rather than adding an entire new test with the additional services, I just adapted the existing one to instead of manually trigger the forwarding tx, allow the relayer service to do it.

PR #6596, opened 2026-02-17
fix: correct version check in verifyUpdatedConfigs test

braveocheretovych:
Fix mismatch between test version (6) and verifyUpdatedConfigs check (v6),
which prevented assertions from running.


cmd/celestia-appd/cmd/update_config_test.go
```diff
@@ -128,7 +128,7 @@ func verifyUpdatedConfigs(t *testing.T, configDir, version string) {
 	cometConfigPath := filepath.Join(configDir, "config.toml")
 	appConfigPath := filepath.Join(configDir, "app.toml")
 
-	if version == "v6" {
```
rootulp, line 131: 😬 yikes this unit test previously didn't work as expected

braveocheretovych: thanks for your help

Issue #6601, opened 2026-02-18
Telemetry causes "duplicate metrics collector registration" error in multiplexer mode


tty47:
### Description

When running a validator node with multiplexer mode enabled and `telemetry.enabled = true` in `app.toml`, the node fails to start with the error:

```
Error: duplicate metrics collector registration attempted
```

The child process (ABCI app) becomes a zombie and the parent process hangs indefinitely.

### celestia-app version

v6.4.10-mocha (also likely affects other versions using multiplexer mode)

### Network

mocha-4

### Steps to Reproduce

1. Configure a validator node with multiplexer mode (default for validators)
2. Set `telemetry.enabled = true` in `app.toml`
3. Start the node with:
   ```bash
   celestia-appd start --home /home/celestia/.celestia-app --address=tcp://127.0.0.1:26658
   ```
4. Observe the error in logs

### Expected Behavior

The node should start normally with telemetry metrics available for Prometheus scraping.

### Actual Behavior

- Child process logs `Error: duplicate metrics collector registration attempted` and exits
- Child process becomes a zombie (`<defunct>`)
- Parent process hangs after `service start impl=PubSub`
- Node never syncs or produces blocks

### Logs

```
INF initializing multiplexer app_version=3 chain_id=mocha-4 module=server
INF initialized remote app client address=127.0.0.1:26658 module=server
INF starting comet node module=server
INF Since the chainID is mocha-4, configuring the default v2 upgrade height to 2585031
INF starting ABCI without Tendermint
Error: duplicate metrics collector registration attempted
INF service start impl=multiAppConn module=proxy msg={}
INF service start connection=query impl=localClient module=abci-client msg={}
INF service start connection=snapshot impl=localClient module=abci-client msg={}
INF service start connection=mempool impl=localClient module=abci-client msg={}
INF service start connection=consensus impl=localClient module=abci-client msg={}
INF service start impl=EventBus module=events msg={}
INF service start impl=PubSub module=pubsub msg={}
[hangs here - no further progress]
```

Process state showing zombie:
```
ps aux | grep celestia-appd
celestia 2269545 19.3  0.5 ... /usr/local/bin/celestia-appd start ...
celestia 2269569  8.9  0.0      0     0 ?  Z  [celestia-appd] <defunct>
```

### Root Cause Analysis

In multiplexer mode, `celestia-appd` spawns two processes:
1. **Parent**: Runs CometBFT/consensus
[...]

PR #6610, opened 2026-02-20
fix: duplicate metrics collector registration for multiplexer

rootulp:
## Summary

Fixes https://github.com/celestiaorg/celestia-app/issues/6601

In multiplexer mode, when `telemetry.enabled = true` and `prometheus-retention-time > 0` in `app.toml`, the node fails with `"duplicate metrics collector registration attempted"`. `telemetry.New()` registers a `PrometheusSink` on the global `prometheus.DefaultRegisterer`; a second registration in the same process fails.

Two-part fix:

- **Parent process**: Add `initTelemetry()` method on `Multiplexer` that is idempotent — it initializes telemetry once and caches `*telemetry.Metrics` so subsequent calls during version switches are no-ops.
- **Child process**: Pass an environment variable to embedded child processes that sets `prometheus-retention-time=0` via Viper's `AutomaticEnv()`, preventing the child from registering a competing Prometheus sink. The env var name is derived dynamically from the binary's filename (e.g. `CELESTIA_APPD_TELEMETRY_PROMETHEUS_RETENTION_TIME=0`).

## Changes

- `multiplexer/abci/multiplexer.go`: Add `metrics` field to `Multiplexer` struct, extract `initTelemetry()` method with idempotent caching
- `multiplexer/appd/run.go`: Add `telemetryDisableEnv()` method and set `cmd.Env` in `Start()`
- `multiplexer/abci/multiplexer_test.go`: Replace `TestDuplicateTelemetryRegistration` with `TestInitTelemetryIsIdempotent` that exercises the actual fix (fails without the caching guard, passes with it)
- `multiplexer/appd/telemetry_test.go`: Add `TestTelemetryDisableEnv`
- `docker/mocha/docker-compose.yml`: Build from source using `multiplexer.Dockerfile` instead of downloading a pre-built release
- `docker/mocha/celestia/entrypoint.sh`: Re-enable state sync, keep telemetry enabled
- `scripts/mocha.sh`: Enable telemetry with Prometheus for local testing

## Test plan

- [x] `go test -v -run TestInitTelemetryIsIdempotent ./multiplexer/abci/` passes
- [x] `go test -v -run TestTelemetryDisableEnv ./multiplexer/appd/` passes
- [x] `TestInitTelemetryIsIdempotent` fails when the caching guard is removed (regression-proof)
- [x] `cd docker/mocha && docker compose down -v && docker compose up --build` starts without "duplicate metrics collector registration attempted" error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -320,7 +320,7 @@ test-race:
 # TODO: Remove the -skip flag once the following tests no longer contain data races.
 # https://github.com/celestiaorg/celestia-app/issues/1369
 	@echo "--> Running tests in race mode"
-	@go test -timeout 15m ./... -v -race -skip "TestPrepareProposalConsistency|TestIntegrationTestSuite|TestSquareSizeIntegrationTest|TestStandardSDKIntegrationTestSuite|TestTxsimCommandFlags|TestTxsimCommandEnvVar|TestPriorityTestSuite|TestTimeInPrepareProposalContext|TestTxClientTestSuite|TestEvictions|TestEstimateGasUsed|TestPrepareProposalCappingNumberOfMessages|TestRejections|TestClaimRewardsAfterFullUndelegation|TestParallelTxSubmission|TestBigBlobSuite|TestTxsimDefaultKeypath|TestGasEstimatorE2E|TestMintIntegrationTestSuite|TestSubmitPayForBlobWithEstimatorService|TestSendToSelfWithLargeFee|TestV2SubmitMethods"
+	@go test -timeout 15m ./... -v -race -skip "TestPrepareProposalConsistency|TestIntegrationTestSuite|TestSquareSizeIntegrationTest|TestStandardSDKIntegrationTestSuite|TestTxsimCommandFlags|TestTxsimCommandEnvVar|TestPriorityTestSuite|TestTimeInPrepareProposalContext|TestTxClientTestSuite|TestEvictions|TestEstimateGasUsed|TestPrepareProposalCappingNumberOfMessages|TestRejections|TestClaimRewardsAfterFullUndelegation|TestParallelTxSubmission|TestBigBlobSuite|TestTxsimDefaultKeypath|TestGasEstimatorE2E|TestMintIntegrationTestSuite|TestSubmitPayForBlobWithEstimatorService|TestSendToSelfWithLargeFee|TestV2SubmitMethods|TestCLITestSuite"
 .PHONY: test-race
 
 ## test-bench: Run benchmark unit tests.


rootulp:
@ninabarbakadze I reproduced with the docker-setup in the previous commit

```
celestia-appd  | Starting celestia-appd...
celestia-appd  | 4:23AM INF initializing multiplexer app_version=1 chain_id=mocha-4 module=server
celestia-appd  | 4:23AM INF initialized remote app client address=127.0.0.1:36658 module=server
celestia-appd  | 4:23AM INF starting comet node module=server
celestia-appd  | 4:23AM INF Since the chainID is mocha-4, configuring the default v2 upgrade height to 2585031
celestia-appd  | 4:23AM INF starting ABCI without Tendermint
celestia-appd  | Error: duplicate metrics collector registration attempted
celestia-appd  | 4:23AM INF service start impl=multiAppConn module=proxy msg={}
celestia-appd  | 4:23AM INF service start connection=query impl=localClient module=abci-client msg={}
celestia-appd  | 4:23AM INF service start connection=snapshot impl=localClient module=abci-client msg={}
celestia-appd  | 4:23AM INF service start connection=mempool impl=localClient module=abci-client msg={}
celestia-appd  | 4:23AM INF service start connection=consensus impl=localClient module=abci-client msg={}
celestia-appd  | 4:23AM INF service start impl=EventBus module=events msg={}
celestia-appd  | 4:23AM INF service start impl=PubSub module=pubsub msg={}
```

rootulp:
Observation: to repro via Docker compose and a local `mocha.sh` script I had to disable state-sync.

rootulp:
Confirmed the fix.

docker/mocha/celestia/entrypoint.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-# Entrypoint script for a Mocha consensus node with state sync.
+# Entrypoint script for a Mocha consensus node using the multiplexer binary.
+# Enables telemetry with Prometheus to verify the fix for
+# https://github.com/celestiaorg/celestia-app/issues/6601
```
rach-id: [nit]
```suggestion
```

Seems unnecessary IMO

multiplexer/appd/run.go
```diff
@@ -54,9 +56,23 @@ func New(version string, compressedBinary []byte) (*Appd, error) {
...
+	prefix := strings.ToUpper(replacer.Replace(basename))
+	return []string{
+		prefix + "_TELEMETRY_PROMETHEUS_RETENTION_TIME=0",
+	}
+}
+
 // Start starts the appd binary with the given arguments.
 func (a *Appd) Start(args ...string) error {
 	cmd := exec.Command(a.path, append([]string{"start"}, args...)...)
+	cmd.Env = append(os.Environ(), a.telemetryDisableEnv()...)
```
rach-id: shouldn't this also append the existing cmd.Env? not to loose any other env set elsewhere?

rootulp: looks like `os.Environ()` already contains env variables set elsewhere

rach-id: so, we make the assumption that all `cmd.Env` are in `os.Environ()`? Also, why wouldn't this be:
```suggestion
	cmd.Env = append(cmd.Env, a.telemetryDisableEnv()...)
```
as the previous code adds all env from OS instead of just the telemetry disable thing?

rootulp: Explanation from Claude below. I added a unit test for env variable handling

## Screenshot

Image

multiplexer/abci/multiplexer.go
```diff
@@ -623,8 +626,21 @@ func (m *Multiplexer) stopTraceWriter() error {
 	return m.traceWriter.Close()
 }
 
-func startTelemetry(cfg serverconfig.Config) (*telemetry.Metrics, error) {
-	return telemetry.New(cfg.Telemetry)
+// initTelemetry initializes telemetry if it hasn't been initialized yet.
+// It is idempotent: subsequent calls are no-ops. This prevents "duplicate
+// metrics collector registration attempted" errors when
+// enableGRPCAndAPIServers is called more than once (e.g. during a version
+// switch).
```
ninabarbakadze, line 633: we could probably make this less verbose

docker/mocha/celestia/entrypoint.sh
```diff
@@ -23,30 +24,29 @@ if [ ! -f "${CELESTIA_APP_HOME}/config/config.toml" ]; then
...
-    s|^(trust_hash[[:space:]]+=[[:space:]]+).*$|\1\"${TRUST_HASH}\"|" "${CELESTIA_APP_HOME}/config/config.toml"
-
     echo "Enabling Prometheus instrumentation..."
     sed -i '/^\[instrumentation\]/,/^\[/{s/^prometheus *=.*/prometheus = true/;}' "${CELESTIA_APP_HOME}/config/config.toml"
 
+    echo "Enabling telemetry in app.toml..."
+    sed -i '/^\[telemetry\]/,/^\[/{s/^enabled *=.*/enabled = true/;}' "${CELESTIA_APP_HOME}/config/app.toml"
+    sed -i '/^\[telemetry\]/,/^\[/{s/^prometheus-retention-time *=.*/prometheus-retention-time = 60/;}' "${CELESTIA_APP_HOME}/config/app.toml"
+
+    echo "Enabling state sync in config.toml..."
```
ninabarbakadze, line 40: so we need to enable state sync after telemetry?

ninabarbakadze (approved):

docker/mocha/celestia/entrypoint.sh
rootulp: Not really. To repro the issue I had to disable state sync. After reproducing and verifying the fix, I reverted this so that the docker/mocha setup still uses state sync.

PR #6611, opened 2026-02-20
test: spam submit txs in tastora

### LINKED ISSUE #5108: Spam submitting many blobs per block per account

ninabarbakadze:
## Overview

Fixes #5108 

This test should act as the first line of defence when(if) we introduce any regressions. We should likely also point this test to Arabica. cc @rootulp @rach-id 

I set the workflow to trigger on push while testing and this is how it looks: https://github.com/celestiaorg/celestia-app/actions/runs/22226059349/job/64292929990?pr=6611

Flup https://github.com/celestiaorg/celestia-app/issues/6613


test/docker-e2e/dockerchain/testchain.go
```diff
@@ -79,9 +82,15 @@ func getPostInitModifications(gasPrices string) []func(context.Context, *tastora
 			cfg.P2P.AllowDuplicateIP = true
 			cfg.P2P.AddrBookStrict = false
 			cfg.Storage.DiscardABCIResponses = false
-			blockTime := time.Duration(2) * time.Second
-			cfg.Consensus.TimeoutCommit = blockTime
-			cfg.Consensus.TimeoutPropose = blockTime
+			timeoutCommit := 2 * time.Second
```
rach-id: why are you using a 2s timeout commit?

test/docker-e2e/dockerchain/config.go
```diff
@@ -21,10 +22,12 @@ const (
...
-	Image           string
-	Tag             string
-	DockerClient    tastoratypes.TastoraDockerClient
-	DockerNetworkID string
+	Image               string
+	Tag                 string
+	DockerClient        tastoratypes.TastoraDockerClient
+	DockerNetworkID     string
+	AdditionalStartArgs []string
+	TimeoutPropose      time.Duration // If not set the default is used
```
rach-id: why only this is configurable?

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -79,9 +82,15 @@ func getPostInitModifications(gasPrices string) []func(context.Context, *tastora
...
+			cfg.Consensus.TimeoutCommit = timeoutCommit
+
+			timeoutPropose := 2 * time.Second
```
rach-id: why 2s timeout propose?

rootulp:
test-race is failing on an unrelated test: https://github.com/celestiaorg/celestia-app/issues/6614 I'll fix separately

test/docker-e2e/e2e_tx_submission_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	ctx := context.Background()
+
+	t.Logf("Starting tx submission test: %d MiB blob size, %v duration",
+		blobSize/(1024*1024), testDuration)
+
+	cfg := dockerchain.DefaultConfig(s.client, s.network).
+		WithTag(dockerchain.GetCelestiaTag()).
+		WithTimeoutPropose(30*time.Second).
+		WithAdditionalStartArgs("--delayed-precommit-timeout", "2500ms")
```
rach-id: why not use the constant defined in app_consts?

test/docker-e2e/latency_monitor_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"time"
+
+	"celestiaorg/celestia-app/test/docker-e2e/dockerchain"
+
+	tastoracontainertypes "github.com/celestiaorg/tastora/framework/docker/container"
+	tastoratypes "github.com/celestiaorg/tastora/framework/types"
+	"github.com/stretchr/testify/require"
+)
+
+const latencyMonitorImage = "ghcr.io/celestiaorg/latency-monitor"
```
rach-id, line 24: is this published with every commit?

test/docker-e2e/e2e_tx_submission_test.go
```diff
@@ -0,0 +1,82 @@
...
+	t := s.T()
+
+	ctx := context.Background()
+
+	t.Logf("Starting tx submission test: %d MiB blob size, %v duration",
+		blobSize/(1024*1024), testDuration)
+
+	cfg := dockerchain.DefaultConfig(s.client, s.network).
+		WithTag(dockerchain.GetCelestiaTag()).
+		WithTimeoutPropose(30*time.Second).
```
rootulp: [question] why is timeout propose 30 seconds? That doesn't seem like it reflects the values on testnets / mainnet

test/docker-e2e/e2e_tx_submission_test.go
rootulp: [question] why is delayed precommit timeout 2500ms? Doesn't seem reflective of testnet / mainnet conditions.

Is this test assuming 32 MB / 3 seconds?

rootulp:
Screenshot 2026-02-20 at 8 04 46 AM

Can the workflow name exactly match the test name? In other words rename `tx-submission-test` to `TestTxSubmission` so that it's easy to find one from the other?

rootulp (dismissed):
LGTM but the TimeoutPropose doesn't match real networks which makes me think the results here won't match what we get on real networks.

Is there anyway to eliminate the timeout overrides that we apply here? 
1. TimeoutPropose
2. DelayedPrecommitTimeout <- if this is just for 3s block times instead of 6s block times that seems fine

rootulp:
I see https://github.com/celestiaorg/celestia-app/issues/6613 as a FLUP. Does that mean the test in this PR won't get run anywhere yet? I assume it won't run by the normal test-docker-e2e workflow that runs on each PR because we haven't added it to that list.

rach-id (approved):
utAck

ninabarbakadze:
> from

Yup in my next PR where I'm adding the workflow I will make sure the name matches.

test/docker-e2e/dockerchain/testchain.go
ninabarbakadze: It was previously set as 2 seconds. I don't know the reasoning for this.

test/docker-e2e/dockerchain/testchain.go
ninabarbakadze: It was previously 2 seconds if you look at the diff. I'm really not sure what the rationale was there.

test/docker-e2e/e2e_tx_submission_test.go
ninabarbakadze: it's set to 6s

ninabarbakadze:
> I see #6613 as a FLUP. Does that mean the test in this PR won't get run anywhere yet? I assume it won't run by the normal test-docker-e2e workflow that runs on each PR because we haven't added it to that list.

It won't get run until https://github.com/celestiaorg/celestia-app/pulls merges. IIRC we didn't wan to run this test on every PR but only nightly.

test/docker-e2e/dockerchain/testchain.go
rach-id: it shouldn't IMO, we should have it equal to a realistic network's timeout commit. Otherwise, the transaction latency results won't be representative at all

test/docker-e2e/e2e_tx_submission_test.go
rach-id: Probably a follow up should be to update this value to point to the app const once it's reduced to 3s

test/docker-e2e/latency_monitor_test.go
ninabarbakadze: no it's built locally for each nightly run which is really fast.

test/docker-e2e/e2e_tx_submission_test.go
ninabarbakadze: oh if it was reduced I can do it in this PR!

test/docker-e2e/dockerchain/config.go
ninabarbakadze: Because I wanted to change it 2s timeout was too aggressive. I can revert this completely and set it to default.

test/docker-e2e/dockerchain/testchain.go
ninabarbakadze: Do we want to set everything to correspond to our app_consts/config? I'd be for this as well.

rach-id: I'm inclined towards this, especially the timeout commit

test/docker-e2e/e2e_tx_submission_test.go
rach-id: it's still not reduced

rootulp:
> It won't get run until https://github.com/celestiaorg/celestia-app/pulls merges.

That link seems incorrect. Did it mean to point to https://github.com/celestiaorg/celestia-app/issues/6613? 

My understanding is this test won't run anywhere until https://github.com/celestiaorg/celestia-app/issues/6613 is addressed.

ninabarbakadze:
> > It won't get run until https://github.com/celestiaorg/celestia-app/pulls merges.
> 
> That link seems incorrect. Did it mean to point to #6613?
> 
> My understanding is this test won't run anywhere until #6613 is addressed.

Yes I meant to point it to https://github.com/celestiaorg/celestia-app/issues/6613. Also, before this PR merges I want to revert config overrides in tastora. Currently chatting with Cian to understand why we even did this in the first place.

test/docker-e2e/e2e_tx_submission_test.go
ninabarbakadze: Considering this addressed because we set it to app consts.

test/docker-e2e/e2e_tx_submission_test.go
ninabarbakadze: Considering this addressed because we set it to app consts.

test/docker-e2e/dockerchain/testchain.go
ninabarbakadze: Considering this addressed because we set it to app consts.

test/docker-e2e/dockerchain/config.go
```diff
@@ -83,18 +81,6 @@ func (c *Config) WithDockerNetworkID(networkID string) *Config {
...
-func (c *Config) WithAdditionalStartArgs(args ...string) *Config {
-	c.AdditionalStartArgs = append(c.AdditionalStartArgs, args...)
-	return c
-}
-
-// WithTimeoutPropose sets the consensus proposal timeout.
-func (c *Config) WithTimeoutPropose(d time.Duration) *Config {
-	c.TimeoutPropose = d
-	return c
-}
```
evan-forbes, line 96: seems unrelated, will defer to author on if we keep this, but technically this is a breaking change so "test!: spam ...."

evan-forbes (approved):

test/docker-e2e/dockerchain/config.go
ninabarbakadze: I think this diff is outdated!

PR #6612, opened 2026-02-20
chore: run the tx submission test nightly

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/6613

### DIFF
--- a/.github/workflows/test-nightly.yml
+++ b/.github/workflows/test-nightly.yml
@@ -147,10 +147,11 @@ jobs:
   notify-slack-on-failure:
     name: Notify Slack on failure
     runs-on: ubuntu-latest
-    needs: [test-sync-to-tip-mocha, test-all-upgrades, test-tx-submission]
+    needs: [test-state-sync-mocha, test-sync-to-tip-mocha, test-all-upgrades, test-tx-submission]
     if: >-
       ${{ always() &&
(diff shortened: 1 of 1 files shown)


.github/workflows/test-nightly.yml
```diff
@@ -111,11 +110,50 @@ jobs:
...
+        run: |
+          if [ -n "${{ inputs.tag }}" ]; then
+            echo "value=${{ inputs.tag }}" >> $GITHUB_OUTPUT
+          else
+            echo "value=$(echo $GITHUB_SHA | cut -c1-8)" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Build Docker image locally
+        if: ${{ inputs.tag == '' }}
+        run: docker build -t "ghcr.io/celestiaorg/celestia-app:${{ steps.tag.outputs.value }}" . -f docker/multiplexer.Dockerfile
```
evan-forbes, line 136: are we doing this for each nightly test? is there a way to only do this once and then run the test we want?

evan-forbes (approved):

rootulp (approved):

.github/workflows/test-nightly.yml
ninabarbakadze: yes I was also thinking that. I'll flup it!

PR #6631, opened 2026-02-23
fix: use correct p2p port for state sync peer

Ocheretovich:
Fix incorrect persistent peer port for local state sync node. The state-sync node listens on 36656 but the script was connecting to 127.0.0.1:26656, causing state sync to fail or hang in local setups.


scripts/single-node-state-sync.sh
```diff
@@ -45,7 +45,7 @@ s|^(rpc_servers[[:space:]]+=[[:space:]]+).*$|\1\"$RPC,$RPC\"| ; \
 s|^(trust_height[[:space:]]+=[[:space:]]+).*$|\1$BLOCK_HEIGHT| ; \
 s|^(trust_hash[[:space:]]+=[[:space:]]+).*$|\1\"$TRUST_HASH\"|" $CELESTIA_APP_HOME/config/config.toml
 
-PEER=$(curl -s http://${RPC}/status | jq -r '.result.node_info.id + "@127.0.0.1:26656"')
+PEER=$(curl -s http://${RPC}/status | jq -r '.result.node_info.id + "@127.0.0.1:36656"')
```
rach-id, line 48: no, this port is for the node started using the `./single-node.sh` script, not the one started in this script

Issue #6633, opened 2026-02-24
Revert tastora config overrides to correspond to celestia netowk


ninabarbakadze:
## Description

Get rid of these overrides. The values from `app_consts` should be used. https://github.com/celestiaorg/celestia-app/blob/429d84e4746f0ffcbb1e65951cb1880766a0cedb/test/docker-e2e/dockerchain/testchain.go#L82

ninabarbakadze:
context: I spoke with @chatton and the reason why some of the overrides were done was to speed up the network. We will revert them and resolve as we go.

rootulp:
@ninabarbakadze will reverting slow down other tests? Perhaps some e2e tests should override the defaults to produce fast blocks and execute the tests quickly. Other tests that depend on a block time similar to Mainnet should not override the defaults.

PR #6634, opened 2026-02-24
fix: tastora should not override consensus consts

ninabarbakadze:
## Overview

Fixes https://github.com/celestiaorg/celestia-app/issues/6633

Unfortunately I couldn't delete the oveerides completely, because [Tastora's `setTestConfig()` in node.go](https://github.com/celestiaorg/tastora/blob/e1756419f62642c4fa0edd3b2f086b2b6a306ebf/framework/docker/cosmos/node.go#L316) hardcodes `TimeoutCommit` and `TimeoutPropose` to 2s, overriding the defaults written by `celestia-appd init`. We re override  them here with appconsts values so the e2e tests use production like consensus timeouts.


test/docker-e2e/dockerchain/testchain.go
```diff
@@ -79,9 +78,6 @@ func getPostInitModifications(gasPrices string) []func(context.Context, *tastora
 			cfg.P2P.AllowDuplicateIP = true
 			cfg.P2P.AddrBookStrict = false
 			cfg.Storage.DiscardABCIResponses = false
-			blockTime := time.Duration(2) * time.Second
```
ninabarbakadze, line 82: do we think that anything else needs to be updated?

rach-id: delayed precommit?

test/docker-e2e/dockerchain/testchain.go
```diff
@@ -79,9 +79,8 @@ func getPostInitModifications(gasPrices string) []func(context.Context, *tastora
...
-			cfg.Consensus.TimeoutCommit = blockTime
-			cfg.Consensus.TimeoutPropose = blockTime
+			cfg.Consensus.TimeoutCommit = appconsts.TimeoutCommit
+			cfg.Consensus.TimeoutPropose = appconsts.TimeoutPropose
```
rach-id, line 83:
```suggestion
			cfg.Consensus.TimeoutCommit = appconsts.TimeoutCommit
			cfg.Consensus.TimeoutPropose = appconsts.TimeoutPropose
			cfg.Consensus.DelayedPrecommitTimeout = appconsts.DelayedPrecommitTimeout
```

ninabarbakadze: I was already setting this temporarily in my other PR but I guess doesn't hurt to set this and forget about it rather than having to do flups

ninabarbakadze:
> No blocking feedback but can you please investigate if this change slows down the other e2e tests? If it does then we likely want a follow-up issue to decrease the block time in some e2e tests (the tests that don't need a 3s or 6s block time).
> 
> The tests that depend on Mainnet block time (3s or 6s) should likely use the default timeouts from appconsts (like this PR).

I'll benchmark all e2e tests on my `working branch` and `main`. We can compare those results and go from there. If there's no significant slowdown because of 3 second block time we can leave everything as is.

ninabarbakadze:
Okay so! I benchmarked tests on this PR against main and realised that ALL tests had the same duration **even if I overrode `DelayedPrecommitTimeout` to 3s**. Then I checked block times on this branch and they were still 6s. From here I realised that overriding `DelayedPrecommitTimeout` in the config won't do anything because if it wasn't passed to the app start up, app will automatically set it to `app_consts` value. You can only override this through the CLI flag.

AFAIK The only thing 2s TimeoutPropose was doing was that even if we submitted transactions nodes didn't have enough time to complete proposals so they got voted nil. cc @rootulp @rach-id

rootulp:
> I realised that overriding DelayedPrecommitTimeout in the config won't do anything because if it wasn't passed to the app start up

Why isn't the config passed to the app start up? Seems odd that this file is setting config values if the config isn't being used

ninabarbakadze:
> DelayedPrecommitTimeout

This is only specific to `DelayedPrecommitTimeout` which can only be overridden via `--delayed-precommit-timeout` CLI flag, not via config.toml. If the flag isn't passed the app defaults it to appconsts (5.85s) and returns that via ABCI.

rootulp:
Do we want a follow-up issue to make it possible to override DelayedPrecommitTimeout via the config file? I find it unintuitive that some configs are CLI only.

ninabarbakadze:
> Do we want a follow-up issue to make it possible to override DelayedPrecommitTimeout via the config file? I find it unintuitive that some configs are CLI only.

Yeah, not sure. I was wondering if it was by design that you can't easily modify `DelayedPrecommitTimeout`.

rach-id:
> Do we want a follow-up issue to make it possible to override DelayedPrecommitTimeout via the config file? I find it unintuitive that some configs are CLI only.

Yes, we specifically made it impossible to change this using config files because operators would fail to update them and they affect heavily the network

rootulp:
> Yes, we specifically made it impossible to change this using config files because operators would fail to update them and they affect heavily the network

If they can't be changed by the config file, let's have a follow up to remove them from the config file.

Alternatively, we could make them configurable in the config file and throw an error on start-up if someone tries to configure the DelayedPrecommitTimeout to a non default value for the public networks (Arabica, Mocha, Mainnet). This allows us to configure DelayedPrecommitTimeout on developer testnets (talis, tastora, locally) but not allow a node operator to configure on a public network.

rach-id:
good idea, we could automatically override them with the chainID is a public chain, otherwise use the value from the config file

rootulp:
> we could automatically override them with the chainID is a public chain

I prefer an error on start-up if the config isn't valid rather than automatically overriding a node operator's config. Error on start-up informs the operator that their config is no longer being respected. If we override, a node operator may miss the log that says "Overriding x to y..." and lead to unexpected behavior.

PR #6637, opened 2026-02-24
ci: add assign random reviewer workflow

tty47:
Automatically assigns one random reviewer from CODEOWNERS when a PR is opened.

  - Triggers on PR open, ready_for_review, and reopen events
  - Skips draft PRs
  - Parses CODEOWNERS to find eligible reviewers for changed files
  - Excludes PR author from candidates
  - Picks one reviewer at random

### DIFF
--- a/.github/workflows/assign-random-reviewer.yml
+++ b/.github/workflows/assign-random-reviewer.yml
@@ -1,0 +1,113 @@
+name: Assign Random Reviewer
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review, reopened]
+
+jobs:
+  assign-reviewer:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    # Skip draft PRs
+    if: github.event.pull_request.draft == false
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


tty47:
> Will this also be adopted in app? https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team or we keep having this workflow?

this is interesting! I cannot see it in the settings (might be permissions), but looks interesting

rootulp:
Seems like we can get rid of the workflow if we use the Github native auto assingment.

Ref: https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team#configuring-auto-assignment

Can be a follow-up

cmwaters:
> Can be a follow-up

~I can look into this. I believe I should have permissions now~

I don't have permissions

rootulp:
This PR merged last week but all PRs since then still have all CODEOWNERs assigned so I don't think this works as expected.

@tty47 @cmwaters

rootulp:
Why don't we just use the native Github solution that Rachid mentioned with round-robin?

PR #6638, opened 2026-02-25
feat: add script to compute correct tx hashes from block data

rootulp:
## Context

https://celestia-team.slack.com/archives/C07TJQPUDJP/p1771999423392399

An external team reported that after enabling `tx_index.indexer="kv"`, querying transactions by hash returned "tx not found" despite the block being indexed. The root cause: they were computing `SHA256(full BlobTx bytes)` but the indexer stores `SHA256(inner SDK tx bytes)`. See celestia-core [`Tx.Hash()`](https://github.com/celestiaorg/celestia-core/blob/v0.39.25/types/tx.go#L97-L105) for the unwrapping logic.

## Summary

- Adds `scripts/txhash/main.go`, a utility that fetches a block by height and computes the correct tx hashes, handling BlobTx unwrapping
- Addresses a common confusion where naively hashing raw block tx bytes produces incorrect hashes for BlobTxs because celestia-core's `Tx.Hash()` unwraps BlobTxs and hashes only the inner SDK transaction bytes
- Supports optional `--verify` flag to confirm each computed hash is queryable via the `/tx?hash=` RPC endpoint

### Usage

```bash
go run ./scripts/txhash/ --height 10230356
go run ./scripts/txhash/ --height 10230356 --rpc http://localhost:26657
go run ./scripts/txhash/ --height 10230356 --verify=false
```

## Test plan

- [x] Verified script compiles
- [x] Verified correct hashes for block 10230356 on mocha (`55BA1E82...` and `3EF1899F...`)
- [x] Verified both hashes are queryable via `/tx?hash=` on a public mocha RPC
- [x] Verified against multiple blocks (10230340, 10230350, 10230356) with varying tx counts

```
go run ./scripts/txhash --height 10230356 --rpc https://rpc-mocha.pops.one
block 10230356 has 2 transaction(s)

tx[0] type=BlobTx hash=55BA1E82A42E137F3A4231138342591FB3EED8D00393BFE631AB9305118CDF17
  verification OK: tx found in index
tx[1] type=BlobTx hash=3EF1899F86AA215941822CC8BA80F1E9B0CCFEB4024965FE120D9DA703110D8D
  verification OK: tx found in index
```

### DIFF
--- a/scripts/txhash/main.go
+++ b/scripts/txhash/main.go
@@ -55,8 +55,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := run(*rpc, *height, *verify, *insecure); err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func run(rpc string, height int64, verify, insecure bool) error {
 	client := http.DefaultClient
-	if *insecure {
+	if insecure {
 		client = &http.Client{
 			Transport: &http.Transport{
 				TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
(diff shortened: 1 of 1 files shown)


scripts/txhash/main.go
```diff
@@ -0,0 +1,157 @@
...
+				continue
+			}
+
+			if result.Error != nil {
+				fmt.Printf("  verification FAILED: %s\n", result.Error.Data)
+			} else {
+				fmt.Printf("  verification OK: tx found in index\n")
+			}
+		}
+	}
```
rach-id, line 134: can this be extracted to a separate function that takes the flags as arguments? would be easier to reason about and would remove the need for os.Exit on many levels

rootulp: https://github.com/celestiaorg/celestia-app/pull/6638/changes/e1f2e71ea8d28c90a817d155663660d02188a043

PR #6654, opened 2026-02-27
fix(x/forwarding): add valuable error information to error response

damiannolan:
## Overview

I ran into a failure case due to a misconfigured hyperlane IGP on mocha.
The issue would have been trivial to identify had we valuable error information returned, however all we got was:
```
raw_log: 'failed to execute message; message index: 0: all tokens failed to forward:
  all 1 tokens failed to forward'
```

Note that the events are discard from the cached context in cosmos-sdk in failed txs so emitted events in a conditional where we will abort is useless. 

This PR iterates the results and compiles a meaningful error msg.


damiannolan:
Personally I think this could be in v7. The point I make in the PR description is that the event removed is always actually dropped regardless.

rootulp:
> Personally I think this could be in v7. The point I make in the PR description is that the event removed is always actually dropped regardless.

SGTM. Seems safe to include in v7.x. Thanks for including the backport: v7.x label 👍 

Also events aren't consensus breaking so even if this PR did modify an event, it would be possible to backport. It may not be ideal because it's API breaking to modify events but we don't have to worry about that with this change.

PR #6655, opened 2026-02-27
chore: prune CLAUDE.md from 379 to 64 lines

rootulp:
Motivation: https://www.youtube.com/watch?v=PQU9o_5rHC4

## Summary

- Remove protocol deep-dives (BlobTx internals, data square details, ABCI lifecycle, ante handler chain, constants) that are discoverable from code
- Keep actionable development workflow guidance: build/test/lint commands, directory structure overview, dependency forks, and conventional commit requirements
- Reduce from 379 lines to 64 lines

## Test plan

- [ ] Verify CLAUDE.md renders correctly on GitHub
- [ ] Confirm no essential workflow guidance was lost


rootulp:
I've noticed that Opus 4.6 starts to behave less than ideal when I have extremely long chat sessions. It appears to auto-compact at certain points when the context window is large so the Claude team seems aware that the models degrade when the context window gets bloated.

AFAIU `CLAUDE.md` gets loaded into the beginning of every session so if we trim it down, it saves on context window usage for every subsequent Claude Code invocation.

To answer your question directly, I haven't noticed an immediate difference between the model performance before and after this PR merged but based on what I've seen / read, I think a smaller CLAUDE.md may improve model performance.

Issue #6597, opened 2026-02-17
The release notes should include that we removed the ability to set a nop mempool. THen, if anyone uses that, we can see what to do to accomodate it


rootulp:
The release notes should include that we removed the ability to set a nop mempool. THen, if anyone uses that, we can see what to do to accomodate it

_Originally posted by @rach-id in https://github.com/celestiaorg/celestia-core/pull/2811#pullrequestreview-3812636678_

PR #6660, opened 2026-03-01
docs: note nop mempool removal in v7 release notes

rootulp:
## Summary
- Add explicit mention of `nop` mempool removal to v7 release notes
- Direct affected users to open an issue so their use case can be accommodated

Closes https://github.com/celestiaorg/celestia-app/issues/6597

## Test plan
- [x] `markdownlint docs/release-notes/release-notes.md` passes


docs/release-notes/release-notes.md
rach-id, line 27: unrelated to this PR but worth noting that with fibre running in a standalone binary, local signing is no longer supported. We should decide wether we will only support tmkms from now on or expose the local signer somehow

rach-id (approved):

PR #6661, opened 2026-03-01
fix!: remove outdated GoalBlockTime and fix MaxExpectedTimePerBlock

### LINKED ISSUE #6644: Consider deprecating outdated `GoalBlockTime`
Since GoalBlockTime is 15 seconds, that seems extremely out of date. We should probably deprecate and remove.

MaxAgeDuration and MaxAgeNumBlocks need research and a CIP if we need to change. If we don't resolve them in this PR then can you please spin out Linear issues to investigate if we need to change them?

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-app/pull/6643#discussion_r2860933153_

rootulp:
Closes #6644
Closes DA-1089
Creates https://github.com/celestiaorg/celestia-app/issues/6689 

## Summary

- Remove the outdated `GoalBlockTime` constant (was 15s, actual block time is ~3s)
- Fix IBC `MaxExpectedTimePerBlock` genesis default from 75s to 15s (5x the ~3s block time)
- Add v8 upgrade handler to migrate the on-chain `MaxExpectedTimePerBlock` value

## Test plan

- [x] `TestSetMaxExpectedTimePerBlock` verifies the upgrade handler sets the correct value
- [x] Existing upgrade and default override tests continue to pass
- [x] `make build-standalone` succeeds

### DIFF
--- a/pkg/appconsts/app_consts.go
+++ b/pkg/appconsts/app_consts.go
@@ -38,7 +38,7 @@ const (
 	// 3-5x the expected block time. The expected block time is primarily
 	// determined by DelayedPrecommitTimeout (~3 seconds), so this value is
 	// 15 seconds (5x). If the timeout constants change, this value must be
-	// updated. 
+	// updated.
 	MaxExpectedTimePerBlock = 15 * time.Second
 
 	// TestUpgradeHeightDelay is the number of blocks that chain-id "test" waits


pkg/appconsts/consensus_consts.go
```diff
@@ -3,12 +3,6 @@ package appconsts
 import "time"
 
 const (
-	// GoalBlockTime is the target time interval between blocks. Since the block
-	// interval isn't enforced at consensus, the real block interval isn't
-	// guaranteed to exactly match GoalBlockTime. GoalBlockTime is currently targeted
-	// through static timeouts (i.e. TimeoutPropose, TimeoutCommit).
-	GoalBlockTime = time.Second * 15
```
rach-id, line 10: this is breaking, no?

app/upgrades.go
```diff
@@ -90,3 +97,17 @@ func (app App) RegisterUpgradeHandlers() {
 		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
 }
+
+// SetMaxExpectedTimePerBlock sets the IBC connection MaxExpectedTimePerBlock
+// parameter to 15 seconds (5x the ~3 second expected block time). This
+// corrects the previous value of 75 seconds which was based on an outdated 15
+// second block time.
```
rach-id: when did we have 15s blocks?

app/default_overrides.go
```diff
@@ -127,13 +127,13 @@ type ibcModule struct {
...
 // DefaultGenesis returns custom x/ibc module genesis state.
 func (ibcModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
-	// per ibc documentation, this value should be 3-5 times the expected block
-	// time. The expected block time is 15 seconds, therefore this value is 75
-	// seconds.
-	maxBlockTime := appconsts.GoalBlockTime * 5
+	// Per IBC documentation, MaxExpectedTimePerBlock should be 3-5 times the
+	// expected block time. The expected block time is ~3 seconds so this value
+	// is 15 seconds.
+	maxExpectedTimePerBlock := 15 * time.Second
```
rach-id: why not `appconsts.DelayedPrecommitTimeout`? or should this be the worst case, which is `max(timeoutPropose + timeoutPrevote, DelayedPrecommit timeout)`?

rach-id (dismissed):
left couple questions to understand more

app/upgrades.go
cmwaters: Pre genesis. I think we were going for 15s but the network actually produced them every 12 seconds

app/default_overrides.go
cmwaters: Yeah I would prefer for it to be in relation to the variable that most dictates the block time so we don't have to worry about changing this if we change the block time

pkg/appconsts/consensus_consts.go
rootulp: true this is breaking b/c it's an exported public const. Added `!` to PR title.

Note: I don't see any usage in celestia-node.

app/default_overrides.go
rootulp: I think that variable has changed. Before it was TimeoutCommit. Now it is DelayedPrecommitTimeout. I could be convinced to go with something like: `max(timeoutPropose + timeoutPrevote, DelayedPrecommit timeout)`

Alternatively we could:
1. Add a unit test that fails if the sum of those values is significantly different from maxExpectedTimePerBlock
2. In the unit test or above these constants, explain in comments that all the times need to be updated at the same time

rootulp:
> and I assume we only need to do this locally (to our network) right?

Correct we only need to fix this on our network.

> What's the actual behaviour change in the IBC when we modify the MaxExpectedTimePerBlock? 

`MaxExpectedTimePerBlock` is only used if an IBC connection has a non-zero `delayPeriodTime`. Currently all 103 IBC connections on Celestia Mainnet use a `delayPeriodTime` of zero. So currently this parameter is not used.

Screenshot 2026-03-02 at 6 32 32 AM

I have a lot more info from Claude but seems unnecessary to include here. TLDR is we should fix this parameter for correctness even though it won't impact existing IBC connections. It may impact future IBC connections (if the delayPeriodTime is configured to be non-zero).

Ref:
- https://github.com/cosmos/ibc-go/blob/main/modules/core/03-connection/types/params.go
- https://pkg.go.dev/github.com/cosmos/ibc-go/v6/modules/core/03-connection/types#Params

app/default_overrides.go
rootulp: Update: I want something that will fail Go compiler or test because we failed to update `GoalBlockTime` in the past. That's why I'm advocating for the unit test.

I think the approach with something like `max(timeoutPropose + timeoutPrevote, DelayedPrecommit timeout)` could also work.

Trying the unit test approach first, then will fallback to that.

pkg/appconsts/app_consts.go
```diff
@@ -19,18 +19,27 @@ const (
...
+	TimeoutCommit                = time.Millisecond
+	// DelayedPrecommitTimeout is the primary determinant of expected block
+	// time. If this value changes, MaxExpectedTimePerBlock should also be
+	// updated. See TestMaxExpectedTimePerBlock.
+	DelayedPrecommitTimeout = time.Millisecond * 2790
+	// MaxExpectedTimePerBlock is the IBC connection parameter that should be
+	// 3-5x the expected block time. The expected block time is primarily
+	// determined by DelayedPrecommitTimeout (~3 seconds), so this value is
+	// 15 seconds (5x). If the timeout constants change, this value must be
+	// updated. See TestMaxExpectedTimePerBlock.
```
rootulp:
```suggestion
	// updated. 
```

app/upgrades.go
```diff
@@ -74,6 +74,12 @@ func (app App) RegisterUpgradeHandlers() {
 		func(ctx context.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
 			sdkCtx := sdk.UnwrapSDKContext(ctx)
 			sdkCtx.Logger().Info("running upgrade handler", "upgrade-name", upgradeName)
+
+			if err := app.SetMaxExpectedTimePerBlock(sdkCtx); err != nil {
+				sdkCtx.Logger().Error("failed to set MaxExpectedTimePerBlock", "error", err)
+				return nil, err
+			}
```
cmwaters, line 81: Will this happen every major upgrade?

cmwaters (approved):

app/upgrades.go
rootulp: Currently it will only run at the v7 -> v8 activation height. We'll remove it after v8 activates similar to how we've removed previous upgrade handlers.

PR #6690, opened 2026-03-02
chore: set max bytes to 32 MiB and decrease `TimeoutPropose` `TimeoutPrevote`

### LINKED ISSUE #6649: Cut the `Prevote` and `Precommit` timeouts in half
I think we should cut the Prevote and Precommit timeouts in half as well, so that the next round is reached faster

_Originally posted by @rach-id in https://github.com/celestiaorg/celestia-app/pull/6643#pullrequestreview-3866198920_

ninabarbakadze:
## Overview

Fixes #6649 

Closes: DA-1095

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-app/pull/6690" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


pkg/appconsts/app_consts.go
```diff
@@ -25,9 +25,9 @@ const (
 	MaxTxSize               int    = 8_388_608 // 8 MiB in bytes
 	TimeoutPropose                 = time.Millisecond * 4500
 	TimeoutProposeDelta            = time.Millisecond * 500
-	TimeoutPrevote                 = time.Millisecond * 3000
+	TimeoutPrevote                 = time.Millisecond * 1500
 	TimeoutPrevoteDelta            = time.Millisecond * 500
-	TimeoutPrecommit               = time.Millisecond * 3000
+	TimeoutPrecommit               = time.Millisecond * 1500
```
evan-forbes: is the reasoning here that we're cutting the block time in half therefore should also cut these timeouts in half? I read the original issue, which stated that this is just to reduce the time to the next round provided the proposer is down.

what experiments have been ran? what are the existing numbers and standard deviation for prevote and precommit times?

pkg/appconsts/initial_consts.go
```diff
@@ -23,7 +23,7 @@ const (
 	// Only applies to app version >= 2
 	DefaultNetworkMinGasPrice = 0.000001 // utia
 
-	DefaultUpperBoundMaxBytes = 128 * mebibyte
+	DefaultUpperBoundMaxBytes = 32 * mebibyte
```
rootulp, line 26: [blocking] we need an upgrade handler that modifies the consensus params Block MaxBytes to 32 MiB. Changing the value here is only one step but it won't impact existing chains.

Ref: https://github.com/celestiaorg/celestia-app/blob/e73e63c54c7021cf7479318af296cb2475b8fab6/test/util/test_app.go#L142

Ref: https://github.com/celestiaorg/celestia-app/blob/e73e63c54c7021cf7479318af296cb2475b8fab6/pkg/appconsts/initial_consts.go#L3-L4

rootulp (changes requested):

pkg/appconsts/app_consts.go
ninabarbakadze: > is the reasoning here that we're cutting the block time in half therefore should also cut these timeouts in half?

yes! naively it seemed adequate to cut the timeouts in half. 

> what experiments have been ran? what are the existing numbers and standard deviation for prevote and precommit times?

I only ran an experiment afterwards submitting 8MB/s to make sure everything was working as expected and didn't notice anything suspicious. No existing numbers on standard deviation for prevote and precommit times. I can get those!

pkg/appconsts/initial_consts.go
ninabarbakadze: apologies, I see! so, we've never set this value as MaxBytes for consensus?

ninabarbakadze: How did we set MaxBytes to 128 on arabica/mocha?

pkg/appconsts/app_consts.go
rach-id: > is the reasoning here that we're cutting the block time in half therefore should also cut these timeouts in half? 

I think these should be cut in half:

1. With 32mb blocks, the network load is much lower than 128mb/6s, so the votes should arrive way faster
2. 3s is the block time. If we keep them, the worst case would be to wait for an extra block time just to get prevotes which is a lot. If we don't get prevotes in half the block time, it's time to move to a new round.

rach-id (commented):
I think a requirement for merging this PR would be to run a couple validators network then to do the upgrade and make sure the block time is correct, the block size, and all constants are correctly set. Or it could be done in a separate issue

ninabarbakadze:
> all constants are correctly set

I assume this also includes all consensus timeouts. What exactly do we want to check for to make sure consensus timeouts are properly configured?

pkg/appconsts/initial_consts.go
ninabarbakadze: https://github.com/celestiaorg/celestia-app/pull/6690/changes/cf89456acac021e361f18eb9be097bb3788b0e15

rach-id:
> What exactly do we want to check for to make sure consensus timeouts are properly configured?

The simplest way would be to just print them in core side and see the logs whether they changed or not

ninabarbakadze:
> > What exactly do we want to check for to make sure consensus timeouts are properly configured?
> 
> The simplest way would be to just print them in core side and see the logs whether they changed or not

Yes, but IIUC Evan also wanted to see experiments for consensus timeout reductions.

pkg/appconsts/app_consts.go
evan-forbes: that reasoning makes sense, and agree that in all likely hood cutting is fine, but imo is not sufficient by itself

we need measure before we cut imo. For example, it could be that we could set these lower or that if we set slightly higher then we get way fewer missed blocks.

evan-forbes: we also have to keep in mind that talis also appears to a small but measurable amount faster than mocha or mainnet

cmwaters:
> I think a requirement for merging this PR would be to run a couple validators network then to do the upgrade and make sure the block time is correct, the block size, and all constants are correctly set. Or it could be done in a separate issue

Can we do this as a FLUP. 

Evan's proposal, as I understand it, is different. His is performance testing and Rachid's is correctness testing and ideally would be done in Tastura as part of the "upgrade to v8 test"

pkg/appconsts/app_consts.go
```diff
@@ -23,9 +23,9 @@ const (
 	TxSizeCostPerByte     uint64 = 10
 	GasPerBlobByte        uint32 = 8
 	MaxTxSize             int    = 8_388_608 // 8 MiB in bytes
-	TimeoutPropose               = time.Millisecond * 4500
+	TimeoutPropose               = time.Millisecond * 3000
 	TimeoutProposeDelta          = time.Millisecond * 500
-	TimeoutPrevote               = time.Millisecond * 3000
+	TimeoutPrevote               = time.Millisecond * 2000
```
ninabarbakadze, line 36: We got a consensus on these values based on multiple experiments. They basically represent the maximum seen per constant plus some extra buffer to account for mocha and mainnet being slower than talis.

see latest analysis for more details https://celestia-team.slack.com/archives/C05CYC8EHBM/p1774951920646719?thread_ts=1772742571.293389&cid=C05CYC8EHBM

rootulp:
I updated the PR title b/c "halve" was not applicable.

pkg/appconsts/initial_consts.go
rootulp: > How did we set MaxBytes to 128 on arabica/mocha?

Governance proposals. See Mocha proposal: https://www.mintscan.io/celestia-testnet/proposals/16

I have a slight preference for introducing a new constant for MaxBytes rather than using the `DefaultUpperBoundMaxBytes` which historically has been used as the _hard-coded_ max bytes. This PR is trying to modify the gov param in the upgrade handler so IMO it should use a new variable.

Also we're violating the comment at the top of this block by changing this variable in an app version

```
// The following defaults correspond to initial parameters of the network that can be changed, not via app versions
// but other means such as on-chain governance, or the node's local config
```

ninabarbakadze:
> One comment about using a different constant for the 32 MiB max bytes

Thanks! Makes sense! Quick question before I push your feedback - do we want to keep `DefaultUpperBoundMaxBytes` for capping gRPC message sizes? I introduced another constant in `app_consts` called `BlockMaxBytes` for setting the gov modifiable param and I'm not super sure on where I should keep using `DefaultUpperBoundMaxBytes` vs `BlockMaxBytes` without creating too much confusion.

rootulp:
> do we want to keep DefaultUpperBoundMaxBytes for capping gRPC message sizes?

Yes. I don't think we strictly need to modify that.

>  I'm not super sure on where I should keep using DefaultUpperBoundMaxBytes vs BlockMaxBytes without creating too much confusion.

Screenshot 2026-04-02 at 11 28 14 AM

Use the new variable in:
- test_app.go

I would leave the other instances of `DefaultUpperBoundMaxBytes` intact.

ninabarbakadze:
> > do we want to keep DefaultUpperBoundMaxBytes for capping gRPC message sizes?
> 
> Yes. I don't think we strictly need to modify that.
> 
> > I'm not super sure on where I should keep using DefaultUpperBoundMaxBytes vs BlockMaxBytes without creating too much confusion.
> 
> Screenshot 2026-04-02 at 11 28 14 AM
> Use the new variable in:
> 
> * test_app.go
> 
> I would leave the other instances of `DefaultUpperBoundMaxBytes` intact.

Adresses! https://github.com/celestiaorg/celestia-app/pull/6690/changes/07e924da8e8ab59551f36497f1f7f5b93066d797

I left `DefaultUpperBoundMaxBytes` at 32MB.

pkg/appconsts/initial_consts.go
rootulp, line 26: [question] @claude do we need to change this?

rach-id (approved):

PR #6694, opened 2026-03-03
fix: nil pointer dereference in RemoteABCIClientV1.FinalizeBlock

### LINKED ISSUE #6540: Potential nil pointer dereference in RemoteABCIClientV1.FinalizeBlock
In multiplexer/abci/remote_v1.go, FinalizeBlock accesses GetConsensusParamUpdates().Version.AppVersion without a nil check:

a.endBlockConsensusAppVersion = endBlockResp.GetConsensusParamUpdates().Version.AppVersion

  GetConsensusParamUpdates() returns nil when the v1 application does not include consensus parameter updates in its EndBlock response. Accessing .Version on nil causes a panic.

  Consensus parameters are typically only updated during network upgrades, so blocks without updates are the common case.

rootulp:
## Summary

- Fix nil pointer dereference in `RemoteABCIClientV1.FinalizeBlock` when `EndBlock` response has nil `ConsensusParamUpdates` or nil `Version` (the common case for non-upgrade blocks)
- Add unit tests for `FinalizeBlock` covering nil `ConsensusParamUpdates`, nil `Version`, and the happy path with a version update

Closes #6540

## Test plan

- [x] `go test -v -run TestFinalizeBlock ./multiplexer/abci/` passes all three subtests
- [x] `go test -v ./multiplexer/abci/` passes all existing tests

### DIFF
--- a/app/default_overrides.go
+++ b/app/default_overrides.go
@@ -127,13 +127,9 @@ type ibcModule struct {
 
 // DefaultGenesis returns custom x/ibc module genesis state.
 func (ibcModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
-	// per ibc documentation, this value should be 3-5 times the expected block
-	// time. The expected block time is 15 seconds, therefore this value is 75
-	// seconds.
-	maxBlockTime := appconsts.GoalBlockTime * 5
 	gs := ibctypes.DefaultGenesisState()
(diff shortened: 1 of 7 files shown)


rootulp:
I haven't seen an example of it getting hit in practice but hypothetically it could get hit when remote_v1 is used (app versions 1 to 3) and a block doesn't return consensus param updates (which is most blocks).

PR #6695, opened 2026-03-03
docs: document `verify_data` config in release notes

### LINKED ISSUE #6691: Document `verify_data` in release notes
## Context

I just learned of a new config. In config.toml

```
# If true, the node will verify the application data in the block via ProcessProposal
# during block sync. If false, this verification is skipped.
# Default: true
verify_data = true
```

## Proposal

1. Research when we added it
2. Add it to release notes

rootulp:
## Summary
- Documents the `verify_data` config option in the `[blocksync]` section of `config.toml` in the v6 128mb/6s release notes.
- This option was introduced in celestia-core v0.39.16 and first included in celestia-app v6.3.0.
- Configures markdownlint MD024 rule with `siblings_only` to allow duplicate heading content in different parent sections (e.g. "Config Changes" in v6 and v7 release notes).

## Test plan
- [ ] Verify the markdown renders correctly on the PR diff.
- [ ] Verify markdownlint passes in CI.

Closes #6691

### DIFF
--- a/.github/workflows/test-nightly.yml
+++ b/.github/workflows/test-nightly.yml
@@ -7,15 +7,14 @@ concurrency:
 on:
   schedule:
     # run nightly at 2 AM UTC
-    - cron: '0 2 * * *'
+    - cron: "0 2 * * *"
   # Allow manual triggering for testing
   workflow_dispatch:
     inputs:
       tag:
-        description: 'Docker image tag to test (leave empty to use commit SHA)'
+        description: "Docker image tag to test (leave empty to use commit SHA)"
         required: false
         type: string
-
 
 jobs:
   test-state-sync-mocha:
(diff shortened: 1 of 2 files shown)


docs/release-notes/release-notes.md
```diff
@@ -119,6 +119,23 @@ After upgrading to 128mb/6s, specific validator hardware specifications are requ
...
+```toml
+[blocksync]
+
+# If true, the node will verify the application data in the block via ProcessProposal
+# during block sync. If false, this verification is skipped.
+# Default: true
+verify_data = true
+```
+
+Setting `verify_data = false` can significantly speed up block sync, especially for large blocks, but reduces security during sync because a malicious peer could send fabricated block data without being caught. Operators should only disable this when syncing from trusted peers.
```
rach-id:
```suggestion
without being caught
```

I think they would get caught if they did. After process proposal is skipped, the block is applied. If the applied block is malicious, then it will generate an invalid state and no more blocks can be applied after it:

```
	// Validate app info
	if !bytes.Equal(block.AppHash, state.AppHash) {
		return fmt.Errorf("wrong Block.Header.AppHash.  Expected %X, got %v",
			state.AppHash,
			block.AppHash,
		)
	}
```

The only difference is: if `verify_data` is `true`, you can verify the block then discard it. If it's `false`, it will be applied and will corrupt your state

docs/release-notes/release-notes.md
```diff
@@ -119,6 +119,23 @@ After upgrading to 128mb/6s, specific validator hardware specifications are requ
...
+```toml
+[blocksync]
+
+# If true, the node will verify the application data in the block via ProcessProposal
+# during block sync. If false, this verification is skipped.
+# Default: true
+verify_data = true
+```
+
+without being caught
```
rootulp:
```suggestion
```

docs/release-notes/release-notes.md
```diff
@@ -119,6 +119,22 @@ After upgrading to 128mb/6s, specific validator hardware specifications are requ
...
+
+```toml
+[blocksync]
+
+# If true, the node will verify the application data in the block via ProcessProposal
+# during block sync. If false, this verification is skipped.
+# Default: true
+verify_data = true
+```
+
```
rootulp, line 136:
```suggestion
```

PR #6720, opened 2026-03-04
test: reproduce stale message authorization in x/zkism

rootulp:
## Summary
- Adds a regression test reproducing CELESTIA-199: messages authorized under state root R1 remain consumable via `Verify()` after the ISM transitions to state root R2.
- The `messages` collection is keyed by `(ismId, messageId)` with no state root component, and `UpdateISM` never clears it on state root change.

## Test plan
- [x] `go test ./x/zkism/keeper/ -run "TestKeeperTestSuite/TestStaleMessagesSurviveStateRootTransition" -v -count=1` passes, confirming the vulnerability is reproducible.

### DIFF
--- a/x/zkism/keeper/stale_messages_test.go
+++ b/x/zkism/keeper/stale_messages_test.go
@@ -1,0 +1,69 @@
+package keeper_test
+
+import (
+	"github.com/bcp-innovations/hyperlane-cosmos/util"
+)
+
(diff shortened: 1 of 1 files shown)


rootulp:
cc: @Manav-Aggarwal @damiannolan

damiannolan (commented):
This is working as designed. If a message id makes its way into the keyset, it is treated as greenlit for processing.

PR #6723, opened 2026-03-05
feat: make the pebbleDB migration script idempotent + resumable + parallel

rach-id:
## Overview

Adds to DA-508.

Tested on network storage with a max read/write speed of 250mb/s:
- 12GB data folder: 2m30s with verification
- 120GB data folder: ~19min without verification

We should expect better results in validators setups as they use bare metal servers and faster SSDs

<details>
  <summary>Implementation plan:</summary>

# migrate-db: LevelDB to PebbleDB Migration Tool

## Goal

Migrate all celestia-app databases (application, blockstore, state, tx_index, evidence) from GoLevelDB to PebbleDB. The tool must be **idempotent**, **resumable** after crashes, and support **parallel** migration of multiple databases.

## Architecture

Single Go binary (`tools/migrate-db/main.go`) with no external dependencies beyond `cosmos-db`, `flock`, and `errgroup`.

### Directory Layout

```
$HOME/.celestia-app/
├── data/                      # Source LevelDB files (e.g. application.db/, blockstore.db/)
├── data_pebble/               # Temporary dest during migration
│   ├── application.db/
│   ├── .migration_state.json  # Persistent progress tracker
│   └── .migration.lock        # Kernel-level flock
└── config/
    └── config.toml            # db_backend rewritten on auto-swap
```

### CLI Flags

| Flag             | Default                    | Description                                        |
|------------------|----------------------------|----------------------------------------------------|
| `-home`          | `$HOME/.celestia-app`      | Node home directory                                |
| `-dry-run`       | `false`                    | Print what would happen, no changes                |
| `-backup`        | `false`                    | Keep source LevelDB after migration                |
| `-batch-size`    | `64` (MB)                  | Dest write batch size                              |
| `-sync-interval` | `1024` (MB)               | Fsync to dest every N MB (0 = only at DB end)      |
| `-parallel`      | `3`                        | Number of databases migrated concurrently           |
| `-verify`        | `false`                    | Sample-verify 1000 keys after each DB migration    |
| `-db`            | `""` (all)                 | Migrate a single database (e.g. `--db blockstore`) |
| `-manual-swap`   | `false`                    | Skip auto-swap; print manual instructions instead  |

## Data Flow

```
main()
  → parse flags, clamp parallel ≥ 1
  → signal.NotifyContext(SIGINT, SIGTERM)
  → runMigration(ctx, opts)
      → validate dataDir exists
      → filter databases if --db is set
      → printBanner
      → if --dry-run: print plan and return
      → MkdirAll(data_pebble)
      → flock(data_pebble/.migration.lock)
      → loadOrInitState
      → errgroup.SetLimit(opts.parallel)
          → migrateOneDB(ctx, dbName) × N
      → auto-swap or print manual steps
```

## Persistent State Machine

Each database moves through a linear state machine, persisted to `.migration_state.json` (atomic write via tmp+rename):

```
pending → in_progress → migrated → source_deleted
```

| State            | Meaning                                      |
|------------------|----------------------------------------------|
| `pending`        | Not started                                  |
| `in_progress`    | Copy loop running (or interrupted)           |
| `migrated`       | All keys copied to dest, verified if enabled |
| `source_deleted` | Source LevelDB removed (no-backup mode)      |

On resume, databases already in `migrated` or `source_deleted` are skipped. A database in `in_progress` whose source directory is missing (crash during no-backup delete) is promoted to `migrated`.

The `stateTracker` wraps `MigrationState` with a `sync.Mutex` for safe concurrent access from parallel goroutines.

## Concurrency

All databases are dispatched through a single `errgroup` with `SetLimit(opts.parallel)`. This naturally serializes when parallel=1 and parallelizes otherwise — no separate code path needed.

If any database fails, the errgroup's context is cancelled, stopping in-flight siblings at their next batch-flush checkpoint.

## Per-Database Migration (`migrateOneDB`)

1. **State check** — skip if already `migrated`/`source_deleted`.
2. **Source existence** — skip if LevelDB dir missing; promote `in_progress` → `migrated` if source was already deleted.
3. **Transition to `in_progress`** and persist.
4. **Call `migrateSingleDB`** — the copy loop.
5. **Transition to `migrated`** with key/byte counts and timestamp.
6. **Optional verification** (`--verify`) — sample 1000 evenly-spaced keys.
7. **Source cleanup** (unless `--backup`) — `os.RemoveAll`, then transition to `source_deleted`.

### Resume Point

On open, `findResumePoint` creates a **reverse iterator** on the dest PebbleDB and reads the last key. `verifyResumePoint` then confirms the value matches the source (skipped if the key was already deleted in no-backup mode). `iteratorFrom` positions the source iterator **after** the resume key.

### Two Copy Strategies

**`copyAllKeys`** (when `--backup` is set):
- Iterates source, batches `Set` calls to dest.
- Flushes when batch reaches `batchBytes` (default 64 MB).
- Fsync when cumulative bytes since last sync reach `syncBytes`.
- Checks `ctx.Err()` after each flush for graceful cancellation.
- Final batch is always fsynced (`WriteSync`).

**`copyAndDeleteKeys`** (default, no backup):
- Same as above, plus tracks copied keys in a `[][]byte` buffer.
- Every `deleteChunkBytes` (1 GB):
  1. Flush pending dest batch with sync.
  2. Close source iterator (required before mutating source).
  3. Call `deleteSourceKeys` to remove copied keys from source.
  4. Reopen source iterator from last position.
- Remaining keys are deleted after the loop completes.

### Constants

| Constant           | Value  | Purpose                                            |
|--------------------|--------|----------------------------------------------------|
| `deleteChunkBytes` | 1 GB   | How much data to copy before deleting from source  |
| `maxDeleteBatch`   | 64 MB  | Cap on in-memory delete batch size for source DB   |
| `progressInterval` | 2 min  | Frequency of throughput logging                    |

### Batch Helpers

- **`flushBatch`** — writes (sync or async), closes the batch, returns a fresh one.
- **`deleteSourceKeys`** — deletes keys in batches capped at `maxDeleteBatch` to bound memory.
- **`logProgress`** — prints keys/bytes/throughput every `progressInterval`.

### Auto-Swap (default)

1. For each database: remove old LevelDB dir (if `--backup`, it still exists), then `os.Rename` PebbleDB dir into `data/`.
2. Rewrite `db_backend = "pebbledb"` in `config/config.toml`.
3. Clean up `.migration_state.json`, `.migration.lock`, and the empty `data_pebble/` dir.

### Manual Swap (`--manual-swap`)

Prints step-by-step shell commands for the operator to run.

## Verification (`--verify`)

`verifyDBSample` does two passes over the source:
1. Count total keys.
2. Sample every `totalKeys/1000`-th key, `Get` it from dest, compare values.

Reports mismatches (up to 3 printed) and fails if any are found.

## Safety Properties

| Property         | Mechanism                                                              |
|------------------|------------------------------------------------------------------------|
| Single-instance  | `flock` on `.migration.lock` (auto-released on crash)                  |
| Crash-safe state | Atomic tmp+rename for `.migration_state.json`                          |
| Resume integrity | Last dest key verified against source before continuing                |
| Graceful stop    | SIGINT/SIGTERM → context cancellation → checked at every batch flush   |
| Disk space       | No-backup mode deletes source every ~1 GB; backup mode needs 2× space |

## Dependencies

- `github.com/cosmos/cosmos-db` — unified DB interface (GoLevelDB + PebbleDB backends)
- `github.com/gofrs/flock` — cross-platform file locking
- `golang.org/x/sync/errgroup` — bounded-concurrency goroutine group

</details>

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-app/pull/6723" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tools/migrate-db/main.go
+++ b/tools/migrate-db/main.go
@@ -302,7 +302,26 @@ func loadOrInitState(pebbleDataDir string, opts migrateOpts) (*stateTracker, err
 func migrateOneDB(ctx context.Context, dbName, dataDir string, tracker *stateTracker, opts migrateOpts) error {
 	ds := tracker.getDBState(dbName)
 
-	if ds.Status == statusMigrated || ds.Status == statusSourceDeleted {
+	if ds.Status == statusSourceDeleted {
+		fmt.Printf("[%s] Already complete (status=%s), skipping\n", dbName, ds.Status)
+		return nil
+	}
+
+	// If migrated but source not yet deleted (crash between save and RemoveAll), retry deletion.
+	if ds.Status == statusMigrated {
+		if !opts.backup {
+			srcPath := filepath.Join(dataDir, dbName+".db")
+			if _, err := os.Stat(srcPath); err == nil {
+				fmt.Printf("[%s] Retrying source deletion (interrupted after migration)...\n", dbName)
+				if err := os.RemoveAll(srcPath); err != nil {
+					return fmt.Errorf("[%s] failed to remove source: %w", dbName, err)
+				}
+				ds.Status = statusSourceDeleted
+				if err := tracker.updateDBState(dbName, ds); err != nil {
+					return fmt.Errorf("[%s] failed to save state: %w", dbName, err)
+				}
+			}
+		}
 		fmt.Printf("[%s] Already complete (status=%s), skipping\n", dbName, ds.Status)
 		return nil
 	}
@@ -877,13 +896,12 @@ func performAutoSwap(homeDir, dataDir, pebbleDataDir string, backup bool, tracke
 		srcPath := filepath.Join(pebbleDataDir, dbName+".db")
 		dstPath := filepath.Join(dataDir, dbName+".db")
 
-		// Remove old LevelDB if it still exists
-		if backup {
-			if _, err := os.Stat(dstPath); err == nil {
-				fmt.Printf("  Removing old %s\n", dstPath)
-				if err := os.RemoveAll(dstPath); err != nil {
-					return fmt.Errorf("failed to remove old %s: %w", dstPath, err)
-				}
+		// Remove old LevelDB if it still exists (always needed — in no-backup mode
+		// the source may survive a crash between statusMigrated and RemoveAll).
+		if _, err := os.Stat(dstPath); err == nil {
+			fmt.Printf("  Removing old %s\n", dstPath)
+			if err := os.RemoveAll(dstPath); err != nil {
+				return fmt.Errorf("failed to remove old %s: %w", dstPath, err)
 			}
 		}
 


rach-id:
> Add to v8 release notes that users should migrate from level to pebble?

We will be rolling this out gradually in v7.

> Modify the default db from level to pebble?

I am working on it separately

I am still finalising the code. The tool works but since it will be released to node operators to run, it needs to be fool proof and I am currently doing the necessary tests for that before merging

rootulp:
> We will be rolling this out gradually in v7.

Sounds good. Will v8 _require_ pebble or is level still supported? I assume v8 will require Pebble in which case we want a breaking change section of the v8 release notes to instruct node operators to run the migration script.

rach-id:
> Sounds good. Will v8 require pebble or is level still supported? I assume v8 will require Pebble in which case we want a breaking change section of the v8 release notes to instruct node operators to run the migration script.

The default will be pebbleDB. But for existing nodes that didn't yet migrate, I'm not sure wether we need to drop support for level entirely, especially if this is of archive nodes/full nodes. The reason being pebble is important for validators for saving the block given the performance it offers so that they move faster to the next height. However, for full nodes/archival, they don't necessarily need it. Also, It can technically help with blocksyncing large blocks, but won't be that much of a difference.

So, I don't think we should stop supporting it for now. Maybe in v9

rootulp:
Thanks for thinking about this deeply. 

> However, for full nodes/archival, they don't necessarily need it.

Then we need validation in the binary that errors on start-up if a validator hasn't configured Pebble.

There is a cost to maintaining multiple DBs:
1. We need to have tests that confirm all types of nodes work with all supported DBs
2. There is cognitive cost for node operators to decide which DB to use

In general, I lean towards minimizing the number of config options we expose to node operators. Is there a compelling reason that a node operator would want to remain on Level? If not, I vote for:
1. Only supporting Pebble
2. Our tests only use Pebble
4. Offer a migration script in v7
5. In v8 require that all node operators have run the migration script and error on start-up if a node operator tries to configure anything but Pebble

cc: @rach-id

rach-id:
@rootulp sounds good 👍

Issue #3164, opened 2024-03-06
Test to verify that ordering of blobs in a namespace is based on priority


rootulp:
## Context

> The order of blobs in a namespace is dictated by the priority of the PFBs that paid for the blob. A PFB with greater priority will have all blobs in that namespace strictly before a PFB with less priority. Priority is determined by the gas-price of the transaction (fee/gas).

From https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#ordering

## Question

Is this actually enforced? It would be nice to have a test that verifies this behavior.

rootulp:
I think this is actually enforced because if a block proposer rearranges blobs in a namespace then that will result in a data root mismatch in process proposal. So the default layout ordering rules are the only layout ordering rules that will be accepted by the network.

rootulp:
If we wanted the default layout ordering rules to only be defaults and allow other valid layouts, I think block proposers would need a mechanism to pass information about the layout (i.e. blob starting share indexes) in their proposal.

evan-forbes:
> if a block proposer rearranges blobs in a namespace then that will result in a data root mismatch in process proposal.

yeah exactly this. the order of the protobuf encoded block determines the order in the square. the commitment to the protobuf encoded block is in the partset header. I think we used to have an exact test for this before refactoring things into blobtxs, but this one is similar but not the same since blobTx are kept inside the block https://github.com/celestiaorg/celestia-app/blob/80968ccea7daf8cfe9d062dc618f8e3e8f783bd9/app/test/process_proposal_test.go#L236-L244

we could add one tho

rootulp:
## Investigation Findings

I investigated how blob ordering within a namespace works end-to-end across celestia-core and celestia-app. Here's what I found:

### 1. The CAT mempool DOES return transactions in priority order

The `t.Skip("TODO: skipping test until prioritisation available in core")` in `app/test/priority_test.go` is outdated. celestia-core v0.39.25's CAT mempool fully implements priority-based ordering:

- `ReapMaxBytesMaxGas` explicitly returns transactions "ordered by decreasing priority, with ties broken by increasing order of arrival"
- The store maintains `orderedTxSets` sorted by aggregated priority via binary search insertion
- `state/execution.go:156` calls `ReapMaxBytesMaxGas` and passes the result to `PrepareProposal`

### 2. Priority ordering is per-sender-group, NOT strict per-transaction

[...]

PR #6726, opened 2026-03-05
test: add blob ordering tests and update specs for priority ordering

rootulp:
## Summary

- Add tests verifying blob ordering behavior within namespaces at the PrepareProposal/ProcessProposal level
- Update specs to accurately describe how priority ordering works (sender-aggregated, not strict per-tx)
- Remove outdated `t.Skip` from `TestPriorityTestSuite` since prioritization is available in celestia-core's CAT mempool

See investigation: https://github.com/celestiaorg/celestia-app/issues/3164#issuecomment-4007121377

Companion celestia-core PR: https://github.com/celestiaorg/celestia-core/pull/2839

### Tests (`app/test/blob_ordering_test.go`)
- **TestBlobOrderingWithinNamespaceByPriority**: Creates 4 PFB transactions with the same namespace but different gas prices, feeds them to `PrepareProposal` in priority order, and verifies the output preserves that ordering. Also verifies `ProcessProposal` accepts the result.
- **TestBlobReorderingWithinNamespaceRejected**: Creates a valid proposal then swaps blob transaction order. Verifies `ProcessProposal` rejects the tampered proposal (data root mismatch).

### Removed skip (`app/test/priority_test.go`)
- Removed `t.Skip("TODO: skipping test until prioritisation available in core")` from `TestPriorityTestSuite`. The CAT mempool in celestia-core v0.39.25 implements priority-based ordering via `ReapMaxBytesMaxGas` which returns transactions "ordered by decreasing priority".
- Used `WaitForTx` helper instead of direct `Client.Tx()` to avoid race condition where the tx indexer hasn't caught up yet.

### Spec updates
- **`specs/src/data_square_layout.md`**: Replaced inaccurate claim that blobs are "strictly" ordered by per-tx priority. The actual ordering is by sender-aggregated priority (gas-weighted average), which can cause a high-priority tx to appear after a lower-priority tx from a different sender if the senders have mixed priority levels.
- **`specs/src/block_proposer.md`**: Clarified that "priority order" means mempool-provided sender-aggregated priority order, not strict per-tx priority.

Closes https://github.com/celestiaorg/celestia-app/issues/3164

## Test plan
- [x] `go test -run "TestBlobOrdering|TestBlobReordering" -count=1 ./app/test/` passes
- [x] `go test -run "TestPriorityTestSuite" -count=1 -timeout 300s ./app/test/` passes (previously skipped)
- [x] `make build` succeeds

### DIFF
--- a/app/test/priority_test.go
+++ b/app/test/priority_test.go
@@ -1,7 +1,6 @@
 package app_test
 
 import (
-	"encoding/hex"
 	"math/rand"
 	"sort"
 	"sync"
@@ -104,10 +103,8 @@ func (s *PriorityTestSuite) TestPriorityByGasPrice() {
 	// note: use rpc types because they contain the tx index
 	heightMap := make(map[int64][]*rpctypes.ResultTx)
 	for hash := range hashes {
-		// use the core rpc type because it contains the tx index
-		hash, err := hex.DecodeString(hash)
-		require.NoError(t, err)
-		coreRes, err := s.cctx.Client.Tx(s.cctx.GoContext(), hash, false)
+		// use WaitForTx to poll until the tx is indexed
+		coreRes, err := s.cctx.WaitForTx(hash, 10)
 		require.NoError(t, err)
 		heightMap[coreRes.Height] = append(heightMap[coreRes.Height], coreRes)
 	}


rootulp:
CI failed so still needs work

app/test/blob_ordering_test.go
```diff
@@ -0,0 +1,222 @@
...
+		addr := testfactory.GetAddress(kr, account)
+		acc := testutil.DirectQueryAccount(testApp, addr)
+		signer, err := user.NewSigner(
+			kr, encConf.TxConfig, testutil.ChainID,
+			user.NewAccount(account, acc.GetAccountNumber(), acc.GetSequence()),
+		)
+		require.NoError(t, err)
+
+		blob, err := blobtypes.NewV0Blob(ns, random.Bytes(blobSize))
+		require.NoError(t, err)
```
ninabarbakadze, line 63: don't we have a method that does all of this for you in blobfactory?

PR #6729, opened 2026-03-05
feat!: add PayForFibre support to PrepareProposal and ProcessProposal

rootulp:
## Context

Part of #6623 — porting celestia-app-fibre contents to celestia-app.

Closes #6751.

This is PR 1 of 2. PR 2 will update `pkg/da/data_availability_header.go` to add squarev4 support.

### `go.mod`
- Updated `go-square/v4` to [v4.0.0-rc4](https://github.com/celestiaorg/go-square/releases/tag/v4.0.0-rc4) which simplifies `TryParseFibreTx` from 3 return values to 2. `Construct` detects plain SDK `MsgPayForFibre` bytes natively.

### `app/filtered_square_builder.go`
- `separateTxs` returns a third category for plain SDK txs containing `MsgPayForFibre`. Filters out invalid txs: undecodable, multi-PFF, and mixed PFF+other messages.
- `Fill` processes PayForFibre txs: calls `tx.TryParseFibreTx` to synthesize the system blob for the builder (size accounting), validates via the ante handler, then **returns plain `rawTx` bytes** (not FibreTx-wrapped). This ensures the tx hash the client submitted is the hash committed in the block, so `ConfirmTx` works.
- Enforces `MaxPayForFibreMessages` cap, consistent with existing `MaxPFBMessages` and `MaxSDKMessages` caps.
- Added `countMsgPayForFibre` helper to detect `MsgPayForFibre` in a tx.

### `app/process_proposal.go`
- Counts `MsgPayForFibre` messages separately from SDK and PFB messages with a dedicated `pffMessageCount`.
- Rejects txs with multiple `MsgPayForFibre` or `MsgPayForFibre` mixed with other messages (consistent with BlobTx requiring exactly one `MsgPayForBlobs`).
- Rejects blocks exceeding `MaxPayForFibreMessages`.
- `squarev4.Construct` + `da.ExtendSharesWithTreePool` replace `da.ConstructEDSWithTreePool`.

### `pkg/appconsts/prepare_proposal_consts.go`
- Added `MaxPayForFibreMessages = 200`.
- **Breaking**: renamed `MaxNonPFBMessages` to `MaxSDKMessages` for clarity now that there are three distinct message categories (SDK, PFB, PayForFibre).

### Tests
[...]

PR #6732, opened 2026-03-06
fix(fibre): submit FibreTx directly to prevent tx hash mismatch and e2e timeout

rootulp:
## Problem

`TestFibreE2ETestSuite/Test03Put` was timing out after 30 minutes.

**Root cause**: `fibre.Put` broadcast a plain SDK `MsgPayForFibre` tx with hash H. `PrepareProposal` then wrapped it into a `FibreTx` — a new protobuf envelope containing both the SDK tx bytes and the system blob — producing a different hash H'. Since H was never committed on-chain (the block contains H', not H), `ConfirmTx(H)` polled indefinitely.

## Fix

Have the fibre client submit an already-wrapped `FibreTx` directly. `PrepareProposal` detects the wrapping via `UnmarshalFibreTx` and passes it through unchanged. The committed tx hash is now H' (the FibreTx hash), which is the same hash returned by `BroadcastTxWithWrap`, so `ConfirmTx` finds it.

## Changes

- **`pkg/user/tx_client.go`**: Add `BroadcastTxWithWrap(ctx, msgs, wrap, opts)` — signs and broadcasts a transaction, applying an optional `wrap func([]byte) ([]byte, error)` to the encoded SDK tx bytes before submission. `BroadcastTx` now delegates to this with `nil` wrap. Also extend `resignTransactionWithNewSequence` to unwrap/rewrap `FibreTx` on sequence mismatch (same pattern as the existing `BlobTx` handling).
- **`app/check_tx.go`**: Handle `FibreTx` in `CheckTx` by validating the inner SDK tx bytes so a `FibreTx` can enter the mempool and be tracked by its inner hash.
- **`app/filtered_square_builder.go`**: `separateTxs` now detects already-wrapped `FibreTx` before the `BlobTx` check and routes them to `payForFibreTxs`. `Fill` detects already-wrapped `FibreTx` and passes them through without re-wrapping.
- **`fibre/client_put.go`**: Use `BroadcastTxWithWrap` with a closure that wraps the signed SDK tx bytes as a `FibreTx`.

## Test plan

- [ ] `go test -run TestSeparateTxs ./app/` — new cases for already-wrapped FibreTx
- [ ] `go test -run TestFilteredSquareBuilderFillWithPayForFibre ./app/` — new cases for already-wrapped FibreTx pass-through and ante rejection
- [ ] `TestFibreE2ETestSuite/Test03Put` should no longer time out

### DIFF
--- a/app/check_tx.go
+++ b/app/check_tx.go
@@ -37,6 +37,21 @@ func (app *App) CheckTx(req *abci.RequestCheckTx) (*abci.ResponseCheckTx, error)
 		return app.handleBlobCheckTx(req, btx)
 	}
 
+	// FibreTx wraps a MsgPayForFibre SDK tx with its system blob. Validate the
+	// inner SDK tx bytes so the tx can enter the mempool and be tracked by hash.
+	ftx, isFibre, err := blobtx.UnmarshalFibreTx(tx)
+	if isFibre && err != nil {
+		return responseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false), err
+	}
+	if isFibre {
+		innerReq := &abci.RequestCheckTx{Tx: ftx.Tx, Type: req.GetType()}
+		sdkTx, err := app.encodingConfig.TxConfig.TxDecoder()(ftx.Tx)
+		if err != nil {
+			return responseCheckTxWithEvents(err, 0, 0, []abci.Event{}, false), err
+		}
+		return app.forwardCheckTx(innerReq, sdkTx)
+	}
+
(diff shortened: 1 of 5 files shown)


rootulp:
Closing in favor of fixing the root cause in #6729. The correct approach is to have PrepareProposal return plain SDK tx bytes (not FibreTx wrapper bytes) so that the tx hash is stable through the PrepareProposal transformation, allowing ConfirmTx to work without any special client-side wrapping.

PR #6754, opened 2026-03-08
fix!: prevent OOM via unbounded Groth16 verifying key deserialization

rootulp:
Fixes https://dashboard.hackenproof.com/manager/companies/celestia/celestia/reports/CELESTIA-214

## Summary

- Add `ValidateGroth16Vkey()` that performs two checks on `Groth16Vkey` in `ValidateBasic()` and genesis `Validate()` **before** passing it to gnark's deserializer:
  1. **Exact size check**: `Groth16VkeySize` (396 bytes) — rejects payloads that are too short or too long
  2. **G1.K length prefix check**: `Groth16VkeyG1KLength` (3) — reads the big-endian `uint32` at bytes 288–291 and verifies it equals the expected value (`nPublic + 1` for SP1's 2 public inputs)
- Without both checks, a crafted payload can cause gnark-crypto to `make([]G1Affine, 0xFFFFFFFF)` (~256 GiB allocation), crashing any node during `CheckTx` with no authentication required
- The size check alone is insufficient: an attacker can craft a 396-byte payload (passes size check) with an inflated G1.K length prefix that still triggers OOM
- Add POC test demonstrating the underlying gnark vulnerability (64 MiB allocation from 292-byte input, skipped in `-short` mode)
- Add regression tests in both `msgs_test.go` and `genesis_test.go` verifying malicious payloads are rejected before deserialization, including the 396-byte bypass variant

## Test plan

- [x] `go test -v ./x/zkism/types/` — all tests pass including new OOM payload rejection tests
- [x] `go test -v ./x/zkism/internal/groth16/` — POC test confirms 65 MiB allocation from 292-byte input (233,472× amplification)
- [x] `go test -v -short ./x/zkism/internal/groth16/` — POC test correctly skipped in short mode
- [x] `make build-standalone` — compiles cleanly

### DIFF
--- a/x/zkism/types/genesis.go
+++ b/x/zkism/types/genesis.go
@@ -38,8 +38,8 @@ func (gs GenesisState) Validate() error {
 			return errorsmod.Wrapf(ErrInvalidMerkleTreeAddress, "ism %s merkle tree address must be 32 bytes", ism.Id.String())
 		}
 
-		if len(ism.Groth16Vkey) != Groth16VkeySize {
-			return errorsmod.Wrapf(ErrInvalidVerifyingKey, "ism %s groth16 vkey must be exactly %d bytes, got %d", ism.Id.String(), Groth16VkeySize, len(ism.Groth16Vkey))
+		if err := ValidateGroth16Vkey(ism.Groth16Vkey); err != nil {
+			return errorsmod.Wrapf(ErrInvalidVerifyingKey, "ism %s: %v", ism.Id.String(), err)
 		}
 
 		if _, err := groth16.NewVerifyingKey(ism.Groth16Vkey); err != nil {
--- a/x/zkism/types/keys.go
+++ b/x/zkism/types/keys.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"strings"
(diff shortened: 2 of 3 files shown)


x/zkism/internal/groth16/groth16_test.go
```diff
@@ -21,6 +22,80 @@ func TestNewVerifyingKey(t *testing.T) {
...
+// total) followed by a uint32 length prefix for the G1.K slice. An attacker
+// can set this to 0xFFFFFFFF, triggering a ~256 GiB allocation from a
+// ~292-byte input.
+//
+// This test uses a moderate G1.K length (1M elements = ~64 MB) to safely
+// demonstrate the code path without crashing the test runner. The allocation
+// happens, then gnark fails with EOF because no element data follows. In
+// production with 0xFFFFFFFF, the allocation would be ~256 GiB and crash the
+// node.
+func TestNewVerifyingKey_OOMVulnerability(t *testing.T) {
```
rootulp, line 40: I'm open to removing this before merging b/c it provides value for reviewers pre-merge but won't provide much value post-merge.

Claude used this to verify the bug before implementing a fix.

x/zkism/types/genesis.go
```diff
@@ -38,6 +38,10 @@ func (gs GenesisState) Validate() error {
 			return errorsmod.Wrapf(ErrInvalidMerkleTreeAddress, "ism %s merkle tree address must be 32 bytes", ism.Id.String())
 		}
 
+		if len(ism.Groth16Vkey) != Groth16VkeySize {
```
rach-id: if I understand correctly, the layout as specified in the PR is:
```
// Layout: 6 curve points (288 bytes) + uint32 G1.K length (4 bytes) +
	// 3 × compressed G1 points (96 bytes) + trailing metadata (8 bytes) = 396
```

what stops a malicious attacker from having a valid payload and only switch the G1.K length to be the maximum? 
In this case, it would pass the size check, but still OOM. So probably we should check whether the length is reasonable before allocating the slice, no?

rootulp: Very good point. I'll try to add a test case to reproduce this

rootulp: Great catch Rachid! thanks a ton. Fixed

damiannolan:
I will take a deeper look at this tomorrow. It makes sense to me, but I don't know if allocation of 256GB is possible. It looks like some reasonable size limits were added in https://github.com/Consensys/gnark-crypto/pull/759 for many of the primitives including that which is used here, right? bn254 prime fields. cc. @jonas089 

```golang
// to avoid allocating too large slice when the header is tampered, we limit
// the maximum allocation. We set the target to 4GB. This incurs a performance
// hit when reading very large slices, but protects against OOM.
targetSize := uint64(1 << 32) // 4GB
if bits.UintSize == 32 {
    // reduce target size to 1GB on 32 bits architectures
    targetSize = uint64(1 << 30) // 1GB
}
maxAllocateSliceLength := targetSize / uint64(Bytes)
```


edit: I see the actual point of slice allocation now in `ecc/bn254/marshal.go`.

jonas089 (approved):

x/zkism/internal/groth16/groth16_test.go
damiannolan: Fine with keeping it if you want. Could optionally always skip the test to avoid every `make test-short` allocating an additional 64MiB

PR #6829, opened 2026-03-12
fix!: prevent nil panic in PaymentPromise.Validate and add unit tests

### LINKED ISSUE #6712: Add unit tests to Validate p.SignerKey
[blocking]
if `p.SignerKey == nil`, `p.SignerKey.Key` would panic. 

By Claude

```suggestion
	if p.SignerKey == nil || len(p.SignerKey.Key) != secp256k1.PubKeySize {
		return fmt.Errorf("signer key must be %d bytes, got %d", secp256k1.PubKeySize, len(p.SignerKey))
```

_Originally posted by @rach-id in https://github.com/celestiaorg/celestia-app/pull/6706#discussion_r2883484302_

### LINKED ISSUE #6832: A nil `SignerKey` causes panic for `ToProto()` and `SignBytes()`
🟣 Pre-existing issue: `ToProto()` (line 114: `*p.SignerKey`) and `SignBytes()` (line 223: `p.SignerKey.Bytes()`) will also panic on nil `SignerKey` if called independently of `Validate()`. Since this PR is specifically about nil `SignerKey` safety, consider adding nil guards to these methods too for consistency.

<details>
<summary>Extended reasoning...</summary>

## What the bug is

The PR fixes a nil pointer panic in `Validate()` when `SignerKey` is nil, but two other methods on `PaymentPromise` have the same nil dereference hazard:

- `ToProto()` at line 114 does `SignerPublicKey: *p.SignerKey`, which dereferences the pointer and will panic if `SignerKey` is nil.
- `SignBytes()` at line 223 calls `p.SignerKey.Bytes()`, which is a method call on a nil pointer and will also panic.

## Concrete example

Consider code that constructs a `PaymentPromise` but fails to set `SignerKey` (or explicitly sets it to nil), then calls `ToProto()` directly — for example, during serialization via `MarshalBinary()` which calls `ToProto()` internally:

```go
pp := &fibre.PaymentPromise{ChainID: "test", ...} // SignerKey is nil
data, err := pp.MarshalBinary() // calls ToProto() -> panics at *p.SignerKey
```

Similarly, calling `SignBytes()` or `Hash()` (which calls `SignBytes()`) on such a promise would panic at `p.SignerKey.Bytes()`.

## Why existing code doesn't prevent it

In the current codebase, all callers of `ToProto()` and `SignBytes()` operate on promises that have been validated or constructed with valid keys (e.g., in `store.go`, `client_upload.go`, `client_put.go`). The expected usage pattern is to call `Validate()` first. Additionally, within `Validate()` itself, the nil check on `SignerKey` now correctly precedes the call to `SignBytes()`, so that internal path is safe.

However, `ToProto()` and `SignBytes()` are exported public methods. Nothing in the type system or API prevents a caller from invoking them directly without calling `Validate()` first. A counterargument is that these methods have an implicit precondition that the struct is fully constructed (standard Go practice), and adding nil checks to every method would be non-idiomatic. This is a fair point — Go structs commonly assume valid construction, and the original bug in `Validate()` was a specific logic error where the error format string accessed `p.SignerKey.Key` *after* confirming `SignerKey` was nil, which is qualitatively different from a general missing-precondition issue.

## Impact and recommendation

The practical impact is low since all current callers construct promises with valid keys. This is purely pre-existing — the PR does not add new callers to these methods, does not increase the surface area, and does not change control flow to make nil `SignerKey` more likely. However, since the PR is specifically about nil `SignerKey` safety and demonstrates the pattern of guarding against it, it would be a natural follow-up to add similar guards (or return errors) in `ToProto()` and `SignBytes()` for defensive consistency. This should not block the PR.

## Suggested fix

For `ToProto()`, return an error or check for nil before dereferencing. For `SignBytes()`, add a nil check before calling `.Bytes()`. Alternatively, document the precondition that `Validate()` must be called before using these methods.

</details>

_Originally posted by @claude[bot] in https://github.com/celestiaorg/celestia-app/pull/6829#discussion_r2927234836_

rootulp:
## Summary
- Fix nil pointer panic in `PaymentPromise.Validate()` when `SignerKey` is nil — the error message previously accessed `p.SignerKey.Key` unconditionally
- Add nil `SignerKey` guard to `ToProto()`, changing its return type to `(*types.PaymentPromise, error)` to prevent nil pointer dereference at `*p.SignerKey`
- Add nil `SignerKey` guard to `SignBytes()` to prevent nil pointer dereference at `p.SignerKey.Bytes()`
- Update all callers of `ToProto()` (`MarshalBinary`, `Store.Put`, `Client.Upload`, `Put`, and test helper) to handle the new error return
- Add unit tests covering every validation branch and nil `SignerKey` paths in `ToProto`, `MarshalBinary`, and `SignBytes`

Closes #6712
Closes #6832

## Test plan
- [x] `go test -v -run TestPaymentPromise ./fibre/` passes with all subtests
- [x] `TestNilSignerKey` subtests for `ToProto`, `MarshalBinary`, and `SignBytes` verify they return errors instead of panicking
- [x] Full `go build ./...` compiles cleanly

### DIFF
--- a/fibre/internal/e2e/fibre_e2e_test.go
+++ b/fibre/internal/e2e/fibre_e2e_test.go
@@ -207,6 +207,8 @@ func (s *FibreE2ETestSuite) Test02FundEscrowAccount() {
 	require.Equal(t, uint32(0), txResp.Code)
 	t.Logf("First deposit tx at height %d", txResp.Height)
 
+	require.NoError(t, s.cctx.WaitForNextBlock())
+
 	// verify escrow balance matches deposit.
 	escrowResp, err = fibreQueryClient.EscrowAccount(ctx, &fibretypes.QueryEscrowAccountRequest{
 		Signer: addr.String(),
@@ -225,6 +227,8 @@ func (s *FibreE2ETestSuite) Test02FundEscrowAccount() {
 	require.NoError(t, err)
 	require.Equal(t, uint32(0), txResp.Code)
 	t.Logf("Second deposit tx at height %d", txResp.Height)
+
+	require.NoError(t, s.cctx.WaitForNextBlock())
 
 	// verify cumulative balance is 75 TIA.
 	escrowResp, err = fibreQueryClient.EscrowAccount(ctx, &fibretypes.QueryEscrowAccountRequest{
--- a/fibre/internal/grpc/host_registry_e2e_test.go
+++ b/fibre/internal/grpc/host_registry_e2e_test.go
@@ -128,6 +128,8 @@ func (s *IntegrationTestSuite) TestGetHostWithRegistration() {
 	require.Equal(t, uint32(0), txResp.Code, "transaction failed with code %d", txResp.Code)
 	t.Logf("Transaction submitted successfully. TxHash: %s, Height: %d", txResp.TxHash, txResp.Height)
 
+	require.NoError(t, s.cctx.WaitForNextBlock())
+
 	host, err := s.hostRegistry.GetHost(s.cctx.GoContext(), s.validator)
 	require.NoError(t, err, "GetHost should now succeed")
 	require.NotEmpty(t, host.String())
@@ -144,6 +146,8 @@ func (s *IntegrationTestSuite) TestGetHostWithRegistration() {
 	require.NoError(t, err, "failed to submit transaction")
 	require.Equal(t, uint32(0), txResp.Code, "transaction failed with code %d", txResp.Code)
 	t.Logf("Transaction submitted successfully. TxHash: %s, Height: %d", txResp.TxHash, txResp.Height)
+
+	require.NoError(t, s.cctx.WaitForNextBlock())
 
 	host, err = s.hostRegistry.GetHost(s.cctx.GoContext(), s.validator)
 	require.NoError(t, err)


fibre/payment_promise.go
```diff
@@ -120,7 +120,10 @@ func (p *PaymentPromise) ToProto() *types.PaymentPromise {
 // It verifies all field constraints and validates the [PaymentPromise.Signature] using [PaymentPromise.SignerKey].
 func (p *PaymentPromise) Validate() error {
 	// signer key must be valid secp256k1 public key (33 bytes)
-	if p.SignerKey == nil || len(p.SignerKey.Key) != secp256k1.PubKeySize {
+	if p.SignerKey == nil {
+		return fmt.Errorf("signer key must be %d bytes, got 0", secp256k1.PubKeySize)
+	}
```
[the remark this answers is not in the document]
rootulp: https://github.com/celestiaorg/celestia-app/pull/6829/commits/6ee304e338b708d4f1adaa746cfd5f977309976e

fibre/payment_promise.go
```diff
@@ -102,7 +105,10 @@ func (p *PaymentPromise) FromProto(pbMsg *types.PaymentPromise) error {
 }
 
 // ToProto converts the [PaymentPromise] to its protobuf representation.
-func (p *PaymentPromise) ToProto() *types.PaymentPromise {
+func (p *PaymentPromise) ToProto() (*types.PaymentPromise, error) {
+	if p.SignerKey == nil {
+		return nil, errors.New("signer key must not be nil")
+	}
```
Wondertan, line 111: While the original catch makes sense, Claude's suggestion doesn't. 

`ToProto` and `SignBytes` already assume validated PP.  We could remove this assumption and make both perform additional validations, but then we should not do it only for `SignerKey` but for all other fields, including `p.Namespace`, which then makes us basically reimplement the whole `Validate` in both `ToProto` and `SignBytes`. So instead of checking all fields again manually in each method, we could simply make them call `Validate`, but then we end up validating fields multiple times across the lifetime of a PP for no good reason.

rootulp: What is your concrete suggestion?

rootulp:
Agreed. Oops I had `!` on the PR that merged into this PR but not this one.

fibre/payment_promise.go
rootulp: I agree we shouldn't re-implement Validate() in every method — that's not the intent here. The distinction is that SignerKey is the only field where a nil value causes a panic rather than just a wrong result:

  - `ToProto()`: SignerPublicKey: *p.SignerKey — nil pointer dereference
  - `strippedSignBytes()`: p.SignerKey.Bytes() — nil pointer dereference

The other fields (Namespace, ChainID, UploadSize, etc.) can't panic — they just produce an incorrect proto message if they hold zero values. So these nil guards aren't field validation, they're preventing a process crash. Converting a panic into an error return seems worth the two extra lines, even if callers are expected to have called Validate() first.

Wondertan: @rootulp, an incorrect proto message is even worse than panic, because it leads to incorrect application state, while panic halts the application, preventing it from entering an incorrect state. The point is that all the fields should be validated universally, and having this check only for the signer key adds zero value and only bloats code in practice. I know it may feel productive to have this change, but it actually is not.

PR #6833, opened 2026-03-12
fix!: add nil SignerKey guards to ToProto() and SignBytes()

### LINKED ISSUE #6832: A nil `SignerKey` causes panic for `ToProto()` and `SignBytes()`

rootulp:
## Summary
- Adds nil `SignerKey` guard to `ToProto()`, changing its return type to `(*types.PaymentPromise, error)` to prevent nil pointer dereference at `*p.SignerKey`
- Adds nil `SignerKey` guard to `SignBytes()` to prevent nil pointer dereference at `p.SignerKey.Bytes()`
- Updates all callers of `ToProto()` (`MarshalBinary`, `Store.Put`, `Client.Upload`, `Put`, and test helper) to handle the new error return

Closes #6832

## Test plan
- [x] Added `TestNilSignerKey` with subtests for `ToProto`, `MarshalBinary`, and `SignBytes` verifying they return errors instead of panicking
- [x] Existing tests pass (`TestPaymentPromise`, `TestValidatorSignPaymentPromise`)
- [x] Full `go build ./...` compiles cleanly

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-app/pull/6833" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/fibre/payment_promise.go
+++ b/fibre/payment_promise.go
@@ -56,7 +56,10 @@ type PaymentPromise struct {
 
 // MarshalBinary encodes the [PaymentPromise] using protobuf.
 func (p *PaymentPromise) MarshalBinary() ([]byte, error) {
-	pbMsg := p.ToProto()
+	pbMsg, err := p.ToProto()
+	if err != nil {
+		return nil, err
+	}
 	return gogoproto.Marshal(pbMsg)
 }
 
@@ -102,7 +105,10 @@ func (p *PaymentPromise) FromProto(pbMsg *types.PaymentPromise) error {
 }
 
 // ToProto converts the [PaymentPromise] to its protobuf representation.
-func (p *PaymentPromise) ToProto() *types.PaymentPromise {
+func (p *PaymentPromise) ToProto() (*types.PaymentPromise, error) {
+	if p.SignerKey == nil {
+		return nil, errors.New("signer key must not be nil")
+	}
 	return &types.PaymentPromise{
 		ChainId:           p.ChainID,
 		Height:            int64(p.Height),
@@ -113,7 +119,7 @@ func (p *PaymentPromise) ToProto() *types.PaymentPromise {
 		CreationTimestamp: p.CreationTimestamp,
 		SignerPublicKey:   *p.SignerKey,
 		Signature:         p.Signature,
-	}
+	}, nil
 }
 
 // Validate performs stateless validation on the [PaymentPromise].
@@ -203,6 +209,9 @@ const (
 // SignBytes caches the result of the computation for subsequent calls,
 // so its not allowed to change the promise after signing.
 func (p *PaymentPromise) SignBytes() ([]byte, error) {
+	if p.SignerKey == nil {
+		return nil, errors.New("signer key must not be nil")
+	}
 	p.signBytesOnce.Do(func() {
 		// use MarshalBinary for timestamp
 		timestampBytes, err := p.CreationTimestamp.UTC().MarshalBinary() // this must be UTC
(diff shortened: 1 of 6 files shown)


fibre/payment_promise.go
```diff
@@ -102,7 +105,10 @@ func (p *PaymentPromise) FromProto(pbMsg *types.PaymentPromise) error {
 }
 
 // ToProto converts the [PaymentPromise] to its protobuf representation.
-func (p *PaymentPromise) ToProto() *types.PaymentPromise {
+func (p *PaymentPromise) ToProto() (*types.PaymentPromise, error) {
```
rootulp, line 108: This is API breaking but we haven't released v8 yet so it's safe to include now

PR #6843, opened 2026-03-13
docs: ADR027 Sequencer Total Ordering on Fibre via Simplex

evan-forbes:
## Overview

total ordering via adding a simplex to fibre! :tada: 

closes https://linear.app/celestia/issue/PROTOCO-1208/write-spec-and-implementation-plan-for-adding-ordering-to-a

the context and protocol summary at the top of the document are meant to be the best entry points into the document.

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-app/pull/6843" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,315 @@
...
+- **ParentPaymentPromise.** A stripped-down representation of the parent `PaymentPromise` containing only the fields needed to reconstruct v1 sign bytes for QC verification. Excludes the sequencer signature (redundant — the parent QC already proves >2/3 validators verified it) and excludes the parent's own ordering fields (which would cause recursive nesting).
+- **Quorum Certificate (QC).** A set of > 2/3 voting power in validator ed25519 signatures over an `OrderedPaymentPromise`. Carried forward as proof of the parent's certification.
+
+The protocol proceeds in four steps per sequence:
+
+1. **Propose.** The sequencer constructs an `OrderedPaymentPromise` at sequence `s` containing `QC_{s-1}` (or empty at genesis), signs it with v1 sign bytes, and uploads shards to validators.
+2. **Vote.** Each validator verifies the parent QC, checks for double-voting, and signs the promise if all checks pass.
+3. **Certify.** The sequencer collects > 2/3 voting power in signatures, forming `QC_s`. Block `B_s` is now **certified/locked**.
```
rach-id, line 33: I find the `block` word confusing in this context because it's not a block, it's a single blob we're getting ordering on

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,315 @@
...
+4. **Finalize.** When the sequencer posts `B_{s+1}` and obtains `QC_{s+1}`, block `B_s` is **finalized**. The second quorum on the child provides two-phase BFT confirmation.
```
rach-id: When would these blocks be pruned? if I understand correctly, fibre prunes the blobs after 24hrs, what happens to lanes that haven't had a blob for more than 24hrs? what would the parent promise be?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,315 @@
...
+
+Finality requires two QC rounds — one to lock, one to finalize — matching the minimal two-phase structure of standard BFT protocols. The rest of this document specifies each component in detail.
+
+## Specification
+
+### Model
+
+The Fibre validator set follows the standard weighted BFT model: `n = 3f + 1`, or equivalently, a quorum requires `> 2/3` of voting power. Liveness requires `> 2/3` of voting power to be online for any QC to form.
+
+There is no namespace authorization mechanism — any signer may start a new chain in any namespace. Lane isolation comes from the signer's public key being part of the lane key and the v1 sign bytes: different signers produce different sign bytes even at the same sequence in the same namespace. The chaining rule (same signer in parent and child) ensures each signer's chain is independent. Authored namespaces could further restrict who may write to a namespace, but this is orthogonal to the ordering protocol.
```
rach-id, line 44: what would Celestia state machine enforce in this model? would it enforce the blocks to be included in order? or the state machine will treat them just as ordinary PPs?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,315 @@
...
+
+**Step 2: Vote.** Each validator runs the ordering `PromiseRule` ([ADR-025](adr-025-promise-rules.md)). If it passes (along with all existing Fibre verification: promise validation, shard assignment, proof verification), the validator signs the `OrderedPaymentPromise` via `SignOrderedPaymentPromiseValidator` and returns the signature.
+
+**Step 3: Certify.** The sequencer collects `> 2/3` voting power in validator signatures, forming `QC_s`. This is the first-phase certificate — `B_s` is now **certified/locked**.
+
+**Step 4: Finalize.** When the sequencer posts `B_{s+1}` with `QC_s` as its parent and obtains `QC_{s+1}`, then `B_s` is **finalized**. The second quorum on the child block provides the second phase of confirmation.
+
+### Safety Rules
+
+1. **No double vote** — A validator MUST NOT sign two different `Commitment` values for the same `(chain_id, namespace, signer_public_key, sequence)`. Signing the same `Commitment` is permitted (idempotent).
```
rach-id: how are these enforced?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,315 @@
...
+// Format: "fibre/pp:v1" || chainID || signerPubKey(33) || namespace(29) ||
+//         blobSize(4) || commitment(32) || blobVersion(4) || height(8) ||
+//         creationTimestamp(15) || sequence(8)
+func (o *OrderedPaymentPromise) SignBytes() ([]byte, error) {
+    // ... v1 computation ...
+}
+
+// SignBytesValidator returns the v1 sign bytes wrapped with CometBFT
+// domain separation using the v1 prefix.
+func (o *OrderedPaymentPromise) SignBytesValidator() ([]byte, error) {
```
rach-id, line 119: this was changed with: https://github.com/celestiaorg/celestia-app/commit/6278a4edc31db0f9a1b69ffdd7c72c065674cc3e

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,315 @@
...
+
+**Step 1: Propose.** The sequencer constructs an `OrderedPaymentPromise` for sequence `s` with `parent_promise` set to the serialized `ParentPaymentPromise` from sequence `s-1` and `parent_signatures` set to the validator signatures from `QC_{s-1}` (or both empty if `s == 1`). It signs the promise with v1 sign bytes, encodes the rollup block as a Fibre blob, and uploads shards to validators via `UploadShard`.
+
+**Step 2: Vote.** Each validator runs the ordering `PromiseRule` ([ADR-025](adr-025-promise-rules.md)). If it passes (along with all existing Fibre verification: promise validation, shard assignment, proof verification), the validator signs the `OrderedPaymentPromise` via `SignOrderedPaymentPromiseValidator` and returns the signature.
+
+**Step 3: Certify.** The sequencer collects `> 2/3` voting power in validator signatures, forming `QC_s`. This is the first-phase certificate — `B_s` is now **certified/locked**.
+
+**Step 4: Finalize.** When the sequencer posts `B_{s+1}` with `QC_s` as its parent and obtains `QC_{s+1}`, then `B_s` is **finalized**. The second quorum on the child block provides the second phase of confirmation.
+
+### Safety Rules
```
rach-id, line 58: how to ensure validators won't selectively sign lanes? censoring certain lanes and signing the others?

rach-id, line 58: The protocol now forces sequences to submit two blobs to ensure finalization. If for example the sequencer doesn't get transactions and doesn't want to post anything, they will be forced to post one last time to finalize

rach-id, line 58: Is there a recovery mechanism for split signatures? For example, a sequencer sends a blob to 1/3 and another to the other 1/3. This means commitment/finalization can't ever happen in that lane

rach-id (commented):
Clear and well written 🎉 

Left a few questions to understand more the edge cases

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,285 @@
...
+  // ... existing fields 1-9 ...
+
+  // sequence is the monotonic ordering position for this lane.
+  // Zero means non-ordered (the proto default). Ordering starts at 1.
+  uint64 sequence = 10;
+
+  // parent_promise is the serialized PaymentPromise from sequence s-1.
+  // Present when sequence > 1. Empty for genesis (sequence == 1) and
+  // non-ordered (sequence == 0).
+  bytes parent_promise = 11;
```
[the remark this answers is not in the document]
evan-forbes: actually ended up switching back to full chaining. the recursive thing here actually wasn't true afaiu since its a hash and a parent promise, maybe I missed something though :thinking:

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: >Is there a recovery mechanism for split signatures? For example, a sequencer sends a blob to 1/3 and another to the other 1/3. This means commitment/finalization can't ever happen in that lane

10000%!!! yeah exactly. sequencers that equivocate halt their lane forever or at least until the validator's local state is pruned

evan-forbes: > they will be forced to post one last time to finalize

yeah true this is a downside to pipelining instead of only getting a second round of voting

evan-forbes: > censoring certain lanes and signing the others?

atm afaiu there is a single solution for censorship for any chain and its social consensus. very possible that this expectation must be clearly set by social consensus. atm censorship by the validator set would result in a hardfork.

censorship could be more devious, in that its not censoring specific content, but specific users.

The benefit here is that the user gets per block censorship resistance since each block is communicated directly to all validators.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: oh nice will update here thank

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: good point, origninally I was leaving this out but in theory we need equivocation fraud proofs that are almost identical to the evidence we have in comet atm

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: the celestia-app state machine doens't need to do anything here, which is nice! however the validators ofc need to maintain some temporary local state in order to guarantee ordering

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: the blobs can be pruned like normal, however the local state for lane watermarks should be kept for the remainder of the unbonding period

fortunately that is very small, I think a hash and an int per lane.

those can be pruned after the unbonding period. it is a bit weirder than a normal chain, since the validator set can change even if there is not a block produced, since the validator set is not determined by the rollup being updated.

but yeah correct, each lane needs to update once per unbonding period to maintain their local state to preserve ordering!

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
rach-id: So comet will slash fibre evidence?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
rach-id: What about slashing or something? I would see this happening because of a bug and we need a recovery mechanism to unblock that, no?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+The `(chain_id, namespace, signer_public_key)` triple defines a total order of posted blobs. The position in this order is the **sequence** `s`. Ordering starts at sequence 1; sequence 0 is reserved for non-ordered promises (the proto default for `uint64`).
+
+### Protocol
+
+**Step 1: Propose.** The sequencer constructs an `OrderedPaymentPromise` for sequence `s` with `previous_payment_promise_hash` set to the ordered hash of sequence `s-1`, `parent_promise` set to the serialized `ParentPaymentPromise` from sequence `s-1`, and `parent_signatures` set to the validator signatures from `QC_{s-1}` (or all parent fields empty if `s == 1`). It signs the promise with v1 sign bytes, encodes the rollup block as a Fibre blob, and uploads shards to validators via `UploadShard`.
+
+**Step 2: Vote.** Each validator runs the ordering `PromiseRule` ([ADR-025](adr-025-promise-rules.md)). If it passes (along with all existing Fibre verification: promise validation, shard assignment, proof verification), the validator signs the `OrderedPaymentPromise` via `SignOrderedPaymentPromiseValidator` and returns the signature.
```
cmwaters, line 52: Does it make sense to split this out into a separate ADR? It's like an antehandler for payment promise and you actually define the concrete verification methods we'd support

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Alternative Approaches
+
+### A. No ordering from fibre signatures
+
+Rollups continue using Celestia block finality for ordering. The sequencer posts blobs via Fibre for availability, but ordering comes from Celestia headers with namespace inclusion proofs. This is the status quo and is already functional. The cost is higher latency (~6s for Celestia finality) and more complex verification (inclusion proofs, share parsing).
+
+### B. Future extensions
+
+The following are not alternatives to this design but extensions that build on top of it:
```
cmwaters, line 306: Can this be a different section outside of Alternative Approaches

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+**Step 3: Certify.** The sequencer collects `> 2/3` voting power in validator signatures, forming `QC_s`. This is the first-phase certificate — `B_s` is now **certified/locked**.
+
+**Step 4: Finalize.** When the sequencer posts `B_{s+1}` with `previous_payment_promise_hash = hash(B_s)`, includes `QC_s` as witness data, and obtains `QC_{s+1}`, then `B_s` is **finalized**. The second quorum on the child block provides the second phase of confirmation.
+
+### Safety Rules
+
+1. **No double vote** — A validator MUST NOT sign two different ordered-promise hashes for the same `(chain_id, namespace, signer_public_key, sequence)`. Signing the same ordered-promise hash is permitted (idempotent).
+2. **Chain extension** — A validator signs sequence `s` only if the proposal includes a valid `QC_{s-1}` (or `s == 1`) and that QC certifies the exact parent whose ordered hash equals `previous_payment_promise_hash`.
```
cmwaters, line 61: So the combination of the two will enforce monotonically increasing sequence numbers right?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+  // promise's height. Entry i corresponds to validator i in that set.
+  // Non-signing validators have empty bytes (length 0). Signing
+  // validators have exactly 64 bytes (ed25519 signature). The length
+  // of this list equals the validator set size at the parent's height.
+  //
+  // This matches the existing SignedPaymentPromise.ValidatorSignatures
+  // encoding used by the Fibre client.
+  //
+  // Present when sequence > 1. Empty for genesis and non-ordered.
+  repeated bytes parent_signatures = 13;
```
cmwaters, line 233: If these are part of `PaymentPromise` then they also make there way on-chain - is this necessary? if so why?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Safety Rules
+
+1. **No double vote** — A validator MUST NOT sign two different ordered-promise hashes for the same `(chain_id, namespace, signer_public_key, sequence)`. Signing the same ordered-promise hash is permitted (idempotent).
+2. **Chain extension** — A validator signs sequence `s` only if the proposal includes a valid `QC_{s-1}` (or `s == 1`) and that QC certifies the exact parent whose ordered hash equals `previous_payment_promise_hash`.
+3. **Commit rule** — `B_s` is **certified/locked** when `QC_s` exists, and **finalized** when `QC_{s+1}` exists for a child whose `previous_payment_promise_hash = hash(B_s)`. A single QC is unique (no conflicting QC can form under the no-double-vote rule), but may not have been observed by enough validators to survive sequencer failure. The second QC on the child proves that `> 2/3` of validators observed the lock on that exact parent, ensuring the chain can continue after sequencer failure.
```
cmwaters, line 62: This feels familiar with someone that understand how BFT consensus works. But I want to probe this a little more. Why do we need the second round? There is actually no leader change supported in this system. Fibre validators could persist and serve their signatures in the event of a sequencer failure.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+- `previous_payment_promise_hash` is empty or all-zero
+- `parent_promise` is empty (no previous sequence to certify)
+- `parent_signatures` is empty
+- Validators accept empty parent data only when `sequence == 1`
+
+The sequencer posts a blob with `sequence = 1` and empty parent fields. Validators sign it using v1 sign bytes if all other checks pass (no double vote). The resulting `QC_1` becomes the anchor for the chain — the signer who obtained `QC_1` is established as the lane's sequencer.
+
+### OrderedPromise Validity Rules
+
+On receiving a `PaymentPromise` with `sequence > 0`, the ordering `PromiseRule` ([ADR-025](adr-025-promise-rules.md)) constructs an `OrderedPaymentPromise` wrapper and verifies:
```
cmwaters, line 256: Why can't the user submit the OrderedPaymentPromise? as a new transaction type

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+5. **No double vote** — The validator has not already signed a different ordered-promise hash at the same `(chain_id, namespace, signer_public_key, sequence)`. Signing the **same** ordered-promise hash is permitted (idempotent) to support sequencer crash recovery. Stale proposals at sequences ≤ the lane watermark are implicitly rejected — see [Local Validator State](#local-validator-state).
+
+If all checks pass, the validator signs the `OrderedPaymentPromise` via `SignOrderedPaymentPromiseValidator` and returns the signature in the `UploadShardResponse`.
+
+### Local Validator State
+
+To enforce the no-double-vote and chain-extension rules, validators must durably track two pieces of state per ordered lane `(chain_id, namespace, signer_public_key)`:
+
+**1. Lane watermark** — the highest certified sequence the validator has adopted. This is the sequence from the most recently verified parent QC, not the highest sequence the validator has personally signed.
```
cmwaters, line 274: Why does it not save the highest sequence the validator has personally signed.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+4. **Existing Fibre checks** — Commitment correctness and data availability checks in the Fibre shard verification pipeline apply unchanged. However, the existing `Validate()` sequencer signature check (which verifies against v0 `SignBytes()`) must be skipped or adapted for ordered promises (`sequence > 0`), since the ordering PromiseRule handles signature verification with v1 sign bytes.
+
+5. **No double vote** — The validator has not already signed a different ordered-promise hash at the same `(chain_id, namespace, signer_public_key, sequence)`. Signing the **same** ordered-promise hash is permitted (idempotent) to support sequencer crash recovery. Stale proposals at sequences ≤ the lane watermark are implicitly rejected — see [Local Validator State](#local-validator-state).
+
+If all checks pass, the validator signs the `OrderedPaymentPromise` via `SignOrderedPaymentPromiseValidator` and returns the signature in the `UploadShardResponse`.
+
+### Local Validator State
+
+To enforce the no-double-vote and chain-extension rules, validators must durably track two pieces of state per ordered lane `(chain_id, namespace, signer_public_key)`:
```
cmwaters, line 272: Are there any bounds here or could I create millions of lanes? Probably not that problematic

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+4. **Existing Fibre checks** — Commitment correctness and data availability checks in the Fibre shard verification pipeline apply unchanged. However, the existing `Validate()` sequencer signature check (which verifies against v0 `SignBytes()`) must be skipped or adapted for ordered promises (`sequence > 0`), since the ordering PromiseRule handles signature verification with v1 sign bytes.
+
+5. **No double vote** — The validator has not already signed a different ordered-promise hash at the same `(chain_id, namespace, signer_public_key, sequence)`. Signing the **same** ordered-promise hash is permitted (idempotent) to support sequencer crash recovery. Stale proposals at sequences ≤ the lane watermark are implicitly rejected — see [Local Validator State](#local-validator-state).
+
+If all checks pass, the validator signs the `OrderedPaymentPromise` via `SignOrderedPaymentPromiseValidator` and returns the signature in the `UploadShardResponse`.
+
+### Local Validator State
```
cmwaters, line 270: What about nodes that state sync and become part of the validator set. How would they retrieve this data?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+Including `parent_promise` (a serialized `ParentPaymentPromise`) alongside `parent_signatures` makes the QC self-contained: any validator can verify the parent QC without needing to have seen or stored the parent promise previously. The validator deserializes the `ParentPaymentPromise`, recomputes the parent's v1 sign bytes and ordered hash from its fields, verifies `parent_signatures` against those recomputed sign bytes, and checks that the resulting ordered hash exactly equals the child's `previous_payment_promise_hash`. This is what enables validators to catch up after missing sequences.
+
+The v1 sign bytes include the `sequence` and `previous_payment_promise_hash` to prevent signature replay across sequences and to authenticate the parent edge. All integer fields use big-endian encoding. `creationTimestamp` is encoded via `time.Time.MarshalBinary()` (15 bytes). Sequence 0 never appears in v1 sign bytes — non-ordered promises use v0. The ordered hash used for chaining is defined over the v1 consensus object, not over recursively nested protobuf bytes, so message size stays constant per sequence apart from the carried QC witness data.
+
+### Genesis
```
cmwaters, line 243: Could the genesis block in fact refer still refer to a previous PaymentPromise in the case of a migration?

cmwaters (commented):
Excellently written and well thought through. There were several edge cases that seem to be covered here. 

Have a few questions that I would like more clarity on before I approve

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
rach-id: And if comet will slash fibre evidence, then malicious committee can have up to the block time to behave maliciously. Should there be some sort of evidence in the fibre layer?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
rach-id: what's the Celestia state machine good for in this model? nothing? just validator set changes? So, inclusion of a blob doesn't necessarily mean the inclusion of the one before it (its parent)?

evan-forbes: exactly yeah that's the goal is to not have to wait for a celestia block to get ordering

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: we can add our own module, not the x/slashing or evidence reactor etc

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: currently its similar to now if a validator double signs. 

the only way to determine if this was a bug or a malicious activity is social consensus, therefore favoring safety over liveness kinda makes sense here.

what we could do maybe is add some sort of view change as you suggest, however I'm a bit hesitant atm as that adds a lot of complication and attack surface

if we remove chaining, so that validators are not checking hashes and are only ensuring there is a single canonical blob per sequence, then we could do that!

initially I didn't do this however to provide a more typical ordering

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: that was the thinking with ADR025 yeah! but arguably not required

one thing I was trying with this flow is to try and break things up into smaller ADRs to speed up the buy-in process for implementation, but we likely don't need one

its here if anyone is interested, should be mostly up to date iirc https://github.com/celestiaorg/celestia-app/pull/6795

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: yeah good point

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: good point yeah a different struct entirely is likely better to avoid this

that's one thing I played around with and would be happy to switch to if others agree

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
rach-id: > its similar to now if a validator double signs.

I agree. But that's for validators, a validator going up/down doesn't hurt the network overall. For clients it's different. If there is a bug in the software, and end up in this situation, then that means their rollup is down indefinitly and they need to hardfork: probably re-create the signer keys, use a different namespace, handle the migration path, etc. 

I would document this decision and run it by our clients as part of this to see if this is a desired feature. Otherwise, I agree with the reasoning, you mess it up, you endure the consequences. 

Also, I don't see a way to determine wether this could be a software bug: fibre submitted different blobs. Or, it's a malicious 1/3rd providing a different signature for some blob. That's why I'm thinking of a recovery mechanism.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: yay nice fun

after the first round the block is certified or locked, and can be used to prove inclusion or even state. This is because there are two possibilities here, either the chain halts or the state is valid and canoncal

the second phase is more or less DA for votes. In order to have a canonical view, we need everyone to have over 2/3 of the votes to proceed. 

With a single round its possible someone has a different view w/o 2/3 despite the one validator pushing the 2/3 + 1 not actually sharing their vote.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: we can that's a good option!

that's actually some of the original thinking with ADR025, is just a standardized way to organize the checks we need to do so we can have different entry points into the validation logic

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: > Why does it not save the highest sequence the validator has personally signed.

ahh I think was from a previous revision as it should be only what the validator signs not what is adopted.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: yeah provided things are resourced priced appropriately and we're able to prune I think we're okay

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: I'm not sure yet, but ofc since its their local data, they would need to back it up and restore it. This is the equivalent of privval state

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: I think so yeah! that was one alternative I was thinking through. I went with creating a special case for height 0 as all the formatting is the same and perhaps this is be more explicit

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
cmwaters: But if they were not part of it before they wouldn't be able to restore it. I think there just needs to be rules that:
a) validators can sign a non zero sequence if they've not stored sequence (because they are new to the set)
b) if a validator drops out and then comes back into the set they can sign a sequence that is not monotonically higher than their previous number but can only do so once (there needs to be a flag)

^ Just suggestions - could be better ways to do this

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
cmwaters: Yeah I think we should not enforce that the previous payment promise be empty

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
cmwaters: Counterargument: The second round is to convince the fibre nodes that the previous blob got 2/3 signatures. But do we need to convince them of that. If a sequencer sends out the 2/3 signatures to other rollup nodes what condition would cause that not to be final and for the sequencer to be able to produce another valid blob with the same sequencer?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
cmwaters: Hmm, we have to be careful though, what if a sequencer waits for the validator set to change by 1/3+ and then rewrites history because the new nodes don't know that there were previous sequence numbers - maybe they need to fetch it from state?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: tbc I'm not arguing that using a centralized sequencer is a good idea

how does this differ from existing centralized rollups that double sign?

> Also, I don't see a way to determine wether this could be a software bug

since this fundamentally can only be done via social consensus, I'm missing how this changes things here

> That's why I'm thinking of a recovery mechanism.

tbc I'm not saying that we shouldn't, however it seems impossible to add one while not also adding meaningfully complexity and implementation specificity. imo the better approach would be to switch to a new proposer. one could add a backup. that backup though would be implemented how the user defines though not us

evan-forbes: for example, we could add a celestia-app tx that increment the sequence by one, unblocking the rollup from the same sequencer

but now we're adding more celestia specific implementation details to the rollup logic (skip height if ___)

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Changelog
+
+- 2026/03/07: Initial draft
+
+## Context
+
+Fibre already provides data availability: users post blobs and collect > 2/3 validator signatures as availability certificates. However, these signatures do not provide **ordering** — they attest that data is available, not that it occupies a specific position in a sequence.
+
+Rollups that need BFT ordering today must wait for Celestia finality (~ 2 * block time). The rollup waits for the blob to be included in a Celestia block, then constructs a namespace inclusion proof against the Celestia header, parses the share from that proof, and compares the commitment. This adds seconds of latency, couples ordering to Celestia's block time, and requires proving inclusion through the Celestia header — an external dependency that adds latency, complexity, and proof overhead beyond what Fibre itself provides.
```
rootulp, line 15: > Rollups that need BFT ordering today must wait for Celestia finality (~ 2 * block time)

@claude why does this say ~ 2 * block time? Celestia has immediate block finality so one block time should be sufficient for finality. Does it take an additional block time to construct an inclusion proof? That seems unrealistic because constructing an inclusion proof does not depend on block time.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Protocol Summary
+
+This ADR extends Fibre's existing `PaymentPromise` signing flow into a chained BFT ordering protocol for a single sequencer. The key concepts:
+
+- **Lane.** A `(chain_id, namespace, signer_public_key)` triple that defines an independent total order. Each signer has their own sequence within a namespace — no registry or namespace authorization is needed.
+- **Sequence.** A monotonic position within a lane, starting at 1 (genesis). Sequence 0 means non-ordered (the proto default for `uint64`) and uses v0 sign bytes, identical to an existing PaymentPromise.
+- **OrderedPaymentPromise.** A wrapper around `PaymentPromise` that adds a `sequence` number, a `previous_payment_promise_hash`, a `ParentPaymentPromise` descriptor (the parent's sign-bytes-relevant fields only), and the parent's quorum certificate. Uses v1 sign bytes (same fields as v0 but with a `fibre/pp:v1` prefix instead of `fibre/pp:v0`, and the sequence number and previous hash appended).
```
rootulp, line 25: @claude what is a quorum certificate?

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
rootulp: If block time is 6 seconds today, do roll-ups really have to wait 12 seconds for Celestia finality? That seems incorrect. Can you please explain the rationale for the ~ 2 * block time figure? @claude

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+The protocol proceeds in four steps per sequence:
+
+1. **Propose.** The sequencer constructs an `OrderedPaymentPromise` at sequence `s` containing `previous_payment_promise_hash = hash(B_{s-1})`, `ParentPaymentPromise_{s-1}`, and `QC_{s-1}` (or empty parent fields at genesis), signs it with v1 sign bytes, and uploads shards to validators.
+2. **Vote.** Each validator verifies the parent QC, checks for double-voting, and signs the promise if all checks pass.
+3. **Certify.** The sequencer collects > 2/3 voting power in signatures, forming `QC_s`. Block `B_s` is now **certified/locked**.
+4. **Finalize.** When the sequencer posts `B_{s+1}` and obtains `QC_{s+1}`, block `B_s` is **finalized**. The second quorum on the child authenticates the exact parent via `previous_payment_promise_hash`, providing two-phase BFT confirmation.
+
+Finality requires two QC rounds — one to lock, one to finalize — matching the minimal two-phase structure of standard BFT protocols. The rest of this document specifies each component in detail.
```
rootulp, line 36: Is there an estimate on how long it takes to perform two QC rounds?

At a very high level it seems like the benefit of Fibre ordering is that roll-ups don't need to wait for Celestia finality and can instead wait for two QC rounds. Numbers would help me understand the benefit.

Celestia finality (currently 6 seconds) -> Two QC rounds (time TBD)

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Specification
+
+### Model
+
+The Fibre validator set follows the standard weighted BFT model: `n = 3f + 1`, or equivalently, a quorum requires `> 2/3` of voting power. Liveness requires `> 2/3` of voting power to be online for any QC to form.
+
+There is no namespace authorization mechanism — any signer may start a new chain in any namespace. Lane isolation comes from the signer's public key being part of the lane key and the v1 sign bytes: different signers produce different sign bytes even at the same sequence in the same namespace. The chaining rule (same signer in parent and child) ensures each signer's chain is independent. Authored namespaces could further restrict who may write to a namespace, but this is orthogonal to the ordering protocol.
```
rootulp, line 44: @claude can you create a diagram to explain to me the difference between lanes, sequences, namespaces, QCs, parent payment promise, etc.

Basically put all these concepts in a diagram to explain their relationships.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: > But do we need to convince them of that

yeah I think so since they could give different numbers of vote to others, therefore some finalize and some don't

if the others are gossiping votes then maybe not, but iirc this is basically the reasoning for the second round

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: a commit (the signatures)

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: in the Tb/s experiment we had half second times, but that was not optimized for latency at all.

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: its true that we have SSF, however its impossible currently to submit a tx and get finality in 6 seconds due to waiting to vote until the end of the 6s

that's also the minimum time, if we include things like light clients, they wait to finalize until the next block finalizes where they get the commit, or onchain light clients take even longer

fibre ordering here gets a lock in < .5s at Tb/s throughputs however I'd bet that we can get way way lower than .5s even at insane throughputs if we optimized for latency

docs/architecture/adr-027-single-sequencer-ordering-fibre.md
evan-forbes: > But if they were not part of it before they wouldn't be able to restore it

the rules currently state that the watermark can increase if there are >2/3 signatures, do you think that they would still need to restore it even if that occurs within the unbonding period?

> sequencer waits for the validator set to change by 1/3+

this is a good point that I think fibre already solves in that the validator set must be known to verify the signatures for the PFF afaiu

rootulp:
Blocked on @cmwaters comments

ninabarbakadze (dismissed):
let's merge this? I don't think it's worth waiting for callum's feedback at this point considering that he's done with celestia work.

ninabarbakadze (commented):
markdown linter complaining

rootulp (approved):
Approving b/c this PR is over a month old and we're not actively working on it or addressing comments. There are no plans to actually implement this ADR but reviewers can open follow-up PRs to clarify the doc or answer any of the open Github comment threads.

PR #6869, opened 2026-03-17
docs: ADR028 Secondary Gas Token

evan-forbes:
## Overview

this ADR is mean to finish scoping https://linear.app/celestia/issue/PROTOCO-1128/explore-supporting-paying-with-stables

it focuses on the absolutely minimal scope required to add a secondary fee token

---

<a href="https://app.devin.ai/review/celestiaorg/celestia-app/pull/6869" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


docs/architecture/adr-028-secondary-gas-token.md
```diff
@@ -0,0 +1,233 @@
...
+Each denom has its own minimum gas price and is validated independently. Mempool priority is local policy: nodes normalize actual gas price against their effective local minimum for that denom. Users choose the fee denom; validators receive fees in that denom.
+
+Out of scope: new modules, new ante decorators, a generalized fee-token whitelist, governance-controlled denom selection, oracles, token conversion, and mixed-denom fees.
+
+### Core Invariants
+
+1. A transaction fee must contain exactly one coin.
+2. The fee denom must be `utia` or the hardcoded secondary denom.
+3. A configured secondary denom must have a positive minimum gas price in `x/minfee` params.
+4. Each denom is validated against its own minimum gas price.
```
ninabarbakadze: Interesting point. So all denoms will have their own min gas price? Would it make sense to just use our gov modifiable min gas price and apply it across secondary denoms.

docs/architecture/adr-028-secondary-gas-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Context
+
+Celestia accepts fees only in `utia`, enforced in `app/ante/fee.go`. The rest of the fee path (`DeductFeeDecorator`, `x/bank`, fee collector, `x/distribution`) already operates on `sdk.Coins`.
+
+This ADR proposes the smallest change that admits one hardcoded secondary fee denom, shipped in a chain upgrade the same way `BondDenom` is defined today.
+
+## Decision
+
+Accept a fee paid entirely in either `utia` or one hardcoded secondary denom. The denom is a compile-time constant in `appconsts`, selected by social consensus and shipped in an upgrade. Governance controls minimum gas prices in `x/minfee`; this ADR adds a `secondary_min_gas_price` field.
```
ninabarbakadze: `secondary_min_gas_price` feels like it's not scalable. I know it's unlikely but what if we add 2 more secondary gas tokens? Do we add more secondary min gas prices?

docs/architecture/adr-028-secondary-fee-token.md
```diff
@@ -0,0 +1,233 @@
...
+## Implementation Surface
+
+### Consensus-Critical
+
+- `pkg/appconsts/global_consts.go` — add `SecondaryFeeDenom`
+- `proto/celestia/minfee/v1/params.proto` — add `secondary_min_gas_price`
+- `x/minfee/types/*.pb.go` (generated)
+- `x/minfee/types/params.go`, `genesis.go`
+- `app/ante/fee.go`
+- `app/ante/fee_test.go`, `get_tx_priority_test.go`
```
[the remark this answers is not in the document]
rach-id: @evan-forbes looks legit

docs/architecture/adr-028-secondary-gas-token.md
rach-id: wouldn't this reduce utility of TIA as a token?

docs/architecture/adr-028-secondary-gas-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+    case appconsts.SecondaryFeeDenom:
+        if appconsts.SecondaryFeeDenom == "" || !params.SecondaryMinGasPrice.IsPositive() {
+            return sdk.Coin{}, math.LegacyDec{}, ErrSecondaryFeeDisabled
+        }
+        return coin, params.SecondaryMinGasPrice, nil
+
+    default:
+        return sdk.Coin{}, math.LegacyDec{}, ErrUnacceptedFeeDenom
+    }
```
ninabarbakadze: I feel like this is not scalable. This switch statement hardcodes each denom as a separate case branch. If we ever support more than one secondary denom every addition scales lines of code. Perhaps some kind of registry would be more extensible. Not sure if I'm jumping the gun here though.

rach-id (dismissed):
LGTM, clear and simple enough

docs/architecture/adr-028-secondary-fee-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Additive (Non-Consensus)
+
+- `app/app.go` — gas-price query helper
+- `app/grpc/gasestimation/gas_estimator.go` — gas estimator
+- `pkg/user/tx_options.go`, `pkg/user/tx_client.go` — tx building / fee defaults
+- `app/errors/insufficient_gas_price.go` — error parsing
+
+## Consequences
```
ninabarbakadze, line 228: how do you get rid of the secondary token if the ecosystem for it disappears?

docs/architecture/adr-028-secondary-gas-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+1. Compute `actualGasPrice = fee.Amount / gas`.
+2. In `CheckTx`, enforce the local validator threshold for that denom.
+3. In `CheckTx` and `DeliverTx`, enforce the network minimum for that denom.
+4. In `CheckTx`, compute priority (see [Priority](#priority)).
+5. Return the original `sdk.Coins` so the SDK deducts the chosen denom.
+
+### Priority
+
+Priority uses **normalized gas price** (`actual / effective local minimum`) so operators can favor one denom over the other without changing admission rules. The effective local minimum is the node's `min-gas-prices` for that denom if above the network minimum, otherwise the network minimum from `x/minfee`.
```
ninabarbakadze: what does it mean operators can favor one denom over the other? does it mean you can submit txs with fees in both denoms?

docs/architecture/adr-028-secondary-gas-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Fee Collection and Distribution
+
+No changes. The SDK deducts the submitted coin into the fee collector, and `x/distribution` allocates all fee collector balances via `sdk.DecCoins`, so rewards naturally accrue in both denoms.
+
+## UX and Rollout
+
+### Wallet / Tx Builder
+
+Clients query `x/minfee`, let users choose a denom, and build a single fee coin. Gas estimation is unchanged.
```
ninabarbakadze: At first glance I'm not sure if gas estimation can be unchanged. In `SortAndExtractGasPrices()` we get gas price for our bonddenom which is utia 

	`gasPrice := float64(feeTx.GetFee().AmountOf(appconsts.BondDenom).Uint64()) / float64(feeTx.GetGas())`

ninabarbakadze (commented):
well-written ADR! thanks!

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: perhaps but I think the typical burning of excess fees (long discussed topic) would fix this. ofc, burning other tokens doesn't do anything, therefore those would have to be put in a gov address and swapped somehow for tia, then burned.

then rewards to validators would be constant inflation.

originally I assumed this was outside of the scope of this document, but perhaps not!

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: presumably they would each need their own or a conversion rate (same diff ofc). therefore having their own gov and config param seems to make sense since each token would be treated effectively the same

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: inherently there must be some way to express relative preference to remain secure or accurate, therefore fundamentally we need one piece of information convey this per gas token.

adding the ability of the operator to increase the min price is technically optional however likely a high risk thing to do since there are no guarantees that prices of each token are remotely stable (ie a hack)

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: this is the same comment as before afaiu so my response is therefore the same

https://github.com/celestiaorg/celestia-app/pull/6869#discussion_r2960637777

docs/architecture/adr-028-secondary-fee-token.md
evan-forbes: same way we add it I suppose

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: same response actually I thinkhttps://github.com/celestiaorg/celestia-app/pull/6869#discussion_r2960637777

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: since we have a config that expresses preference, we have a conversion rate chosen by the user with a governance mandated floor, therefore we can always convert using gas

docs/architecture/adr-028-secondary-gas-token.md
rach-id: makes sense 👍

docs/architecture/adr-028-secondary-fee-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+No action required. Operators who want to favor one denom can set per-denom values in `min-gas-prices`; raising the local minimum tightens admission and lowers that denom's relative priority.
+
+### Governance
+
+Governance does not select the denom (that stays in the upgrade constant). Governance sets `secondary_min_gas_price` via `MsgUpdateMinfeeParams` to define the network-wide admission floor.
+
+### Rollout Sequence
+
+1. Social consensus on the secondary fee denom.
```
rootulp, line 205: We could have a FLUP CIP to recommend the secondary fee denom. The CIP could propose the secondary fee denom based these criteria: 
- The most popular USD stablecoin on Celestia
- The USD stablecoin with the greatest liquidity on a chain directly connected to Celestia via IBC or Hyperlane
- The USD stablecoin with the greatest adoption (e.g. USDT, USDC)

Regardless it shouldn't block this ADR but you could create a follow-up issue if/when this merges.

docs/architecture/adr-028-secondary-fee-token.md
```diff
@@ -0,0 +1,233 @@
...
+- `app/app.go` — gas-price query helper
+- `app/grpc/gasestimation/gas_estimator.go` — gas estimator
+- `pkg/user/tx_options.go`, `pkg/user/tx_client.go` — tx building / fee defaults
+- `app/errors/insufficient_gas_price.go` — error parsing
+
+## Consequences
+
+**Positive**: Minimal state changes. Backward compatible when disabled. Operators can express local preference between accepted fee denoms through existing `min-gas-prices` configuration.
+
+**Negative**: Mempool ordering may differ across nodes because priority normalization depends on local operator policy. Rewards become multi-denom. Full UX requires follow-up outside the ante handler.
```
rootulp, line 232: Is it possible to mitigate this negative consequence by removing 

> local operator policy

I assume this refers to a node operator's `min-gas-prices` configuration. 

Put differently, if we made that a network wide constant would this negative consequence be eliminated?

docs/architecture/adr-028-secondary-gas-token.md
rootulp: A user can't submit a tx with fees in both denoms because 

```
    if len(fee) != 1 {
        return sdk.Coin{}, math.LegacyDec{}, ErrInvalidFeeShape
    }
```

so I don't understand 

> so operators can favor one denom over the other without changing admission rules

docs/architecture/adr-028-secondary-gas-token.md
rootulp: I don't think we should be thinking about scaling this beyond one coin. IMO safe to add fields for 2 coins. We may consider refactoring if we expanded to 3 coins which isn't currently being requested or planned for so let's not block on scaling to 3+ coins.

docs/architecture/adr-028-secondary-gas-token.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Status
+
+Draft
+
+## Context
+
+Celestia accepts fees only in `utia`, enforced in `app/ante/fee.go`. The rest of the fee path (`DeductFeeDecorator`, `x/bank`, fee collector, `x/distribution`) already operates on `sdk.Coins`.
+
+This ADR proposes the smallest change that admits one hardcoded secondary fee denom, shipped in a chain upgrade the same way `BondDenom` is defined today.
```
rootulp: @claude the way this is phrased makes it seem like `BonDenom` is related to shipping the change in a chain upgrade but those seem like distinct concepts. Can you rephrase?

rootulp (commented):
1. Can users pay for blob fees via USD to an off-chain service? https://linear.app/celestia/issue/PROTOCO-1082/stable-blob-fees#comment-41550c18
2. Are there any alternative designs we should consider? For example, is it worth exploring an option that adds a USD/TIA price oracle to Celestia so that there doesn't need to be a second min gas price config? Likely out of scope and more complicated.

docs/architecture/adr-028-secondary-fee-token.md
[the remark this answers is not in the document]
evan-forbes: I'd just defer this to the implementation per the todo as this is important but adds noise to this document

docs/architecture/adr-028-secondary-fee-token.md
evan-forbes: inherently we must have a way to adapt quickly if there is a hack on the new token and it dives in prices relative to the other token

otherwise the feemarket critical for prioritizing important usage (oracle updates in eth or solana land for example) can't be prioritized.

this was the thinking in including this. Therefore, if that is a requirement then there is no way to avoid the ordering of the consistent. However I don't see this as a major issues since the mempool is out of protocol and cannot be enforced. It must be assumed that a peer's mempool's state if different from one's own. This has caused a lot of other issues before if we don't plan for this occuring

docs/architecture/adr-028-secondary-fee-token.md
```diff
@@ -0,0 +1,233 @@
+# ADR: Secondary Fee Token Support
```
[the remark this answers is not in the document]
evan-forbes: addressed to review feedback commit

docs/architecture/adr-028-secondary-fee-token.md
evan-forbes: added an issue!

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: fixed via removing the above constrain @rootulp thanks!

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: added a comment to document this but basically this argument could be summarized as "revenue is revenue"

we could also end up reducing inflation to cover costs with sufficient fees and adaptable hardware consumption

docs/architecture/adr-028-secondary-gas-token.md
evan-forbes: I don't think claude fixed this here so I fixed this in the reedback review comment

evan-forbes:
I can go back and fix the adr numbers if necessary after we merge this

docs/architecture/adr-028-secondary-fee-token.md
rootulp: > inherently we must have a way to adapt quickly if there is a hack on the new token and it dives in prices relative to the other token

I don't think there's a big time difference between these two approaches:

Approach A
1. protocol team creates a new release of celestia-app with a modified min gas price for one token. 
2. Node operators upgrade to new binary

Approach B
1. Node operators modify their config to set min gas price to a new setting

Both approaches depend on node operators performing a manual step. Approach A has the addition of an action from our team which we control so we can do it quickly if need be.

evan-forbes: sure that makes sense and I agree

however stopping the people's mempool from differing I don't think adds any value

this really isn't a problem since syncing folk's mempools requires consensus, thus we already have mempools with very different states. All mempools must be capable of handling such differences, and our mempool currently is capable of such diffs

rootulp: > however stopping the people's mempool from differing I don't think adds any value

Then why is this listed as a negative consequence?

> **Negative**: Mempool ordering may differ across nodes because priority normalization depends on local operator policy. Rewards become multi-denom. Full UX requires follow-up outside the ante handler.

rootulp: FWIW this isn't blocking feedback and I don't really expect this to get implemented soon so I don't think we'll get a lot of value from continuing this thread. I'm content with marking this thread as resolved.

evan-forbes:
> Can users pay for blob fees via USD to an off-chain service? https://linear.app/celestia/issue/PROTOCO-1082/stable-blob-fees#comment-41550c18
Are there any alternative designs we should consider? For example, is it worth exploring an option that adds a USD/TIA price oracle to Celestia so that there doesn't need to be a second min gas price config? Likely out of scope and more complicated.

Yeah definitely the off chain payment service could be good originally I thought that was out of scope since this assumed in protocol 

Agree that adding an oracle is likely off the table since it can be avoided and introduces a lot of fragility. Outdated or incorrect values could in theory break the chain as the cost to use is either way too high or way too cheap. It would also allow the validators to have a monopoly over the price rather than the existing mech which would currently allows a single validator to include low fee txs

I can add these the adr to record them but had ruled them out subconsciously so didn't originally

rootulp:
> since this assumed in protocol

I don't view in protocol as a strict requirement.

I think it would be helpful to start from the [customer feedback](https://linear.app/celestia/issue/PROTOCO-1082/stable-blob-fees#comment-d5d1eb05), distill product requirements, and then research potential solutions.

rootulp (approved):

docs/architecture/adr-028-secondary-fee-token.md
evan-forbes: I'd actually prefer to come to a conclusion if you have time :slightly_smiling_face: 

are there ways that people's mempool ordering would cause problems? my current assumptions are that this is okay, since inherently consensus is required to reach agreement on ordering, therefore the mempool MUST be designed to accommodate this.

